PR #7697, opened 2023-04-12
YJIT: Fix missing argc check in known cfuncs

jhawthorn:
Previously we were missing a compile-time check that the known cfuncs receive the correct number of arguments.

```
$ ruby --yjit-call-threshold=1 -e '"foo".to_s(*[])'
thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `2`', ./yjit/src/codegen.rs:7225:5
-e:1: [BUG] YJIT panicked
ruby 3.3.0dev (2023-04-08T18:54:01Z master 671cfc2000) +YJIT [x86_64-linux]
```

We noticed this because in particular when using ARGS_SPLAT, which also wasn't checked, YJIT would crash on code which was otherwise correct (didn't raise exceptions in the VM).

This still supports vararg (argc == -1) cfuncs. I added an additional assertion that when we use the specialized codegen for one of these known functions that the argc are popped off the stack correctly, which should help ensure they're implemented correctly (previously the crash was usually observed on a future `leave` insn).

We should backport this to Ruby 3.2 as well as it seems affected by the non-splat case (it's a little harder to make happen since the code would normally raise an exception on the VM). ex. `ruby --yjit-call-threshold=1 -e '"foo".to_s(1)'`

Ruby 3.1 doesn't seem to have this issue because at that time we [checked argc early enough](https://github.com/ruby/ruby/blob/ruby_3_1/yjit_codegen.c#L3232-L3236).

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1227,6 +1227,38 @@ assert_equal '42', %q{
   run
 }
 
+# splatting an empty array on a specialized method
+assert_equal 'ok', %q{
+  def run
+    "ok".to_s(*[])
+  end
+
+  run
+  run
+}
+
+# splatting an single element array on a specialized method
+assert_equal '[1]', %q{
+  def run
+    [].<<(*[1])
+  end
+
+  run
+  run
+}
+
+# specialized method with wrong args
+assert_equal 'ok', %q{
+  def run(x)
+    "bad".to_s(123) if x
+  rescue
+    :ok
+  end
+
+  run(false)
+  run(true)
+}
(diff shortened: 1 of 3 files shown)


maximecb (approved):
Thank you for catching this and adding extra tests 🙏

k0kubun (approved):
> We should backport this to Ruby 3.2 as well as it seems affected by the non-splat case

Could you file a ticket for backport management and mention it from the commit message?

maximecb:
There's a Rust build error on arm64

PR #7698, opened 2023-04-12
YJIT: Add support for rest with option and splat args

jimmyhmiller:
Now that we can exit anywhere with CantCompile, I'd like to reorder around this code. The changes are scattered throughout the file and it is getting incredibly hard to work with gen_send_iseq. 

Personally, I'd love to refactor the whole function. But I'd imagine the way I'd be interested in seeing the function work is different from ideas others have. I also know we are making lots of changes to codegen and don't want to conflict with those changes.

I will say, I think I lost a good amount of time to dealing with the state of this code and I think had this method been refactored in the right way, this change wouldn't have taken so long.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -14,6 +14,7 @@ use YARVOpnd::*;
 
 use std::cell::Cell;
 use std::cmp;
+use std::cmp::min;
 use std::collections::HashMap;
 use std::ffi::CStr;
 use std::mem;
@@ -4802,6 +4803,7 @@ fn gen_push_frame(
     asm: &mut Assembler,
     set_sp_cfp: bool, // if true CFP and SP will be switched to the callee
     frame: ControlFrame,
+    rest_arg: Option<(i32, Opnd)>,
 ) {
     assert!(frame.local_size >= 0);
 
@@ -4900,6 +4902,13 @@ fn gen_push_frame(
             let offs = SIZEOF_VALUE_I32 * (i - num_locals - 3);
             asm.store(Opnd::mem(64, sp, offs), Qnil.into());
         }
+    }
+
+    if let Some((opts_missing, rest_arg)) = rest_arg {
+        // We want to set the rest_param just after the optional arguments
+        let index = opts_missing - num_locals - 3;
+        let offset = SIZEOF_VALUE_I32 * index;
+        asm.store(Opnd::mem(64, sp, offset), rest_arg);
     }
 
     if set_sp_cfp {
@@ -5131,7 +5140,7 @@ fn gen_send_cfunc(
         },
         iseq: None,
         local_size: 0,
-    });
+    }, None);
 
     if !kw_arg.is_null() {
         // Build a hash from all kwargs passed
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -5712,23 +5733,68 @@ fn gen_send_iseq(
...
+        opts_filled = filled;
+        num_params -= opts_missing as u32;
+
+        // We are going to jump to the correct offset based on how many optional
+        // params are remaining.
+        unsafe {
+            let opt_table = get_iseq_body_param_opt_table(iseq);
+            start_pc_offset = (*opt_table.offset(filled as isize)).try_into().unwrap();
+        };
+    }
```
maximecb, line 5796: I'm a bit confused by this piece of code.  What's on the right hand side of the if let? What does it evaluate to?

maximecb:
I'm open to refactoring the send logic. I think it's dangerously complex and hard to follow 😅 

Would you mind opening an issue on Shopify/ruby and outlining your ideas in there so we can discuss them?

XrXr:
Since we're talking about refactoring argument transfer, I'm going to quote some of my old (probably somewhat outdated) thinking here:
https://github.com/Shopify/yjit/issues/301

My main thesis there is that the transfer should be done piece-meal following the delimiters already present in parameter information on each iseq. It's more natural, and has a closer data flow to `setup_parameters_complex()`, so easier to see correctness by contrasting the two. Graphics from `rb_iseq_constant_body`:
```
    /**
     * parameter information
     *
     *  def m(a1, a2, ..., aM,                    # mandatory
     *        b1=(...), b2=(...), ..., bN=(...),  # optional
     *        *c,                                 # rest
     *        d1, d2, ..., dO,                    # post
     *        e1:(...), e2:(...), ..., eK:(...),  # keyword
     *        **f,                                # keyword_rest
     *        &g)                                 # block
     * =>
     *
     *  lead_num     = M
     *  opt_num      = N
     *  rest_start   = M+N
     *  post_start   = M+N+(*1)
     *  post_num     = O
     *  keyword_num  = K
     *  block_start  = M+N+(*1)+O+K
     *  keyword_bits = M+N+(*1)+O+K+(&1)
     *  size         = M+N+O+(*1)+K+(&1)+(**1) // parameter size.
     */
```
I think "working on slice X of the argument array" is useful language for explaining what the generator/generated code is doing, in any case.

For keyword args and splats, the logic split into at least two parts, guarding and transfer. I think we probably want to make some Rust structs to make transfer's dependency on the guard (or compile-time pre-check) more explicit. Structs that act as tokens of proof that some checks have happened and/or some guards have been issued, basically.

Anyways, carry on.

jimmyhmiller:
@XrXr I don't think I fully understand the idea, but what I do understand I like. I particularly like the idea of thinking explicitly in guarding and transfers. The code right now is very implicit about the fact that what we are doing is transferring values to the callee.

k0kubun:
I'm also in favor of refactoring send implementation.

When we do that, it'd be nice to somehow propagate a `CantCompile`-like thing with just `?`, which would be useful to split the huge send logic into multiple functions. One idea is to use `Option<CodegenStatus>` and consider `None` a new `CantCompile`, removing `CantCompile` from `CodegenStatus`.

jimmyhmiller:
> When we do that, it'd be nice to somehow propagate a CantCompile-like thing with just ?, which would be useful to split the huge send logic into multiple functions. One idea is to use Option<CodegenStatus> and consider None a new CantCompile, removing CantCompile from CodegenStatus.

Very much agree with this. Would love to take advantage of `?` a lot more.

maximecb (approved):

PR #7702, opened 2023-04-13
YJIT: Fix edge and total counts in exit_locations

jhawthorn:
The stackprof-format raw samples are suffixed with a count (ie. how many times did the previously recorded side-exit repeat). Previously we were correctly using this value to increment the `:samples` count, but not the `:total_samples` count or edges.

This made the stackprof aggregate results incorrect (though any flamegraphs generated should have been correct, since those only rely on raw samples).

test script:
```
def bar(**args)
end

1000.times do
  bar() # side exit
end
```

**Before**

```
$ ./miniruby --yjit-trace-exits test.rb && stackprof yjit_exit_locations.dump --method=opt_send_without_block
opt_send_without_block (nonexistent.def:1)
  samples:   971 self (98.3%)  /      1 total (0.1%)
  callers:
       1  (  100.0%)  block in <main>
  callees (-970 total):
  code:
        SOURCE UNAVAILABLE
```

**After**

```
$ ./miniruby --yjit-trace-exits test.rb && stackprof yjit_exit_locations.dump --method=opt_send_without_block
opt_send_without_block (nonexistent.def:1)
  samples:   971 self (98.3%)  /    971 total (98.3%)
  callers:
     971  (  100.0%)  block in <main>
  callees (0 total):
  code:
        SOURCE UNAVAILABLE
```

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -70,6 +70,8 @@ module RubyVM::YJIT
       stack_length = raw_samples[i] + 1
       i += 1 # consume the stack length
 
+      sample_count = raw_samples[i + stack_length]
+
       prev_frame_id = nil
       stack_length.times do |idx|
         idx += i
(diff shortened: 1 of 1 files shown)


maximecb:
Thanks for fixing these issues, and if there's anything we can do to improve YJIT perf on GitHub, if you run into any performance or memory usage issues, please let us know :)

PR #7705, opened 2023-04-13
Avoid seeding OpenSSL for SecureRandom

shyouhei:
OpenSSL's man page previously stated that "the application is responsible for seeding the PRNG by calling RAND_add" (see [1]). So we had this code.  However things changed.  They no longer say so, instead "manual (re-)seeding of the default OpenSSL random generator is not necessary" now (see [2]).  It seems all OpenSSL versions that we support now already behaves like this. Let's follow that.

[1]: https://www.openssl.org/docs/man1.0.2/man3/RAND_add.html
[2]: https://www.openssl.org/docs/manmaster/man3/RAND_add.html

### DIFF
--- a/lib/securerandom.rb
+++ b/lib/securerandom.rb
@@ -47,17 +47,6 @@ module SecureRandom
     private
 
     def gen_random_openssl(n)
-      @pid = 0 unless defined?(@pid)
-      pid = $$
-      unless @pid == pid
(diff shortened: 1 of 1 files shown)


shyouhei:
Cirrus CI failures seem unrelated.

mame (approved):

PR #7707, opened 2023-04-13
[Feature #19538] Add "performance" warning category

nobu:
https://bugs.ruby-lang.org/issues/19538

### DIFF
--- a/include/ruby/internal/error.h
+++ b/include/ruby/internal/error.h
@@ -50,6 +50,9 @@ typedef enum {
     /** Warning is for experimental features. */
     RB_WARN_CATEGORY_EXPERIMENTAL,
 
+    /** Warning is for performance. */
+    RB_WARN_CATEGORY_PERFORMANCE,
+
     RB_WARN_CATEGORY_ALL_BITS = (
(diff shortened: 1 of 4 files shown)


casperisfine:
Oh, I was working on it 😄

include/ruby/internal/error.h
```diff
@@ -50,6 +50,9 @@ typedef enum {
     /** Warning is for experimental features. */
     RB_WARN_CATEGORY_EXPERIMENTAL,
 
+    /** Warning is for performance. */
+    RB_WARN_CATEGORY_PERFORMANCE,
```
paracycle, line 54: Shouldn't this also be added to the `RB_WARN_CATEGORY_ALL_BITS` below?

casperisfine: Nope, it was specifically asked that this category is not enabled on bare`-W`. `RB_WARN_CATEGORY_ALL_BITS` is what controls that.

hsbt:
We merged https://github.com/ruby/ruby/pull/7708 instead of this.

PR #7708, opened 2023-04-13
Emit a performance warning when a class reached max variations

casperisfine:
[[Feature #19538]](https://bugs.ruby-lang.org/issues/19538)

This new `peformance` warning category is disabled by default. It needs to be specifically enabled via `-W:performance` or `Warning[:performance] = true`

cc @nobu @tenderlove

### DIFF
--- a/include/ruby/internal/error.h
+++ b/include/ruby/internal/error.h
@@ -50,6 +50,9 @@ typedef enum {
     /** Warning is for experimental features. */
     RB_WARN_CATEGORY_EXPERIMENTAL,
 
+    /** Warning is for performance issues (not enabled by -w). */
+    RB_WARN_CATEGORY_PERFORMANCE,
+
     RB_WARN_CATEGORY_ALL_BITS = (
         (1U << RB_WARN_CATEGORY_DEPRECATED) |
         (1U << RB_WARN_CATEGORY_EXPERIMENTAL) |
--- a/test/ruby/test_object.rb
+++ b/test/ruby/test_object.rb
@@ -422,6 +422,18 @@ class TestObject < Test::Unit::TestCase
     assert_equal(1+3+5+7+9, n)
   end
 
+  def test_max_shape_variation_with_performance_warnings
+    assert_in_out_err([], <<-INPUT, %w(), /Maximum shapes variations \(8\) reached by Foo, instance variables accesses will be slower\.$/)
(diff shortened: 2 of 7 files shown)


include/ruby/internal/error.h
```diff
@@ -50,7 +50,10 @@ typedef enum {
     /** Warning is for experimental features. */
     RB_WARN_CATEGORY_EXPERIMENTAL,
 
-    RB_WARN_CATEGORY_ALL_BITS = (
```
nobu, line 53: Since this file is a public header, we may have to keep the name including the new bit.

casperisfine: Ah, I didn't notice that. Thanks, I'll restore the old name.

nobu: Sorry, I meant to keep the old name and add the new name for the bits enabled by `-w`.

casperisfine: Oh. I'll make another PR.

eregon:
Could you add an entry in NEWS for this?
And also document `-W:performance` e.g. in the man page and docs of `Warning.[]`?

byroot:
> Could you add an entry in NEWS for this?

So I wanted to do this, but none of the existing section made sense. Any idea?

eregon:
> So I wanted to do this, but none of the existing section made sense. Any idea?

https://github.com/ruby/ruby/pull/7715 seems fine, could also be under command-line changes or so (I guess older NEWS files must have a way to document those) & Warning.

test/ruby/test_object.rb
```diff
@@ -422,6 +422,18 @@ def test_check_to_integer
     assert_equal(1+3+5+7+9, n)
   end
 
+  def test_max_shape_variation_with_performance_warnings
+    assert_in_out_err([], <<-INPUT, %w(), /Maximum shapes variations \(8\) reached by Foo, instance variables accesses will be slower\.$/)
+      $VERBOSE = true
```
eregon, line 427: Does it also work with `$VERBOSE = false`? See https://github.com/ruby/ruby/pull/7715#discussion_r1169777099
If only with `$VERBOSE = true`, it seems too complicated to enable performance warnings (need to enable two things, very easy to mess up)

casperisfine: I believe it needs both yes.

PR #7709, opened 2023-04-13
[Bug #19597] Freeze script name

nobu:
https://bugs.ruby-lang.org/issues/19597

### DIFF
--- a/test/ruby/test_process.rb
+++ b/test/ruby/test_process.rb
@@ -1424,6 +1424,11 @@ class TestProcess < Test::Unit::TestCase
     REPRO
   end
 
+  def test_argv0_frozen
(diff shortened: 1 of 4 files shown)


test/ruby/test_process.rb
```diff
@@ -1424,6 +1424,11 @@ def test_argv0_keep_alive
     REPRO
   end
 
+  def test_argv0_frozen
+    assert_predicate Process.argv0, :frozen?
+    assert_predicate $0, :frozen?
+  end
```
byroot, line 1430: Probably doesn't matter much, but that's technically a behavior change, right?

nobu: Yes, bug fixes change behaviors.

PR #7710, opened 2023-04-13
Add RB_WARN_CATEGORY_DEFAULT_BITS

casperisfine:
Followup: ac123f167a364c3d7a43eca78d564e41f6dbb91e

RB_WARN_CATEGORY_ALL_BITS is exposed in a public header, so it makes sense for it to be updated to contain all valid bits.

Instead we introduce RB_WARN_CATEGORY_DEFAULT_BITS to list the categories that are enabled by default.

### DIFF
--- a/include/ruby/internal/error.h
+++ b/include/ruby/internal/error.h
@@ -53,9 +53,15 @@ typedef enum {
     /** Warning is for performance issues (not enabled by -w). */
     RB_WARN_CATEGORY_PERFORMANCE,
 
+    RB_WARN_CATEGORY_DEFAULT_BITS = (
+        (1U << RB_WARN_CATEGORY_DEPRECATED) |
+        (1U << RB_WARN_CATEGORY_EXPERIMENTAL) |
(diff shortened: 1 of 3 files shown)


include/ruby/internal/error.h
```diff
@@ -53,9 +53,15 @@ typedef enum {
     /** Warning is for performance issues (not enabled by -w). */
     RB_WARN_CATEGORY_PERFORMANCE,
 
+    RB_WARN_CATEGORY_DEFAULT_BITS = (
+        (1U << RB_WARN_CATEGORY_DEPRECATED) |
+        (1U << RB_WARN_CATEGORY_EXPERIMENTAL) |
+        0)
```
nobu, line 59:
```suggestion
        0),
```

nobu (approved):

include/ruby/internal/error.h
casperisfine: Oops 🤦

PR #7711, opened 2023-04-13
[Bug #19417] Make word prop match join_control ...

jaynetics:
... to conform to UTS 18 as mentioned in https://bugs.ruby-lang.org/issues/19417#note-3

https://unicode.org/reports/tr18/#word states word should match join_control chars.

It currently does not:

```ruby
[*0x0..0xD799, *0xE000..0x10FFFF].map { |n| n.chr 'utf-8' } => all_chars
all_chars.grep(/\p{join_control}/) => jc
jc.count # => 2
jc.grep(/\p{word}/).count # => 0
```


tool/enc-unicode.rb
```diff
@@ -5,9 +5,10 @@
 #
 # To use this, get UnicodeData.txt, Scripts.txt, PropList.txt,
 # PropertyAliases.txt, PropertyValueAliases.txt, DerivedCoreProperties.txt,
-# DerivedAge.txt and Blocks.txt  from unicode.org.
+# DerivedAge.txt, Blocks.txt, emoji/emoji-data.txt,
+# auxiliary/GraphemeBreakProperty.txt from unicode.org
 # (http://unicode.org/Public/UNIDATA/) And run following command.
-# ruby1.9 tool/enc-unicode.rb data_dir > enc/unicode/name2ctype.kwd
+# tool/enc-unicode.rb data_dir emoji_data_dir > enc/unicode/name2ctype.kwd
```
jaynetics, line 11: these comments were outdated; some more files and args are needed to run this now

jaynetics:
see also #7287 which updates the documentation to reflect the standard-conforming behavior

enc/unicode/15.0.0/name2ctype.h
```diff
@@ -41768,16 +41769,17 @@ uniname2ctype_hash (register const char *str, register size_t len)
       6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099,
       6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099,
       6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099, 6099,
-#ifndef USE_UNICODE_AGE_PROPERTIES
```
jaynetics, line 41771: the following diff was in the gperf output for me even before i updated enc-unicode.rb

nobu: This seems cased by used `diff`.
What `diff` did you use?

nobu: Confirmed that the following hunks don't exist when using GNU diffutils 3.8.

nobu:
Another test.
```patch
diff --git a/test/ruby/test_regexp.rb b/test/ruby/test_regexp.rb
index 6ec77a41a3c..ff555b3419d 100644
--- a/test/ruby/test_regexp.rb
+++ b/test/ruby/test_regexp.rb
@@ -1230,6 +1230,9 @@
     assert_match(/\A[[:space:]]+\z/, "\r\n\v\f\r\s\u0085")
     assert_match(/\A[[:ascii:]]+\z/, "\x00\x7F")
     assert_no_match(/[[:ascii:]]/, "\x80\xFF")
+
+    assert_match(/[[:word:]]/, "\u{200C}")
+    assert_match(/[[:word:]]/, "\u{200D}")
   end
 
   def test_cclass_R
```

enc/unicode/15.0.0/name2ctype.h
jaynetics: thanks for checking!

i hadn't noticed that `enc-unicode.rb` uses `diff` internally and was using FreeBSD diff. GNU diff 3.**9** also produces slightly different output btw.

jaynetics:
> Another test.

thanks, added!

jaynetics:
@nobu from my point of view this is ready to merge when you want

enc/unicode/15.0.0/name2ctype.h
nobu: GNU diff 3.9 has a bug that `-D` option does not work at all.

jeremyevans:
@nurse @duerst Could you review this? Now that #7287 was merged, we definitely want the implementation to match the documentation.

nurse:
It looks `\p{Word}` is updated in TR#18 Version 15.
https://www.unicode.org/reports/tr18/tr18-15.html
The fix looks good. If the conflict is resolved, I'll merge this.
thanks,

PR #7691, opened 2023-04-12
YJIT: Let Assembler own Context

k0kubun:
## Background
We've sometimes wanted to use `Context` when we modify `Assembler`, e.g. `asm.spill_temps(ctx)` and `ctx.stack_push(asm, ..)`. We want to do this more. For example, we could write an immediate instead of a stack operand when Context knows the exact value. 

We could keep passing `ctx` to `asm` and copying it through IR operands, however, having `Context` in `Assembler` seems simpler to approach it.

## Changes
This PR adds `ctx` field to `Assembler` and use `asm.ctx` instead of passing `ctx` separately.

### Out of scope
We probably want to move more `Context` methods like `ctx.stack_opnd` to `Assembler` for consistency and to avoid lengthy `asm.ctx`. But this PR is already too big, so I'd like to separate a PR for that change.

PR #7712, opened 2023-04-14
YJIT: Introduce Target::SideExit

k0kubun:
This is a refactoring discussed at https://github.com/ruby/ruby/pull/7691#issuecomment-1505792005. 

We could use `Target::SideExit(Counter)` instead of `Target::SideExit(Option<Counter>)`, but we currently have too many existing `counter: Option<Counter>` arguments. I'd like to scope it out from this PR.

### DIFF
--- a/yjit/src/backend/ir.rs
+++ b/yjit/src/backend/ir.rs
@@ -3,13 +3,15 @@
 #![allow(unused_imports)]
 
 use std::cell::Cell;
+use std::collections::HashMap;
 use std::fmt;
 use std::convert::From;
 use std::io::Write;
 use std::mem::take;
+use crate::codegen::{gen_outlined_exit, gen_counted_exit};
 use crate::cruby::{VALUE, SIZEOF_VALUE_I32};
 use crate::virtualmem::{CodePtr};
-use crate::asm::{CodeBlock, uimm_num_bits, imm_num_bits};
+use crate::asm::{CodeBlock, uimm_num_bits, imm_num_bits, OutlinedCb};
 use crate::core::{Context, Type, TempMapping, RegTemps, MAX_REG_TEMPS, MAX_TEMP_TYPES};
 use crate::options::*;
 use crate::stats::*;
@@ -280,9 +282,10 @@ impl From<VALUE> for Opnd {
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum Target
 {
-    CodePtr(CodePtr),     // Pointer to a piece of YJIT-generated code
-    SideExitPtr(CodePtr), // Pointer to a side exit code
-    Label(usize),         // A label within the generated code
+    CodePtr(CodePtr),          // Pointer to a piece of YJIT-generated code
+    SideExit(Option<Counter>), // Side exit with a counter
+    SideExitPtr(CodePtr),      // Pointer to a side exit code
+    Label(usize),              // A label within the generated code
 }
 
 impl Target
@@ -467,6 +470,9 @@ pub enum Insn {
     /// Shift a value right by a certain amount (signed).
     RShift { opnd: Opnd, shift: Opnd, out: Opnd },
 
+    /// Set a side exit context for subsequent instructions.
+    SideExitContext { pc: *mut VALUE, stack_size: u8 },
+
     /// Spill a stack temp from a register into memory
     SpillTemp(Opnd),
 
@@ -552,6 +558,7 @@ impl Insn {
             Insn::PadInvalPatch => "PadEntryExit",
             Insn::PosMarker(_) => "PosMarker",
             Insn::RShift { .. } => "RShift",
+            Insn::SideExitContext { .. } => "SideExitContext",
             Insn::SpillTemp(_) => "SpillTemp",
             Insn::Store { .. } => "Store",
             Insn::Sub { .. } => "Sub",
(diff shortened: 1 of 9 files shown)


yjit/src/backend/ir.rs
```diff
@@ -899,21 +917,30 @@ pub struct Assembler
 
     /// Context for generating the current insn
     pub ctx: Context,
+
+    /// Parallel vec with insns
+    /// Past Context for this insn
+    pub(super) insn_ctx: Vec<Context>,
```
maximecb, line 923: Hmmmm. Is this storing a context for each YARV insn or each backend IR insn?

yjit/src/backend/ir.rs
```diff
@@ -280,9 +282,10 @@ impl From<VALUE> for Opnd {
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub enum Target
 {
-    CodePtr(CodePtr),     // Pointer to a piece of YJIT-generated code
-    SideExitPtr(CodePtr), // Pointer to a side exit code
-    Label(usize),         // A label within the generated code
+    CodePtr(CodePtr),          // Pointer to a piece of YJIT-generated code
+    SideExit(Option<Counter>), // Side exit with a counter
```
maximecb, line 286: Why wrap the counter in an `Option` here? Why not make it mandatory?

yjit/src/backend/ir.rs
```diff
@@ -467,6 +470,9 @@ pub enum Insn {
     /// Shift a value right by a certain amount (signed).
     RShift { opnd: Opnd, shift: Opnd, out: Opnd },
 
+    /// Set a side exit context for subsequent instructions.
+    SideExitContext { pc: *mut VALUE, stack_size: u8 },
```
maximecb, line 474: I wonder if this really needs to be an instruction? Could potentially call `asm.set_side_exit_context()` and set some mutable fields in the `Assembler` instead? That would avoid allocating and copying these pseudo-instructions?

yjit/src/backend/ir.rs
k0kubun: The latter. At the time of `asm.compile`, the latest `Context` isn't useful for anything. So if we want to do something using `Context` in `Assembler`, e.g. identifying a stack slot is an immediate, this is needed regardless of side exits.

Another way I thought of was to introduce a backend IR that's inserted every time `Context` is detected to be changed on `push_insn`, which would limit the number of `Context`s on memory during compilation time. That would add another backend IR, so I didn't go for it, but the logic to propagate `Context` from an `Assembler` to another would be simpler than the current implementation.

Did you have other ideas as to how we're going to use `Context` from `Assembler`?

yjit/src/backend/ir.rs
k0kubun: Many exiting places are lacking a counter and adding new counters for all of them would make the diff much larger, so I'm deliberately leaving it for another PR. But once we add counters to all places, changing `SideExit(Option<Counter>)` to `SideExit(Counter)` afterwards shouldn't be difficult.

yjit/src/backend/ir.rs
k0kubun: This is similar to https://github.com/ruby/ruby/pull/7712#discussion_r1166913625. If we were to introduce another iterator for `Target` and convert all `Target::SideExit` to `Target::SideExitPr` on `push_insn`, having only a single mutable field would be enough, but it's generally too late at the time of `asm.compile`.

However, assuming we shouldn't use `CodegenGlobals::get_outlined_cb()` at this place, resolving a `Target` on `push_insn` would require `ocb` for every `Assembler` API that takes `Target`s, e.g. `asm.je(target, ocb)`. And I wasn't sure if we want that API.

yjit/src/backend/ir.rs
maximecb: Sounds good.

yjit/src/backend/ir.rs
maximecb: Presumably, instead of inserting `Insn::SideExitContext { pc, stack_size }` pseudo-instructions, you could store those values on the `Assembler`, and when a `Target::SideExit` is inserted, you would write those values into the target?

k0kubun: Sounds like a good idea. I didn't think about adding an iterator for `Target`s, but it might be useful for what's discussed in https://github.com/ruby/ruby/pull/7712#discussion_r1167092787 too. I'll have a look at removing this instruction.

yjit/src/backend/ir.rs
k0kubun: I stopped cloning `Context` for each insn at https://github.com/ruby/ruby/pull/7712/commits/f322d2ab24bcb179bcfcb60dbaca0ca788bf120a.

The performance impact became much smaller.

```
before: ruby 3.3.0dev (2023-04-14T18:21:29Z master d83e59e6b8) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-04-14T19:17:44Z yjit-side-exit f322d2ab24) +YJIT [x86_64-linux]

----------  -----------  ----------  ----------  ----------  ------------  -------------
bench       before (ms)  stddev (%)  after (ms)  stddev (%)  before/after  after 1st itr
30k_ifelse  1482.5       0.0         1541.2      0.0         0.96          0.96
----------  -----------  ----------  ----------  ----------  ------------  -------------
```

yjit/src/backend/ir.rs
k0kubun: removed `Insn::SideExitContext` at https://github.com/ruby/ruby/pull/7712/commits/a61a4cb7adcd78a307eb7e4e728f611ed0aeabfd.

maximecb (approved):
Good work :)

PR #7715, opened 2023-04-14
Add spec for `Warning[:performance]`

casperisfine:
[Feature #19538]

cc @eregon

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,12 @@ Note that each entry is kept to a minimum, see links for details.
 
 ## Language changes
 
+## Command line options
+
+* A new `performance` warning category was introduced.
+  They are not displayed by default even in verbose mode.
+  Turn them on with `-W:performance` or `Warning[:performance] = true`. [[Feature #19538]]
+
 ## Core classes updates
 
 Note: We're only listing outstanding class updates.
@@ -79,5 +85,7 @@ changelog for details of the default gems or bundled gems.
 ## JIT
 
 [Bug #19150]:     https://bugs.ruby-lang.org/issues/19150
+[Feature #18498]: https://bugs.ruby-lang.org/issues/18498
 [Feature #19314]: https://bugs.ruby-lang.org/issues/19314
 [Feature #19347]: https://bugs.ruby-lang.org/issues/19347
+[Feature #19538]: https://bugs.ruby-lang.org/issues/19538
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -7,6 +7,10 @@ Note that each entry is kept to a minimum, see links for details.
 
 ## Language changes
 
+* A new `performance` warning category was introduced.
+  They are not displayed by default even in verbose mode.
+  Turn them on with `-wW:performance` or `Warning[:performance] = true`. [Feature #19538]
```
eregon, line 12:
```suggestion
  Turn them on with `-W:performance` or `Warning[:performance] = true`. [Feature #19538]
```

spec/ruby/core/warning/element_reference_spec.rb
```diff
@@ -8,6 +8,14 @@
     end
   end
 
+  ruby_version_is '3.3' do
+    it "returns default values for :performance category" do
+      ruby_exe('p Warning[:performance]').chomp.should == "false"
+      ruby_exe('p Warning[:performance]', options: "-w").chomp.should == "false"
+      ruby_exe('p Warning[:performance]', options: "-W").chomp.should == "false"
```
eregon, line 15: `-W` is the same as `-w` so I'd test just `-w`, because subprocess tests like this are not so cheap.

spec/ruby/core/warning/element_reference_spec.rb
```diff
@@ -8,6 +8,14 @@
     end
   end
 
+  ruby_version_is '3.3' do
+    it "returns default values for :performance category" do
+      ruby_exe('p Warning[:performance]').chomp.should == "false"
```
eregon, line 13: This one could be just `Warning[:performance].should == false` to avoid an extra subprocess.
mspec will most likely never set `Warning[:performance]` so that's not a concern.

eregon (approved):
Thanks!

NEWS.md
eregon: I wonder, did you maybe mean teh `-wW:performance` here? Are performance warnings emitted with VERBOSE=false or they VERBOSE=true (in addition to the category being enabled). It seems best to emit them at VERBOSE=false level, so it's much easier to enable them, since they are already opt-in.

casperisfine: Maybe I misinterpreted https://bugs.ruby-lang.org/issues/19538#note-9 ?

>  said that ruby -w should not enable this warning mode. ruby -W:performance should do.

PR #7718, opened 2023-04-14
YJIT: Fix false object collection when setting ivar

XrXr:
Previously, setinstancevariable could generate code that calls
`rb_ensure_iv_list_size()` without first updating `cfp->sp`. This means
in the event that a GC start from within said routine the top few
objects would not be marked, causing them to be falsly collected.

Call `jit_prepare_routine_call()` first.

[[Bug #19601]](https://bugs.ruby-lang.org/issues/19601)

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2290,6 +2290,11 @@ fn gen_setinstancevariable(
                 if needs_extension {
                     // Generate the C call so that runtime code will increase
                     // the capacity and set the buffer.
+                    asm.comment("call rb_ensure_iv_list_size");
+
+                    // It allocates so can trigger GC, which takes the VM lock
+                    // so could yield to a different ractor.
+                    jit_prepare_routine_call(jit, asm);
                     asm.spill_temps(); // for ccall
                     asm.ccall(rb_ensure_iv_list_size as *const u8,
                               vec![
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2295,7 +2295,6 @@ fn gen_setinstancevariable(
                     // It allocates so can trigger GC, which takes the VM lock
                     // so could yield to a different ractor.
                     jit_prepare_routine_call(jit, asm);
-                    asm.spill_temps(); // for ccall
                     asm.ccall(rb_ensure_iv_list_size as *const u8,
                               vec![
                                   recv,
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1,3 +1,23 @@
+# Regression test for GC mishap while doing shape transition
+assert_equal '[:ok]', %q{
+  # [Bug #19601]
+  class RegressionTest
+    def initialize
+      @a = @b = @fourth_ivar_does_shape_transition = nil
+    end
+
+    def extender
+      @first_extended_ivar = [:ok]
+    end
+  end
+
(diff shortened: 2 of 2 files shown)


yjit/src/codegen.rs
```diff
@@ -2342,6 +2342,11 @@ fn gen_setinstancevariable(
                 if needs_extension {
                     // Generate the C call so that runtime code will increase
                     // the capacity and set the buffer.
+                    asm.comment("call rb_ensure_iv_list_size");
+
+                    // It allocates so can trigger GC, which takes the VM lock
+                    // so could yield to a differnt ractor.
+                    jit_prepare_routine_call(jit, asm);
                     asm.spill_temps(); // for ccall
```
XrXr, line 2350: This `asm.spill_temps()` is dominated by the other call above on line 2302. Maybe one of the calls should be removed?

k0kubun: That seems right. 

`jit_prepare_routine_call` calls `asm.spill_temps()` internally, so your change should obviate it anyway.

```suggestion
```

k0kubun:
Not for this PR, but with this amount of `jit_prepare_routine_call`, we might be calling `jit_save_pc` multiple times for a single instruction. I guess we should have a PC offset-like concept in `JITState` and skip `jit_save_pc` when it's been already written.

k0kubun (approved):

XrXr:
Uh oh, the regression test is so good that it crashes RJIT even after my fix attempt at [f889a96](https://github.com/ruby/ruby/pull/7718/commits/f889a967dc33cbb1c3ca933d82b963cd33e92a05). Will need to debug that further, or skip it for now for RJIT?

k0kubun:
No worries. Please skip it for RJIT for now.

maximecb (approved):

XrXr:
> Not for this PR, but with this amount of jit_prepare_routine_call, we might be calling jit_save_pc multiple times for a single instruction. I guess we should have a PC offset-like concept in JITState and skip jit_save_pc when it's been already written.

Yeah. Kind of related to that, given how often we've had this class of bugs, I think ccall should do `jit_prepare_routine_call` by default. Should be a much safer default, since calling without prepping requires carefully checking the called routine. It's not trivial to change the default right now because the assembler doesn't have enough info to schedule the cfp updates correctly, but maybe in the future. When it can do that, it can elide duplicate PC updates, too.

k0kubun:
If it doesn't cause a significant performance drawback, that'd be nice.

> I think ccall should do jit_prepare_routine_call by default.

`asm.ccall` currently doesn't take `JITState`, which is required for `jit_prepare_routine_call`. Instead of adding a `JITState` parameter to it, another approach could be to just assert that `sp_offset` is `0` and local types have been cleared, which is now possible thanks to `Assembler` owning `Context`. It could still leave a stale PC, but it would still catch a missing `jit_prepare_routine_call` in most cases.

maximecb:
> asm.ccall currently doesn't take JITState, which is required for jit_prepare_routine_call. Instead of adding a JITState parameter to it, another approach could be to just assert that sp_offset is 0 and local types have been cleared, which is now possible thanks to Assembler owning Context. It could still leave a stale PC, but it would still catch a missing jit_prepare_routine_call in most cases.

Seems like a decent solution.

Another option could be to remove the `ccall` method from assembler, and make it a method in JITState instead.

PR #7720, opened 2023-04-14
YJIT: Add a counter to all side exits

k0kubun:
This changes `Target::SideExit(Option<Counter>)` to `Target::SideExit(Counter)` to obviate `Some()` and to make exits counted by default.


maximecb (approved):
Nice 👌

PR #7725, opened 2023-04-17
YJIT: Move exits in gen_send_iseq to functions and use ?

jimmyhmiller:
Happy to bike shed on names. Just wanted to get the shape out there.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -245,21 +245,17 @@ pub enum JCCKinds {
     JCC_JNA,
 }
 
-macro_rules! gen_counter_incr {
-    ($asm:tt, $counter_name:ident) => {
-        if (get_option!(gen_stats)) {
-            // Get a pointer to the counter variable
-            let ptr = ptr_to_counter!($counter_name);
+#[inline(always)]
+fn gen_counter_incr(asm: &mut Assembler, counter: Counter) {
+    if get_option!(gen_stats) {
+        asm.comment(&format!("increment counter {}", counter.get_name()));
+        let ptr = get_counter_ptr(&counter.get_name());
+        let ptr_reg = asm.load(Opnd::const_ptr(ptr as *const u8));
+        let counter_opnd = Opnd::mem(64, ptr_reg, 0);
 
-            // Load the pointer into a register
-            $asm.comment(&format!("increment counter {}", stringify!($counter_name)));
-            let ptr_reg = $asm.load(Opnd::const_ptr(ptr as *const u8));
-            let counter_opnd = Opnd::mem(64, ptr_reg, 0);
-
-            // Increment and store the updated value
-            $asm.incr_counter(counter_opnd, Opnd::UImm(1));
-        }
-    };
+        // Increment and store the updated value
+        asm.incr_counter(counter_opnd, Opnd::UImm(1));
+    }
 }
 
 // Save the incremented PC on the CFP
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -245,6 +245,16 @@ pub enum JCCKinds {
     JCC_JNA,
 }
 
+fn gen_increment_count(asm: &mut Assembler, counter: Counter) {
```
k0kubun, line 248: I think this should check `if (get_option!(gen_stats))` like `gen_counter_incr` does. `gen_counter_incr` could share this implementation if it calls `gen_increment_count(asm, Counter::$counter_name)`.

yjit/src/codegen.rs
```diff
@@ -6227,6 +6135,155 @@ fn gen_send_iseq(
...
+fn exit_if_has_rest_and_optional_and_block(asm: &mut Assembler, iseq_has_rest: bool, opt_num: i32, iseq: *const rb_iseq_t, block_arg: bool) -> Option<()> {
+    exit_if(
+        asm,
+        iseq_has_rest && opt_num != 0 && (unsafe { get_iseq_flags_has_block(iseq) } || block_arg),
+        Counter::send_iseq_has_rest_opt_and_block
+    )
+}
+
+#[must_use]
+fn exit_if_unsupported_block_arg_type(asm: &mut Assembler, block_arg_type: Option<Type>) -> Option<()> {
```
k0kubun, line 6259: `gen_send_cfunc` could use this too, but is it out of scope in this PR?

yjit/src/codegen.rs
jimmyhmiller: Done. Should we just replace gen_counter_incr! with a non-macro version?

yjit/src/codegen.rs
jimmyhmiller: I assume cfunc and iseq will probably get support for block types at different times. So I didn't extract that out into a function.

k0kubun: [This place](https://github.com/ruby/ruby/blob/165ef42b3c29414a29c3720bd353c990bb9a6638/yjit/src/codegen.rs#L5545-L5556) currently has the exact same code, so I thought you could just call `exit_if_unsupported_block_arg_type(asm, block_arg_type)?;` there, but do you mean it's gonna change soon?

k0kubun: But adding support for other block types to cfunc and iseq at different times does make sense. If sharing it makes it more difficult, it's fine to leave it as is. I wonder how you're going to do the same kind of refactoring on `gen_send_cfunc` though (the function names will conflict if it's not shared).

yjit/src/codegen.rs
maximecb: We could replace gen_counter_incr with the method version and use the same name as before.

yjit/src/codegen.rs
```diff
@@ -6227,6 +6126,155 @@ fn gen_send_iseq(
     Some(EndBlock)
 }
 
+fn exit_if(asm: &mut Assembler, pred: bool, counter: Counter) -> Option<()> {
```
maximecb, line 6129: Should comment to explain what this is used for. Put yourself in the shoes of someone who's never seen this codebase and maybe is relatively new to Rust.

yjit/src/codegen.rs
jimmyhmiller: @maximecb Do you want me to do this as part of this PR? Or separately?

maximecb: Would be nice if it was in this PR.

jimmyhmiller: @maximecb Done

XrXr:
> For some reason the actions are pulling code that doesn't match what I see on the branch

PR action jobs run against a merged SHA. It's shown in the environment: `GITPULLOPTIONS: --no-tags origin refs/pull/7725/merge`.

jimmyhmiller:
I can recreate the failures I was getting on master. Running a bisect

PR #7729, opened 2023-04-17
[Bug #19592] Fix ext/Setup support

XrXr:
https://bugs.ruby-lang.org/issues/19592


configure.ac
```diff
@@ -3009,6 +3009,9 @@ AS_CASE([",$EXTSTATIC,"], [,static,|*,enc,*], [
 ])
 AC_SUBST(ENCOBJS)
 AC_SUBST(EXTOBJS)
+AS_IF([[grep '^\s*[^#]' "$srcdir/ext/Setup" >/dev/null]], [dnl selective extension static linking
+  AC_DEFINE_UNQUOTED(EXTSTATIC, 1)dnl build support code
+])
```
nobu, line 3014: This doesn't work for `option nodynamic` in ext/Setup.
And ext/Setup can be placed not only in the build directory too, but also can be given with `--with-setup=` option.

nobu: Also `ext/Setup.$target_os`.

XrXr: With this many options available, trying to do this in `configure` seems infeasible, then. I'll look into unconditionally including the support code and exporting `ruby_init_ext`.

PR #7733, opened 2023-04-18
Make the maximum shapes variation warning non-verbose

casperisfine:
[Feature #19538]

Since that category is not enabled by default, making it a verbose warning is redundant. Enabling performance warning should work with the default verbosity level.

Ref: https://github.com/ruby/ruby/pull/7715#discussion_r1169777099

cc @eregon @mame


eregon:
Thanks!

PR #7740, opened 2023-04-19
YJIT: invokesuper: Remove cme mid matching check

jhawthorn:
This check was introduced to match an assertion in the C YJIT when this was originally introduced. I don't believe it's necessary for correctness of the generated code.

In a test page on GitHub we found that [this invokesuper call in Rails](https://github.com/rails/rails/blob/076003d8e66ba2ad7f8350742675c6bfaaa493ba/activesupport/lib/active_support/core_ext/erb/util.rb#L15) was responsible for over 15% of our side-exits. So I think we should find a way to make it work. We won't see this in Railsbench because it uses an old version of Rails.

cc @XrXr I know we discussed this a bit when initially introduced. Per the comment I don't believe this ever actually needed that method lookup to hit, and the invalidation I think is just so we don't leak the object if it is invalidated (the correctness in invokesuper comes from checking the comptime cfp's ME against the runtime ME).

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -777,6 +777,25 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_super_with_alias
+    assert_compiles(<<~'RUBY', insns: %i[invokesuper opt_plus opt_mult], result: 15)
+      class A
+        def foo = 1 + 2
+      end
+
+      module M
+        def foo = super() * 5
+        alias bar foo
+
+        def foo = :bad
+      end
+
+      A.prepend M
+
+      A.new.bar
+    RUBY
+  end
+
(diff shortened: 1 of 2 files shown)


k0kubun:
> We won't see this in Railsbench because it uses an old version of Rails.

Not related to this PR, but it sounds like yjit-bench's Railsbench is too deviated from today's production environments at this point. We should upgrade it to Rails 7.

maximecb:
> Not related to this PR, but it sounds like yjit-bench's Railsbench is too deviated from today's production environments at this point. We should upgrade it to Rails 7.

Agreed.

XrXr (approved):
Makes sense, and thanks for including a test! The assertion that's already gone didn't accommodate `invokesuper` well, and now makes even less sense now that we only track method entry validity.

PR #7741, opened 2023-04-19
YJIT: Remove Insn::RegTemps

k0kubun:
This PR refactors `RegTemps` handling. We currently copy `RegTemps` for every insn in Assembler, but this PR removes it by moving it to `Opnd::Stack`. This obviates assertion that checks `asm.ctx.reg_temps` and `asm.reg_temps` are consistent.

This also makes compilation faster.

```
before: ruby 3.3.0dev (2023-04-19T06:59:59Z master 7477284043) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-04-19T18:18:26Z yjit-reg-temps d91062cfb2) +YJIT [x86_64-linux]

30k_ifelse  1519.3       0.0         1292.5      0.0         1.18           1.18
```


yjit/src/core.rs
```diff
@@ -1623,9 +1623,11 @@ impl Context {
 
     /// Stop using a register for a given stack temp.
     pub fn dealloc_temp_reg(&mut self, stack_idx: u8) {
```
maximecb, line 1625: What is the difference with this and spill?

k0kubun: `spill_temp` is lowered to `Insn::Mov` whereas `dealloc_temp_reg` only mutates `ctx.reg_temps` and generates no code.

k0kubun: Exactly. It's mostly used for stack slots that are above the stack top. It's not safe to automatically do this thing at `ctx.stack_pop` because that could still be used, but it never crosses YARV instruction boundaries, so we use this mainly at the YARV instruction boundary.

yjit/src/core.rs
```diff
@@ -1623,9 +1623,11 @@ impl Context {
 
     /// Stop using a register for a given stack temp.
```
maximecb, line 1624:
```suggestion
    /// Stop using a register for a given stack temp.
    /// This allows us to reuse the register for a value that we know is dead
    /// and will no longer be used (e.g. popped stack temp).
```

PR #7659, opened 2023-04-04
YJIT: Stack temp register allocation for arm64

k0kubun:
Following up https://github.com/ruby/ruby/pull/7651, this PR adds arm64 support for stack temp register allocation. Because benchmark results are generally better in both x86_64 and arm64, this PR also changes the default of `--yjit-temp-regs` to `5`.

## Benchmark
The speedups on M1 seem less dramatic compared to my Linux x86_64 environment (maybe because M1 is faster on memory access?). But still benchmark results seem generally better than not enabling it.

```
regs=0: ruby 3.3.0dev (2023-04-04T23:36:32Z yjit-stack-arm64 ca7446cd37) +YJIT [arm64-darwin22]
regs=5: ruby 3.3.0dev (2023-04-04T23:36:32Z yjit-stack-arm64 ca7446cd37) +YJIT [arm64-darwin22]

--------------  -----------  ----------  -----------  ----------  -------------  --------------
bench           regs=0 (ms)  stddev (%)  regs=5 (ms)  stddev (%)  regs=0/regs=5  regs=5 1st itr
activerecord    17.7         2.3         17.5         2.2         1.01           0.89
erubi_rails     6.1          10.6        6.1          9.8         1.00           0.65
hexapdf         922.7        1.9         916.6        2.1         1.01           0.97
liquid-c        24.5         3.6         24.4         3.6         1.01           0.82
liquid-render   49.0         3.3         48.4         3.6         1.01           0.86
mail            57.1         2.2         55.3         1.7         1.03           0.87
psych-load      888.2        0.4         880.4        0.5         1.01           1.01
railsbench      646.9        1.6         647.6        1.7         1.00           0.92
ruby-lsp        32.0         26.4        32.2         34.7        1.00           1.02
sequel          30.3         1.6         30.4         1.4         0.99           1.00
binarytrees     112.4        2.6         110.2        2.6         1.02           1.02
chunky_png      370.1        0.3         362.7        0.4         1.02           1.00
erubi           121.4        1.5         121.8        2.1         1.00           1.00
etanni          210.6        0.9         207.3        0.9         1.02           1.02
fannkuchredux   446.0        0.4         401.5        0.2         1.11           1.00
lee             482.6        0.7         474.6        0.9         1.02           1.48
nbody           41.3         0.7         42.3         0.7         0.98           0.96
optcarrot       1434.6       0.6         1364.9       0.6         1.05           1.03
ruby-json       1496.1       0.4         1506.0       0.4         0.99           0.99
[...]

PR #7745, opened 2023-04-20
YJIT: Avoid splitting mov for small values on arm64

k0kubun:
When `Mov`'s destination is a register and the source is an immediate which cannot be encoded with a bitmask but is <= `0xffff`, it can be a single `mov` instruction. 

Currently, `split_bitmask_immediate` splits such source operands, which are still necessary for other instructions, so this PR avoids calling it for that case.

While I don't see a significant impact on benchmarks, this should help the performance of stack temp register allocation on arm64 https://github.com/ruby/ruby/pull/7659 a little.

### before
```asm
  # Insn: 0000 putobject (stack_size: 0)
  # reg_temps: 00000000 -> 00000001
  0x1065d0090: mov x11, #5
  0x1065d0094: mov x1, x11
```

### after
```asm
  # Insn: 0000 putobject (stack_size: 0)
  # reg_temps: 00000000 -> 00000001
  0x103840090: mov x1, #5
```

### DIFF
--- a/yjit/src/backend/arm64/mod.rs
+++ b/yjit/src/backend/arm64/mod.rs
@@ -605,7 +605,11 @@ impl Assembler
                         },
                         // Otherwise we'll use the normal mov instruction.
                         (Opnd::Reg(_), _) => {
-                            let value = split_bitmask_immediate(asm, src, dest.rm_num_bits());
+                            let value = match src {
+                                // Unlike other instructions, we can avoid splitting this case, using movz.
+                                Opnd::UImm(uimm) if uimm <= 0xffff => src,
+                                _ => split_bitmask_immediate(asm, src, dest.rm_num_bits()),
+                            };
                             asm.mov(dest, value);
                         },
                         _ => unreachable!()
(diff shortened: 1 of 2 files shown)


yjit/src/backend/arm64/mod.rs
```diff
@@ -956,7 +960,18 @@ impl Assembler
                     };
                 },
                 Insn::Mov { dest, src } => {
-                    mov(cb, dest.into(), src.into());
+                    // This supports the following two kinds of immediates:
+                    //   * The value fits into a single movz instruction
+                    //   * It can be encoded with the special bitmask immediate encoding
+                    // arm64_splat should have split other immediates that require multiple instructions.
```
XrXr, line 966:
```suggestion
                    // arm64_split() should have split other immediates that require multiple instructions.
```

yjit/src/backend/arm64/mod.rs
```diff
@@ -1588,4 +1603,16 @@ mod tests {
             0x0: ldur x1, [x19, #8]
         "});
     }
+
+    #[test]
+    fn test_not_split_mov() {
+        let (mut asm, mut cb) = setup_asm();
+
+        asm.mov(Opnd::Reg(Assembler::TEMP_REGS[0]), Opnd::UImm(5));
```
XrXr, line 1611: I would test with 0xffff and 0x10000 to show that it handles the boundary condition properly.

XrXr (approved):
Nice!

Off topic, sometimes our `asm.mov` generates `orr` in the disassembly because I think [for small values `movz` gets the `mov` mnemonic](https://developer.arm.com/documentation/ddi0596/2021-12/Base-Instructions/MOVZ--Move-wide-with-zero-). Might want to change the selection there just for the looks of the disassembly.

maximecb (approved):

k0kubun:
> Off topic, sometimes our asm.mov generates orr in the disassembly because I think [for small values movz gets the mov mnemonic](https://developer.arm.com/documentation/ddi0596/2021-12/Base-Instructions/MOVZ--Move-wide-with-zero-). Might want to change the selection there just for the looks of the disassembly.

I would like a non-`orr` output too, but is there an instruction that's useful for this? I believe we can't control what libcapstone outputs when multiple instructions have the same encoding, so I'm assuming you're talking about a different instruction, but is that still `mov`-like one?

XrXr:
I'm not very sure what exactly is going on either. It might be wishful thinking that Capstone tries to follow the spec when picking mnemonics. I'll start taking notes when I see orr and not spend dedicated time on this hah.

PR #7755, opened 2023-04-24
YJIT: Show definedivar exit reasons

k0kubun:
`definedivar` makes up 4.3% of exit reasons on SFR. It'd be nice to show a breakdown of it. 

I bet it's mostly `definedivar_megamorphic` and we want to put a general case at the end of chains, but I want to double-check it first.


maximecb (approved):
Wise to gather more data 👌

PR #7758, opened 2023-04-26
Prevent segfault when calling `defined?(super)` in `BasicObject`

garyhtou:
Prior to this commit, a segmentation fault occurred in `vm_defined`'s `zsuper` implementation after NULL is returned as `BasicObject`'s superclass. This fix returns false from `vm_defined` if the superclass is NULL.

For example, the following code resulted in a segfault.
```ruby
class BasicObject
  def seg_fault
    defined?(super)
  end
end

seg_fault
```

### Intended behavior
`defined?(super)` and the `seg_fault` method returns `nil`.

<br/>
I've also added a test for this:

```sh
$ make test-all TESTS="ruby/test_defined.rb --name=/test_super_in_basic_object/ -v"
```

---

Special thanks to @rafaelfranca and @matthewd who helped me debug this!

### DIFF
--- a/test/ruby/test_defined.rb
+++ b/test/ruby/test_defined.rb
@@ -311,9 +311,9 @@ class TestDefined < Test::Unit::TestCase
     end
 
     assert_nil(a)
-
+  ensure
     BasicObject.class_eval do
-      undef_method :a
+      undef_method :a if defined?(a)
     end
   end
 
--- a/test/ruby/test_defined.rb
+++ b/test/ruby/test_defined.rb
@@ -303,6 +303,20 @@ class TestDefined < Test::Unit::TestCase
     assert_equal("super", o.x, bug8367)
   end
 
+  def test_super_in_basic_object
+    BasicObject.class_eval do
+      def a
+        defined?(super)
+      end
+    end
+
+    assert_nil(a)
+
+    BasicObject.class_eval do
+      undef_method :a
+    end
+  end
+
   def test_super_toplevel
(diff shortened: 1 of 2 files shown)


test/ruby/test_defined.rb
```diff
@@ -303,6 +303,20 @@ def x; b {return defined?(super)}; end
...
+  def test_super_in_basic_object
+    BasicObject.class_eval do
+      def a
+        defined?(super)
+      end
+    end
+
+    assert_nil(a)
+
+    BasicObject.class_eval do
```
rafaelfranca, line 315: Let’s make sure this is called in an ensure block in case assert_nil (or other code) raises an exception.

vm_insnhelper.c
```diff
@@ -4791,6 +4791,8 @@ vm_defined(rb_execution_context_t *ec, rb_control_frame_t *reg_cfp, rb_num_t op_
 
             if (me) {
                 VALUE klass = vm_search_normal_superclass(me->defined_class);
+                if (klass == (VALUE)NULL) return false;
```
rafaelfranca, line 4794: maybe the following would work as well?

```suggestion
                if (!klass) return false;
```

test/ruby/test_defined.rb
garyhtou: great idea! https://github.com/ruby/ruby/pull/7758/commits/7c78de578e0f1c0f1a16f9610a32e18fa23a7a25

PR #7760, opened 2023-04-26
Fix infinite loop in peephole optimizer doing useless jump elimination

KJTsanaktsidis:
If you have a pair of jumps which point at each other, preceeded by a jump which points at one of them too, iseq_peephole_optimize can loop forever switching the target of the first jump alternately between the other two.

Fixes [Bug #19611]

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -2795,6 +2795,24 @@ find_destination(INSN *i)
     return 0;
 }
 
+static bool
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis:
https://github.com/ruby/ruby/pull/7754 is a better fix for this.

PR #7764, opened 2023-04-26
marshal.c: shallow freeze user objects

casperisfine:
When `freeze: true` argument is passed.

[Bug #19427]

cc @eregon


eregon:
Thank you!

PR #7772, opened 2023-04-28
Change double quotes to single quotes.

zedbydade:
We don't have any string interpolation here so we can use single quotes.

### DIFF
--- a/doc/fiber.md
+++ b/doc/fiber.md
@@ -9,21 +9,21 @@ Fibers execute a user-provided block. During the execution, the block may call `
 ``` ruby
 #!/usr/bin/env ruby
 
-puts "1: Start program."
+puts '1: Start program.'
 
 f = Fiber.new do
-  puts "3: Entered fiber."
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thank you for submitting a pull request. I don't see a benefit from switching from double quoted strings to single quoted strings here. Can you explain why this change should be made?

zedbydade:
I think It's just coding best practice since most of the Ruby Linters suggest using single quote literals for your strings and going for the double ones in the case of interpolation/escaping sequences.

tenderlove:
An argument could be made (and has been at previous places I've worked) that you always want to use double quoted strings as they _could_ be used for interpolation and have no cost to use when there is no interpolation.

We (on the Ruby core team) don't want to make any judgements about your use (or non use) of double / single quoted strings, so using a liter as justification doesn't hold up.  Thank you for sending this pull request, but I think we should leave this code as-is.

zedbydade:
Thanks for your review.

PR #7783, opened 2023-05-03
 Ensure the VM is alive before accessing objspace in C API (Feature #19627)

ianks:
It is not possible to ensure the Ruby VM is accessible in certain scenarios, such as [thread local destructors](https://github.com/bytecodealliance/wasmtime-rb/actions/runs/4867822840/jobs/8680761059) and [async signal handlers](https://github.com/tmm1/stackprof/pull/200). 

Previously, it was possible to workaround this by probing the hidden `ruby_current_vm_ptr` symbol. But as of https://github.com/ruby/ruby/pull/7459, those symbols are no longer visible.

To remedy this, this patch adds some null checks when accessing `objspace` to eliminate known safety issues in the public C API (`rb_during_gc`, `rb_gc` and `rb_gc_adjust_memory_usage`).

prior art: https://github.com/ruby/ruby/pull/7663
issue tracker: https://bugs.ruby-lang.org/issues/19627

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1007,11 +1007,11 @@ asan_unlock_freelist(struct heap_page *page)
 /* Aliases */
 #define rb_objspace (*rb_objspace_of(GET_VM()))
 #define rb_objspace_of(vm) ((vm)->objspace)
-#define rb_objspace_ptr() \
-  ({ \
-    rb_vm_t *rb_objspace_ptr_vm = GET_VM(); \
-    (UNLIKELY(rb_objspace_ptr_vm == NULL)) ? NULL : ((rb_objspace_ptr_vm)->objspace); \
-  })
+#define unless_objspace(objspace) \
+    rb_objspace_t *objspace; \
+    rb_vm_t *unless_objspace_vm = GET_VM(); \
+    if (unless_objspace_vm) objspace = unless_objspace_vm->objspace; \
+    else /* return; or objspace will be warned uninitialized */
 
 #define ruby_initial_gc_stress	gc_params.gc_stress
 
@@ -10919,7 +10919,7 @@ rb_gc_start(void)
 void
 rb_gc(void)
 {
-    rb_objspace_t *objspace = rb_objspace_ptr();
+    unless_objspace(objspace) { return; }
     if (!objspace) return;
     unsigned int reason = GPR_DEFAULT_REASON;
     garbage_collect(objspace, reason);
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -1007,6 +1007,11 @@ asan_unlock_freelist(struct heap_page *page)
 /* Aliases */
 #define rb_objspace (*rb_objspace_of(GET_VM()))
 #define rb_objspace_of(vm) ((vm)->objspace)
+#define rb_objspace_ptr() \
+  ({ \
+    rb_vm_t *rb_objspace_ptr_vm = GET_VM(); \
+    (UNLIKELY(rb_objspace_ptr_vm == NULL)) ? NULL : ((rb_objspace_ptr_vm)->objspace); \
+  })
```
nobu, line 1014: Statement expression is a gcc extension.

nobu (changes requested):

gc.c
```diff
@@ -10914,15 +10919,17 @@ rb_gc_start(void)
 void
 rb_gc(void)
 {
-    rb_objspace_t *objspace = &rb_objspace;
+    rb_objspace_t *objspace = rb_objspace_ptr();
```
byroot, line 10922: Rather than a `rb_objspace_ptr` macro, wouldn't it be more convenient to implement some kind of `EARLY_RETURN_IF_SHUTTING_DOWN` kind of help macro?

nobu: Just an idea 😜 

```C
#define unless_objspace(objspace) \
    rb_objspace_t *objspace; \
    rb_vm_t *rb_objspace_ptr_vm = GET_VM(); \
    if (rb_objspace_ptr_vm) objspace = rb_objspace_ptr_vm->objspace; \
    else /* return; or objspace will be warned uninitialized */

unless_objspace(objspace) {return;}
// or
unless_objspace(objspace) {return FALSE;}
```

ianks: I tried out the early return approach, but doing it right involved some `do while (0)` trickery, and I couldn’t quite get it to properly return default values.

This approach seemed simpler to implement and easier to understand, but happy to try another approach. `unless_objspace` seems good, I’ll give it a whirl.

ianks: Implemented in https://github.com/ruby/ruby/pull/7783/commits/77ad19c453d1165407608c82163dda5122ccdce7

gc.c
```diff
@@ -10914,15 +10919,16 @@ rb_gc_start(void)
 void
 rb_gc(void)
 {
-    rb_objspace_t *objspace = &rb_objspace;
+    unless_objspace(objspace) { return; }
+    if (!objspace) return;
```
byroot, line 10923:
```suggestion
    unless_objspace(objspace) { return; }
```

nobu (approved):

PR #7793, opened 2023-05-09
For YJIT stats, set avg_len_in_yjit to 0 if denominator would be 0

noahgibbs:
This field is only compiled for yjit-stats builds.

Currently the result is NaN if the total exit is 0, which is inconvenient. This should only occur if YJIT has not yet executed any compiled methods.


noahgibbs:
Nope, never mind, found another bug

noahgibbs:
And it was in my config. Reopening.

PR #7795, opened 2023-05-09
Added more sample ruby code

EzanaK:


sample/is_palindrome.rb
```diff
@@ -0,0 +1,14 @@
...
+    tail = length - 1
+    while head <= tail do
+        if s[head] != s[tail]
+            return false
+        end
+        head += 1
+        tail -= 1
+    end
+    return true
+end
```
nobu, line 14:
```suggestion
def is_palindrome(n)
  s = n.to_s
  (s.size / 2).times.all? {|i| s[i] == s[-1-i]}
end
```

sample/freq.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+    keys = map.keys
+    frequent = s[0]
+    for i in keys do
+        if (map[i] > map[frequent])
+            frequent = i
+        end
+    end
+    return frequent
+end
```
nobu, line 21:
```suggestion
def freq(s)
  if s.empty?
    return ""
  end
  map = Hash.new(0)
  s.each_char do |i|
    map[s[i]] += 1
  end
  map.max_by {|_, i| i}.first
end
```

sample/fib_numbers.rb
```diff
@@ -0,0 +1,13 @@
...
+        if i == 1
+            fib_arr.push(0)
+        elsif i == 2
+            fib_arr.push(1)
+        else
+            fib_arr.push(fib_arr[i - 3] + fib_arr[i - 2])
+        end
+    end
+    return fib_arr
+end
```
nobu, line 13:
```suggestion
def fib_numbers(n)
  case
  when n < 1
    return []
  when n == 1
    return [0]
  end

  fib_arr = [0, 1]
  3.upto(n) do
    fib_arr.push(fib_arr[-2] + fib_arr[-1])
  end
  return fib_arr
end
```

sample/hash_to_array.rb
```diff
@@ -0,0 +1,7 @@
+def hashToArray(hash)
+    arr = Array.new
+    for i in hash.keys do
+        arr.push(Array[i, hash[i]])
+    end
+    return arr
```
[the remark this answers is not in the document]
nobu: This sample just calls a core method, usually it is used directly instead, I think.

sample/freq.rb
```diff
@@ -0,0 +1,10 @@
+def freq(s)
+  if s.empty?
+    return ""
+  end
+  map = Hash.new(0)
+  s.each_char do |i|
+    map[s[i]] += 1
+  end
+  map.max_by {|_, i| i}.first
+end
```
nobu, line 10: Also `tally` method.

```ruby
def freq(s)
  each_char.tally.max_by {|_, i| i}&.first || ""
end
```

sample/zip_hash.rb
```diff
@@ -0,0 +1,10 @@
+def zipHash(arr1, arr2)
+    if arr1.size != arr2.size
+        return nil
+    end
+    map = Hash.new
+    for i in 0..arr1.size - 1 do
+        map.store(arr1[i], arr2[i])
+    end
+    return map
+end
```
nobu, line 10: Other ways:

```ruby
def zipHash(arr1, arr2)
  arr1.zip(arr2).to_h if arr1.size == arr2.size
end
```

or

```ruby
def zipHash(arr1, arr2)
  (0...arr1.size).to_h {|i| [arr1[i], arr2[i]]} if arr1.size == arr2.size
end
```

nobu (commented):
Why no newline at end of files?

sample/hash_to_array.rb
```diff
@@ -0,0 +1,3 @@
+def hashToArray(hash)
```
junaruga, line 1: Why is the method name [camel case](https://en.wikipedia.org/wiki/Camel_case) rather than [snake case](https://en.wikipedia.org/wiki/Snake_case) here?

hsbt:
Closing due to lack of feedback while 1 year. Basically, we only keep the current code with historical reason.

PR #7799, opened 2023-05-11
Introduce `math_arc` macro

S-H-GAMELINKS:
`math_acos` and `math_asin` functions has almost same code in `math.c`.

I thought better that cut out these codea and introduce `math_macro` that more simply.

```c
#define math_arc(num, msg, func) \
    double d; \
    d = Get_Double((num)); \
    domain_check_range(d, -1.0, 1.0, msg); \
    return DBL2NUM(func(d));
```

### DIFF
--- a/math.c
+++ b/math.c
@@ -170,6 +170,12 @@ math_tan(VALUE unused_obj, VALUE x)
     return DBL2NUM(tan(Get_Double(x)));
 }
 
+#define math_arc(num, func) \
+    double d; \
+    d = Get_Double((num)); \
+    domain_check_range(d, -1.0, 1.0, #func); \
+    return DBL2NUM(func(d));
+
 /*
  *  call-seq:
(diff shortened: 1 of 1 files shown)


math.c
```diff
@@ -170,6 +170,12 @@ math_tan(VALUE unused_obj, VALUE x)
     return DBL2NUM(tan(Get_Double(x)));
 }
 
+#define math_arc(num, msg, func) \
+    double d; \
+    d = Get_Double((num)); \
+    domain_check_range(d, -1.0, 1.0, msg); \
```
nobu, line 176: You can omit to repeat `func` and `msg`.

```suggestion
#define math_arc(num, func) \
    double d; \
    d = Get_Double((num)); \
    domain_check_range(d, -1.0, 1.0, #func); \
```

S-H-GAMELINKS: Thanks!
Fixed it!

PR #7800, opened 2023-05-12
[DOC] Add entry for CSV backslash quote liberal parsing option

markaschneider:


nobu:
Could you send this PR to the upstream, https://github.com/ruby/csv?
It will be incorporated once merged there.

kou:
nobu is correct.
Please resend this to https://github.com/ruby/csv .

markaschneider:
Thanks.  Apologies for my mistake.

PR #7815, opened 2023-05-15
Bypass `rb_sigwait_fd_get`.

ioquatix:
See https://github.com/ruby/ruby/pull/7482 for more details.


ioquatix:
Too much of `thread*.c` depends on this working, even thought it should not.

PR #7816, opened 2023-05-15
Remove explicit SIGCHLD handling.

ioquatix:
Related to https://github.com/ruby/ruby/pull/7815.


ioquatix:
cc @k0kubun @jeremyevans

PR #7821, opened 2023-05-17
[Feature #19678] Don't immediately promote children of old objects

peterzhu2118:
Alternative implementation of #7683 where the feature is always enabled.

References from an old object to a write barrier protected young object will not immediately promote the young object. Instead, the young object will age just like any other object, meaning that it has to survive three collections before being promoted to the old generation. References from an old object to a write barrier unprotected object will place the parent object in the remember set for marking during minor collections. This allows the child object to be reclaimed in minor collections at the cost of increased time for minor collections.

On one of [Shopify's highest traffic Ruby apps, Storefront Renderer](https://shopify.engineering/how-shopify-reduced-storefront-response-times-rewrite), we saw significant improvements after deploying this feature in production. We compare the GC time and response time of web workers that have the original behaviour (non-experimental group) and this new behaviour (experimental group). We see that with this feature we spend significantly less time in the GC, 0.81x on average, 0.88x on p99, and 0.45x on p99.9.

This translates to improvements in average response time (0.96x) and p99 response time (0.92x).

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -959,6 +959,8 @@ struct heap_page {
     bits_t uncollectible_bits[HEAP_PAGE_BITMAP_LIMIT];
     bits_t marking_bits[HEAP_PAGE_BITMAP_LIMIT];
 
+    bits_t remembered_bits[HEAP_PAGE_BITMAP_LIMIT];
+
     /* If set, the object is not movable */
     bits_t pinned_bits[HEAP_PAGE_BITMAP_LIMIT];
 };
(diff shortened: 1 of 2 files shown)


gc.c
```diff
@@ -6885,29 +6863,14 @@ rgengc_check_relation(rb_objspace_t *objspace, VALUE obj)
 
     if (old_parent) { /* parent object is old */
         if (RVALUE_WB_UNPROTECTED(obj)) {
-            if (gc_remember_unprotected(objspace, obj)) {
-                gc_report(2, objspace, "relation: (O->S) %s -> %s\n", obj_info(old_parent), obj_info(obj));
-            }
+            rgengc_remember(objspace, old_parent);
         }
         else {
             if (!RVALUE_OLD_P(obj)) {
```
byroot, line 6869: Would it make sense to merge both conditions?

```c
if (RVALUE_WB_UNPROTECTED(obj) || !RVALUE_OLD_P(obj)) {
    rgengc_remember(objspace, old_parent);
}
```

gc.c
```diff
@@ -8944,16 +8906,11 @@ gc_writebarrier_incremental(VALUE a, VALUE b, rb_objspace_t *objspace)
...
             if (!RVALUE_WB_UNPROTECTED(b)) {
-                gc_report(1, objspace, "gc_writebarrier_incremental: [GN] %p -> %s\n", (void *)a, obj_info(b));
-                RVALUE_AGE_SET_OLD(objspace, b);
-
-                if (RVALUE_BLACK_P(b)) {
-                    gc_grey(objspace, b);
-                }
+                rgengc_remember(objspace, a);
             }
             else {
```
byroot, line 8911: Similarly here the only difference between the two branches is one calls `gc_report`. I think it could be simplified.

byroot (approved):
❤️

peterzhu2118:
I did, both the PR and commit references #19678.

casperisfine:
This was accepted right? Any blocker left on merging?

PR #7823, opened 2023-05-18
Add Fiber#kill, similar to Thread#kill.

ioquatix:


jsaak:
Thank you! <3

ioquatix:
This is mostly useful for implementing your own fiber scheduler, in general, you should use whatever the scheduler provides for stopping tasks.

jsaak:
Is it not part of the Fiber scheduler interface? When using Async for example, can you not use this method like in the tests?

ioquatix:
If you use the Fiber scheduler agnostic interfaces, you can use `Fiber#kill`. But if you use Async gem, use `Async::Task#stop`.

PR #7825, opened 2023-05-18
Remove writev support.

ioquatix:
https://bugs.ruby-lang.org/issues/19642


ioquatix:
I am no longer planning on doing this.

PR #7826, opened 2023-05-19
Improvements to `IO::Buffer` `read`/`write`/`pread`/`pwrite`.

ioquatix:
As discussed in <https://github.com/socketry/async/issues/224>.

### DIFF
--- a/io_buffer.c
+++ b/io_buffer.c
@@ -991,17 +991,23 @@ io_buffer_readonly_p(VALUE self)
     return RBOOL(rb_io_buffer_readonly_p(self));
 }
 
+static void
+io_buffer_lock(struct rb_io_buffer *buffer)
+{
+    if (buffer->flags & RB_IO_BUFFER_LOCKED) {
+        rb_raise(rb_eIOBufferLockedError, "Buffer already locked!");
+    }
+
+    buffer->flags |= RB_IO_BUFFER_LOCKED;
+}
+
 VALUE
 rb_io_buffer_lock(VALUE self)
 {
     struct rb_io_buffer *buffer = NULL;
     TypedData_Get_Struct(self, struct rb_io_buffer, &rb_io_buffer_type, buffer);
 
-    if (buffer->flags & RB_IO_BUFFER_LOCKED) {
-        rb_raise(rb_eIOBufferLockedError, "Buffer already locked!");
-    }
-
-    buffer->flags |= RB_IO_BUFFER_LOCKED;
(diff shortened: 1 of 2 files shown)


ioquatix:
https://github.com/ruby/ruby/pull/7827

ioquatix:
I want to write some more tests.

io_buffer.c
```diff
@@ -2508,23 +2617,12 @@ rb_io_buffer_read(VALUE self, VALUE io, size_t length, size_t offset)
 static VALUE
```
u3shit, line 2617: I think documentation needs update (`at least +length+ bytes`, length is optional).

io_buffer.c
```diff
@@ -2629,41 +2727,55 @@ rb_io_buffer_pread(VALUE self, VALUE io, rb_off_t from, size_t length, size_t of
...
```
u3shit, line 2727: Documentation

io_buffer.c
```diff
@@ -2433,18 +2439,120 @@ io_buffer_default_size(size_t page_size)
...
+    size_t total = 0;
     struct io_buffer_read_internal_argument *argument = _argument;
-    ssize_t result = read(argument->descriptor, argument->base, argument->size);
-    return rb_fiber_scheduler_io_result(result, errno);
+
+    while (true) {
+        ssize_t result = read(argument->descriptor, argument->base, argument->size);
+
+        if (result < 0) {
+            return rb_fiber_scheduler_io_result(result, errno);
```
u3shit, line 2540: `rb_fiber_scheduler_io_result` returns either the read bytes or errno but not both, which can be problematic if a read fails in the second or later iteration. But as a workaround the caller can call read with length=0 and manually handle the retries if it needs this info, so I think this is a good trade-off, because the alternative would be to return both the read bytes and the errno, which would make an ugly api (errors would no longer be exceptions).

ioquatix: I agree with you and I'm not sure if there is a better way to return "partial read or partial write + error". I think if you ask to read or write at least N bytes, if you don't get N bytes, it's an error. Otherwise, as you say, you should just manage the read/write process yourself.

io_buffer.c
ioquatix: I've updated the documentation for `read`/`write`/`pread`/`pwrite` and it should be more consistent: https://github.com/ruby/ruby/pull/7860

u3shit: Cool, it looks alright.

PR #7827, opened 2023-05-19
Add support for pread/pwrite on windows.

ioquatix:
This is an experimental PR to see if we can get it working.

### DIFF
--- a/include/ruby/win32.h
+++ b/include/ruby/win32.h
@@ -147,8 +147,16 @@ typedef int clockid_t;
 #define open			rb_w32_uopen
 #define close(h)		rb_w32_close(h)
 #define fclose(f)		rb_w32_fclose(f)
-#define read(f, b, s)		rb_w32_read(f, b, s)
-#define write(f, b, s)		rb_w32_write(f, b, s)
+
+#define read(f, b, s) rb_w32_read(f, b, s, NULL)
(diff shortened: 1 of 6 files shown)


include/ruby/win32.h
```diff
@@ -147,8 +147,16 @@ typedef int clockid_t;
...
-#define write(f, b, s)		rb_w32_write(f, b, s)
+
+#define read(f, b, s) rb_w32_read(f, b, s, NULL)
+#define write(f, b, s) rb_w32_write(f, b, s, NULL)
+
+#define HAVE_PREAD
+#define pread(f, b, s, o) rb_w32_read(f, b, s, &o)
+
+#define HAVE_PWRITE
+#define pwrite(f, b, s, o) rb_w32_write(f, b, s, &o)
```
nobu, line 158:
```suggestion
#define pread(f, b, s, o) rb_w32_read(f, b, s, &(o))

#define HAVE_PWRITE
#define pwrite(f, b, s, o) rb_w32_write(f, b, s, &(o))
```

ioquatix: @nobu it's a good idea, but in truth, if anyone puts an expression, it will fail to compile as `&(expression)` is invalid. So, maybe better to make actual function?

PR #7828, opened 2023-05-19
Hash.new: reject keyword arguments

casperisfine:
[[Feature #19236]](https://bugs.ruby-lang.org/issues/19236)

In Ruby 3.3, `Hash.new` shall raise an ArgumentError if keyword arguments are passed instead of treating them as an implicit positional Hash.

This will allow to safely introduce a `capacity` keyword argument in 3.4

Note: I'm not very familiar with `rb_scan_args / rb_get_kwargs`, I'm unsure if this is the best way to implement this feature.


tenderlove (approved):

ioquatix (approved):

mame:
In my understanding, Ruby 3.3 shall issue a deprecation warning, not an exception, for `Hash.new(keyword: value)`.

casperisfine:
@mame what was recorded in the dev meeting log is:

> Conclusion:

> In 3.3 it throws error all keyword argments to Hash.new. Then Ruby 3.4 allows that Hash.new will accept capacity keyword argument.

But if you think what was actually approved was a warning, can try to implement that instead.

mame:
The language barrier :cry: I think a warning is what @matz approved, and @nobu also said so.

In fact, we found several programs that use `Hash.new(something: val)`, but I don't think there are any programs that use it with the `capacity` keyword (i.e., `Hash.new(capacity: val)`). If we could make `Hash.new(something: val)` raise an exception suddenly, it would also be possible to introduce the `capacity` keyword at the same time.

casperisfine:
> The language barrier 😢

Ahah no worries, it's hard for all of us 🤗 .

Alright, I'll change the PR to emit a deprecation warning instead, and I'll also update the meeting log. Thanks for letting me know.

casperisfine:
Ok. PR updated to print a warning instead. I wasn't super inspired on the warning message, so suggestions welcome.

mame (approved):

mame:
Thanks! LGTM.

PR #7840, opened 2023-05-22
Add documentation for some stats counters, and about YJIT stats

noahgibbs:
Docs updates. Are there other counters that have been there a long time and aren't likely to change that we should document?

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -54,8 +54,7 @@ series = {VMIL 2021}
 
 ## Current Limitations
 
-YJIT may not be suitable for certain applications. It currently only supports macOS and Linux on x86-64 and arm64/aarch64 CPUs. YJIT will use more memory than the Ruby interpreter because the JIT compiler needs to generate machine code in memory and maintain additional state
-information.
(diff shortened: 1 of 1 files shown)


doc/yjit/yjit.md
```diff
@@ -94,12 +93,21 @@ make -j install
...
 
 ```sh
-# Configure in dev (debug) mode for development, build and install
+# Configure in lower-performance dev (debug) mode for development, build and install
 ./autogen.sh
 ./configure --enable-yjit=dev --prefix=$HOME/.rubies/ruby-yjit --disable-install-doc
 make -j install
 ```
 
+Dev mode includes extended YJIT statistics, but can be slow. For only statistics you can configure in stats mode:
```
k0kubun, line 102: This sounds like `dev` shows more statistics than `stats`, but I believe they both support the same set of statistics. Both of them are slower than release build for YARV insn counters. `dev` is still slower than `stats` due to `RUBY_DEBUG` enabling all interpreter assertions though.

k0kubun (approved):

doc/yjit/yjit.md
noahgibbs: Yes, sorry. Both have exactly the same stats. I'll amend to make that clearer.

maximecb:
Good work 👍

PR #7844, opened 2023-05-23
Move `ruby_node_name` to node.c and rename prefix of the function

yui-knk:


nobu:
`ruby_node_name` is prefixed with `ruby` instead of `rb`, because it can be called without a running interpreter, I think.
But it is ok.

yui-knk:
Thanks for your information. OK, I will merge this.

nobu:
Oops, the name was exported.
So, strictly speaking, this breaks binary compatibility.

nobu:
Sorry, source level incompatibility.
But this would be only for ripper probably.

yui-knk:
I renamed the function to the original by https://github.com/ruby/ruby/pull/7852.

PR #7846, opened 2023-05-23
Fix crash when replacing ST hash with AR hash

peterzhu2118:
With VWA, AR hashes are much larger than ST hashes. Hash#replace attempts to directly copy the contents of AR hashes into ST hashes so there will be memory corruption caused by writing past the end of memory.

This commit changes it so that if a ST hash is being replaced with an AR hash it will insert each element into the ST hash.


byroot (approved):
❤️

PR #7847, opened 2023-05-23
Add Regexp#timeout= setter

aharpole:
This makes it possible to set a custom timeout value for Regexp instances after initialization, or to be able to set a timeout for regexes that weren’t initialized using Regexp.new, like regex literals (though they're frozen and you'd need to `dup` them first to be able to do so).

### DIFF
--- a/test/ruby/test_regexp.rb
+++ b/test/ruby/test_regexp.rb
@@ -1813,4 +1813,16 @@ class TestRegexp < Test::Unit::TestCase
       re =~ s
     end
   end
+
+  def test_timeout_setter
+    re = /(a)\1{100000}/.dup
+    re.timeout = 0.1
+    assert_equal(0.1, re.timeout)
+    assert_raise(Regexp::TimeoutError) {re.match? "a"* 100000}
+  end
+
+  def test_timeout_setter_frozen
+    re = /(a)\1{100000}/
+    assert_raise(FrozenError) {re.timeout = 0.1}
+  end
 end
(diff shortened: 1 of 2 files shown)


jeremyevans:
Regexp literals have been frozen since Ruby 3.0, and you should not be able to modify the timeout for frozen regexps.

aharpole:
@jeremyevans ah, thank you! I have updated this to ensure that we don't mutate the regexp when frozen.

test/ruby/test_regexp.rb
```diff
@@ -1813,4 +1813,16 @@ def test_linear_performance
...
+    re.timeout = 0.1
+    assert_equal(0.1, re.timeout)
+    assert_raise(Regexp::TimeoutError) {re.match? "a"* 100000}
+  end
+
+  def test_timeout_setter_frozen
+    re = /(a)\1{100000}/
+    assert_raise(FrozenError) {re.timeout = 0.1}
+  end
+end
```
nobu, line 1828: Why removed a newline?

aharpole: whoops! this happened when I was trying to fix an unrelated whitespace issue.

nobu:
To make a regexp with a timeout, you can use `Regexp.new`.
As you can’t set the timeout to literals, I’m not sure if this method is useful.
what is the use-case?

nobu:
https://bugs.ruby-lang.org/issues/19694

PR #7856, opened 2023-05-24
Document throw/catch in the control expressions document [ci skip]

jeremyevans:
This are implemented as Kernel methods and not keywords, but I still think they are worth documenting with the other control flow expressions.


nobu:
I think that there should be links to `Kernel#throw` and `Kernel#catch`.

jeremyevans:
@nobu I added links to `Kernel#throw` and `Kernel#catch`.

nobu (approved):

PR #7857, opened 2023-05-25
YJIT: Introduce RubyVM::YJIT.stats_string

k0kubun:
To periodically report production stats output in SFR, we've been using an implementation similar to this. Currently, it requires SFR to modify `$stderr` and call the private method `_print_stats`.

Since it's already proven to be useful, we should make it a public interface to avoid the hacky implementation.

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -186,7 +186,7 @@ module RubyVM::YJIT
 
   # Format and print out counters as a String. This returns a non-empty
   # content only when --yjit-stats is enabled.
-  def self.printed_stats
+  def self.stats_string
     # Lazily require StringIO to avoid breaking miniruby
     require 'stringio'
     strio = StringIO.new
--- a/yjit.rb
+++ b/yjit.rb
@@ -184,6 +184,17 @@ module RubyVM::YJIT
     stats
   end
 
+  # Format and print out counters as a String. This returns a non-empty
+  # content only when --yjit-stats is enabled.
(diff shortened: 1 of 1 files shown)


yjit.rb
```diff
@@ -184,6 +184,17 @@ def self.runtime_stats(context: false)
...
 
+  # Format and print out counters as a String. This returns a non-empty
+  # content only when --yjit-stats is enabled.
+  def self.printed_stats
+    # Lazily require StringIO to avoid breaking miniruby
+    require 'stringio'
+    strio = StringIO.new
+    _print_stats(out: strio)
+    strio.rewind
+    strio.read
```
nobu, line 195:
```suggestion
    strio.string
```

yjit.rb
```diff
@@ -184,6 +184,16 @@ def self.runtime_stats(context: false)
     stats
   end
 
+  # Format and print out counters as a String. This returns a non-empty
+  # content only when --yjit-stats is enabled.
+  def self.printed_stats
```
maximecb, line 189:
```suggestion
  def self.stats_string
```

Seems like a good idea. Maybe this would be a more descriptive name?

k0kubun: :+1: Renamed https://github.com/ruby/ruby/pull/7857/commits/f3705b63d206eba9fedea05ca378dc0da78e7f86

maximecb (approved):

PR #7862, opened 2023-05-25
[DOC] Link fix

BurdetteLamar:


lib/irb.rb
```diff
@@ -973,7 +973,7 @@ class Binding
   #     Cooked potato: true
   #
   #
-  # See IRB@IRB+Usage for more information.
+  # See IRB@Usage for more information.
```
peterzhu2118, line 976: Can you send IRB changes upstream to https://github.com/ruby/irb?

nobu (approved):

peterzhu2118 (approved):

PR #7863, opened 2023-05-25
mkmf: Add the extra option `--with-verbose` to enable verbose mode.

junaruga:
The Ruby ticket is <https://bugs.ruby-lang.org/issues/19695>.

This PR is to add the new extra argument `--with-verbose` to the `mkmf.rb`, and the PR has 3 commits.

1. mkmf: Add the extra option `--with-verbose` to enable verbose mode.
2. mkmf: Add tests for `MakeMakefile#with_config`.
3. mkmf: Add tests to test the verbose mode in MakeMakefile#configuration.

## mkmf: Add the extra option `--with-verbose` to enable verbose mode.

The first commit is the main one of the feature.
    
This commit is to add an extra option to enable verbose mode (V=1) in the generated `Makefile` at runtime of the Ruby to print compiler command lines by the commands below when building native extensions. It's possible to enable the verbose mode by setting the environment variable `MAKEFLAGS="V=1"`[1] implemented in GNU make. However, I wanted to make a consistent user-interface not depending on the specific make's implementation.

```
$ ruby /path/to/extconf.rb -- --with-verbose
```

You can also add the extra option via rake-compiler gem.

```
$ rake compiler -- --with-verbose
```

If the extra option is not given, the value of the `RbConfig::CONFIG["MKMF_VERBOSE"]` enabled by the configure option below is used.

```
$ ./configure --enable-mkmf-verbose
```

[1] https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html

Co-authored-by: Nobuyoshi Nakada <nobu@ruby-lang.org>

## mkmf: Add tests for `MakeMakefile#with_config`.

This second commit is to test the `MakeMakefile#with_config` that is used in the main commit. As I couldn't find the test of the `with_config`, I created it. 

## mkmf: Add tests to test the verbose mode in MakeMakefile#configuration.

The third commit is the tests to test the use cases of the first commit.

### DIFF
--- a/test/mkmf/base.rb
+++ b/test/mkmf/base.rb
@@ -147,7 +147,10 @@ class TestMkmf < Test::Unit::TestCase
 
   include Base
 
-  def assert_separately(args, src, *rest, **options)
-    super(args + ["-r#{__FILE__}"], "extend TestMkmf::Base; setup\nEND{teardown}\n#{src}", *rest, **options)
+  def assert_separately(args, extra_args, src, *rest, **options)
+    super(args + ["-r#{__FILE__}"] + %w[- --] + extra_args,
+          "extend TestMkmf::Base; setup\nEND{teardown}\n#{src}",
+          *rest,
+          **options)
   end
 end
--- a/test/mkmf/test_config.rb
+++ b/test/mkmf/test_config.rb
@@ -1,15 +1,30 @@
 # frozen_string_literal: false
-$extmk = true
+require_relative 'base'
 
-require 'test/unit'
-require 'mkmf'
-
-class TestMkmfConfig < Test::Unit::TestCase
+class TestMkmfConfig < TestMkmf
   def test_dir_config
     bug8074 = '[Bug #8074]'
     lib = RbConfig.expand(RbConfig::MAKEFILE_CONFIG["libdir"], "exec_prefix"=>"/test/foo")
-    assert_separately %w[-rmkmf - -- --with-foo-dir=/test/foo], %{
+    assert_separately([], %w[--with-foo-dir=/test/foo], <<-"end;")
       assert_equal(%w[/test/foo/include #{lib}], dir_config("foo"), #{bug8074.dump})
-    }
+    end;
+  end
+
+  def test_with_config_with_arg_and_value
+    assert_separately([], %w[--with-foo=bar], <<-'end;')
+      assert_equal("bar", with_config("foo"))
+    end;
+  end
+
+  def test_with_config_with_arg_and_no_value
+    assert_separately([], %w[--with-foo], <<-'end;')
+      assert_equal(true, with_config("foo"))
+    end;
+  end
+
+  def test_with_config_without_arg
+    assert_separately([], %w[--without-foo], <<-'end;')
+      assert_equal(false, with_config("foo"))
+    end;
   end
 end
--- a/test/mkmf/test_configuration.rb
+++ b/test/mkmf/test_configuration.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: false
+require_relative 'base'
+
+class TestMkmfConfiguration < TestMkmf
+  def test_verbose_with_rbconfig_verbose_disabled
+    makefile = mkmf do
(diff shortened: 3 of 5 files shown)


test/mkmf/test_configuration.rb
```diff
@@ -0,0 +1,45 @@
...
+require_relative 'base'
+
+class TestMkmfConfiguration < TestMkmf
+  def test_verbose_arg
+    assert_separately([], %w[- -- --with-verbose], <<-'end;') #do
+      assert_equal(true, with_config("verbose"))
+    end;
+  end
+
+  def test_verbose_with_rbconfig_disabled
```
junaruga, line 11: I want to change the method name `test_verbose_with_rbconfig_disabled` to `test_verbose_with_rbconfig_verbose_disabled`, and `test_verbose_with_rbconfig_enabled` to `test_verbose_with_rbconfig_verbose_enabled`.

test/mkmf/test_config.rb
```diff
@@ -12,4 +12,22 @@ def test_dir_config
...
+    assert_separately %w[-rmkmf - -- --with-foo], %{
+      assert_equal(true, with_config("foo"))
+    }
+  end
+
+  def test_with_config_without_arg
+    assert_separately %w[-rmkmf - -- --without-foo], %{
+      assert_equal(false, with_config("foo"))
+    }
+  end
```
nobu, line 32: Maybe using base.rb here too?

test/mkmf/base.rb
```diff
@@ -147,7 +147,7 @@ def config_value(name)
 
   include Base
 
-  def assert_separately(args, src, *rest, **options)
-    super(args + ["-r#{__FILE__}"], "extend TestMkmf::Base; setup\nEND{teardown}\n#{src}", *rest, **options)
+  def assert_separately(args, extra_args, src, *rest, **options)
+    super(args + ["-r#{__FILE__}"] + extra_args, "extend TestMkmf::Base; setup\nEND{teardown}\n#{src}", *rest, **options)
```
nobu, line 151: We could always add `- --` here.

Or, instead of adding `extra_args`, what about move `args` after `"-r#{__FILE__}"`?

test/mkmf/test_config.rb
junaruga: Yes, I modified on your way.

test/mkmf/test_configuration.rb
junaruga: Renamed the method names.

test/mkmf/base.rb
junaruga: > We could always add - -- here.

I modified on your suggestion above.

test/mkmf/base.rb
```diff
@@ -147,7 +147,10 @@ def config_value(name)
 
   include Base
 
-  def assert_separately(args, src, *rest, **options)
-    super(args + ["-r#{__FILE__}"], "extend TestMkmf::Base; setup\nEND{teardown}\n#{src}", *rest, **options)
+  def assert_separately(args, extra_args, src, *rest, **options)
+    super(args + ["-r#{__FILE__}"] + %w[- --] + extra_args,
```
junaruga, line 151: In this part, I wanted to write `args + %w[-r#{__FILE__} - --] + extra_args`. However, I saw the `#{__FILE__}` was not converted to the value. So, the `["-r#{__FILE__}"] + %w[- --]` is a bit compromised way.

```
$ ruby -v
ruby 3.3.0dev (2023-05-25T12:45:21Z wip/mkmf-verbose-r.. cbeba2ed16) [x86_64-linux]

$ which irb
~/.local/ruby-cbeba2ed16/bin/irb

$ irb
irb(main):001:0> %w[-r#{__FILE__} - --]
=> ["-r\#{__FILE__}", "-", "--"]

irb(main):002:0> %[-r#{__FILE__} - --]
=> "-r(irb) - --"
```

junaruga:
I rebased, fixing the review items. Thanks for that. I squashed the 3 commits to 1 commit.

PR #7868, opened 2023-05-26
[DOC] Link fixes

BurdetteLamar:


file.c
```diff
@@ -7135,7 +7135,7 @@ const char ruby_null_device[] =
  *
  *  Note that file permissions are quite different from the _mode_
  *  of a file stream (\File object).
- *  See IO@Modes.
+ *  See IO.
```
peterzhu2118, line 7138: Is there a specific part we can link to rather than the generic IO page?

io.c
```diff
@@ -11957,7 +11957,7 @@ io_s_foreach(VALUE v)
  *
  *  - {Open Options}[rdoc-ref:IO@Open+Options].
  *  - {Encoding options}[rdoc-ref:encodings.rdoc@Encoding+Options].
- *  - {Line Options}[rdoc-ref:IO@Line+Options].
+ *  - {Line IO}[rdoc-ref:IO@Line+IO].
```
peterzhu2118, line 11960: The `Line IO` section doesn't seem to talk about keyword arguments, the `sep` and `limit` are both positional arguments. Can we even pass these as keyword arguments?

BurdetteLamar: Out of scope for this PR?

peterzhu2118: The Line IO section (removed in ded895baa96cb9622fcd44d308fb48341565605b) looked like this:

```
 *  === Line Options
 *
 *  A number of \IO methods accept optional keyword arguments
 *  that determine how lines in a stream are to be treated:
 *
 *  - +:chomp+: If +true+, line separators are omitted; default is +false+.
```

This is now linking to a different section that is not about keyword arguments.

BurdetteLamar: Should this simply be removed?

peterzhu2118: Well we should keep the part about the `chomp` keyword (maybe just paste it inline), since that's actually useful information.

BurdetteLamar: Done.

io.c
```diff
@@ -12055,7 +12054,7 @@ io_s_readlines(VALUE v)
...
```
peterzhu2118, line 12057: Ditto here.

BurdetteLamar: Fixed.  This is _really_ careless.  The report showed both breaks.  Thanks for being a lert; we need more lerts.

peterzhu2118:
@BurdetteLamar can you respond to the comment above about linking to the generic `IO` page?

file.c
BurdetteLamar: Sorry, I thought I'd responded here.  I looked, but didn't see as suitable target.

peterzhu2118: Can you fix this one in a separate PR then? (Or alternatively delete this completely). I don't think asking the reader to read all of the docs of IO is particularly helpful.

BurdetteLamar: Sure.  I see the problem, but not the solution.

BurdetteLamar:
To be revisited.

PR #7872, opened 2023-05-29
Unify error messages of rb_num2ulong and rb_num2ull

peterzhu2118:
The error messages were slightly different due, which causes different behaviour on 32-bit and 64-bit systems.

### DIFF
--- a/spec/ruby/core/time/new_spec.rb
+++ b/spec/ruby/core/time/new_spec.rb
@@ -523,9 +523,16 @@ describe "Time.new with a timezone argument" do
       end
 
       it "raise TypeError is can't convert precision keyword argument into Integer" do
+        error_msg =
+          if "3.3" <= RUBY_VERSION
+            "no implicit conversion of String into Integer"
+          else
+            "no implicit conversion from string"
+          end
+
         -> {
           Time.new("2021-12-25 00:00:00.123456789876 +09:00", precision: "")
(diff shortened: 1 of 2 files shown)


spec/ruby/core/time/new_spec.rb
```diff
@@ -523,9 +523,16 @@ def obj.to_int; 3; end
       end
 
       it "raise TypeError is can't convert precision keyword argument into Integer" do
+        error_msg =
+          if "3.3" <= RUBY_VERSION
+            "no implicit conversion of String into Integer"
+          else
+            "no implicit conversion from string"
+          end
```
eregon, line 531: I think a regexp would be simpler here to just allow both

eregon (approved):
Thank you!

spec/ruby/core/time/new_spec.rb
peterzhu2118: You're right, but regex would allow a regression and making it explicit means that we can remove this code once we drop support for Ruby 3.2 in the future.

eregon: Right. I'll push a commit to use a `ruby_version_is "3.3"` then to make it clear it changed on purpose.
It's definitely good to be consistent in error messages, otherwise it's really hard for alternative Rubies to match them, and it makes harder for end users to test error cases.

PR #7875, opened 2023-05-31
Mark the classpath field for T_ICLASS.

wks:
The classpath field was not marked if the object is a T_ICLASS.  As a result, the classpath field may contain dangling pointer.  This commit make sure the field is marked during GC.

An alternative solution is reverting the commit https://github.com/ruby/ruby/commit/081cc4eb283cb01ddffb364397e5175dbfacab66 so that the `classpath` field is never used for `T_ICLASS`.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -7211,6 +7211,7 @@ gc_mark_children(rb_objspace_t *objspace, VALUE obj)
         }
         mark_m_tbl(objspace, RCLASS_CALLABLE_M_TBL(obj));
         cc_table_mark(objspace, obj);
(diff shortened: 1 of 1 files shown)


byroot:
Your patch looks correct to me, but are we sure `classpath` is always initialized on `I_CLASS` ?

I fear we may be marking garbage in most cases.

peterzhu2118:
It won't be marking garbage as the `RCLASS_EXT` is zero'd out during allocation so it will be `Qfalse` when not set.

PR #7876, opened 2023-05-31
Skip a flaky test on macOS

KJTsanaktsidis:
I believe this is flaky because the operating system is buggy.

https://bugs.ruby-lang.org/issues/19700 - I put the details about my thinking there.

### DIFF
--- a/test/ruby/test_process.rb
+++ b/test/ruby/test_process.rb
@@ -691,6 +691,7 @@ class TestProcess < Test::Unit::TestCase
   end unless windows? # does not support fifo
 
   def test_execopts_redirect_open_fifo_interrupt_print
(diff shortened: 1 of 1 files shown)


hsbt:
@KJTsanaktsidis Does this change still need? My Ventura and Sonoma are always passed this test.

KJTsanaktsidis:
@hsbt - yes, last I checked a couple weeks ago this was still broken in macOS. It's only flaky under m1 macs afaict - if you have an intel mac you won't see this failure. I've (I guess unsurprisingly) not had any response from apple about my bugreport to them...

hsbt:
I didn't face this with my Apple Silicon machine and our CI that are Intel/Apple Silicon.

PR #7878, opened 2023-05-31
No need to define "BISON" on extconf.rb

yui-knk:
"BISON" is defined in "ext/ripper/depend".


hsbt (approved):

yui-knk:
Thanks for the review.

PR #7880, opened 2023-06-01
Hide the usage of `rb_io_t` where possible.

ioquatix:
This is purely the additive parts of the previous PR, which caused downstream compatibility issues, https://github.com/ruby/ruby/pull/6511

This introduces the minimum amount of public interface for avoiding accessing `rb_io_t` internals in most cases.


spec/ruby/optional/capi/ext/io_spec.c
```diff
@@ -28,9 +28,13 @@ static int set_non_blocking(int fd) {
 }
 
 static int io_spec_get_fd(VALUE io) {
+#ifdef RUBY_VERSION_IS_3_3
```
eregon, line 31: Should be 3_1 since it exists since 3.1, I changed it in 957b6a491fe6a9a2df467f83a37e7c34861f995c

ioquatix: Yes, but this will break truffle ruby right? That's why I changed it to 3.3 to give you time to implement it.. unless you want to backport it? IIUC.

PR #7882, opened 2023-06-01
Revert "Revert https://github.com/ruby/io-console/pull/43"

ioquatix:
This restores the changes in `io-console` without breaking backwards compatibility.


nobu:
> This restores the changes in `io-console` without breaking backwards compatibility.

Could you add this description to the commit log?

ioquatix:
@nobu sure thing.

PR #7884, opened 2023-06-01
Use a real Ruby mutex in rb_io_close_wait_list

KJTsanaktsidis:
Because a thread calling IO#close now blocks in a native condvar wait, it's possible for there to be _no_ threads left to actually handle incoming signals/ubf calls/etc.

This manifested as failing tests on Solaris 10 (SPARC), because:

* One thread called IO#close, which sent a SIGVTALRM to the other thread to interrupt it, and then waited on the condvar to be notified that the reading thread was done.
* One thread was calling IO#read, but it hadn't yet reached the actual call to select(2) when the SIGVTALRM arrived, so it never unblocked itself.

This results in a deadlock.

The fix is to use a real Ruby mutex for the close lock; that way, the closing thread goes into sigwait-sleep and can keep trying to interrupt the select(2) thread.

See the discussion in: https://github.com/ruby/ruby/pull/7865/

----

In that linked PR above, I also said "nogvl_wait_for should wait not only on the actual FD that was requested, but also on the sigwait FD (if it's available).". I think I want to do that _as well_. But this patch seems sufficient to solve the deadlock I introduced, and the tests pass on the Solaris VM i'm using.

CC @mame  and @ioquatix

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -1663,6 +1663,27 @@ rb_thread_call_without_gvl(void *(*func)(void *data), void *data1,
     return rb_nogvl(func, data1, ubf, data2, 0);
 }
 
+static void
+rb_thread_io_wake_pending_closer(struct waiting_fd *wfd)
+{
+    bool has_waiter = wfd->busy && RB_TEST(wfd->busy->wakeup_mutex);
+    if (has_waiter) {
+        rb_mutex_lock(wfd->busy->wakeup_mutex);
+    }
+
+    /* Needs to be protected with RB_VM_LOCK because we don't know if
+       wfd is on the global list of pending FD ops or if it's on a
+       struct rb_io_close_wait_list close-waiter. */
+    RB_VM_LOCK_ENTER();
+    ccan_list_del(&wfd->wfd_node);
+    RB_VM_LOCK_LEAVE();
+
+    if (has_waiter) {
+        rb_thread_wakeup(wfd->busy->closing_thread);
+        rb_mutex_unlock(wfd->busy->wakeup_mutex);
+    }
+}
+
 VALUE
 rb_thread_io_blocking_region(rb_blocking_function_t *func, void *data1, int fd)
 {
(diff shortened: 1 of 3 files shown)


thread.c
```diff
@@ -2494,34 +2505,39 @@ rb_notify_fd_close(int fd, struct rb_io_close_wait_list *busy)
...
-    has_any = !ccan_list_empty(&busy->list);
+
+    has_any = !ccan_list_empty(&busy->pending_fd_users);
+    busy->closing_thread = rb_thread_current();
+    wakeup_mutex = Qnil;
     if (has_any) {
-        rb_native_mutex_initialize(&busy->mu);
-        rb_native_cond_initialize(&busy->cv);
+        wakeup_mutex = rb_mutex_new();
+        RBASIC_CLEAR_CLASS(wakeup_mutex); /* hide from ObjectSpace */
```
KJTsanaktsidis, line 2514: Is it OK to use a Ruby mutex like this inside `thread.c`/`io.c`? It works, and does what I want, but I don't know if this kind of thing is considered wrong for some reason?

ioquatix: I think it's okay, and I'd even say you don't need to worry about exposing it to object space.

ioquatix:
Thanks for your work on this difficult issue.

ko1:
It doesn't work on Ractors so I'll change them later.

KJTsanaktsidis:
Is the issue that two Ractors could be doing IO to the same fd (e.g. by both creating an IO with `IO.for_fd`), and the mutex would wind up being shared between Ractors unsafely? Would the fix for that be to just protect the `rb_io_close_wait_list` with the `RB_VM_LOCK_ENTER()` lock?

luanzeba:
Thanks for this change. We were seeing deadlocks when running cc698c6cc29b7e838932ce29745d2376c094671d but this change fixed it.

PR #7890, opened 2023-06-02
singleton cleanup

dpep:
I was trying to understand the `Singleton` code (neat!) and saw a few ways we could simplify.  If we roll `Singleton.__init__` into `included`, it means we don't have to expose that helper method to the public and creates a single entrypoint (`include Singleton`).  We could also simplify the `instance` method to have only a single return statement, at the end.

Would this help?

### DIFF
--- a/lib/singleton.rb
+++ b/lib/singleton.rb
@@ -112,20 +112,19 @@ module Singleton
   module SingletonClassMethods # :nodoc:
 
     def clone # :nodoc:
-      Singleton.__init__(super)
+      super.include(Singleton)
     end
 
     # By default calls instance(). Override to retain singleton state.
-    def _load(str)
+    def _load(*)
       instance
     end
 
     def instance # :nodoc:
-      return @singleton__instance__ if @singleton__instance__
-      @singleton__mutex__.synchronize {
-        return @singleton__instance__ if @singleton__instance__
-        @singleton__instance__ = new()
-      }
+      unless @singleton__instance__
+        @singleton__mutex__.synchronize { @singleton__instance__ ||= new }
+      end
+
       @singleton__instance__
     end
 
@@ -133,19 +132,11 @@ module Singleton
 
     def inherited(sub_klass)
       super
-      Singleton.__init__(sub_klass)
+      sub_klass.include(Singleton)
     end
   end
 
   class << Singleton # :nodoc:
-    def __init__(klass) # :nodoc:
-      klass.instance_eval {
-        @singleton__instance__ = nil
-        @singleton__mutex__ = Thread::Mutex.new
-      }
-      klass
-    end
-
     private
 
     # extending an object with Singleton is a bad idea
@@ -156,14 +147,16 @@ module Singleton
       unless mod.instance_of?(Class)
         raise TypeError, "Inclusion of the OO-Singleton module in module #{mod}"
       end
+
       super
     end
 
     def included(klass)
       super
+
       klass.private_class_method :new, :allocate
       klass.extend SingletonClassMethods
-      Singleton.__init__(klass)
+      klass.instance_eval { @singleton__mutex__ = Thread::Mutex.new }
     end
   end
 


ioquatix:
This looks reasonable to me.

Do you mind creating a feature request on `bugs.ruby-lang.org`?

cc @jeremyevans maybe a nice case for your "feature request triage process"?

jeremyevans:
Singleton is a standard library with its own upstream, so this pull request should be closed and a separate pull request should be filed upstream https://github.com/ruby/singleton/pulls

@ioquatix Feature request triaging is aimed exclusively at older feature requests.  New feature requests are ignored completely so that anyone has a chance to comment and consider them.  Also, this seems not like a feature request at all, as it attempts to refactor without changing behavior.

lib/singleton.rb
```diff
@@ -112,40 +112,31 @@ def _dump(depth = -1)
   module SingletonClassMethods # :nodoc:
 
     def clone # :nodoc:
-      Singleton.__init__(super)
+      super.include(Singleton)
     end
 
     # By default calls instance(). Override to retain singleton state.
-    def _load(str)
+    def _load(*)
```
jeremyevans, line 119: This is not a good change, as it changes the arity, and results in an unnecessary array allocation per call.

lib/singleton.rb
```diff
@@ -112,40 +112,31 @@ def _dump(depth = -1)
...
       instance
     end
 
     def instance # :nodoc:
-      return @singleton__instance__ if @singleton__instance__
-      @singleton__mutex__.synchronize {
-        return @singleton__instance__ if @singleton__instance__
-        @singleton__instance__ = new()
-      }
+      unless @singleton__instance__
```
jeremyevans, line 124: If you are going to refactor:

```ruby
  @singleton__instance__ || @singleton__mutex__.synchronize { @singleton__instance__ ||= new }
```

lib/singleton.rb
```diff
@@ -112,40 +112,31 @@ def _dump(depth = -1)
   module SingletonClassMethods # :nodoc:
 
     def clone # :nodoc:
-      Singleton.__init__(super)
+      super.include(Singleton)
```
jeremyevans, line 115: I'm guessing this does not work correctly, since `@singleton__instance__` is not unset for the clone.  At the least, you would want to remove the `@singleton__instance__` instance variable from the cloned object before the method returned, if the instance variable exists.

lib/singleton.rb
```diff
@@ -112,40 +112,31 @@ def _dump(depth = -1)
...
+
       @singleton__instance__
     end
 
     private
 
     def inherited(sub_klass)
       super
-      Singleton.__init__(sub_klass)
+      sub_klass.include(Singleton)
```
jeremyevans, line 135: This isn't safe if `super` here copies instance variables from the superclass to the subclass, for similar reasons as for `clone`.

lib/singleton.rb
```diff
@@ -156,14 +147,16 @@ def append_features(mod)
...
       super
     end
 
     def included(klass)
       super
+
       klass.private_class_method :new, :allocate
       klass.extend SingletonClassMethods
-      Singleton.__init__(klass)
+      klass.instance_eval { @singleton__mutex__ = Thread::Mutex.new }
```
jeremyevans, line 159: `instance_variable_set` may be preferable to `instance_eval` here, since only a single instance variable is set.

jeremyevans (commented):
Here are some review notes in case this is sent upstream.

ioquatix:
Thanks @jeremyevans - great feedback.

I don't have any particular stake in the game, are you suggesting that as this is neither a feature or bug it doesn't require an issue on <bugs.ruby-lang.org>?

jeremyevans:
> Thanks @jeremyevans - great feedback.
> 
> I don't have any particular stake in the game, are you suggesting that as this is neither a feature or bug it doesn't require an issue on https://bugs.ruby-lang.org?

Correct, I don't think pure refactors need to have an issue created on redmine.  These days, bugs and features in standard libraries with upstream github repos are encouraged to be filed as upstream pull requests, with redmine only used for core Ruby issues. Matz is the maintainer of singleton, so a pull request to upstream should be sufficient for him to consider it.

lib/singleton.rb
dpep: good catch!  it worked only because the tests cloned before calling `.instance` (so `@singleton_instance__` was nil during cloning).  I added a test to cover this

lib/singleton.rb
dpep: even better!

dpep:
thanks for the feedback gents!  ported upstream to https://github.com/ruby/singleton/pull/7

PR #7894, opened 2023-06-02
Introduce `enum rb_io_mode`.

ioquatix:
It would be nice if auto-complete worked correctly, and we got a little more type-checking from the compiler.

- [x] Upstream changes to `ruby/stringio` first.

### DIFF
--- a/include/ruby/io.h
+++ b/include/ruby/io.h
@@ -137,129 +137,6 @@ struct rb_io_encoding {
     VALUE ecopts;
 };
 
-#ifndef HAVE_RB_IO_T
-#define HAVE_RB_IO_T 1
-/** Ruby's IO, metadata and buffers. */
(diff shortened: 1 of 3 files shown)


nobu:
Why need to add `enum`?
If it is a `typedef`, it would work just by adding a fallback definition.

```C
#ifndef HAVE_TYPE_RB_IO_MODE
typedef int rb_io_mode;
#endif
```
Or
```ruby
have_type("rb_io_mode") or $defs << "-Drb_io_mode=int"
```

ioquatix:
Ah yes, that's a great idea, thanks.

ioquatix:
Upstream changes to `ruby/stringio`: https://github.com/ruby/stringio/pull/129

nobu (approved):

PR #7895, opened 2023-06-02
Stabilize test_latest_gc_info_need_major_by

casperisfine:
Fix:
```
    1) Failure:
  TestGc#test_latest_gc_info_need_major_by [/home/runner/work/ruby/ruby/src/test/ruby/test_gc.rb:266]:
  <nil> expected to not be nil.
```

`GC.stat(:major_gc_count)` can be bumped while `GC.latest_gc_info(:need_major_by)` is still nil.

@ioquatix @peterzhu2118

### DIFF
--- a/test/ruby/test_gc.rb
+++ b/test/ruby/test_gc.rb
@@ -259,7 +259,7 @@ class TestGc < Test::Unit::TestCase
     # allocate objects until need_major_by is set or major GC happens
     major_count = GC.stat(:major_gc_count)
     objects = []
-    while GC.stat(:major_gc_count) == major_count && GC.latest_gc_info(:need_major_by).nil?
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):
We should keep an eye on CI after merging this. I worry that there could be cases where we don't pause during marking and the while loop runs forever.

ioquatix:
Thanks everyone for your efforts!

@peterzhu2118 is there some way we could improve the test to avoid that case? What about putting an upper bound on the loop, e.g. 30 seconds or X iterations?

peterzhu2118:
I don't think there is a way to improve it. IIRC the test runner has a timeout, so we don't need to add one.

PR #7910, opened 2023-06-05
YJIT: Avoid using _C_ARG_OPNDS for InsnOut on x86_64

k0kubun:
## Issue
On x86_64, there's a potential bug in our register allocator for `InsnOut`. The following code is a synthetic example.

```rs
let rax = asm.load(Opnd::UImm(1));
let rcx = asm.load(Opnd::UImm(2));
let rdx = asm.load(Opnd::UImm(3));
asm.ccall(0 as _, vec![
    rax, // mov rdi, rax (ok)
    rcx, // mov rsi, rcx (ok)
    rcx, // mov rdx, rcx (NG)
    rdx, // mov rcx, rdx (this becomes 2, not 3)
]);
```

Currently, both `get_alloc_regs()` (`InsnOut` registers) and `_C_ARG_OPNDS` have `rdx` (3rd argument) and `rcx` (4th argument). If a 3rd or 4th argument is set and then an `InsnOut` that was using the same register is set to another argument after that, the argument will have an unexpected value.

We're probably not hitting this today, but I'd like to sort this out now since I'm working on something related to this.

## Changes
This PR removes the overlap between `get_alloc_regs()` and `_C_ARG_OPNDS` on x86_64. To do so, I needed to update `SCRATCH0` and `TEMP_REGS` as well. They do overlap with `_C_ARG_OPNDS`, but `SCRATCH0` is not used when C arguments are set up, and `TEMP_REGS` are all spilled before `asm.ccall` for now.

arm64 doesn't have this problem from the beginning.

### DIFF
--- a/yjit/src/backend/x86_64/mod.rs
+++ b/yjit/src/backend/x86_64/mod.rs
@@ -89,19 +89,15 @@ impl Assembler
     // A special scratch register for intermediate processing.
     // Note: right now this is only used by LeaLabel because label_ref accepts
     // a closure and we don't want it to have to capture anything.
-    const SCRATCH0: X86Opnd = X86Opnd::Reg(R11_REG);
+    const SCRATCH0: X86Opnd = X86Opnd::Reg(R9_REG);
 
-    /// List of registers that can be used for stack temps.
-    pub const TEMP_REGS: [Reg; 5] = [RSI_REG, RDI_REG, R8_REG, R9_REG, R10_REG];
+    /// List of registers that can be used for Opnd::Stack.
+    pub const TEMP_REGS: [Reg; 5] = [RDI_REG, RSI_REG, RDX_REG, RCX_REG, R8_REG];
 
-    /// Get the list of registers from which we can allocate on this platform
-    pub fn get_alloc_regs() -> Vec<Reg>
-    {
-        vec![
-            RAX_REG,
-            RCX_REG,
-            RDX_REG,
-        ]
+    /// Get the list of registers that can be used for Opnd::InsnOut.
+    pub fn get_alloc_regs() -> Vec<Reg> {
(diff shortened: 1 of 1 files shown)


yjit/src/backend/x86_64/mod.rs
```diff
@@ -89,19 +89,15 @@ impl Assembler
...
-        vec![
-            RAX_REG,
-            RCX_REG,
-            RDX_REG,
-        ]
+    /// Get the list of registers that can be used for Opnd::InsnOut.
+    pub fn get_alloc_regs() -> Vec<Reg> {
+        // _C_ARG_OPNDS registers shouldn't be used to avoid conflict on C calls
+        vec![RAX_REG, R10_REG, R11_REG]
     }
```
maximecb, line 101: I don't want to make your life harder, but... 😅 

The problem with this is that R10 and R11 require the REX prefix to encode on x86-64, which forces all instructions using them to be one byte longer. Ideally we want to use registers with register number <= 7 for temp regs as much as possible.

Ideally, the best way to fix this is to have a little algorithm to generate the sequence of moves for method arguments. I implemented something like this in Higgs. You only need one temporary register to break cycles: https://github.com/higgsjs/Higgs/blob/1b4e5d85dd192592882fa2d30b839ed843be402f/source/jit/moves.d#L86

We could merge this as a temporary solution, but ideally we should implement a more optimal solution. What do you think?

k0kubun: Thank you for the feedback. I haven't paid enough attention to REX.B with regard to the difference between R8-15 and the others. With that in mind, I agree that we should try to use as many registers without the prefix as possible.

I was trying to make my implementation simpler by compromising some edge cases (if you rule out InsnOut, you can know the exact registers in x86_split), but I'll experiment with an optimal solution first.

k0kubun:
I'll revisit this in a different PR.

PR #7911, opened 2023-06-05
YJIT: Avoid identity-based known-class guards for IO objects

XrXr:
`IO#reopen` is very special in that it is able to change the class and
singleton class of IO instances. In its presence, it is not correct to
assume that IO instances has a stable class/singleton class and guard
by comparing identity.

---

Discovered after looking at <https://bugs.ruby-lang.org/issues/19712>. Thanks Peter! 

- [ ] make a ticket for this for backport


k0kubun (approved):
It's annoying that `IO#reopen` does stuff like this. I wonder if it could cause other problems when it changes the class of a receiver, but I can't think of any.

maximecb:
> It's annoying that `IO#reopen` does stuff like this. I wonder if it could cause other problems when it changes the class of a receiver, but I can't think of any.

I hate this behavior as well 😅 Should not be in the language.

PR #7913, opened 2023-06-06
YJIT: Reserve C argument registers on register allocation

k0kubun:
This is a revised attempt to fix the issue described in https://github.com/ruby/ruby/pull/7910.

We could reorder instructions and/or use an extra register to support such scenarios, but it seems like we don't need it for the current codegen.rs. So this PR just asserts that we don't have that problem by taking C argument registers from the pool when used.

---

Note: w.r.t. https://github.com/ruby/ruby/pull/7910#discussion_r1219762144, I still intend to implement an algorithm to generate optimal code for C arguments, but it'll be used only for stack temps. I want to reorder instructions in `x86_split` for it, but `InsnOut` is not lowered until `alloc_regs`.

### DIFF
--- a/yjit/src/backend/arm64/mod.rs
+++ b/yjit/src/backend/arm64/mod.rs
@@ -74,9 +74,10 @@ impl From<Opnd> for A64Opnd {
             Opnd::Mem(Mem { base: MemBase::InsnOut(_), .. }) => {
                 panic!("attempted to lower an Opnd::Mem with a MemBase::InsnOut base")
             },
-            Opnd::InsnOut { .. } => panic!("attempted to lower an Opnd::InsnOut"),
-            Opnd::Value(_) => panic!("attempted to lower an Opnd::Value"),
-            Opnd::Stack { .. } => panic!("attempted to lower an Opnd::Stack"),
+            Opnd::CArg { .. } |
+            Opnd::InsnOut { .. } |
+            Opnd::Value(_) |
+            Opnd::Stack { .. } => panic!("attempted to lower {:?}", opnd),
             Opnd::None => panic!(
                 "Attempted to lower an Opnd::None. This often happens when an out operand was not allocated for an instruction because the output of the instruction was not used. Please ensure you are using the output."
             ),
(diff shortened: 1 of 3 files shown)


k0kubun:
I'll make it a draft for now to fix issues in arm64.

k0kubun:
I think we're discussing different difficulties. I'm not questioning the feasibility of breaking a cycle with a single scratch register, for example.

What's a little complicated here is the architecture of our passes. As of `x86_split`, the reordering target is a vector of operands in `Insn::CCall`, which is easy to deal with. On the other hand, in `alloc_regs`, it's `Mov` instructions in the iterator. You have to first figure out which `Mov` instructions need to be moved (I think it needs a new IR) and then pop multiple such `Mov` instructions from the iterator before reordering them. You still have to assign registers before reordering them, but popping instructions first would make it harder to reuse the current logic that is designed for iterating over instructions one by one.

So if we were to reorder both InsnOut and stack temp registers together, I would add another pass after `alloc_regs` because doing it inside `alloc_regs` seems too complicated. But then it would slow down compilation. I don't think it's a good idea to slow down compilation for that reason when you don't need to reorder `InsnOut` registers in the first place. WDYT?

maximecb:
I think we should avoid adding another pass as you said.

Another possibility. Not sure if this would work but... Maybe we could generate the moves to break cycles in the last pass to emit code for x86? Assuming we have access to a scratch register. Then we'd have registers allocated for everything and we could generate moves for the call arguments. Would that make sense or do you think it would be impractical?

k0kubun:
Thinking about my initial design further, it turns out it actually doesn't work as I hoped 😅 Yes, we _currently_ don't have a cycle in `InsnOut` registers, however, those registers could still be used as a mov destination of stack temp registers. So you can't just handle them in different passes separately. Sorry for confusing you with a broken idea. 

> Maybe we could generate the moves to break cycles in the last pass to emit code for x86? Assuming we have access to a scratch register. Then we'd have registers allocated for everything and we could generate moves for the call arguments. Would that make sense or do you think it would be impractical?

The last half of the second paragraph in https://github.com/ruby/ruby/pull/7913#issuecomment-1581170484 explains what it would look like. `alloc_regs` might need some refactoring to support allocating registers for multiple peeked instructions and modifying `live_ranges` there, but it does seem feasible. I'll work on it instead.

k0kubun:
I managed to make it work inside `alloc_regs` https://github.com/ruby/ruby/pull/7918.

PR #7915, opened 2023-06-07
Deprecate Kernel#open and IO support for subprocess creation or forking

flavorjones:
[Feature #19630] / https://bugs.ruby-lang.org/issues/19630

Deprecate Kernel#open and IO support for subprocess creation and
forking. This also deprecates subprocess creation and forking in

- Kernel#open
- URI.open
- IO.binread
- IO.foreach
- IO.readlines
- IO.read
- IO.write

Example warnings:

```
IO.read("|echo foo")
> warning: IO process creation with a leading '|' is deprecated and will be removed in Ruby 4.0; use IO.popen instead


open("|echo foo")
> warning: Calling Kernel#open with a leading '|' is deprecated and will be removed in Ruby 4.0; use IO.popen instead

require 'open-uri'
URI.open("|echo foo")
> ruby/3.3.0+0/open-uri.rb:31: warning: Calling Kernel#open with a leading '|' is deprecated and will be removed in Ruby 4.0; use IO.popen instead
```

### DIFF
--- a/io.c
+++ b/io.c
@@ -8109,20 +8109,10 @@ check_pipe_command(VALUE filename_or_command)
  *    open(path, mode = 'r', perm = 0666, **opts)             -> io or nil
  *    open(path, mode = 'r', perm = 0666, **opts) {|io| ... } -> obj
  *
- *  Creates an IO object connected to the given stream, file, or subprocess.
+ *  Creates an IO object connected to the given file.
  *
- *  Required string argument +path+ determines which of the following occurs:
- *
- *  - The file at the specified +path+ is opened.
- *  - The process forks.
- *  - A subprocess is created.
- *
- *  Each of these is detailed below.
- *
- *  <b>File Opened</b>
-
- *  If +path+ does _not_ start with a pipe character (<tt>'|'</tt>),
- *  a file stream is opened with <tt>File.open(path, mode, perm, **opts)</tt>.
+ *  This method has potential security vulnerabilities if called with untrusted input;
+ *  see {Command Injection}[rdoc-ref:command_injection.rdoc].
  *
  *  With no block given, file stream is returned:
  *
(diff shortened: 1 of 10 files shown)


io.c
```diff
@@ -8222,6 +8222,7 @@ rb_f_open(int argc, VALUE *argv, VALUE _)
             else {
                 VALUE cmd = check_pipe_command(tmp);
                 if (!NIL_P(cmd)) {
+                    rb_warn_deprecated_to_remove("4.0", "Calling Kernel#open with a leading '|'", "IO.popen");
```
nobu, line 8225: The bellow will cause an error when 4.0.
```suggestion
                    rb_warn_deprecated_to_remove_at(4.0, "Calling Kernel#open with a leading '|'", "IO.popen");
```

io.c
```diff
@@ -8139,67 +8129,6 @@ check_pipe_command(VALUE filename_or_command)
...
- *      else
- *        $stderr.puts "In child, pid is #{$$}."
- *      end
- *    end
- *
- *  Output:
- *
- *    In parent, child pid is 28427.
- *    In child, pid is 28427.
- *
```
nobu, line 8175: The vulnerability is for **command** injections, but should be unrelated to `fork`.

nobu (changes requested):

io.c
matz: You don't have to keep `|=`.

spec/ruby/core/io/binread_spec.rb
```diff
@@ -44,4 +44,14 @@
   it "raises an Errno::EINVAL when not passed a valid offset" do
     -> { IO.binread @fname, 0, -1  }.should raise_error(Errno::EINVAL)
   end
+
+  ruby_version_is "3.3" do
+    it "warns about deprection given a path with a pipe" do
+      cmd = "|date"
+      platform_is(:windows) { cmd = "|date /t" }
```
nobu, line 51: `echo`  command should work even on Windows.
```suggestion
      cmd = "|echo ok"
```

flavorjones: OK, I've updated the PR to use `|echo ok` on all platforms for all the new specs.

nobu (approved):

matz (approved):

PR #7917, opened 2023-06-08
objspace_dump.c: don't skip object's class in references

oleg-vinted:
**Problem:** when fixing memory leaks, you sometimes need to analyze heap dumps to find a reference chain from a root object to the object that's potentially leaked. However, using just the `references` array as the source of objects that are referenced (and thus retained) is not enough because sometimes object's class is also referenced but it is not put into the `references` array.

**Solution:** do not skip references to object's class in the `references` array. This would simplify heap analysis tools a lot because the tools would no longer need to make guesses if the `class` key needs to be used or not.

Ref https://github.com/zombocom/heapy/pull/33.

### DIFF
--- a/ext/objspace/objspace_dump.c
+++ b/ext/objspace/objspace_dump.c
@@ -313,9 +313,6 @@ reachable_object_i(VALUE ref, void *data)
 {
     struct dump_config *dc = (struct dump_config *)data;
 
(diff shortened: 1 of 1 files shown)


oleg-vinted:
Closing as stale

PR #7918, opened 2023-06-08
YJIT: Break register cycles for C arguments

k0kubun:
The InsnOut register allocator `alloc_regs()` has a bug that C argument registers and InsnOut registers could substitute each other, leading to wrong values set to C arguments. See `test_c_arg_cycle` added in this PR.

For each C call, this PR buffers C arguments and modify operations on such arguments. To maintain fast compilation, we do so inside `alloc_regs()` instead of adding another pass after that.

See also: https://github.com/ruby/ruby/pull/7910, https://github.com/ruby/ruby/pull/7913


k0kubun:
See https://github.com/ruby/ruby/pull/7920 for speedups enabled by this PR later.

PR #7920, opened 2023-06-08
YJIT: Use registers to pass stack temps to C calls

k0kubun:
## Background

The current `asm.ccall` expects `asm.spill_temps()` to be called outside it. It was necessary when some arguments were popped before a C call and using registers for them would cause stack temp registers and argument registers to depend on each other.

## Changes

https://github.com/ruby/ruby/pull/7918 can automatically remove such issues from C arguments, so this PR changes `asm.ccall` to lazily spill stack temps that are still on stack and avoid spilling stack temps that are already popped. This not only makes the code cleaner but also optimizes the following methods by removing memory reads and writes:

* `Array#[]`
* `Module#===`
* `Integer#/`
* `Integer#[]`
* `Integer#%`
* `String#bytesize`
* `Struct#[]=`

Many other methods use `jit_prepare_routine_call` and it's generally not safe to skip spilling stack temps in such cases because arguments also need to be on the VM stack for GC. We could still let them use registers in arguments even after spilling them in some cases, but it's out of scope in this PR.

### Headline
liquid-compile and mail reliably speed up by 2%.

```
before: ruby 3.3.0dev (2023-06-07T01:10:08Z master 46583f7dcf) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-06-08T23:46:17Z yjit-temp-ccall ccdd408693) +YJIT [x86_64-linux]

activerecord    32.3         5.4         32.0        5.8         1.01           1.01
chunky_png      420.2        0.7         417.8       0.5         1.00           1.01
erubi_rails     11.5         26.3        11.5        28.4        0.95           1.00
hexapdf         1358.5       2.0         1365.5      0.9         0.99           0.99
liquid-c        41.2         1.2         40.6        1.4         0.98           1.01
liquid-compile  37.9         2.6         37.1        3.0         0.95           1.02
liquid-render   73.4         0.6         73.2        0.7         0.99           1.00
mail            90.5         0.3         88.4        0.2         0.98           1.02
psych-load      1232.1       0.3         1221.6      0.2         1.01           1.01
railsbench      1243.6       1.4         1242.8      1.4         0.99           1.00
ruby-lsp        38.6         19.7        38.3        21.0        0.98           1.01
sequel          52.0         1.9         51.8        2.3         1.00           1.00
```

### Other
A number of them show a speedup. fannkuchredux is 7% faster, and nbody is 5% faster.

```
before: ruby 3.3.0dev (2023-06-07T01:10:08Z master 46583f7dcf) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-06-08T23:46:17Z yjit-temp-ccall ccdd408693) +YJIT [x86_64-linux]

-------------  -----------  ----------  ----------  ----------  -------------  ------------
bench          before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
binarytrees    137.0        1.4         137.3       1.4         1.02           1.00
erubi          161.0        0.3         156.8       0.1         1.02           1.03
etanni         261.3        0.6         255.4       0.3         1.02           1.02
fannkuchredux  472.0        0.3         439.9       0.3         1.01           1.07
lee            611.3        1.1         607.7       1.2         1.01           1.01
nbody          45.7         0.6         43.5        0.3         1.04           1.05
optcarrot      1400.7       0.8         1390.9      0.9         1.02           1.01
ruby-json      2428.6       0.1         2499.8      0.2         0.97           0.97
rubykon        4409.4       0.3         4265.8      0.3         1.03           1.03
-------------  -----------  ----------  ----------  ----------  -------------  ------------
```

### DIFF
--- a/yjit/src/backend/ir.rs
+++ b/yjit/src/backend/ir.rs
@@ -1646,19 +1646,25 @@ impl Assembler {
     }
 
     pub fn ccall(&mut self, fptr: *const u8, opnds: Vec<Opnd>) -> Opnd {
-        // Save what registers are available for the C arguments.
-        let reg_temps = self.ctx.get_reg_temps();
+        let old_temps = self.ctx.get_reg_temps(); // with registers
         // Spill stack temp registers since they are caller-saved registers.
         // Note that this doesn't spill stack temps that are already popped
         // but may still be used in the C arguments.
         self.spill_temps();
+        let new_temps = self.ctx.get_reg_temps(); // all spilled
 
+        // Temporarily manipulate RegTemps so that we can use registers
+        // to pass stack operands that are already spilled above.
+        self.ctx.set_reg_temps(old_temps);
+
+        // Call a C function
         let out = self.next_opnd_out(Opnd::match_num_bits(&opnds));
-        // For C arguments, use stack temp registers that are still alive.
-        // After the C call that may clobber them, use spilled stack temps.
-        self.ctx.set_reg_temps(reg_temps);
         self.push_insn(Insn::CCall { fptr, opnds, out });
-        self.ctx.set_reg_temps(RegTemps::default());
+
+        // Registers in old_temps may be clobbered by the above C call,
+        // so rollback the manipulated RegTemps to a spilled version.
+        self.ctx.set_reg_temps(new_temps);
+
         out
     }
 
--- a/yjit/src/backend/ir.rs
+++ b/yjit/src/backend/ir.rs
@@ -1646,9 +1646,19 @@ impl Assembler {
     }
 
     pub fn ccall(&mut self, fptr: *const u8, opnds: Vec<Opnd>) -> Opnd {
-        assert_eq!(self.ctx.get_reg_temps(), RegTemps::default(), "temps must be spilled before ccall");
+        // Save what registers are available for the C arguments.
+        let reg_temps = self.ctx.get_reg_temps();
+        // Spill stack temp registers since they are caller-saved registers.
+        // Note that this doesn't spill stack temps that are already popped
+        // but may still be used in the C arguments.
+        self.spill_temps();
+
         let out = self.next_opnd_out(Opnd::match_num_bits(&opnds));
+        // For C arguments, use stack temp registers that are still alive.
+        // After the C call that may clobber them, use spilled stack temps.
+        self.ctx.set_reg_temps(reg_temps);
         self.push_insn(Insn::CCall { fptr, opnds, out });
+        self.ctx.set_reg_temps(RegTemps::default());
         out
     }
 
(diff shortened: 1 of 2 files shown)


yjit/src/backend/ir.rs
```diff
@@ -1646,9 +1646,19 @@ impl Assembler {
...
+        // Note that this doesn't spill stack temps that are already popped
+        // but may still be used in the C arguments.
+        self.spill_temps();
+
         let out = self.next_opnd_out(Opnd::match_num_bits(&opnds));
+        // For C arguments, use stack temp registers that are still alive.
+        // After the C call that may clobber them, use spilled stack temps.
+        self.ctx.set_reg_temps(reg_temps);
         self.push_insn(Insn::CCall { fptr, opnds, out });
+        self.ctx.set_reg_temps(RegTemps::default());
```
maximecb, line 1661: Trying to make sure I understand what this PR does.

From the PR title it sounded like you are passing arguments via registers instead of through the temp stack, but looking at the code, it looks like the PR is doing something different.

You've moved the call to `spill_temps()` inside of `ccall()`, so that you can be smarter about which temps you spill instead of always spilling, correct?

I don't quite understand what the two calls to `set_reg_temps()` before and after the `Insn::CCall` are doing. I feel like the comment there could potentially be improved.

k0kubun: Essentially, this PR achieves a couple of optimizations:

1. Skip spilling off-stack `Opnd::Stack`s on `ccall()`. They'll also be passed to C arguments as registers.
2. Pass on-stack `Opnd::Stack`s to C arguments as registers _after spilling them_ on `ccall()`.

> You've moved the call to spill_temps() inside of ccall(), so that you can be smarter about which temps you spill instead of always spilling, correct?

So this discusses (1), but you may be missing (2).

Let's say we have this stack when entering an insn, `[A opnd (spilled), B opnd (register), C opnd (register)]`, and `C opnd` is popped before a `ccall()`. All of the operands are used by the call.

With (1), it no longer spills `C opnd`. `C opnd` is passed as a register as well.

With (2), it first spills `B opnd`. But at that moment, `B opnd` is also in a register. So we use a register to pass `B opnd`, which wasn't the case before this PR.

The resulting code looks like:

* before
    1. spill B
    2. spill C -- removed by (1)
    3. pass A (spilled)
    4. pass B (spilled) -- optimized by (2)
    5. pass C (spilled) -- optimized by (1)
* after
    1. spill B
    2. pass A (spilled)
    3. pass B (register)
    4. pass C (register)

k0kubun: > I feel like the comment there could potentially be improved.

How about this? https://github.com/ruby/ruby/pull/7920/commits/bcce57a4f5b2b8d436044aa94dad0e2c1c372464

PR #7923, opened 2023-06-09
[DOC] Regexp doc

BurdetteLamar:
Note: this breaks links in several other files;  fixes must be added before merge (but after the reviewed work is stable, I think).


doc/regexp/unicode_properties.rdoc
```diff
@@ -0,0 +1,590 @@
...
+- <tt>/\p{Any}/</tt>
+- <tt>/\p{Assigned}/</tt>
+- <tt>/\p{C}/</tt>
+- <tt>/\p{Cf}/</tt>
+- <tt>/\p{Cn}/</tt>
+- <tt>/\p{Co}/</tt>
+- <tt>/\p{Cs}/</tt>
+- <tt>/\p{L}/</tt>
+- <tt>/\p{LC}/</tt>
+- <tt>/\p{Ll}/</tt>
```
jaynetics, line 27: maybe these would be more intelligible if they used the long form, e.g. `\p{Lowercase_Letter}`?

doc/regexp/unicode_properties.rdoc
```diff
@@ -0,0 +1,590 @@
...
+- <tt>/\p{In_CJK_Unified_Ideographs_Extension_E}/</tt>
+- <tt>/\p{In_CJK_Unified_Ideographs_Extension_F}/</tt>
+- <tt>/\p{In_CJK_Compatibility_Ideographs_Supplement}/</tt>
+- <tt>/\p{In_CJK_Unified_Ideographs_Extension_G}/</tt>
+- <tt>/\p{In_CJK_Unified_Ideographs_Extension_H}/</tt>
+- <tt>/\p{In_Tags}/</tt>
+- <tt>/\p{In_Variation_Selectors_Supplement}/</tt>
+- <tt>/\p{In_Supplementary_Private_Use_Area_A}/</tt>
+- <tt>/\p{In_Supplementary_Private_Use_Area_B}/</tt>
+- <tt>/\p{In_No_Block}/</tt>
```
jaynetics, line 590: are unicode age properties, e.g. `\p{Age=1.1}`, omitted on purpose?

BurdetteLamar: Yes.  I tried all properties.  About 12-15 raised exceptions, mostly the age ones, but one or two others as well.

jaynetics: Which error did you get? `invalid character property name`? This would be strange, because age properties are still supported AFAICT. 

```sh
ruby -e "puts RUBY_VERSION, /\p{Age=1.1}/ =~ 'a'"
# => 3.2.0
# => 0
```

[This list](https://github.com/ammar/regexp_parser/blob/7413c52e0f6dca97c1d21e2daaea040c83eb46cb/lib/regexp_parser/syntax/token/unicode_property.rb#L40-L62) shows at which points each age property was added to Ruby. The only property for which support has ever been removed is `\p{newline}` (dropped in Ruby 2.0). I think the removal of `\p{newline}` is actually the only subtractive change ever made to Ruby's Regexp syntax, at least after Ruby 1.8.6.

doc/regexp/unicode_properties.rdoc
```diff
@@ -0,0 +1,590 @@
+== \Regexps Based on Character Properties
+
+- <tt>/\p{Alpha}}/</tt>
```
jaynetics, line 3:
```suggestion
- <tt>/\p{Alpha}/</tt>
```

doc/regexp/unicode_properties.rdoc
jaynetics: These supported properties are missing from the doc:

```
age=1.1
age=10.0
age=11.0
age=12.0
age=12.1
age=13.0
age=14.0
age=2.0
age=2.1
age=3.0
age=3.1
age=3.2
age=4.0
age=4.1
age=5.0
age=5.1
age=5.2
age=6.0
age=6.1
age=6.2
age=6.3
age=7.0
age=8.0
age=9.0
alphabetic
ascii_hex_digit
control
decimal_number
in_basic_latin
in_braille_patterns
in_deseret
in_meroitic_hieroglyphs
in_ol_chiki
in_old_north_arabian
in_old_south_arabian
in_palmyrene
in_sharada
in_shavian
lowercase
punctuation
uppercase
white_space
```

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+\Method Regexp#casefold? returns whether the mode is case-insensitive.
+
+=== Multiline Mode
+
+By default, a Ruby regexp is in multiline mode,
+which means that anchors <tt>^</tt> and <tt>$</tt> match, respectively,
+the beginning or end of a line in the target string.
+
+Modifier +m+ therefore does not enable multiline mode -- it's already enabled.
+The modifier does, however, make one small difference:
```
peterzhu2118, line 1061: I think it would be a good idea to write what the modifier does in words, rather than making the reader try to figure it out from the example.

doc/regexp.rdoc
```diff
@@ -757,71 +1175,80 @@ performance.
...
-    /(b|a+)/ =~ s #=> 0
-    /(b|a+)*/ =~ s #=> 0
+  (start = Time.now); /(b|a)/.match(s);   (Time.now - start) # => 4.4e-06
+  (start = Time.now); /(b|a+)/.match(s);  (Time.now - start) # => 4.6e-06
+  (start = Time.now); /(b|a+)*/.match(s); (Time.now - start) # => 5.2e-06
 
 However, the following pattern takes appreciably longer:
 
-    /(b|a+)*c/ =~ s #=> 26
+  (start = Time.now); /(b|a+)*c/.match(s); (Time.now - start) # => 5.3e-06
```
peterzhu2118, line 1189: I think there's a typo in the result, because it should take much longer than that.

doc/regexp/methods.rdoc
```diff
@@ -0,0 +1,42 @@
+== \Regexp Methods
```
peterzhu2118, line 1: I'm not sure what the purpose of having a full list of methods that accept regexp is?

doc/regexp/unicode_properties.rdoc
BurdetteLamar: $ ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x64-mingw-ucrt]
$ ruby -e "puts RUBY_VERSION, /\p{Age=1.1}/ =~ 'a'"
-e:1: invalid character property name {Age=1.1}: /\p{Age=1.1}/
-e: compile error (SyntaxError)

doc/regexp/methods.rdoc
BurdetteLamar: When I looked at these, I saw some methods that I've used but have not been aware that I can use a regexp (giving more power).  Just thought I'd remind others.  It's off-page, and so is pretty much free?

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+
+- <tt>\B</tt>: Matches non-word boundary:
+
+    /foo\B/.match('foobar')  # => #<MatchData "foo">
+    /foo\B/.match('foo bar') # => nil
+
+- <tt>\G</tt>: Matches first matching position:
+
+  In methods like String#gsub and String#scan, it changes on each iteration.
+  It initially matches the beginning of subject, and in each following iteration it matches where the last match finished.
```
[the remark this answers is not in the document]
BurdetteLamar: It's unchanged from current doc.  Not clear to me either, so I left it alone.  Frankly, I don't understand \G;  will research.

doc/regexp/unicode_properties.rdoc
BurdetteLamar: I see that at least some of those you've listed do work in my Ruby version.
I formed my list of candidates from enc/unicode/15.0.0/name2ctype.h.  Is there a better source?

doc/regexp.rdoc
BurdetteLamar: s = 'a' * 25 + 'd' + 'a' * 4 + 'c'
# => "aaaaaaaaaaaaaaaaaaaaaaaaadaaaac"
(start = Time.now); /(b|a+)*c/.match(s); (Time.now - start)
# => 7.0e-06
(start = Time.now); /(b|a+)*c/.match(s); (Time.now - start)
# => 1.07e-05
(start = Time.now); /(b|a+)*c/.match(s); (Time.now - start)
# => 7.3e-06

Is it possible that the engine has been improved?

doc/regexp/unicode_properties.rdoc
BurdetteLamar: Added.

doc/regexp/unicode_properties.rdoc
jaynetics: > ruby 3.2.2 (2023-03-30 revision https://github.com/ruby/ruby/commit/e51014f9c05aa65cbf203442d37fef7c12390015) [x64-mingw-ucrt]
> $ ruby -e "puts RUBY_VERSION, /\p{Age=1.1}/ =~ 'a'"
> -e:1: invalid character property name {Age=1.1}: /\p{Age=1.1}/

i've run specs involving age properties on many environments, and this is the first time i've seen this error. it is probably specific to mingw. i guess it could be a bug, or the flag `USE_UNICODE_AGE_PROPERTIES` is purposefully not set there for some memory reason...

> I formed my list of candidates from enc/unicode/15.0.0/name2ctype.h. Is there a better source?

i think `name2ctype` is fine, though for optimization reasons it does not list all properties in the same way. properties that happen to match the same codepoints as others are defined as aliases, e.g. `#define CR_In_Basic_Latin CR_ASCII`. i also have a gem, [regexp_property_values](https://github.com/jaynetics/regexp_property_values), that can provide the list (but it basically just reproduces what ruby's `tool/enc-unicode.rb` does, so might not be failsafe against all future changes).

BurdetteLamar: Thanks @jaynetics, for your help.

I'm incline to add the missing properties manually, just to keep it simple.  I've retained the order in the .h file, in case it matters.  Any ideas about where to add, or even about the overall ordering?

BurdetteLamar: I see that you've used alpha order over at your gem.  Will do the same.  May I take values from your file `values`?

BurdetteLamar: I have done so.

doc/regexp/unicode_properties.rdoc
BurdetteLamar: Fixed in revised file.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+
+- <tt>#match</tt>: see {Method match}[rdoc-ref:regexp.rdoc@Method+match].
+- <tt>#=~</tt>: see {Operator =~}[rdoc-ref:regexp.rdoc@Operator+-3D~].
+
+The affected global variables are:
+
+- <tt>$~</tt>: Returns a MatchData object, or +nil+.
+- <tt>$&</tt>: Returns the matched string, or +nil+.
+- <tt>$`</tt>: Returns the string to the left of the matched string, or +nil+.
+- <tt>$</tt>: Returns the string to the right of the matched string, or +nil+.
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+Method Regexp.escape returns an escaped string:
+
+  Regexp.escape('.?-+*^\|$()[]{}')
+  # => "\\.\\?\\-\\+\\*\\^\\\\\\|\\$\\(\\)\\[\\]\\{\\}"
+
+=== Source Literals
+
+The source literal largely behaves like a double-quoted string;
+see {String Literals}[rdoc-ref:syntax/literals.rdoc@String+Literals].
+`
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+    /\w{1,2}/.match('')    # => nil
+    /\w{1,2}/.match('x')   # => #<MatchData "x">
+    /\w{1,2}/.match('xyz') # => #<MatchData "xy">
+
+==== Greedy, Lazy, or Possessive Matching
+
+Quantifier matching may be greedy, lazy, or possessive:
+
+- In _greedy_ matching, as many occurrences as possible are matched.
+  while still allowing the overall match to succeed.
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1171 @@
...
+
+This table shows how each subexpression in the regexp above
+matches a substring in the target string:
+
+  | Subexpression in Regexp   | Matching Substring in Target String |
+  |---------------------------|-------------------------------------|
+  |       First '[csh]'       |            Character 'c'            |
+  |          '(..)'           |        First substring 'at'         |
+  |      First space ' '      |      First space character ' '      |
+  |       Second '[csh]'      |            Character 'c'            |
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
BurdetteLamar: Fixed.

doc/regexp/unicode_properties.rdoc
jaynetics: > May I take values from your file values?

of course!

BurdetteLamar:
@jeremyevans, would you like to review this before it's merged?

BurdetteLamar:
@peterzhu2118, ok to move to Ready for Review?

duerst:
A more general comment: When updating to a new Unicode version, there's of course a new value for the age property, but also in most cases new values for scripts and blocks (which start with "in"). Are we assuming that the documentation must be updated for every update to a new Unicode version? I think it would at least be very good to add a note saying that older Ruby versions, depending on what version of Unicode they support, may not support all these properties, and that newer versions may support additional properties.

doc/regexp/unicode_properties.rdoc
duerst: Just FYI, you can check whether a property is supported or not in the source. For Age=1.1, for example, it shows up at https://github.com/ruby/ruby/blob/master/enc/unicode/15.0.0/name2ctype.h#L43101. Please note that you have to lowercase and eliminate spaces (and I think also hyphens,...).

duerst: For the ordering of values, please have a look at https://github.com/k-takata/Onigmo/blob/master/doc/UnicodeProps.txt. I think that's better than just simply alphabetic order, because that gives the user at least a faint idea as to what all these properties are about.

duerst: > i've run specs involving age properties on many environments, and this is the first time i've seen this error. it is probably specific to mingw. i guess it could be a bug, or the flag `USE_UNICODE_AGE_PROPERTIES` is purposefully not set there for some memory reason...

The age properties indeed take quite a bit of memory. The inversion lists of the age properties (in name2ctype.h) are 24373 - 37636 = 13263 lines long, which is more than a third of the about 39000 lines overall for the inversion list data. The problem is that these lists repeat mostly the same data over and over again, because what's in version 15 is of course also in version 14 is in version 13... I'll look into how to improve that, but this may take time.

doc/regexp.rdoc
duerst: Yes, the engine indeed has been improved. See in particular https://bugs.ruby-lang.org/issues/19104. I think this part of the documentation should be carefully rewritten by somebody who understands the underlying issues. This should be handled separately from this pull request. @make_now_just

BurdetteLamar: Thanks, @duerst.  I agree; there's much more to some of Regexp than I understand.  I aspire only to improving its documentation, not to perfecting it.

doc/regexp/unicode_properties.rdoc
BurdetteLamar: > For the ordering of values, please have a look at https://github.com/k-takata/Onigmo/blob/master/doc/UnicodeProps.txt. I think that's better than just simply alphabetic order, because that gives the user at least a faint idea as to what all these properties are about.

@duerst, revised and much better; thanks.

jaynetics: IIUC, the parts of Onigmo that are used by Ruby are now maintained within the ruby repository, and `UnicodeProps.txt` from the original repository hasn't been updated since 2019, so it is missing 40 or so properties added since then.

doc/regexp/methods.rdoc
jaynetics: i would have enjoyed such a list as a Ruby beginner, though maybe it should be made clear that it is incomplete, e.g. "These are some of the methods that accept a Regexp as an argument"

jaynetics:
> When updating to a new Unicode version, there's of course a new value for the age property, but also in most cases new values for scripts and blocks (which start with "in"). Are we assuming that the documentation must be updated for every update to a new Unicode version? I think it would at least be very good to add a note saying that older Ruby versions, depending on what version of Unicode they support, may not support all these properties, and that newer versions may support additional properties.

@duerst Another option would be to have a script that generates the docs and is perhaps even run automatically when generating a new `name2ctype` file in `enc-unicode.rb`. This script could be a ruby version of https://github.com/k-takata/Onigmo/blob/master/tool/update-doc.py, or it could be a very simple script that reads `name2ctype`, if `enc-unicode.rb` was modified to add parseable comments before each group of properties that it writes.

peterzhu2118:
@BurdetteLamar I would prefer to have this PR fix the dead links that is caused by this change. That way we won't have an intermediate state where there is an increase in dead links in the Ruby master docs.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+    (XC|XL|L?X{0,3})    # tens - 90 (XC), 40 (XL), 0-30 (0 to 3 Xs),
+                        #        or 50-80 (L, followed by 0 to 3 Xs)
+    (IX|IV|V?I{0,3})    # ones - 9 (IX), 4 (IV), 0-3 (0 to 3 Is),
+                        #        or 5-8 (V, followed by 0 to 3 Is)
+    $                   # end of string
+  EOT
+  re = /#{pattern}/x
+  re.match('MCMXLIII') # => #<MatchData "MCMXLIII" 1:"CM" 2:"XL" 3:"III">
+
+=== Interpolation Mode
```
jeremyevans, line 1104: The `o` flag means that the first time the literal regexp with interpolations is encountered, the generated Regexp object is saved and used for all future evaluations of the literal regexp.  Without the `o` flag, the generated Regexp is not saved, so each evaluation of the literal regexp generates a new Regexp object.  Note that if the literal regexp does not have interpolations, the `o` behavior is the default.

I recommend this section be rewritten with that in mind.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+With modifier +o+, the interpolated method is called only once,
+and so takes about 5 seconds to execute:
+
+  start = Time.now
+  words.each {|word| word.match(/\A[#{letters}]+\z/o) }
+  Time.now - start # => 5.0010866
+
+== Encodings
+
+By default, a regexp has US-ASCII encoding:
```
jeremyevans, line 1126: `By default, a regexp with only US-ASCII characters has US-ASCII encoding:`

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+
+  re = /foo/
+  re.source.encoding # => #<Encoding:US-ASCII>
+  re.encoding        # => #<Encoding:US-ASCII>
 
 Regular expressions are assumed to use the source encoding. This can be
```
jeremyevans, line 1132: `Regular expressions containing non-US-ASCII characters are assumed to use the source encoding`

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
 overridden with one of the following modifiers.
 
-* <tt>/</tt><i>pat</i><tt>/u</tt> - UTF-8
-* <tt>/</tt><i>pat</i><tt>/e</tt> - EUC-JP
-* <tt>/</tt><i>pat</i><tt>/s</tt> - Windows-31J
-* <tt>/</tt><i>pat</i><tt>/n</tt> - ASCII-8BIT
+- <tt>/_pat_/n</tt>: ASCII-8BIT
```
jeremyevans, line 1135: Only `ASCII-8BIT` if the regexp contains non-US-ASCII characters:

```ruby
/foo/n.encoding
# => #<Encoding:US-ASCII>
/foo\xff/n.encoding
# => #<Encoding:ASCII-8BIT>
/foo\x7f/n.encoding
# => #<Encoding:US-ASCII>
```

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
-            # same as m[2]
-    $3      #=> nil
-            # no third group in pattern
-    $+      #=> "c"
-            # same as m[-1]
-
-These global variables are thread-local and method-local variables.
+- They have the same encoding.
+- The regexp's encoding is _US-ASCII_ and the string's encoding
+  is ASCII-compatible.
```
jeremyevans, line 1156: Not strictly true:

```ruby
re = eval("# encoding: ISO-8859-1\n/foo\\xff?/")
re.encoding
# => #<Encoding:ISO-8859-1>
re =~ "foo".encode("UTF-8")
# => 0
re =~ "foo\u0100"
# Encoding::CompatibilityError
```

The regexp encoding does not need to be US-ASCII, it may just need to have fixed encoding.  And whether you get Encoding::CompatibilityError depends not just on the string's encoding, but whether the string contains only ASCII characters.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+
+  # Regexp with encoding ISO-8859-1.
+  re = Regexp.new("a".force_encoding('iso-8859-1'), Regexp::FIXEDENCODING)
+  re.encoding  # => #<Encoding:ISO-8859-1>
+  # Target string with encoding UTF-8.
+  s = "a\u3042"
+  s.encoding   # => #<Encoding:UTF-8>
+  re.match(s)  # Raises Encoding::CompatibilityError.
 
 == Performance
```
jeremyevans, line 1175: I recommend removing this section as it does not reflect optimizations made in Ruby 3.2.

jeremyevans (changes requested):
See inline comments.

BurdetteLamar:
> A more general comment: When updating to a new Unicode version, there's of course a new value for the age property, but also in most cases new values for scripts and blocks (which start with "in"). Are we assuming that the documentation must be updated for every update to a new Unicode version? I think it would at least be very good to add a note saying that older Ruby versions, depending on what version of Unicode they support, may not support all these properties, and that newer versions may support additional properties.

Added.

BurdetteLamar:
> @BurdetteLamar I would prefer to have this PR fix the dead links that is caused by this change. That way we won't have an intermediate state where there is an increase in dead links in the Ruby master docs.

I have fixed links in doc/literals.rdoc, re.c, string.c, and string.rb.  Are there still some broken?

doc/regexp/methods.rdoc
BurdetteLamar: I meant for it to be complete, and would like it to be.  Are there others?

peterzhu2118:
> I have fixed links in doc/literals.rdoc, re.c, string.c, and string.rb. Are there still some broken?

Ah you fixed the links after my review, I didn't see that commit. Thanks for fixing the links. I think this is ready for review.

doc/regexp.rdoc
makenowjust: Thank you for mentioning it. I recently updated the Regexp optimization (see #7931). Thus, this status is a little unstable, and I will try to update the docs after stabilizing this.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
 
-    /[0-9a-f]/.match('9f') #=> #<MatchData "9">
-    /[9f]/.match('9f')     #=> #<MatchData "9">
+  /bar/ =~ 'foo bar' # => 4
+  'foo bar' =~ /bar/ # => 4
+  /baz/ =~ 'foo bar' # => nil
 
-If the first character of a character class is a caret (<tt>^</tt>) the
-class is inverted: it matches any character _except_ those named.
+Operator NilClass#=~ returns +nil+ and does not set global variables.
```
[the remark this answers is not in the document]
BurdetteLamar: Removed.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+
+\Regexp special characters, called _metacharacters_,
+has special meanings in certain contexts;
+depending on the context, these are sometimes metacharacters:
+
+  . ? - + * ^ \ | $ ( ) [ ] { }
+
+To match a metacharacter literally, backslash-escape it:
+
+  # Metacharacter period matches any character except newline.
```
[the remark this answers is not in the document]
BurdetteLamar: Modified.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+    /bar$/.match("foo\nbar") # => #<MatchData "bar">
+    /ba$/.match("foo\nbar")  # => nil
+
+- <tt>\A</tt>: Matches the beginning of the string:
+
+    /\Afoo/.match('foo bar')  # => #<MatchData "foo">
+    /\Afoo/.match(' foo bar') # => nil
+
+- <tt>\Z</tt>: Matches the end of the string;
+  if string ends with a single newline,  it matches just before newline:
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
+
+    'The cat sat in the hat'.gsub(/[csh]at/, '\0s')
     # => "The cats sats in the hats"
 
-=== Named Captures
+==== Named Captures
+
+As seen above, a capture can be referred to by its number.
+A capture can also have a name,
+prefixed as <tt>?<_name_></tt> or <tt>?'_name_'></tt>,
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp.rdoc
BurdetteLamar: Fixed.

doc/regexp.rdoc
BurdetteLamar: Fixed.

doc/regexp.rdoc
BurdetteLamar: Modified.

doc/regexp.rdoc
BurdetteLamar: Modified.

doc/regexp.rdoc
BurdetteLamar: Removed.

doc/regexp/methods.rdoc
BurdetteLamar: > i would have enjoyed such a list as a Ruby beginner, though maybe it should be made clear that it is incomplete, e.g. "These are some of the methods that accept a Regexp as an argument"

I've added a note that core methods are listed.  Are some core methods missing?

doc/regexp/unicode_properties.rdoc
BurdetteLamar: Both forms now in revised list.

BurdetteLamar:
@jaynetics, have all your comments been resolved?

doc/regexp/methods.rdoc
jaynetics: > Are some core methods missing?

Only `String#[]=` jumps out to me. 

There might be other methods that have a more or less specific handling for regexps, either now or in the future.

Of course everything that takes an object also takes a Regexp, e.g. `Array#push`. This is another reason why I'd advocate for presenting this as a curated list.

Finally, I'm not sure `Kernel#{sub,gsub}` still exist.

jaynetics:
> @jaynetics, have all your comments been resolved?

i had one more remark [here](https://github.com/ruby/ruby/pull/7923/files#r1226872151), but in general this looks good to me. an impressive effort!

doc/regexp/methods.rdoc
BurdetteLamar: Added String#[]=;  removed Kernel.  Added a note:  is it adequate?

BurdetteLamar:
@duerst, have all your comments been resolved?

doc/regexp/methods.rdoc
```diff
@@ -0,0 +1,41 @@
...
+- Enumerable#one?
+- Enumerable#slice_after
+- Enumerable#slice_before
+- Regexp#=~
+- Regexp#match
+- Regexp#match?
+- Regexp.new
+- Regexp.union
+- String#=~
+- String[]=
```
jaynetics, line 19:
```suggestion
- String#[]=
```

doc/regexp/methods.rdoc
jaynetics: I can't see a note, did you add it in the last commit?

BurdetteLamar: Should not have characterized it as a note;  I changed the language in methods.rdoc line 3, to avoid suggesting that the list is complete.

doc/regexp/methods.rdoc
```diff
@@ -0,0 +1,41 @@
...
+- String#byterindex
+- String#gsub
+- String#gsub!
+- String#index
+- String#match
+- String#match?
+- String#partition
+- String#rindex
+- String#rpartition
+- StringString#scan
```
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

doc/regexp/methods.rdoc
BurdetteLamar: Fixed.

duerst:
> @duerst, have all your comments been resolved?

Thanks a lot for all your work. Just one point: It says "The properties shown here are those currently supported in Ruby." This may be true currently, but because the documentation may get out of sync, I'd suggest to instead write "The properties shown here should be those currently supported in Ruby." or some such.

BurdetteLamar:
> > @duerst Another option would be to have a script that generates the docs and is perhaps even run automatically when generating a new `name2ctype` file in `enc-unicode.rb`. This script could be a ruby version of https://github.com/k-takata/Onigmo/blob/master/tool/update-doc.py, or it could be a very simple script that reads `name2ctype`, if `enc-unicode.rb` was modified to add parseable comments before each group of properties that it writes.
> 
> I'm opened an issue for this at https://bugs.ruby-lang.org/issues/19728. Comments are welcome.

We already have `tool/enc-unicode.rb`, which as I understand makes `enc/unicode/<VERSION>/name2ctype.h`.  Would it make sense for it to also make `doc/regexp/unicode_properties.rdoc`?

duerst:
> > I'm opened an issue for this at https://bugs.ruby-lang.org/issues/19728. Comments are welcome.
> 
> We already have `tool/enc-unicode.rb`, which as I understand makes `enc/unicode/<VERSION>/name2ctype.h`. Would it make sense for it to also make `doc/regexp/unicode_properties.rdoc`?

@BurdetteLamar Please continue the discussion at https://bugs.ruby-lang.org/issues/19728, not here.

BurdetteLamar:
> > @duerst Another option would be to have a script that generates the docs and is perhaps even run automatically when generating a new `name2ctype` file in `enc-unicode.rb`. This script could be a ruby version of https://github.com/k-takata/Onigmo/blob/master/tool/update-doc.py, or it could be a very simple script that reads `name2ctype`, if `enc-unicode.rb` was modified to add parseable comments before each group of properties that it writes.
> 
> I'm opened an issue for this at https://bugs.ruby-lang.org/issues/19728. Comments are welcome.

We already have `tool/enc-unicode.rb`, which as I understand makes `enc/unicode/<VERSION>/name2ctype.h`.  Would it make sense for it to also make `doc/regexp/unicode_properties.rdoc`?

> > @duerst, have all your comments been resolved?
> 
> Thanks a lot for all your work. Just one point: It says "The properties shown here are those currently supported in Ruby." This may be true currently, but because the documentation may get out of sync, I'd suggest to instead write "The properties shown here should be those currently supported in Ruby." or some such.

I'm not comfortable with this.  If the properties shown here should be those currently supported, I think we ought to make sure they are, now and always.

BurdetteLamar:
@jaynetics and @duerst, have all your comments been resolved?

BurdetteLamar:
@peterzhu2118 and @jeremyevans, anything more for this PR?

doc/regexp.rdoc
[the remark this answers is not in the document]
jeremyevans: This still looks wrong. `?'_name_'>` should be replaced with `?'_name_'` (the unmatched `>` should be removed).

doc/regexp.rdoc
```diff
@@ -1,753 +1,1176 @@
...
-        #=> #<MatchData "ototo" vowel:"o">
+When (and only when) a regexp contains named capture groups
+and appears before the <tt>=~</tt> operator,
+the captured substrings are assigned to local variables with corresponding names:
 
-*Note*: A regexp can't use named backreferences and numbered
-backreferences simultaneously. Also, if a named capture is used in a
-regexp, then parentheses used for grouping which would otherwise result
-in a unnamed capture are treated as non-capturing.
+  /\$(?<dollars>\d+)\.(?<cents>\d+)/.match('$3.67')
```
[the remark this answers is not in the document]
jeremyevans: This still needs to be fixed.

doc/regexp.rdoc
jeremyevans: This section still needs to be rewritten IMO.

doc/regexp.rdoc
```diff
@@ -1,827 +1,1230 @@
...
 
-A similar case is typified by the following example, which takes
-approximately 60 seconds to execute for me:
+A regular expression containing non-US-ASCII characters
+is assumed to use the source encoding.
+This can be overridden with one of the following modifiers.
 
-Match a string of 29 <i>a</i>s against a pattern of 29 optional <i>a</i>s
-followed by 29 mandatory <i>a</i>s:
+- <tt>/_pat_/n</tt>: US-ASCII or ASCII-8BIT
```
jeremyevans, line 1134: Maybe: `US -ASCII if only containing US-ASCII characters, otherwise ASCII-8BIT`

doc/regexp.rdoc
```diff
@@ -1,827 +1,1230 @@
...
+    /foo/s.encoding # => #<Encoding:Windows-31J>
+
+A regexp can be matched against a target string when either:
+
+- They have the same encoding.
+- The regexp's encoding is a fixed encoding and the string
+  contains only ASCII characters.
+  Method Regexp#fixed_encoding? returns whether the regexp
+  has a <i>fixed</i> encoding.
+
```
jeremyevans, line 1159: Remove extra blank line.

jeremyevans (changes requested):
A few of my comments weren't addressed, please see inline comments.

doc/regexp.rdoc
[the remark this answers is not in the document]
BurdetteLamar: Fixed.

jeremyevans:
@BurdetteLamar I'm guessing since you are asking for another review, you want to push additional changes? :)

doc/regexp.rdoc
BurdetteLamar: Rewritten.

BurdetteLamar:
Thought I had done so, thanks (typo in push command).  Now done.

    On Sunday, June 18, 2023 at 12:16:16 PM CDT, Jeremy Evans ***@***.***> wrote:  
 
 


@BurdetteLamar I'm guessing since you are asking for another review, you want to push additional changes? :)

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+- <tt>#match</tt>: see {Method match}[rdoc-ref:regexp.rdoc@Method+match].
+- <tt>#=~</tt>: see {Operator =~}[rdoc-ref:regexp.rdoc@Operator+-3D~].
+
+The affected global variables are:
+
+- <tt>$~</tt>: Returns a MatchData object, or +nil+.
+- <tt>$&</tt>: Returns the matched string, or +nil+.
+- <tt>$`</tt>: Returns the string to the left of the matched string, or +nil+.
+- <tt>$'</tt>: Returns the string to the right of the matched string, or +nil+.
```
jaynetics, line 123:
```suggestion
- <tt>$&</tt>: Returns the matched part of the string, or +nil+.
- <tt>$`</tt>: Returns the part of the string to the left of the match, or +nil+.
- <tt>$'</tt>: Returns the part of the string to the right of the match, or +nil+.
```

i think this would be less ambiguous, especially for `$&`

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+- {Quantifiers}[rdoc-ref:regexp.rdoc@Quantifiers]
+- {Groups and captures}[rdoc-ref:regexp.rdoc@Groups+and+Captures]
+- {Unicode}[rdoc-ref:regexp.rdoc@Unicode]
+- {POSIX Bracket Expressions}[rdoc-ref:regexp.rdoc@POSIX+Bracket+Expressions]
+- {Comments}[rdoc-ref:regexp.rdoc@Comments]
+
+=== Special Characters
+
+\Regexp special characters, called _metacharacters_,
+has special meanings in certain contexts;
```
jaynetics, line 190:
```suggestion
have special meanings in certain contexts;
```

typo

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+  while still allowing the overall match to succeed.
+  Greedy quantifiers: <tt>*</tt>, <tt>+</tt>, <tt>?</tt>,
+  <tt>{min, max}</tt> and its variants.
+- In _lazy_ matching, the minimum number of occurrences are matched.
+  Lazy quantifiers: <tt>*?</tt>, <tt>+?</tt>, <tt>??</tt>,
+  <tt>{min, max}?</tt> and its variants.
+- In _possessive_ matching, once a match is found, there is no backtracking;
+  that match is retained, even if it jeopardises the overall match.
+  Possessive quantifiers: <tt>*+</tt>, <tt>++</tt>, <tt>?+</tt>,
+  <tt>{min, max}+</tt> and its variants.
```
jaynetics, line 552:
```suggestion
  Possessive quantifiers: <tt>*+</tt>, <tt>++</tt>, <tt>?+</tt>.
  Note that <tt>{min, max}</tt> and its variants do _not_ support possessive matching.
```

see https://github.com/k-takata/Onigmo/blob/791140951eefcf17db4e762e789eb046ea8a114c/doc/RE#L155-L156

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+  re = /\A(?<xyzzy>foo)?(?(<xyzzy>)(T)|(F))\z/
+  re.match('fooT') # => #<MatchData "fooT" xyzzy:"foo">
+  re.match('F')    # => #<MatchData "F" xyzzy:nil>
+  re.match('fooF') # => nil
+  re.match('T')    # => nil
 
-This pattern matches a <i>(</i> character and assigns it to the <tt>paren</tt>
-group, tries to call that the <tt>paren</tt> sub-expression again but fails,
-then matches a literal <i>)</i>:
+=== Unicode
```
jaynetics, line 786:
```suggestion

==== Absence Operator

The absence operator is a special group that matches anything which does _not_ match the contained subexpressions.

  /(?~real)/.match('surrealist') # => #<MatchData "surrea">
  /(?~real)ist/.match('surrealist') # => #<MatchData "ealist">
  /sur(?~real)ist/.match('surrealist') # => nil

=== Unicode
```

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+
+    /\s/.match('foo bar') # => #<MatchData " ">
+    /\s/.match('foo')     # => nil
+
+- <tt>/\S/</tt>: Matches a non-whitespace character: equivalent to <tt>/[^ \t\r\n\f\v]/</tt>:
+
+    /\S/.match(" \t\r\n\f\v foo") # => #<MatchData "f">
+    /\S/.match(" \t\r\n\f\v")     # => nil
+
+- <tt>/\R/</tt>: Matches a linebreak:
```
jaynetics, line 327:
```suggestion
- <tt>/\R/</tt>: Matches a linebreak, platform-independently:
```

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
-end of the capture group, and the <tt>\0</tt> backreference uses the
-implicit capture group of the entire match.  However, you can use
-this backreference when doing substitution:
+  re = /(foo)+/
+  re.match('')       # => nil
+  re.match('foo')    # => #<MatchData "foo" 1:"foo">
+  re.match('foofoo') # => #<MatchData "foofoo" 1:"foo">
 
-  "The cat sat in the hat".gsub(/[csh]at/, '\0s')
+The returned \MatchData object give access to matched the substrings:
```
jaynetics, line 598:
```suggestion
The returned \MatchData object gives access to the matched substrings:
```

typos

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
-side of an expression and the <tt>=~</tt> operator, the captured text is
-also assigned to local variables with corresponding names.
+This causes the subexpression to be matched
+independently of the rest of the expression,
+so that the matched substring becomes fixed for the remainder of the match,
+unless the entire subexpression must be abandoned and subsequently revisited.
 
-    /\$(?<dollars>\d+)\.(?<cents>\d+)/ =~ "$3.67" #=> 0
-    dollars #=> "3"
+In this way pat is treated as a non-divisible whole.
```
jaynetics, line 698:
```suggestion
In this way _subexpression_ is treated as a non-divisible whole.
```

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+  {Extended Mode}[rdoc-ref:regexp.rdoc@Extended+Mode].
+- +o+: <tt>/_pattern_/o</tt> sets
+  {Interpolation Mode}[rdoc-ref:regexp.rdoc@Interpolation+Mode].
+
+Any, all, or none of these may be applied.
+
+Modifiers +i+, +m+, and +x+ may be applied to subexpressions:
+
+- A preceding <tt>(?_modifier_)</tt> turns the mode "on."
+- A preceding <tt>(?-_modifier_)</tt> turns the mode "off."
```
jaynetics, line 1026:
```suggestion
- <tt>(?_modifier_)</tt> turns the mode "on" for ensuing subexpressions
- <tt>(?-_modifier_)</tt> turns the mode "off" for ensuing subexpressions
- <tt>(?_modifier_:_subexp_)</tt> turns the mode "on" for _subexp_ within the group
- <tt>(?-_modifier_:_subexp_)</tt> turns the mode "off" for _subexp_ within the group
```

added info about "group-local options"

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
 Example:
 
-    m = /s(\w{2}).*(c)/.match('haystack') #=> #<MatchData "stac" 1:"ta" 2:"c">
-    $~                                    #=> #<MatchData "stac" 1:"ta" 2:"c">
-    Regexp.last_match                     #=> #<MatchData "stac" 1:"ta" 2:"c">
+  re = /(?i)te(?-i)st/
+  re.match('test') # => #<MatchData "test">
+  re.match('TEst') # => #<MatchData "TEst">
+  re.match('TEST') # => nil
+  re.match('teST') # => nil
```
jaynetics, line 1034:
```suggestion
  re.match('teST') # => nil
  
  re = /t(?i:e)st/
  re.match('test') # => #<MatchData "test">
  re.match('tEst') # => #<MatchData "tEst">
  re.match('tEST') # => nil
```

added example about "group-local options"

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+\Method Regexp#casefold? returns whether the mode is case-insensitive.
+
+=== Multiline Mode
+
+By default, a Ruby regexp is in multiline mode,
+which means that anchors <tt>^</tt> and <tt>$</tt> match, respectively,
+the beginning or end of a line in the target string.
+
+Modifier +m+ therefore does not enable multiline mode -- it's already enabled.
+The modifier does, however, make one small difference:
```
jaynetics, line 1059:
```suggestion
The multiline-mode in Ruby is what is commonly called a "dot-all mode". Unlike other languages, the _m_ modifier does not affect the anchors <tt>^</tt> and <tt>$</tt>. These anchors always match at line-boundaries in Ruby.
```

to me this was a bit long-winded and confusing.

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+
+- Without the +m+ modifier, a newline does not match subexpression <tt>.</tt>:
+
+    /a.c/.match("a\nc")  # => nil
+
+- With the modifier, it does match:
+
+    /a.c/m.match("a\nc") # => #<MatchData "a\nc">
+
```
jaynetics, line 1068:
```suggestion
The multiline-mode in Ruby is what is commonly called a "dot-all mode":

- Without the +m+ modifier, the subexpression <tt>.</tt> does not match newlines:

    /a.c/.match("a\nc")  # => nil

- With the modifier, it does match:

    /a.c/m.match("a\nc") # => #<MatchData "a\nc">

Unlike other languages, the _m_ modifier does not affect the anchors <tt>^</tt> and <tt>$</tt>.
These anchors always match at line-boundaries in Ruby.

```

to me this was a bit long-winded and confusing.

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+=== Extended Mode
+
+Modifier +x+ enables extended mode, which means that:
+
+- Literal white space in the pattern is to be ignored.
+- Character <tt>#</tt> marks the remainder of its containing line as a comment,
+  which is also to be ignored.
```
jaynetics, line 1075:
```suggestion
  which is also to be ignored for matching purposes.
```

doc/regexp.rdoc
```diff
@@ -1,827 +1,1236 @@
...
+    Regexp.timeout = 3.0
+    Regexp.timeout # => 3.0
+
+- An instance timeout, which defaults to +nil+ and may be set in Regexp.new:
+
+    re = Regexp.new('foo', timeout: 5.0)
+    re.timeout # => 5.0
+
+When regexp.timeout is +nil+, the timeout "falls through" to Regexp.timeout;
+when regexp.timeout is non-+nil+, that value control timing out:
```
jaynetics, line 1209:
```suggestion
when regexp.timeout is non-+nil+, that value controls timing out:
```

typo

jaynetics:
@BurdetteLamar please find another 10 or so suggestions above. sorry for coming in late with all of these; i only found time to read the full new document now.

doc/regexp.rdoc
```diff
@@ -1,827 +1,1250 @@
...
+
+    /\R/.match("\r")     # => #<MatchData "\r">     # Carriage return (CR)
+    /\R/.match("\n")     # => #<MatchData "\n">     # Newline (LF)
+    /\R/.match("\f")     # => #<MatchData "\f">     # Formfeed (FF)
+    /\R/.match("\v")     # => #<MatchData "\v">     # Vertical tab (VT)
+    /\R/.match("\r\n")   # => #<MatchData "\r\n">   # CRLF
+    /\R/.match("\u0085") # => #<MatchData "\u0085"> # Next line (NEL)
+    /\R/.match("\u2028") # => #<MatchData "\u2028"> # Line separator (LSEP)
+    /\R/.match("\u2029") # => #<MatchData "\u2029"> # Paragraph separator (PSEP)
+
```
jaynetics, line 337:
```suggestion

- <tt>/\X/</tt>: Matches extended grapheme clusters, i.e. any sequence 
  of codepoints that produce a single, human-readable character:

    /\X/.match("A") # => #<MatchData "A"> # plain A 
    /\X/.match("A\u0308") # => #<MatchData "Ä"> # A + combining diaeresis
    /\X/.match("\u{1F441 200D 1F5E8}") # => #<MatchData "👁️‍🗨️"> #  👁️ + ZWJ +  🗨️

```

jaynetics (commented):
i don't have approval rights, but to me it looks good now and considerably improved from before.

jeremyevans (approved):
No further changes requested by me.  Due to the size of the PR, please wait for an additional approval from @peterzhu2118 or @duerst.

Also, please remember to squash when merging.

peterzhu2118 (approved):
One minor comment, but otherwise looks good!

doc/regexp.rdoc
```diff
@@ -1,827 +1,1250 @@
...
+  |         nil          |          nil         |       Never times out.      |
+  |         nil          |         Float        | Times out in Float seconds. |
+  |        Float         |          Any         | Times out in Float seconds. |
+
+Example:
 
-== Timeout
+  Regexp.timeout = 2.0
+  s = 'a' * 10 + 'x'
+  re.match(s) # => nil
```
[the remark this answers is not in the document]
BurdetteLamar: I did not succeed in crafting an example that actually times out, so I've removed the  example.

BurdetteLamar:
@jeremyevans, I did not mean to ask for your re-review.

peterzhu2118 (approved):

PR #7924, opened 2023-06-09
Move GC age out of the RVALUE flags

eightbitraptor:
Previously 2 bits of the flags on each RVALUE are reserved to store the number of GC cycles that each object has survived. This commit introduces a new bit array on the heap page, called age_bits, to store that information instead.

This patch allows us to free up a bit from the RVALUE flags that can be reused, and it also abstracts a generational GC concern out of the actual object model used by Ruby.

One thing to note is that we can only free up one RVALUE flag rather than both the age flags.

This is because the Write barrier needs to know when the object is old or young. It's always done this by accessing the flags, which is fast, and making it do a lookup on a bitmask on the heap page would seriously impact the performance.

So this commit reserves one of the bits, the old FL_PROMOTED0, which has been renamed to FL_PROMOTED, in order to provide a fast method for the write barrier to detect whether the object is old or not. Old objects (age 3+) will have FL_PROMOTED set, and young objects (age <3) will not.


tool/runruby.rb
```diff
@@ -140,12 +140,7 @@ def File.realpath(*args)
 if debugger
   case debugger
   when :gdb, nil
-    debugger = %W'gdb -x #{srcdir}/.gdbinit'
-    if File.exist?(gdb = 'run.gdb') or
-      File.exist?(gdb = File.join(abs_archdir, 'run.gdb'))
-      debugger.push('-x', gdb)
-    end
-    debugger << '--args'
+    debugger = %W'rr record '
```
nobu, line 143: What is this?

vm.c
```diff
@@ -3667,6 +3667,8 @@ Init_VM(void)
     fcore = rb_class_new(rb_cBasicObject);
     rb_set_class_path(fcore, rb_cRubyVM, "FrozenCore");
     RBASIC(fcore)->flags = T_ICLASS;
+    rb_reset_age(fcore);
+
```
nobu, line 3671:
```suggestion

```

include/ruby/internal/fl_type.h
```diff
@@ -57,8 +57,8 @@
 
 #define FL_SINGLETON    RBIMPL_CAST((VALUE)RUBY_FL_SINGLETON)            /**< @old{RUBY_FL_SINGLETON} */
 #define FL_WB_PROTECTED RBIMPL_CAST((VALUE)RUBY_FL_WB_PROTECTED)         /**< @old{RUBY_FL_WB_PROTECTED} */
-#define FL_PROMOTED0    RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED0)            /**< @old{RUBY_FL_PROMOTED0} */
-#define FL_PROMOTED1    RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED1)            /**< @old{RUBY_FL_PROMOTED1} */
+#define FL_PROMOTED    RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED0)             /**< @old{RUBY_FL_PROMOTED0} */
```
nobu, line 60:
```suggestion
#define FL_PROMOTED     RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED0)            /**< @old{RUBY_FL_PROMOTED0} */
```

gc.c
```diff
@@ -2951,10 +2999,15 @@ rb_class_instance_allocate_internal(VALUE klass, VALUE flags, bool wb_protected)
...
+    switch(flags & RUBY_T_MASK) {
+      case T_OBJECT:
         return rb_class_instance_allocate_internal(klass, (flags | ROBJECT_EMBED) & ~FL_WB_PROTECTED, flags & FL_WB_PROTECTED);
-    }
-    else {
+        break;
+      case T_MODULE:
+      case T_ICLASS:
+      case T_CLASS:
+        if (RGENGC_WB_PROTECTED_CLASS) flags |= FL_WB_PROTECTED;
```
peterzhu2118, line 3009: Why is this change needed?

gc.c
```diff
@@ -3443,6 +3496,8 @@ obj_free(rb_objspace_t *objspace, VALUE obj)
         break;
     }
 
+    RVALUE_AGE_RESET(obj);
```
peterzhu2118, line 3499: `heap_page_add_freeobj` also resets the age. Do we need to do it in both places?

gc.c
```diff
@@ -9852,6 +9909,7 @@ gc_move(rb_objspace_t *objspace, VALUE scan, VALUE free, size_t src_slot_size, s
     }
 
     memset(src, 0, src_slot_size);
+    RVALUE_CLEAR_AGE_BITS(GET_HEAP_PAGE(src)->age_bits, NUM_IN_PAGE(src));
```
peterzhu2118, line 9912:
```suggestion
    RVALUE_AGE_RESET((VALUE)src);
```

gc.c
```diff
@@ -9882,6 +9940,8 @@ gc_move(rb_objspace_t *objspace, VALUE scan, VALUE free, size_t src_slot_size, s
         CLEAR_IN_BITMAP(GET_HEAP_UNCOLLECTIBLE_BITS((VALUE)dest), (VALUE)dest);
     }
 
+    RVALUE_SET_AGE_BITS(GET_HEAP_PAGE(dest)->age_bits, NUM_IN_PAGE(dest), age);
```
peterzhu2118, line 9943:
```suggestion
    RVALUE_SET_AGE((VALUE)dest, age);
```

gc.c
```diff
@@ -13447,7 +13507,7 @@ rb_raw_obj_info_common(char *const buff, const size_t buff_size, const VALUE obj
         }
     }
     else {
-        const int age = RVALUE_FLAGS_AGE(RBASIC(obj)->flags);
+        const int age = (int)RVALUE_GET_AGE_BITS(GET_HEAP_PAGE(obj)->age_bits, NUM_IN_PAGE(obj));
```
peterzhu2118, line 13510:
```suggestion
        const int age = (int)RVALUE_GET_AGE(obj);

File: gc.c [hunk 9]
```diff
@@ -9810,6 +9866,7 @@ gc_move(rb_objspace_t *objspace, VALUE scan, VALUE free, size_t src_slot_size, s
     wb_unprotected = RVALUE_WB_UNPROTECTED((VALUE)src);
     uncollectible = RVALUE_UNCOLLECTIBLE((VALUE)src);
     bool remembered = RVALUE_REMEMBERED((VALUE)src);
+    age = (int)RVALUE_GET_AGE_BITS(GET_HEAP_PAGE(src)->age_bits, NUM_IN_PAGE(src));
```
peterzhu2118, line 9869:
```suggestion
    age = (int)RVALUE_GET_AGE((VALUE)src);
```

include/ruby/internal/fl_type.h
```diff
@@ -57,8 +57,8 @@
 
 #define FL_SINGLETON    RBIMPL_CAST((VALUE)RUBY_FL_SINGLETON)            /**< @old{RUBY_FL_SINGLETON} */
 #define FL_WB_PROTECTED RBIMPL_CAST((VALUE)RUBY_FL_WB_PROTECTED)         /**< @old{RUBY_FL_WB_PROTECTED} */
-#define FL_PROMOTED0    RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED0)            /**< @old{RUBY_FL_PROMOTED0} */
-#define FL_PROMOTED1    RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED1)            /**< @old{RUBY_FL_PROMOTED1} */
+#define FL_PROMOTED     RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED0)             /**< @old{RUBY_FL_PROMOTED0} */
```
peterzhu2118, line 60: Why is this still called `RUBY_FL_PROMOTED0`? We should call it `RUBY_FL_PROMOTED` for the name to be in sync.

include/ruby/internal/fl_type.h
```diff
@@ -57,8 +57,8 @@
...
+#define FL_UNUSED       RBIMPL_CAST((VALUE)RUBY_FL_PROMOTED1)            /**< @old{RUBY_FL_PROMOTED1} */
```
peterzhu2118, line 61: `RUBY_FL_PROMOTED1` doesn't exist anymore.

include/ruby/internal/fl_type.h
```diff
@@ -237,24 +237,7 @@ ruby_fl_type {
...
-     * collections.
-     *
-     * @internal
-     *
-     * But honestly, @shyouhei  doesn't think this flag should  be visible from
-     * 3rd parties.  It must be an implementation detail that they should never
-     * know.  Might better be hidden.
-     */
-    RUBY_FL_PROMOTED     = RUBY_FL_PROMOTED0 | RUBY_FL_PROMOTED1,
+    RUBY_FL_UNUSED    = (1<<6),
```
peterzhu2118, line 240: The comment above doesn't apply anymore.

test/objspace/test_objspace.rb
```diff
@@ -644,41 +644,6 @@ def assert_test_string_entry_correct_in_dump_all(output)
...
-
-        puts dump_my_heap_please
-      end;
-      assert_nil(error)
-      dump = File.readlines(output)
-      File.unlink(output)
-
-      assert_test_string_entry_correct_in_dump_all(dump)
-    end
-
```
peterzhu2118, line 681: Why was this test removed?

transient_heap.c
```diff
@@ -422,6 +422,7 @@ rb_transient_heap_alloc(VALUE obj, size_t req_size)
 
             /* ptr is set up; OK to unpoison. */
             asan_unpoison_memory_region(ptr, size - sizeof *header, true);
+
```
peterzhu2118, line 425: Nit: can you remove this newline.

eightbitraptor:
Closing this draft PR until I'm ready for feedback.

PR #7928, opened 2023-06-12
Reduce dependencies from test-bundled-gems

hsbt:
We should ignore `racc` dependency by `pry` for `test-bundled-gems` target.


hsbt:
I disabled to test `Prime_test.rb` in `rbs`. It should be removed from `rbs` because `prime` is bundled gems, not the standard library of Ruby.

PR #7932, opened 2023-06-12
Fix very large Struct classes crashing the GC

casperisfine:
Reproductible with:

```
./miniruby -e 'Struct.new(*100.times.map { |i| :"a_#{i}" }).new(*100.times.to_a)'
```

(still working on the fix, just pushing the failing test case for now)


casperisfine:
Closing in favor of https://github.com/ruby/ruby/pull/7933

PR #7933, opened 2023-06-12
When alloc size is too large, only allocate struct

eightbitraptor:


peterzhu2118 (commented):
Can you add a test for this? I think we can extend `test_morethan10members`.

test/ruby/test_struct.rb
```diff
@@ -41,6 +41,15 @@ def test_morethan10members
     end
   end
 
+  def test_larger_than_largest_pool
+    count = (GC::INTERNAL_CONSTANTS[:SIZE_POOL_COUNT] - 1) * GC::INTERNAL_CONSTANTS[:RVALUE_SIZE]
```
peterzhu2118, line 45: This calculation is not correct, it should be:

```suggestion
    count = ((2**(GC::INTERNAL_CONSTANTS[:SIZE_POOL_COUNT] - 1)) * GC::INTERNAL_CONSTANTS[:RVALUE_SIZE]) / RbConfig::SIZEOF["void*"]
```

But I'm not sure if we should be using all this GC internal values in these tests. We could just change `test_morethan10members` to test a much larger size.

peterzhu2118 (approved):

PR #7440, opened 2023-03-03
Implement embedded TypedData objects 

peterzhu2118:
This PR adds a new flag RUBY_TYPED_EMBEDDABLE that allows the data of a TypedData object to be embedded after the object itself. This will improve cache locality and allow us to save the 8 byte data pointer.

This PR also implements this feature on Time objects, which drops the total size of a Time object from 86 bytes to 80 bytes.

This PR is based on top of #7438.

Co-Authored-By: @byroot

PR #7938, opened 2023-06-13
Store object age in a bitmap

eightbitraptor:
Previously 2 bits of the flags on each RVALUE are reserved to store the number of GC cycles that each object has survived. This commit introduces a new bit array on the heap page, called age_bits, to store that information instead.

This patch still reserves one of the age bits in the flags (the old FL_PROMOTED0 bit, now renamed FL_PROMOTED).

This is set to 0 for young objects and 1 for old objects, and is used as a performance optimisation for the write barrier. Fetching the age_bits from the heap page and doing the required math to calculate if the object was old or not would slow down the write barrier. So we keep this bit synced in the flags for fast access.


gc.c
```diff
@@ -1723,39 +1756,34 @@ RVALUE_OLD_UNCOLLECTIBLE_SET(rb_objspace_t *objspace, VALUE obj)
...
 static inline void
-RVALUE_DEMOTE_RAW(rb_objspace_t *objspace, VALUE obj)
+RVALUE_AGE_SET_CANDIDATE(rb_objspace_t *objspace, VALUE obj)
 {
-    RBASIC(obj)->flags = RVALUE_FLAGS_AGE_SET(RBASIC(obj)->flags, 0);
-    CLEAR_IN_BITMAP(GET_HEAP_UNCOLLECTIBLE_BITS(obj), obj);
+    check_rvalue_consistency(obj);
+    GC_ASSERT(!RVALUE_OLD_P(obj));
+    RVALUE_AGE_SET(obj, RVALUE_OLD_AGE - 1);
+    FL_UNSET(obj, RUBY_FL_PROMOTED);
```
peterzhu2118, line 1785: Doesn't `RVALUE_AGE_SET` unset the `RUBY_FL_PROMOTED` already?

gc.c
```diff
@@ -1777,19 +1806,13 @@ RVALUE_DEMOTE(rb_objspace_t *objspace, VALUE obj)
...
 static inline void
 RVALUE_AGE_RESET(VALUE obj)
 {
     check_rvalue_consistency(obj);
-    GC_ASSERT(!RVALUE_OLD_P(obj));
-
-    RVALUE_AGE_RESET_RAW(obj);
+    bits_t *bits = GET_HEAP_PAGE(obj)->age_bits;
+    bits[RVALUE_AGE_BITMAP_INDEX(NUM_IN_PAGE(obj))] &= ~RVALUE_AGE_MASK_BITS(NUM_IN_PAGE(obj));
+    FL_UNSET(obj, RUBY_FL_PROMOTED);
```
peterzhu2118, line 1815: Can we call `RVALUE_AGE_SET(obj, 0)` here?

gc.c
```diff
@@ -2505,6 +2530,14 @@ newobj_init(VALUE klass, VALUE flags, int wb_protected, rb_objspace_t *objspace,
...
     *((VALUE *)&p->as.basic.klass) = klass;
 
+    int t = flags & RUBY_T_MASK;
+    if (t == T_CLASS || t == T_MODULE || t == T_ICLASS) {
+        RVALUE_AGE_SET_CANDIDATE(objspace, obj);
+    }
+    else {
+        RVALUE_AGE_SET(obj, 0);
+    }
+
```
peterzhu2118, line 2540: Do we need this else clause? The age should have been set to 0 when we put it on the freelist right?

gc.c
```diff
@@ -2521,12 +2554,13 @@ newobj_init(VALUE klass, VALUE flags, int wb_protected, rb_objspace_t *objspace,
         GC_ASSERT(RVALUE_OLD_P(obj) == FALSE);
         GC_ASSERT(RVALUE_WB_UNPROTECTED(obj) == FALSE);
 
-        if (flags & FL_PROMOTED1) {
-            if (RVALUE_AGE(obj) != 2) rb_bug("newobj: %s of age (%d) != 2.", obj_info(obj), RVALUE_AGE(obj));
+        if (flags & RUBY_FL_PROMOTED) {
+            if (RVALUE_AGE_GET(obj) != 2) rb_bug("newobj: %s of age (%d) != 2.", obj_info(obj), RVALUE_AGE_GET(obj));
```
peterzhu2118, line 2558: This assertion isn't correct. The original assertion was checking that if `FL_PROMOTED1` was set then the age must be 2 (the object is a candidate).

vm.c
```diff
@@ -3667,6 +3667,10 @@ Init_VM(void)
     fcore = rb_class_new(rb_cBasicObject);
     rb_set_class_path(fcore, rb_cRubyVM, "FrozenCore");
     RBASIC(fcore)->flags = T_ICLASS;
+    // Classes are allocated pre-tenured, we're overwriting this by manually assigning
+    // flags, so we reset the age in order to make sure the promoted bit and the age
+    // bits are in sync.
+    //rb_reset_age(fcore);
```
peterzhu2118, line 3673: This can be removed right?

gc.c
```diff
@@ -1011,6 +1014,48 @@ asan_unlock_freelist(struct heap_page *page)
 
 #define GC_SWEEP_PAGES_FREEABLE_PER_STEP 3
 
+#define RVALUE_AGE_BITMAP_INDEX(n)  ((n) / (BITS_BITLENGTH / RVALUE_AGE_BIT_COUNT))
+#define RVALUE_AGE_BITMAP_OFFSET(n) ((n) % (BITS_BITLENGTH / RVALUE_AGE_BIT_COUNT) << 1)
+#define RVALUE_AGE_BITMAP_BIT(n)    ((bits_t)1 << (n))
```
peterzhu2118, line 1019: This macro doesn't appear to be used.

gc.c
eightbitraptor: Reading through this part of the code again. I don't think these assertions have much value now that we're pre-tenuring classes in `newobj_init`. I'll delete.

maximecb:
> This is set to 0 for young objects and 1 for old objects, and is used as a performance optimisation for the write barrier. Fetching the age_bits from the heap page and doing the required math to calculate if the object was old or not would slow down the write barrier. So we keep this bit synced in the flags for fast access.

Given that accessing these bits from the heap adds extra performance overhead, can you provide more background as to why these bits are being maintained in a separate location in addition to the flag bits?

vm.c
peterzhu2118: I opened #7940 which should solve this issue.

eightbitraptor: Thanks for this @peterzhu2118 - Now that #7940 has been merged I've rebased so this is no longer an issue.

eightbitraptor:
> > This is set to 0 for young objects and 1 for old objects, and is used as a performance optimisation for the write barrier. Fetching the age_bits from the heap page and doing the required math to calculate if the object was old or not would slow down the write barrier. So we keep this bit synced in the flags for fast access.
> 
> Given that accessing these bits from the heap adds extra performance overhead, can you provide more background as to why these bits are being maintained in a separate location in addition to the flag bits?

@maximecb @byroot @peterzhu2118 and I discussed this outside of Github, but for context I'll summarise here:

We're not maintaining these bits in two separate places. The age of the object is now only being stored in the bitmap attached to a heap page, and the only part of the code that should have to access or modify this bitmap is the age related GC code. This data should be ignored by the rest of the VM.

The `FL_PROMOTED` bit is being used by the write barrier to decide whether or not it needs to take action for the object.

Currently  `FL_PROMOTED` is being used as a proxy for whether the old object is in the Young or the Old gen, because that's what the write barrier cares about (because of the Old object, Young reference issue); and so superficially, this can be seen as data duplication. We could use the age bits directly from within the write barrier to determine what generation the object is in, but we chose not to do this for a couple of reasons.

1. Performance - the object is passed into the write barrier function - we don't need to do any extra lookups to check the flags. If we were to look at the age bits, we'd have to find them by calculating where the objects heap page is, and then load them in. This would be much slower.
2. It would break encapsulation of the GC. The main motivation behind this PR is to get some GC related concerns out of other parts of the VM and back into the GC. Requiring the write barrier to know about age bits, and which ages correspond to which generation would break this encapsulation.

Conceptually, `FL_PROMOTED` is now just a write barrier trigger, which we could use for any reason if we wanted to trigger a write barrier in the future.

This approach also does not negatively impact performance, as can be seen by the benchmark results on the [bug tracker ticket](https://bugs.ruby-lang.org/issues/19729)

gc.c
```diff
@@ -939,6 +939,9 @@ static const bool HEAP_PAGE_ALLOC_USE_MMAP = false;
 static bool heap_page_alloc_use_mmap;
 #endif
 
+#define RVALUE_AGE_BIT_COUNT 2
+#define RVALUE_AGE_BIT_MASK (~((bits_t)-1L << RVALUE_AGE_BIT_COUNT))
```
peterzhu2118, line 943: This relies on the fact that most implementations use 2's complement for negative integers. The C standard does not enforce this so it would not work on implementations that don't use it.

From paragraph 2 of section 6.2.6.2 of the [C spec](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf):

> If the sign bit is one, the value shall be modified in one of the following ways:
> — the corresponding value with sign bit 0 is negated (sign and magnitude);
> — the sign bit has the value −(2N ) (two’s complement);
> — the sign bit has the value −(2N − 1) (ones’ complement).

```suggestion
#define RVALUE_AGE_BIT_MASK ((1UL << RVALUE_AGE_BIT_COUNT) - 1)
```

gc.c
```diff
@@ -1011,6 +1015,44 @@ asan_unlock_freelist(struct heap_page *page)
 
 #define GC_SWEEP_PAGES_FREEABLE_PER_STEP 3
 
+#define RVALUE_AGE_BITMAP_INDEX(n)  ((n) / (BITS_BITLENGTH / RVALUE_AGE_BIT_COUNT))
+#define RVALUE_AGE_BITMAP_OFFSET(n) ((n) % (BITS_BITLENGTH / RVALUE_AGE_BIT_COUNT) << 1)
```
peterzhu2118, line 1019: I think we need to make this vary based on the value of `RVALUE_AGE_BIT_COUNT`.

```suggestion
#define RVALUE_AGE_BITMAP_OFFSET(n) (((n) % (BITS_BITLENGTH / RVALUE_AGE_BIT_COUNT)) * RVALUE_AGE_BIT_COUNT)
```

gc.c
```diff
@@ -1011,6 +1015,44 @@ asan_unlock_freelist(struct heap_page *page)
...
+
+static void RVALUE_AGE_RESET(VALUE obj);
+static void
+RVALUE_AGE_SET(VALUE obj, int age)
+{
+    RUBY_ASSERT(age <= RVALUE_OLD_AGE);
+    bits_t *age_bits = GET_HEAP_PAGE(obj)->age_bits;
+    // clear the bits
+    age_bits[RVALUE_AGE_BITMAP_INDEX(NUM_IN_PAGE(obj))] &=
+        ~(RVALUE_AGE_BIT_MASK << (NUM_IN_PAGE(obj) * RVALUE_AGE_BIT_COUNT));
```
peterzhu2118, line 1038: Don't we need to shift by `RVALUE_AGE_BITMAP_OFFSET` here? `NUM_IN_PAGE` could return a value that's much larger than 64 when it's in a separate bit plane, which will cause the left shift to result in 0.

```suggestion
        ~(RVALUE_AGE_BIT_MASK << RVALUE_AGE_BITMAP_OFFSET(NUM_IN_PAGE(obj)));
```

gc.c
```diff
@@ -1011,6 +1015,44 @@ asan_unlock_freelist(struct heap_page *page)
...
+        RB_FL_UNSET_RAW(obj, RUBY_FL_PROMOTED);
+    }
+}
+
+void
+rb_reset_age(VALUE obj)
+{
+    RVALUE_AGE_RESET(obj);
+}
+
```
peterzhu2118, line 1055: We don't need this function anymore right?

gc.c
```diff
@@ -1723,39 +1744,34 @@ RVALUE_OLD_UNCOLLECTIBLE_SET(rb_objspace_t *objspace, VALUE obj)
...
 static inline void
-RVALUE_DEMOTE_RAW(rb_objspace_t *objspace, VALUE obj)
+RVALUE_AGE_SET_CANDIDATE(rb_objspace_t *objspace, VALUE obj)
 {
-    RBASIC(obj)->flags = RVALUE_FLAGS_AGE_SET(RBASIC(obj)->flags, 0);
-    CLEAR_IN_BITMAP(GET_HEAP_UNCOLLECTIBLE_BITS(obj), obj);
+    check_rvalue_consistency(obj);
+    GC_ASSERT(!RVALUE_OLD_P(obj));
+    RVALUE_AGE_SET(obj, RVALUE_OLD_AGE - 1);
+    FL_UNSET(obj, RUBY_FL_PROMOTED);
```
peterzhu2118, line 1773: Doesn't `RVALUE_AGE_SET` unset the `RUBY_FL_PROMOTED` already?

internal/gc.h
```diff
@@ -232,6 +232,8 @@ bool rb_gc_is_ptr_to_obj(void *ptr);
 VALUE rb_gc_id2ref_obj_tbl(VALUE objid);
 VALUE rb_define_finalizer_no_check(VALUE obj, VALUE block);
 
+void rb_reset_age(VALUE obj);
+
```
peterzhu2118, line 236: We should also be able to remove this.

peterzhu2118 (approved):
Looks good!

ioquatix:
Assuming we go ahead with this, can I humbly request one of the available flag bits for immutability?

byroot:
Only one of the two bits is being freed, and we plan to use it to embed T_DATA objects.

byroot:
> which predates `Data`

This has nothing to do with `Data.define`. And either way I don't think that's what should define what free flags can be used for.

>  I'm interested to know what "embed T_DATA" is

https://github.com/ruby/ruby/pull/7440

ioquatix:
> This has nothing to do with Data.define. And either way I don't think that's what should define what free flags can be used for.

Of course, but if it comes to it, it should at least be a discussion.

include/ruby/internal/fl_type.h
```diff
@@ -207,54 +206,30 @@ ruby_fl_type {
...
-     * collections.
-     *
-     * @internal
-     *
-     * But honestly, @shyouhei  doesn't think this flag should  be visible from
-     * 3rd parties.  It must be an implementation detail that they should never
-     * know.  Might better be hidden.
      */
-    RUBY_FL_PROMOTED     = RUBY_FL_PROMOTED0 | RUBY_FL_PROMOTED1,
+    RUBY_FL_UNUSED    = (1<<6),
```
ko1, line 232: how about to use `FL_UNUSED6` to avoid conflict from other bits?

PR #7940, opened 2023-06-13
Directly allocate FrozenCore as an ICLASS

peterzhu2118:
It's a bad idea to overwrite the flags as the garbage collector may have set other flags.

### DIFF
--- a/class.c
+++ b/class.c
@@ -1130,9 +1130,15 @@ rb_define_module_id_under(VALUE outer, ID id)
 }
 
 VALUE
+rb_iclass_alloc(VALUE klass)
+{
+    return class_alloc(T_ICLASS, klass);
+}
+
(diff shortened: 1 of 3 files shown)


nobu:
clang and gcc don't seem to optimize global variable accesses.

peterzhu2118:
@nobu I'm not sure what you mean. I switched back to using a local variable and checked the compiled instructions using `objdump`, it doesn't look any different. I'm on clang on ARM on macOS. What kind of optimizations are you referring to?

nobu:
Sorry to be unclear.
x86_64 macOS clang 14.0.3 generates code to access the global variable for each time.

ko1:
BTW maybe frozen  core objects are not needed any more because we has builtin function feature.

PR #7944, opened 2023-06-14
[Bug #19728] Auto-generate unicode property docs

jaynetics:
https://bugs.ruby-lang.org/issues/19728

This PR should probably wait for #7923 to be completed.

The main open question to me is whether there is a better way to categorize the properties, or to name or order the sections? The current choices are based on https://github.com/k-takata/Onigmo/blob/master/doc/UnicodeProps.txt

### DIFF
--- a/tool/enc-unicode.rb
+++ b/tool/enc-unicode.rb
@@ -269,23 +269,12 @@ def parse_block(data)
   blocks << constname
 end
 
-# shim for Ruby 1.8
-unless {}.respond_to?(:key)
-  class Hash
-    alias key index
-  end
-end
-
 $const_cache = {}
 # make_const(property, pairs, name): Prints a 'static const' structure for a
 # given property, group of paired codepoints, and a human-friendly name for
 # the group
 def make_const(prop, data, name)
-  if name.empty?
-    puts "\n/* '#{prop}' */"
-  else
-    puts "\n/* '#{prop}': #{name} */"
-  end
+  puts "\n/* '#{prop}': #{name} */" # comment used to generate documentation
   if origprop = $const_cache.key(data)
     puts "#define CR_#{prop} CR_#{origprop}"
   else
@@ -437,8 +426,6 @@ define_posix_props(data)
 POSIX_NAMES.each do |name|
   if name == 'XPosixPunct'
     make_const(name, data[name], "[[:Punct:]]")
-  elsif name == 'Punct'
-    make_const(name, data[name], "")
   else
     make_const(name, data[name], "[[:#{name}:]]")
   end
--- a/common.mk
+++ b/common.mk
@@ -372,7 +372,7 @@ ruby.imp: $(COMMONOBJS)
 	sort -u -o $@
 
 install: install-$(INSTALLDOC)
-docs: $(DOCTARGETS)
+docs: srcs-doc $(DOCTARGETS)
 pkgconfig-data: $(ruby_pc)
 $(ruby_pc): $(srcdir)/template/ruby.pc.in config.status
 
@@ -566,15 +566,15 @@ do-install-dbg: $(PROGRAM) pre-install-dbg
 post-install-dbg::
 	@$(NULLCMD)
 
-rdoc: PHONY main
+rdoc: PHONY main srcs-doc
 	@echo Generating RDoc documentation
 	$(Q) $(RDOC) --ri --op "$(RDOCOUT)" $(RDOC_GEN_OPTS) $(RDOCFLAGS) "$(srcdir)"
 
-html: PHONY main
+html: PHONY main srcs-doc
 	@echo Generating RDoc HTML files
 	$(Q) $(RDOC) --op "$(HTMLOUT)" $(RDOC_GEN_OPTS) $(RDOCFLAGS) "$(srcdir)"
 
-rdoc-coverage: PHONY main
+rdoc-coverage: PHONY main srcs-doc
 	@echo Generating RDoc coverage report
 	$(Q) $(RDOC) --quiet -C $(RDOCFLAGS) "$(srcdir)"
 
@@ -1084,7 +1084,7 @@ common-srcs: $(srcs_vpath)parse.c $(srcs_vpath)lex.c $(srcs_vpath)enc/trans/newl
 
 missing-srcs: $(srcdir)/missing/des_tables.c
 
-srcs: common-srcs missing-srcs srcs-enc
+srcs: common-srcs missing-srcs srcs-enc srcs-doc
 
 EXT_SRCS = $(srcdir)/ext/ripper/ripper.c \
 	   $(srcdir)/ext/rbconfig/sizeof/sizes.c \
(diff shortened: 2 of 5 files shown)


tool/enc-unicode.rb
```diff
@@ -269,23 +269,12 @@ def parse_block(data)
...
 # make_const(property, pairs, name): Prints a 'static const' structure for a
 # given property, group of paired codepoints, and a human-friendly name for
 # the group
 def make_const(prop, data, name)
-  if name.empty?
-    puts "\n/* '#{prop}' */"
-  else
-    puts "\n/* '#{prop}': #{name} */"
-  end
+  puts "\n/* '#{prop}': #{name} */"
```
jaynetics, line 277: `Punct` was the only case for which an empty "name" was given and thus a different type of code comment rendered

BurdetteLamar:
Outstanding, @jaynetics!  And thanks for folding in the aliases.

nobu:
I think that doc should be generated separately, using erb.

```diff
diff --git c/common.mk i/common.mk
index 212ab18bb79..e76a51f2b72 100644
--- c/common.mk
+++ i/common.mk
@@ -1660,6 +1660,11 @@ $(UNICODE_HDR_DIR)/name2ctype.h:
 		$(UNICODE_SRC_DATA_DIR) $(UNICODE_SRC_EMOJI_DATA_DIR) > $@.new
 	$(MV) $@.new $@
 
+$(srcdir)/doc/regexp/unicode_properties.rdoc: $(UNICODE_HDR_DIR)/name2ctype.h
+	$(Q) $(BASERUBY) $(tooldir)/generic_erb.rb -o $@ \
+	  $(srcdir)/doc/regexp/unicode_properties.rdoc \
+	  $(UNICODE_HDR_DIR)/name2ctype.h $(UNICODE_SRC_DATA_DIR)
+
 # the next non-comment line was:
 # $(UNICODE_HDR_DIR)/casefold.h: $(tooldir)/enc-case-folding.rb \
 # but was changed to make sure CI works on systems that don't have gperf
diff --git c/tool/enc-unicode-docs.rb i/template/unicode_properties.rdoc.tmpl
similarity index 55%
rename from tool/enc-unicode-docs.rb
rename to template/unicode_properties.rdoc.tmpl
index 2d1c9ac34c4..6fb5bcbbaa5 100755
--- c/tool/enc-unicode-docs.rb
+++ i/template/unicode_properties.rdoc.tmpl
@@ -1,22 +1,19 @@
-#!/usr/bin/env ruby
+== \Regexps Based on Unicode Properties
 
-# Generate a documentation file for the unicode properties generated by ./enc-unicode.rb.
+The properties shown here are those currently supported in Ruby.
+Older versions may not support all of these.
+<%
+# Generate a documentation file for the unicode properties.
 #
 # Usage:
 #
-# This script is usually called automatically by ./enc-unicode.rb.
-#
 # To run it individually, get PropertyAliases.txt, PropertyValueAliases.txt from unicode.org
-# (http://unicode.org/Public/UNIDATA/) and run `tool/enc-unicode-doc.rb data_dir`.
+# (http://unicode.org/Public/UNIDATA/) and run
+# ```
+# ruby tool/generic_erb.rb template/unicode_properties.rdoc.tmpl data_dir name2ctype.h
+# ```
 
-data_dir = ARGV[0]&.tap { |d| Dir.exist?(d) } || abort("Usage: #{$0} data_directory")
-
-doc_introduction = <<~'EOS'
-  == \Regexps Based on Unicode Properties
-
-  The properties shown here are those currently supported in Ruby.
-  Older versions may not support all of these.
-EOS
+data_dir = ARGV.shift&.tap { |d| Dir.exist?(d) } || abort("Usage: #{$0} data_directory [name2ctype.h]")
 
 # Map group names, given as last argument to #make_const in enc-unicode.rb,
 # to sections in the doc. The order in this hash controls the order in the doc.
@@ -39,9 +36,9 @@
   File.read(File.join(data_dir, 'PropertyValueAliases.txt')).scan(/^(?:gc|sc)\s*; (\w+)\s*; (\w+)/)
 ).to_h
 
-header_data = File.read(Dir["#{__dir__}/../enc/unicode/**/name2ctype.h"].last)
 props_by_section = {}
-header_data.scan(/'([^']+)': (.+) \*/).each do |prop, name|
+ARGF.each_line do |line|
+  next unless /'(?<prop>[^']+)': (?<name>.+) \*/ =~ line
   next if prop == 'NEWLINE' # ignore custom internal prop
 
   section = map.find { |k, v| k === name }&.last || warn("no doc section for #{name}")
@@ -50,16 +47,11 @@
   long_prop_name = aliases[prop] || prop
   (props_by_section[section] ||= []) << long_prop_name
 end
+map.each_value do |section| -%>
 
-File.open("#{__dir__}/../doc/regexp/unicode_properties.rdoc", 'w') do |f|
-  f.puts doc_introduction
+=== <%=section%>
 
-  map.each_value do |section|
-    f.print "\n=== #{section}\n\n"
-
-    props_by_section[section].sort.each do |prop|
-      long, short = [prop, aliases.key(prop)]
-      f.puts "- #{[long, short].compact.uniq.map { |v| "<tt>\\p{#{v}}</tt>" }.join(', ')}"
-    end
-  end
-end
+%   props_by_section[section].sort.each do |prop|
+- <%= [prop, aliases.key(prop)].compact.uniq.map { |v| "<tt>\\p{#{v}}</tt>" }.join(', ') %>
+%   end
+% end
diff --git c/tool/enc-unicode.rb i/tool/enc-unicode.rb
index 10e4d455850..581d1a7b823 100755
--- c/tool/enc-unicode.rb
+++ i/tool/enc-unicode.rb
@@ -615,6 +615,4 @@
       }
     }
   }
-
-  load "#{__dir__}/enc-unicode-docs.rb"
 end
```

jaynetics:
@nobu thank you for the feedback - i've applied your suggestions and can confirm that `$ make docs` still works and creates the rdoc file.

tool/enc-unicode.rb
```diff
@@ -269,23 +269,12 @@ def parse_block(data)
...
-    alias key index
-  end
-end
-
 $const_cache = {}
 # make_const(property, pairs, name): Prints a 'static const' structure for a
 # given property, group of paired codepoints, and a human-friendly name for
 # the group
 def make_const(prop, data, name)
-  if name.empty?
```
jaynetics, line 284: `Punct` was the only case for which an empty "name" was given and thus a different type of code comment rendered

jaynetics:
hmm, appveyor fails with 

> C:/projects/ruby/tool/lib/output.rb:42:in `binwrite': Permission denied @ rb_sysopen - ../doc/regexp/unicode_properties.rdoc (Errno::EACCES)

maybe this would work?

```diff
diff --git a/common.mk b/common.mk
index bdd0ce5a20..952e2fb454 100644
--- a/common.mk
+++ b/common.mk
@@ -1662,9 +1662,10 @@ $(UNICODE_HDR_DIR)/name2ctype.h:

 srcs-doc: $(srcdir)/doc/regexp/unicode_properties.rdoc
 $(srcdir)/doc/regexp/unicode_properties.rdoc: $(UNICODE_HDR_DIR)/name2ctype.h $(UNICODE_PROPERTY_FILES)
-       $(Q) $(BOOTSTRAPRUBY) $(tooldir)/generic_erb.rb -o $@ \
+       $(Q) $(BOOTSTRAPRUBY) $(tooldir)/generic_erb.rb -o $@.new \
                $(srcdir)/template/unicode_properties.rdoc.tmpl \
                $(UNICODE_SRC_DATA_DIR) $(UNICODE_HDR_DIR)/name2ctype.h
+       $(MV) $@.new $@
```

nobu:
`-c` option may workaround it.
This option let the tool not to overwrite the target file unless its content about to change.
It will fail if something changes, but we’ll need to notice it anyway.

nobu:
Is it intentional to change the already merged unicode_properties.rdoc?

jaynetics:
> Is it intentional to change the already merged unicode_properties.rdoc?

@nobu yes! The intended changes are:

- put the less commonly used sections (e.g. derived ages) further down
- sort the props within each section alphabetically
- list long and short forms of a single prop in-line, not as distinct items (e.g. `currency_symbol` and `sc`) 
- add all missing props (the previous doc was, i think, based on the Onigmo repo, still at at unicode version 12.1)
- fix some minor typos (such as double space in L384)

One thing i've done by accident is ommitting the surrounding slashes for each prop. E.g. it now lists `\p{ASCII}` instead of `/\p{ASCII}/`. I'm not sure what the best representation is. The doc is linked from a section where it is already explained how to use these properties, so maybe it would even suffice to list just the names, e.g. just `ASCII`.

nobu:
@BurdetteLamar What do you think about the change?

BurdetteLamar:
> @BurdetteLamar What do you think about the change?

I like it.  I think the organizaytion is good.

PR #7953, opened 2023-06-17
Add support for UUID version 7

nevans:
Although the specification for UUIDv7 is still in draft, the UUIDv7 algorithm has been relatively stable as it progresses to completion.

Version 7 UUIDs can be very useful, because they are lexographically sortable, which can improve e.g: database index locality.  See section 6.10 of the draft specification for further explanation:

  https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/

Ruby issue: https://bugs.ruby-lang.org/issues/19735

### DIFF
--- a/lib/random/formatter.rb
+++ b/lib/random/formatter.rb
@@ -174,6 +174,113 @@ module Random::Formatter
     "%08x-%04x-%04x-%04x-%04x%08x" % ary
   end
 
+  alias uuid_v4 uuid
+
+  # Generate a random v7 UUID (Universally Unique IDentifier).
+  #
+  #   require 'random/formatter'
+  #
+  #   Random.uuid_v7 # => "0188d4c3-16fe-7f96-85c7-242a7aa58f1e"
+  #   Random.uuid_v7 # => "0188d4c3-16fe-744f-86af-38fa04c62bb5"
+  #   Random.uuid_v7 # => "0188d4c3-16fe-764f-b049-c204ce0afa23"
+  #   Random.uuid_v7 # => "0188d4c3-16fe-7085-b14f-ef6415dc6f31"
+  #   #                    |<--sorted-->| |<----- random ---->|
+  #
+  #   # or
+  #   prng = Random.new
+  #   prng.uuid_v7 # => "0188ca51-5e72-7950-a11d-def7ff977c98"
+  #
+  # The version 7 UUID starts with the least significant 48 bits of a 64 bit
+  # Unix timestamp (milliseconds since the epoch) and fills the remaining bits
+  # with random data, excluding the version and variant bits.
+  #
+  # This allows version 7 UUIDs to be sorted by creation time.  Time ordered
+  # UUIDs can be used for better database index locality of newly inserted
+  # records, which may have a significant performance benefit compared to random
+  # data inserts.
+  #
+  # The result contains 74 random bits (9.25 random bytes).
+  #
+  # See draft-ietf-uuidrev-rfc4122bis[https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/]
+  # for details of UUIDv7.
+  #
+  # ==== Monotonicity
+  #
+  # By default, UUIDv7 has millisecond precision and multiple UUIDs created
+  # within the same millisecond are not time-ordered.  To create UUIDs that are
+  # time-ordered with sub-millisecond precision, up to 12 bits of additional
+  # clock precision may added with +extra_timestamp_bit_count+.  The extra
+  # timestamp precision comes at the expense of random bits.  Setting
+  # <tt>extra_timestamp_bit_count: 12</tt> provides ~244ns of precision,
+  # and only 62 random bits (7.75 random bytes).
+  #
+  #   prng = Random.new
+  #   Array.new(4) { prng.uuid_v7(extra_timestamp_bit_count: 12) }
+  #   # =>
+  #   ["0188d4c7-13da-74f9-8b53-22a786ffdd5a",
+  #    "0188d4c7-13da-753b-83a5-7fb9b2afaeea",
+  #    "0188d4c7-13da-754a-88ea-ac0baeedd8db",
+  #    "0188d4c7-13da-7557-83e1-7cad9cda0d8d"]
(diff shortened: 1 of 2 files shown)


lib/random/formatter.rb
```diff
@@ -174,6 +174,47 @@ def uuid
...
+  # records, which may have a significant performance benefit compared to random
+  # data inserts.
+  #
+  # The result contains 74 random bits (9.25 random bytes).
+  #
+  # See draft-ietf-uuidrev-rfc4122bis[https://datatracker.ietf.org/doc/draft-ietf-uuidrev-rfc4122bis/]
+  # for details of UUID.
+  #
+  def uuid_v7
+    unix_ts_ms = Process.clock_gettime(Process::CLOCK_REALTIME, :millisecond)
```
slewsys, line 206: For the purpose of sorting, might CLOCK_MONOTONIC be preferable?

nevans: > For the purpose of sorting, might CLOCK_MONOTONIC be preferable?

Not if we stick to the (draft) RFC:
> UUID version 7 features a time-ordered value field derived from the widely implemented and well known Unix Epoch timestamp source, the number of milliseconds since midnight 1 Jan 1970 UTC, leap seconds excluded.

The problem is that `CLOCK_MONOTONIC` isn't based on the "wall clock" at all.  On Linux systems it's the time since the system was booted, not including time when the system is suspended.  So, while it's great for timers on a single host, there's no direct relationship to the monotonic time from one host to another, or even from one system boot to another.  So the spec uses the Unix Epoch timestamp.

But the spec _does_ allow optional alterations to UUIDv7s for improving monotonicity, by replacing random bits with:
> * An OPTIONAL sub-millisecond timestamp fraction (12 bits at maximum) as per [Section 6.2](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-07.html#monotonicity_counters) (Method 3).
> * An OPTIONAL carefully seeded counter as per [Section 6.2](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-07.html#monotonicity_counters) (Method 1 or 2).

Unfortunately, both of those options reduce randomness.  Randomness is often more important than monotonicity: the most commonly used UUID version, version 4, doesn't provide *any* monotonicity and 122 bits of randomness.  I think the default should follow the RFC and err on the side of more randomness and less monotonicity.

[Section 6.2](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-07.html#monotonicity_counters) goes into more details for a variety of techniques to improve monotonicity, but most of them come with significant increase in complexity, and some with other significant tradeoffs (e.g: reliance on global state, increased guessability).  And distributed monotonicity isn't really possible for UUIDv7.

The version in the PR right now is the simplest version, and it provides "good enough" monotonicity for most purposes.  If the primary use-case is better performance through better DB index locality, millisecond precision for single node and normal clock skew between different nodes is probably "good enough".

But, should I should probably update the rdoc to note that the UUIDs are not guaranteed to be monotonic beyond millisecond precision or clock skew between hosts?

nevans: And I think a kwarg to add extra_timestamp_bits _may_ be an acceptable increase in complexity.  Here's what it might look like if we expand it to include subms precision:

```ruby
def uuid_v7(extra_timestamp_bits: nil)
  case extra_timestamp_bits
  in nil | false | 0
    ts_ms = Process.clock_gettime(Process::CLOCK_REALTIME, :millisecond)
    rand_a, rand_b1, rand_b2, rand_b3 = Random.bytes(10).unpack("nnnN")
    rand_a &= 0x0fff
  in (1..12)
    ts_ms, ts_ns = Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
      .divmod(1_000_000)
    subms = (ts_ns.to_f / 1_000_000 * (2 ** extra_timestamp_bits))
      .floor
    rand_a, rand_b1, rand_b2, rand_b3 = Random.bytes(10).unpack("nnnN")
    rand_a_bitcount = 12 - extra_timestamp_bits
    rand_a_bitmask  = 2**rand_a_bitcount - 1
    rand_a = (subms << rand_a_bitcount) | (rand_a & rand_a_bitmask)
  in true
    return uuid_v7(extra_timestamp_bits: 12)
  end

  "%08x-%04x-7%03x-%04x-%04x%08x" % [
    (ts_ms & 0x0000_ffff_ffff_0000) >> 16,
    (ts_ms & 0x0000_0000_0000_ffff,
    rand_a,
    rand_b1,
    rand_b2,
    rand_b3
  ]
end
```

That would certainly need better tests than I've written for the PR as-is.  Is it worth it?

In my opinion, 12 more bits of timestamp precision should be good enough for all but the most strict use-cases.  That's enough precision that `Array.new(1000) { uuid_v7(extra_timestamp_bits: 12) }` was monotonically increasing when I tested it on my laptop.

Implementing any better guarantee of monotonicity comes with [too many options and tradeoffs](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-07.html#name-monotonicity-and-counters), and requires tracking global state.  If they truly need it, applications should handle that complexity themselves (or in special-purpose gems).  What do you think?

nevans: I just pushed an updated version with more details about monotonicity in the rdoc and better testing.  And, at the cost of some additional code complexity, I decided to add the +extra_timestamp_bit_count+ kwarg to the PR too. 🙂

nevans:
I just pushed an updated version with more details about monotonicity in the rdoc and better testing.  And, at the cost of some additional code complexity, I decided to add the +extra_timestamp_bit_count+ kwarg to the PR too. 🙂

shyouhei:
(Not against this feature itself but) https://github.com/ruby/securerandom is the upstream of the library these days, and it seems some discussion about v7 UUID is already ongoing there.

nevans:
> (Not against this feature itself but) https://github.com/ruby/securerandom is the upstream of the library these days, and it seems some discussion about v7 UUID is already ongoing there.

Ah... Thanks @shyouhei!

nevans:
I updated my PR and made it on the `securerandom` repo, here: https://github.com/ruby/securerandom/pull/19

PR #7956, opened 2023-06-18
Simplify parser_st.c

nobu:
IIUC, parser_st.o will be included in libruby (or libruby_parser) to provide the shim to other products.
That means there should be also st.c when compiling parser_st.c.


yui-knk:
Correct, parser_st is created for providing symbols same with st.c but not depending on ruby functions like `ruby_xmalloc`.

yui-knk (approved):

PR #7957, opened 2023-06-18
[Cirrus] Runs only for YJIT (provisional)

nobu:
Our tasks very often reach a concurrency limit on Cirrus-CI, and get delayed.
Submitting new tasks during the delay seems to make delay longer and longer.
So I think we should restrict the use of Cirrus, provisionally at least.
As YJIT needs ARM build to test code for ARM, use only for it.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,8 +21,11 @@ env:
 
 config_template: &CONFIG_TEMPLATE
   auto_cancellation: $CIRRUS_BRANCH != 'master'
-  only_if: $CIRRUS_CRON != ''
-  skip: "changesIncludeOnly('doc/**', '**.{md,rdoc,ronn,[1-8]}', '.document')"
+  only_if: changesInclude('.cirrus.yml', 'yjit.{c,h,rb}', 'yjit/**.{mk,rs,toml}')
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -21,8 +21,11 @@ env:
 
 config_template: &CONFIG_TEMPLATE
   auto_cancellation: $CIRRUS_BRANCH != 'master'
-  only_if: $CIRRUS_CRON != ''
-  skip: "changesIncludeOnly('doc/**', '**.{md,rdoc,ronn,[1-8]}', '.document')"
+  only_if: changesInclude('yjit.{c,h,rb}', 'yjit/**.{mk,rs,toml}')
```
k0kubun, line 24: This seems irrelevant to `task`. I think this should be something like https://github.com/ruby/ruby/pull/8019.

If the `task` is there only for YJIT, I think we should delete `task` and resurrect a clang-12 job for `task_yjit` so that it can test bindgen with libclang.

k0kubun: 🤔 I changed my mind. I think this is okay as is.

k0kubun (approved):

PR #7960, opened 2023-06-19
Add `IO::Buffer#cat` for concat two IO::Buffer instances

unasuke:
See also https://bugs.ruby-lang.org/issues/19737


ioquatix:
Thanks for working on this.

`cat` is probably not the right name for such a feature. Please let me consider it.

PR #7962, opened 2023-06-20
Stop `rb_ivar_foreach`

nobu:
Stop `rb_ivar_foreach` when callback returned `ST_STOP`.


unak (approved):
seems good. go ahead!

PR #6178, opened 2022-07-25
rb_str_resize: Only clear coderange on truncation

casperisfine:
If we are expanding the string or only stripping extra capacity then coderange won't change, so clearing it is wasteful.

cc @nirvdrum

As far as I can tell, this `CLEAR` was there from the beginning: https://github.com/ruby/ruby/commit/e41b84895efbf8d8bd80b5e88f52a1baf46f5da7

PR #7965, opened 2023-06-21
String#slice! should clear the coderange when truncating

casperisfine:
[[Bug #19739]](https://bugs.ruby-lang.org/issues/19739)

This bug was incidentally fixed in Ruby 3.2 via b0b9f7201acab05c2a3ad92c3043a1f01df3e17f but remains on 3.1 and older.

Ref: https://github.com/ruby/ruby/pull/6178


hsbt:
This PR has been merged at https://github.com/ruby/ruby/commit/7e926c4d8c2e785d174161b38a8a688b19ea4b99

PR #7977, opened 2023-06-24
Implement POC of "mutex reservation"

KJTsanaktsidis:
This implements a potential solution to Bug #19717, where it's possible for a resource protected by a condition variable & a mutex to ping-ping between two threads, whilst a third thread never gets it. This occurs if the resource is released & re-acquired in the same thread, because a thread which releases a mutex will almost always be able to re-acquire it afterwards unless it does some blocking work in between.

This implements a kind of "reservation" system for mutexes orchestrated with condition variables. A condition variable waiter recoreds the fact that it would like the mutex it released to be "reserved" for it on re-awakening. In that case, the thread with the mutex will not wake up any _other_ thread when it releases it, and no other thread will attempt to claim the mutex.

I'm sure this has bugs and probably stalls with nobody waiting for the mutex in some circumstances, but it's a start. This is just a prototype intended to facilitate some discussion in https://bugs.ruby-lang.org/issues/19717

### DIFF
--- a/thread_sync.c
+++ b/thread_sync.c
@@ -8,6 +8,8 @@ static VALUE rb_eClosedQueueError;
 /* Mutex */
 typedef struct rb_mutex_struct {
     rb_fiber_t *fiber;
+    /* VALUE, not rb_fiber_t, because it could be dead and we need to keep it not-GGC'd*/
+    VALUE reserved_for_fiber;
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
Superseded by https://github.com/ruby/ruby/pull/8331

PR #7313, opened 2023-02-15
Remove USE_RINCGC flag

eightbitraptor:
Ruby doesn't compile when USE_RINCGC is disabled. It's also not being tested in CI.

Let's remove it, and the underlying GC_ENABLE_INCREMENTAL_MARK flag.

PR #7982, opened 2023-06-27
Remove reference to USE_RINCGC

eightbitraptor:
This compile time flag was removed in https://github.com/ruby/ruby/pull/7313

This commit cleans up some related dead code.


nobu (approved):
It doesn't seem nice to expose the private compile-time flag.

PR #7990, opened 2023-06-28
Introduce script_lines function for refactor

S-H-GAMELINKS:
Introduce `script_lines_with` macro for refactor `script_lines_defined` and `script_lines_get` functions

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -149,22 +149,24 @@ literal_hash(VALUE a)
     return rb_iseq_cdhash_hash(a);
 }
 
-static int
-script_lines_defined(void)
+static ID
+script_lines(void)
 {
     ID script_lines;
     CONST_ID(script_lines, "SCRIPT_LINES__");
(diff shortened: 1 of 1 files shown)


nobu:
Isn't it better to extract a function returns `script_lines`?

parse.y
```diff
@@ -149,22 +149,24 @@ literal_hash(VALUE a)
     return rb_iseq_cdhash_hash(a);
 }
 
-static int
-script_lines_defined(void)
+static ID
+script_lines(void)
 {
-    ID script_lines;
+    ID script_lines; \
```
nobu, line 155:
```suggestion
    ID script_lines;
```

S-H-GAMELINKS: Thanks, fixed it in https://github.com/ruby/ruby/pull/7990/commits/83d67923f26bfa31c7e5502392cc943656ff7bb3

PR #7992, opened 2023-06-28
Increase memory leak test timeout

peterzhu2118:
The test times out on some platforms, so increase the timeout.

### DIFF
--- a/test/ripper/test_ripper.rb
+++ b/test/ripper/test_ripper.rb
@@ -143,7 +143,7 @@ end
 
   def test_no_memory_leak
     assert_no_memory_leak(%w(-rripper), "", "#{<<~'end;'}", rss: true)
-      10_000_000.times do
+      2_000_000.times do
         Ripper.parse("")
       end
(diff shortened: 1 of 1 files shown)


test/ripper/test_ripper.rb
```diff
@@ -142,7 +142,7 @@ def test_dedent_string
   end
 
   def test_no_memory_leak
-    assert_no_memory_leak(%w(-rripper), "", "#{<<~'end;'}", rss: true)
+    assert_no_memory_leak(%w(-rripper), "", "#{<<~'end;'}", rss: true, timeout: 60)
```
nobu, line 145: This test consumes 22 secs.
Repeating 10M times is needed?

peterzhu2118: I changed it to 2M times. At 1M times, the test doesn't work on my M1 MacBook Pro (but works on my x86 Linux machine). 2M times is the minimum for my M1.

PR #7998, opened 2023-06-29
Refactor extract apt-get

shyouhei:
Same as #7994, but for `apt-get` instead.

### DIFF
--- a/.github/actions/setup/ubuntu/action.yml
+++ b/.github/actions/setup/ubuntu/action.yml
@@ -1,0 +1,53 @@
+name: Setup ubuntu environment
+description: >-
+  At the beginning there was no way but to copy & paste `apt-get`
+  everywhere.  But now that we have composite actions, it seems better
+  merge them into one.
+
+inputs:
+  arch:
+    required: false
+    default: ''
+    description: >-
+      Architecture.  Because we run this on a GitHub-hosted runner
+      acceptable value for this input is very limited.
+
+outputs:
+  arch:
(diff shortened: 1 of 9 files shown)


.github/actions/setup/ubuntu/action.yml
```diff
@@ -0,0 +1,53 @@
...
+      run: |
+        set -x
+        ${arch:+sudo dpkg --add-architecture ${arch#:}}
+        sudo apt-get update -qq || :
+        sudo apt-get install --no-install-recommends -qq -y -o=Dpkg::Use-Pty=0 \
+        ${arch:+cross}build-essential${arch/:/-} \
+        libssl-dev${arch} libyaml-dev${arch} libreadline6-dev${arch} \
+        zlib1g-dev${arch} libncurses5-dev${arch} libffi-dev${arch} \
+        autoconf ruby
+        sudo apt-get install -qq -y pkg-config${arch} || :
```
nobu, line 53: Just today I was thinking of putting `pkg-config` in the above line too.

nobu (approved):

.github/actions/setup/ubuntu/action.yml
shyouhei: @nobu It seems `apt-get install pkg-config` is failing @ i386.  The `|| :` is important here.
https://github.com/ruby/ruby/actions/runs/5417548413/jobs/9848636867#step:4:496

PR #8003, opened 2023-06-29
Handle unterminated unicode escapes in regexps

jeremyevans:
This fixes an infinite loop possible after ec3542229b29ec93062e9d90e877ea29d3c19472. For \u{} escapes in regexps, skip validation in the parser, and rely on the regexp code to handle validation. This is necessary so that invalid unicode escapes in comments in extended regexps are allowed.

Fixes [Bug #19750]

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -7296,44 +7296,71 @@ tokadd_utf8(struct parser_params *p, rb_encoding **encp,
     if (regexp_literal) { tokadd(p, '\\'); tokadd(p, 'u'); }
 
     if (peek(p, open_brace)) {  /* handle \u{...} form */
-        const char *second = NULL;
-        int c, last = nextc(p);
-        if (p->lex.pcur >= p->lex.pend) goto unterminated;
-        while (ISSPACE(c = *p->lex.pcur) && ++p->lex.pcur < p->lex.pend);
-        while (c != close_brace) {
-            if (c == term) goto unterminated;
-            if (second == multiple_codepoints)
-                second = p->lex.pcur;
-            if (regexp_literal) tokadd(p, last);
-            if (!tokadd_codepoint(p, encp, regexp_literal, TRUE)) {
-                break;
+        if (regexp_literal && p->lex.strterm->u.literal.u1.func == str_regexp) {
+            /*
+             * Skip parsing validation code and copy bytes as-is until term or
+             * closing brace, in order to correctly handle extended regexps where
+             * invalid unicode escapes are allowed in comments. The regexp parser
+             * does its own validation and will catch any issues.
+             */
+            int c = *p->lex.pcur;
+            tokadd(p, c);
+            for (c = *++p->lex.pcur; p->lex.pcur < p->lex.pend; c = *++p->lex.pcur) {
+                if (c == close_brace) {
+                    tokadd(p, c);
+                    ++p->lex.pcur;
+                    break;
+                }
+                else if (c == term) {
+                    break;
+                }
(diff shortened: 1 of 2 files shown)


test/ruby/test_parse.rb
```diff
@@ -1052,6 +1052,22 @@ def test_yyerror_at_eol
     assert_syntax_error("    0b\n", /\^/)
   end
 
+  def test_unclosed_unicode_escape_at_eol_bug_19750
+    assert_separately([], "#{<<-"begin;"}\n#{<<~"end;"}")
```
nobu, line 1056:
```suggestion
    assert_separately([], "#{<<-"begin;"}\n#{<<~'end;'}")
```

and you can remove doubled backslashes.

test/ruby/test_parse.rb
```diff
@@ -1052,6 +1052,22 @@ def test_yyerror_at_eol
...
+    begin;
+      assert_syntax_error("\\/\\\\u", /too short escape sequence/)
+      assert_syntax_error("\\/\\\\u{", /unterminated regexp meets end of file/)
+      assert_syntax_error("\\/\\\\u{\\\\n", /invalid Unicode list/)
+      assert_syntax_error("\\/a#\\\\u{\\\\n/", /invalid Unicode list/)
+      re = eval("\\/a#\\\\u{\\n$/x")
+      assert_match(re, 'a')
+      assert_not_match(re, 'a#')
+      re = eval("\\/a#\\\\u\\n$/x")
```
nobu, line 1065: Also, are `/`s needed to escape?
```suggestion
      assert_syntax_error("/\\u", /too short escape sequence/)
      assert_syntax_error("/\\u{", /unterminated regexp meets end of file/)
      assert_syntax_error("/\\u{\\n", /invalid Unicode list/)
      assert_syntax_error("/a#\\u{\\n/", /invalid Unicode list/)
      re = eval("/a#\\u{\n$/x")
      assert_match(re, 'a')
      assert_not_match(re, 'a#')
      re = eval("/a#\\u\n$/x")
```

parse.y
```diff
@@ -7296,44 +7296,71 @@ tokadd_utf8(struct parser_params *p, rb_encoding **encp,
...
+            for (c = *++p->lex.pcur; p->lex.pcur < p->lex.pend; c = *++p->lex.pcur) {
+                if (c == close_brace) {
+                    tokadd(p, c);
+                    ++p->lex.pcur;
+                    break;
+                }
+                else if (c == term) {
+                    break;
+                }
+                if (c == '\\' && p->lex.pcur + 1 < p->lex.pend) {
```
nobu, line 7317: Byte-wise scan for `\` can cause unexpected behavior with some encoding, e.g. Shift_JIS.

parse.y
```diff
@@ -7296,44 +7296,71 @@ tokadd_utf8(struct parser_params *p, rb_encoding **encp,
...
+                    tokadd(p, c);
+                    c = *++p->lex.pcur;
+                }
+                tokadd(p, c);
```
nobu, line 7321:
```suggestion
                tokaddmbc(p, c);
```

parse.y
jeremyevans: Can you let me know if there is a better solution?

nobu: `tokadd_mbchar` is it, but forgot to mention the declaration.

jeremyevans: No problem, thanks for fixing this.  I'll push a commit for the declaration.

nobu (approved):

PR #8006, opened 2023-06-30
YJIT: Fix autosplat miscomp for blocks with optionals

XrXr:
When passing an array as the sole argument to `yield`, and the yieldee
takes more than 1 optional parameter, the array is expanded similar
to `*array` splat calls. This is called "autosplat" in
`setup_parameters_complex()`.

Previously, YJIT did not detect this autosplat condition. It passed the
array without expanding it, deviating from interpreter behavior.
Detect this conditon and refuse to compile it.

Fixes: Shopify/yjit#313

---
The `setup_parameters_complex()` logic, for reference:
https://github.com/ruby/ruby/blob/d49a92d036fb316757d7c4a12a2f9808875f35bf/vm_args.c#L643-L649

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1,3 +1,13 @@
+# Regression test for yielding with autosplat to block with
+# optional parameters. https://github.com/Shopify/yjit/issues/313
+assert_equal '[:a, :b, :a, :b]', %q{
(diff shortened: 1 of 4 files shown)


k0kubun (approved):

tomstuart:
Thank you Alan! ❤️

PR #8007, opened 2023-06-30
Manual YARP resync

jemmaissroff:
There were some silent failures when cherry-picking YARP. Force a re-sync of YARP.


lib/yarp/serialize.rb
```diff
@@ -9,34 +9,38 @@
...
-        @source = source.dup
+        @input = input.dup
         @serialized = serialized
         @io = io
 
         @constant_pool_offset = nil
         @constant_pool = nil
+
+        offsets = [0]
+        input.b.scan("\n") { offsets << $~.offset(0)[0] + 1 }
```
nobu, line 34:
```suggestion
        offsets = [0]
        input.b.scan("\n") { offsets << $~.end(0) }
```

or

```suggestion
        offsets = []
        input.b.scan(/^/) { offsets << $~.begin(0) }
```

PR #8008, opened 2023-06-30
ruby_3_2 backport for #8006

XrXr:
This is a backported version of https://github.com/ruby/ruby/pull/8006
The patch is different because the two branches have diverged.

Original bug report: https://github.com/Shopify/yjit/issues/313

---

YJIT: Fix autosplat miscomp for blocks with optionals

When passing an array as the sole argument to `yield`, and the yieldee takes more than 1 optional parameter, the array is expanded similar to `*array` splat calls. This is called "autosplat" in `setup_parameters_complex()`.

Previously, YJIT did not detect this autosplat condition. It passed the array without expanding it, deviating from interpreter behavior. Detect this conditon and refuse to compile it.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1,3 +1,13 @@
+# Regression test for yielding with autosplat to block with
+# optional parameters. https://github.com/Shopify/yjit/issues/313
+assert_equal '[:a, :b, :a, :b]', %q{
+  def yielder(arg) = yield(arg) + yield(arg)
+
+  yielder([:a, :b]) do |c = :c, d = :d|
+    [c, d]
+  end
+}
+
 assert_equal 'true', %q{
   # regression test for tracking type of locals for too long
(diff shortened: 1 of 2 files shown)


k0kubun (approved):

maximecb:
Everything passed except for something which seems to be a Cirrus CI issue, so I'm going to merge this.

XrXr:
Sorry, I forgot to mention that we're not supposed to merge changes to backport branches (though it's fine to make requests). The branch maintainer is given complete control over what lands. At least that's the policy last I checked.

hsbt:
@maximecb Don't merge without @nagachika approval at `ruby_3_2` branch.

nagachika:
I have reviewed the changeset.

Also, if the changesets in the master branch cannot be cleanly applied, creating patches for the stable branches is always welcome. I will handle PRs when I am mentioned or assigned.

PR #8010, opened 2023-07-01
Support `--backtrace-limit` in RUBYOPT, allow -1 as value, and add it to man page

tomstuart:
The `--backtrace-limit` option is not currently allowed to appear in the `RUBYOPT` environment variable, but that appears to be a mistake. Unlike other long options which are not allowed in `RUBYOPT` (e.g. `--copyright`, `--version`, `--dump` and `--help`) it does not cause the interpreter to terminate, and cannot cause harm if read from the environment. During the initial discussion about the `--backtrace-limit` feature, Matz [suggested](https://bugs.ruby-lang.org/issues/8661#note-27) that he expected `RUBYOPT` to allow this option.

The `--backtrace-limit` option should also allow a value of -1 since that is legitimate (and the default), and be documented on Ruby’s `man` page.

### DIFF
--- a/man/ruby.1
+++ b/man/ruby.1
@@ -271,6 +271,11 @@ In auto-split mode, Ruby executes
 .Dl $F = $_.split
 at beginning of each loop.
 .Pp
+.It Fl -backtrace-limit Ns = Ns Ar num
+Limits the maximum length of backtraces to
+.Ar num
+lines (default -1, meaning no limit).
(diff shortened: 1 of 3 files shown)


tomstuart:
I’ve rebased this onto #8068 and added @nobu’s [test](https://bugs.ruby-lang.org/issues/19752#note-6).

PR #8014, opened 2023-07-02
[DOC] Dir doc

BurdetteLamar:
The link at line 304 is fouled by the expression in line 302.  I don't know why, and I don't have a workaround.

### DIFF
--- a/dir.rb
+++ b/dir.rb
@@ -13,7 +13,7 @@
 #
 # == About the Examples
 #
-# Some examples on this page use this file tree:
+# Some examples on this page use this simple file tree:
 #
 #   example/
 #   ├── config.h
@@ -23,7 +23,8 @@
 #   │   └── song.rb
 #   └── main.rb
 #
-# Others use the file tree for the Ruby project itself.
+# Others use the file tree for the
+# {Ruby project itself}[https://github.com/ruby/ruby].
 #
 # == \Dir As \Array-Like
 #
@@ -37,12 +38,14 @@
 # A \Dir object is in some ways stream-like.
 #
 # The stream is initially open for reading,
-# but may be closed (method Dir#close),
+# but may be closed manually (using method #close),
 # and will be closed on block exit if created by Dir.open called with a block.
+# The closed stream may not be further manipulated,
+# and may not be reopened.
 #
-# The stream has a _position_, which is an index of an entry in the directory:
+# The stream has a _position_, which is the index of an entry in the directory:
 #
-# - The initial position is zero.
+# - The initial position is zero (before the first entry).
 # - \Method #tell (aliased as #pos) returns the position.
 # - \Method #pos= sets the position (but ignores a value outside the stream),
 #   and returns the position.
@@ -52,7 +55,7 @@
 #   if at end-of-stream, does not increment the position.
 # - \Method #rewind sets the position to zero.
 #
-# Example:
+# Examples (using the {simple file tree}[rdoc-ref:Dir@About+the+Examples]):
 #
 #   dir = Dir.new('example') # => #<Dir:example>
 #   dir.pos                  # => 0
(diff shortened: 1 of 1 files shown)


dir.rb
```diff
@@ -1,11 +1,85 @@
...
+# The closed stream may not be further manipulated,
+# and may not be reopened.
+#
+# The stream has a _position_, which is the index of an entry in the directory:
+#
+# - The initial position is zero (before the first entry).
+# - \Method #tell (aliased as #pos) returns the position.
+# - \Method #pos= sets the position (but ignores a value outside the stream),
+#   and returns the position.
+# - \Method #seek is like #pos, but returns +self+ (convenient for chaining).
```
peterzhu2118, line 52: I think `#seek` is more like `#pos=` right? It sets the position rather than gets the position.

dir.rb
```diff
@@ -108,114 +194,157 @@ def self.open(name, encoding: nil, &block)
...
+  #    if followed by  the slash character <tt>'/'</tt>:
+  #
+  #      Dir.glob('**/').take(3) # => ["basictest/", "benchmark/", "benchmark/gc/"]
+  #
+  #    If the string pattern contains other characters
+  #    or is not followed by a slash character,
+  #    it is equivalent to <tt>'*'</tt>.
+  #
+  # - <tt>'?'</tt> Matches any single character;
+  #   similar in meaning to regexp <tt>/.{1}/</tt>:
```
peterzhu2118, line 291: Doesn't the regex `/./` mean the same as `/.{1}/`?

dir.rb
```diff
@@ -108,114 +194,157 @@ def self.open(name, encoding: nil, &block)
...
   #
-  # Note that the pattern is not a regexp, it's closer to a shell glob.
-  # See File::fnmatch for the meaning of the +flags+ parameter.
-  # Case sensitivity depends on your system (+File::FNM_CASEFOLD+ is ignored).
+  #   Dir.glob('*') # => ["config.h", "lib", "main.rb"]
   #
-  # <code>*</code>::
-  #   Matches any file. Can be restricted by other values in the glob.
-  #   Equivalent to <code>/.*/mx</code> in regexp.
+  # with a block, calls the block with each of the _entry_names_
```
peterzhu2118, line 240: Nit: capitalization
 
```suggestion
  # With a block, calls the block with each of the _entry_names_
```

dir.rb
```diff
@@ -108,114 +194,157 @@ def self.open(name, encoding: nil, &block)
...
-  # The results which matched single wildcard or character set are sorted in
-  # binary ascending order, unless +false+ is given as the optional +sort+
-  # keyword argument.  The order of an Array of pattern strings and braces
-  # are preserved.
+  # With no block, returns array _entry_names_;
+  # example (using the {simple file tree}[rdoc-ref:Dir@About+the+Examples]):
   #
-  # Note that the pattern is not a regexp, it's closer to a shell glob.
-  # See File::fnmatch for the meaning of the +flags+ parameter.
-  # Case sensitivity depends on your system (+File::FNM_CASEFOLD+ is ignored).
```
peterzhu2118, line 153: We should talk about the possible `flags` found in `File::Constants`.

dir.rb
BurdetteLamar: Fixed.

dir.rb
BurdetteLamar: Fixed.

dir.rb
BurdetteLamar: Fixed.

dir.rb
BurdetteLamar: I had not known where FNM_* were.

peterzhu2118: I think we should list the flags under `File::Constants` that can be used in this method.

peterzhu2118: Ok I guess it's fair that this isn't causing a regression

BurdetteLamar: @peterzhu2118, I will not forget about your suggestion.

BurdetteLamar: @peterzhu2118, which flags in File::Constants apply for Dir.glob?  I tried a couple of unlikelies (TRUNC, CREAT) and got no error.

peterzhu2118: It looks like invalid flags just get silently ignored.

PR #8015, opened 2023-07-03
Fix IO::Buffer#get_string with negative offset

noteflakes:
Fixes the behaviour of IO::Buffer#get_string when a negative offset
is given, counting from the end of the buffer. If the given offset
is less than negative buffer size, it is normalized to 0.

https://bugs.ruby-lang.org/issues/19753

### DIFF
--- a/io_buffer.c
+++ b/io_buffer.c
@@ -2253,6 +2253,8 @@ io_buffer_copy(int argc, VALUE *argv, VALUE self)
  *
  *  Read a chunk or all of the buffer into a string, in the specified
  *  +encoding+. If no encoding is provided +Encoding::BINARY+ is used.
+ *  If the given offset is negative, it is regarded as counting backward
+ *  from the end of the buffer.
  *
  *    buffer = IO::Buffer.for('test')
  *    buffer.get_string
@@ -2261,6 +2263,10 @@ io_buffer_copy(int argc, VALUE *argv, VALUE self)
  *    # => "st"
  *    buffer.get_string(2, 1)
  *    # => "s"
+ *    buffer.get_string(-3)
+ *    # => "est"
+ *    buffer.get_string(-1000, 2)
+ *    # => "te"
  */
 static VALUE
 io_buffer_get_string(int argc, VALUE *argv, VALUE self)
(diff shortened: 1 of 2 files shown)


io_buffer.c
```diff
@@ -2279,7 +2285,12 @@ io_buffer_get_string(int argc, VALUE *argv, VALUE self)
     rb_encoding *encoding = rb_ascii8bit_encoding();
 
     if (argc >= 1) {
-        offset = NUM2SIZET(argv[0]);
+        long offset_int = NUM2LONG(argv[0]);
```
nobu, line 2288:
```suggestion
        ssize_t offset_int = NUM2SSIZET(argv[0]);
```

io_buffer.c
```diff
@@ -2279,7 +2285,12 @@ io_buffer_get_string(int argc, VALUE *argv, VALUE self)
...
+        if (offset_int < 0) {
+            offset_int = size + offset_int;
+            if (offset_int < 0) offset_int = 0;
```
nobu, line 2291: Shouldn't raise `ArgumentError`?

noteflakes: As I wrote in the issue (https://bugs.ruby-lang.org/issues/19753), I think it should behave like `String#[]` or `Array#[]`, so negative offset should be acceptable, but counting from the end of the buffer.

noteflakes:
Hey @ioquatix, could you take a look and let me know what you think?

ioquatix (changes requested):
What is the range of `ssize_t`?

noteflakes:
> What is the range of `ssize_t`?

It's a `long int`. `SSIZE_MAX` is `0x7fffffffffffffffL`.

https://man.archlinux.org/man/core/man-pages/ssize_t.3type.en

ioquatix:
IIRC, it’s only guaranteed to be able to represent -1 and nothing smaller, although in practice it may be able to.

ioquatix:
Let me summarise a few points:

Argument extraction for methods that take `(length, offset)` is done in https://github.com/ruby/ruby/blob/57bf73225ddb57bfb75171ff7d4c1e72d4f1dbd5/io_buffer.c#L2490 and does not accept negative values for offset.

The argument order for `get_string` is `(offset, length)`. However, I was considering changing it to (length, offset) to be more consistent. It would be a painful change but within the bounds of "experimental interface".

Other value accessors only need offset, typically.

Regarding negative offsets - it introduces more complexity to the interface, but I'm not sure it's valuable for `IO::Buffer` as the length of the buffer is not "user controlled" like it is for a string. The expected and encouraged usage is to allocate page-width buffers, and any fixed reference to the size of the buffer is unlikely to be useful.

So, my initial conclusion is: reject negative values as we do in `io_buffer_extract_arguments`, and my follow up question is whether we should change `get_string` (and others?) to follow the same order as `io_buffer_extract_arguments`.

Even if we decided to reject negative values today, it does not prevent us from introducing that later. It could also be useful to consider it in the wider picture of `io_buffer_extract_arguments` (should it support negative offsets?).

I need a bit more time to think about some of these questions, but I welcome your input.

PR #8016, opened 2023-07-03
[Bug #19754] Make `IO::Buffer#get_string` check `offset` range

nobu:
https://bugs.ruby-lang.org/issues/19754

### DIFF
--- a/io_buffer.c
+++ b/io_buffer.c
@@ -1156,6 +1156,9 @@ VALUE rb_io_buffer_free_locked(VALUE self)
 static inline void
 io_buffer_validate_range(struct rb_io_buffer *buffer, size_t offset, size_t length)
 {
+    if (offset > buffer->size) {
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks please let me check it.

ioquatix:
This looks okay to me.

ioquatix:
I merged this as it appears to fix a bug, but I think the bug is that length is computed incorrectly:

```
TestIOBuffer#test_get_string
io_buffer_validate_range(0x559895c17de0, 0, 16)
io_buffer_validate_range(0x559895c17de0, 0, 16)
io_buffer_validate_range(0x559895c17de0, 0, 16)
io_buffer_validate_range(0x559895c17de0, 0, 129)
io_buffer_validate_range(0x559895c17de0, 129, 18446744073709551615) # Something wrong here.
```

ioquatix:
See https://github.com/ruby/ruby/pull/8427 for the root cause fix.

PR #8022, opened 2023-07-04
Suppressing security alert of atty dependency by env_logger-0.9.0

hsbt:
https://github.com/ruby/ruby/security/dependabot/2


XrXr:
Thanks!

PR #8028, opened 2023-07-05
Hotfix disable yjit for non-flonum

shyouhei:
I'm aware I broke our CI.  This is because the container image now includes `rustc`, which is detected by our `configure`, which doesn't interface well with `-DUSE_FLONUM=0`.

This should fix (I don't think it's a wise idea to rely on `rustc`'s being absent).


nobu:
llvm-objcopy <= 9 also don't seem working.

shyouhei:
Mmm yes, that's a sad thing.  I guess they are too old.

PR #8029, opened 2023-07-05
Move ISASCII defination to parse.y

S-H-GAMELINKS:
Move to `ISASCII` defination to parse.y for universal parser.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -6352,6 +6352,17 @@ ripper_dispatch_delayed_token(struct parser_params *p, enum yytokentype t)
 #endif /* RIPPER */
 
 static inline int
+parse_isascii(int c)
+{
+    return '\0' <= c && c <= '\x7f';
+}
+
+#ifdef ISASCII
+#undef ISASCII
+#define ISASCII parse_isascii
+#endif
+
+static inline int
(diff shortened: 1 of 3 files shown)


yui-knk:
Thanks for the PR. Can I ask you to update check these points?

1. Remove `int (*isascii)(int c);` from `struct rb_parser_config_struct` (rubyparser.h) then remove `config->isascii = parse_isascii;` from `rb_parser_config_initialize` (ruby_parser.c). One of goals of the refactoring is to minimize the number of functions required by `rb_parser_config_t`.
2. Define `parse_isascii` in parse.y not rubyparser.h. rubyparser.h is a public interface. I don't want to make `parse_isascii` public. We can put `parse_isascii` definition just before `#ifdef ISASCII` in parse.y

S-H-GAMELINKS:
Thanks for your review!
I fixed these point.

>1. Remove int (*isascii)(int c); from struct rb_parser_config_struct (rubyparser.h) then remove config->isascii = parse_isascii; from rb_parser_config_initialize (ruby_parser.c). One of goals of the refactoring is to minimize the number of functions required by rb_parser_config_t.
>2. Define parse_isascii in parse.y not rubyparser.h. rubyparser.h is a public interface. I don't want to make parse_isascii public. We can put parse_isascii definition just before #ifdef ISASCII in parse.y

yui-knk:
Thank you!

PR #8030, opened 2023-07-05
Remove duplicated Rust installation

k0kubun:
https://cirrus-ci.com/task/4785727414730752

```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y $rustup_init
info: downloading installer
warning: it looks like you have an existing installation of Rust at:
warning: /rust/cargo/bin
warning: It is recommended that rustup be the primary Rust installation.
warning: Otherwise you may have confusion unless you are careful with your PATH
warning: If you are sure that you want both rustup and your already installed Rust
warning: then please reply `y' or `yes' or set RUSTUP_INIT_SKIP_PATH_CHECK to yes
warning: or pass `-y' to ignore all ignorable checks.
error: cannot install while Rust is installed
warning: continuing (because the -y flag is set and the error is ignorable)
error: could not write settings file: '/rust/rustup/settings.toml'

Caused by:
    Permission denied (os error 13)
```

It seems to be installed to the system by default.


XrXr:
Looks like the breakage is coming from the recent change here: https://github.com/ruby/ruby-ci-image/pull/8

The image should already have `rustc`.

k0kubun:
Yeah. Apparently it also has `cargo` itself, but the remaining problem is missing `clippy`. Maybe we should only attempt to install clippy in that place.

k0kubun:
It looks like `rustup component add clippy` doesn't work https://cirrus-ci.com/task/5907292055404544:

```
cd yjit && rustup component add clippy && cargo clippy --all-targets --all-features
info: downloading component 'clippy'
info: retrying download for 'https://static.rust-lang.org/dist/2023-06-01/clippy-1.70.0-aarch64-unknown-linux-gnu.tar.xz'
info: retrying download for 'https://static.rust-lang.org/dist/2023-06-01/clippy-1.70.0-aarch64-unknown-linux-gnu.tar.xz'
info: retrying download for 'https://static.rust-lang.org/dist/2023-06-01/clippy-1.70.0-aarch64-unknown-linux-gnu.tar.xz'
info: retrying download for 'https://static.rust-lang.org/dist/2023-06-01/clippy-1.70.0-aarch64-unknown-linux-gnu.tar.xz'
error: component download failed for clippy-aarch64-unknown-linux-gnu

Caused by:
    could not download file from 'https://static.rust-lang.org/dist/2023-06-01/clippy-1.70.0-aarch64-unknown-linux-gnu.tar.xz' to '/rust/rustup/downloads/ca344314dfc6db7432bbb69c404549bbcbf4a1a0443e4fd63df2069b082628b8.partial'
```

Given that we run the same thing for non-arm64 code on GitHub Actions, I don't feel this is too urgent. Let me make the CI green first and leave this issue for another PR.

shyouhei:
@k0kubun Do we want to do `rustup component add clippy` in the container side?  That in fact sounds nice to me.

k0kubun:
Ah, I just didn't realize it was our `ghcr.io/ruby/ruby-ci-image`. Yes, adding that to the container image would be helpful.

PR #8032, opened 2023-07-05
YJIT: add new stats counter for compiled ISEQ entry points

maximecb:
This is related to https://github.com/Shopify/ruby/issues/520

Adding a counter so we can keep track of how many times we start compiling a new ISEQ from the interpreter entry point (and so based on the call threshold).

Some results from our benchmarks:
```
For railsbench:
compiled_iseq_count:           2,151
compiled_iseq_entry:           1,214

For hexapdf:
compiled_iseq_count:             650
compiled_iseq_entry:             433
```

As can be seen, most ISEQs get compiled directly from the entry point, which I find a little disappointing. I would have expected more ISEQs to get compiled due to calls from callers. I think this definitely weighs in favor of adding some factor to favor larger ISEQs being compiled first (having lower effective threshold).


XrXr:
For solving "we're compiling too much", maybe comparing the two metrics is not comprehensive enough. `compiled_iseq_count` increments only for the first block we add to a particular iseq, so an iseq that has many blocks doesn't get as much representation in this stat, even though we might retain a lot of code for it.

maximecb:
> For solving "we're compiling too much", maybe comparing the two metrics is not comprehensive enough. `compiled_iseq_count` increments only for the first block we add to a particular iseq, so an iseq that has many blocks doesn't get as much representation in this stat, even though we might retain a lot of code for it.

With this extra counter I just wanted to see how many ISEQs get compiled from interpreter entry points.

So many test failures 😮 I'm assuming this is due to some bogus commit that already got merged.

k0kubun:
> So many test failures 😮 I'm assuming this is due to some bogus commit that already got merged.

https://github.com/ruby/ruby/commit/f314656c238f799e26f49658b798379ccaf376ea should fix it. https://github.com/ruby/ruby/pull/8032/commits/d918f43aece564f03edc37e4159231ed7cc3f50b didn't include it, so I pushed https://github.com/ruby/ruby/pull/8032/commits/e54bac934a2b2ba979a744be790cacd839ff1a31.

yjit.rb
```diff
@@ -286,6 +286,7 @@ def _print_stats(out: $stderr) # :nodoc:
       out.puts "bindings_set:          " + format_number(13, stats[:binding_set])
       out.puts "compilation_failure:   " + format_number(13, compilation_failure) if compilation_failure != 0
       out.puts "compiled_iseq_count:   " + format_number(13, stats[:compiled_iseq_count])
+      out.puts "compiled_iseq_entry:   " + format_number(13, stats[:compiled_iseq_entry])
```
k0kubun, line 289:
```suggestion
      out.puts "compiled_iseq_entry:   " + format_number(13, stats[:compiled_iseq_entry])
      out.puts "compiled_iseq_count:   " + format_number(13, stats[:compiled_iseq_count])
```

It seems nicer to have them swapped to make it the later, the larger. It's consistent with the other in stats.rs as well.

PR #7115, opened 2023-01-12
YJIT: Fix shared/static library symbol leaks

XrXr:
https://bugs.ruby-lang.org/issues/19255
Based on nobu's branch (thank you!). It's going to bit a bit hacky, especially on macOS, but maybe alright?

---

Rust 1.58.0 unfortunately doesn't provide facilities to control symbol
visibility/presence, but we care about controlling the list of
symbols exported from libruby-static.a and libruby.so.

This commit uses `ld -r` to make a single object out of rustc's
staticlib output, libyjit.a. This moves libyjit.a out of MAINLIBS and adds
libyjit.o into COMMONOBJS, which obviates the code for merging libyjit.a
into libruby-static.a. The odd appearance of libyjit.a in SOLIBS is also
gone.

To filter out symbols we do not want to export on ELF platforms, we use
objcopy after the partial link. On darwin, we supply a symbol list to
the linker which takes care of hiding unprefixed symbols.

[Bug #19255]

PR #8034, opened 2023-07-06
Disable YJIT if objcopy command is old

nobu:


.github/workflows/compilers.yml
```diff
@@ -94,10 +94,10 @@ jobs:
           - { name: clang-11,  env: { default_cc: clang-11 } }
           - { name: clang-10,  env: { default_cc: clang-10 } }
           # llvm-objcopy<=9 doesn't have --wildcard. It compiles, but leaves Rust symbols in libyjit.o.
```
k0kubun, line 96:
```suggestion
```

k0kubun (commented):
It seems fair to me, but I wonder what other folks in @ruby/yjit think.

noahgibbs:
Looks reasonable to me.

configure.ac
```diff
@@ -3790,13 +3790,23 @@ AS_IF([test "$cross_compiling" = no],
...
+])
+
 dnl build YJIT in release mode if rustc >= 1.58.0 is present and we are on a supported platform
 AC_ARG_ENABLE(yjit,
     AS_HELP_STRING([--enable-yjit],
     [enable in-process JIT compiler that requires Rust build tools. enabled by default on supported platforms if rustc 1.58.0+ is available]),
     [YJIT_SUPPORT=$enableval],
-    [AS_CASE(["$YJIT_TARGET_OK:$YJIT_RUSTC_OK"],
-        [yes:yes], [
+    [AS_CASE(["$YJIT_TARGET_OK:$YJIT_RUSTC_OK:$YJIT_OBJCOPY_OK"],
```
k0kubun, line 3808: @nobu https://github.com/ruby/ruby/pull/7115 used objcopy only for ELF platforms (@XrXr Let me know if I'm wrong), so this check should be skipped for darwin. hsbt temporarily reverted this PR https://github.com/ruby/ruby/commit/a642a94b68a31c0e7c092087114c269132cdb159 since it now disables YJIT build on macOS by default.

nobu: Is it for all ELF platforms, not only Linux?

XrXr: ELF platforms other than Linux, is really just FreeBSD, and it has a new enough objcopy (I couldn't test on OpenBSD). YJIT on macOS doesn't need objcopy at all because the linker has options for hiding everything.

PR #8035, opened 2023-07-06
Ensure the name given to Module#set_temporary_name is not a valid constant path

eregon:
cc @fxn

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -230,7 +230,7 @@ rb_mod_set_temporary_name(VALUE mod, VALUE name)
         }
 
         if (is_constant_path(name)) {
-            rb_raise(rb_eArgError, "name must not be valid constant path");
+            rb_raise(rb_eArgError, "the temporary name must not be a constant path to avoid confusion");
         }
 
         // Set the temporary classpath to the given name:
--- a/variable.c
+++ b/variable.c
@@ -35,6 +35,7 @@
 #include "ruby/util.h"
 #include "transient_heap.h"
 #include "shape.h"
+#include "symbol.h"
 #include "variable.h"
 #include "vm_core.h"
 #include "ractor_core.h"
@@ -134,6 +135,38 @@ rb_mod_name(VALUE mod)
     return classname(mod, &permanent);
 }
 
+// Similar to logic in rb_mod_const_get()
+static bool
+is_constant_path(VALUE name)
+{
+    const char *path = RSTRING_PTR(name);
+    const char *pend = RSTRING_END(name);
+    rb_encoding *enc = rb_enc_get(name);
+
+    const char *p = path;
+
+    if (p >= pend || !*p) {
+        return false;
+    }
+
+    while (p < pend) {
+        if (p + 2 <= pend && p[0] == ':' && p[1] == ':') {
+            p += 2;
+        }
+
+        const char *pbeg = p;
+        while (p < pend && *p != ':') p++;
+
+        if (pbeg == p) return false;
+
+        if (rb_enc_symname_type(pbeg, p - pbeg, enc, 0) != ID_CONST) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
 /*
  *  call-seq:
  *     mod.set_temporary_name(string) -> self
@@ -196,8 +229,8 @@ rb_mod_set_temporary_name(VALUE mod, VALUE name)
             rb_raise(rb_eArgError, "empty class/module name");
         }
 
-        if (rb_is_const_name(name)) {
-            rb_raise(rb_eArgError, "name must not be valid constant name");
+        if (is_constant_path(name)) {
+            rb_raise(rb_eArgError, "name must not be valid constant path");
         }
 
         // Set the temporary classpath to the given name:
--- a/common.mk
+++ b/common.mk
@@ -17548,6 +17548,7 @@ variable.$(OBJEXT): {$(VPATH)}rubyparser.h
 variable.$(OBJEXT): {$(VPATH)}shape.h
 variable.$(OBJEXT): {$(VPATH)}st.h
 variable.$(OBJEXT): {$(VPATH)}subst.h
+variable.$(OBJEXT): {$(VPATH)}symbol.h
 variable.$(OBJEXT): {$(VPATH)}thread_$(THREAD_MODEL).h
 variable.$(OBJEXT): {$(VPATH)}thread_native.h
 variable.$(OBJEXT): {$(VPATH)}transient_heap.h
--- a/spec/ruby/core/module/name_spec.rb
+++ b/spec/ruby/core/module/name_spec.rb
@@ -6,55 +6,6 @@ describe "Module#name" do
     Module.new.name.should be_nil
   end
 
-  ruby_version_is "3.3" do
-    it "can assign a temporary name" do
-      m = Module.new
-      m.name.should be_nil
-
-      m.set_temporary_name("fake_name")
-      m.name.should == "fake_name"
-
-      m.set_temporary_name(nil)
-      m.name.should be_nil
-    end
-
-    it "can't assign empty string as name" do
-      m = Module.new
-      -> { m.set_temporary_name("") }.should raise_error(ArgumentError, "empty class/module name")
-    end
-
-    it "can't assign a constant name as a temporary name" do
-      m = Module.new
-      -> { m.set_temporary_name("Object") }.should raise_error(ArgumentError, "name must not be valid constant name")
-    end
-
-    it "can't assign name to permanent module" do
-      -> { Object.set_temporary_name("fake_name") }.should raise_error(RuntimeError, "can't change permanent name")
-    end
-
-    it "can assign a temporary name to a nested module" do
-      m = Module.new
-      module m::N; end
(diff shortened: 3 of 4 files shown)


eregon:
There seems to be a bug, because when running the spec on its own it fails:
```
make test-spec MSPECOPT="spec/ruby/core/module/set_temporary_name_spec.rb" 
revision.h updated
generating x86_64-linux-fake.rb
x86_64-linux-fake.rb updated
$ /home/eregon/code/ruby/miniruby -I/home/eregon/code/ruby/lib /home/eregon/code/ruby/tool/runruby.rb --archdir=/home/eregon/code/ruby --extout=.ext -- /home/eregon/code/ruby/spec/mspec/bin/mspec-run -B ./spec/default.mspec spec/ruby/core/module/set_temporary_name_spec.rb
ruby 3.3.0dev (2023-07-06T11:45:07Z constant-path-check 484cd0be36) [x86_64-linux]
                                                                                             
1)
Module#set_temporary_name can't assign a constant path as a temporary name FAILED
Expected ArgumentError (name must not be valid constant path)
but no exception was raised (A::B was returned)
/home/eregon/code/ruby/spec/ruby/core/module/set_temporary_name_spec.rb:37:in `block (3 levels) in <top (required)>'
/home/eregon/code/ruby/spec/ruby/core/module/set_temporary_name_spec.rb:4:in `block in <top (required)>'
/home/eregon/code/ruby/spec/ruby/core/module/set_temporary_name_spec.rb:3:in `<top (required)>'
[\ | ==================100%================== | 00:00:00]      1F      0E 

Finished in 0.000996 seconds

1 file, 8 examples, 14 expectations, 1 failure, 0 errors, 0 tagged
```
But somehow it passes when all specs are run, weird.

eregon:
The problem is almost surely from `rb_check_id_cstr` which seems quite difficult to use correctly, probably I should not use that.

eregon:
I have cleaned it up, now it's ready for review.

nobu:
> The problem is almost surely from `rb_check_id_cstr` which seems quite difficult to use correctly, probably I should not use that.

`ID`s are permanent resources, not for temporary uses.

variable.c
```diff
@@ -134,6 +134,38 @@ rb_mod_name(VALUE mod)
     return classname(mod, &permanent);
 }
 
+// Similar to logic in rb_mod_const_get()
+static bool
+is_constand_path(VALUE name) {
```
nobu, line 139:
```suggestion
is_constand_path(VALUE name)
{
```

variable.c
```diff
@@ -134,6 +134,38 @@ rb_mod_name(VALUE mod)
...
+    const char *path = RSTRING_PTR(name);
+    const char *pend = RSTRING_END(name);
+
```
nobu, line 142:
```suggestion
    rb_encoding *enc = rb_enc_get(name);

```

variable.c
```diff
@@ -134,6 +134,38 @@ rb_mod_name(VALUE mod)
...
+        }
+
+        const char *pbeg = p;
+        while (p < pend && *p != ':') p++;
+
+        if (pbeg == p) return false;
+
+        VALUE part = rb_str_subseq(name, pbeg - path, p - pbeg);
+        OBJ_FREEZE(part);
+        if (!rb_is_const_name(part)) {
```
nobu, line 161:
```suggestion
        if (rb_enc_symname_type(pbeg, p - pbeg, enc, 0) != ID_CONST) {
```

nobu (commented):
It is able to see ID type without making a temporary `String`.

variable.c
nobu: Of course `#include "symbol.h"` and a dependency in common.mk are needed, too.

eregon: > and a dependency in common.mk are needed, too.

How to do that one?

eregon:
@nobu Of course I get:
```
compiling variable.c
variable.c: In function ‘is_constand_path’:
variable.c:161:13: warning: implicit declaration of function ‘rb_enc_symname_type’; did you mean ‘rb_enc_symname_p’? [-Wimplicit-function-declaration]
  161 |         if (rb_enc_symname_type(pbeg, p - pbeg, enc, 0) != ID_CONST) {
      |             ^~~~~~~~~~~~~~~~~~~
      |             rb_enc_symname_p
At top level:
cc1: note: unrecognized command-line option ‘-Wno-self-assign’ may have been intended to silence earlier diagnostics
cc1: note: unrecognized command-line option ‘-Wno-parentheses-equality’ may have been intended to silence earlier diagnostics
cc1: note: unrecognized command-line option ‘-Wno-constant-logical-operand’ may have been intended to silence earlier diagnostics
```
I need to add the header.
But, why is that not an error? I thought we made implicit-function-declaration always an error, no?

ioquatix (approved):
LGTM

fxn:
Ah, I believe an article "a" is missing too, my suggestion would be:

> name must not be a constant path

fxn:
> True, but name must not be (a) valid constant path seems clearer than name must not be a constant path to me.
Any suggestion for a correct and clear error message?

Your call :).

eregon:
I'll change to `the temporary name must not be a constant path to avoid confusion` that seems clear to me

fxn:
> I'll change to the temporary name must not be a constant path to avoid confusion that seems clear to me

Sounds good too!

fxn:
I have had conversation with Matz about terminology related to constants.

Relevant to this patch is that `::A::B` is not considered to be a constant path. Therefore, given that input, you cannot reject with a message that says "must not be a constant path", since it is indeed not a constant path.

If you want to keep the error message in the patch, `::A` and `::A::B` should be allowed. Indeed, "::A::B" cannot be a permanent name, so there would be no confusion. Remember, Ruby says nothing about whether a name could be evaluated, or that if evaluated it gives you the original module object.

Now, _if_ you want the name to be evaluable, then you are responsible for its contents. The `Template` use case.

If you prefer to disallow a leading `::`, then the error message needs to be revised, I think.

eregon:
I think we shouldn't allow `::A`/`::B` because it would be pretty confusing, and there seems be little to no value to have a temporary name starting with `::` (since temporary modules are not rooted in an Object constant).

> If you prefer to disallow a leading `::`, then the error message needs to be revised, I think.

Any suggestion?
TBH Ruby error messages are rarely exact or very precise (or even correct grammatically), so this tiny edge case is IMO better than most exception messages in core, and probably matters little in practice.

I'll merge this now so I don't forget about it, we can tweak the message again, feel free to open a PR for that.

eregon:
We could reject "starts with ::" with a different message, unsure if that's worth the extra complexity in the code.

nobu:
> I need to add the header. But, why is that not an error? I thought we made implicit-function-declaration always an error, no?

It'd depend on compilers, I got an error there.

```
../src/variable.c:161:13: error: call to undeclared function 'rb_enc_symname_type'; ISO C99 and later do not support
      implicit function declarations [-Wimplicit-function-declaration]
        if (rb_enc_symname_type(pbeg, p - pbeg, enc, 0) == ID_CONST) {
            ^
../src/variable.c:161:13: note: did you mean 'rb_enc_symname_p'?
../src/include/ruby/internal/encoding/symbol.h:56:5: note: 'rb_enc_symname_p' declared here
int rb_enc_symname_p(const char *str, rb_encoding *enc);
    ^
1 error generated.
```

fxn:
Update: While we chatted about the three concepts, in the end Matz concluded this very explictly:

> There's no constant path in Ruby.

He only uses two terms: _constant reference_, and _class/module path_.

eregon:
@nobu I have:
```
$ ruby -e 'pp RbConfig::CONFIG["CC"]'
"gcc"
$ cc --version
gcc (GCC) 12.2.1 20221121 (Red Hat 12.2.1-4)
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
I'd think that compiler supports `-Werror=implicit-function-declaration`. Any idea why it would detect not or choose to not pass -Werror?

eregon:
I was curious what term `const_get` docs use and it's:
> This method will recursively look up constant names if a namespaced
> class name is provided.  For example:
> 
>   module Foo; class Bar; end end
>   Object.const_get 'Foo::Bar'

I'm not sure `namespaced class name` is any clearer (in fact I think it's less clear than `constant path`).

Re `class/module path`, although a term I've not heard before, that sounds OK to me, PR welcome if you'd like to change to that.

---

I can't resist to quote this part too :D

> If the argument is not a valid constant name a NameError will be raised
> with a warning "wrong constant name".
> 
>   Object.const_get 'foobar' #=> NameError: wrong constant name foobar

I guess because many Rubyists have seen "wrong <...> name" then "valid <...> name" is a thing.

fxn:
Agree, I believe "constant path" is a term that would be useful for docs and error messages, even if it is not a thing in the syntax tree.

Because class/mod path is not generic enough for the use cases we find. The constant may not be storing a class/mod necessarily, or may not be a permanent name of any object.

I have asked to Matz if he'd be open to a proposal to introduce this concept formally, let's see what he thinks.

PR #8039, opened 2023-07-07
Fix autosplat conditions to handle ruby2_keywords case

jeremyevans:
Autosplat should not occur if there are two arguments but second argument is an array containing a ruby2_keywords splat.

Fixes [Bug #19759]

### DIFF
--- a/vm_args.c
+++ b/vm_args.c
@@ -640,7 +640,7 @@ setup_parameters_complex(rb_execution_context_t * const ec, const rb_iseq_t * co
       case arg_setup_method:
         break; /* do nothing special */
       case arg_setup_block:
-        if (given_argc == ((NIL_P(keyword_hash) || !calling->kw_splat) ? 1 : 2) &&
+        if (given_argc == 1 &&
             allow_autosplat &&
             (min_argc > 0 || ISEQ_BODY(iseq)->param.opt_num > 1) &&
             !ISEQ_BODY(iseq)->param.flags.ambiguous_param0 &&
--- a/vm_args.c
+++ b/vm_args.c
@@ -640,7 +640,7 @@ setup_parameters_complex(rb_execution_context_t * const ec, const rb_iseq_t * co
       case arg_setup_method:
         break; /* do nothing special */
       case arg_setup_block:
-        if (given_argc == (NIL_P(keyword_hash) ? 1 : 2) &&
+        if (given_argc == ((NIL_P(keyword_hash) || !calling->kw_splat) ? 1 : 2) &&
             allow_autosplat &&
             (min_argc > 0 || ISEQ_BODY(iseq)->param.opt_num > 1) &&
             !ISEQ_BODY(iseq)->param.flags.ambiguous_param0 &&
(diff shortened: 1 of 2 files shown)


vm_args.c
```diff
@@ -640,7 +640,7 @@ setup_parameters_complex(rb_execution_context_t * const ec, const rb_iseq_t * co
       case arg_setup_method:
         break; /* do nothing special */
       case arg_setup_block:
-        if (given_argc == (NIL_P(keyword_hash) ? 1 : 2) &&
+        if (given_argc == ((NIL_P(keyword_hash) || !calling->kw_splat) ? 1 : 2) &&
```
XrXr, line 643: Is there any situation where this evaluates to `given_argc == 2`, and autosplat goes through? I looked at this in the first place because the old condition where it checked for 2, `!NIL_P(keyword_hash) && allow_autosplat`, looked like a contradiction, since `allow_autosplat` is defined as `0 == (vm_ci_flag(ci) & (VM_CALL_KWARG | VM_CALL_KW_SPLAT | VM_CALL_KW_SPLAT_MUT))`

Checking for 2 is also weird because `given_argc` is decremented in different paths above.

jeremyevans: `given_argc` is decremented, but it is also incremented based on the size of the splat (`given_argc += len - 1`). However, it does appear unlikely that the check for `given_argc == 2` is needed anymore.  I assume it was needed in beae6cbf0fd8b6619e5212552de98022d4c4d4d4 to make the added test pass, but changes since then may now make it unnecessary.  I'll push a commit that just uses `given_argc == 1` and see if that passes CI.

XrXr: I did look at `given_argc == 1` and found that it changes the behavior of `yield(*[[:a, :b], keywords_hash])`, so it's no quite right (noted in the ticket).

The condition that makes the most sense to me right now is:
```c
if (given_argc == 1 &&             // passing one arg
    allow_autosplat &&             // rules out given_argc changes from kwsplat handling 
    !splat_flagged_keyword_hash && // rules out ruby2_keywords_hash given_argc changes,
                                   // not captured by allow_autosplat as this comes
                                   // from processing non-kw splat.
```

It's supposed to activate splatting an array that has a single array in it, and the `keyword_hash` check sort of undos the `given_argc` decrement to look at the size of the original splat array:

```ruby
def yielder(splat)
  yield(*splat)
end

print = Proc.new { |a, b, **| p [a, b] }
yielder([ [:a, :b] ], &print) # autosplat after splatting array of one array


yielder([ [:a, :b], Hash.ruby2_keywords_hash({}) ], &print)    # no autosplat, array of two elements
yielder([ [:a, :b], Hash.ruby2_keywords_hash({k:1}) ], &print) # no autosplat, array of two elements
# ^ the two cases above change behavior if we only check for given_argc == 1

puts

yielder([ [:a, :b], Hash.ruby2_keywords_hash({}) ]) \
  do |a, b, **nil| # in this case autosplat activates on 3.2. It's an odd one.
    p [a, b]       # if instead of NIL_P(keyword_hash) we checked for !splat_flagged_keyword_hash
  end              # autosplat would deactivate in this case too.
```

So, I think we need to figure out when autosplat should activate in principle. In my model, it's checking for the argc immediately after splat expansion. This model predicts that the 3.2 behavior in the `**nil` case is incorrect, though.

jeremyevans: Your analysis makes sense to me.  I added more test cases and the `!splat_flagged_keyword_hash` change.  Let's see if this still passes CI.

PR #8042, opened 2023-07-07
Add yarp/yarp_compiler.c

jemmaissroff:
Add structure code necessary for compiling YARP. I will continue to fill in the actual compilation code. This is just the stencil of how we'll implement `RubyVM::InstructionSequence.compile_yarp`


tenderlove:
LGTM.  @ko1 do you think it's OK for us to add `compile_yarp` to `RubyVM::InstructionSequence`?  Or should we add a `:yarp` option to `compile`?  Or something else?

tenderlove:
@ko1 I think renaming isn't a big deal, we just need to have a better name.  Any suggestions? 😅

ko1:
Please ask matz!

eregon:
I replied on https://bugs.ruby-lang.org/issues/19772.
In short:
* I don't think it makes sense to rename YARP, IMO it's too late for that.
* `RubyVM::InstructionSequence.compile(yarp: true)` seems best
* YARP.compile is not OK, it would force to partially define the YARP module, and the implementation should be in compile.c AFAIK (because it depends on the specific bytecodes in a given CRuby version/commit, the YARP gem can't maintain that logic for multiple Ruby versions + this stuff only makes sense when part of CRuby (cannot be compiled outside of CRuby))

kddnewton (approved):

PR #8043, opened 2023-07-08
Remove unused ISASCII definition

alitaso345:
- Removed ISASCII definition from universal_parser.c
- The ISASCII definition was moved to parse.y, but the old definition wasn't removed.
    - https://github.com/ruby/ruby/pull/8029


yui-knk:
Thank you!!

PR #8044, opened 2023-07-08
Move some macro for universal parser 

S-H-GAMELINKS:
Move some macro that defined in ``internal/ctype.h`` to `parse.y` .

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -201,6 +201,87 @@ new_strterm(VALUE v1, VALUE v2, VALUE v3, VALUE v0, int heredoc)
 }
 #endif /* !UNIVERSAL_PARSER */
 
+static inline int
+parse_isascii(int c)
+{
+    return '\0' <= c && c <= '\x7f';
+}
+
+#undef ISASCII
+#define ISASCII parse_isascii
+
+static inline int
+parse_isspace(int c)
+{
+    return c == ' ' || ('\t' <= c && c <= '\r');
+}
+
+#undef ISSPACE
+#define ISSPACE parse_isspace
+
+static inline int
+parse_iscntrl(int c)
+{
+    return ('\0' <= c && c < ' ') || c == '\x7f';
+}
+
+#undef ISCNTRL
+#define ISCNTRL(c) parse_iscntrl(c)
+
+static inline int
+parse_isupper(int c)
+{
+    return 'A' <= c && c <= 'Z';
+}
+
+static inline int
+parse_islower(int c)
+{
+    return 'a' <= c && c <= 'z';
+}
+
+static inline int
+parse_isalpha(int c)
+{
+    return parse_isupper(c) || parse_islower(c);
+}
+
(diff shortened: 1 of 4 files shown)


parse.y
```diff
@@ -80,6 +80,94 @@ struct lex_context;
...
+static inline int
+parse_isspace(int c)
+{
+    return c == ' ' || ('\t' <= c && c <= '\r');
+}
+
+#ifdef ISSPACE
+#undef ISSPACE
+#define ISSPACE parse_isspace
+#endif
```
nobu, line 92: Is `#ifdef` needed?
This means `ISSPACE` macro needs to be defined previously.

S-H-GAMELINKS: Thanks for comment.
`ISSPACE` macro already defined in `include/ruby/internal/ctype.h` and that included in `ruby/ruby.h`.
`parse.y` include `ruby/ruby.h`, and I thought `#undef ISSPACE` is needed.

yui-knk: At the moment `include/ruby/internal/ctype.h` is included into parse.y in both cases of with `UNIVERSAL_PARSER` and without `UNIVERSAL_PARSER`. However we may remove such inclusion in the future if possible. Current code wrapped with `#ifdef` might not work once `include/ruby/internal/ctype.h` is not included because `#define ISSPACE parse_isspace` is not defined without `ISSPACE` definition ahead of this line.
Could you try to remove `#ifdef ISSPACE` ... `#ifdef STRNCASECMP` ?

S-H-GAMELINKS: Thanks comment and I understand. I'll try it.
And shoud I remove also `#ifdef ISASCII `?

S-H-GAMELINKS: Removed `#ifdef ISSPACE` ... `#ifdef STRNCASECMP`(also `#ifdef ISASCII`).
It' be fine that all test has passed, I think.

yui-knk: Not only `#ifdef ISSPACE` and `#ifdef ISASCII` but also other `#ifdef XXX` like `#ifdef ISCNTRL`, `#ifdef ISALPHA` ...

By the way, current positions for `parse_XXX` is only for without `UNIVERSAL_PARSER`. Could you move these definitions to the place outside of `UNIVERSAL_PARSER` macro branches and `RIPPER` macro branches like https://github.com/ruby/ruby/blob/2903e9456f6990c068008eb823b90715e5a21932/parse.y#L542?

```
# without UNIVERSAL_PARSER
$ nm parse.o | grep isspace
0000000000010ab0 t _parse_isspace
000000000002a9b8 t _rb_enc_isspace

# with UNIVERSAL_PARSER
$ nm parse.o | grep isspace
0000000000010a08 t _rb_isspace
```

S-H-GAMELINKS: Thanks comment, It's fixed to move outside of UNIVERSAL_PARSER macro branches and RIPPER macro branche.

PR #8053, opened 2023-07-10
[DOC] fix formatting - bsearch.rdoc & getoptLong.rb

beeburrt:
These are intended to be as sub-headings, but as they are, they blend in with the rest of the text.

This change makes them stand out more.

https://docs.ruby-lang.org/en/master/bsearch_rdoc.html

### DIFF
--- a/doc/bsearch.rdoc
+++ b/doc/bsearch.rdoc
@@ -29,7 +29,7 @@ Find-any mode:: method +bsearch+ some element, if any, for which
 The block should not mix the modes by sometimes returning +true+ or +false+
 and other times returning a numeric value, but this is not checked.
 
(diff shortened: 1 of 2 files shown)


nobu:
The tag `[docs]` doesn't work for our CIs, it should be `[DOC]`.

beeburrt:
> The tag `[docs]` doesn't work for our CIs, it should be `[DOC]`.

Awesome, thank you. Is it ok that I pushed to the same branch like that?

nobu:
> Do you mean for the commit or for the PR?

The tag is for commits.
Currently, PRs don't see titles and need to be labeled as "Documentation".

> Also, is it ok that I pushed to the same branch like that?

Of course you can push to your branch anytime.

PR #8054, opened 2023-07-10
[Bug #19392] Make args connected with `and`/`or` as a body of method

yui-knk:
`def test = puts("foo") or puts("bar")` is interrupted as `def test = (puts("foo") or puts("bar"))`.


nobu (approved):

zverok:
Unfortunately, this PR doesn't fix `if`/`unless` problem (which, despite my ticket's initial contents, is more critical than `and`/`or`):

```ruby
def foo(arg) = p("OK") if arg

foo(false)
foo(true)
```
This is expected (with "natural" for the author precedences) to print "OK" once, but instead it fails with 
```
test.rb:1:in `<main>': undefined local variable or method `arg' for main (NameError)
```

Apparently, it is still interpreted as
```ruby
(def foo(arg) = p("OK")) if arg
```

PR #8055, opened 2023-07-11
Try to fix flaky GC test

peterzhu2118:
assert_not_nil could allocate objects which may trigger the major GC, so don't run the assertions until the major GC has been ran.

### DIFF
--- a/test/ruby/test_gc.rb
+++ b/test/ruby/test_gc.rb
@@ -262,9 +262,12 @@ class TestGc < Test::Unit::TestCase
       objects.append(100.times.map { '*' })
     end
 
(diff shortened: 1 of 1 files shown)


test/ruby/test_gc.rb
```diff
@@ -262,9 +262,12 @@ def test_latest_gc_info_need_major_by
       objects.append(100.times.map { '*' })
     end
 
-    assert_not_nil GC.latest_gc_info(:need_major_by)
+    need_major_by = GC.latest_gc_info(:need_major_by)
```
nirvdrum, line 265: I appreciate the git log message indicating why the change was made, but I think a comment here would make it less likely someone accidentally makes it flaky again.

PR #8058, opened 2023-07-12
Reduce util.c dependency

S-H-GAMELINKS:
Reduce `util.c` dependency for Universal Parser.


nobu (commented):
It doesn't look nice to copy every duplicate code.

PR #8061, opened 2023-07-12
[DOC] Add spaces to fix formatting in example - getoptlong.rb

beeburrt:
The very last example on the page here, under "Command line:":

https://docs.ruby-lang.org/en/master/GetoptLong.html

the example doesn't show the code properly formatted, so I added two spaces to match the other examples

### DIFF
--- a/lib/getoptlong.rb
+++ b/lib/getoptlong.rb
@@ -352,7 +352,7 @@
 #
 # Command line:
 #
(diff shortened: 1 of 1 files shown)


beeburrt:
@nobu, Did I do something wrong? Why was this closed instead of merged directly? Just wondering so I can do better next time :)

shahroon:
Unsubscribe

PR #8064, opened 2023-07-13
YJIT: Make ratio_in_yjit always available

k0kubun:
This PR changes `ratio_in_yjit` to be available on any YJIT-enabled build. To minimize overhead when `--yjit` or `--yjit-stats` is not used, we increment a global variable without branching or locking. The counter could be inaccurate on Ractors.

### interpreter
```
before: ruby 3.3.0dev (2023-07-13T18:48:14Z master 62ecf78b87) [x86_64-linux]
after: ruby 3.3.0dev (2023-07-13T20:22:04Z yjit-reg-counter 6c13fcc476) [x86_64-linux]

railsbench  1912.2       0.6         1913.4      0.6         1.00           1.00
```

### --yjit
```
before: ruby 3.3.0dev (2023-07-13T18:48:14Z master 62ecf78b87) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-07-13T20:22:04Z yjit-reg-counter 6c13fcc476) +YJIT [x86_64-linux]

railsbench  1237.2       1.1         1239.3      1.1         1.00           1.00
```

### --yjit-exec
```
before: ruby 3.3.0dev (2023-07-13T18:48:14Z master 62ecf78b87) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-07-13T20:22:04Z yjit-reg-counter 6c13fcc476) +YJIT [x86_64-linux]

railsbench  1326.9       1.0         1317.0      1.0         1.00           1.01
```

This PR doesn't seem to have a significant performance impact on railsbench.

### DIFF
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -369,7 +369,6 @@ make_counters! {
     binding_allocations,
     binding_set,
 
-    vm_insns_count,
     compiled_iseq_entry,
     compiled_iseq_count,
     compiled_blockid_count,
@@ -507,7 +506,6 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
 
     let hash = unsafe { rb_hash_new() };
 
-    // CodeBlock stats
     unsafe {
         // Get the inline and outlined code blocks
         let cb = CodegenGlobals::get_inline_cb();
@@ -547,6 +545,9 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
             hash_aset_usize!(hash, "live_context_count", live_context_count);
             hash_aset_usize!(hash, "live_context_size", live_context_count * context_size);
         }
+
+        // VM instructions count
+        hash_aset_usize!(hash, "vm_insns_count", rb_vm_insns_count as usize);
     }
 
     // If we're not generating stats, the hash is done
@@ -565,13 +566,6 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
             // Get the counter value
             let counter_ptr = get_counter_ptr(counter_name);
             let counter_val = *counter_ptr;
-
-            #[cfg(not(feature = "stats"))]
-            if counter_name == &"vm_insns_count" {
-                // If the stats feature is disabled, we don't have vm_insns_count
-                // so we are going to exclude the key
-                continue;
-            }
 
             // Put counter into hash
             let key = rust_str_to_sym(counter_name);
@@ -748,12 +742,6 @@ pub extern "C" fn rb_yjit_reset_stats_bang(_ec: EcPtr, _ruby_self: VALUE) -> VAL
     }
 
     return Qnil;
-}
-
-/// Increment the number of instructions executed by the interpreter
-#[no_mangle]
-pub extern "C" fn rb_yjit_collect_vm_usage_insn() {
-    incr_counter!(vm_insns_count);
 }
 
 #[no_mangle]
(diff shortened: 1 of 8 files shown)


maximecb:
So the good news is it looks like you managed to make the overhead negligible, but that comes at the expense of some accuracy. Are we sure that the loss in accuracy will always be small?

Looking at the PR. I think maybe things could be simplified a bit? It looks like there are a few functions exported on both sides. Ideally, maybe we should try to do as much of the vm_insn_count accounting on the C side?

I'm also wondering if you tried to make it so that the C interpreter loop just does an atomic increment on a `vm_insn_count` global variable? I assume you've tried this and it was slower?

yjit.c
```diff
@@ -1116,6 +1116,17 @@ rb_yjit_assert_holding_vm_lock(void)
...
+// If true, count the number of insns executed on vm_exec_core.
+// Using a C variable instead of Rust code to minimize the overhead.
+bool rb_yjit_insn_counter_enabled = false;
+
+// Enable counting the number of insns executed on vm_exec_core.
+void
+rb_yjit_insn_counter_enable(void)
+{
+    rb_yjit_insn_counter_enabled = true;
+}
```
maximecb, line 1128: Maybe we could get rid of this state if we always do the update/increment?

yjit/src/stats.rs
```diff
@@ -787,3 +774,9 @@ fn global_allocation_size() -> usize {
     let stats = GLOBAL_ALLOCATOR.stats();
     stats.bytes_allocated.saturating_sub(stats.bytes_deallocated)
 }
+
+// Increase vm_insns_count. vm_exec_core uses this.
+#[no_mangle]
+pub extern "C" fn rb_yjit_add_vm_insns_count(count: u64) {
+    incr_counter_by!(vm_insns_count, count);
+}
```
maximecb, line 782: We could also get rid of this function if the insn count variable is owned/managed by the C side?

yjit.c
k0kubun: Incrementing a register is fast, but this flag is for propagating a value from a register to Rust. The latter is a function call, so I'd like to avoid it by just checking a C variable.

k0kubun: Hm, I guess you're thinking about moving ownership to C. Maybe that works? I'll try.

k0kubun:
> I'm also wondering if you tried to make it so that the C interpreter loop just does an atomic increment on a vm_insn_count global variable? I assume you've tried this and it was slower?

I assumed it wouldn't be faster than incrementing a local variable, so I didn't try it. I tried it now, and here's the result.

Using `lock addq` does slow down the interpreter on railsbench by 2%. However, using `addq` without locking does not slowdown the interpreter on railsbench. Given that it could be inaccurate only when you enable Ractors, I think it's a better type of inaccuracy for us? I'll update the PR like that after lunch.

maximecb:
> However, using addq without locking does not slowdown the interpreter on railsbench. Given that it could be inaccurate only when you run multiple Ractors, I think it's a better type of inaccuracy for us? I'll update the PR like that after lunch.

Sounds like it could be a pretty good tradeoff 👍 🙂

yjit.c
k0kubun: removed this function.

yjit/src/stats.rs
k0kubun: removed this too.

k0kubun:
I simplified the implementation https://github.com/ruby/ruby/pull/8064/commits/6148884aac913cd327f335d1817bae976c957b6e and updated the benchmark results on the PR description. The current version seems much nicer 🙂 

The failure of `test-bundled-gems` CI is an existing problem on master. I'll fix it separately.

maximecb (approved):
Nice! It's awesome that we can now have all the stats including percentage in YJIT on all builds! 😎

PR #8070, opened 2023-07-13
Use the caller location as default filename for eval family of methods

casperisfine:
[[Feature #19755]](https://bugs.ruby-lang.org/issues/19755)

Before (in /tmp/test.rb):

```ruby
Object.class_eval("p __FILE__") # => "(eval)"
```

After:

```ruby
Object.class_eval("p __FILE__") # => "(eval at /tmp/test.rb:1)"
```

This makes it much easier to track down generated code in case the author forgot to provide a filename argument.

### DIFF
--- a/vm_eval.c
+++ b/vm_eval.c
@@ -1613,7 +1613,23 @@ rb_each(VALUE obj)
     return rb_call(obj, idEach, 0, 0, CALL_FCALL);
 }
 
-static VALUE eval_default_path;
+static VALUE eval_default_path = Qfalse;
+
+static VALUE
+get_eval_default_path(void)
+{
+    int location_lineno;
+    VALUE location_path = rb_source_location(&location_lineno);
+    if (!NIL_P(location_path)) {
+        return rb_fstring(rb_sprintf("(eval at %"PRIsVALUE":%d)", location_path, location_lineno));
+    }
+
+    if (!eval_default_path) {
+        eval_default_path = rb_fstring_lit("(eval)");
+        rb_gc_register_mark_object(eval_default_path);
+    }
+    return eval_default_path;
+}
 
 static const rb_iseq_t *
 eval_make_iseq(VALUE src, VALUE fname, int line, const rb_binding_t *bind,
@@ -1653,11 +1669,7 @@ eval_make_iseq(VALUE src, VALUE fname, int line, const rb_binding_t *bind,
         if (!NIL_P(fname)) fname = rb_fstring(fname);
     }
     else {
-        if (!eval_default_path) {
-            eval_default_path = rb_fstring_lit("(eval)");
-            rb_gc_register_mark_object(eval_default_path);
-        }
-        fname = eval_default_path;
+        fname = get_eval_default_path();
         coverage_enabled = FALSE;
     }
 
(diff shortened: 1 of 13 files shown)


nobu:
Also syntax_suggest fails.

vm_eval.c
```diff
@@ -2508,9 +2527,16 @@ rb_current_realfilepath(void)
         if (path == eval_default_path) {
             return Qnil;
         }
-        else {
-            return path;
+
+        // [Feature #19755] implicit eval location is "(eval at #{__FILE__}:#{__LINE__})"
+        if (RSTRING_LEN(path) >= 9) {
+            if (RSTRING_PTR(path)[RSTRING_LEN(path) - 1] == ')' &&
+                memcmp(RSTRING_PTR(path), "(eval at ", 9) == 0) {
```
casperisfine, line 2534: I'm not super happy with this, but the only alternative I can think of would be to record either on the string or on the Iseq that it's an "eval location".

casperisfine:
The last blocker seem to be syntax_suggest, I opened a PR there: https://github.com/ruby/syntax_suggest/pull/200

casperisfine:
Ok, I think the PR is ready now.

vm_eval.c
```diff
@@ -1615,6 +1615,23 @@ rb_each(VALUE obj)
...
+static VALUE
+get_eval_default_path(void)
+{
+    int location_lineno;
+    VALUE location_path = rb_source_location(&location_lineno);
+    if (!NIL_P(location_path)) {
+        return rb_fstring(rb_sprintf("(eval at %s:%d)", RSTRING_PTR(location_path), location_lineno));
+    }
+
+    static VALUE eval_default_path;
```
nobu, line 1627: Should not shadow the file scope variable; it is used in `rb_current_realfilepath` too.

```suggestion
```

vm_eval.c
```diff
@@ -1982,6 +1995,12 @@ specific_eval(int argc, const VALUE *argv, VALUE self, int singleton, int kw_spl
             file = argv[1];
             if (!NIL_P(file)) StringValue(file);
         }
+
+
```
nobu, line 1999:
```suggestion
```

vm_eval.c
```diff
@@ -2508,9 +2527,16 @@ rb_current_realfilepath(void)
         if (path == eval_default_path) {
             return Qnil;
         }
-        else {
-            return path;
+
+        // [Feature #19755] implicit eval location is "(eval at #{__FILE__}:#{__LINE__})"
+        if (RSTRING_LEN(path) >= 9) {
```
nobu, line 2532: If it equals 9, the next `if` condition can't be true.
```suggestion
        if (RSTRING_LEN(path) > 9) {
```

casperisfine:
Thanks for the review @nobu !

vm_eval.c
```diff
@@ -1613,7 +1613,23 @@ rb_each(VALUE obj)
...
-static VALUE eval_default_path;
+static VALUE eval_default_path = Qfalse;
+
+static VALUE
+get_eval_default_path(void)
+{
+    int location_lineno;
+    VALUE location_path = rb_source_location(&location_lineno);
+    if (!NIL_P(location_path)) {
+        return rb_fstring(rb_sprintf("(eval at %s:%d)", RSTRING_PTR(location_path), location_lineno));
```
nobu, line 1624: Preserve the encoding.

```suggestion
        return rb_fstring(rb_sprintf("(eval at %"PRIsVALUE":%d)", location_path, location_lineno));
```

ioquatix:
Maybe this is a crazy idea, but if we could tell the difference between heredoc, string constant and string variable, we could probably set the default line number correctly in like 90% of cases.

PR #8072, opened 2023-07-13
Remove RGENGC_OLD_NEWOBJ_CHECK

peterzhu2118:
The code doesn't compile, so probably nobody is using this.


ko1:
we can re-introduce with random oldgen newly created objects to test gengc more.

PR #8076, opened 2023-07-14
[Doc] Corrected to the same operation as in the Quick start guide

jinroq:
I hope it to be the same operation as the [Quick start guide](https://docs.ruby-lang.org/en/master/contributing/building_ruby_md.html#label-Quick+start+guide).

* see [issues/19768](https://bugs.ruby-lang.org/issues/19768)

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -86,7 +86,8 @@ The YJIT `ruby` binary can be built with either GCC or Clang. It can be built ei
 ```sh
 # Configure in release mode for maximum performance, build and install
 ./autogen.sh
-./configure --enable-yjit --prefix=$HOME/.rubies/ruby-yjit --disable-install-doc
+mkdir build && cd build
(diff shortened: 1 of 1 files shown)


noahgibbs:
Looks like this is to make sure there's a build dir, so you can easily do a full clean build without having to do "make distclean" or equivalent. Not a bad thing.

XrXr:
Configuring outside of the source directory is what I do, too, but I think this breaks Maxime's workflow, because now you'd need to `cd ruby/build` instead of `cd ruby` before you can run `make`. If we're fine with that we can change it.

The root of the problem is in-source-dir builds being incompatible with out-of-source-dir builds and it's not really a YJIT issue. It's on my todo list to have `configure` detect and error out in this situation, but alas.

maximecb:
Yeah. I don't want to be annoying, but I don't love adding more steps to the build process 😅

PR #8077, opened 2023-07-14
[Doc] Modified shell script so that it can be executed by copy and paste.

jinroq:
#8076


XrXr:
Thanks for taking the time!

PR #8078, opened 2023-07-14
YJIT: refactoring to allow for fancier call threshold logic

maximecb:
Just a small refactoring to make it easier to implement more complex logic to determine when an ISEQ should be compiled from the Rust side.

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -387,7 +387,7 @@ jit_compile(rb_execution_context_t *ec)
 
     // Trigger JIT compilation as needed
     if (yjit_enabled) {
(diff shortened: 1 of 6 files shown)


vm.c
```diff
@@ -385,9 +385,15 @@ jit_compile(rb_execution_context_t *ec)
         return 0;
     }
 
+    // Don't try to compile the function if it's already compiled
+    if (body->jit_func)
+    {
```
XrXr, line 390:
```suggestion
    if (body->jit_func) {
```

k0kubun (approved):

PR #8080, opened 2023-07-14
[Bug #19769] Fix range of size 1 in String#tr

alexandre184:
Hi

The details are here :
https://bugs.ruby-lang.org/issues/19769

### DIFF
--- a/spec/ruby/core/string/tr_spec.rb
+++ b/spec/ruby/core/string/tr_spec.rb
@@ -16,11 +16,13 @@ describe "String#tr" do
     "hello ^-^".tr("---", "_").should == "hello ^_^"
   end
 
-  it "accepts c1-c1 notation to denote range of one character" do
-    "hello".tr('e-e', 'x').should == "hxllo"
-    "123456789".tr("2-23","xy").should == "1xy456789"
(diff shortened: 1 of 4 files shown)


spec/ruby/core/string/tr_spec.rb
```diff
@@ -16,6 +16,13 @@
     "hello ^-^".tr("---", "_").should == "hello ^_^"
   end
 
+  it "accepts c1-c1 notation to denote range of one character" do
+    "hello".tr('e-e', 'x').should == "hxllo"
+    "123456789".tr("2-23","xy").should == "1xy456789"
+    "hello ^-^".tr("e-", "a-a_").should == "hallo ^_^"
+    "hello ^-^".tr("---o", "_a").should == "hella ^_^"
+  end
```
nobu, line 24:
```suggestion
  ruby_bug "#19769", ""..."3.3" do
    it "accepts c1-c1 notation to denote range of one character" do
      "hello".tr('e-e', 'x').should == "hxllo"
      "123456789".tr("2-23","xy").should == "1xy456789"
      "hello ^-^".tr("e-", "a-a_").should == "hallo ^_^"
      "hello ^-^".tr("---o", "_a").should == "hella ^_^"
    end
  end
```

PR #8083, opened 2023-07-16
Fix `#line` directive filename of ripper.c

yui-knk:
Before:

```c
/* First part of user prologue.  */
#line 14 "parse.y"
```

After:

```c
/* First part of user prologue.  */
#line 14 "ripper.y"
```

### DIFF
--- a/ext/ripper/depend
+++ b/ext/ripper/depend
@@ -12,7 +12,7 @@ ripper.o: ripper.c
 
 .y.c:
 	$(ECHO) compiling compiler $<
-	$(Q) $(BISON) -t -v -o$@ -h$*.h - parse.y < $<
+	$(Q) $(BISON) -t -v -o$@ -h$*.h - $< < $<
 
 all: check
 static: check


ext/ripper/depend
```diff
@@ -12,7 +12,7 @@ ripper.o: ripper.c
 
 .y.c:
 	$(ECHO) compiling compiler $<
-	$(Q) $(BISON) -t -v -o$@ -h$*.h - parse.y < $<
+	$(Q) $(BISON) -t -v -o$@ -h$*.h - ripper.y < $<
```
nobu, line 15:
```suggestion
	$(Q) $(BISON) -t -v -o$@ -h$*.h - $< < $<
```

nobu: Or `$(top_srcdir)/parse.y`

yui-knk: `$(srcdir)/tools/preproc.rb` processes `parse.y` to generate `ripper.y` then I think `ripper.y` (`$<`) is better than `parse.y`.

PR #8084, opened 2023-07-16
Fix SEGV of ripper_s_allocate

yui-knk:
`rb_ruby_ripper_parser_allocate` calls `ruby_xcalloc` which runs GC. If GC runs inside of `rb_ruby_ripper_parser_allocate`, struct ripper allocated by `TypedData_Make_Struct` is marked even so the struct is not fully initialized. This causes SEGV when `rb_parser_t *p` is marked. Fix the order of allocation so that mark for struct ripper doesn't cause SEGV on `rb_ruby_parser_mark`.

Fix http://ci.rvm.jp/results/trunk-random1@ruby-sp2-docker/4647472

### DIFF
--- a/ext/ripper/ripper_init.c.tmpl
+++ b/ext/ripper/ripper_init.c.tmpl
@@ -109,18 +109,19 @@ static VALUE
 ripper_s_allocate(VALUE klass)
 {
     struct ripper *r;
-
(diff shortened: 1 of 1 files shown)


yui-knk:
https://github.com/ruby/ruby/pull/8085 solves the issue.

PR #8087, opened 2023-07-17
use experimental_copy instead of experimental_everything magic comment

Ivanov-Anton:
Improve Ruby Documentation


nobu:
For what purpose?

Ivanov-Anton:
@nobu because there is `# shareable_constant_value: experimental_everything`
inside `Mode experimental_copy ` definition. And it looks like a typo.

PR #8089, opened 2023-07-17
Allocate Ripper struct before use

eightbitraptor:
Ruby CI failing build: [4643684](http://ci.rvm.jp/results/trunk-random2@ruby-sp2-docker/4643684)

We need to allocate the ripper struct before we assign a pointer to it inside the T_DATA object. This is because
`rb_ruby_ripper_parse_allocate` can trigger GC, via `ruby_xcalloc`. If GC is triggered before the r->p pointer has been set, we will attempt to mark a `NULL` reference.

This bug can be replicated using the following test script.

```
require "ripper"
GC.stress = true
Ripper.new("foobar")
```

### DIFF
--- a/ext/ripper/ripper_init.c.tmpl
+++ b/ext/ripper/ripper_init.c.tmpl
@@ -108,10 +108,7 @@ ripper_lex_io_get(struct parser_params *p, VALUE src)
 static VALUE
 ripper_s_allocate(VALUE klass)
 {
-    struct ripper *r;
-
-    VALUE self = TypedData_Make_Struct(klass, struct ripper,
-                                       &parser_data_type, r);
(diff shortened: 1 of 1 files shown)


yui-knk:
Thanks for the PR. The fix was applied by https://github.com/ruby/ruby/pull/8085. At a glance the issue does not happen after https://github.com/ruby/ruby/pull/8085/.

eightbitraptor:
@yui-knk Ah. Apologies. I missed that PR. I'll close this one.

yui-knk:
No worries, thank you!

nobu:
Note that "alloc then wrap" can cause a memory leak if the "wrap" failed.

PR #8096, opened 2023-07-19
Warn for bundled gems and Gemfile

hsbt:
https://bugs.ruby-lang.org/issues/19776

### DIFF
--- a/lib/bundler/rubygems_ext.rb
+++ b/lib/bundler/rubygems_ext.rb
@@ -31,6 +31,14 @@ end
 module Gem
 
   BUNDLED_GEMS = {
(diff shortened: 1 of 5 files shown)


lib/bundler/rubygems_ext.rb
```diff
@@ -29,6 +29,29 @@
 end
 
 module Gem
+
+  BUNDLED_GEMS = {
```
voxik, line 33: I don't think this should be part of Bundler. Bundle is not the one who decides that some gem is default/bundled. It is Ruby upstream decision.

hsbt:
I merged this at https://github.com/rubygems/rubygems/pull/6831

PR #8112, opened 2023-07-24
Process.warmup: precompute strings coderange

byroot:
This both save time for when it will be eventually needed, and avoid mutating heap pages after a potential fork.

Instrumenting some large Rails app, I've witnessed up to 58% of String instances having their coderange still unknown.

### DIFF
--- a/process.c
+++ b/process.c
@@ -8555,6 +8555,7 @@ static VALUE rb_mProcID_Syscall;
  *  * Perform a major GC.
  *  * Compacts the heap.
  *  * Promotes all surviving objects to the old generation.
+ *  * Precompute the coderange of all strings.
  */
 
 static VALUE
@@ -8565,7 +8566,6 @@ proc_warmup(VALUE _)
     RB_VM_LOCK_LEAVE();
     return Qtrue;
 }
-
 
 /*
  * Document-module: Process
--- a/test/ruby/test_process.rb
+++ b/test/ruby/test_process.rb
@@ -2709,4 +2709,16 @@ EOS
       assert_equal compact_count + 1, GC.stat(:compact_count)
     end;
   end
+
+  def test_warmup_precompute_string_coderange
+    assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
+    require 'objspace'
+    begin;
+      obj = "a" * 12
+      obj.force_encoding(Encoding::BINARY)
+      assert_include(ObjectSpace.dump(obj), '"coderange":"unknown"')
+      Process.warmup
+      assert_include(ObjectSpace.dump(obj), '"coderange":"7bit"')
+    end;
+  end
 end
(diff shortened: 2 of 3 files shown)


process.c
```diff
@@ -8563,10 +8584,12 @@ proc_warmup(VALUE _)
     RB_VM_LOCK_ENTER();
```
byroot, line 8584: I wonder if we still actually need to lock the VM here, wasn't that for the transient heap?

process.c
```diff
@@ -8563,10 +8584,12 @@ proc_warmup(VALUE _)
...
     rb_gc_prepare_heap();
     RB_VM_LOCK_LEAVE();
+
+    rb_objspace_each_objects(proc_warmup_each_object, NULL);
```
byroot, line 8588: This means we iterate the whole the heap a second time.

We could do this at the same time we go over objects to mark them as candidate to oldgen as to only loop once, but it means we'd potentially compute the coderange of objects that are about to be collected, so I'm not 100% sure which is most efficient.

byroot: Hum, the issue is that I'd need to include more headers in `gc.c` to have access to `STR_NOFREE` etc.

byroot: Ok, I tried that, what do you think now?

peterzhu2118: We should test in CI if not checking `STR_NOFREE` is a problem. It shouldn't be a problem. I would prefer if we didn't handle `STR_NOFREE` as a special case.

byroot: Well, if we compute the coderange on NOFREE strings potentially about to be swept, there is a risk the pointer is garbage.

That said, are NOFREE strings even public API?

peterzhu2118: Yes, it can be created with `rb_str_new_static`.

peterzhu2118 (approved):

process.c
byroot: Hum, right. And the doc clearly state you should only pass it "string literals" meaning them being freed is no concern.

If an extension uses this for malloced memory, it's their fault.

byroot: I'll remove the flag check and merge tomorrow.

peterzhu2118: Good point, so I guess it should be fine to compute it for `STR_NOFREE` strings.

test/ruby/test_process.rb
```diff
@@ -2709,4 +2709,16 @@ def test_warmup_run_major_gc_and_compact
...
+
+  def test_warmup_precompute_string_coderange
+    assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
+    require 'objspace'
+    begin;
+      obj = "a" * 12
+      obj.force_encoding(Encoding::BINARY)
+      assert_includes(ObjectSpace.dump(obj), '"coderange":"unknown"')
+      Process.warmup
+      assert_includes(ObjectSpace.dump(obj), '"coderange":"7bit"')
```
nobu, line 2721:
```suggestion
      assert_include(ObjectSpace.dump(obj), '"coderange":"unknown"')
      Process.warmup
      assert_include(ObjectSpace.dump(obj), '"coderange":"7bit"')
```

ioquatix:
This is awesome!

PR #8113, opened 2023-07-24
Implement Weak References in the GC

peterzhu2118:
See [[Feature #19783]](https://bugs.ruby-lang.org/issues/19783).


method.h
```diff
@@ -73,7 +73,7 @@ typedef struct rb_callable_method_entry_struct { /* same fields with rb_method_e
 #define METHOD_ENTRY_COMPLEMENTED_SET(me)    ((me)->flags |= IMEMO_FL_USER3)
 #define METHOD_ENTRY_CACHED(me)              ((me)->flags & IMEMO_FL_USER4)
 #define METHOD_ENTRY_CACHED_SET(me)          ((me)->flags |= IMEMO_FL_USER4)
-#define METHOD_ENTRY_INVALIDATED(me)         ((me)->flags & IMEMO_FL_USER5)
+#define METHOD_ENTRY_INVALIDATED(me)         (UNDEF_P((VALUE)(me)) || (me)->flags & IMEMO_FL_USER5)
```
byroot, line 76: cc @tenderlove since you were worried this may add overhead to checking entries, apparently benchmarks indicate they don't?

Another possibility we mentioned with @peterzhu2118 was to have two `rb_gc_mark_weak`, one that would set to `Qundef` on sweep, that the other that would set to `0`.

That would however complexify the GC.

byroot: Hum, I was assuming whatever the default value is is already checked elsewhere. No?

byroot: Ah I see. Yeah I suppose `!= Qundef` and `!= 0` should have similarish perf.

gc.c
```diff
@@ -7018,7 +7048,7 @@ gc_mark_imemo(rb_objspace_t *objspace, VALUE obj)
         {
             const struct rb_callcache *cc = (const struct rb_callcache *)obj;
             // should not mark klass here
-            gc_mark(objspace, (VALUE)vm_cc_cme(cc));
+            rb_gc_mark_weak((VALUE *)&cc->cme_);
```
wks, line 7051: Actually the `klass` field has weak reference semantics, too.  Currently, the `klass` field is cleared by the `obj_free` function for classes.  However, the following line in `gc_ref_update_imemo` (introduced in https://github.com/ruby/ruby/commit/087a2deccfb9d99961f1ce8526b80c5f72ee9a5d) may indicate that the `klass` filed may not always contain a valid object reference:

```c
// excerpt fro gc_ref_update_imemo
if (moved_or_living_object_strictly_p(objspace, cc->klass) &&
```

It may be a result that some `imemo:callcache` instance are not invalidated when `cc->klass` dies.

We may add `rb_gc_mark_weak((VALUE *)&cc->klass);` and see if it ever contains a reference to a dead class.

peterzhu2118: Never mind the above comment. Ultimately I chose to mark the `klass` field as weak as well. I think that will fix the crashes on CI.

After 6dc15cc8895b28800d5c187929d846ac4eb7cd3f, during reference updating of compaction, we no longer directly set the klass of the CC as 0 when the CC has been invalidated. Therefore, we need mark it as weak so that it gets overwritten with a Qundef when the klass has been reclaimed.

PR #8116, opened 2023-07-25
Store initial slots per size pool

peterzhu2118:
This commit stores the initial slots per size pool, configured with the environment variables `RUBY_GC_HEAP_INIT_SIZE_%d_SLOTS`. This ensures that the configured initial slots remains a low bound for the
number of slots in the heap, which can prevent heaps from thrashing in size.

cc. @byroot


byroot:
👍

PR #8117, opened 2023-07-25
YJIT: Implement VM_CALL_ARGS_BLOCKARG with Proc for ISeq calls

XrXr:
Rack uses this. Guard that the `obj` in `the_call(&obj)` is always going
to be a Proc if the compile time sample is a Proc.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5068,6 +5068,9 @@ enum SpecVal {
     BlockParamProxy,
     PrevEP(*const VALUE),
     PrevEPOpnd(Opnd),
+    // To avoid holding values across C calls for complex calls,
+    // we might need to set the SpecVal earlier in the call sequence
+    AlreadySet,
 }
 
 // Each variant represents a branch in vm_caller_setup_arg_block.
@@ -5160,9 +5163,14 @@ fn gen_push_frame(
         }
         SpecVal::PrevEPOpnd(ep_opnd) => {
             asm.or(ep_opnd, 1.into())
-        },
+        }
+        SpecVal::AlreadySet => 0.into(), // unused
     };
-    asm.store(Opnd::mem(64, sp, SIZEOF_VALUE_I32 * -2), specval);
+    if let SpecVal::AlreadySet = frame.specval {
+        asm.comment("specval should have been set");
+    } else {
+        asm.store(Opnd::mem(64, sp, SIZEOF_VALUE_I32 * -2), specval);
+    }
 
     // Write env flags at sp[-1]
     // sp[-1] = frame_type;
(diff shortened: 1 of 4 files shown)


yjit/src/codegen.rs
```diff
@@ -6003,12 +6005,35 @@ fn gen_send_iseq(
...
+            // Simple predicate, no need for jit_prepare_routine_call().
+            let is_proc = asm.ccall(rb_obj_is_proc as _, vec![asm.stack_opnd(0)]);
+            asm.cmp(is_proc, Qfalse.into());
+            jit_chain_guard(
+                JCC_JE,
+                jit,
+                asm,
+                ocb,
+                SEND_MAX_CHAIN_DEPTH,
+                Counter::send_block_arg_type,
```
k0kubun, line 6024:
```suggestion
                Counter::guard_send_block_arg_type,
```

Counters used for guards should be prefixed with `guard_send_` to be classified as "method call exit reasons" instead of "method call fallback reasons".

yjit/src/codegen.rs
```diff
@@ -5160,9 +5163,14 @@ fn gen_push_frame(
         }
         SpecVal::PrevEPOpnd(ep_opnd) => {
             asm.or(ep_opnd, 1.into())
-        },
+        }
+        SpecVal::AlreadySet => 0.into(), // unused
     };
-    asm.store(Opnd::mem(64, sp, SIZEOF_VALUE_I32 * -2), specval);
+    if let SpecVal::AlreadySet = frame.specval {
+        asm.comment("specval should been set");
```
k0kubun, line 5170:
```suggestion
        asm.comment("specval should have been set");
```
?

yjit/src/codegen.rs
```diff
@@ -5068,6 +5068,9 @@ enum SpecVal {
     BlockParamProxy,
     PrevEP(*const VALUE),
     PrevEPOpnd(Opnd),
+    // To avoid holding values across C calls for complex calls,
+    // we might need to set the SpecVal earlier in the call sequence
+    AlreadySet,
```
k0kubun, line 5073: This is one kind of `block_handler`, so I think this (and `BlockParamProxy`) should be a variant of `BlockHandler` enum. But feel free to leave it in this PR since I've been procrastinating on refactoring `BlockParamProxy` too.

yjit/src/stats.rs
```diff
@@ -210,6 +210,7 @@ make_counters! {
     send_args_splat_super,
     send_iseq_zsuper,
     send_block_arg,
+    send_block_arg_type,
```
k0kubun, line 213: You need to rename this as well (and please place it around L299-315)

maximecb (approved):

yjit/src/codegen.rs
XrXr: 👍🏼 I'll do the refactor today. `SpecVal::None` should also go in `BlockHandler` as it puts `VM_BLOCK_HANDLER_NONE`.

PR #8106, opened 2023-07-21
YJIT: Fallback send instructions to vm_sendish

k0kubun:
`send`, `opt_send_without_block`, `invokesuper`, and `invokeblock` instructions are currently the top exit reasons for most benchmarks and production applications. It's because they side-exit when YJIT doesn't know how to optimize them.

This PR changes YJIT to call the implementation of the interpreter in such scenarios so that the execution stays in JIT code as long as possible.

## Ratio in YJIT
It improves `ratio_in_yjit` from 92.6% to 97.1% on railsbench.

### Before
<details>

```
***YJIT: Printing YJIT statistics on exit***
method call exit reasons:
                          block_arg:     63,519 (20.5%)
             iseq_has_rest_and_send:     51,188 (16.5%)
                        iseq_zsuper:     45,901 (14.8%)
                   iseq_arity_error:     30,517 ( 9.9%)
                iseq_ruby2_keywords:     24,696 ( 8.0%)
                     iseq_has_no_kw:     21,095 ( 6.8%)
          args_splat_cfunc_var_args:     16,018 ( 5.2%)
            iseq_materialized_block:     15,942 ( 5.2%)
                           kw_splat:     12,611 ( 4.1%)
                    iseq_has_kwrest:      7,172 ( 2.3%)
                        not_fixnums:      4,968 ( 1.6%)
                  klass_megamorphic:      4,286 ( 1.4%)
           iseq_missing_optional_kw:      3,942 ( 1.3%)
             args_splat_cfunc_zuper:      1,972 ( 0.6%)
        iseq_has_rest_opt_and_block:      1,959 ( 0.6%)
                      iseq_has_post:      1,958 ( 0.6%)
              cfunc_ruby_array_varg:      1,118 ( 0.4%)
                       mid_mismatch:        389 ( 0.1%)
                           keywords:         90 ( 0.0%)
    args_splat_cfunc_ruby2_keywords:         33 ( 0.0%)
                        interrupted:         17 ( 0.0%)
                      zsuper_method:         15 ( 0.0%)
invokeblock exit reasons:
      proc:      1,807 (93.5%)
    symbol:        126 ( 6.5%)
invokesuper exit reasons:
    me_changed:      4,676 (69.2%)
         block:      2,081 (30.8%)
leave exit reasons:
    interp_return:  1,678,795 (100.0%)
     se_interrupt:         17 ( 0.0%)
getblockparamproxy exit reasons:
    block_param_modified:         77 (96.2%)
[...]

PR #8122, opened 2023-07-26
YJIT: Count the number of dynamic send dispatches

k0kubun:
This counter checks the utilization of https://github.com/ruby/ruby/pull/8106. Decreasing this counter should generally contribute to performance even if it doesn't change `ratio_in_yjit`.


maximecb (approved):
👌

PR #8123, opened 2023-07-26
Add function rb_reg_onig_match

peterzhu2118:
rb_reg_onig_match performs preparation, error handling, and cleanup for matching a regex against a string. This reduces repetitive code and removes the need for StringScanner to access internal data of regex.

### DIFF
--- a/include/ruby/re.h
+++ b/include/ruby/re.h
@@ -18,6 +18,7 @@
 
 #include <stdio.h>
 
+#include "ruby/onigmo.h"
 #include "ruby/regex.h"
 #include "ruby/internal/core/rmatch.h"
 #include "ruby/internal/dllexport.h"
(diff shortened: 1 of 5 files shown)


include/ruby/re.h
```diff
@@ -105,25 +106,28 @@ long rb_reg_adjust_startpos(VALUE re, VALUE str, long pos, int dir);
...
+ * @return     Match position on a successful match, `ONIG_MISMATCH` otherwise.
  *
  * @internal
  *
- * The  return  type,  `regex_t  *`, is  defined  in  `<ruby/onigmo.h>`,  _and_
+ * The type `regex_t  *` is  defined  in  `<ruby/onigmo.h>`,  _and_
  * _conflicts_ with POSIX's  `<regex.h>`.  We can no longer  save the situation
  * at this point.  Just don't mix the two.
  */
-regex_t *rb_reg_prepare_re(VALUE re, VALUE str);
```
k0kubun, line 126: I understand that you changed strscan to stop using this, but this is still a C API exposed by this public header. Are you sure existing gems don't use this?

k0kubun: Well, strscan 3.0.7 has not been released yet, so I cannot even run railsbench with Ruby master anymore. Let me re-expose the API for now.

k0kubun: done https://github.com/ruby/ruby/commit/97219721752e7cc1eda51131d6a3cd753d37276c

peterzhu2118: I looked through gem-codesearch and found that the only gems that use it are abandoned ([match_at](https://rubygems.org/gems/match_at) and [mmapscanner](https://rubygems.org/gems/mmapscanner)).

I forgot that previous versions of strscan installed in a Gemfile still needs this function. Thanks for fixing it!

PR #8124, opened 2023-07-26
YJIT: getblockparamproxy for when block is a Proc

XrXr:


yjit/src/codegen.rs
```diff
@@ -7774,19 +7774,19 @@ fn gen_getblockparamproxy(
...
     // EP level
     let level = jit.get_arg(1).as_u32();
 
     // Peek at the block handler so we can check whether it's nil
     let comptime_handler = jit.peek_at_block_handler(level);
 
-    // When a block handler is present, it should always be a GC-guarded
-    // pointer (VM_BH_ISEQ_BLOCK_P)
-    if comptime_handler.as_u64() != 0 && comptime_handler.as_u64() & 0x3 != 0x1 {
-        incr_counter!(gbpp_not_gc_guarded);
```
XrXr, line 7789: Because this was incrementing the counter once at compile time, it wasn't counting when the generated code was exiting through this path. That was the cause for the [breakdown counters not summing to the total](https://github.com/Shopify/ruby/issues/524#issuecomment-1652100626).

yjit/src/codegen.rs
```diff
@@ -7891,7 +7928,7 @@ fn gen_getblockparam(
     asm.test(flags_opnd, VM_ENV_FLAG_WB_REQUIRED.into());
 
     // if (flags & VM_ENV_FLAG_WB_REQUIRED) != 0
-    asm.jnz(Target::side_exit(Counter::gbpp_wb_required));
+    asm.jnz(Target::side_exit(Counter::gbp_wb_required));
```
k0kubun, line 7931: > Revert "YJIT: Fix naming for a getblockparamproxy counter"

I'm confused. Are they used for different insns?

yjit/src/codegen.rs
```diff
@@ -7844,6 +7848,39 @@ fn gen_getblockparamproxy(
...
+            asm,
+            ocb,
+            SEND_MAX_DEPTH,
+            Counter::gbpp_block_handler_not_proc,
+        );
+
+        let top = asm.stack_push(Type::Unknown);
+        asm.mov(top, proc_or_false);
+    } else {
+        unreachable!("absurd given initial filtering");
```
k0kubun, line 7883: I'm tempted to put `gen_counter_incr(asm, Counter::gbpp_unsupported_type);` and `return None;` here to simplify the logic (and possibly optimize away the wasted instructions by rewinding `Assembler` in the future), but do you think this case is frequent enough?

k0kubun (approved):

yjit/src/stats.rs
```diff
@@ -355,11 +355,13 @@ make_counters! {
     expandarray_not_array,
     expandarray_rhs_too_small,
 
-    gbpp_wb_required,
-    gbpp_not_gc_guarded,
+    gbp_wb_required,
```
maximecb, line 358: Assuming this is a typo

maximecb (approved):

yjit/src/codegen.rs
XrXr: Right, the diff header for this is `gen_getblockparam(. It's for a different generator for a different instruction.

yjit/src/stats.rs
XrXr: Nah, `gbp_` is for `getblockpraram`, different from `gbpp_` for `getblockparamproxy` I got confused and renamed this counter and I'm reverting that here.

yjit/src/codegen.rs
XrXr: Move the check here, you mean? I think that makes sense. The symbol case seems uncommon enough.

yjit/src/stats.rs
maximecb, line 358:
```suggestion
    // getblockparam
    gbp_wb_required,
```

yjit/src/stats.rs
```diff
@@ -355,11 +355,13 @@ make_counters! {
...
+
+    gbpp_unsupported_type,
```
maximecb, line 360:
```suggestion
    // getblockparamproxy
    gbpp_unsupported_type,
```

yjit/src/codegen.rs
XrXr: Oh, that ends up triggering a different issue: `attempted to lower an Opnd::InsnOut` probably because the `block_handler` ends up unused :/ so it doesn't seem enticing anymore.

k0kubun: Ah, that issue. Never mind then.

maximecb:
Stats on `rack` before:
```
ratio_in_yjit:                 88.7%
avg_len_in_yjit:                43.1
Top-9 most frequent exit ops (100.0% of exits):
        getblockparamproxy:      9,972 (98.0%)
             setlocal_WC_0:         90 ( 0.9%)
               invokesuper:         87 ( 0.9%)
    opt_send_without_block:          8 ( 0.1%)
      opt_getconstant_path:          6 ( 0.1%)
                     leave:          5 ( 0.0%)
                      send:          2 ( 0.0%)
                   opt_mod:          1 ( 0.0%)
                  opt_size:          1 ( 0.0%)
```

Stats on `rack` after:
```
ratio_in_yjit:                 89.3%
avg_len_in_yjit:                43.4
Top-8 most frequent exit ops (100.0% of exits):
             setlocal_WC_0:         90 (45.0%)
               invokesuper:         87 (43.5%)
    opt_send_without_block:          8 ( 4.0%)
                     leave:          6 ( 3.0%)
      opt_getconstant_path:          6 ( 3.0%)
                  opt_aref:          1 ( 0.5%)
                      send:          1 ( 0.5%)
                  opt_plus:          1 ( 0.5%)
```

It's a nice improvement, though I'm a bit puzzled as to why our percentage in YJIT is still so low on this benchmark with so few exits 🤔

PR #8125, opened 2023-07-27
YJIT: Use dynamic dispatch for megamorphic send

k0kubun:
This PR generates the dynamic send dispatch https://github.com/ruby/ruby/pull/8106 when a callsite becomes megamorphic for receiver classes. It's currently the top exit reason on SFR.

In addition to generating that code, this PR adds another counter, `num_send_megamorphic`, which counts the number of such calls. For consistency, I renamed `get_ivar_max_depth` to `num_getivar_megamorphic`, and introduced `num_setivar_megamorphic` as well.


maximecb (approved):
Well done 👍

PR #8128, opened 2023-07-27
Reduce dependency to is_notop_id for Universal Parser

S-H-GAMELINKS:
Add `is_notop_id` macro defination in parse.y that reduce dependency C API for Universal Parser.


nobu:
Before `is_notop_id`, parse.y is using pre-defined static IDs.
It doesn't make sense to extract only this function without extracting them,

PR #8130, opened 2023-07-27
YJIT: implement missing `asm.jg` instruction in backend

maximecb:
While trying to implement a specialize integer left shift, I ran into a problem where we have no way to do a greater-than comparison at the moment. Surprising we went this far without ever needing it!


k0kubun (approved):
Is it intentional that `jit_rb_int_lshift` is pushed as well?

maximecb:
> Is it intentional that `jit_rb_int_lshift` is pushed as well?

Yes. It's commented out for now.

PR #8141, opened 2023-07-29
RDoc for module Process

BurdetteLamar:
Adds a "What's Here" section.  This is partly for me to begin to understand this module.  As I add sections in the main discussion, this may become redundant and unnecessary. Therefore I'm marking this as a draft, but also asking @peterzhu2118 for review.

### DIFF
--- a/process.c
+++ b/process.c
@@ -8570,7 +8570,7 @@ proc_warmup(VALUE _)
 /*
  * Document-module: Process
  *
- * \Module \Process represents a process in the underlying operating system.
+ * \Module +Process+ represents a process in the underlying operating system.
  * Its methods support management of the current process and its child processes.
  *
  * == What's Here
--- a/process.c
+++ b/process.c
@@ -8571,7 +8571,7 @@ proc_warmup(VALUE _)
  * Document-module: Process
  *
  * \Module \Process represents a process in the underlying operating system.
- * Its methods support management of the current process and its subprocesses.
+ * Its methods support management of the current process and its child processes.
  *
  * == What's Here
  *
@@ -8584,7 +8584,7 @@ proc_warmup(VALUE _)
  * - ::getrlimit: Returns the resource limit.
  * - ::gid: Returns the (real) group ID.
  * - ::pid: Returns the process ID.
- * - ::ppid: Returns the process ID of the parent.
+ * - ::ppid: Returns the process ID of the parent process.
  * - ::uid: Returns the (real) user ID.
  *
  * === Current-Process Setters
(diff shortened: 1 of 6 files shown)


process.c
```diff
@@ -8570,19 +8570,88 @@ proc_warmup(VALUE _)
...
+ * - ::exit: Initiates process termination by raising exception SystemExit
+ *   (which may be caught).
+ * - ::exit!: Immediately exits the process.
+ * - ::warmup: Notifies the Ruby virtual machine that the boot sequence
+ *   for the application is completed,
+ *   and that the VM may begin optimizing the application.
+ *
+ * === Subprocesses
+ *
+ * - ::_fork: An internal method for forking; should not be called directly.
```
peterzhu2118, line 8617: We shouldn't document this. It's not a recommended API to use.

peterzhu2118 (commented):
I think this document looks good so far.

BurdetteLamar:
Should we say "child process" instead of "subprocess"; seems in keeping with the rest of the world.

process.c
BurdetteLamar: My thought, too.  But existing doc does say it can be overridden.

peterzhu2118:
I like "child process" more than "subprocess" to be consistent with the Linux manual.

process.c
BurdetteLamar: Removed (from What's Here).

BurdetteLamar:
> I like "child process" more than "subprocess" to be consistent with the Linux manual.

Changed.

process.c
```diff
@@ -8570,19 +8570,87 @@ proc_warmup(VALUE _)
...
- *   * Low-level introspection of other processes, like Process.getpgid, Process.getpriority;
- *   * Management of the current process: Process.abort, Process.exit, Process.daemon, etc.
- *     (for convenience, most of those are also available as global functions
- *     and module functions of Kernel);
- *   * Creation and management of child processes: Process.fork, Process.spawn, and
- *     related methods;
- *   * Management of low-level system clock: Process.times and Process.clock_gettime,
- *     which could be important for proper benchmarking and other elapsed
- *     time measurement tasks.
+ * \Module \Process represents a process in the underlying operating system.
```
nobu, line 8573: To escape `Module` is that it is a common word, but `Process` is not a common word here.
I think it is better to mark up as code.
```suggestion
 * \Module +Process+ represents a process in the underlying operating system.
```

nobu:
Frankly speaking, I'm not sure why you want to avoid to link to classes/modules.

process.c
BurdetteLamar: @nobu, we have hundreds (if not thousands) of instances like \Process.  If ok with you, I'd like to defer this and propose a usage in the documentation guide.

process.c
```diff
@@ -8570,19 +8570,87 @@ proc_warmup(VALUE _)
...
+ * - ::exit!: Immediately exits the process.
+ * - ::warmup: Notifies the Ruby virtual machine that the boot sequence
+ *   for the application is completed,
+ *   and that the VM may begin optimizing the application.
+ *
+ * === Child Processes
+ *
+ * - ::detach: Guards against a child process becoming a zombie.
+ * - ::fork: Creates a child process.
+ * - ::kill: Sends a given signal to processes.
```
nobu, line 8619: `kill` is not only for child processes.
But I can't say it should not be here since it is the typical usage...

BurdetteLamar:
@nobu @peterzhu2118 I think all is resolved;  ok to squash and merge?

process.c
```diff
@@ -8570,19 +8570,87 @@ proc_warmup(VALUE _)
...
+ * - ::spawn: Creates a chile process.
```
peterzhu2118, line 8620: Typo

```suggestion
 * - ::spawn: Creates a child process.
```

peterzhu2118 (approved):

BurdetteLamar:
> Frankly speaking, I'm not sure why you want to avoid to link to classes/modules.

@nobu, are you content with https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html#label-Auto-Linking?  If so, I'll be more strict with myself.

PR #8152, opened 2023-08-01
YJIT: Fallback setivar if the next shape is too complex

k0kubun:
Currently, `setivar_too_complex` is one of the most frequent exit ops on SFR (e.g. 21.7% overall). This exit happens when it needs to transition to a too-complex shape.

This PR hoists out the logic to get the `new_shape_id` (as well as other information that is used to get `new_shape_id` and reused later) and check if it's `OBJ_TOO_COMPLEX_SHAPE_ID` before generating any code. If it is, it fallbacks to the generic setivar C call.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2188,10 +2188,66 @@ fn gen_setinstancevariable(
         gen_counter_incr(asm, Counter::num_setivar_megamorphic);
     }
 
+    // Get the iv index
+    let shape_too_complex = comptime_receiver.shape_too_complex();
+    let ivar_index = if !shape_too_complex {
+        let shape_id = comptime_receiver.shape_id_of();
+        let shape = unsafe { rb_shape_get_shape_by_id(shape_id) };
+        let mut ivar_index: u32 = 0;
+        if unsafe { rb_shape_get_iv_index(shape, ivar_name, &mut ivar_index) } {
+            Some(ivar_index as usize)
+        } else {
+            None
+        }
+    } else {
+        None
+    };
+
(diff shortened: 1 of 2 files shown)


maximecb (approved):

casperisfine:
I've witnessed YJIT specific crashes since this was merged.

Here's a relatively small repro script:

```ruby
require 'bundler/inline'

require 'objspace'
# run with --yjit --yjit-call-threshold=1

gemfile do
  source 'https://rubygems.org'
  gem 'mocha', '2.0.4'
  gem 'minitest', '5.18.1'
end

module MyConfig
end

module OtherModule
end
223.times do |iv|
  MyConfig.instance_variable_set("@_iv_#{iv}", 42)
end

require "minitest/autorun"
require "mocha/minitest"

class Module
  def mocha(instantiate = true)
    if instantiate
      @mocha ||= Mocha::Mockery.instance.mock_impersonating(self)
    else
      defined?(@mocha) ? @mocha : nil
    end
  end
end

class BugTest < Minitest::Test
  def test_mocha
    10.times do
      OtherModule.mocha
    end
    puts ObjectSpace.dump(MyConfig)
    puts ObjectSpace.dump(MyConfig.singleton_class)
    MyConfig.mocha
  end
end
```

casperisfine:
Ok, so from what I gathered, the interpreter doesn't run into this, because `vm_setivar` just bails out when on a `T_MODULE` or `T_CLASS`, so we end up in the slow path.

From why I understand of the code, YJIT seem to always use the fast path, even with `T_MODULE` / `T_CLASS` hence why it runs into that bug.

PR #8160, opened 2023-08-02
YJIT: Fallback setivar if the receiver isn't T_OBJECT

casperisfine:
Followup: https://github.com/ruby/ruby/pull/8152

If the receiver is a T_MODULE or T_CLASS and has a lot of ivars, `get_next_shape_internal` will return `NULL`.

cc @k0kubun @maximecb (I still need to add a regression test).


k0kubun (approved):
Thanks!

PR #8161, opened 2023-08-02
YJIT: handle expandarray_rhs_too_small case

maximecb:


yjit/src/codegen.rs
```diff
@@ -1500,49 +1500,89 @@ fn gen_expandarray(
...
+        let heap_ptr_opnd = Opnd::mem(
+            usize::BITS as u8,
+            asm.load(array_opnd),
+            RUBY_OFFSET_RARRAY_AS_HEAP_PTR,
+        );
+        let ary_opnd = asm.csel_nz(ary_opnd, heap_ptr_opnd);
+
+        // Loop backward through the array and push each element onto the stack.
+        for i in (0..num).rev() {
+            let top = asm.stack_push(Type::Unknown);
```
k0kubun, line 1578: If it's useful, we could upgrade the type to `Type::Nil` for `i >= comptime_len`.

maximecb: Good idea 👍 will make that change

yjit/src/codegen.rs
```diff
@@ -1500,49 +1500,89 @@ fn gen_expandarray(
...
+        asm.test(flags_opnd, Opnd::UImm(RARRAY_EMBED_FLAG as u64));
+        let heap_ptr_opnd = Opnd::mem(
+            usize::BITS as u8,
+            asm.load(array_opnd),
+            RUBY_OFFSET_RARRAY_AS_HEAP_PTR,
+        );
+        let ary_opnd = asm.csel_nz(ary_opnd, heap_ptr_opnd);
+
+        // Loop backward through the array and push each element onto the stack.
+        for i in (0..num).rev() {
```
k0kubun, line 1577: I wonder if it makes sense to share this loop for `comptime_len == 0` case as well, e.g.

```rs
let ary_opnd = if comptime_len > 0 {
    // ...
    Some(asm.csel_nz(ary_opnd, heap_ptr_opnd))
} else {
    None
};

for i in (0..num).rev() {
    // ...
    let elem_opnd = if (i as u32) < comptime_len { Opnd::mem(64, ary_opnd.unwrap(), offset) } else { Qnil.into() };
    asm.mov(top, elem_opnd);
}
```

yjit/src/codegen.rs
```diff
@@ -7394,7 +7434,7 @@ fn gen_leave(
     ocb: &mut OutlinedCb,
 ) -> Option<CodegenStatus> {
     // Only the return value should be on the stack
-    assert_eq!(1, asm.ctx.get_stack_size());
+    assert_eq!(1, asm.ctx.get_stack_size(), "leave instruction expects stack size 1");
```
k0kubun, line 7437: I would like to keep the information about the actual value.

```suggestion
    assert_eq!(1, asm.ctx.get_stack_size(), "leave instruction expects stack size 1, but was: {}", asm.ctx.get_stack_size());
```

yjit/src/codegen.rs
```diff
@@ -1500,49 +1500,89 @@ fn gen_expandarray(
...
+    let array_opnd = asm.stack_opnd(0);
     let array_reg = asm.load(array_opnd);
     let array_len_opnd = get_array_len(asm, array_reg);
 
-    // Only handle the case where the number of values in the array is greater
-    // than or equal to the number of values requested.
-    asm.cmp(array_len_opnd, num.into());
-    asm.jl(Target::side_exit(Counter::expandarray_rhs_too_small));
+    // Guard on the comptime/expected array length
+    asm.comment(&format!("guard array length == {}", comptime_len));
```
k0kubun, line 1539: Nice solve for the register spill issue :+1:

yjit/src/codegen.rs
```diff
@@ -1500,49 +1500,89 @@ fn gen_expandarray(
...
+    if !jit.at_current_insn() {
+        defer_compilation(jit, asm, ocb);
+        return Some(EndBlock);
+    }
+
+    let comptime_recv = jit.peek_at_stack(&asm.ctx, 0);
+
+    // If the comptime receiver is not an array, bail
+    if comptime_recv.class_of() != unsafe { rb_cArray } {
+        return None;
```
k0kubun, line 1513: Can we have a counter for this?

k0kubun (approved):

PR #8162, opened 2023-08-02
Remove --disable-gems for assert_separately

peterzhu2118:
assert_separately adds --disable=gems so we don't need to add --disable-gems when calling assert_separately.


nobu (approved):
I haven't noticed there're "gem" and "gems".

PR #8165, opened 2023-08-03
No computing embed_capa_max in str_subseq

wks:
Fix str_subseq so that it does not attempt to predict the size of the object returned by str_alloc_heap.

This is an alternative approach for https://github.com/ruby/ruby/pull/8151  This PR is simpler in the sense that it changes only one function, `str_subseq`.  It duplicates some of the code in `str_replace_shared_without_enc`.  `str_subseq` now uses the same if-condition as `str_replace_shared_without_enc`, namely `str_embed_capa(str2) >= len + termlen`, so the two functions should always make the same decision about whether to return an embedded string or heap string.

### DIFF
--- a/string.c
+++ b/string.c
@@ -2732,15 +2732,30 @@ str_subseq(VALUE str, long beg, long len)
 {
     VALUE str2;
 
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):
Looks good. This looks much better! Thank you!

PR #8166, opened 2023-08-03
Don't check stack for moved after compaction

peterzhu2118:
We don't need to check stack for moved objects after compaction because the mutator cannot run between marking the stack and the end of compaction. However, the stack may have moved objects leftover from marking and sweeping phases. This means that their pages will be invalidated and all objects moved back. We don't need to move these objects back.

This also fixes the issue on Windows where some compaction tests sometimes fail due to the page of the object being invalidated.

cc. @tenderlove

### DIFF
--- a/test/ruby/test_gc_compact.rb
+++ b/test/ruby/test_gc_compact.rb
@@ -310,7 +310,6 @@ class TestGCCompact < Test::Unit::TestCase
 
   def test_moving_arrays_down_size_pools
     omit if GC::INTERNAL_CONSTANTS[:SIZE_POOL_COUNT] == 1
-    pend "Flaky on some platforms" if /solaris|mswin/i =~ RUBY_PLATFORM
 
     assert_separately(%w[-robjspace], "#{<<~"begin;"}\n#{<<~"end;"}", timeout: 10, signal: :SEGV)
     begin;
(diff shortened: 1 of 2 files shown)


test/ruby/test_gc_compact.rb
```diff
@@ -354,7 +352,7 @@ def test_moving_arrays_up_size_pools
 
   def test_moving_objects_between_size_pools
     omit if GC::INTERNAL_CONSTANTS[:SIZE_POOL_COUNT] == 1
-    pend "Flaky on some platforms" if /solaris|mswin/i =~ RUBY_PLATFORM
+    omit "Flaky on Solaris" if /solaris/i =~ RUBY_PLATFORM
```
nobu, line 355: Still flaky on Solaris?

peterzhu2118: The issue on Solaris is different. The issue on Solaris is:

```
TestGCCompact#test_moving_objects_between_size_pools [/export/home/users/chkbuild/cb-gcc/tmp/build/20230731T230003Z/ruby/test/ruby/test_gc_compact.rb:378]:
Expected 499 to be >= 500.
```

This means that the object was not moved at all, which this PR does not fix. This PR fixes when the object is moved, but is moved back due to reverting the stack. I haven't debugged why this test is flaky on Solaris.

My suspicion is that the last object being created remains in the stack or in a register on Solaris causing the object to be pinned, which I don't know how to fix.

PR #8169, opened 2023-08-03
YJIT: guard for array_len >= num in expandarray

maximecb:
Avoid generating long dispatch chains for all array lengths seen.


k0kubun (approved):
Nice!

PR #8170, opened 2023-08-04
Remove uneeded int2big property for Universal Parser

S-H-GAMELINKS:
I was reading Universal Parser and parse.y and it seems that int2big is not used in parse.y.
Therefore, it has been removed.


yui-knk:
Thank you!

PR #8171, opened 2023-08-04
YJIT: Compile exception handlers

k0kubun:
This PR supports compiling exception handlers in YJIT.

## ratio_in_yjit
This improves `ratio_in_yjit`, e.g.

* liquid-compile: 94.2% -> 99.9%
* liquid-render: 89.5% -> 99.9%
* sequel: 97.8% -> 99.9%

## Benchmark
This speeds up liquid-render by 25%, liquid-compile by 6%, and sequel by 4%.

```
before: ruby 3.3.0dev (2023-08-08T00:21:45Z master 69292676a2) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-08-08T03:57:11Z yjit-exceptions 9b1b055055) +YJIT [x86_64-linux]

liquid-compile  41.4         0.6         39.1        0.5         1.01           1.06
liquid-render   71.7         0.2         57.5        0.2         1.12           1.25
sequel          49.0         1.8         47.1        2.1         1.02           1.04
```

Other benchmarks don't seem to have a significant performance impact.

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -2427,7 +2427,8 @@ vm_exec_loop(rb_execution_context_t *ec, enum ruby_tag_type state,
 
     rb_ec_raised_reset(ec, RAISED_STACKOVERFLOW | RAISED_NOMEMORY);
     while (UNDEF_P(result = vm_exec_handle_exception(ec, state, result))) {
-        /* caught a jump, exec the handler */
+        // caught a jump, exec the handler. JIT code in jit_exec_exception()
+        // may return Qundef to run remaining frames with vm_exec_core().
         if (UNDEF_P(result = jit_exec_exception(ec))) {
             result = vm_exec_core(ec);
         }
--- a/vm.c
+++ b/vm.c
@@ -370,7 +370,14 @@ extern VALUE rb_vm_invoke_bmethod(rb_execution_context_t *ec, rb_proc_t *proc, V
 static VALUE vm_invoke_proc(rb_execution_context_t *ec, rb_proc_t *proc, VALUE self, int argc, const VALUE *argv, int kw_splat, VALUE block_handler);
 
 #if USE_RJIT || USE_YJIT
-// Try to compile the current ISeq in ec. Return 0 if not compiled.
+// Generate JIT code that supports the following kinds of ISEQ entries:
+//   * The first ISEQ on vm_exec (e.g. <main>, or Ruby methods/blocks
+//     called by a C method). The current frame has VM_FRAME_FLAG_FINISH.
+//     The current vm_exec stops if JIT code returns a non-Qundef value.
+//   * ISEQs called by the interpreter on vm_sendish (e.g. Ruby methods or
+//     blocks called by a Ruby frame that isn't compiled or side-exited).
+//     The current frame doesn't have VM_FRAME_FLAG_FINISH. The current
+//     vm_exec does NOT stop whether JIT code returns Qundef or not.
 static inline rb_jit_func_t
 jit_compile(rb_execution_context_t *ec)
 {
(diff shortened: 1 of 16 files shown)


yjit/src/codegen.rs
```diff
@@ -662,31 +662,33 @@ pub fn gen_entry_prologue(cb: &mut CodeBlock, ocb: &mut OutlinedCb, iseq: IseqPt
...
-        Opnd::mem(64, CFP, RUBY_OFFSET_CFP_JIT_RETURN),
-        Opnd::const_ptr(CodegenGlobals::get_leave_exit_code().raw_ptr()),
+    asm.ccall(
+        rb_yjit_set_jit_return as _,
+        vec![
+            CFP,
+            CodegenGlobals::get_leave_compile_code().into_usize().into(),
+            CodegenGlobals::get_leave_exit_code().into_usize().into(),
+            handle_exception,
+        ],
```
maximecb, line 681: I don't love that you're adding an extra parameter to every JIT entry point and also calling a C function in there... We want entry points to be as streamlined as possible.

Could you compile a different kind of entry point if `handle_exception` is true?

yjit/src/codegen.rs
```diff
@@ -8293,6 +8293,9 @@ pub struct CodegenGlobals {
     /// Code for exiting back to the interpreter from the leave instruction
     leave_exit_code: CodePtr,
 
+    /// Code for compiling and jumping to the caller
+    leave_compile_code: CodePtr,
+
```
maximecb, line 8298: Would it make sense to call this `leave_exception_code` instead of `leave_compile_code` so that its purpose is more clear?

yjit/src/core.rs
```diff
@@ -2397,6 +2407,81 @@ fn new_pending_branch(jit: &mut JITState, gen_fn: BranchGenFn) -> PendingBranchR
...
+            leave_compile_hit_body(ec).unwrap_or(
+                CodegenGlobals::get_stub_exit_code().raw_ptr()
+            )
+        })
+    }
+}
+
+/// This is called after a frame is popped. The current CFP points to the caller,
+/// and this function attempts to compile and execute the caller.
+fn leave_compile_hit_body(ec: EcPtr) -> Option<*const u8> {
```
maximecb, line 2425: Worth documenting in the comment that this is used for exceptions and for compiling somewhere in the middle of the caller?

yjit/src/codegen.rs
k0kubun: > Could you compile a different kind of entry point if handle_exception is true?

Yeah, I think we could. We could let each ISEQ have two function pointers instead of one, e.g. `jit_func` and `jit_exception`. `jit_exec(ec, false)` calls `jit_func`, and `jit_exec(ec, true)` calls `jit_exception`. Then `jit_func` can be as streamlined as it currently is. `jit_exception` still needs to be complicated, but at least this would leave existing paths intact.

I hesitated to do that because that would complicate the implementation a bit, but I'll work on it to streamline `jit_func`, limiting the complex setup to `jit_exception` paths.

maximecb: But will each ISEQ be guaranteed to have only one entry point for handling exceptions? Can't exceptions be caught at multiple different locations in the ISEQ?

Also it's not that great if there is still a conditional check in `jit_exec`. Hmmm. Would it be possible to separate out the logic for handling exections into a function separate from `jit_exec`? 😅  That way there would be no boolean argument and no if statement for this?

k0kubun: Hopefully, I think my proposed solution actually addresses all of your concerns 🙂  I'll try to push that as soon as possible.

vm.c
```diff
@@ -372,7 +372,7 @@ static VALUE vm_invoke_proc(rb_execution_context_t *ec, rb_proc_t *proc, VALUE s
 #if USE_RJIT || USE_YJIT
 // Try to compile the current ISeq in ec. Return 0 if not compiled.
 static inline rb_jit_func_t
-jit_compile(rb_execution_context_t *ec)
+jit_compile(rb_execution_context_t *ec, bool handle_exception)
```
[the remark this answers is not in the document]
k0kubun: I tried my best to explain relevant details above `jit_compile` at https://github.com/ruby/ruby/pull/8171/commits/e5cbb368b6bcefa7736522181d1af07258eb9bad.

yjit/src/codegen.rs
k0kubun: This is what I was thinking of https://github.com/ruby/ruby/pull/8171/commits/5d79afbe66435e9a4735cc1bcc31814db39c503e. With this patch, existing `jit_exec` behaviors do not change.

yjit/src/codegen.rs
k0kubun: Sounds better :+1: https://github.com/ruby/ruby/pull/8171/commits/a83e4626a0564cd97a678ec0c3ef07c65620ce89

yjit/src/core.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/8171/commits/b55383c85d31f2df9f5dbbd5775d70dc1b8eb3f2

k0kubun:
> interfacing with the interpreter

I reworked the callback in https://github.com/ruby/ruby/pull/8171/commits/9b1b055055e179154d44a51d81eea50abdadbb83. It's still as complex as it needs to be, but at least it's much faster than the previous version.

vm.c
```diff
@@ -2382,7 +2428,9 @@ vm_exec_loop(rb_execution_context_t *ec, enum ruby_tag_type state,
     rb_ec_raised_reset(ec, RAISED_STACKOVERFLOW | RAISED_NOMEMORY);
     while (UNDEF_P(result = vm_exec_handle_exception(ec, state, result))) {
         /* caught a jump, exec the handler */
-        result = vm_exec_core(ec);
+        if (UNDEF_P(result = jit_exec_exception(ec))) {
+            result = vm_exec_core(ec);
+        }
```
maximecb, line 2433: Could you add a comment to explain what `result == undef` means in this context?

yjit/src/codegen.rs
```diff
@@ -672,19 +712,35 @@ pub fn gen_entry_prologue(cb: &mut CodeBlock, ocb: &mut OutlinedCb, iseq: IseqPt
     asm.mov(SP, Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SP));
 
     // Setup cfp->jit_return
-    asm.mov(
-        Opnd::mem(64, CFP, RUBY_OFFSET_CFP_JIT_RETURN),
-        Opnd::const_ptr(CodegenGlobals::get_leave_exit_code().raw_ptr()),
-    );
+    if jit_exception {
```
maximecb, line 715:
```suggestion
    // If this is an exception handler entry point
    if jit_exception {
```

maximecb (approved):
Good work 👍 I'm excited to see the results on SFR!

vm.c
k0kubun: sure, added a comment in https://github.com/ruby/ruby/pull/8171/commits/0bc8f420a22157a06ae2bb340441ffd3ccd0b173

PR #8177, opened 2023-08-05
Fixup lib/bundled_gems.rb

MSP-Greg:
Current master builds throw an error when a file like `drb/drb` is required.  Error similar to:
```
/usr/local/lib/ruby/3.3.0+0/bundled_gems.rb:80:in `<': comparison of String with nil failed (ArgumentError)
```
One way to test this is to copy the current lib/bundled_gems.rb file into a recent master build, then run
```
ruby -rdrb -e "puts 'Throws error'"
```

First, all ruby-loco builds failed.  Then I did the above with recent master builds on both Windows & Ubuntu.  the above command line failed.

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -67,6 +67,7 @@ module Gem::BUNDLED_GEMS
     return unless gem = find_gem(path)
     caller, = caller_locations(3, 1)
     return if find_gem(caller&.absolute_path)
+    name = name[%r[\A[^/]+]]
     return if WARNED[name]
     WARNED[name] = true
     if gem == true


MSP-Greg:
With the current code, when passed a file name like `drb/drb`, the variable `gem` will be set to that value.  It then hits the code below, and `SINCE[gem]` returns `nil`, which throws an error.
https://github.com/ruby/ruby/blob/f07ef1d54c3bbae62e2ddd7266932c294db91daf/lib/bundled_gems.rb#L80

nobu:
`Gem::BUNDLED_GEMS.find_gem` should return `"drb"` for `"drb/drb"` instread of `true`.

nobu:
https://github.com/ruby/ruby/pull/8178

MSP-Greg:
@nobu 

With your PR and the below
```
ruby -rdrb/drb -e "puts ''"
```
nothing is warned.  With this PR, it is.

nobu:
```shell-session
$ ~/.rbenv/versions/3.3.0-2023-08-05/bin/ruby -v
ruby 3.3.0dev (2023-08-05T09:56:37Z master 4e6861d337) [x86_64-darwin22]

$ ~/.rbenv/versions/3.3.0-2023-08-05/bin/ruby -rdrb/drb -ep
warning: drb/drb is found in drb which will be not part of the default gems since Ruby 3.4.0
```

MSP-Greg:
Closing.

See 4e6861d3376eb7 and #8178

PR #8179, opened 2023-08-05
[DOC] RDoc for Process

BurdetteLamar:
Can the examples at Process.wait be shorter?


process.c
```diff
@@ -540,18 +541,15 @@ get_ppid(void)
...
- *     Process.ppid   -> integer
+ *    Process.ppid -> integer
  *
- *  Returns the process id of the parent of this process. Returns
- *  untrustworthy value on Win32/64. Not available on all platforms.
+ *  Returns the process ID of the parent of the current process:
  *
- *     puts "I am #{Process.pid}"
- *     Process.fork { puts "Dad is #{Process.ppid}" }
+ *    Process.ppid # => 16112
```
peterzhu2118, line 548: We should use a similar example as the previous one and show that the `ppid` does not change across forking.

process.c
```diff
@@ -1274,48 +1279,135 @@ proc_wait(int argc, VALUE *argv)
...
+ *      Process.wait(pid0)                            # Raises Errno::ECHILD
+ *
+ *  - <tt>0</tt>: Waits for any child process whose group ID
+ *    is the same as that of the current process:
+ *
+ *      parent_pgpid = Process.getpgid(Process.pid)
+ *      puts "Parent process group ID is #{parent_pgpid}."
+ *      child0_pid = fork do
+ *      puts "Child 0 pid is #{Process.pid}"
+ *      child0_pgid = Process.getpgid(Process.pid)
```
peterzhu2118, line 1306: Indentation is off:

```suggestion
 *        puts "Child 0 pid is #{Process.pid}"
 *        child0_pgid = Process.getpgid(Process.pid)
```

process.c
```diff
@@ -1274,48 +1279,135 @@ proc_wait(int argc, VALUE *argv)
...
+ *      child1_pid = fork do
+ *        puts "Child 1 pid is #{Process.pid}"
+ *        Process.setpgid(0, Process.pid)
+ *        child1_pgid = Process.getpgid(Process.pid)
+ *        puts "Child 1 process group ID is #{child1_pgid} (different from parent's)."
+ *      end
+ *      retrieved_pid = Process.wait(-1)
+ *      puts "Process.wait(-1) returned pid #{retrieved_pid}, which is child 0 pid."
+ *      retrieved_pid = Process.wait(-1)
+ *      puts "Process.wait(-1) returned pid #{retrieved_pid}, which is child 1 pid."
```
peterzhu2118, line 1351: I don't think we can assume which of the `retrieved_pid` is child 0 and child 1. The ordering will depend on which one of the two children finish first.

process.c
BurdetteLamar: Similar to which example, please?

process.c

process.c
BurdetteLamar: Fixed (child 1 sleeps).

process.c
peterzhu2118: The example that was removed. So something like:

```ruby
puts "PID is #{Process.pid}"
fork { puts "Parent PID is #{Process.ppid}" }
```

process.c
```diff
@@ -1344,6 +1344,7 @@ proc_wait(int argc, VALUE *argv)
  *        Process.setpgid(0, Process.pid)
  *        child1_pgid = Process.getpgid(Process.pid)
  *        puts "Child 1 process group ID is #{child1_pgid} (different from parent's)."
+ *        sleep 3 # To force child 1 to exit later than child 0 exit.
```
peterzhu2118, line 1347: While this `sleep` does make it very likely that child 1 exits later than child 0, it does not guarantee it. We should instead change the example to not assume that the first `Process.wait` returns child 0 and the second returns child 1.

BurdetteLamar: Fixed.

process.c
BurdetteLamar: Restored.

process.c
```diff
@@ -518,12 +518,13 @@ clear_pid_cache(void)
...
+ *    Process.pid -> integer
  *
- *  Returns the process id of this process. Not available on all
- *  platforms.
+ *  Returns the process ID of the current process:
+ *
+ *    Process.pid # => 15668
  *
- *     Process.pid   #=> 27415
+ *  Not available on all platforms.
```
nobu, line 527: AFAIK, `Process.pid` seems available on all supported platforms.

process.c
```diff
@@ -540,18 +541,21 @@ get_ppid(void)
...
+ *  Output:
  *
- *     I am 27417
- *     Dad is 27417
+ *    Pid is 271290.
+ *    Parent pid is 271290.
+ *
+ *  Not available on all platforms.
+ *
+ *  Returns an untrustworthy value on Windows.
```
nobu, line 558: The availability is same as `pid`, but it is always 0 on Wasm.
As for Windows, I guess "an untrustworthy value" means that it is 0 on old Windows 9x, but not sure.

```suggestion
 *  Does not return a trustworthy value on all platforms.
```

process.c
```diff
@@ -624,18 +628,25 @@ rb_last_status_get(void)
...
- *     Process.wait Process.spawn("ruby", "-e", "exit 13")
- *     Process.last_status   #=> #<Process::Status: pid 4825 exit 13>
+ *    Process.spawn('ruby', '-e', 'exit 14')
+ *    Process.wait
+ *    Process.last_status # => #<Process::Status: pid 4692 exit 14>
  *
- *  If no child process has ever been executed in the current
- *  thread, this returns +nil+.
+ *    Process.spawn('ruby', '-e', 'exit 15')
+ *    # 'exit 15' has not exited.
```
nobu, line 645: The `last_status` is set by `Process.wait`, or not changed until `wait`.
```suggestion
 *    # 'exit 15' has not been reaped by #wait.
```

process.c
peterzhu2118: "Does not return a trustworthy value on all platforms." sounds like it never returns a trustworthy value on any platforms. How about "May not return a trustworthy value on certain platforms."

process.c
BurdetteLamar: Removed.

process.c
BurdetteLamar: Modified (per @peterzhu2118).  Okay with you, @nobu?

process.c

process.c
```diff
@@ -638,11 +635,12 @@ rb_last_status_get(void)
  *    Process.last_status # => #<Process::Status: pid 14396 exit 13>
  *
  *    Process.spawn('ruby', '-e', 'exit 14')
+
```
nobu, line 638: An empty line?

BurdetteLamar: Removed.

PR #8182, opened 2023-08-06
Refactor VM exec loop

nobu:


vm.c
```diff
@@ -2299,113 +2299,97 @@ hook_before_rewind(rb_execution_context_t *ec, const rb_control_frame_t *cfp,
...
-
-static void
-vm_exec_enter_vm_loop(rb_execution_context_t *ec, struct rb_vm_exec_context *ctx,
-                      struct rb_vm_tag *_tag, bool skip_first_ex_handle)
+static VALUE
+vm_exec_loop(rb_execution_context_t *ec, enum ruby_tag_type state,
+             struct rb_vm_tag *_tag, VALUE result)
 {
-    if (skip_first_ex_handle) {
+    if (state == TAG_NONE) { /* no jumps, result is discarded */
```
kateinoigakukun, line 2306: I was worried about the overhead of this extra branch on non-wasm paths, but according to vm_eval benchmark result, it doesn't seem to be a concern.

```
$ make benchmark ITEM=vm_eval.yml COMPARE_RUBY=path/to/build/x86_64-linux-master/miniruby
```

|         |compare-ruby|built-ruby|
|:--------|-----------:|---------:|
|vm_eval  |    289.780k|  293.964k|
|         |           -|     1.01x|

kateinoigakukun (approved):

vm.c
```diff
@@ -2299,113 +2299,97 @@ hook_before_rewind(rb_execution_context_t *ec, const rb_control_frame_t *cfp,
...
+
+    return result;
 }
 
+struct rb_vm_exec_context {
+    rb_execution_context_t *const ec;
+    struct rb_vm_tag *const tag;
+
+    VALUE result;
+};
```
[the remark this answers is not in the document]
nobu: Thanks, it’s a hangover that I tried the struct in the both sides.

vm.c
nobu: By the optimization hints, the branch is optimized away on gcc 12 and Apple's clang at least.

nobu: Correction:

In the non-wasm side, since the `vm_exec_loop` calls are in branches for the same condition, the condition in this function is invariant and no additional branches are added if this function is inlined.

The optimization hints optimize away branches after `ruby_setjmp()` returned non-zero in `EC_EXEC_TAG() == TAG_NONE`, i.e., `RBIMPL_ASSUME` in `rb_ec_tag_state`.

k0kubun (approved):
:+1:

nobu:
@kateinoigakukun If `try_f` and `catch_f` could return a value and `rb_wasm_try_catch_loop_run` would return it, `result` in `struct rb_vm_exec_context` can be removed.

PR #8183, opened 2023-08-07
Resync YARP

jemmaissroff:


nobu (changes requested):
Please re-sync with the latest sync_default_gems.rb.

jemmaissroff:
> Please re-sync with the latest sync_default_gems.rb.

@nobu it usually syncs through this mechanism. However, I was OOO last week, and early in the week there was a change to common.mk that was necessary. Without this, and due to a few warnings that appeared in Ruby CI, the sync script wasn't able to succeed. It's now behind. My goal with this PR (which was generated by running the script locally, and making the necessary common.mk changes) is to catch up the sync script and allow it to proceed as normally. Do you have other suggestions for how to accomplish this?

k0kubun:
Closing this one since we are working on this at https://github.com/ruby/ruby/pull/8226 instead.

PR #8184, opened 2023-08-08
[DOC] Mention the omission of a superclass when reopening a class

kyanagi:


doc/syntax/modules_and_classes.rdoc
```diff
@@ -40,8 +40,8 @@ functionality:
     remove_method :my_method
   end
 
-Reopening classes is a very powerful feature of Ruby, but it is best to only
-reopen classes you own.  Reopening classes you do not own may lead to naming
+Reopening modules (or classes) is a very powerful feature of Ruby, but it is best to only
+reopen modules you own.  Reopening modules you do not own may lead to naming
 conflicts or difficult to diagnose bugs.
```
nobu, line 45:
```suggestion
Reopening modules (or classes) is a very powerful feature of Ruby, but it is
best to only reopen modules you own.  Reopening modules you do not own may lead
to naming conflicts or difficult to diagnose bugs.
```

doc/syntax/modules_and_classes.rdoc
```diff
@@ -259,6 +259,28 @@ includes a minimum of built-in methods.  You can use BasicObject to create an
 independent inheritance structure.  See the BasicObject documentation for
 further details.
 
+Just like modules, classes can also be reopened.
+You can omit its superclass when you reopen a class.
+Specifying a different superclass than the previous definition will raise an error.
```
nobu, line 264:
```suggestion
Just like modules, classes can also be reopened.  You can omit its superclass
when you reopen a class.  Specifying a different superclass than the previous
definition will raise an error.
```

nobu (changes requested):
Could you fold the paragraphs?

PR #8186, opened 2023-08-08
YJIT: Correct name of a counter

XrXr:


k0kubun (approved):

XrXr:
This naming mistake actually came from a refactor that's not meant to change behavior. The code before the refactor did check for send: https://github.com/ruby/ruby/pull/7725/files#diff-aa74543b1a984391e79ba4d4337fefa0dcc350e8255029e111572a72df3c8778L5497

But we've carried this for a while and it seems to work for simple cases I threw at it so hopefully it should be fine. The current check does effectively makes a bunch of code for optional parameters + splat unreachable, but I'll address it along with the bugfix/improvement for opts+rest+kws.

PR #8189, opened 2023-08-08
Skip running brew upgrade

k0kubun:
This has been unstable:
https://github.com/ruby/ruby/actions/runs/5797755676/job/15713988590

and I'm not sure if we need that in the first place, assuming the OS image itself is maintained by GitHub.


k0kubun:
The CI is passing. Gonna merge this for now, but we could always revert this if it turns out to be actually needed.

PR #8197, opened 2023-08-09
YJIT: Fallback megamorphic super/yield to dynamic dispatch

k0kubun:
A single `yield` could jump to multiple ISEQ blocks depending on who's calling it. A `super` could call multiple method entries if a module is inherited by multiple classes.

This PR changes `invokeblock` and `invokesuper` to fallback to dynamic dispatch when they end up being megamorphic. `guard_invokeblock_iseq_block_changed` and `guard_invokesuper_me_changed` are one of the top exit reasons on SFR, and this tries to stay in JIT code in such cases.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -7226,6 +7226,12 @@ fn gen_invokeblock_specialized(
         return Some(EndBlock);
     }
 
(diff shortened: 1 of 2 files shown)


noahgibbs:
LGTM

PR #8201, opened 2023-08-10
YJIT: implement codegen for rb_int_lshift

maximecb:
This is an implementation of bitwise left shift for integers, which will hopefully improve YJIT's performance on optcarrot a little bit more 🤞  

This implementation speculates that the shift amount will not change, which is always true on optcarrot (the side-exit is never taken).

The generated code is more compact than with a variable shift amount:
```
  # regenerate_branch
  # Block: foo@test.rb:2 (chain_depth: 1)
  # reg_temps: 00000011
  # Insn: 0004 opt_ltlt (stack_size: 2)
  # call to Integer#<<
  0x10daa0063: cmp rdi, 5
  0x10daa0067: jne 0x10daa2087
  0x10daa006d: mov rax, rsi
  0x10daa0070: sub rax, 1
  0x10daa0074: mov rcx, rax
  0x10daa0077: shl rcx, 2
  0x10daa007b: mov rdx, rcx
  0x10daa007e: sar rdx, 2
  0x10daa0082: cmp rdx, rax
  0x10daa0085: jne 0x10daa2087
  0x10daa008b: add rcx, 1
  0x10daa008f: mov rsi, rcx
```

### DIFF
--- a/yjit/src/asm/x86_64/mod.rs
+++ b/yjit/src/asm/x86_64/mod.rs
@@ -1277,7 +1277,7 @@ fn write_shift(cb: &mut CodeBlock, op_mem_one_pref: u8, op_mem_cl_pref: u8, op_m
         }
 
         _ => {
-            unreachable!("{:?} {:?}", opnd0, opnd1);
+            unreachable!("unsupported operands: {:?}, {:?}", opnd0, opnd1);
         }
     }
 }
--- a/yjit/src/asm/x86_64/mod.rs
+++ b/yjit/src/asm/x86_64/mod.rs
@@ -1277,7 +1277,7 @@ fn write_shift(cb: &mut CodeBlock, op_mem_one_pref: u8, op_mem_cl_pref: u8, op_m
         }
 
         _ => {
-            unreachable!();
+            unreachable!("{:?} {:?}", opnd0, opnd1);
         }
     }
 }
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4444,7 +4444,6 @@ fn jit_rb_int_div(
     true
 }
 
-/*
 fn jit_rb_int_lshift(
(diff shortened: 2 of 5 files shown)


yjit/src/codegen.rs
```diff
@@ -4460,28 +4459,42 @@ fn jit_rb_int_lshift(
...
     let lhs = asm.stack_pop(1);
 
-    // Ruby supports using a negative shift value
-    asm.comment("Guard shift negative");
-    let shift_val = asm.sub(rhs, 1.into());
-    asm.cmp(shift_val, 0.into());
-    asm.jl(Target::side_exit(Counter::lshift_range));
+    // Guard on the shift value we speculated on
+    asm.cmp(rhs, comptime_shift.into());
+    asm.jne(Target::side_exit(Counter::lshift_amt_changed));
```
k0kubun, line 4480: I wonder if it's useful to have a chain guard here, but I guess we'd rather like to save memory, assuming the shift amount is always an integer literal?

yjit/src/asm/x86_64/mod.rs
```diff
@@ -1277,7 +1277,7 @@ fn write_shift(cb: &mut CodeBlock, op_mem_one_pref: u8, op_mem_cl_pref: u8, op_m
         }
 
         _ => {
-            unreachable!();
+            unreachable!("{:?} {:?}", opnd0, opnd1);
```
k0kubun, line 1280:
```suggestion
            unreachable!("unsupported operands: {:?}, {:?}", opnd0, opnd1);
```

k0kubun (approved):

yjit/src/codegen.rs
maximecb: It seems in optcarrot it is always constant. It's kind of rare to see people using bitwise shifts with a variable amount.

PR #8203, opened 2023-08-11
YJIT: Implement checkmatch instruction

k0kubun:
`checkmatch` instruction is the top exit reason on SFR (35% of all exits). This PR calls the interpreter's C implementation to just stay in JIT code.


maximecb:
I really wish the tests were designed not to rely on HTTP connections 🤦‍♀️

PR #8204, opened 2023-08-11
YJIT: implement fast path for integer multiplication in opt_mult

maximecb:


yjit/src/codegen.rs
```diff
@@ -3364,8 +3364,42 @@ fn gen_opt_mult(
...
+    let two_fixnums = match asm.ctx.two_fixnums_on_stack(jit) {
+        Some(two_fixnums) => two_fixnums,
+        None => {
+            defer_compilation(jit, asm, ocb);
+            return Some(EndBlock);
+        }
+    };
+
+    if two_fixnums {
+        if !assume_bop_not_redefined(jit, asm, ocb, INTEGER_REDEFINED_OP_FLAG, BOP_PLUS) {
```
XrXr, line 3376:
```suggestion
        if !assume_bop_not_redefined(jit, asm, ocb, INTEGER_REDEFINED_OP_FLAG, BOP_MULT) {
```

maximecb: I wonder how that got there 😅 😆 

Thanks Alan!

maximecb:
Seems like it's not detecting the overflow correctly on arm 🤔 

```
#265 test_yjit.rb:4031: 
     def foo(a, b)
       a * b
     end
   
     r1 = foo(2, 3)
     r2 = foo(2, -3)
     r3 = foo(2 << 40, 2 << 41)
     r4 = foo(2 << 40, -2 << 41)
     r5 = foo(1 << 62, 1 << 62)
   
     [r1, r2, r3, r4, r5]
  #=> "[6, -6, 0, 0, 21267647932558653966460912964485513216]" (expected "[6, -6, 9671406556917033397649408, -9671406556917033397649408, 21267647932558653966460912964485513216]")  
test_yjit.rb  FAIL 1/265
```

PR #8207, opened 2023-08-11
[DOC] Don't suppress autolinks

BurdetteLamar:


array.c
```diff
@@ -8158,7 +8158,7 @@ rb_ary_deconstruct(VALUE ary)
  *      %i[foo bar baz] # => [:foo, :bar, :baz]
  *      %i[1 % *]       # => [:"1", :%, :*]
  *
- *  - \Method Kernel#Array:
```
peterzhu2118, line 8161: I think `Method` should be an exception. We don't usually deal with `Method` objects so we shouldn't link to it.

array.c
```diff
@@ -8167,7 +8167,7 @@ rb_ary_deconstruct(VALUE ary)
  *      Array(1)                      # => [1]
  *      Array({:a => "a", :b => "b"}) # => [[:a, "a"], [:b, "b"]]
  *
- *  - \Method Array.new:
```
peterzhu2118, line 8170: Ditto.

array.c
BurdetteLamar: Fixed.  Also, noted in doc guide.

doc/contributing/documentation_guide.md
```diff
@@ -164,6 +164,10 @@ We might consider whether to suppress when:
   (e.g., _Array_ in the documentation for class `Array`).
 - The same reference is repeated many times
   (e.g., _RDoc_ on this page).
+- The reference is to a class or module that users
+  usually don't deal with, including these:
+
+  - \Method.
```
peterzhu2118, line 170: I think we can also include `Class` and `Module` in this list.

BurdetteLamar: Added.

peterzhu2118 (approved):

PR #8208, opened 2023-08-11
[DOC] Don't suppress autolinks

BurdetteLamar:


string.c
```diff
@@ -11295,12 +11295,12 @@ rb_str_unicode_normalized_p(int argc, VALUE *argv, VALUE str)
  *
  * == What's Here
  *
- * First, what's elsewhere. \Class \Symbol:
```
peterzhu2118, line 11298: This is the documentation for `Symbol` right? I'm not a fan of the class linking to itself.

BurdetteLamar: Fixed (throughout class Symbol).

peterzhu2118 (approved):

PR #8221, opened 2023-08-15
[DOC] Improve doc guide compliance

BurdetteLamar:
Avoid suppressing autolinks to classes and modules.

I think this is the last for ruby/*.{c,rb}.  Everything else is upstream, right?

### DIFF
--- a/object.c
+++ b/object.c
@@ -4145,7 +4145,7 @@ f_sprintf(int c, const VALUE *v, VALUE _)
  *
  *  These are the methods defined for BasicObject:
  *
- *  - ::new: Returns a new BasicObject instance.
+ *  - ::new: Returns a new \BasicObject instance.
  *  - #!: Returns the boolean negation of +self+: +true+ or +false+.
  *  - #!=: Returns whether +self+ and the given object are _not_ equal.
  *  - #==: Returns whether +self+ and the given object are equivalent.
--- a/object.c
+++ b/object.c
@@ -4143,7 +4143,7 @@ f_sprintf(int c, const VALUE *v, VALUE _)
  *
  *  === What's Here
  *
- *  These are the methods defined for BasicObject:
+ *  These are the methods defined for \BasicObject:
  *
  *  - ::new: Returns a new \BasicObject instance.
  *  - #!: Returns the boolean negation of +self+: +true+ or +false+.
--- a/object.c
+++ b/object.c
@@ -4298,7 +4298,7 @@ InitVM_Object(void)
      *
      * == What's Here
      *
-     * \Module Kernel provides methods that are useful for:
+     * \Module \Kernel provides methods that are useful for:
      *
      * - {Converting}[rdoc-ref:Kernel@Converting]
      * - {Querying}[rdoc-ref:Kernel@Querying]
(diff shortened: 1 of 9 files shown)


object.c
```diff
@@ -4298,7 +4298,7 @@ InitVM_Object(void)
      *
      * == What's Here
      *
-     * \Module \Kernel provides methods that are useful for:
```
peterzhu2118, line 4301: This is documentation for the `Kernel` module, so we don't need to unescape it.

re.c
```diff
@@ -3498,7 +3498,7 @@ match_hash(VALUE match)
  *  call-seq:
  *    matchdata == object -> true or false
  *
- *  Returns +true+ if +object+ is another \MatchData object
```
peterzhu2118, line 3501: This is documentation for the `MatchData#==` right?

struct.c
```diff
@@ -1547,7 +1547,7 @@ rb_struct_dig(int argc, VALUE *argv, VALUE self)
 /*
  *  Document-class: Data
  *
- *  \Class \Data provides a convenient way to define simple classes
+ *  \Class Data provides a convenient way to define simple classes
```
peterzhu2118, line 1550: This is documentation for the `Data` class, so we should escape it.

struct.c
```diff
@@ -1619,7 +1619,7 @@ rb_struct_dig(int argc, VALUE *argv, VALUE self)
  * call-seq:
  *   define(*symbols) -> class
  *
- *  Defines a new \Data class.
+ *  Defines a new Data class.
```
peterzhu2118, line 1622: Ditto.

struct.c
```diff
@@ -1634,7 +1634,7 @@ rb_struct_dig(int argc, VALUE *argv, VALUE self)
  *     #=> #<data Measure amount=1, unit="km">
  *
  *
- *  Note that member-less \Data is acceptable and might be a useful technique
+ *  Note that member-less Data is acceptable and might be a useful technique
```
peterzhu2118, line 1637: Ditto.

object.c
BurdetteLamar: Fixed.

struct.c
BurdetteLamar: Fixed.

struct.c
BurdetteLamar: Fixed.

BurdetteLamar:
@peterzhu2118, why are we having more than one class/module defined in a file?  It makes this operation (and perhaps others) more difficult.

peterzhu2118:
Some classes need to share internal implementation in the C code, so they live in the same file. Unfortunately we can't resolve this issue without some (possibly significant) refactors in the C code.

object.c
```diff
@@ -4143,9 +4143,9 @@ f_sprintf(int c, const VALUE *v, VALUE _)
  *
  *  === What's Here
  *
- *  These are the methods defined for \BasicObject:
+ *  These are the methods defined for BasicObject:
  *
- *  - ::new: Returns a new \BasicObject instance.
+ *  - ::new: Returns a new BasicObject instance.
```
peterzhu2118, line 4148: This is documentation for the BasicObject class.

BurdetteLamar: Fixed.

peterzhu2118 (approved):

PR #8222, opened 2023-08-15
Fix a typo for `RJIT_TARGET_OK`

nobu:


k0kubun:
It has been deliberately left as is for allowing arm64 JIT, but I'll merge this first and fix it properly.

PR #8239, opened 2023-08-17
YJIT: Fix Kernel#respond_to? handling of rb_f_notimplement

XrXr:
We should return false for this type of special methods but wasn't
previously. Was reproducible with:

    make test-all TESTS=../test/-ext-/test_notimplement.rb RUN_OPTS='--yjit-call-threshold=1'

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4885,19 +4885,27 @@ fn jit_obj_respond_to(
     };
 
     let result = match (visibility, allow_priv) {
-        (METHOD_VISI_UNDEF, _) => Qfalse, // No method => false
-        (METHOD_VISI_PUBLIC, _) => Qtrue, // Public method => true regardless of include_all
-        (_, Some(true)) => Qtrue, // include_all => always true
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -4885,20 +4885,28 @@ fn jit_obj_respond_to(
...
+                return false;
+            }
+            Qfalse
+        }
+        (METHOD_VISI_PUBLIC, _) | // Public method => fine regardless of include_all
+        (_, Some(true)) => { // include_all => all visibility are acceptable
+            // Method exists and has acceptable visibility
+            if cme_def_type == VM_METHOD_TYPE_NOTIMPLEMENTED {
+                // C method with rb_f_notimplement(). `respond_to?` returns false
+                // without consulting `respond_to_missing?`.
```
k0kubun, line 4901:
```suggestion
                // without consulting `respond_to_missing?`. See also: rb_add_method_cfunc()
```

k0kubun (approved):
So complicated lol

PR #8240, opened 2023-08-17
YJIT: Fix String#<< return type

XrXr:
We previously falsely asserted that String#<< always returns a ::String
instance. Issue was discovered on CI with `--yjit-verify-ctx`.

https://github.com/ruby/ruby/actions/runs/5893760435/job/15986002531


k0kubun (approved):

noahgibbs:
Oof. Good catch!

PR #8241, opened 2023-08-17
[DOC] IO::console.getpass usage example

olivierlacan:
There were no clear example of this very useful method's usage anywhere in the `IO` or `IO::Console` docs, which was a shame.

See: https://docs.ruby-lang.org/en/master/IO.html#method-i-getpass

### DIFF
--- a/ext/io/console/console.c
+++ b/ext/io/console/console.c
@@ -1578,6 +1578,12 @@ str_chomp(VALUE str)
  * see String#chomp!.
  *
  * You must require 'io/console' to use this method.
+ * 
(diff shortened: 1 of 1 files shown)


jeremyevans:
This looks good and I strongly agree an example is needed.  However, could you please submit this upstream?:  https://github.com/ruby/io-console

olivierlacan:
Oh sorry @jeremyevans, will do. Didn’t realize.

olivierlacan:
Now opened as https://github.com/ruby/io-console/pull/46

PR #8242, opened 2023-08-18
[Feature #19839] Add Range#overlap?

shouichi:
Add a method that returns true if two range overlap, otherwise false.

```
(0..10).overlap?(5..15) #=> true
(0..10).overlap?(20..30) #=> false
```

### DIFF
--- a/range.c
+++ b/range.c
@@ -2149,6 +2149,34 @@ range_count(int argc, VALUE *argv, VALUE range)
     }
 }
 
+/*
+ *  call-seq:
+ *    overlap?(range) -> true or false
+ *
+ *  Returns +true+ if +range+ overlaps with +self+, +false+ otherwise:
+ *
+ *    (0..2).overlap?(1..3) #=> true
+ *    (0..2).overlap?(3..4) #=> false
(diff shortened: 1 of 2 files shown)


baweaver:
Oooo, I like this a lot, and have very frequently made helpers for this.

range.c
```diff
@@ -2149,6 +2149,30 @@ range_count(int argc, VALUE *argv, VALUE range)
...
+ *  Related: Range#cover?.
+ */
+
+static VALUE
+range_overlap(VALUE range, VALUE other)
+{
+    VALUE beg_self, beg_other;
+
+    beg_self = RANGE_BEG(range);
+    beg_other = RANGE_BEG(other);
```
nobu, line 2171: `(0..2).overlap?(nil)` segfaults.

nobu (changes requested):

range.c
shouichi: Thank you. Fxi to raise TypeError when a non-range argument is passed.

nobu (approved):

PR #8243, opened 2023-08-18
Add rb_no_keywords_accepted to C-API

jeremyevans:
This can be used by extensions for C functions that want to implement the equivalent of **nil syntax.

Extract this function from vm_args.c.

Update Enumerator::Product.new and Enumerator.product to use this function.

Loosen spec to pass with both the previous and new message.

Fixes [Bug #19829]


eregon (approved):

jeremyevans:
Matz didn't think it worth it to differentiate between the error messages.

PR #8246, opened 2023-08-18
Allow waitpid(-1) to be woken if a waitpid(pid) call is pending (3.1 backport)

KJTsanaktsidis:
If two threads are running, with one calling waitpid(-1), and another calling waitpid($some_pid), and then $some_other_pid exits, we would expect the waitpid(-1) call to retrieve that exit status; however, it cannot actually do so until $some_pid _also_ exits.

This patch fixes the issue by unconditionally checking for pending process group waits on SIGCHLD, and then allowing pending pid-only waits to "steal" the notification.

https://bugs.ruby-lang.org/issues/19837

[Fixes #19387]

### DIFF
--- a/process.c
+++ b/process.c
@@ -1120,7 +1120,7 @@ rb_sigwait_fd_migrate(rb_vm_t *vm)
 extern volatile unsigned int ruby_nocldwait; /* signal.c */
 /* called by timer thread or thread which acquired sigwait_fd */
 static void
-waitpid_each(struct list_head *head)
+waitpid_each(rb_vm_t *vm, struct list_head *head)
 {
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis:
Apologies @hsbt - I had messed up my merging and a previous version of this PR failed the test I added. I've fixed that now. The current CI failures all seem to be build infrastructure issues...

stanhu:
@unak  Do you know if there is a timeline for another Ruby 3.1 or 3.2 release? We hit this bug in production (https://github.com/puma/puma/issues/3313), and it caused some significant performance issues.

PR #8248, opened 2023-08-18
Allow waitpid(-1) to be woken if a waitpid(pid) call is pending (3.0 backport)

KJTsanaktsidis:
If two threads are running, with one calling waitpid(-1), and another calling waitpid($some_pid), and then $some_other_pid exits, we would expect the waitpid(-1) call to retrieve that exit status; however, it cannot actually do so until $some_pid _also_ exits.

This patch fixes the issue by unconditionally checking for pending process group waits on SIGCHLD, and then allowing pending pid-only waits to "steal" the notification.

https://bugs.ruby-lang.org/issues/19837

[Fixes #19387]

### DIFF
--- a/process.c
+++ b/process.c
@@ -1133,7 +1133,7 @@ rb_sigwait_fd_migrate(rb_vm_t *vm)
 extern volatile unsigned int ruby_nocldwait; /* signal.c */
 /* called by timer thread or thread which acquired sigwait_fd */
 static void
-waitpid_each(struct list_head *head)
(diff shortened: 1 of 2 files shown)


jeremyevans:
Note that Ruby 3.0 is in security maintenance mode, and this doesn't appear to be security related, so I don't think the branch maintainer will consider backporting to Ruby 3.0.

KJTsanaktsidis:
Closing this since it's not a security problem.

PR #8250, opened 2023-08-18
YJIT: Fix return type of Integer#/ with T_FIXNUM inputs

XrXr:
Issue found by running ruby/spec with `--yjit-verify-ctx`. Thanks!


noahgibbs:
This is definitely more correct, but I wonder if it will have knock-on effects with performance.

maximecb:
> This is definitely more correct, but I wonder if it will have knock-on effects with performance.

Unlikely because division is fairly rare in practice.

PR #8251, opened 2023-08-18
YJIT: Quiet mode when running with `--yjit-stats`

ywenc:
This PR adds a the `--yjit-stats=quiet` option while running with YJIT in stats mode. 

Currently, when running with `--yjit-stats`, stats are always printed on exit. With this change, `--yjit-stats=quiet` will give us the option to collect the stats-mode runtime stats without needing to print them. Passing in an invalid option to `--yjit-stats` will return an error message, which was the case for passing anything to `--yjit-stats` before the changes in this PR. Running with `--yjit-stats` will still print the stats on exit by default.

This option would be useful in cases where we would like to collect stats without printing them, e.g. if also running tests that compare what gets printed with test expectations.

### DIFF
--- a/yjit.c
+++ b/yjit.c
@@ -1169,6 +1169,7 @@ rb_yjit_set_exception_return(rb_control_frame_t *cfp, void *leave_exit, void *le
 
 // Primitives used by yjit.rb
 VALUE rb_yjit_stats_enabled_p(rb_execution_context_t *ec, VALUE self);
+VALUE rb_yjit_print_stats_p(rb_execution_context_t *ec, VALUE self);
 VALUE rb_yjit_trace_exit_locations_enabled_p(rb_execution_context_t *ec, VALUE self);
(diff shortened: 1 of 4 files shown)


yjit.c
```diff
@@ -1169,6 +1169,7 @@ rb_yjit_set_exception_return(rb_control_frame_t *cfp, void *leave_exit, void *le
 
 // Primitives used by yjit.rb
 VALUE rb_yjit_stats_enabled_p(rb_execution_context_t *ec, VALUE self);
+VALUE rb_yjit_verbose_stats_p(rb_execution_context_t *ec, VALUE self);
```
maximecb, line 1172: Very much a nitpick but could you rename it to `print_stats_p`. Verbose kind of implies that there is a non-verbose mode. I think this would be more descriptive. Everything else looks good 👌

PR #8023, opened 2023-07-04
Ruby 3.2 - Speed up rebuilding the loaded feature index and realpath cache

krk:
Backport "Speed up rebuilding the loaded feature index" (https://github.com/ruby/ruby/commit/1f8823f241c52e9c763888a47426c92baf284a48) and "Add a realpath cache to reduce number of syscalls" (https://github.com/ruby/ruby/pull/8017) to Ruby 3.2.

PR #8252, opened 2023-08-19
Re-apply "Ruby 3.2 - Speed up rebuilding the loaded feature index and realpath cache (#8023)"

MSP-Greg:
@nagachika

Thanks for your work with this.

Before the first revert, an encoding error caused a test-all failure.

For a number of years, Windows MSYS2 (mingw & ucrt) build's CI needed encoding changes in the shell.  Removed from test-all, and also changed test & test-all to run from a cmd shell instead of a bash shell.

After doing so, readline-ext had one encoding failure.  It is not installed in Ruby 3.3, so I removed the file `readline.so`, which then uses reline.

All tests pass.  Also, this should be the case, as now the mingw CI is identical to the Windows/mswin CI...

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -139,12 +139,11 @@ jobs:
         run: |
           make test
         if: ${{matrix.test_task == 'check' || matrix.test_task == 'test'}}
+        shell: cmd
 
       - name: test-all
         timeout-minutes: 45
         run: |
-          # Actions uses UTF8, causes test failures, similar to normal OS setup
-          chcp.com 437
           make ${{ StartsWith(matrix.test_task, 'test/') && matrix.test_task || 'test-all' }}
         env:
           RUBY_TESTOPTS: >-
@@ -152,6 +151,7 @@ jobs:
             ${{ matrix.test-all-opts }}
           BUNDLER_VERSION:
         if: ${{matrix.test_task == 'check' || matrix.test_task == 'test-all' || StartsWith(matrix.test_task, 'test/')}}
+        shell: cmd
 
       - name: test-spec
         timeout-minutes: 10
--- a/load.c
+++ b/load.c
@@ -164,6 +164,12 @@ get_loaded_features_realpaths(rb_vm_t *vm)
 }
 
 static VALUE
+get_loaded_features_realpath_map(rb_vm_t *vm)
+{
+    return vm->loaded_features_realpath_map;
+}
+
+static VALUE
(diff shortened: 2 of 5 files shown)


MSP-Greg:
@nobu

Thank you, much better than the double 'revert'.

Re the changes in the workflow file, the encoding code dates back to the [commit](https://github.com/ruby/ruby/commit/eb2b4258214337) running mingw on Actions, which was when 3.0 just started development (2019-12-31).  I still had similar code in ruby-loco, and removed all of it this morning. CI passes for all three builds (mingw, ucrt, & mswin).  I think it probably existed due to readline (now readline-ext), which now isn't used on Windows?

So, maybe add the changes to master?

nagachika:
@MSP-Greg It's OK to change the shell in GitHub Actions for MinGW. I would prefer to skip the failing readline test instead of removing readline.so on MinGW.
Please see my branch.
https://github.com/nagachika/ruby/commits/00-win-ruby
I have tried to create a PR for your branch, but I cannot find the base repository on the GitHub comparing page. 🤔

MSP-Greg:
@nagachika

Re the shell change, glad that's ok.  I keep an eye on a few extension gems, and also https://github.com/oneclick/rubyinstaller2. Most people are using Windows shells, not the Git or MSYS2 bash shells.   Also, most run CI from the default Actions shell, which is PowerShell.  There are exceptions...

Re the readline issue, I checked [readline-ext](https://github.com/ruby/readline-ext/blob/c8b49fe7f51e58/.github/workflows/test.yml), and it doesn't run CI on Windows.

Skipping the test is fine by me.

> I have tried to create a PR for your branch, but I cannot find the base repository on the GitHub comparing page.

Hmm.  Not sure.  When I've been in your position, I don't think I've ever done a PR, I've only added commits.  I think I've done so both from cmd line and the desktop app.  But, force pushes made a mess of things.

So, should I change the first commit title/subject to match @nobu's suggestion, squash the CI commits, and add your https://github.com/nagachika/ruby/commit/c7a3de8f6ef42dd9dc237915f4e20f5c6f2afaf3?

BTW, thanks.

MSP-Greg:
@nagachika

The changes I mentioned above are at https://github.com/MSP-Greg/ruby/commits/00-win-ruby-2

Running CI now.  Does that look ok?  Three commits, rebased also...

nagachika:
Thanks for rebasing. LGTM.

nagachika (approved):

PR #8253, opened 2023-08-19
[DOC] RDoc for Process

BurdetteLamar:
Revises doc for a few methods, including exec.  Adds class-doc sections "Execution Environment" and "Execution Options," which exec links to.

### DIFF
--- a/process.c
+++ b/process.c
@@ -3006,7 +3006,6 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
 /*
  *  call-seq:
  *    exec([env, ] command_line, options = {})
- *    exec([env, ] exe_path, options  = {})
  *    exec([env, ] exe_path, *args, options  = {})
  *
  *  Replaces the current process by doing one of the following:
--- a/process.c
+++ b/process.c
@@ -8650,8 +8650,7 @@ proc_warmup(VALUE _)
  * Each of these methods creates a process:
  *
  * - Process.exec: Replaces the current process by running a given external command.
- * - Process.spawn: Creates a child process.
- * - Kernel#spawn: Executes the given command and returns its pid without waiting for completion.
+ * - Process.spawn, Kernel#spawn: Executes the given command and returns its pid without waiting for completion.
  * - Kernel#system: Executes the given command in a subshell.
  *
  * Each of these methods accepts:
--- a/process.c
+++ b/process.c
@@ -1516,6 +1516,9 @@ rb_detach_process(rb_pid_t pid)
  *
  *  Avoids the potential for a child process to become a
  *  {zombie process}[https://en.wikipedia.org/wiki/Zombie_process].
+ *  Process::detach prevents this by setting up a separate Ruby thread
+ *  whose sole job is to reap the status of the process _pid_ when it terminates.
+ *
  *  This method is needed only when the parent process will never wait
  *  for the child process.
  *
@@ -1534,13 +1537,11 @@ rb_detach_process(rb_pid_t pid)
  *  This example also does not reap the second child process,
  *  but it does detach the process so that it does not become a zombie:
  *
- *    pid0 = Process.spawn('ruby', '-e', 'exit 13') # => 313213
- *    pid1 = Process.spawn('ruby', '-e', 'exit 14') # => 313262
- *    Process.wait(pid0)                            # => 313213
- *    sleep 2
- *    thread = Process.detach(pid1)
+ *    pid = Process.spawn('ruby', '-e', 'exit 13') # => 313213
+ *    sleep(1)
+ *    thread = Process.detach(pid)
  *    # => #<Process::Waiter:0x00007f038f48b838 run>
- *    system("ps -ho pid,state -p #{pid1}")         # Finds no zombies.
+ *    system("ps -ho pid,state -p #{pid}")        # Finds no zombies.
  *
  *  The waiting thread can return the pid of the detached child process:
  *
@@ -3014,7 +3015,7 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
  *  - Invoking the executable at +exe_path+.
  *
  *  The new process is created using the
- *  {exec system call}[https://www.man7.org/linux/man-pages/man1/exec.1p.html];
+ *  {exec system call}[https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/execve.html];
  *  it may inherit some of its environment from the calling program
  *  (possibly including open file descriptors).
  *
--- a/process.c
+++ b/process.c
@@ -1538,8 +1538,8 @@ rb_detach_process(rb_pid_t pid)
  *  but it does detach the process so that it does not become a zombie:
  *
  *    pid = Process.spawn('ruby', '-e', 'exit 13') # => 313213
- *    sleep(1)
  *    thread = Process.detach(pid)
+ *    sleep(1)
  *    # => #<Process::Waiter:0x00007f038f48b838 run>
  *    system("ps -ho pid,state -p #{pid}")        # Finds no zombies.
  *
--- a/process.c
+++ b/process.c
@@ -8790,6 +8790,7 @@ proc_warmup(VALUE _)
  * to create a new process group for the new process.
  *
  * ==== Resource Limits
+ *
  * Use execution options to set resource limits.
  *
  * The keys for these options are symbols of the form
--- a/process.c
+++ b/process.c
@@ -1419,21 +1419,18 @@ proc_m_wait(int c, VALUE *v, VALUE _)
     return proc_wait(c, v);
 }
 
-
 /*
  *  call-seq:
- *     Process.wait2(pid=-1, flags=0)      -> [pid, status]
- *     Process.waitpid2(pid=-1, flags=0)   -> [pid, status]
+ *    Process.wait2(pid = -1, flags = 0) -> [pid, status]
  *
- *  Waits for a child process to exit (see Process::waitpid for exact
- *  semantics) and returns an array containing the process ID and the
- *  exit status (a Process::Status object) of that
- *  child. Raises a SystemCallError if there are no child processes.
+ *  Like Process.waitpid, but returns an array
+ *  containing the child process pid and Process::Status status:
  *
- *     Process.fork { exit 99 }   #=> 27437
- *     pid, status = Process.wait2
- *     pid                        #=> 27437
- *     status.exitstatus          #=> 99
+ *    pid = Process.spawn('ruby', '-e', 'exit 13') # => 309581
+ *    Process.wait2(pid)
+ *    # => [309581, #<Process::Status: pid 309581 exit 13>]
+ *
+ *  Process.waitpid2 is an alias for Process.waitpid.
  */
 
 static VALUE
@@ -1447,22 +1444,17 @@ proc_wait2(int argc, VALUE *argv, VALUE _)
 
 /*
  *  call-seq:
- *     Process.waitall   -> [ [pid1,status1], ...]
+ *    Process.waitall -> array
  *
- *  Waits for all children, returning an array of
- *  _pid_/_status_ pairs (where _status_ is a
- *  Process::Status object).
+ *  Waits for all children, returns an array of 2-element arrays;
+ *  each subarray contains the integer pid and Process::Status status
+ *  for one of the reaped child processes:
  *
- *     fork { sleep 0.2; exit 2 }   #=> 27432
- *     fork { sleep 0.1; exit 1 }   #=> 27433
- *     fork {            exit 0 }   #=> 27434
- *     p Process.waitall
+ *    pid0 = Process.spawn('ruby', '-e', 'exit 13') # => 325470
+ *    pid1 = Process.spawn('ruby', '-e', 'exit 14') # => 325495
+ *    Process.waitall
+ *    # => [[325470, #<Process::Status: pid 325470 exit 13>], [325495, #<Process::Status: pid 325495 exit 14>]]
  *
- *  <em>produces</em>:
- *
- *     [[30982, #<Process::Status: pid 30982 exit 0>],
- *      [30979, #<Process::Status: pid 30979 exit 1>],
- *      [30976, #<Process::Status: pid 30976 exit 2>]]
  */
 
 static VALUE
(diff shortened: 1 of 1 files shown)


process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
+ *    Process.detach(pid) -> thread or nil
  *
- *  Some operating systems retain the status of terminated child
- *  processes until the parent collects that status (normally using
- *  some variant of <code>wait()</code>). If the parent never collects
- *  this status, the child stays around as a <em>zombie</em> process.
- *  Process::detach prevents this by setting up a separate Ruby thread
- *  whose sole job is to reap the status of the process _pid_ when it
- *  terminates. Use #detach only when you do not intend to explicitly
- *  wait for the child to terminate.
```
peterzhu2118, line 1532: I think this part of the documentation is important. It explains how `Process.detach` works. It looks like it's missing from the new documentation.

process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
- *  ID, the thread returns +nil+ immediately.
+ *  This example does not reap the second child process;
+ *  that process appears as a zombie in the process status (+ps+) output:
  *
- *  The waiting thread has #pid method which returns the pid.
+ *    pid0 = Process.spawn('ruby', '-e', 'exit 13') # => 312691
+ *    pid1 = Process.spawn('ruby', '-e', 'exit 14') # => 312716
+ *    Process.wait(pid0)                            # => 312691
+ *    # Find zombies.
+ *    system("ps -ho pid,state -p #{pid1}")
```
peterzhu2118, line 1529: What's the purpose of `Process.wait(pid0)` here? It appears to serve the same purpose as a pause for `pid1` to run. Can we just do something like this instead?

```suggestion
 *    pid = Process.spawn('ruby', '-e', 'exit 13') # => 312691
 *    sleep(1)
 *    # Find zombies.
 *    system("ps -ho pid,state -p #{pid}")
```

process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
+ *  but it does detach the process so that it does not become a zombie:
  *
- *     27389 Z
+ *    pid0 = Process.spawn('ruby', '-e', 'exit 13') # => 313213
+ *    pid1 = Process.spawn('ruby', '-e', 'exit 14') # => 313262
+ *    Process.wait(pid0)                            # => 313213
+ *    sleep 2
+ *    thread = Process.detach(pid1)
+ *    # => #<Process::Waiter:0x00007f038f48b838 run>
+ *    system("ps -ho pid,state -p #{pid1}")         # Finds no zombies.
```
peterzhu2118, line 1544: Ditto. We don't need `pid0` here right?

process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
- *  wait for the child to terminate.
+ *  Avoids the potential for a child process to become a
+ *  {zombie process}[https://en.wikipedia.org/wiki/Zombie_process].
+ *  This method is needed only when the parent process will never wait
+ *  for the child process.
  *
- *  The waiting thread returns the exit status of the detached process
- *  when it terminates, so you can use Thread#join to
- *  know the result.  If specified _pid_ is not a valid child process
- *  ID, the thread returns +nil+ immediately.
```
peterzhu2118, line 1537: This doesn't look correct. The function `rb_detach_process` can never return `nil`, it always returns a new thread. It looks like the case in testing as well:

```ruby
Process.detach(213131212) # => #<Process::Waiter:0x00007f190ad24338 run>
```

process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
- *  the child automatically.
+ *  The waiting thread can return the pid of the detached child process:
  *
- *     p1 = fork { sleep 0.1 }
- *     p2 = fork { sleep 0.2 }
- *     Process.detach(p1)
- *     Process.waitpid(p2)
- *     sleep 2
- *     system("ps -ho pid,state -p #{p1}")
+ *    thread.join(thread).pid                       # => 313262
```
peterzhu2118, line 1548: Did you mean to do this? `Thread#join` does not accept an argument (at least does not accept a `Thread` object as an argument).

```suggestion
 *    thread.join.pid                       # => 313262
```

process.c
```diff
@@ -3020,77 +3005,67 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
 
 /*
  *  call-seq:
- *     exec([env,] command... [,options])
- *
- *  Replaces the current process by running the given external _command_, which
- *  can take one of the following forms:
+ *    exec([env, ] command_line, options = {})
+ *    exec([env, ] exe_path, options  = {})
```
peterzhu2118, line 3009: Do we need this `call-seq`? Doesn't the one below cover this case since `*args` could contain no elements.

process.c
```diff
@@ -3020,77 +3005,67 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
+ *    exec([env, ] exe_path, *args, options  = {})
```
peterzhu2118, line 3010: I think it's more accurate to say that `options` is keyword arguments.

```suggestion
 *    exec([env, ] command_line, **options)
 *    exec([env, ] exe_path, **options)
 *    exec([env, ] exe_path, *args, **options)
```

process.c
```diff
@@ -3020,77 +3005,67 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
- *  (return value of spawn() and IO#pid for IO.popen) is the pid of the invoked
- *  command, not shell.
+ *  <b>Argument +command_line+</b>
  *
- *  In the second form (<code>exec("command1", "arg1", ...)</code>), the first
- *  is taken as a command name and the rest are passed as parameters to command
- *  with no shell expansion.
+ *  \String argument +command_line+ is a command line to be passed to a shell;
+ *  it must begin with a shell reserved word or special built-in,
+ *  and may also contain arguments and options for that command.
```
peterzhu2118, line 3039:
```suggestion
 *  it must begin with a shell reserved word, begin with a special built-in,
 *  or contain meta characters.
 *  It may also contain arguments and options for that command.
```

process.c
```diff
@@ -8669,6 +8644,172 @@ proc_warmup(VALUE _)
  * \Module +Process+ represents a process in the underlying operating system.
  * Its methods support management of the current process and its child processes.
  *
+ * == \Process Creation
+ *
+ * Each of these methods creates a process:
+ *
+ * - Process.exec: Replaces the current process by running a given external command.
+ * - Process.spawn: Creates a child process.
+ * - Kernel#spawn: Executes the given command and returns its pid without waiting for completion.
```
peterzhu2118, line 8653: I'm not sure why the documentation for `Process.spawn` is different from `Kernel#spawn`. They both call the same C function `rb_f_spawn`.

process.c
```diff
@@ -8669,6 +8644,172 @@ proc_warmup(VALUE _)
...
+ *
+ *   Process.spawn('ruby -e "puts Dir.pwd"', {chdir: '/tmp'})
+ *
+ * Output:
+ *
+ *   /tmp
+ *
+ * The working directory of the current process is not changed:
+ *
+ *   Dir .pwd # => "/var"
```
peterzhu2118, line 8710:
```suggestion
 *   Dir.pwd # => "/var"
```

process.c
```diff
@@ -8669,6 +8644,172 @@ proc_warmup(VALUE _)
...
+ * use execution option +:pgroup+ with one of the following values:
+ *
+ * - +true+: Create a new process group for the new process.
+ * - _pgid_: Create the new process in the process group
+ *   whose id is _pgid_.
+ *
+ * On Windows only, use execution option +:new_pgroup+ with value +true+
+ * to create a new process group for the new process.
+ *
+ * ==== Resource Limits
```
peterzhu2118, line 8792: Nit: newline after heading.

```suggestion
 * ==== Resource Limits
 *
```

process.c
```diff
@@ -1419,21 +1419,18 @@ proc_m_wait(int c, VALUE *v, VALUE _)
...
- *  child. Raises a SystemCallError if there are no child processes.
- *
- *     Process.fork { exit 99 }   #=> 27437
- *     pid, status = Process.wait2
- *     pid                        #=> 27437
- *     status.exitstatus          #=> 99
+ *    Process.wait2(pid = -1, flags = 0) -> [pid, status]
+ *
+ *  Like Process.waitpid, but returns an array
+ *  containing the child process pid and Process::Status status:
```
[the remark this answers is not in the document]
nobu: _pid_ and _status_ as variables, I guess.

process.c
```diff
@@ -3020,77 +3005,67 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
- *  shell expansion before being executed.
+ *  - Passing string +command_line+ to the shell.
+ *  - Invoking the executable at +exe_path+.
  *
- *  The standard shell always means <code>"/bin/sh"</code> on Unix-like systems,
- *  otherwise, <code>ENV["RUBYSHELL"]</code> or <code>ENV["COMSPEC"]</code> on
- *  Windows and similar.  The command is passed as an argument to the
- *  <code>"-c"</code> switch to the shell, except in the case of +COMSPEC+.
+ *  The new process is created using the
+ *  {exec system call}[https://www.man7.org/linux/man-pages/man1/exec.1p.html];
```
nobu, line 3018: That page is not the system call.
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/functions/execve.html

process.c
```diff
@@ -3020,77 +3005,67 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
+ *  On a Unix-like system, the shell is <tt>/bin/sh</tt>;
+ *  otherwise the shell is determined by environment variable
+ *  <tt>ENV['RUBYSHELL']</tt>, if defined, or <tt>ENV['COMSPEC']</tt> otherwise.
  *
- *  In order to execute the command, one of the <code>exec(2)</code> system
- *  calls are used, so the running command may inherit some of the environment
- *  of the original program (including open file descriptors).
+ *  Except for the +COMSPEC+ case,
+ *  the entire string +command_line+ is passed as an argument
+ *  to {shell option -c}[https://www.man7.org/linux/man-pages/man1/bash.1.html#OPTIONS].
```
nobu, line 3047: Ruby uses `/bin/sh` not `bash`.
https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/sh.html

process.c
```diff
@@ -8669,6 +8644,172 @@ proc_warmup(VALUE _)
...
+ * - Kernel#system: Executes the given command in a subshell.
+ *
+ * Each of these methods accepts:
+ *
+ * - An optional hash of environment variable names and values.
+ * - An optional hash of execution options.
+ *
+ * === Execution Environment
+ *
+ * Optional leading argument +env+ is a hash of name/value pairs,
```
nobu, line 8663:
```suggestion
 * Optional leading argument _env_ is a hash of name/value pairs,
```

process.c
nobu: As it can contain `IO` instances as keys, it isn’t strictly “keyword” arguments.

process.c
BurdetteLamar: Restored.

process.c
BurdetteLamar: Modified (correctly, I hope).

process.c
BurdetteLamar: I don't know what needs to be changed here.

process.c
BurdetteLamar: I think that passing an empty array is different from omitting the argument altogether?

process.c
BurdetteLamar: I took these from the two What's Here sections.  If not ok, what should we say here?

process.c

process.c
BurdetteLamar: Fixed.

process.c
BurdetteLamar: @nobu, there are thousands of existing argument mentions using plus-signs.  Maybe I should add something to the documentation guide about that; also, when to use italics for something like _n_?

nobu: I too often overlook inconsistencies in the notation.
I agree that the criteria should be clear.

process.c
```diff
@@ -1520,48 +1512,41 @@ rb_detach_process(rb_pid_t pid)
...
  *
- *     27389 Z
+ *  This example also does not reap the second child process,
+ *  but it does detach the process so that it does not become a zombie:
  *
- *  In the next example, Process::detach is used to reap
- *  the child automatically.
+ *    pid = Process.spawn('ruby', '-e', 'exit 13') # => 313213
+ *    sleep(1)
+ *    thread = Process.detach(pid)
```
peterzhu2118, line 1542: We need the `Process.detach` call before the `sleep` call to ensure that the thread has a chance to run right?

process.c
peterzhu2118: I think the call-seq needs to be changed. It always returns a `thread` and never `nil`.

process.c
peterzhu2118: I'm not sure I understand what you mean. If we're using `exe_path`, then empty `args` array means we pass no arguments to the executable.

process.c
peterzhu2118: I see. Yeah I think we can leave as-is.

process.c
peterzhu2118: Maybe we can group them together into a single bullet point? Or maybe be separate bullet points have their descriptions be the same? Maybe we can mention that they are the same method? I'm not sure what the best solution here is.

process.c
peterzhu2118: @BurdetteLamar I don't see this as fixed.

process.c
peterzhu2118: Is there a reason we use italics over monofont text for variable names? I took a look at the current documentation and it looks like most of it uses monofont instead of italics. I personally prefer monofont over italics because I think that monofont is for code while italics is used for emphasis.

nobu: Just originally RDoc used `_var_`.
Also I don't necessarily want it to be rendered in italics, and monotype (with other attributes) would be fine.
I just focus on the difference between variables and other arbitrary code notations.

process.c
BurdetteLamar: Grouped.

process.c
BurdetteLamar: Fixed.

process.c
BurdetteLamar: Stet, per our discussion at the documentation guide.

process.c
BurdetteLamar: Deleted.

process.c

process.c
BurdetteLamar: Fixed.

process.c
peterzhu2118: @BurdetteLamar I think this one still needs fixing.

BurdetteLamar: Fixed?

peterzhu2118 (approved):

PR #8254, opened 2023-08-20
Removed unused false-branch of always-true conditional in version.c

sampersand:
This removes dead code that was added [8 years ago](https://github.com/ruby/ruby/commit/c5634371dcd1e49501989f3c713f1fd1f76c572e).

### DIFF
--- a/version.c
+++ b/version.c
@@ -121,7 +121,7 @@ Init_version(void)
     /*
      * The version of the engine or interpreter this ruby uses.
      */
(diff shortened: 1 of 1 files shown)


nobu:
It is a marker for fake.rb.

sampersand:
Im confused, could you elaborate?

nobu:
https://github.com/ruby/ruby/blob/70ce3c8947c7f95897e56b397270c061dd50f280/template/fake.rb.in#L24 needs `MKSTR` to extract constants to be overridden in generated fake.rb

PR #8256, opened 2023-08-21
Fix compile error on older systems without clock_get*

Bo98:
clock_get* calls are already gated in ifdef conditionals here:

https://github.com/ruby/ruby/blob/3c53d31ffa72946f17e662b119fe55b9a3bf31c8/process.c#L8377-L8381

and here:

https://github.com/ruby/ruby/blob/3c53d31ffa72946f17e662b119fe55b9a3bf31c8/process.c#L8556-L8560

as are all other assignments of the variable `c`.

However declaration `clockid_t c;` was not wrapped in an ifdef. Besides being an unused variable, systems that are without `clock_get*` are unlikely to have `clockid_t`, resulting in a compile error. This pull request fixes that.

This regressed in Ruby 3.2.

### DIFF
--- a/process.c
+++ b/process.c
@@ -8216,7 +8216,9 @@ rb_clock_gettime(int argc, VALUE *argv, VALUE _)
 
     VALUE unit = (rb_check_arity(argc, 1, 2) == 2) ? argv[1] : Qnil;
     VALUE clk_id = argv[0];
+#ifdef HAVE_CLOCK_GETTIME
     clockid_t c;
+#endif
 
     if (SYMBOL_P(clk_id)) {
 #ifdef CLOCK_REALTIME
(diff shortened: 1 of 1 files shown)


nobu:
What is your platform?
I'm curious where `CLOCK_REALTIME` etc are defined while `clock_gettime` is unavailable.

Bo98:
Sorry, I wrote this patch months ago and it appears I've misread it. I've corrected the description - the issue is solely the `clockid_t` usage. I've also simplified the patch to reflect this. I've tested it again and the patch fixes the issue described.

I've found a bug report for it: https://bugs.ruby-lang.org/issues/19478. I happen to be using the same system, but it may affect other systems not updated for POSIX 2008 compliance.

nobu (approved):
Thank you for the explanation.
Makes sense.

PR #8257, opened 2023-08-21
Free all empty heap pages in Process.warmup

peterzhu2118:
This commit adds `free_empty_pages` which frees all empty heap pages and moves the number of pages freed to the allocatable pages counter. This is used in Process.warmup to improve performance because page invalidation from copy-on-write is slower than allocating a new page.


byroot (approved):
❤️

PR #8258, opened 2023-08-21
Trying to fix OpenBSD and windows

jemmaissroff:


k0kubun:
I think https://github.com/ruby/ruby/pull/8266 fixes the problem. At least it did when I tried that on openbsd.rubyci.org. OpenBSD will make the next build in two hours, so I'm not sure if it actually succeeds on chkbuild yet, but assuming it's done, let me close this for now.

PR #8259, opened 2023-08-22
Stop incrementing `jit_entry_calls` once threshold is hit

byroot:
Otherwise the ISeq page will constantly be written into preventing it from being shared.

cc @k0kubun as we mentioned that recently.

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -381,19 +381,16 @@ static VALUE vm_invoke_proc(rb_execution_context_t *ec, rb_proc_t *proc, VALUE s
 static inline rb_jit_func_t
 jit_compile(rb_execution_context_t *ec)
 {
-    // Increment the ISEQ's call counter
     const rb_iseq_t *iseq = ec->cfp->iseq;
     struct rb_iseq_constant_body *body = ISEQ_BODY(iseq);
     bool yjit_enabled = rb_yjit_compile_new_iseqs();
-    if (yjit_enabled || rb_rjit_call_p) {
-        body->jit_entry_calls++;
-    }
-    else {
+    if (!(yjit_enabled || rb_rjit_call_p)) {
         return NULL;
     }
 
-    // Trigger JIT compilation if not compiled
+    // Increment the ISEQ's call counter and trigger JIT compilation if not compiled
     if (body->jit_entry == NULL) {
+        body->jit_entry_calls++;
         if (yjit_enabled) {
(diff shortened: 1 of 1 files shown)


byroot:
Hum, RJIT tests are failing in a weird way, not sure why.

k0kubun:
You're returning `body->jit_entry` even when `rb_rjit_call_p` is false. `rb_rjit_call_p` means "call RJIT-compiled code", not "compile code with RJIT". When it's false, it must return `NULL` for invalidation reasons.

Also, with today's master, you'd want to change `jit_compile_exception` and `body->jit_exception_calls` as well.

byroot:
> When it's false, it must return `NULL` for invalidation reasons.

Ah thank you, I totally missed that.

k0kubun (approved):
:+1:

vm.c
```diff
@@ -439,16 +436,16 @@ jit_compile_exception(rb_execution_context_t *ec)
     // Increment the ISEQ's call counter
```
k0kubun, line 436:
```suggestion
```

vm.c
```diff
@@ -381,19 +381,16 @@ static VALUE vm_invoke_proc(rb_execution_context_t *ec, rb_proc_t *proc, VALUE s
...
-    if (yjit_enabled || rb_rjit_call_p) {
-        body->jit_entry_calls++;
-    }
-    else {
+    if (!(yjit_enabled || rb_rjit_call_p)) {
         return NULL;
     }
 
-    // Trigger JIT compilation if not compiled
+    // Increment the ISEQ's call counter or Trigger JIT compilation if not compiled
```
k0kubun, line 391:
```suggestion
    // Increment the ISEQ's call counter and trigger JIT compilation if not compiled
```

vm.c
```diff
@@ -439,16 +436,16 @@ jit_compile_exception(rb_execution_context_t *ec)
...
     struct rb_iseq_constant_body *body = ISEQ_BODY(iseq);
-    if (rb_yjit_compile_new_iseqs()) {
-        body->jit_exception_calls++;
-    }
-    else {
+    if (!rb_yjit_compile_new_iseqs()) {
         return NULL;
     }
 
     // Trigger JIT compilation if not compiled
```
k0kubun, line 443:
```suggestion
    // Increment the ISEQ's call counter and trigger JIT compilation if not compiled
```

PR #8262, opened 2023-08-22
YJIT: add code_region_overhead stat output

maximecb:
Investigating for https://github.com/Shopify/ruby/issues/529, I thought it would be interesting to compute how many bytes in the code region are being unused, and to report what percentage of the total code region size this represents.

railsbench:
```
inline_code_size:          5,146,888
outlined_code_size:        3,766,480
code_region_size:          9,207,808
code_region_overhead:        294,440 ( 3.2%)
```

lobsters:
```
inline_code_size:         14,101,872
outlined_code_size:       10,151,248
code_region_size:         25,870,336
code_region_overhead:      1,617,216 ( 6.3%)
```

Some potential savings there, though it's not as bad as I thought. I'm actually surprised we're wasting this little space. I'm assuming here that the code region size is the total amount of space (pages) we've allocated for machine code.

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -273,6 +273,8 @@ module RubyVM::YJIT
       # Number of failed compiler invocations
       compilation_failure = stats[:compilation_failure]
 
+      code_region_overhead = stats[:code_region_size] - (stats[:inline_code_size] + stats[:outlined_code_size])
+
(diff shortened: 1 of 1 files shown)


k0kubun:
:+1:

PR #8269, opened 2023-08-23
YJIT: move `gc_obj_offsets` to `IseqPayload` to save memory

maximecb:
We can save a bit of memory by avoiding the administrative overhead and memory fragmentation of having a separate heap-allocated list for each block.

Lobsters before:
```
yjit_alloc_size:          29,162,617
```

Lobsters after:
```
yjit_alloc_size:          28,329,645
```

Memory savings of about 2.9% 🎉 

Now let's hope this passes the CI gauntlet 😆


k0kubun:
Not sure why CI is crashing, but I guess we want to clear `gc_obj_offsets` on Code GC (maybe through `rb_yjit_tracing_invalidate_all` for off-stack ISEQs)? We shouldn't read values from freed pages.

k0kubun:
Also, now that they are shared among blocks for the same ISEQ, maybe we could deduplicate some offsets when they point to the same values for extra savings.

maximecb:
> Not sure why CI is crashing, but I guess we want to clear gc_obj_offsets on Code GC (maybe through rb_yjit_tracing_invalidate_all for off-stack ISEQs)? We shouldn't read values from freed pages.

Ah yes that would make sense. Somehow it doesn't fail the tests on my local machine (!)

maximecb:
> Also, now that they are shared among blocks for the same ISEQ, maybe we could deduplicate some offsets when they point to the same values. IIRC, blocks for the same ISEQ are always freed together, so that deduplication should be safe.

The tricky thing is that if the values can be moved, the we need to update all occurrences. If the values are pinned deduplication may be possible. Otherwise would need to take this into account in the codegen and point to the same values, which may be doable.

maximecb:
Still segfaulting, but in other places. I'm wondering if this is worth pursuing. It seems like at some point we'll need to switch to a constant pool in order to have better space efficiency and better copy-on-write behavior 🤔

PR #8270, opened 2023-08-23
Handle Array#* raising ArgumentError in test

jeremyevans:
Treat ArgumentError as NoMemoryError, so it will resize the array and try again.

Fixes [Bug #12500]


nobu:
Or `rescue` the exception as same as `NoMemoryError`?

jeremyevans:
> Or `rescue` the exception as same as `NoMemoryError`?

We could do that, but if ArgumentError is raised by another of the method calls, it would hide that issue.

nobu (approved):

PR #8271, opened 2023-08-23
YJIT: Move block handler SpecVal variants into BlockHandler

XrXr:
A refactor so that the variants correspond to
branches in vm_caller_setup_arg_block().

As discussed in https://github.com/ruby/ruby/pull/8117#discussion_r1302274193

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5063,7 +5063,7 @@ unsafe extern "C" fn build_kwhash(ci: *const rb_callinfo, sp: *const VALUE) -> V
 // at sp[-2]. Depending on the frame type, it can serve different purposes,
 // which are covered here by enum variants.
 enum SpecVal {
-    BlockHandler(BlockHandler),
+    BlockHandler(Option<BlockHandler>),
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -5063,23 +5063,25 @@ unsafe extern "C" fn build_kwhash(ci: *const rb_callinfo, sp: *const VALUE) -> V
...
-    // To avoid holding values across C calls for complex calls,
-    // we might need to set the SpecVal earlier in the call sequence
-    AlreadySet,
 }
 
 // Each variant represents a branch in vm_caller_setup_arg_block.
 #[derive(Clone, Copy)]
 pub enum BlockHandler {
+    // passing no block
+    None,
```
k0kubun, line 5075: We are using `Option<BlockHandler>` to propagate a `block_handler`, but I think it should be changed to just `BlockHandler` if we add `None` variant to it. `None` and `Some(BlockHandler::None)` seem to mean the same thing.

k0kubun (approved):
:+1:

maximecb (approved):

PR #8273, opened 2023-08-24
Fix support for dynamic keys.

ioquatix:
Unfortunately there is a bug in `Fiber[key] = value` when key is dynamically defined. I didn't fully understand the implementation of `rb_check_id`.

https://bugs.ruby-lang.org/issues/19845

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2118,7 +2118,7 @@ rb_fiber_storage_get(VALUE self)
 static int
 fiber_storage_validate_each(VALUE key, VALUE value, VALUE _argument)
 {
-    rb_check_id(&key);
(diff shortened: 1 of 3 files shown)


nobu (approved):

eregon:
:+1: it should only allow Symbols, I think that's how we discussed it in the past, and so much simpler.

ioquatix:
Yeah, totally agree in retrospect. I monkey patched a fix in my `fiber-storage` shim: https://github.com/ioquatix/fiber-storage/commit/c98221b1eaf08762a52924b5a833c55ced525987

PR #8275, opened 2023-08-24
Revert "Re-apply "Ruby 3.2 - Speed up rebuilding the loaded feature index and realpath cache (#8023)""

nagachika:
Reverts ruby/ruby#8252

I suspect that #8252 could introduce a `[BUG] object allocation during garbage collection phase` error.

https://github.com/ruby/actions/actions/runs/5942679604/job/16116356743
https://github.com/ruby/actions/actions/runs/5955117204/job/16153242195
https://github.com/ruby/ruby/actions/runs/5932543614/job/16086524440

Let's consider reverting it and observe the results for a while.

### DIFF
--- a/.github/workflows/mingw.yml
+++ b/.github/workflows/mingw.yml
@@ -139,11 +139,12 @@ jobs:
         run: |
           make test
         if: ${{matrix.test_task == 'check' || matrix.test_task == 'test'}}
-        shell: cmd
 
       - name: test-all
         timeout-minutes: 45
         run: |
+          # Actions uses UTF8, causes test failures, similar to normal OS setup
+          chcp.com 437
           make ${{ StartsWith(matrix.test_task, 'test/') && matrix.test_task || 'test-all' }}
         env:
           RUBY_TESTOPTS: >-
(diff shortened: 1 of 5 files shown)


MSP-Greg:
@nagachika

Just pinging back on this.  Looking at https://github.com/ruby/actions/actions/workflows/snapshot-ruby_3_2.yml, there are still intermittent failures/errors.  Not sure about the cause.

But, the timing shown in https://bugs.ruby-lang.org/issues/19246#note-7 still applies.  Ruby 2.7, 3.0, and head are working as expected.

I haven't built a recent ruby_3_1 stable, but testing with `ruby 3.2.2 (2023-10-28 revision 8bbf909bb5) [x64-mswin64_140]` still shows the problematic behavior.

thomthom:
#8252 addressed a blocker for us migrating to Ruby 3.2. Would very interested in an update on the status of this issue.

nagachika:
Applied the patch again at https://github.com/ruby/ruby/commit/5baf94f9131fb45d50c8c408e007a138ced46606 and see no failure in CI at this time.

PR #8278, opened 2023-08-24
Optimize handle_interrupt(Exception => ..) as a common case

matthewd:
When interrupt behaviour is configured for all possible exceptions using `Exception`, there's no need to iterate the pending exception's ancestors for hash lookups.

More significantly, by storing the catch-all timing symbol directly in the mask stack, we can skip allocating the hash we would otherwise need.

---

The shortcut in `rb_threadptr_pending_interrupt_check_mask` is nice as a bonus, but not likely to ever be hot enough to show a difference in a benchmark. At best there _might_ be some situation where it helps speed up re-checks when an pending interrupt is held for a long time.

The more interesting performance difference is in "regular" code that must always protect against the possibility of an interrupt, rather than that rarely-hit code that gets invoked when an interrupt does actually occur.

<details><summary>benchmark script</summary>

```ruby
require "benchmark/ips"

def safely_do_nothing(defer_interrupts, allow_interrupts)
  Thread.handle_interrupt(defer_interrupts) do
    Thread.handle_interrupt(allow_interrupts) do
      nil
    end
  ensure
    begin
      Thread.handle_interrupt(allow_interrupts) do
        nil
      end
    ensure
      nil
    end
  end
end

# Reproduces the same level of block nesting: we don't want to measure
# the cost of method calls
def unsafely_do_nothing
  1.times do
    1.times do
      nil
    end
  ensure
    begin
      1.times do
        nil
      end
    ensure
      nil
    end
  end
end

Benchmark.ips do |x|
  x.report("control") do
    unsafely_do_nothing
  end

  exception_immutable_never = { Exception => :never }.compare_by_identity.freeze
  exception_immutable_immediate = { Exception => :immediate }.compare_by_identity.freeze

  x.report("frozen Exception") do
    safely_do_nothing(exception_immutable_never, exception_immutable_immediate)
  end

  exception_mutable_never = { Exception => :never }
  exception_mutable_immediate = { Exception => :immediate }

  x.report("mutable Exception") do
    safely_do_nothing(exception_mutable_never, exception_mutable_immediate)
  end

  runtimeerror_immutable_never = { RuntimeError => :never }.compare_by_identity.freeze
  runtimeerror_immutable_immediate = { RuntimeError => :immediate }.compare_by_identity.freeze

  x.report("frozen RuntimeError") do
    safely_do_nothing(runtimeerror_immutable_never, runtimeerror_immutable_immediate)
  end

  runtimeerror_mutable_never = { RuntimeError => :never }
  runtimeerror_mutable_immediate = { RuntimeError => :immediate }

  x.report("mutable RuntimeError") do
    safely_do_nothing(runtimeerror_mutable_never, runtimeerror_mutable_immediate)
  end
end
```
</details>

**Before (5b407450f5a28295cdeae8e8bb3ea4591600e395)**

```
             control      3.964M (± 6.9%) i/s -     19.741M in   5.009446s
    frozen Exception      1.078M (± 5.7%) i/s -      5.382M in   5.008824s
   mutable Exception      1.018M (± 9.8%) i/s -      5.043M in   5.013756s
 frozen RuntimeError      1.070M (± 5.9%) i/s -      5.386M in   5.050367s
mutable RuntimeError      1.038M (± 6.2%) i/s -      5.243M in   5.073596s
```

**After (5c16f28680bdf8d9e08602c641f00a6355362c53)**
 
```
             control      3.800M (±12.5%) i/s -     18.844M in   5.059755s
    frozen Exception      2.438M (± 5.1%) i/s -     12.229M in   5.031170s
   mutable Exception      2.225M (± 6.9%) i/s -     11.231M in   5.074758s
 frozen RuntimeError      2.379M (± 4.6%) i/s -     12.075M in   5.087612s
mutable RuntimeError      1.017M (± 5.9%) i/s -      5.078M in   5.010767s
```

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -1892,6 +1892,23 @@ enum handle_interrupt_timing {
 };
 
 static enum handle_interrupt_timing
+rb_threadptr_pending_interrupt_from_symbol(rb_thread_t *th, VALUE sym)
+{
+    if (sym == sym_immediate) {
+        return INTERRUPT_IMMEDIATE;
+    }
+    else if (sym == sym_on_blocking) {
+        return INTERRUPT_ON_BLOCKING;
+    }
+    else if (sym == sym_never) {
+        return INTERRUPT_NEVER;
+    }
+    else {
+        rb_raise(rb_eThreadError, "unknown mask signature");
+    }
+}
+
+static enum handle_interrupt_timing
 rb_threadptr_pending_interrupt_check_mask(rb_thread_t *th, VALUE err)
 {
     VALUE mask;
@@ -1902,6 +1919,16 @@ rb_threadptr_pending_interrupt_check_mask(rb_thread_t *th, VALUE err)
 
     for (i=0; i<mask_stack_len; i++) {
         mask = mask_stack[mask_stack_len-(i+1)];
+
+        if (SYMBOL_P(mask)) {
+            /* do not match RUBY_FATAL_THREAD_KILLED etc */
+            if (err != rb_cInteger) {
+                return rb_threadptr_pending_interrupt_from_symbol(th, mask);
+            }
+            else {
+                continue;
+            }
+        }
 
         for (mod = err; mod; mod = RCLASS_SUPER(mod)) {
             VALUE klass = mod;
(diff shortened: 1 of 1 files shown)


ioquatix:
This looks pretty nice to me.

thread.c
```diff
@@ -1903,6 +1920,16 @@ rb_threadptr_pending_interrupt_check_mask(rb_thread_t *th, VALUE err)
     for (i=0; i<mask_stack_len; i++) {
         mask = mask_stack[mask_stack_len-(i+1)];
 
+        if (SYMBOL_P(mask)) {
+            /* do not match RUBY_FATAL_THREAD_KILLED etc */
+            if (err != rb_cInteger) {
```
matthewd, line 1925: It seems surprising (and undocumented) that an "exception" key of `Integer` [and by extension, `Object`/`Kernel`/similar] to `handle_interrupt` affects behaviour -- https://github.com/ruby/ruby/actions/runs/5973661284/job/16206337391?pr=8278#step:9:128

This conditional maintains the existing behaviour, though I suspect the logic could be simplified in future to specifically treat the special exceptions as matching Exception, rather than Integer.

PR #8280, opened 2023-08-24
Fix use of the rb_profile_frames start parameter

jeremyevans:
Previously, it was decrementing the start argument until it reached zero without actually changing the control frame pointer.

This is the same as #2713, just rebased against master.

[Bug #14607]


ko1:
Sorry I missed it.
I couldn't say I understand completely but okay to merge (@jeremyevans knows more).

PR #8282, opened 2023-08-24
[DOC] RDoc for Process

BurdetteLamar:


process.c
```diff
@@ -4402,43 +4402,47 @@ rb_f_exit(int argc, const VALUE *argv)
...
- *
- *     begin
- *       exit
- *       puts "never get here"
- *     rescue SystemExit
- *       puts "rescued a SystemExit exception"
- *     end
- *     puts "after begin block"
+ *    exit(status = true)
+ *    Process::exit(status = true)
```
peterzhu2118, line 4406: I think we're standardizing on using a period rather than two colons for class methods right?

```suggestion
 *    Process.exit(status = true)
```

process.c
```diff
@@ -4402,43 +4402,47 @@ rb_f_exit(int argc, const VALUE *argv)
...
  *
- *  <em>produces:</em>
+ *  Initiates termination of the Ruby script by raising SystemExit;
+ *  the exception may be caught.
+ *  Returns exit status +status+ to the underlying operating system.
+ *
+ *  Values +true+ and +false+ for argument +status+
+ *  indicate, respectively, success and failure;
+ *  The meanings of integer values are system-dependent.
```
peterzhu2118, line 4414: We should add this line to the documentation for `exit!` too.

process.c
```diff
@@ -4345,13 +4345,13 @@ exit_status_code(VALUE status)
 NORETURN(static VALUE rb_f_exit_bang(int argc, VALUE *argv, VALUE obj));
 /*
  *  call-seq:
- *     Process.exit!(status=false)
+ *    Process.exit!(status = false)
```
peterzhu2118, line 4348: Just like `exit`, this method also exists as `Kernel#exit!` should we add a call-seq for `exit!(status = false)` here to be consistent?

process.c

process.c
BurdetteLamar: Done.

process.c
BurdetteLamar: Done.

peterzhu2118 (approved):

PR #8288, opened 2023-08-25
Check that __builtin_mul_overflow can handle long long

jeremyevans:
Fixes [Bug #17646]

Patch from xtkoba (Tee KOBAYASHI)


nobu (approved):
I missed the issue, thanks!

PR #8290, opened 2023-08-25
Add documentation regarding how to build Ruby for debugging.

ioquatix:
Is this the best way @nobu? Do you have any other advice?

### DIFF
--- a/doc/contributing/building_ruby.md
+++ b/doc/contributing/building_ruby.md
@@ -157,6 +157,14 @@ with the Ruby script you'd like to run. You can use the following make targets:
 * `make lldb-ruby`: Runs `test.rb` using Ruby in lldb
 * `make gdb-ruby`: Runs `test.rb` using Ruby in gdb
 
+### Compiling for Debugging
+
+You should configure Ruby without optimization and other flags that may interfere with debugging:
+
+``` shell
+./configure --enable-debug-env cppflags="-fno-omit-frame-pointer" optflags=-O0 LDFLAGS="-fno-omit-frame-pointer"
+```
+
 ### Building with Address Sanitizer
 
 Using the address sanitizer is a great way to detect memory issues.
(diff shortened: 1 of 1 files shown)


doc/contributing/building_ruby.md
```diff
@@ -157,6 +157,14 @@ with the Ruby script you'd like to run. You can use the following make targets:
 * `make lldb-ruby`: Runs `test.rb` using Ruby in lldb
 * `make gdb-ruby`: Runs `test.rb` using Ruby in gdb
 
+### Compiling for Debugging
+
+You should configure Ruby without optimization and other flags that may interfere with debugging:
+
+``` shell
+./configure --enable-debug-env cppflags="-fno-omit-frame-pointer" optflags=-O0 LDFLAGS="-fno-omit-frame-pointer"
```
nobu, line 165: `-fno-omit-frame-pointer` isn't a flag for `cpp`.
And `$(optflags)` is included in `LDFLAGS` not only in `CFLAGS`.

```suggestion
./configure --enable-debug-env optflags="-O0 -fno-omit-frame-pointer"
```

`--enable-devel` (which implies `debug-env`) may be worth to mention too.

ioquatix: Ah, yes, you are totally correct.

ioquatix:
@nobu do you think `--enable-debug-env` should set `-O0` etc?

nobu:
Hmmm, now `optflags` is defaulted to `-O3`.
I think it's ok to replace the default.

nobu:
> now `optflags` is defaulted to `-O3`.
if the compiler is gcc-variant.

nobu:
Ah, `--enable-debug-env`?
It is to "enable RUBY_DEBUG environment variable", different purpose.

ioquatix:
Thanks @nobu - I believe documentation like this is really helpful. If you have any other ideas let's discuss it.

PR #8294, opened 2023-08-25
lib/bundled_gems.rb - fixup when rb & so are both loaded

MSP-Greg:
Example - big_decimal

See https://bugs.ruby-lang.org/issues/19852


nobu:
Does `-rbigdecimal.rb` work?

nobu:
#8295

MSP-Greg:
@nobu

Sorry, I rarely use file extensions with `require`.  To clarify, using `.rb` does not work currently or with this PR.

Closing in favor of #8295.

PR #8300, opened 2023-08-26
YJIT: Handle getblockparamproxy with ifunc

jhawthorn:
Meant to fix this when we last looked at getblockparamproxy 😅. `getblockparamproxy` for "ifunc" behaves identically to iseq, in just pushing rb_block_param_proxy.

In `insns.def` it ends up in the same switch statement branch: https://github.com/ruby/ruby/blob/master/insns.def#L157-L160

Then examining `VM_BH_IFUNC_P` it's basically the same as `VM_BH_ISEQ_BLOCK_P`, except it's `x & 0x03 == 0x03` instead of `x & 0x03 == 0x01`. So we can combine those and just look for `x & 1 == 1` (which really makes sense, both of the types of block handler which don't require an allocation are the kinds which can be optimized into the block_param_proxy object).

This has a nice side-benefit that supporting the extra type actually lets us reduce the size of the generated code by one instruction 😇.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -8149,9 +8149,10 @@ fn gen_getblockparamproxy(
     // Peek at the block handler so we can check whether it's nil
     let comptime_handler = jit.peek_at_block_handler(level);
 
-    // Filter for the 3 cases we currently handle
+    // Filter for the 4 cases we currently handle
     if !(comptime_handler.as_u64() == 0 ||              // no block given
             comptime_handler.as_u64() & 0x3 == 0x1 ||   // iseq block (no associated GC managed object)
+            comptime_handler.as_u64() & 0x3 == 0x3 ||   // ifunc block (no associated GC managed object)
             unsafe { rb_obj_is_proc(comptime_handler) }.test() // block is a Proc
         ) {
         // Missing the symbol case, where we basically need to call Symbol#to_proc at runtime
@@ -8179,7 +8180,7 @@ fn gen_getblockparamproxy(
 
     // Use block handler sample to guide specialization...
     // NOTE: we use jit_chain_guard() in this decision tree, and since
-    // there are only 3 cases, it should never reach the depth limit use
+    // there are only a few cases, it should never reach the depth limit use
     // the exit counter we pass to it.
     //
     // No block given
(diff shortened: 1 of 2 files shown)


jhawthorn:
On RailsBench:

Before:
```
Top-10 most frequent exit ops (100.0% of exits):
               expandarray:      9,945 (57.3%)
                    opt_eq:      4,966 (28.6%)
        getblockparamproxy:      2,090 (12.0%)
      opt_getconstant_path:        243 ( 1.4%)
                      once:         58 ( 0.3%)
                     leave:         21 ( 0.1%)
               objtostring:         19 ( 0.1%)
    opt_send_without_block:          6 ( 0.0%)
                      send:          1 ( 0.0%)
                       pop:          1 ( 0.0%)
```

After:

```
Top-12 most frequent exit ops (100.0% of exits):
               expandarray:      9,944 (64.8%)
                    opt_eq:      4,966 (32.4%)
      opt_getconstant_path:        243 ( 1.6%)
        getblockparamproxy:         77 ( 0.5%)
                      once:         58 ( 0.4%)
                     leave:         19 ( 0.1%)
               objtostring:         18 ( 0.1%)
    opt_send_without_block:         11 ( 0.1%)
          opt_regexpmatch2:          1 ( 0.0%)
                   opt_mod:          1 ( 0.0%)
                      send:          1 ( 0.0%)
                       pop:          1 ( 0.0%)

```

yjit/src/codegen.rs
```diff
@@ -8147,6 +8147,7 @@ fn gen_getblockparamproxy(
     // Filter for the 3 cases we currently handle
```
byroot, line 8147:
```suggestion
    // Filter for the 4 cases we currently handle
```
?

maximecb (approved):

XrXr (approved):
This is neat, thanks!
I just have a patch for correctness anxiety:
```diff
diff --git a/yjit/src/codegen.rs b/yjit/src/codegen.rs
index 7b60c7282e..2af83f8992 100644
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -8175,7 +8175,7 @@ fn gen_getblockparamproxy(
 
     // Use block handler sample to guide specialization...
     // NOTE: we use jit_chain_guard() in this decision tree, and since
-    // there are only 3 cases, it should never reach the depth limit use
+    // there are only a few cases, it should never reach the depth limit use
     // the exit counter we pass to it.
     //
     // No block given
@@ -8199,10 +8199,10 @@ fn gen_getblockparamproxy(
         //   VM_BH_ISEQ_BLOCK_P(): block_handler & 0x03 == 0x01
         //   VM_BH_IFUNC_P():      block_handler & 0x03 == 0x03
         // So to check for either of those cases we can use: val & 0x1 == 0x1
+        const _: () = assert!(RUBY_SYMBOL_FLAG & 1 == 0, "guard below rejects symbol block handlers");
+        // Procs are aligned heap pointers so testing the bit rejects them too.
 
-        // Bail unless VM_BH_ISEQ_BLOCK_P(bh). This also checks for null.
         asm.test(block_handler, 0x1.into());
-
         jit_chain_guard(
             JCC_JZ,
             jit,

```

maximecb:
@jhawthorn

PR #8302, opened 2023-08-26
[DOC] More on method exec

BurdetteLamar:


BurdetteLamar:
Marked as draft b/c examples needed.

process.c
```diff
@@ -3035,8 +3038,23 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
  *
  *  \String argument +command_line+ is a command line to be passed to a shell;
  *  it must begin with a shell reserved word, begin with a special built-in,
- *  or contain meta characters.
- *  It may also contain arguments and options for that command.
+ *  or contain meta characters:
+ *
+ *    exec('date')                         # Built-in.
+ *    exec('if true; then echo "Foo"; fi') # Shell reserved word.
+ *    [Needs example.]                     # Contains meta character.
```
peterzhu2118, line 3045: I think an example for this could be `exec('echo $0')` (since `$` is a meta character). It outputs `sh` UNIX systems.

process.c
```diff
@@ -3058,10 +3082,22 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
+ *
+ *    exec('/usr/bin/date')
+ *
+ *  Output:
+ *
+ *    Sat Aug 26 09:38:00 AM CDT 2023
+ *
+ *  Ruby invokes the executable directly, with no shell and no shell expansion:
+ *
+ *    [Needs example.]
```
peterzhu2118, line 3095: An example is that this raises a Ruby error:

```ruby
exec('doesnt_exist') # => Raises Errno::ENOENT (No such file or directory - doesnt_exist)
```

While adding a meta character invokes the shell:

```ruby
exec('doesnt_exist && echo ok') # => sh: doesnt_exist: command not found
```

process.c
```diff
@@ -3058,10 +3082,22 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
  *
  *  If one or more +args+ is given, each is an argument or option
- *  to be passed to the executable.
+ *  to be passed to the executable:
+ *
+ *    [Needs example.]
```
peterzhu2118, line 3100: Maybe something like:

```ruby
exec('echo', 'hello', 'world') # => hello world
```

process.c
nobu: `date` is not a built-in command on POSIX (but on cmd.exe).
`echo` is built-in, it may be confusing as an example of meta character.

process.c
nobu:
```ruby
exec('echo', 'C*') # => C*
```
in contrast to the above example:
```ruby
exec('echo C*') # CONTRIBUTING.md COPYING COPYING.ja
```

process.c
nobu: A newline character (`\n`) is a meta character on all shells AFAIK.

process.c
peterzhu2118: Ah yeah that's a good example.

process.c
```diff
@@ -3035,8 +3038,23 @@ NORETURN(static VALUE f_exec(int c, const VALUE *a, VALUE _));
...
- *  It may also contain arguments and options for that command.
+ *  or contain meta characters:
+ *
+ *    exec('echo')                         # Built-in.
+ *    exec('if true; then echo "Foo"; fi') # Shell reserved word.
+ *    exec('date > date.tmp')              # Contains meta character.
+ *
+ *  Output:
+ *
+ *    Sat Aug 26 09:26:05 AM CDT 2023
```
nobu, line 3049: This output is for the 3rd one only?
Since `date` is redirected, no output.

BurdetteLamar: Duh.  Removed.

peterzhu2118 (approved):

PR #8304, opened 2023-08-27
Skip weak references to old objects in minor GC

peterzhu2118:
If we are in a minor GC and the object to mark is old, then the old object should already be marked and cannot be reclaimed in this GC cycle so we don't need to add it to the weak refences list.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -6901,6 +6901,16 @@ rb_gc_mark_weak(VALUE *ptr)
 
     GC_ASSERT(objspace->rgengc.parent_object == 0 || FL_TEST(objspace->rgengc.parent_object, FL_WB_PROTECTED));
 
+    /* If we are in a minor GC and the other object is old, then obj should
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -6901,6 +6901,16 @@ rb_gc_mark_weak(VALUE *ptr)
 
     GC_ASSERT(objspace->rgengc.parent_object == 0 || FL_TEST(objspace->rgengc.parent_object, FL_WB_PROTECTED));
 
+    /* If we are in a minor GC and the other object is old, then obj should
+     * already be marked and cannot be reclaimed in this GC cycle so we don't
+     * need to add it to the weak refences list. */
+    if (!is_full_marking(objspace) && RVALUE_OLD_P(obj)) {
```
wks, line 6907: The expression `RVALUE_OLD_P(obj)` will load the flags of `obj`.  This may cause a cache miss, but it's not as bad as marking `obj` (when treating the current field as strong reference).  Anyway, it's better to measure the cost.

PR #8309, opened 2023-08-28
[DOC] RDoc for Kernel#system

BurdetteLamar:


process.c
```diff
@@ -4679,56 +4679,120 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *  or contain meta characters:
+ *
+ *    system('echo')                             # => true  # Built-in.
+ *    $?                                         # => #<Process::Status: pid 640610 exit 0>
+ *    system('if true; then echo "Foo"; fi')     # => true  # Shell reserved word.
+ *    system('date > /tmp/date.tmp')             # => true  # Contains meta character.
+ *    system('date > /nop/date.tmp')             # => false
+ *    $?                                         # => #<Process::Status: pid 640742 exit 2>
+ *    system('foo')                              # => nil   # Command failed.
+ *    system('date > date.tmp', exception: true) # Raises RuntimeError.
```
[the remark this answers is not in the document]
BurdetteLamar: Revised.

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *    system('date > /tmp/date.tmp')             # => true  # Contains meta character.
+ *    system('date > /nop/date.tmp')             # => false
+ *    system('date > date.tmp', exception: true) # Raises RuntimeError.
+ *
+ *  Assigns the command’s error status to global variable <tt>$?</tt>:
+ *
+ *    system('echo')                             # => true  # Built-in.
+ *    $?                                         # => #<Process::Status: pid 640610 exit 0>
+ *    system('date > /nop/date.tmp')             # => false
+ *    $?                                         # => #<Process::Status: pid 640742 exit 2>
```
peterzhu2118, line 4740: These examples are under `Argument +command_line+` which makes it seem like `$?` is only set for the command line case and not for the executable path case. IMO we should just move this up, where we discuss `$?`.

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *  it must begin with a shell reserved word, begin with a special built-in,
+ *  or contain meta characters:
+ *
+ *    system('echo')                             # => true  # Built-in.
+ *    system('if true; then echo "Foo"; fi')     # => true  # Shell reserved word.
+ *    system('date > /tmp/date.tmp')             # => true  # Contains meta character.
+ *    system('date > /nop/date.tmp')             # => false
+ *    system('date > date.tmp', exception: true) # Raises RuntimeError.
+ *
+ *  Assigns the command’s error status to global variable <tt>$?</tt>:
```
peterzhu2118, line 4735: This is a non ASCII character:

```suggestion
 *  Assigns the command's error status to global variable <tt>$?</tt>:
```

BurdetteLamar: Fixed in new commit.

process.c
BurdetteLamar: Mentioned also for executable path (ll. 4785 ff.).  The thing about the first mention, early in the general discussion (l. 4703), is that we haven't yet had any examples of any kind.  I think we need to mention at 4703, but give examples in the two later spots.

peterzhu2118:
@BurdetteLamar I don't see any new commits.

BurdetteLamar:
> @BurdetteLamar I don't see any new commits.

Something is fouled.  When I go to https://github.com/ruby/ruby/blob/67cf62867e1f167930d2a5c2a85a221c45e72a4c/process.c, I see "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."  How to fix?

peterzhu2118:
That's expected. The commit hasn't been merged into Ruby so it doesn't exist in ruby/ruby (only in your fork).

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
- *    command name, <code>argv[0]</code> and zero or more arguments (no shell)
+ *  Returns:
  *
- *  system returns +true+ if the command gives zero exit status,
- *  +false+ for non zero exit status.
- *  Returns +nil+ if command execution fails.
- *  An error status is available in <code>$?</code>.
+ *  - +true+ if the command exits with status zero.
+ *  - +false+ if the exit status is a non-zero integer.
+ *  - +nil+ if the command fails.
```
nobu, line 4698: It is not guaranteed that successful exit is status zero, although I haven't used such platforms.
Also, "fails" feels ambiguous a little, whether the command couldn't execute or exited unsuccessfully.

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *
+ *  \String argument +command_line+ is a command line to be passed to a shell;
+ *  it must begin with a shell reserved word, begin with a special built-in,
+ *  or contain meta characters:
+ *
+ *    system('echo')                             # => true  # Built-in.
+ *    system('if true; then echo "Foo"; fi')     # => true  # Shell reserved word.
+ *    system('date > /tmp/date.tmp')             # => true  # Contains meta character.
+ *    system('date > /nop/date.tmp')             # => false
+ *    system('date > date.tmp', exception: true) # Raises RuntimeError.
```
nobu, line 4733: This wouldn't raise usually.

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *
+ *  Argument +exe_path+ is one of the following:
+ *
+ *  - The string path to an executable to be called.
+ *  - A 2-element array containing the path to an executable
+ *    and the string to be used as the name of the executing process.
+ *
+ *  Example:
+ *
+ *    system('/usr/bin/date') # => true
```
nobu, line 4778: Note that the location of `date` is system dependent.

process.c
ioquatix: > Assigns the command’s error status to global variable

I believe it's thread local.

process.c
```diff
@@ -4679,56 +4679,120 @@ rb_spawn(int argc, const VALUE *argv)
...
- *  Returns +nil+ if command execution fails.
- *  An error status is available in <code>$?</code>.
+ *  - +true+ if the command exits with status zero.
+ *  - +false+ if the exit status is a non-zero integer.
+ *  - +nil+ if the command fails.
  *
- *  If the <code>exception: true</code> argument is passed, the method
- *  raises an exception instead of returning +false+ or +nil+.
+ *  Raises an exception (instead of returning +false+ or +nil+)
+ *  if keyword argument +exception+ is passed a truthy value.
```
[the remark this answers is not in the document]

process.c
BurdetteLamar: > > Assigns the command’s error status to global variable
> 
> I believe it's thread local.

Removed  'global variable' (5 places).

process.c
BurdetteLamar: 'fails' => 'could not execute'.

process.c

process.c
BurdetteLamar: Noted.

process.c
ioquatix: Thanks @BurdetteLamar!

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *  Argument +exe_path+ is one of the following:
+ *
+ *  - The string path to an executable to be called.
+ *  - A 2-element array containing the path to an executable
+ *    and the string to be used as the name of the executing process.
+ *
+ *  Example:
+ *
+ *    system('/usr/bin/date') # => true # Path to date on Unix-style system.
+ *    system('foo')           # => nil   # Command failed.
```
peterzhu2118, line 4779: Nit: alignment is off from the line above.

```suggestion
 *    system('foo')           # => nil  # Command failed.
```

peterzhu2118 (approved):
A minor nit, otherwise looks good.

BurdetteLamar:
@nobu, are we good to go, do you think?

process.c
```diff
@@ -4679,56 +4679,132 @@ rb_spawn(int argc, const VALUE *argv)
...
+ *
+ *    system('echo "Foo"') # => true
+ *
+ *  Output:
+ *
+ *    Foo
+ *
+ *  On a Unix-like system, the shell is <tt>/bin/sh</tt>;
+ *  otherwise the shell is determined by environment variable
+ *  <tt>ENV['RUBYSHELL']</tt>, if defined, or <tt>ENV['COMSPEC']</tt> otherwise.
```
nobu, line 4752: Nit: I'm not sure if "otherwise" is appropriate since `RUBYSHELL` is used only on Windows now (OS/2 is mentioned in ruby.1 but not supported).

nobu (approved):

ioquatix:
Nice work team!

PR #8311, opened 2023-08-28
[YARP] Compile basic types

jemmaissroff:
Now that we have the infrastructure to compile YARP nodes in CRuby, this is a first PR to start compiling different node types. We will have several follow on PRs for the more complicated node types.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -45,8 +45,6 @@
 #include "insns_info.inc"
 #include "yarp/yarp.h"
 
-VALUE rb_iseq_compile_yarp_node(rb_iseq_t * iseq, const yp_node_t * yarp_pointer);
-
 #undef RUBY_UNTYPED_DATA_WARNING
 #define RUBY_UNTYPED_DATA_WARNING 0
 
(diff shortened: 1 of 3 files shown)


kddnewton (approved):
Looks good. I ran through everything and only found a couple of small things that could be addressed later. I'll list them out below:

* Our scope node function works for program - maybe we should only accept scope nodes to translate_yarp and call it before it gets into there.
* Within the YARP codebase, we've been consistently downcasting with a `cast` variable, which seems nice and consistent. That way you don't have to name them differently every time. For example: `yp_statements_node_t *cast = (yp_statements_node_t *) node;`. It might be nice to define `YP_DOWNCAST(cast, type)` even, just to keep it consistent.
* Maybe we should raise an error/return if a missing node is found in the tree? Those trees should really never be compiled.

kddnewton:
> It makes it easier for me to read and reason about if the node name is slightly more descriptive so I don't have to think twice when reading or writing the code about the node type and which fields it has. I know yarp itself does this differently though. If you think it's important that style on this stays consistent in CRuby and YARP, I can make the change here, although probably will do it in a later PR to avoid a ton of merge conflicts.

Nah this is totally fine. No worries.

PR #8315, opened 2023-08-29
Validate the typed data before dereferencing the internal struct.

ioquatix:
See <https://bugs.ruby-lang.org/issues/19853> for more context.

### DIFF
--- a/process.c
+++ b/process.c
@@ -1222,9 +1222,7 @@ rb_waitpid(rb_pid_t pid, int *st, int flags)
     VALUE status = rb_process_status_wait(pid, flags);
     if (NIL_P(status)) return 0;
 
-    struct rb_process_status *data;
-    TypedData_Get_Struct(status, struct rb_process_status, &rb_process_status_type, data);
-
+    struct rb_process_status *data = rb_check_typeddata(status, &rb_process_status_type);
     pid = data->pid;
 
     if (st) *st = data->status;
@@ -4751,8 +4749,7 @@ rb_f_system(int argc, VALUE *argv, VALUE _)
     if (pid > 0) {
         VALUE status = rb_process_status_wait(pid, 0);
 
-        struct rb_process_status *data;
-        TypedData_Get_Struct(status, struct rb_process_status, &rb_process_status_type, data);
+        struct rb_process_status *data = rb_check_typeddata(status, &rb_process_status_type);
 
         // Set the last status:
         rb_obj_freeze(status);
--- a/process.c
+++ b/process.c
@@ -1222,8 +1222,8 @@ rb_waitpid(rb_pid_t pid, int *st, int flags)
     VALUE status = rb_process_status_wait(pid, flags);
     if (NIL_P(status)) return 0;
 
-    rb_check_typeddata(status, &rb_process_status_type);
-    struct rb_process_status *data = RTYPEDDATA_DATA(status);
+    struct rb_process_status *data;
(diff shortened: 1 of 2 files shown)


nobu:
Shouldn't it be guaranteed that the return value of `rb_process_status_wait` is `Process::Status` or `nil`?

ioquatix:
> Shouldn't it be guaranteed that the return value of rb_process_status_wait is Process::Status or nil?

No, the fiber scheduler can return anything, or raise an exception.

nobu:
> > Shouldn't it be guaranteed that the return value of rb_process_status_wait is Process::Status or nil?
> 
> No, the fiber scheduler can return anything, or raise an exception.

And `Process::Status.wait` is expected to return it as-is?

ioquatix:
> And Process::Status.wait is expected to return it as-is?

Yes, because it's the pure interface with no side effects (doesn't set `last_status`, etc).

process.c
```diff
@@ -1222,7 +1222,9 @@ rb_waitpid(rb_pid_t pid, int *st, int flags)
     VALUE status = rb_process_status_wait(pid, flags);
     if (NIL_P(status)) return 0;
 
+    rb_check_typeddata(status, &rb_process_status_type);
     struct rb_process_status *data = RTYPEDDATA_DATA(status);
+
```
nobu, line 1227:
```suggestion
    struct rb_process_status *data = rb_check_typeddata(status, &rb_process_status_type);
```

process.c
```diff
@@ -4748,6 +4750,8 @@ rb_f_system(int argc, VALUE *argv, VALUE _)
 
     if (pid > 0) {
         VALUE status = rb_process_status_wait(pid, 0);
+
+        rb_check_typeddata(status, &rb_process_status_type);
         struct rb_process_status *data = RTYPEDDATA_DATA(status);
```
nobu, line 4755: Ditto.

```suggestion
        struct rb_process_status *data = rb_check_typeddata(status, &rb_process_status_type);
```

And `NIL_P` check is not needed here?

process.c
ioquatix: Ah great, thanks.

process.c
ioquatix: That is a good question. My understanding is that `rb_f_system` always blocks, so `nil` cannot be returned. `nil` is only returned for non-blocking `Process#wait` IIRC.

nobu:
BTW, why are `rb_process_status_wait` and `rb_process_status_waitv` non-static?
They are not exported in the headers at least.

ioquatix:
> BTW, why are rb_process_status_wait and rb_process_status_waitv non-static?
> They are not exported in the headers at least

In `io-event` gem, it's used. I believe it should be in public header, if it's not, I'll fix that in a separate PR.

```
#ifdef HAVE_RB_PROCESS_STATUS_WAIT
#define IO_Event_Selector_process_status_wait(pid) rb_process_status_wait(pid)
#else
VALUE IO_Event_Selector_process_status_wait(rb_pid_t pid);
#endif
```

nobu:
> In `io-event` gem, it's used. I believe it should be in public header, if it's not, I'll fix that in a separate PR.

They are not exported from libruby.so.
And does it use also `rb_process_status_waitv` function, not only `rb_process_status_wait`?

ioquatix:
I don't know about `rb_process_status_waitv` but I'll investigate it.

PR #8327, opened 2023-08-29
[YARP] Compile ProgramNode as ScopeNode

jemmaissroff:
We need to also have a scope on top level variables in a program. This commit allows us to do that.

### DIFF
--- a/yarp/yarp_compiler.c
+++ b/yarp/yarp_compiler.c
@@ -186,7 +186,7 @@ again:
             DECL_ANCHOR(cond_seq);
             INIT_ANCHOR(cond_seq);
 
-            yp_compile_node(iseq, cond, cond_seq, src, Qfalse, compile_context);
+            yp_compile_node(iseq, cond, cond_seq, src, false, compile_context);
             ADD_SEQ(ret, cond_seq);
         }
         break;
@@ -301,7 +301,7 @@ yp_compile_while(rb_iseq_t *iseq, int lineno, yp_node_flags_t flags, enum yp_nod
 
     ADD_LABEL(ret, redo_label);
     if (statements) {
-        yp_compile_node(iseq, (yp_node_t *)statements, ret, src, Qtrue, compile_context);
+        yp_compile_node(iseq, (yp_node_t *)statements, ret, src, true, compile_context);
     }
 
     ADD_LABEL(ret, next_label);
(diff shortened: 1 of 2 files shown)


yarp/yarp_compiler.c
```diff
@@ -970,13 +970,14 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
       case YP_NODE_PROGRAM_NODE: {
           yp_program_node_t *program_node = (yp_program_node_t *) node;
 
+          yp_scope_node_t scope_node;
+          yp_scope_node_init((yp_node_t *)node, &scope_node);
           if (program_node->statements->body.size == 0) {
               ADD_INSN(ret, &dummy_line_node, putnil);
           } else {
-              yp_compile_node(iseq, (yp_node_t *) program_node->statements, ret, src, popped, compile_context);
+              yp_compile_node(iseq, (yp_node_t *) &scope_node, ret, src, popped, compile_context);
```
kddnewton, line 978: I believe this is a strict-aliasing issue in C. I think the solution is:

```suggestion
              yp_scope_node_t *node = &scope_node;
              yp_compile_node(iseq, (yp_node_t *) node, ret, src, popped, compile_context);
```

Truthfully I still don't truly understand strict aliasing, but this is what I've done in the past that @XrXr has pointed me to. Alan can you confirm or deny what I'm saying here? I'm out of my depth.

yarp/yarp_compiler.c
```diff
@@ -1185,11 +1186,12 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
           yp_statements_node_t *statements_node = (yp_statements_node_t *) node;
           yp_node_list_t node_list = statements_node->body;
           for (size_t index = 0; index < node_list.size; index++) {
-              // We only want to have popped == false for the last instruction
               if (!popped && (index != node_list.size - 1)) {
-                  popped = true;
+                  yp_compile_node(iseq, node_list.nodes[index], ret, src, Qtrue, compile_context);
```
kddnewton, line 1190: It seems like you're using `Qtrue` and `Qfalse` to represent `popped`, but that relies on the fact that `Qfalse` is 0. to convert into 0. Can we use a boolean instead?

kddnewton (changes requested):

kddnewton (approved):

yarp/yarp_compiler.c
XrXr: It might be fine here for strict aliasing because there is actually a `yp_node_t` inside `scope_node_t`. But you could get rid of any casting by passing `&scope_node->base`.

PR #8328, opened 2023-08-29
[YARP] Implement SourceNodes (File, Line, Encoding)

jemmaissroff:


yarp/yarp_compiler.c
```diff
@@ -1240,6 +1240,38 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
 
           return;
       }
+      case YP_NODE_SOURCE_ENCODING_NODE: {
+          const char *encoding = compile_context->parser->encoding.name;
+          if (!popped) {
+              // TODO: This encoding isn't an exact match of what CRuby outputs, for
+              // example "utf8" vs "#<Encoding:UTF-8>". Look into if there's a standard
+              // conversion to the expected format here.
+              ADD_INSN1(ret, &dummy_line_node, putobject, rb_str_new_cstr(encoding));
```
kddnewton, line 1249: I think you want `rb_find_encoding` here, but you need to convert the `encoding` into a string first.

kddnewton (changes requested):

yarp/yarp_compiler.c
```diff
@@ -1240,6 +1240,38 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
 
           return;
       }
+      case YP_NODE_SOURCE_ENCODING_NODE: {
+          const char *encoding = compile_context->parser->encoding.name;
+          if (!popped) {
+              rb_encoding *enc = rb_find_encoding(rb_str_new_cstr(encoding));
```
kddnewton, line 1246: This can return `NULL`. If it does, we should raise an `rb_bug`, because it shouldn't ever happen.

yarp/yarp_compiler.c
```diff
@@ -1240,6 +1240,38 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+              VALUE enc_name = rb_str_new_cstr(enc->name);
+              VALUE enc_str = rb_str_freeze(rb_sprintf("#<Encoding:%"PRIsVALUE">", enc_name));
```
kddnewton, line 1248: I don't think we want the name of the encoding, I think we want the actual encoding object here.

kddnewton (changes requested):

yarp/yarp_compiler.c
jemmaissroff: Hm, I'm not following what you mean. [The `rb_obj_as_string` call](https://github.com/ruby/ruby/blob/master/sprintf.c#L1128) doesn't work for an `rb_encoding`. Can you say more about what you mean here?

kddnewton: I mean the value going into the instruction is the encoding itself, not the name of the encoding. Check `RubyVM::InstructionSequence.compile("__ENCODING__").to_a[13][2]`. It shouldn't be the `rb_encoding`, it should be the Ruby `Encoding` object. You probably need `rb_enc_from_encoding` to convert it.

kddnewton (approved):

PR #8331, opened 2023-08-30
Implement "mutex reservation"

KJTsanaktsidis:
This implements a solution to Bug #19717, where it's possible for a resource protected by a condition variable & a mutex to ping-ping between two threads, whilst a third thread never gets it. This occurs if the resource is released & re-acquired in the same thread, because a thread which releases a mutex will almost always be able to re-acquire it afterwards unless it does some blocking work in between.

This implements a kind of "reservation" system for mutexes orchestrated with condition variables. A condition variable waiter (or, actually, any caller to Mutex#sleep) records the mutex it's waiting on in a thread variable; when it's woken up by the condition variable signaller, it gets the mutex "reserved" for it. If the waking thread subsequently attempts to reacquire the mutex immediately, it will thus fail because of the reservation and go to sleep.

[Bug #19717]

### DIFF
--- a/thread_sync.c
+++ b/thread_sync.c
@@ -8,6 +8,39 @@ static VALUE rb_eClosedQueueError;
 /* Mutex */
 typedef struct rb_mutex_struct {
     rb_fiber_t *fiber;
+    /**
+     *  The "reservation" system for mutexes is essentially designed to prevent the following
+     *  scenario:
+     *    - Thread A holds a mutex
+     *    - Threads B and C are waiting for that mutex, via a call to ConditionVariable#wait
+     *    - Thread A does some work
+     *    - Thread A then calls ConditionVariable#signal and releses the mutex
+     *    - In signaling the condvar, thread A marks thread B as runnable.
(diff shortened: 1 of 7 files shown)


thread_sync.c
```diff
@@ -38,25 +71,41 @@ struct queue_sleep_arg {
...
 {
     struct sync_waiter *cur = 0, *next;
+    long woken = 0;
 
     ccan_list_for_each_safe(head, cur, next, node) {
-        ccan_list_del_init(&cur->node);
-
         if (cur->th->status != THREAD_KILLED) {
             if (cur->th->scheduler != Qnil && cur->fiber) {
                 rb_fiber_scheduler_unblock(cur->th->scheduler, cur->self, rb_fiberptr_self(cur->fiber));
```
KJTsanaktsidis, line 89: I wasn't sure if I should implement this for the fiber-scheduler case as well - I was half-wondering if this kind of fix should actually be applied in the fiber scheduler implementaiton(s)? 

I think I can make it work in here though if that's preferable.

PR #8332, opened 2023-08-30
Remove dead loopval_popped code in compiler

jemmaissroff:


kddnewton (approved):
Looks good to me but maybe @ko1 should check it out

PR #6407, opened 2022-09-20
BasicSocket#recv* return `nil` rather than an empty packet

casperisfine:
[[Bug #19012]](https://bugs.ruby-lang.org/issues/19012)

man recvmsg(2) states:

> Return Value
> These calls return the number of bytes received, or -1 if an error occurred.
> The return value will be 0 when the peer has performed an orderly shutdown.

Not too sure how one is supposed to make the difference between a packet of size 0 and a closed connection.

PR #8338, opened 2023-08-31
TestSocket_UNIXSocket: stop testing empty packets

casperisfine:
Followup: https://github.com/ruby/ruby/pull/6407
[[Bug #19012]](https://bugs.ruby-lang.org/issues/19012)

OpenBSD and Solaris behave differently here.

Linux does deliver the empty packet, which is questionable as it's undistinguishable from a closed connection.

It seems that OpenBSD and Solaris simply drop it.

We could test the platform before doing the assertion, but it would likely be fragile, and the entire web recommend to not ever send an empty packet, so the value of this assertion is low.

### DIFF
--- a/test/socket/test_unix.rb
+++ b/test/socket/test_unix.rb
@@ -488,9 +488,7 @@ class TestSocket_UNIXSocket < Test::Unit::TestCase
       assert_kind_of(IO::WaitReadable, e)
     end
 
-    s2.send("", 0)
     s2.send("haha", 0)
-    assert_equal(nil, s1.recv(10)) # no way to distinguish empty packet from EOF with SOCK_SEQPACKET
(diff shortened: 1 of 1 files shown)


ioquatix:
> Linux does deliver the empty packet, which is questionable as it's undistinguishable from a closed connection.

What's the actual semantics? `read` returns 0?

jeremyevans (approved):

ioquatix:
> Yhe issue is with recvmsg

Thanks for the correction, sorry I was half asleep :)

> basically concluded with: "Don't send empty packets".

Makes sense. I wonder if we can get around that by using something like `wait_readable`... a closed connection will return different from a connection that's still open IIRC.

PR #8339, opened 2023-08-31
Enable passing the `target` argument to `Array#bsearch`, `Array#bsearch_index`, and `Range#bsearch`

kyanagi:
related issue: https://bugs.ruby-lang.org/issues/19075

This pull request implements the option (3) in the above issue.

* `bsearch` and `bsearch_index` take the argument `target`, whose possible values are `:first` or `:last`.
* find-by-number mode (former find-any mode) returns the first (or last) element for which the block returns zero.
* find-any mode has been renamed to find-by-number mode.
* find-minimum mode has been renamed to find-by-boolean mode.

```
a = [2, 3, 5, 7, 11]
a.bsearch {|x| x >= 6 } # => 7
a.bsearch(target: :first) {|x| x >= 6 } # same as a.bsearch { ... }
a.bsearch(target: :last) {|x| x <= 6 } # => 5
a.bsearch(target: :first) {|x| 0 } # => 2
a.bsearch(target: :last) {|x| 0 } # => 11
```

### DIFF
--- a/array.c
+++ b/array.c
@@ -3573,7 +3573,6 @@ rb_ary_bsearch_index(int argc, VALUE *argv, VALUE ary)
 
     static ID keyword_ids[1];
     VALUE target;
-    ID target_id;
 
     if (!keyword_ids[0]) {
         keyword_ids[0] = rb_intern_const("target");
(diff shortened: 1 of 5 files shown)


array.c
```diff
@@ -3488,60 +3551,52 @@ rb_ary_bsearch(VALUE ary)
...
+
+    if (target_id == rb_intern("first")) {
+        return ary_bsearch_index(ary, true);
+    }
+    else if (target_id == rb_intern("last")) {
+        return ary_bsearch_index(ary, false);
+    }
+
+  invalid_target:
+    rb_raise(rb_eArgError, "target must be :first or :last");
```
nobu, line 3598: Do not inadvertently convert an input to ID.
It can cause ID table overflow.

```suggestion
    if (target == ID2SYM(rb_intern("first"))) {
        return ary_bsearch_index(ary, true);
    }
    else if (target == ID2SYM(rb_intern("last"))) {
        return ary_bsearch_index(ary, false);
    }
    else {
        rb_raise(rb_eArgError, "target must be :first or :last");
        UNREACHABLE_RETURN(Qnil);
    }
```

range.c
```diff
@@ -647,44 +649,94 @@ bsearch_integer_range(VALUE beg, VALUE end, int excl)
...
+
+    if (target_id == rb_intern("first")) {
+        return bsearch_impl(argc, argv, range, true);
+    }
+    else if (target_id == rb_intern("last")) {
+        return bsearch_impl(argc, argv, range, false);
+    }
+
+  invalid_target:
+    rb_raise(rb_eArgError, "target must be :first or :last");
```
nobu, line 735: Ditto.

array.c
kyanagi: Thank you for your suggestion. I have fixed it.

range.c
kyanagi: This is also fixed.

PR #8342, opened 2023-08-31
[DOC] RDoc for #spawn

BurdetteLamar:


peterzhu2118 (approved):

mame:
@BurdetteLamar Has this erased the note added in 0bd6f36aa73059022fd5dd9cc2462cb4b785ba25? Could you restore the paragraph?

I appreciate your improving and formatting the document, but as a general rule, please be careful not to remove any content.

peterzhu2118:
@mame I think the contents of the note is still there, but the wording is different.

In the new documentation:

> The first required argument is one of the following:
>
> - `command_line` if it is a string, and if it begins with a shell reserved word or special built-in, or if it contains one or more meta characters.
> - `exe_path` otherwise.
>
> [...]
>
> # Argument `command_line`
>
> String argument `command_line` is a command line to be passed to a shell; it must begin with a shell reserved word, begin with a special built-in, or contain meta characters
>
> [...]
>
> # Argument `exe_path`
>
> [...]
>
> Ruby invokes the executable directly, with no shell and no shell expansion.

Is there anything else that should be mentioned?

mame:
@peterzhu2118 

> > # Argument `exe_path`
> > [...]
> > Ruby invokes the executable directly, with no shell and no shell expansion.

This is wrong. 0bd6f36aa73059022fd5dd9cc2462cb4b785ba25 says that even with `exe_path` style, shell expansion can take place depending on OS and arguments. See https://bugs.ruby-lang.org/issues/8550

peterzhu2118:
In the ticket, it says that on Windows, `echo` is a "special built-in". So in this documentation it will be a `command_line` which executes the shell. On Linux/macOS, `echo` is a normal program, so `echo` will be a `exe_path` and not execute the shell.

peterzhu2118:
Ah I guess the call-seq is `spawn([env, ] command_line, options = {})`, so maybe we can change it to `spawn([env, ] command_line, *args, options = {})` or add a special note?

mame:
The built-in command is just an example. Not only special built-in commands like `echo`, but also any call of `system(cmdname, arg1, arg2, arg3)` can cause shell expansion (or cmd.exe expansion) in Windows.

According to @unak and @nobu, this seems to be unavoidable since Windows does not have the paradigm of "an array of command-line arguments" unlike UN*X systems. It is too hard to describe the exact conditions, so users should not pass unreliable input to `Kernel#spawn`, etc. in any way. There seems to be no portable, generic sanitization method.

peterzhu2118:
So does that mean that all calls to `Process#spawn` runs in a shell on Windows?

nobu:
Not all, but it is not predictable without concrete knowledge about the command to be invoked.

nobu:
https://github.com/ruby/ruby/pull/9626

BurdetteLamar:
I am so Not Getting Into This.  As just a simple documentor, I leave it to you expert Rubyists to sort this out and make the appropriate PRs.

mame:
@BurdetteLamar No one blames you for introducing a bug, but you need to be responsible for a bug you introduced. You may think that the documentation change does not introduce a bug, but in fact it does, because the documentation is the basis for determining the correctness of the behavior. Even worse, the documentation cannot be mechanically tested. In a sense, it is more troublesome than the code.

I hope that anyone reviewing his PRs will be very careful in their review. If you see anything suspicious, it should be escalated to a committer in the know. If there is even a little doubt, it should not be merged.

junaruga:
If the PR included a bug, and Burdette is not motivated in fixing the bug, how about just reverting this commit?

BurdetteLamar:
I will open a new PR to amend this.

BurdetteLamar:
Opened https://github.com/ruby/ruby/pull/9666.

PR #8344, opened 2023-08-31
[YARP] Implement BackReferenceReadNode, NumberedReferenceReadNode

jemmaissroff:


yarp/yarp_compiler.c
```diff
@@ -472,6 +472,16 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
           }
           return;
       }
+      case YP_NODE_BACK_REFERENCE_READ_NODE: {
+          if (!popped) {
+              // Since a back reference is `$'`, ruby represents the ID as the following:
+              // It's not hardcoded in any variable we can use in yarp, just a rb_intern
+              // on the value after the `$`.
+              ID backref_val = INT2FIX(rb_intern("'")) << 1 | 1;
```
kddnewton, line 480: I think we want to read the value here. Back references can be `$&`, `$'`, `` $` ``, or `$+`. You should be able to use `location->start[1]`.

kddnewton (changes requested):

yarp/yarp_compiler.c
kddnewton: I'm not sure what we would expose exactly tbh. We could add flags for the character, but it would actually slow it down to look for each case here.

kddnewton (approved):

PR #8345, opened 2023-08-31
Fix YARP compiled send for popping

jemmaissroff:
Prior to this commit, we were incorrectly popping callers and receivers on CallNodes. We shouldn't pop any of these as they might have side effects.

### DIFF
--- a/yarp/yarp_compiler.c
+++ b/yarp/yarp_compiler.c
@@ -495,14 +495,14 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret, 
       case YP_NODE_CALL_NODE: {
           yp_call_node_t *call_node = (yp_call_node_t *) node;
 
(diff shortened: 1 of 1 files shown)


yarp/yarp_compiler.c
```diff
@@ -1320,7 +1323,7 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
           yp_statements_node_t *statements_node = (yp_statements_node_t *) node;
           yp_node_list_t node_list = statements_node->body;
           for (size_t index = 0; index < node_list.size - 1; index++) {
-              yp_compile_node(iseq, node_list.nodes[index], ret, src, !popped, compile_context);
+              yp_compile_node(iseq, node_list.nodes[index], ret, src, true, compile_context);
```
jemmaissroff, line 1326: This was wrong previously - we should pop all instructions on statements except the last one.

kddnewton (approved):

PR #8347, opened 2023-08-31
[Bug #19281] Allow semicolon in parenthesis at the first argument of command call

yui-knk:
Allow compstmt in the first argument of command call wrapped with parenthesis like following arguments with parenthesis.

https://bugs.ruby-lang.org/issues/19281

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -3348,14 +3348,7 @@ primary		: literal
                     /*% %*/
                     /*% ripper: begin!($3) %*/
                     }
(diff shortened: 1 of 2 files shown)


nobu (approved):

yui-knk:
Thanks for the review!

yahonda:
This commit introduces one of Rails CI failure https://github.com/rails/rails/issues/49128 , I have not found if this issue should be resolved at Rails side or not yet.

PR #8348, opened 2023-09-01
Use end of char boundary in start_with?

jhawthorn:
Previously we used the next character following the found prefix to determine if the match ended on a broken character.

This had caused surprising behaviour when a valid character was followed by a UTF-8 continuation byte.

This commit changes the behaviour to instead look for the end of the last character in the prefix.

[[Bug #19784]](https://bugs.ruby-lang.org/issues/19784)

### DIFF
--- a/internal/string.h
+++ b/internal/string.h
@@ -125,6 +125,15 @@ at_char_boundary(const char *s, const char *p, const char *e, rb_encoding *enc)
     return rb_enc_left_char_head(s, p, e, enc) == p;
 }
 
+static inline bool
+at_char_right_boundary(const char *s, const char *p, const char *e, rb_encoding *enc)
+{
+    RUBY_ASSERT(s <= p);
+    RUBY_ASSERT(p <= e);
+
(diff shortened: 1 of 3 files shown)


internal/string.h
```diff
@@ -125,6 +125,22 @@ at_char_boundary(const char *s, const char *p, const char *e, rb_encoding *enc)
...
+    RUBY_ASSERT(p <= e);
+
+    if (s == p) {
+        return true;
+    } else {
+        const char *prev_byte = p - 1;
+        const char *start_of_prev_char = rb_enc_left_char_head(s, prev_byte, e, enc);
+        const char *end_of_prev_char = start_of_prev_char + rb_enc_mbclen(start_of_prev_char, e, enc);
+        return end_of_prev_char == p;
+    }
```
nobu, line 141:
```suggestion
    return s == p || rb_enc_right_char_head(s, p, e, enc) == p;
```

jhawthorn: Thank you!

PR #8350, opened 2023-09-01
Keep write-barrier status after splicing array

peterzhu2118:
We don't need to remove the write-barrier protected status after splicing an array. We can simply add it to the rememberset for marking during the next GC.

The benchmark illustrates the performance impact on minor GC:

```ruby
require "benchmark"

arys = 1_000_000.times.map do
  ary = Array.new(50)
  ary.insert(1, 3)
  ary
end

4.times { GC.start }

puts(Benchmark.measure do
  1000.times do
    GC.start(full_mark: false)
  end
end)
```

This branch:

```
  1.309910   0.004342   1.314252 (  1.314580)
```

Master branch:

```
 54.376091   0.219037  54.595128 ( 54.742996)
```

### DIFF
--- a/array.c
+++ b/array.c
@@ -2176,9 +2176,14 @@ rb_ary_splice(VALUE ary, long beg, long len, const VALUE *rptr, long rlen)
             ARY_SET_LEN(ary, alen);
         }
         if (rlen > 0) {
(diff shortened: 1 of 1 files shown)


byroot (approved):
❤️

PR #8362, opened 2023-09-03
[Misc #19860] [DOC] Update license phrases to SPDX BSD-2-Clause

nobu:
There seem to be many variations on the "2-clause BSDL" these days. The current BSDL file seems to be the old style of the BSD 2 Clause "Simplified" license (BSD-2-Clause), so clarified this and updated it to a more recent version: https://spdx.org/licenses/BSD-2-Clause.html


BSDL
```diff
@@ -1,22 +1,24 @@
-Copyright (C) 1993-2013 Yukihiro Matsumoto. All rights reserved.
+Copyright (C) 1993-2023 Yukihiro Matsumoto. All rights reserved.
```
hsbt, line 1:
```suggestion
Copyright (C) 1993 Yukihiro Matsumoto. All rights reserved.
```

hsbt: End date is not necessary for copyright. We should leave to update it.

nurse:
License needs to be carefully reviewed

hsbt:
This is not only for GitHub metadata. We may need to update the recent version provided by SPDX.

hsbt:
It's not time to update this now.

PR #8366, opened 2023-09-04
[DOC] RDoc for process.c

BurdetteLamar:


peterzhu2118 (approved):

process.c
```diff
@@ -8639,38 +8639,108 @@ get_PROCESS_ID(ID _x, VALUE *_y)
...
+ *  If +signal+ is:
+ *
+ *  - A non-negative integer, or a signal name or symbol
+ *    without prefixed <tt>'-'</tt>,
+ *    each process with process ID +id+ is signalled.
+ *  - A negative integer, or a signal name or symbol
+ *    with prefixed <tt>'-'</tt>,
+ *    each process group with group ID +id+ is signalled.
+ *
+ *  The signal non-negative integers and strings are:
```
nobu, line 8685: Supported signals are platform dependent, I don't think list them up here.
Also the actual values too of course.

process.c
```diff
@@ -8639,38 +8639,108 @@ get_PROCESS_ID(ID _x, VALUE *_y)
...
+ *
+ *  - +0+ (<tt>'SIGEXIT'</tt>): Exit the current process.
```
nobu, line 8687: This is not a real signal, nor exit the process.
- For `Process.kill`, `0` examines if the destination process is alive or not, but `SIGEXIT` is nog recognized as a signal.
- For `Signal.trap`, Hook for when the current process exits.

process.c
BurdetteLamar: Will look into how best to make this change.  (Can't reopen a PR, right?)

process.c
BurdetteLamar: Ditto,

PR #8374, opened 2023-09-05
[YARP] Use constant_id lookups where possible

jemmaissroff:


yarp/yarp_compiler.c
```diff
@@ -359,6 +359,12 @@ yp_lookup_local_index_with_depth(rb_iseq_t *iseq, yp_compile_context_t *compile_
     return yp_lookup_local_index(iseq, compile_context, constant_id);
 }
 
+static ID
+yp_constant_id_lookup(yp_compile_context_t *compile_context, ID constant)
```
kddnewton, line 363: I don't think this will necessarily cause an issue because I think `ID` is an unsigned long long and `yp_constant_id_t` is a `uint32_t` so it will always be wider, but better to be safe than sorry.

```suggestion
yp_constant_id_lookup(yp_compile_context_t *compile_context, yp_constant_id_t constant)
```

Other than that, a comment on this function would be good just to explain what is being returned.

kddnewton (changes requested):

kddnewton (approved):

PR #8375, opened 2023-09-05
Introduce rb_gc_remove_weak

peterzhu2118:
If we're during incremental marking, then Ruby code can execute that deallocates certain memory buffers that have been called with rb_gc_mark_weak, which can cause use-after-free bugs.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -6924,6 +6924,27 @@ rb_gc_mark_weak(VALUE *ptr)
     objspace->profile.weak_references_count++;
 }
 
+void
+rb_gc_remove_weak(VALUE parent_obj, VALUE *ptr)
+{
+    rb_objspace_t *objspace = &rb_objspace;
+
+    /* If we're not incremental marking, then the state of the objects can't
+     * change so we don't need to do anything. */
+    if (!is_incremental_marking(objspace)) return;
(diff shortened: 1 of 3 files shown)


gc.c
```diff
@@ -6924,6 +6924,27 @@ rb_gc_mark_weak(VALUE *ptr)
...
+
+    /* If we're not incremental marking, then the state of the objects can't
+     * change so we don't need to do anything. */
+    if (!is_incremental_marking(objspace)) return;
+    /* If parent_obj has not been marked, then ptr has not yet been marked
+     * weak, so we don't need to do anything. */
+    if (!RVALUE_MARKED(parent_obj)) return;
+
+    VALUE **ptr_ptr;
+    rb_darray_foreach(objspace->weak_references, i, ptr_ptr) {
```
byroot, line 6940: I wonder how often we'd end up in that linear search. 

Right now weakref aren't used a lot so that list is probably small anyway.

peterzhu2118: This solution isn't optimal. However, it should be relatively rare that we run into this case since it requires that we're in a major GC, the parent object is marked, and we're calling `WeakMap#delete` or `WeakKeyMap#delete`.

byroot (approved):

PR #8379, opened 2023-09-06
YJIT: Make compiled_* stats available by default

k0kubun:
This PR introduces `DEFAULT_COUNTERS` that makes the counters available in `RubyVM::YJIT.runtime_stats` without `--yjit-stats`. I want to monitor `compiled_*` stats among different Ruby revisions, and I don't have enough clusters to accommodate `--yjit-stats` versions besides them. So I want to make them available without `--yjit-stats`.

It also leverages this constant to refactor `code_gc_count` as discussed in https://github.com/ruby/ruby/pull/6665#discussion_r1013224734.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -249,7 +249,7 @@ pub enum JCCKinds {
 
 #[inline(always)]
 fn gen_counter_incr(asm: &mut Assembler, counter: Counter) {
-    // Assert that default counters do not require --yjit-stats
+    // Assert that default counters are not incremented by generated code as this would impact performance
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -249,6 +249,9 @@ pub enum JCCKinds {
 
 #[inline(always)]
 fn gen_counter_incr(asm: &mut Assembler, counter: Counter) {
+    // Assert that default counters do not require --yjit-stats
```
maximecb, line 252:
```suggestion
    // Assert that default counters are not incremented by generated code as this would impact performance
```

maximecb (approved):
Good idea 👍

PR #8380, opened 2023-09-06
memory_view: Avoid using bit field

kou:
Bit field's memory layout is implementation-defined.

See also:
https://wiki.sei.cmu.edu/confluence/display/c/EXP11-C.+Do+not+make+assumptions+regarding+the+layout+of+structures+with+bit-fields

If memory layout is implementation-defined, it's difficult to use from FFI library such as Ruby-FFI.

### DIFF
--- a/include/ruby/memory_view.h
+++ b/include/ruby/memory_view.h
@@ -47,10 +47,10 @@ typedef struct {
     char format;
 
     /** :FIXME: what is a "native" size is unclear. */
-    unsigned native_size_p: 1;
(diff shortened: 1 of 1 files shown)


kou:
@mrkn What do you think about this?

mrkn:
@kou LGTM!!

kou:
Thanks!

I'll merge this in this week if nobody objects this.

kou:
I'll merge this.

PR #8389, opened 2023-09-06
Remove function call for String#bytesize

tenderlove:
String size is stored in a consistent location, so we can eliminate the function call.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4697,10 +4697,21 @@ fn jit_rb_str_bytesize(
     asm.comment("String#bytesize");
 
     let recv = asm.stack_pop(1);
-    let ret_opnd = asm.ccall(rb_str_bytesize as *const u8, vec![recv]);
+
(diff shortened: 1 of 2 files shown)


yjit/src/codegen.rs
```diff
@@ -4697,10 +4697,21 @@ fn jit_rb_str_bytesize(
...
+
+    asm.comment("get string length");
+    let str_len_opnd = Opnd::mem(
+        std::os::raw::c_long::BITS as u8,
+        asm.load(recv),
+        RUBY_OFFSET_RSTRING_LEN as i32,
+    );
+
+    let len = asm.load(str_len_opnd);
+    let shifted_val = asm.lshift(len, Opnd::UImm(1 as u64));
```
k0kubun, line 4709:
```suggestion
    let shifted_val = asm.lshift(len, Opnd::UImm(1));
```

k0kubun (approved):

maximecb (approved):
👌

PR #7452, opened 2023-03-06
YJIT: Bump SEND_MAX_DEPTH to 10

k0kubun:
On SFR, this reduced `send_klass_megamorphic / side_exit_count` from 20.5% to 15.0%. 

`send_klass_megamorphic` is not that common on railsbench, so it had no significant impact there.

PR #8390, opened 2023-09-07
YJIT: Decrease SEND_MAX_DEPTH to 5

k0kubun:
Since we have the send insn fallback (https://github.com/ruby/ruby/pull/8106), using large `SEND_MAX_DEPTH` no longer contributes to `ratio_in_yjit`.

We tested 4, 5, 8, 10, and 20 as a `SEND_MAX_DEPTH` on SFR, and 5 had the best average latency. It also decreases the max code size by 28%. We also know that something similar happened to Core when we tested 5 and 20 on it.

So this PR changes `SEND_MAX_DEPTH` back to 5, reverting https://github.com/ruby/ruby/pull/7452 and https://github.com/ruby/ruby/pull/7469.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2035,10 +2035,10 @@ fn jit_chain_guard(
     }
 }
 
-// up to 5 different classes, and embedded or not for each
+// up to 10 different classes, and embedded or not for each
 pub const GET_IVAR_MAX_DEPTH: i32 = 10;
 
-// up to 5 different classes, and embedded or not for each
+// up to 10 different classes, and embedded or not for each
 pub const SET_IVAR_MAX_DEPTH: i32 = 10;
 
 // hashes and arrays
@@ -2047,11 +2047,8 @@ pub const OPT_AREF_MAX_CHAIN_DEPTH: i32 = 2;
 // expandarray
 pub const EXPANDARRAY_MAX_CHAIN_DEPTH: i32 = 4;
 
-// up to 10 different classes
-pub const SEND_MAX_DEPTH: i32 = 20;
-
-// up to 20 different methods for send
-pub const SEND_MAX_CHAIN_DEPTH: i32 = 20;
+// up to 5 different methods for send
+pub const SEND_MAX_DEPTH: i32 = 5;
 
 // up to 20 different offsets for case-when
 pub const CASE_WHEN_MAX_DEPTH: i32 = 20;
@@ -4385,7 +4382,7 @@ fn jit_rb_kernel_instance_of(
         jit,
         asm,
         ocb,
-        SEND_MAX_CHAIN_DEPTH,
+        SEND_MAX_DEPTH,
         Counter::guard_send_instance_of_class_mismatch,
     );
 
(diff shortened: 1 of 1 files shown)


maximecb:
I think that a 28% reduction in max code size is really great. Could you provide some data for the impact on the performance of headline benchmarks (and also their rss) as well? I would just like to have an idea of the order of magnitude of the impact.

k0kubun:
Sure.

```
before: ruby 3.3.0dev (2023-09-06T22:26:11Z master 54274b8c65) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-09-06T23:54:53Z yjit-depth-5 ac4436f208) +YJIT [x86_64-linux]

--------------  -----------  ----------  ---------  ----------  ----------  ---------  -------------  ------------
bench           before (ms)  stddev (%)  RSS (MiB)  after (ms)  stddev (%)  RSS (MiB)  after 1st itr  before/after
activerecord    30.8         5.3         56.9       30.5        5.3         56.4       1.01           1.01
chunky-png      403.3        0.5         46.4       402.5       0.5         51.1       1.00           1.00
erubi-rails     10.4         32.4        99.0       10.4        26.0        97.0       1.01           1.01
hexapdf         1335.0       2.0         269.0      1335.7      1.1         256.6      1.00           1.00
liquid-c        38.8         0.7         39.7       38.5        0.6         39.5       1.00           1.01
liquid-compile  40.4         0.5         34.8       40.3        0.5         34.6       1.01           1.00
liquid-render   56.7         0.3         38.1       56.3        0.4         37.9       1.00           1.01
lobsters        602.2        4.0         333.7      607.3       4.2         331.0      1.08           0.99
mail            83.4         0.2         50.5       83.6        0.3         50.4       1.02           1.00
psych-load      1246.1       0.2         36.9       1242.6      0.3         36.4       1.00           1.00
railsbench      1130.3       1.3         105.8      1132.2      1.3         102.8      0.99           1.00
ruby-lsp        34.5         24.7        134.3      35.9        26.1        131.7      1.07           0.96
sequel          48.3         1.9         41.0       47.7        1.8         40.8       1.02           1.01
--------------  -----------  ----------  ---------  ----------  ----------  ---------  -------------  ------------
```

1st itr is improved for larger benchmarks like lobsters and ruby-lsp, which would translate to better p90/p99/average latency in production. ruby-lsp seems to get a lower peak performance while its warmup is improved.

RSS is decreased for most benchmarks. chunky-png was the only benchmark that uses more RSS with this change. Its `code_region_size` was actually improved by 8%. `code_region_overhead` happens to be 58 bytes (0.0%) in master while it becomes 25KB (4.7%) in this PR, so it might be related?

maximecb:
> RSS is decreased for most benchmarks. chunky-png was the only benchmark that uses more RSS with this change.

Benoit Daloze said that RSS is noisy and opened a PR to measure max_rss instead. We may want to switch to using this metric by defaul tin the future. Could also be the overhead as you mentioned, there can be effects that are hard to predict.

I think the results look good. Our benchmarks are probably generally not very polymorphic, so not affected much, which is nice. In the future (YJIT 4.4?) we may want to have a proper inline cache of some sort, which would allow us to have better performance and smaller code size for polymorphic ivar accesses and calls. Either way this is a nice improvement for 3.3.

Worth it to do a similar investigation for getivar max depth.

PR #8393, opened 2023-09-07
Extract `do_mutex_lock_check_interrupts` to try and fix `ppc64le`.

ioquatix:
We found some tests were hanging in `do_mutex_lock`, specifically the fiber scheduler autoload test. After much investigation, it may be a code generation bug. Because we didn't change the code, but only extracted it into a separate function, and it appears to fix the problem.

### DIFF
--- a/thread_sync.c
+++ b/thread_sync.c
@@ -290,6 +290,28 @@ delete_from_waitq(VALUE value)
     return Qnil;
 }
 
+static void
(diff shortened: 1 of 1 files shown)


junaruga:
This PR is not related to fixing the failures in Travis ppc64le?
I see this PR's Travis ppc64le failing.
https://app.travis-ci.com/github/ruby/ruby/jobs/609392097#L2335

ioquatix:
@junaruga it's due to other failures in the Ruby CI. With this PR:

image

PR #8395, opened 2023-09-07
[YARP] Implement compilation for InterpolatedXStringNode

jemmaissroff:


yarp/yarp_compiler.c
```diff
@@ -1335,6 +1335,45 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+          yp_interpolated_x_string_node_t *interp_x_string_node = (yp_interpolated_x_string_node_t *) node;
+          size_t parts_size = interp_x_string_node->parts.size;
+
+          ADD_INSN(ret, &dummy_line_node, putself);
+
+          if (parts_size > 0) {
+              for (size_t index = 0; index < parts_size; index++) {
+                  yp_node_t *part = interp_x_string_node->parts.nodes[index];
+
+                  switch (part->type) {
```
kddnewton, line 1348: Can you use `YP_NODE_TYPE` here? That will make it so that the switch will be exhaustive.

yarp/yarp_compiler.c
```diff
@@ -1335,6 +1335,45 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+                    case YP_NODE_STRING_NODE: {
+                        yp_x_string_node_t *x_string_node = (yp_x_string_node_t *) part;
```
kddnewton, line 1350: This is the wrong type here, since you're checking it's a `STRING_NODE` it should be `yp_string_node_t*`. From the other interpolated nodes, it looks like it matches those loops exactly, could you extract this out? Looks like it could be a function that takes a `yp_node_list_t*`.

kddnewton (changes requested):

kddnewton (approved):

PR #8397, opened 2023-09-07
Add `String#getbyte` YJIT implementation

ipc103:
Adds an implementation for `String#getbyte` for YJIT, along with a bootstrap test. This should be helpful for pure Ruby implementations and to avoid unneeded allocations.

See the following benchmark for comparison. The benchmark is loosely based on an example from TinyGraphql - see this [blog post](https://tenderlovemaking.com/2023/09/02/fast-tokenizers-with-stringscanner.html) and [code snippet](https://github.com/tenderlove/tinygql/blob/3fc680c73cf16bfb2c7b57ecb07e8721e67d384c/lib/tinygql/lexer.rb#L116-L126).

```ruby
require 'benchmark'

def get_key(str, pos)
    key = (str.getbyte(pos + 2) << 8) | str.getbyte(pos + 1)
end

def get_keys
    str = "one string" * 10_000
    0.upto(str.length - 10).each do |i|
        get_key(str , i)
    end
end

Benchmark.bm do |x|
  x.report('ruby-yjit') { 100.times { get_keys } }
end
```

```
user               system     total      real
ruby-control       1.213070   0.000037   1.213107 (  1.215320)
ruby-yjit-control  0.517514   0.007920   0.525434 (  0.569921)
ruby-yjit-getbyte  0.424347   0.003604   0.427951 (  0.428764)

ruby-control: ruby 3.3.0dev (2023-08-31T18:31:18Z master 945945dad4)
ruby-yjit-control: ruby 3.3.0dev (2023-08-31T18:31:18Z master 945945dad4) --yjit
ruby-yjit-getbyte: ruby 3.3.0dev (2023-09-06T23:49:59Z yjit_str_getbyte 76251be9e2) [x86_64-linux] --yjit
```

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1832,6 +1832,24 @@ assert_equal 'true', %q{
   jittable_method
 }
 
+# test getbyte on string class
+assert_equal '[97, :nil, 97, :nil, :raised]', %q{
+  def getbyte(s, i)
+   byte = begin
+    s.getbyte(i)
+   rescue TypeError
+    :raised
+   end
+
+   byte || :nil
+  end
+
+  getbyte("a", 0)
+  getbyte("a", 0)
+
+  [getbyte("a", 0), getbyte("a", 1), getbyte("a", -1), getbyte("a", -2), getbyte("a", "a")]
+}
+
 # Test << operator on string subclass
 assert_equal 'abab', %q{
   class MyString < String; end
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -1848,7 +1848,7 @@ assert_equal '[97, :nil, 97, :nil, :raised]', %q{
   getbyte("a", 0)
 
   [getbyte("a", 0), getbyte("a", 1), getbyte("a", -1), getbyte("a", -2), getbyte("a", "a")]
-}
+} unless defined?(RubyVM::RJIT) && RubyVM::RJIT.enabled? # Not yet working on RJIT
 
 # Test << operator on string subclass
 assert_equal 'abab', %q{
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4702,6 +4702,34 @@ fn jit_rb_str_bytesize(
     true
 }
 
+fn jit_rb_str_getbyte(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    _ocb: &mut OutlinedCb,
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    asm.comment("String#getbyte");
+    extern "C" {
+        fn rb_str_getbyte(str: VALUE, index: VALUE) -> VALUE;
+    }
+    // Raises when non-integers are passed in
+    jit_prepare_routine_call(jit, asm);
+
+    let index = asm.stack_pop(1);
+    let recv = asm.stack_pop(1);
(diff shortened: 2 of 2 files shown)


bootstraptest/test_yjit.rb
```diff
@@ -1832,6 +1832,18 @@ def jittable_method
...
+# test getbyte on string class
+assert_equal '[97, nil]', %q{
+  def getbyte(s, i)
+    s.getbyte(i) || :nil
+  end
+
+  getbyte("a", 0)
+  getbyte("a", 0)
+
+  [getbyte("a", 0), getbyte("a", 1)]
```
maximecb, line 1844: Could you also test with indices -1 and -2 just for completeness? I know it will work, but it's good to test things thoroughly in case of future refactorings.

yjit/src/codegen.rs
```diff
@@ -4702,6 +4702,32 @@ fn jit_rb_str_bytesize(
...
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    asm.comment("String#getbyte");
+    extern "C" {
+        fn rb_str_getbyte(str: VALUE, index: VALUE) -> VALUE;
+    }
+    // rb_str_getbyte() never allocates
```
jhawthorn, line 4719: @tenderlove pointed out that though this doesn't allocate it (likely) does raise on non-integers so I think we might need to jit_prepare_routine_call after all. Maybe a test would be good to add too that it raises properly when the argument is a non-integer.

(sorry I didn't notice earlier 😅)

XrXr, line 4719: It for non int indexes it calls to_int() so you need to jit_prepare_routine_call().

```
$ ruby -e 'a = []; def a.to_int = p(100); "".getbyte(a)'
100
```

k0kubun: FYI this would change the behavior of a backtrace when it does get raised because we don't push a frame for `MethodGenFn`. It's fine for methods that have a corresponding `opt_*` insn since the interpreter breaks it too, but it's not the case for this method. We already seem to have `String#+@` as another such example (that doesn't have `opt_` insn), so it's not new in this PR, but we should still be careful not to be too aggressive about this.

bootstraptest/test_yjit.rb
```diff
@@ -1832,6 +1832,18 @@ def jittable_method
   jittable_method
 }
 
+# test getbyte on string class
+assert_equal '[97, nil]', %q{
+  def getbyte(s, i)
+    s.getbyte(i) || :nil
```
k0kubun, line 1838:
```suggestion
    s.getbyte(i) || nil
```

ipc103: Thank you! I updated as part of [`38590fe`](https://github.com/ruby/ruby/compare/76251be9e26e088d319600d5c9b895d609cf89ac..38590fe1b9e0bcb3a226390714f9bf4f19011eae)

bootstraptest/test_yjit.rb
ipc103: Thank you! I updated as part of [`38590fe`](https://github.com/ruby/ruby/compare/76251be9e26e088d319600d5c9b895d609cf89ac..38590fe1b9e0bcb3a226390714f9bf4f19011eae)

yjit/src/codegen.rs
jhawthorn: Ah, right. Maybe we should consider a fixnum guard instead?

maximecb (approved):

yjit/src/codegen.rs
k0kubun: If it is easy enough and still helps the benchmarks you want to improve, that'd be great. But I'm personally open to merging it as is as long as there are compelling benchmark improvements.

PR #8404, opened 2023-09-08
[YARP] Implement compiliation for RationalNodes, fix other num types

jemmaissroff:


yarp/yarp_compiler.c
```diff
@@ -27,19 +39,79 @@ parse_number(const yp_node_t *node) {
...
+        size_t fraclen = length - seen_decimal - 1;
+        memmove(decimal, decimal + 1, fraclen + 1);
+
+        VALUE v = rb_cstr_to_inum(buffer, 10, false);
+
+        res = rb_rational_new(v, rb_int_positive_pow(10, fraclen));
+    }
+    else {
+        RUBY_ASSERT(YP_NODE_TYPE_P(((yp_rational_node_t *)node)->numeric, YP_NODE_INTEGER_NODE));
+        VALUE number = rb_cstr_to_inum(buffer, 10, false);
```
kddnewton, line 74: Now that we know we can parse integers without allocating, you could use the same technique we use in `parse_integer` here to avoid allocating out into the buffer.

kddnewton (approved):

kddnewton (approved):

PR #8407, opened 2023-09-08
[DOC] Fix method docs of scheduler#io_read and scheduler#io_write

kudojp:
## Details

* Fixed method docs of `Fiber::Scheduler#io_read` and `io_write`. Each of them takes the `offset` as the fourth argument, which is not documented.


nobu:
In general, we don't update past news.

kudojp:
@nobu
Thanks for your comment..

> In general, we don't update past news.

I removed the change of `NEWS-3.2.0.md` and force pushed.

PR #8411, opened 2023-09-10
Add __owner__

konsolebox:
For https://bugs.ruby-lang.org/issues/19871.

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -1976,7 +1976,7 @@ rb_f_owner(VALUE _)
 {
     rb_control_frame_t *prev_cfp = previous_frame(GET_EC());
     if (!prev_cfp) return Qnil;
-    rb_callable_method_entry_t *me = rb_vm_frame_method_entry(prev_cfp);
+    const rb_callable_method_entry_t *me = rb_vm_frame_method_entry(prev_cfp);
     return me ? me->owner : Qnil;
 }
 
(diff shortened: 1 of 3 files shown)


eval.c
```diff
@@ -1963,6 +1963,23 @@ rb_f_callee_name(VALUE _)
...
+ *  Returns the owner of the current method.
+ *  If called outside of a method, it returns <code>nil</code>.
+ */
+
+static VALUE
+rb_f_owner(VALUE _)
+{
+    rb_control_frame_t *prev_cfp = previous_frame(GET_EC());
+    if (!prev_cfp) return Qnil;
+    rb_callable_method_entry_t *me = rb_vm_frame_method_entry(prev_cfp);
```
nobu, line 1979:
```suggestion
    const rb_callable_method_entry_t *me = rb_vm_frame_method_entry(prev_cfp);
```

nobu (changes requested):
You will need tests, at least.

konsolebox:
> You will need tests, at least.

I'll try to add them asap.

konsolebox (commented):
Already applied

konsolebox:
> You will need tests, at least.

I just added the spec but it's failing in "Rubyspec Version Guards Check / Rubyspec (ruby-3.2) (pull_request)".  Should this one be fixed as well?

konsolebox:
So apparently a constant can't be shadowed by an inheriting class when it's expanded without a scope, so I'll just close this for now until another use case needs it.

PR #8414, opened 2023-09-11
[YARP] Fix up yarp_compiler.c whitespace

kddnewton:
Previously the switch cases were indented by only 2 spaces. This standardizes on 4. Easiest to review with -w.


jemmaissroff:
Thanks for fixing this. Do you mind also adding this commit to `.git-blame-ignore-revs` so the history is easier to preserve?

kddnewton:
Ugh, need to close this and reopen it, too many conflicts

PR #8415, opened 2023-09-11
Return line 0 for JIT frames

tenderlove:
Frames pushed by YJIT have an unreliable PC.  The PC could be garbage, and if we try to read the line number with a garbage PC, then the program can crash.

This commit returns line 0 for programs where there is a `jit_return` function.  If `jit_return` has been set then this frame was pushed by the JIT, and we cannot trust the PC.

Here is a debugger session for a program that crashed due to a broken PC:

```
(lldb) p ruby_current_vm_ptr->ractor.main_thread->ec->cfp->iseq->body->iseq_encoded
(VALUE *) $0 = 0x0000000118a30e00
(lldb) p/x ruby_current_vm_ptr->ractor.main_thread->ec->cfp->pc
(const VALUE *) $1 = 0x0000600000b02d00
(lldb) p/x ruby_current_vm_ptr->ractor.main_thread->ec->cfp->jit_return
(void *) $2 = 0x000000010622942c
```

You can see the PC is completely out of range, but there is a `jit_return` pointer so we can avoid this crash.

### DIFF
--- a/vm_backtrace.c
+++ b/vm_backtrace.c
@@ -1583,6 +1583,7 @@ rb_profile_frames(int start, int limit, VALUE *buff, int *lines)
     int i;
     const rb_execution_context_t *ec = GET_EC();
     const rb_control_frame_t *cfp = ec->cfp, *end_cfp = RUBY_VM_END_CONTROL_FRAME(ec);
+    const rb_control_frame_t *top = cfp;
     const rb_callable_method_entry_t *cme;
 
     // If this function is called inside a thread after thread creation, but
@@ -1613,7 +1614,18 @@ rb_profile_frames(int start, int limit, VALUE *buff, int *lines)
                 buff[i] = (VALUE)cfp->iseq;
             }
 
-            if (lines) lines[i] = calc_lineno(cfp->iseq, cfp->pc);
+            if (lines) {
+                // The topmost frame may not have an updated PC because the JIT
+                // may not have set one.  The JIT compiler will update the PC
+                // before entering a new function (so that `caller` will work),
+                // so only the topmost frame could possibly have an out of date PC
+                if (cfp == top && cfp->jit_return) {
+                    lines[i] = 0;
+                }
+                else {
+                    lines[i] = calc_lineno(cfp->iseq, cfp->pc);
+                }
+            }
 
             i++;
         }


vm_backtrace.c
```diff
@@ -1613,7 +1613,17 @@ rb_profile_frames(int start, int limit, VALUE *buff, int *lines)
                 buff[i] = (VALUE)cfp->iseq;
             }
 
-            if (lines) lines[i] = calc_lineno(cfp->iseq, cfp->pc);
+            if (lines) {
+                // If this frame has a jit_return pointer on it, then the PC
+                // _is not_ reliable.  We can get a crash if we try to calculate
+                // the line number.
+                if (cfp->jit_return) {
+                    lines[i] = 0;
```
k0kubun, line 1621: Could we maybe use `cfp->iseq->body->iseq_encoded` instead of `cfp->pc`?

k0kubun:
We could improve the situation if we land frame outlining, but I think this is what we want for now. cc: @ruby/yjit 

We should also file a ticket to backport this to 3.2.

vm_backtrace.c
tenderlove: I don't understand.  You mean just to decode the line, rather than return 0?

k0kubun: I meant decoding the first line of the iseq, assuming it's not 0? could be just `rb_iseq_first_lineno(cfp->iseq)`.

tenderlove: I think we could, but it seems like `0` is more honest.  Since line numbers should start at 1, I think that 0 means "I don't know".  If we return a real line number, then it might be more confusing.

maximecb (approved):

vm_backtrace.c
k0kubun: ok, I just didn't know it starts at 1 (I thought it starts at 100 if it's 100 in the file). It seems fine then.

k0kubun:
You need to update `test_profile_frames` to make the CI successful, but the change looks good to me.

jhawthorn (approved):

vm_backtrace.c
```diff
@@ -1613,7 +1613,17 @@ rb_profile_frames(int start, int limit, VALUE *buff, int *lines)
...
+                }
+                else {
+                    lines[i] = calc_lineno(cfp->iseq, cfp->pc);
```
jhawthorn, line 1624: alternatively we could check that `cfp->pc` is within `iseq_encoded ... iseq_encoded+iseq_size` and _sometimes_ get a line number (even though it may be inaccurate or just completely incorrect), but it might be better just to always return 0 indicating that we can't be certain.

tenderlove: I was considering this as well.  Maybe we should only return 0 for the topmost frame?  I think all other frames should have a correct PC since it gets flushed before making the call.

(Indeed the crash I saw was in the topmost frame)

PR #8416, opened 2023-09-11
[DOC] RDoc for Process::Status

BurdetteLamar:


process.c
```diff
@@ -566,30 +566,22 @@ proc_get_ppid(VALUE _)
...
- *  exited, signaled) and the upper bits possibly contain additional
- *  information (for example the program's return code in the case of
- *  exited processes). Pre Ruby 1.8, these bits were exposed directly
- *  to the Ruby program. Ruby now encapsulates these in a
- *  Process::Status object. To maximize compatibility,
- *  however, these objects retain a bit-oriented interface. In the
- *  descriptions that follow, when we talk about the integer value of
- *  _stat_, we're referring to this 16 bit value.
+ *  A Process::Status may contain (and usually does contain)
+ *  information about a system process.
```
peterzhu2118, line 570: I'm not sure when `Process::Status` doesn't contain information about a system process?

process.c
```diff
@@ -716,14 +708,12 @@ pst_status(VALUE pst)
...
  *
- *  Returns the bits in _stat_ as an Integer. Poking
- *  around in these bits is platform dependent.
+ *  Returns the system-dependent integer status of +self+:
  *
- *     fork { exit 0xab }         #=> 26566
- *     Process.wait               #=> 26566
- *     sprintf('%04x', $?.to_i)   #=> "ab00"
+ *    `cat /nop`
+ *    $?.to_i # => 256*
```
peterzhu2118, line 716: Is the asterisk here a typo?

process.c
```diff
@@ -826,12 +817,12 @@ pst_to_s(VALUE st)
...
  *  call-seq:
- *     stat.inspect   -> string
+ *    inspect -> string
  *
- *  Override the inspection method.
+ *  Returns a string representation of +self+:
  *
  *    system("false")
- *    p $?.inspect #=> "#<Process::Status: pid 12861 exit 1>"
+ *    $? # => #<Process::Status: pid 1246759 exit 1>
```
peterzhu2118, line 825: We should explicitly call `inspect` because it's only IRB that implicitly calls `inspect` for the output.

process.c
```diff
@@ -942,10 +933,10 @@ pst_wifstopped(VALUE st)
 
 /*
  *  call-seq:
- *     stat.stopsig   -> integer or nil
+ *    stopsig -> integer or nil
  *
- *  Returns the number of the signal that caused _stat_ to stop
- *  (or +nil+ if self is not stopped).
+ *  Returns the number of the signal that caused the process to stop,
+ *  or +nil+ if the process is not stopped,
```
peterzhu2118, line 939:
```suggestion
 *  or +nil+ if the process is not stopped.
```

process.c
BurdetteLamar: Line 1174 has what I thought of as an "empty" Process::Status object;  not so?

process.c
BurdetteLamar: Yup;  fixed.

process.c
BurdetteLamar: Duh.  inspect is the method being docked.  Fixed.

process.c
peterzhu2118: I see. While you're technically correct, I think that saying "may contain" is confusing to the reader and it leaves then guessing in what cases this doesn't contain valid information.

peterzhu2118 (approved):

process.c
peterzhu2118: Thank you!

PR #8417, opened 2023-09-12
YJIT: Add compilation time counter

k0kubun:
This PR adds a counter of cumulative time taken for compiling JIT code. Similar to `GC.stat(:time)`, if you call this before and after your request, you can measure how long you spent in YJIT compilation for each request.

When we compare multiple Ruby versions, we could monitor `GC.stat(:time)` and this to check how much impact each of GC and YJIT had on their difference in performance.

### Example

| benchmark | compile_time_ms |
|:------------|:------------------|
| 30k_ifelse | 1,234ms |
| 30k_methods | 429ms |
| railsbench | 621ms |
| lobsters | 1,699ms |

### DIFF
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -3,6 +3,8 @@
 
 #![allow(dead_code)] // Counters are only used with the stats features
 
+use std::time::SystemTime;
+
 use crate::codegen::CodegenGlobals;
 use crate::core::Context;
 use crate::core::for_each_iseq_payload;
@@ -411,6 +413,7 @@ make_counters! {
     compiled_blockid_count,
     compiled_block_count,
     compiled_branch_count,
+    compile_time_ns,
     compilation_failure,
     block_next_count,
     defer_count,
@@ -838,4 +841,18 @@ pub extern "C" fn rb_yjit_count_side_exit_op(exit_pc: *const VALUE) -> *const VA
 fn global_allocation_size() -> usize {
     let stats = GLOBAL_ALLOCATOR.stats();
     stats.bytes_allocated.saturating_sub(stats.bytes_deallocated)
+}
+
+/// Measure the time taken by func() and add that to yjit_compile_time.
+pub fn with_compile_time<F, R>(func: F) -> R where F: FnOnce() -> R {
+    if get_option!(gen_stats) {
+        let start = SystemTime::now();
+        let ret = func();
+        let nanos = SystemTime::now().duration_since(start)
+            .expect("time went backwards").as_nanos();
+        incr_counter_by!(compile_time_ns, nanos);
+        ret
+    } else {
+        func()
+    }
 }
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -3,7 +3,7 @@
 
 #![allow(dead_code)] // Counters are only used with the stats features
 
-use std::time::SystemTime;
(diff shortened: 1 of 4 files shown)


yjit/src/stats.rs
```diff
@@ -839,3 +842,17 @@ fn global_allocation_size() -> usize {
     let stats = GLOBAL_ALLOCATOR.stats();
     stats.bytes_allocated.saturating_sub(stats.bytes_deallocated)
 }
+
+/// Measure the time taken by func() and add that to yjit_compile_time.
+pub fn with_compile_time<F, R>(func: F) -> R where F: FnOnce() -> R {
+    if get_option!(gen_stats) {
+        let start = SystemTime::now();
+        let ret = func();
+        let nanos = SystemTime::now().duration_since(start)
```
casperisfine, line 851: I know very little of Rust, but looking at the doc this seem to use the "real time" system clock.

Generally for these things we use the monotonic clock, as it never decrease and is marginally faster.

Apparently you just need to use `Instant` instead ? https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/std/time/struct.Instant.html

yjit/src/stats.rs
```diff
@@ -411,6 +413,7 @@ make_counters! {
     compiled_blockid_count,
     compiled_block_count,
     compiled_branch_count,
+    compile_time_ns,
```
maximecb, line 416: Is it `compile_time_ms` or `compile_time_ns` ? 😅

maximecb: Ok I see you're doing the conversion. Was a bit confused because the recorded stat is not the same as reported, but it makes sense 👍

k0kubun:
> could you also measure if there is any impact on total run time for the first iteration of e.g. 30k_ifelse

I ran one iteration of the following benchmarks:

### --yjit

```
before: ruby 3.3.0dev (2023-09-11T17:09:22Z master 1ae5dd6f32) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-09-12T16:02:16Z yjit-compile-time 42b9e18d96) +YJIT [x86_64-linux]

lobsters     1738.2       0.0         1722.0      0.0         1.01           1.01
railsbench   1498.7       0.0         1493.5      0.0         1.00           1.00
30k_ifelse   1345.2       0.0         1327.1      0.0         1.01           1.01
```

### --yjit-stats

```
before: ruby 3.3.0dev (2023-09-11T17:09:22Z master 1ae5dd6f32) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-09-12T16:02:16Z yjit-compile-time 42b9e18d96) +YJIT [x86_64-linux]

lobsters    1918.4       0.0         1909.9      0.0         1.00           1.00
railsbench  1606.6       0.0         1603.7      0.0         1.00           1.00
30k_ifelse  1725.2       0.0         1705.3      0.0         1.01           1.01
```

This PR doesn't seem to impact the 1st itr performance whether this feature is enabled or not.

yjit/src/stats.rs
k0kubun: Thanks! https://github.com/ruby/ruby/pull/8417/commits/42b9e18d966fda0cb1ee1baae1ff02c8433b83df

maximecb (approved):
Very nice 👌

PR #8421, opened 2023-09-12
Ensure signaled processes at opening FIFO terminated

nobu:
When the process hangs up, and the test timed out, the process is left unterminated.


KJTsanaktsidis:
@nobu did you notice this on an ARM macbook perhaps? If so, have you seen https://bugs.ruby-lang.org/issues/19700? I'm fairly sure there's a macOS bug here and the test should just be skipped.

KJTsanaktsidis:
Wait you have seen that issue, you responded to my ticket in May. Never mind!

nobu:
I don't remember the reason for this commit well now, but guess it wasn't ARM machine but on some CIs.

PR #8424, opened 2023-09-12
Don't filter backtrace in make test-spec

XrXr:
Especially on CI, the backtrace is very valuable for debugging.


nobu:
Adding `--debug` always even non-CI?
Why not in `.github/actions/setup/directories/action.yml` and `.*.yml` files?

XrXr:
> Why not in .github/actions/setup/directories/action.yml and .*.yml files?

We're trying to track down a flaky failure on http://ci.rvm.jp/. I guess it's better to make changes there.

PR #8430, opened 2023-09-13
[Backport 3.1] Fix missing write barrier in iseq instruction list

peterzhu2118:
[[Bug #19880]](https://bugs.ruby-lang.org/issues/19880)

There's a missing write barrier for operands in the iseq instruction list, which can cause crashes.

It can be reproduced when Ruby is compiled with `-DRUBY_DEBUG_ENV=1`. Using the following command:

```
RUBY_GC_HEAP_OLDOBJECT_LIMIT_FACTOR=0 RUBY_DEBUG=gc_stress ruby -w --disable=gems -Itool/lib -W0 test.rb
```

The following script crashes:

```ruby
require "test/unit"
```

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -1245,6 +1245,31 @@ new_adjust_body(rb_iseq_t *iseq, LABEL *label, int line)
     return adjust;
 }
 
+static void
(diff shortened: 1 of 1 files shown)


unak (approved):
hmm, LGTM
thx!

PR #8433, opened 2023-09-13
YJIT: Skip adding past_page_bytes for past pages

k0kubun:
Fix https://github.com/Shopify/ruby/issues/534

Currently, `past_page_bytes` is updated whenever `CodeBlock` changes its code page. However, when it moves from a past page to the next page, the bytes written in the page should have been already added to `past_page_bytes`. So this PR skips updating `past_page_bytes` when moving from a past page.


maximecb (approved):
Thanks for the fix and the new test 🙏 :)

PR #8434, opened 2023-09-13
Fix GC.stress= segfaulting on non-bool/integers

sampersand:
## Issue
Whilst updating the RBS type definitions for `GC`, I came across a segfault in `GC.stress=`: If you pass it a non-immediate type, accessing any method on `GC.stress` will segfault.

## How to Reproduce
```ruby
GC.stress = "hello world" # or any allocated object.
GC.stress.class
```

## Cause
The segfault is caused by a use-after-free, as `gc_mark_roots()` doesn't mark `objspace->gc_stress_mode`. I suspect it was never marked because the documentation says that it only accepts `true`, `false`, or a bitwise OR of `0x1` `0x2` and `0x4`, all of which are immediate values.

## The Fix
This PR adds a single line, `rb_gc_mark(ruby_gc_stress_mode);`, to `gc_mark_roots()`. This ensures that the stress mode is always marked, and thus won't be erroneously freed. While I was here, I also updated `GC.stress=` as well to only accept `true`, `false`, and types that define `.to_int`, as the documentation describes.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -7460,6 +7460,7 @@ gc_mark_roots(rb_objspace_t *objspace, const char **categoryp)
     rb_gc_mark(objspace->next_object_id);
     mark_tbl_no_pin(objspace, objspace->obj_to_id_tbl); /* Only mark ids */
 
+    rb_gc_mark(ruby_gc_stress_mode);
     if (stress_to_class) rb_gc_mark(stress_to_class);
 
     MARK_CHECKPOINT("finish");
@@ -11460,6 +11461,10 @@ gc_stress_get(rb_execution_context_t *ec, VALUE self)
 static void
 gc_stress_set(rb_objspace_t *objspace, VALUE flag)
 {
+    if (flag != Qtrue && flag != Qfalse) {
+        flag = rb_to_int(flag);
+    }
+
     objspace->flags.gc_stressful = RTEST(flag);
     objspace->gc_stress_mode = flag;
 }
--- a/spec/ruby/core/gc/stress_spec.rb
+++ b/spec/ruby/core/gc/stress_spec.rb
@@ -20,8 +20,26 @@ describe "GC.stress=" do
     GC.stress = false
   end
 
-  it "sets the stress mode" do
+  it "accepts true, false, and integers" do
     GC.stress = true
     GC.stress.should be_true
+    GC.stress = false
+    GC.stress.should be_false
+    GC.stress = 4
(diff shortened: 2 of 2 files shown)


shyouhei (approved):
LGTM, CI failure seems unrelated to me.

gc.c
```diff
@@ -11460,6 +11461,10 @@ gc_stress_get(rb_execution_context_t *ec, VALUE self)
 static void
 gc_stress_set(rb_objspace_t *objspace, VALUE flag)
 {
+    if (flag != Qtrue && flag != Qfalse) {
+        flag = rb_to_int(flag);
```
nobu, line 11465: GC won't matter by converting to a `Fixnum`.

```suggestion
        flag = UINT2NUM(NUM2UINT(flag));
```

nobu (changes requested):
1. Please file this segfault issue on https://bugs.ruby-lang.org/issues/new to back port the fix.
2. And add `ruby_bug "#xxxx", ""..."3.3" do` ... `end` guard around the new tests.
3. Or move the tests to `test/ruby/test_gc.rb`

gc.c
```diff
@@ -7460,6 +7460,7 @@ gc_mark_roots(rb_objspace_t *objspace, const char **categoryp)
     rb_gc_mark(objspace->next_object_id);
     mark_tbl_no_pin(objspace, objspace->obj_to_id_tbl); /* Only mark ids */
 
+    rb_gc_mark(ruby_gc_stress_mode);
```
nobu, line 7463:
```suggestion
```

gc.c
nobu, line 11465: Fixnums don't need to be marked.
```suggestion
        flag = UINT2NUM(NUM2UINT(flag));
```

nobu (changes requested):

gc.c
sampersand: What's the point of `UINT2NUM` -> `NUM2UINT`?

nobu: Reject non-integer objects.
This may be safer, to avoid Bignum possibility:

```C
    flag = FIX2LONG(NUM2UINT(flag) & ~(~0U << gc_stress_max));
```

nobu, line 11465:
```suggestion
        /* mask valid bits only */
        flag = LONG2FIX(NUM2UINT(flag) & ~(~0U << gc_stress_max));
```

PR #8441, opened 2023-09-14
YJIT: Skip Insn::Comment and format! if disasm is disabled

k0kubun:
`format!` is taking some time even on release build. We should skip the call when `disasm` is not supported.

```
before: ruby 3.3.0dev (2023-09-14T18:16:21Z master 533c4072a9) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-09-14T20:14:08Z yjit-comment b9cbeee272) +YJIT [x86_64-linux]

----------  -----------  ----------  ---------  ----------  ----------  ---------  -------------  ------------
bench       before (ms)  stddev (%)  RSS (MiB)  after (ms)  stddev (%)  RSS (MiB)  after 1st itr  before/after
30k_ifelse  1225.0       0.0         98.8       1118.6      0.0         98.6       1.10           1.10
----------  -----------  ----------  ---------  ----------  ----------  ---------  -------------  ------------
```

### DIFF
--- a/yjit/src/backend/ir.rs
+++ b/yjit/src/backend/ir.rs
@@ -1051,7 +1051,7 @@ impl Assembler
     /// Append an instruction onto the current list of instructions and update
     /// the live ranges of any instructions whose outputs are being used as
     /// operands to this instruction.
-    pub(super) fn push_insn(&mut self, insn: Insn) {
+    pub fn push_insn(&mut self, insn: Insn) {
         // Index of this instruction
         let insn_idx = self.insns.len();
 
@@ -1187,7 +1187,7 @@ impl Assembler
 
         // Spill live stack temps
         if self.ctx.get_reg_temps() != RegTemps::default() {
-            self.comment(&format!("spill_temps: {:08b} -> {:08b}", self.ctx.get_reg_temps().as_u8(), RegTemps::default().as_u8()));
+            asm_comment!(self, "spill_temps: {:08b} -> {:08b}", self.ctx.get_reg_temps().as_u8(), RegTemps::default().as_u8());
             for stack_idx in 0..u8::min(MAX_REG_TEMPS, self.ctx.get_stack_size()) {
                 if self.ctx.get_reg_temps().get(stack_idx) {
                     let idx = self.ctx.get_stack_size() - 1 - stack_idx;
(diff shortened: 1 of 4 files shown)


yjit/src/backend/ir.rs
```diff
@@ -2002,6 +2004,16 @@ impl Assembler {
...
 
+/// Macro to use format! for asm.comment, which skips a format! call
+/// when disasm is not supported.
+macro_rules! asm_comment {
+    ($asm:expr, $($fmt:tt)*) => {
+        #[cfg(feature = "disasm")]
+        $asm.comment(&format!($($fmt)*));
+    };
+}
+pub(crate) use asm_comment;
```
maximecb, line 2015: AFAIK the asm_comment macro shouldn't be necessary. With the cfg! check you put above, asm.comment becomes an empty method if the disasm feature is not enabled.

maximecb:
Very nice perf gain from such a small change. Nicely done once again!

yjit/src/backend/ir.rs
XrXr: It's necessary because the `format!` macro allocates a `String` at the call site before entering the noop function. We have more certainty that we're skipping the allocation with this macro.

yjit/src/backend/ir.rs
```diff
@@ -1724,7 +1724,9 @@ impl Assembler {
     }
 
     pub fn comment(&mut self, text: &str) {
-        self.push_insn(Insn::Comment(text.to_string()));
+        if cfg!(feature = "disasm") {
+            self.push_insn(Insn::Comment(text.to_string()));
```
kddnewton, line 1728: You might want to remove the `comment` function entirely so that folks are forced into using `asm_comment!`. You could replace the `$asm.comment(&format!($($fmt)*));` with `self.push_insn(Insn::Comment(text.to_string()));`

maximecb: Not a bad idea 👍

maximecb (approved):

k0kubun:
Now Ruby 3.3 YJIT compiles faster than Ruby 3.2 YJIT.

```
3.2.2: ruby 3.2.2 (2023-03-30 revision e51014f9c0) +YJIT [x86_64-linux]
3.3.0: ruby 3.3.0dev (2023-09-14T22:49:40Z master 982d6503b9) +YJIT [x86_64-linux]

----------  ----------  ----------  ----------  ----------  -------------  -----------
bench       3.2.2 (ms)  stddev (%)  3.3.0 (ms)  stddev (%)  3.3.0 1st itr  3.2.2/3.3.0
30k_ifelse  1165.4      0.0         1127.4      0.0         1.03           1.03
----------  ----------  ----------  ----------  ----------  -------------  -----------
```

maximecb:
> https://github.com/Shopify/shopify/pull/451618

Wonderful :)

PR #8443, opened 2023-09-14
YJIT: Check for native stack overflow before compiling

XrXr:
Previously, TestStack#test_machine_stack_size failed pretty consistently
on ARM64 macOS, with Rust code and part of the interpreter used for
per-instruction fallback (rb_vm_invokeblock() and friends) touching the
stack guard page and crashing with SEGV. I've also seen the same test
fail on x64 Linux, though with a different symptom.

### DIFF
--- a/vm_insnhelper.c
+++ b/vm_insnhelper.c
@@ -93,6 +93,7 @@ rb_ec_stack_overflow(rb_execution_context_t *ec, int crit)
 #endif
 }
 
+static inline void stack_check(rb_execution_context_t *ec);
 
 #if VM_CHECK_MODE > 0
 static int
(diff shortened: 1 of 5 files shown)


vm_insnhelper.c
```diff
@@ -5565,6 +5566,7 @@ vm_sendish(
 VALUE
 rb_vm_send(rb_execution_context_t *ec, rb_control_frame_t *reg_cfp, CALL_DATA cd, ISEQ blockiseq)
 {
+    stack_check(ec);
```
maximecb, line 5569: You seem to be adding extra checks to the interpreter as well, why?

XrXr: These are used for instruction fallback. You can see usage in our gen_send().

k0kubun: I was worried since this could slow down the fallback calls, but maybe not too significant?

```
before: ruby 3.3.0dev (2023-09-14T15:11:32Z master 6df0927be2) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-09-14T19:33:53Z yjit-stack-overflow ec4cf7c1f6) +YJIT [x86_64-linux]

lobsters    610.8        4.0         610.0       4.1         1.00           1.00
railsbench  1128.5       1.4         1121.7      1.4         1.00           1.01
```

maximecb (approved):

PR #8451, opened 2023-09-15
Process.warmup: invoke `malloc_trim` if available

casperisfine:
Similar to releasing free GC pages, releasing free malloc pages reduce the amount of page faults post fork.

NB: Some popular allocators such as `jemalloc` don't implement it, so it's a noop for them.


gc.c
```diff
@@ -9740,6 +9740,10 @@ rb_gc_prepare_heap(void)
     rb_objspace_each_objects(gc_set_candidate_object_i, NULL);
     gc_start_internal(NULL, Qtrue, Qtrue, Qtrue, Qtrue, Qtrue);
     free_empty_pages();
+
+#if defined(HAVE_MALLOC_TRIM) && !defined(RUBY_ALTERNATIVE_MALLOC_HEADER)
+    malloc_trim(0);
```
casperisfine, line 9745: `jemalloc` has a semi-equivalent capacity via `mallctl`

> `arena.<i>.purge (void) --`

`Purge all unused dirty pages for arena <i>, or for all arenas if <i> equals MALLCTL_ARENAS_ALL.`

So: 

```c
mallctl("arena." STRINGIFY(MALLCTL_ARENAS_ALL) ".purge", NULL, NULL, NULL, 0);
```

peterzhu2118 (approved):

PR #8452, opened 2023-09-15
Free all heap pages at shutdown

HParker:
previously heap_allocated_pages was decremented from heap_page_free causing only half the heap pages to be freed at shutdown


peterzhu2118 (approved):
Thank you!

PR #8457, opened 2023-09-16
Warn the implicit usage of default gem only when bundler is used

mame:
https://bugs.ruby-lang.org/issues/19885


hsbt:
I'll close this because this proposal is outdated from the current master version of `bundled_gems.rb`.

PR #8468, opened 2023-09-18
Support `alias $global $global2` in YARP

tenderlove:
This patch adds alias global support to the YARP compiler.  It only works with normal globals, not special globals like `$1` though.

The test might be a little bit overboard 😅

### DIFF
--- a/test/yarp/compiler_test.rb
+++ b/test/yarp/compiler_test.rb
@@ -242,6 +242,22 @@ module YARP
       assert_equal (1), compile("(1)")
     end
 
+    def test_alias_globals
+      rd, wr = IO.pipe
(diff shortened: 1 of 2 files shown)


test/yarp/compiler_test.rb
```diff
@@ -242,6 +242,22 @@ def test_ParenthesesNode
       assert_equal (1), compile("(1)")
     end
 
+    def test_alias_globals
+      rd, wr = IO.pipe
+      pid = fork {
```
jemmaissroff, line 247: Looks like the windows machines don't have forking?

jemmaissroff:
This was implemented in #8675

PR #8470, opened 2023-09-18
[YARP] Implement ConstantTargetNode

eightbitraptor:


test/yarp/compiler_test.rb
```diff
@@ -103,6 +103,20 @@ def test_ClassVariableOperatorWriteNode
       test_yarp_eval("class YARP::CompilerTest; @@yct = 0; @@yct += 1; end")
     end
 
+    def test_ConstantTargetNode
+      # We don't call test_yarp_eval directly in this case becuase we
```
jemmaissroff, line 107: Should we extract this pattern into a helper too? Doesn't need to be within the scope of this PR

jemmaissroff (approved):
Approving with an optional comment

test/yarp/compiler_test.rb
eightbitraptor: I'm happy to look at this in a seperate PR 👍

PR #8472, opened 2023-09-18
[DOC] Add production performance tips to yjit.md

k0kubun:
* Added "Speeding Up YJIT" and "Saving YJIT Memory Usage" that contain tips for tuning YJIT in production.
* Tweaked a couple of places in "Performance Tips" since exceptions and `Integer#times` are sped up in Ruby 3.3.

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -186,6 +186,56 @@ ruby 3.3.0dev (2023-01-31T15:11:10Z master 2a0bf269c9) +YJIT dev [x86_64-darwin2
 
 We have collected a set of benchmarks and implemented a simple benchmarking harness in the [yjit-bench](https://github.com/Shopify/yjit-bench) repository. This benchmarking harness is designed to disable CPU frequency scaling, set process affinity and disable address space randomization so that the variance between benchmarking runs will be as small as possible. Please kindly note that we are at an early stage in this project.
 
+## Performance Tips for Production Deployments
+
+While YJIT options default to what we think would work well for most workloads,
+they might not necessarily be the best configuration for your application.
+
+This section covers tips on improving YJIT performance in case YJIT does not
+speed up your application in production.
+
+### Increasing --yjit-exec-mem-size
+
+When JIT code size (`RubyVM::YJIT.runtime_stats[:code_region_size]`) reaches this value,
+YJIT triggers "code GC" that frees all JIT code and starts recompiling everything.
+Compiling code takes some time, so scheduling code GC too frequently slows down your application.
+Increasing `--yjit-exec-mem-size` may speed up your application if `RubyVM::YJIT.runtime_stats[:code_gc_count]` is not 0 or 1.
+
+### Running workers as long as possible
+
+It's helpful to call the same code as many times as possible before a process restarts.
+If a process is killed too frequently, the time taken for compiling methods may outweigh
+the speedup obtained by compiling them.
+
+You should monitor the number of requests each process has served.
+If you're periodically killing worker processes, e.g. with `unicorn-worker-killer` or `puma_worker_killer`,
+you may want to reduce the killing frequency or increase the limit.
+
+## Saving YJIT Memory Usage
+
+YJIT allocates memory for JIT code and metadata. Enabling YJIT generally results in more memory usage.
+
+This section goes over tips on minimizing YJIT memory usage in case it uses more than your capacity.
+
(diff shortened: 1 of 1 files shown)


doc/yjit/yjit.md
```diff
@@ -242,6 +241,56 @@ Performance counter names are not guaranteed to remain the same between Ruby ver
...
+
+### Increasing --yjit-call-threshold
+
+As of Ruby 3.2, `--yjit-call-threshold` defaults to 30. With this default, some applications end up
+compiling methods that are used only during the application boot. Increasing this option may help
+you reduce the size of JIT code and metadata. It's worth trying different values like `--yjit-call-threshold=100`.
+
+Note that increasing the value too much may result in compiling code too late.
+You should monitor how many requests each worker processes before it's restarted. For example,
+if each process only handles 1000 requests, `--yjit-call-threshold=1000` might be too large for your application.
```
maximecb, line 283: I think I would move this up first

doc/yjit/yjit.md
```diff
@@ -242,6 +241,56 @@ Performance counter names are not guaranteed to remain the same between Ruby ver
 
 The printed text after a --yjit-stats run includes other information that may be named differently than the information in runtime_stats.
 
+## Speeding Up YJIT
```
maximecb, line 244:
```suggestion
## Performance Tips for Production Deployments
```

doc/yjit/yjit.md
```diff
@@ -186,19 +186,18 @@ ruby 3.3.0dev (2023-01-31T15:11:10Z master 2a0bf269c9) +YJIT dev [x86_64-darwin2
 
 We have collected a set of benchmarks and implemented a simple benchmarking harness in the [yjit-bench](https://github.com/Shopify/yjit-bench) repository. This benchmarking harness is designed to disable CPU frequency scaling, set process affinity and disable address space randomization so that the variance between benchmarking runs will be as small as possible. Please kindly note that we are at an early stage in this project.
 
-### Performance Tips
+## Performance Tips
```
maximecb, line 189:
```suggestion
## Code Optimization Tips
```

I think we should rename this and move it below the production performance tips, because people seem not very likely to actually modify their existing Ruby code.

maximecb (commented):
Thank you for writing this Kokubun 🙏 

Wrote some small suggestions.

doc/yjit/yjit.md
k0kubun: It's what this section suggests (in the first paragraph) too. Tweaked the first paragraph https://github.com/ruby/ruby/pull/8472/commits/a8b81eb2ace2405e89b3b346b36bd01ef93ed5da a little in case it helps?

But I don't think it's a good idea to encourage increasing this value infinitely since we initially went with `--yjit-call-threshold=1000` for Core, and observed a speedup only after decreasing it to `--yjit-call-threshold=100`. This is why the second paragraph exists.

k0kubun: Umm, if it's about moving the section, I'm not sure what exactly you mean by "first", first as in first in the page, or just one section before the other tuning tips?

k0kubun: I'm guessing this is what you meant https://github.com/ruby/ruby/pull/8472/commits/53439a73f01aef611bd9989d06c1eb88caa2484b. Please feel free to edit this branch or leave a suggested change comment if it's not the case.

doc/yjit/yjit.md
k0kubun: applied the rename in https://github.com/ruby/ruby/pull/8472/commits/e972db4aa3ef32c10ab085bd5abf1baea8de2b62 and reordered them in https://github.com/ruby/ruby/pull/8472/commits/89428d0416bf57f7626fa4ef14582ecb93526015.

doc/yjit/yjit.md
k0kubun: applied in https://github.com/ruby/ruby/pull/8472/commits/e972db4aa3ef32c10ab085bd5abf1baea8de2b62

doc/yjit/yjit.md
k0kubun: Reading https://github.com/ruby/ruby/pull/8472#discussion_r1329322007, maybe this was what you want https://github.com/ruby/ruby/pull/8472/commits/89428d0416bf57f7626fa4ef14582ecb93526015?

ioquatix (approved):
LGTM!

maximecb (approved):

PR #8473, opened 2023-09-18
Move IO#readline to Ruby

tenderlove:
This commit moves IO#readline to Ruby.  In order to call C functions, keyword arguments must be converted to hashes.  Prior to this commit, code like `io.readline(chomp: true)` would allocate a hash.  This commits moves the keyword "denaturing" to Ruby, allowing us to send positional arguments to the C API and avoiding the hash allocation.

Here is an allocation benchmark for the method:

```
x = GC.stat(:total_allocated_objects)
File.open("/usr/share/dict/words") do |f|
  f.readline(chomp: true) until f.eof?
end
p ALLOCATIONS: GC.stat(:total_allocated_objects) - x
```

Before this commit, the output was this:

```
$ make run
./miniruby -I./lib -I. -I.ext/common  -r./arm64-darwin22-fake  ./test.rb
{:ALLOCATIONS=>707939}
```

Now it is this:

```
$ make run
./miniruby -I./lib -I. -I.ext/common  -r./arm64-darwin22-fake  ./test.rb
{:ALLOCATIONS=>471962}
```

[Bug #19890] [ruby-core:114803]

### DIFF
--- a/io.c
+++ b/io.c
@@ -4357,22 +4357,28 @@ rb_io_set_lineno(VALUE io, VALUE lineno)
     return lineno;
 }
 
-/*
- *  call-seq:
- *    readline(sep = $/, chomp: false)   -> string
- *    readline(limit, chomp: false)      -> string
- *    readline(sep, limit, chomp: false) -> string
- *
- *  Reads a line as with IO#gets, but raises EOFError if already at end-of-stream.
- *
- *  Optional keyword argument +chomp+ specifies whether line separators
- *  are to be omitted.
- */
-
+/* :nodoc: */
 static VALUE
-rb_io_readline(int argc, VALUE *argv, VALUE io)
+io_readline(rb_execution_context_t *ec, VALUE io, VALUE sep, VALUE lim, VALUE chomp)
 {
-    VALUE line = rb_io_gets_m(argc, argv, io);
+    if (NIL_P(lim)) {
+        // If sep is specified, but it's not a string and not nil, then assume
+        // it's the limit (it should be an integer)
+        if (!NIL_P(sep) && NIL_P(rb_check_string_type(sep))) {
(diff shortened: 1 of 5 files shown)


io.c
```diff
@@ -4356,22 +4356,22 @@ rb_io_set_lineno(VALUE io, VALUE lineno)
...
-    VALUE line = rb_io_gets_m(argc, argv, io);
+    VALUE rs = rb_rs;
+    VALUE tmp = Qnil;
+
+    if (NIL_P(sep) || !NIL_P(tmp = rb_check_string_type(sep))) {
+        rs = tmp;
+    }
+    else {
+        lim = sep;
+    }
```
nobu, line 4371: Currently, when non-string separator is given with limit, for instance `STDIN.readline(1, 2)`, raises `TypeError`.
This PR changes the behavior  to silently ignore the limit.

tenderlove: Ah, oops.  Thanks.  I didn't mean to change that.  I'll fix and add a test.

tenderlove:
@nobu would you mind reviewing again?  I'm not confident about the test fix.  🙇

test/ruby/test_io.rb
```diff
@@ -1898,6 +1898,116 @@ def test_set_lineno_gets
...
+      assert_raise(TypeError) do
+        f.readline 1, 2
+      end
+    end
+  end
+
+  def test_readline_raises
+    File.open(__FILE__) do |f|
+      File.open(__FILE__) do |f2|
+        assert_equal f2.read, f.readline(nil)
```
nobu, line 1918: Why not just `assert_equal File.read(__FILE__), f.readline(nil)`?

tenderlove: Fixed, thanks!

PR #8474, opened 2023-09-19
Make popcount bit-masks stricter

nobu:
Each bit run is upto the right shift count, so the each mask does not need more upper bits.


nobu:
In fact, the mask at `>> 4` in `rb_popcount64` is `0x0707070707070707` but not `0x0f0f0f0f0f0f0f0f`.

shyouhei:
FYI These bit masks first appeared in commit https://github.com/ruby/ruby/commit/4cf460a7bb258d3d61414d2f74df4c0f83c6a3af

nurse (approved):

PR #8476, opened 2023-09-19
[YARP] Use the integer base flag

kddnewton:


jemmaissroff:
Code change looks good, can we add tests here too?

kddnewton:
@jemmaissroff this should be tested by virtue of https://github.com/ruby/ruby/blob/3e4ccd6ddc4abbd3f08c628c4e56d54ef00c5bfa/test/yarp/compiler_test.rb#L34-L37.

jemmaissroff:
Ah this was working before because of `-10`. Right, thanks

jemmaissroff (approved):

PR #8478, opened 2023-09-19
[YARP] Implement regex flags

eileencodes:
This fixes the TODO in the code that was passing 0 instead of the regex flags.


jemmaissroff (approved):

kddnewton (approved):
This looks great!

I see that this PR implements `i`, `m`, and `x`. We will also need to support `o`, `n`, `e`, `s`, and `u`. We can either merge this PR as is, or you could implement them here, up to you.

PR #8485, opened 2023-09-20
Do not hide the IO in rb_f_backquote if a scheduler is registered

casperisfine:
Otherwise we end up passing an hidden object to Ruby code which is invalid.

```
NotImplementedError: method `hash' called on hidden T_FILE object (0x0000000119cd5ca0 flags=0xb)
    src/graphql-ruby/spec/support/dummy_scheduler.rb:159:in `io_wait'
```

This `hide` was recently added as a response to https://bugs.ruby-lang.org/issues/19624

### DIFF
--- a/io.c
+++ b/io.c
@@ -10467,7 +10467,11 @@ rb_f_backquote(VALUE obj, VALUE str)
     if (NIL_P(port)) return rb_str_new(0,0);
 
     GetOpenFile(port, fptr);
-    rb_obj_hide(port);
+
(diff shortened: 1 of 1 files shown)


byroot:
FYI: @ioquatix @nobu 

My first reaction was that there was probably no need to hide this instance, but turns out it was added recently to fix a bug, so, I changed it to only hide if a scheduler is there, but I don't like it that much.

Overall I'm not sure hiding it makes sense, the user it iterating the heap to close all IOs, IMO it works as intended.

PR #8487, opened 2023-09-20
fix memory leak in vm_method

HParker:
issue: https://bugs.ruby-lang.org/issues/19896

This introduces a unified reference_count to clarify who is referencing a method. This also allows us to treat the refinement method as the def owner since it counts itself as a reference.

reproducible with this test from @peterzhu2118 

```ruby
10.times do                                                                                                                                                                                          
  20_000.times do                                                                                                                                                                                    
    $c = Class.new do                                                                                                                                                                                
      def foo; end                                                                                                                                                                                   
    end                                                                                                                                                                                              
                                                                                                                                                                                                     
    $m = Module.new do                                                                                                                                                                               
      refine $c do                                                                                                                                                                                   
        def foo; end                                                                                                                                                                                 
      end                                                                                                                                                                                            
    end                                                                                                                                                                                              
                                                                                                                                                                                                     
    Class.new do                                                                                                                                                                                     
      using $m                                                                                                                                                                                       
                                                                                                                                                                                                     
      def initialize                                                                                                                                                                                 
        o = $c.new                                                                                                                                                                                   
        o.method(:foo).unbind                                                                                                                                                                        
      end                                                                                                                                                                                            
    end.new                                                                                                                                                                                          
  end                                                                                                                                                                                                
                                                                                                                                                                                                     
  puts `ps -o rss= -p #{$$}`                                                                                                                                                                         
end
```

before:

```
19792
20944
22224
23504
24784
26064
27344
28624
29776
31056
```

after:

```
18560
18560
18560
18560
18560
18560
18560
18560
18560
18560
```

### DIFF
--- a/method.h
+++ b/method.h
@@ -179,9 +179,9 @@ typedef struct rb_method_optimized {
 struct rb_method_definition_struct {
     BITFIELD(rb_method_type_t, type, VM_METHOD_TYPE_MINIMUM_BITS);
     unsigned int iseq_overload: 1;
-    int alias_count : 27;
-    int complemented_count : 28;
     unsigned int no_redef_warning: 1;
+    unsigned int aliased : 1;
+    int reference_count : 28;
 
     union {
         rb_method_iseq_t iseq;
@@ -214,7 +214,7 @@ void rb_add_method_optimized(VALUE klass, ID mid, enum method_optimized_type, un
 void rb_add_refined_method_entry(VALUE refined_class, ID mid);
 
 rb_method_entry_t *rb_method_entry_set(VALUE klass, ID mid, const rb_method_entry_t *, rb_method_visibility_t noex);
-rb_method_entry_t *rb_method_entry_create(ID called_id, VALUE klass, rb_method_visibility_t visi, const rb_method_definition_t *def);
+rb_method_entry_t *rb_method_entry_create(ID called_id, VALUE klass, rb_method_visibility_t visi, rb_method_definition_t *def);
 
 const rb_method_entry_t *rb_method_entry_at(VALUE obj, ID id);
 
--- a/vm_insnhelper.c
+++ b/vm_insnhelper.c
@@ -3811,7 +3811,7 @@ aliased_callable_method_entry(const rb_callable_method_entry_t *me)
         VM_ASSERT(RB_TYPE_P(orig_me->owner, T_MODULE));
         cme = rb_method_entry_complement_defined_class(orig_me, me->called_id, defined_class);
 
-        if (me->def->alias_count + me->def->complemented_count == 0) {
+        if (me->def->reference_count == 1) {
             RB_OBJ_WRITE(me, &me->def->body.alias.original_me, cme);
         }
         else {
--- a/vm_method.c
+++ b/vm_method.c
@@ -166,7 +166,6 @@ invalidate_negative_cache(ID mid)
     }
 }
 
-static rb_method_entry_t *rb_method_entry_alloc(ID called_id, VALUE owner, VALUE defined_class, const rb_method_definition_t *def);
 const rb_method_entry_t * rb_method_entry_clone(const rb_method_entry_t *src_me);
 static const rb_callable_method_entry_t *complemented_callable_method_entry(VALUE klass, ID id);
 static const rb_callable_method_entry_t *lookup_overloaded_cme(const rb_callable_method_entry_t *cme);
@@ -403,33 +402,25 @@ rb_add_method_optimized(VALUE klass, ID mid, enum method_optimized_type opt_type
 }
 
 static void
-rb_method_definition_release(rb_method_definition_t *def, int complemented)
+rb_method_definition_release(rb_method_definition_t *def)
 {
     if (def != NULL) {
-        const int alias_count = def->alias_count;
-        const int complemented_count = def->complemented_count;
-        VM_ASSERT(alias_count >= 0);
-        VM_ASSERT(complemented_count >= 0);
+        const int reference_count = def->reference_count;
+        def->reference_count--;
 
-        if (alias_count + complemented_count == 0) {
-            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d,%d (remove)\n", (void *)def,
-                                      rb_id2name(def->original_id), alias_count, complemented_count);
+        VM_ASSERT(reference_count >= 0);
+
+        if (def->reference_count == 0) {
+            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d (remove)\n", (void *)def,
+                                      rb_id2name(def->original_id), def->reference_count);
             if (def->type == VM_METHOD_TYPE_BMETHOD && def->body.bmethod.hooks) {
                 xfree(def->body.bmethod.hooks);
             }
             xfree(def);
         }
         else {
-            if (complemented) {
-                VM_ASSERT(def->complemented_count > 0);
-                def->complemented_count--;
-            }
-            else if (def->alias_count > 0) {
-                def->alias_count--;
-            }
-
-            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d->%d,%d->%d (dec)\n", (void *)def, rb_id2name(def->original_id),
-                                      alias_count, def->alias_count, complemented_count, def->complemented_count);
+            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d->%d (dec)\n", (void *)def, rb_id2name(def->original_id),
+                                      reference_count, def->reference_count);
         }
     }
 }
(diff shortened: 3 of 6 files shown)


vm_insnhelper.c
```diff
@@ -3811,7 +3811,7 @@ aliased_callable_method_entry(const rb_callable_method_entry_t *me)
         VM_ASSERT(RB_TYPE_P(orig_me->owner, T_MODULE));
         cme = rb_method_entry_complement_defined_class(orig_me, me->called_id, defined_class);
 
-        if (me->def->alias_count + me->def->complemented_count == 0) {
+        if (me->def->reference_count == 0) {
```
peterzhu2118, line 3814: I think when it allocates, the `reference_count` is 1 right? So we should  be checking for `== 1` here.

vm_method.c
```diff
@@ -406,30 +406,21 @@ static void
...
-
-        if (alias_count + complemented_count == 0) {
-            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d,%d (remove)\n", (void *)def,
-                                      rb_id2name(def->original_id), alias_count, complemented_count);
+        const int reference_count = def->reference_count;
+        def->reference_count--;
+
+        VM_ASSERT(reference_count >= 0);
+
+        if (reference_count == 0) {
```
peterzhu2118, line 414: Why are we checking the original `reference_count`? Shouldn't we be checking the updated `def->reference_count` here?

vm_method.c
```diff
@@ -406,30 +406,21 @@ static void
...
-                VM_ASSERT(def->complemented_count > 0);
-                def->complemented_count--;
-            }
-            else if (def->alias_count > 0) {
-                def->alias_count--;
-            }
-
-            if (METHOD_DEBUG) fprintf(stderr, "-%p-%s:%d->%d,%d->%d (dec)\n", (void *)def, rb_id2name(def->original_id),
-                                      alias_count, def->alias_count, complemented_count, def->complemented_count);
+        } else {
```
peterzhu2118, line 421: Nit:

```suggestion
        }
        else {
```

vm_method.c
```diff
@@ -771,7 +763,7 @@ make_method_entry_refined(VALUE owner, rb_method_entry_t *me)
             rb_method_entry_alloc(me->called_id, me->owner,
                                   me->defined_class ?
                                   me->defined_class : owner,
-                                  method_definition_addref(me->def));
+                                  me->def);
```
peterzhu2118, line 766: I think we need to increment the reference count here.

HParker: If I increment here, the leak comes back. I agree it makes sense that this would increment the count, but without this increment we do seem to clean up correctly. My only theory is it is still the correct count because `reference_count` starts at 1, so it still counts itself correctly?

HParker: Peter helped me find `rb_method_definition_set` can correct this count when it goes wrong. Thanks! ❤️

vm_method.c
```diff
@@ -741,7 +722,7 @@ rb_method_entry_complement_defined_class(const rb_method_entry_t *src_me, ID cal
 }
 
 void
-rb_method_entry_copy(rb_method_entry_t *dst, const rb_method_entry_t *src)
+rb_method_entry_copy(rb_method_entry_t *dst, rb_method_entry_t *src)
```
peterzhu2118, line 725: Why was the `const` removed from `src`?

vm_method.c
```diff
@@ -692,13 +681,7 @@ rb_method_entry_create(ID called_id, VALUE klass, rb_method_visibility_t visi, c
...
 rb_method_entry_clone(const rb_method_entry_t *src_me)
 {
-    rb_method_entry_t *me = rb_method_entry_alloc(src_me->called_id, src_me->owner, src_me->defined_class, src_me->def);
-    if (METHOD_ENTRY_COMPLEMENTED(src_me)) {
-        method_definition_addref_complement(src_me->def);
-    }
-    else {
-        method_definition_addref(src_me->def);
-    }
+    rb_method_entry_t *me = rb_method_entry_alloc(src_me->called_id, src_me->owner, src_me->defined_class, src_me->def, false);
```
peterzhu2118, line 684:
```suggestion
    rb_method_entry_t *me = rb_method_entry_alloc(src_me->called_id, src_me->owner, src_me->defined_class, src_me->def, METHOD_ENTRY_COMPLEMENTED(src_me));
```

vm_method.c
```diff
@@ -899,7 +880,7 @@ rb_method_entry_make(VALUE klass, ID mid, VALUE defined_class, rb_method_visibil
 
         if (RTEST(ruby_verbose) &&
             type != VM_METHOD_TYPE_UNDEF &&
-            (old_def->alias_count == 0) &&
+            (old_def->aliased == 0) &&
```
peterzhu2118, line 883: Nit:

```suggestion
            (old_def->aliased == false) &&
```

method.h
```diff
@@ -179,10 +179,9 @@ typedef struct rb_method_optimized {
 struct rb_method_definition_struct {
     BITFIELD(rb_method_type_t, type, VM_METHOD_TYPE_MINIMUM_BITS);
     unsigned int iseq_overload: 1;
-    int alias_count : 27;
-    int complemented_count : 28;
     unsigned int no_redef_warning: 1;
-
+    unsigned int aliased : 1;
+    int reference_count : 28;
```
peterzhu2118, line 184: Nit: can you add a newline here?

```suggestion
    int reference_count : 28;

```

vm_method.c
```diff
@@ -406,30 +405,22 @@ static void
 rb_method_definition_release(rb_method_definition_t *def, int complemented)
```
peterzhu2118, line 405: I don't think we need to pass `complemented` here anymore.

vm_method.c
```diff
@@ -642,25 +642,13 @@ rb_method_definition_create(rb_method_type_t type, ID mid)
...
-    if (METHOD_DEBUG) fprintf(stderr, "+%p-%s:%d\n", (void *)def, rb_id2name(def->original_id), def->complemented_count);
-    return def;
-}
-
 static rb_method_entry_t *
-rb_method_entry_alloc(ID called_id, VALUE owner, VALUE defined_class, const rb_method_definition_t *def)
+rb_method_entry_alloc(ID called_id, VALUE owner, VALUE defined_class, rb_method_definition_t *def, bool complemented)
 {
+    if (def) {
+        if (!complemented) def->aliased = TRUE;
```
peterzhu2118, line 649: I think we need to set `def->aliased = true` in `method_definition_addref` and accept `complemented` in `rb_method_entry_copy` and `rb_method_definition_set` as well.

vm_method.c
```diff
@@ -744,6 +724,7 @@ rb_method_entry_complement_defined_class(const rb_method_entry_t *src_me, ID cal
 void
 rb_method_entry_copy(rb_method_entry_t *dst, const rb_method_entry_t *src)
 {
+    VM_ASSERT(dst->def == NULL);
```
peterzhu2118, line 727: It doesn't look like we can assume this assertion.

class.c
```diff
@@ -1375,7 +1375,7 @@ move_refined_method(ID key, VALUE value, void *data)
             RB_OBJ_WRITE(me, &me->def->body.refined.orig_me, NULL);
             new_me = rb_method_entry_clone(me);
             rb_method_table_insert(klass, tbl, key, new_me);
-            rb_method_entry_copy(me, orig_me);
+            rb_method_entry_copy(me, orig_me, true);
```
peterzhu2118, line 1378: We shouldn't need to pass `complemented` to `rb_method_entry_copy`. We can determine that using `METHOD_ENTRY_COMPLEMENTED(src}` in `rb_method_entry_copy`.

vm_method.c
```diff
@@ -509,10 +500,21 @@ setup_method_cfunc_struct(rb_method_cfunc_t *cfunc, VALUE (*func)(ANYARGS), int
     cfunc->invoker = call_cfunc_invoker_func(argc);
 }
 
+static rb_method_definition_t *
+method_definition_addref(rb_method_definition_t *def, bool complemented)
+{
+    def->reference_count++;
+    if (!complemented)
+        def->aliased = true;
```
peterzhu2118, line 508: I think this should be:

```suggestion
    if (!complemented &&  def->reference_count > 0) def->aliased = true;
    def->reference_count++;
```

We don't want to set `aliased` for the very first reference to `rb_method_definition_t` when it was just created.

vm_method.c
```diff
@@ -642,25 +644,11 @@ rb_method_definition_create(rb_method_type_t type, ID mid)
...
-    if (METHOD_DEBUG) fprintf(stderr, "+%p-%s:%d\n", (void *)def, rb_id2name(def->original_id), def->complemented_count);
-    return def;
-}
-
 static rb_method_entry_t *
-rb_method_entry_alloc(ID called_id, VALUE owner, VALUE defined_class, const rb_method_definition_t *def)
+rb_method_entry_alloc(ID called_id, VALUE owner, VALUE defined_class, rb_method_definition_t *def, bool complement)
 {
+    if (def)
+        def = method_definition_addref(def, complement);
```
peterzhu2118, line 651: Nit: `method_definition_addref` really should not be returning anything. IMO it makes it confusing because it makes it seem like it can create a different `rb_method_definition_t`.

```suggestion
    if (def) method_definition_addref(def, complement);
```

peterzhu2118 (approved):

PR #8488, opened 2023-09-20
[YARP] Implement ConstantPathTargetNode

eightbitraptor:


yarp/yarp_compiler.c
```diff
@@ -516,6 +516,58 @@ yp_compile_class_path(LINK_ANCHOR *const ret, rb_iseq_t *iseq, const yp_node_t *
...
+yp_compile_multi_write_lhs(rb_iseq_t *iseq, int lineno, const yp_node_t *node, LINK_ANCHOR *const ret, yp_compile_context_t *compile_context, uint8_t *pushed, bool nested) {
+    NODE dummy_line_node = generate_dummy_line_node(lineno, lineno);
+
+    switch (YP_NODE_TYPE(node)) {
+        case YP_MULTI_TARGET_NODE: {
+            yp_multi_target_node_t *cast = (yp_multi_target_node_t *) node;
+            for (size_t index = 0; index < cast->targets.size; index++) {
+                yp_compile_multi_write_lhs(iseq, lineno, cast->targets.nodes[index], ret, compile_context, pushed, false);
+            }
+            break;
```
jemmaissroff, line 536: Can these `break`s be `return`s?

yarp/yarp_compiler.c
```diff
@@ -516,6 +516,58 @@ yp_compile_class_path(LINK_ANCHOR *const ret, rb_iseq_t *iseq, const yp_node_t *
...
+/**
+ * In order to properly compile multiple-assignment, some preprocessing needs to
+ * be performed in the case of call or constant path targets. This is when they
+ * are read, the "parent" of each of these nodes should only be read once (the
+ * receiver in the case of a call, the parent constant in the case of a constant
+ * path).
+ */
+static void
+yp_compile_multi_write_lhs(rb_iseq_t *iseq, int lineno, const yp_node_t *node, LINK_ANCHOR *const ret, yp_compile_context_t *compile_context, uint8_t *pushed, bool nested) {
+    NODE dummy_line_node = generate_dummy_line_node(lineno, lineno);
```
jemmaissroff, line 528: Can we pass in the `dummy_line_node` itself as an arg to `yp_compile_multi_write_lhs` avoid continuing to create it on recursion since we're not using the `lineno` isolated anywhere?

yarp/yarp_compiler.c
```diff
@@ -516,6 +516,58 @@ yp_compile_class_path(LINK_ANCHOR *const ret, rb_iseq_t *iseq, const yp_node_t *
...
+                ADD_INSN1(ret, &dummy_line_node, putobject, rb_cObject);
+            }
+            yp_compile_multi_write_lhs(iseq, lineno, cast->child, ret, compile_context, pushed, cast->parent);
+            break;
+        }
+        case YP_CONSTANT_READ_NODE: {
+            yp_constant_read_node_t *cast = (yp_constant_read_node_t *) node;
+            ADD_INSN1(ret, &dummy_line_node, putobject, nested ? Qfalse : Qtrue);
+            ADD_INSN1(ret, &dummy_line_node, getconstant, ID2SYM(yp_constant_id_lookup(compile_context, cast->name)));
+            *pushed = *pushed + 2;
```
jemmaissroff, line 563: Could this method return `pushed` instead of passing a pointer to it around?

yarp/yarp_compiler.c
```diff
@@ -516,6 +516,58 @@ yp_compile_class_path(LINK_ANCHOR *const ret, rb_iseq_t *iseq, const yp_node_t *
...
+            } else {
+                ADD_INSN(ret, &dummy_line_node, pop);
+                ADD_INSN1(ret, &dummy_line_node, putobject, rb_cObject);
+            }
+            yp_compile_multi_write_lhs(iseq, lineno, cast->child, ret, compile_context, pushed, cast->parent);
+            break;
+        }
+        case YP_CONSTANT_READ_NODE: {
+            yp_constant_read_node_t *cast = (yp_constant_read_node_t *) node;
+            ADD_INSN1(ret, &dummy_line_node, putobject, nested ? Qfalse : Qtrue);
```
jemmaissroff, line 561: Not strictly necessary, but could use `RBOOL` here:

`ADD_INSN1(ret, &dummy_line_node, putobject, RBOOL(!nested));

jemmaissroff (changes requested):

yarp/yarp_compiler.c
eightbitraptor: I haven't changed this, because I refactored the code to rely on `pushed` being a return value rather than a pointer manipulation - `break`ing out of the `case` and having a single `return` reads better to me than having a `return pushed` inside each `case` branch.

jemmaissroff (approved):

PR #8496, opened 2023-09-22
[YARP] Implement MatchWriteNode

eightbitraptor:


yarp/yarp_compiler.c
```diff
@@ -1503,6 +1503,60 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+
+          for (size_t index = 0; index < capture_count; index++) {
+              yp_constant_id_t constant = cast->locals.ids[index];  
+              int local_index = yp_lookup_local_index(iseq, compile_context, constant);
+
+              if (index < (capture_count - 1)) {
+                  ADD_INSN(ret, &dummy_line_node, dup);
+              }
+              ADD_INSN1(ret, &dummy_line_node, putobject, rb_id2sym(yp_constant_id_lookup(compile_context, constant)));
+              ADD_SEND(ret, &dummy_line_node, rb_intern("[]"), INT2FIX(1));
```
kddnewton, line 1527:
```suggestion
              ADD_SEND(ret, &dummy_line_node, idAREF, INT2FIX(1));
```

yarp/yarp_compiler.c
```diff
@@ -1503,6 +1503,60 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+
+              if (capture_count == 1) {
```
kddnewton, line 1529: I think this might be easier to read if this if statement were outside the loop instead. Right now, it's checking it on each iteration. Could you check out what it would look like to move the `if` outside?

yarp/yarp_compiler.c
```diff
@@ -1503,6 +1503,60 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
+                  
+          }
+
+          ADD_INSNL(ret, &dummy_line_node, jump, end_label);
+          ADD_LABEL(ret, fail_label);
+          ADD_INSN(ret, &dummy_line_node, pop);
+
+          for (size_t index = 0; index < capture_count; index++) {
+              yp_constant_id_t constant = cast->locals.ids[index];  
+              int local_index = yp_lookup_local_index(iseq, compile_context, constant);
```
kddnewton, line 1552: It looks like we're fetching the index for all of the constants twice (once in each loop). Instead, could we stack allocate a list that mirrors the constants? You should be able to use `ALLOCV_N` for this.

kddnewton (changes requested):

yarp/yarp_compiler.c
eightbitraptor: Yes, of course you're right! 🤦 I ended up with this structure based on what `compile.c` does in this case - but of course `compile.c` is using a linked list to represent the capture groups and we're not.

eightbitraptor: Actually, on second glance -  I think this is going to be the easiest approach to read - because the `ADD_SETLOCAL` instruction is common between the 2 branches. So basically we need to add the `putobj` and the `send` for each named capture group - and we need to add a `setlocal` in every case, but in the case where there is only one capture group there's an extra jump instruction.

I'm not sure I can think of a cleaner way of implementing this (without duplicating a bunch of code) other than checking whether we only have a single named capture inside the locals list.

kddnewton: Yeah that's basically what I'm suggesting.

```c
          if (capture_count == 1) {
              yp_constant_id_t constant = cast->locals.ids[index];  
              int local_index = yp_lookup_local_index(iseq, compile_context, constant);

              ADD_INSN1(ret, &dummy_line_node, putobject, rb_id2sym(yp_constant_id_lookup(compile_context, constant)));
              ADD_SEND(ret, &dummy_line_node, rb_intern("[]"), INT2FIX(1));

              DECL_ANCHOR(nom);
              INIT_ANCHOR(nom);

              ADD_INSNL(nom, &dummy_line_node, jump, end_label);
              ADD_LABEL(nom, fail_label);
              ADD_LABEL(nom, end_label);
              ADD_SETLOCAL(nom, &dummy_line_node, local_index, 0);

              ADD_SEQ(ret, nom);
              return;
          }

          for (size_t index = 0; index < capture_count; index++) {
              yp_constant_id_t constant = cast->locals.ids[index];  
              int local_index = yp_lookup_local_index(iseq, compile_context, constant);

              if (index < (capture_count - 1)) {
                  ADD_INSN(ret, &dummy_line_node, dup);
              }
              ADD_INSN1(ret, &dummy_line_node, putobject, rb_id2sym(yp_constant_id_lookup(compile_context, constant)));
              ADD_SEND(ret, &dummy_line_node, rb_intern("[]"), INT2FIX(1));
              ADD_SETLOCAL(ret, &dummy_line_node, local_index, 0);
          }
```

I know 2 of those instructions are duplicated, but I think it's easier to follow.

yarp/yarp_compiler.c
```diff
@@ -1503,6 +1503,60 @@ yp_compile_node(rb_iseq_t *iseq, const yp_node_t *node, LINK_ANCHOR *const ret,
...
           return;
       }
+      case YP_MATCH_WRITE_NODE: {
+          yp_match_write_node_t *cast = (yp_match_write_node_t *)node;
+          LABEL *fail_label = NEW_LABEL(lineno);
+          LABEL *end_label = NEW_LABEL(lineno);
+          size_t capture_count = cast->locals.size;
+
+          YP_COMPILE((yp_node_t *)cast->call);
+          VALUE global_variable_name = rb_id2sym(rb_intern("$~"));
```
kddnewton, line 1513: I think this is `tBACKREF`

yarp/yarp_compiler.c
eightbitraptor: Ok 👍

kddnewton (approved):

kddnewton:
Sorry about the rename! This will need a rebase

test/prism/iseq_test.rb
```diff
@@ -230,6 +230,12 @@ def test_LocalVariableOperatorWriteNode
       test_prism_eval("pit = 0; pit += 1")
     end
 
+    def test_MatchWriteNode
+      test_prism_eval("/(?<foo>bar)(?<baz>bar>)/ =~ 'barbar'")
+      test_prism_eval("/(?<foo>bar)/ =~ 'barbar'")
+    end
+
+
```
kddnewton, line 238:
```suggestion
```

kddnewton (approved):

PR #8504, opened 2023-09-23
[DOC] Correction for doc guide + TOC fix in File

BurdetteLamar:
The doc guide is corrected to omit reference to GFM tables, which RDoc does not support.

In linking to verbatim example, I discovered an TOC jump-level that caused most of the headings in File doc to be omitted from the doc's left-TOC for the class.  Now corrected.


BurdetteLamar:
Closed b/c title did not have [DOC].

PR #8508, opened 2023-09-24
[Doc] Fix link to #close in scheduler.c

olleolleolle:
This points to the existing method #close, rather than #io_close, which does not exist.


hsbt:
This was fixed at https://github.com/ruby/ruby/commit/01d1cbbbec3d40cb2e3f564371beb9dc3729c7ba

PR #8516, opened 2023-09-27
Change RNode structure from union to struct

yui-knk:
All kind of AST nodes use same struct RNode, which has u1, u2, u3 union members for holding different kind of data.
This has two problems.

1. Low flexibility of data structure

Some nodes, for example NODE_TRUE, don’t use u1, u2, u3. On the other hand, NODE_OP_ASGN2 needs more than three union members. However they use same structure definition, need to allocate three union members for NODE_TRUE and need to separate NODE_OP_ASGN2 into another node.
This change removes the restriction so make it possible to change data structure by each node type.

2. No compile time check for union member access

It’s developer’s responsibility for using correct member for each node type when it’s union. This change clarifies which node has which type of fields and enables compile time check.

This commit also changes node_buffer_elem_struct buf management to handle different size data with alignment.


node.h
```diff
@@ -19,12 +19,14 @@ typedef void (*bug_report_func)(const char *fmt, ...);
 
 typedef struct node_buffer_elem_struct {
     struct node_buffer_elem_struct *next;
-    long len;
-    NODE buf[FLEX_ARY_LEN];
+    long len; /* Lenght of nodes */
```
nobu, line 22:
```suggestion
    long len; /* Length of nodes */
```

rubyparser.h
```diff
@@ -130,79 +131,13 @@ enum node_type {
     NODE_HSHPTN,
     NODE_FNDPTN,
     NODE_ERROR,
+    NODE_DEF_TEMP,
+    NODE_DEF_TEMP2,
+    NODE_RIPPER,
+    NODE_RIPPER2,
```
nobu, line 137: I think we should use better name...

nobu (approved):

rubyparser.h
yui-knk: What do feel about `NODE_RIPPER_VALUES` instead of `NODE_RIPPER2`? VALUES comes from the fact all 3 fields are VALUE.

PR #8521, opened 2023-09-27
[DOC] RDoc for Open3.popen3

BurdetteLamar:


lib/open3.rb
```diff
@@ -33,56 +33,124 @@
...
+  #   Open3.popen3([env, ] command_line, options = {}) -> [stdin, stdout, stderr, wait_thread]
+  #   Open3.popen3([env, ] exe_path, *args, options = {}) -> [stdin, stdout, stderr, wait_thread]
+  #   Open3.popen3([env, ] command_line, options = {}) {|stdin, stdout, stderr, wait_thread| ... } -> object
+  #   Open3.popen3([env, ] exe_path, *args, options = {}) {|stdin, stdout, stderr, wait_thread| ... } -> object
+  #
+  # Basically a wrapper for Process.spawn that:
+  #
+  # - Creates a child process, by calling Process.spawn with the given arguments.
+  # - Creates streams +stdin+, +stdout+, and +stderr+,
+  #   which are streams $stdin, $stdout, and $stderr in the child process.
```
peterzhu2118, line 46: `$stdin`, `$stdout`, and `$stderr` are global variables. They can be redefined and it does not affect the result of `Open3.popen3`. For example:

```ruby
require "open3"
require "tempfile"

$stdin = Tempfile.new
$stdout = Tempfile.new
$stderr = Tempfile.new

a = Open3.popen3("echo hi")
STDOUT.puts(a[1].read)
```

It's more accurate to say it captures the standard input, standard output, and standard error.

lib/open3.rb
```diff
@@ -33,56 +33,124 @@
...
+  # The caller should close each of the three returned streams.
+  #
+  #   stdin, stdout, stderr, wait_thread = Open3.popen3('echo')
+  #   # => [#<IO:fd 8>, #<IO:fd 10>, #<IO:fd 12>, #<Process::Waiter:0x00007f58d5428f58 run>]
+  #   stdin.close
+  #   stdout.close
+  #   stderr.close
+  #   wait_thread.pid   # => 2210481
+  #   wait_thread.value # => #<Process::Status: pid 2210481 exit 0>
+  #
```
peterzhu2118, line 62: Nit: extra newline.

```suggestion
```

lib/open3.rb
BurdetteLamar: Modified.

lib/open3.rb
BurdetteLamar: Fixed.

BurdetteLamar:
@peterzhu2118, there are a bunch of these methods, so let's get this first one as good as possible.  Any thoughts on the ordering of the material?  The examples work only on the +stdout+ stream;  should we show something about using the other two streams?  Any other thoughts?

peterzhu2118:
I think it's reasonable to assume that the reader has an idea of what standard in, out, and error are and how to interact with them. I think it's out of scope for this documentation to explain how `IO` objects work.

peterzhu2118 (approved):

PR #8522, opened 2023-09-27
YJIT: add heuristic to avoid compiling cold ISEQs

maximecb:
Substantial memory savings and p99 improvements for large production applications possible.

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -165,7 +165,7 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
 
 - `--yjit`: enable YJIT (disabled by default)
 - `--yjit-call-threshold=N`: number of calls after which YJIT begins to compile a function (default 30)
-- `--yjit-cold-threshold=N`: number of calls after which an ISEQ is considered cold and not
+- `--yjit-cold-threshold=N`: number of global calls after which an ISEQ is considered cold and not
 compiled, lower values mean less code is compiled (default 200000)
 - `--yjit-exec-mem-size=N`: size of the executable memory block to allocate, in MiB (default 64 MiB in Ruby 3.2, 128 MiB in Ruby 3.3+)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -165,9 +165,11 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
 
 - `--yjit`: enable YJIT (disabled by default)
 - `--yjit-call-threshold=N`: number of calls after which YJIT begins to compile a function (default 30)
+- `--yjit-cold-threshold=N`: number of calls after which an ISEQ is considered cold and not
+compiled, lower values mean less code is compiled (default 200000)
 - `--yjit-exec-mem-size=N`: size of the executable memory block to allocate, in MiB (default 64 MiB in Ruby 3.2, 128 MiB in Ruby 3.3+)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
-- `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost) 
+- `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost)
 - `--yjit-trace-exits`: produce a Marshal dump of backtraces from specific exits. Automatically enables `--yjit-stats` (must configure and build with `--enable-yjit=stats` to use this)
 - `--yjit-max-versions=N`: maximum number of versions to generate per basic block (default 4)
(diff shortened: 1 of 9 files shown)


vm.c
```diff
@@ -442,10 +442,11 @@ jit_compile_exception(rb_execution_context_t *ec)
     // Increment the ISEQ's call counter and trigger JIT compilation if not compiled
     if (body->jit_exception == NULL) {
         body->jit_exception_calls++;
-        if (rb_yjit_threshold_hit(iseq, body->jit_exception_calls)) {
+        if (body->jit_exception_calls == rb_yjit_call_threshold()) {
```
maximecb, line 445: I reverted this change because the logic in `rb_yjit_call_threshold()` wouldn't work correctly if there are separate thresholds for entry calls and exceptions. We'd need a separate "hotness estimation" path.

yjit/src/stats.rs
```diff
@@ -198,9 +198,10 @@ macro_rules! make_counters {
 
 /// The list of counters that are available without --yjit-stats.
 /// They are incremented only by `incr_counter!` and don't use `gen_counter_incr`.
-pub const DEFAULT_COUNTERS: [Counter; 7] = [
+pub const DEFAULT_COUNTERS: [Counter; 8] = [
     Counter::code_gc_count,
     Counter::compiled_iseq_entry,
+    Counter::iseq_entry_cold,
```
k0kubun, line 204: This is kind of the opposite of `compiled_iseq_entry`, so maybe it's useful to call it `cold_iseq_entry` to contrast them.

yjit/src/yjit.rs
```diff
@@ -46,11 +46,62 @@ pub fn yjit_enabled_p() -> bool {
...
+    if total_calls < call_threshold {
+        return false;
+    }
+
+    // We expect threshold 1 to compile everything immediately
+    // We need to compare for equality here because otherwise we
+    // can get a deaclock with ractors
+    if call_threshold < CALL_COUNT_INTERV {
+        return total_calls == call_threshold;
+    }
```
k0kubun, line 79: I think this should be located inside `if total_calls == call_threshold {` so that most calls execute one less branching. We use a small call threshold only for testing purposes, so we shouldn't optimize for that case.

doc/yjit/yjit.md
```diff
@@ -165,9 +165,11 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
 
 - `--yjit`: enable YJIT (disabled by default)
 - `--yjit-call-threshold=N`: number of calls after which YJIT begins to compile a function (default 30)
+- `--yjit-cold-threshold=N`: number of calls after which an ISEQ is considered cold and not
```
k0kubun, line 168: We should add this to the help (`yjit_options` in ruby.c) as well. It would have more readers than this document.

k0kubun: I guess we should implement that printing logic in options.rs and let ruby.c call it since we tend to forget it and it sometimes gets out of date.

yjit/src/yjit.rs
```diff
@@ -46,11 +46,62 @@ pub fn yjit_enabled_p() -> bool {
...
+
+    // Record the number of calls at the beginning of the interval
+    if total_calls == call_threshold - CALL_COUNT_INTERV {
```
k0kubun, line 82: I'm confused. When `call_threshold` is 30 and `CALL_COUNT_INTERV` is 20, this is hit when `total_calls` is 10. However, `total_calls < call_threshold` is true in that case, so it would have already returned above at "Exit early if the threshold is not reached" (`if total_calls < call_threshold {`). So it seems like this expression is never true. What am I missing here?

yjit/src/stats.rs
maximecb: Good idea

doc/yjit/yjit.md
maximecb: It would be neat if it could be automated somehow, the tricky bit is that you may want to have a shorter description string in the console output than you would in a readme file.

casperisfine, line 168:
```suggestion
- `--yjit-cold-threshold=N`: number of global calls after which an ISEQ is considered cold and not
```
thinking it might be a bit puzlli gotherwise as it;s right after the call threshold that uses `calls` as "calls to that one function (we should rename to method probably). While `cold-threshold` is the global number of method calls.

yjit/src/yjit.rs
maximecb: Whoops! 😅

yjit/src/yjit.rs
```diff
@@ -46,11 +46,55 @@ pub fn yjit_enabled_p() -> bool {
...
+    // Record the number of calls at the beginning of the interval
+    if total_calls + CALL_COUNT_INTERV == call_threshold {
+        let payload = get_or_create_iseq_payload(iseq);
+        let call_count = unsafe { TOTAL_ENTRY_HITS };
+        payload.call_count_at_interv = call_count;
+    }
+
+    // Try to estimate the total time taken (total number of calls) to reach 20 calls to this ISEQ
+    // This give us a ratio of how hot/cold this ISEQ is
+    if total_calls == call_threshold {
```
k0kubun, line 78: It probably doesn't matter too much, but this could be `else if`.

k0kubun (approved):

PR #8523, opened 2023-09-27
Clarify doc of `Class.inherited`

amomchilov:
It's not obvious that ancestors don't get their callbacks invoked unless you know (and remember!) to call `super`, so let's clarify it in the doc.


object.c
```diff
@@ -909,24 +909,34 @@ rb_class_search_ancestor(VALUE cl, VALUE c)
  *
  * Callback invoked whenever a subclass of the current class is created.
  *
+ * The callback is only called on the immediate superclass of the new
+ * subclass, and not further ancestors.
+ * So if you want to propagate it, you'll have to manually call `super`.
```
nobu, line 914:
```suggestion
 * So if you want to propagate it, you'll have to manually call +super+.
```

object.c
```diff
@@ -909,24 +909,34 @@ rb_class_search_ancestor(VALUE cl, VALUE c)
  *
  * Callback invoked whenever a subclass of the current class is created.
  *
+ * The callback is only called on the immediate superclass of the new
```
nobu, line 912: The word "immediate superclass" is questionable.
When a subclass of `Child` in this example is defined, `Parent.inherited` will be called.
Is `Parent` the "immediate superclass" of that subclass?

object.c
```diff
@@ -909,24 +909,34 @@ rb_class_search_ancestor(VALUE cl, VALUE c)
...
+ *    class Child < Parent
  *    end
  *
  * <em>produces:</em>
  *
- *    New subclass: Bar
- *    New subclass: Baz
+ * New GrandParent subclass: Parent
+ * New GrandParent subclass: Child
+ * New Parent subclass: Child
```
nobu, line 939: Needs indentation for verbatim.

```suggestion
 *   New GrandParent subclass: Parent
 *   New GrandParent subclass: Child
 *   New Parent subclass: Child
```

nobu (changes requested):

amomchilov:
Hey guys, I'll try to polish this off for merging tonight :)

object.c
amomchilov: Correct. This wording is a bit tricky, and I don't like how I phrase it here. At the same time, I'm not sure how to improve it 😅 

My current explanation is focusing on the subclass, and talking about _its_ superclass, but that's not a good fit, because `inherited` is a callback on the superclass (i.e. `self` is the superclass).

I was considering this alternative:

> The system calls this method when a new direct subclass is made.
> If you would like the method to be called for all subclasses, including indirect descendants
> you will need to manually call +super+ in all of the descendants' overrides.

It's written from the perspective of `self` (the superclasses not the subclasses) but it's quite wordy and still not very clear. I'd love any suggestions!

amomchilov: Does this look ok now?

amomchilov:
Hi @hsbt, were you interested in merging this PR?

PR #8533, opened 2023-09-28
[PRISM] Fix static hash compilation and compile implicit nodes

kddnewton:


test/ruby/test_compile_prism.rb
```diff
@@ -303,6 +303,27 @@ def test_XStringNode
...
 
+    ############################################################################
+    # Structures                                                               #
+    ############################################################################
+
+    def test_ArrayNode
+      test_prism_eval("[]")
+      test_prism_eval("[1, 2, 3]")
+      test_prism_eval("%i[foo bar baz]")
+      test_prism_eval("%w[foo bar baz]")
```
jemmaissroff, line 314: Can we add a test case for a non-static array too?

jemmaissroff (approved):
Would be helpful to add one more test case as inidcated, but LGTM!

test/ruby/test_compile_prism.rb
kddnewton: The `%w` is non-static because the strings aren't frozen by default

PR #8542, opened 2023-09-29
Merge NODE_DEF_TEMP and NODE_DEF_TEMP2

yui-knk:


nobu:
I thought using `ctxt` instead of `VALUE nd_cval`, but that struct isn’t visible in the header?

yui-knk:
There are 2 options for using `ctxt` instead of `VALUE nd_cval`.

## Option 1

Move these definitions to `rubyparser.h` from `parse.y`

* `enum shareability`
* `enum rescue_context`
* `struct lex_context`

## Option 2

Move these definitions to `parse.y` from `rubyparser.h`

* `struct RNode_DEF_TEMP`
* `RNODE_DEF_TEMP`

I prefer #2 because `NODE_DEF_TEMP` is only used parse.y internal.

What you think about it?

nobu:
I second option 2.

yui-knk:
Added a commit for option 2.

nobu (approved):

PR #8554, opened 2023-09-29
YJIT: Chain-guard opt_mult overflow

k0kubun:
`opt_mult_overflow` is 80% of side exits on SFR today. It appears that we intentionally do that for hashing 64-bit integers, so this PR lets `opt_mult` fallback to a regular C method call in that case.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4181,3 +4181,6 @@ assert_equal '[6, -6, 9671406556917033397649408, -9671406556917033397649408, 212
 
   [r1, r2, r3, r4, r5]
 }
+
+# Integer multiplication and overflow (minimized regression test from test-basic)
+assert_equal '8515157028618240000', %q{2128789257154560000 * 4}
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1347,6 +1347,10 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_opt_mult_overflow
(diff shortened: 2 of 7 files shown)


k0kubun:
`make test-basic RUN_OPTS=--yjit-call-threshold=1` passes on x86_64, but fails on arm64 🤔 I'll look into what's wrong on the arm64 side.

maximecb:
> `make test-basic RUN_OPTS=--yjit-call-threshold=1` passes on x86_64, but fails on arm64 🤔 I'll look into what's wrong on the arm64 side.

We had to play some special tricks in the arm64 backend because `mul` doesn't actually set the overflow flag. May have something to do with that.

k0kubun:
Yeah, it was related.

The immediate problem was that the token next to `Insn::Mul` was changed from `Insn::Jo` to `Insn::PosMarker`. So it failed to peek `Insn::Jo`. It was trivial to fix.

But the more fundamental problem with the special trick was that it effectively replaced `Insn::Jo` (`emit_conditional_jump::<{Condition::VS}>`) with `Insn::Jne` (`emit_conditional_jump::<{Conditional::NE}>`). For chain guards, we need to regenerate `Insn::Jo`, but that would change the condition to a different one, `Condition::VS`.

So in https://github.com/ruby/ruby/pull/8554/commits/639cecd964c98a6e990d0d3b63a41419ce962bb4, I introduced a new instruction `Insn::JoMul` that does:

* x86_64: It works in the same way as `Insn::Jo`.
* arm64: It works like `Insn::Jne`, but signals the preceding `Insn::Mul` to generate the special trick.

I think this implementation is slightly less hacky than the previous design since it no longer replaces the peeked `Insn::Jo` with something else.

PR #8568, opened 2023-09-30
Use rb_node_opt_arg_t and rb_node_kw_arg_t instead of NODE

yui-knk:


ext/ripper/tools/preproc.rb
```diff
@@ -59,9 +59,9 @@ def prelude(f, out)
     when /\A%%/
       out << "%%\n"
       return
-    when /\A%token/, /\A%type/, /\A} <node>/
+    when /\A%token/, /\A%type/, /\A} <(?:node|node_.*)>/
```
nobu, line 62:
```suggestion
    when /\A%token/, /\A%type/, /\A} <node(?>_\w+)?>/
```

ext/ripper/tools/preproc.rb
```diff
@@ -59,9 +59,9 @@ def prelude(f, out)
...
       # types in %union which have corresponding set_yylval_* macro.
-      out << line.sub(/<(?:node|num|id)>/, '<val>')
+      out << line.sub(/<(?:node|node_.*|num|id)>/, '<val>')
```
nobu, line 64:
```suggestion
      out << line.sub(/<(?:node(?>_\w+)?|num|id)>/, '<val>')
```

nobu (approved):

PR #8574, opened 2023-10-01
Refactor node calls

yui-knk:


compile.c
```diff
@@ -5352,8 +5419,8 @@ compile_cpath(LINK_ANCHOR *const ret, rb_iseq_t *iseq, const NODE *cpath)
 static inline int
 private_recv_p(const NODE *node)
 {
-    if (nd_type_p(RNODE_CALL(node)->nd_recv, NODE_SELF)) {
-        NODE *self = RNODE_CALL(node)->nd_recv;
+    if (nd_type_p(get_nd_recv(node), NODE_SELF)) {
+        NODE *self = get_nd_recv(node);
         return RNODE_SELF(self)->nd_state != 0;
```
nobu, line 5424: As `get_nd_recv()` may return 0, isn't this safer?
```suggestion
    NODE *recv = get_nd_recv(node);
    if (recv && nd_type_p(recv, NODE_SELF)) {
        return RNODE_SELF(recv)->nd_state != 0;
```
Although this function wouldn't be called when no receiver node.

yui-knk: Sounds reasonable. Included the change.

PR #8578, opened 2023-10-02
Allow changing RUBY_PATCHLEVEL_STR if RUBY_PATCHLEVEL == -1

k0kubun:
One of Shopify's services uses Ruby 3.3.0-preview2. So we use `ruby ">= 3.3.0.preview2"` in Gemfile. We also want to deploy Ruby master (newer than preview2) to a few instances of that service, but 3.3.0.dev does not satisfy `>= 3.3.0.preview2`.

I want to keep using `>= 3.3.0.preview2` instead of relaxing it to `>= 3.3.0.dev` since I want to use `ruby ">= 3.3.0.preview3"` when I bump it again. To allow installing Ruby master while keeping `ruby ">= 3.3.0.preview2"`, I'd like to change `RUBY_PATCHLEVEL_STR` to something `>= preview2`, e.g. `cppflags=-DRUBY_PATCHLEVEL_STR=pshopify1`.

### DIFF
--- a/version.h
+++ b/version.h
@@ -55,11 +55,13 @@
 #endif
 
 #if RUBY_PATCHLEVEL == -1
-#define RUBY_PATCHLEVEL_STR "dev"
+# ifndef RUBY_PATCHLEVEL_STR
+#  define RUBY_PATCHLEVEL_STR "dev"
+# endif
 #elif defined RUBY_ABI_VERSION
-#error RUBY_ABI_VERSION is defined in non-development branch
+# error RUBY_ABI_VERSION is defined in non-development branch
 #else
-#define RUBY_PATCHLEVEL_STR ""
+# define RUBY_PATCHLEVEL_STR ""
 #endif
 
 #endif /* RUBY_TOPLEVEL_VERSION_H */
--- a/version.h
+++ b/version.h
@@ -55,7 +55,9 @@
 #endif
 
 #if RUBY_PATCHLEVEL == -1
-# ifndef RUBY_PATCHLEVEL_STR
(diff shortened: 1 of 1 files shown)


byroot (approved):
If there is no objections from other committers I would really  like this feature.

nobu:
> I want to keep using `>= 3.3.0.preview2` instead of relaxing it to `>= 3.3.0.dev` since I want to use `ruby ">= 3.3.0.preview3"` when I bump it again. To allow installing Ruby master while keeping `ruby ">= 3.3.0.preview2"`, I'd like to change `RUBY_PATCHLEVEL_STR` to something `>= preview2`, e.g. `cppflags=-DRUBY_PATCHLEVEL_STR=pshopify1`.

I suspect you'll need quotes, `cppflags=-DRUBY_PATCHLEVEL_STR='"pshopify1"'`.

version.h
```diff
@@ -55,11 +55,13 @@
 #endif
 
 #if RUBY_PATCHLEVEL == -1
-#define RUBY_PATCHLEVEL_STR "dev"
+# ifndef RUBY_PATCHLEVEL_STR
```
nobu, line 58:
```suggestion
# ifdef RUBY_PATCHLEVEL_NAME
#  define RUBY_PATCHLEVEL_STR STRINGIZE(RUBY_PATCHLEVEL_NAME)
# else
```

Then `make DEFS=-DRUBY_PATCHLEVEL_NAME=shopify`.
`DEFS` is defaulted to undefined, while `cppflags` may be set.

k0kubun: Thanks for the suggestion 👍

nobu (approved):

PR #8580, opened 2023-10-02
Try macOS M1 runner for YJIT

k0kubun:


k0kubun:
looks like it doesn't work.

k0kubun:
will open a completely new PR.

PR #8582, opened 2023-10-03
YJIT: Call mprotect after entry stub failure

k0kubun:
When `entry_stub_hit` runs out of memory, it currently doesn't call `mark_all_executable` (after `code_gc`). It causes SEGV.

This PR fixes that issue by removing an early return and let the failure case go through the same flow and return at the end.

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -2283,7 +2283,7 @@ fn entry_stub_hit_body(entry_ptr: *const c_void, ec: EcPtr) -> Option<*const u8>
     let pending_entry = gen_entry_chain_guard(&mut asm, ocb, iseq, insn_idx)?;
     asm.compile(cb, Some(ocb));
 
-    // Try to find an existing compiled version of this block
+    // Find or compile a block version
     let blockid = BlockId { iseq, idx: insn_idx };
     let mut ctx = Context::default();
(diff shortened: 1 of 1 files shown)


maximecb (approved):

yjit/src/core.rs
```diff
@@ -2293,33 +2293,32 @@ fn entry_stub_hit_body(entry_ptr: *const c_void, ec: EcPtr) -> Option<*const u8>
...
 
-    // Regenerate the previous entry
-    assert!(!entry_ptr.is_null());
-    let entryref = NonNull::<Entry>::new(entry_ptr as *mut Entry).expect("Entry should not be null");
-    regenerate_entry(cb, &entryref, next_entry);
+    // Commit or retry the entry
+    if blockref.is_some() {
+        // Regenerate the previous entry
+        assert!(!entry_ptr.is_null());
+        let entryref = NonNull::<Entry>::new(entry_ptr as *mut Entry).expect("Entry should not be null");
```
XrXr, line 2306:
```suggestion
        let entryref = NonNull::<Entry>::new(entry_ptr as *mut Entry).expect("Entry should not be null");
```
The expect() on NonNull::new already checks for null. (Sorry, I know it's not part of your change 🙏)

XrXr (approved):

PR #8585, opened 2023-10-03
YJIT: Avoid writing return value to memory in `leave`

XrXr:
Previously, at the end of `leave` we did
`*caller_cfp->sp = return_value`, like the interpreter. With future changes that leaves the SP field uninitialized for C frames, this will become problematic. For cases like returning from `rb_funcall()`, the return value was written above the stack never read anyway (the copy in the return register is what callers use).

Leave the return value in a register at the end of `leave` and have the code at `cfp->jit_return` decide what to do with it. This avoids the unncessary memory write mentioned above. For JIT-to-JIT return, it goes through `asm.stack_push()` and benefits from register allocation for stack temporaries.

Mostly flat on benchmarks, with maybe some marginal speed improvements.

```
yjit-pre: ruby 3.3.0dev (2023-10-02T21:09:51Z master 492e9437dd) +YJIT [x86_64-linux]
yjit-post: ruby 3.3.0dev (2023-10-03T21:32:59Z yjit-register-gen-.. babb9cabf4) +YJIT [x86_64-linux]

--------------  -------------  ----------  ---------  --------------  ----------  ---------  -----------------  ------------------
bench           yjit-pre (ms)  stddev (%)  RSS (MiB)  yjit-post (ms)  stddev (%)  RSS (MiB)  yjit-post 1st itr  yjit-pre/yjit-post
activerecord    41.0           6.8         55.4       41.2            6.7         55.3       0.99               1.00              
chunky-png      662.1          0.3         45.3       651.8           0.3         41.9       1.02               1.02              
erubi-rails     14.2           27.2        104.5      14.2            26.5        105.2      1.07               1.00              
hexapdf         2209.0         0.7         249.6      2204.3          0.6         262.3      1.02               1.00              
liquid-c        56.7           0.8         39.6       56.9            0.4         39.0       1.01               1.00              
liquid-compile  59.4           0.4         34.5       59.6            0.4         34.5       1.01               1.00              
liquid-render   83.5           0.4         38.4       83.2            0.4         38.4       1.00               1.00              
lobsters        905.6          4.6         323.5      903.9           4.5         323.7      1.00               1.00              
mail            125.6          0.2         56.4       125.0           0.3         55.9       1.00               1.00              
psych-load      1794.8         0.1         35.0       1769.8          0.2         38.4       0.99               1.01              
railsbench      1749.3         1.3         107.5      1739.5          1.3         107.6      0.99               1.01              
sequel          66.1           1.7         40.5       65.9            1.7         40.5       1.01               1.00              
--------------  -------------  ----------  ---------  --------------  ----------  ---------  -----------------  ------------------
Legend:
- yjit-post 1st itr: ratio of yjit-pre/yjit-post time for the first benchmarking iteration.
- yjit-pre/yjit-post: ratio of yjit-pre/yjit-post time. Higher is better for yjit-post. Above 1 represents a speedup.
```

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -450,8 +450,11 @@ pub struct Context {
     /// Bitmap of which stack temps are in a register
     reg_temps: RegTemps,
 
-    // Depth of this block in the sidechain (eg: inline-cache chain)
-    chain_depth: u8,
+    /// Fields packed into u8
+    /// - Lower 7 bits: Depth of this block in the sidechain (eg: inline-cache chain)
+    /// - Top bit: Whether this code is the target of a JIT-to-JIT Ruby return
+    ///   ([Self::is_return_landing])
+    chain_depth_return_landing: u8,
 
     // Local variable types we keep track of
     local_types: [Type; MAX_LOCAL_TYPES],
@@ -1399,7 +1402,7 @@ fn find_block_version(blockid: BlockId, ctx: &Context) -> Option<BlockRef> {
 /// Produce a generic context when the block version limit is hit for a blockid
 pub fn limit_block_versions(blockid: BlockId, ctx: &Context) -> Context {
     // Guard chains implement limits separately, do nothing
-    if ctx.chain_depth > 0 {
+    if ctx.get_chain_depth() > 0 {
         return ctx.clone();
     }
 
@@ -1607,6 +1610,9 @@ impl Context {
         generic_ctx.stack_size = self.stack_size;
         generic_ctx.sp_offset = self.sp_offset;
         generic_ctx.reg_temps = self.reg_temps;
+        if self.is_return_landing() {
+            generic_ctx.set_as_return_landing();
+        }
         generic_ctx
     }
 
@@ -1637,15 +1643,30 @@ impl Context {
     }
 
     pub fn get_chain_depth(&self) -> u8 {
-        self.chain_depth
+        self.chain_depth_return_landing & 0x7f
     }
 
     pub fn reset_chain_depth(&mut self) {
-        self.chain_depth = 0;
+        self.chain_depth_return_landing &= 0x80;
     }
 
     pub fn increment_chain_depth(&mut self) {
-        self.chain_depth += 1;
+        if self.get_chain_depth() == 0x7f {
+            panic!("max block version chain depth reached!");
+        }
+        self.chain_depth_return_landing += 1;
+    }
+
+    pub fn set_as_return_landing(&mut self) {
+        self.chain_depth_return_landing |= 0x80;
+    }
+
+    pub fn clear_return_landing(&mut self) {
+        self.chain_depth_return_landing &= 0x7f;
+    }
+
+    pub fn is_return_landing(&self) -> bool {
+        self.chain_depth_return_landing & 0x80 > 0
     }
 
     /// Get an operand for the adjusted stack pointer address
@@ -1842,13 +1863,17 @@ impl Context {
         let src = self;
 
         // Can only lookup the first version in the chain
-        if dst.chain_depth != 0 {
+        if dst.get_chain_depth() != 0 {
             return TypeDiff::Incompatible;
         }
 
         // Blocks with depth > 0 always produce new versions
         // Sidechains cannot overlap
-        if src.chain_depth != 0 {
+        if src.get_chain_depth() != 0 {
+            return TypeDiff::Incompatible;
+        }
+
+        if src.is_return_landing() != dst.is_return_landing() {
             return TypeDiff::Incompatible;
         }
 
@@ -2495,6 +2520,9 @@ fn branch_stub_hit_body(branch_ptr: *const c_void, target_idx: u32, ec: EcPtr) -
         let running_iseq = rb_cfp_get_iseq(cfp);
         let reconned_pc = rb_iseq_pc_at_idx(running_iseq, target_blockid.idx.into());
         let reconned_sp = original_interp_sp.offset(target_ctx.sp_offset.into());
+        // Unlike in the interpreter, our `leave` doesn't write to the caller's
+        // SP -- we do it in the returned-to code. Account for this difference.
+        let reconned_sp = reconned_sp.add(target_ctx.is_return_landing().into());
 
         assert_eq!(running_iseq, target_blockid.iseq as _, "each stub expects a particular iseq");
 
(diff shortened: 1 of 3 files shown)


yjit/src/core.rs
```diff
@@ -2675,36 +2709,51 @@ pub fn gen_branch_stub_hit_trampoline(ocb: &mut OutlinedCb) -> CodePtr {
...
 
     // Jump to the address returned by the branch_stub_hit() call
     asm.jmp_opnd(jump_addr);
 
+    // HACK: restoring the value of C_RET_REG clobbers the
+    // return value of branch_stub_hit we want to jump to,
+    // so we need a scratch regsiter. This unrechable `test`
+    // maintain liveness of C return reg so we get something
+    // else for the scratch register.
+    asm.test(stub_hit_ret, 0.into());
```
XrXr, line 2735: This is unfortunate, but I don't know if there is another way.

yjit/src/core.rs
```diff
@@ -450,8 +450,11 @@ pub struct Context {
     /// Bitmap of which stack temps are in a register
     reg_temps: RegTemps,
 
-    // Depth of this block in the sidechain (eg: inline-cache chain)
-    chain_depth: u8,
+    /// Fields packed into u8
+    /// - Lower 7 bits: Depth of this block in the sidechain (eg: inline-cache chain)
+    /// - Top bit: Whether this code is the target of a JIT-to-JIT Ruby return
+    ///   ([Self::is_return_landing])
+    chain_depth_plus: u8,
```
k0kubun, line 457: I'm not sure what `plus` means here. Is it "chain_depth plus (is_return_landing)"?

I would probably call it `chain_depth_return_landing` or `chain_depth_and_return_landing`, but they are too long I guess.

yjit/src/core.rs
k0kubun, line 2735: This would waste fewer bytes on x86_64.

```suggestion
    asm.test(stub_hit_ret, stub_hit_ret);
```

k0kubun: I don't either, but I guess it doesn't matter too much since it's a shared trampoline (wastes bytes only once) and the instruction is never executed.

k0kubun (approved):

yjit/src/core.rs
maximecb: image

yjit/src/core.rs
```diff
@@ -2675,36 +2709,51 @@ pub fn gen_branch_stub_hit_trampoline(ocb: &mut OutlinedCb) -> CodePtr {
...
 
     // Jump to the address returned by the branch_stub_hit() call
     asm.jmp_opnd(jump_addr);
 
+    // HACK: restoring the value of C_RET_REG clobbers the
+    // return value of branch_stub_hit we want to jump to,
+    // so we need a scratch regsiter. This unrechable `test`
+    // maintain liveness of C return reg so we get something
+    // else for the scratch register.
+    asm.test(stub_hit_ret, stub_hit_ret);
```
maximecb, line 2735: I feel like we've run into this before. Isn't this what `asm.live_reg_opnd` is for? 🤔

XrXr:
This is indeed a similar situation that prompted the addition of `live_reg_opnd()`, but subtly different. We've only used it to make a register live out of thin air before, but not to stretch the live range of a register like we need here.

But it can be used for arbitrarily setting the end of a live range! With:

```diff
-    asm.test(stub_hit_ret, stub_hit_ret);
+    let _ = asm.live_reg_opnd(stub_hit_ret);
```

I get: 
```
  # branch_stub_hit() trampoline (x64)
  mov rdx, r12
  call 0x55bc8837c7a0
  mov rcx, rax
  pop rax
  pop r10
  pop r9
  pop r8
  pop rdi
  pop rsi
  jmp rcx

  # branch_stub_hit() trampoline (A64)
  mov x2, x20
  mov x16, #0xff78
  movk x16, #0x9a, lsl #16
  movk x16, #1, lsl #32
  blr x16
  mov x11, x0
  ldr x0, [sp], #0x10
  ldr x15, [sp], #0x10
  ldr x14, [sp], #0x10
  ldr x10, [sp], #0x10
  ldr x9, [sp], #0x10
  ldr x1, [sp], #0x10
  br x11
```

Usually unused `InsnOut` leads to a panic, but I guess not in this case due to the special nature of `live_reg_opnd()`.

> For changes like this, I think it would make sense to also look at microbenchmarks that make heavy use of calls, such as fib and 30k_methods. Not that these have to be the deciding factor, but they might highlight performance differences more clearly.

Interestingly the change actually reduces perf on `fib` and other `--category micro` benchmarks:

```
--------------  -------------  ----------  --------------  ----------  -----------------  ------------------
bench           yjit-pre (ms)  stddev (%)  yjit-post (ms)  stddev (%)  yjit-post 1st itr  yjit-pre/yjit-post
fib             44.0           0.1         48.0            0.3         0.91               0.92              
setivar_young   59.4           41.9        63.0            38.4        1.00               0.94              
keyword_args    37.3           9.7         41.1            8.9         0.95               0.91              
30k_ifelse      328.1          0.1         331.4           0.1         0.99               0.99              
30k_methods     736.3          0.0         738.4           0.0         0.99               1.00              
respond_to      16.4           14.5        15.7            14.2        1.04               1.05              
cfunc_itself    27.3           11.0        27.2            11.3        1.00               1.01              
getivar         24.8           69.4        24.8            71.0        0.98               1.00              
setivar         13.4           73.0        13.4            70.6        1.03               1.00              
setivar_object  60.4           41.0        60.5            41.4        1.01               1.00              
str_concat      46.6           0.4         46.4            0.3         1.00               1.00              
throw           26.7           0.3         26.5            0.3         1.00               1.01              
--------------  -------------  ----------  --------------  ----------  -----------------  ------------------
```

Needs more investigation.

XrXr:
For `fib`, the deficit is from changing the alignment of the jump at the end of `leave`. Because it's a recursive method the native code forms a loops. Drilling in with [`perf` and Top-Down](https://perf.wiki.kernel.org/index.php/Top-Down_Analysis) functionality eventually reveals that post-change, there is 3x more `frontend_retired.dsb_miss`.

Searching around reveals that the [DSB](https://www.intel.com/content/www/us/en/docs/vtune-profiler/user-guide/2023-0/cpu-metrics-reference.html#INFO-DSB-COVERAGE) is sensitive to alignment of jump instructions, so I added some padding to the beginning of `gen_prologue()` to change the layout:

```
diff --git a/yjit/src/codegen.rs b/yjit/src/codegen.rs
index e4fa1a3665..ccea4b53bf 100644
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -720,6 +720,9 @@ pub fn gen_entry_prologue(
 
     asm.frame_setup();
 
+    asm.test(SP, SP);
+    asm.test(SP, SP);
+
     // Save the CFP, EC, SP registers to the C stack
     asm.cpush(CFP);
     asm.cpush(EC);

```

Sure enough, `frontend_retired.dsb_miss` dropped down to pre-change levels and overall performance becomes even.

So, I think the speed drop from `fib` is not really a useful signal and more just a revelation of an uarch quirk. There is so little code that arcane things like DSB friendliness becomes a major factor for speed.

maximecb:
Thanks for investigating this in depth Alan 🙏 

When you say jump alignment, you mean the start address of fib? Do you think there's any value in trying to align function start addresses to cache line boundaries?

XrXr:
> Do you think there's any value in trying to align function start addresses to cache line boundaries?

It's sometimes an optimization and sometimes not, depending on the uarch. For DSB specifically, it tries to cache uops that repeatedly runs inside a native loop, and is sensitive to the placement of the contents of the loop. By inserting padding we might bump the code for the loop out of / into alignment, depending on the volume of code in the loop and what comes before the loop etc. Also, DSBs are specific to Intel uarches, and the size / alignment they like vary between models.
Due to the diversity of uarch quirks the outcome might as well be unpredictable. Hard to justify putting something there we don't know how it would behave.

PR #8586, opened 2023-10-04
[DOC] Fix description for rb_postponed_job_register_one()

osyoyu:
The current documentation for rb_postponed_job_register_one() is explaining the differences with itself, where it should be explaining the differences with rb_postponed_job_register().

### DIFF
--- a/include/ruby/debug.h
+++ b/include/ruby/debug.h
@@ -628,7 +628,7 @@ typedef void (*rb_postponed_job_func_t)(void *arg);
 int rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void *data);
 
 /**
- * Identical to rb_postponed_job_register_one(),  except it additionally checks
+ * Identical to rb_postponed_job_register(),  except it additionally checks
  * for  duplicated registration.   In case  the passed  job is  already in  the
  * postponed job buffer this function does nothing.
  *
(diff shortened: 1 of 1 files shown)


osyoyu:
I'm wondering how these documentation comments are formatted (they contain mutiple spaces between words to maintain a block-like appearance). Should I follow a particular editor configuration to not break this?

nobu:
In Emacs:

1. set `fill-column` to 79 (`C-x f 79 RET`, or go to the end of the next line and `C-u C-x f RET`).
2. execute `c-fill-paragraph` with prefix arg to justify (`C-u M-q`).

include/ruby/debug.h
```diff
@@ -628,7 +628,7 @@ typedef void (*rb_postponed_job_func_t)(void *arg);
 int rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void *data);
 
 /**
- * Identical to rb_postponed_job_register_one(),  except it additionally checks
+ * Identical to rb_postponed_job_register(),  except it additionally checks
  * for  duplicated registration.   In case  the passed  job is  already in  the
  * postponed job buffer this function does nothing.
```
nobu, line 633:
```suggestion
 * Identical to rb_postponed_job_register(), except  it additionally checks for
 * duplicated registration.  In case the passed job is already in the postponed
 * job buffer this function does nothing.
```

nobu (approved):

nobu:
I don't the way with other editors; you may find it by searching "justify" + your editor.

osyoyu:
Thanks! Seems the way for Vim is using justify.vim or external commands such as par/fmt.
https://vi.stackexchange.com/questions/509/can-i-justify-text-in-vim

PR #8587, opened 2023-10-04
Optimize `Range#count` by using `range_size` if possible

kyanagi:
If no arguments and no block are given, if `Range#begin` is `Integer`, then `Range#count` is equal to `Range#size`.
That can be determined through simple calculations.

# benchmark

```
prelude: |
  r_1 = 1..1
  r_1k = 1..1000
  r_1m = 1..1000000
  r_str = 'a'..'z'
benchmark:
  'int 1': r_1.count
  'int 1K': r_1k.count
  'int 1M': r_1m.count
  string: r_str.count
```

## Iteration per second (i/s)

|        |master|PR|
|:-------|----------------------------------:|-------------------------:|
|int 1   |                             8.948M|                   32.001M|
|        |                                  -|                     3.58x|
|int 1K  |                           106.865k|                   32.580M|
|        |                                  -|                   304.87x|
|int 1M  |                            108.473|                   26.998M|
|        |                                  -|                248888.37x|
|string  |                           929.009k|                  895.954k|
|        |                              1.04x|                         -|

### DIFF
--- a/range.c
+++ b/range.c
@@ -607,6 +607,10 @@ double_as_int64(double d)
 static int
 is_integer_p(VALUE v)
 {
+    if (FIXNUM_P(v) || RB_TYPE_P(v, T_BIGNUM)) {
+        return true;
+    }
+
     ID id_integer_p;
     VALUE is_int;
     CONST_ID(id_integer_p, "integer?");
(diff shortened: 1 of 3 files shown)


range.c
```diff
@@ -607,6 +607,10 @@ double_as_int64(double d)
 static int
 is_integer_p(VALUE v)
 {
+    if (FIXNUM_P(v) || RB_TYPE_P(v, T_BIGNUM)) {
```
nobu, line 610:
```suggestion
    if (rb_integer_type_p(v)) {
```

kyanagi: Thank you!

nobu (approved):

PR #8591, opened 2023-10-05
YJIT: Build on BSD platforms with GNU make

Freaky:
A note in yjit.mk states "BSD handling is in yjit/not_gmake.mk", which appears to just replicate the Linux branch of the gmake version.

This patch expands that branch to cover the BSDs so either option can be used, and has been lightly tested across FreeBSD, OpenBSD, NetBSD, and DragonFly BSD.


jeremyevans (approved):
I was able to build with `--enable-yjit` with BSD make with this patch.

PR #8597, opened 2023-10-05
YJIT: shink local types from 8 to 4 bytes

maximecb:
Looking at `yjit_alloc_size` on `lobsters.rs`:
```
master, context size 21 bytes:
  yjit_alloc_size:          25,887,349
This change, without repr(packed), contest size 20 bytes:
  yjit_alloc_size:          25,555,097
  2.3% memory savings in yjit_alloc_size
This change, with repr(packed), context size 17 bytes
  yjit_alloc_size:          25,524,169
  2.4% memory savings in yjit_alloc_size
```

So obviously, 2.4% savings is not massive, but I think it's good considering this is not a tricky refactoring. Adding `repr(packed)` doesn't get us much, presumably because we are losing some bytes due to alignment constraints elsewhere.

If we played the same trick to compress `TempMapping` some more, we could probably shave one more byte on the context, get it down to 16 bytes, at which point we might suddenly see more substantial memory savings... Alternatively, if we keep this as is, our context is effectively 20 bytes, which means we have 3 bytes free that we could store additional information into.

Obviously, diminishing returns, but I we are very close to 16 bytes if we want to make that happen, and it's funny to think we got the context this small considering it used to be a whole 38 bytes in size!

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -64,6 +64,10 @@ pub enum Type {
 
     BlockParamProxy, // A special sentinel value indicating the block parameter should be read from
                      // the current surrounding cfp
+
+    // The context currently relies on types taking at most 4 bits (max value 15)
+    // to encode, so if we add any more, we will need to refactor the context,
+    // or we could remove HeapSymbol, which is currently unused.
 }
 
 // Default initialization
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -42,7 +42,7 @@ pub type IseqIdx = u16;
 #[derive(Copy, Clone, Hash, PartialEq, Eq, Debug)]
 #[repr(u8)]
 pub enum Type {
-    Unknown,
+    Unknown = 0,
     UnknownImm,
     UnknownHeap,
     Nil,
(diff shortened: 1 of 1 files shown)


yjit/src/core.rs
```diff
@@ -3193,11 +3210,45 @@ impl<T> RefUnchecked for Cell<T> {
 mod tests {
     use crate::core::*;
 
+    #[test]
+    fn type_size() {
+        // Check that we can store types in 4 bits,
+        // and all local types in 32 bits
+        assert_eq!(mem::size_of::<Type>(), 1);
+        assert!(Type::BlockParamProxy as usize <= 0b1111);
```
XrXr, line 3218: Unfortunate we don't have a way[^1](https://github.com/rust-lang/rust/issues/73662) to count the number of variants in a enum to make this robust against adding variants to the end of `Type`, so it may be good to add a comment at the declaration of `Type` to warn against it.

XrXr (approved):
Excellent branch name 😄

yjit/src/core.rs
maximecb: Sure. At the moment I did add an assertion in `set_local_type()` to guard against it. Will add a comment as well.

k0kubun (approved):
:+1:

PR #8598, opened 2023-10-05
Add Array.pack1

jeremyevans:
This operates the same as Array#pack called on a single element array, but avoids the need to allocate an array.  In the included benchmark, Array.pack1 is about 30% faster than Array#pack.

Implements [Feature #18573]


test/ruby/test_pack.rb
```diff
@@ -892,3 +892,499 @@ def test_unpack_offset
...
+  end
+
+  def test_pack_unpack_lL
+    assert_equal(4, Array.pack1(1, "l").bytesize)
+    assert_equal(4, Array.pack1(1, "L").bytesize)
+    assert_operator(4, :<=, Array.pack1(1, "l!").bytesize)
+    assert_operator(4, :<=, Array.pack1(1, "L!").bytesize)
+  end
+
+  require 'rbconfig'
```
nobu, line 1173: I think this and the original line in `TestPack` should be at the top of this file.

nobu (approved):

nobu:
Could you apply the changes in #8600 too?

jeremyevans:
> Could you apply the changes in #8600 too?

@nobu I rebased this on master (after your merge of #8600), and made the other changes you requested.

jeremyevans:
Closing as matz was against having this as a class method.  An alternative approach that doesn't require users to rewrite their code is being worked on in #8734.

PR #8605, opened 2023-10-08
Thread events hooks: Pass concerned Thread to callbacks

osyoyu:
Let Thread event callback functions receive VALUE thread in concern. This will be a more stable way than calling gettid(2) in callbacks.

### DIFF
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -242,8 +242,8 @@ rb_native_cond_timedwait(rb_nativethread_cond_t *cond, pthread_mutex_t *mutex, u
 // thread scheduling
 
 static rb_internal_thread_event_hook_t *rb_internal_thread_event_hooks = NULL;
-static void rb_thread_execute_hooks(rb_event_flag_t event);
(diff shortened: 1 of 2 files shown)


thread_pthread.c
```diff
@@ -2451,9 +2451,12 @@ rb_thread_execute_hooks(rb_event_flag_t event)
 
     if (rb_internal_thread_event_hooks) {
         rb_internal_thread_event_hook_t *h = rb_internal_thread_event_hooks;
+        rb_internal_thread_event_data_t *data = ALLOC_N(rb_internal_thread_event_data_t, 1);
+        data->thread = th->self;
```
nobu, line 2455: Why dynamically allocating?
This seems a memory leak.

```suggestion
        rb_internal_thread_event_data_t data[1] = {{.thread = th->self}};
```

jpcamara:
This was added in a separate PR so this should no longer be necessary https://github.com/ruby/ruby/pull/8885

PR #8619, opened 2023-10-10
YJIT: Allow --yjit-trace-exits on release builds

k0kubun:
I don't think we need to require `--enable-yjit=stats` or `--enable-yjit=dev` for using `--yjit-trace-exits`. It should have no impact when you don't use `--yjit-trace-exits` whether we require it or not.


maximecb (approved):
👍

maximecb:
@tenderlove just so you are aware this is now possible.

PR #8621, opened 2023-10-10
YJIT: Fix constant invalidation for ic_cref

k0kubun:
`rb_yjit_constant_ic_update` currently does not invalidate `opt_getconstant_path` when its inline cache has `ic_cref` because it assumes it cannot generate code for that case.

However, that case has been supported since https://github.com/Shopify/yjit/pull/240, so we should invalidate such blocks and re-generate them.

This helps `ratio_in_yjit` of lobsters a little. `opt_getconstant_path` has another problem to be fixed, but it's out of scope in this PR.

### Before
```
ratio_in_yjit:                 98.1%
avg_len_in_yjit:                55.0
Top-20 most frequent exit ops (100.0% of exits):
                opt_getconstant_path:     14,810 (33.5%)
                              opt_eq:     13,420 (30.4%)
                  getblockparamproxy:     10,987 (24.9%)
                                  ...
```

### After
```
ratio_in_yjit:                 98.3%
avg_len_in_yjit:                56.3
Top-20 most frequent exit ops (100.0% of exits):
                              opt_eq:     13,419 (36.8%)
                  getblockparamproxy:     10,987 (30.2%)
                opt_getconstant_path:      7,064 (19.4%)
                                  ...
```

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -476,6 +476,25 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_opt_getconstant_path_slowpath_invalidation
+    assert_compiles(<<~RUBY, exits: { opt_getconstant_path: 1 }, result: [1, 1], call_threshold: 1)
+      class A
+        FOO = 1
+        class << self
+          def foo
+            _foo = nil
+            FOO
+          end
+        end
+      end
+
+      result = []
+      result << A.foo # exit and then invalidate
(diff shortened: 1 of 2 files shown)


k0kubun:
Hmm, interesting 🤔 I didn't expect other tests to fail.

https://github.com/ruby/ruby/blob/8b7a979cb8e2c9dc34626929d7860d3f6680f18f/test/ruby/test_yjit.rb#L908-L928

I'll reopen another PR with a different approach.

PR #8623, opened 2023-10-11
YJIT: Fallback opt_getconstant_path for const_missing

k0kubun:
`opt_getconstant_path_no_ic_entry` is the top exit reason for SFR and lobsters.

On lobsters, one of the classes used by ActiveRecord uses a class generated by `DelegateClass`. Some constant accesses in it always go to `Delegator.const_missing`. When it calls `const_missing`, `vm_ic_update` skips updating `ic->entry`. So the IC for such accesses is never filled.

To avoid exiting JIT code in such cases, this PR lets it fallback to the interpreter's implementation. That invalidates the block if the IC is filled, so it will be re-compiled to optimized code whenever possible. With a normal call threshold, the IC entry should have been already filled when YJIT starts compiling it, so this shouldn't generate wasted code in most cases.

This improves `ratio_in_yjit` on lobsters:

### Before
```
opt_getconstant_path exit reasons:
    no_ic_entry:     14,807 (100.0%)
        ic_miss:          4 ( 0.0%)

ratio_in_yjit:                 98.1%
```

### After
```
opt_getconstant_path exit reasons:
    ic_miss:          4 (100.0%)

ratio_in_yjit:                 98.4%
```


vm_insnhelper.c
```diff
@@ -5846,6 +5846,33 @@ vm_ic_update(const rb_iseq_t *iseq, IC ic, VALUE val, const VALUE *reg_ep, const
...
+    struct iseq_inline_constant_cache_entry *ice = ic->entry;
+    if (ice && vm_ic_hit_p(ice, GET_EP())) {
+        val = ice->value;
+
+        VM_ASSERT(val == vm_get_ev_const_chain(ec, segments));
+    } else {
+        ruby_vm_constant_cache_misses++;
+        val = vm_get_ev_const_chain(ec, segments);
+        vm_ic_track_const_chain(GET_CFP(), ic, segments);
+        // Because leaf=false, we need to undo the PC increment to get the address to this instruction
```
XrXr, line 5863:
```suggestion
        // Undo the PC increment to get the address to this instruction
```
Since we always increment PC now

vm_insnhelper.c
```diff
@@ -5846,6 +5846,33 @@ vm_ic_update(const rb_iseq_t *iseq, IC ic, VALUE val, const VALUE *reg_ep, const
...
+        // INSN_ATTR(width) == 2
+        vm_ic_update(GET_ISEQ(), ic, val, GET_EP(), GET_PC() - 2);
+    }
+    return val;
+}
+
+VALUE
+rb_vm_opt_getconstant_path(rb_execution_context_t *ec, rb_control_frame_t *const reg_cfp, IC ic)
```
XrXr, line 5871: Nit: maybe have just the non-static `rb_` variant for code size in this case? Up to you.

XrXr (approved):
Thanks!

vm_insnhelper.c
k0kubun: It was a deliberate choice for potentially saving a function call overhead since `static` could be inlined whereas non-`static` functions are never inlined. Looking at the code generated by GCC 11.4, however, `vm_opt_getconstant_path` was compiled as a separate function. So I removed the unneeded wrapper. `rb_vm_opt_getconstant_path` had only `endbr64` and `jmp` (+ `nop`) and the code size impact wasn't too bad, but it'd be nice to remove a `jmp` instruction for JIT.

vm_insnhelper.c
k0kubun: :+1: https://github.com/ruby/ruby/pull/8623/commits/ef8d915bd22624cbf214776297e00bdbc668c8d4

maximecb (approved):

PR #8628, opened 2023-10-11
YJIT: port call threshold logic from Rust to C for performance

maximecb:
This way the C compiler can potentially inline the YJIT call threshold logic, and we avoid allocating/touching the ISEQ payload object or needing to take the VM lock to do call threshold accounting.

### DIFF
--- a/tool/rjit/bindgen.rb
+++ b/tool/rjit/bindgen.rb
@@ -637,7 +637,7 @@ generator = BindingGenerator.new(
   skip_fields: {
     'rb_execution_context_struct.machine': %w[regs], # differs between macOS and Linux
     rb_execution_context_struct: %w[method_missing_reason], # non-leading bit fields not supported
-    rb_iseq_constant_body: %w[jit_exception jit_exception_calls yjit_payload], # conditionally defined
+    rb_iseq_constant_body: %w[jit_exception jit_exception_calls yjit_payload yjit_calls_at_interv], # conditionally defined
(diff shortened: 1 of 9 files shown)


k0kubun:
Benchmarked this with `--yjit-cold-threshold=0` and `fib` (since the result of `30k_methods` was unreliable for some reason 🤔). It's looking good.

```
before: ruby 3.3.0dev (2023-10-11T14:16:39Z master cc311e1c45) +YJIT [x86_64-linux]
last_commit=[ruby/irb] Avoid locking the debug UI to a single thread (https://github.com/ruby/irb/pull/725)
after: ruby 3.3.0dev (2023-10-11T21:16:38Z yjit_c_threshold_l.. ba3587b89d) +YJIT [x86_64-linux]
last_commit=Exclude yjit_calls_at_interv from RJIT bindgen

-----  -----------  ----------  ----------  ----------  -------------  ------------
bench  before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
fib    237.9        0.1         226.7       0.2         1.06           1.05
-----  -----------  ----------  ----------  ----------  -------------  ------------
```

k0kubun (approved):

maximecb:
Oh wow that much speed difference! I'm surprised, but seems it was really worth it!

We should expose `rb_yjit_compile_new_iseqs()` as a Rust-side global next 🤔

PR #8629, opened 2023-10-11
M:N thread scheduler for Ractors

ko1:
This patch introduce M:N thread scheduler for Ractor system.

In general, M:N thread scheduler employs N native threads (OS threads) to manage M user-level threads (Ruby threads in this case). On the Ruby interpreter, 1 native thread is provided for 1 Ractor and all Ruby threads are managed by the native thread.

From Ruby 1.9, the interpreter uses 1:1 thread scheduler which means 1 Ruby thread has 1 native thread. M:N scheduler change this strategy.

Because of compatibility issue (and stableness issue of the implementation) main Ractor doesn't use M:N scheduler on default. On the other words, threads on the main Ractor will be managed with 1:1 thread scheduler.

There are additional settings by environment variables:

`RUBY_MN_THREADS=1` enables M:N thread scheduler on the main ractor. Note that non-main ractors use the M:N scheduler without this configuration. With this configuration, single ractor applications run threads on M:1 thread scheduler (green threads, user-level threads).

`RUBY_MAX_CPU=n` specifies maximum number of native threads for M:N scheduler (default: 8).

This patch will be reverted soon if non-easy issues are found.

[Bug #19842]

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -53,6 +53,7 @@
 #include "internal/loadpath.h"
 #include "internal/missing.h"
 #include "internal/object.h"
+#include "internal/thread.h"
 #include "internal/ruby_parser.h"
 #include "internal/variable.h"
 #include "ruby/encoding.h"
@@ -2148,6 +2149,7 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
 #endif
 
     ruby_gc_set_params();
+    ruby_mn_threads_params();
     ruby_init_loadpath();
 
     Init_enc();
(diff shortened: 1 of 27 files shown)


ruby.c
```diff
@@ -1431,6 +1432,7 @@ proc_long_options(ruby_cmdline_options_t *opt, const char *s, long argc, char **
     else if (strcmp("verbose", s) == 0) {
         opt->verbose = 1;
         ruby_verbose = Qtrue;
+        bp();
```
nobu, line 1435:
```suggestion
```

ko1: OMG

jarthod:
@ko1 for the record I just posted this regression which *might* be related to this change: https://bugs.ruby-lang.org/issues/20816 (in case you want to have a look and provide your input). Thanks :bow:

PR #8630, opened 2023-10-11
Fix ObjectSpace.dump with super() callinfo

jhawthorn:
`super()` uses 0 as mid for its callinfo, so we need to check for that to avoid a segfault when using `dump_all`.

Previously this would crash since #8571

``` ruby
require "objspace"
class Foo
  def foo
    super(bar: 123)
  end
end
ObjectSpace.dump_all(output: :string)
```

I also switched to using `dump_append_string_value` even though that's probably not necessary, it removes an assumption about what callinfo mids the parser will generate.

cc @peterzhu2118


peterzhu2118 (approved):

casperisfine:
I was just tracking that same crash 5 minutes ago ❤️

PR #8637, opened 2023-10-12
[Feature #19422] Enable shared by default on macOS

nobu:


misdoro:
can this be tried again since #9284 was merged?

PR #8640, opened 2023-10-12
disable MN scheduler on !`USE_MN_THREADS`

ko1:
!`USE_MN_THREADS` platforms (== non Linux platform now) `enable_mn_threads` should be false.


jeremyevans:
Is there a plan to support `USE_MN_THREADS` on non-Linux platforms?  Which Linux-specific APIs does it currently use?

ko1:
It uses `epoll` and `kqueue` support is needed.

PR #8643, opened 2023-10-12
[PRISM] Add --dump=prism mode

jemmaissroff:
Similar to `--dump=parsetree`, `--dump=prism` will dump the Prism AST for a given file. It is also compatible with `-e <code>` syntax.

I have also created this issue https://github.com/ruby/prism/issues/1681 to make the Prism AST printing nicer in the C API.


kddnewton (approved):

ko1:
How about to rename `prisim_parsetree` or similar to make clear it is parse tree?
It will help documentation.

Maybe you can use `--dump=prisim` for the command line option because shorthand will be accepted.

faraazahmad:
Just curious, If/when Prism becomes the default parser for Ruby, will `--dump=parsetree` and `--dump=prism` return the same output?

kddnewton:
@faraazahmad From my understanding either way they will become the same, because CRuby has committed to prism's node structure (regardless of the internals of the parser itself)

kddnewton:
@ko1 happy to change this to `prism_parsetree`

PR #8647, opened 2023-10-13
YJIT: Fix argument clobbering in some block_arg+rest_param calls

XrXr:
Previously, for block argument callsites with some specific argument count and callee local variable count combinations, YJIT ended up writing over arguments that are supposed to be collected into a rest parameter array unmodified.

Detect when clobbering would happen and avoid it. Also, place the block handler after the stack overflow check, since it writes to new stack space.

Reported-by: Takashi Kokubun


k0kubun (approved):
Thanks!

PR #7429, opened 2023-03-03
Proof of concept: shape iv ancestor search

jhawthorn:
_Opening this as draft because I'm not sure of the idea, but wanted to share it early. I haven't benchmarked realistic code or profiled or optimized yet._

This implements `rb_shape_get_iv_index_with_hint`: a version of rb_shape_get_iv_index which uses a previous shape with a valid index for the given IV in order to speed up searches on a IV inline cache miss.

This works by walking up both the hint shape's and current shape's tree branches in parallel until they meet at a common ancestor. If the common ancestor is deeper than the edge name we're looking for we can skip the remainder of the search. We also check for the edge name as we search.

The thinking behind this is that in many cases we have different, but very similar shapes, who share the common ancestor defining the IV we're looking for. This should also work well for a frozen version of a shape.

This is significantly faster when the divergence between the current and hint shape is shallow, but when it is not (there is no relation between the two shapes) it still resorts to a linear scan. The full scan is slower with this, as more work is being done, but I believe not 2x slower (the LL means we'll be stalling all the time on memory so walking two lists in parallel isn't _so_ bad).

Benchmark: https://gist.github.com/jhawthorn/e9db7a1443eaa957cef772dccdc14d9a

With a shallow tree divergence

before: `ruby --disable-gems test_shape_search.rb  1.64s user 0.00s system 99% cpu 1.648 total`
after: `./miniruby test_shape_search.rb  0.67s user 0.00s system 99% cpu 0.666 total`

With a deep tree divergence (worse, assumed uncommon)

before: `ruby --disable-gems test_shape_search_deep.rb  1.64s user 0.00s system 99% cpu 1.643 total`
after: `./miniruby test_shape_search_deep.rb  1.88s user 0.01s system 99% cpu 1.889 total`

cc @tenderlove @jemmaissroff

PR #7344, opened 2023-02-21
Bump SHAPE_MAX_NUM_IVS to 80

k0kubun:
A quick fix to https://github.com/ruby/ruby/pull/7188#issuecomment-1422087741. I know we'll have a long-term solution shortly, but it wouldn't hurt to normalize yjit-bench sooner.

71 seems to be the minimum number that fixes the immediate problem, but I wanted to use a somewhat round number, while still not encouraging to use an infinite number of ivars.

```
before: ruby 3.3.0dev (2023-02-21T04:46:11Z master 6384f7981c) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-02-21T07:15:39Z shape-num-iv 6dc96a29f7) +YJIT [x86_64-linux]

---------  -----------  ----------  ----------  ----------  ------------  -------------
bench      before (ms)  stddev (%)  after (ms)  stddev (%)  before/after  after 1st itr
optcarrot  3287.2       0.4         1453.8      0.8         2.26          2.22
---------  -----------  ----------  ----------  ----------  ------------  -------------
```

cc: @jemmaissroff @tenderlove @ruby/yjit

PR #8650, opened 2023-10-13
vm_getivar: look for a common shape ancestor on miss

casperisfine:
Ref: https://github.com/ruby/ruby/pull/7429

vm_getivar: assume the cached shape_id like have a common ancestor

When an inline cache misses, it is very likely that the stale shape_id
and the current instance shape_id have a close common ancestor.

For example if the instance variable is sometimes frozen sometimes
not, one of the two shape will be the direct parent of the other.

Another pattern that commonly cause IC misses is "memoization",
in such case the object will have a "base common shape" and then
a number of close descendants.

In addition, when we find a common ancestor, we store it in the
inline cache instead of the current shape. This help prevent the
cache from flip-flopping, ensuring the next lookup will be marginally
faster and more generally avoid writing in memory too much.

However, now that shapes have an ancestors index, we only check
for a few ancestors before falling back to use the index.

So overall this change speeds up what is assumed to be the more common
case, but makes what is assumed to be the less common case a bit slower.

```
compare-ruby: ruby 3.3.0dev (2023-10-26T05:30:17Z master 701ca070b4) [arm64-darwin22]
built-ruby: ruby 3.3.0dev (2023-10-26T09:25:09Z shapes_double_sear.. a723a85235) [arm64-darwin22]
warming up......

|                                     |compare-ruby|built-ruby|
|:------------------------------------|-----------:|---------:|
|vm_ivar_stable_shape                 |     11.672M|   11.679M|
|                                     |           -|     1.00x|
|vm_ivar_memoize_unstable_shape       |      7.551M|   10.506M|
|                                     |           -|     1.39x|
|vm_ivar_memoize_unstable_shape_miss  |     11.591M|   11.624M|
|                                     |           -|     1.00x|
|vm_ivar_unstable_undef               |      9.037M|    7.981M|
|                                     |       1.13x|         -|
|vm_ivar_divergent_shape              |      8.034M|    6.657M|
|                                     |       1.21x|         -|
|vm_ivar_divergent_shape_imbalanced   |     10.471M|    9.231M|
|                                     |       1.13x|         -|
```

### DIFF
--- a/benchmark/vm_ivar_memoize.yml
+++ b/benchmark/vm_ivar_memoize.yml
@@ -1,0 +1,85 @@
+prelude: |
+  IVARS = 60
+  class Record
+    def initialize(offset = false)
+      @offset = 1 if offset
+      @first = 0
+      IVARS.times do |i|
+        instance_variable_set("@ivar_#{i}", i)
+      end
+    end
+
+    def first
+      @first
+    end
+
+    def lazy_set
+      @lazy_set ||= 123
+    end
+
+    def undef
+      @undef
+    end
+  end
+
+  Record.new # Need one alloc to right size
+
+  BASE = Record.new
+  LAZY = Record.new
+  LAZY.lazy_set
+
+  class Miss < Record
+    @first = 0
+    IVARS.times do |i|
+      instance_variable_set("@i_#{i}", i)
+    end
+  end
+
+  Miss.new # Need one alloc to right size
+  MISS = Miss.new
+
+  DIVERGENT = Record.new(true)
+
+benchmark:
+  vm_ivar_stable_shape: |
+    BASE.first
+    BASE.first
+    BASE.first
+    BASE.first
+    BASE.first
+    BASE.first
+  vm_ivar_memoize_unstable_shape: |
+    BASE.first
+    LAZY.first
+    BASE.first
+    LAZY.first
+    BASE.first
+    LAZY.first
+  vm_ivar_memoize_unstable_shape_miss: |
+    BASE.first
+    MISS.first
+    BASE.first
+    MISS.first
+    BASE.first
+    MISS.first
+  vm_ivar_unstable_undef: |
+    BASE.undef
+    LAZY.undef
+    BASE.undef
+    LAZY.undef
+    BASE.undef
+    LAZY.undef
+  vm_ivar_divergent_shape: |
+    BASE.first
+    DIVERGENT.first
+    BASE.first
+    DIVERGENT.first
+    BASE.first
+    DIVERGENT.first
+  vm_ivar_divergent_shape_imbalanced: |
+    BASE.first
+    DIVERGENT.first
+    DIVERGENT.first
+    DIVERGENT.first
+    DIVERGENT.first
+    DIVERGENT.first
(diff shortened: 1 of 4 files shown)


casperisfine:
@jhawthorn @tenderlove I'd love your opinion here.

casperisfine:
Also forgot to note: what does this mean for `setivar` and `YJIT`. I suppose they could both benefit from this logic, but they'd need to be adapted.

casperisfine:
@jhawthorn @tenderlove I reworked the PR based on our recent discussion (see the updated description).

I settled on checking up to 2 ancestors based on the micro-benchmark results, but also on YJIT polymorphic cache size, and more generally on the idea that one or two memoized variables is rather common and should be supported.

If I raise the limit to `3` the hit on fully divergent shapes starts to be a bit too much:

```
|                                     |compare-ruby|built-ruby|
|:------------------------------------|-----------:|---------:|
|vm_ivar_stable_shape                 |     11.737M|   11.632M|
|                                     |       1.01x|         -|
|vm_ivar_memoize_unstable_shape       |      7.570M|   10.375M|
|                                     |           -|     1.37x|
|vm_ivar_memoize_unstable_shape_miss  |     11.809M|   11.687M|
|                                     |       1.01x|         -|
|vm_ivar_unstable_undef               |      8.921M|    7.964M|
|                                     |       1.12x|         -|
|vm_ivar_divergent_shape              |      7.808M|    5.996M|
|                                     |       1.30x|         -|
|vm_ivar_divergent_shape_imbalanced   |     10.318M|    8.933M|
|                                     |       1.16x|         -|
```

k0kubun:
Looking at https://rubybench.github.io/benchmarks/yjit-bench.html#optcarrot, I noticed that the interpreter slowed down a lot on Optcarrot on 11/3, and it was because of this change.

```
before: ruby 3.3.0dev (2023-11-03T11:39:57Z master 0ae6a2a1cf) [x86_64-linux]
after: ruby 3.3.0dev (2023-11-03T11:47:43Z master b92b9e1e9e) [x86_64-linux]

---------  -----------  ----------  ----------  ----------  -------------  ------------
bench      before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
optcarrot  5021.2       0.9         5356.0      0.8         0.94           0.94
---------  -----------  ----------  ----------  ----------  -------------  ------------
```

Not that this benchmark is important, but I wonder if you intended such a slowdown in some scenarios.

byroot:
That's really curious. Do you have any idea as of why? Is there a shape flipping a lot or something?

k0kubun:
I don't, since I only did bisect and haven't done any investigation yet. One thing that might be relevant is this benchmark was the closest benchmark to the SHAPE_MAX_NUM_IVS limit https://github.com/ruby/ruby/pull/7344, which once existed before https://github.com/ruby/ruby/pull/8755. So it's kind of a unique benchmark not only because it uses getivar/setivar a lot but also because it uses so many different ivars.

casperisfine:
> Could it be deoptimizing some class into dictionary mode

No, this patch doesn't have an impact on the actual shape of objects, only on how it looks it up.

But overall I think it must be two very deep shapes flip-floping. This PR has a know tradeoff: we're faster when we miss for a shape that is a close ancestor of the old one, but also slower when we miss for a shape that doesn't have a common ancestor.

That's what `vm_ivar_divergent_shape` shows, and I suspect optcarrot somehow goes into that case?

PR #8651, opened 2023-10-13
use clear filename for btest

ko1:
`make btest` (bootstraptest/) makes tempfile and run the file. This patch introduces two changes:

* use fullpath to make easy to find the file (with `ps`)
* `_<line>_<id>` to `_L<line>_N<id>` (`_123_456` is not easy to find which is line number).


ko1:
It seems that wasm doesn't allow to use fullpath, so close it.

PR #8663, opened 2023-10-14
[Misc #18984] Raise `TypeError` from `Range#size` if the range is not iterable

kyanagi:
https://bugs.ruby-lang.org/issues/18984
https://github.com/ruby/dev-meeting-log/blob/master/2023/DevMeeting-2023-10-12.md

### DIFF
--- a/range.c
+++ b/range.c
@@ -827,7 +827,12 @@ sym_each_i(VALUE v, VALUE arg)
  *    (1..4).size      # => 4
  *    (1...4).size     # => 3
  *    (1..).size       # => Infinity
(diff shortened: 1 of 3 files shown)


jeremyevans:
Apologies that this was not reviewed in a timely fashion.  Could you rebase and update the specs to they apply to Ruby 3.4 instead of 3.3? Also, in the specs, please remove `should ==` inside a `.should raise_error` block.

kyanagi:
@jeremyevans Thank you for your review!
I rebased my commit and added minor fixes.

PR #7641, opened 2023-03-31
Generate ruby_method_ids constants separately

nobu:
The pre-defined ID sequences in id.h are generated automatically at the compile time so that the repository is magic-number free, but currently cruby_ids.inc.rs generated from the header file is committed into the repository.  This is that the dependencies are inverted (or broken).

This PR separates pre-defined ID list from the rust source, and makes an intermidiate tool named as cruby_ids_gen, which generate the rust source.  This depends on that YJIT does not support cross-compilation.

PR #8676, opened 2023-10-16
YJIT: Lookup IDs on boot instead of binding to them

XrXr:
Previously, the version-controlled `cruby_bindings.inc.rs` file
contained the build-time artifact `id.h`, which nobu mentioned hinders
the goal of having fewer magic numbers in the repository.

Lookup the IDs YJIT needs on boot. It costs cycles, but it's fine since
YJIT only uses a handful of IDs at the moment. No perceptible
degradation to boot time found in my testing.

```
$ hyperfine "$HOME/.rubies/yjit-pre/bin/ruby --yjit -e 0" "$HOME/.rubies/yjit-post/bin/ruby --yjit -e 0"
Benchmark 1: /Users/alan/.rubies/yjit-pre/bin/ruby --yjit -e 0
  Time (mean ± σ):      37.8 ms ±   1.4 ms    [User: 27.0 ms, System: 7.0 ms]
  Range (min … max):    35.2 ms …  41.2 ms    69 runs
 
Benchmark 2: /Users/alan/.rubies/yjit-post/bin/ruby --yjit -e 0
  Time (mean ± σ):      37.4 ms ±   1.2 ms    [User: 26.5 ms, System: 6.7 ms]
  Range (min … max):    34.9 ms …  39.8 ms    67 runs
 
Summary
  /Users/alan/.rubies/yjit-post/bin/ruby --yjit -e 0 ran
    1.01 ± 0.05 times faster than /Users/alan/.rubies/yjit-pre/bin/ruby --yjit -e 0
```

Resolves: #7641


yjit/src/cruby.rs
```diff
@@ -735,3 +735,42 @@ mod manual_defs {
...
+        name: respond_to_missing content: b"respond_to_missing?"
+    }
+}
+
+macro_rules! rb_id {
+    ($id_name:ident) => {
+        $crate::cruby::ids::$id_name.load(std::sync::atomic::Ordering::Relaxed)
+    }
+}
+pub(crate) use rb_id;
```
maximecb, line 776: Could you add a comment explaining what this macro does?

yjit/src/cruby.rs
```diff
@@ -735,3 +735,42 @@ mod manual_defs {
...
+
+                    $ident.store(
+                        unsafe { $crate::cruby::rb_intern2(ptr.cast(), content.len() as _) },
+                        std::sync::atomic::Ordering::Relaxed
+                    );
+                )*
+
+            }
+        }
+    }
```
maximecb, line 760: Could you add some comments with explanations? Why the atomic u64? I'm assuming you're looking up the id and caching it?

yjit/src/cruby.rs
```diff
@@ -735,3 +735,42 @@ mod manual_defs {
...
+    def_ids! {
+        name: NULL               content: b""
+        name: min                content: b"min"
+        name: max                content: b"max"
+        name: hash               content: b"hash"
+        name: respond_to_missing content: b"respond_to_missing?"
+    }
+}
+
+macro_rules! rb_id {
```
k0kubun, line 771: WDYT about calling it `ID!`? Would that be too confusing? `rb_id` feels like a C API of CRuby, but it doesn't exist in the interpreter code. 

I personally feel it's easier to understand it's an `ID` when it's `ID!(min)` instead of `rb_id!(min)`.

maximecb: Maybe `RUBY_ID!()` ? Just because ID is kind of a generic concept.

k0kubun: If we want to have a prefix, I would prefer `RB_` over `RUBY_` since `RUBY_` seems too long compared to the main `ID`, `RB_` prefix is also common enough. So, I'd vote for `RB_ID!()` (it still seems better than `rb_id!()` as to the point I made above).

XrXr: I like `ID!` too. The macro being nested under `cruby` probably makes it specific enough, and LSP hover shows the full path for anyone who wants to dig. It matching the C type name feels nice.

maximecb (approved):

nobu (approved):
Wish C had richer macro capabilities too.

k0kubun (approved):

PR #8680, opened 2023-10-16
Remove compile context

jemmaissroff:
Remove pm_compile_context_t, move the context onto ScopeNode
    
    We changed ScopeNodes to point to their parent (previous) ScopeNodes.
    Accordingly, we can remove pm_compile_context_t, and store all
    necessary context in ScopeNodes, allowing us to access locals from
    outer scopes.

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -939,10 +939,10 @@ rb_iseq_new_with_opt(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE rea
     return iseq_translate(iseq);
 }
 
-VALUE rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t scope_node, pm_parser_t *parser);
+VALUE rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t *scope_node, pm_parser_t *parser);
 
 rb_iseq_t *
-pm_iseq_new_with_opt(pm_scope_node_t scope_node, pm_parser_t *parser, VALUE name, VALUE path, VALUE realpath,
+pm_iseq_new_with_opt(pm_scope_node_t *scope_node, pm_parser_t *parser, VALUE name, VALUE path, VALUE realpath,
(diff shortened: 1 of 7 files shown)


iseq.c
```diff
@@ -939,10 +939,10 @@ rb_iseq_new_with_opt(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE rea
     return iseq_translate(iseq);
 }
 
-VALUE rb_iseq_compile_prism_node(rb_iseq_t * iseq, const pm_node_t *node, pm_parser_t *parser);
+VALUE rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t scope_node, pm_parser_t *parser);
 
 rb_iseq_t *
-pm_iseq_new_with_opt(pm_node_t *node, pm_parser_t *parser, VALUE name, VALUE path, VALUE realpath,
+pm_iseq_new_with_opt(pm_scope_node_t scope_node, pm_parser_t *parser, VALUE name, VALUE path, VALUE realpath,
```
tenderlove, line 945: Same question as above

tenderlove:
@jemmaissroff I pushed a commit to your branch that I think will fix the build.  Kokubun and I were able to reproduce the problem like this:

```
$ make test-all TESTOPTS='--seed=2026' TESTS='test/ruby/test_defined.rb test/ruby/test_compile_prism.rb'
```

Basically what was happening is that the new test introduced in this PR would happen to [define the method foo](https://github.com/ruby/ruby/pull/8680/files#diff-0ac5c87a2c444663081de383e3c3c1459461e8fb9eb4d059481a1fac2154de27R513), and other tests [expected foo to *not* be defined](https://github.com/jemmaissroff/ruby/blob/05829ca29a8edd9874aab08cce83ed3b4a3f3548/test/ruby/test_defined.rb#L33-L37).

I changed the test case to use `assert_separately`.  Unfortunately I had to copy/paste some code to make it work.  We should probably consider moving those methods to a helper file that `assert_separately` can require, or writing a method that returns the source code so that we can eval it.

PR #8688, opened 2023-10-17
YJIT: Return Option from asm.compile() for has_dropped_bytes()

XrXr:
So that we get a reminder to check CodeBlock::has_dropped_bytes().
Internally, asm.compile() already checks it, and this patch just
propagates it out to the caller with a `#[must_use]`.

### DIFF
--- a/yjit/src/backend/arm64/mod.rs
+++ b/yjit/src/backend/arm64/mod.rs
@@ -188,6 +188,12 @@ fn emit_load_value(cb: &mut CodeBlock, rd: A64Opnd, value: u64) -> usize {
 /// These are caller-saved registers.
 pub static TEMP_REGS: [Reg; 5] = [X1_REG, X9_REG, X10_REG, X14_REG, X15_REG];
 
+#[derive(Debug, PartialEq)]
+enum EmitError {
+    RetryOnNextPage,
+    OutOfMemory,
+}
+
 impl Assembler
 {
     // Special scratch registers for intermediate processing.
@@ -693,7 +699,7 @@ impl Assembler
 
     /// Emit platform-specific machine code
     /// Returns a list of GC offsets. Can return failure to signal caller to retry.
-    fn arm64_emit(&mut self, cb: &mut CodeBlock, ocb: &mut Option<&mut OutlinedCb>) -> Result<Vec<u32>, ()> {
+    fn arm64_emit(&mut self, cb: &mut CodeBlock, ocb: &mut Option<&mut OutlinedCb>) -> Result<Vec<u32>, EmitError> {
         /// Determine how many instructions it will take to represent moving
         /// this value into a register. Note that the return value of this
         /// function must correspond to how many instructions are used to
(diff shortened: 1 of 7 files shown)


yjit/src/codegen.rs
```diff
@@ -547,7 +546,7 @@ pub fn gen_counted_exit(side_exit: CodePtr, ocb: &mut OutlinedCb, counter: Optio
 
     // Jump to the existing side exit
     asm.jmp(Target::CodePtr(side_exit));
-    asm.compile(ocb, None);
+    let _ = asm.compile(ocb, None); // Caller checks cb.has_dropped_bytes()
```
k0kubun, line 549: I'm actually not sure if this is safe. For example, `jit_ensure_block_entry_exit` -> `get_side_exit` -> `gen_counted_exit` doesn't seem to check `has_dropped_bytes()`.

Can't we just propagate `Option<CodePtr>` to the callers (i.e. this function should also return `Option<CodePtr>`)? It seems like the point of this PR.

Also, I thought you were going to let `asm.compile()` return `CodePtr` as well. If it returns `Option<(CodePtr, Vec<u32>)>`, you would map it to `.0` of it, always respecting the `Option` state instead of discarding it with `let _ =`. Using `ocb.get_write_ptr()` seems prone to bugs compared to always using an `Option`al return value.

XrXr: Sorry, I definitely cut some corners to keep the diff small. The cases where we ignore the error are probably fine because they're caught by checks at the end of `gen_single_block()`, but propagating the error is definitely better.

I'll go work on a v2.

k0kubun: I appreciate you trying to minimize the diff. That makes reviews easier. I'd be comfortable approving the current diff and waiting for another PR on top of it too. I just wasn't sure if you want to fix it since having the `// Caller checks cb.has_dropped_bytes()` comment (instead of a TODO comment) seemed to me like you're saying "this is good". 

If the number of lines wouldn't change that much, it seems fine to do that in this PR too.

yjit/src/core.rs
```diff
@@ -2298,19 +2298,35 @@ c_callable! {
...
+        with_compile_time(|| {
+            with_vm_lock(src_loc!(), || {
+                let cb = CodegenGlobals::get_inline_cb();
+                let ocb = CodegenGlobals::get_outlined_cb();
+
+                let addr = entry_stub_hit_body(entry_ptr, ec, cb, ocb)
+                    .unwrap_or_else(|| {
+                        // Trigger code GC (e.g. no space).
+                        // This entry point will be recompiled later.
+                        cb.code_gc(ocb);
```
k0kubun, line 2310: We could hit this in undesirable conditions, e.g. `stack_size` does not fit in `u8`. Can we at least check `has_dropped_bytes()` here?

yjit/src/backend/arm64/mod.rs
```diff
@@ -188,6 +188,12 @@ fn emit_load_value(cb: &mut CodeBlock, rd: A64Opnd, value: u64) -> usize {
 /// These are caller-saved registers.
 pub static TEMP_REGS: [Reg; 5] = [X1_REG, X9_REG, X10_REG, X14_REG, X15_REG];
 
+#[derive(Debug, PartialEq)]
+enum EmitError {
+    RetryOnNextPage,
+    OutOfMem,
```
k0kubun, line 194: nits: How about just `OutOfMemory`? Generally it's easier to understand a name if you don't abbreviate it, and this one would be still short enough (at least shorter than `RetryOnNextPage`).

k0kubun (approved):
Looking much better :+1:

yjit/src/core.rs
XrXr: We probably should, and for all the other code gc trigger points too. But I think it should go with the `--yjit-disable-code-gc` PR.

PR #8689, opened 2023-10-17
YJIT: Use RbConfig.ruby instead of EnvUtil.rubybin

k0kubun:
`EnvUtil.invoke_ruby` uses `EnvUtil.rubybin` by default, which has [a weird fallback logic](https://github.com/ruby/ruby/blob/2abd061e8beefbdead0296377948ce7a0098277b/tool/lib/envutil.rb#L23-L31) that is not necessary for test_yjit.

Since there are folks reporting [test_yjit failures we can't reproduce](https://bugs.ruby-lang.org/issues/19921), I want to rule out that possibility by always using `RbConfig.ruby`.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1519,6 +1519,6 @@ class TestYJIT < Test::Unit::TestCase
   # A wrapper of EnvUtil.invoke_ruby that uses RbConfig.ruby instead of EnvUtil.ruby
   # that might use a wrong Ruby depending on your environment.
   def invoke_ruby(*args, **kwargs)
-    EnvUtil.invoke_ruby(*args, **{ rubybin: RbConfig.ruby, **kwargs })
+    EnvUtil.invoke_ruby(*args, rubybin: RbConfig.ruby, **kwargs)
   end
 end
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -75,7 +75,7 @@ class TestYJIT < Test::Unit::TestCase
   end
 
   def test_yjit_stats_and_v_no_error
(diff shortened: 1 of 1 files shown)


test/ruby/test_yjit.rb
```diff
@@ -1517,4 +1515,10 @@ def eval_with_jit(script, call_threshold: 1, timeout: 1000, mem_size: nil)
     stats_r&.close
     stats_w&.close
   end
+
+  # A wrapper of EnvUtil.invoke_ruby that uses RbConfig.ruby instead of EnvUtil.ruby
+  # that might use a wrong Ruby depending on your environment.
+  def invoke_ruby(*args, **kwargs)
+    EnvUtil.invoke_ruby(*args, **{ rubybin: RbConfig.ruby, **kwargs })
```
XrXr, line 1522:
```suggestion
    EnvUtil.invoke_ruby(*args, rubybin: RbConfig.ruby, **kwargs)
```
Unless it was intentional?

XrXr (approved):
Sounds good to me. By the way, I dug a little bit after looking at the logic in `EnvUtil.invoke_ruby` and it was actually using `RUBY=build_dir/exe/ruby` from the environment set by `runruby.rb`, and `make test-all` only builds `ruby`, but doesn't copy it over to `exe/ruby`.

test/ruby/test_yjit.rb
k0kubun: I was initially thinking of using `**{ rubybin: RbConfig.ruby }.merge!(kwargs)` so that `rubybin` can be safely overridden. But then I ended up rewriting that to this for simplifying the code, and now it prints a warning if `rubybin` is provided, just like yours. We shouldn't pass `rubybin` there anyway, so it seems fine to use either of these. Applied your suggestion https://github.com/ruby/ruby/pull/8689/commits/1256de9e12b60c754025cb6118854f5309516488.

PR #8695, opened 2023-10-18
Make the name resolution interruptible

mame:
When pthread_create is available, rb_getaddrinfo creates a pthread and
executes getaddrinfo(3) in it. The caller thread waits for the pthread
to complete, but detaches it if interrupted. This allows name resolution
to be interuppted by Timeout.timeout, etc. even if it takes a long time
(for example, when the DNS server does not respond).

### DIFF
--- a/ext/socket/raddrinfo.c
+++ b/ext/socket/raddrinfo.c
@@ -472,13 +472,13 @@ start:
 
     arg = allocate_getaddrinfo_arg(hostp, portp, hints);
     if (!arg) {
-        return ENOMEM;
+        return EAI_MEMORY;
     }
 
     pthread_t th;
     if (pthread_create(&th, 0, do_getaddrinfo, arg) != 0) {
         free_getaddrinfo_arg(arg);
-        return EAGAIN;
+        return EAI_AGAIN;
     }
 
     pthread_detach(th);
@@ -499,7 +499,7 @@ start:
             if (err == 0) *ai = arg->ai;
         }
         else if (arg->cancelled) {
-            err = EAGAIN;
+            err = EAI_AGAIN;
         }
         else {
             // If already interrupted, rb_thread_call_without_gvl2 may return without calling wait_getaddrinfo.
@@ -691,13 +691,13 @@ start:
 
     arg = allocate_getnameinfo_arg(sa, salen, hostlen, servlen, flags);
     if (!arg) {
-        return ENOMEM;
+        return EAI_MEMORY;
     }
 
     pthread_t th;
     if (pthread_create(&th, 0, do_getnameinfo, arg) != 0) {
         free_getnameinfo_arg(arg);
-        return EAGAIN;
+        return EAI_AGAIN;
     }
 
     pthread_detach(th);
@@ -720,7 +720,7 @@ start:
         }
     }
     else if (arg->cancelled) {
-        err = EAGAIN;
+        err = EAI_AGAIN;
     }
     else {
         // If already interrupted, rb_thread_call_without_gvl2 may return without calling wait_getnameinfo.
--- a/ext/socket/raddrinfo.c
+++ b/ext/socket/raddrinfo.c
@@ -10,6 +10,17 @@
 
 #include "rubysocket.h"
 
+// GETADDRINFO_IMPL == 0 : call getaddrinfo/getnameinfo directly
+// GETADDRINFO_IMPL == 1 : call getaddrinfo/getnameinfo without gvl (but uncancellable)
+
+#ifndef GETADDRINFO_IMPL
+#  ifdef GETADDRINFO_EMU
+#    define GETADDRINFO_IMPL 0
+#  else
+#    define GETADDRINFO_IMPL 1
+#  endif
+#endif
+
 #if defined(INET6) && (defined(LOOKUP_ORDER_HACK_INET) || defined(LOOKUP_ORDER_HACK_INET6))
 #define LOOKUP_ORDERS (sizeof(lookup_order_table) / sizeof(lookup_order_table[0]))
 static const int lookup_order_table[] = {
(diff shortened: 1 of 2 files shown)


ioquatix:
Is it guaranteed to be safe to interrupt a thread which is stuck in `getaddrinfo`?

mame:
https://bugs.ruby-lang.org/issues/19965

mame:
> Is it guaranteed to be safe to interrupt a thread which is stuck in `getaddrinfo`?

My patch does not interrupt a thread which is stuck in `getaddrinfo(3)`. See my ticket.

ext/socket/raddrinfo.c
```diff
@@ -10,6 +10,22 @@
...
 
+// GETADDRINFO_IMPL == 0 : call getaddrinfo/getnameinfo directly
+// GETADDRINFO_IMPL == 1 : call getaddrinfo/getnameinfo without gvl (but uncancellable)
+// GETADDRINFO_IMPL == 2 : call getaddrinfo/getnameinfo in a dedicated pthread
+//                         (and if the call is interrupted, the pthread is detached)
+
+#ifndef GETADDRINFO_IMPL
+#  ifdef GETADDRINFO_EMU
+#    define GETADDRINFO_IMPL 0
+#  elif !defined(HAVE_PTHREAD_CREATE) || !defined(HAVE_PTHREAD_DETACH) || defined(__MINGW32__) || defined(__MINGW64__)
```
ko1, line 21: if `!defined(HAVE_PTHREAD_CREATE) || !defined(HAVE_PTHREAD_DETACH)` is given, both functions are available.
You can also use `#define HAVE_PTHREAD_H 1` in `.ext/include/x86_64-linux/ruby/config.h` (arch).

ext/socket/raddrinfo.c
```diff
@@ -302,7 +292,243 @@ rb_freeaddrinfo(struct rb_addrinfo *ai)
...
+
+    arg = allocate_getaddrinfo_arg(hostp, portp, hints);
+    if (!arg) {
+        return ENOMEM;
+    }
+
+    pthread_t th;
+    if (pthread_create(&th, 0, do_getaddrinfo, arg) != 0) {
+        free_getaddrinfo_arg(arg);
+        return EAGAIN;
```
ko1, line 477: EAGAIN?

ext/socket/raddrinfo.c
mame: I was worried about environments that only provide one of the two, or environments where pthread.h are available but not actually usable. So I thought this would be a safe bet. I may try it after Ruby CI passes.

ext/socket/raddrinfo.c
mame: Good catch, it should return `EAI_AGAIN` as per `getaddrinfo(3)`. I have changed it.

mame:
I talked to @ko1 and he said it looked good.

ext/socket/raddrinfo.c
ko1: I'm not sure about `EAI_AGAIN`, but I feel it is like `EAI_MEMORY` (resource is not enough to do it).

XrXr (commented):
Because glibc getaddrinfo() calls getenv() without any synchronization while running without GVL, it could race with Ruby `ENV` writers that setenv() and crash. `ENV` write methods use a lock, but the reads inside getaddrinfo() do not.

There is a recent issue stemming from the same underlying issue from Go:
 - https://rachelbythebay.com/w/2023/10/16/env/
 - https://github.com/golang/go/issues/63567
 - https://rachelbythebay.com/w/2017/01/30/env/

The race is also apparent if you combine `man 7 attributes` to interpret what `MT-Safe env locale` means in `man 3 getaddrinfo`, but maybe only with the benefit of hindsight.

Sorry for pointing out an issue without offer a potential solution :/
Even without this new feature the implementation already releases the GVL, so maybe it won't be more problematic than it already is currently.

ioquatix:
On the blog post I read:

> you are not allowed to setenv in a multi-threaded program

I'm not sure how official this is, but maybe we can prevent `ENV[...] = ...` when multiple threads exist, i.e. still allow it at start-up, but prevent it later to avoid this entire class of issues.

mame:
@XrXr Thanks. @shyouhei also shared the news with me.

> Even without this new feature the implementation already releases the GVL, so maybe it won't be more problematic than it already is currently.

I agree with this. Currently, users must not change `setenv` and `locale` during `getaddrinfo`. We might want to document it, but I guess we need to check functions other than `getaddrinfo` that are conditionally unsafe.

@ioquatix

> maybe we can prevent `ENV[...] = ...` when multiple threads exist

It might be theoretically correct, but practically, incompatibility could be a problem. I think we should carefully consider it separately.

mame:
I'd like to merge it and see what happens with CI.

ioquatix:
@mame in some "worst case" can this in theory create a large (potentially unbounded, or bounded to the number of requests) number of threads?

mame:
Yes. A typical case where a large number of pthreads will be created is to repeatedly interrupt name resolution.

```
loop do
  begin
    Timeout.timeout(2) do
      Addrinfo.getaddrinfo("unsolvabledomain")
    end
  rescue Timeout::Error
  end
end
```

The timeout of `getaddrinfo(3)` depends on the system configuration. On my Ubuntu, it seems to be 10 seconds. So, if `Timeout.timeout(2)` repeatedly interrupts `Addrinfo.getaddrinfo`, 10 / 2 = 5 pthreads will be stay.

ioquatix:
That seems reasonable to me.

flux-benj:
👋 Possible to backport this into earlier ruby versions? We're running 3.0.x in prod and this looks like it would solve a nasty issue for us with forking and tracing.

jjb:
i tried 

```bash
ruby -rsocket -e 'Addrinfo.getaddrinfo("www.ruby-lang.org", 80)'
```

with ruby 3.3 and a bad dns server, and i still could not interrupt it- is that expected?

PR #8697, opened 2023-10-18
YJIT: Add --yjit-perf

k0kubun:
This PR puts together Alan's [yjit-linux-perf-c-cf-push-self-time](https://github.com/XrXr/ruby/tree/yjit-linux-perf-c-cf-push-self-time) and [yjit-x64-frame-pointer](https://github.com/XrXr/ruby/tree/yjit-x64-frame-pointer) branches, changes the symbols to method names, and wraps them with a `--yjit-perf` option.

`--yjit-perf` enables the following two features:

* frame_pointer: It sets up RBP register on x86_64. (Arm64 sets up X29 by default)
* perf_map: Writes address ranges for each method on `/tmp/perf-{pid}.map`, which Linux perf

Each of them has an undocumented suboption `--yjit-perf=fp` and `--yjit-perf=map` to enable only part of them, but it's for YJIT developers and users are supposed to use `--yjit-perf`.

This should enable profiling JIT-ed code with Linux perf, including caller frames.

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -173,6 +173,7 @@ compiled, lower values mean less code is compiled (default 200000)
 - `--yjit-trace-exits`: produce a Marshal dump of backtraces from specific exits. Automatically enables `--yjit-stats`
 - `--yjit-max-versions=N`: maximum number of versions to generate per basic block (default 4)
 - `--yjit-greedy-versioning`: greedy versioning mode (disabled by default, may increase code size)
+- `--yjit-perf`: Enable frame pointers and perf profiling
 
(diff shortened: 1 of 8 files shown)


maximecb (approved):

maximecb:
One minor thing that could be useful: could you briefly document how to run YJIT with perf and produce useful output? I know I might want to look this up later.

k0kubun:
:+1: done https://github.com/ruby/ruby/pull/8697/commits/0156d64f30c867635ba0684e68b279711c290f7a

Here's an example output produced by the documented usage: https://share.firefox.dev/3rTXJxr

maximecb:
Good job 👍

PR #8700, opened 2023-10-18
Revert "Make attr_index_t uint8_t"

octokatherine:
We ran into a bug at GitHub in a module with hundreds of ivars. We traced the bug back to this commit: https://github.com/ruby/ruby/commit/e3afc212ec059525fe4e5387b2a3be920ffe0f0e. The assumption about maximum ivars in https://github.com/ruby/ruby/pull/8618 didn't seem to hold for Modules or object like Hashes.

### DIFF
--- a/rjit_c.rb
+++ b/rjit_c.rb
@@ -898,7 +898,7 @@ module RubyVM::RJIT # :nodoc: all
   end
 
   def C.attr_index_t
-    @attr_index_t ||= CType::Immediate.parse("uint8_t")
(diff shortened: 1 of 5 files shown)


jhawthorn:
cc @byroot @tenderlove at least currently afaict exivar and modules/classes don't go to TOO_COMPLEX after 80 ivars so we ran into issues there. Regression test provided.

byroot (approved):

adnanpolice:
Karachi Pakistan

PR #8708, opened 2023-10-19
[Bug #18286] Relax IBF platform dependencies

nobu:
https://bugs.ruby-lang.org/issues/18286


ccaviness:
Though I've only done basic testing, this seems to allow an Intel Mac to build a universal `ruby` successfully, that works on both Intel and Apple Silicon.

PR #8710, opened 2023-10-19
[PRISM] Remove foo method name from test

jemmaissroff:


test/ruby/test_compile_prism.rb
```diff
@@ -488,30 +488,8 @@ def test_PinnedVariableNode
...
-        compare_eval("#{source}; 1")
-      ensure
-        $VERBOSE = verbose_bak
-      end
-    end
-        test_prism_eval("a = 1; tap do; { a: }; end")
-        test_prism_eval("a = 1; def foo(a); a; end")
-                        end;
+      test_prism_eval("a = 1; tap do; { a: }; end")
+      test_prism_eval("a = 1; def prism_test_scope_node(a); a; end")
```
nobu, line 492: Doesn’t this define the global method?

I tried a singleton method but this still seems to define it in the top-level.

```suggestion
      test_prism_eval("a = Object; def a.prism_test_scope_node(a); a; end")
```

PR #8711, opened 2023-10-19
`test_yjit.rb` improvements

XrXr:
Please see each commit. Would prefer merge-with-rebase (or let me merge)
- YJIT: Print exit reasons on failure in test_yjit.rb
- YJIT: Make test_yjit.rb faster with --yjit-stats=quiet

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1501,7 +1501,7 @@ class TestYJIT < Test::Unit::TestCase
     args = [
       "--disable-gems",
       "--yjit-call-threshold=#{call_threshold}",
-      "--yjit-stats"
+      "--yjit-stats=quiet"
     ]
     args << "--yjit-exec-mem-size=#{mem_size}" if mem_size
     args << "-e" << script_shell_encode(script)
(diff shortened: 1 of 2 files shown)


test/ruby/test_yjit.rb
```diff
@@ -1461,8 +1463,15 @@ def collect_insns(iseq)
...
         exits != recorded_exits &&
         (exits.keys != recorded_exits.keys || !exits.all? { |k, v| v === recorded_exits[k] }) # triple-equal checks range membership or integer equality
-        flunk "Expected #{exits.empty? ? "no" : exits.inspect} exits" \
-          ", but got\n#{recorded_exits.inspect}"
+        stats_reasons = StringIO.new
+        ::RubyVM::YJIT.send(:_print_stats_reasons, runtime_stats, stats_reasons)
+        stats_reasons = stats_reasons.string
+        flunk <<~EOM
+          Expected #{exits.empty? ? "no" : exits.inspect} exits ", but got
+          #{recorded_exits.inspect}"
```
k0kubun, line 1471:
```suggestion
          Expected #{exits.empty? ? "no" : exits.inspect} exits, but got
          #{recorded_exits.inspect}
```

Are these quotes necessary? 🤔

k0kubun (approved):
> I suspect the test is failing because of a timing related interrupt, which on paper could happen slow-enough GC runs.

Ughh nice guess. This seems useful for investigating that.

maximecb (approved):

PR #8714, opened 2023-10-19
[Doc] Add documentation for `IO#cursor`

MatheusRich:
There are a few other methods here that need docs. I'll open one (or several) PRs for those if this one is accepted.

### DIFF
--- a/ext/io/console/console.c
+++ b/ext/io/console/console.c
@@ -1023,6 +1023,16 @@ console_goto(VALUE io, VALUE y, VALUE x)
     return io;
 }
 
+/*
+ * call-seq:
+ *   io.cursor -> [row, column]
+ *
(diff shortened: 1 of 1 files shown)


ext/io/console/console.c
```diff
@@ -1023,6 +1023,16 @@ console_goto(VALUE io, VALUE y, VALUE x)
     return io;
 }
 
+/*
+ * call-seq:
```
MatheusRich, line 1027: I'm not sure how to document methods inside `#if defined _WIN32`. Do I need to duplicate it or is there an alternative way of doing it?

nobu:
Could you send this to the [upstream](https://github.com/ruby/io-console)?

MatheusRich:
@nobu [done](https://github.com/ruby/io-console/pull/50)!

PR #8722, opened 2023-10-20
[Prism] Compile ForNode

eightbitraptor:
Fixes https://github.com/ruby/prism/issues/1648

This currently supports ForNodes with a single instance variable. Multiple assigments aren't fully implemented yet.

eg.

```
for i, j in {a: 'b'} do
  puts i
  puts j
end
```

will result in a stack underflow. This is dependant on the implementation of `MultiTargetNode` (https://github.com/ruby/prism/issues/1646)

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -968,7 +968,7 @@ rb_iseq_compile_node(rb_iseq_t *iseq, const NODE *node)
     return iseq_setup(iseq, ret);
 }
 
-static VALUE rb_translate_prism(rb_iseq_t *iseq, const pm_scope_node_t *scope_node, LINK_ANCHOR *const ret);
+static VALUE rb_translate_prism(pm_parser_t *parser, rb_iseq_t *iseq, pm_scope_node_t *scope_node, LINK_ANCHOR *const ret);
 
 VALUE
 rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t *scope_node, pm_parser_t *parser)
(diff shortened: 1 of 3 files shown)


prism_compile.c
```diff
@@ -1453,6 +1455,48 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
         }
         return;
       }
+      case PM_FOR_NODE: {
+          pm_for_node_t *for_node = (pm_for_node_t *)node;
```
nobu, line 1459: This block is not aligned as the others.

compile.c
```diff
@@ -984,8 +984,14 @@ rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t *scope_node, pm_par
         constants[index] = rb_intern3((const char *) constant->start, constant->length, encoding);
     }
 
-    scope_node->constants = (void *)constants;
+    st_table *index_lookup_table = st_init_numtable();
+    pm_constant_id_list_t *locals = &((pm_scope_node_t *) scope_node)->locals;
```
jemmaissroff, line 988: Is the `(pm_scope_node_t *)` cast necessary?

jemmaissroff (changes requested):

compile.c
```diff
@@ -984,8 +984,14 @@ rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t *scope_node, pm_par
         constants[index] = rb_intern3((const char *) constant->start, constant->length, encoding);
     }
 
-    scope_node->constants = (void *)constants;
+    st_table *index_lookup_table = st_init_numtable();
+    pm_constant_id_list_t *locals = &scope_node->locals;
+    for (size_t i = 0; i < locals->size; i++) {
+        st_insert(index_lookup_table, locals->ids[i], i);
+    }
```
nobu, line 991: Isn't this prism internal?
If so, shouldn't be done in `rb_translate_prism`?

jemmaissroff (approved):

kddnewton (changes requested):
What an achievement! Great work! One small comment

prism_compile.c
```diff
@@ -48,6 +48,8 @@
 #define PM_SWAP_UNLESS_POPPED \
     if (!popped) PM_SWAP;
 
+#define TEMP_CONSTANT_IDENTIFIER ((pm_constant_id_t)(1 << 31))
```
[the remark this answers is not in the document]
eightbitraptor: Sounds good. I should probably think of a better name too!

kddnewton (approved):

PR #8723, opened 2023-10-20
[PRISM] Fix compilation for IfNode, UnlessNode

jemmaissroff:
This properly implements the branch condition for FlipFlopNodes on If / UnlessNodes, and also fixes the bug in UnlessNodes


kddnewton (approved):
This looks good.

There's one other thing here - I believe the getspecial instruction requires an index for the flip flop. At the moment it's always 0, so I think this will fail if you have 2 flip-flops in the same context. Up to you if you want to address this in this PR or in a subsequent one.

PR #8724, opened 2023-10-20
[PRISM] Add several tests

jemmaissroff:


test/ruby/test_compile_prism.rb
```diff
@@ -56,8 +56,10 @@ def test_SourceFileNode
     end
 
     def test_SourceLineNode
-      # TODO:
-      # test_prism_eval("__LINE__")
+      ruby_eval = RubyVM::InstructionSequence.compile_prism("__LINE__").eval
```
kddnewton, line 59: I think this should probably be `compile`

kddnewton (changes requested):

kddnewton (approved):

PR #8726, opened 2023-10-20
YJIT: On test_bug_19316, only check the result

XrXr:
Because the `&` call checks for interrupts, the test was accidentally
timing dependent. Stop checking for exits.

[Bug #19921]
https://bugs.ruby-lang.org/issues/19921#note-16


k0kubun (approved):

maximecb (approved):
👌 good work tracking this down :)

PR #8728, opened 2023-10-21
Raise TypeError for bad IO::Buffer.map argument

headius:
The previous behavior here was to simply extract from an IO or else blindly call #fileno. An object that was not IO and did not implement #fileno would cause NoMethodError, a bit confusing. This PR changes it to use a checked funcall and the eventual error is a TypeError.

cc @ioquatix

### DIFF
--- a/io.c
+++ b/io.c
@@ -2871,7 +2871,7 @@ rb_io_descriptor(VALUE io)
     }
     else {
         VALUE fileno = rb_check_funcall(io, id_fileno, 0, NULL);
-        if (fileno != Qundef) {
+        if (!UNDEF_P(fileno)) {
             return RB_NUM2INT(fileno);
         }
     }
--- a/io.c
+++ b/io.c
@@ -2878,7 +2878,7 @@ rb_io_descriptor(VALUE io)
 
     rb_raise(rb_eTypeError, "expected IO or #fileno, %"PRIsVALUE" given", rb_obj_class(io));
 
-    return -1;
+    UNREACHABLE_RETURN(-1);
 }
 
 int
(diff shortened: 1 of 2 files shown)


ioquatix (approved):
Looks good to me. For a little bit of extra context, this issue was identified because we were implementing the fiber scheduler on JRuby and the conversion to an IO object is different, so the test:

```ruby
  def test_file_mapped_invalid
    assert_raise NoMethodError do
      IO::Buffer.map("foobar")
    end
  end
```

was failing.

io.c
```diff
@@ -2870,8 +2870,15 @@ rb_io_descriptor(VALUE io)
         return fptr->fd;
     }
     else {
-        return RB_NUM2INT(rb_funcall(io, id_fileno, 0));
+        VALUE fileno = rb_check_funcall(io, id_fileno, 0, NULL);
+        if (fileno != Qundef) {
```
nobu, line 2874:
```suggestion
        if (!UNDEF_P(fileno)) {
```

io.c
```diff
@@ -2870,8 +2870,15 @@ rb_io_descriptor(VALUE io)
...
+            return RB_NUM2INT(fileno);
+        }
     }
+
+    rb_raise(rb_eTypeError, "expected IO or #fileno, %"PRIsVALUE" given", rb_obj_class(io));
+
+    return -1;
```
nobu, line 2881:
```suggestion
    UNREACHABLE_RETURN(-1);
```

Or invert the condition of check for `fileno`.

nobu (approved):

PR #8729, opened 2023-10-21
Extract IO::Buffer.for string locking test

headius:
String locking with locktmp is not really part of the public API, and the test relies in a side effect of using it to protect the buffer. On other implementations without locktmp this does not fail. Separate into its own test so it can be excluded from public API expectations.

It might be worth discussing what parts of the internal String lock are actually public, specified behavior.


ioquatix (approved):
LGTM. For the future, we should consider how we specify this in RubySpec.

PR #8734, opened 2023-10-21
Add an optimized instruction for Array#pack

tenderlove:
This commit adds a rule to the peephole optimizer to emit a special instruction when it sees code like this:

```ruby
[a = 1].pack("C")
```

The iseqs for the above code are this:

```
0000 putobject_INT2FIX_1_                                             (   1)[Li]
0001 dup
0002 setlocal_WC_0                          a@0
0004 putstring                              "C"
0006 opt_newarray_send                      2, :pack
0009 leave
```

The new instruction is the `opt_newarray_send`.  The instruction isn't really new, we're just adding a different parameter `:pack`.

Since Array#pack is defined in Ruby I had to rearrange some of the boot code in order to support BOP invalidation for code that is defined in Ruby.


nobu:
FYI: https://github.com/nobu/ruby/tree/opt_newarray_pack

tenderlove:
> FYI: https://github.com/nobu/ruby/tree/opt_newarray_pack

hah, you beat me to it.  I'll close this PR in favor of yours.  Thanks @nobu!

tenderlove:
@nobu btw I think we should add a test for monkey patching array to your branch (I didn't see a test)

nobu:
Well, I did it just to see if possible, and no idea about RJit support yet.
I'm OK if you want to take it over.

tenderlove:
@nobu OK, I'll take it over.  I'll submit a patch with the rbinc stuff first then send another with the `pack` optimization.  Also I like what you did with the fake array stuff so I'll incorporate that as well.

Thank you!

PR #8740, opened 2023-10-23
[PRISM] NoKeywordsParameterNode, ParametersNode

jemmaissroff:
Closes https://github.com/ruby/prism/issues/1669, https://github.com/ruby/prism/issues/1670


prism_compile.c
```diff
@@ -2118,6 +2118,8 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
       case PM_NIL_NODE:
         PM_PUTNIL_UNLESS_POPPED
         return;
+      case PM_NO_KEYWORDS_PARAMETER_NODE:
+        return;
```
kddnewton, line 2122: I don't think we can return here immediately without marking the iseq as not accepting keyword parameters. I'm seeing this for CRuby right now:

```
irb(main):003> RubyVM::AbstractSyntaxTree.parse("def foo(**nil); end")
=> 
(SCOPE@1:0-1:19
 tbl: []
 args: nil
 body:
   (DEFN@1:0-1:19
    mid: :foo
    body:
      (SCOPE@1:0-1:19
       tbl: []
       args:
         (ARGS@1:8-1:13
          pre_num: 0
          pre_init: nil
          opt: nil
          first_post: nil
          post_num: 0
          post_init: nil
          rest: nil
          kw: false
          kwrest: false
          block: nil)
       body: nil)))
irb(main):004> RubyVM::AbstractSyntaxTree.parse("def foo; end")
=> 
(SCOPE@1:0-1:12
 tbl: []
 args: nil
 body:
   (DEFN@1:0-1:12
    mid: :foo
    body:
      (SCOPE@1:0-1:12
       tbl: []
       args:
         (ARGS@1:7-1:7
          pre_num: 0
          pre_init: nil
          opt: nil
          first_post: nil
          post_num: 0
          post_init: nil
          rest: nil
          kw: nil
          kwrest: nil
          block: nil)
       body: nil)))
```

looks like `kw` gets set to `false` if this parameter is present.

kddnewton (changes requested):

prism_compile.c
```diff
@@ -2193,8 +2193,10 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
       case PM_NIL_NODE:
         PM_PUTNIL_UNLESS_POPPED
         return;
-      case PM_NO_KEYWORDS_PARAMETER_NODE:
+      case PM_NO_KEYWORDS_PARAMETER_NODE: {
+        ISEQ_BODY(iseq)->param.flags.accepts_no_kwarg = TRUE;
```
jemmaissroff, line 2197: Got this from [here](https://github.com/jemmaissroff/ruby/blob/5ded3fa417ef7e3b9509e1876072df636150cfb8/compile.c#L2079-L2081). It's still showing the wrong message, `wrong number of arguments (given 1, expected 0)` instead of `no keywords accepted`. But I'm fairly sure that's a bug due to the fact that we haven't implemented `KeywordRestParameterNode`, since `def foo(**args); end; foo(a: 1)` currently gives the same error (incorrectly)

prism_compile.c
```diff
@@ -2193,6 +2193,10 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
       case PM_NIL_NODE:
         PM_PUTNIL_UNLESS_POPPED
         return;
+      case PM_NO_KEYWORDS_PARAMETER_NODE: {
+        ISEQ_BODY(iseq)->param.flags.accepts_no_kwarg = TRUE;
```
kddnewton, line 2197:
```suggestion
        ISEQ_BODY(iseq)->param.flags.accepts_no_kwarg = true;
```

kddnewton (approved):

PR #8748, opened 2023-10-24
[Bug #19969] Rehash after deleted a half or more

nobu:
https://bugs.ruby-lang.org/issues/19969

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -1423,6 +1423,16 @@ rb_hash_foreach(VALUE hash, rb_foreach_func *func, VALUE farg)
     hash_verify(hash);
 }
 
+void rb_st_compact_table(st_table *tab);
+
+static void
+compact_after_delete(VALUE hash)
+{
+    if (RHASH_ITER_LEV(hash) == 0 && RHASH_ST_TABLE_P(hash)) {
+        rb_st_compact_table(RHASH_ST_TABLE(hash));
+    }
+}
+
 static VALUE
 hash_alloc_flags(VALUE klass, VALUE flags, VALUE ifnone, bool st)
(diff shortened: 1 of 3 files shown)


hsbt:
This patch didn't fix memory usage.

```
# master HEAD
$ ruby -v -rset -e 's1 = Set.new(10000.times); s2 = Set.new(9999.times); Array.new(10000) { s1 - s2 - [0] }; puts `ps -o rss= -p #{$$}`.to_i'
ruby 3.3.0dev (2023-11-07T20:48:06Z master aa6642de63) +YJIT [arm64-darwin22]
4529376
# This branch
$ ./ruby -v -rset -e 's1 = Set.new(10000.times); s2 = Set.new(9999.times); Array.new(10000) { s1 - s2 - [0] }; puts `ps -o rss= -p #{$$}`.to_i'
ruby 3.3.0dev (2023-11-08T04:39:19Z nobu-rehash-after-.. eccfdd6087) [arm64-darwin22]
4237136
```

eregon:
I think that's due to a missing `rehash_after_delete()` call in `Hash#delete` in this PR.

Set#- uses Hash#delete:
https://github.com/ruby/ruby/blob/8f11a422d3ffc32dce5aba721cf3626ea6e0f758/lib/set.rb#L638-L639
https://github.com/ruby/ruby/blob/8f11a422d3ffc32dce5aba721cf3626ea6e0f758/lib/set.rb#L617-L618
https://github.com/ruby/ruby/blob/8f11a422d3ffc32dce5aba721cf3626ea6e0f758/lib/set.rb#L537-L538

PR #8752, opened 2023-10-24
[PRISM] Implement compilation for PostExecutionNode

jemmaissroff:
Compiling the PostExecutionNode requires defining a way to create a new prism iseq with a callback. This commit includes that definition and all related ones, as well as the implementation of compilation for the PostExecutionNode.

Closes https://github.com/ruby/prism/issues/1655

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1106,6 +1106,11 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
             scope->locals = cast->locals;
             break;
         }
+        case PM_STATEMENTS_NODE: {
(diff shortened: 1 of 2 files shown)


kddnewton:
It seems like we're using this callback function in order to inject behavior when a scope node gets compiled. Is that generally correct?

We previously ran into this with for loops, because the first instruction has to be assigning the loop variable to the index field. We solved this with for loops by checking the node attached to the scope node when it gets compiled. Could we do something like that here as well? (Detecting the a `POST_EXECUTION_NODE` is attached to the scope node when the scope node gets compiled.) That way we could avoid having to deal with callbacks, which are a lot harder to follow.

kddnewton (approved):
Looks great!

PR #8754, opened 2023-10-24
Avoid array allocation for f(*a, **kw)

jeremyevans:
In Ruby 3.2, `f(*a, **kw)` was compiled to something like `f(*[*a, kw])`:

```
$ ruby32 -v --dump=i -e'f(*A,**KW)'
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-openbsd]
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)> (catch: false)
0000 putself                                                          (   1)[Li]
0001 opt_getconstant_path                   <ic:0 A>
0003 splatarray                             true
0005 putspecialobject                       1
0007 newhash                                0
0009 opt_getconstant_path                   <ic:1 KW>
0011 opt_send_without_block                 <calldata!mid:core#hash_merge_kwd, argc:2, ARGS_SIMPLE>
0013 newarray                               1
0015 concatarray
0016 opt_send_without_block                 <calldata!mid:f, argc:1, ARGS_SPLAT|FCALL|KW_SPLAT>
0018 leave
```

This alocated 4 arrays and 1 hash.  In e87d0882910001ef3b0c2ccd43bf00cee8c34a0c, this was changed to pass the splatted array and keywords as separate arguments:

```
$ ruby33 -v --dump=i -e'f(*A,**KW)'
ruby 3.3.0preview2 (2023-09-14 master e50fcca9a7) [x86_64-openbsd]
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)>
0000 putself                                                          (   1)[Li]
0001 opt_getconstant_path                   <ic:0 A>
0003 splatarray                             true
0005 opt_getconstant_path                   <ic:1 KW>
0007 opt_send_without_block                 <calldata!mid:f, argc:2, ARGS_SPLAT|FCALL|KW_SPLAT>
0009 leave
```

This reduced the allocations to a 1 array and 0 hashes.  The array was allocated because of the true flag to splatarray.  However, that flag is no longer needed now that the keywords are not added to the array. So this patch changes the splatarray flag to false in this case, making this call allocationless.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -5937,7 +5937,13 @@ setup_args_core(rb_iseq_t *iseq, LINK_ANCHOR *const args, const NODE *argn,
       }
       case NODE_ARGSPUSH: {
         if (flag_ptr) *flag_ptr |= VM_CALL_ARGS_SPLAT;
-        int argc = setup_args_core(iseq, args, RNODE_ARGSPUSH(argn)->nd_head, 1, NULL, NULL);
+        /* Do not dup_rest for f(*a, **kw) or f(*a, kw: 1)
+         * dup_rest for f(*a, arg) and f(*a, {kw: 1})
+         */
+        int argc = setup_args_core(iseq, args, RNODE_ARGSPUSH(argn)->nd_head,
+                                   !(nd_type(RNODE_ARGSPUSH(argn)->nd_body) == NODE_HASH &&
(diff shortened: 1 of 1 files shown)


jeremyevans:
Closing this as I don't think it handles `f(*a, **a.pop)` correctly.  I'll work on a fixed version.

PR #8756, opened 2023-10-24
[PRISM] Move pm_scope_node_init to prism_compile.c

jemmaissroff:
pm_scope_node_init is only used for CRuby, so should not live in the ruby/prism repo. We will merge the changes here first so they're not breaking, and will then remove from ruby/prism.

https://github.com/ruby/prism/pull/1735 is the corresponding ruby/prism change to be merged after this merges.


eregon:
`pm_scope_node_t` itself should also be moved here, so the correct types can be used for e.g. index_lookup_table

prism_compile.h
```diff
@@ -0,0 +1,20 @@
...
+    struct pm_scope_node *previous;
+    pm_node_t *ast_node;
+    struct pm_parameters_node *parameters;
+    pm_node_t *body;
+    pm_constant_id_list_t locals;
+    pm_parser_t *parser;
+
+    // We don't have the CRuby types ID and st_table within Prism
+    // so we use void *
+    void *constants; // ID *constants
```
kddnewton, line 16: We can do this here right? Without needing `void*`?

kddnewton (approved):
One suggestion

PR #8764, opened 2023-10-25
Fix few test_prism_eval that snuck in

HParker:


eileencodes:
Thanks @HParker I was just about to fix these too.

PR #8767, opened 2023-10-25
Clean up rb_ivar_count

tenderlove:
We can use object shapes to count the number of instance variables on an object.  This allows us to avoid taking the VM lock most of the time (for modules and classes), and we also don't have to do Qundef checks because shapes provides us with precise IV counts

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -2074,41 +2074,10 @@ rb_ivar_count(VALUE obj)
 
     switch (BUILTIN_TYPE(obj)) {
       case T_OBJECT:
-        if (rb_shape_obj_too_complex(obj)) {
-            return ROBJECT_IV_COUNT(obj);
-        }
-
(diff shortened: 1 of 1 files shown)


internal/class.h
```diff
@@ -111,6 +112,44 @@ struct RClass_and_rb_classext_t {
 #define RCLASS_SUPERCLASSES_INCLUDE_SELF FL_USER2
 #define RICLASS_ORIGIN_SHARED_MTBL FL_USER3
 
+static inline st_table *
+RCLASS_IV_HASH(VALUE obj)
+{
+    RBIMPL_ASSERT_TYPE(obj, RUBY_T_OBJECT);
```
jhawthorn, line 118:
```suggestion
  RUBY_ASSERT(RB_TYPE_P(obj, RUBY_T_CLASS) || RB_TYPE_P(obj, RUBY_T_MODULE));
```

PR #8768, opened 2023-10-26
Backport #18991 for 3.1: Fix false LocalJumpError when branch coverage is enabled

KJTsanaktsidis:
This is a backport PR for the fix for https://bugs.ruby-lang.org/issues/18991 in https://github.com/ruby/ruby/pull/6688 to be added to Ruby 3.1. The issue still affects 3.1 and the fix here applied cleanly, passed all the tests, and fixes the issue, so I think it should be OK to backport?

---

Fix false LocalJumpError when branch coverage is enabled

`throw TAG_BREAK` instruction makes a jump only if the continuation of catch of TAG_BREAK exactly matches the instruction immediately following the "send" instruction that is currently being executed. Otherwise, it seems to determine break from proc-closure.

Branch coverage may insert some recording instructions after "send" instruction, which broke the conditions for TAG_BREAK to work properly.

This change forces to set the continuation of catch of TAG_BREAK immediately after "send" (or "invokesuper") instruction.

[Bug #18991]

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -7216,7 +7216,30 @@ compile_iter(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, in
 			   ISEQ_TYPE_BLOCK, line);
 	CHECK(COMPILE(ret, "iter caller", node->nd_iter));
     }
-    ADD_LABEL(ret, retry_end_l);
+
+    {
+        // We need to put the label "retry_end_l" immediately after the last "send" instruction.
+        // This because vm_throw checks if the break cont is equal to the index of next insn of the "send".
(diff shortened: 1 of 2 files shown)


hsbt:
I marked https://github.com/ruby/ruby/pull/8768/commits/2d039099631e512a795a43536eaaf9722598f9cf for backport target at https://github.com/ruby/ruby/pull/8768/commits/2d039099631e512a795a43536eaaf9722598f9cf

KJTsanaktsidis:
Thank you!

KJTsanaktsidis:
This got backported: https://github.com/ruby/ruby/commit/d494cf4ddababb80660381e963f910ccacc3f7bc

Thanks again.

PR #8772, opened 2023-10-26
[DOC] Fix markup for == and ===

kyanagi:
It seems that `+==+` and `+===+` are not being parsed as intended.

https://docs.ruby-lang.org/en/master/Numeric.html#method-i-eql-3F


kyanagi:
Fixed on master

PR #8773, opened 2023-10-26
Move some defines from shape.h to shape.c

casperisfine:
If they are only used there, we might as well not expose them.

@tenderlove


tenderlove (approved):
Yes, great.  Thank you!

tenderlove:
I manually pushed this commit.

PR #8781, opened 2023-10-26
compile defined with nested constants and stovetop

HParker:
This matches the way compile.c compiles this, but using prism AST.


prism_compile.c
```diff
@@ -1206,6 +1222,27 @@ pm_compile_defined_expr(rb_iseq_t *iseq, const pm_defined_node_t *defined_node,
...
 
+void
+pm_compile_defined_expr(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, const uint8_t *src, bool popped, pm_scope_node_t *scope_node,  NODE dummy_line_node, int lineno, bool in_condition)
+{
+    LABEL *lfinish[3];
+    LINK_ELEMENT *last = ret->last;
+
+    lfinish[0] = NEW_LABEL(lineno);
+    lfinish[1] = 0;
+    lfinish[2] = 0;
```
jemmaissroff, line 1233: I think we don't need `lfinish[2]` and can instead make this an array with 2 elements

jemmaissroff (changes requested):
Looking good to me, one small comment!

prism_compile.c
HParker: updated thanks!

HParker:
@jemmaissroff I updated with support for `defined? []` as well since it uses the same recursive structure. It is in a second commit, so we can move them into a different PR if you prefer.

PR #8782, opened 2023-10-27
Support tracing of struct member accessor methods

jeremyevans:
This follows the same approach used for attr_reader/attr_writer in 2d98593bf54a37397c6e4886ccc7e3654c2eaf85, skipping the checking for tracing after the first call using the call cache, and clearing the call cache when tracing is turned on/off.

Fixes [Bug #18886]

### DIFF
--- a/benchmark/struct_accessor.yml
+++ b/benchmark/struct_accessor.yml
@@ -1,0 +1,25 @@
+prelude: |
+  C = Struct.new(:x) do
+    class_eval <<-END
+      def r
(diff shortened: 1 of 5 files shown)


ioquatix:
I no longer need this feature because I changed my implementation to no longer need `TracePoint`. However, this looks like a good improvement for consistency's sake, so I'm happy to see this PR.

ioquatix:
Nice work!

PR #8789, opened 2023-10-27
Windows: Add libffi as a build dependency

larskanis:
Otherwise the fiddle extension isn't build.


junaruga:
Note that the `[DOC]` keyword in the commit message is useful to skip the CI.

PR #8790, opened 2023-10-27
.travis.yml: Upgrade the Ubuntu arm32 to 22.04 (Jammy).

junaruga:
Upgrade the gcc version to the latest version 11.4.0 to align the gcc version used in the RubyCI arm64 server.

I want to disable allow_failures if possible to avoid unintentional new
failures. It seems that the arm32 case is stable.

Feel free to enable allow_failures again when you see the case is unstable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ language: c
 
 os: linux
 
-dist: focal
+dist: jammy
 
 git:
   quiet: true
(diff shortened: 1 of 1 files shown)


junaruga:
Travis looks good.

https://app.travis-ci.com/github/ruby/ruby/builds/266893188

* arm32: 12:32
* ppc64le: 23:07 (no change, and no error)

PR #8802, opened 2023-10-30
[PRISM] Implement all argument types for CallNodes

jemmaissroff:
This implements all argument types for CallNodes. It does not add tests to `test/test_compile_prism.rb` because all argument types have not been correctly implemented on DefNodes. There are ways we could work around this (compile the iseq with CRuby that defines the method, and then call it using Prism's compiler), but implementing that does not seem like the best use of time since we won't need it when we continue implementing DefNodes next. It has been thoroughly manually tested as a stopgap, and I will continue to debug this once we finish DefNodes.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1369,6 +1369,7 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         ID method_id = pm_constant_id_lookup(scope_node, call_node->name);
         int flags = 0;
         int orig_argc = 0;
+        struct rb_callinfo_kwarg *kw_arg = NULL;
 
         if (call_node->receiver == NULL) {
             ADD_INSN(ret, &dummy_line_node, putself);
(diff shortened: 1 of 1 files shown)


prism_compile.c
```diff
@@ -1380,21 +1381,155 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                          break;
+                      }
+                      else {
+                          kw_arg = rb_xmalloc_mul_add(len, sizeof(VALUE), sizeof(struct rb_callinfo_kwarg));
+                          flags |= VM_CALL_KWARG;
+                          kw_arg->keyword_len += len;
+
+                          for (size_t i = 0; i < len; i++) {
+                              pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+                              kw_arg->keywords[i] = pm_static_literal_value(assoc->key, scope_node, parser);
```
kddnewton, line 1479: I'm expecting this won't work for things like `foo(bar => baz)`. Can you add a comment here indicating that we'll need to revisit this?

kddnewton (approved):
One comment then good

PR #8804, opened 2023-10-30
Ensure keyword splat method argument is hash

jeremyevans:
Commit e87d0882910001ef3b0c2ccd43bf00cee8c34a0c introduced a regression where the keyword splat object passed by the caller would be directly used by callee as keyword splat parameters, if it implemented #to_hash.  The return value of #to_hash would be ignored in this case.


ko1 (approved):
Thank you!!

PR #8813, opened 2023-11-01
fix regex from regex memory corruption

HParker:
before this change, creating a regex from a regex with a named capture, `Regexp.new(/(?<name>)/)`, causes memory to be shared between the two named capture groups which can cause a segfault if the original is GCed.

This copies the named capture groups so each regex has their own copy.

Reproduction script:

```
10_000.times do
  Regexp.new(Regexp.new("(?<name>)"))
end
```

bug introduced in: https://github.com/ruby/ruby/commit/ab6eb3786c94e69c561080cbb796c2381702a3a4

CC: @nobu


peterzhu2118 (approved):

nobu:
Good catch.

Wouldn’t it be better to add the function to copy the named capture table rather than moving the named capture stuffs to the header?

HParker:
> Wouldn’t it be better to add the function to copy the named capture table rather than moving the named capture stuffs to the header?

@nobu 

Yeah, I tried to just add,

```
static UChar* strdup_with_null(OnigEncoding enc, const UChar* s, const UChar* end);
```

to the header and leave the definition in regparse.c, but I get,

```
regparse.h:345:15: warning: ‘strdup_with_null’ used but never defined
```

Maybe there is an easy way to fix this warning?

nobu (changes requested):
Also `NameTable` may or may not be `st_table` depending on `USE_ST_LIBRARY`.
It is not good to hardcode `NameTable` as `st_table`.

HParker:
@nobu is it possible to have named capture groups without `ST_LIBRARY`? 

I see `name_add` always does `onig_st_insert_strend`
https://github.com/ruby/ruby/blob/master/regparse.c#L806

Also, I am replacing a `st_copy` in this PR which I think also requires `ST_LIBRARY`

I tried compiling on the `master` branch without the `#define USE_ST_LIBRARY` line in `regint.h` and I get link time errors.

Maybe we can add support for named captures without `ST_LIBRARY` in another PR since I think it is going to touch things unrelated to this change.

HParker:
I will try to create a PR makes disabling `USE_ST_LIBRARY` work.

HParker:
oh! you beat me to it!
https://github.com/ruby/ruby/pull/8821

Thank you! ❤️ 

I'll update this PR to also work without ST_LIBRARY once your PR merges.

HParker:
Thanks @nobu! ❤️ I think this PR also looks much better now. Thanks for fixing the ST_LIBRARY stuff.

nobu (changes requested):
Onigmo files obey the different style, indenting 2 spaces.

test/ruby/test_regexp.rb
```diff
@@ -723,6 +723,12 @@ def test_initialize
     assert_raise(RegexpError) { Regexp.new("((?<v>))\\g<0>") }
   end
 
+  def test_initialize_from_regex_memory_corruption
+    10_000.times do
+      Regexp.new(Regexp.new("(?<name>)"))
+    end
```
HParker, line 729: @nobu does this test look ok? I am not sure if there is an existing pattern for tests like this.

nobu: One possible way is `assert_ruby_status`, to assert it doesn't crash.
```ruby
    assert_ruby_status([], "#{<<-"begin;"}\n#{<<-'end;'}")
    begin;
      10_000.times do
        Regexp.new(Regexp.new("(?<name>)"))
      end
    end;
```

PR #8817, opened 2023-11-02
Update racc revision to test for dummy rake/extensiontask.rb

nobu:


hsbt:
racc version is now `1.7.3` at `bundled_gems`

PR #8819, opened 2023-11-02
Make String.new size pools aware.

casperisfine:
If the required capacity would fit in an embded string, returns one.

This can reduce malloc churn for code that use string buffers.

```
$ time ../miniruby-master -v -e '10_000_000.times { String.new(capacity: 512) }'
ruby 3.3.0dev (2023-11-02T16:15:48Z master ad4f973ecd) [arm64-darwin22]
last_commit=YJIT: Always define method codegen table at boot (#8807)

real	0m2.821s
user	0m2.460s
sys	0m0.249s
$ time ./miniruby -v -e '10_000_000.times { String.new(capacity: 512) }'
ruby 3.3.0dev (2023-11-02T16:58:58Z string-buffer-opt ee1dacb047) [arm64-darwin22]

real	0m1.127s
user	0m1.080s
sys	0m0.020s
```

However backward compatibility is tricky. `new` should forward all arguments to the subclass init.

### DIFF
--- a/string.c
+++ b/string.c
@@ -1874,6 +1874,94 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
     return str;
 }
 
+static size_t
+str_embed_size(long capa)
+{
+    return offsetof(struct RString, as.embed.ary) + (sizeof(char) * capa);
+}
+
+static bool
+str_embeddable_p(long capa)
+{
+    return rb_gc_size_allocatable_p(str_embed_size(capa));
+}
+
+static VALUE
+string_create(rb_execution_context_t *ec, VALUE klass, VALUE orig, VALUE orig_notset, VALUE encoding, VALUE capacity)
+{
+    rb_encoding *enc = NULL;
+    long termlen = 1;
+
+    if (orig_notset == Qtrue) {
+        orig = Qnil;
+    }
+    else {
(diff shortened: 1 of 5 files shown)


string.c
```diff
@@ -1874,6 +1874,94 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
...
+    if (NIL_P(encoding)) {
+        if (!NIL_P(orig)) {
+            encoding = rb_obj_encoding(orig);
+        }
+    }
+
+    if (!NIL_P(encoding)) {
+        enc = rb_to_encoding(encoding);
+        termlen = rb_enc_mbminlen(enc);
+    }
```
paracycle, line 1911: Isn't this the same as:
```suggestion
    if (NIL_P(encoding)) {
        if (!NIL_P(orig)) {
            encoding = rb_obj_encoding(orig);
        }
    } else {
        enc = rb_to_encoding(encoding);
        termlen = rb_enc_mbminlen(enc);
    }
```

casperisfine: Haha, no.

If `!NIL_P(orig)`, then we'll set `encoding`. It's super confusing I know :)

byroot:
Closing in favor of https://github.com/ruby/ruby/pull/8825

PR #8821, opened 2023-11-02
Fix onigmo name table without st

nobu:


nobu:
Draft because testing without st now.

regparse.c
```diff
@@ -736,10 +747,51 @@ onig_names_free(regex_t* reg)
...
+        onig_names_free(reg);
+        return ONIGERR_MEMORY;
+      }
+      e->back_refs = xmalloc(e->back_alloc * sizeof(e->back_refs[0]));
+      if (IS_NULL(e->back_refs)) {
+        xfree(e->name);
+        onig_names_free(reg);
+        return ONIGERR_MEMORY;
+      }
+      memcpy(e->back_refs, oe->back_refs, e->back_num * sizeof(e->back_refs[0]));
```
HParker, line 784: I think we need to manually set `back_ref1` here. I believe it is always the same as `back_refs[0]`

```suggestion
      memcpy(e->back_refs, oe->back_refs, e->back_num * sizeof(e->back_refs[0]));
      e->back_ref1 = e->back_refs[0];
```

PR #8822, opened 2023-11-02
[PRISM] Implement compilation for MultiWriteNodes, fix MultiTargetNodes

jemmaissroff:
Compilation now works for MultiWriteNodes and MultiTargetNodes, with nesting on MultiWrites. See the tests added in this commit for example behavior.

Closes https://github.com/ruby/prism/issues/1646

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -5274,7 +5274,6 @@ compile_massign0(rb_iseq_t *iseq, LINK_ANCHOR *const pre, LINK_ANCHOR *const rhs
 
     int llen = 0;
     int lpos = 0;
-    int expand = 1;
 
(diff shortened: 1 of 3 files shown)


compile.c
```diff
@@ -5274,7 +5274,6 @@ compile_massign0(rb_iseq_t *iseq, LINK_ANCHOR *const pre, LINK_ANCHOR *const rhs
 
     int llen = 0;
     int lpos = 0;
-    int expand = 1;
```
jemmaissroff, line 5277: Not directly related to this PR, just noticed that this int is unnecessary.

PR #8824, opened 2023-11-02
[PRISM] Fix test infrastucture, comment out failing tests

jemmaissroff:
[84dfa0fa5e451adb87beaf497165cb5a1bc93770](https://github.com/ruby/ruby/commit/84dfa0fa5e451adb87beaf497165cb5a1bc93770) introduced a bug where we were no longer testing the "popped" case because the "; 1" meant to be appended to the source was no longer functioning as intended.

The first commit here re-introduces the popped case, and comments out all now failing tests.

All subsequent commits fix popped cases for different nodes

cc @nobu

### DIFF
--- a/test/ruby/test_compile_prism.rb
+++ b/test/ruby/test_compile_prism.rb
@@ -214,8 +214,8 @@ module Prism
     end
 
     def test_ConstantPathOrWriteNode
-      # assert_prism_eval("Prism::CPOrWN = nil; Prism::CPOrWN ||= 1")
-      # assert_prism_eval("Prism::CPOrWN ||= 1")
+      assert_prism_eval("Prism::CPOrWN = nil; Prism::CPOrWN ||= 1")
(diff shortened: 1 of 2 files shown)


kddnewton (approved):

test/ruby/test_compile_prism.rb
```diff
@@ -962,7 +964,6 @@ def compare_eval(source)
     def assert_prism_eval(source)
       $VERBOSE, verbose_bak = nil, $VERBOSE
 
-      source = "class Prism::TestCompilePrism\n#{source}\nend"
```
kddnewton, line 965: Did you mean to do this?

jemmaissroff: Yes, otherwise we could be missing other bugs by virtue of always testing something within a class, and never isolated.

jemmaissroff: But it looks like it's causing tests to fail, so will revert for now!

PR #8825, opened 2023-11-02
Make String.new size pools aware.

casperisfine:
Fix: https://github.com/ruby/ruby/pull/8819

If the required capacity would fit in an embded string, returns one.

This can reduce malloc churn for code that use string buffers.

```
$ time ../miniruby-master -v -e '10_000_000.times { String.new(capacity: 512) }'
ruby 3.3.0dev (2023-11-02T16:15:48Z master ad4f973ecd) [arm64-darwin22]
last_commit=YJIT: Always define method codegen table at boot (#8807)

real	0m2.674s
user	0m2.377s
sys	0m0.250s
$ time ./miniruby -v -e '10_000_000.times { String.new(capacity: 512) }'
ruby 3.3.0dev (2023-11-02T18:19:48Z string-buffer-opt2 e6633b02a3) [arm64-darwin22]

real	0m1.377s
user	0m1.323s
sys	0m0.034s
```

The speedup is smaller than in https://github.com/ruby/ruby/pull/8819, but it no longer breaks subclasses.

### DIFF
--- a/string.c
+++ b/string.c
@@ -1874,6 +1874,98 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
     return str;
 }
 
+/* :nodoc: */
+static VALUE
+rb_str_s_new(int argc, VALUE *argv, VALUE klass)
+{
+    if (klass != rb_cString) {
+        return rb_class_new_instance_pass_kw(argc, argv, klass);
+    }
+
+    static ID keyword_ids[2];
+    VALUE orig, opt, encoding = Qnil, capacity = Qnil;
+    VALUE kwargs[2];
+    rb_encoding *enc = NULL;
+
+    int n = rb_scan_args(argc, argv, "01:", &orig, &opt);
+    if (NIL_P(opt)) {
+        return rb_class_new_instance_pass_kw(argc, argv, klass);
+    }
+
+    keyword_ids[0] = rb_id_encoding();
+    CONST_ID(keyword_ids[1], "capacity");
+    rb_get_kwargs(opt, keyword_ids, 0, 2, kwargs);
+    encoding = kwargs[0];
+    capacity = kwargs[1];
+
+    int termlen = 1;
+
+    if (n == 1) {
+        orig = StringValue(orig);
+    }
+    else {
+        orig = Qnil;
+    }
+
+    if (UNDEF_P(encoding)) {
+        if (!NIL_P(orig)) {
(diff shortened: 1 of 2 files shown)


string.c
```diff
@@ -1874,6 +1874,114 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
     return str;
 }
 
+static size_t
+str_embed_size(long capa)
+{
+    return offsetof(struct RString, as.embed.ary) + (sizeof(char) * capa);
+}
```
peterzhu2118, line 1881: There already is `rb_str_embed_size`, which is the same.

string.c
```diff
@@ -1874,6 +1874,114 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
...
+
+static bool
+str_embeddable_p(long capa)
+{
+    return rb_gc_size_allocatable_p(str_embed_size(capa));
+}
```
peterzhu2118, line 1887: Same, with `STR_EMBEDDABLE_P`.

string.c
```diff
@@ -1874,6 +1874,114 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
...
+    if (UNDEF_P(capacity)) {
+        if (NIL_P(orig)) {
+            str = str_new(klass, "", 0);
+            if (enc) {
+                rb_enc_associate(str, enc);
+            }
+            return str;
+        }
+        VALUE copy = str_duplicate(klass, orig);
+        str_modifiable(copy);
```
peterzhu2118, line 1945: I don't think we need this line? `str_modifiable` just checks that `copy` is not locked nor frozen, both of which should be true since it's newly created.

string.c
```diff
@@ -1874,6 +1874,114 @@ rb_str_init(int argc, VALUE *argv, VALUE str)
...
+
+    if (str_embeddable_p(capa)) {
+        str = str_alloc_embed(klass, capa);
+    }
+    else {
+        str = str_alloc_heap(klass);
+        RSTRING(str)->as.heap.aux.capa = capa - termlen;
+        RSTRING(str)->as.heap.ptr = ALLOC_N(char, (size_t)capa);
+        STR_SET_LEN(str, 0);
+    }
```
peterzhu2118, line 1972: You can do:

```suggestion
    str = str_new0(klass, NULL, capa - termlen, termlen);
```

peterzhu2118 (approved):

PR #8830, opened 2023-11-02
[PRISM] Compile Ensure Nodes

eightbitraptor:
- ~~This is dependant on https://github.com/ruby/ruby/pull/8829 and https://github.com/ruby/ruby/pull/8828 being merged. Tests will fail until then~~ 

   EDIT: These PR's are now merged
- ~~This is also dependant on fixing the last remaining hack in `pm_scope_node_init` where I am forcibly adding the `$!` `ERROR_STATUS` constant into the child iseq's local table. I need a better way of referencing this global constant from inside the child iseq rather than hardcoding the ID.~~ 

   EDIT: Fixed this. I had overlooked the availablity of `idERROR_STATUS` in `compile.c`

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1079,15 +1079,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         }
         case PM_ENSURE_NODE: {
             pm_ensure_node_t *cast = (pm_ensure_node_t *)node;
-
-            // HAXXX
-            pm_constant_id_t err_info = 3863;
-            pm_constant_id_list_t locals;
-            pm_constant_id_list_init(&locals);
-            pm_constant_id_list_append(&locals, err_info);
-
             scope->body = (pm_node_t *)cast->statements;
-            scope->locals = locals;
             break;
         }
         case PM_FOR_NODE: {
@@ -1590,6 +1582,11 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
             iseq_set_exception_local_table(iseq);
             pm_scope_node_t next_scope_node;
             pm_scope_node_init((pm_node_t *)begin_node->ensure_clause, &next_scope_node, scope_node, parser);
+
+            pm_constant_id_list_t locals;
+            pm_constant_id_list_init(&locals);
+            pm_constant_id_list_append(&locals, idERROR_INFO);
+            next_scope_node.locals = locals;
 
             child_iseq = NEW_CHILD_ISEQ(next_scope_node,
                     rb_str_new2("ensure in"),
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -12,7 +12,10 @@
 #define NEW_CHILD_ISEQ(node, name, type, line_no) \
     pm_new_child_iseq(iseq, (node), parser, rb_fstring(name), iseq, (type), (line_no))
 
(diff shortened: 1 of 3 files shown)


prism_compile.c
```diff
@@ -1074,6 +1074,19 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
             scope->locals = cast->locals;
             break;
         }
+        case PM_ENSURE_NODE: {
+            pm_ensure_node_t *cast = (pm_ensure_node_t *)node;
+
+            // HAXXX
+            pm_constant_id_t err_info = 3863;
```
kddnewton, line 1081: We already have the top bit indicate that it's special. What do you think about:

| top bit | rest bits | result |
| --- | --- | --- |
| 0 | anything | actual constant |
| 1 | 0 | make a temporary |
| 1 | anything non-zero | mask off top bit, use rest as ID |

Is that too hacky?

prism_compile.c
```diff
@@ -3295,9 +3359,22 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
             ADD_CATCH_ENTRY(CATCH_TYPE_NEXT, start, end, NULL, end);
             break;
         }
+        case ISEQ_TYPE_ENSURE: {
+            iseq_set_exception_local_table(iseq);
+            pm_compile_node(iseq, (pm_node_t *)(scope_node->body), ret, src, popped, scope_node);
```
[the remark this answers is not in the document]
jemmaissroff: This changes slightly depending on the comment above about `PM_COMPILE`

test/ruby/test_compile_prism.rb
```diff
@@ -562,6 +562,59 @@ def test_BreakNode
...
+      assert_prism_eval(<<~ENS)
+        begin
+          puts 1
+        ensure
+          puts 2
+        end
+        puts "hi"
+      ENS
+
+    # popped
```
jemmaissroff, line 614: I don't think this case is necessary because of [this](https://github.com/ruby/ruby/blob/master/test/ruby/test_compile_prism.rb#L1005-L1006)

jemmaissroff (changes requested):

prism_compile.c
```diff
@@ -12,8 +12,8 @@
 #define NEW_CHILD_ISEQ(node, name, type, line_no) \
     pm_new_child_iseq(iseq, (node), parser, rb_fstring(name), iseq, (type), (line_no))
 
-#define PM_COMPILE(node) \
-    pm_compile_node(iseq, (node), ret, src, popped, scope_node)
+#define PM_COMPILE(_ret, node) \
```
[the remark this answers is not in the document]
eightbitraptor: I have no real preference on this. I guess having `PM_COMPILE_WITH_LINK_ANCHOR` would result in a smaller diff.

When speaking to @tenderlove about this he showed a preference towards making the `PM_COMPILE` family of macros have more similarities with the CRuby compiler macros, which is why I chose this refactor instead (as `COMPILE` does take a link anchor, as well as other parameters).

Again. I don't have strong feelings one way or the other.

prism_compile.c
eightbitraptor: Looking at this again. I don't think we need this at all as we can just use `idERROR_INFO` I had overlooked this initially.

jemmaissroff (approved):
Looks great, thanks for the work here!

PR #8839, opened 2023-11-03
YJIT: implement two-step call threshold

maximecb:
Automatically switch call threshold to a larger value for larger, production-sized apps, while still allowing smaller apps and command-line programs to start with a lower threshold.

### DIFF
--- a/yjit/src/options.rs
+++ b/yjit/src/options.rs
@@ -108,7 +108,7 @@ static YJIT_OPTIONS: [(&str, &str); 11] = [
     ("--yjit-call-threshold=num",       "Number of calls to trigger JIT"),
     ("--yjit-small-threshold=num",      "Call threshold for small apps (default: 30)"),
     ("--yjit-large-threshold=num",      "Call threshold for large deployments (default: 120)"),
-    ("--yjit-large-iseq-count=num",     "Live ISEQs to consider an app large (default: 40000)"),
+    ("--yjit-large-iseq-count=num",     "Live ISEQs to consider an app large (default: 40K)"),
     ("--yjit-cold-threshold=num",       "Global call after which ISEQs not compiled (default: 200K)"),
 
     ("--yjit-max-versions=num",         "Maximum number of versions per basic block (default: 4)"),
--- a/yjit/src/options.rs
+++ b/yjit/src/options.rs
@@ -2,11 +2,20 @@ use std::{ffi::{CStr, CString}, ptr::null};
 use crate::backend::current::TEMP_REGS;
 use std::os::raw::{c_char, c_int, c_uint};
 
+// Call threshold for small deployments and command-line apps
+pub static SMALL_CALL_THRESHOLD: u64 = 30;
+
+// Call threshold for larger deployments and production-sized applications
+pub static LARGE_CALL_THRESHOLD: u64 = 120;
+
+// Number of live ISEQs after which we consider an app to be large
+pub static LARGE_ISEQ_COUNT: u64 = 40_000;
+
 // This option is exposed to the C side a a global variable for performance, see vm.c
 // Number of method calls after which to start generating code
 // Threshold==1 means compile on first execution
 #[no_mangle]
-pub static mut rb_yjit_call_threshold: u64 = 30;
+pub static mut rb_yjit_call_threshold: u64 = SMALL_CALL_THRESHOLD;
 
 // This option is exposed to the C side a a global variable for performance, see vm.c
 // Number of execution requests after which a method is no longer
@@ -21,15 +30,6 @@ pub struct Options {
     // Size of the executable memory block to allocate in bytes
     // Note that the command line argument is expressed in MiB and not bytes
     pub exec_mem_size: usize,
-
-    // Call threshold for small deployments and command-line apps
-    pub small_threshold: u64,
-
-    // Call threshold for larger deployments and production-sized applications
-    pub large_threshold: u64,
-
-    // Number of live ISEQs after which we consider an app to be large
-    pub large_iseq_count: u64,
 
     // Disable the propagation of type information
     pub no_type_prop: bool,
@@ -79,9 +79,6 @@ pub struct Options {
 // Initialize the options to default values
 pub static mut OPTIONS: Options = Options {
     exec_mem_size: 128 * 1024 * 1024,
-    small_threshold: 30,
-    large_threshold: 120,
-    large_iseq_count: 40_000,
     no_type_prop: false,
     max_versions: 4,
     num_temp_regs: 5,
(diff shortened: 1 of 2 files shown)


maximecb:
Tagging @byroot just so he's aware of this change. With this, YJIT should automatically use a larger threshold on SFR, Core, and other deployments similar in size or larger.

yjit/src/options.rs
```diff
@@ -87,13 +99,18 @@ pub static mut OPTIONS: Options = Options {
...
     ("--yjit-trace-exits",              "Record Ruby source location when exiting from generated code"),
     ("--yjit-trace-exits-sample-rate",  "Trace exit locations only every Nth occurrence"),
     ("--yjit-exec-mem-size=num",        "Size of executable memory block in MiB (default: 128)"),
-    ("--yjit-call-threshold=num",       "Number of calls to trigger JIT (default: 30)"),
+
+    ("--yjit-call-threshold=num",       "Number of calls to trigger JIT"),
+    ("--yjit-small-threshold=num",      "Call threshold for small apps (default: 30)"),
+    ("--yjit-large-threshold=num",      "Call threshold for large deployments (default: 120)"),
+    ("--yjit-large-iseq-count=num",     "Live ISEQs to consider an app large (default: 40000)"),
     ("--yjit-cold-threshold=num",       "Global call after which ISEQs not compiled (default: 200K)"),
```
XrXr, line 112:
```suggestion
    ("--yjit-large-iseq-count=num",     "Live ISEQs to consider an app large (default: 40K)"),
    ("--yjit-cold-threshold=num",       "Global call after which ISEQs not compiled (default: 200K)"),
```

byroot:
Makes sense 👍

yjit/src/options.rs
```diff
@@ -87,13 +99,18 @@ pub static mut OPTIONS: Options = Options {
...
+static YJIT_OPTIONS: [(&str, &str); 11] = [
     ("--yjit-stats",                    "Enable collecting YJIT statistics"),
     ("--yjit-trace-exits",              "Record Ruby source location when exiting from generated code"),
     ("--yjit-trace-exits-sample-rate",  "Trace exit locations only every Nth occurrence"),
     ("--yjit-exec-mem-size=num",        "Size of executable memory block in MiB (default: 128)"),
-    ("--yjit-call-threshold=num",       "Number of calls to trigger JIT (default: 30)"),
+
+    ("--yjit-call-threshold=num",       "Number of calls to trigger JIT"),
+    ("--yjit-small-threshold=num",      "Call threshold for small apps (default: 30)"),
+    ("--yjit-large-threshold=num",      "Call threshold for large deployments (default: 120)"),
```
k0kubun, line 110: I think we agreed on not documenting these in `ruby --help`? I guess you changed your mind since it's hard to document the defaults without them?

byroot: can't it just be hardcoded? If you pass an explicit `--call-threshold` then we ignore them.

Then these can be documented as part of the `--yjit-call_thresold` default.

XrXr (approved):

yjit/src/options.rs
k0kubun: I'm mildly leaning towards undoing this format so that we could refer to a constant (which we need to define and use for setting the actual default) without formatting it if we refactor it and allow these strings to be dynamic.

yjit/src/options.rs
maximecb: Yes you're right, we did agree, and I remembered it after opening the PR. I'll undocument these options 😆

maximecb: Jean's suggestion is not a bad one. I could define two constants instead.

yjit/src/yjit.rs
```diff
@@ -134,6 +134,13 @@ pub extern "C" fn rb_yjit_iseq_gen_entry_point(iseq: IseqPtr, ec: EcPtr, jit_exc
         return std::ptr::null();
     }
 
+    // If this is a large application (has very many ISEQs), switch to
+    // using the call threshold for large applications after this entry point
+    use crate::stats::rb_yjit_live_iseq_count;
+    if unsafe { rb_yjit_live_iseq_count } > LARGE_ISEQ_COUNT {
+        unsafe { rb_yjit_call_threshold = LARGE_CALL_THRESHOLD; };
```
k0kubun, line 141: I think we'd like to skip this when `--yjit-call-threshold` is explicitly specified. SFR uses 1000, but the current patch would decrease it to hard-coded 120.

yjit/src/yjit.rs
```diff
@@ -134,6 +134,13 @@ pub extern "C" fn rb_yjit_iseq_gen_entry_point(iseq: IseqPtr, ec: EcPtr, jit_exc
         return std::ptr::null();
     }
 
+    // If this is a large application (has very many ISEQs), switch to
+    // using the call threshold for large applications after this entry point
+    use crate::stats::rb_yjit_live_iseq_count;
+    if unsafe { rb_yjit_live_iseq_count } > LARGE_ISEQ_COUNT {
```
k0kubun, line 140:
```suggestion
    if rb_yjit_call_threshold < LARGE_CALL_THRESHOLD && unsafe { rb_yjit_live_iseq_count } > LARGE_ISEQ_COUNT {
```

yjit/src/yjit.rs
k0kubun: made a suggested patch https://github.com/ruby/ruby/pull/8839/files#r1382161082

maximecb:
I benchmarked this change on AWS:
```
yjit: ruby 3.3.0dev (2023-11-03T18:00:13Z master bc4d1c09fa) [x86_64-linux]
2step: ruby 3.3.0dev (2023-11-03T19:06:35Z yjit_two_step b0afaacbb9) [x86_64-linux]

--------------  ---------  ----------  ----------  ----------  -------------  ----------
bench           yjit (ms)  stddev (%)  2step (ms)  stddev (%)  2step 1st itr  yjit/2step
activerecord    81.7       3.1         81.8        3.5         1.00           1.00      
chunky-png      1095.4     0.4         1089.2      0.3         1.00           1.01      
erubi-rails     26.6       15.4        26.8        14.8        1.02           0.99      
hexapdf         3496.0     1.2         3552.6      1.2         1.02           0.98      
liquid-c        80.5       1.0         80.9        1.1         1.01           0.99      
liquid-compile  84.0       0.7         83.5        1.0         1.03           1.01      
liquid-render   200.7      0.5         198.5       0.6         1.01           1.01      
lobsters        1285.4     2.3         1291.9      2.1         1.04           0.99      
mail            175.1      0.6         175.6       0.6         0.98           1.00      
psych-load      2754.6     0.1         2792.8      0.2         1.01           0.99      
railsbench      2871.8     0.5         2885.9      0.6         0.98           1.00      
ruby-lsp        78.9       3.4         79.7        3.4         1.65           0.99      
sequel          89.9       1.7         89.3        1.7         0.97           1.01      
--------------  ---------  ----------  ----------  ----------  -------------  ----------
```

I'm was a bit surprised by the big change in 1st itr performance for ruby-lsp since it only has 24K live ISEQs according to `--yjit-stats`, so I reran that one:
```
yjit: ruby 3.3.0dev (2023-11-03T18:00:13Z master bc4d1c09fa) [x86_64-linux]
2step: ruby 3.3.0dev (2023-11-03T19:06:35Z yjit_two_step b0afaacbb9) [x86_64-linux]

--------  ---------  ----------  ----------  ----------  -------------  ----------
bench     yjit (ms)  stddev (%)  2step (ms)  stddev (%)  2step 1st itr  yjit/2step
ruby-lsp  79.2       3.2         79.5        3.1         1.00           1.00      
--------  ---------  ----------  ----------  ----------  -------------  ----------
```

Seems there is no significant performance change, which is good.

yjit/src/yjit.rs
maximecb: You're right that we shouldn't change the threshold if one is specified, but I think we should just check that it isn't the default value. I'll make that change 👍

yjit/src/yjit.rs
```diff
@@ -134,6 +134,14 @@ pub extern "C" fn rb_yjit_iseq_gen_entry_point(iseq: IseqPtr, ec: EcPtr, jit_exc
         return std::ptr::null();
     }
 
+    // If a custom call threshold was not specified on the command-line and
+    // this is a large application (has very many ISEQs), switch to
+    // using the call threshold for large applications after this entry point
+    use crate::stats::rb_yjit_live_iseq_count;
+    if unsafe { rb_yjit_call_threshold } == SMALL_CALL_THRESHOLD && unsafe { rb_yjit_live_iseq_count } > LARGE_ISEQ_COUNT {
```
maximecb, line 141: @k0kubun based on your feedback

k0kubun (approved):
:+1:

PR #8841, opened 2023-11-03
Add general_ivar_set to unify instance variable setting

peterzhu2118:
This PR implements `general_ivar_set` and switches Object, Class, and generic instance variables use it. This unifies the logic for how instance variables are set, which cleans up the implementation and reduces bugs in certain code paths.

cc. @byroot

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1370,42 +1370,6 @@ rb_attr_delete(VALUE obj, ID id)
     return rb_ivar_delete(obj, id, Qnil);
 }
 
-static int
-rb_complex_ivar_set(VALUE obj, ID id, VALUE val)
-{
-    struct gen_ivtbl *ivtbl;
-    st_table *table;
-    RUBY_ASSERT(rb_shape_obj_too_complex(obj));
-
-    switch (BUILTIN_TYPE(obj)) {
-      case T_OBJECT:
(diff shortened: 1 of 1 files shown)


variable.c
```diff
@@ -1599,59 +1672,49 @@ rb_obj_evacuate_ivs_to_hash_table(ID key, VALUE val, st_data_t arg)
...
+    RUBY_ASSERT(!rb_shape_obj_too_complex(obj));
 
-        RUBY_ASSERT(!rb_shape_obj_too_complex(obj));
+    return ROBJECT_IVPTR(obj);
+}
 
-        if (UNLIKELY(shape->next_iv_index >= num_iv)) {
-            RUBY_ASSERT(shape->next_iv_index == num_iv);
+static void
+obj_ivar_set_shape_resize_table(VALUE obj, attr_index_t old_capa, attr_index_t new_capa, void *_data)
```
byroot, line 1684: nitpick: the `_table` suffix make me think of `st_table`, as in TOO_COMPLEX.

byroot (approved):
❤️

variable.c
peterzhu2118: I renamed it `ivptr` 👍

PR #6792, opened 2022-11-23
YJIT: Use NonNull pointer for CodePtr

k0kubun:
Use `NonNull` inside `CodePtr` to reduce memory consumption, leveraging the following optimization:

> https://doc.rust-lang.org/stable/std/ptr/struct.NonNull.html
> Unlike *mut T, the pointer must always be non-null, even if the pointer is never dereferenced. This is so that enums may use this forbidden value as a discriminant – Option<NonNull<T>> has the same size as *mut T.

Also fixed a warning on `cargo test` with `MarkUnused`.

## Code Size
`yjit_alloc_size` is reduced from 17.8MB to 17.2MB.

### Before
```
code_region_size:         8830976
yjit_alloc_size:         17815913
```

### After
```
code_region_size:         8814592
yjit_alloc_size:         17228805
```

PR #8842, opened 2023-11-03
YJIT: Use u32 for CodePtr to save 4 bytes each

XrXr:
We've long had the restriction on the size of the code memory region
such that a u32 could refer to everything. This commit capitalizes on
this restriction by shrinking the storage of `CodePtr` to be 4 bytes from
8.

To derive a full raw pointer from a `CodePtr`, one needs a base pointer.
Both `CodeBlock` and `VirtualMemory` can be used for this purpose. The
base pointer is readily available everywhere, except for in the case of
the `jit_return` "branch". Generalize lea_label() to lea_jump_target()
in the IR to delay deriving the `jit_return` address until `compile()`,
when the base pointer is available.

On railsbench, this yields roughly a 1% reduction to `yjit_alloc_size`
(58,397,765 to 57,742,248).

### DIFF
--- a/yjit/src/backend/arm64/mod.rs
+++ b/yjit/src/backend/arm64/mod.rs
@@ -5,6 +5,7 @@ use crate::asm::arm64::*;
 use crate::cruby::*;
 use crate::backend::ir::*;
 use crate::virtualmem::CodePtr;
+use crate::utils::*;
 
 // Use the arm64 register type for this platform
 pub type Reg = A64Reg;
@@ -97,13 +98,13 @@ fn emit_jmp_ptr_with_invalidation(cb: &mut CodeBlock, dst_ptr: CodePtr) {
     #[cfg(not(test))]
     {
         let end = cb.get_write_ptr();
-        unsafe { rb_yjit_icache_invalidate(start.raw_ptr() as _, end.raw_ptr() as _) };
+        unsafe { rb_yjit_icache_invalidate(start.raw_ptr(cb) as _, end.raw_ptr(cb) as _) };
     }
 }
 
 fn emit_jmp_ptr(cb: &mut CodeBlock, dst_ptr: CodePtr, padding: bool) {
-    let src_addr = cb.get_write_ptr().into_i64();
-    let dst_addr = dst_ptr.into_i64();
+    let src_addr = cb.get_write_ptr().as_offset();
+    let dst_addr = dst_ptr.as_offset();
 
     // If the offset is short enough, then we'll use the
     // branch instruction. Otherwise, we'll move the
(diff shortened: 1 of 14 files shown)


yjit/src/virtualmem.rs
```diff
@@ -57,14 +57,39 @@ pub trait Allocator {
...
 
-/// Pointer into a [VirtualMemory].
-/// We may later change this to wrap an u32.
-/// Note: there is no NULL constant for CodePtr. You should use Option<CodePtr> instead.
+/// Pointer into a [VirtualMemory] represented as an offset from the base.
+/// Note: there is no NULL constant for [CodePtr]. You should use `Option<CodePtr>` instead.
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Debug)]
 #[repr(C, packed)]
-pub struct CodePtr(NonNull<u8>);
+pub struct CodePtr(u32);
```
k0kubun, line 64: I believe this would effectively revert https://github.com/ruby/ruby/pull/6792. is that intentional?

XrXr: It's fine, since the optimization from `NonNull` was only important when we stored `Option<CodePtr>` in a lot of places. We've stopped wrapping `CodePtr`s in the types we retain long term.

I was saw the size of `Block` and `Branch` go down before writing up the patch through rust-analyzer hover :)

yjit/src/backend/arm64/mod.rs
```diff
@@ -1023,14 +1024,18 @@ impl Assembler
...
-                    });
-
-                    mov(cb, out.into(), Self::SCRATCH0);
+                        mov(cb, out.into(), Self::SCRATCH0);
+                    } else {
+                        let target_code = target.unwrap_code_ptr();
+                        let target_addr = target_code.raw_addr(cb).as_u64();
+                        emit_load_value(cb, out.into(), target_addr);
+                    }
                 },
```
maximecb, line 1039: Can you add a bit of commenting here to explain each branch of the if-else statement is doing?

yjit/src/backend/ir.rs
```diff
@@ -447,9 +447,8 @@ pub enum Insn {
     // Add a label into the IR at the point that this instruction is added.
     Label(Target),
 
-    // Load effective address relative to the current instruction pointer. It
-    // accepts a single signed immediate operand.
-    LeaLabel { target: Target, out: Opnd },
+    /// Get the code address of a jump target
+    LeaJumpTarget { target: Target, out: Opnd },
```
maximecb, line 451: Can you explain why the renaming? (not criticizing, genuinely curious)

maximecb:
This is a big PR but the changes seem very clean so I would be inclined to merge it.

Only caveat is that we're dealing with some code GC bugs right now, so we should probably be careful about potentially introducing new bugs. Still inclined to merge but we should probably avoid major changes to the code GC after this.

yjit/src/backend/ir.rs
XrXr: >To derive a full raw pointer from a CodePtr, one needs a base pointer.
Both CodeBlock and VirtualMemory can be used for this purpose. The
base pointer is readily available everywhere, except for in the case of
the jit_return "branch". Generalize lea_label() to lea_jump_target()
in the IR to delay deriving the jit_return address until compile(),
when the base pointer is available.

Basically just so the `jit_return` branch can talk about full address with just access to an assembler (the code block is available only later).

maximecb (approved):

PR #8847, opened 2023-11-05
[DOC] RDoc for module Process

BurdetteLamar:
Adding to the module's RDoc with a view to helping the user in choosing among the ways to create a process.  Should this also mention backticks?  Or anything else?

### DIFF
--- a/process.c
+++ b/process.c
@@ -8792,8 +8792,7 @@ proc_warmup(VALUE _)
  *
  * - \Method Kernel#system executes a given command-line (string) in a subshell;
  *   returns +true+, +false+, or +nil+.
- * - \Method {Kernel#`}[rdoc-ref:Kernel@method-i-60]
- *   executes a given command-line (string) in a subshell,
+ * - \Method Kernel#` executes a given command-line (string) in a subshell,
  *   returns its $stdout string.
  * - \Module Open3 supports creating child processes
  *   with access to their $stdin, $stdout, and $stderr streams.
--- a/process.c
+++ b/process.c
@@ -8792,7 +8792,7 @@ proc_warmup(VALUE _)
  *
  * - \Method Kernel#system executes a given command-line (string) in a subshell;
  *   returns +true+, +false+, or +nil+.
- * - \Method Kernel#` executes a given command-line (string) in a subshell,
+ * - \Method Kernel#` executes a given command-line (string) in a subshell;
  *   returns its $stdout string.
  * - \Module Open3 supports creating child processes
  *   with access to their $stdin, $stdout, and $stderr streams.
--- a/process.c
+++ b/process.c
@@ -8784,44 +8784,19 @@ proc_warmup(VALUE _)
  * The choice of process or subshell depends on the form of the command;
  * see {Argument command_line or exe_path}[rdoc-ref:Process@Argument+command_line+or+exe_path].
  *
- * - Each of these methods executes a single command in a process or subshell,
- *   accepts a string for input to $stdin,
- *   and returns string output from $stdout, $stderr, or both:
- *
- *   - Open3.capture2: Executes the command;
- *     returns the string from $stdout.
- *   - Open3.capture2e: Executes the command;
- *     returns the string from merged $stdout and $stderr.
- *   - Open3.capture3: Executes the command;
- *     returns strings from $stdout and $stderr.
- *
- * - Each of these methods executes a single command in a process or subshell,
- *   and returns pipes for $stdin, $stdout, and/or $stderr:
- *
- *   - Open3.popen2: Executes the command;
- *     returns pipes for $stdin and $stdout.
- *   - Open3.popen2e: Executes the command;
- *     returns pipes for $stdin and merged $stdout and $stderr.
- *   - Open3.popen3: Executes the command;
- *     returns pipes for $stdin, $stdout, and $stderr.
- *
- * - Each of these methods executes one or more commands in processes and/or subshells,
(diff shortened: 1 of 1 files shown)


process.c
```diff
@@ -8779,16 +8779,58 @@ proc_warmup(VALUE _)
...
+ *     returns pipes for $stdin, $stdout, and $stderr.
+ *
+ * - Each of these methods executes one or more commands in processes and/or subshells,
+ *   returns pipes for the first $stdin, the last $stdout, or both:
+ *
+ *   - Open3.pipeline_r: Returns a pipe for the last $stdout.
+ *   - Open3.pipeline_rw: Returns pipes for the first $stdin and the last $stdout.
+ *   - Open3.pipeline_w: Returns a pipe for the first $stdin.
+ *   - Open3.pipeline_start: Does not wait for processes to complete.
+ *   - Open3.pipeline: Waits for processes to complete.
```
peterzhu2118, line 8815: `Open3` is not part of `Process`, so I don't think this should belong in the documentation for `Process`.

process.c
```diff
@@ -8779,16 +8779,58 @@ proc_warmup(VALUE _)
...
+ *   returns +true+, +false+, or +nil+.
+ *
+ * Each of the methods above accepts:
+ *
+ * - An optional hash of environment variable names and values;
+ *   see {Exection Environment}[rdoc-ref:Process@Execution+Environment].
+ * - A required string argument that is a +command_line+ or +exe_path+;
+ *   see {Argument command_line or exe_path}[rdoc-ref:Process@Argument+command_line+or+exe_path].
+ * - An optional hash of execution options;
+ *   see {Execution Options}[rdoc-ref:Process@Execution+Options].
```
peterzhu2118, line 8833: I think this kind of detail should belong in the documentation for each respective method, rather than here where it's just an overview.

peterzhu2118 (commented):
> Should this also mention backticks?

I think it would make sense to mention backticks since it also mentions `Kernel#system`.

process.c
BurdetteLamar: I tend to agree, but I think somewhere we should be giving this overview of process creation, to help the users understand and choose.  I'd thought about doc/process.rdoc as a home for this, but the search engines love our classes and hate our doc pages.  Should this material be in doc/ anyway?

peterzhu2118: Rather than going into details about `Open3`, I think this level of detail should belong in `Open3` and then in `Process` just briefly state that `Open3` has convenient methods for shelling out.

BurdetteLamar: Done;  Kernel#` mention also added.

process.c
peterzhu2118: @BurdetteLamar can you address this comment too?

BurdetteLamar: Removed here.  I think the three methods already have the details, so no changes there.

BurdetteLamar:
@peterzhu2118, ready for you.

process.c
```diff
@@ -8779,16 +8779,24 @@ proc_warmup(VALUE _)
...
+ * - Process.exec: Replaces the current process by executing the command.
  *
- * Each of these methods accepts:
+ * In addition:
  *
- * - An optional hash of environment variable names and values.
- * - An optional hash of execution options.
+ * - \Method Kernel#system executes a given command-line (string) in a subshell;
+ *   returns +true+, +false+, or +nil+.
+ * - \Method {Kernel#`}[file:///home/burdette/Documents/GitHub/ruby/html/Kernel.html#method-i-60]
```
peterzhu2118, line 8795: This link only works on your local computer.

BurdetteLamar: Fixed.

peterzhu2118: I don't see this as fixed. @BurdetteLamar did you forget to push?

BurdetteLamar: Duh.  Now fixed.

process.c
```diff
@@ -8779,16 +8779,24 @@ proc_warmup(VALUE _)
...
+ * - Process.exec: Replaces the current process by executing the command.
  *
- * Each of these methods accepts:
+ * In addition:
  *
- * - An optional hash of environment variable names and values.
- * - An optional hash of execution options.
+ * - \Method Kernel#system executes a given command-line (string) in a subshell;
+ *   returns +true+, +false+, or +nil+.
+ * - \Method {Kernel#`}[https://docs.ruby-lang.org/en/master/Kernel.html#method-i-60]
```
peterzhu2118, line 8795: We can do a direct `rdoc-ref` here since it's inside of Ruby (and not in a gem).

process.c
```diff
@@ -8779,16 +8779,24 @@ proc_warmup(VALUE _)
...
+ * - Process.exec: Replaces the current process by executing the command.
  *
- * Each of these methods accepts:
+ * In addition:
  *
- * - An optional hash of environment variable names and values.
- * - An optional hash of execution options.
+ * - \Method Kernel#system executes a given command-line (string) in a subshell;
+ *   returns +true+, +false+, or +nil+.
+ * - \Method {Kernel#`}[rdoc-ref:Kernel@method-i-60]
```
peterzhu2118, line 8795: This link doesn't work for me, but this does:

```suggestion
 * - \Method Kernel#`
```

BurdetteLamar: Present works for me, suggestion does not.  I'm running rdoc 6,3,1, you?

peterzhu2118: I'm on RDoc 6.5.0.

peterzhu2118 (approved):

PR #8849, opened 2023-11-06
Add "Optimization" section to regexp.rdoc

makenowjust:
This PR adds "Optimization" section to regexp.rdoc. The added contents are followed by #7931 changes.

@BurdetteLamar Could you polish this and the overall regexp.rdoc? I believe you are an expert on this file.

### DIFF
--- a/doc/regexp.rdoc
+++ b/doc/regexp.rdoc
@@ -1228,6 +1228,34 @@ when regexp.timeout is non-+nil+, that value controls timing out:
   |         nil          |         Float        | Times out in Float seconds. |
   |        Float         |          Any         | Times out in Float seconds. |
 
+== Optimization
+
+Regexp matching time may increase explosively depending on a pattern and an
+target string. A vulnerability caused by this is known as {regular expression
+denial-of-service (ReDoS)}[https://en.wikipedia.org/wiki/ReDoS].
+
+Ruby's regexp matching applies an optimization to prevent ReDoS. When the
+optimization is applied, matching time increases in linear to the input's
+length; that is ReDoS does not cause.  For applying this optimization, a
+pattern satisfy the following conditions:
+
+- no backreferences
+- no subexpression calls
+- no nested lookaround anchors and atomic groups
+- no nested quantifiers with counting (i.e. no nested <tt>{n}</tt>,
+  <tt>{min,}</tt>, <tt>{,max}</tt>, or <tt>{min,max}</tt> style quantifiers)
+
+For a regexp source or object, Regexp.linear_time? can check whether the above
+conditions are satisfied.
+
+  Regexp.linear_time?(/a*/)     # => true
+  Regexp.linear_time?('a*')     # => true
+  Regexp.linear_time?(/(a*)\1/) # => false
+
+However, you should not consider an untrusted source is safe for use even if
+this method returns +true+. Since we use memoization for this optimization,
+the untrusted source may invoke large memory consumption.
+
 == References
 
 Read (online PDF books):
--- a/doc/regexp.rdoc
+++ b/doc/regexp.rdoc
@@ -1230,31 +1230,30 @@ when regexp.timeout is non-+nil+, that value controls timing out:
 
 == Optimization
 
-Regexp matching time may increase explosively depending on a pattern and an
-target string. A vulnerability caused by this is known as {regular expression
-denial-of-service (ReDoS)}[https://en.wikipedia.org/wiki/ReDoS].
+For certain values of the pattern and target string,
+matching time can grow polynomially or exponentially in relation to the input size;
+the potential vulnerability arising from this is the {regular expression denial-of-service}[https://en.wikipedia.org/wiki/ReDoS] (ReDoS) attack.
 
-Ruby's regexp matching applies an optimization to prevent ReDoS. When the
-optimization is applied, matching time increases in linear to the input's
-length; that is ReDoS does not cause.  For applying this optimization, a
-pattern satisfy the following conditions:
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
On second thought, I think it's better to merge this (after tech review) and let me work on it in a new PR.  Makes sense?

BurdetteLamar:
Made a mistake.  I thought the suggested reviewers at page top were actually requested reviewers.  Since there are none, I will get on this very soon.

BurdetteLamar:
I have done `git clone -b regexp-doc-optimization https://github.com/makenowjust/ruby.git`, and so have the new file on my machine.  Is it appropriate for me to push from there (instead of reviewing the PR on Github)?  If not, what's the best way to proceed?

makenowjust:
@BurdetteLamar If you think it needs a major rewrite, you can create another branch/PR.

doc/regexp.rdoc
```diff
@@ -1228,6 +1228,34 @@ when regexp.timeout is non-+nil+, that value controls timing out:
   |         nil          |         Float        | Times out in Float seconds. |
   |        Float         |          Any         | Times out in Float seconds. |
 
+== Optimization
+
+Regexp matching time may increase explosively depending on a pattern and an
+target string. A vulnerability caused by this is known as {regular expression
+denial-of-service (ReDoS)}[https://en.wikipedia.org/wiki/ReDoS].
```
BurdetteLamar, line 1235: For certain values of the pattern and target string,
matching time can grow polynomially or exponentially in relation to the input size;
the potential vulnerability arising from this is the {regular expression denial-of-service}[https://en.wikipedia.org/wiki/ReDoS] (ReDoS) attack.

doc/regexp.rdoc
```diff
@@ -1228,6 +1228,34 @@ when regexp.timeout is non-+nil+, that value controls timing out:
...
+Ruby's regexp matching applies an optimization to prevent ReDoS. When the
+optimization is applied, matching time increases in linear to the input's
+length; that is ReDoS does not cause.  For applying this optimization, a
+pattern satisfy the following conditions:
+
+- no backreferences
+- no subexpression calls
+- no nested lookaround anchors and atomic groups
+- no nested quantifiers with counting (i.e. no nested <tt>{n}</tt>,
+  <tt>{min,}</tt>, <tt>{,max}</tt>, or <tt>{min,max}</tt> style quantifiers)
```
BurdetteLamar, line 1246: \Regexp matching can apply an optimization to prevent ReDoS attacks.
When the optimization is applied, matching time increases linearly (not polynomially or exponentially)
in relation to the input size, and a ReDoS attach is not possible.

This optimization is applied if and only if the pattern meets these criteria:

- No backreferences.
- No subexpression calls.
- No nested lookaround anchors or atomic groups.
- No nested quantifiers with counting (i.e. no nested <tt>{n}</tt>,
  <tt>{min,}</tt>, <tt>{,max}</tt>, or <tt>{min,max}</tt> style quantifiers).

doc/regexp.rdoc
```diff
@@ -1228,6 +1228,34 @@ when regexp.timeout is non-+nil+, that value controls timing out:
...
+
+For a regexp source or object, Regexp.linear_time? can check whether the above
+conditions are satisfied.
```
BurdetteLamar, line 1249: You can use method Regexp.linear_time? to determine whether a pattern meets these criteria:

doc/regexp.rdoc
```diff
@@ -1228,6 +1228,34 @@ when regexp.timeout is non-+nil+, that value controls timing out:
...
+
+  Regexp.linear_time?(/a*/)     # => true
+  Regexp.linear_time?('a*')     # => true
+  Regexp.linear_time?(/(a*)\1/) # => false
+
+However, you should not consider an untrusted source is safe for use even if
+this method returns +true+. Since we use memoization for this optimization,
+the untrusted source may invoke large memory consumption.
```
BurdetteLamar, line 1257: However, an untrusted source may not be safe even if the method returns +true+,
because the optimization uses memoization (which may invoke large memory consumption).

BurdetteLamar (commented):
Suggested minor edits throughout.

doc/regexp.rdoc
makenowjust: Generally, *polynomial* contains linear. However, this is not problematic.

doc/regexp.rdoc
makenowjust: "if and only if" sounds too strict. Actually, since some small counting repetition is unfolded internally, this case does not violate the conditions. The complete condition is too complicated to explain and detailed on the implementation. "if" seems better to me.

makenowjust:
Thank you @BurdetteLamar. Your suggestions are applied.

I would give you notice that ReDoS is also mentioned at the head of the "Timeout" section.
Maybe it is better to merge the "Timeout" and "Optimization" sections.

BurdetteLamar:
> Thank you @BurdetteLamar. Your suggestions are applied.
> 
> I would give you notice that ReDoS is also mentioned at the head of the "Timeout" section. Maybe it is better to merge the "Timeout" and "Optimization" sections.

Is that for a future PR?  Or for here?  If here, should I be looking at that?

makenowjust:
I thought that was for a future PR since this PR is completed in one topic. Sorry for my misleading.

@BurdetteLamar If there seems to be no problem, I will merge this. Thank you.

BurdetteLamar:
> I thought that was for a future PR since this PR is completed in one topic. Sorry for my misleading.
> 
> @BurdetteLamar If there seems to be no problem, I will merge this. Thank you.

Looks good to me.

PR #8850, opened 2023-11-06
generic_ivar_set: properly check for TOO_COMPLEX on capacity transition

casperisfine:
cc @peterzhu2118


peterzhu2118 (commented):
🤦 Thanks for catching this typo!

PR #7097, opened 2023-01-10
YJIT: Inline basic Ruby methods

k0kubun:
## Changes
This PR inlines method calls of ISEQs that have only: `putnil`, `putobject`, `putself`, or `leave`.

## Background
Sometimes such simple methods are called. `Symbol#to_sym` is one of such examples, which consists of 1% of all Ruby calls in railsbench https://github.com/Shopify/yjit-bench/pull/163. Such small ISEQs are faster in Ruby than in C on the interpreter, so we generally rewrite them to Ruby https://github.com/ruby/ruby/pull/6683 (more examples are in `ruby/ruby/*.rb`). Unlike MJIT, YJIT currently doesn't have a way to inline such Ruby methods, and this PR implements that.

Another way to implement this would be to convert such Ruby methods to C and use cfunc specialized codegen. But the downside is that it would slow down the interpreter and we can't do that for methods defined in the Rails side, such as `Object#html_safe?` and `Object#present?`.

## Design
* `putnil`, `putobject`, `putself`, or `leave` are the targets because:
  * They never side-exist (besides check ints on leave) and they're "leaf" (i.e. they don't call arbitrary methods).
  * We could include some other insns as well, but I assume most methods using such insns would involve other insns that may side-exit.
* On `opt_send_without_block`, jump to the callee block while keeping the receiver and the arguments on the stack.
  * The receiver is kept for `putself`. We could eliminate a few instructions if we write the return value there if `putself` is not used, but I made the implementation simple, assuming the benefit of that optimization would be trivial.
  * Arguments are kept on the stack so that we could use them for `getlocal_WC_0` later. Out of scope in this PR.
* The callee jumps back to the caller. Normally, both jumps are optimized away by patching on `gen_direct_jmp`.
* Always inline such ISEQs since they're fairly small. Ignore such inlined contexts from the versioning limit.

## Example
```rb
def symbol(sym)
  sym.to_sym
end
symbol(:sym)
```

### Before
<details>

```asm
  # ...
  # Block: symbol@/home/k0kubun/tmp/a.rb:2 (ISEQ offset: 0)
  # Insn: getlocal_WC_0
  0x5575c4639059: mov rax, qword ptr [r13 + 0x20]
  0x5575c463905d: mov rax, qword ptr [rax - 0x18]
  0x5575c4639061: mov qword ptr [rbx], rax
  # regenerate_branch
  # Block: symbol@/home/k0kubun/tmp/a.rb:2 (ISEQ offset: 2)
[...]

PR #8855, opened 2023-11-07
YJIT: Inline basic Ruby methods

k0kubun:
This PR optimizes calls of an ISEQ that consists of only `putnil`/`putobject` and `leave` into a single `mov` instruction. By specializing the inlining logic for such cases, it simplifies the implementation more than the previous attempt https://github.com/ruby/ruby/pull/7097.

## Benchmark
erubi-rails seems to have a higher ratio of such method calls than other headline benchmarks:

```
num_send_inline:          18,258,990 ( 3.8%)
```

Here's a benchmark result on erubi-rails:

```
before: ruby 3.3.0dev (2023-11-06T19:39:09Z master 49b6dc8f07) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-11-07T01:17:48Z yjit-basic-inlining a8ea580d9d) +YJIT [x86_64-linux]

erubi-rails  1613.3       1.0         1567.3      1.4         1.02           1.03
```

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4196,3 +4196,24 @@ assert_equal '[6, -6, 9671406556917033397649408, -9671406556917033397649408, 212
 
 # Integer multiplication and overflow (minimized regression test from test-basic)
 assert_equal '8515157028618240000', %q{2128789257154560000 * 4}
+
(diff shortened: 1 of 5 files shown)


maximecb (approved):
Very well done. This is great! :)

PR #8857, opened 2023-11-07
YJIT: handle out of shape situation in gen_setinstancevariable

casperisfine:
If the VM ran out of shape, `rb_shape_transition_shape_capa` might return `OBJ_TOO_COMPLEX_SHAPE`.

cc @maximecb @XrXr @k0kubun 

NB: this case is tested in `test/ruby/test_shapes.rb`, it's unclear to me why it didn't catch that YJIT bug. I tried to figure out how to make it fail, but couldn't figure it out.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2473,7 +2473,11 @@ fn gen_setinstancevariable(
         };
 
         let dest_shape = if let Some(capa_shape) = capa_shape {
-            unsafe { rb_shape_get_next(capa_shape, comptime_receiver, ivar_name) }
+            if OBJ_TOO_COMPLEX_SHAPE_ID == unsafe { rb_shape_id(capa_shape) } {
+              capa_shape
+            } else {
+              unsafe { rb_shape_get_next(capa_shape, comptime_receiver, ivar_name) }
+            }
         } else {
             unsafe { rb_shape_get_next(shape, comptime_receiver, ivar_name) }
         };


yjit/src/codegen.rs
```diff
@@ -2473,7 +2473,11 @@ fn gen_setinstancevariable(
         };
 
         let dest_shape = if let Some(capa_shape) = capa_shape {
```
maximecb, line 2475: Can you add some comments explaining what's going on a bit? Not your fault, but this code is becoming hard to make sense of. I stared at it for a good five minutes trying to make sense of what is `capa_shape` representing.

I'm also confused as to why we need to call so many C functions to get the next shape ID. Why can't `rb_get_shape_next` automatically return the same too complex shape if the object is already too complex?

@jemmaissroff I imagine you have other things to think about right now, but: it feels like there should be a C function that encapsulates all (or most of) the logic for "give me the next shape". Most of that logic should be handled in the shape code on the C side rather than in YJIT. Just because it would be tidier for that logic to be grouped with the rest of the shape-related code. What do you think?

casperisfine: > Why can't `rb_get_shape_next` automatically return the same too complex shape if the object is already too complex?

Yeah, I want to do such refactor. I also want `rb_get_shape_next` to do the `capa` transition as well, the tricky part if that when this happens you need to rewrite the object, so we probably need an API that get the object.

Here I tried to just fix the issue with the smallest change possible. But I'm happy to refactor in a followup.

casperisfine: Here's the refactor I started doing but put on the side because it was becoming too big for a fix PR: https://github.com/Shopify/ruby/commit/dea6c4db8850eb2f23b5a0deb0995046aaf8ebf4

maximecb: Understood. I will merge this PR so the bug is fixed. It does seem problematic that we didn't catch it in `test_shapes` though. It sounds like we need more tests (and specifically tests for shapes + YJIT).

PR #8858, opened 2023-11-07
Fix concurrent calls to ObjectSpace.dump_all crashing

KJTsanaktsidis:
This should fix https://bugs.ruby-lang.org/issues/19922 I believe. Since calling `mark_func` can yield the GVL, and set `cr->mfd` to something else, we need to set it back.

Fixes: #19922

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -11950,19 +11950,23 @@ objspace_reachable_objects_from_root(rb_objspace_t *objspace, void (func)(const 
 {
     if (during_gc) rb_bug("objspace_reachable_objects_from_root() is not supported while during_gc == true");
 
-    rb_ractor_t *cr = GET_RACTOR();
-    struct root_objects_data data = {
-        .func = func,
-        .data = passing_data,
-    };
-    struct gc_mark_func_data_struct mfd = {
-        .mark_func = root_objects_from,
-        .data = &data,
-    }, *prev_mfd = cr->mfd;
+    RB_VM_LOCK_ENTER();
+    {
+        rb_ractor_t *cr = GET_RACTOR();
+        struct root_objects_data data = {
+            .func = func,
+            .data = passing_data,
+        };
+        struct gc_mark_func_data_struct mfd = {
+            .mark_func = root_objects_from,
+            .data = &data,
+        }, *prev_mfd = cr->mfd;
 
-    cr->mfd = &mfd;
-    gc_mark_roots(objspace, &data.category);
-    cr->mfd = prev_mfd;
+        cr->mfd = &mfd;
+        gc_mark_roots(objspace, &data.category);
+        cr->mfd = prev_mfd;
+    }
+    RB_VM_LOCK_LEAVE();
 }
 
 /*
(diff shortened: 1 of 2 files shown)


mame:
@KJTsanaktsidis @byroot Looks like our Ubuntu 22.04 CI has started failing since this change. Could you please check it out?

https://rubyci.s3.amazonaws.com/ubuntu2204/ruby-master/log/20231112T183003Z.fail.html.gz
```
  1) Failure:
TestGCCompact#test_moving_hashes_down_size_pools [/home/chkbuild/chkbuild/tmp/build/20231112T183003Z/ruby/test/ruby/test_gc_compact.rb:442]:
Expected 499 to be >= 500.
```

KJTsanaktsidis:
@mame I'll have a look tonight. The test in question has this comment in it:

```
    # This test fails on Solaris SPARC with the following error and I can't figure out why:
    #   TestGCCompact#test_moving_hashes_down_size_pools
    #   Expected 499 to be >= 500.
    omit if /sparc-solaris/ =~ RUBY_PLATFORM
```

which makes me think I'm going to have a fun time with this.

More generally - this is really not the first time one of my PR's has been merged and then it turns out it breaks things on some particular platform on rubyci.org.... is there something I should be doing to get this kind of feedback before merging? Or is this just an "expected" part of the ruby development process?

mame:
In terms of this particular issue, the test may need reconsideration. @peterzhu2118 What do you think?

On a general point. I think this is somewhat "expected". Some post-commit follow-up is desperately necessary. You may hope to see RubyCI results before the merge, but it's not nearly feasible in terms of computing resources. Also, even if all CIs pass, bug reports often come in later.

BTW, I do not intend to blame you. Almost all interpreter tests are integration tests, so it is quite common for unrelated tests to start failing. However, someone still needs to stop the error. And the initial response to the error will then be made by the person who made the change that somehow triggered the error.

byroot:
I'm reverting in https://github.com/ruby/ruby/pull/8908 in the meantime. I prefer a known low criticity bug than an unknown issue when we're getting close to the final release like this.

@peterzhu2118 will be away for most of the week, he's at Ruby Conf.

As mentioned in https://github.com/ruby/ruby/pull/8908 , I think `dump_all` should just not allow the GVL to be released.

KJTsanaktsidis:
> However, someone still needs to stop the error. And the initial response to the error will then be made by the person who made the change that somehow triggered the error

Of course - I’m just sorry I didn’t check rubyci myself this morning and you had to come and point it out to me!

reverting seems like the right thing to do for now. I’ll try and find some time to look at that test this week regardless. The fact that it was apparently already failing on Solaris is a good clue, I think those sparc machines are just slow enough to trigger race conditions that other architectures don’t (they pointed out a real issue with some threading changes I made earlier this year, for example).

as far as locking the VM & getting a consistent shapshot goes - I don’t really mind either way. Nobody should be caring about any kind of performamce when using these APIs, so it should be fine to just lock. I’m happy to try another PR doing that later (if i can get to the bottom of the hash compact tests first)

byroot:
> Nobody should be caring about any kind of performamce when using these APIs

I wouldn't go as far as that, because `dump_all` should try to be reasonably fast, but locking the VM and being slow are two different things.

PR #8859, opened 2023-11-07
[PRISM] Implement compilation for different parameters

jemmaissroff:
This commit compiles most parameter types, setting appropriate values on the ISEQ_BODY. It also adds tests for callers and callees of methods, using many versions of tests from bootstraptest


jemmaissroff:
This PR also fixes [this failure](http://ci.rvm.jp/results/trunk-iseq_binary@ruby-sp2-docker/4760294)

PR #8862, opened 2023-11-07
Refactor `rb_shape_transition_shape_capa` out

casperisfine:
Followup: https://github.com/ruby/ruby/pull/8857

Right now the `rb_shape_get_next` shape caller need to first check if there is capacity left, and if not call `rb_shape_transition_shape_capa` before it can call `rb_shape_get_next`.

And on each of these it needs to checks if we got a TOO_COMPLEX back.

All this logic is duplicated in the interpreter, YJIT and RJIT.

Instead we can have `rb_shape_get_next` do the capacity transition when needed. The caller can compare the old and new shapes capacity to know if resizing is needed. It also can check for TOO_COMPLEX only once.

@peterzhu2118 @maximecb @k0kubun @XrXr @jemmaissroff

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2450,45 +2450,33 @@ fn gen_setinstancevariable(
         None
     };
 
-    // Get the next shape information if it needs transition
+    // The current shape doesn't contain this iv, we need to transition to another shape.
     let new_shape = if !shape_too_complex && receiver_t_object && ivar_index.is_none() {
-        let shape = comptime_receiver.shape_of();
+        let current_shape = comptime_receiver.shape_of();
+        let next_shape = unsafe { rb_shape_get_next(current_shape, comptime_receiver, ivar_name) };
+        let next_shape_id = unsafe { rb_shape_id(next_shape) };
 
-        let current_capacity = unsafe { (*shape).capacity };
-
-        // If the object doesn't have the capacity to store the IV,
-        // then we'll need to allocate it.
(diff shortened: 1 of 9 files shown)


peterzhu2118 (approved):

casperisfine:
Somehow I broke RJIT :/

yjit/src/codegen.rs
```diff
@@ -2450,45 +2450,33 @@ fn gen_setinstancevariable(
...
     let new_shape = if !shape_too_complex && receiver_t_object && ivar_index.is_none() {
-        let shape = comptime_receiver.shape_of();
-
-        let current_capacity = unsafe { (*shape).capacity };
-
-        // If the object doesn't have the capacity to store the IV,
-        // then we'll need to allocate it.
-        let needs_extension = unsafe { (*shape).next_iv_index >= current_capacity };
+        let current_shape = comptime_receiver.shape_of();
+        let dest_shape = unsafe { rb_shape_get_next(current_shape, comptime_receiver, ivar_name) };
```
maximecb, line 2456: Small detail, but would use the name `next_shape` since the function is called `rb_shape_get_next`

maximecb (commented):
I feel like the YJIT side could probably be simplified further, but this seems like a good first step. Would be in favor of merging but I'll defer to @jemmaissroff as to whether this should be merged or not since she's directly worked on the shape code.

casperisfine:
> I feel like the YJIT side could probably be simplified further,

Without a doubt, but I'm very unfamiliar with YJIT so I'm very worried about attempting anything bigger :/

PR #8867, opened 2023-11-08
Remove no longer used RUBY_ABI_VERSION

nobu:


eregon:
Sharing my thoughts about this here too in addition to Slack:

Users of the ABI version are e.g. gems testing ruby-head using ruby/setup-ruby. If they have many CI runs and lots of gems to install it's a significant overhead on their CI and on [RubyGems.org](http://RubyGems.org) to always reinstall all gems vs using the cache.
If there is no ABI version we can only use the git commit, which is quite wasteful, do we want to waste resources like this for little effort to maintain the ABI version? TruffleRuby maintains a correct ABI version for years now, it's very little overhead, especially if it doesn't change often.
It needs an approach where it's explicitly tracked in CI though, whenever a `include/` file changes, either the ABI version is bumped or another file is bumped to acknowledge it's ABI-compatible.

e.g. in TruffleRuby, if a header changes, then one must either bump https://github.com/oracle/truffleruby/blob/master/lib/cext/ABI_version.txt or https://github.com/oracle/truffleruby/blob/master/lib/cext/ABI_check.txt.
The check is at https://github.com/oracle/truffleruby/blob/5dbdd3187e8a03435abe4560f9472a7b835f3073/tool/jt.rb#L3121

Other users of the ABI versions are Rubyists using `ruby-build 3.3-dev/ruby-dev` etc. Those risk segfaults/memory corruption if gems are reused due to CRuby not setting a correct ABI version (i.e. same "version" for ABI-imcompatible changes).

There is more reasoning at https://bugs.ruby-lang.org/issues/18249. Removing the ABI version breaks all of these.
So if this is removed, I think the least is to create a ticket at https://bugs.ruby-lang.org/ and explain how all the issues of https://bugs.ruby-lang.org/issues/18249 will be solved then.

nurse:
I heard Shopify once used RUBY_ABI_VERSION, but after some practice they give up to use it because what people can do is only clean and build when they hit ABI incompatibility.
Do you use RUBY_ABI_VERSION and how it makes your life easier?

eregon:
> because what people can do is only clean and build when they hit ABI incompatibility.

They'll probably hit a segfault in such cases, with the ABI version it's a nice and clean error instead.
If it's not someone working on CRuby I think they would be most confused to get a segfault, and even if they are it might still be confusing and waste a fair amount of time until they figure out what it is.
So it seems a loss-loss situation to remove it. The added complexity is negligible.

Also note that Bundler uses the ABI version automatically e.g. with `path` set (often set to vendor/bundle).
So as long as the version is correctly maintained, you cannot hit a ABI incompatibility segfault when using Bundler with path, which is a really nice property, which is lost without the ABI version.

It's really easy to maintain the ABI version correctly, we have been doing it for years in TruffleRuby. It can easily be checked in CI. I don't understand why some CRuby devs want to remove this so much, I think they largely overestimate the costs. It would also be possible to automatically compute an ABI version based on all include/ files, which once integrated really has 0 cost.
So why not do that and prefer to let Rubyists using ruby-head to get confusing segfaults instead?
(which might also mean more segfault reports at https://bugs.ruby-lang.org/).

> Do you use RUBY_ABI_VERSION and how it makes your life easier?

Probably no one uses the RUBY_ABI_VERSION macro, but many places use the ABI version accessible via `RbConfig::CONFIG['ruby_version']`. Merging this PR means that ABI version becomes incorrect for ruby-head, while being the correct thing to use for releases. So in that view, it breaks every usage of `RbConfig::CONFIG['ruby_version']`, which is many: https://github.com/search?q=RbConfig%3A%3ACONFIG%5B%27ruby_version%27%5D+OR+RbConfig%3A%3ACONFIG%5B%22ruby_version%22%5D&type=code

eregon:
So I see 2 big problems if the ABI version is removed:
* ruby/setup-ruby bundler caching for ruby-head will need to use the git commit as the ABI version. This is very wasteful and costs money to rubygems.org. Do you realize merging this PR has that impact?
* Dealing with native extensions + unreleased ruby versions is painful, most notably it takes a long time to recompile them and without an ABI version there is no way to cache it efficiently (i.e. using the commit hash is very inefficient although better than nothing). This is notably the 3rd major problem mentioned in https://github.com/Shopify/ruby/issues/525#issuecomment-1634964830. Both productivity for developers using ruby-head and CI/staging could benefit significantly from this. Benchmarking with ruby-head and many gems is also much more expensive and/or much slower due to that.

Let me cc @k0kubun @casperisfine @byroot and @peterzhu2118 so they are aware of this PR.

Of course we need to fix so the ABI version is tracked properly, e.g., using one of the two solutions described above.
I can contribute the one with a CI check + bumping the ABI version file or the ABI check file.

PR #8868, opened 2023-11-08
Free everything at shutdown

HParker:
when the --free-on-shutdown flag is enabled, manually free memory at shutdown.


mame:
@HParker Please add `static` for internal function definitions.

```
  Checking leaked global symbols...leaked
    free_shared_fiber_pool
    free_global_enc_table
    free_warning
    free_encoded_insn_data
    free_loaded_features_index
    free_compat_allocator_table
    free_default_rand_key
    free_static_symid_str
    free_transcoder_table
    free_rb_global_tbl
    free_generic_iv_tbl_
    free_vm_opt_tables
    free_on_shutdown
```

HParker:
oh! I misunderstood the failure and added `rb_` infront of the names 🤦 thank you! I'll switch to static instead.

HParker:
Actually @mame, am I able to use `static`? I think the functions are declared in different files so C will complain if I use `static` right? maybe there is a way around this I don't know about?

mame:
@HParker Sorry, if they need to refer across files, it is OK to add `rb_` prefix.

internal/vm.h
```diff
@@ -83,6 +83,23 @@ void rb_check_stack_overflow(void);
...
 
+extern bool rb_free_on_shutdown;
+
+void rb_free_default_rand_key(void);
+void rb_free_encoded_insn_data(void);
+void rb_free_environ(void);
+void rb_free_global_enc_table(void);
+// rb_free_loaded_builtin_table needs a week definition in case minibuiltin.c is not used
+// this is required for the wasm build.
+void rb_free_loaded_builtin_table() __attribute__((weak));
```
HParker, line 94: Looks like `__attribute__((weak))` is not supported in Windows, but removing this break the wasm build.

Maybe there is another builtin.c equivalent place I need to define this so that wasm has it and Windows is happy?

encoding.c
```diff
@@ -71,6 +71,23 @@ static struct enc_table {
...
 
+static int
+enc_names_free_i(st_data_t name, st_data_t idx, st_data_t args)
+{
+    ruby_xfree((void *)name);
+    return ST_DELETE;
+}
+
+void
+rb_free_global_enc_table(void) {
```
peterzhu2118, line 82: Nit:

```suggestion
rb_free_global_enc_table(void)
{
```

internal/vm.h
```diff
@@ -83,6 +83,23 @@ void rb_check_stack_overflow(void);
 extern uint64_t rb_vm_insns_count;
 #endif
 
+extern bool rb_free_on_shutdown;
+
+void rb_free_default_rand_key(void);
```
peterzhu2118, line 88: Can we move these function declarations to the header file for the respective C files rather than placing them all in `vm.h`?

symbol.c
```diff
@@ -531,6 +531,16 @@ register_sym(rb_symbols_t *symbols, VALUE str, VALUE sym)
     }
 }
 
+void
+rb_free_static_symid_str(void)
+{
+    GLOBAL_SYMBOLS_ENTER(symbols)
+        {
+            st_free_table(symbols->str_sym);
+        }
```
peterzhu2118, line 540: Nit:
```suggestion
    {
        st_free_table(symbols->str_sym);
    }
```

vm.c
```diff
@@ -2968,20 +2975,70 @@ free_loading_table_entry(st_data_t key, st_data_t value, st_data_t arg)
...
-            rb_fiber_reset_root_local_storage(th);
-            thread_free(th);
+            st_free_table(vm->static_ext_inits);
+            st_free_table(vm->ensure_rollback_table);
+
+            ruby_xfree(vm->postponed_job_buffer);
+            st_free_table(vm->defined_module_hash);
+
+            rb_id_table_free(vm->constant_cache);
+        } else {
```
peterzhu2118, line 3033: Nit:

```suggestion
        }
        else {
```

vm.c
```diff
@@ -2995,6 +3055,16 @@ ruby_vm_destruct(rb_vm_t *vm)
         }
         RB_ALTSTACK_FREE(vm->main_altstack);
         if (objspace) {
+            if (rb_free_on_shutdown) {
+                rb_objspace_free_objects(objspace);
+                rb_free_generic_iv_tbl_();
+                if (th) {
+                    xfree(nt);
```
HParker, line 3062: on Mac, this fails with,
```
pointer being freed was not allocated
```

but I am confused since it seems like,

```
vm->ractor.main_thread
```

is always `ZALLOC`ed in `Init_BareVM`.

test-code.rb
```diff
@@ -0,0 +1 @@
+defined? Foo.foo()
```
peterzhu2118, line 1: You probably didn't want to commit this right?

HParker: oops! thanks!

internal/vm.h
nobu: @mame It seems fine.
Just one thing, don't use pre-ANSI-style declaration.

missing/setproctitle.c
```diff
@@ -135,6 +135,16 @@ compat_init_setproctitle(int argc, char *argv[])
...
 #ifndef HAVE_SETPROCTITLE
+
+void
+rb_free_environ(void) {
+    size_t i;
+    char **envp = environ;
+    for (i = 0; envp[i] != NULL; i++)
+        free(environ[i]);
+    free(environ);
+}
```
nobu, line 146: - `environ` is copied conditionally.
- Why not using `envp` inside the loop?
- Adjust the style.

```suggestion
rb_free_environ(void)
{
#if SPT_TYPE == SPT_REUSEARGV
	size_t i;
	char **envp = environ;
	for (i = 0; envp[i] != NULL; i++)
		free(envp[i]);
	free(environ);
#endif
}
```

variable.c
```diff
@@ -454,6 +454,31 @@ struct rb_global_entry {
...
 
+static enum rb_id_table_iterator_result
+free_global_entry_i(ID key, VALUE val, void *arg)
+{
+    if (((struct rb_global_entry *)val)->var->counter == 1) {
+        ruby_xfree(((struct rb_global_entry *)val)->var);
+    } else {
+        ((struct rb_global_entry *)val)->var->counter--;
+    }
+    ruby_xfree((struct rb_global_entry *)val);
```
nobu, line 465:
```suggestion
    struct rb_global_entry *entry = (struct rb_global_entry *)val;
    if (entry->var->counter == 1) {
        ruby_xfree(entry->var);
    }
    else {
        entry->var->counter--;
    }
    ruby_xfree(entry);
```

HParker:
I tested this branch with a Rails app, and I get a double free from freeing environ,

```
==76982== Invalid free() / delete / delete[] / realloc()
==76982==    at 0x483CA3F: free (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==76982==    by 0x4E3BD7B: tdestroy_recurse (tsearch.c:767)
==76982==    by 0x4E3BD7B: tdestroy_recurse (tsearch.c:766)
==76982==    by 0x4E3BD7B: tdestroy_recurse (tsearch.c:764)
==76982==    by 0x4E3BD7B: tdestroy_recurse (tsearch.c:764)
==76982==    by 0x4E3BD7B: tdestroy_recurse (tsearch.c:766)
==76982==    by 0x4E3BD7B: tdestroy (tsearch.c:780)
==76982==    by 0x4EB98B5: free_mem (in /usr/lib/x86_64-linux-gnu/libc-2.31.so)
==76982==    by 0x4EB9B41: __libc_freeres (in /usr/lib/x86_64-linux-gnu/libc-2.31.so)
==76982==    by 0x48331C6: _vgnU_freeres (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_core-amd64-linux.so)
==76982==    by 0x4D679C1: __run_exit_handlers (exit.c:132)
==76982==    by 0x4D67A5F: exit (exit.c:139)
==76982==    by 0x4D45089: (below main) (libc-start.c:342)
==76982==  Address 0x303435b0 is 0 bytes inside a block of size 36 free'd
==76982==    at 0x483CA3F: free (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==76982==    by 0x462E3B: rb_free_environ (setproctitle.c:146)
==76982==    by 0x35AD16: ruby_vm_destruct (vm.c:2995)
==76982==    by 0x15AFDE: rb_ec_cleanup (eval.c:268)
==76982==    by 0x15B868: ruby_run_node (eval.c:328)
==76982==    by 0x156206: rb_main (main.c:39)
==76982==    by 0x156206: main (main.c:58)
==76982==  Block was alloc'd at
==76982==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==76982==    by 0x4D67368: __add_to_environ (setenv.c:215)
==76982==    by 0x484459F: setenv (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==76982==    by 0x19AE5F: ruby_setenv (hash.c:5134)
==76982==    by 0x19B19C: env_aset (hash.c:5321)
==76982==    by 0x19B19C: env_aset_m (hash.c:5302)
==76982==    by 0x33E5AA: vm_call_cfunc_with_frame_ (vm_insnhelper.c:3504)
==76982==    by 0x351682: vm_sendish (vm_insnhelper.c:5582)
==76982==    by 0x351682: vm_exec_core (insns.def:822)
==76982==    by 0x34269D: rb_vm_exec (vm.c:2481)
==76982==    by 0x345F89: invoke_block (vm.c:1499)
==76982==    by 0x345F89: invoke_iseq_block_from_c (vm.c:1569)
==76982==    by 0x345F89: invoke_block_from_c_bh (vm.c:1587)
==76982==    by 0x34694F: vm_yield_with_cref (vm.c:1624)
==76982==    by 0x34694F: vm_yield (vm.c:1632)
==76982==    by 0x34694F: rb_yield_0 (vm_eval.c:1362)
==76982==    by 0x34694F: rb_yield (vm_eval.c:1378)
==76982==    by 0x46BBBB: rb_ary_each (array.c:2532)
==76982==    by 0x33E5AA: vm_call_cfunc_with_frame_ (vm_insnhelper.c:3504)
```

I am going to stop freeing environ which means it will show up as a leak, but it allows testing with less trivial Ruby scripts. I would love a better solution, but I am not sure how to determine when/if something in environ is "owned" by the vm.

Edwing123:
I would've liked to see `FREE_ON_SHUTDOWN` being prefixed with `RB_`. So it's easy to see that is related to Ruby.

Edwing123:
I regret saying `RB_`, let's better use `RUBY_`. what do you think? @peterzhu2118.

HParker:
I looked around the code and found more examples of `getenv("RUBY_*" so I chose that as the prefix. Happy to change it, but that seems like the most consistent name.

vm.c
```diff
@@ -2968,20 +2977,67 @@ free_loading_table_entry(st_data_t key, st_data_t value, st_data_t arg)
...
+                    rb_id_table_free(cursor->edges);
+                cursor += 1;
+            }
+
+            xfree(GET_SHAPE_TREE());
+
+            st_free_table(vm->static_ext_inits);
+            st_free_table(vm->ensure_rollback_table);
+
+            ruby_xfree(vm->postponed_job_buffer);
```
KJTsanaktsidis, line 3027: While working on another issue with the postponed job infrastructure, I realised there's something a little bit annoying about doing this... It's the most miniscule of corner cases, but I figure I'll bring it up.

If some other gem has registered a signal handler, it's possible a signal might arrive after this and before the process exits, the gem's signal signal handler could then try and call `rb_register_postponed_job`, which will be a use-after-free here.

Stackprof in particular _tries_ to get this right, and almost does: it registers a `ruby_vm_at_exit` hook here: https://github.com/tmm1/stackprof/blob/91d12e77b3d3c5014870b4ee7c19eab24ad04b5b/ext/stackprof/stackprof.c#L903 to disable itself.

There are just a couple teeny tiny problems though:

The first problem is, that the free-on-shutdown stuff is happening before `ruby_vm_run_at_exit_hooks`. Easy fix, let's hoist that further up I suppose? Or push this stuff further down?

The second problem is what happens in the presence of non-ruby-managed threads. If there are other native threads in the process, a signal handler could be in-progress on that thread already when `ruby_vm_run_at_exit_hooks` is called, so setting `ruby_vm_running = 0;` wouldn't stop it from happening.

I actually can't think of anything at all we can do about that though. If there was a way to run something at exit after all _other_ threads have been destroyed, we could do that, but there isn't such a thing. I guess well-behaved extensions need to make sure they `pthread_join` all their threads in a `ruby_vm_at_exit` handler?

mame:
@HParker @peterzhu2118 @byroot Matz suggested `RUBY_FREE_AT_EXIT` instead of `ON_EXIT` because there is `Kernel#at_exit`. Can we fix it?

HParker:
@mame oops! 🤦 sorry! I even named the PR correctly, but must have got mixed up on the name in the PR. I opened this to fix it: https://github.com/ruby/ruby/pull/9302

PR #8872, opened 2023-11-08
String#force_encoding don't clear coderange if encoding is unchanged

casperisfine:
Some code out there blind calls `force_encoding` without checking what the original encoding was, which clears the coderange uselessly.

If the String is big, it can be a rather costly mistake.

For instance the `rack-utf8_sanitizer` gem does this on request bodies.

### DIFF
--- a/string.c
+++ b/string.c
@@ -10843,7 +10843,23 @@ static VALUE
 rb_str_force_encoding(VALUE str, VALUE enc)
 {
     str_modifiable(str);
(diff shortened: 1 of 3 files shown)


peterzhu2118 (approved):

byroot:
I don't get the failures, looks like a parser bug but it's weird. I'll dig more tomorrow.

nobu (approved):

PR #8873, opened 2023-11-08
[PRISM] Implement compilation for ForwardingArgumentsNode

jemmaissroff:
Closes https://github.com/ruby/prism/issues/1660

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1413,6 +1413,14 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
 
                   break;
               }
+              case PM_FORWARDING_ARGUMENTS_NODE: {
+                  orig_argc++;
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -1413,6 +1413,14 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
 
                   break;
               }
+              case PM_FORWARDING_ARGUMENTS_NODE: {
+                  orig_argc++;
+                  *flags |= VM_CALL_ARGS_BLOCKARG | VM_CALL_ARGS_SPLAT;
+                  ADD_GETLOCAL(ret, &dummy_line_node, 3, 0);
```
kddnewton, line 1419: These depths/indices being hardcoded is going to work in some scenarios but not in others.

For the depths:

```ruby
def foo(...) = tap { bar(...) }
```

for the indices:

```ruby
def foo(a, b, c, ...) = bar(...)
```

We'll probably need to lookup the local index for `*` and `&`.

PR #8875, opened 2023-11-08
Remove SHAPE_CAPACITY_CHANGE shapes

peterzhu2118:
We don't need to create a shape to transition capacity as we can transition the capacity when the capacity of the SHAPE_IVAR changes.

cc. @byroot @jemmaissroff


byroot (approved):
Love it 👍

PR #8877, opened 2023-11-09
Ensure f(**kw, &block) calls kw.to_hash before block.to_proc

jeremyevans:
Previously, block.to_proc was called first, by vm_caller_setup_arg_block. kw.to_hash was called later inside CALLER_SETUP_ARG or setup_parameters_complex.

This adds a splatkw instruction that is inserted before all sends with ARGS_BLOCKARG and KW_SPLAT and without KW_SPLAT_MUT. This is not needed in the KW_SPLAT_MUT case, because then you know the value is a hash, and you don't need to call to_hash on it.

The splatkw instruction checks whether the second to top block is a hash, and if not, replaces it with the value of calling to_hash on it (using rb_to_hash_type).  As it is always before a send with ARGS_BLOCKARG and KW_SPLAT, second to top is the keyword splat, and top is the passed block.

This is implemented via the peephole optimizer, because it was easier to ensure correct behavior there compared to having it in the compiler proper.

This currently fails YJIT tests because YJIT panics instead of falling back to the interpreter for instructions it doesn't recognize.


ko1 (approved):

KJTsanaktsidis:
I think this is another manifestation of https://bugs.ruby-lang.org/issues/20021

The tl;dr of the problem is that the compaction tests create a bunch of objects and expect them all to be moved. But, if one has a live reference on the machine stack (even if it's a "live" reference in uninitialized stack memory), then it is of course pinned and can't be moved.

I wrote up a patch to count the number of pinned objects during compaction, and do the assertion on `(moved + pinned) >= num_objects_created` rather than `moved >= num_objects_created`, and it makes your tests pass on i686: https://github.com/ruby/ruby/pull/9175

Honestly though I don't love what I had to do to the mark & pin bitset traversal to get these counts, and I'm leaning towards just adding a fudge factor of 10 or so objects in the tests to account for this instead. WDYT @peterzhu2118 ?

KJTsanaktsidis:
I opened https://github.com/ruby/ruby/pull/9176 to add the fudge factor. I'm happy for either https://github.com/ruby/ruby/pull/9176 or https://github.com/ruby/ruby/pull/9175 to be committed, either approach should fix the issue.

jeremyevans:
@KJTsanaktsidis thank you very much for working on this so quickly.  I think the fudge factor approach is simpler, though the pinned approach is more exact and invasive. I'm fine with either approach.

PR #8880, opened 2023-11-09
RubyVM::InstructionSequence.compile_file_prism

kddnewton:
* Provide a new API compile_file_prism which mirrors compile_file but uses prism to parse/compile.
* Provide the ability to run test-all with RUBY_ISEQ_DUMP_DEBUG set to "prism". If it is, we'll use the new compile_file_prism API to load iseqs during the test run.

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -943,6 +943,20 @@ rb_iseq_new_with_opt(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE rea
 
 VALUE rb_iseq_compile_prism_node(rb_iseq_t * iseq, pm_scope_node_t *scope_node, pm_parser_t *parser);
 
+/**
+ * Initialize an rb_code_location_t with a prism location.
+ */
+static void
+pm_code_location(rb_code_location_t *code_location, const pm_newline_list_t *newline_list, const pm_location_t *location) {
+    pm_line_column_t start = pm_newline_list_line_column(newline_list, location->start);
+    pm_line_column_t end = pm_newline_list_line_column(newline_list, location->end);
+
+    *code_location = (rb_code_location_t) {
+        .beg_pos = { .lineno = (int) start.line, .column = (int) start.column },
+        .end_pos = { .lineno = (int) end.line, .column = (int) end.column }
+    };
+}
+
 rb_iseq_t *
 pm_iseq_new_with_opt(pm_scope_node_t *scope_node, pm_parser_t *parser, VALUE name, VALUE path, VALUE realpath,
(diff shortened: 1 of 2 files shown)


kddnewton:
Note that I ran the following script:

```ruby
require "timeout"

filepaths = Dir["test/**/test_*.rb"]
puts "running #{filepaths.length} test files"

passed = []
failed = []
timeout = []

at_exit {
  puts "=" * 500
  puts "passed files:"
  puts passed
  puts "timeout files:"
  puts timeout
  puts "passed: #{passed.length}"
  puts "failed: #{failed.length}"
  puts "timeout: #{timeout.length}"
}

filepaths.each.with_index(1) do |filepath, index|
  puts "[%03d] %s" % [index, filepath]

  begin
    Timeout.timeout(30) do
      if system("RUBY_ISEQ_DUMP_DEBUG=prism make test-all TESTS=#{filepath} 2>&1 > /dev/null")
        passed << filepath
      else
        failed << filepath
      end
    end
  rescue Timeout::Error
    timeout << filepath
  end
end
```

By my count there are 872 and we are currently passing 119.

iseq.c
```diff
@@ -1391,6 +1391,47 @@ iseqw_s_compile(int argc, VALUE *argv, VALUE self)
...
+    rb_compile_option_t option;
+    make_compile_option(&option, opt);
+
+    prepare_iseq_build(
+        iseq,
+        rb_fstring_lit("<compiled>"),
+        file,
+        path,
+        first_lineno,
+        &(rb_code_location_t) {
```
jemmaissroff, line 1409: Should we extract this and the `start/end_line_column` to its own function that takes a `pm_newline_list_t` and a `pm_location_t` since we're reusing it

jemmaissroff (commented):
One small comment, really excited for this change!

iseq.c
```diff
@@ -1391,6 +1391,47 @@ iseqw_s_compile(int argc, VALUE *argv, VALUE self)
...
+        Qnil,
+        &option
+    );
+
+    pm_scope_node_t scope_node;
+    pm_scope_node_init(node, &scope_node, NULL, parser);
+    rb_iseq_compile_prism_node(iseq, &scope_node, parser);
+
+    finish_iseq_build(iseq);
+    pm_node_destroy(parser, node);
```
jemmaissroff, line 1432: Also, should this not be returning a `VALUE`? Surprised there's no compilation warning

jemmaissroff:
@kddnewton just pinging this one, it's a great change and would love to try get it merged asap

kddnewton:
@jemmaissroff Updated!

jemmaissroff (approved):

PR #6189, opened 2022-07-27
GVL Instrumentation: pass thread_id as part of event data

casperisfine:
Context: https://github.com/ivoanjo/gvl-tracing/pull/4

Some hooks may want to collect data on a per thread basis. Right now the only way to identified the concerned thread is to
use `rb_nativethread_self()` or similar, but even then because of the thread cache, two distinct Ruby thread may report the same native thread id.

This patch pass a `thread_id` as part of the `event_data` struct. It's a serial id generated when the thread is created.

cc @ivoanjo 

@ko1 what do you think? Would this be acceptable? There is an atomic operation an overhead even if the API is not in use, but that's on thread creation, I assume it's minor compared to the overall thread creation cost?

PR #8885, opened 2023-11-09
GVL Instrumentation: pass thread->self as part of event data

casperisfine:
Replaces: https://github.com/ruby/ruby/pull/6189
Context: https://github.com/ivoanjo/gvl-tracing/pull/4

Some hooks may want to collect data on a per thread basis. Right now the only way to identify the concerned thread is to use `rb_nativethread_self()` or similar, but even then because of the thread cache or MaNy, two distinct Ruby threads may report the same native thread id.

By passing `thread->self`, hooks can use it as a key to store the metadata.

NB: Most hooks are executed outside the GVL, so such data collection need to use a thread-safe data-structure, and shouldn't use the reference in other ways from inside the hook.

They must also either pin that value or handle compaction.

cc @ko1 @tenderlove @ivoanjo


jpcamara:
This is great. This also should make it much simpler to use `rb_profile_thread_frames` when instrumenting 👌🏼

ivoanjo:
+1 This is great, thanks as always :)

PR #8886, opened 2023-11-09
Update checkout

xuanxu:
This PR updates the checkout action to the latest version, as used in the rest of the workflows


nobu:
Thank you for the pointing out.
I thought that files was automatically updated too, but the default is limited to ".github/workflows/".
I've changed it to include other directories and am waiting to see if it works.

PR #8893, opened 2023-11-10
Fix freeing embedded TypedData and embed backtrace

casperisfine:
https://github.com/ruby/ruby/pull/7440 didn't handle embedded types with a custom `dfree` it was passing `DATA_PTR()` to `dfree` which was garbage.

Additionally `rb_data_free` would bail out if  `DATA_PTR(obj) == NULL` which in the case of an embedded type doesn't mean there's nothing to free.

This was discovered while making Backtrace embedded.


gc.c
```diff
@@ -10699,7 +10699,7 @@ gc_update_object_references(rb_objspace_t *objspace, VALUE obj)
       case T_DATA:
         /* Call the compaction callback, if it exists */
         {
-            void *const ptr = DATA_PTR(obj);
+            void *const ptr = RTYPEDDATA_P(obj) ? RTYPEDDATA_GET_DATA(obj) : DATA_PTR(obj);
```
casperisfine, line 10702: We later noticed compaction had the same issue.

peterzhu2118 (approved):

PR #8898, opened 2023-11-10
[DOC] More on JSON extensions

BurdetteLamar:
Adds JSON extensions to What's Here.

### DIFF
--- a/range.c
+++ b/range.c
@@ -2609,6 +2609,7 @@ range_overlap(VALUE range, VALUE other)
  * - {Comparing}[rdoc-ref:Range@Methods+for+Comparing]
  * - {Iterating}[rdoc-ref:Range@Methods+for+Iterating]
  * - {Converting}[rdoc-ref:Range@Methods+for+Converting]
+ * - {Methods for Working with JSON}[rdoc-ref:Range@Methods+for+Working+with+JSON]
  *
  * === Methods for Creating a \Range
  *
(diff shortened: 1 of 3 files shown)


nobu:
Shouldn’t it be in JSON docs?

BurdetteLamar:
> Shouldn’t it be in JSON docs?

RDoc puts these three methods here (and not in JSON), and includes them in the left TOC, so I thought since they're "here," they belong in "What's Here."

range.c
```diff
@@ -2651,6 +2652,12 @@ range_overlap(VALUE range, VALUE other)
  * - #to_a (aliased as #entries): Returns elements of +self+ in an array.
  * - #to_s: Returns a string representation of +self+ (uses #to_s).
  *
+ * === Methods for Working with JSON
```
peterzhu2118, line 2655: We should be clear here that you need to load JSON to get these methods.

range.c
```diff
@@ -2651,6 +2652,16 @@ range_overlap(VALUE range, VALUE other)
...
+ *
+ * To make these methods available:
+ *
+ *   require 'json/add/range'
```
peterzhu2118, line 2659: I think this is an internal implementation detail, we should just say that they need to load all of JSON.

```suggestion
 *   require 'json'
```

peterzhu2118 (approved):

PR #8903, opened 2023-11-12
Fix GC issues with ASAN fake stacks.

KJTsanaktsidis:
This fixes some of the issues with ASAN I outlined in https://bugs.ruby-lang.org/issues/20001

With this merged, `make` now works on my machine with the following:

```
CC=clang ./configure cppflags="-fsanitize=address -fno-omit-frame-pointer" optflags=-O2 LDFLAGS="-fsanitize=address -fno-omit-frame-pointer" --prefix="$(realpath installed)"
ASAN_OPTIONS=use_sigaltstack=0:detect_leaks=0:abort_on_error=1 make
```

Without these fixes, I get crashes in various steps which use the built miniruby (e.g. parsing rdoc or runing extconf files) because `VALUE`'s on the machine stack are not getting properly marked (because they are instead living in ASAN fake stacks).

This seems to build fine with ASAN both enabled and disabled on linux/mac, and builds without asan on windows (i wasn't brave enough to try turning asan on with msvc heh).

[Bug #20001]

### DIFF
--- a/include/ruby/internal/interpreter.h
+++ b/include/ruby/internal/interpreter.h
@@ -141,7 +141,7 @@ void ruby_show_copyright(void);
  *
  * @param[in]  addr  A pointer somewhere on the stack, near its bottom.
  */
-void ruby_init_stack(volatile VALUE *addr);
+void ruby_init_stack(volatile void *addr);
 
 /**
  * Initializes the VM and builtin libraries.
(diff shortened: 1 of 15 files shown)


include/ruby/internal/interpreter.h
```diff
@@ -141,7 +141,7 @@ void ruby_show_copyright(void);
  *
  * @param[in]  addr  A pointer somewhere on the stack, near its bottom.
  */
-void ruby_init_stack(volatile VALUE *addr);
+void ruby_init_stack(volatile void *addr);
```
KJTsanaktsidis, line 144: This is a public header; changing `VALUE` to `void` is not a breaking change I think (and where this is called inside ruby, it's called with a `(void *)` cast anyway today). However, removing `volatile` _would_ break a program which calls this with a volatile variable I think.

It doesn't _need_ to be volatile (we never dereference this address anyway), but it doesn't do any harm I suppose.

PR #8904, opened 2023-11-12
Array#rassoc should try to convert to array implicitly. Fixes #20003

bolshakov:
Fixes https://bugs.ruby-lang.org/issues/20003

There is a difference between behaviour of `Array#assoc` and `Array#rassoc`. The first one performs implicit conversion (calls `#to_ary`) while the former does not.

```ruby
class ArrayConvertible
  def initialize(*values)
    @values = values;
  end

  def to_ary
    @values
  end
end

s1 = [1, 2]
s2 = ArrayConvertible.new(2, 3)
a = [s1, s2]
```

The `a.assoc(2)` call returns `[2, 3]` as expected. However, `a.rassoc(3)` returns `nil`

**Expected behaviour**: `a.rassoc(3)` returns `[2, 3]` in such cases.

### DIFF
--- a/array.c
+++ b/array.c
@@ -5049,7 +5049,7 @@ rb_ary_rassoc(VALUE ary, VALUE value)
     VALUE v;
 
     for (i = 0; i < RARRAY_LEN(ary); ++i) {
-        v = RARRAY_AREF(ary, i);
+        v = rb_check_array_type(RARRAY_AREF(ary, i));
         if (RB_TYPE_P(v, T_ARRAY) &&
             RARRAY_LEN(v) > 1 &&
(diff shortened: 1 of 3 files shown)


bolshakov:
@mame now everything passed, except the spec I introduced here https://github.com/ruby/spec/pull/1110 I will update it after the merge

mame:
@bolshakov Could you update the mirrored spec file in this ruby/ruby repo directly? Changing ruby/spec is not reflected in real time. I want to keep the CI green.

bolshakov:
@mame I updated the specs, and all checks are green now.

PR #8907, opened 2023-11-13
[DOC] Correct Time class timezone example

olivierlacan:
The Time documentation seemed to suggest there was such a method as `timezone()` which doesn't exist but it does have an equivalent in the TZInfo gem which is (officially?) supported by the Ruby Time class per [time.c][1].

We could also remove this example since it depends on a third-party gem, but TZInfo is a valuable Ruby tool, so I don't think that would be advisable.

[1]: https://github.com/olivierlacan/ruby/blob/5bb756287244923f2817a5989fe9dc78930b4380/time.c#L5700-L5705

### DIFF
--- a/timev.rb
+++ b/timev.rb
@@ -66,9 +66,10 @@
 #
 #   Time.new(2002, 10, 31, 2, 2, 2, "+02:00") #=> 2002-10-31 02:02:02 +0200
 #
-# Or a timezone object:
+# Or a {TZInfo}[https://github.com/tzinfo/tzinfo] timezone object:
 #
-#   zone = timezone("Europe/Athens")      # Eastern European Time, UTC+2
+#   require "tzinfo"
+#   zone = TZInfo::Timezone.get("Europe/Athens") # Eastern European Time, UTC+2
 #   Time.new(2002, 10, 31, 2, 2, 2, zone) #=> 2002-10-31 02:02:02 +0200
 #
 # You can also use Time.local and Time.utc to infer


nobu:
I don't suggest TZInfo exclusively.
Of course is is a supported libraray, but also `Timezone` gem.

What about this?
```diff
diff --git i/doc/timezones.rdoc w/doc/timezones.rdoc
index 18233b97aeb..b69ffefb305 100644
--- i/doc/timezones.rdoc
+++ w/doc/timezones.rdoc
@@ -57,7 +57,9 @@ in the range <tt>-86399..86399</tt>:
 
 ==== Timezone Objects
 
-The zone value may be an object responding to certain timezone methods.
+The zone value may be an object responding to certain timezone methods, an
+instance of {Timezone}[https://github.com/panthomakos/timezone] and
+{TZInfo}[https://tzinfo.github.io] for example.
 
 The timezone methods are:
 
diff --git i/timev.rb w/timev.rb
index e710dc69900..80de049a072 100644
--- i/timev.rb
+++ w/timev.rb
@@ -66,7 +66,7 @@
 #
 #   Time.new(2002, 10, 31, 2, 2, 2, "+02:00") #=> 2002-10-31 02:02:02 +0200
 #
-# Or a timezone object:
+# Or {a timezone object}[rdoc-ref:timezones.rdoc@Timezone+Objects]:
 #
 #   zone = timezone("Europe/Athens")      # Eastern European Time, UTC+2
 #   Time.new(2002, 10, 31, 2, 2, 2, zone) #=> 2002-10-31 02:02:02 +0200
```

olivierlacan:
@nobu Sounds good to me. Thanks! I'll close and let you apply your patch.

PR #8914, opened 2023-11-13
[PRISM] Implement once node for interpolated regex

eileencodes:
This PR implements the once node on interpolated regexes.

There is a bug in Prism where the interpolated regex with the once flag only works when there is not a local variable so the test uses a "1". We'll need to fix that.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1091,6 +1091,11 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
             scope->body = (pm_node_t *)cast->statements;
             break;
         }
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -1091,6 +1091,10 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
             scope->body = (pm_node_t *)cast->statements;
             break;
         }
+        case PM_INTERPOLATED_REGULAR_EXPRESSION_NODE: {
```
jemmaissroff, line 1094: Can we add an assertion that this has a once flag?

jemmaissroff (changes requested):

jemmaissroff (approved):

PR #8917, opened 2023-11-14
Reject 'class << (return); end` by "void value expression"

makenowjust:


nobu (approved):
Corresponding tests would be nice.

makenowjust:
@nobu Which is better to put tests to `test_syntax.rb` or `test_parse.rb`?

PR #8921, opened 2023-11-14
Bundle rbs-3.3.1

soutaro:
https://github.com/ruby/rbs/pull/1619


hsbt:
Should we remove https://github.com/ruby/ruby/blob/master/tool/rbs_skip_tests#L20 - L29?

soutaro:
Yes. I plan to release 3.3.1 for it.

PR #8929, opened 2023-11-15
Fix regex match cache out-of-bounds access

XrXr:
Previously the following read and wrote 1 byte out-of-bounds:

    $ valgrind ruby -e 'p /(\W+)[bx]\?/i.match? "aaaaaa aaaaaaaaa aaaa aaaaaaaa aaa aaaaxaaaaaaaaaaa aaaaa aaaaaaaaaaaa a ? aaa aaaa a ?"' 2> >(grep Invalid -A 30)

Because of the `match_cache_point_index + 1` in
memoize_extended_match_cache_point() and
check_extended_match_cache_point(), we need one more byte of space.

---

Sorry, I don't understand this code well enough to give a better reproducer. Since the corruption is just one byte, I also couldn't make a test case, but it does eventually lead to crashes in a private app.

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -4092,7 +4092,7 @@ match_at(regex_t* reg, const UChar* str, const UChar* end,
 	  if (num_match_cache_points >= LONG_MAX_LIMIT) {
 	    return ONIGERR_MEMORY;
 	  }
-	  size_t match_cache_buf_length = (num_match_cache_points >> 3) + (num_match_cache_points & 7 ? 1 : 0);
+	  size_t match_cache_buf_length = (num_match_cache_points >> 3) + (num_match_cache_points & 7 ? 1 : 0) + 1;
 	  uint8_t* match_cache_buf = (uint8_t*)xmalloc(match_cache_buf_length * sizeof(uint8_t));
 	  if (match_cache_buf == NULL) {
 	    return ONIGERR_MEMORY;


makenowjust:
Your fix seems good, but it is not essential.
I guess this kind of bug is caused by failure to count the number of cache op-codes in `count_num_cache_opcodes`. I will find it, but if it is not found until the 3.3.0 release, I will merge this PR.

makenowjust (approved):
I did Approve just in case. If this fix is needed just now, I will not stop merging this.

byroot:
I'll take the liberty to merge this, because it cause so many crashes on our internal CI that it prevent us from seeing if any other issue arise.

byroot:
Also @makenowjust feel free to ping me if you wish to test your better fix against our internal CI.

wojtekLs:
@byroot why not make CI public?

byroot:
Because it's running the test suite of our main application, so it would leak lots of private information.

byroot:
For the record the regexp triggering the bug was:

```ruby
sql_query.gsub(/(\W+)[bx]\?/i, "\\1?")
```

makenowjust:
I found the reason for this bug. https://github.com/ruby/ruby/pull/9367 will fix this and we can revert this commit.

PR #8932, opened 2023-11-16
Pin object ivars while inside `rb_evict_ivars_to_hash`

casperisfine:
Otherwise if compaction triggers, the reference we just copied inside the table may be moved without us knowing.

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1370,10 +1370,61 @@ rb_attr_delete(VALUE obj, ID id)
     return rb_ivar_delete(obj, id, Qnil);
 }
 
+static void
+ivar_pinner_mark(void *data) {
+    if (data) {
+        VALUE obj = (VALUE)data;
+        RUBY_ASSERT(!rb_shape_obj_too_complex(obj));
+        rb_gc_mark(obj);
+
+        if (FL_TEST(obj, FL_EXIVAR)) {
+            struct gen_ivtbl *ivtbl;
+
+            if (rb_gen_ivtbl_get(obj, 0, &ivtbl)) {
+                for (uint32_t i = 0; i < ivtbl->as.shape.numiv; i++) {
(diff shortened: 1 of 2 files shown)


casperisfine:
@XrXr if you have a minute to review. I'm not super proud of this solution, but if you have a better suggestion, I'm all ears.

casperisfine:
I don't think my fix is working properly, I'm still seeing some crashes in the added tests.

XrXr:
I wrote a patch that pins the objects that are in the st_table rather than the original object by allocating a separate too-complex T_OBJECT as the pinner. No need for a new type, but still a bit hacky.

I then realized that the crash here and what I see locally is `<OBJ_INFO:gc_mark_ptr@/Users/alan/ruby/gc.c:6966> 0x0000000103e44a90 [0 M    ] T_NONE` which is
```
gc_mark(objspace, (VALUE)def->body.iseq.cref);
```
Feels like a separate issue entirely that the test just happens to uncover.

casperisfine:
> Feels like a separate issue entirely that the test just happens to uncover.

Yes, I had the same conclusion. I'm trying to figure it out but it's tricky.

casperisfine:
> by allocating a separate too-complex T_OBJECT as the pinner.

Ah, smart.

variable.c
```diff
@@ -1377,6 +1383,8 @@ rb_obj_convert_to_too_complex(VALUE obj, st_table *table)
 
     VALUE *old_ivptr = NULL;
 
+    rb_ivar_foreach(obj, rb_obj_write_barrier_ivars_i, (st_data_t)obj);
```
casperisfine, line 1386: I don't understand why this is needed. But if I remove it, I get some GC crash trying to mark T_NONE.

PR #6046, opened 2022-06-20
WIP: thread storage API

casperisfine:
### Context

Something I discussed with @ko1 after experimenting on https://github.com/Shopify/gvltools.

It seems a common need for the thread callbacks to have to store some thread local data. It's doable using `_Thread_local` (or equivalent) however native threads don't map very well with Ruby thread.

Currently MRI has an optimization where when a Ruby thread is dead, rather than to free the native thread, it's kept in cache and re-used for a future Ruby thread. In this scenario the `_Thread_local` are not re-initialized, which makes it an easy mistake to leak state.

There's also some project to implement N:M threads in MRI, if this was to happen, then `_Thread_local` would become entirely unusable.

### Feature

The idea here is to expose an API similar to `pthread_key_create / pthread_getspecific / pthread_setspecific (3)`:

```c
typedef void (*rb_thread_storage_destructor)(void *data);
bool rb_thread_storage_create_key(rb_internal_thread_store_key_t *key, rb_thread_storage_destructor func);
void *rb_internal_thread_store_get(rb_thread_storage_key_t key);
bool rb_internal_thread_store_set(rb_thread_storage_key_t key, void *data);
```

If a destructor is provided, it's called with the data pointer unless the data pointer is `NULL` (inspired from `pthread_key_create`).

The destructor is called after `RUBY_INTERNAL_THREAD_EVENT_EXITED` is fired.

PR #8936, opened 2023-11-16
Thread specific storages APIs

ko1:
This patch introduces thread specific storage APIs for tools which use `rb_internal_thread_event_hook` APIs.

* `rb_internal_thread_specific_key_create()` to create a tool specific thread local storage key.
* `rb_internal_thread_specific_set()` sets a data to thread and tool specific storage.
* `rb_internal_thread_specific_get()` gets a data in thread and tool specific storage.

Note that `rb_internal_thread_specific_get(thread_val, key)` can be called without GVL and safe for async signal and safe for multi-threading (native threads). So you can call it in any internal thread event hooks. Further more you can call it from other native threads. Of course `thread_val` should be living while accessing the data from this function.

Note that you should not forget to clean up the set data.

### DIFF
--- a/include/ruby/thread.h
+++ b/include/ruby/thread.h
@@ -267,6 +267,46 @@ rb_internal_thread_event_hook_t *rb_internal_thread_add_event_hook(
 bool rb_internal_thread_remove_event_hook(
         rb_internal_thread_event_hook_t * hook);
 
+
+typedef int rb_internal_thread_specific_key_t;
+#define RB_INTERNAL_THREAD_SPECIFIC_KEY_MAX 8
+/**
+ * Create a key to store thread specific data.
+ *
+ * These APIs are designed for tools using
+ * rb_internal_thread_event_hook APIs.
+ *
+ * Note that only `RB_INTERNAL_THREAD_SPECIFIC_KEY_MAX` keys
+ * can be created. raises `ThreadError` if exceeded.
+ *
+ * Usage:
+ *   // at initialize time:
+ *   int tool_key; // gvar
+ *   Init_tool() {
+ *     tool_key = rb_internal_thread_specific_key_create();
+ *   }
+ *
+ *   // at any timing:
+ *   rb_internal_thread_pecific_set(thread, tool_key, per_thread_data);
+ *   ...
+ *   per_thread_data = rb_internal_thread_specific_get(thread, tool_key);
+ */
+rb_internal_thread_specific_key_t rb_internal_thread_specific_key_create(void);
+
+/**
+ * Get thread and tool specific data.
+ *
+ * This function is async signal safe and thread safe.
+ */
+void *rb_internal_thread_specific_get(VALUE thread_val, rb_internal_thread_specific_key_t key);
+
+/**
+ * Set thread and tool specific data.
+ *
+ * This function is async signal safe and thread safe.
+ */
+void rb_internal_thread_specific_set(VALUE thread_val, rb_internal_thread_specific_key_t key, void *data);
+
 RBIMPL_SYMBOL_EXPORT_END()
 
 #endif /* RUBY_THREAD_H */
--- a/vm_core.h
+++ b/vm_core.h
@@ -1152,6 +1152,7 @@ typedef struct rb_thread_struct {
 
     /* misc */
     VALUE name;
+    void **specific_storage;
 
     struct rb_ext_config ext_config;
 } rb_thread_t;
--- a/rjit_c.rb
+++ b/rjit_c.rb
@@ -1530,6 +1530,7 @@ module RubyVM::RJIT # :nodoc: all
       scheduler: [self.VALUE, Primitive.cexpr!("OFFSETOF((*((struct rb_thread_struct *)NULL)), scheduler)")],
       blocking: [CType::Immediate.parse("unsigned int"), Primitive.cexpr!("OFFSETOF((*((struct rb_thread_struct *)NULL)), blocking)")],
       name: [self.VALUE, Primitive.cexpr!("OFFSETOF((*((struct rb_thread_struct *)NULL)), name)")],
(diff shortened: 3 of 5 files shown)


ko1:
This is an alternative proposal to https://github.com/ruby/ruby/pull/6046.
@casperisfine what do you think about it?

vm_core.h
```diff
@@ -1152,6 +1152,7 @@ typedef struct rb_thread_struct {
 
     /* misc */
     VALUE name;
+  void **specific_storage;
```
casperisfine, line 1155:
```suggestion
    void **specific_storage;
```

casperisfine:
> what do you think about it?

It's perfect for https://github.com/Shopify/gvltools. 

Also the API allows to query another thread's storage, is that safe? If so it's great.

ko1:
> Also the API allows to query another thread's storage, is that safe? If so it's great.

Yes.

> Further more you can call it from other native threads.

include/ruby/thread.h
```diff
@@ -264,6 +264,46 @@ rb_internal_thread_event_hook_t *rb_internal_thread_add_event_hook(
...
+ * Create a key to store thread specific data.
+ *
+ * These APIs are designed for tools using
+ * rb_internal_thread_event_hook APIs.
+ *
+ * Note that only `RB_INTERNAL_THREAD_SPECIFIC_KEY_MAX` keys
+ * can be created. raises `ThreadError` if exceeded.
+ *
+ * Usage:
+ *   // at initialize timing:
```
byroot, line 280:
```suggestion
 *   // at initialize time:
```

include/ruby/thread.h
```diff
@@ -264,6 +264,46 @@ rb_internal_thread_event_hook_t *rb_internal_thread_add_event_hook(
...
+ *   int tool_key; // gvar
+ *   Init_tool() {
+ *     tool_key = rb_internal_thread_specific_key_create();
+ *   }
+ *
+ *   // at thread start timing (with GVL):
```
byroot, line 286:
```suggestion
 *   // at thread start time (with GVL):
```

include/ruby/thread.h
```diff
@@ -264,6 +264,46 @@ rb_internal_thread_event_hook_t *rb_internal_thread_add_event_hook(
...
+ *   rb_internal_thread_pecific_set(thread, tool_key, per_thread_data);
+ *
+ *   // at any timing:
```
byroot, line 289:
```suggestion
 *   // at any time:
```

byroot (commented):
timing -> time

include/ruby/thread.h
```diff
@@ -264,6 +264,46 @@ rb_internal_thread_event_hook_t *rb_internal_thread_add_event_hook(
...
+/**
+ * Get thread and tool specific data.
+ *
+ * This function is async signal safe and thread safe.
+ */
+void *rb_internal_thread_specific_get(VALUE thread_val, rb_internal_thread_specific_key_t key);
+
+/**
+ * Set thread and tool specific data.
+ * This function should be called with GVL.
```
casperisfine, line 303: Hum, actually, this may be tricky for `gvltools`, ideally I'd want to be able to initialize the key on the first event received for that thread.

Could we use atomics to allow this?

ko1: or should we allocate storage for all existing threads (and created threads in future) at `rb_internal_thread_specific_key_create()` is created?

ko1: This design introduce some more complexity because of Ractors.

ko1: The current comment is wrong.
Calling this set function is safe if the storage is allocated. For example, if `get()` returns non-null, it is safe to call `set()`.

byroot: I think preallocating storage would make sense.

ko1: I don't want to introduce pre-allocation because of complexity. Could you elaborate why it is needed more?

byroot: Right, so if you want to do record something on `RUBY_INTERNAL_THREAD_EVENT_STARTED` event, you can't because you don't have the GVL, you need to wait for the thread to acquire the GVL first, there you can initialize memory.

(pseudo code)
```c
static void
gt_thread_callback(rb_event_flag_t event, const rb_internal_thread_event_data_t *event_data, void *user_data) {
   gvl_stats stats = rb_internal_thread_specific_get(instrumentation_key);
   if (!stats) {
     stats = alloc_thread_storage();
     rb_internal_thread_specific_set(instrumentation_key, stats);
   }

    switch(event) {
        case RUBY_INTERNAL_THREAD_EVENT_STARTED:
          stats.started_at = time_now();
// snip...
}

ko1: I see... I missed that STARTED state does not have GVL.

byroot: Yes. But even beyond that. It is common to have the hook only registered at some point, so the first event of any thread can be any of the events.

ko1: 1. It is easy to pre-allocate at creation timing. No complex issue.
2. It is hard to allocate for *existing* threads.

Does dropping 2 make sense?

byroot: I think 2 can be worked around by iterating on `Thread.list`.

But couldn't `rb_internal_thread_specific_key_create` lock the whole VM? This is really intended as a debug API so I don't think it's a problem to temporarily stop other threads and ractors. Also there is a max of 8 keys, so it's not like it would happen repeatedly.

casperisfine:
cc @tenderlove @jhawthorn as you may have interest into this as well?

include/ruby/thread.h
ko1: I added to the limitation "the first call of `rb_internal_thread_specific_key_create()` should not be called with multiple ractors". I think it would be reasonable limitation (Init_foo of extension should call it).

PR #8939, opened 2023-11-17
File.directory? doc was overriding File::Stat#directory? doc

robotdana:
`File::Stat#directory?` now gets its own documentation that was already in the code, just hidden previously because it picked up the File.directory? documentation comment instead.

before: 

https://ruby-doc.com/3.2.2/File/Stat.html#method-i-directory-3F

after:

### DIFF
--- a/file.c
+++ b/file.c
@@ -1593,8 +1593,6 @@ rb_access(VALUE fname, int mode)
  */
 
 /*
- * Document-method: directory?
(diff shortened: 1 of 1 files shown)


XrXr:
Thanks! By the way, there is a first-party documentation site at <https://docs.ruby-lang.org/>

PR #8948, opened 2023-11-20
[3.2 backport] Fix compile error on older systems without clock_get*

Bo98:
Cherry pick of #8256, which has been stable in master for a while.

Fixes a regression introduced in 3.2. Earlier branches are not affected.


hsbt:
Can you file it https://bugs.ruby-lang.org/?

see https://github.com/ruby/ruby/wiki/How-To-Request-Backport

Bo98:
https://bugs.ruby-lang.org/issues/20010

nagachika (approved):

PR #8949, opened 2023-11-20
Take 2: Fix thread unsafety with rb_register_postponed_job (without locking)

KJTsanaktsidis:
This is an alternative fix to the problem I tried to solve in https://github.com/ruby/ruby/pull/8856. This implementation does not rely on any blocking or syscalls in the critical path. Let me copy-paste what I wrote in the commit message to outline the new solution:

-----

This commit fixes the issue by re-implementing both the postponed-job and workqueue mechanism, and wrapping them into an overarching `rb_postponed_job_queues_t` structure. The structure is split into "blocking" and "async" parts.

The "blocking" component is a re-implementation of the workqueue mechanism. Instead of using a linked-list of jobs, it stores them in a hashtable, so that fast deduplication of jobs in _one functions is possible. This is an approach which Eric Wong originally committed in https://github.com/ruby/ruby/commit/5a1dfb04bc2b09fcf8f3427cac72d0ce52a45eb2, but had to be reverted due to async-signal-safety concerns. When protected by a mutex, however, this is perfectly fine.

The "blocking" queue is exposed through `rb_workqueue_register` and `rb_workqueue_register_one`; these functions are safe to be called at any time and from any thread, but _NOT_ from a signal handler (since they use a mutex). The job is guaranteed to be enqueued; the function may block or allocate memory in order to fulfil that guarantee.

The "async" component is a re-implementation of the existing postponed job buffer. It represents a lock-free, single-consumer, multi-producer ringbuffer. It's safe to insert jobs into this ringbuffer at any time, from any thread, even in a signal handler. Because allocating memory is not safe inside a signal handler, the ringbuffer is fixed-size and it's
possible enqueuing jobs might fail for want of space; callers must accept that.

The "async" queue is exposed through the rb_postponed_job_register and rb_postponed_job_register_one functions. These functions can be called from any thread, including from signal handlers.

The _execution_ of both the async and blocking jobs is unified in `rb_postponed_job_flush`. The async queue is actually drained into the blocking queue, which has the effect of deduplicating jobs with the same callback enqueued with `rb_workqueue_register_one` and `rb_postponed_job_register_one`.

-----

I compared the yjit-bench results of this branch with current master, and it seems this implementation produces a very slight speedup in interpreter performance (well, at the very least, it doesn't look like it makes things slower):

```
interp: ruby 3.3.0dev (2023-11-20T03:48:56Z ktsanaktsidis/post.. 7869e776b2) [x86_64-linux]
yjit: ruby 3.3.0dev (2023-11-20T03:48:56Z ktsanaktsidis/post.. 7869e776b2) +YJIT [x86_64-linux]

--------------  -----------  ----------  ---------  ----------  ------------  -----------
bench           interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
activerecord    191.7        2.8         89.5       5.0         1.61          2.14
erubi_rails     58.7         6.7         32.6       13.8        0.39          1.80
hexapdf         7535.4       0.6         4334.5     0.5         1.50          1.74
liquid-c        197.0        0.5         135.2      1.8         0.85          1.46
liquid-compile  193.7        3.3         132.1      0.8         0.82          1.47
liquid-render   469.5        0.3         192.0      0.6         1.48          2.44
mail            398.5        0.3         281.1      0.4         0.80          1.42
psych-load      6447.5       0.3         4255.3     0.6         1.50          1.52
railsbench      6484.0       0.6         3772.4     1.0         1.40          1.72
ruby-lsp        20.8         4.6         15.6       11.7        0.52          1.33
sequel          214.1        1.0         152.1      1.9         1.36          1.41
--------------  -----------  ----------  ---------  ----------  ------------  -----------
```

vs

```
interp: ruby 3.3.0dev (2023-11-20T02:02:55Z master 701b0650de) [x86_64-linux]
yjit: ruby 3.3.0dev (2023-11-20T02:02:55Z master 701b0650de) +YJIT [x86_64-linux]

--------------  -----------  ----------  ---------  ----------  ------------  -----------
bench           interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
activerecord    196.2        4.1         88.5       5.3         1.64          2.22
erubi_rails     59.3         6.5         33.2       13.4        0.32          1.79
hexapdf         7637.4       2.0         4344.3     0.6         1.52          1.76
liquid-c        197.4        0.5         137.3      4.2         0.86          1.44
liquid-compile  192.8        0.3         133.7      2.1         0.77          1.44
liquid-render   465.7        0.4         191.2      0.6         1.34          2.44
mail            400.8        0.5         280.4      0.4         0.81          1.43
psych-load      6551.1       0.3         4253.2     0.2         1.52          1.54
railsbench      6526.4       1.0         3776.6     0.8         1.40          1.73
ruby-lsp        20.7         3.5         15.5       11.7        0.49          1.33
sequel          216.7        3.4         152.8      3.2         1.37          1.42
--------------  -----------  ----------  ---------  ----------  ------------  -----------
```

Fixes #19991 (https://bugs.ruby-lang.org/issues/19991)

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -952,6 +952,7 @@ typedef struct rb_objspace {
 #endif
 
     rb_darray(VALUE *) weak_references;
+    rb_postponed_job_handle_t finalize_deferred_pjob;
 } rb_objspace_t;
 
 
@@ -1424,6 +1425,8 @@ static inline void gc_prof_set_heap_info(rb_objspace_t *);
 PRINTF_ARGS(static void gc_report_body(int level, rb_objspace_t *objspace, const char *fmt, ...), 3, 4);
 static const char *obj_info(VALUE obj);
 static const char *obj_type_name(VALUE obj);
+
+static void gc_finalize_deferred(void *dmy);
 
 /*
  * 1 - TSC (H/W Time Stamp Counter)
@@ -1906,6 +1909,10 @@ rb_objspace_alloc(void)
     rb_objspace_t *objspace = calloc1(sizeof(rb_objspace_t));
     objspace->flags.measure_gc = 1;
     malloc_limit = gc_params.malloc_limit_min;
+    objspace->finalize_deferred_pjob = rb_postponed_job_preregister(gc_finalize_deferred, objspace);
+    if (objspace->finalize_deferred_pjob == POSTPONED_JOB_HANDLE_INVALID) {
+        rb_bug("Could not preregister postponed job for GC");
+    }
 
     for (int i = 0; i < SIZE_POOL_COUNT; i++) {
         rb_size_pool_t *size_pool = &size_pools[i];
@@ -4527,9 +4534,8 @@ gc_finalize_deferred(void *dmy)
 static void
 gc_finalize_deferred_register(rb_objspace_t *objspace)
 {
-    if (rb_postponed_job_register_one(0, gc_finalize_deferred, objspace) == 0) {
-        rb_bug("gc_finalize_deferred_register: can't register finalizer.");
-    }
+    /* will enqueue a call to gc_finalize_deferred */
+    rb_postponed_job_trigger(objspace->finalize_deferred_pjob);
 }
 
 static int pop_mark_stack(mark_stack_t *stack, VALUE *data);
--- a/include/ruby/debug.h
+++ b/include/ruby/debug.h
@@ -10,6 +10,7 @@
  *             modify this file, provided that  the conditions mentioned in the
  *             file COPYING are met.  Consult the file for details.
  */
+#include "ruby/internal/attr/deprecated.h"
 #include "ruby/internal/attr/nonnull.h"
 #include "ruby/internal/attr/returns_nonnull.h"
 #include "ruby/internal/dllexport.h"
@@ -615,48 +616,151 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
 
 /*
  * Postponed Job API
- * rb_postponed_job_register and rb_postponed_job_register_one are
- * async-signal-safe and used via SIGPROF by the "stackprof" RubyGem
+ *
+ * This API is designed to be called from contexts where it is not safe to run Ruby
+ * code (e.g. because they do not hold the GVL or because GC is in progress), and
+ * defer a callback to run in a context where it _is_ safe. The primary intended
+ * users of this API is for sampling profilers like the "stackprof" gem; these work
+ * by scheduling the periodic delivery of a SIGPROF signal, and inside the C-level
+ * signal handler, deferring a job to collect a Ruby backtrace when it is next safe
+ * to do so.
+ *
+ * Historically, this API provided two functions `rb_postponed_job_register` and
+ * `rb_postponed_job_register_one`, which claimed to be fully async-signal-safe and
+ * would call back the provided `func` and `data` at an appropriate time. However,
+ * these functions were subject to race conditions which could cause crashes when
+ * racing with Ruby's internal use of them.
+ *
+ * Therefore, this API has now been changed, and now requires that jobs scheduled
+ * from a signal handler context are pre-registered in advance into a fixed-size
+ * table. This table is quite small (it only has 32 entries on most systems)
+ * and so gems should generally only preregister one or two funcs. This process is
+ * managed by the `rb_postponed_job_preregister` and `rb_postponed_job_trigger`
+ * functions.
+ *
+ * We also provide the old `rb_postponed_job_register` and
+ * `rb_postponed_job_register_one` functions for backwards compatability, but with
+ * changed semantics; `rb_postponed_job_register` now behaves the same as
+ * `rb_postponed_job_register_once`. These changes should remain compatible with all
+ * of the observed in-the-wild usages of the postponed job APIs, which almost all
+ * use the _one API and pass `0` for data anyway.
  */
+
 
 /**
  * Type of postponed jobs.
  *
- * @param[in,out]  arg What was passed to rb_postponed_job_register().
+ * @param[in,out]  arg What was passed to `rb_postponed_job_preregister`
  */
 typedef void (*rb_postponed_job_func_t)(void *arg);
 
 /**
- * Registers a postponed job.
+ * The type of a handle returned from `rb_postponed_job_preregister` and
+ * passed to `rb_postponed_job_trigger`
+ */
+typedef unsigned int rb_postponed_job_handle_t;
+#define POSTPONED_JOB_HANDLE_INVALID ((rb_postponed_job_handle_t)UINT_MAX)
+
+/**
+ * Pre-registers a func in Ruby's postponed job preregistration table,
+ * returning an opaque handle which can be used to trigger the job later. Generally,
+ * this function will be called during the initialization routine of an extension.
  *
- * There  are situations  when running  a ruby  program is  not possible.   For
- * instance when  a program is in  a signal handler; for  another instance when
- * the GC  is busy.   On such situations  however, there might  be needs  to do
- * something.  We cannot but defer such operations until we are 100% sure it is
- * safe  to execute  them.   This  mechanism is  called  postponed jobs.   This
- * function  registers a  new one.   The registered  job would  eventually gets
- * executed.
+ * The returned handle can be used later to call `rb_postponed_job_trigger`. This will
+ * cause Ruby to call back into the registered `func` with `data` at a later time, in
+ * a context where the GVL is held and it is safe to perform Ruby allocations.
  *
- * @param[in]      flags      (Unused) reserved for future extensions.
+ * If the given func was already pre-registered, this method will overwrite the
+ * stored data with the newly passed data, and return the same handle instance as
+ * was previously returned.
+ *
+ * If this function is called concurrently with the same `func`, then the stored data
+ * could be the value from either call (but will definitely be one of them).
+ *
+ * If this function is called to update the data concurrently with a call to
+ * `rb_postponed_job_trigger` on the same handle, it's undefined whether `func` will
+ * be called with the old data or the new data.
+ *
+ * Although the current implementation of this method is in fact async-signal-safe and
+ * has defined semantics when called concurrently on the same `func`, a future Ruby
+ * version might require that this method be called under the GVL; thus, programs which
+ * aim to be forward-compatible should call this method whilst holding the GVL.
+ *
+ * @param[in]   func        The function to be pre-registered
+ * @param[in]   data        The data to be pre-registered
+ * @retval      POSTPONED_JOB_HANDLE_INVALID    The job table is full; this registration
+ *                          did not succeed and no further registration will do so for
+ *                          the lifetime of the program.
+ * @retval      otherwise   A handle which can be passed to `rb_postponed_job_trigger`
+ */
+rb_postponed_job_handle_t rb_postponed_job_preregister(rb_postponed_job_func_t func, void *data);
+
+/**
+ * Triggers a pre-registered job registered with rb_postponed_job_preregister,
+ * scheduling it for execution the next time the Ruby VM checks for interrupts.
+ * The context in which the job is called in holds the GVL and is safe to perform
+ * Ruby allocations within (i.e. it is not during GC).
+ *
+ * This method is async-signal-safe and can be called from any thread, at any
+ * time, including in signal handlers.
+ *
+ * If this method is called multiple times, Ruby will coalesce this into only
+ * one call to the job the next time it checks for interrupts.
+ *
+ * @params[in]  h   A handle returned from rb_postponed_job_preregister
+ */
+void rb_postponed_job_trigger(rb_postponed_job_handle_t h);
+
+/**
+ * Schedules the given `func` to be called with `data` when Ruby next checks for
+ * interupts. If this function is called multiple times in between Ruby checking
+ * for interrupts, then `func` will be called only once with the `data` vlaue from
+ * the first call to this function.
+ *
+ * Like `rb_postponed_job_trigger`, the context in which the job is called
+ * holds the GVL and can allocate Ruby objects.
+ *
+ * This method essentially has the same semantics as:
+ *
+ * ```
+ *   rb_postponed_job_trigger(rb_postponed_job_preregister(func, data));
+ * ```
+ *
+ * @note    Prevoius versions of Ruby promised that the (`func`, `data`) pairs would
+ *          be executed as many times as they were registered with this function; in
+ *          reality this was always subject to race conditions and this function no
+ *          longer provides this guarantee. Instead, we only promise that `func` will
+ *          be called once.
+ *
+ * @deprecated  This interface implies that arbitrarily many `func`'s can be enqueued
+ *              over the lifetime of the program, whilst in reality the registration
+ *              slots for postponed jobs are a finite resource. This is made clearer
+ *              by the `rb_postponed_job_preregister` and `rb_postponed_job_trigger`
+ *              functions, and a future version of Ruby might delete this function.
+ *
+ * @param[in]      flags      Unused and ignored.
  * @param[in]      func       Job body.
  * @param[in,out]  data       Passed as-is to `func`.
- * @retval         0          Postponed job buffer is full.  Failed.
- * @retval         otherwise  Opaque return value.
- * @post           The passed job is postponed.
+ * @retval         0          Postponed job registration table is full. Failed.
+ * @retval         1          Registration succeeded.
+ * @post           The passed job will run on the next interrupt check.
  */
+ RBIMPL_ATTR_DEPRECATED(("use rb_postponed_job_preregister and rb_postponed_job_trigger"))
 int rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void *data);
 
 /**
- * Identical to rb_postponed_job_register(), except  it additionally checks for
- * duplicated registration.  In case the passed job is already in the postponed
- * job buffer this function does nothing.
+ * Identical to `rb_postponed_job_register`
  *
- * @param[in]      flags      (Unused) reserved for future extensions.
+ * @deprecated  This is deprecated for the same reason as `rb_postponed_job_register`
+ *
+ * @param[in]      flags      Unused and ignored.
  * @param[in]      func       Job body.
  * @param[in,out]  data       Passed as-is to `func`.
- * @retval         0          Postponed job buffer is full.  Failed.
- * @retval         otherwise  Opaque return value.
+ * @retval         0          Postponed job registration table is full. Failed.
+ * @retval         1          Registration succeeded.
+ * @post           The passed job will run on the next interrupt check.
  */
+ RBIMPL_ATTR_DEPRECATED(("use rb_postponed_job_preregister and rb_postponed_job_trigger"))
 int rb_postponed_job_register_one(unsigned int flags, rb_postponed_job_func_t func, void *data);
 
 /** @} */
--- a/vm_core.h
+++ b/vm_core.h
@@ -726,9 +726,8 @@ typedef struct rb_vm_struct {
     /* relation table of ensure - rollback for callcc */
     struct st_table *ensure_rollback_table;
 
-    /* postponed_job (async-signal-safe, NOT thread-safe) */
-    struct rb_postponed_job_struct *postponed_job_buffer;
-    rb_atomic_t postponed_job_index;
+    /* postponed_job (async-signal-safe, and thread-safe) */
+    struct rb_postponed_job_queue *postponed_job_queue;
 
     int src_encoding_index;
 
@@ -2171,6 +2170,10 @@ rb_exec_event_hook_script_compiled(rb_execution_context_t *ec, const rb_iseq_t *
 }
 
 void rb_vm_trap_exit(rb_vm_t *vm);
+void rb_vm_postponed_job_atfork(void); /* vm_trace.c */
+void rb_vm_postponed_job_free(void); /* vm_trace.c */
+size_t rb_vm_memsize_postponed_job_queue(void); /* vm_trace.c */
+void rb_vm_postponed_job_queue_init(rb_vm_t *vm); /* vm_trace.c */
 
 RUBY_SYMBOL_EXPORT_BEGIN
 
--- a/vm_trace.c
+++ b/vm_trace.c
@@ -23,11 +23,15 @@
 
 #include "eval_intern.h"
 #include "internal.h"
+#include "internal/bits.h"
 #include "internal/class.h"
+#include "internal/gc.h"
 #include "internal/hash.h"
 #include "internal/symbol.h"
+#include "internal/thread.h"
 #include "iseq.h"
 #include "rjit.h"
+#include "ruby/atomic.h"
 #include "ruby/debug.h"
 #include "vm_core.h"
 #include "ruby/ractor.h"
@@ -1617,17 +1621,22 @@ Init_vm_trace(void)
     rb_undef_alloc_func(rb_cTracePoint);
 }
 
-typedef struct rb_postponed_job_struct {
-    rb_postponed_job_func_t func;
-    void *data;
-} rb_postponed_job_t;
-
-#define MAX_POSTPONED_JOB                  1000
-#define MAX_POSTPONED_JOB_SPECIAL_ADDITION   24
+/*
+ * Ruby actually has two separate mechanisms for enqueueing work from contexts
+ * where it is not safe to run Ruby code, to run later on when it is safe. One
+ * is async-signal-safe but more limited, and accessed through the
+ * `rb_postponed_job_preregister` and `rb_postponed_job_trigger` functions. The
+ * other is more flexible but cannot be used in signal handlers, and is accessed
+ * through the `rb_workqueue_register` function.
+ *
+ * The postponed job functions form part of Ruby's extension API, but the
+ * workqueue functions are for internal use only.
+ */
 
 struct rb_workqueue_job {
     struct ccan_list_node jnode; /* <=> vm->workqueue */
-    rb_postponed_job_t job;
+    rb_postponed_job_func_t func;
+    void *data;
 };
 
 // Used for VM memsize reporting. Returns the size of a list of rb_workqueue_job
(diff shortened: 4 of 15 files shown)


vm_core.h
```diff
@@ -2170,6 +2165,49 @@ rb_exec_event_hook_script_compiled(rb_execution_context_t *ec, const rb_iseq_t *
...
+ * threads completely unknown to Ruby. However, this function is NOT ok to call
+ * from a (C-level) signal handler, since it can allocate and block.
+ *
+ * @param[in]      flags      (Unused) reserved for future extensions.
+ * @param[in]      func       Job body.
+ * @param[in,out]  data       Passed as-is to `func`.
+ * @return         Always returns 1
+ * @post           The passed job queued for execution.
+ */
+int rb_workqueue_register(unsigned int flags, rb_postponed_job_func_t func, void *data);
```
KJTsanaktsidis, line 2196: These functions were previously not exposed as part of the public headers, and I didn't want to make any API changes as part of this PR, so I added them to the internal `vm_core.h` header. 

If we were going to expose them, I'd want to maybe think about some better naming; it's not obvious from the name when you should call `workqueue` and when you should call `postponed_job`. So for now, let's leave them in internal headers.

gc.c
```diff
@@ -4506,9 +4506,7 @@ gc_finalize_deferred(void *dmy)
 static void
 gc_finalize_deferred_register(rb_objspace_t *objspace)
 {
-    if (rb_postponed_job_register_one(0, gc_finalize_deferred, objspace) == 0) {
-        rb_bug("gc_finalize_deferred_register: can't register finalizer.");
-    }
+    rb_workqueue_register_one(0, gc_finalize_deferred, objspace);
```
KJTsanaktsidis, line 4509: Since this just `rb_bug`'s if enqueuing the job fails, and it's not in a signal handler, it seems like it makes sense to use the workqueue API for this, which blocks and cannot fail.

yjit/src/cruby_bindings.inc.rs
```diff
@@ -961,6 +961,12 @@ extern "C" {
         klass: VALUE,
         id: ID,
     ) -> *const rb_callable_method_entry_t;
+    pub fn rb_profile_frames(
```
KJTsanaktsidis, line 964: I guess I reordered some includes, which changed how this file was generated.

vm_trace.c
```diff
@@ -1726,87 +1921,160 @@ rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void
...
+    return 1;
 }
 
-/*
- * thread-safe and called from non-Ruby thread
- * returns FALSE on failure (ENOMEM), TRUE otherwise
- */
 int
-rb_workqueue_register(unsigned flags, rb_postponed_job_func_t func, void *data)
+rb_workqueue_register_impl(rb_postponed_job_func_t func, void *data, rb_atomic_t flags)
```
ko1, line 1981: what is the purpose of `flags`?

ko1:
Honestly speaking it is too complex for me to understand :p
At first, could you make clear related functions and using data structures?

My understanding:

* 1. workqueue series is for multi-thread-safe job queuing system using `st_table`. Need to handle GC for that. No difficulties.
* 2. postponed_job is for async-signal-safe job queuing system with atomic operations.

How about to introduce spin-lock for 2 to make impl simpler? I think lock-free queue can introduce difficult bugs...

KJTsanaktsidis:
> Honestly speaking it is too complex for me to understand :p
At first, could you make clear related functions and using data structures?

I totally understand - I’ll separate out the two data structures (the lock-free ring buffer and the lock-protected st_table list) into their own types, with their own functions, rather than having it be part of the postponed job implementation directly. Would that make it easier to review?

> How about to introduce spin-lock for 2 to make impl simpler? 

unfortunately this isn’t safe - in fact, it’s never safe to use a spin lock in user space, because you cannot disable preemption. For example, if a thread is holding the postponed job lock (because it’s executing rb_postponed_job_flush), and a signal interrupts the thread, the signal handler will spin forever attempting to acquire the lock (that’s already held by the interrupted function). You can get around this by masking signals, but then you’re back to my previous PR.

Even without signal handlers, spin locks are not a good idea because the kernel might put a thread holding the lock to sleep, and any other thread which tries to take the lock will then spin the cpu at max until it exhausts its time slice.

———

Unfortunately, the API we expose with postponed jobs is just really complicated to do correctly…

KJTsanaktsidis:
> workqueue series is for multi-thread-safe job queuing > system using st_table. Need to handle GC for that. No > difficulties.
> postponed_job is for async-signal-safe job queuing > system with atomic operations.

this is exactly right I think, and how the current implementations behave.

KJTsanaktsidis:
Oh, just one more thought - the super-complex stuff is for async-safe postponed job, and in this patch I moved all the internal uses of these apis inside Ruby to the simpler work queue api. So at least any bugs in the ring buffer stuff will only affect programs using eg stackprof or other gems which call the postponed_job api; they can’t stop normal ruby programs from working properly.

ko1:
Thank you for very quick response.

> unfortunately this isn’t safe - in fact, it’s never safe to use a spin lock in user space

Yes, I realized it just after posting.

ko1:
Obsolete current postponed job interface is another idea. Maybe most of libraries use `_one` interface, so

```C
Init_foo(){
  job_location = rb_job_allocate_location();
}

on any location:
  rb_job_enqueue(job_location, job); // will kick job

rb_job_allocate_location()
{
  loc = alloc_loc();
  add_linked_list(jq->location, loc);
}

rb_job_enqueue(location, job)
{
  location->enabled = true;
  location->job = job;
  jq->job_counts++;
}

rb_job_flush()
{
  while (jq->job_counts > 0) {
    for_each loc in jq->locations {
      if (loc->enabled) { fire(loc->job); loc->enabled = false; }
    }
}
```

simpler one.

ko1:
There is 2 racy points though.

```C
#define MAX_POSTPONED_JOB (SIZEOF_VOIDP * 8)

#if SIZEOF_VOIDP == 8
typedef uint64_t ppj_bits_t;
#elif SIZEOF_VOIDP == 4
typedef uint32_t ppj_bits_t;
#else
#error "unsupported"
#endif

struct {
    union {
        ppj_bits_t bits;
        void *ptr_bits;
    } enabled;

    rb_postponed_job_func_t funcs[MAX_POSTPONED_JOB];
    void *data[MAX_POSTPONED_JOB];
} ppj;

static bool
ppj_enable(int i)
{
    while (1) {
        ppj_bits_t bits = ppj.enabled.bits;
        if (bits & (1UL << i)) {
            return false;
        }
        if (RUBY_ATOMIC_PTR_CAS(ppj.enabled.ptr_bits, (void *)bits, (void *)(bits | (1 << i))) == (void *)bits) {
            return true;
        }
    }
}

bool
ppj_disable(int i)
{
    while (1) {
        ppj_bits_t bits = ppj.enabled.bits;
        if (bits & (1 << i)) {
            if (RUBY_ATOMIC_PTR_CAS(ppj.enabled.ptr_bits, (void *)bits, (void *)(bits & ~(1UL << i))) == (void *)bits) {
                return true;
            }
        }
        else {
            return false;
        }
    }
}

static bool
ppj_enabled_p(int i)
{
    return ppj.enabled.bits & (1 << i);
}

// return 0 if registration is failed.
int
rb_postponed_job_register_one(unsigned int flags, rb_postponed_job_func_t func, void *data)
{
    int i;

    for (i=0; i<MAX_POSTPONED_JOB; i++) {
        if (ppj.funcs[i] == func) { // use existing func
            if (ppj_enabled_p(i)) return FALSE; // already enabled
            ppj.data[i] = data;
            return ppj_enable(i); // racy because data is not consistent
        }
        else if (ppj.funcs[i] == NULL) { // register new func
            if (RUBY_ATOMIC_PTR_CAS(ppj.enabled.bits, NULL, func) == NULL) {
                ppj.data[i] = data;
                return ppj_enable(i); // racy because data is not consistent
            }
            else {
                // can not allocate. try next slot
            }
        }
    }

    return FALSE; // full
}

int
rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void *data)
{
    return rb_postponed_job_register_one(flags, func, data);
}

void
rb_postponed_job_flush(rb_vm_t *vm)
{
    int i;

  retry:
    for (i=0; i<MAX_POSTPONED_JOB && ppj.enabled.bits; i += 1) {
        if (ppj_enabled_p(i)) {
            rb_postponed_job_func_t func = ppj.funcs[i];
            void *data = ppj.data[i];
            if (!ppj_disable(i)) rb_bug("should not be conflicted");

            (*func)(data);
        }
    }

    if (ppj.enabled.bits) goto retry;
}
```

ko1:
There are the following users in gems.
Nobody checks the results :p

```
ko1@gem-codesearch:~$ gem-codesearch 'rb_postponed_job_register\('
2021-09-01 /srv/gems/appoptics_apm-4.13.0/ext/oboe_metal/src/profiling.cc:        rb_postponed_job_register(0, Profiling::profiler_gc_handler, (void *)0);
2021-09-01 /srv/gems/appoptics_apm-4.13.0/ext/oboe_metal/src/profiling.cc:        rb_postponed_job_register(0, Profiling::profiler_job_handler, (void *)0);
2016-02-11 /srv/gems/gctools-0.2.4/ext/gcprof/gcprof.c:      rb_postponed_job_register(0, invoke_proc, (void *)_gcprof.serial);
2023-05-29 /srv/gems/solarwinds_apm-5.1.9/ext/oboe_metal/src/profiling.cc:        rb_postponed_job_register(0, Profiling::profiler_gc_handler, (void *)0);
2023-05-29 /srv/gems/solarwinds_apm-5.1.9/ext/oboe_metal/src/profiling.cc:        rb_postponed_job_register(0, Profiling::profiler_job_handler, (void *)0);
2018-02-22 /srv/gems/tunemygc-1.0.71/ext/tunemygc/tunemygc_ext.c:    rb_postponed_job_register(0, tunemygc_invoke_gc_snapshot, (void *)stat);

ko1@gem-codesearch:~$ gem-codesearch 'rb_postponed_job_register_one'
2015-04-18 /srv/gems/allocation_tracer-0.6.3/ext/allocation_tracer/allocation_tracer.c: rb_postponed_job_register_one(0, aggregate_freed_info, arg);
2023-11-10 /srv/gems/ddtrace-1.16.2/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c:  int result = rb_postponed_job_register_one(0, sample_from_postponed_job, NULL);
2023-11-10 /srv/gems/ddtrace-1.16.2/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c:    rb_postponed_job_register_one(0, after_gc_from_postponed_job, NULL);
2017-07-09 /srv/gems/llrb-0.0.1/ext/llrb/profiler.c:    rb_postponed_job_register_one(0, llrb_job_handler, 0);
2023-10-04 /srv/gems/pf2-0.1.0/ext/pf2/pf2.c:    rb_postponed_job_register_one(0, pf2_postponed_job, 0);
2015-01-15 /srv/gems/rbkit-0.1.10/ext/extconf.rb:unless(have_func('rb_postponed_job_register_one') &&
2020-08-03 /srv/gems/rbtrace-0.4.14/ext/extconf.rb:have_func('rb_postponed_job_register_one', 'ruby.h')
2020-08-03 /srv/gems/rbtrace-0.4.14/ext/rbtrace.c:  rb_postponed_job_register_one(0, rbtrace__receive, 0);
2023-04-06 /srv/gems/stackprof-0.2.25/ext/stackprof/extconf.rb:if have_func('rb_postponed_job_register_one') &&
2023-04-06 /srv/gems/stackprof-0.2.25/ext/stackprof/stackprof.c:        rb_postponed_job_register_one(0, stackprof_job_record_gc, (void*)0);
2023-04-06 /srv/gems/stackprof-0.2.25/ext/stackprof/stackprof.c:            rb_postponed_job_register_one(0, stackprof_job_sample_and_record, (void*)0);
2023-04-06 /srv/gems/stackprof-0.2.25/ext/stackprof/stackprof.c:            rb_postponed_job_register_one(0, stackprof_job_record_buffer, (void*)0);
2015-01-28 /srv/gems/stackprofx-0.3.0/ext/extconf.rb:if have_func('rb_postponed_job_register_one') &&
2015-01-28 /srv/gems/stackprofx-0.3.0/ext/stackprofx.c: rb_postponed_job_register_one(0, stackprofx_job_handler, 0);
```

ko1:
https://github.com/DataDog/dd-trace-rb/blob/8e470dcbc20f4d6fad95fcfe83e963c8482e6363/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c#L483

Do you find unsuccess cases?

ko1:
Ah racy issue can be solved if we change the semantics of `rb_postponed_job_register_one()` to "register at least one function and we do not guarantee which `data` is used if there is an existing `data`. In this case user should use same `data` for safety.

ko1:
* obsolete
  * `rb_postponed_job_register_one`
  * `rb_postponed_job_register`
* New APIs
* `rb_postponed_job_preregister(func, ptr) -> index`
* `rb_postponed_job_activate(index)`

With new APIs.

```C
#define MAX_POSTPONED_JOB (SIZEOF_VOIDP * 8)

#if SIZEOF_VOIDP == 8
typedef uint64_t ppj_bits_t;
#elif SIZEOF_VOIDP == 4
typedef uint32_t ppj_bits_t;
#else
#error "unsupported"
#endif

struct {
    union {
        ppj_bits_t bits;
        void *ptr_bits;
    } enabled;

    rb_postponed_job_func_t funcs[MAX_POSTPONED_JOB];
    void *data[MAX_POSTPONED_JOB];
} ppj;

static bool
ppj_enable(int i)
{
    VM_ASSERT(i < MAX_POSTPONED_JOB);

    while (1) {
        ppj_bits_t bits = ppj.enabled.bits;
        if (bits & (1UL << i)) {
            return false;
        }
        if (RUBY_ATOMIC_PTR_CAS(ppj.enabled.ptr_bits, (void *)bits, (void *)(bits | (1UL << i))) == (void *)bits) {
            return true;
        }
    }
}

static bool
ppj_disable(int i)
{
    VM_ASSERT(i < MAX_POSTPONED_JOB);

    while (1) {
        ppj_bits_t bits = ppj.enabled.bits;
        if (bits & (1 << i)) {
            if (RUBY_ATOMIC_PTR_CAS(ppj.enabled.ptr_bits, (void *)bits, (void *)(bits & ~(1UL << i))) == (void *)bits) {
                return true;
            }
        }
        else {
            return false;
        }
    }
}

static bool
ppj_enabled_p(int i)
{
    VM_ASSERT(i < MAX_POSTPONED_JOB);
    return ppj.enabled.bits & (1UL << i);
}

// return 0 if registration is failed.
int
rb_postponed_job_register_one(unsigned int flags, rb_postponed_job_func_t func, void *data)
{
    for (int i=0; i<MAX_POSTPONED_JOB; i++) {
        if (ppj.funcs[i] == func) { // use existing func
            if (ppj_enabled_p(i)) return FALSE; // already enabled
            ppj.data[i] = data;
            return ppj_enable(i); // racy because data is not consistent
        }
        else if (ppj.funcs[i] == NULL) { // register new func
            if (RUBY_ATOMIC_PTR_CAS(ppj.funcs[i], NULL, func) == NULL) {
                ppj.data[i] = data;
                return ppj_enable(i); // racy because data is not consistent
            }
            else {
                // can not allocate. try next slot
            }
        }
    }

    return FALSE; // full
}

int
rb_postponed_job_register(unsigned int flags, rb_postponed_job_func_t func, void *data)
{
    return rb_postponed_job_register_one(flags, func, data);
}

// new API

// -1: no room to preregister function
// otherwise: postponed_job index which can be used from rb_postponed_job_activate()
int
rb_postponed_job_preregister(unsigned int flags, rb_postponed_job_func_t func, void *data)
{
    for (int i=0; i<MAX_POSTPONED_JOB; i++) {
        if (ppj.funcs[i] == func) { // use existing func
            ppj.data[i] = data;
            return i;
        }
        else if (ppj.funcs[i] == NULL) { // register new func
            ppj.funcs[i] = func;
            ppj.data[i] = data;
            return i;
        }
    }

    return -1;
}

int
rb_postponed_job_activate(int i)
{
    return ppj_enable(i);
}

void
rb_postponed_job_flush(rb_vm_t *vm)
{
    int i;

    while (UNLIKELY(ppj.enabled.bits)) {
        for (i=0; i<MAX_POSTPONED_JOB; i += 1) {
            if (ppj_enabled_p(i)) {
                rb_postponed_job_func_t func = ppj.funcs[i];
                void *data = ppj.data[i];
                if (!ppj_disable(i)) rb_bug("should not be conflicted");

                (*func)(data);

                if (!ppj.enabled.bits) return;
            }
        }
    }
}
```

KJTsanaktsidis:
So based on the codesearch.... _none_ of the gems calls register with a non-fixed callback function, and only one (`gctools`) does so with a non-fixed data. And as for `gctools`, that gem is a) last updated in 2016, and b) could use workqueue instead since it's not in a signal handler.


_technically speaking_, the solarwinds & appoptics usages of `register_postponed_job not-once cannot be replaced with `_once` in a non-breaking way. It uses this in a signal handler to collect profiling samples; if a thread spends a _really_ long time between calls to RUBY_VM_CHECK_INTS, its possible currently for _two_ samples to be collected with the same stack, which is _probably_ indicative of something very slow you might care about in APM. If we care about this, we could use an array of `rb_atomic_t`'s for the "triggered count" of each func, rather than a bitset.

If you used every single one of these gems (lol), you would have 16 callbacks. So using a fixed-size table of func/data pairs, where once it is registered it permanently occupies a slot (i.e. what you proposed in your latest message), seems fine. I thought about using a simple hashtable to store these, but since the common case is that you will have literally zero or one or maybe two (stackprof uses two different funcs), i guess iteration will just be simpler & faster.

If we do this, I think we should expose the workqueue API publicly though - provide a flexible way to register jobs from threads, and document postponed_job as being more limited and really only meant for signal handlers.

Should I put together a third PR for this?

ko1:
Thank you.

> but since the common case is that you will have literally zero or one or maybe two (stackprof uses two different funcs), i guess iteration will just be simpler & faster.

and 2 for GC, RJIT.

> If we do this, I think we should expose the workqueue API publicly though - provide a flexible way to register jobs from threads, and document postponed_job as being more limited and really only meant for signal handlers.

I think workqueue is not needed until someone requests it.

KJTsanaktsidis:
> and 2 for GC and RJIT.

I think it makes to move these to workqueue though anyway, because they don't need to be async-signal-safe. That way we keep the rule: postponed_job for async-signal-safe, workqueue otherwise

ko1:
> I think it makes sense to move these to workqueue though anyway, because they don't need to be async-signal-safe. That way we keep the rule: postponed_job for async-signal-safe, workqueue otherwise

My proposal means drop workqueue implementation to reduce the code. I think there is no performance impact with only (limited) postponed_job (if the number of jobs are enough small).

KJTsanaktsidis:
> My proposal means drop workqueue implementation to reduce the code. I think there is no performance impact with only (limited) postponed_job (if the number of jobs is enough small). Do you have any concern?

RJIT has some usages of workqueue which require variable data - in `rb_rjit_cme_invalidate` and `rb_rjit_constant_state_changed`. That won't work with the "small table with fixed func/data pairs" we want to replace postponed_job with

KJTsanaktsidis:
Hm - it actually looks like Eric discovered this thread-unsafety in rb_register_postponed_job that I'm trying to solve five years ago: https://bugs.ruby-lang.org/issues/15316

That's why these workqueue APIs exist; MJIT was the first thing, really, to actually pass variable data to postponed_jobs from threads without the GVL i thin.

KJTsanaktsidis:
@ko1 OK, I pushed up another commit here which I think implements roughly what you were talking about. I also tried to do my best to document the overall theme of the postponed job APIs I added in the debug.h header.

Is this roughly what you're thinking?

(I still need to fix the conflicts & check that it works with stackprof, but :crossed_fingers: it should be OK since the tests worked how I expected)

KJTsanaktsidis:
OK, I rebased this and it should be ready for another look. I verified that stackprof still installs & works correctly without complaint.

@tenderlove pinging you since you seem to be the most active maintainer of stackprof... can you see any problems in this PR with:

1. the proposed new "preregistration" API: https://github.com/ruby/ruby/pull/8949/files#diff-0a664fffd3b89256eb738e146535994539ffc48424e16ace9c75506351ebe00cR705
2. the proposed compatability implementation of `rb_postponed_job_register` in terms of `rb_postponed_job_preregister` and `rb_postponed_job_trigger` https://github.com/ruby/ruby/pull/8949/files#diff-3f265ff4189b4876f627ad86f2feaf99df0bbb009ed2af54c84190133f41ca8dR1798

Thanks for taking a look :pray:

KJTsanaktsidis:
```
   Assertion Failed: ../src/shape.c:810:shape_cache_get_iv_index:!shape_get_iv_index(shape, id, value)
```

hmm... is this something known to be flaky or should I investigate this?

ko1:
* I'm okay to remain `rb_workqueue_register` for now but I think we can remove it by replacing with new postponed job interface. I'll revisit it and measure it after ruby 3.3 release.
* Do not introduce new interface (functionality) if there is no requirements (`rb_postponed_job_enqueue` and `rb_postponed_job_enqueue_one`).
  * Do not use `rb_gc_disable()` to this purpose because it introduce performance impact on GC.

vm_trace.c
```diff
@@ -1617,197 +1621,289 @@ Init_vm_trace(void)
...
-    else {
-        return PJRR_INTERRUPTED;
+    rb_postponed_job_queues_t *pjq = GET_VM()->postponed_job_queues;
+    for (unsigned int i = 0; i < PJOB_PREREG_TABLE_SIZE; i++) {
+        /* Needs to be atomic: must synchronize with RUBY_ATOMIC_PTR_CAS below */
+        rb_postponed_job_func_t existing_func = (rb_postponed_job_func_t)RUBY_ATOMIC_PTR_LOAD(pjq->prereg_table[i].func);
+        /* Needs to be atomic: must synchronize with RUBY_ATOMIC_PTR_EXCHANGE below */
+        void *existing_data = RUBY_ATOMIC_PTR_LOAD(pjq->prereg_table[i].data);
+cas_retry:
+        if (existing_func == func && existing_data == data) {
```
ko1, line 1759: should we check existing_data too?

include/ruby/debug.h
```diff
@@ -615,50 +616,202 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
...
+ *                     `rb_postponed_job_enqueue`.
  */
 typedef void (*rb_postponed_job_func_t)(void *arg);
 
 /**
- * Registers a postponed job.
+ * The type of a handle returned from `rb_postponed_job_preregister` and
+ * passed to `rb_postponed_job_trigger`a
+ */
+typedef unsigned int rb_postponed_job_handle;
```
ko1, line 679: on ruby code `_t` suffix is popular for this kind of type naming.

vm_trace.c
KJTsanaktsidis: That is what I documented in the header file:

```
* Pre-registers a (func, data) pair in Ruby's postponed job preregistration table,
 * returning an opaque handle which can be used to trigger the job later. This function
 * must be called under the GVL (generally, it will be called during the initialization
 * routine of an extension).
 *
 * If the given (func,data) pair was already pre-registered, this method will return
 * the same handle instance as was previously returned.
 ```

Honestly I wrote the documentation first, and this "felt" like the most obvious API design, to think in terms of `(func, data)` pairs. Otherwise, you'd need to write something like "If this function is called multiple times with the same func and different data, the data argument is ignored and it returns the same handle which was pre-registered the first time `func` was registered", which just seems a little wordier.

**Comment by KJTsanaktsidis** [comment]
> I'm okay to remain rb_workqueue_register for now but I think we can remove it by replacing with new postponed job interface. I'll revisit it and measure it after ruby 3.3 release.
> Do not introduce new interface (functionality) if there is no requirements (rb_postponed_job_enqueue and rb_postponed_job_enqueue_one). 

OK - it sounds like the best thing to do here is to leave workqueue exactly how it is. I had tried to fold workqueue & postponed job together into one thing, since they're so similar, but this refactoring is doing more harm than good it seems. I've removed all the workqueue-related changes from my PR; it works exactly like it did before now, and it's used in exactly the places it was before, and the `rb_postponed_job_enqueue` methods which functioned like workqueue are now gone.


>   Do not use rb_gc_disable() to this purpose because it introduce performance impact on GC.

This actually was needed for `rb_postponed_job_enqueue` because it would have been possible for a call to `ruby_xmalloc` to trigger a GC which might try and enqueue a finalizer and recursively call `rb_postponed_job_enqueue`, and thus deadlock. But in any case this is now gone because all the workqueue changes are gone.

Thanks again for your patience, I think this should be ready again for another look :pray:

File: vm_trace.c (hunk 5 above)
Comment (ko1) [reply vm_trace.c:1762]: > "If this function is called multiple times with the same func and different data, the data argument is ignored and it returns the same handle which was pre-registered the first time func was registered", which just seems a little wordier.

Yes. This is what I assume. Otherwise, 64 (or 32) entries are not enough. Now we confirmed that the number of `func` is enough small. If we allow multiple `data` we can not register finalizers for example. This is not acceptable.

Comment (ko1) [reply vm_trace.c:1762]: (for finalizer (and other important `func`s) will be reserved by preregister, though)

Comment (KJTsanaktsidis) [reply vm_trace.c:1762]: that's right - the internal usages of this are pre-registered during early boot, before any user code or other extensions can run - so extensions can't possibly starve out Ruby's own need for postponed job slots.

Comment (ko1) [reply vm_trace.c:1762]: Ruby 3.3 RC1 will be released soon so:

* you will change the spec.
* merge this commit and change the spec after that by me.

which do you want?

I'm sorry to make non-frequent responses.

Comment (KJTsanaktsidis) [reply vm_trace.c:1762]: So about this code example... (note. I purposely changed `i * 1000` to `i + 1000` to make something clearer)

```
  // 1st iteration
  for (i=0; i<100; i++) rb_postponed_job_register(a_func, i);
  // 2nd iteration
  for (i=0; i<100; i++) rb_postponed_job_register(a_func, i + 1000);
```

I definitely understand how my change as it currently is will break this. What I don't think I understand is how you _want_ this to behave.

Right now, on master, this code will cause `a_func` to be called 100 times with the numbers 0 ... 100, and then 100 more times with the numbers 1000...1100.

What do you want the behaviour of this to be?

1. make `a_func` be called 100 times with argument `0`, and then, later on,  100 more times with argument `0`
2. make `a_func` be called one time with argument `0` , and then, later on, one more time with argument `0`
3. make `a_func` be called one time with argument `0`, and then, later on, one more time with argument `1000`
4. make `a_func` be called ~32 times with the numbers 0..32 as an argument, and then never again after that? (that's what my current patch is going to do)

I think if we remove ` && existing_data == data` from the check on line 1759, the behaviour will be option 2.

I guess what worries me is that anything actually using `rb_postponed_job_register` is going to probably wind up doing something very different than the author originally intended with any of these options. 

> Ruby 3.3 RC1 will be released soon 

To step back a bit, maybe what we should do is:

* Wait for Ruby 3.3 to be released,
* Early next year, merge the circular, lock-free ringbuffer implementation I originally proposed. I definitely agree it was too complicated to merge in December a month before Ruby 3.3, but maybe it's OK to merge in January?
* Gives us plenty of time to stabilise it and work out the bugs, or revert it if it's becoming a problem.
* Because the ringbuffer implementation keeps the existing semantics of both `rb_postponed_job_register` and `rb_postponed_job_register_one`, we can consider it for backport into a point release once we're happy with it.

What do you think? Should we leave this until after 3.3? As far as I can tell I'm the only one who ran into the signal-safety crash in the first place, and this bug will only affect people using profiling tools in production, so it's probably not worth rushing it into 3.3 I think.

Comment (ko1) [reply vm_trace.c:1762]: > make a_func be called one time with argument 0, and then, later on, one more time with argument 1000

Let decide to choose option 3. I proposed it.

>  I guess what worries me is that anything actually using rb_postponed_job_register is going to probably wind up doing something very different than the author originally intended with any of these options.

Yes the behavior will be changed. However the size of queue is implementation details and this change is 1000 -> 1. It is not a semantic change but a parameter change (yes, it change the behavior).

> What do you think? Should we leave this until after 3.3? As far as I can tell I'm the only one who ran into the signal-safety crash in the first place, and this bug will only affect people using profiling tools in production, so it's probably not worth rushing it into 3.3 I think.

I want to introduce my proposal on Ruby 3.3.0. If someone blamed the change, we can revert fix on Ruby 3.3.1.

Comment (ko1) [reply vm_trace.c:1762]: > make a_func be called one time with argument 0, and then, later on, one more time with argument 1000

Sorry. Let's make clear:

* On iteration 1, register `a_func` with 0.
* one call `a_func(0)` on safe-place
* On iteration 2, register `a_func` with 1000.
* one call `a_func(1000)` on safe-place
* ...

(`one more time with argument 1000` is not clear, so I make it clear)

Comment (KJTsanaktsidis) [reply vm_trace.c]: @ko1 OK, I _think_ I have implemented that functionality correctly and pushed the change up. It still needs a couple more test cases I think, but otherwise does this look how you wanted it to?

Also: Do we need to open a discussion on Redmine about this kind of change, do you think?

File: include/ruby/debug.h [hunk 7]
```diff
@@ -615,48 +616,151 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
...
+ *
+ * This API is designed to be called from contexts where it is not safe to run Ruby
+ * code (e.g. because they do not hold the GVL or because GC is in progress), and
+ * defer a callback to run in a context where it _is_ safe. The primary intended
+ * users of this API is for sampling profilers like the "stackprof" gem; these work
+ * by scheduling the periodic delivery of a SIGPROF signal, and inside the C-level
+ * signal handler, deferring a job to collect a Ruby backtrace when it is next safe
+ * to do so.
+ *
+ * Historically, this API provided two functions `rb_postponed_job_register` and
```
ko1, line 628: In general historical reason and functions for compatibility should not be located at the beginning.
Recommended functions and usage should be at the top on this kind of document.
Please consider to reorder the doc later.

include/ruby/debug.h
```diff
@@ -615,48 +616,151 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
...
+ * If this function is called concurrently with the same `func`, then the stored data
+ * could be the value from either call (but will definitely be one of them).
+ *
+ * If this function is called to update the data concurrently with a call to
+ * `rb_postponed_job_trigger` on the same handle, it's undefined whether `func` will
+ * be called with the old data or the new data.
+ *
+ * Although the current implementation of this method is in fact async-signal-safe and
+ * has defined semantics when called concurrently on the same `func`, a future Ruby
+ * version might require that this method be called under the GVL; thus, programs which
```
ko1, line 686: s/method/function/

include/ruby/debug.h
```diff
@@ -615,48 +616,151 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
...
- * function  registers a  new one.   The registered  job would  eventually gets
- * executed.
+ * Like `rb_postponed_job_trigger`, the context in which the job is called
+ * holds the GVL and can allocate Ruby objects.
  *
- * @param[in]      flags      (Unused) reserved for future extensions.
+ * This method essentially has the same semantics as:
+ *
+ * ```
+ *   rb_postponed_job_trigger(rb_postponed_job_preregister(func, data));
```
ko1, line 726: This is not correct because of 2nd call.
I recommend you to remove about it.

include/ruby/debug.h
```diff
@@ -615,48 +616,151 @@ VALUE rb_tracearg_object(rb_trace_arg_t *trace_arg);
...
+ * ```
+ *
+ * @note    Prevoius versions of Ruby promised that the (`func`, `data`) pairs would
+ *          be executed as many times as they were registered with this function; in
+ *          reality this was always subject to race conditions and this function no
+ *          longer provides this guarantee. Instead, we only promise that `func` will
+ *          be called once.
```
ko1, line 733: `once` is not clear. "once" in safe-point interval but not "once" in interpreter lifetime.

include/ruby/debug.h
ko1: Ah, it's my mistake. I didn't check the 2nd call of `rb_postponed_job_preregister(func, data))` with different `data`. 
(I assume it doesn't change the 2nd `data` but proposed spec override the old one).
Proposed spec is acceptable (and it is deprecated. no need to care more)

vm_trace.c
```diff
@@ -1701,83 +1709,113 @@ get_valid_ec(rb_vm_t *vm)
...
+rb_postponed_job_register_one(unsigned int flags, rb_postponed_job_func_t func, void *data)
+{
+    return pjob_register_legacy_impl(flags, func, data);
+}
+
+
 void
 rb_postponed_job_flush(rb_vm_t *vm)
 {
+    rb_postponed_job_queues_t *pjq = GET_VM()->postponed_job_queue;
```
ko1, line 1816: why do you use VM? (why not global variable?)

vm_trace.c
```diff
@@ -1645,52 +1654,51 @@ rb_vm_memsize_workqueue(struct ccan_list_head *workqueue)
...
+#define PJOB_PREREG_TABLE_SIZE              (sizeof(rb_atomic_t) * CHAR_BIT)
+/* pre-registered jobs table, for async-safe jobs */
+typedef struct rb_postponed_job_queue {
+    struct {
+        rb_postponed_job_func_t func;
+        void *data;
+    } prereg_table[PJOB_PREREG_TABLE_SIZE];
+    /* Bits in this are set when the corresponding entry in prereg_table has non-zero
+     * triggered_count; i.e. somebody called rb_postponed_job_trigger */
+    rb_atomic_t prereg_triggered_bitset;
```
ko1, line 1690: `prereg_` prefix seems verbose.

vm_trace.c
ko1: Thank you for the long discussion. I'll merge it.

One thing I find we need `flag` parameter for future change but I'll fix it later.

Making redmine ticket is good idea. Could you summary the fixed spec and reasons?

ko1: I added some trivial comments so please refer to refine if they make sense.

ko1:
Thank you for the long discussion and I think the implementation is enough small to make maintenance easy

vm_trace.c
KJTsanaktsidis: The previous `postponed_job_buffer` was stored as a field on the VM struct so I put the new struct there as well.

PR #8950, opened 2023-11-20
Add RUBY_ASSERT_GC_FREE_REGION

casperisfine:
In some situation, we need to perform a set of operations while ensuring GC doesn't trigger in the middle. The typical example is writting some reference and later trigger write barriers.

If GC triggers in the middle it may lead to memory corruption.

Currently this is hard to test for, the only way is to write a test case with GC put in stress mode, but it's slow and not 100% reliable.

With `RUBY_ASSERT_GC_FREE_REGION`, we can make it an assertion:

```c
void
rb_obj_copy_ivs_to_hash_table(VALUE obj, st_table *table)
{
    RUBY_ASSERT_GC_FREE_REGION_START();
    rb_ivar_foreach(obj, rb_obj_copy_ivs_to_hash_table_i, (st_data_t)table);
    RUBY_ASSERT_GC_FREE_REGION_END();
}
```

@peterzhu2118 what do you think?

In such section, any operation that could potentially trigger GC must fail an assertion.


peterzhu2118 (commented):
The example you give could trigger GC because `rb_obj_copy_ivs_to_hash_table_i` calls `st_insert` which allocates memory. In this case we need to ensure that it's in a state that's safe for GC (which, as you said, is harder to ensure) rather than checking that GC does not run.

casperisfine:
> The example you give could trigger GC

Yeah, I know, it's not 100% perfect, but I think it's better an more reliable than GC stress? https://github.com/ruby/ruby/pull/8932/files#diff-527698bc9a2348f784821fdfcb904c66c81fc43b1fa64723c4a44aa24832415fR278

gc.c
```diff
@@ -1177,6 +1180,26 @@ VALUE *ruby_initial_gc_stress_ptr = &ruby_initial_gc_stress;
 #define dont_gc_val()         (objspace->flags.dont_gc)
 #endif
 
+#if RUBY_DEBUG
+void
+RUBY_ASSERT_GC_FREE_REGION_START(void)
```
ko1, line 1185: s/START/BEGIN/ for Ruby's terminology?

casperisfine: I modeled it after `DURING_GC_COULD_MALLOC_REGION_START`, but yes, I absolutely don't mind renaming.

casperisfine: Renamed as `RUBY_ASSERT_GC_FREE_REGION_BEGIN`

PR #8910, opened 2023-11-13
Use Variable Width Allocation on procs 

wildmaples:
```ruby
require 'benchmark'

a = 100_000.times.map { proc{} }
puts Benchmark.measure { 1_000.times { a.each { |p| p.call } } }
``` 

this change
```ruby
  3.856084   0.006743   3.862827 (  3.863474) 
```

old 
```ruby
  3.895393   0.013311   3.908704 (  3.909022)
```

Co-authored-by: Peter Zhu <peter.zhu@shopify.com>

PR #8959, opened 2023-11-20
Use Variable Width Allocation on procs

casperisfine:
Continuation of https://github.com/ruby/ruby/pull/8910

Trying to see if I can figure out the CI crashes.

```
compare-ruby: ruby 3.3.0dev (2023-11-20T02:02:55Z master 701b0650de) [arm64-darwin22]
last_commit=[ruby/prism] feat: add encoding for IBM865 (https://github.com/ruby/prism/pull/1884)
built-ruby: ruby 3.3.0dev (2023-11-20T17:47:00Z vwa-proc 3b295e4643) [arm64-darwin22]
warming up.
# Iteration per second (i/s)

|            |compare-ruby|built-ruby|
|:-----------|-----------:|---------:|
|proc_alloc  |      7.791M|   10.633M|
|            |           -|     1.36x|
```

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2500,7 +2500,7 @@ rb_fiber_start(rb_fiber_t *fiber)
         rb_context_t *cont = &VAR_FROM_MEMORY(fiber)->cont;
         int argc;
         const VALUE *argv, args = cont->value;
-        GetProcPtr(fiber->first_proc, proc);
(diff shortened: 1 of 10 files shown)


byroot:
Closing for now as we'd either need to pin all Proc instance or to refactor quite a bit of code so we no longer hold pointers into inside the Proc memory. So perhaps for 3.4, but not now.

PR #8961, opened 2023-11-20
[PRISM] Fix LocalVariableWriteNodes within blocks

jemmaissroff:
Prior to this commit, we weren't recursing up scopes to look for the local definition. With this commit, we do so, fixing local writes within blocks

Closes https://github.com/ruby/prism/issues/1890

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -2883,8 +2883,7 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         pm_local_variable_target_node_t *local_write_node = (pm_local_variable_target_node_t *) node;
 
         pm_constant_id_t constant_id = local_write_node->name;
-        for (size_t index = 0; index < local_write_node->depth; index++) scope_node = scope_node->previous;
-        int index = pm_lookup_local_index(iseq, scope_node, constant_id);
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Do you need to do local targets too?

prism_compile.c
```diff
@@ -2863,10 +2883,9 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
         pm_local_variable_target_node_t *local_write_node = (pm_local_variable_target_node_t *) node;
 
         pm_constant_id_t constant_id = local_write_node->name;
-        for (size_t index = 0; index < local_write_node->depth; index++) scope_node = scope_node->previous;
-        int index = pm_lookup_local_index(iseq, scope_node, constant_id);
+        int index = pm_lookup_local_index_any_scope(iseq, scope_node, constant_id);
 
-        ADD_SETLOCAL(ret, &dummy_line_node, (int)index, local_write_node->depth);
+        ADD_SETLOCAL(ret, &dummy_line_node, index, local_write_node->depth);
```
eregon, line 2888: It seems a bit odd to use and trust depth but not use it fir the index.

What was the problem with the previous code?
The PR description says `we weren't recursing up scopes to look for the local definition`, but old line 2866 was recursing up scopes via `scope_node = scope_node->previous`, no?

Maybe the depth is wrong in some cases?

PR #8975, opened 2023-11-21
Refactor NameError::message and make it embeded

casperisfine:
These aren't particularly common, but avoiding the malloc churn for small types is always nice, and this commit also modernize and cleanup the TypedData API usage.

@peterzhu2118

### DIFF
--- a/error.c
+++ b/error.c
@@ -2202,50 +2202,50 @@ rb_nomethod_err_new(VALUE mesg, VALUE recv, VALUE method, VALUE args, int priv)
     return nometh_err_init_attr(exc, args, priv);
 }
 
-/* :nodoc: */
-enum {
-    NAME_ERR_MESG__MESG,
-    NAME_ERR_MESG__RECV,
-    NAME_ERR_MESG__NAME,
-    NAME_ERR_MESG_COUNT
-};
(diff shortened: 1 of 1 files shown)


error.c
```diff
@@ -2267,34 +2267,33 @@ name_err_mesg_alloc(VALUE klass)
...
     rb_obj_init_copy(obj1, obj2);
 
-    TypedData_Get_Struct(obj1, VALUE, &name_err_mesg_data_type, ptr1);
-    TypedData_Get_Struct(obj2, VALUE, &name_err_mesg_data_type, ptr2);
-    MEMCPY(ptr1, ptr2, VALUE, NAME_ERR_MESG_COUNT);
+    name_error_message_t *ptr1, *ptr2;
+    TypedData_Get_Struct(obj1, name_error_message_t, &name_err_mesg_data_type, ptr1);
+    TypedData_Get_Struct(obj2, name_error_message_t, &name_err_mesg_data_type, ptr2);
+
+    MEMCPY(ptr1, ptr2, VALUE, sizeof(name_error_message_t));
```
peterzhu2118, line 2277: I think you need to fire write barriers here rather than use `memcpy`.

PR #8976, opened 2023-11-21
Get rid of env_data_type

casperisfine:
Defining `ENV` as a T_DATA doesn't give us anything.


casperisfine:
Ok, so being a T_DATA somehow allows it to be ractor shareable. I standcorrected.

PR #8978, opened 2023-11-21
Embed `bf_dump` compiler objects

casperisfine:
They are both quite small and rather ephemeral, so no reason not to embed them.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -13190,14 +13190,13 @@ ibf_dump_free(void *ptr)
         st_free_table(dump->iseq_table);
         dump->iseq_table = 0;
     }
-    ruby_xfree(dump);
 }
 
 static size_t
 ibf_dump_memsize(const void *ptr)
 {
     struct ibf_dump *dump = (struct ibf_dump *)ptr;
-    size_t size = sizeof(*dump);
+    size_t size = 0;
     if (dump->iseq_table) size += st_memsize(dump->iseq_table);
     if (dump->global_buffer.obj_table) size += st_memsize(dump->global_buffer.obj_table);
     return size;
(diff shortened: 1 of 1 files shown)


casperisfine:
@peterzhu2118

compile.c
```diff
@@ -13270,7 +13269,6 @@ rb_iseq_ibf_dump(const rb_iseq_t *iseq, VALUE opt)
 
     str = dump->global_buffer.str;
     ibf_dump_free(dump);
```
peterzhu2118, line 13271: Since we are no longer setting the pointer to `NULL`, we will have a double free here since the GC will free everything again.

```suggestion
```

casperisfine: Actually, the code handles that:
```c
static void
ibf_dump_free(void *ptr)
{
    struct ibf_dump *dump = (struct ibf_dump *)ptr;
    if (dump->global_buffer.obj_table) {
        st_free_table(dump->global_buffer.obj_table);
        dump->global_buffer.obj_table = 0;
    }
    if (dump->iseq_table) {
        st_free_table(dump->iseq_table);
        dump->iseq_table = 0;
    }
}
```

casperisfine: I'm not sure if it's really worth keeping that early free thought.

peterzhu2118: IMO it's a better idea to leave freeing to the GC.

casperisfine: Alright, I suspected it was added for perf reasons, hence why I was hesistant, but after blaming it appears it's a "left-over": https://github.com/ruby/ruby/commit/2c3f0749c56a1a094ca7be12e669f0d1545bb756

I'll remove it.

PR #8980, opened 2023-11-21
Fix memory leak when evacuating generic ivars

peterzhu2118:
The lookup in the table is using the wrong key when converting generic instance variables to too complex, which means that it never looks up the entry which leaks memory when the entry is overwritten.

cc. @byroot

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1379,7 +1379,7 @@ rb_obj_convert_to_too_complex(VALUE obj, st_table *table)
         RB_VM_LOCK_ENTER();
         {
             struct st_table *gen_ivs = generic_ivtbl_no_ractor_check(obj);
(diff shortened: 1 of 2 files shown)


variable.c
```diff
@@ -1379,7 +1379,7 @@ rb_obj_convert_to_too_complex(VALUE obj, st_table *table)
         RB_VM_LOCK_ENTER();
         {
             struct st_table *gen_ivs = generic_ivtbl_no_ractor_check(obj);
-            st_lookup(gen_ivs, (st_data_t)&obj, (st_data_t *)&old_ivptr);
+            st_lookup(gen_ivs, (st_data_t)obj, (st_data_t *)&old_ivptr);
```
byroot, line 1382: Urk.

PR #8981, opened 2023-11-21
VWA Embed the rest of the data objects in Enumerator

eightbitraptor:


enumerator.c
```diff
@@ -1309,7 +1309,7 @@ static const rb_data_type_t yielder_data_type = {
         yielder_memsize,
```
peterzhu2118, line 1309: The `memsize` functions can be removed since the size of the struct is calculated by the GC when `RUBY_TYPED_EMBEDDABLE` is enabled. The `memsize` function should only report memory used outside of the allocated struct.

PR #8982, opened 2023-11-21
Fix compacting during evacuation of generic ivars 

peterzhu2118:
When evacuating generic instance variables, the instance variables exist in both the array and the ST table. We need to ensure it has switched to the ST table before performing any operations that can trigger GC compaction.

Alternate fix to #8932.

cc. @byroot

### DIFF
--- a/test/ruby/test_shapes.rb
+++ b/test/ruby/test_shapes.rb
@@ -256,6 +256,85 @@ class TestShapes < Test::Unit::TestCase
     end;
   end
 
+  def test_evacuate_class_ivar_and_compaction
+    assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
+    begin;
+      count = 20
+
+      c = Class.new
+      count.times do |ivar|
+        c.instance_variable_set("@i#{ivar}", "ivar-#{ivar}")
+      end
+
+      RubyVM::Shape.exhaust_shapes
+
+      GC.auto_compact = true
+      GC.stress = true
+      # Cause evacuation
+      c.instance_variable_set(:@a, o = Object.new)
+      assert_equal(o, c.instance_variable_get(:@a))
+      GC.stress = false
+
+      count.times do |ivar|
+        assert_equal "ivar-#{ivar}", c.instance_variable_get("@i#{ivar}")
+      end
+    end;
+  end
+
+  def test_evacuate_generic_ivar_and_compaction
+    assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
+    begin;
+      count = 20
+
+      c = Hash.new
+      count.times do |ivar|
+        c.instance_variable_set("@i#{ivar}", "ivar-#{ivar}")
+      end
+
+      RubyVM::Shape.exhaust_shapes
+
+      GC.auto_compact = true
+      GC.stress = true
+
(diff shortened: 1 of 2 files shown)


byroot:
It seems to end up in infinite recursion sometimes:

```
   | /home/runner/work/ruby/ruby/build/ruby(rb_shape_traverse_from_new_root+0x7a) [0x564ed76c1b0a] ../src/shape.c:818
  | /home/runner/work/ruby/ruby/build/ruby(rb_shape_traverse_from_new_root+0x7a) [0x564ed76c1b0a] ../src/shape.c:818
  | /home/runner/work/ruby/ruby/build/ruby(rb_shape_traverse_from_new_root+0x7a) [0x564ed76c1b0a] ../src/shape.c:818
  | /home/runner/work/ruby/ruby/build/ruby(rb_shape_traverse_from_new_root+0x7a) [0x564ed76c1b0a] ../src/shape.c:818
  | /home/runner/work/ruby/ruby/build/ruby(rb_shape_traverse_from_new_root+0x7a) [0x564ed76c1b0a] ../src/shape.c:818
```

That's surprising.

peterzhu2118:
It's probably not an infinite recursion but rather the stack gets too deep because it's recursive based on the number of ivars, and that test adds a lot of ivars to one object. I will look into a solution.

byroot:
Ah, I see. Yeah, in the meantime you can likely work around this issue by using the new `exhaust_shapes` helper. This way you avoid defining a object with 30k ivars that then need to be transitioned.

XrXr:
`Assertion Failed: ../src/variable.c:1646:rb_obj_copy_ivs_to_hash_table_i:!st_lookup((st_table *)arg, (st_data_t)key, NULL)` in i686

Is this indication of movement while copying out ivars? I guess `st_add_direct()` could still trigger GC.
Also, is this supposed to fix too_complex transition for T_CLASS and T_OBJECT in addition to generic ivars?

peterzhu2118:
> Is this indication of movement while copying out ivars?

I don't think so. `iterate_over_shapes_with_callback` caches the shape so even if the shape changes (which it shouldn't), it shouldn't affect anything. It's an indication that there's two nodes in the shape tree with the same instance variable name (which is not allowed). So there's a bug somewhere that causes that to happen.

> I guess st_add_direct() could still trigger GC.

It should not trigger GC because we allocate the table with enough size. If it did, then there would be an issue with how Hashes transition from AR -> ST as well.

> Also, is this supposed to fix too_complex transition for T_CLASS and T_OBJECT in addition to generic ivars?

I'm not sure if Classes and Objects have a problem. Are you able to find an issue in those code paths?

XrXr:
> I'm not sure if Classes and Objects have a problem. Are you able to find an issue in those code paths?

I've seen a bunch of `mark T_NONE` assertion failures on classes in core dumps, so I suspected that something is up there. Also, I bring it up since this diff add tests evacuation tests for classes and objects.

I've seen the `st_insert()` in `rb_obj_copy_ivs_to_hash_table_i()` trigger GC too but maybe that was fixed with 81b35fe7297 when the table size is changed to use `rb_ivar_count()`.

So, sorry, maybe I'm working under false pretenses. I'm going to check my assumptions again.

PR #8989, opened 2023-11-21
Remove gen_ivtbl_count

tenderlove:
We can calculate number of IVs from the shape of the object.


variable.c
```diff
@@ -1968,13 +1948,16 @@ rb_copy_generic_ivar(VALUE clone, VALUE obj)
 
     rb_check_frozen(clone);
 
-    if (!FL_TEST(obj, FL_EXIVAR)) {
-        goto clear;
+    if (BUILTIN_TYPE(obj) <= T_MODULE) {
```
byroot, line 1951: Is `BUILTIN_TYPE(obj) <= ` a common thing? I can see any other occurrence of it, and it seems a bit risky to couple that with the current values of the various T_* enums.

variable.c
```diff
@@ -2080,10 +2062,13 @@ rb_ivar_count(VALUE obj)
         return RCLASS_IV_COUNT(obj);
       default:
         if (FL_TEST(obj, FL_EXIVAR)) {
-            struct gen_ivtbl *ivtbl;
-
-            if (rb_gen_ivtbl_get(obj, 0, &ivtbl)) {
-                return gen_ivtbl_count(obj, ivtbl);
+            if (rb_shape_obj_too_complex(obj)) {
+                struct gen_ivtbl *ivtbl;
+                rb_gen_ivtbl_get(obj, 0, &ivtbl);
```
Empact, line 2067: If `rb_gen_ivtbl_get` fails, `ivtbl` will not be set, so dereferencing it will be undefined behavior.

PR #8990, opened 2023-11-21
Fix compiler warnings in complex_pow_for_special_angle

kyanagi:
It seems that there are warnings with some compilers.
https://jinroq.hatenablog.jp/entry/2023/11/21/194058


kyanagi:
fixed by 31517c862760626c228b017a4a6e7c6003abeb84.

PR #8991, opened 2023-11-22
[Feature #20005] Add C API to resolve symbols from other extensions

tagomoris:
This is to add a C API for extensions to resolve symbols `extern`-ed by other extensions.

https://bugs.ruby-lang.org/issues/20005

### DIFF
--- a/dln.c
+++ b/dln.c
@@ -463,8 +463,8 @@ dln_symbol(void *handle, const char *symbol)
     }
     if (handle == NULL) {
 # if defined(USE_DLN_DLOPEN)
-        handle = dlopen(NULL, 0);
-# elif defined(_WIN32) && defined(RUBY_EXPORT)
+        handle = dlopen(NULL, RTLD_LAZY | RTLD_GLOBAL);
+# elif defined(_WIN32)
         handle = rb_libruby_handle();
 # else
         return NULL;
--- a/dln.c
+++ b/dln.c
@@ -419,33 +419,63 @@ dln_open(const char *file)
 static void *
 dln_sym(void *handle, const char *symbol)
 {
-    void *func;
-    const char *error;
-
 #if defined(_WIN32)
-    char message[1024];
-
-    func = GetProcAddress(handle, symbol);
-    if (func == NULL) {
-        error = dln_strerror();
-        goto failed;
-    }
-
+    return GetProcAddress(handle, symbol);
 #elif defined(USE_DLN_DLOPEN)
-    func = dlsym(handle, symbol);
+    return dlsym(handle, symbol);
+#endif
+}
+
+static void *
+dln_sym_func(void *handle, const char *symbol)
+{
+    void *func = dln_sym(handle, symbol);
+
     if (func == NULL) {
+        const char *error;
+#if defined(_WIN32)
+        char message[1024];
+        error = dln_strerror();
+#elif defined(USE_DLN_DLOPEN)
         const size_t errlen = strlen(error = dln_strerror()) + 1;
         error = memcpy(ALLOCA_N(char, errlen), error, errlen);
-        goto failed;
+#endif
+        dln_loaderror("%s - %s", error, symbol);
     }
-#endif
-
     return func;
-
-  failed:
-    dln_loaderror("%s - %s", error, symbol);
 }
+
+#define dln_sym_callable(rettype, argtype, handle, symbol) \
+    (*(rettype (*)argtype)dln_sym_func(handle, symbol))
 #endif
+
+void *
+dln_symbol(void *handle, const char *symbol)
+{
+#if defined(_WIN32) || defined(USE_DLN_DLOPEN)
+    if (EXTERNAL_PREFIX[0]) {
+        const size_t symlen = strlen(symbol);
+        char *const tmp = ALLOCA_N(char, symlen + sizeof(EXTERNAL_PREFIX));
+        if (!tmp) dln_memerror();
+        memcpy(tmp, EXTERNAL_PREFIX, sizeof(EXTERNAL_PREFIX) - 1);
+        memcpy(tmp + sizeof(EXTERNAL_PREFIX) - 1, symbol, symlen + 1);
+        symbol = tmp;
+    }
+    if (handle == NULL) {
+# if defined(USE_DLN_DLOPEN)
+        handle = dlopen(NULL, 0);
+# elif defined(_WIN32) && defined(RUBY_EXPORT)
+        handle = rb_libruby_handle();
+# else
+        return NULL;
+# endif
+    }
+    return dln_sym(handle, symbol);
+#else
+    return NULL;
+#endif
+}
+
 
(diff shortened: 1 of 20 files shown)


tagomoris:
I have no idea how we can test this kind of API. What's the best example?

tagomoris:
I got some idea about `ext/-test-` and `test/-ext-`. So I'll try them anyway.

KJTsanaktsidis:
That sounds right - make two modules in `ext/-test-`, and make one try and resolve symbols from the other. I think the  feature name for an extension `foo` in `ext/-test-/foo` will be `-test-/foo`

tagomoris:
I added extensions for tests under `ext/-test-` and also added a test case under `text/-ext-`. These look working correctly on my laptop.

load.c
```diff
@@ -1511,6 +1515,29 @@ rb_f_autoload_p(int argc, VALUE *argv, VALUE obj)
...
+    }
+    if (rb_ary_entry(resolved, 0) != ID2SYM(rb_intern("so"))) {
+        return NULL;
+    }
+    path = rb_ary_entry(resolved, 1);
+    handle = rb_hash_aref(ruby_dln_libmap, path);
+    if (NIL_P(handle)) {
+        return NULL;
+    }
+    return dln_symbol((void *)NUM2LONG(handle), symbol);
```
nobu, line 1538:
```suggestion
    return dln_symbol((void *)NUM2SVALUE(handle), symbol);
```

load.c
```diff
@@ -1511,6 +1515,29 @@ rb_f_autoload_p(int argc, VALUE *argv, VALUE obj)
...
+
+    resolved = rb_resolve_feature_path((VALUE)NULL, fname_str);
+    if (NIL_P(resolved) || RARRAY_LEN(resolved) != 2) {
+        return NULL;
+    }
+    if (rb_ary_entry(resolved, 0) != ID2SYM(rb_intern("so"))) {
+        return NULL;
+    }
+    path = rb_ary_entry(resolved, 1);
+    handle = rb_hash_aref(ruby_dln_libmap, path);
```
nobu, line 1534: `rb_hash_aref` unnecessarily uses the default value.
 
```suggestion
    handle = rb_hash_lookup(ruby_dln_libmap, path);
```

load.c
```diff
@@ -1511,6 +1515,29 @@ rb_f_autoload_p(int argc, VALUE *argv, VALUE obj)
...
+{
+    VALUE handle;
+    VALUE resolved;
+    VALUE path;
+    VALUE fname_str = rb_str_new_cstr(fname);
+
+    resolved = rb_resolve_feature_path((VALUE)NULL, fname_str);
+    if (NIL_P(resolved) || RARRAY_LEN(resolved) != 2) {
+        return NULL;
+    }
```
nobu, line 1529: You might need to consider statically linked extensions here.

nobu (changes requested):

dln.c
```diff
@@ -447,6 +447,33 @@ dln_sym(void *handle, const char *symbol)
...
+#if defined(_WIN32) || defined(USE_DLN_DLOPEN)
+    if (EXTERNAL_PREFIX[0]) {
+        const size_t symlen = strlen(symbol);
+        char *const tmp = ALLOCA_N(char, symlen + sizeof(EXTERNAL_PREFIX));
+        if (!tmp) dln_memerror();
+        memcpy(tmp, EXTERNAL_PREFIX, sizeof(EXTERNAL_PREFIX) - 1);
+        memcpy(tmp + sizeof(EXTERNAL_PREFIX) - 1, symbol, symlen + 1);
+        symbol = tmp;
+    }
+    if (handle == NULL) {
```
tagomoris, line 462: The API to pass `NULL` as the `handle` argument looks dirty - should I add another argument to search the main executable explicitly?

dln.c
```diff
@@ -447,6 +447,33 @@ dln_sym(void *handle, const char *symbol)
...
+# if defined(USE_DLN_DLOPEN)
+        handle = dlopen(NULL, RTLD_FIRST);
+# else
+        // WIN32 seems not having similar feature to get the handle of the main process
```
tagomoris, line 466: Is this true? I couldn't find any way to do that.

include/ruby/internal/intern/load.h
```diff
@@ -176,6 +176,19 @@ VALUE rb_f_require(VALUE self, VALUE feature);
...
+/**
+ * Resolves and returns a symbol in the native extension specified by the
+ * feature and symbol names. Extensions will use this function to access the
+ * symbols provided by other native extensions.
+ *
+ * @param[in]  feature           Name of a feature, e.g. `"json"`.
+ * @param[in]  symbol            Name of a symbol defined by the feature.
+ * @return     The resolved symbol, defined and externed by the specified
+ *             feature. It may be NULL when the feature is not loaded, the
+ *             feature was not extension, or the symbol was not found.
```
tagomoris, line 188: Actually, the current implementation raises LoadError when the feature exists but the symbol is not found.
Should it suppress the LoadError?

load.c
tagomoris: I updated the code on the assumption that the feature name of statically linked extensions are also included in `$LOADED_FEATURES`. Is that true?

include/ruby/internal/intern/load.h
nobu: It is questionable to me.

dln.c
nobu: `rb_libruby_handle()` should return the handle to `libruby`.
Rather `RTLD_FIRST` seems platform dependent.

load.c
nobu: You can use `rb_feature_p()` instead, I guess.

include/ruby/internal/intern/load.h
tagomoris: Finally, I think that returning NULL in both cases (the feature not found and the symbol not found) should be consistent.

tagomoris:
CIs are failing on Windows environments. These show `rb_ext_resolve_symbol` is failing to resolve symbols. Hmm?

tagomoris:
And YJIT-ed runs crashed.

tagomoris:
About the CIs on Windows environments, the LoadError message `<LoadError: resolve_symbol_target is not loaded: 80>` explains that `GetProcAddress(handle, symbol);` returns NULL with a valid handle (not statically linked).

tagomoris:
I thought it was enough to specify `RUBY_EXTERN` to export a symbol from a .dll, but failing tests show it's not. Does it need to have `*.def` to declare `EXPORTS` explicitly? Or is there anything I'm missing?
I'll investigate more later.

tagomoris:
All tests passed!

tagomoris:
I've updated my commits and pushed again.
@nobu Could you review this change once more?

nobu (approved):

tagomoris:
Can this get merged into Ruby 3.3?

nobu (approved):

PR #8993, opened 2023-11-22
[Bug #19114] Fix for multiple calls of `TracePoint#enable`

kyanagi:
https://bugs.ruby-lang.org/issues/19114


vm_trace.c
```diff
@@ -1195,6 +1195,10 @@ rb_tracepoint_enable(VALUE tpval)
         rb_raise(rb_eArgError, "can't nest-enable a targeting TracePoint");
     }
 
+    if (tp->tracing) {
+        return Qundef;
```
tenderlove, line 1199: I think this should return `tpval` and not `Qundef`.

tenderlove (changes requested):
I don't think it's a good idea to return `Qundef` from functions that return a `VALUE`. Usually people assume that `VALUE` refers to a valid Ruby object and `Qundef` is not valid for a Ruby program to "see".

vm_trace.c
kyanagi: Originally, this method returned `Qundef`, so I chose the same return value for the newly added branch.
I think it would be possible to modify it to return `tpval`, or to return `Qtrue` or `Qfalse` depending on whether a new hook is added or not.
However, it would be necessary to confirm if such a change is acceptable.

tenderlove: Wow, you're totally right I didn't catch that.  I'll just merge this, thank you!

PR #8995, opened 2023-11-22
windows-arm64 support

pbo-linaro:
This series allows to build and run Ruby natively on windows-arm64.

It contains fixes for build system (some of them taken from [MSYS2 Ruby package](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-ruby)).
Most of the work is detection of `__pioinfo` pointer, which was the blocking step. Details are available in commit message. This was tested on windows-arm64 11 retail build (**not** insiders).

You can find a build log [here](https://gitlab.com/Linaro/windowsonarm/packages/ruby/-/jobs/5594478181), that uses this [script](https://gitlab.com/Linaro/windowsonarm/packages/ruby/-/blob/master/build_ruby.sh?ref_type=heads) from an MSYS2 clangarm64 shell. 

I'm not sure if Ruby community accepts a serie of patches, or if several PR are needed. Let me know.
Note: This is an effort part of my work at [Linaro](https://www.linaro.org/windows-on-arm/).

### DIFF
--- a/win32/win32.c
+++ b/win32/win32.c
@@ -2615,8 +2615,72 @@ set_pioinfo_extra(void)
      * * https://bugs.ruby-lang.org/issues/18605
      */
     char *p = (char*)get_proc_address(UCRTBASE, "_isatty", NULL);
-    char *pend = p;
     /* _osfile(fh) & FDEV */
+
+#ifdef _M_ARM64
+#define IS_INSN(pc, name) ((*(pc) & name##_mask) == name##_id)
+    const int max_num_inst = 500;
+    uint32_t *start = (uint32_t*)p;
+    uint32_t *end_limit = (start + max_num_inst);
+    uint32_t *pc = start;
+
+    if (!p) {
+        fprintf(stderr, "_isatty proc not found in " UCRTBASE "\n");
+        _exit(1);
+    }
+
+    /* end of function */
+    const uint32_t ret_id = 0xd65f0000;
+    const uint32_t ret_mask = 0xfffffc1f;
+    for(; pc < end_limit; pc++) {
+        if (IS_INSN(pc, ret)) {
+            break;
+        }
+    }
+    if (pc == end_limit) {
+        fprintf(stderr, "end of _isatty not found in " UCRTBASE "\n");
+        _exit(1);
+    }
+
+    /* pioinfo instruction mark */
+    const uint32_t adrp_id = 0x90000000;
+    const uint32_t adrp_mask = 0x9f000000;
+    const uint32_t add_id = 0x11000000;
+    const uint32_t add_mask = 0x3fc00000;
+    for(; pc > start; pc--) {
+        if (IS_INSN(pc, adrp) && IS_INSN(pc + 1, add)) {
+            break;
+        }
+    }
+    if(pc == start) {
+        fprintf(stderr, "pioinfo mark not found in " UCRTBASE "\n");
+        _exit(1);
+    }
+
+    /* We now point to instructions that load address of __pioinfo:
+     * adrp	x8, 0x1801d8000
+     * add	x8, x8, #0xdb0
+     * https://devblogs.microsoft.com/oldnewthing/20220809-00/?p=106955
+     * The last adrp/add sequence before ret is what we are looking for.
+     */
+    const uint32_t adrp_insn = *pc;
+    const uint32_t adrp_immhi = (adrp_insn & 0x00ffffe0) >> 5;
+    const uint32_t adrp_immlo = (adrp_insn & 0x60000000) >> (5 + 19 + 5);
+    /* imm = immhi:immlo:Zeros(12), 64 */
+    const uint64_t adrp_imm = ((adrp_immhi << 2) | adrp_immlo) << 12;
+    /* base = PC64<63:12>:Zeros(12) */
+    const uint64_t adrp_base = (uint64_t)pc & 0xfffffffffffff000;
+
+    const uint32_t add_insn = *(pc + 1);
+    const uint32_t add_sh = (add_insn & 0x400000) >> (12 + 5 + 5);
+    /* case sh of
+      when '0' imm = ZeroExtend(imm12, datasize);
+      when '1' imm = ZeroExtend(imm12:Zeros(12), datasize); */
+    const uint64_t add_imm = ((add_insn & 0x3ffc00) >> (5 + 5)) << (add_sh ? 12 : 0);
+
+    __pioinfo = (ioinfo**)(adrp_base + adrp_imm + add_imm);
+#else /* _M_ARM64 */
+    char *pend = p;
 
 # ifdef _WIN64
     int32_t rel;
@@ -2667,7 +2731,8 @@ set_pioinfo_extra(void)
 #else
     __pioinfo = *(ioinfo***)(p);
 #endif
-#endif
+#endif /* _M_ARM64 */
+#endif /* RUBY_MSVCRT_VERSION */
     int fd;
 
     fd = _open("NUL", O_RDONLY);
--- a/configure.ac
+++ b/configure.ac
@@ -2667,6 +2667,9 @@ AS_CASE([$coroutine_type], [yes|''], [
         [*86-mingw*], [
             coroutine_type=win32
         ],
+        [aarch64-mingw*], [
+            coroutine_type=arm64
+        ],
         [arm*-linux*], [
             coroutine_type=arm32
         ],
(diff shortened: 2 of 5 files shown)


pbo-linaro:
For now, 2 tests are failing. I can investigate and fix this, once this PR will be merged.

```
#209 test_fiber.rb:37: 
     Fiber.new(&Object.method(:class_eval)).resume("foo")
  #=> killed by SIGSEGV (signal 11)  [ruby-dev:34128]
#1524 test_thread.rb:336: 
     g = enum_for(:binding)
     loop { g.next }
  #=> killed by SIGSEGV (signal 11)  [ruby-dev:34128]
FAIL 2/0 tests failed
```

pbo-linaro:
@dennisameling @lazka You could be interested by some of those patches for [MSYS2 Ruby package](https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-ruby). It would enable clangarm64 version of it (I didn't try yet to build additional gems).

pbo-linaro:
@nurse You're welcome to review it, as this `__pioinfo` stuff should be familiar to you ;).

Biswa96:
> Most of the work is detection of __pioinfo pointer, which was the blocking step.

Out of curiosity, would it be possible to remove/replace that __pioinfo mess entirely? The assembly instructions in ucrtbase may change anytime in any Windows version.

lazka:
@larskanis did something in that area last year: https://bugs.ruby-lang.org/issues/18605#note-6 I don't know what the status there is.

pbo-linaro:
I completely agree that the best solution would be to get rid of `pioinfo` pointer. Python, Perl, and MinGW (at least) got rid of it from what I investigated. As @lazka mentioned, @larskanis started an implementation [here](https://github.com/larskanis/ruby/tree/pioinfo). However, it seems it has been untouched for a while.

For now, it's a tentative to have a pragmatic solution, and hopefully, @ArminG-MSFT can make sure ucrtbase _isatty function will keep the same pattern for windows-arm64 too.

nurse:
As lazka quoted and larskanis said, to remove `__pioninfo` it needs to reimplement C standard library I/O functions to use our HANDLE/SOCKET mapping table. Why the table is needed is to emulate Unix-like behavior which unifies file handle and socket. In other words to provide `IO.select`. The strategy larskanis wrote in https://bugs.ruby-lang.org/issues/18605#note-6  is aligned our (I and usa) understanding. The most tough part is implement CRLF handling (text mode).

> I completely agree that the best solution would be to get rid of pioinfo pointer. Python, Perl, and MinGW (at least) got rid of it from what I investigated.

As far as I understand Python is under slightly different situation because they don't provide HANDLE/SOCKET unification. (their `select` doesn't handle files on Windows) https://docs.python.org/3/library/select.html

Perl seems similar situation, but they introduced another hack this year (!). https://github.com/Perl/perl5/commit/8552f09f5cfe61a536a65f11290ef026f7aa0356
It looks interesting and it's worth investigating how stable the strategy is.

lazka:
fyi, the USE_LLVM_WINDRES workaround in this PR should no longer be needed starting with llvm v18, and the relevant llvm changes are backported in MSYS2 as of today -> https://github.com/msys2/MINGW-packages/pull/19524

win32/win32.c
```diff
@@ -2615,9 +2615,73 @@ set_pioinfo_extra(void)
...
+     * adrp	x8, 0x1801d8000
+     * add	x8, x8, #0xdb0
+     * https://devblogs.microsoft.com/oldnewthing/20220809-00/?p=106955
+     * The last adrp/add sequence before ret is what we are looking for.
+     */
+    const uint32_t adrp_insn = *pc;
+    const uint32_t adrp_immhi = (adrp_insn & 0x00ffffe0) >> 5;
+    const uint32_t adrp_immlo = (adrp_insn & 0x60000000) >> (5 + 19 + 5);
+    /* imm = immhi:immlo:Zeros(12), 64 */
+    const uint64_t adrp_imm = ((adrp_immhi << 2) | adrp_immlo) << 12;
```
jeremyd2019, line 2670: I'm messing with some similar code, and noticed what I think is a bug here: adrp's immediate should be sign-extended.  I think something like 
```suggestion
    /* imm = SignExtend(immhi:immlo:Zeros(12), 64) */
    int64_t adrp_imm = (adrp_immhi << 2) | adrp_immlo;
    /* sign extend */
    if (adrp_imm & (1 << 20))
        adrp_imm |= ~((1 << 21) - 1);
    adrp_imm <<= 12;
```

if you really want that to be `const` you could do that in a ternary but I don't know if it's more legible that way:
```suggestion
    /* imm = SignExtend(immhi:immlo:Zeros(12), 64) */
    const int64_t adrp_imm = (((adrp_immhi & (1 << 18)) ? (1 << 21) - 1 : 0) | (adrp_immhi << 2) | adrp_immlo) << 12;
```

pbo-linaro: You're right, I missed that, and luckily, we didn't it the case where adrp immhi sign matters.
I won't push a fix, but you're welcome to contribute it.

jeremyd2019: Maybe... The code I'm messing with didn't hit a negative offset either, but I was thinking it wouldn't be very useful if it wasn't signed, so checked the docs instead of just relying on what this code was doing... I didn't check the specifics of the add instruction, the code I was looking at didn't use that, instead an `ldr` with an immediate unsigned offset to add in the low-order bits.

pbo-linaro: By curiosity, I checked the QEMU source as well, and sign is carefully extended for the immediate when decoding instruction.

win32/win32.c
```diff
@@ -2615,9 +2615,73 @@ set_pioinfo_extra(void)
...
+    if (pc == end_limit) {
+        fprintf(stderr, "end of _isatty not found in " UCRTBASE "\n");
+        _exit(1);
+    }
+
+    /* pioinfo instruction mark */
+    const uint32_t adrp_id = 0x90000000;
+    const uint32_t adrp_mask = 0x9f000000;
+    const uint32_t add_id = 0x11000000;
+    const uint32_t add_mask = 0x3fc00000;
```
jeremyd2019, line 2649: I'm having trouble understanding the add id/mask.  I think this is trying to match https://developer.arm.com/documentation/ddi0602/2024-09/Index-by-Encoding/Data-Processing----Immediate?lang=en#addsub_imm.  
```
0  0  0  1  0  0  0  1  0  0
0  0  1  1  1  1  1  1  1  1
sf op S  1  0  0  0  1  0  sh
```
So you mask off `sf` (size, 32 or 64 bit), and `op` (apparently add vs subtract), and match `S` must be off (not setting flags) and `sh` must be off (not shifting by 12 bits).  You then later implement checking and handling the `sh` bit left shifting by 12, but never check or implement the `op` bit turning it into a subtract.  Perhaps you meant mask `0x7f800000` instead?  Or am I off by one somewhere?

pbo-linaro: I used doc pages for instructions encoding: [add](https://developer.arm.com/documentation/ddi0602/2022-06/Base-Instructions/ADD--immediate---Add--immediate--) - vs [sub](https://developer.arm.com/documentation/ddi0602/2022-06/Base-Instructions/SUB--immediate---Subtract--immediate--).

`_mask` are the bits that matters (same between sub and add).
`_id` are the bits expected, i.e. the concrete encoding. (bit 30 allow to differentiate sub vs add).

To identify an add, we need to check bits 23 to 30, which represent the instruction (fixed) encoding.
add_id: 0x11000000 (bits 23 to 30 are 0b00100010, sub_id would be 0b10100010)
add_mask: 0x3fc00000 (0b0111111110000000000000000000000, bits 23 to 30), sub_mask would be the same.

Note: The code included here is not supposed to be a complete disassembler, just a simple way to find a couple of sequences expected (it's supposed that only immediates may be changed in the future). So it makes assumptions (S bit at 0) and is "enough" to work, no more.

jeremyd2019: > `_mask` are the bits that matters to identify encoding (same between sub and add). `_id` are the bits expected, i.e. the concrete encoding. (bit 30 allow to differentiate sub vs add).
> 
> To identify an add, we need to check bits 23 to 30, which represent the instruction (fixed) encoding. `add_id`: 0x11000000 (bits 23 to 30 are 0b00100010, `sub_id` would be 0b10100010) `add_mask`: 0x3fc00000 (0b0111111110000000000000000000000, bits 23 to 30), sub_mask would be the same.

`sub     x1, x2, #0x159` is:
`0xd1056441` is `0b11010001000001010110010001000001`
your mask is:
`0x3fc00000` is `0b00111111110000000000000000000000`
`and`ing them results in:
`11000000` is the same as `add_id`.

pbo-linaro: oops, you're right, it seems that the mask is shifted to the right, and missing the `op` bit which is wrong.
And counting again yesterday, I did the same mistake...

The correct mask I get is `0x7f800000` (not including the sh bit). (0xff << 23, 23 being bits 23 to 30).

pbo-linaro: Feel free to make a PR for it (and ping me, I can give you a +1).

PR #9011, opened 2023-11-22
More on links

BurdetteLamar:


doc/contributing/documentation_guide.md
```diff
@@ -199,6 +199,36 @@ will be autolinked:
...
+#### +rdoc-ref+ Scheme
+
+Use the +rdoc-ref+ scheme for:
+
+- A link in core documentation to other core documentation.
+- A link in core documentation to documentation in a standard library package.
+- A link in a standard library package to other documentation in that same
+  standard library package.
+
+See section "+rdoc-ref+ Scheme" in {Links}[https://docs.ruby-lang.org/en/master/RDoc/MarkupReference.html#class-RDoc::MarkupReference-label-Links].
```
jeremyevans, line 215: Isn't this a link in the core documentation to documentation in a standard library package, and therefore, shouldn't it use `rdoc-ref` according to this?

jeremyevans (approved):
Looks good.  Please see inline question, and update the documentation if you agree.

doc/contributing/documentation_guide.md
BurdetteLamar: I have not been able to get that to work.  Please make correction and merge.

jeremyevans: I'm not sure how to do so either.  Considering that, maybe we shouldn't be recommending use of rdoc-ref in a case where we don't know how to get it to work?

BurdetteLamar: @drbrain, can you help us out here?

PR #9015, opened 2023-11-23
YJIT: Apply patches ignoring page_end_reserve

k0kubun:
## Problem
YJIT randomly crashes with "Failed to apply patch at CodePtr(33546231)" on Shopify's CI using `--yjit-code-gc --yjit-exec-mem-size=32`.

### Background

`jump_to_next_insn` can put a patch point at the end of the block when the next block is a fallthrough. So patching it would overwrite the next block instead of the block being invalidated. This can cause the following two problems:

1. When it's at the end of the last code page, there may be no enough space to fit a `jmp` instruction.
2. Fail the "patches should not overlap" assertion when there's another patch point in the first 5 bytes of the next block.

A core dump of our CI proved that (1) was the cause of the crash. We haven't observed (2) so far.

## Solution
Disable `page_end_reserve` when invalidating patch points.

We don't need to jump to the next page for writing a jmp instruction, so it's safe to disable `page_end_reserve`. This addresses (1), but not (2). Guaranteeing the safety for (2) requires to fill 5 nop instructions on every patch point, so I want to see if we actually have that problem before attempting to fix a hypothetical problem. In a separate PR, we could add an assertion to verify the assumption.

### DIFF
--- a/yjit/src/invariants.rs
+++ b/yjit/src/invariants.rs
@@ -526,14 +526,15 @@ pub extern "C" fn rb_yjit_tracing_invalidate_all() {
         for patch in &patches {
             assert!(last_patch_end <= patch.inline_patch_pos.raw_ptr(cb), "patches should not overlap");
 
-            let mut asm = crate::backend::ir::Assembler::new();
-            asm.jmp(patch.outlined_target_pos.as_side_exit());
-
             cb.set_write_ptr(patch.inline_patch_pos);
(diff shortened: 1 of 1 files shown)


k0kubun:
This patch passed Core CI's main tests with `--yjit-exec-mem-size=32 --yjit-code-gc`, which was previously reproducing the crash.

XrXr (approved):

PR #9020, opened 2023-11-23
[DOC] More on links 

BurdetteLamar:
Replaces https://github.com/ruby/ruby/pull/9011, which I fouled up.

### DIFF
--- a/doc/contributing/documentation_guide.md
+++ b/doc/contributing/documentation_guide.md
@@ -205,7 +205,7 @@ When writing an explicit link, follow these guidelines.
 
 #### +rdoc-ref+ Scheme
 
-Use the +rdoc-ref+ scheme for:
+Prefer the +rdoc-ref+ scheme for:
 
 - A link in core documentation to other core documentation.
 - A link in core documentation to documentation in a standard library package.
--- a/doc/contributing/documentation_guide.md
+++ b/doc/contributing/documentation_guide.md
@@ -205,14 +205,14 @@ When writing an explicit link, follow these guidelines.
 
 #### +rdoc-ref+ Scheme
 
-Prefer the +rdoc-ref+ scheme for:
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
I have changed "Use the rdoc-ref" to "Prefer the rdoc-ref."

doc/contributing/documentation_guide.md
```diff
@@ -199,6 +199,36 @@ will be autolinked:
...
+### Explicit Links
+
+When writing an explicit link, follow these guidelines.
+
+#### +rdoc-ref+ Scheme
+
+Prefer the +rdoc-ref+ scheme for:
+
+- A link in core documentation to other core documentation.
+- A link in core documentation to documentation in a standard library package.
```
jeremyevans, line 211: My feelings have not changed.  We should not recommend an approach unless we can figure out how to make it work ourselves.

One issue with using rdoc-ref in this case is if you build just the core documentation without the standard library documentation, the rdoc-ref link will not work.  I don't think that affects docs.ruby-lang.org.  It looks like for recent releases, ruby-doc.org also builds them both together.  So I don't think we should avoid this recommendation, but we must figure out how to get it to work first.  I'll see if I can determine how to get it to work.

BurdetteLamar: I'll work on it, too.

BurdetteLamar: Got it.  I'll push soon.

BurdetteLamar: Fixed.

jeremyevans (approved):
Looks good.  Please remember to squash when merging.

BurdetteLamar:
> Looks good. Please remember to squash when merging.

Always hate to merge.  @jeremyevans, will you do it?

PR #9023, opened 2023-11-23
Abort GC on shutdown

peterzhu2118:
On large Ruby applications, shutdown may be slow if a major GC has just started because rb_objspace_call_finalizer completes the GC.

This commit adds gc_abort which discards the mark stack if during incremental marking and stops sweeping if during lazy sweeping.


byroot (approved):
❤️

PR #9025, opened 2023-11-24
Add support for passing a block to `IO#readlines` and `IO.readlines`.

ioquatix:
`IO#readlines` can allocate memory proportional to the file size. This can be a problem. While `IO#each_line` exists, `IO#readlines` is convenient. So, let's extend `IO#readlines` and `IO.readlines` to accept a block, and yield one line at a time, to avoid memory usage issues when dealing with large files.

cc @youchan

### DIFF
--- a/io.c
+++ b/io.c
@@ -4474,15 +4474,24 @@ rb_io_readlines(int argc, VALUE *argv, VALUE io)
 static VALUE
 io_readlines(const struct getline_arg *arg, VALUE io)
 {
-    VALUE line, ary;
+    VALUE line;
 
     if (arg->limit == 0)
         rb_raise(rb_eArgError, "invalid limit: 0 for readlines");
-    ary = rb_ary_new();
-    while (!NIL_P(line = rb_io_getline_1(arg->rs, arg->limit, arg->chomp, io))) {
-        rb_ary_push(ary, line);
+
+    // If a block is provided, yield lines.
+    if (rb_block_given_p()) {
+        while (!NIL_P(line = rb_io_getline_1(arg->rs, arg->limit, arg->chomp, io))) {
+            rb_yield(line);
+        }
+        return Qnil;
+    } else {
+        VALUE ary = rb_ary_new();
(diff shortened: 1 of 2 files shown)


ioquatix:
Here is a small example program I was using for testing:

```ruby
#!/usr/bin/env ruby

require 'async'
require 'async/queue'
require 'thread/local'

PATH = ARGV.pop || '/tmp/fifo'

class Producer
  def initialize
    @queue = Thread::Queue.new
  end

  def dequeue
    @queue.pop
  end

  def run
    Thread.new do
      # read data from named pipe
      File.readlines(PATH) do |line|
        @queue.push line
      end
    end
  end
end

# consumer
producer = Producer.new
producer.run
loop do
  p producer.dequeue
end
```

Most of the credit goes to @youchan.

ioquatix:
@mame mentioned that this function can already be done by `IO.foreach`. I agree there is significant overlap.

nobu:
Can be closed?

ioquatix:
I am fine to close this, however I think what this shows is that there is a lack of consistency in the IO interface. I've actually made exactly the same mistake before, expecting `IO.readlines` takes a block.

The other problem is, whether we should encourage this approach at all. `IO.readlines` could work for trivial cases, but fail with big files. Users may not be aware of the limitations. For Ruby 3.3, I'm fine with not making any changes. However, I wonder if we should consider this more carefully. After all, the public interface on core classes will guide our users/programmers and I think we should guide them towards incremental/lazy `each_line` style patterns.

ioquatix:
If you check https://github.com/search?q=%2FIO%5C.readlines%28.*%3F%29do%2F&type=code you can see plenty of examples of poor quality code, e.g. `IO.readlines.each do` which with this PR becomes `IO.readlines do`. Without this PR, it should be `IO.foreach(...) do`. So, please ask yourself the question, why do people ignore `IO.foreach`? I think it's because `IO.readlines` name matches closely with how people are thinking about the problem. In other words, `IO.foreach` is not obvious.

PR #9019, opened 2023-11-23
GVL Instrumentation events are fired out of order on Ruby 3.3

jpcamara:
Right now this PR just provides an instrumentation spec to show what the behavior should be, and was in Ruby 3.2.

* The changes that came along with M:N threads have caused some issues with the GVL instrumentation api. The events are now fired out of order.

* This spec on ruby 3.2 would consistently record as %[started ready resumed suspended exited] per thread.

* On Ruby 3.3, it comes through as %w[ready resumed started exited suspended]

* With `RUBY_MN_THREADS=1` you get %w[ready started exited suspended]

```
make test-all TESTS="test/-ext-/thread/test_instrumentation_api.rb --name=TestThreadInstrumentation#test_order_of_events"
# => %w[ready resumed started exited suspended]

RUBY_MN_THREADS=1 make test-all TESTS="test/-ext-/thread/test_instrumentation_api.rb --name=TestThreadInstrumentation#test_order_of_events"
# => %w[ready started exited suspended]
```

@casperisfine

PR #9029, opened 2023-11-24
Refactor and fix the GVL instrumentation API

casperisfine:
Ref: https://github.com/jhawthorn/vernier/issues/33
Fix: https://github.com/ruby/ruby/pull/9019

This entirely changes how it is tested. Rather than to use counters we now record the timeline of events with associated threads which makes it much easier to assert that certains events are only preceded by a specific event, and makes it much easier to debug unexpected timelines.

co-authored by: @etiennebarrie @jpcamara  
FYI: @jhawthorn @tenderlove @ko1 

NB: This passes on macOS, but I don't have a Linux or Windows machine, so I kinda expect CI to not immediately pass, especially with M:N threads on.

### DIFF
--- a/ext/-test-/thread/instrumentation/instrumentation.c
+++ b/ext/-test-/thread/instrumentation/instrumentation.c
@@ -2,85 +2,136 @@
 #include "ruby/atomic.h"
 #include "ruby/thread.h"
 
-static rb_atomic_t started_count = 0;
-static rb_atomic_t ready_count = 0;
-static rb_atomic_t resumed_count = 0;
-static rb_atomic_t suspended_count = 0;
-static rb_atomic_t exited_count = 0;
-
 #ifndef RB_THREAD_LOCAL_SPECIFIER
 #  define RB_THREAD_LOCAL_SPECIFIER
 #endif
 
-static RB_THREAD_LOCAL_SPECIFIER unsigned int local_ready_count = 0;
-static RB_THREAD_LOCAL_SPECIFIER unsigned int local_resumed_count = 0;
-static RB_THREAD_LOCAL_SPECIFIER unsigned int local_suspended_count = 0;
-
 static VALUE last_thread = Qnil;
+static VALUE timeline_value = Qnil;
+
+struct thread_event {
+    VALUE thread;
+    rb_event_flag_t event;
+};
+
+#define MAX_EVENTS 1024
+static struct thread_event event_timeline[MAX_EVENTS];
+static rb_atomic_t timeline_cursor;
+
+static void
+event_timeline_gc_mark(void *ptr) {
+    rb_atomic_t cursor;
+    for (cursor = 0; cursor < timeline_cursor; cursor++) {
+        rb_gc_mark(event_timeline[cursor].thread);
+    }
+}
+
+static const rb_data_type_t event_timeline_type = {
+    "TestThreadInstrumentation/event_timeline",
+    {event_timeline_gc_mark, NULL, NULL,},
+    0, 0,
+    RUBY_TYPED_FREE_IMMEDIATELY,
+};
+
+static void
+reset_timeline(void)
+{
+    timeline_cursor = 0;
+    memset(event_timeline, 0, sizeof(struct thread_event) * MAX_EVENTS);
+}
+
+static rb_event_flag_t
+find_last_event(VALUE thread)
+{
+    rb_atomic_t cursor = timeline_cursor;
+    if (cursor) {
+        do {
+            if (event_timeline[cursor].thread == thread){
+                return event_timeline[cursor].event;
+            }
+            cursor--;
+        } while (cursor > 0);
+    }
+    return 0;
+}
+
+static const char *
+event_name(rb_event_flag_t event)
+{
+    switch (event) {
+      case RUBY_INTERNAL_THREAD_EVENT_STARTED:
+        return "started";
+      case RUBY_INTERNAL_THREAD_EVENT_READY:
+        return "ready";
+      case RUBY_INTERNAL_THREAD_EVENT_RESUMED:
+        return "resumed";
+      case RUBY_INTERNAL_THREAD_EVENT_SUSPENDED:
+        return "suspended";
+      case RUBY_INTERNAL_THREAD_EVENT_EXITED:
+        return "exited";
+    }
+    return "no-event";
+}
+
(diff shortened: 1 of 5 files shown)


casperisfine:
Oh, an important note, it predates this PR, but the callback are no longer guaranteed to be triggered from the associated native thread, they can sometimes be triggered from a parent thread. But consumer of the API should only rely on `event_data->thread` anyway.

I need to update the documentation to reflect that.

jhawthorn:
Hmm, I'm seeing more inconsistencies with this branch vs master (with traditional dedicated threads). In particular multiple threads from the main Ractor look like they're "resumed" at the same time. I'll try to make a reproduction.

jpcamara:
> NB: This passes on macOS, but I don't have a Linux or Windows machine, so I kinda expect CI to not immediately pass, especially with M:N threads on.

I'm on a Mac as well, but I have a Dockerfile I use to test Linux features like M:N. Do you think that'd be useful to PR to Ruby @casperisfine ?

ext/-test-/thread/instrumentation/instrumentation.c
```diff
@@ -2,85 +2,119 @@
...
-        last_thread = event_data->thread;
-        RUBY_ATOMIC_INC(started_count);
+        if (last_event != 0) {
+            rb_bug("TestThreadInstrumentation: `started` event can't be preceded by `%s`", event_name(last_event));
+        }
         break;
       case RUBY_INTERNAL_THREAD_EVENT_READY:
-        RUBY_ATOMIC_INC(ready_count);
-        local_ready_count++;
+        if (last_event != 0 && last_event != RUBY_INTERNAL_THREAD_EVENT_STARTED && last_event != RUBY_INTERNAL_THREAD_EVENT_SUSPENDED) {
```
jhawthorn, line 87: I believe `READY` can be preceded directly by `RESUMED` in the case that the 100ms quantum runs out

casperisfine: If that was the case it was a bug. How I see it, it should go through suspended always. Unless you think there is value in such behavior to distinguish a thread releasing the GVL volontarily VS being switched because of the quantum?

jhawthorn: I thought it made sense going directly to READY, since we aren't really suspending or waiting for anything, but I don't feel that strongly about it.

jhawthorn:
> Oh, an important note, it predates this PR, but the callback are no longer guaranteed to be triggered from the associated native thread, they can sometimes be triggered from a parent thread. But consumer of the API should only rely on `event_data->thread` anyway.
> 
> I need to update the documentation to reflect that.

Could we provide that guarantee for at least the RESUMED event (ideally also SUSPENDED, READY, though I can appreciate if that may be complicated if the GVL is held while spawning a Thread)? To make a thread-aware profiler it's important to know on which native thread we are actually running.

jhawthorn:
@byroot I have two tests for you 🙇‍♂️.

https://github.com/jhawthorn/ruby/commit/f03c73a476618d24874ca987a6e7b9fd18c648d1

First is a quick ping-pong test using two Queues. We spawn a thread get it to send a message to one queue and then after receive from another. This crashes on the assertion of where we transition from.

```
$ make test-all TESTS='test/-ext-/thread/test_instrumentation_api.rb -n /queue/'
----8<----
# Running tests:

[1/1] TestThreadInstrumentation#test_queue_releases_gvl/home/jhawthorn/src/ruby/test/-ext-/thread/test_instrumentation_api.rb:92: [BUG] TestThreadInstrumentation: `ready` must be preceded by `start
ed` or `suspended`, got: `resumed`
ruby 3.3.0dev (2023-11-24T22:21:12Z refactor-instrumen.. f03c73a476) [x86_64-linux]

c:0029 p:---- s:0170 e:000169 CFUNC  :push
c:0028 p:0030 s:0165 E:001818 METHOD /home/jhawthorn/src/ruby/test/-ext-/thread/test_instrumentation_api.rb:92
c:0027 p:0037 s:0156 e:000155 METHOD /home/jhawthorn/src/ruby/tool/lib/test/unit/testcase.rb:199
c:0026 p:0052 s:0150 e:000149 METHOD /home/jhawthorn/src/ruby/tool/lib/test/unit/testcase.rb:167
```

The second test is simpler (and likely showing a different symptom of the same problem?). We spawn a thread to attempt acquiring a Mutex, which should be blocked forever (we disable the instrumentation while still holding it from the main thread). The thread _should_ end up in the `suspended` state, but is left on resumed. (apologies for the hacky prints, sleeps, and Thread.pass)

```
$ make test-all TESTS='test/-ext-/thread/test_instrumentation_api.rb -n /blocked/'
----8<----
# Running tests:

[1/1] TestThreadInstrumentation#test_thread_blocked_forever#<Thread:0x00007f16eddc6378 /home/jhawthorn/src/ruby/test/-ext-/thread/test_instrumentation_api.rb:105 sleep_forever>
 = 0.10 s
  1) Failure:
TestThreadInstrumentation#test_thread_blocked_forever [/home/jhawthorn/src/ruby/test/-ext-/thread/test_instrumentation_api.rb:119]:
<[:started, :ready, :resumed, :suspended]> expected but was
<[:started, :ready, :resumed]>.
```

byroot:
Thanks @jhawthorn I'll have a look at those Monday!

casperisfine:
Ok, so it was tricky, but I managed to make both your tests pass, and even better I no longer need to reference `RB_INTERNAL_THREAD_HOOK` from `thread.c` which was a smell.

I still haven't tested M:N, but I suspect it's missing some events. But given it's experimental I think it can wait.

jhawthorn:
I still don't think this is working. I'm seeing multiple threads incorrectly reported as running simultaneously (without M:N)

byroot:
And yeah, I merged knowing it probably isn't perfect yet, but likely much better than before.

jhawthorn:
I was seeing significantly more correct timelines from the previous code

byroot:
> I was seeing significantly more correct timelines from the previous code

Oh -_-.

rubyonrails3:
Hi, I noticed some changes how resources are shared in ruby 3.3 vs older versions of ruby and I am wondering if that is because of this change.  I created a reddit post as well 

https://www.reddit.com/r/ruby/comments/1cx6jvy/does_ruby_33_have_an_implicit_mutex/

PR #9030, opened 2023-11-24
Fix compaction for generic ivars

peterzhu2118:
When generic instance variable has a shape, it is marked movable. If it it transitions to too complex, it needs to update references otherwise it may have incorrect references.

cc. @byroot @XrXr

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1066,17 +1066,34 @@ gen_ivtbl_resize(struct gen_ivtbl *old, uint32_t n)
 }
 
 void
-rb_mark_and_update_generic_ivar(VALUE obj)
+rb_mark_generic_ivar(VALUE obj)
 {
     struct gen_ivtbl *ivtbl;
 
     if (rb_gen_ivtbl_get(obj, 0, &ivtbl)) {
         if (rb_shape_obj_too_complex(obj)) {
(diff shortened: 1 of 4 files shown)


byroot (approved):
Wow, nice find.

variable.c
```diff
@@ -1066,13 +1066,30 @@ gen_ivtbl_resize(struct gen_ivtbl *old, uint32_t n)
...
+rb_ref_update_generic_ivar(VALUE obj)
+{
+    struct gen_ivtbl *ivtbl;
+
+    if (rb_gen_ivtbl_get(obj, 0, &ivtbl)) {
+        if (rb_shape_obj_too_complex(obj)) {
+            rb_gc_ref_update_table_values_only(ivtbl->as.complex.table);
         }
         else {
             for (uint32_t i = 0; i < ivtbl->as.shape.numiv; i++) {
```
XrXr, line 1095: Nit, feel free to ignore: could change the rb_gc_mark_and_move() call below since this is purely an update callback now.

XrXr (approved):
❤️

variable.c
peterzhu2118: Ah good point! I totally missed this.

PR #9031, opened 2023-11-24
[PRISM] Compile RescueNode

eightbitraptor:


kddnewton (approved):
Looks great!

prism_compile.c
```diff
@@ -1852,6 +1857,63 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                                                                  ISEQ_BODY(iseq)->location.label),
+                                                    ISEQ_TYPE_RESCUE, 1);
+            lstart->rescued = LABEL_RESCUE_BEG;
+            lend->rescued = LABEL_RESCUE_END;
+            ADD_LABEL(ret, lstart);
+
+            bool prev_in_rescue = ISEQ_COMPILE_DATA(iseq)->in_rescue;
+            ISEQ_COMPILE_DATA(iseq)->in_rescue = true;
+            {
+                pm_node_list_t node_list = begin_node->statements->body;
```
jemmaissroff, line 1876: Is it possible that the begin_node has no statements?

prism_compile.c
```diff
@@ -1852,6 +1857,63 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                if (node_list.size > 0) {
```
jemmaissroff, line 1877: I think this if / else can be simplified to `PM_COMPILE_NOT_POPPED(begin_node->statements)`

prism_compile.c
```diff
@@ -1852,6 +1857,63 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+            if (begin_node->else_clause) {
+                PM_POP_UNLESS_POPPED;
+                PM_COMPILE((pm_node_t *)begin_node->else_clause);
+            }
+
+            ADD_LABEL(ret, lend);
+            ADD_INSN(ret, &dummy_line_node, nop);
+            ADD_LABEL(ret, lcont);
+
+            if (begin_node->else_clause) {
```
jemmaissroff, line 1898: If we're calling `PM_POP_IF_POPPED` in either case, do we need this if / else block?

prism_compile.c
```diff
@@ -3572,6 +3633,50 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+
+        pm_node_list_t exception_list = rescue_node->exceptions;
+        if (exception_list.size > 0) {
+            for (size_t i = 0; i < exception_list.size; i++) {
+                ADD_GETLOCAL(ret, &dummy_line_node, LVAR_ERRINFO, 0);
+                PM_COMPILE(exception_list.nodes[i]);
+                ADD_INSN1(ret, &dummy_line_node, checkmatch, INT2FIX(VM_CHECKMATCH_TYPE_RESCUE));
+                ADD_INSN1(ret, &dummy_line_node, branchif, excep_match);
+            }
+            ADD_INSN1(ret, &dummy_line_node, jump, rescue_end);
```
jemmaissroff, line 3651: Looks like this is happening in either case, can we pull it out of the if / else?

test/ruby/test_compile_prism.rb
```diff
@@ -769,6 +769,69 @@ def test_RedoNode
...
+      assert_prism_eval(<<-CODE)
+        a = 1
+        begin
+          b = 2
+          raise "bang"
+        rescue
+          c = 3
+        end
+        a + b + c
+      CODE
```
jemmaissroff, line 832: Can we add a test case with an empty begin?

jemmaissroff (changes requested):

prism_compile.c
```diff
@@ -1852,6 +1857,55 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                                                    rb_str_concat(rb_str_new2("rescue in"),
+                                                                  ISEQ_BODY(iseq)->location.label),
+                                                    ISEQ_TYPE_RESCUE, 1);
+            lstart->rescued = LABEL_RESCUE_BEG;
+            lend->rescued = LABEL_RESCUE_END;
+            ADD_LABEL(ret, lstart);
+
+            bool prev_in_rescue = ISEQ_COMPILE_DATA(iseq)->in_rescue;
+            ISEQ_COMPILE_DATA(iseq)->in_rescue = true;
+            {
```
jemmaissroff, line 1875: Nit, but what are these braces for?

jemmaissroff (approved):
One nit, otherwise lgtm, thanks!

PR #9033, opened 2023-11-24
test/net/http/test_https.rb - fixup session reuse tests

MSP-Greg:
Add `test_session_reuse_with_timeout` test, which verifies that setting `ssl_timeout` will allow session reuse.  Another test checks that not setting `ssl_timeout` works, as it uses OpenSSL's default value.

Rename `test_session_reuse_but_expire` to `test_session_reuse_with_expired_timeout` and fixup for WIndows mingw/ucrt builds.

**EDIT:** duplicate of https://github.com/ruby/net-http/pull/162. Not sure which is preferred.

### DIFF
--- a/test/net/http/test_https.rb
+++ b/test/net/http/test_https.rb
@@ -164,7 +164,7 @@ class TestNetHTTPS < Test::Unit::TestCase
     http.finish
   end
 
-  def test_session_reuse_but_expire
+  def test_session_reuse_with_timeout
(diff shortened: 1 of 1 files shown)


MSP-Greg:
This PR does get CI passing.  MSYS2 updated their OpenSSL package to 3.2.0, which affects Windows ucrt builds.  MSFT/vcpkg will probably update to 3.2.0 sometime soon, which only affects the Windows mswin build.

Since this is an upstream issue, and the test has been skipped, closing.

PR #9034, opened 2023-11-25
Fix cache incoherency for ME resolved through VM_METHOD_TYPE_REFINED

XrXr:
Previously, we didn't invalidate the method entry wrapped by
VM_METHOD_TYPE_REFINED method entries which could cause calls to
land in the wrong method like in the included test.

Fixup: cfd7729ce7a31c8b6ec5dd0e99c67b2932de4732
See-also: e201b81f79828c30500947fe8c8ea3c515e3d112


KJTsanaktsidis:
Ah, this is the "other issue" you were talking about on https://github.com/ruby/ruby/pull/8899 right? Let me see if I understand the issue here:

* We have a method `foo`
* When we define the refinement on Foo, that actualy creates a new method entry of `VM_METHOD_TYPE_REFINED`, _even` though the refinement is unactivated_, and store this in the refined class's method table.
* So, when performing method resolution for `foo` when it is called, it finds the refined method entry, but since the refinement is not activated in this scope, it "unwaps" it and calls `body.refined.original_me` instead. That unwrapped method entry is what gets saved in the callcache at the callsite.
* When we define `foo` again, `clear_method_cache_by_id_in_class` gets called to mark the method entry of `foo` as invalid. The idea is that the next time we call `foo`, the callcache points to an invalidated method entry, so method resolution is retriggered.
* However, in the current code, we don't invalidate the _wrapped_ original method entry when we redefine (this is what you're fixing)
* Which means if we return to the same callsite and call `foo` again, we don't notice that the method entry we're about to call is invalidated, so we don't do method resolution again and so call the wrong thing.

XrXr:
> Let me see if I understand the issue here:

Yep, you got it. `VM_METHOD_TYPE_REFINED` entries signals each call site to augment their method lookup process to consider refinements, so these entries are in the table regardless of the set of refinement activations at each site.

PR #9040, opened 2023-11-27
Fix flaky "Expected 499 to be >= 500" assertion in test_gc_compact.rb

KJTsanaktsidis:
There have been some sproradically flaky tests related to GC compaction, which fail with:

```
  1) Failure:
TestGCCompact#test_moving_hashes_down_size_pools [/test/ruby/test_gc_compact.rb:442]:
    Expected 499 to be >= 500.
```

What's happening here, is that, _sometimes_, depending on very unlucky combinations of machine things, one of the expected-to-be-moved hashes might be found on the machine stack during GC, and thus pinned.

One factor which seems to make this _more_ likely is that GCC 11 on Ubuntu 22.04 seems to want to allocate 440 bytes of stack space for `gc_start`, which is much more than it actually uses on the common code path. The result is that there are some 50-odd VALUE-sized cells "live" on the stack which may well contain valid heap pointers from previous function calls, and will need to be pinned.

This is, of course, totally normal and expected; Ruby's GC is conservative and if there is the possibility that a VALUE might be live on the machine stack, it can't be moved. However, it does make these tests flaky.

This commit "fixes" the tests by performing the work in a fiber; the fiber goes out of scope and should be collected by the call to verify_compaction_references, so there should be no references to the to-be-moved objects floating around on the machine stack.

### DIFF
--- a/test/ruby/test_gc_compact.rb
+++ b/test/ruby/test_gc_compact.rb
@@ -317,14 +317,16 @@ class TestGCCompact < Test::Unit::TestCase
 
       GC.verify_compaction_references(expand_heap: true, toward: :empty)
 
-      arys = ARY_COUNT.times.map do
-        ary = "abbbbbbbbbb".chars
-        ary.uniq!
-      end
+      Fiber.new {
+        $arys = ARY_COUNT.times.map do
+          ary = "abbbbbbbbbb".chars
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):
Thank you! This makes sense to me.

PR #9041, opened 2023-11-27
[Bug #20022] Fix GC.verify_compaction_references not moving every object

KJTsanaktsidis:
The intention of `GC.verify_compaction_references` is, I believe, to force every single movable object to be moved, so that it's possible to debug native extensions which not correctly updating their references to objects they mark as movable.

To do this, it doubles the number of allocated pages for each size pool, and sorts the heap pages so that the free ones are swept first; thus, every object in an old page should be moved into a free slot in one of the new pages.

This worked fine until movement of objects _between_ size pools during compaction was implemented. That causes some problems for `verify_compaction_references`:

* We were doubling the number of pages in each size pool, but actually if some objects need to move into a _different_ pool, there's no guarantee that they'll be enough room in that one.
* It's possible for the sweep & compact cursors to meet in one size pool before all the objects that want to move into that size pool from another are processed by the compaction.

You can see these problems by changing some of the movement tests in test_gc_compact.rb to try and move e.g. 50,000 objects instead of 500; the test is not able to actually move all of the objects in a single compaction run.

To fix this, we do two things in `verify_compaction_references`:

* Firstly, we add enough pages to every size pool to make them the same size. This ensures that their compact cursors will all have space to move during compaction (even if that means empty pages are pointlessly compacted)
* Then, we examine every object and determine where it _wants_ to be compacted into. We use this information to add additional pages to each size pool to handle all objects which should live there.

With these two changes, we can move arbitrary amounts of objects into the correct size pool in a single call to verify_compaction_references.

My _motivation_ for performing this work was to try and fix some test stability issues in test_gc_compact.rb. I now no longer think that we actually see this particular bug in rubyci.org, but I also think `verify_compaction_references` should do what it says on the tin, so it's worth keeping.

[Bug #20022]

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -11059,6 +11059,39 @@ gc_compact(VALUE self)
 #endif
 
 #if GC_CAN_COMPILE_COMPACTION
+
+struct desired_compaction_pages_i_data {
+    rb_objspace_t *objspace;
+    size_t required_slots[SIZE_POOL_COUNT];
+};
+
+static int
+desired_compaction_pages_i(struct heap_page *page, void *data)
+{
+    struct desired_compaction_pages_i_data *tdata = data;
+    rb_objspace_t *objspace = tdata->objspace;
+    VALUE vstart = (VALUE)page->start;
+    VALUE vend = vstart + (VALUE)(page->total_slots * page->size_pool->slot_size);
+
+
+    for (VALUE v = vstart; v != vend; v += page->size_pool->slot_size) {
+        /* skip T_NONEs; they won't be moved */
+        void *poisoned = asan_unpoison_object_temporary(v);
+        if (BUILTIN_TYPE(v) == T_NONE) {
+            if (poisoned) {
+                asan_poison_object(v);
+            }
+            continue;
+        }
+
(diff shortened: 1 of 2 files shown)


gc.c
```diff
@@ -11050,18 +11109,54 @@ gc_verify_compaction_references(rb_execution_context_t *ec, VALUE self, VALUE do
...
+                pages_to_add += slots_to_pages_for_size_pool(objspace, size_pool, desired_compaction.required_slots[i]);
+                /*
+                 * Step 3: Add two more pages so that the compact & sweep cursors will meet _after_ all objects
+                 * have been moved, and not on the last iteration of the `gc_sweep_compact` loop
+                 */
+                pages_to_add += 2;
+
+                heap_add_pages(objspace, size_pool, heap, pages_to_add);
+            }
+        } else if (RTEST(double_heap)) {
```
peterzhu2118, line 11153: Nit: style.

```suggestion
        }
        else if (RTEST(double_heap)) {
```

peterzhu2118 (approved):
One small nitpick, otherwise this makes sense to me. Thank you for working on this!

PR #9043, opened 2023-11-27
[PRISM] Compile Index{And,Or,Operator}WriteNodes

jemmaissroff:
Closes https://github.com/ruby/prism/issues/1780
Closes https://github.com/ruby/prism/issues/1781
Closes https://github.com/ruby/prism/issues/1782


prism_compile.c
```diff
@@ -836,6 +836,255 @@ pm_compile_call_and_or_write_node(bool and_node, pm_node_t *receiver, pm_node_t
...
+                      }
+                  }
+              }
+            }
+        }
+    }
+    return orig_argc;
+}
+
+static void
```
kddnewton, line 999: Some comments might be helpful here. I think I understand what `boff` is (block offset?) but it's kind of hard to understand.

kddnewton (approved):
Optionally add some comments, otherwise looks good

prism_compile.c
jemmaissroff: Yeah, I copied the variable name over but should've updated it! Will add more comments and then merge

PR #9046, opened 2023-11-27
[PRISM] Compile YieldNode with different arguments

jemmaissroff:
Closes https://github.com/ruby/prism/issues/1930


kddnewton (approved):
Looks like some checks failing, not sure if related

jemmaissroff:
> Looks like some checks failing, not sure if related

Interesting. These are all failing on master. One is due to a newly introduced naming conflict with a Prism test case though, so changed the names on the Prism side in a new commit, and will then merge

PR #9047, opened 2023-11-27
YJIT: add top C function call counts to `--yjit-stats`

maximecb:
My little pet project. Add top (uninlined) C function calls to `--yjit-stats`. This can reveal some easy optimization targets without needing to use a profiler.

Lobsters:
```
Top-20 most frequent C calls (31.3% of C calls):
                        Hash#fetch:    770,426 ( 9.4%)
                         Class#new:    366,541 ( 4.5%)
                        Array#each:    232,277 ( 2.8%)
                  Class#superclass:    160,770 ( 2.0%)
                        Array#any?:    117,291 ( 1.4%)
    Concurrent::Array#reverse_each:     97,111 ( 1.2%)
                     Regexp#match?:     78,375 ( 1.0%)
                         Hash#key?:     66,934 ( 0.8%)
                        Class#name:     66,534 ( 0.8%)
          SQLite3::Statement#done?:     61,089 ( 0.7%)
           SQLite3::Statement#step:     61,089 ( 0.7%)
                       Hash#empty?:     58,755 ( 0.7%)
                       Symbol#to_s:     57,037 ( 0.7%)
                     TrueClass#===:     56,353 ( 0.7%)
                         Object#!=:     56,066 ( 0.7%)
                      Array#reject:     55,639 ( 0.7%)
                    FalseClass#===:     54,948 ( 0.7%)
                          Hash#dup:     52,600 ( 0.6%)
                         Array#map:     52,183 ( 0.6%)
                    Array#include?:     45,110 ( 0.6%)

```

Here `Class#superclass` may be an easy optimization target, `TrueClass#===`, `Hash#empty?` and `Class#name` as well. Also surprised that we are still calling `Symbol#to_s`.

Optcarrot:
```
Top-20 most frequent C calls (20.5% of C calls):
             Integer#*:  2,643,814 ( 8.2%)
             Array#[]=:  1,540,322 ( 4.8%)
         Array#rotate!:  1,493,457 ( 4.6%)
            Integer#>>:    673,674 ( 2.1%)
            Array#size:    143,732 ( 0.4%)
             Integer#^:    109,241 ( 0.3%)
             Method#[]:     28,777 ( 0.1%)
          Fiber#resume:      5,443 ( 0.0%)
           Class#yield:      5,443 ( 0.0%)
            Array#each:      2,732 ( 0.0%)
     Integer#object_id:      2,029 ( 0.0%)
           Float#floor:      1,449 ( 0.0%)
       Array#object_id:      1,014 ( 0.0%)
           Array#uniq!:        995 ( 0.0%)
           Symbol#to_s:        612 ( 0.0%)
           Array#clear:        513 ( 0.0%)
    Symbol#start_with?:        397 ( 0.0%)
        Array#include?:        180 ( 0.0%)
               Float#/:        171 ( 0.0%)
               Float#-:        171 ( 0.0%)
```

On optcarrot, we already inline almost 80% of C function calls, which is very impressive and probably explains why we're so fast on that benchmark. The remaining top C functions still show many potential optimization targets, however. There is room to be even faster eventually.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5419,8 +5419,7 @@ fn gen_send_cfunc(
 
     // Log the name of the method we're calling to,
     // note that we intentionally don't do this for inlined cfuncs
-    #[cfg(feature = "stats")]
-    {
+    if get_option!(gen_stats) {
         // TODO: extract code to get method name string into its own function
 
         // Assemble the method name string
@@ -7875,7 +7874,6 @@ fn gen_invokesuper_specialized(
             gen_send_iseq(jit, asm, ocb, iseq, ci, frame_type, None, cme, Some(block), ci_flags, argc, None)
         }
         VM_METHOD_TYPE_CFUNC => {
-            // FIXME: why do we put unknown comptime class here?
             gen_send_cfunc(jit, asm, ocb, ci, cme, Some(block), ptr::null(), ci_flags, argc)
         }
         _ => unreachable!(),
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5421,10 +5421,20 @@ fn gen_send_cfunc(
     // note that we intentionally don't do this for inlined cfuncs
     #[cfg(feature = "stats")]
     {
+        // TODO: extract code to get method name string into its own function
+
         // Assemble the method name string
         let mid = unsafe { vm_ci_mid(ci) };
-        let class_name = unsafe { cstr_to_rust_string(rb_class2name(*recv_known_klass)) }.unwrap();
-        let method_name = unsafe { cstr_to_rust_string(rb_id2name(mid)) }.unwrap();
+        let class_name = if recv_known_klass != ptr::null() {
+            unsafe { cstr_to_rust_string(rb_class2name(*recv_known_klass)) }.unwrap()
+        } else {
+            "Unknown".to_string()
+        };
+        let method_name = if mid != 0 {
+            unsafe { cstr_to_rust_string(rb_id2name(mid)) }.unwrap()
+        } else {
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -7847,6 +7875,7 @@ fn gen_invokesuper_specialized(
             gen_send_iseq(jit, asm, ocb, iseq, ci, frame_type, None, cme, Some(block), ci_flags, argc, None)
         }
         VM_METHOD_TYPE_CFUNC => {
+            // FIXME: why do we put unknown comptime class here?
```
k0kubun, line 7878: I think it's because it's not known. `known_recv_class` expects the receiver's class (otherwise things like `jit_rb_str_to_s` would be broken), but we aren't guarding against it on invokesuper (we only guard against a method entry in the local environment).

For this feature, we'd actually like to document `cme->defined_class` instead of the receiver's class, so we probably need to pass a different variable if we want to provide that information. Even though we don't pass `known_recv_class`, it's not safe to overload this variable for this purpose (because we don't guard it).

yjit/src/codegen.rs
```diff
@@ -5416,6 +5417,33 @@ fn gen_send_cfunc(
         }
     }
 
+    // Log the name of the method we're calling to,
+    // note that we intentionally don't do this for inlined cfuncs
+    #[cfg(feature = "stats")]
```
k0kubun, line 5422: Is it intentional that we enable this on stats (or dev) build instead of `--yjit-stats`? Did you mean `if get_option!(gen_stats) {`?

k0kubun (approved):
seems very useful :+1:

casperisfine:
> Class#new:    366,541 ( 4.5%)

😸

PR #9050, opened 2023-11-28
Further fix the GVL instrumentation API

casperisfine:
Followup: https://github.com/ruby/ruby/pull/9029

[Bug #20019]

Some events still weren't triggered from the right place.

The test suite was also improved a bit more.

cc @jpcamara @jhawthorn, hopefully third time the charm?

I also manually checked that the full timeline makes sense in the CPU preemption case, but I'm not sure if I should actually turn it into assertions as I'm worried it may be flaky. Thoughts?

```
TestThreadInstrumentation#test_muti_thread_timeline
thread-0: started              
thread-0: ready
thread-1: started
thread-1: ready
thread-2: started
thread-2: ready
main    : suspended
main    : ready
thread-0: resumed
thread-0: suspended
thread-0: ready
thread-1: resumed
thread-1: suspended
thread-1: ready
thread-2: resumed
thread-2: suspended
thread-2: ready
main    : resumed
main    : suspended
main    : ready
thread-0: resumed
thread-0: suspended
thread-0: ready
thread-1: resumed
thread-1: suspended
thread-1: ready
thread-2: resumed
thread-2: suspended
thread-2: ready
main    : resumed
main    : suspended
main    : ready
thread-0: resumed
thread-0: suspended
thread-0: ready
thread-1: resumed
thread-1: suspended
thread-1: ready
thread-2: resumed
thread-2: suspended
thread-2: ready
main    : resumed
main    : suspended
thread-0: resumed
main    : ready
thread-0: suspended
thread-0: exited
thread-1: resumed
thread-1: suspended
thread-1: exited
thread-2: resumed
thread-2: suspended
thread-2: exited
main    : resumed
```

### DIFF
--- a/ext/-test-/thread/instrumentation/instrumentation.c
+++ b/ext/-test-/thread/instrumentation/instrumentation.c
@@ -73,53 +73,51 @@ event_name(rb_event_flag_t event)
     return "no-event";
 }
 
-NORETURN(static void unexpected(const char *format, const char *event_name, VALUE thread));
-
 static void
-unexpected(const char *format, const char *event_name, VALUE thread)
+unexpected(bool strict, const char *format, VALUE thread, rb_event_flag_t last_event)
 {
-#if 0
-    fprintf(stderr, "----------------\n");
-    fprintf(stderr, format, event_name, thread);
-    fprintf(stderr, "\n");
-    rb_backtrace();
-    fprintf(stderr, "----------------\n");
-#else
-    rb_bug(format, event_name, thread);
-#endif
+     const char *last_event_name = event_name(last_event);
+    if (strict) {
+        rb_bug(format, thread, last_event_name);
+    }
+    else {
+        fprintf(stderr, format, thread, last_event_name);
+        fprintf(stderr, "\n");
+    }
 }
 
 static void
 ex_callback(rb_event_flag_t event, const rb_internal_thread_event_data_t *event_data, void *user_data)
 {
     rb_event_flag_t last_event = find_last_event(event_data->thread);
+    bool strict = (bool)user_data;
 
-    switch (event) {
-      case RUBY_INTERNAL_THREAD_EVENT_STARTED:
-        if (last_event != 0) {
-            unexpected("TestThreadInstrumentation: `started` event can't be preceded by `%s` (thread=%"PRIxVALUE")", event_name(last_event), event_data->thread);
+    if (last_event != 0) {
+        switch (event) {
+          case RUBY_INTERNAL_THREAD_EVENT_STARTED:
+            unexpected(strict, "[thread=%"PRIxVALUE"] `started` event can't be preceded by `%s`", event_data->thread, last_event);
+            break;
+          case RUBY_INTERNAL_THREAD_EVENT_READY:
+            if (last_event != RUBY_INTERNAL_THREAD_EVENT_STARTED && last_event != RUBY_INTERNAL_THREAD_EVENT_SUSPENDED) {
+                unexpected(strict, "[thread=%"PRIxVALUE"] `ready` must be preceded by `started` or `suspended`, got: `%s`", event_data->thread, last_event);
+            }
+            break;
+          case RUBY_INTERNAL_THREAD_EVENT_RESUMED:
+            if (last_event != RUBY_INTERNAL_THREAD_EVENT_READY) {
+                unexpected(strict, "[thread=%"PRIxVALUE"] `resumed` must be preceded by `ready`, got: `%s`", event_data->thread, last_event);
+            }
+            break;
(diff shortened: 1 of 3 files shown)


jpcamara:
> I also manually checked that the full timeline makes sense in the CPU preemption case, but I'm not sure if I should actually turn it into assertions as I'm worried it may be flaky. Thoughts?

That makes sense. As a specification, here were my expectations previously:

1. `STARTED` and `EXITED` should always be in the right order (first and last), which these PRs addressed
2. Only one thread should ever show as being `RESUMED` at any given time, which this PR should address
3. Thread events should be internally consistent. (1) dictates how a thread starts and ends. Other events that fire should happen in the order of `READY`, `RESUMED`, and `SUSPENDED`.
4. In any spec that tries to run long enough to cause preemption, the actual number of events fired between "STARTED" and "EXITED" can't be guaranteed

I haven't had a chance to fully dig into the changes and test things out. But my current understanding is that my expectations for (1), (2) and (4) line up with the current state of things. 

- Is (3) expected as well, or is it possible to get `READY`, `RESUMED` and `SUSPENDED` in different orders? That's not necessarily a problem if they can happen in different orders, I just want to clarify what the expectation is
- Is there any way to have a spec which guarantees (2)? I agree that assertions could be flakey for preemption cases, just thinking through how to prevent regressions

casperisfine:
> * Is (3) expected as well, or is it possible to get `READY`, `RESUMED` and `SUSPENDED` in different orders?

It's expected as well yes. At least in the current design. You could argue that `SUSPENDED -> RESUMED` may make sense in some cases (e.g. a thread directly yielding to another one), But i think It's best to be consistent here, otherwise implementing instrumenters become tricky.

> Is there any way to have a spec which guarantees (2)? 

So with the "full_timeline" it's possible to write such assertion, but I suspect it could be flaky because most of these events are emitted without the GVL held, so it could be racy.

jhawthorn (approved):
Thank you! The timelines I see from this branch look correct (at least at a glance).

Example thread making HTTP request:
Screenshot 2023-11-28 at 09 17 14

casperisfine:
🎉

jpcamara:
🎉  Looks like GVL tracing is working better now as well. No longer reporting as "died" and then "waiting". And "started" is coming first.

Screenshot 2023-11-28 at 4 05 53 PM

As might be expected, all of the "started" events are reporting on the main thread since the `STARTED` event is being fired there but so gvl tracing is reporting that the main thread started 3 times. Vernier should have the same issue based on how it determines thread id.

Screenshot 2023-11-28 at 4 06 21 PM

@casperisfine what is the best way to utilize the `event_data->thread` to get an appropriate native thread id? I was hoping I could use the ruby apis, but it seems like none of the thread id apis are exposed to extensions.

byroot:
> As might be expected, all of the "started" events are reporting on the main thread since the STARTED event is being fired 

Yep, as mentioned in: https://github.com/ruby/ruby/pull/9029#issuecomment-1825755997

> what is the best way to utilize the event_data->thread to get an appropriate native thread id? 

Do you actually need a native thread id? Or just a reliable thread id? How I was seeing it, apps would just record `event_data->thread` (need to be marked though).

jpcamara:
> > As might be expected, all of the "started" events are reporting on the main thread since the STARTED event is being fired
> 
> Yep, as mentioned in: [#9029 (comment)](https://github.com/ruby/ruby/pull/9029#issuecomment-1825755997)
> 
> > what is the best way to utilize the event_data->thread to get an appropriate native thread id?
> 
> Do you actually need a native thread id? Or just a reliable thread id? How I was seeing it, apps would just record `event_data->thread` (need to be marked though).

So just record the `VALUE` directly? I was considering that as well but wasn't sure it made sense - but you'd know best so if that's a sensible stable value I will use that. 

I seem to remember reading a discussion somewhere in the gvl-tracing repo that being able to access native thread is helpful but I am open to this approach instead.

byroot:
> So just record the `VALUE` directly?

Yup. You can even cast it to `long long` or something if you wish. I don't remember if `Thread` are movable, but I suspect they aren't so compaction is probably not a worry.

It's only if you wish to call stuff on it later that you have to concern yourself with marking.

> a discussion somewhere in the gvl-tracing repo that being able to access native thread is helpful 

If you find it let me know, but I don't see why it would. If anything having the Ruby thread is way more useful as you can then get a much nicer representation of it (e.g. grab `Thread#name` and fallback to `Thread#inspect`).

jpcamara:
Will do, I'll see if I can find it.

> If you find it let me know, but I don't see why it would. If anything having the Ruby thread is way more useful as you can then get a much nicer representation of it (e.g. grab Thread#name and fallback to Thread#inspect).

I'll probably have some questions around some of this, but I will open a PR and ask them there. Thanks!

PR #9055, opened 2023-11-28
[PRISM] Don't calculate params size based on locals

jemmaissroff:
Prior to this commit, we were conflating the size of the locals list with the number of parameters. This commit distinguishes the two, and fixes a related bug which would occur if we set a local that was not a parameter


test/ruby/test_compile_prism.rb
```diff
@@ -756,6 +756,7 @@ def test_DefNode
       assert_prism_eval("def self.prism_test_def_node() 1 end; prism_test_def_node()")
       assert_prism_eval("def self.prism_test_def_node(a,b) [a, b] end; prism_test_def_node(1,2)")
       assert_prism_eval("def self.prism_test_def_node(a,x=7,y=1) x end; prism_test_def_node(7,1)")
+      assert_prism_eval("def self.prism_test_def_node(a = 1); x = 2; end; prism_test_def_node")
```
jemmaissroff, line 759: Confirmed that this test was previously failing

tenderlove (approved):

PR #9056, opened 2023-11-28
YJIT: fix bug in top cfunc logging in `--yjit-stats`

maximecb:
Correctly handle case where there are no cfunc calls


yjit.rb
```diff
@@ -375,7 +375,9 @@ def _print_stats(out: $stderr) # :nodoc:
 
     def print_sorted_cfunc_calls(stats, out:, how_many: 20, left_pad: 4) # :nodoc:
       calls = stats[:cfunc_calls]
-      #puts calls
+      if calls.size == 0
```
k0kubun, line 378:
```suggestion
      if calls.empty?
```

k0kubun (approved):

PR #9058, opened 2023-11-29
[PRISM] Fixes `SyntaxError` when parsing methods with keyword arguments with prism

thdaraujo:
Fixes https://github.com/ruby/prism/issues/1894

Fixes the following error on Prism (Ruby 3.3.0-preview3):
```
ruby -e "RubyVM::InstructionSequence.compile_prism('a(\"a\" => 0)')"

#=> <compiled>: <compiled>: argument stack underflow (-4927) (SyntaxError)
```

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -1399,7 +1399,12 @@ iseqw_s_compile_prism_compile(pm_parser_t *parser, VALUE opt, rb_iseq_t *iseq, V
 
     rb_compile_option_t option;
     make_compile_option(&option, opt);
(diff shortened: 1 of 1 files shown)


jemmaissroff:
Hey @thdaraujo! I think the solution is a bit more involved than this. I think this will be quite a specific change that deals with the appropriate nodes in `prism_compile.c`. With this change on master, Prism will still emit an error on the test case I linked

thdaraujo:
that makes sense, thanks @jemmaissroff ! I'll dig a little deeper and see what I can find. :smile:

PR #9065, opened 2023-11-29
[PRISM] Fix EnsureNode, pass depth to get locals

jemmaissroff:
This commit fixes a bug with locals in ensure nodes by setting the local tables correctly. It also changes accessing locals to look at local tables in parent scopes, and account for this correctly on depths of get or setlocals.

Closes https://github.com/ruby/prism/issues/1932


eightbitraptor (approved):
This looks good! Thanks for fixing

PR #9068, opened 2023-11-29
YJIT: edit `yjit.md` and bring it up to date

maximecb:
Also make various minor edits to improve readability.

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -4,22 +4,22 @@
   </a>
 </p>
 
-
 YJIT - Yet Another Ruby JIT
 ===========================
 
 YJIT is a lightweight, minimalistic Ruby JIT built inside CRuby.
 It lazily compiles code using a Basic Block Versioning (BBV) architecture.
-The target use case is that of servers running Ruby on Rails.
 YJIT is currently supported for macOS, Linux and BSD on x86-64 and arm64/aarch64 CPUs.
 This project is open source and falls under the same license as CRuby.
 
 <p align="center"><b>
     If you're using YJIT in production, please
     <a href="mailto:maxime.chevalierboisvert@shopify.com">share your success stories with us!</a>
- </b></p>
+</b></p>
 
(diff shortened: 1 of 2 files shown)


doc/yjit/yjit.md
```diff
@@ -231,7 +245,7 @@ if each process only handles 1000 requests, `--yjit-call-threshold=1000` might b
 
 ### Decreasing --yjit-exec-mem-size
 
-`--yjit-exec-mem-size` specifies the JIT code size, but YJIT also uses memory for its metadata,
+The `--yjit-exec-mem-size` option specifies the JIT code size, but YJIT also uses memory for its metadata,
 which often consumes more memory than JIT code. Generally, YJIT adds memory overhead by roughly
 3-4x of `--yjit-exec-mem-size` in production as of Ruby 3.2. You should multiply that by the number
```
maximecb, line 250: @k0kubun would you say that the 3-4x figure is still accurate?

With the work we've done, it's maybe closer to 2x or 2-3x?

yjit/src/stats.rs
```diff
@@ -750,8 +750,6 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
 
             for (name, idx) in cfunc_name_to_idx {
                 let count = call_counts[*idx];
-                println!("{}: {}", name, count);
-
```
maximecb, line 754: Sneak a fix for this `println!()` I forgot to remove 😅

byroot (approved):

doc/yjit/yjit.md
k0kubun: It's closer than 3.2, yes, but it still seems to be in the range of 3-4x.

On SFR experimental clusters running very recent master, max yjit_alloc_size is 72MiB when max code_region_size is 28MiB. So yjit_alloc_size + code_region_size is 3.6x as large as code_region_size.

k0kubun:
looks good to me :+1:

PR #9073, opened 2023-11-30
Add missing GVL hooks for M:N threads and ractors

jhawthorn:
Followup: #9050

This fixes GVL instrumentation in three locations it was missing related to M:N threads and Ractors:
- Suspending when blocking on a Ractor
- Suspending when doing a coroutine transfer from an M:N thread
- Resuming after an M:N thread starts

cc @jpcamara @byroot @ko1 @tenderlove


byroot (approved):

byroot:
I think the CI errors may be legit though? Something Ractor related, not super clear, but too consistent to just be a flake

byroot:
The two remaining errors were flakes, they passed on a retry.

:shipit:

PR #9079, opened 2023-11-30
[PRISM] Refactor scope depth, and fix POST_EXECUTION_NODE

eightbitraptor:
This PR implements a slightly different approach to fix the local depth issues seen in ensure nodes and rescue nodes amongst others. Instead of passing a `depth` value by reference and then incrementing it every time we look up the scope tree, instead let's specify an offset directly inside the scope node itself - that way we can track offsets explicitly for nodes that need it.

This PR then uses this approach to fix the local lookup bug in POST_EXECUTION_NODES by providing the correct offset when creating the relevant scope node.

This fixes this code:

```
foo = 1
END { puts foo }
```

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1398,6 +1398,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         case PM_POST_EXECUTION_NODE: {
             pm_post_execution_node_t *cast = (pm_post_execution_node_t *) node;
             scope->body = (pm_node_t *) cast->statements;
+            scope->local_depth_offset += 2;
             break;
         }
         case PM_PROGRAM_NODE: {
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -689,7 +689,7 @@ pm_interpolated_node_compile(pm_node_list_t *parts, rb_iseq_t *iseq, NODE dummy_
 // It also takes a pointer to depth, and increments depth appropriately
 // according to the depth of the local
 static int
-pm_lookup_local_index_any_scope(rb_iseq_t *iseq, pm_scope_node_t *scope_node, pm_constant_id_t constant_id, int *depth)
+pm_lookup_local_index_any_scope(rb_iseq_t *iseq, pm_scope_node_t *scope_node, pm_constant_id_t constant_id)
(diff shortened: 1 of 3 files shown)


prism_compile.c
```diff
@@ -1396,6 +1398,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         case PM_POST_EXECUTION_NODE: {
             pm_post_execution_node_t *cast = (pm_post_execution_node_t *) node;
             scope->body = (pm_node_t *) cast->statements;
+            scope->local_depth_offset = 2;
```
kddnewton, line 1401: Should be `+ 2`, since you can have `foo = 0; END { END { foo } }`

prism_compile.c
```diff
@@ -1368,6 +1369,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         case PM_ENSURE_NODE: {
             pm_ensure_node_t *cast = (pm_ensure_node_t *)node;
             scope->body = (pm_node_t *)cast->statements;
+            scope->local_depth_offset = 1;
```
kddnewton, line 1372: Should be `+ 1`, since you can have:

```ruby
foo = 0

begin
ensure
  begin
  ensure
    foo
  end
end
```

prism_compile.h
```diff
@@ -13,6 +13,11 @@ typedef struct pm_scope_node {
 
     ID *constants;
     st_table *index_lookup_table;
+
+    // Some locals are defined at higher scopes than they are used. We can use
+    // this offset to control which parent scopes local table we should be
+    // referencing from the current scope.
+    int local_depth_offset;
```
kddnewton, line 20: Can we make this unsigned? Just to make it slightly clearer that this can never be `< 0`.

```suggestion
    unsigned int local_depth_offset;
```

kddnewton (changes requested):
Nice!

prism/prism.c
```diff
@@ -1,3 +1,4 @@
+
```
kddnewton, line 1:
```suggestion
```

kddnewton (approved):

PR #9082, opened 2023-11-30
[PRISM] Fix behavior of BlockParameters with only one parameter

jemmaissroff:
This commit sets the ambiguous param flag if there is only one parameter on a block node. It also fixes a small bug with a trailing comma on params.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -991,7 +991,7 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
                       PM_COMPILE_NOT_POPPED(splat_node->expression);
                   }
 
-                  ADD_INSN1(ret, &dummy_line_node, splatarray, popped ? Qfalse : Qtrue);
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -3678,6 +3678,12 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
                     PM_COMPILE(required_param);
                 }
             }
+
+            // If there's only one required element in the parameters
+            // CRuby needs to recognize it as an ambiguous parameter
+            if (body->type == ISEQ_TYPE_BLOCK && requireds_list->size == 1) {
```
kddnewton, line 3684: I think there are more things we have to check here. I think it's that `required_list->size` is 1 and that there are no other parameters in the list at all.

kddnewton (changes requested):

kddnewton (approved):

PR #9085, opened 2023-11-30
YJIT: optimized codegen for `rb_ary_length()`

maximecb:
Rubykon benchmark before:
```
num_send_cfunc:           53,687,773 (26.2%)
num_send_cfunc_inline:    36,127,214 (67.3%)
...
Top-20 most frequent C calls (31.2% of C calls):
                 Array#size:  2,688,119 ( 5.0%)
               Array#inject:  1,702,545 ( 3.2%)
                  Array#map:  1,652,550 ( 3.1%)
                 Integer#!=:  1,581,468 ( 2.9%)
```

After:
```
num_send_cfunc:           53,687,773 (26.2%)
num_send_cfunc_inline:    38,815,333 (72.3%)
...
Top-20 most frequent C calls (26.4% of C calls):
               Array#inject:  1,702,545 ( 3.2%)
                  Array#map:  1,652,550 ( 3.1%)
                 Integer#!=:  1,581,468 ( 2.9%)
                 Array#any?:  1,581,097 ( 2.9%)
```

Also brings us from 79.5% inlined cfunc calls to 79.9% on optcarrot 🥕 , and this function is also used in railsbench, liquid, lobsters, etc.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4949,6 +4949,31 @@ fn jit_rb_ary_empty_p(
     return true;
 }
 
+// Codegen for rb_ary_length()
+fn jit_rb_ary_length(
+    _jit: &mut JITState,
+    asm: &mut Assembler,
+    _ocb: &mut OutlinedCb,
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    let array_opnd = asm.stack_pop(1);
+    let array_reg = asm.load(array_opnd);
+    let len_opnd = get_array_len(asm, array_reg);
+
+    // Convert the length to a fixnum
+    let shifted_val = asm.lshift(len_opnd, Opnd::UImm(1));
+    let out_val = asm.or(shifted_val, Opnd::UImm(RUBY_FIXNUM_FLAG as u64));
+
+    let out_opnd = asm.stack_push(Type::Fixnum);
+    asm.store(out_opnd, out_val);
+
+    return true;
+}
+
 fn jit_rb_ary_push(
     jit: &mut JITState,
     asm: &mut Assembler,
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -4949,6 +4949,31 @@ fn jit_rb_ary_empty_p(
...
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    let array_opnd = asm.stack_pop(1);
+    let array_reg = asm.load(array_opnd);
+    let len_opnd = get_array_len(asm, array_reg);
+
+    // Convert the length to a fixnum
+    let shifted_val = asm.lshift(len_opnd, Opnd::UImm(1));
```
k0kubun, line 4968: I think we need to check if it overflows. `rb_ary_length` uses `LONG2NUM`, which checks if it's `RB_FIXABLE` before conversion.

maximecb: Thinking about this further. The array size in bytes is the array length multiplied by 8 bytes per slot, so it's basically impossible to allocate `FIXNUM_MAX` slots on a 64-bit system?

k0kubun (approved):
:+1:

yjit/src/codegen.rs
XrXr: I agree that it's probably impossible enough that we can skip the runtime guard. On 64-bit Windows, where `long` is 32 bit and Fixnum is 31 bits, this is more possible, though. Maybe worth a static assert like:
```rust
const _:() = {
    let _: unsafe extern "C" fn(VALUE) -> std::ffi::c_long = rb_yjit_array_len; // static assert `c_long` can represent all array sizes
    assert!(
        std::ffi::c_long::BITS == 64,
        "no guard for array length overflow because process should run out of memory before it can make an array larger than FIXNUM_MAX"
    );
};
```

Another plausible change is supporting the default value form like `huge_array = Array.new(2**70, :my_favorite_value)` without making a huge allocation. Though there is no way to statically guard for that, there is probably no need to worry about it.

maximecb: There's a constant for `FIXNUM_MAX`. We could directly use that to reject codegen (or not register the codegen function at all) in Rust?

maximecb: Tho @k0kubun just said `ARY_MAX_SIZE` is `LONG_MAX / (int)sizeof(VALUE)` which presumably completely eliminates the problem, even on Windows, no?

PR #9086, opened 2023-11-30
YJIT: Cancel on-stack jit_return on invalidation

k0kubun:
Even after code GC invalidates patch points, `jit_exec_exception` leaves the possibility of entering old code referenced by `cfp->jit_return` while entering a JIT frame compiling a new block. When such old code has no patch points before a jump to another freed block, it crashes.

This PR prevents it by updating `cfp->jit_return` from compiled blocks or branch stubs to `leave_exit`, forcing every on-stack frame to exit to the interpreter on `leave`.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -1291,17 +1291,35 @@ rb_jit_cont_each_iseq(rb_iseq_callback callback, void *data)
             continue;
 
         const rb_control_frame_t *cfp;
-        for (cfp = RUBY_VM_END_CONTROL_FRAME(cont->ec) - 1; ; cfp = RUBY_VM_NEXT_CONTROL_FRAME(cfp)) {
+        for (cfp = RUBY_VM_END_CONTROL_FRAME(cont->ec) - 1; cfp >= cont->ec->cfp; cfp = RUBY_VM_NEXT_CONTROL_FRAME(cfp)) {
             const rb_iseq_t *iseq;
             if (cfp->pc && (iseq = cfp->iseq) != NULL && imemo_type((VALUE)iseq) == imemo_iseq) {
                 callback(iseq, data);
             }
-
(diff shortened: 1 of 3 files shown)


cont.c
```diff
@@ -1291,17 +1291,35 @@ rb_jit_cont_each_iseq(rb_iseq_callback callback, void *data)
             continue;
 
         const rb_control_frame_t *cfp;
-        for (cfp = RUBY_VM_END_CONTROL_FRAME(cont->ec) - 1; ; cfp = RUBY_VM_NEXT_CONTROL_FRAME(cfp)) {
+        for (cfp = RUBY_VM_END_CONTROL_FRAME(cont->ec) - 1; cfp >= cont->ec->cfp; cfp = RUBY_VM_NEXT_CONTROL_FRAME(cfp)) {
```
XrXr, line 1294: Iterating like this is tricky, with the -1 and the comparison and stuff. If the order doesn't matter, we could we go from top from bottom and use the more abstracted way like the following:

```
rb_vm_get_binding_creatable_next_cfp(ec, cfp)
{
    while (!RUBY_VM_CONTROL_FRAME_STACK_OVERFLOW_P(ec, cfp)) {
        if (cfp->iseq) {
            return (rb_control_frame_t *)cfp;
        }
        cfp = RUBY_VM_PREVIOUS_CONTROL_FRAME(cfp);
    }
```

XrXr (approved):

cont.c
k0kubun: :+1: https://github.com/ruby/ruby/pull/9086/commits/f7d3661b0bfe66c41ee4c6c7754613507a2b224f

maximecb (approved):
Thanks for all the hard work you've both put into tracking down this bug 🙏 You're really good at this! :)

PR #9088, opened 2023-12-01
Fix test: Relax test conditions to velify Socket::ResolutionError#error_code

shioimm:
The test for Socket::ResolutionError#error_code fails in the FreeBSD environment with this test condition. Because Socket::ResolutionError#error_code returns Socket::EAI_FAIL instead of Socket::EAI_FAMILY.
ref: https://rubyci.s3.amazonaws.com/freebsd12/ruby-master/log/20231130T103002Z.fail.html.gz

This PR avoids the test failure by relaxing the condition.

Also changed the domain for testing to `example.com`.

### DIFF
--- a/test/socket/test_socket.rb
+++ b/test/socket/test_socket.rb
@@ -771,13 +771,10 @@ class TestSocket < Test::Unit::TestCase
   end
 
   def test_resolurion_error_error_code
-    # https://rubyci.s3.amazonaws.com/freebsd12/ruby-master/log/20231130T103002Z.fail.html.gz
-    omit if /freebsd/ =~ RUBY_PLATFORM
-
     begin
-      Socket.getaddrinfo("www.kame.net", 80, "AF_UNIX")
+      Socket.getaddrinfo("example.com", 80, "AF_UNIX")
     rescue => e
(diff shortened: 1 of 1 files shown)


test/socket/test_socket.rb
```diff
@@ -771,13 +771,10 @@ def test_udp_recvmsg_truncation
...
   def test_resolurion_error_error_code
-    # https://rubyci.s3.amazonaws.com/freebsd12/ruby-master/log/20231130T103002Z.fail.html.gz
-    omit if /freebsd/ =~ RUBY_PLATFORM
-
     begin
-      Socket.getaddrinfo("www.kame.net", 80, "AF_UNIX")
+      Socket.getaddrinfo("example.com", 80, "AF_UNIX")
     rescue => e
-      assert_equal(Socket::EAI_FAMILY, e.error_code)
+      assert_equal([Socket::EAI_FAMILY, Socket::EAI_FAIL].include?(e.error_code), true)
```
hsbt, line 777: It's better to use `assert_include`. 

https://github.com/test-unit/test-unit/blob/bdd1668e1e3c0875fd1b3a59df8f6ec6472014a1/lib/test/unit/assertions.rb#L1671

hsbt (approved):

PR #9089, opened 2023-12-01
Add support for `defined?(a.b.c)`

tenderlove:
This PR adds support for `defined?` with method calls inside.

I think we can consider this to fix #1888.  However, as I was implementing this, I noticed some un-implemented code paths and various bugs.  I'll file bugs and fix them.  For example `defined?([a, b, c])` will just crash.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1465,12 +1465,29 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
     }
 }
 
+static void pm_compile_call(rb_iseq_t *iseq, const pm_call_node_t *call_node, LINK_ANCHOR *const ret, const uint8_t *src, bool popped, pm_scope_node_t *scope_node);
+
 void
-pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, const uint8_t *src, bool popped, pm_scope_node_t *scope_node,  NODE dummy_line_node, int lineno, bool in_condition, LABEL **lfinish)
+pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, const uint8_t *src, bool popped, pm_scope_node_t *scope_node,  NODE dummy_line_node, int lineno, bool in_condition, LABEL **lfinish, bool explicit_receiver)
 {
     // in_condition is the same as compile.c's needstr
     enum defined_type dtype = DEFINED_NOT_DEFINED;
     switch (PM_NODE_TYPE(node)) {
+      case PM_ARGUMENTS_NODE: {
+        const pm_arguments_node_t *cast = (pm_arguments_node_t *) node;
+        const pm_node_list_t *arguments = &cast->arguments;
+        for (size_t idx = 0; idx < arguments->size; idx++) {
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -1877,49 +1980,13 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
-        }
-
-        if (call_node->receiver == NULL) {
-            flags |= VM_CALL_FCALL;
-        }
-
-        ADD_SEND_R(ret, &dummy_line_node, method_id, INT2FIX(orig_argc), block_iseq, INT2FIX(flags), kw_arg);
-        PM_POP_IF_POPPED;
-        return;
+        pm_compile_call(iseq, call_node, ret, src, popped, scope_node);
```
kddnewton, line 1989: This looks like it will fall through, I'm assuming that's not intentional?

prism_compile.c
```diff
@@ -1500,14 +1502,42 @@ pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *co
...
         }
         ADD_INSN3(ret, &dummy_line_node, defined, INT2FIX(DEFINED_CONST_FROM),
                   ID2SYM(pm_constant_id_lookup(scope_node, ((pm_constant_read_node_t *)constant_path_node->child)->name)), PUSH_VAL(DEFINED_CONST));
         return;
       }
 
+      case PM_CALL_NODE: {
+        pm_call_node_t *call_node = ((pm_call_node_t *)node);
+        if (call_node->receiver) {
+          if (PM_NODE_TYPE_P(call_node->receiver, PM_SELF_NODE)) {
```
kddnewton, line 1517: Total nit, but the indentation is off here (should be 4 spaces).

kddnewton (changes requested):
In general looks good!

prism_compile.c
kddnewton: I think the style so far has been:

2 spaces between switch and case label, 2 spaces between case label and block, 4 spaces otherwise. Hopefully it's been consistent in that?

tenderlove: @kddnewton no, it hasn't been consistent.  I _think_ my link above shows an example of the inconsistency.  AFAIK, this whole function is not consistent with the rest of the file.

kddnewton: Ahh okay, ignore this then

kddnewton (approved):

HParker:
@tenderlove thanks! I think this is the last defined case we are missing. After this, I think we can make the default case `EXPR` instead of rb_bug, but we should double triple check we didn't miss any cases.

tenderlove:
> @tenderlove thanks! I think this is the last defined case we are missing. After this, I think we can make the default case `EXPR` instead of rb_bug, but we should double triple check we didn't miss any cases.

TBH I'd like to specify the behavior for all nodes and remove the `default` branch.  Then, since it's an `enum`, the compiler will warn us if any new nodes are introduced.  Might be overkill though, idk.

prism_compile.c
tenderlove: No, it's a bug.  Eileen pushed a fix for it 😅

kddnewton (approved):

jemmaissroff (approved):

PR #9090, opened 2023-12-01
Revert "[Bug #20004] Revert "[Feature #19422] Enable shared by default on macOS""

kateinoigakukun:
DO NOT MERGE until the next release.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -3680,7 +3680,7 @@ AS_IF([test "$EXEEXT" = .exe], [
 
 AS_CASE("$cross_compiling:${LIBPATHENV}", [yes:* | no:], [], [
     AC_MSG_CHECKING(whether wrapper for $LIBPATHENV is needed)
-    AS_IF([test x"$enable_shared" = xno],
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -3677,7 +3680,7 @@ AS_IF([test "$EXEEXT" = .exe], [
 
 AS_CASE("$cross_compiling:${LIBPATHENV}", [yes:* | no:], [], [
     AC_MSG_CHECKING(whether wrapper for $LIBPATHENV is needed)
-    AS_IF([env ${LIBPATHENV}=/lib /bin/sh -c ': ${'${LIBPATHENV}'?}' 2>/dev/null],
+    AS_IF([test x"$enable_shared" = xno],
```
nobu, line 3683: We may need to consider WCHAR env on Windows.

```suggestion
    AS_CASE(["$enable_shared:$target_os"], [no:*|*:mingw*|*:cygwin],
```

nobu (approved):

PR #9093, opened 2023-12-01
[PRISM] Use depth_offset for FOR Node

eightbitraptor:
This PR removes the transparent scope node mechanism that we implemented inside Prism to deal with the local lookup offset caused by FOR node's pushing scopes and instead uses the depth_offset mechanism recently introduced inside the CRuby prism compiler.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1375,6 +1375,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         case PM_FOR_NODE: {
             pm_for_node_t *cast = (pm_for_node_t *)node;
             scope->body = (pm_node_t *)cast->statements;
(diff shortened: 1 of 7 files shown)


prism_compile.c
```diff
@@ -1375,6 +1375,7 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
         case PM_FOR_NODE: {
             pm_for_node_t *cast = (pm_for_node_t *)node;
             scope->body = (pm_node_t *)cast->statements;
+            scope->local_depth_offset = 1;
```
kddnewton, line 1378: This probably needs to be `+1` in case of

```ruby
begin
rescue
  for foo in bar do end
end
```

kddnewton (changes requested):

jemmaissroff (approved):

PR #9105, opened 2023-12-02
[Fix] Support when nil is assigned to variable `name`

jinroq:
If we run the test `test/ruby/test_optimization.rb` in the following environment, it will fail.

* OS: Ubuntu 22.04
* Compiler: gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

error log:
````
   1) Failure:
TestRubyOptimization#test_objtostring [/home/jinroq/dev/ruby/ruby/test/ruby/test_optimization.rb:938]:
assert_separately failed with error message
pid 3585101 exit 1
| /home/jinroq/dev/ruby/ruby/lib/ruby_vm/rjit/c_pointer.rb:343:in `with_class_name': undefined method `empty?' for nil (NoMethodError)
````


k0kubun (approved):
Thanks!

PR #9114, opened 2023-12-04
[PRISM] Implement Retry node.

eightbitraptor:
Couple of things worth noting here

- compile rescue nodes before ensure: this makes sure the catch tables are defined in the correct order
- guard against double compilation of the ensure node body statements if both an ensure and a rescue are defined at the same time

This PR also fixes https://github.com/ruby/prism/issues/1964


test/ruby/test_compile_prism.rb
```diff
@@ -849,6 +849,26 @@ def test_RescueNode
       assert_prism_eval("begin; rescue; end")
     end
 
+    def test_RetryNode
```
jemmaissroff, line 852: Can we add a few more test cases? Specifically with empty begins, rescues with just a retry without anything else in the rescue?

jemmaissroff (changes requested):

PR #9115, opened 2023-12-04
[PRISM] allow enabling Prism via flag or env var

HParker:
Enable Prism using either --prism

    ruby --prism test.rb

or via env var

    RUBY_PRISM=1 ruby test.rb

This uses Prism to load the first file and files loaded with `require`. This does not use Prism for `eval`, but that can be done in a followup.


iseq.h
```diff
@@ -13,6 +13,7 @@
 #include "internal/gc.h"
 #include "shape.h"
 #include "vm_core.h"
+#include "prism/prism.h"
```
jemmaissroff, line 16: Is there a way we can do this without adding this dependency?

jemmaissroff (commented):
Looks good, just one q about if we can reduce dependencies

iseq.h
HParker: I think, as long as `rb_iseq_new_main_prism` takes a `pm_string_t` and `pm_options_t`, we need those types available.

Maybe we can hide the types behind a `void *` or a `VALUE`? Would that be better?

I looked into if we could move this to `prism_compile.c` instead of defining this in `iseq.c`, but `prism_compile.c` doesn't know about `iseq_alloc`.

Also, it looks like `iseq.c` already depends on `prism/prism.h` via `prism_compile.h`. 

Is the idea in reducing dependencies to decouple things, or is the worry this will make compile times longer since changes in prism will require a recompile on all things that use `iseq.h`?

jemmaissroff: The smell is actually a bit of `common.mk`. Historically we've been trying to keep the dependencies to a minimum to make it easier to pull Prism out if the Ruby core team ever decides to do something like that. But I hink in this case, you're right, it's not really worth the `void *` casting.

jemmaissroff (approved):

mame:
@jemmaissroff @HParker Hey, this is a user-facing feature. Please file a ticket on bugs.ruby-lang.org and get matz's approval to merge it. Please revert it for now. cc/ @kddnewton

hsbt:
Command line option and Environmental variables need to discuss naming and get approval from Matz.

HParker:
Thank you @mame and @hsbt. I opened https://github.com/ruby/ruby/pull/9133 so the conversation about the feature flag can happen. Please feel free to merge it.

I apologize for not following the correct process.

hsbt:
@HParker Thank you for your quick action ❤️

HParker:
@mame and @hsbt,

> Sorry for the hassle

No thank you both for the noticing and helping me learn the correct process! ❤️ 

I included mame's questions in the ticket. I think those are all really good points and many things I didn't consider enough.

PR #9118, opened 2023-12-04
Make env_clone and env_copy compaction safe

peterzhu2118:


vm.c
```diff
@@ -1218,6 +1218,10 @@ env_copy(const VALUE *src_ep, VALUE read_only_variables)
 
     VALUE *env_body = ZALLOC_N(VALUE, src_env->env_size); // fill with Qfalse
     VALUE *ep = &env_body[src_env->env_size - 2];
+    ep[VM_ENV_DATA_INDEX_ME_CREF] = src_ep[VM_ENV_DATA_INDEX_ME_CREF];
+    ep[VM_ENV_DATA_INDEX_FLAGS]   = src_ep[VM_ENV_DATA_INDEX_FLAGS] | VM_ENV_FLAG_ISOLATED;
+
+    const rb_env_t *copied_env = vm_env_new(ep, env_body, src_env->env_size, src_env->iseq);
     volatile VALUE prev_env = Qnil;
 
     if (read_only_variables) {
```
XrXr, line 1227: I think you need a RB_OBJ_WRITTEN when writing into `env_body` now

XrXr (commented):
Thanks! CI is showing some relevant crashes, though.

vm.c
peterzhu2118: Good point. I also needed to temporarily set `VM_ENV_FLAG_LOCAL` since the previous EP has not been set yet.

PR #9121, opened 2023-12-05
Pretenure callcache IMEMOs

jhawthorn:
For some time at least Ruby has pretenured Class objects, setting their age bits to the oldest immediately on allocation.

On Ruby 3.2 and earlier, objects which were referenced by an old object immediately were themselved promoted to old as well. This had the effect of implicitly always aging callcache IMEMOs as well, since they are (always?) referenced by classes.

This commit adds the same pretenuring to imemo_callcache via RVALUE_AGE_SET_CANDIDATE which is currently performed for CLASS/MODULE/ICLASS.

At GitHub we've seen significantly more time spent in GC right after boot on Ruby versions including [[Feature #19678] Don't immediately promote children of old objects](https://github.com/ruby/ruby/pull/7821). I don't know how much this PR will help but callcache objects are well represented in the long lived objects of our heaps (we end up with about 60k almost immediately), so if they do not age immediately we will need to pay for them on each GC until they become old.

---

I wrote an synthetic benchmark, where I allocate a bunch of callcaches via method calls and them measure a series of manual minor GCs.

<details>
<summary>benchmark script</summary>

```
NUM_METHODS = 1000
NUM_CLASSES = 600

class Foo
  NUM_METHODS.times do |i|
    class_eval "def foo#{i}; end"
  end

  class_eval <<~RUBY
    def call
      #{NUM_METHODS.times.map{|i| "foo#{i}" }.join(";")}
    end
  RUBY
end

classes = NUM_CLASSES.times.map do
  Class.new(Foo)
end

def benchmark
  before = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  yield
  after = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  puts "%.2fms" % ((after - before) * 1000.0)
end

4.times { GC.start }

GC.disable

classes.each { _1.new.call }

benchmark { GC.start(full_mark: false) }
benchmark { GC.start(full_mark: false) }
benchmark { GC.start(full_mark: false) }
benchmark { GC.start(full_mark: false) }
benchmark { GC.start(full_mark: false) }
benchmark { GC.start(full_mark: false) }
```

</details>

**Ruby 3.2**

```
$ ruby --disable-gems test_callcache.rb
30.02ms
27.07ms
0.08ms
0.05ms
0.05ms
0.05ms
```

**Ruby master**

```
$ ruby --disable-gems test_callcache.rb
21.23ms
17.22ms
18.31ms
16.97ms
0.09ms
0.06ms
```

**This branch**
```
$ ruby --disable-gems test_callcache.rb
21.90ms
16.88ms
0.09ms
0.05ms
0.05ms
0.05ms
```

---

I could see pulling the aging up a level so that we don't have this conditional in a hot path, but maybe there is a need or future intention in having it here.

cc @peterzhu2118 @eightbitraptor @byroot @tenderlove


byroot:
So I was actually advocating for removing this pre-aging, as I suspect it causes a bunch of ephemeral classes to make it to the old generation.

> At GitHub we've seen significantly more time spent in GC right after boot on Ruby versions including [[Feature #19678] Don't immediately promote children of old objects](https://github.com/ruby/ruby/pull/7821).

Are you not using `Process.warmup` or https://github.com/ko1/nakayoshi_fork ? That's the kind of problems it's meant to solve.

jhawthorn:
> Are you not using `Process.warmup` or https://github.com/ko1/nakayoshi_fork ? That's the kind of problems it's meant to solve.

callcaches tend to be allocated after boot as there's no good way (yet?) to warm them up. So those techniques do not help (as in the benchmark provided)

byroot:
> Many callcaches tend to be allocated after boot as there's no good way (yet?) to warm them up.

Ah my bad, I'm much less familiar with this part of the VM.

Any reason why we couldn't pre-allocate them? I suppose that it would waste memory if these callsites end up never executed?

But yeah, your explanation make sense, I suppose if anything it makes more sense to age these than even T_CLASS.

jhawthorn:
I'd like to explore pre-allocating callcaches, but yeah at least the naive way would be to allocate `ObjectSpace.each_object(Class).sum{ _1.instance_methods.count }` callcaches which at least for us would about 10x the size of our heap 😅 (though it's possible that could be pruned down a while bunch and also I haven't tested it).

Yeah, that was my thinking too, these could make even more sense than classes. If we decide to not pretenure Classes (which would be nice to do at least for singleton classes, for example), we could change this to instead copy the age of the class to the imemo on creation.

jhawthorn:
Looking through a production heap dump I found more cases than expected (still by far the minority minority) of callcaches which weren't being referenced (anymore?) by an oldgen Class, so I'll do some more investigation to better understand before merging.

jhawthorn:
Just to update this with some things I missed on a first pass. There are two types of callcaches w/r/t their lifetime: "regular" callcaches, which are cached on a class forever, and "orphan" callcaches, which are only used as inline caches (by `super` and refinements I believe) and are more likely to be replaced.

What I wrote about their lifetime matching that of the class obviously only applies to the "regular" type, so those are the only ones, if any, we should be pretenuring. I also agree it would be nice to not pretenure classes as they to may be short-lived.

Maybe the ideal version of this would be to, only when the callcache is attached to a class, copy the class age bits onto the callcache. That should avoid all downsides.

I'll close this for now.

PR #9124, opened 2023-12-05
YJIT: Simplify code page switching logic, remove an assert

XrXr:
We have received a report of `assert!( !cb.has_dropped_bytes())` in
set_page() failing. The only explanation for this seems to be memory
allocation failing in write_byte(). The if condition implies that
`current_write_pos < dst_pos < mem_size`, which rules out failing to
encode the relative jump. The has_capacity() assert above not tripping
implies that we were in a place in the page where write_byte() did
attempt to write the byte and potentially made a syscall in the process.

Remove the assert, since memory allocation could fail. Also, return
failure if the destination is outside of the code region to detect that
out-of-memory situation quicker.

Also: `YJIT: Assert code pages are not partially in-bounds`

ref: https://github.com/Shopify/ruby/issues/548

### DIFF
--- a/yjit/src/asm/mod.rs
+++ b/yjit/src/asm/mod.rs
@@ -189,22 +189,21 @@ impl CodeBlock {
         // but you need to waste some space for keeping write_pos for every single page.
         // It doesn't seem necessary for performance either. So we're currently not doing it.
         let dst_pos = self.get_page_pos(page_idx);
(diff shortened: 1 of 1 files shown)


maximecb (approved):

k0kubun (approved):
> since memory allocation could fail

Good point. This seems like a nice patch to address the reported issue.

PR #9125, opened 2023-12-05
Check need_major_gc during GC stress

peterzhu2118:
need_major_gc is set when a major GC is required. However, if gc_stress_no_major is also set, then it will not actually run a major GC.

For example, the following script will sometimes crash:

```ruby
def foo(garbage_count)
  garbage = garbage_count.times.map { [] }
  v0 = []
  v1 = []
  v2 = []
  v3 = []
  v4 = []
  v5 = []

  rand(1..10).times { GC.start(full_mark: false) }

  GC.stress = 1
  GC.auto_compact = true
  garbage = nil
  b = binding
  6.times { raise unless b.local_variable_get(:"v#{_1}") == [] }
  GC.start
  GC.stress = false
  GC.auto_compact = false
end

time = Time.now
(10_000..).each do
  foo _1
  break if (Time.now - time) > 2
end
```

With the following message:

```
test.rb:18: [BUG] cannot create a new page after major GC
```

cc. @XrXr


XrXr (approved):
A more reliable repro you might want to include in your commit instead: `./miniruby -e 'GC.stress = 1; 50000.times.map { [] }'` (also crashes 3.2.2)

I guess the flag should probably be renamed since it's not "no major" it's "don't force a major at every opportunity"...

PR #9126, opened 2023-12-05
[PRISM] Implement more `defined?` nodes

eileencodes:
Implements the following nodes for `defined?`:

* `PM_IMAGINARY_NODE`
* `PM_KEYWORD_HASH_NODE`
* `PM_MATCH_PREDICATE_NODE`
* `PM_SOURCE_ENCODING_NODE`
* `PM_SOURCE_FILE_NODE`
* `PM_SOURCE_LINE_NODE`
* `PM_SPLAT_NODE`

For arrays with splats in them, the upstream parser will ignore the rest of the elements and return `expression`. See [commit](https://github.com/ruby/ruby/commit/34f13f7bd818900806fe5c5d760a12925496b6a0). 

Each commit includes the Ruby code and instructions produced. For `PM_KEYWORD_HASH_NODE` I've included the instructions with and without optimizations.

I went through every example test for the prism compiler and checked how that code behaves with `defined?`. Other than the ones that return `PM_CALL_NODE`, I think this is all the nodes I can come up with.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1466,12 +1466,14 @@ pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *co
         break;
       case PM_ARRAY_NODE: {
         pm_array_node_t *array_node = (pm_array_node_t *) node;
-        for (size_t index = 0; index < array_node->elements.size; index++) {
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Looks great! My only comment would be that I see the array node is now falling through. Usually in prism we mark that with a `/* fallthrough */` comment, but honestly I have no idea if CRuby has that convention or not.

PR #9127, opened 2023-12-05
[PRISM] move rb_ruby_prism_ptr out of prism/prism

HParker:
This moves rb_ruby_prism_ptr out of prism/prism.h where i mistakenly put it.


HParker:
closing as the PR this fixes needs to be reverted until discussion about the feature flag can happen: https://github.com/ruby/ruby/pull/9133

PR #9142, opened 2023-12-06
Update string and xstring encodings

kddnewton:
This merges https://github.com/ruby/prism/pull/1990 and correctly sets encodings for string and xstring nodes. It does not set the correct encoding for individual pieces of interpolated string and interpolated xstring nodes, only because I want to get this merged first and then figure out what is missing.


prism/encoding.c
```diff
@@ -4212,9 +4212,9 @@ pm_encoding_shift_jis_char_width(const uint8_t *b, ptrdiff_t n) {
 }
 
 /**
- * This is the definition of all of the encodings that we support.
+ * This is the table of all of the encodings that prisms supports.
```
paracycle, line 4215:
```suggestion
 * This is the table of all of the encodings that prism supports.
```

prism/encoding.h
```diff
@@ -213,13 +213,27 @@ typedef enum {
...
     PM_ENCODING_WINDOWS_1258,
     PM_ENCODING_WINDOWS_31J,
-    PM_ENCODING_WINDOWS_874
+    PM_ENCODING_WINDOWS_874,
+    PM_ENCODING_MAXIMUM
 } pm_encoding_type_t;
 
 /**
- * This is the default UTF-8 encoding. We need it to quickly create parsers.
+ * This is the table of all of the encodings that prisms supports.
```
paracycle, line 221:
```suggestion
 * This is the table of all of the encodings that prism supports.
```

prism/encoding.c
kddnewton: These changes are on prism's side, so I'll update them there.

PR #9144, opened 2023-12-06
[PRISM] Correct depth offset for block local vars

eightbitraptor:
Blocks should always look at their own local table first, even when defined inside an ensure/rescue or something else that uses depth offset. We can ignore the depth offset if we're doing local lookups inside a block


kddnewton (approved):

tenderlove:
I'm not 100% sure the issue, but reverting this PR fixes this case:

```ruby
begin
  foo = 1
  m = nil
rescue
  tap { p foo }
end
```

We need foo to look up 2 environments, and clearing the local offset seems to break the math

tenderlove:
Related to this bug: https://github.com/ruby/prism/issues/2053

PR #9146, opened 2023-12-06
Improve Fiber#kill docs and simplify the NEWS entry to let the documentation explain

eregon:


cont.c
```diff
@@ -3264,12 +3266,18 @@ rb_fiber_m_raise(int argc, VALUE *argv, VALUE self)
  *  call-seq:
  *     fiber.kill -> nil
  *
- *  Terminates +fiber+ by raising an uncatchable exception, returning
- *  the terminated Fiber.
+ *  Terminates +fiber+ by raising an uncatchable exception.
```
ioquatix, line 3269: Just because it's uncatchable, does it mean it can be seen in `$!`? I think we should be careful about this wording, as in practice, no exception should be allocated, at least in CRuby.

eregon: I think it's the closest thing we can say, things like throw tags are CRuby impl details.
Conceptually it behaves like an uncatchable exception.
I think "uncatchable" kinda implies "will not set $!", because that would sort of capture it and also $! is only ever set to an instance of Ruby Exception.

eregon:
I'll merge this now so I don't forget about it.

eregon:
Ah my mistake, it doesn't auto-merge because of some transient issue failed a required job.

ioquatix (commented):
LGTM.

PR #9147, opened 2023-12-06
[PRISM] Add handling of implicit hash arguments

paracycle:
Arguments that are passed as a hash need special consideration since in certain case they are not treated as keyword arguments. For example, if a call is passing `"a" => 1` as an argument, this will be turned into an implicit hash argument and not a keyword argument.

The existing compiler checks to see if all hash nodes can be treated as keyword arguments. If they can, then it will treat them as keyword arguments. If not, then it will treat them as implicit hash arguments.

This commit implements the same logic inside the Prism compiler.

Fixes https://github.com/ruby/prism/issues/1894

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -993,16 +993,49 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
                       break;
                   }
                   else {
-                      *kw_arg = rb_xmalloc_mul_add(len, sizeof(VALUE), sizeof(struct rb_callinfo_kwarg));
-                      *flags = VM_CALL_KWARG;
-                      (*kw_arg)->keyword_len = (int) len;
+                    // We need to first figure out if all elements of the KeywordHashNode are AssocNodes
+                    // with static literal keys.
+                    // TODO: Figure this out from flags on the KeywordHashNode when Prism supports it
+                    bool all_keys_static_literals = true;
 
-                      // TODO: Method callers like `foo(a => b)`
-                      for (size_t i = 0; i < len; i++) {
-                          pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
-                          (*kw_arg)->keywords[i] = pm_static_literal_value(assoc->key, scope_node, parser);
-                          PM_COMPILE_NOT_POPPED(assoc->value);
-                      }
+                    for (size_t i = 0; i < len; i++) {
+                        pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+                        pm_node_t *key = assoc->key;
+
+                        if (!key || !PM_NODE_TYPE_P(key, PM_ASSOC_NODE) || !pm_static_literal_p(key)) {
+                            all_keys_static_literals = false;
+                            break;
+                        }
+                    }
+
+                    if (all_keys_static_literals) {
+                        // If they are all static literal keys then we can pass them as keyword arguments.
+                        *kw_arg = rb_xmalloc_mul_add(len, sizeof(VALUE), sizeof(struct rb_callinfo_kwarg));
+                        *flags = VM_CALL_KWARG;
+                        VALUE *keywords = (*kw_arg)->keywords;
+                        (*kw_arg)->references = 0;
+                        (*kw_arg)->keyword_len = (int)len;
+
+                        for (size_t i = 0; i < len; i++) {
+                            pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+                            pm_node_t *key = assoc->key;
+                            keywords[i] = pm_static_literal_value(key, scope_node, parser);
+                            PM_COMPILE_NOT_POPPED(assoc->value);
+                        }
+                    } else {
+                        // If they aren't all static literal keys then we need to construct a new hash
+                        // and pass that as an argument.
+                        orig_argc++;
+                        *flags |= VM_CALL_KW_SPLAT | VM_CALL_KW_SPLAT_MUT;
+
+                        for (size_t i = 0; i < len; i++) {
+                            pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+                            PM_COMPILE_NOT_POPPED(assoc->key);
+                            PM_COMPILE_NOT_POPPED(assoc->value);
+                        }
+
+                        ADD_INSN1(ret, &dummy_line_node, newhash, INT2FIX(len * 2));
+                    }
                   }
                   break;
               }
--- a/test/ruby/test_compile_prism.rb
+++ b/test/ruby/test_compile_prism.rb
@@ -1329,6 +1329,16 @@ module Prism
         end
         foo(1)
       CODE
+
+      assert_prism_eval(<<-CODE)
+        def self.prism_opt_var_trail_hash(a = nil, *b, c, **d); end
+        prism_opt_var_trail_hash("a")
+        prism_opt_var_trail_hash("a", c: 1)
+        prism_opt_var_trail_hash("a", "b")
+        prism_opt_var_trail_hash("a", "b", "c")
+        prism_opt_var_trail_hash("a", "b", "c", c: 1)
+        prism_opt_var_trail_hash("a", "b", "c", "c" => 0, c: 1)
+      CODE
     end
 
     def test_CallAndWriteNode


prism_compile.c
```diff
@@ -870,6 +870,58 @@ pm_compile_index_write_nodes_add_send(bool popped, LINK_ANCHOR *const ret, rb_is
...
+                    int *flags, const uint8_t *src, pm_scope_node_t *scope_node, NODE dummy_line_node, pm_parser_t *parser)
+{
+    size_t len = keyword_arg->elements.size;
+    int seen_nodes = 0;
+
+    for (size_t i = 0; i < len; i++) {
+        pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+        pm_node_t *key = assoc->key;
+
+        seen_nodes++;
```
jemmaissroff, line 886: Is `seen_nodes` equivalent to `i`?

prism_compile.c
```diff
@@ -970,16 +1022,21 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
                       break;
                   }
                   else {
-                      *kw_arg = rb_xmalloc_mul_add(len, sizeof(VALUE), sizeof(struct rb_callinfo_kwarg));
-                      *flags = VM_CALL_KWARG;
-                      (*kw_arg)->keyword_len = (int) len;
+                    int compiled = pm_compile_keyword_arg(iseq, ret, keyword_arg, kw_arg, flags, src, scope_node, dummy_line_node, parser);
```
jemmaissroff, line 1025: The way I'm understanding this code is that we're doing two passes over the `keyword_arg->elements` list. In the first pass (currently done within `pm_compile_keyword_arg` L882-894) we're looking to determine whether it contains at least one non-static literal value. This is the true / false that we're returning. In the true case, we do the compilation and flag setting in `pm_compile_keyword_arg`. In the false case, we do the compilation and flag setting here.

I think we could clean up this code a little bit by making the two passes more explicit, and probably inlining them both here (or at least being consistent about what happens where).

Pass one:
- We set a bool `arguments_contain_all_keywords` or a better name. We don't do any flag setting or compilation.

Pass two:
- We have an if / else based on the value of `arguments_contain_all_keywords`
- In each case we set flags appropriately
- In each case we loop over the arguments again, and compile according to the specific case
- We set  values on `kw_arg` appropriately

I think this will make the code more readable / understandable. It _also_ will have the added benefit of allowing us to pull the first pass into Prism itself. We should know that information at parse time, and can set it as a flag on the arguments node.

jemmaissroff (changes requested):
Thanks for doing this! I left a structural comment which I think will help us reason about what's happening here more easily, and potentially also extract some of this work into Prism (so we don't have to do two passes).

Also, can you add test cases to `test/ruby/test_compile_prism.rb`? These should go within the `test_CallNode` method

prism_compile.c
kddnewton: We already set `PM_NODE_FLAG_STATIC_LITERAL` in the parser for this reason. I think you should be able to just use that directly instead of the first pass

paracycle: Wait, hold on, I did know about the `PM_NODE_FLAG_STATIC_LITERAL` since that's exactly what I am using to see if the keys of the `AssocNode`s inside the `KeywordHashNode` are static literals or not.

@kddnewton Are you saying that the `KeywordHashNode` would also have a `PM_NODE_FLAG_STATIC_LITERAL` flag set if all of the elements inside have static literal keys?

kddnewton: It should, yeah. Unless I'm misremembering. I know we did it for arrays and hashes, but maybe we missed the keyword hash node?

jemmaissroff: Yeah, I don't think a KeywordHashNode has that flag. But also, it doesn't seem like that's exactly the logic of this? It seems like we're checking that each node is an AssocNode _and_ a static literal. 

I guess it depends how we'd implement static_literal on KeywordHashNode. But I'd think something like this `method_call(**{})` would have a static literal KeywordHashNode, but not fit into the case that the code describes

kddnewton: I don't think that would be a static literal, because that would contain an assoc-splat. (In that case the inner hash _would_ be a static literal.)

paracycle: Ok, let me summarize the way forward that I'll take:
1. I'll inline the logic into the original method and simplify as @jemmaissroff suggested and try to ship this PR.
2. I'll make another PR on `prism` to add the `PM_NODE_FLAG_STATIC_LITERAL` flag to `KeywordHashNode`s if all elements are `AssocNode`s with static literal keys (we don't care if the values are static literals or not).
3. Once that work is done, I'll come back and make another PR on `ruby/ruby` to use that flag for this logic.

Does that sound good to all of you?

prism_compile.c
paracycle: Yeah, got rid of this altogether.

prism_compile.c
paracycle: Made this refactor

paracycle:
> Also, can you add test cases to `test/ruby/test_compile_prism.rb`? These should go within the `test_CallNode` method

Added this test as well. I copied the complex cases from `test_opt_var_trail_hash` case in `test_scan_args` file that was failing to compile initially as test cases to ensure that we have various calling patterns covered.

prism_compile.c
```diff
@@ -993,16 +993,49 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
                       break;
                   }
                   else {
-                      *kw_arg = rb_xmalloc_mul_add(len, sizeof(VALUE), sizeof(struct rb_callinfo_kwarg));
-                      *flags = VM_CALL_KWARG;
-                      (*kw_arg)->keyword_len = (int) len;
+                    // We need to first figure out if all elements of the KeywordHashNode are AssocNodes
+                    // with static literal keys.
+                    // TODO: Figure this out from flags on the KeywordHashNode when Prism supports it
+                    int all_keys_static_literals = TRUE;
```
jemmaissroff, line 999: I think this can be a boolean

prism_compile.c
```diff
@@ -993,16 +993,49 @@ pm_setup_args(pm_arguments_node_t *arguments_node, int *flags, struct rb_callinf
...
+                        for (size_t i = 0; i < len; i++) {
+                            pm_assoc_node_t *assoc = (pm_assoc_node_t *)keyword_arg->elements.nodes[i];
+                            pm_node_t *key = assoc->key;
+                            keywords[i] = pm_static_literal_value(key, scope_node, parser);
+                            PM_COMPILE_NOT_POPPED(assoc->value);
+                        }
+                    } else {
+                        // If they aren't all static literal keys then we need to construct a new hash
+                        // and pass that as an argument.
+                        orig_argc++;
```
jemmaissroff, line 1028: Why do we only need to increment `orig_argc` in this case?

jemmaissroff (changes requested):

prism_compile.c
kddnewton: Actually @paracycle I think we should skip the flag on keyword hash node. I think you should do it as you have here.

STATIC_LITERAL is enabled on a couple of nodes:

* false
* float
* imaginary
* rational
* integer
* nil
* regexp
* source encoding
* source file
* source line
* string (if frozen)
* symbol
* true

I think all of these intuitively make sense. Then we have the containers. It gets a little weird:

* arrays - if they contain all static literals AND do not contain any other containers
* assocs - if the key and value are static literals AND do not contain any other containers
* hash - if all elements are both static literals AND do not contain any other containers
* range - if both left and right are either omitted, nil, or integers

I think the range one is already weird, but at least somewhat consistent. I think adding it for keyword hash where it would only check the keys would be very surprising. 

That being said, I'm very okay with adding a _different_ flag from static literal to keyword hashes that indicates that all of its keys are static literal. I just don't want to reuse that same flag.

Also I kind of want to rename the range flag for the same reason.

prism_compile.c
paracycle: I _think_ the reason is that, in this case, unlike the other one, we pass the hash that we build up as an extra positional argument and not as keyword arguments. So the `orig_argc` should be incremented to account for that extra argument.

Based on the current `compile.c`, that's my understanding but I am not 100% clear on this.

prism_compile.c
paracycle: Sure! Updated. To be fair, back in my day, we didn't have `bool` as a type, so I was erring on the side of caution and falling back to old habits by using an `int` type 😄.

prism_compile.c
paracycle: @kddnewton Thanks. I was thinking about the same thing regarding the `STATIC_LITERAL` flag on keyword hash nodes representing all assoc nodes with static literal keys only and it felt weird to me as well.

I will define a different flag (something like `PM_NODE_FLAG_STATIC_LITERAL_ASSOC_NODE_KEYS` or similar -- open to suggestions) to denote this on keyword hashes, and I'll add/remove the flag inside `pm_keyword_hash_node_elements_append` depending on the element being added.

Does that sound good?

jemmaissroff (approved):
LGTM, thanks for your work here!

PR #9160, opened 2023-12-07
[DOC] Enhance docs for WeakMap and WeakKeyMap

zverok:
(Part of the work on preparing yearly changelog.)

In preparation for the new version, I wanted to slightly enhance the newly introduced class/method docs:
* `WeakKeyMap`: more class-level explanations, more details on `#getkey`, fix a slight bug in code of `#delete` example (using `h` instead of `m`)
* `WeekMap`: I started with docs for a new method `#delete`, but other methods' docs were very outdated/badly rendered, and as the class became of public use in recent versions, I took the liberty of improving them a bit and writing a new (short) class intro.

The new docs can be temporarily seen on my site:
* [WeakMap](https://zverok.space/ruby-rdoc/ObjectSpace/WeakMap.html)
* [WeakKeyMap](https://zverok.space/ruby-rdoc/ObjectSpace/WeakKeyMap.html)

Compare to the current `master` versions:
* [WeakMap](https://docs.ruby-lang.org/en/master/ObjectSpace/WeakMap.html)
* [WeakKeyMap](https://docs.ruby-lang.org/en/master/ObjectSpace/WeakKeyMap.html)

### DIFF
--- a/weakmap.c
+++ b/weakmap.c
@@ -276,7 +276,14 @@ wmap_each_i(VALUE key, VALUE val, st_data_t _)
     rb_yield_values(2, key, val);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each {|key, val| ... } -> self
+ *
+ * Iterates over keys and values. Note that unlike other collections,
+ * +each+ without block isn't supported.
+ *
+ */
 static VALUE
 wmap_each(VALUE self)
 {
@@ -294,7 +301,14 @@ wmap_each_key_i(VALUE key, VALUE _val, st_data_t _data)
     rb_yield(key);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each_key {|key| ... } -> self
+ *
+ * Iterates over keys. Note that unlike other collections,
+ * +each_key+ without block isn't supported.
+ *
+ */
 static VALUE
 wmap_each_key(VALUE self)
 {
@@ -312,7 +326,14 @@ wmap_each_value_i(VALUE _key, VALUE val, st_data_t _data)
     rb_yield(val);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each_value {|val| ... } -> self
+ *
+ * Iterates over values. Note that unlike other collections,
+ * +each_value+ without block isn't supported.
+ *
+ */
 static VALUE
 wmap_each_value(VALUE self)
 {
@@ -332,7 +353,13 @@ wmap_keys_i(st_data_t key, st_data_t _, st_data_t arg)
     rb_ary_push(ary, key);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.keys -> new_array
+ *
+ * Returns a new Array containing all keys in the map.
+ *
+ */
 static VALUE
 wmap_keys(VALUE self)
 {
@@ -353,7 +380,13 @@ wmap_values_i(st_data_t key, st_data_t val, st_data_t arg)
     rb_ary_push(ary, (VALUE)val);
 }
 
-/* Iterates over values and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.values -> new_array
+ *
+ * Returns a new Array containing all values in the map.
+ *
+ */
 static VALUE
 wmap_values(VALUE self)
 {
@@ -400,7 +433,15 @@ wmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t new_key_ptr, int exi
     return ST_CONTINUE;
 }
 
-/* Creates a weak reference from the given key to the given value */
+/*
+ *  call-seq:
+ *    map[key] = value -> value
+ *
+ *  Associates the given +value+ with the given +key+.
+ *
+ *  If the given +key+ exists, replaces its value with the given +value+;
+ *  the ordering is not affected.
+ */
 static VALUE
 wmap_aset(VALUE self, VALUE key, VALUE val)
 {
@@ -434,7 +475,14 @@ wmap_lookup(VALUE self, VALUE key)
     return *(VALUE *)data;
 }
 
-/* Retrieves a weakly referenced object with the given key */
+/*
+ *  call-seq:
+ *    map[key] -> value
+ *
+ *  Returns the value associated with the given +key+ if found.
+ *
+ *  If +key+ is not found, returns +nil+.
+ */
 static VALUE
 wmap_aref(VALUE self, VALUE key)
 {
@@ -442,7 +490,34 @@ wmap_aref(VALUE self, VALUE key)
     return !UNDEF_P(obj) ? obj : Qnil;
 }
 
-/* Delete the given key from the map */
+/*
+ *  call-seq:
+ *    map.delete(key) -> value or nil
+ *    map.delete(key) {|key| ... } -> object
+ *
+ *  Deletes the entry for the given +key+ and returns its associated value.
+ *
+ *  If no block is given and +key+ is found, deletes the entry and returns the associated value:
+ *    m = ObjectSpace::WeakMap.new
+ *    key = "foo"
+ *    m[key] = 1
+ *    m.delete(key) # => 1
+ *    m[key] # => nil
+ *
+ *  If no block is given and +key+ is not found, returns +nil+.
+ *
+ *  If a block is given and +key+ is found, ignores the block,
+ *  deletes the entry, and returns the associated value:
+ *    m = ObjectSpace::WeakMap.new
+ *    key = "foo"
+ *    m[key] = 2
+ *    m.delete(key) { |key| raise 'Will never happen'} # => 2
+ *
+ *  If a block is given and +key+ is not found,
+ *  yields the +key+ to the block and returns the block's return value:
+ *    m = ObjectSpace::WeakMap.new
+ *    m.delete("nosuch") { |key| "Key #{key} not found" } # => "Key nosuch not found"
+ */
 static VALUE
 wmap_delete(VALUE self, VALUE key)
 {
@@ -473,14 +548,24 @@ wmap_delete(VALUE self, VALUE key)
     }
 }
 
-/* Returns +true+ if +key+ is registered */
+/*
+ *  call-seq:
+ *    map.key?(key) -> true or false
+ *
+ *  Returns +true+ if +key+ is a key in +self+, otherwise +false+.
+ */
 static VALUE
 wmap_has_key(VALUE self, VALUE key)
 {
     return RBOOL(!UNDEF_P(wmap_lookup(self, key)));
 }
 
-/* Returns the number of referenced objects */
+/*
+ * call-seq:
+ *   map.size -> number
+ *
+ * Returns the number of referenced objects
+ */
 static VALUE
 wmap_size(VALUE self)
 {
@@ -714,7 +799,7 @@ wkmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t data_args, int exis
  *  call-seq:
  *    map[key] = value -> value
  *
- *  Associates the given +value+ with the given +key+; returns +value+.
+ *  Associates the given +value+ with the given +key+
  *
  *  The reference to +key+ is weak, so when there is no other reference
  *  to +key+ it may be garbage collected.
@@ -755,7 +840,8 @@ wkmap_aset(VALUE self, VALUE key, VALUE val)
  *
  *  If no block is given and +key+ is found, deletes the entry and returns the associated value:
  *    m = ObjectSpace::WeakKeyMap.new
- *    m["foo"] = 1
+ *    key = "foo" # to hold reference to the key
+ *    m[key] = 1
  *    m.delete("foo") # => 1
  *    m["foo"] # => nil
  *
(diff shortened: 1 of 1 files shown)


weakmap.c
```diff
@@ -274,7 +274,15 @@ wmap_each_i(VALUE key, VALUE val, st_data_t _)
...
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each {|key, val| ... } -> self
+ *   map.each_pair {|key, val| ... } -> self
+ *
+ * Iterates over keys and values. Note that unlike other collections,
+ * +each+ without block is forbidden.
```
byroot, line 283:
```suggestion
 * +each+ without block isn't supported.
```

weakmap.c
```diff
@@ -292,7 +300,14 @@ wmap_each_key_i(VALUE key, VALUE _val, st_data_t _data)
     rb_yield(key);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each_key {|key| ... } -> self
+ *
+ * Iterates over keys. Note that unlike other collections,
+ * +each_key+ without block is forbidden.
```
byroot, line 308:
```suggestion
 * +each_key+ without block isn't supported.
```

weakmap.c
```diff
@@ -310,7 +325,14 @@ wmap_each_value_i(VALUE _key, VALUE val, st_data_t _data)
     rb_yield(val);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each_value {|val| ... } -> self
+ *
+ * Iterates over values. Note that unlike other collections,
+ * +each_value+ without block is forbidden.
```
byroot, line 333:
```suggestion
 * +each_value+ without block isn't supported.
```

weakmap.c
```diff
@@ -398,7 +432,30 @@ wmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t new_key_ptr, int exi
...
 
-/* Creates a weak reference from the given key to the given value */
+/*
+ *  call-seq:
+ *    map[key] = value -> value
+ *
+ *  Associates the given +value+ with the given +key+; returns +value+.
+ *
+ *  The reference to both +key+ and +value+ are weak, so when there is no
+ *  other references to +key+ or +value+ left, they may be garbage collected.
```
byroot, line 442:
```suggestion
 *  The reference to both +key+ and +value+ are weak, so when there is no
 *  other references to either +key+ or +value+ left, they may be garbage collected.
```

weakmap.c
```diff
@@ -398,7 +432,30 @@ wmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t new_key_ptr, int exi
...
+ *     key = "foo"
+ *     val = Object.new
+ *     m[key] = val
+ *
+ *     m[key] #=> #<Object:0x00007fea7156c580>
+ *
+ *     val = nil # remove the other reference to value
+ *     GC.start
+ *
+ *     m[key] #=> nil
```
byroot, line 454: Note that since Ruby GC is conservative, this is not at all guaranteed to work as showcased. So I'd rather not include this kind of examples with `GC.start`.

weakmap.c
```diff
@@ -432,15 +489,49 @@ wmap_lookup(VALUE self, VALUE key)
...
+ *  Deletes the entry for the given +key+ and returns its associated value.
+ *
+ *  If no block is given and +key+ is found, deletes the entry and returns the associated value:
+ *    m = ObjectSpace::WeakMap.new
+ *    key = "foo"
+ *    m[key] = 1
+ *    m.delete(key) # => 1
+ *    m[key] # => nil
+ *
+ *  If no block given and +key+ is not found, returns +nil+.
```
byroot, line 521:
```suggestion
 *  If no block is given and +key+ is not found, returns +nil+.
```

weakmap.c
```diff
@@ -876,20 +994,71 @@ wkmap_inspect(VALUE self)
...
  *
- *  An ObjectSpace::WeakMap object holds references to
- *  any objects, but those objects can get garbage collected.
+ *  An ObjectSpace::WeakMap is a key-value map that holds weak references
+ *  to its keys and values, so they can be garbage-collected when there are
+ *  no more references left.
  *
- *  This class is mostly used internally by WeakRef, please use
- *  +lib/weakref.rb+ for the public interface.
+ *  Keys in the map are compare by identity.
```
byroot, line 1001:
```suggestion
 *  Keys in the map are compared by identity.
```

weakmap.c
```diff
@@ -876,20 +994,71 @@ wkmap_inspect(VALUE self)
...
- *  any objects, but objects uses as keys can be garbage collected.
+ *  An ObjectSpace::WeakKeyMap is a key-value map that holds weak references
+ *  to its keys, so they can be garbage collected when there is no more references.
+ *
+ *  Unlike ObjectSpace::WeakMap:
+ *
+ *  * references to values are _strong_, so they aren't garbage collected while
+ *    they are in the map;
+ *  * keys are compared by value (using Object#eql?), not by identity;
+ *  * only garbage-collecatable objects can be used as keys.
```
byroot, line 1018:
```suggestion
 *  * only garbage-collectable objects can be used as keys.
```

weakmap.c
byroot: I'd prefer something that say objects are never guaranteed to be GCed.

zverok: @byroot What's your preferred wording then? 

I still feel it would be good to leave some examples. 
I am open to any suggestions that would make them less misleading, though I think that `GC.start` (which would work in a simple test script to "experiment how it works") + an explanation that `GC.start` is only for demo purposes might be enough.

byroot: IMO a comment that say there is no longer any strong reference on that object is enough. I really don't like examples with `GC.start` it tend to really confuse people.

byroot: That said, I don't consider myself good at documentation, so do what you want, I don't have a very strong opinion here.

zverok: @byroot It is just from all my experience of learning and teaching programming, the actual code examples (however artificial they are) for many people are telling much more than the most careful wording. 

And I think everybody understands it is artificial (and a note that "it is just for demonstrational purposes" would further clarify that), just like most people understand that examples showing `hash["foo"] = "bar"` don't claim "you should work with `"foo"` and `"bar"` in your real code."

`GC.start` is the only way to demonstrate the _actual behavior_ of the "weak maps" (and how they are different from "just maps") and can be invaluable for the initial understanding.

We had a similar discussion when designing [Ractor](https://docs.ruby-lang.org/en/master/Ractor.html) docs and eventually agreed [on introducing](https://docs.ruby-lang.org/en/master/Ractor.html#class-Ractor-label-Note+on+code+examples) very artificial (and, in general, non-deterministic) `wait` method, which, though, allowed to write artificial yet very demonstrative code.

So... Is there a chance you'll approve it in this form?.. (Considering I've already added the notice about `GC.start` to every example using it.)

byroot: > So... Is there a chance you'll approve it in this form?.

You don't need my approval. You asked for my opinion, I gave it, but I won't approve nor block documentation changes.

weakmap.c
```diff
@@ -274,7 +274,15 @@ wmap_each_i(VALUE key, VALUE val, st_data_t _)
     rb_yield_values(2, key, val);
 }
 
-/* Iterates over keys and objects in a weakly referenced object */
+/*
+ * call-seq:
+ *   map.each {|key, val| ... } -> self
+ *   map.each_pair {|key, val| ... } -> self
```
jeremyevans, line 280: Remove the `each_pair` line.  The documentation guide states "Omit aliases from the call-seq".

weakmap.c
```diff
@@ -398,7 +432,33 @@ wmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t new_key_ptr, int exi
     return ST_CONTINUE;
 }
 
-/* Creates a weak reference from the given key to the given value */
+/*
+ *  call-seq:
+ *    map[key] = value -> value
+ *
+ *  Associates the given +value+ with the given +key+; returns +value+.
```
jeremyevans, line 439: The method itself does not return `value`, it returns an integer.  The expression in the call-seq does, but only because assignment expressions return the right hand side.

weakmap.c
```diff
@@ -398,7 +432,33 @@ wmap_aset_replace(st_data_t *key, st_data_t *val, st_data_t new_key_ptr, int exi
...
+ *
+ *  The reference to both +key+ and +value+ are weak, so when there is no
```
jeremyevans, line 441: I think this should be class level documentation, not documentation specific to the `[]=` method, as it describes how the class works in general.

weakmap.c
```diff
@@ -432,15 +492,49 @@ wmap_lookup(VALUE self, VALUE key)
...
+ *
+ *  If a block is given and +key+ is found, ignores the block,
+ *  deletes the entry, and returns the associated value:
+ *    m = ObjectSpace::WeakMap.new
+ *    key = "foo"
+ *    m[key] = 2
+ *    m.delete(key) { |key| raise 'Will never happen'} # => 2
+ *
+ *  If a block is given and +key+ is not found,
+ *  calls the block and returns the block's return value:
```
jeremyevans, line 534: Maybe `yields the +key+ to the block and returns the block's return value`?

weakmap.c
```diff
@@ -471,14 +565,27 @@ wmap_delete(VALUE self, VALUE key)
     }
 }
 
-/* Returns +true+ if +key+ is registered */
+/*
+ *  call-seq:
+ *    map.key?(key) -> true or false
+ *    map.include?(key) -> true or false
```
jeremyevans, line 571: Probably best to only include the `include?` line, because the other are aliases.

weakmap.c
```diff
@@ -471,14 +565,27 @@ wmap_delete(VALUE self, VALUE key)
...
 static VALUE
 wmap_has_key(VALUE self, VALUE key)
 {
     return RBOOL(!UNDEF_P(wmap_lookup(self, key)));
 }
 
-/* Returns the number of referenced objects */
+/*
+ * call-seq:
+ *   map.size -> number
```
jeremyevans, line 584: Likewise, only include `size`, not `length`.

weakmap.c
```diff
@@ -761,12 +868,12 @@ wkmap_aset(VALUE self, VALUE key, VALUE val)
  *  deletes the entry, and returns the associated value:
  *    m = ObjectSpace::WeakKeyMap.new
  *    m["foo"] = 2
- *    h.delete("foo") { |key| raise 'Will never happen'} # => 2
+ *    m.delete("foo") { |key| raise 'Will never happen'} # => 2
```
jeremyevans, line 871: This is a bad example (not your fault, it was wrong before your typo fix).  Without frozen string literals, the `"foo"` refer to separate objects, and the code raises.  Probably easiest to use a local variable to ensure a reference to the same object.

weakmap.c
```diff
@@ -801,6 +908,20 @@ wkmap_delete(VALUE self, VALUE key)
  *    map.getkey(key) -> existing_key or nil
  *
  *  Returns the existing equal key if it exists, otherwise returns +nil+.
+ *
+ *  This might be useful for implementing caches, so that only one copy of
+ *  some object would be used everywhere in the program:
+ *
+ *    Money = Data.define(:amount, :currency)
```
jeremyevans, line 915: I appreciate the `Data` was committed by you, but this could be expressed more simply using a hash/array/string/time instead of subclass of `Data`.

weakmap.c
```diff
@@ -876,20 +997,74 @@ wkmap_inspect(VALUE self)
...
+ *  not always lead to demonstrated results.)
+ *
+ *  The collection is especially useful for implementing caches of lightweight value
+ *  objects, so that only one copy of each value representation would be stored in
+ *  memory, but the copies that aren't used would be garbage-collected.
+ *
+ *    Measurement = Data.define(:amount, :unit)
+ *
+ *    def make_measurement(**)
+ *       @cache ||= ObjectSpace::WeakKeyMap
```
jeremyevans, line 1054: Probably best to avoid thread-unsafe constructions in examples.  Maybe make `CACHE` a constant outside the method?

weakmap.c
```diff
@@ -876,20 +997,74 @@ wkmap_inspect(VALUE self)
...
+ *       map["name"] #=> nil
+ *
+ *  (Note that GC.start is used here only for demonstrational purposes and might
+ *  not always lead to demonstrated results.)
+ *
+ *  The collection is especially useful for implementing caches of lightweight value
+ *  objects, so that only one copy of each value representation would be stored in
+ *  memory, but the copies that aren't used would be garbage-collected.
+ *
+ *    Measurement = Data.define(:amount, :unit)
```
jeremyevans, line 1051: Again, no reason to use a `Data` subclass here.  You could use `val` as a method parameter to simplify the example.

jeremyevans (changes requested):

weakmap.c
byroot: Note that this is `WeakKeyMap`, it does use regular equality semantic, so it doesn't matter if it's the same instance of `"foo"`.

jeremyevans: Good point.  However, without frozen string literals, couldn't the `"foo"` string passed to `m.[]` be garbage collected before the `m.delete` call?

byroot: If the `dupstring` endup up triggering GC, then yes it's possible. But it's like with the `GC.start` examples, it may not work if there is something on the stack than happen to holding the `VALUE`.

Which raise the same question of how accurate code snippets need to be, and also why I'm not a big fan of using code snippet to explain weak references. But again, I don't consider myself good at documenting so I don't have a strong opinion either way.

jeremyevans: Seems like assigning to a local variable and using that local variable for `m.[]` would be best in this case, since then you can be sure the value would not be garbage collected.  You don't need the variable for `m.delete`, since, as you mentioned, it does not compare by identity.

weakmap.c
zverok: Well, in this case, I just copied that from `WeakKeyMap` docs, and that was apparently just copied from the [Hash docs](https://docs.ruby-lang.org/en/master/Hash.html#method-i-5B-5D-3D) :shrug: So I didn't put much thought in it.

But I agree it is factually incorrect.

weakmap.c
zverok: My bad, I test-rendered the changes with older version of RDoc gem which wasn't yet be able to make the aliases docs just a link, so I used the old approach of providing every call-sequence (_now_ I remember we even discussed that before and the change was introduced, but :shrug: )

weakmap.c
zverok: Yeah, it wasn't to promote `Data` (I swear :joy:), just to demonstrate "small value object," but I agree it can be done with less extra concepts.

weakmap.c
zverok: Yeah, as I said, _now_ I remember it was the old RDoc behavior I relied upon, and we even discussed this earlier :shrug:

PR #9161, opened 2023-12-07
[PRISM] Account for multiple arguments when compiling ArgumentsNode

jemmaissroff:
BreakNode, ReturnNode and NextNode all compile the ArgumentsNode directly, but we weren't accounting for multiple arguments. If there is more than one argument, we need to also emit a newarray instruction to put the arguments onto the stack


prism_compile.c
```diff
@@ -1762,6 +1762,9 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
         for (size_t index = 0; index < node_list.size; index++) {
             PM_COMPILE(node_list.nodes[index]);
         }
+        if (node_list.size > 1) {
```
kddnewton, line 1765: Can you put a comment here that arguments nodes are not compiled directly by their parent call nodes? This is confusing to me only because usually I associate arguments with calls and seeing the `newarray` here is not something I would expect.

kddnewton (approved):
Small comment but good other than that

PR #9163, opened 2023-12-07
Bundled gems warning improvements

deivid-rodriguez:
The improvements are detailed in the separate commit messages.

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -89,10 +89,10 @@ module Gem::BUNDLED_GEMS
   end
 
   def self.warning?(name, specs: nil)
-    name = File.path(name) # name can be a feature name or a file path with String or Pathname
-    return if specs.to_a.map(&:name).include?(name.sub(LIBEXT, ""))
-    name = name.tr("/", "-")
-    _t, path = $:.resolve_feature_path(name)
+    feature = File.path(name) # name can be a feature name or a file path with String or Pathname
+    return if specs.to_a.map(&:name).include?(feature.sub(LIBEXT, ""))
+    _t, path = $:.resolve_feature_path(feature)
+    name = feature.tr("/", "-")
     if gem = find_gem(path)
       caller = caller_locations(3, 3).find {|c| c&.absolute_path}
       return if find_gem(caller&.absolute_path)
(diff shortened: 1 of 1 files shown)


hsbt:
Thank you @deivid-rodriguez 

I added https://github.com/ruby/ruby/commit/8f6cf72e661e36410848c1451335fd28a32cecdf

Example case is:

```
$ cat -p Gemfile
# frozen_string_literal: true

source "https://rubygems.org"

gem "bigdecimal"
gem "activesupport"
gem "mutex_m"
```

and run `bundle exec irb`. It shows like this:

```
$ bundle exec irb
/Users/hsbt/.local/share/gem/gems/activesupport-7.0.7.2/lib/active_support/core_ext/big_decimal/conversions.rb:4: warning: bigdecimal/util is found in bigdecimal, which will no longer be part of the default gems since Ruby 3.4.0. Add bigdecimal to your Gemfile or gemspec. Also contact author of activesupport-7.0.7.2 to add bigdecimal into its gemspec.
>>
```

We should detect `bigdecimal/util` as bigdecimal feature. But I have no smart idea now.

hsbt:
Ah sorry, I missed full commit message of https://github.com/ruby/ruby/pull/9163/commits/7dd861d4a2c76a9e2137e467d492b20fa93e8185.

We need to cover above case with `bigdecimal` declaration on Gemfile.

deivid-rodriguez:
I feel we should have some tests to cover this feature, because it's very complicated!

hsbt:
Agreed. I only test them with test-cases of Gemfile and ruby script in my local machine. I try to port them to ruby/ruby repo.

deivid-rodriguez:
@hsbt The problem you mention is because ruby master branch vendors bigdecimal 3.1.5 but that does not exist in rubygems.org. So because of this, Bundler defaults to using the default version even if it's specified in the Gemfile. If 3.1.5 is released to rubygems.org or you specify `gem "bigdecimal", "3.1.4"` in your Gemfile, then my patch works. I think this problem is very specific to the current state but should not happen in real life, so https://github.com/ruby/ruby/commit/7dd861d4a2c76a9e2137e467d492b20fa93e8185 could be reverted.

We could also consider that `Bundler` should always prefer rubygems.org gems over default gems for stuff specified directly in the `Gemfile`.

deivid-rodriguez:
Thank you, I will also look into improving Bundler here.

albus522:
Bootsnap passing in fully resolved paths making that the name seems to be the ultimate kicker here

hsbt:
I filed https://bugs.ruby-lang.org/issues/20060

PR #9165, opened 2023-12-07
Normalize bundler bindir

deivid-rodriguez:
This makes bundler consistent with all other gems, and makes the default installation of Bundler in the release package look like any other bundler installation.

Before (on preview3, for example), Bundler executable is installed at:

lib/ruby/gems/3.3.0+0/gems/bundler-2.5.0.dev/libexec/bundle

Now it's installed in the standard location:

lib/ruby/gems/3.3.0+0/gems/bundler-2.5.0.dev/exe/bundle

I'm working on adding a Ruby 3.3 build matrix to Bundler at https://github.com/rubygems/rubygems/pull/7233 and I believe this should fix some failures.

### DIFF
--- a/lib/bundler/bundler.gemspec
+++ b/lib/bundler/bundler.gemspec
@@ -39,7 +39,7 @@ Gem::Specification.new do |s|
   # include the gemspec itself because warbler breaks w/o it
   s.files += %w[lib/bundler/bundler.gemspec]
 
-  s.bindir        = "libexec"
+  s.bindir        = "exe"
   s.executables   = %w[bundle bundler]
   s.require_paths = ["lib"]
 end
--- a/spec/bundler/quality_spec.rb
+++ b/spec/bundler/quality_spec.rb
@@ -191,9 +191,10 @@ RSpec.describe "The library itself" do
   end
 
   it "ships the correct set of files" do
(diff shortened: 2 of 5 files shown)


deivid-rodriguez:
@hsbt I should've fixed CI here now.

Now there's only one default gem, erb, that [sets its bindir to `libexec`](https://github.com/ruby/ruby/blob/c69d1367a7994c70f956fc6486568823e8fa5812/lib/erb.gemspec#L26). However, that one is expected because the upstream repo also uses libexec: https://github.com/ruby/erb/tree/8db8b8b5086713175523f7cab4bc8b779f069709/libexec.

deivid-rodriguez:
I found, even if I still think we should fix this inconsistency, that this is actually a problem in rubygems default bundler installer, which I'm fixing at https://github.com/rubygems/rubygems/pull/7244.

That should allow us to add 3.3.0-rc1 to our build matrix. After that I will rebase this PR and make sure it passes on both sides. For now I'll set it to WIP.

deivid-rodriguez:
We're already testing with 3.3.0-rc1 upstream but I still think this is a good change, so I'm setting this as ready for you to review @hsbt!

deivid-rodriguez:
This is also passing on our side 👍.

hsbt (approved):
👍 I also confirmed this is working with my local.

PR #9167, opened 2023-12-08
introduce --parser runtime flag

HParker:
Introduce --parser runtime flag
    
```
ruby --parser=prism
```
    
also update the description:
    
```
$ ruby --parser=prism --version
ruby 3.3.0dev (2023-12-08T04:47:14Z add-parser-runtime.. 0616384c9f) +PRISM [x86_64-darwin23]
```

[Bug #20044](https://bugs.ruby-lang.org/issues/20044)


kddnewton (approved):

ractor_core.h
```diff
@@ -186,6 +186,7 @@ struct rb_ractor_struct {
     VALUE r_stderr;
     VALUE verbose;
     VALUE debug;
+    bool prism;
```
ko1, line 189: I believe it should be global, not ractor local.

ko1 (changes requested):

HParker:
I have been thinking about this change and if it should go in before or after 3.3.0.

I want to have it in 3.3.0, but I worry that it might make a bad first impression for Prism. I worry that someone will turn on prism via this option and be disappointed and decide Prism is not good when they run into issues. If we get this in shortly after the 3.3.0 release and have all the time between then and 3.3.1 to test, we might be able to make a better first impression.

I am open to whatever the team thinks is best, but right now I am nervous about including it in 3.3.0.

jeremyevans:
My opinion is that we should not add features to tiny releases, only to minor ones. So if `--parser=prism` is not ready for 3.3.0, it should wait until 3.4.0.

HParker:
In that case, I think we should probably add this in 3.3.0. My hope is that there is a way to use prism optionally as the parser before 3.4.0. I would like to at least be testing with prism in CI regularly in 2024.

kddnewton:
> My opinion is that we should not add features to tiny releases, only to minor ones. So if `--parser=prism` is not ready for 3.3.0, it should wait until 3.4.0.

I agree, I don't think we should add features in tiny releases. Given that we now compile all nodes and are just fixing bugs, I think we need to get this merged for 3.3.0 so that we're not stuck waiting for it for another year. The function of this flag is not for people to run it in production, the function is for us to gather feedback about what may or may not be breaking with people's code that we don't have access to. We need this flag for testing and I wouldn't want to add it in a tiny release.

kddnewton:
@ko1 can you rereview? It looks like @HParker addressed your concern. I just want to make sure you are happy with this before I merge it. Given timezones, I'll wait for your reply tomorrow.

kddnewton:
Sorry I just clicked the wrong button....

kddnewton (approved):
Looks great!

ruby.c
```diff
@@ -1407,6 +1408,18 @@ proc_long_options(ruby_cmdline_options_t *opt, const char *s, long argc, char **
     else if (is_option_with_arg("external-encoding", Qfalse, Qtrue)) {
         set_external_encoding_once(opt, s, 0);
     }
+    else if (is_option_with_arg("parser", Qfalse, Qtrue)) {
+        if (strcmp("prism", s) == 0) {
+            (*rb_ruby_prism_ptr()) = true;
+            rb_warn("The Prism compiler is currently experimental and compatibility with parse.y is not yet complete. Please report an issues you find on the prism issue tracker.");
```
eregon, line 1414: `an issues`, should be without `an`

kddnewton: I assume that's meant to be `any`

eregon: Yeah `any issue` would also work

test/ruby/test_rubyoptions.rb
```diff
@@ -287,6 +287,19 @@ def test_rjit_version
     end
   end
 
+  def test_parser_flag
+    warning = "/Users/hparker/code/ruby/exe/ruby: warning: The Prism compiler is currently experimental and compatibility with parse.y is not yet complete. Please report an issues you find on the prism issue tracker."
```
eregon, line 291: I guess this hardcoded path won't work in CI. It's likely possible to use a regexp or #include? Instead

HParker: Thanks yeah, updated to not include the path.

ruby.c
paracycle, line 1414: Can we not compare the "Prism compiler" to "parse.y" please? It is the "Prism parser" that is comparable to "parse.y", not the compiler based on that parse tree.

How does the following read?
```suggestion
            rb_warn("The compiler based on the Prism parser is currently experimental and compatibility with the current Ruby compiler is not yet complete. Please report any issues you find on the `ruby/prism` issue tracker.");
```

HParker: I feel like this message is a bit more complicated this way, but I agree it was inaccurate about which bits compare with each other.

test/ruby/test_rubyoptions.rb
```diff
@@ -287,6 +287,19 @@ def test_rjit_version
     end
   end
 
+  def test_parser_flag
+    warning = /The Prism compiler is currently experimental and compatibility with parse.y is not yet complete. Please report an issues you find on the prism issue tracker./
```
paracycle, line 291: This message needs to be updated as well.

ko1:
> @ko1 can you rereview? It looks like @HParker addressed your concern. I just want to make sure you are happy with this before I merge it. Given timezones, I'll wait for your reply tomorrow.

Sorry I missed the comment.
It's okay but `rb_ruby_` can be located on ruby.c (natural for me).

PR #9172, opened 2023-12-08
Fix op asgn calls with keywords

jeremyevans:
Examples of such calls:

```ruby
obj[kw: 1] += fo
obj[**kw] &&= bar
```

Before this patch, literal keywords would segfault in the compiler, and keyword splat usage would result in TypeError.

This handles all cases I can think of:

* literal keywords
* keyword splats
* combined with positional arguments
* combined with regular splats
* both with and without blocks
* both popped and non-popped cases

Fixes [Bug #20051]


compile.c
```diff
@@ -8915,62 +8933,116 @@ compile_op_asgn1(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node
...
             }
-            ADD_SEND_WITH_FLAG(ret, node, idASET, argc, INT2FIX(flag));
+            ADD_INSN(ret, node, swap);
+            ADD_SEND_R(ret, node, idASET, FIXNUM_INC(argc, 1), NULL, INT2FIX(flag), keywords);
+        }
+        else if (keyword_len) {
+            ADD_INSN(ret, node, dup);
+            ADD_INSN1(ret, node, opt_reverse, INT2FIX(keyword_len+boff+2));
+            ADD_INSN1(ret, node, opt_reverse, INT2FIX(keyword_len+boff+1));
+            ADD_INSN(ret, node, pop);
```
nobu, line 8977: This rotates the stack and moves the TOS before `keyword_len+boff` from the top?
Then `dup` and `pop` seem unnecessary to me.

```suggestion
            ADD_INSN1(ret, node, opt_reverse, INT2FIX(keyword_len+boff+1));
            ADD_INSN1(ret, node, opt_reverse, INT2FIX(keyword_len+boff+0));
```

jeremyevans:
@nobu thanks for the suggestion, I merged it into the pull request.  Is this OK to merge now?

PR #9175, opened 2023-12-08
Include pinned count in GC compaction test

KJTsanaktsidis:
(Just want to see if this fixes jeremy's tests on ubuntu i686, so opening a draft PR to get CI to run it)


KJTsanaktsidis:
We merged https://github.com/ruby/ruby/pull/9176 instead.

PR #9183, opened 2023-12-09
[DOC] Adjust some new features wording/examples.

zverok:
(Part of working on this year's changelog.)

Some adjustments of the wording of the new features docs.

The changes are relatively small, so I packed them into one PR. If this is unacceptable, I can split in several (every change is its own commit), but I hope that it can be reviewed easily.

The changes in this PR:
* [Range#overlap?](https://docs.ruby-lang.org/en/master/Range.html#method-i-overlap-3F): changed wording of the last paragraph from less packed and extracted inline code as examples.
* [Thread::Queue#freeze](https://docs.ruby-lang.org/en/master/Thread/Queue.html#method-i-freeze): adjusted method docs with explanations why the queue can't be frozen.
* [TracePoint](https://docs.ruby-lang.org/en/master/TracePoint.html): added missing new `:rescue` event.
* [Module#set_temporary_name](https://docs.ruby-lang.org/en/master/Module.html#method-i-set_temporary_name): slightly expanded examples and explanations with more visible consequences and possible usecases.
* [Process::Status#&](https://docs.ruby-lang.org/en/master/Process/Status.html#method-i-26) and [#>>](https://docs.ruby-lang.org/en/master/Process/Status.html#method-i-3E-3E): slightly expanded deprecation notice, removed `>>` usage from class header example (as it is deprecated, we shoudn't encourage it with the initial example).

Added Dec 12:
* [MatchData#named_captures](https://docs.ruby-lang.org/en/master/MatchData.html#method-i-named_captures): Fix rendering glitch due to wrong indentation.
* [Refinement#target](https://docs.ruby-lang.org/en/master/Refinement.html#method-i-target): Adjust docs a bit, add deprecation notice to `#refined_class`
* [Dir#fchdir](https://docs.ruby-lang.org/en/master/Dir.html#method-c-fchdir): Fix examples (they were confusing: the first one, after the description of the no-block form, demonstrated only a block form; the second one didn't demonstrate `#fchdir` at all)

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -1346,9 +1346,16 @@ rb_using_module(const rb_cref_t *cref, VALUE module)
 
 /*
  *  call-seq:
- *     target    -> class
+ *     target    -> class_or_module
  *
  *  Return the class or module refined by the receiver.
+ *
+ *     module M
+ *       refine String do
+ *       end
+ *     end
+ *
+ *     M.refinements[0].target # => String
  */
 VALUE
 rb_refinement_module_get_refined_class(VALUE module)
@@ -1362,6 +1369,8 @@ rb_refinement_module_get_refined_class(VALUE module)
 /*
  *  call-seq:
  *     refined_class    -> class
+ *
+ *  Deprecated; prefer #target.
  *
  *  Return the class refined by the receiver.
  */
--- a/range.c
+++ b/range.c
@@ -2391,18 +2391,17 @@ empty_region_p(VALUE beg, VALUE end, int excl)
  *    (1..2).overlap?(3..4)      # => false
  *    (1...3).overlap?(3..4)     # => false
  *
- *  This method assumes that there is no minimum value because
- *  Ruby lacks a standard method for determining minimum values.
- *  This assumption is invalid.
- *  For example, there is no value smaller than <tt>-Float::INFINITY</tt>,
- *  making <tt>(...-Float::INFINITY)</tt> an empty set.
- *  Consequently, <tt>(...-Float::INFINITY)</tt> has no elements in common with itself,
- *  yet <tt>(...-Float::INFINITY).overlap?((...-Float::INFINITY))<tt> returns
- *  +true+ due to this assumption.
- *  In general, if <tt>r = (...minimum); r.overlap?(r)</tt> returns +true+,
- *  where +minimum+ is a value that no value is smaller than.
- *  Such values include <tt>-Float::INFINITY</tt>, <tt>[]</tt>, <tt>""</tt>, and
- *  classes without subclasses.
+ *  Note that the method wouldn't make any assumptions about the beginless
+ *  range being actually empty, even if its upper bound is the minimum
+ *  possible value of its type, so all this would return +true+:
+ *
+ *     (...-Float::INFINITY).overlap?(...-Float::INFINITY) # => true
+ *     (..."").overlap?(..."") # => true
+ *     (...[]).overlap?(...[]) # => true
+ *
+ *  Even if those ranges are effectively empty (no number can be smaller than
+ *  <tt>-Float::INFINITY</tt>), they are still considered overlapping
+ *  with themselves.
  *
  *  Related: Range#cover?.
  */
--- a/thread_sync.c
+++ b/thread_sync.c
@@ -1163,8 +1163,9 @@ NORETURN(static VALUE rb_queue_freeze(VALUE self));
  * call-seq:
  *   freeze
  *
- * Raises an exception:
- *   Queue.new.freeze # Raises TypeError (cannot freeze #<Thread::Queue:0x...>)
+ * The queue can't be frozen, so this method raises an exception:
+ *   Thread::Queue.new.freeze # Raises TypeError (cannot freeze #<Thread::Queue:0x...>)
+ *
  */
 static VALUE
 rb_queue_freeze(VALUE self)
--- a/trace_point.rb
+++ b/trace_point.rb
@@ -37,6 +37,7 @@
 # +:c_call+:: call a C-language routine
 # +:c_return+:: return from a C-language routine
 # +:raise+:: raise an exception
+# +:rescue+:: rescue an exception
 # +:b_call+:: event hook at block entry
 # +:b_return+:: event hook at block ending
 # +:a_call+:: event hook at all calls (+call+, +b_call+, and +c_call+)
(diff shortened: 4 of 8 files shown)


thread_sync.c
```diff
@@ -1163,8 +1163,13 @@ NORETURN(static VALUE rb_queue_freeze(VALUE self));
  * call-seq:
  *   freeze
  *
- * Raises an exception:
- *   Queue.new.freeze # Raises TypeError (cannot freeze #<Thread::Queue:0x...>)
+ * The queue can't be frozen, so this method raises an exception:
+ *   Thread::Queue.new.freeze # Raises TypeError (cannot freeze #<Thread::Queue:0x...>)
+ *
+ * Unlike other collections, freezing a queue (to stop accepting #pop and #push)
```
jeremyevans, line 1169: I don't think the documentation needs to justify the implementation, only explain it, so I would remove this second paragraph.

trace_point.rb
```diff
@@ -396,7 +397,7 @@ def return_value
     Primitive.tracepoint_attr_return_value
   end
 
-  # Value from exception raised on the +:raise+ event
+  # Value from exception raised on the +:raise+ event, or rescued on the +rescue+ event.
```
jeremyevans, line 400: For consistency, either remove the colon before `raise`, or add it before `rescue`.

variable.c
```diff
@@ -162,21 +162,21 @@ is_constant_path(VALUE name)
...
+ *  as instances, constants, and methods.
  *
- *  After a permanent name is assigned, a temporary name can no longer be set,
- *  and this method raises a RuntimeError.
+ *  The name should be +nil+ or non-empty string that is not a valid constant
+ *  name (to avoid confusing between permanent and temporary names).
  *
- *  If the name given is not a string or is a zero length string, this method
- *  raises an ArgumentError.
+ *  The method can be useful to name classes and modules auto-generated from
```
jeremyevans, line 172: I would remove this paragraph. "The DSL" is ambiguous, "RSpec" is not in ruby core or stdlib, "Web-framework routes" could only use this if they were classes or modules.

variable.c
```diff
@@ -189,15 +189,27 @@ is_constant_path(VALUE name)
...
- *    n.set_temporary_name("fake_name")
+ *    c = Class.new do
+ *       def foo
+ *       end
+ *    end
+ *    c.set_temporary_name("MyClass(with description)")
+ *
+ *    c.new # => #<MyClass(with description):0x0....>
+ *    c.instance_method(:foo) # => #<UnboundMethod: MyClass(with description)#foo() ...>
+ *    c.new.method(:foo) #<Method: MyClass(with description)#foo() ...>
```
jeremyevans, line 200: Missing `# =>` in this line

jeremyevans (changes requested):

nobu:
FYI: `Docs:` tags do nothing for us.  `[DOC]` is to suppress CIs.

range.c
```diff
@@ -2391,18 +2391,17 @@ empty_region_p(VALUE beg, VALUE end, int excl)
...
- *  For example, there is no value smaller than <tt>-Float::INFINITY</tt>,
- *  making <tt>(...-Float::INFINITY)</tt> an empty set.
- *  Consequently, <tt>(...-Float::INFINITY)</tt> has no elements in common with itself,
- *  yet <tt>(...-Float::INFINITY).overlap?((...-Float::INFINITY))<tt> returns
- *  +true+ due to this assumption.
- *  In general, if <tt>r = (...minimum); r.overlap?(r)</tt> returns +true+,
- *  where +minimum+ is a value that no value is smaller than.
- *  Such values include <tt>-Float::INFINITY</tt>, <tt>[]</tt>, <tt>""</tt>, and
- *  classes without subclasses.
+ *  Note that the method wouldn't make any assumptions about the semi-open
```
mame, line 2394: Rather than introducing a new term "semi-open range," I would prefer "exclusive range," which is used elsewhere.

variable.c
zverok: @mame My point was to demonstrate that _all_ related objects are affected (as an emphasis on "why would this be useful").
Though we probably can just leave one of the two lines (`instance_method` vs `method`), did you mean that?

range.c
zverok: I used it here as a synonym for "beginless or endless" (not for "exclusive range"), yet the statement is related only to beginless one, so I'll use more precise wording.

zverok:
@jeremyevans I tried to address all of your concerns. I also added 3 more small commits (listed under `Dec 12` in the PR description), I hope that's OK for "fixes a bit of everything in docs" PR :shrug:

eval.c
```diff
@@ -1346,9 +1346,19 @@ rb_using_module(const rb_cref_t *cref, VALUE module)
...
 /*
  *  call-seq:
- *     target    -> class
+ *     target    -> class_or_module
  *
  *  Return the class or module refined by the receiver.
+ *
+ *     module M
+ *       refine String do
+ *         def indent(level)
```
jeremyevans, line 1355: Best remove the method definition here, it doesn't affect the result and is unrelated to the method being documented.

variable.c
```diff
@@ -189,15 +189,25 @@ is_constant_path(VALUE name)
  *    m.set_temporary_name(nil) # => #<Module:0x0000000102c68f38>
  *    m.name #=> nil
  *
- *    n = Module.new
- *    n.set_temporary_name("fake_name")
+ *    c = Class.new do
```
jeremyevans, line 192: I'm with @mame here, the method already has significant examples, I don't think we need to add to them.

jeremyevans (changes requested):
Looks almost ready.  Just a couple changes requested.

variable.c
zverok: OK. My point here was to demonstrate one of the cases where redefining `Module#name` and/or `Module#inspect` wouldn't fix the related object's `#inspect` to be meaningful (and adds to a justification "why somebody needs a new method to set a pretend-name"). But probably, it doesn't look important enough anyway.

zverok:
@jeremyevans I think I fixed all the comments. So I rebased it from `master`, and squashed all the fixups so it should be ready to merge.

Thanks for the reviews!

jeremyevans (approved):
Please squash when merging.

zverok:
@jeremyevans Do you mean we need to have just one commit with all the changes (because I tried to structure it in a "commit-per-class" fashion)?

jeremyevans:
Yes.

PR #9193, opened 2023-12-11
Array#drop_whille! and Hash.drop_while!

HoneyryderChuck:
Addresses https://bugs.ruby-lang.org/issues/20049


HoneyryderChuck:
I didn't do `Hash#drop_while` yet (as in, returning an hash), as i'd like feedback about these changes, and backcompat posture (the current enumerable `drop_while` would return an array for it).

PR #9194, opened 2023-12-11
YJIT: Fix missing arity check for splat calls to methods with optionals (+1 more)

XrXr:
Previously, for splat callsites that land in methods with optional
parameters, we didn't reject the case where the caller supplies too many
arguments. Accepting those calls previously caused YJIT to construct
corrupted control frames, which leads to crashes if the callee uses
certain stack walking methods such as Kernel#raise and String#gsub (for
setting up the frame-local `$~`).

Example crash in a debug build:

    Assertion Failed: ../vm_core.h:1375:VM_ENV_FLAGS:FIXNUM_P(flags)

Also see the other commit: YJIT: Rename helper function and correct counter name


XrXr:
For interest, the bug existed as early as #7698, though at that point there was no corruption, just letting calls through that shouldn't. Shows that these type of issues are hard to detect, since code with ArgumentError generally don't make it to CI.

k0kubun (approved):

k0kubun:
> code with ArgumentError generally don't make it to CI

I wonder if we can test ArgumentError cases in the call fuzzer 🤔

bootstraptest/test_yjit.rb
```diff
@@ -1,3 +1,29 @@
...
+  rescue ArgumentError
+    :ae
+  end
+
+  [test(Array.new 5), test([])]
+} unless defined?(RubyVM::RJIT) && RubyVM::RJIT.enabled? # Not yet working on RJIT
+
+# regression test for arity check with splat and send
+assert_equal '[:ae, :ae]', %q{
+  def two_reqs(a, b_, _ = 1) = a.gsub(a, a)
```
maximecb, line 16: I didn't know you could do that. The underscore argument... This is an argument that's being ignored but has a default value? 😅

maximecb:
> > code with ArgumentError generally don't make it to CI
> 
> I wonder if we can test ArgumentError cases in the call fuzzer 🤔

It's tricky to design it in a way to really support every possible combination of arguments but we can definitely extend it still.

We could try to catch ArgumentError and return a sentinel value as you seem to be doing here, and with a small probability sample a case with extra arguments or too few arguments.

bootstraptest/test_yjit.rb
XrXr: Normally all the parameters need to have different names, but `_` is an exception. For example, this parses fine:
```
$ ruby -ce 'def foo(_, _, _) = 1'
Syntax OK
```
It's probably useful for API stability maintenance.

I thought you could make a create a method that have fewer locals than parameters this way, but luckily no. Each `_` gets a local even though they're inaccessible from Ruby land, so these are not that special at the VM level.

PR #9206, opened 2023-12-12
Sync with upstream rubygems

deivid-rodriguez:
Bundler suite was broken by https://github.com/ruby/ruby/commit/67940b135cae091f0f13d2c531b2ef33660e82fc.


k0kubun (approved):
Thank you!

PR #9207, opened 2023-12-12
[PRISM] Fix segv with regex once flag

eileencodes:
When you have an interpolated regex with a `once` flag and local variable is outside the block created by the `once` flag, Prism would see a segv. This is because it was not taking the depth into account.

To fix this, we need to add 1 to the `local_depth_offset` on the `scope`.

Fixes: ruby/prism#2047

cc/ @jemmaissroff @tenderlove


jemmaissroff (approved):
Nice, thanks!

PR #9209, opened 2023-12-12
[PRISM] Fix attrset edge cases

eileencodes:
In some cases code may look like an attrset ID but should actually return the value of the method, not the passed values.

In ruby/prism#2051 a flag was added when we have a attribute write call. I used this flag to add the proper instructions when we have a real attrset instead of using `rb_is_attrset_id` which was kind of hacky anyway.

The value that should be returned in the newly added test is 42, not 2. Previously the changes we had made returned 2.

Related to ruby/prism#1715

cc/ @kddnewton @tenderlove

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1801,6 +1801,8 @@ pm_compile_call(rb_iseq_t *iseq, const pm_call_node_t *call_node, LINK_ANCHOR *c
     int lineno = (int)pm_newline_list_line_column(&newline_list, ((pm_node_t *)call_node)->location.start).line;
     NODE dummy_line_node = generate_dummy_line_node(lineno, lineno);
 
+    pm_node_t *pm_node = (pm_node_t *)call_node;
+
     int flags = 0;
     struct rb_callinfo_kwarg *kw_arg = NULL;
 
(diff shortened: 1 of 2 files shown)


kddnewton (approved):

prism_compile.c
```diff
@@ -1834,15 +1836,15 @@ pm_compile_call(rb_iseq_t *iseq, const pm_call_node_t *call_node, LINK_ANCHOR *c
...
-
-    ADD_SEND_R(ret, &dummy_line_node, method_id, INT2FIX(orig_argc), block_iseq, INT2FIX(flags), kw_arg);
-
-    if (rb_is_attrset_id(method_id)) {
+        ADD_SEND_R(ret, &dummy_line_node, method_id, INT2FIX(orig_argc), block_iseq, INT2FIX(flags), kw_arg);
         PM_POP;
     }
+    else {
+        ADD_SEND_R(ret, &dummy_line_node, method_id, INT2FIX(orig_argc), block_iseq, INT2FIX(flags), kw_arg);
+    }
```
eileencodes, line 1846: This is a more visible in split mode, but I felt like it was easier to read the 2 additional instructions surrounding the `method_id` into the conditional over conditionally setting the `setn` and `pop` on their own lines. But I can put this back if y'all prefer that style.

kddnewton: Ehh this seems fine.

PR #9212, opened 2023-12-12
YJIT: Fix off-by-one in Kernel#send type handling

XrXr:
Previously, if the method ID argument happens to be on one below the top
of the stack, we didn't overwrite the type of the stack slot, which
leaves an incorrect type for the stack slot. The included script tripped
asserts both with and without --yjit-verify-ctx.


k0kubun (approved):

maximecb (approved):
Beautiful PR 👌 Thanks for adding extra tests :)

PR #9221, opened 2023-12-13
RJIT: Just skip generating code for aarch64/arm64

k0kubun:
We merged https://github.com/ruby/ruby/pull/8222, but this resurrects automatic RJIT enablement on aarch64/arm64 as per @tenderlove's request.


tenderlove:
Thank you!

PR #9222, opened 2023-12-13
[PRISM] Add anon KW args to the block local table

eightbitraptor:
Fixes https://github.com/ruby/prism/issues/2054

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -4365,6 +4365,7 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         pm_node_list_t *posts_list = NULL;
         pm_node_list_t *requireds_list = NULL;
         pm_node_list_t *block_locals = NULL;
+        pm_node_t *block_param_keyword_rest = NULL;
 
         struct rb_iseq_constant_body *body = ISEQ_BODY(iseq);
 
@@ -4374,6 +4375,9 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
                 pm_block_parameters_node_t *block_parameters_node = (pm_block_parameters_node_t *)scope_node->parameters;
                 parameters_node = block_parameters_node->parameters;
                 block_locals = &block_parameters_node->locals;
+                if (parameters_node) {
+                    block_param_keyword_rest = parameters_node->keyword_rest;
+                }
                 break;
               }
               case PM_PARAMETERS_NODE: {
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -4699,18 +4699,25 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
                         body->param.flags.has_kwrest = true;
 
-                        pm_constant_id_t constant_id = ((pm_keyword_rest_parameter_node_t *)parameters_node->keyword_rest)->name;
+                        pm_constant_id_t constant_id = kw_rest_node->name;
                         if (constant_id) {
                             pm_insert_local_index(constant_id, local_index, index_lookup_table, local_table_for_iseq, scope_node);
-                            local_index++;
                         }
+                        else {
+                            const ID kw_splat_id = parse_location_symbol(&kw_rest_node->operator_loc, parser);
```
jemmaissroff, line 4715: I think this can be `ID kw_splat_id = idPow;`. It's very hard to find, but there are a bunch of builtin IDs in `defs/id.def`

prism_compile.c
```diff
@@ -4699,18 +4699,25 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                            local_table_for_iseq->ids[local_index] = kw_splat_id;
+                            local_table_for_iseq->size++;
```
jemmaissroff, line 4717: Why do we increment this here?

prism_compile.c
```diff
@@ -4699,18 +4699,25 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                            st_insert(index_lookup_table, constant_id, local_index);
```
jemmaissroff, line 4718: Do we have to add it to the index_lookup_table? Is there anything that will lookup the `**` in this table?

jemmaissroff (changes requested):
Thanks for handling this!

prism_compile.c
```diff
@@ -4446,6 +4450,10 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
             }
         }
 
+        if (keyword_rest) {
```
jemmaissroff, line 4453: Such a nitpick, but can we call this something like `block_param_keyword_rest` to disambiguate it from the below?

jemmaissroff (approved):
One optional nit, feel free to merge with / without

jemmaissroff (approved):
LGTM, thanks!

PR #9224, opened 2023-12-13
Fix memory leak in Hash#compare_by_identity

XrXr:
We didn't free the old ST before overwriting it which caused a leak.
Found with RUBY_FREE_ON_EXIT.

Co-authored-by: Peter Zhu <peter@peterzhu.ca>

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -4372,6 +4372,7 @@ rb_hash_compare_by_id(VALUE hash)
 
     rb_hash_foreach(hash, rb_hash_rehash_i, (VALUE)tmp);
 
+    rb_hash_free(hash);
     RHASH_ST_TABLE_SET(hash, identtable);
     RHASH_ST_CLEAR(tmp);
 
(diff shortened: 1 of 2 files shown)


maximecb:
@k0kubun this memory leak may have affected our production deployments. Could you deploy it on our SFR experimental YJIT 3.3.0 pod when you have a chance? I'm curious to see if it improves the memory usage vs 3.2.2 🤞

maximecb:
@XrXr thinking about it, since this is a memory leak in something that is core to the Ruby VM, and web workloads typically make heavy use of hashes, this actually seems super likely to affect SFR/Core, and generally anything that runs on Rails.

k0kubun:
Filed a PR for it. 

Both versions, including 3.2 which did not have this leak, eventually reach the PSS limit (1600MiB today, which we lowered from 2000MiB this Monday) and their unicorns get restarted, so I'm guessing we'll probably not see a significant difference in the max PSS/RSS. 3.3 runs major GC significantly less often than 3.2, so I'd be rather surprised if they end up using the same amount of memory.

XrXr:
I don't think this method is particular popular so not everyone using Hash is using it. The functionality of the method seems niche. https://docs.ruby-lang.org/en/3.2/Hash.html#method-i-compare_by_identity

PR #9232, opened 2023-12-13
[PRISM] Fix bugs in compiling optional keyword parameters

jemmaissroff:
This PR fixes two bugs when compiling optional keyword parameters:
- It moves keyword parameter compilation to STEP 5 in the parameters sequence, where the rest of compilation happens. This is important because keyword parameter compilation relies on the value of body->param.keyword->bits_start which gets set in an earlier step
- It compiles array and hash values for keyword parameters, which it didn't previously

Closes https://github.com/ruby/prism/issues/2063
Closes https://github.com/ruby/prism/issues/2068
Closes https://github.com/ruby/prism/issues/2065

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -4684,21 +4684,21 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
                     pm_node_t *value = cast->value;
                     name = cast->name;
 
-                    if (pm_static_literal_p(value)) {
-                        rb_ary_push(default_values, pm_static_literal_value(value, scope_node, parser));
-                    }
-                    else {
-                        LABEL *end_label = NEW_LABEL(nd_line(&dummy_line_node));
-
-                        int index = pm_lookup_local_index(iseq, scope_node, name);
-                        int kw_bits_idx = table_size - body->param.keyword->bits_start;
-                        ADD_INSN2(ret, &dummy_line_node, checkkeyword, INT2FIX(kw_bits_idx + VM_ENV_DATA_SIZE - 1), INT2FIX(i - 1));
-                        ADD_INSNL(ret, &dummy_line_node, branchif, end_label);
-                        PM_COMPILE(value);
-                        ADD_SETLOCAL(ret, &dummy_line_node, index, 0);
-
-                        ADD_LABEL(ret, end_label);
+                    switch PM_NODE_TYPE(value) {
+                      case PM_FALSE_NODE:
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -4684,21 +4684,21 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
                     pm_node_t *value = cast->value;
                     name = cast->name;
 
-                    if (pm_static_literal_p(value)) {
```
kddnewton, line 4687: I think this check was okay, could you reuse it but explicitly ignore hashes, arrays, and ranges? I think something we could do to make this more clear on the prism side is to make hashes, arrays, and ranges have different flags from static literal so that we don't have to keep ignoring them.

prism_compile.c
```diff
@@ -4684,21 +4684,21 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
-
-                        ADD_LABEL(ret, end_label);
+                    switch PM_NODE_TYPE(value) {
+                      case PM_FALSE_NODE:
+                      case PM_FLOAT_NODE:
+                      case PM_INTEGER_NODE:
+                      case PM_IMAGINARY_NODE:
+                      case PM_NIL_NODE:
+                      case PM_RATIONAL_NODE:
+                      case PM_STRING_NODE:
```
kddnewton, line 4694: The problem with this is that strings are only static literal if they're frozen

prism_compile.c
```diff
@@ -4920,6 +4920,57 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
+                switch (PM_NODE_TYPE(keyword_parameter_node)) {
+                  // def foo(a, (b, *c, d), e = 1, *f, g, (h, *i, j),  k:, l: 1, **m, &n)
+                  //                                                       ^^^^
+                  case PM_OPTIONAL_KEYWORD_PARAMETER_NODE: {
+                    pm_optional_keyword_parameter_node_t *cast = ((pm_optional_keyword_parameter_node_t *)keyword_parameter_node);
+
+                    pm_node_t *value = cast->value;
+                    name = cast->name;
+
+                    switch PM_NODE_TYPE(value) {
```
kddnewton, line 4937: Same thing here, we can check static_literal_p?

kddnewton (changes requested):

kddnewton (approved):

PR #9233, opened 2023-12-14
[DOC] [ci-skip] Shorter example for `Module#instance_method`

okuramasafumi:
The previous example code was too complex and includes extra logics that's not relevant to its main usage: `bind`.
The new example code focuses on `bind_call` so that readers can understand how it works more easily.


nobu (approved):

nobu:
Not for this method directly; more usage of `UnboundMethod#bind_call` to call an overridden method may be useful.

PR #9234, opened 2023-12-14
Fix gem detection for names with dash

deivid-rodriguez:
Fixes https://bugs.ruby-lang.org/issues/20065.


hsbt:
Thank you!

FYI: I added my checking scripts to https://github.com/ruby/ruby/tree/master/tool/test_for_warn_bundled_gems

deivid-rodriguez:
Thank you! Are they run in CI or something, or for now they are just something we need to run manually when changing this code?

PR #9236, opened 2023-12-14
Pattern matching

kddnewton:
Fixes https://github.com/ruby/prism/issues/1674
Fixes https://github.com/ruby/prism/issues/1675
Fixes https://github.com/ruby/prism/issues/1676
Fixes https://github.com/ruby/prism/issues/1677
Fixes https://github.com/ruby/prism/issues/1678
Fixes https://github.com/ruby/prism/issues/1679

This implements all of the remaining pattern matching expressions.

Also fixes https://github.com/ruby/prism/issues/1335.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -214,7 +214,7 @@ parse_symbol(const uint8_t *start, const uint8_t *end, pm_parser_t *parser)
 }
 
 static inline ID
-parse_string_symbol(pm_string_t *string, pm_parser_t *parser)
+parse_string_symbol(const pm_string_t *string, pm_parser_t *parser)
 {
     const uint8_t *start = pm_string_source(string);
     return parse_symbol(start, start + pm_string_length(string), parser);
@@ -407,6 +407,53 @@ pm_static_literal_value(const pm_node_t *node, pm_scope_node_t *scope_node, pm_p
         rb_raise(rb_eArgError, "Don't have a literal value for this type");
         return Qfalse;
     }
+}
+
+/**
+ * Currently, the ADD_INSN family of macros expects a NODE as the second
+ * parameter. It uses this node to determine the line number and the node ID for
+ * the instruction.
+ *
+ * Because prism does not use the NODE struct (or have node IDs for that matter)
+ * we need to generate a dummy node to pass to these macros. We also need to use
+ * the line number from the node to generate labels.
+ *
+ * We use this struct to store the dummy node and the line number together so
+ * that we can use it while we're compiling code.
+ *
+ * In the future, we'll need to eventually remove this dependency and figure out
+ * a more permanent solution. For the line numbers, this shouldn't be too much
+ * of a problem, we can redefine the ADD_INSN family of macros. For the node ID,
+ * we can probably replace it directly with the column information since we have
+ * that at the time that we're generating instructions. In theory this could
+ * make node ID unnecessary.
+ */
+typedef struct {
+    NODE node;
+    int lineno;
+} pm_line_node_t;
+
+/**
+ * The function generates a dummy node and stores the line number after it looks
+ * it up for the given scope and node. (The scope in this case is just used
+ * because it holds a reference to the parser, which holds a reference to the
+ * newline list that we need to look up the line numbers.)
+ */
+static void
+pm_line_node(pm_line_node_t *line_node, const pm_scope_node_t *scope_node, const pm_node_t *node)
+{
+    // First, clear out the pointer.
+    memset(line_node, 0, sizeof(pm_line_node_t));
+
+    // Next, retrieve the line and column information from prism.
+    pm_line_column_t line_column = pm_newline_list_line_column(&scope_node->parser->newline_list, node->location.start);
+
+    // Next, use the line number for the dummy node.
+    int lineno = (int) line_column.line;
+
+    nd_set_line(&line_node->node, lineno);
+    nd_set_node_id(&line_node->node, lineno);
+    line_node->lineno = lineno;
 }
 
 static void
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -409,6 +409,53 @@ pm_static_literal_value(const pm_node_t *node, pm_scope_node_t *scope_node, pm_p
     }
 }
 
+/**
```
jemmaissroff, line 412: I really like this in concept, but I'm a little confused by its place in this PR. More specifically, I think my order of preferences is:

1. Have this used everywhere that we're currently using dummy_line_node
2. Have this not exist
3. Have this used in some places, and not used in other places

It's clear this is better than the existing solution. However, having it used in some places and not others has ripple effects, like not allowing us to use many of the macros we've written to make prism compilation easier (`PM_POP`, `PM_PUTNIL`, `PM_DUP` etc). It then leaves us  with more things to update when we do, inevitably, move everything over to use this.

prism_compile.c
```diff
@@ -1195,60 +1242,706 @@ pm_compile_multi_write_lhs(rb_iseq_t *iseq, NODE dummy_line_node, const pm_node_
...
+    ADD_INSN1(ret, &line.node, topn, INT2FIX(3));
+    ADD_INSN1(ret, &line.node, topn, INT2FIX(5));
+    ADD_SEND(ret, &line.node, id_core_sprintf, INT2FIX(3));
+    ADD_INSN1(ret, &line.node, setn, INT2FIX(base_index + PM_PATTERN_BASE_INDEX_OFFSET_ERROR_STRING + 1));
+    ADD_INSN1(ret, &line.node, putobject, Qfalse);
+    ADD_INSN1(ret, &line.node, setn, INT2FIX(base_index + PM_PATTERN_BASE_INDEX_OFFSET_KEY_ERROR_P + 2));
+    ADD_INSN(ret, &line.node, pop);
+    ADD_INSN(ret, &line.node, pop);
+
+    ADD_LABEL(ret, match_succeeded_label);
```
jemmaissroff, line 1355: Why is the `match_succeeded_label` in a different place here than the above methods?

prism_compile.c
```diff
@@ -1195,60 +1242,706 @@ pm_compile_multi_write_lhs(rb_iseq_t *iseq, NODE dummy_line_node, const pm_node_
...
+ * This is a variation on compiling a pattern matching expression that is used
+ * to have the pattern matching instructions fall through to immediately after
+ * the pattern if it passes. Otherwise it jumps to the given unmatched_label
+ * label.
+ */
+static int
+pm_compile_pattern_match(rb_iseq_t *iseq, pm_scope_node_t *scope_node, const pm_node_t *node, LINK_ANCHOR *const ret, const uint8_t *src, LABEL *unmatched_label, bool in_single_pattern, bool in_alternation_pattern, bool use_deconstructed_cache, unsigned int base_index)
+{
+    LABEL *matched_label = NEW_LABEL(nd_line(node));
+    CHECK(pm_compile_pattern(iseq, scope_node, node, ret, src, matched_label, unmatched_label, in_single_pattern, in_alternation_pattern, use_deconstructed_cache, base_index));
```
jemmaissroff, line 1373: Curious about the use of `CHECK` and returning `int`s within this PR - seems like the only value returned is `COMPILE_OK`, so we're adding complexity by using `CHECK` and not making the functions `void`

test/ruby/test_compile_prism.rb
```diff
@@ -1697,6 +1697,153 @@ def test_AlternationPatternNode
...
+      assert_prism_eval("begin; Object.new => [1, 2, 3]; rescue NoMatchingPatternError; true; end")
+      assert_prism_eval("begin; [1, 2, 3] => Object[1, 2, 3]; rescue NoMatchingPatternError; true; end")
+    end
+
+    def test_CapturePatternNode
+      assert_prism_eval("[1] => [Integer => foo]")
+    end
+
+    def test_CaseMatchNode
+      assert_prism_eval(<<~RUBY)
```
jemmaissroff, line 1734: Can we add tests with empty arrays or empty hashes? We've run into several bugs in compilation with things being empty vs not

jemmaissroff (changes requested):
Looks really good overall! I left a couple more general questions. 

Also, did you test this against the pattern matching cases within Ruby's existing test files?

prism_compile.c
kddnewton: I believe because in this case it checks the result of matching the pattern, and if it fails then it sets `error_p` to false.

prism_compile.c
kddnewton: I copied it over from the other compiler. I know we're not using it everywhere, but we're going to have to use it everywhere eventually so I'd figured I'd add it now so that we don't miss it.

It can return `COMPILE_NG` in the case of an invalid local variable, this just propagates that check further.

test/ruby/test_compile_prism.rb
kddnewton: I have the empty hash in the hash pattern node but I'll add the empty array as well

prism_compile.c
kddnewton: Yeah, I'm sorry about putting that into this PR. The thing is, there were a lot of places where I was repeatedly calling out to determine the line number and it because really obvious in this PR.

My thinking here is that `dummy_line_node` is used everywhere, so it's not something we could fix in a single pull request. We're going to have to go incrementally with it. Right now the use of `pm_line_node_t` is isolated to pattern matching, so I was hoping we could start with it in this place.

The other thing is, all of those macros are going to have to change when we get line numbers correct. They're not going to be able to just reference `dummy_line_node`. So that work is going to have to be made anyway, so I figured I wouldn't dig us further down into using it with this PR.

I definitely understand your point and concern about the inconsistency here. I share it because I don't want to be doing things in two different ways. But I thought in this case it might be worth it since it's only used in pattern matching. What do you think?

jemmaissroff: Totally agreed that the macros will have to change anyways, but with this PR it's not only that the macros will have to change, but also that we'll have to go through all of this code and convert to macros too, which is fine, just to me seemed less optimal than leaving in dummy line node generation here, including the macro usage in this commit, and then updating the macros just where they are defined when introducing this way of working with line numbers.

I think fixing dummy_line_node in one PR is going to be pretty doable, a search and replace for `dummy_line_node` itself, and then fixing places where we pass it / generate it.

That all said, this is definitely not a hill I'd die on, so if you prefer it like this, that's okay too, and we can update all this code to use the macros when we switch over

jemmaissroff (approved):
Approving, feel free to decide what works best for you with the dummy_line_node usage

PR #9240, opened 2023-12-14
check modifcation while ar->st

ko1:
* delete `ar_try_convert` but use `ar_force_convert_table`
  to make program simple.
* `ar_force_convert_table` checks hash modification while
  calling `#hash` method with the following strategy:

1. copy keys (and vals) of ar_table
2. calc hashes from keys
3. check copied keys and hash's keys. if not matched, repeat from 1

fix [Bug #20050]


martinemde:
Thank you for looking at that issue and offering a fix. I appreciate your work!

PR #9241, opened 2023-12-14
Implement IO::Buffer on VWA

peterzhu2118:


casperisfine:
Are we sure this is fully safe? I'm unfamiliar with `IO::Buffer` but aren't pointers into the memory sometimes passed to native code etc? What if the buffer is relocated?

casperisfine:
Ah, nevermind, the actual buffer isn't in the struct, it's a pointer. Ignore me.

PR #9244, opened 2023-12-14
[PRISM] Implement safe navigation in CallNodes

jemmaissroff:
This commit implements safe navigation for CallNodes, CallAndWriteNodes and CallOperatorWriteNodes

Closes https://github.com/ruby/prism/issues/2084


kddnewton (approved):
You can use the new PM_Flag_P if you want but this is great as is

PR #9247, opened 2023-12-15
Reduce Implicit Array/Hash Allocations For Method Calls Involving Splats

jeremyevans:
This is a series of commits to reduce implicit array and hash allocations for method calls involving splats.  The following optimizations are included:

VM_CALL_ARGS_SPLAT_MUT callinfo flag

This is similar to the VM_CALL_KW_SPLAT_MUT flag added in Ruby 3.0.  This makes it so if the caller-side allocates an array for the method call, the flag is used to signal to the callee that it can reuse the allocated array and does not need to duplicate it.

concattoarray VM instruction

This instruction is similar to concatarray, but assumes the object being concatenated to is already a mutable array (such as those created by the splatarray VM instruction).  This optimizes method calls with multiple splats such as `f(*a,*a,*a)` (which previously allocated 3 arrays), allocating a single array instead of an array per splatted array.

pushtoarray VM instruction

This is similar, but handles non-splat arguments after a splat.  Previously, the VM would wrap those arguments in an array using newarray, and then call concatarray, such that `f(*a, a)` allocated 3 arrays caller-side.  This instruction just appends to the mutable array, reducing the number of arrays allocated to 1.

Allocationless Anonymous Splat Forwarding

This allows `def f(*, **) end` to not allocate an array or hash callee side.  This works because it is not possible to mutate the local variables, only pass them as splats to other methods.  This can make the following call chain allocation less:

```ruby
def f(a, b: 1) end
def g(*, **) f(*, **) end
ea

a = [1]
kw = {b: 2}
g(*a, **kw) # No allocations in this call
```

Switch ... argument forwards to not use ruby2_keywords

Using ruby2_keywords has probably been slower since Koichi's changes early in the Ruby 3.3 development cycle to not combine keyword splats into the positional splat array.  This removes the FORWARD_ARGS_WITH_RUBY2_KEYWORDS define, so that `def f(...) end` operates similarly to  `def f(*, **) end`, allowing allocationless splat forwarding

I created a benchmark for all of these changes.  In the method calls optimized by these changes, it is significantly faster:

```
named_multi_arg_splat
after:   5344097.6 i/s 
before:   3088134.0 i/s - 1.73x  slower

named_post_splat
after:   5401882.3 i/s 
before:   2629321.8 i/s - 2.05x  slower

anon_arg_splat
after:  12242780.9 i/s 
before:   6845413.2 i/s - 1.79x  slower

anon_arg_kw_splat
after:  11277398.7 i/s 
before:   4329509.4 i/s - 2.60x  slower

anon_multi_arg_splat
after:   5132699.5 i/s 
before:   3018103.7 i/s - 1.70x  slower

anon_post_splat
after:   5602915.1 i/s 
before:   2645185.5 i/s - 2.12x  slower

anon_kw_splat
after:  15403727.3 i/s 
before:   6249504.6 i/s - 2.46x  slower

anon_fw_to_named_splat
after:   2985715.3 i/s 
before:   2049159.9 i/s - 1.46x  slower

anon_fw_to_named_no_splat
after:   2941030.4 i/s 
before:   2100380.0 i/s - 1.40x  slower

fw_to_named_splat
after:   2801008.7 i/s 
before:   2012416.4 i/s - 1.39x  slower

fw_to_named_no_splat
after:   2742670.4 i/s 
before:   1957707.2 i/s - 1.40x  slower

fw_to_anon_to_named_splat
after:   2309246.6 i/s 
before:   1375924.6 i/s - 1.68x  slower

fw_to_anon_to_named_no_splat
after:   2193227.6 i/s 
before:   1351184.1 i/s - 1.62x  slower
```

Only fallout from these changes:

* Minor change to AST output for `...` not using `ruby2_keywords`
* Prism and rbs need updating for `...` not using `ruby2_keywords`
* typeprof need updating for new VM instructions (at least pushtoarray)

VM_CALL_ARGS_SPLAT_MUT, concattoarray, and pushtoarray only affect uncommon callsites (multiple splats, argument after splat).  Other commits only optimize calls to methods using anonymous splats or `...` argument forwarding.  Previously, there was no performance reason to use anonymous splats or `...` argument forwarding, but with this change, using them can be faster, and can offer a new way for users to optimize their code.


k0kubun:
I benchmarked this PR with headline benchmarks of yjit-bench:

## Interpreter

It speeds up psych-load and sequel, but it slows down activerecord. Is the slowdown intended?

```
before: ruby 3.3.0dev (2023-12-14T16:09:17Z master 247ce712fc) [x86_64-linux]
after: ruby 3.3.0dev (2023-12-15T00:48:09Z vm-call-args-splat.. b28e0be92e) [x86_64-linux]

activerecord    106.6        0.3         109.4       0.3         0.98           0.97
chunky-png      952.3        0.3         950.5       0.2         1.00           1.00
erubi-rails     2900.2       0.6         2891.2      0.6         1.01           1.00
hexapdf         3958.5       0.7         3947.6      0.5         1.00           1.00
liquid-c        106.1        0.4         106.1       0.4         1.00           1.00
liquid-compile  107.0        0.2         107.3       0.2         1.00           1.00
liquid-render   257.4        0.3         256.7       0.2         1.00           1.00
lobsters        1468.6       0.5         1474.4      0.5         1.09           1.00
mail            210.6        0.3         210.8       0.2         1.01           1.00
psych-load      3286.2       0.1         3191.2      0.1         1.03           1.03
railsbench      3369.0       0.1         3379.6      0.0         1.00           1.00
ruby-lsp        93.2         0.9         93.6        0.2         0.99           1.00
sequel          116.3        0.4         114.0       0.4         1.01           1.02
```

## YJIT

For your information, this is the impact on benchmarks with YJIT enabled.

```
before: ruby 3.3.0dev (2023-12-14T16:09:17Z master 247ce712fc) +YJIT [x86_64-linux]
after: ruby 3.3.0dev (2023-12-15T00:48:09Z vm-call-args-splat.. b28e0be92e) +YJIT [x86_64-linux]

activerecord    52.7         0.5         54.9        0.5         0.96           0.96
chunky-png      556.6        0.1         556.3       0.3         1.00           1.00
erubi-rails     1600.3       0.6         1620.2      0.5         0.98           0.99
hexapdf         2068.8       0.9         2126.6      0.9         0.99           0.97
liquid-c        71.8         0.6         72.1        0.6         0.99           1.00
liquid-compile  74.6         0.4         74.3        0.4         1.01           1.00
liquid-render   94.2         0.6         94.6        0.3         1.00           1.00
lobsters        947.4        1.6         963.8       1.0         1.03           0.98
mail            149.1        0.3         149.4       0.3         1.00           1.00
psych-load      1906.3       0.1         1939.8      0.1         0.98           0.98
railsbench      2037.4       0.1         2079.6      0.0         0.98           0.98
ruby-lsp        59.4         24.9        61.4        26.8        0.99           0.97
sequel          88.2         0.4         87.7        0.5         1.01           1.01
```

jeremyevans:
> I benchmarked this PR with headline benchmarks of yjit-bench:
> 
> ## Interpreter
> It speeds up psych-load and sequel, but it slows down activerecord. Is the slowdown intended?

No.  I'm not sure why activerecord would be slower and nothing else would.  I'm surprised it speeds up sequel and psych-load as much as it did.

> ## YJIT
> For your information, this is the impact on benchmarks with YJIT enabled.

YJIT slowdown is somewhat expected until YJIT supports the concattoarray and pushtoarray VM instructions.  YJIT could probably also get a speedup using the new callinfo flags this adds to avoid allocations.

maximecb:
@jeremyevans are you intending to merge this PR in for 3.3? I'm a wary of merging this just 10 days away from the release.

At Shopify we've been in the process of testing and fixing a lot of bugs. We've discovered and fixed a number of issues affecting both CRuby and YJIT. Many of which had to do with method calls.

IMO merging this just before the release means we don't have enough time to test. This is a fairly significant design change. It's also unfortunate that we didn't have time to discuss implications for YJIT or implement any kind of support. 

For those not intimately familiar with how Ruby splat calls get compiled, can you provide some YARV examples of how the bytecode might look before/after with the new instructions you want to introduce?

tenderlove:
In the future, I think we should try to get tests passing rather than disabling them. If you need help getting the Prism tests to pass, please feel free to ping any of the Prism maintainers (me, @kddnewton, etc). We'd like Prism to be treated as a first class library embedded in CRuby.

Thanks.

jeremyevans:
Agreed.  In this case, I did not know prism would break.  This PR was last rebased on January 8, and the prism CI that broke wasn't added until January 10.  I would have discovered this if I rebased again right before merging, but I didn't do that (I usually only rebase right before merging if the last rebase was over a month ago).

PR #9250, opened 2023-12-15
Fix `ARCH_FLAG`

nobu:


tool/m4/ruby_default_arch.m4
```diff
@@ -1,12 +1,19 @@
...
+    [arm64:arm*],        [ARCH_FLAG=-m64],
+    [arm*:arm*],         [ARCH_FLAG=-m32],
+    [x86_64:[i[3-6]86]], [ARCH_FLAG=-m64],
+    [x64:x86_64],        [],
+    [[i[3-6]86]:x86_64], [ARCH_FLAG=-m32],
+    [ppc64:ppc*],        [ARCH_FLAG=-m64],
+    [ppc*:ppc64],        [ARCH_FLAG=-m32],
+    AS_CASE([$build_os],
+        [darwin*], [ARCH_FLAG="-arch "$1],
+        [ARCH_FLAG=-march=$1]
```
stanhu, line 15: This change broke cross-compiling with `arm64-darwin` since `target_cpu` is set to `arm64`. As a result, `clang` attempts to compile with `-march=arm64`, which isn't a valid option: https://github.com/rake-compiler/rake-compiler-dock/pull/112#issuecomment-1869373411. 

I've confirmed that reverting this pull request makes cross-compiling work with `arm64-darwin`.

Previously cross-compiling worked when `-march=<arch>` were omitted entirely. I wonder if this flag is only needed for x86?

stanhu: Actually, the issue here is that `build_os` is not guaranteed to to be `arm64-darwin` on a cross-compiled system either. Should `target_cpu` be used?

stanhu: Filed https://bugs.ruby-lang.org/issues/20088.

nobu: Is this OK?
https://github.com/ruby/ruby/pull/9364

PR #9253, opened 2023-12-15
rb_bug_for_fatal_signal: exit with the right signal

casperisfine:
`die()` calls `abort()` which always exit as if `SIGABRT` was received.

This isn't very friendly with systems that automatically collect crashes as the `%s` parameter will be changed.

### DIFF
--- a/error.c
+++ b/error.c
@@ -1066,6 +1066,7 @@ rb_bug_for_fatal_signal(ruby_sighandler_t default_sighandler, int sig, const voi
 
     if (default_sighandler) default_sighandler(sig);
 
+    ruby_default_signal(sig);
     die();
 }
 


error.c
```diff
@@ -1066,6 +1066,7 @@ rb_bug_for_fatal_signal(ruby_sighandler_t default_sighandler, int sig, const voi
 
     if (default_sighandler) default_sighandler(sig);
 
+    ruby_default_signal(sig);
     die();
```
nobu, line 1070: `die()` may never be called?

nobu (approved):

error.c
casperisfine: Yes, it's in case for some reason `sig` isn't fatal? Maybe it doesn't make sense but I wasn't 100% sure so seemed safer.

PR #9258, opened 2023-12-15
Move internal ST functions to internal/st.h

peterzhu2118:
st_replace and st_init_existing_table_with_size are functions used internally in Ruby and should not be publicly visible.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -7650,6 +7650,7 @@ hash.$(OBJEXT): $(top_srcdir)/internal/imemo.h
 hash.$(OBJEXT): $(top_srcdir)/internal/object.h
 hash.$(OBJEXT): $(top_srcdir)/internal/proc.h
 hash.$(OBJEXT): $(top_srcdir)/internal/serial.h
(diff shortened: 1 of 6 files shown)


jeremyevans:
This changes public ABI after 3.3.0-rc1, so you should get @nurse approval before merging.

peterzhu2118:
I think the impact will be minimal since these were introduced in Ruby 3.3 and public by mistake. These are supposed to be private functions.

jeremyevans:
I agree regarding the minimal impact, and I think it is worth merging. However, it does change the ABI, and therefore it requires @nurse approval, and requires a mention in the release notes. FWIW, issues like this are why I recommended against stating that we would try to keep the same ABI.

peterzhu2118:
> FWIW, issues like this are why I recommended against stating that we would try to keep the same ABI.

I think there is significant value to freezing the ABI from features and deprecation. For mistakes like this where a new API was added for only this version, there are likely no users and so removing it should not break anyone's code.

PR #9259, opened 2023-12-15
Compile IndexTargetNode and CallTargetNode

eightbitraptor:
These are used within the context of the MultiWriteNode

Fixes https://github.com/ruby/prism/issues/2017


kddnewton (approved):
Heck yeah!

PR #9261, opened 2023-12-15
[PRISM] Pre-concatenate Strings in InterpolatedStringNode

jemmaissroff:
This commit concatenates String VALUEs within
InterpolatedStringNodes to allow us to preserve frozenness of concatenated strings such as `"a""b"`

Closes https://github.com/ruby/prism/issues/2072


kddnewton (approved):
Nice! I think I'd like to have a flag on interpolated* nodes on whether or not they actually have interpolation in them so we could fast-path this. This part shouldn't be necessary if we actually had that I think?

prism_compile.c
```diff
@@ -729,30 +729,62 @@ pm_compile_while(rb_iseq_t *iseq, int lineno, pm_node_flags_t flags, enum pm_nod
...
+                    current_string = string_value;
+                }
             }
             else {
+                if (RTEST(current_string)) {
+                    if (parser->frozen_string_literal) {
+                        ADD_INSN1(ret, &dummy_line_node, putobject, rb_str_freeze(current_string));
+                    }
+                    else {
+                        ADD_INSN1(ret, &dummy_line_node, putstring, current_string);
```
tenderlove, line 759: I think we should freeze this too.  If anyone manages to get a reference to it, they could mutate an object that iseqs point to.  Basically, I think _any_ objects that iseqs directly point to should be frozen.  `putstring` will just dup the frozen string, so we'll still end up with a mutable string on the stack.

test/ruby/test_compile_prism.rb
```diff
@@ -589,6 +589,17 @@ def test_InterpolatedStringNode
...
+      assert_prism_eval(<<-CODE)
+        # frozen_string_literal: true
+
+        ("a""b").frozen?
+      CODE
+      assert_prism_eval(<<-CODE)
+        # frozen_string_literal: true
+
+        ("a""b""#{1}").frozen?
+      CODE
```
paracycle, line 602: Can we add a test case to exercise the `putobject`/`putstring` that happens after the loop if we still have items in the `current_string`? 

That is:
```suggestion
      CODE
      assert_prism_eval(<<-CODE)
        # frozen_string_literal: true
        
        ("a""#{1}""b").frozen?
      CODE
```

PR #9266, opened 2023-12-16
We don't use tracked_files in before(:suite). It breaks with release package

hsbt:
https://github.com/rubygems/rubygems/issues/7297


hsbt:
This change is working with package test.

PR #9274, opened 2023-12-18
Optimize allocations in `Hash#compare_by_identity`

amomchilov:
* If `self` is an empty Hash, then we don't need to allocate the `tmp` Hash (since there's nothing to re-hash).
* If `self` is non-empty, then we don't need to bother with `ar_force_convert_table`. We're going to discard that table and replace it with the new `identtable` st table that we're stealing out of `tmp`.

So in either case, we get one less allocation.

This simple micro-benchmark clocks it as ~1.7x faster than before:

```
Comparison:
 After: {}.compare_by_identity: 13725103.3 i/s
Before: {}.compare_by_identity:  8085211.5 i/s - 1.70x  slower
```

Oddly, the difference is even larger (~1.85x) if you `GC.disable` (though this "same-ish: difference falls within error" conclusion seems pretty wrong...):

```
Comparison:
 After: {}.compare_by_identity:  9306969.8 i/s
Before: {}.compare_by_identity:  5036135.4 i/s - same-ish: difference falls within error
```

<details><summary><code>benchmark.rb</code></summary>

```ruby
#!/usr/bin/ruby

require "benchmark/ips"

# GC.disable

Benchmark.ips do |x|
  x.report("Before: {}.compare_by_identity") do |times|
    i = 0
    while (i += 1) < times
      {}.compare_by_identity
    end
  end
  
  x.compare!
  x.hold!("compare_by_identity.json")
  
  x.report(" After: {}.compare_by_identity") do |times|
    i = 0
    while (i += 1) < times
      {}.compare_by_identity
    end
  end
  
  x.compare!
end
```

</p>
</details>

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -4377,13 +4377,16 @@ rb_hash_compare_by_id(VALUE hash)
     if (hash_iterating_p(hash)) {
         rb_raise(rb_eRuntimeError, "compare_by_identity during iteration");
     }
-    ar_force_convert_table(hash, __FILE__, __LINE__);
-    HASH_ASSERT(RHASH_ST_TABLE_P(hash));
 
     if (RHASH_TABLE_EMPTY_P(hash)) {
         // Fast path: There's nothing to rehash, so we don't need a `tmp` table.
+        // We're most likely an AR table, so this will need an allocation.
+        ar_force_convert_table(hash, __FILE__, __LINE__);
+        HASH_ASSERT(RHASH_ST_TABLE_P(hash));
+
         RHASH_ST_TABLE(hash)->type = &identhash;
-    } else {
+    }
+    else {
         // Slow path: Need to rehash the members of `self` into a new
         // `tmp` table using the new `identhash` compare/hash functions.
         tmp = hash_alloc(0);
@@ -4391,8 +4394,10 @@ rb_hash_compare_by_id(VALUE hash)
         identtable = RHASH_ST_TABLE(tmp);
 
         rb_hash_foreach(hash, rb_hash_rehash_i, (VALUE)tmp);
-
         rb_hash_free(hash);
+
+        // We know for sure `identtable` is an st table,
+        // so we can skip `ar_force_convert_table` here.
         RHASH_ST_TABLE_SET(hash, identtable);
         RHASH_ST_CLEAR(tmp);
     }
--- a/hash.c
+++ b/hash.c
@@ -4380,15 +4380,22 @@ rb_hash_compare_by_id(VALUE hash)
     ar_force_convert_table(hash, __FILE__, __LINE__);
     HASH_ASSERT(RHASH_ST_TABLE_P(hash));
 
-    tmp = hash_alloc(0);
-    hash_st_table_init(tmp, &identhash, RHASH_SIZE(hash));
-    identtable = RHASH_ST_TABLE(tmp);
+    if (RHASH_TABLE_EMPTY_P(hash)) {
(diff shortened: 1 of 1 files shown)


hash.c
```diff
@@ -4377,18 +4377,29 @@ rb_hash_compare_by_id(VALUE hash)
...
+        ar_force_convert_table(hash, __FILE__, __LINE__);
+        HASH_ASSERT(RHASH_ST_TABLE_P(hash));
 
-    rb_hash_foreach(hash, rb_hash_rehash_i, (VALUE)tmp);
+        RHASH_ST_TABLE(hash)->type = &identhash;
+    } else {
+        // Slow path: Need to rehash the members of `self` into a new
+        // `tmp` table using the new `identhash` compare/hash functions.
+        tmp = hash_alloc(0);
+        hash_st_table_init(tmp, &identhash, RHASH_SIZE(hash));
```
amomchilov, line 4392: Could this just re-use `rb_ident_hash_new_with_size`?

 ```suggestion
        tmp = rb_ident_hash_new_with_size( RHASH_SIZE(hash));
        identtable = RHASH_ST_TABLE(tmp);
```

What's the difference between `hash_alloc(0)` and `rb_hash_new`?

It calls `hash_alloc(rb_cHash)` instead, but I don't understand the significance of passing `0` (null?) in this case.

File: hash.c [hunk 2]
```diff
@@ -4377,18 +4377,29 @@ rb_hash_compare_by_id(VALUE hash)
...
-    identtable = RHASH_ST_TABLE(tmp);
+    if (RHASH_TABLE_EMPTY_P(hash)) {
+        // Fast path: There's nothing to rehash, so we don't need a `tmp` table.
+        // We're most likely an AR table, so this will need an allocation.
+        ar_force_convert_table(hash, __FILE__, __LINE__);
+        HASH_ASSERT(RHASH_ST_TABLE_P(hash));
 
-    rb_hash_foreach(hash, rb_hash_rehash_i, (VALUE)tmp);
+        RHASH_ST_TABLE(hash)->type = &identhash;
+    } else {
```
peterzhu2118, line 4388: Nit:

```suggestion
    }
    else {
```

hash.c
peterzhu2118: We want to have a `klass` of 0 because we want this hash to be hidden (i.e. it does not get outputted in the `ObjectSpace` traversals).

peterzhu2118 (approved):
The code makes sense to me.

It looks like this gives a good speedup when the Hash is empty:

```ruby
require "benchmark"

puts(Benchmark.measure do
  10_000_000.times { {}.compare_by_identity }
end)
```

```
ruby 3.3.0dev (2023-12-18T08:21:42Z optimize-compare_b.. 025242017f) [arm64-darwin23]
1.174059   0.004965   1.179024 (  1.184111)

ruby 3.3.0dev (2023-12-18T17:26:37Z master 6c4b04de5c) [arm64-darwin23]
1.747141   0.006490   1.753631 (  1.764480)
```

hash.c
amomchilov: Is that a common convention? Seems a bit magical. Is it worth a comment?

Or perhaps being more explicit with `rb_obj_hide(rb_ident_hash_new_with_size(RHASH_SIZE(hash)))`?

peterzhu2118: Maybe we can have a `rb_hash_hidden_new` to mirror `rb_ary_hidden_new`.

PR #9275, opened 2023-12-18
Make the SHAPE_TOO_COMPLEX performance warning more actionable

casperisfine:
As suggested by @mame, we should try to help users fix the issues without having to lookup the meaning of the warning.

### DIFF
--- a/shape.c
+++ b/shape.c
@@ -733,9 +733,10 @@ rb_shape_get_next(rb_shape_t *shape, VALUE obj, ID id)
                 if (RCLASS_EXT(klass)->variation_count >= SHAPE_MAX_VARIATIONS) {
                     rb_category_warn(
                         RB_WARN_CATEGORY_PERFORMANCE,
-                        "Maximum shapes variations (%d) reached by %"PRIsVALUE", instance variables accesses will be slower.",
-                        SHAPE_MAX_VARIATIONS,
-                        rb_class_path(klass)
+                        "The class %"PRIsVALUE" reached %d shape variations, instance variables accesses will be slower and memory usage increased.\n"
+                        "It is recommended to define instance variable in a consistent order, for instance by eagerly defining them all in the `initialize` method.",
+                        rb_class_path(klass),
+                        SHAPE_MAX_VARIATIONS
                     );
                 }
             }
--- a/test/ruby/test_object.rb
+++ b/test/ruby/test_object.rb
@@ -458,7 +458,7 @@ class TestObject < Test::Unit::TestCase
   end
 
   def test_max_shape_variation_with_performance_warnings
-    assert_in_out_err([], <<-INPUT, %w(), /Maximum shapes variations \(8\) reached by Foo, instance variables accesses will be slower\.$/)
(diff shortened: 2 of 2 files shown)


shape.c
```diff
@@ -733,9 +733,10 @@ rb_shape_get_next(rb_shape_t *shape, VALUE obj, ID id)
                 if (RCLASS_EXT(klass)->variation_count >= SHAPE_MAX_VARIATIONS) {
                     rb_category_warn(
                         RB_WARN_CATEGORY_PERFORMANCE,
-                        "Maximum shapes variations (%d) reached by %"PRIsVALUE", instance variables accesses will be slower.",
-                        SHAPE_MAX_VARIATIONS,
-                        rb_class_path(klass)
+                        "The class %"PRIsVALUE" reached %d shape variations, instance variables accesses will be slower amd memory usage increased.\n"
```
mame, line 736:
```suggestion
                        "The class %"PRIsVALUE" reached %d shape variations, instance variables accesses will be slower and memory usage increased.\n"
```

test/ruby/test_object.rb
```diff
@@ -458,7 +458,7 @@ def test_check_to_integer
   end
 
   def test_max_shape_variation_with_performance_warnings
-    assert_in_out_err([], <<-INPUT, %w(), /Maximum shapes variations \(8\) reached by Foo, instance variables accesses will be slower\.$/)
+    assert_in_out_err([], <<-INPUT, %w(), /The class Foo reached 8 shape variations, instance variables accesses will be slower amd memory usage increased/)
```
mame, line 461:
```suggestion
    assert_in_out_err([], <<-INPUT, %w(), /The class Foo reached 8 shape variations, instance variables accesses will be slower and memory usage increased/)
```

mame (commented):
😇

shape.c
casperisfine: Thanks, for my defense I'm not done drinking my ☕  😉

mame (approved):

shape.c
```diff
@@ -733,9 +733,10 @@ rb_shape_get_next(rb_shape_t *shape, VALUE obj, ID id)
                 if (RCLASS_EXT(klass)->variation_count >= SHAPE_MAX_VARIATIONS) {
                     rb_category_warn(
                         RB_WARN_CATEGORY_PERFORMANCE,
-                        "Maximum shapes variations (%d) reached by %"PRIsVALUE", instance variables accesses will be slower.",
-                        SHAPE_MAX_VARIATIONS,
-                        rb_class_path(klass)
+                        "The class %"PRIsVALUE" reached %d shape variations, instance variables accesses will be slower and memory usage increased.\n"
+                        "It is recommended to define instance variable in a consistent order, for instance by eagerly defining them all in the `initialize` method.",
```
eregon, line 737: Minor, but backquotes seem rare in Ruby exception messages, here I think `#initialize` would be better.

byroot: https://github.com/ruby/ruby/pull/9280

PR #9276, opened 2023-12-18
Reduce usage of `object_id` in core gems

amomchilov:
Related: ruby/psych#663

Since Ruby 2.7, `object_id` became more expensive, particularly on its first call for a particular object. @JemmaIssroff has a great [blog post](https://jemma.dev/blog/gc-object-id) about it.

This PR aims to minimize its use by:

1. Replacing Hashed keyed with `.objetc_id` values with `Hash#compare_by_identity`
2. Replacing `object_id` comparisons with `equal?`


eval.c
```diff
@@ -1232,10 +1232,7 @@ ensure_class_or_module(VALUE obj)
 static VALUE
 hidden_identity_hash_new(void)
 {
-    VALUE hash = rb_ident_hash_new();
-
-    RBASIC_CLEAR_CLASS(hash); /* hide from ObjectSpace */
```
amomchilov, line 1237: Looking through uses of `rb_ident_hash_new`, I ran into this. Most cases use `rb_obj_hide`, but this uses `RBASIC_CLEAR_CLASS`. Is the difference intentional?

lib/ostruct.rb
```diff
@@ -386,17 +386,16 @@ def delete_field(name, &block)
   # Returns a string containing a detailed summary of the keys and values.
   #
   def inspect
-    ids = (Thread.current[InspectKey] ||= [])
-    if ids.include?(object_id)
+    ids = (Thread.current[InspectKey] ||= Set.new.compare_by_identity)
```
amomchilov, line 389: Using an identity Set here matches what `pp` does:

https://github.com/ruby/ruby/blob/e8064c6c2c317be78953b4d19226368580af0dca/lib/pp.rb#L146-L161

lib/reline/kill_ring.rb
```diff
@@ -14,7 +14,7 @@ def initialize(str)
     end
 
     def ==(other)
-      object_id == other.object_id
+      equal?(other)
     end
```
amomchilov, line 18: Alternatively, I could have deleted the `Struct#==` implementation with:

```suggestion
    remove_method(:==)
```

This would cause the  `Object#==` implementation to be inherited (which is identity-based), but it's much less clear, though.

lib/set.rb
```diff
@@ -361,16 +361,15 @@ def to_set(klass = Set, *args, &block)
     klass.new(self, *args, &block)
   end
 
-  def flatten_merge(set, seen = Set.new) # :nodoc:
+  def flatten_merge(set, seen = Set.new.compare_by_identity) # :nodoc:
     set.each { |e|
       if e.is_a?(Set)
-        if seen.include?(e_id = e.object_id)
+        unless seen.add?(e)
```
amomchilov, line 367: There's 3 nice performance wins here:

* We don't trigger a look up (and potentially an assignment) of an object ID
* We can do an identity-based look-up, which is faster
* We combine the `include?` and `add` steps into one, saving one hash lookup/comparison.

object.c
```diff
@@ -98,6 +98,15 @@ rb_obj_embedded_size(uint32_t numiv)
     return offsetof(struct RObject, as.ary) + (sizeof(VALUE) * numiv);
 }
 
+/*!
```
amomchilov, line 101: Unrelated, but this was unclear (what does "hide" mean?) and didn't have any docs, so I just added some.

eval.c
```diff
@@ -1232,10 +1232,7 @@ ensure_class_or_module(VALUE obj)
...
-    return hash;
+    rb_obj_hide(rb_ident_hash_new());
```
nobu, line 1235:
```suggestion
    return rb_obj_hide(rb_ident_hash_new());
```

nobu (changes requested):
The files under `lib` have each own upstreams.
Please send PR to the upstreams for such files.

eval.c
amomchilov: Good catch!

amomchilov:
@nobu ah interesting…
Is there any way to synchronize the changes to `set` and `ostruct`? The two share use of `Thread.current[InspectKey]`, whose usage pattern would need to change (because it’s type would change from an Array to and ident Set).

Would I need to write the patch in each of the two gems in a way that’s backwards-compatible with the other gem being in the older version?

eval.c
amomchilov: @nobu Do you have any context on this?

amomchilov:
Oh cool, I managed to cause a segfault :D

Any idea why that might be happening?

eval.c
nobu: This line predates commit:adf1c94ffe75, where `rb_obj_hide` was added.
Just added to fix https://bugs.ruby-lang.org/issues/8204, but didn't change already existing code all.
Maybe for unnecessary `SPECIAL_CONST_P` check.
It would be OK to replace this line with `rb_obj_hide`, I think.

object.c
nobu: Hide the object from ObjectSpace, as written in `hidden_identity_hash_new` comment.

nobu:
> * Interestingly, https://github.com/ruby/monitor is archived. Should this gem be removed from [`./ext/monitor`](https://github.com/ruby/ruby/tree/master/ext/monitor) in this repo?

It was moved there once, but partially rewritten as an extension library at #2576.
So that repo is outdated.

object.c
```diff
@@ -98,6 +98,15 @@ rb_obj_embedded_size(uint32_t numiv)
...
+ * :nodoc:
```
nobu, line 102: Since `rb_obj_hide` is not a method implementation, RDoc does not read this comment and this directive does nothing.

object.c
amomchilov: Heh yeah I got it, but I didn’t in first read, which is why I added the doc comment for future readers

amomchilov:
Hey @nobu, I incorporated the changes you mentioned.

Could you please merge what's left of this PR, and review the smaller PRs I opened on the upstream gems?

amomchilov:
Rewrote those rebased commits, so they're signed correctly and no longer "unverified".

PR #9277, opened 2023-12-18
Update `BasicSocket#recv` documentation on return value

casperisfine:
Ref: https://github.com/ruby/ruby/pull/6407

[[Bug #19012]](https://bugs.ruby-lang.org/issues/19012)

`0` is now interpreted as closed connection an not an empty packet, as these are very rare and pretty much useless.


zverok (approved):
Thanks!

PR #9282, opened 2023-12-18
[PRISM] Fix crash when --parser=prism called with stdin

eightbitraptor:
Currently Ruby crashes when the --parser=prism flag is used either with no input, or with input that is being redirected from stdin. So all of the following will crash

```
ruby --parser=prism
ruby --parser=prism < test_code.rb
cat test_code.rb | ruby --parser=prism
```

This commit checks whether the input is assumed to be from stdin, and then processes that as a file.

This will fix the second and third case above, but will cause a slight behavioural changes for the first case - Ruby will treat stdin as an empty file in this case and exit, rather than waiting for data to be piped into stdin.

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -2344,7 +2344,16 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
         pm_string_t input;
         pm_options_t options = { 0 };
 
-        if (opt->e_script) {
+        if (strcmp(opt->script, "-") == 0) {
+            int xflag = opt->xflag;
+            VALUE rb_source = open_load_file(opt->script_name, &xflag);
+            opt->xflag = xflag != 0;
+
+            rb_warn("Prism support for streaming code from stdin is not currently supported");
+            pm_string_constant_init(&input, RSTRING_PTR(rb_source), RSTRING_LEN(rb_source));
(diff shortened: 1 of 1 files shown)


HParker:
Thanks for finding this! The fix looks good to me. We could maybe add a test in `test/ruby/test_rubyoptions.rb`, but I am not sure if that is necessary.

Are you planning to open a PR to fix the first case as well? I can look, but I don't believe I have any special knowledge here beyond what you have in this PR.

tenderlove (approved):
I think this is a good stopgap for now.

eightbitraptor:
> Thanks for finding this! The fix looks good to me. We could maybe add a test in `test/ruby/test_rubyoptions.rb`, but I am not sure if that is necessary.
> 
> Are you planning to open a PR to fix the first case as well? I can look, but I don't believe I have any special knowledge here beyond what you have in this PR.

Thanks! I'll take a look at adding a test to this and then get it shipped

As far as a proper fix - I haven't looked into it in major detail yet, but I think that Prism needs to handle this case rather than dealing with it in the options processing code. And I think [this issue will need addressed first.](https://github.com/ruby/prism/issues/1247)

PR #9284, opened 2023-12-18
Revert "[Bug #19831] Remove duplicate library options"

misdoro:
This reverts commit 5bb946228550c7f171c27725860b153a675404f3.

Fixes https://bugs.ruby-lang.org/issues/20072

Building ruby with `--enable-shared --with-jemalloc` resulted in crash when requiring `sassc` gem:

```
$ irb
irb(main):001> require 'sassc'
free(): invalid pointer
Aborted
```

This is fixed by reverting the original change introduced in #8233


misdoro:
:thinking: Might be worth adding a test to `ubuntu` workflow enabling both `shared` and `jemalloc`.

PR #9285, opened 2023-12-18
Unmask signal before re-raising fatal signal

XrXr:
On Linux, while the signal handler runs, that signal is masked, so we
translated fatal signals to SIGABRT unintentionally. See signal(7).
It worked fine on macOS, though.

Before:

    $ ./miniruby -e 'Process.kill :SIGSEGV, Process.pid'
    <snip>
    Aborted (core dumped)

After:

    $ ./miniruby -e 'Process.kill :SIGSEGV, Process.pid'
    <snip>
    Segmentation fault (core dumped)

Follow-up for 1ac0afab4da "rb_bug_for_fatal_signal: exit with the right
signal".

cc: @byroot

### DIFF
--- a/signal.c
+++ b/signal.c
@@ -403,6 +403,9 @@ interrupt_init(int argc, VALUE *argv, VALUE self)
 }
 
 void rb_malloc_info_show_results(void); /* gc.c */
+#ifdef POSIX_SIGNAL
+static void reset_sigmask(int sig);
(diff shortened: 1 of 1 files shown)


byroot (approved):
❤️ I tested it on macOS, didn't expect it to work differently on Linux. Nice to see I'm not crazy

PR #9288, opened 2023-12-18
We need to load builtins so that they work

tenderlove:
Before this commit no methods defined in Ruby were being loaded. For example `class` or `tap` methods would not exist.

I think this plugs a leak in the current implementation but we should probably revisit how Prism is integrated in to the Ruby boot process.

cc @eightbitraptor


eightbitraptor (approved):
Thanks @tenderlove 

This fixes https://bugs.ruby-lang.org/issues/20073

PR #9289, opened 2023-12-19
[WIP] Implement Class#new in Ruby

tenderlove:
This commit adds a new primitive function `Primitive.send_delegate!` which emits a send instruction based on the first two parameters. It forces the send instruction to be an FCALL, thereby passing visibility checks.

This allows us to write `Class#new` in Ruby and have it work for both BasicObject and regular objects.

Here are the instructions for `Class#new`:

```
$ ./miniruby -e'puts RubyVM::InstructionSequence.of(Class.method(:new)).disasm'
== disasm: #<ISeq:new@<internal:class>:2 (2,2)-(6,5)>
local table (size: 4, argc: 0 [opts: 0, rest: 0, post: 0, block: 1, kw: -1@-1, kwrest: -1])
[ 4] "*"@0<Rest>[ 3] "&"@1<Block>[ 2] "..."@2    [ 1] obj@3
0000 opt_invokebuiltin_delegate             <builtin!rb_class_alloc2/0>, 0(   3)[LiCa]
0003 setlocal_WC_0                          obj@3
0005 getlocal_WC_0                          obj@3                     (   4)[Li]
0007 getlocal_WC_0                          "*"@0
0009 splatarray                             false
0011 getblockparamproxy                     "&"@1, 0
0014 send                                   <calldata!mid:initialize, argc:1, ARGS_SPLAT|ARGS_BLOCKARG|FCALL>, nil
0017 pop
0018 getlocal_WC_0                          obj@3                     (   5)[Li]
0020 leave                                                            (   6)[Re]
```


class.rb
```diff
@@ -0,0 +1,7 @@
+class Class
+  def new(...)
+    obj = Primitive.rb_class_alloc2
+    Primitive.send_delegate!(obj, :initialize, ...)
```
eregon, line 4: Does this mean there will be a single inline cache, so essentially a 1-entry global cache for calls to `initialize`?
Also I'm not sure but it might be updated on every call with a different receiver class, which would be expensive.
I guess currently the call to `initialize` from `new` is completely uncached (`rb_funcall`), so maybe it's about the same in this PR for that aspect?

eregon: I realize on YJIT this inline cache concern might matter less because it creates its own inline cache by generating for the receiver it sees. Although that inline cache would only be per caller-of-new if YJIT always inlines `new`, otherwise it would still be global.

tenderlove: > I guess currently the call to `initialize` from `new` is completely uncached (`rb_funcall`), so maybe it's about the same in this PR for that aspect?

The C calls use the global call cache.  I did [some refactoring on the global call cache](https://github.com/ruby/ruby/pull/10623), so we can use it for the initialize case.  This should make it hit cache as well as the C version did. See bb5fdce4d88d16472f7071e9b8156a12ff9eeab0

eregon: Ah right, there is still the global cache, that's at least (in the hit case) one (simplified) hashmap lookup vs multiple.
FWIW, I looked, `Class#new` uses `rb_obj_call_init_kw()` which uses `rb_funcallv_kw()` and that doesn't have any inline cache.

PR #9299, opened 2023-12-19
Add tool/missing-baseruby.bat, used when BASERUBY not available

jeremyevans:
Previously, the embedded semicolon in BASERUBY if BASERUBY is not available breaks tarball builds without BASERUBY when using OpenBSD make, due to the inability to escape MFLAGS correctly. This moves the same BASERUBY code into a separate file, avoiding the MFLAGS quoting issue.

BASERUBY must be passed to build-ext because it is required by ripper since the introduction of lrama.

Fixes [Bug #19683]

### DIFF
--- a/tool/missing-baseruby.bat
+++ b/tool/missing-baseruby.bat
@@ -1,0 +1,3 @@
+#!/bin/sh
+echo "executable host ruby is required.  use --with-baseruby option."
+exit 1
--- a/tool/missing-baseruby.bat
+++ b/tool/missing-baseruby.bat
@@ -1,3 +1,5 @@
-#!/bin/sh
+: "
+@echo off
+: "
 echo "executable host ruby is required.  use --with-baseruby option."
 exit 1
--- a/tool/missing-baseruby.bat
+++ b/tool/missing-baseruby.bat
@@ -1,5 +1,5 @@
 : "
 @echo off
 : "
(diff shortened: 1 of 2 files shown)


tool/missing-baseruby.bat
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
nobu, line 1: Do you intend to use this on Windows too?
But `#!` is not a valid command on Windows.
Usually shebang for `/bin/sh` is not necessary. 

Instead, `@echo off` will be better.
```suggestion
: "
@echo off
: "
```

The `:` is an invalid label to be ignored up to end of line on `cmd.exe`, and the command nothing to do on `sh` and
the 2nd and 3rd lines are its unused argument.

tool/missing-baseruby.bat
```diff
@@ -0,0 +1,3 @@
...
+echo "executable host ruby is required.  use --with-baseruby option."
```
nobu, line 2: One more for `cmd.exe`, the command will be passed `echo` as written including quotes, and `echo` prints the quotes too.

```suggestion
echo executable host ruby is required.  use --with-baseruby option.
```

nobu (approved):

mame:
What a useful polyglot! 👏

PR #9306, opened 2023-12-20
coroutine/arm64/Context.S: Append PAC/BTI note section if needed

kateinoigakukun:
Fixes https://bugs.ruby-lang.org/issues/20029

### DIFF
--- a/coroutine/arm64/Context.h
+++ b/coroutine/arm64/Context.h
@@ -17,7 +17,7 @@
 
 #define COROUTINE __attribute__((noreturn)) void
 
-enum {COROUTINE_REGISTERS = 0xb0 / 8};
+enum {COROUTINE_REGISTERS = 0xa0 / 8};
 
 #if defined(__SANITIZE_ADDRESS__)
     #define COROUTINE_SANITIZE_ADDRESS
@@ -89,7 +89,7 @@ static inline void coroutine_initialize(
     context->stack_pointer -= COROUTINE_REGISTERS;
     memset(context->stack_pointer, 0, sizeof(void*) * COROUTINE_REGISTERS);
 
-    context->stack_pointer[0xa0 / 8] = ptrauth_sign_instruction_addr((void*)start, (void*)top);
+    context->stack_pointer[0x98 / 8] = ptrauth_sign_instruction_addr((void*)start, (void*)top);
 }
 
 struct coroutine_context * coroutine_transfer(struct coroutine_context * current, struct coroutine_context * target);
--- a/coroutine/arm64/Context.h
+++ b/coroutine/arm64/Context.h
@@ -50,6 +50,23 @@ static inline void coroutine_initialize_main(struct coroutine_context * context)
     context->stack_pointer = NULL;
 }
 
+#if defined(__ARM_FEATURE_PAC_DEFAULT) && __ARM_FEATURE_PAC_DEFAULT != 0
+// Sign the given instruction address with the given modifier and key A
+static inline void *ptrauth_sign_instruction_addr(void *addr, void *modifier) {
+    register void *r17 __asm("r17") = addr;
+    register void *r16 __asm("r16") = modifier;
+    // Use HINT mnemonic instead of PACIA1716 for compatibility with older assemblers.
+    asm ("hint #8;" : "+r"(r17) : "r"(r16));
+    addr = r17;
+    return addr;
+}
+#else
+// No-op if PAC is not enabled
+static inline void *ptrauth_sign_instruction_addr(void *addr, void *modifier) {
+    return addr;
+}
+#endif
+
 static inline void coroutine_initialize(
     struct coroutine_context *context,
     coroutine_start start,
@@ -66,12 +83,13 @@ static inline void coroutine_initialize(
 
     // Stack grows down. Force 16-byte alignment.
     char * top = (char*)stack + size;
-    context->stack_pointer = (void**)((uintptr_t)top & ~0xF);
+    top = (char *)((uintptr_t)top & ~0xF);
+    context->stack_pointer = (void**)top;
 
     context->stack_pointer -= COROUTINE_REGISTERS;
     memset(context->stack_pointer, 0, sizeof(void*) * COROUTINE_REGISTERS);
 
-    context->stack_pointer[0xa0 / 8] = (void*)start;
+    context->stack_pointer[0xa0 / 8] = ptrauth_sign_instruction_addr((void*)start, (void*)top);
 }
 
 struct coroutine_context * coroutine_transfer(struct coroutine_context * current, struct coroutine_context * target);
--- a/coroutine/arm64/Context.S
+++ b/coroutine/arm64/Context.S
@@ -91,3 +91,37 @@ PREFIXED_SYMBOL(SYMBOL_PREFIX,coroutine_transfer):
 #if defined(__linux__) && defined(__ELF__)
 .section .note.GNU-stack,"",%progbits
 #endif
+
+#if __ARM_FEATURE_BTI_DEFAULT != 0 || __ARM_FEATURE_PAC_DEFAULT != 0
+/*  See "ELF for the Arm 64-bit Architecture (AArch64)"
+    https://github.com/ARM-software/abi-aa/blob/2023Q3/aaelf64/aaelf64.rst#program-property */
+#  define GNU_PROPERTY_AARCH64_FEATURE_1_BTI (1<<0)
+#  define GNU_PROPERTY_AARCH64_FEATURE_1_PAC (1<<1)
+
+#  if __ARM_FEATURE_BTI_DEFAULT != 0
+#    define BTI_FLAG GNU_PROPERTY_AARCH64_FEATURE_1_BTI
+#  else
+#    define BTI_FLAG 0
+#  endif
+#  if __ARM_FEATURE_PAC_DEFAULT != 0
+#    define PAC_FLAG GNU_PROPERTY_AARCH64_FEATURE_1_PAC
+#  else
+#    define PAC_FLAG 0
+#  endif
+
+  # The note section format is described by Note Section in Chapter 5
+  # of "System V Application Binary Interface, Edition 4.1".
+  .pushsection .note.gnu.property, "a"
+  .p2align 3
+  .long 0x4        /* Name size ("GNU\0") */
+  .long 0x10       /* Descriptor size */
+  .long 0x5        /* Type: NT_GNU_PROPERTY_TYPE_0 */
+  .asciz "GNU"     /* Name */
+  # Begin descriptor
+  .long 0xc0000000 /* Property type: GNU_PROPERTY_AARCH64_FEATURE_1_AND */
+  .long 0x4        /* Property size */
+  .long (BTI_FLAG|PAC_FLAG)
+  .long 0x0        /* 8-byte alignment padding */
+  # End descriptor
+  .popsection
+#endif
--- a/coroutine/arm64/Context.S
+++ b/coroutine/arm64/Context.S
@@ -28,8 +28,12 @@
 PREFIXED_SYMBOL(SYMBOL_PREFIX,coroutine_transfer):
 
 #if defined(__ARM_FEATURE_PAC_DEFAULT) && (__ARM_FEATURE_PAC_DEFAULT != 0)
-	# paciasp
+	# paciasp (it also acts as BTI landing pad, so no need to insert BTI also)
 	hint #25
+#elif defined(__ARM_FEATURE_BTI_DEFAULT) && (__ARM_FEATURE_BTI_DEFAULT != 0)
+	# For the the case PAC is not enabled but BTI is.
+	# bti c
+	hint #34
 #endif
 	# Make space on the stack for caller registers
 	sub sp, sp, 0xb0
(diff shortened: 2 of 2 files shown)


ioquatix:
My understanding is that this is insufficient.

What you are asking for, is for a shadow stack for call/return/jump instructions, but I don't think this metadata solves that problem. IIUC, this metadata change indicates that the code implements BTI/PAC without actually implementing it in the coroutine. Feel free to correct me if I'm wrong.

kateinoigakukun:
Ah, good point. I thought that this is a leaf function and we don't need to care about the protection stuff. But on second thought, the function reads the return address from the shadow stack, so we still need to sign the pointer. Also apparently we also need BTI landing pad. Will push some followup fixes.

kateinoigakukun:
@ioquatix Just curious: why x30 (LR) is saved twice in (`str x30, [sp, 0xa0]` and `stp x29, x30, [sp, 0x90]`)? IIUC we can skip the second store/load and use `ret` instead of `ret x17`.

If we stick to use the return address stored in `[sp, 0xa0]`, we need some extra instructions like:

```diff
-       # Load return address into x4
-       ldr x4, [sp, 0xa0]
+       # Load return address into x17
+       ldr x17, [sp, 0xa0]

        # Pop stack frame
        add sp, sp, 0xb0
+       mov x16, sp
+
+       pacia1716

-       # Jump to return address (in x4)
-       ret x4
+       # Jump to return address (in x17)
+       ret x17
```


If we can just use `[sp, 0x98]` return register loaded to x30, we can simplify it much more:

```diff
        # Load return address into x4
-       ldr x4, [sp, 0xa0]

        # Pop stack frame
        add sp, sp, 0xb0

+       autiasp

-       # Jump to return address (in x4)
-       ret x4
+       # Jump to return address (in x30)
+       ret
```

Also for the case if the target doesn't need ISA backward compatibility, we can combine `autiasp` and `ret` into `retaa`.

ggardet:
I tested current patch, but unfortunately, the problem persists.

kateinoigakukun:
@ggardet Which configure option did you use? I tested `./configure cppflags=-mbranch-protection=standard cflags=-mbranch-protection=standard ASFLAGS=-mbranch-protection=standard LDFLAGS="-mbranch-protection=standard"` on 
aarch64 openSUSE Tumbleweed with GCC 13. 

---

Side note: The configure command I proposed uses `cppflags` to append the option. It works but it's semantically wrong. This is because `configure` automatically append `-mbranch-protection=pac-ret` to XCFLAGS after user-provided CFLAGS. Ideally we can change `configure` to respect user-provided CFLAGS by pretending auto-guessed options before them. But we have a new version release soon and we have a workaround by `cppflags`, so I don't want to change build stuff at this moment.

ggardet:
Here is my `configure` command:
```
[   12s] + export 'CFLAGS=-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fno-strict-aliasing -std=gnu99'
[   12s] + CFLAGS='-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fno-strict-aliasing -std=gnu99'
[   12s] + export 'LDFLAGS= -z force-bti '
[   12s] + LDFLAGS=' -z force-bti '
[   12s] + CONFIG_SHELL=/usr/bin/bash
[   12s] + export CONFIG_SHELL
[   12s] + CFLAGS='-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fno-strict-aliasing -std=gnu99'
[   12s] + export CFLAGS
[   12s] + CXXFLAGS='-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g'
[   12s] + export CXXFLAGS
[   12s] + FFLAGS='-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g '
[   12s] + export FFLAGS
[   12s] + FCFLAGS='-mbranch-protection=standard -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g '
[   12s] + export FCFLAGS
[   12s] + LDFLAGS=' -z force-bti '
[   12s] + export LDFLAGS
[   12s] + ./configure --host=aarch64-suse-linux-gnu --build=aarch64-suse-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --with-baseruby=/usr/bin/ruby --enable-pie --with-compress-debug-sections=none --program-suffix=.ruby3.2 --with-soname=ruby3.2 --target=aarch64-suse-linux --with-valgrind --with-mantype=man --enable-shared --disable-static --disable-rpath
```
I added `-z force-bti` to `LDFLAGS` to get the BTI warning and this should not be used for production/release.
And yes, this is on aarch64 openSUSE Tumbleweed with GCC 13.

kateinoigakukun:
@ggardet 
> EDIT: -mbranch-protection=pac-ret added by configure script may overwrite -mbranch-protection=standard which would enable only PAC and not PAC+BTI.

Right, that's why I added `cppflags=-mbranch-protection=standard`

kateinoigakukun:
@ggardet Also please note that I added `ASFLAGS=-mbranch-protection=standard` too

ggardet:
> @ggardet
> 
> > EDIT: -mbranch-protection=pac-ret added by configure script may overwrite -mbranch-protection=standard which would enable only PAC and not PAC+BTI.
> 
> Right, that's why I added `cppflags=-mbranch-protection=standard`


> @ggardet Also please note that I added `ASFLAGS=-mbranch-protection=standard` too

Ok, that works now.
I still have few warnings, such as:
```
[  153s] gcc -I. -I.ext/include/aarch64-linux-gnu -I./include -I.   -DONIG_ENC_REGISTER=rb_enc_register -fPIC -O3 -fno-fast-math -ggdb3 -Wall -Wextra -Wdeprecated-declarations -Wdiv-by-zero -Wduplicated-cond -Wimplicit-function-declaration -Wimplicit-int -Wmisleading-indentation -Wpointer-arith -Wwrite-strings -Wold-style-definition -Wimplicit-fallthrough=0 -Wmissing-noreturn -Wno-cast-function-type -Wno-constant-logical-operand -Wno-long-long -Wno-missing-field-initializers -Wno-overlength-strings -Wno-packed-bitfield-compat -Wno-parentheses-equality -Wno-self-assign -Wno-tautological-compare -Wno-unused-parameter -Wno-unused-value -Wsuggest-attribute=format -Wsuggest-attribute=noreturn -Wunused-variable -Wundef -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fno-strict-aliasing -std=gnu99 -fPIC  -o enc/trans/iso2022.o -c ./enc/trans/iso2022.c
[  153s] gcc -shared -o .ext/aarch64-linux-gnu/enc/encdb.so enc/encdb.o -L. -L. -L.  -mbranch-protection=standard -z force-bti  -fstack-protector-strong -rdynamic -Wl,-export-dynamic -Wl,--no-as-needed -mbranch-protection=standard -z force-bti  -Wl,--compress-debug-sections=none    -lruby3.2 -lm -lpthread  
[  153s] gcc -I. -I.ext/include/aarch64-linux-gnu -I./include -I.   -DONIG_ENC_REGISTER=rb_enc_register -fPIC -O3 -fno-fast-math -ggdb3 -Wall -Wextra -Wdeprecated-declarations -Wdiv-by-zero -Wduplicated-cond -Wimplicit-function-declaration -Wimplicit-int -Wmisleading-indentation -Wpointer-arith -Wwrite-strings -Wold-style-definition -Wimplicit-fallthrough=0 -Wmissing-noreturn -Wno-cast-function-type -Wno-constant-logical-operand -Wno-long-long -Wno-missing-field-initializers -Wno-overlength-strings -Wno-packed-bitfield-compat -Wno-parentheses-equality -Wno-self-assign -Wno-tautological-compare -Wno-unused-parameter -Wno-unused-value -Wsuggest-attribute=format -Wsuggest-attribute=noreturn -Wunused-variable -Wundef -O2 -Wall -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=3 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -g -fno-strict-aliasing -std=gnu99 -fPIC  -o enc/trans/japanese.o -c ./enc/trans/japanese.c
[  153s] /usr/lib64/gcc/aarch64-suse-linux/13/../../../../aarch64-suse-linux/bin/ld: /tmp/ccM1RRpJ.ltrans0.ltrans.o: warning: BTI turned on by -z force-bti when all inputs do not have BTI in NOTE section.
[  153s] /usr/lib64/gcc/aarch64-suse-linux/13/../../../../aarch64-suse-linux/bin/ld: /tmp/ccjJtYJo.debug.temp.o: warning: BTI turned on by -z force-bti when all inputs do not have BTI in NOTE section.
```
It looks like some `gcc` command lines are missing the `-mbranch-protection=standard`.

kateinoigakukun:
@ggardet I guess you are setting `CFLAGS` to append `-mbranch-protection=standard` but I use `cflags` (lower case version). Could you try with this again?

coroutine/arm64/Context.h
```diff
@@ -66,12 +78,17 @@ static inline void coroutine_initialize(
...
     // Stack grows down. Force 16-byte alignment.
     char * top = (char*)stack + size;
-    context->stack_pointer = (void**)((uintptr_t)top & ~0xF);
+    top = (char *)((uintptr_t)top & ~0xF);
+    context->stack_pointer = (void**)top;
 
     context->stack_pointer -= COROUTINE_REGISTERS;
     memset(context->stack_pointer, 0, sizeof(void*) * COROUTINE_REGISTERS);
 
+#if defined(__ARM_FEATURE_PAC_DEFAULT) && __ARM_FEATURE_PAC_DEFAULT != 0
```
ioquatix, line 87: What do you think about avoiding `#if` here and making `ptrauth_sign_instruction_addr` a no-op on platforms where it's not enabled? That way we can avoid increasing cyclomatic complexity.

ioquatix:
I read the documentation, thanks for the links. I was under the impression we'd need to implement a shadow stack like how it works with CET. However, it seems like this is sufficient. I'm a little cautious about merging this so soon before 3.3.0 - what are your thoughts on trying to get in before the next release?

ioquatix:
> Just curious: why x30 (LR) is saved twice in (str x30, [sp, 0xa0] and stp x29, x30, [sp, 0x90])? IIUC we can skip the second store/load and use ret instead of ret x4.

Probably because I was not an expert at the assembler and just did my best to get something that looked good enough. Feel free to improve it as you see fit :)

coroutine/arm64/Context.h
kateinoigakukun: Sounds reasonable to me 👍

kateinoigakukun:
@ioquatix 
> I'm a little cautious about merging this so soon before 3.3.0 - what are your thoughts on trying to get in before the next release?

I think it's relatively low-risk to get this in 3.3.0 release, let's merge soon after some fixup commits.

> Probably because I was not an expert at the assembler and just did my best to get something that looked good enough. Feel free to improve it as you see fit :)

Okay, thank you for clarifying!

kateinoigakukun:
The RJIT workflow failure is unrelated to this change, it's also failing on master. https://github.com/ruby/ruby/actions/runs/7294360546/job/19879066887

ioquatix (approved):
LGTM

kateinoigakukun:
Thanks!

ggardet:
> @ggardet I guess you are setting `CFLAGS` to append `-mbranch-protection=standard` but I use `cflags` (lower case version). Could you try with this again?

Ok, that works now. 
But why `cflags` (lower case version) instead of being consistent across the code and use `CFLAGS` (upper case) everywhere?

kateinoigakukun:
`CFLAGS` and `cflags` are different variables having different semantics and both are used in Ruby. `CFLAGS` can be configured by user and configure.ac appends guessed flags. `cflags` is appended after `CFLAGS`.

PR #9308, opened 2023-12-20
[DOC] Improve NEWS.md rendering and add missing features

zverok:
* Fix rendering of the file on [docs.ruby-lang.org](https://docs.ruby-lang.org/en/master/NEWS_md.html):
  * More indentation for code;
  * Properly qualified methods to link to docs (`Queue#freeze` → `Thread::Queue#freeze`, remove unnecessary backticks, etc.)
* Add two minor yet notable changes:
  * `Time.new` stricter parsing of strings;
  * `NoMethodError` rendering logic change.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -36,26 +36,27 @@ Note: We're only listing outstanding class updates.
 
 * Encoding
 
-    * Encoding#replicate has been removed, it was already deprecated.  [[Feature #18949]]
+    * `Encoding#replicate` has been removed, it was already deprecated.  [[Feature #18949]]
 
 * Fiber
 
     * Introduce Fiber#kill. [[Bug #595]]
-      ```ruby
-      fiber = Fiber.new do
-        while true
-          puts "Yielding..."
-          Fiber.yield
-        end
-      ensure
-        puts "Exiting..."
-      end
 
-      fiber.resume
-      # Yielding...
-      fiber.kill
-      # Exiting...
-      ```
+        ```ruby
+        fiber = Fiber.new do
+          while true
+            puts "Yielding..."
+            Fiber.yield
+          end
+        ensure
+          puts "Exiting..."
+        end
+
+        fiber.resume
+        # Yielding...
+        fiber.kill
+        # Exiting...
+    ```
 
 * MatchData
 
@@ -67,6 +68,16 @@ Note: We're only listing outstanding class updates.
     * Module#set_temporary_name added for setting a temporary name for a
       module. [[Feature #19521]]
 
+* NoMethodError
+
+    * Error message have changed to not use the target object's `#inspect`
+      for efficiency, and says "instance of ClassName" instead. [[Feature #18285]]
+
+        ```ruby
+        ([1] * 100).nonexisting
+        # undefined method `nonexisting' for an instance of Array (NoMethodError)
+        ```
+
 * ObjectSpace::WeakKeyMap
 
     * New core class to build collections with weak references.
@@ -75,7 +86,7 @@ Note: We're only listing outstanding class updates.
 
 * ObjectSpace::WeakMap
 
-    * `ObjectSpace::WeakMap#delete` was added to eagerly clear weak map
+    * ObjectSpace::WeakMap#delete was added to eagerly clear weak map
       entries. [[Feature #19561]]
 
 * Proc
@@ -84,7 +95,7 @@ Note: We're only listing outstanding class updates.
 
 * Process
 
-    * New `Process.warmup` method that notify the Ruby virtual machine that the boot sequence is finished,
+    * New Process.warmup method that notify the Ruby virtual machine that the boot sequence is finished,
       and that now is a good time to optimize the application. This is useful
       for long-running applications. The actual optimizations performed are entirely
       implementation-specific and may change in the future without notice. [[Feature #18885]]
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -67,6 +68,16 @@ Note: We're only listing outstanding class updates.
     * Module#set_temporary_name added for setting a temporary name for a
       module. [[Feature #19521]]
 
+* NoMethodError
```
jeremyevans, line 71: I don't think this change is worth mentioning in NEWS.

NEWS.md
```diff
@@ -488,3 +510,5 @@ changelog for details of the default gems or bundled gems.
 [Feature #19965]: https://bugs.ruby-lang.org/issues/19965
 [Feature #20005]: https://bugs.ruby-lang.org/issues/20005
 [Feature #20057]: https://bugs.ruby-lang.org/issues/20057
+[Bug #19293]: https://bugs.ruby-lang.org/issues/19293
+[Feature #18285]: https://bugs.ruby-lang.org/issues/18285
```
jeremyevans, line 514: When removing the NoMethodError change, we would want to remove this line as well.

jeremyevans (changes requested):
Looks mostly good.  Only a couple changes requested.

NEWS.md
nobu: It's better to sort by URLs.

NEWS.md
```diff
@@ -75,7 +86,7 @@ Note: We're only listing outstanding class updates.
 
 * ObjectSpace::WeakMap
 
-    * `ObjectSpace::WeakMap#delete` was added to eagerly clear weak map
```
ko1, line 78: why do you remove backquote even though it is `.md`?

NEWS.md
mame: I think it is better to explain this in "Compatibility issues" section than in "Core classes updates".

NEWS.md
nobu: It will be marked up as `<code>` without backquotes.

NEWS.md
zverok: @jeremyevans My thinking here is this:
1. It is a part of public behavior
2. It is a behavior that is frequently encountered (i.e., almost every Rubyist sees this error almost every day, while some other notable changes like `WeakKeyMap` or `Fiber#kill` are something that wouldn't be used by as many people)
3. The change affects the quality of life. I understand the justification, yet the change might make somebody's debugging harder (by not seeing the real problematic object) and somebody's easier (by removing 20-lines `#inspect`s of complicated objects)
4. The change also might affect application/library code (this is purely theoretical, but not unimaginable that some code caught the error and extracted the `#inspect` of the affected object; or that somebody's `#inspect` was defined in the way that considered frequent rendering in this message)
5. The entry in `NEWS` not only explains the change but links to the corresponding discussion, so when one sees in the new version, "wait, is this changed?" they can a) find the entry in NEWS (and make sure the change was conscious, not some side-effect or bug) and b) find the justification.

NEWS.md
zverok: @ko1 Without backticks RDoc is able to linkify the method name. 
The current version:
image
After the fix:
image

ko1: I see.

----

For me, it is better to allow backquotes and rdoc also accepts backquoted method names, because it is easy to copy & paste to other articles.

nobu: You mean the case of "only method name enclosed in backquotes", right?
E.g., `` `ObjectSpace::WeakMap#delete` ``, but not `` `(ObjectSpace::WeakMap#delete)` ``.
It would be better but should be requested to ruby/rdoc, not here.

NEWS.md
jeremyevans: @zverok I read what you wrote and still do not think this is worth mentioning.  There is a tradeoff in mentioning anything in NEWS.  We want NEWS to be short enough that people will read it, hitting the important parts of the release.  NEWS is not designed to be comprehensive and discuss every single change since the last release. If you think this is a compatibility issue (I do not), take @mame's advice and put it in the compatibility issues section. Otherwise, I think it should be removed.

NEWS.md
```diff
@@ -109,16 +106,31 @@ Note: We're only listing outstanding class updates.
...
+    * Thread::SizedQueue#freeze now raises TypeError. [[Bug #17146]]
+
+* Time
+
+    * Time.new with a string argument became stricter. [[Bug #19293]]
+
+        ```ruby
+        Time.new('2023-12-20')
+        #  no time information (ArgumentError)
+        Time.new('2023-12-20')
```
jeremyevans, line 130: This uses the same code twice, which doesn't appear intentional.  I would remove the duplicate lines.

jeremyevans (changes requested):
One additional change requested.  After that, OK to merge without further review.

NEWS.md
zverok: Ugh. Meant to show "year-month" and "year-month-day", it is a typo. But anyway, I think one example is enough.

jeremyevans (approved):
I did mention in my previous review that this could be merged without further review after the change I requested was made.

Please squash when merging.

NEWS.md
nurse: I'm for because it is a part of public behavior.
Generally speaking NEWS.md is allowed to write about something if a committer thinks it should be.
I think what jeremy says  is rather a release note.

PR #9311, opened 2023-12-20
Use main_thread->ec from rb_vm_main_ractor_ec

jhawthorn:
`rb_vm_main_ractor_ec` was introduced to allow `rb_postponed_job_*` to work when fired on non-Ruby threads, which have no EC set, and that is its only use.

When `RUBY_MN_THREADS=1` is set `ractor->threads.running_ec` is `NULL` when the shared thread is sleeping. This instead grabs the EC directly from the main thread which seems to always be set.

With this change StackProf works with `RUBY_MN_THREADS=1` set (this should fix https://github.com/tmm1/stackprof/issues/221)

Fixes [[Bug #20016]](https://bugs.ruby-lang.org/issues/20016)

cc @byroot @ko1 @KJTsanaktsidis @tenderlove

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -305,7 +305,7 @@ changelog for details of the default gems or bundled gems.
     * added: `rb_postponed_job_preregister()`
     * added: `rb_postponed_job_trigger()`
     * deprecated: `rb_postponed_job_register()` (and semantic change. see below)
-    * deprecated: `rb_postponed_job_register_once()`
+    * deprecated: `rb_postponed_job_register_one()`
   * The postponed job APIs have been changed to address some rare crashes.
     To solve the issue, we introduced new two APIs and deprecated current APIs.
     The semantics of these functions have also changed slightly; `rb_postponed_job_register`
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -305,7 +305,7 @@ changelog for details of the default gems or bundled gems.
     * added: `rb_postponed_job_preregister()`
     * added: `rb_postponed_job_trigger()`
     * deprecated: `rb_postponed_job_register()` (and semantic change. see below)
-    * deprecated: `rb_postponed_job_register_once()`
```
KJTsanaktsidis, line 308: Eurgh, sorry about this and thank you for fixing it

KJTsanaktsidis (approved):
This looks correct to me I think - `main_thread` is set in exactly two places (VM bootstrap and after forking) so it should always be available for use in postponed job execution.

NEWS.md
jhawthorn: Not at all! The `preregister` improvement looks great.

ko1 (approved):

PR #9314, opened 2023-12-22
Avoid underflow of rb_yjit_live_iseq_count

jhawthorn:
This value is only incremented when `rb_iseq_translate_threaded_code` is called, which doesn't happen for iseqs which result in a syntax error.

This is easy to hit by running a debug build with `RUBY_FREE_AT_EXIT=1`, but any build and options could underflow this value by running enough evals.

cc @ruby/yjit


k0kubun (approved):
Thanks!

PR #9319, opened 2023-12-22
.travis.yml: Try another way to be notified.

junaruga:
I <jaruga@ruby-lang.org> haven't received the notifications since August 2022. So, try another way to be notified written on the document below.

https://docs.travis-ci.com/user/notifications/#configuring-email-notifications

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -146,4 +146,7 @@ notifications:
     on_success: never
     on_failure: always
   email:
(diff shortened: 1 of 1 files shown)


junaruga:
The Travis CI passed below.
https://app.travis-ci.com/github/ruby/ruby/builds/268011098

PR #9321, opened 2023-12-22
Rename thread_profile_frames() to profile_thread_frames()

osyoyu:
Match the word order in rb_profile_thread_frames().

rb_profile_thread_frames() was named rb_thread_profile_frames() in the original proposal, but it was renamed to its current name through discussion. The internal implementation should match its name too.


osyoyu:
Ref: #7784

PR #9322, opened 2023-12-22
Make bundled gems loadable

nobu:


mame:
As the comment says, do not add anything to `test/runner.rb`. This will make the environment inconsistent between parallel test and non-parallel test.

If you want to make bundled gems loadable, you should remove these two lines completely.

hsbt:
We can test bundled gems without this today. And that test is working at `ruby/rubygems` repo.

If we really need to load bundled gems for testing, this will apply this.

PR #9324, opened 2023-12-22
RUBY_FREE_AT_EXIT fixes

jhawthorn:
This fixes a couple errors which could occur when RUBY_FREE_AT_EXIT=1 is enabled. These fixes aren't perfect, but I think fixes which avoided more leaks and never caused errors would be more complex and would be dangerous to merge so close to 3.3's release. (I'd like to investigate this more next year and maybe we can backport the fixes For 3.3.1)

First, during VM destruction EC seems to be NULL, this avoids two cases we were always calling `GET_EC()` at that time (one we just didn't need and the other I added an if case for it being NULL). This fixes an error which always occurred when Ruby was built in debug.

Fixes `RUBY_FREE_AT_EXIT=1 ./miniruby -e ''`

Next, we need to move the freeing of `default_rand_key` after the freeing of Ractors in `rb_objspace_free_objects`, as that will iterate over over the local storage keys so the memory needs to be freed later to avoid a use-after-free.

Fixes `RUBY_FREE_AT_EXIT=1 ./miniruby -e rand`

Lastly when we forked for a Thread we would end up double-freeing the main thread's stack, because the *main thread* was no longer the *initial thread* from `Init_BareVM`. The new Thread's stack was freed naturally in `rb_objspace_free_objects`

Fixes `RUBY_FREE_AT_EXIT=1 ./miniruby -e 'Thread.new { fork { } }.join; Process.waitpid'`

With these changes `RUBY_FREE_AT_EXIT=1 make btest` works with RUBY_DEBUG and ASAN enabled (other than a few Ractor tests which detect false positives).

cc @HParker @peterzhu2118

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2587,12 +2587,12 @@ rb_threadptr_root_fiber_release(rb_thread_t *th)
         /* ignore. A root fiber object will free th->ec */
     }
     else {
-        rb_execution_context_t *ec = GET_EC();
+        rb_execution_context_t *ec = rb_current_execution_context(false);
 
         VM_ASSERT(th->ec->fiber_ptr->cont.type == FIBER_CONTEXT);
         VM_ASSERT(th->ec->fiber_ptr->cont.self == 0);
 
-        if (th->ec == ec) {
+        if (ec && th->ec == ec) {
             rb_ractor_set_current_ec(th->ractor, NULL);
         }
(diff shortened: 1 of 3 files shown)


HParker:
Nice find! The description makes a ton of sense to me! 👍

peterzhu2118 (approved):
Great finds. Make sense to me!

PR #9328, opened 2023-12-23
Replicate EEXIST epoll_ctl behavior in kqueue

jpcamara:
Before registering an event, the kqueue code now checks to see if that fd has already been registered. This seems pretty inefficient (would lead to alot of looping in heavy thread usage), but it does fix certain test cases like `test_tcp#test_accept_multithread`. Throwing it out there as a possibility.

I think ideally there would be an efficient lookup available to find existing threads waiting on a particular fd and what events they've registered. From what I can tell this information is all available in `timer_th.waiting`, but it can only be accessed by iterating the entire list.

I originally had an approach where I tried to find all registered threads whenever the events fired. That meant no additional native threads got created, but also seemed more inefficient (even more looping, even more `ccan_list_for_each`). Something more like a hash lookup would be ideal.

cc: @ko1

### DIFF
--- a/thread_pthread_mn.c
+++ b/thread_pthread_mn.c
@@ -616,6 +616,21 @@ kqueue_unregister_waiting(int fd, enum thread_sched_waiting_flag flags)
     }
 }
 
+static bool
+kqueue_already_registered(int fd)
+{
+    rb_thread_t *wth, *found_wth = NULL;
+    ccan_list_for_each(&timer_th.waiting, wth, sched.waiting_reason.node) {
+        // Similar to EEXIST in epoll_ctl, but more strict because it checks fd rather than flags
+        //   for simplicity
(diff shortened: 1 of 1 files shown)


ko1:
Thank you for the report. O(n) is not acceptable so I'll check it more.

jpcamara:
> Thank you for the report. O(n) is not acceptable so I'll check it more.

Makes sense. This is ultimately something needed for epoll as well, since these scenarios (multiple registrations) "work" there because of EEXIST, but always lead to native threads getting created. It's just worse with kevent because instead of rejecting the registration it overwrites the previous entry.

ko1:
I've tried to make a bitmap (available or not) but now it is buggy so I merge this PR because it is better than not working.
Thank you the patch.

PR #9329, opened 2023-12-23
IO::Buffer improvements and documentation.

ioquatix:
Tidy up the documentation of the C implementation.


nobu:
Doesn't look documentation only?

ioquatix:
You are correct, I reorganised and made some minor improvements of the code too.

io_buffer.c
```diff
@@ -287,7 +294,23 @@ static const rb_data_type_t rb_io_buffer_type = {
     .flags = RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED | RUBY_TYPED_EMBEDDABLE,
 };
 
-// Extract an offset argument, which must be a positive integer.
+static inline enum rb_io_buffer_flags
+io_buffer_extract_flags(VALUE argument)
+{
+    if (rb_int_negative_p(argument)) {
+        rb_raise(rb_eArgError, "Flags can't be negative!");
```
nobu, line 301: We usually don’t end error messages with exclamation (or period).

ioquatix: Have we documented this anywhere?

ioquatix: Maybe we can reference something like this: https://learn.microsoft.com/en-us/dotnet/standard/exceptions/best-practices-for-exceptions

> Use grammatically correct error messages
> Write clear sentences and include ending punctuation. Each sentence in the string assigned to the [Exception.Message](https://learn.microsoft.com/en-us/dotnet/api/system.exception.message#system-exception-message) property should end in a period. For example, "The log table has overflowed." would be an appropriate message string.

PR #9344, opened 2023-12-24
This single change fixes every MN test on macos...

jpcamara:
No way that's right, as in truly _fixing_ it. But it's curious that they start working. Maybe it just triggers them to switch to native threads, which is the opposite of the intended affect. But here's a commit to test it out anyways

Related to https://bugs.ruby-lang.org/issues/20076


hsbt:
@jpcamara Thanks. I'm closing this. see update of https://bugs.ruby-lang.org/issues/20076

PR #9345, opened 2023-12-24
[DOC] Markdown table for File#flock

BurdetteLamar:


file.c
```diff
@@ -5227,50 +5227,25 @@ rb_thread_flock(void *data)
...
- *  Locks or unlocks a file according to the given `locking_constant`,
+ *  Locks or unlocks file +self+ according to the given `locking_constant`,
  *  a bitwise OR of the values in the table below.
  *
  *  Not available on all platforms.
  *
  *  Returns `false` if `File::LOCK_NB` is specified and the operation would have blocked;
  *  otherwise returns `0`.
+ *
  *  <br>
```
peterzhu2118, line 5238: Is there a reason why we should keep the line break tags before and after the table?

BurdetteLamar: I had to experiment to get any vertical whitespace at all;  this is what I came up with.

peterzhu2118:
@BurdetteLamar it looks like this was closed because you deleted the repository, did you mean to do this?

peterzhu2118:
Yeah, I think a new PR is the only way to fix this.

PR #9356, opened 2023-12-25
[DOC] RDoc for complex.c

BurdetteLamar:
Prefer `Complex.rect` to `Complex`, as being more explicit.

There are about 70 more of these, but I want to get agreement before doing more.


peterzhu2118 (approved):
> I want to get agreement before doing more.

I'm indifferent on whether we use `Kernel#Complex` or `Complex.rect`, since they're both valid Ruby API. Since you think this is better, we can standardize on this.

PR #9357, opened 2023-12-26
Use ubf list on cygwin

fd00:
Use ubf list on cygwin as on other platforms.
This allows test_io to pass.

- https://bugs.ruby-lang.org/issues/4957
- https://bugs.ruby-lang.org/issues/5055
- https://bugs.ruby-lang.org/issues/7781
- https://bugs.ruby-lang.org/issues/15465
- https://bugs.ruby-lang.org/issues/17996

Cygwin appears to support SIGVTALRM.
- https://github.com/Alexpux/Cygwin/commit/8a0efa53e44919bcf5ccb1d3353618a82afdf8bc

### DIFF
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -305,7 +305,7 @@ event_name(rb_event_flag_t event)
 
 static rb_serial_t current_fork_gen = 1; /* We can't use GET_VM()->fork_gen */
 
-#if defined(SIGVTALRM) && !defined(__CYGWIN__) && !defined(__EMSCRIPTEN__)
+#if defined(SIGVTALRM) && !defined(__EMSCRIPTEN__)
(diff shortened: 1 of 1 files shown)


nobu (approved):
> Use ubf list on cygwin as on other platforms. This allows test_io to pass.

Since we don't have cygwin CI, we can only take what you say.

> Cygwin appears to support SIGVTALRM.
> 
> * [Alexpux/Cygwin@8a0efa5](https://github.com/Alexpux/Cygwin/commit/8a0efa53e44919bcf5ccb1d3353618a82afdf8bc)

It seems Feb 2000, but the line was introduced in Aug 2018.
I guess that SIGVTALRM didn't work at that time at least.

PR #9359, opened 2023-12-26
Use `PRIuSIZE` instead of `%zu` for `size_t`

nobu:
Since the 'z' modifier for `size_t` is specified in C99, but is not supported in VC2013, until we drop support for VC2013, we have to use `PRIuSIZE` instead.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -213,7 +213,7 @@ rb_malloc_grow_capa(size_t current, size_t type_size)
     new_capacity -= malloc_offset;
     new_capacity /= type_size;
     if (current > new_capacity) {
-        rb_bug("rb_malloc_grow_capa: current_capacity=%zu, new_capacity=%zu, malloc_offset=%zu", current, new_capacity, malloc_offset);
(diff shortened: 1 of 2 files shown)


ioquatix:
Are we going to drop VS2013 for 3.4?

prism/prism.c
```diff
@@ -977,7 +977,7 @@ static inline void *
 pm_alloc_node(PRISM_ATTRIBUTE_UNUSED pm_parser_t *parser, size_t size) {
     void *memory = calloc(1, size);
     if (memory == NULL) {
-        fprintf(stderr, "Failed to allocate %zu bytes\n", size);
+        fprintf(stderr, "Failed to allocate "PRIuSIZE" bytes\n", size);
```
nobu, line 980:
```suggestion
        fprintf(stderr, "Failed to allocate %zu bytes\n", size);
```

hsbt:
see https://bugs.ruby-lang.org/issues/19982

hsbt:
We no longer support `vc2013` for Ruby 3.4. But We may backport this to Ruby 3.3 with `vc2013`.

@k0kubun Can you look this?

k0kubun (approved):

ioquatix:
Thanks everyone for your efforts!

PR #9362, opened 2023-12-26
Introduce `get_node_nd_value` and `get_node_nd_vid` macro

S-H-GAMELINKS:
I thought that introducing `get_node_nd_value` and `get_node_nd_vid` macros to make similar case statements more simpler.
How is this?


Empact:
This strikes me as too cute. Ymmv.

PR #9364, opened 2023-12-26
[Bug #20088] Fix to use target_os for cross compiling

nobu:
Fix https://github.com/ruby/ruby/pull/9250#discussion_r1436342036

### DIFF
--- a/tool/m4/ruby_default_arch.m4
+++ b/tool/m4/ruby_default_arch.m4
@@ -10,10 +10,12 @@ AS_CASE([$1:"$host_cpu"],
     [[i[3-6]86]:x86_64], [ARCH_FLAG=-m32],
     [ppc64:ppc*],        [ARCH_FLAG=-m64],
     [ppc*:ppc64],        [ARCH_FLAG=-m32],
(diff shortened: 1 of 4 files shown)


tool/m4/ruby_default_arch.m4
```diff
@@ -10,7 +10,7 @@ AS_CASE([$1:"$host_cpu"],
     [[i[3-6]86]:x86_64], [ARCH_FLAG=-m32],
     [ppc64:ppc*],        [ARCH_FLAG=-m64],
     [ppc*:ppc64],        [ARCH_FLAG=-m32],
-    AS_CASE([$build_os],
+    AS_CASE([$target_os],
```
stanhu, line 13: I tried this, but it's still not working yet. I wonder if `$target_os` has to be passed in as an argument as `$host_cpu` is.

nobu: Changed to detect a possible option.

stanhu:
Thanks, this worked for `arm64-darwin`.

PR #9366, opened 2023-12-26
[DOC] Add What's Here for Complex

BurdetteLamar:


complex.c
```diff
@@ -2506,6 +2506,80 @@ float_arg(VALUE self)
...
+ *   - #real?: Returns +false+; for compatibility with Numeric#real?.
+ *   - #rect (and its alias #rectangular):
+ *     Returns the array <tt>[self.real, self.imag]</tt>.
+ *
+ * - Comparing:
+ *
+ *   - {#<=>}[Complex.html#method-i-3C-3D-3E]:
+ *     Returns whether +self+ is less than, equal to, or greater than the given argument.
+ *   - {#==}[Complex.html#method-i-3D-3D]:
+ *     Returns whether +self+ is equal to the given argument.
```
peterzhu2118, line 2551: This generates correctly for me, so I don't think we need to manually link to `#<=>` and `#==`.

```suggestion
 *   - #<=>: Returns whether +self+ is less than, equal to, or greater than the given argument.
 *   - #==: Returns whether +self+ is equal to the given argument.
```

complex.c
```diff
@@ -2506,6 +2506,80 @@ float_arg(VALUE self)
...
+ *
+ * - Converting:
+ *
+ *   - #rationalize: Returns a Rational object whose value is exactly
+ *     or approximately equivalent to that of <tt>self.real</tt>.
+ *   - #to_c: Returns +self+.
+ *   - #to_d: Returns the value as a BigDecimal object.
+ *   - #to_f: Returns the value of <tt>self.real</tt> as a Float, if possible.
+ *   - #to_i: Returns the value of <tt>self.real</tt> as an Integer, if possible:
```
peterzhu2118, line 2560:
```suggestion
 *   - #to_i: Returns the value of <tt>self.real</tt> as an Integer, if possible.
```

complex.c
```diff
@@ -2506,6 +2506,80 @@ float_arg(VALUE self)
...
+ *   - #conj (and its alias #conjugate): Returns the conjugate of +self+.
+ *   - #fdiv: Returns <tt>Complex.rect(self.real/numeric, self.imag/numeric)</tt>.
+ *
+ * - Working with JSON:
+ *
+ *   - ::json_create: Returns a new \Complex object,
+ *     deserialized from the given serialized hash.
+ *   - #as_json: Returns a serialized hash constructed from +self+.
+ *   - #to_json: Returns a JSON string representing +self+.
+ *
```
peterzhu2118, line 2582: We should add the little blurb about these methods not being available by default.

```suggestion
 *
 * These methods are provided by the {JSON gem}[https://github.com/flori/json]. To make these methods available:
 *
 *   require 'json/add/complex'
```

complex.c
```diff
@@ -2506,6 +2506,79 @@ float_arg(VALUE self)
...
+ * First, what's elsewhere:
+ *
+ * - \Class \Complex inherits (directly or indirectly)
+ *   from classes {Numeric}[rdoc-ref:Numeric@What-27s+Here]
+ *   and {Object}[rdoc-ref:Object@What-27s+Here].
+ * - Includes (indirectly) module {Comparable}[rdoc-ref:Comparable@What-27s+Here].
+ *
+ * Here, class \Complex has methods for:
+ *
+ * - Creating \Complex objects:
```
peterzhu2118, line 2520: I noticed that other docs (e.g. [Array](https://docs.ruby-lang.org/en/master/Array.html#class-Array-label-What-27s+Here), [Hash](https://docs.ruby-lang.org/en/master/Hash.html#class-Hash-label-What-27s+Here), [Range](https://docs.ruby-lang.org/en/master/Range.html#class-Range-label-What-27s+Here)) use subtitles instead of bullet points. Should we do that here too for consistency?

BurdetteLamar: Done.

peterzhu2118 (approved):

PR #9371, opened 2023-12-27
[Bug #20085] Use consistent default options for `-mbranch-protection`

kateinoigakukun:
We need to use the same options for both C compiler and assembler when `-mbranch-protection` is guessed by configure. Otherwise, `coroutine/arm64/Context.{h,S}` will use incompatible PAC strategies.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -830,7 +830,10 @@ AS_IF([test "$GCC" = yes], [
 	AS_FOR(option, opt, [-mbranch-protection=pac-ret -msign-return-address=all], [
             RUBY_TRY_CFLAGS(option, [branch_protection=yes], [branch_protection=no])
             AS_IF([test "x$branch_protection" = xyes], [
+                # C compiler and assembler must be consistent for -mbranch-protection
+                # since they both check `__ARM_FEATURE_PAC_DEFAULT` definition.
                 RUBY_APPEND_OPTION(XCFLAGS, option)
+                RUBY_APPEND_OPTION(ASFLAGS, option)
                 break
             ])
         ])


flavorjones:
@kateinoigakukun I saw your comment "Unfortunately, we don't have a CI node that affects the issue" and wanted to say that it's possible to test this architecture using qemu, see for example https://github.com/sparklemotion/nokogiri/blob/34571990eaac1c95a3dea31fd8f31b6baa18df47/.github/workflows/ci.yml#L597-L617

I'm not sure where in the ruby github actions suite to place this, but the general idea would be:

```yaml
      - run: |
          docker run --rm --privileged multiarch/qemu-user-static --reset -p yes
          docker run --rm -v "$(pwd):/ruby" -w /ruby \
            --platform=linux/arm64/v8 ubuntu:22.04 \
            ./script-to-build-and-test-ruby
```

kateinoigakukun:
@flavorjones Thanks for your info :)

saviogl:
When is this fix coming out? Would love not have to revert back to 3.2.2 unless necessary

kobaltz:
Notating this here since this PR is the first thing that comes up on a "Apple Silicon Docker Ruby 3.3.0" search. For anyone else having issues on Apple Silicon and Docker with Ruby 3.3.0, this is what I'm currently doing until the next release. This is just for my local development environment and I'll use `ruby:3.3.0-slim` on production images.

```
FROM debian:bullseye-slim as base
RUN apt update && apt install -y build-essential wget autoconf
RUN wget https://github.com/postmodern/ruby-install/releases/download/v0.9.3/ruby-install-0.9.3.tar.gz \
  && tar -xzvf ruby-install-0.9.3.tar.gz \
  && cd ruby-install-0.9.3/ \
  && make install
RUN ruby-install -p https://github.com/ruby/ruby/pull/9371.diff ruby 3.3.0
ENV PATH="/opt/rubies/ruby-3.3.0/bin:${PATH}"
```

It definitely takes longer to do a fresh build locally, but at least it is a workaround until the patch is available in the mainstream Ruby docker images.

PR #4038, opened 2021-01-08
Implement Happy Eyeballs Version 2 (RFC8305) in Socket.tcp

mmasaki:
This change implements Happy Eyeballs Version 2 (RFC8305) in Socket.tcp.
It enables fallback from IPv6 to IPv4 without a long waiting time.

PR #9374, opened 2023-12-27
Introduction of Happy Eyeballs Version 2 (RFC8305) in Socket.tcp

shioimm:
This is an implementation of Happy Eyeballs version 2 (RFC 8305) in Socket.tcp.

### Background
Currently, `Socket.tcp` synchronously resolves names and makes connection attempts with `Addrinfo::foreach.`
This implementation has the following two problems.

1. In hostname resolution, the program stops until the DNS server responds to all DNS queries.
2. In a connection attempt, while an IP address is trying to connect to the destination host and is taking time, the program stops, and other resolved IP addresses cannot try to connect.

### Proposal
"Happy Eyeballs" ([RFC 8305](https://datatracker.ietf.org/doc/html/rfc8305)) is an algorithm to solve this kind of problem. It avoids delays to the user whenever possible and also uses IPv6 preferentially.
I implemented it into `Socket.tcp` by using `Addrinfo.getaddrinfo` in each thread spawned per address family to resolve the hostname asynchronously, and using `Socket::connect_nonblock` to try to connect with multiple addrinfo in parallel. 

### Outcome

This change eliminates a fatal defect in the following cases.

#### Case 1. One of the A or AAAA DNS queries does not return

```ruby
require 'socket'

class Addrinfo
  class << self
    # Current Socket.tcp depends on foreach
    def foreach(nodename, service, family=nil, socktype=nil, protocol=nil, flags=nil, timeout: nil, &block)
      getaddrinfo(nodename, service, Socket::AF_INET6, socktype, protocol, flags, timeout: timeout)
        .concat(getaddrinfo(nodename, service, Socket::AF_INET, socktype, protocol, flags, timeout: timeout))
        .each(&block)
    end

    def getaddrinfo(_, _, family, *_)
      case family
      when Socket::AF_INET6 then sleep
      when Socket::AF_INET then [Addrinfo.tcp("127.0.0.1", 4567)]
      end
    end
  end
end

Socket.tcp("localhost", 4567)
```

Because the current `Socket.tcp` cannot resolve IPv6 names, the program stops in this case. It cannot start to connect with IPv4 address.
Though `Socket.tcp` with HEv2 can promptly start a connection attempt with IPv4 address in this case.

#### Case 2. Server does not promptly return ack for syn of either IPv4 / IPv6 address family

```ruby
require 'socket'

fork do
  socket = Socket.new(Socket::AF_INET6, :STREAM)
  socket.setsockopt(:SOCKET, :REUSEADDR, true)
  socket.bind(Socket.pack_sockaddr_in(4567, '::1'))
  sleep
  socket.listen(1)
  connection, _ = socket.accept
  connection.close
  socket.close
end

fork do
  socket = Socket.new(Socket::AF_INET, :STREAM)
  socket.setsockopt(:SOCKET, :REUSEADDR, true)
  socket.bind(Socket.pack_sockaddr_in(4567, '127.0.0.1'))
  socket.listen(1)
  connection, _ = socket.accept
  connection.close
  socket.close
end

Socket.tcp("localhost", 4567)
```

The current `Socket.tcp` tries to connect serially, so when its first name resolves an IPv6 address and initiates a connection to an IPv6 server, this server does not return an ACK, and the program stops.
Though `Socket.tcp` with HEv2 starts to connect sequentially and in parallel so a connection can be established promptly at the socket that attempted to connect to the IPv4 server.

In exchange, the performance of `Socket.tcp` with HEv2 will be degraded.

```
100.times { Socket.tcp("www.ruby-lang.org", 80) }
# Socket.tcp (Before) 0.123809
# Socket.tcp (After)  0.246387
```

This is due to the addition of the creation of IO objects, Thread objects, etc., and calls to `IO::select` in the implementation.

### DIFF
--- a/ext/socket/lib/socket.rb
+++ b/ext/socket/lib/socket.rb
@@ -703,7 +703,12 @@ class Socket < BasicSocket
           family_name, res = hostname_resolution_queue.get
 
           if res.is_a? Exception
-            last_error = res unless (res.is_a?(Socket::ResolutionError)) && (res.error_code == Socket::EAI_ADDRFAMILY)
+            unless (Socket.const_defined?(:EAI_ADDRFAMILY)) &&
+                (res.is_a?(Socket::ResolutionError)) &&
+                (res.error_code == Socket::EAI_ADDRFAMILY)
+              last_error = res
+            end
+
             if hostname_resolution_retry_count.zero?
               state = :failure
               break
--- a/ext/socket/lib/socket.rb
+++ b/ext/socket/lib/socket.rb
@@ -663,6 +663,7 @@ class Socket < BasicSocket
     state = :start
     connected_socket = nil
     last_error = nil
+    is_windows_environment ||= (RUBY_PLATFORM =~ /mswin|mingw|cygwin/)
 
     if local_host && local_port
       local_addrinfos = Addrinfo.getaddrinfo(local_host, local_port, nil, :STREAM, nil)
@@ -814,7 +815,16 @@ class Socket < BasicSocket
 
         if connectable_sockets&.any?
           while (connectable_socket = connectable_sockets.pop)
-            if (sockopt = connectable_socket.getsockopt(Socket::SOL_SOCKET, Socket::SO_ERROR)).int.zero?
+            is_connected =
+              if is_windows_environment
+                sockopt = connectable_socket.getsockopt(Socket::SOL_SOCKET, Socket::SO_CONNECT_TIME)
+                sockopt.unpack('i').first >= 0
+              else
+                sockopt = connectable_socket.getsockopt(Socket::SOL_SOCKET, Socket::SO_ERROR)
+                sockopt.int.zero?
+              end
+
+            if is_connected
               connected_socket = connectable_socket
               connecting_sockets.delete connectable_socket
               connectable_sockets.each do |other_connectable_socket|
--- a/ext/socket/lib/socket.rb
+++ b/ext/socket/lib/socket.rb
@@ -602,9 +602,6 @@ class Socket < BasicSocket
   RESOLUTION_DELAY = 0.05
   private_constant :RESOLUTION_DELAY
 
-  PATIENTLY_RESOLUTION_DELAY = 3
-  private_constant :PATIENTLY_RESOLUTION_DELAY
-
   CONNECTION_ATTEMPT_DELAY = 0.25
   private_constant :CONNECTION_ATTEMPT_DELAY
 
@@ -668,7 +665,7 @@ class Socket < BasicSocket
     hostname_resolution_threads = []
     hostname_resolution_queue = nil
     hostname_resolution_waiting = nil
-    wait_for_hostname_resolution_patiently = false
+    hostname_resolution_expires_at = nil
     selectable_addrinfos = SelectableAddrinfos.new
     connecting_sockets = ConnectingSockets.new
     local_addrinfos = []
@@ -700,7 +697,7 @@ class Socket < BasicSocket
         resolving_family_names = ADDRESS_FAMILIES.keys
         hostname_resolution_queue = HostnameResolutionQueue.new(resolving_family_names.size)
         hostname_resolution_waiting = hostname_resolution_queue.waiting_pipe
-        hostname_resolution_started_at = current_clocktime
+        hostname_resolution_started_at = current_clocktime if resolv_timeout
         hostname_resolution_args = [host, port, hostname_resolution_queue]
 
         hostname_resolution_threads.concat(
@@ -713,6 +710,7 @@ class Socket < BasicSocket
         )
 
         hostname_resolution_retry_count = resolving_family_names.size - 1
+        hostname_resolution_expires_at = hostname_resolution_started_at + resolv_timeout if resolv_timeout
 
         while hostname_resolution_retry_count >= 0
           remaining = resolv_timeout ? second_to_timeout(hostname_resolution_started_at + resolv_timeout) : nil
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
This PR supersedes https://github.com/ruby/ruby/pull/4038.

shioimm:
NOTE
I have added the following tests:

- If an IPv6 address is resolved before an IPv4 address, connect using the IPv6 address.
  - test_tcp_socket_v6_hostname_resolved_earlier
- If an IPv4 address is resolved before an IPv6 address, connect using the IPv4 address.
  - test_tcp_socket_v4_hostname_resolved_earlier
- If an IPv4 address is resolved first, but the IPv6 address is resolved during the Resolution Delay, connect using the IPv6 address.
  - test_tcp_socket_v6_hostname_resolved_in_resolution_delay
- If the IPv6 address is resolved first but the IPv6 server is not listening, and the IPv4 address is then resolved, connect using the IPv4 address.
  - test_tcp_socket_v6_hostname_resolved_earlier_and_v6_server_is_not_listening
- If name resolution does not complete within resolv_timeout, an Errno::ETIMEDOUT error should occur.
  - test_tcp_socket_resolv_timeout
- If name resolution fails for any address family, connect using the address that was successfully resolved.
  - test_tcp_socket_one_hostname_resolution_succeeded_at_least
- If both IPv4 and IPv6 address resolutions fail, raise an exception with the error of the latter failed resolution.
  - test_tcp_socket_all_hostname_resolution_failed
- If a resolved IPv6 address is passed as an argument, connect using the IPv6 address.
  - test_tcp_socket_v6_address_passed
- Ensure connection is possible when passing fast_fallback: false as an argument.
  - test_tcp_socket_fast_fallback_is_false

Tests not added:

- Connecting using an IPv4 address when a resolved IPv4 address is specified as host.
- Raising an IPv4 connection failure exception if the connection using an IPv4 address fails after specifying a resolved IPv4 address as host.
- Raising an IPv6 connection failure exception if the connection using an IPv6 address fails after specifying a resolved IPv6 address as host.
- Connecting using an IPv6 address if IPv4 address resolution fails and then IPv6 address resolution succeeds.
- Raising an IPv6 name resolution error exception if IPv4 address resolution fails and then IPv6 address resolution fails.
- Connecting using an IPv4 address if IPv4 address resolution succeeds and then IPv6 address resolution fails during the Resolution Delay.
- Raising an IPv4 connection failure exception if IPv4 address resolution succeeds, IPv6 address resolution fails during the Resolution Delay, and then the connection using the IPv4 address fails.
- Raising an IPv4 connection failure exception if IPv4 address resolution succeeds, IPv6 address resolution succeeds during the Resolution Delay, but the connection using the IPv6 address fails, followed by a failed connection using the IPv4 address.
- Connecting using the respective IPv6 address if there are unattempted IPv6 and IPv4 addresses after IPv4 address resolution succeeds, IPv6 address resolution succeeds during the Resolution Delay, but the connection using both addresses fails.
- Connecting using the respective IPv6 address if there are unattempted IPv6 addresses after IPv4 address resolution succeeds, IPv6 address resolution succeeds during the Resolution Delay, but the connection using both addresses fails.
- Connecting using the respective IPv4 address if there are unattempted IPv4 addresses after IPv4 address resolution succeeds, IPv6 address resolution succeeds during the Resolution Delay, but the connection using both addresses fails.
- Connecting using an IPv6 address if IPv4 address resolution succeeds, the connection using the IPv4 address fails, but the IPv6 address resolution succeeds within the PATIENTLY_RESOLUTION_DELAY.
- Raising an IPv6 connection failure exception if IPv4 address resolution succeeds, the connection using the IPv4 address fails, but the IPv6 address resolution succeeds within the PATIENTLY_RESOLUTION_DELAY, followed by a failed connection using the IPv6 address.
- Connecting using the respective IPv6 address if there are unattempted IPv6 addresses after IPv4 address resolution succeeds, the connection using the IPv4 address fails, but the IPv6 address resolution succeeds within the PATIENTLY_RESOLUTION_DELAY, followed by a failed connection using the IPv6 address.
- Raising an IPv4 connection failure exception if IPv4 address resolution succeeds, the connection using the IPv4 address fails, but IPv6 address resolution does not succeed within the PATIENTLY_RESOLUTION_DELAY.
- Connecting using an IPv4 address if IPv6 address resolution fails and then IPv4 address resolution succeeds.
- Raising an IPv4 name resolution error exception if IPv6 address resolution fails and then IPv4 address resolution fails.
- Connecting using an IPv4 address if IPv6 address resolution succeeds and the connection using the IPv6 address fails, but IPv4 address resolution succeeds within the PATIENTLY_RESOLUTION_DELAY.
- Raising an IPv6 connection failure exception if IPv6 address resolution succeeds and the connection using the IPv6 address fails, but IPv4 address resolution fails within the PATIENTLY_RESOLUTION

ext/socket/lib/socket.rb
```diff
@@ -599,6 +599,47 @@ def accept_nonblock(exception: true)
...
+  HOSTNAME_RESOLUTION_QUEUE_UPDATED = 0
+  private_constant :HOSTNAME_RESOLUTION_QUEUE_UPDATED
+
+  @tcp_fast_fallback = true
+
+  class << self
+    attr_accessor :tcp_fast_fallback
+
+    def initialize_socket
+      @local_ip_address_list = Socket.ip_address_list
```
hanazuki, line 626: I suppose caching `Socket.ip_address_list` permanently is not a good idea. The host may later gain or lose IP addresses on network configuration changes (e.g., Wi-Fi / cellular network) or when the OS is booting and the network interfaces are not yet ready.

shioimm: Thank you for your comment.
As you mentioned, persisting the cache could lead to unexpected issues, so we have decided to remove it. https://github.com/ruby/ruby/pull/9374/commits/a3d7056591c1f798b13893ec4ab603623980a7e2

hanazuki: I'd also support removing the dual-/single-stack check because it could result in false negatives in certain corner cases, for example,
- When the host only has a link-local address, and /etc/hosts or mDNS allows resolving names of peer hosts to link-local addresses. In this case, the hosts can communicate using link-local addresses (APIPA network).
- When the host doesn't have an IPv6 GUA, and tries to connect to a hostname that resolves to ::1 (e.g., "localhost6" in /etc/hosts). In this case a PF_INET6 socket should be used.

I think it's difficult to enumerate this kind of unusual network configuration in the wild.

ext/socket/lib/socket.rb
```diff
@@ -624,8 +648,487 @@ def accept_nonblock(exception: true)
...
+      hostname_resolution_queue&.close_all
+
+      connecting_sockets.each do |connecting_socket|
+        connecting_socket.close unless connecting_socket.closed?
+      end
+    end
+  end
+
+  def self.specified_family_name_and_next_state(hostname)
+    if    hostname.match?(/:/)                             then [:ipv6, :v6c]
```
hanazuki, line 940: This is just a corner case, but several platforms (at least Linux and macOS) accept a colon (:) in hostnames, and DNS also allows a domain name with a colon. For instance, `Socket.tcp("x::y", 80)` and `Socket.tcp("1::1::1", 80)` specify hostnames (not IPv6 address literals), in which case both of the IPv4 and IPv6 name resolutions should be tried; otherwise, they fail to connect if the names only have IPv4 addresses.
So I think here we need the exact pattern to match IPv6 address literals (including link-local addresses with scope). False negatives (redundant name resolutions) are better than false positives (connection failures).

PR #9382, opened 2023-12-28
Introduce `Fiber#resuming?`.

ioquatix:
https://bugs.ruby-lang.org/issues/20102

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2999,9 +2999,12 @@ rb_fiber_alive_p(VALUE fiber_value)
 }
 
 /*
- *  call-seq: fiber.resumed? -> true or false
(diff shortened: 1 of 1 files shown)


cont.c
```diff
@@ -2998,6 +2998,24 @@ rb_fiber_alive_p(VALUE fiber_value)
     return RBOOL(!FIBER_TERMINATED_P(fiber_ptr(fiber_value)));
 }
 
+/*
+ *  call-seq: fiber.resumed? -> true or false
```
herwinw, line 3002: I guess this should be `fiber.resuming?` instead of `fiber.resumed?`

PR #9387, opened 2023-12-28
Fix [Bug #20095]: support nested/consecutive lookarounds

makenowjust:
https://bugs.ruby-lang.org/issues/20095


makenowjust:
This fix is buggy.

PR #9392, opened 2023-12-29
Move `negate_lit` to stub

nobu:
Instead replacing functions to negate each types, just replace `nagate_lit`.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -76,48 +76,6 @@
 #include "symbol.h"
 
 #ifndef RIPPER
-static void
-bignum_negate(VALUE b)
-{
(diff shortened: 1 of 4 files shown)


yui-knk:
@S-H-GAMELINKS plans to introduce Numeric Nodes. `negate_lit` is moved to outside of parse.y in the PR (https://github.com/S-H-GAMELINKS/ruby/pull/35/). What do you think the approach?

nobu:
I agree the basic policy.

yui-knk:
https://github.com/ruby/ruby/pull/9417 is merged.

PR #9395, opened 2023-12-29
[DOC] clarify situation for Ractor#send of T_DATA objects.

luke-gru:
For copying, use language like "some" T_DATA objects not supported. This is because objects like Time are supported, they don't refer to other objects so it's okay.

For `send(obj, move: true)`, it's currently not supported to move any T_DATA objects.

Fix [Bug #20078]


ko1:
Thank you!

PR #9398, opened 2023-12-30
Don't create T_MATCH object if /regexp/.match(string) doesn't match

luke-gru:
Fixes [Bug #20104]

### DIFF
--- a/re.c
+++ b/re.c
@@ -1747,14 +1747,17 @@ rb_reg_search_set_match(VALUE re, VALUE str, long pos, int reverse, int set_back
         .range = reverse ? 0 : len,
     };
 
-    VALUE match = match_alloc(rb_cMatch);
-    struct re_registers *regs = RMATCH_REGS(match);
+    struct re_registers regs = {0};
 
-    OnigPosition result = rb_reg_onig_match(re, str, reg_onig_search, &args, regs);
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
This looks correct to me, but I 'm not that knowledgeable about regexp internals. @makenowjust can you review?

re.c
```diff
@@ -1747,15 +1747,19 @@ rb_reg_search_set_match(VALUE re, VALUE str, long pos, int reverse, int set_back
         .range = reverse ? 0 : len,
     };
 
-    VALUE match = match_alloc(rb_cMatch);
-    struct re_registers *regs = RMATCH_REGS(match);
+    struct re_registers regs;
+    memset(&regs, 0, sizeof(struct re_registers));
```
nobu, line 1751: You can use `MEMZERO` macro.
```suggestion
    MEMZERO(&regs, 0, struct re_registers);
```
or
```suggestion
    memset(&regs, 0, sizeof(regs));
```
or
```suggestion
   struct re_registers regs = {0}; 
```

nobu (approved):

makenowjust (approved):

re.c
luke-gru: Okay thanks, I'll use c99 initializer 👍

PR #9407, opened 2024-01-03
Check hash key duplication for `__LINE__` and `__FILE__`

yui-knk:


parse.y
```diff
@@ -1956,6 +2037,28 @@ rb_parser_string_pointer(rb_parser_string_t *str)
     return str->ptr;
 }
 
+#ifndef UNIVERSAL_PARSER
+# define PARSER_STRING_GETMEM(str, ptrvar, lenvar, encvar) \
+    ((ptrvar) = str->ptr,                            \
+     (lenvar) = str->len,                            \
+     (encvar) = str->enc)                            \
```
nobu, line 2044: A macro ending with an empty line feels like a pitfall.
```suggestion
     (encvar) = str->enc)                            \
     /* end */
```

parse.y
```diff
@@ -123,25 +123,106 @@ hash_literal_key_p(VALUE k)
...
-    return rb_iseq_cdhash_cmp(val, lit);
+    return node_cdhash_cmp(val, lit);
 }
 
 static st_index_t
 literal_hash(VALUE a)
 {
     if (!hash_literal_key_p(a)) return (st_index_t)a;
-    return rb_iseq_cdhash_hash(a);
+    return node_cdhash_hash(a);
```
nobu, line 225: Is it necessary to separate `node_cdhash_`-functions?

nobu (approved):

parse.y
yui-knk: I will remove `\` from the last line.

parse.y
yui-knk: Once all NODE_LIT is separated into dedecated NODE type, I will remove `rb_iseq_cdhash_hash` and `rb_iseq_cdhash_cmp` from parse.y. In such case, it might be good to define `node_cdhash_`-functions.
Am I misunderstanding something?

PR #9413, opened 2024-01-04
Free pthread_attr after setting up the thread

HParker:
```
require 'socket'

10.times do
  10_000.times do
    IPSocket.getaddress("localhost")
  end

  puts `ps -o rss= -p #{$$}`
end
```

before:

```
26744
33400
39928
46328
53240
59640
66168
72696
79352
85880
```

after:

```
20864
20864
20864
20864
20864
20992
20992
20992
20992
21120
```

bug: https://bugs.ruby-lang.org/issues/20149

### DIFF
--- a/ext/socket/raddrinfo.c
+++ b/ext/socket/raddrinfo.c
@@ -511,6 +511,11 @@ start:
     }
     pthread_detach(th);
 
+    int r;
+    if ((r = pthread_attr_destroy(&attr)) != 0) {
+        rb_bug_errno("pthread_attr_destroy", r);
+    }
+
     rb_thread_call_without_gvl2(wait_getaddrinfo, arg, cancel_getaddrinfo, arg);
 
     int need_free = 0;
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):

ext/socket/raddrinfo.c
```diff
@@ -732,12 +737,17 @@ rb_getnameinfo(const struct sockaddr *sa, socklen_t salen,
 #endif
 
     pthread_t th;
-    if (do_pthread_create(&th, 0, do_getnameinfo, arg) != 0) {
+    if (do_pthread_create(&th, &attr, do_getnameinfo, arg) != 0) {
```
HParker, line 740: I believe this was intended to be used when creating the thread, but I might be wrong.

I think this dates back to when we only did `pthread_create` here. i.e. https://github.com/ruby/ruby/commit/3dc311bdc8badb680267f5a10e0c467ddd9dfe4c

If this isn't needed, I think we can avoid calling `pthread_attr_init` all together.

CC: @mame

mame (approved):
You are right, thanks!

PR #9426, opened 2024-01-05
YJIT: Fallback Integer#<< if a shift amount varies

k0kubun:
* Chain-guard `Integer#<<` once and fallback to a C call if a shift amount varies.
* Rename an exit reason `amt_changed` to `amount_changed` since the name seemed a bit cryptic.

On `nqueens` https://github.com/Shopify/yjit-bench/pull/260 with `--yjit-call-threshold=1`, this change pushes `ratio_in_yjit` from 0.0% to 99.9% and makes it 3.60x faster.

```
before: ruby 3.4.0dev (2024-01-05T16:51:37Z master 4d03140009) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-01-08T17:07:08Z yjit-ltlt 43edc3419f) +YJIT [x86_64-linux]

nqueens  581.8        0.7         161.8       0.1         3.55           3.60
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4637,17 +4637,28 @@ fn jit_rb_int_lshift(
 
     // Untag the fixnum shift amount
     let shift_amt = comptime_shift.as_isize() >> 1;
-
     if shift_amt > 63 || shift_amt < 0 {
+        return false;
+    }
+
+    // Fallback to a C call if the shift amount varies
+    if asm.ctx.get_chain_depth() > 0 {
         return false;
     }
 
     let rhs = asm.stack_pop(1);
     let lhs = asm.stack_pop(1);
 
-    // Guard on the shift value we speculated on
+    // Guard on the shift amount we speculated on
     asm.cmp(rhs, comptime_shift.into());
-    asm.jne(Target::side_exit(Counter::lshift_amt_changed));
+    jit_chain_guard(
+        JCC_JNE,
+        jit,
+        asm,
+        ocb,
+        1,
+        Counter::lshift_amount_changed,
+    );
 
     let in_val = asm.sub(lhs, 1.into());
     let shift_opnd = Opnd::UImm(shift_amt as u64);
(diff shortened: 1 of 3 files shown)


noahgibbs (approved):

yjit/src/codegen.rs
```diff
@@ -4637,17 +4637,28 @@ fn jit_rb_int_lshift(
 
     // Untag the fixnum shift amount
     let shift_amt = comptime_shift.as_isize() >> 1;
-
     if shift_amt > 63 || shift_amt < 0 {
         return false;
     }
 
+    // Fallback to a C call if the shift amount varies
+    if asm.ctx.get_chain_depth() > 0 {
```
XrXr, line 4645: Looks like this effectively disables the specialized path, with `./miniruby --yjit-dump-disasm --yjit-call-threshold=1 -e '23434 << 3'` generating a call:

```
  # regenerate_branch
  # Block: <main>@-e:1 (chain_depth: 1)
  # reg_temps: 00000011
  # Insn: 0004 opt_ltlt (stack_size: 2)
  # call to Integer#<<
  # RUBY_VM_CHECK_INTS(ec)
  0x104d1803c: ldur w11, [x20, #0x20]
  0x104d18040: tst w11, w11
  <snip>
```

Should be `> 1` probably, and it looks like the defer_compilation is from gen_send_general() and it's unconditional.

k0kubun: Ugh, thanks for catching it. https://github.com/ruby/ruby/pull/9426/commits/43edc3419f581962122102fb0d00b0c53938dbed

I don't get the point of this increment by the way. If it's only for the "Double defer!" detection, we could assign one bit from `chain_depth_return_landing` to it instead of overloading `chain_depth`. It seems counter-intuitive when you look at `*_MAX_DEPTH` constants.

maximecb (approved):

k0kubun:
Did you use `--yjit-call-threshold=1`? As I said in https://github.com/Shopify/yjit-bench/pull/260#issuecomment-1879352986, the script doesn't JIT the method by default, so it needs to be updated.

maximecb:
kk. I'll tweak the benchmark parameters.

PR #9427, opened 2024-01-06
Fix GC.measure_total_time regression

rianmcguire:
Commit 93ac7405b80cc61930d73da04441fa09af1851e1 introduced a regression where measurements would still be taken after setting GC.measure_total_time = false.

This PR re-introduces the checks for the `objspace->flags.measure_gc` flag around the measurement code.

Fixes [[Bug #20157](https://bugs.ruby-lang.org/issues/20157)]

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -9749,10 +9749,6 @@ gc_enter_count(enum gc_enter_event event)
     }
 }
 
-#ifndef MEASURE_GC
-#define MEASURE_GC (objspace->flags.measure_gc)
-#endif
-
 static bool current_process_time(struct timespec *ts);
 
 static void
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis:
Thanks for your fix - it looks correct to me. Could you add your reproduction script from the issue you opened as a testcase (probably in `test/ruby/test_gc.rb` maybe?). It should more or less work as-is in an `assert_separately` block I think.

KJTsanaktsidis:
(just the check on `GC.stat(:time)`, asserting on what syscalls are made is obviously going to send the test directly to flaketown)

rianmcguire:
CI failures on the latest commit look like unrelated flake

PR #9434, opened 2024-01-07
Refactor numeric node functions

yui-knk:


ruby_parser.c
```diff
@@ -60,10 +60,10 @@ VALUE
 rb_node_float_literal_val(rb_node_float_t* node)
 {
     double d = strtod(node->val, 0);
+    VALUE val = DBL2NUM(d);
     if (node->minus) {
-        d = -d;
+        val = compile_negative_numeric(val);
```
nobu: Doesn't something work with `-d`?

yui-knk: I wanted to same logic (`float_value`) in `rb_node_float_literal_val` and `rb_node_imaginary_literal_val. Is it better to restore original code for performance?

yui-knk: Updated to keep `d = -d`.

PR #9442, opened 2024-01-08
Do not convert NODE_STR to NODE_LIT when the string is hash key

yui-knk:
parse.y converted NODE_STR when the string is hash key like

```
h1 = {"str1" => 1}
m1("str2" => 2)
m2({"str3" => 3})
```

This commit stop the conversion.
`static_literal_node_p` needs to know the node is for hash key or not for the optimization.


nobu (approved):
You can use `true` and `false` for `bool`.

yui-knk:
Thanks for the review, I updated to use `true` and `false`.

PR #9445, opened 2024-01-08
[PRISM] Blocks should track the found local depth

eightbitraptor:
Rather than rely purely on local depth offset. This is because we can't assume a specific depth offset for all variable accesses happening within a block in the same way that we can for rescue/ensure/for or other nodes that push scopes.

This is because block parameters are defined in the scope level, so we always need to start from the top most scope and walk backwards.

The difference in behaviour between local variable nodes in blocks vs elsewhere is implemented by two different nodes in the CRuby compiler NODE_LASGN and NODE_DASGN.

This fixes the example code in ruby/prism@2053 However the referenced `comile_test.rb` is still failing for a different reason.


kddnewton (approved):
Yeah I think this makes sense

peterzhu2118 (approved):

PR #9447, opened 2024-01-08
[Bug #20161] Fix memory leak in regexp grapheme clusters

peterzhu2118:
The cc->mbuf gets overwritten, so we need to free it to not leak memory.

For example:

```ruby
str = "hello world".encode(Encoding::UTF_32LE)

10.times do
  1_000.times do
    str.grapheme_clusters
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

```
15536
15760
15920
16144
16304
16480
16640
16784
17008
17280
```

After:

```
15584
15584
15760
15824
15888
15888
15888
15888
16048
16112
```


jeremyevans (approved):
Not a regexp expert, but this looks right to me.

PR #9456, opened 2024-01-09
Extract mutex_m as bundled gems

hsbt:


hsbt:
We should wait to merge this until releasing new version of `minitest`

PR #9461, opened 2024-01-09
[DOC] Enhance documentation for Array#zip

the-spectator:
This PR adds an example for `Array#zip` to accept a range object. It also mentions that `Array#zip` can accept an object that responds with `#each` and uses it to extract elements.

It is also covered in ruby/spec. https://github.com/the-spectator/ruby/blob/88d7838445ec84b1cc630ce3bd97bb71cd0aefd4/spec/ruby/core/array/zip_spec.rb#L31-L39

### DIFF
--- a/array.c
+++ b/array.c
@@ -4399,6 +4399,13 @@ take_items(VALUE obj, long n)
  *    d = a.zip(b, c)
  *    d # => [[:a0, :b0, :c0], [:a1, :b1, :c1], [:a2, :b2, :c2], [:a3, :b3, :c3]]
  *
+ *  If an argument is not an array, it extracts the values by calling #each:
+ *
+ *  a = [:a0, :a1, :a2, :a2]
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -4399,6 +4399,13 @@ take_items(VALUE obj, long n)
  *    d = a.zip(b, c)
  *    d # => [[:a0, :b0, :c0], [:a1, :b1, :c1], [:a2, :b2, :c2], [:a3, :b3, :c3]]
  *
+ *  If an argument is not an array, it tries to extract the values by calling #each:
```
peterzhu2118, line 4402: Nit: 

```suggestion
 *  If an argument is not an array, it extracts the values by calling #each:
```

the-spectator: Thanks for the suggestion. Added.

peterzhu2118 (approved):

PR #9463, opened 2024-01-09
Free environ when RUBY_FREE_AT_EXIT

peterzhu2118:
The environ is malloc'd, so it gets reported as a memory leak. This commit adds ruby_free_proctitle which frees it during shutdown when RUBY_FREE_AT_EXIT is set.

    STACK OF 1 INSTANCE OF 'ROOT LEAK: <calloc in ruby_init_setproctitle>':
    5   dyld                                  0x18b7090e0 start + 2360
    4   ruby                                  0x10000e3a8 main + 100  main.c:58
    3   ruby                                  0x1000b4dfc ruby_options + 180  eval.c:121
    2   ruby                                  0x1001c5f70 ruby_process_options + 200  ruby.c:3014
    1   ruby                                  0x10035c9fc ruby_init_setproctitle + 76  setproctitle.c:105
    0   libsystem_malloc.dylib                0x18b8c7b78 _malloc_zone_calloc_instrumented_or_legacy + 100

cc. @HParker

### DIFF
--- a/missing/setproctitle.c
+++ b/missing/setproctitle.c
@@ -80,10 +80,20 @@ static char **argv1_addr = NULL;
 
 #endif /* HAVE_SETPROCTITLE */
 
+#if defined(SPT_TYPE) && SPT_TYPE == SPT_REUSEARGV
+# define ALLOCATE_ENVIRON 1
+#else
+# define ALLOCATE_ENVIRON 0
+#endif
+
+#if ALLOCATE_ENVIRON
+static char **orig_environ = NULL;
+#endif
+
 void
 compat_init_setproctitle(int argc, char *argv[])
 {
-#if defined(SPT_TYPE) && SPT_TYPE == SPT_REUSEARGV
+#if ALLOCATE_ENVIRON
 	extern char **environ;
 	char *lastargv = NULL;
 	char *lastenvp = NULL;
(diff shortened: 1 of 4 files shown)


missing/setproctitle.c
```diff
@@ -80,10 +80,20 @@ static char **argv1_addr = NULL;
...
 #endif /* HAVE_SETPROCTITLE */
 
+#if defined(SPT_TYPE) && SPT_TYPE == SPT_REUSEARGV
+# define ALLOCATE_ENVIRON 1
+#else
+# define ALLOCATE_ENVIRON 0
+#endif
+
+#if ALLOCATE_ENVIRON
+char **orig_environ = NULL;
```
nobu, line 90:
```suggestion
static char **orig_environ = NULL;
```

missing/setproctitle.c
```diff
@@ -134,6 +145,26 @@ compat_init_setproctitle(int argc, char *argv[])
 #endif /* SPT_REUSEARGV */
 }
 
+void
+ruby_free_proctitle(void)
+{
+#if ALLOCATE_ENVIRON
+	extern char **environ;
```
nobu, line 152:
```suggestion
	extern char **environ;

	if (!orig_environ) return; /* environ is allocated by OS */
	
```

nobu (changes requested):

PR #9464, opened 2024-01-09
s/SafeStringValue/StringValue/

fxn:
The macro `SafeStringValue()` became just `StringValue()` in c5c05460ac2:

```c
#define SafeStringValue(v) StringValue(v)
```

and [it is deprecated](https://github.com/ruby/ruby/blob/55b7121358e2bed9b90001405b651f57fafbeaad/include/ruby/internal/core/rstring.h#L91-L98) nowadays.

This patch replaces remaining macro usage. Some occurrences are left in `ext/stringio` and `ext/win32ole`, they should be fixed upstream.

The macro itself is not deleted, because it may be used in extensions.

/cc @jeremyevans

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -3509,7 +3509,7 @@ dir_s_home(int argc, VALUE *argv, VALUE obj)
     rb_check_arity(argc, 0, 1);
     user = (argc > 0) ? argv[0] : Qnil;
     if (!NIL_P(user)) {
-        SafeStringValue(user);
+        StringValue(user);
         rb_must_asciicompat(user);
         u = StringValueCStr(user);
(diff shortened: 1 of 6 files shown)


jeremyevans (changes requested):
Looks mostly good to me.  We should definitely leave the macro defined, as there is no point in purposely breaking external extensions that use it.

However, for standard lib gems with upstream repositories (stringio, win32ole), the changes should be made in those repositories.

fxn:
@jeremyevans awesome, I have updated the patch and the PR description too.

jeremyevans (approved):

PR #9467, opened 2024-01-10
`st_index_t` is not `VALUE`

nobu:


nobu:
I need to investigate [mysterious warnings].
Also `make check` is cancelled twice, and the logs are deleted immediately for unknown reason.

```
warning: unused import: `condition::Condition`
  --> ../src/yjit/src/asm/arm64/arg/mod.rs:13:9
   |
13 | pub use condition::Condition;
   |         ^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `rb_yjit_fix_mul_fix as rb_fix_mul_fix`
   --> ../src/yjit/src/cruby.rs:188:9
    |
188 | pub use rb_yjit_fix_mul_fix as rb_fix_mul_fix;
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

[mysterious warnings]: https://github.com/ruby/ruby/actions/runs/7469933652/job/20328633704#step:7:89

nobu:
The YJIT warnings seems occurring randomly regardless this PR.
And the `make check` failures also look like unrelated GitHub Actions issues.

PR #9475, opened 2024-01-10
Fix issues with Ractor#send with move option for Array/Hash/Struct

luke-gru:
Array/Hash/Struct were not being moved properly. If arrays were embedded they weren't being moved. Hashes were not being updated properly and write barriers were not being fired on newly moved object. Structs fields were not being set properly.

Also fix issue where moved object_id was different from original object's object id.

Fixes [Bug #20165]

### DIFF
--- a/array.c
+++ b/array.c
@@ -392,6 +392,21 @@ rb_ary_make_embedded(VALUE ary)
     }
 }
 
+void rb_ary_unembed(VALUE ary)
+{
+    if (ARY_EMBED_P(ary)) {
+        long len = ARY_EMBED_LEN(ary);
+        long capacity = ARY_CAPA(ary);
+        VALUE *ptr = ary_heap_alloc(capacity);
+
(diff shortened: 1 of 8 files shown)


luke-gru:
WIP for now, I'm exporting some weird symbols in internal/hash.h for instance.
Also, I'm not firing write barriers on newly moved Array objects right now, I'll fix that soon.

luke-gru:
Yeah I agree about keeping arrays embedded. I was just creating a first pass at getting things correct, then I wanted to look into keeping the optimizations. With the way that objects are moved now, this seemed like the easy way to get it to work. For embedded arrays, I still need to look into how to copy them as I'm unfamiliar with how they work exactly (for example, if all arrays have the same embedded size, or if different objects in different size pools have different capacities). In reviewing this PR though, there are other issues other than the arrays (memory leak issues), so I think I'll close this PR and put the object_id stuff in a separate pull request. Then when I have time I'll look into doing these fixes properly.

PR #9482, opened 2024-01-11
Use 3.3.0 explicitly for Rubyspec Version Guards Check

hsbt:
https://github.com/ruby/ruby/actions/runs/7484732410/job/20372022839?pr=9481

`ruby/setup-ruby` returns `3.3.0-preview2` with `ruby-3.3`.


hsbt:
This is caused to use old version of `setup-ruby`. I fixed dependabot auto-update and submit https://github.com/ruby/ruby/pull/9486

PR #9488, opened 2024-01-11
Omit flaky test on mingw

hsbt:


unak (approved):
very very very very thanks!

PR #9492, opened 2024-01-11
Fix interrupts during Ractor.select

luke-gru:
Fixes [Bug #20168]


bootstraptest/test_ractor.rb
```diff
@@ -1467,6 +1467,25 @@ class C
...
+  r = Ractor.new do
+    loop do
+      sleep 1
+    end
+  end
+
+  Thread.new do
+    sleep 0.5
+    Process.kill(:INT, Process.pid)
+  do
```
nobu, line 1485:
```suggestion
  end
```

nobu (changes requested):

PR #9495, opened 2024-01-11
YJIT: expandarray for non-arrays

ywenc:
This PR updates the `expandarray` instruction for YJIT. 

`expandarray` is an instruction that appears for array decomposition or multiple assignment. 

Most commonly, `expandarray` will be encountered when assigning multiple variables at once from array, 
example - 

```ruby 
a, b, c = [1, 2, 3]
```

But it can often appear with non-arrays, when assigning multiple variables to a single value. It is common enough to see `expandarray` on a single `nil` that `gen_expandarray` previously had a special case if the operand was `nil`, but this is potentially incorrect because `to_ary` can be redefined on `NilClass`.

This PR updates the non-array case by bailing if `to_ary` is defined on a non-array operand. Otherwise, we push `nil`s to fill the unset values, then push the original operand, which also helps get rid of side exits from a pattern like:

```ruby
def some_method
   if ok
      true
   else 
      [false, err]
   end
end

value, err = some_method
```

This pattern was one of the most commonly seen side-exits for GitHub benchmarks.

Quick benchmark:

```bash
% time ./miniruby --disable-gems --yjit -e '10_000_000.times { a,b = 1 }'
./miniruby --disable-gems --yjit -e '10_000_000.times { a,b = 1 }'  0.35s user 0.02s system 61% cpu 0.594 total

% time ruby --disable-gems --yjit -e '10_000_000.times { a,b = 1 }'      
ruby --disable-gems --yjit -e '10_000_000.times { a,b = 1 }'  1.01s user 0.03s system 95% cpu 1.098 total
```

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -2509,7 +2509,7 @@ assert_equal '[1, 2]', %q{
 
   expandarray_redefined_nilclass
   expandarray_redefined_nilclass
-}
+} unless rjit_enabled?
 
 assert_equal '[1, 2, nil]', %q{
   def expandarray_rhs_too_small
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -2494,6 +2494,23 @@ assert_equal '[:not_array, nil, nil]', %q{
   expandarray_not_array(obj)
 }
 
+assert_equal '[1, 2]', %q{
+  class NilClass
+    private
+    def to_ary
+      [1, 2]
+    end
+  end
+
+  def expandarray_redefined_nilclass
+    a, b = nil
+    [a, b]
+  end
+
+  expandarray_redefined_nilclass
+  expandarray_redefined_nilclass
+}
+
 assert_equal '[1, 2, nil]', %q{
   def expandarray_rhs_too_small
     a, b, c = [1, 2]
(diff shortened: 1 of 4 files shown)


maximecb (approved):

yjit/src/codegen.rs
```diff
@@ -1657,10 +1645,51 @@ fn gen_expandarray(
...
-        return None;
+    // If the comptime receiver is not an array
+    if !unsafe { RB_TYPE_P(comptime_recv, RUBY_T_ARRAY) } {
+        // at compile time, ensure to_ary is not defined
+        let target_cme = unsafe { rb_callable_method_entry_or_negative(comptime_recv.class_of(), ID!(to_ary)) };
+        let cme_def_type = unsafe { get_cme_def_type(target_cme) };
+
+        if cme_def_type != VM_METHOD_TYPE_UNDEF {
+            // if to_ary is defined, return can't compile so to_ary can be called
+            return None;
```
k0kubun, line 1656: Could you leave a counter for this exit? Maybe rename the counter from `expandarray_comptime_not_array`.

k0kubun (approved):

bootstraptest/test_yjit.rb
```diff
@@ -2494,6 +2494,23 @@ def obj.to_ary
...
+  end
+
+  def expandarray_redefined_nilclass
+    a, b = nil
+    [a, b]
+  end
+
+  expandarray_redefined_nilclass
+  expandarray_redefined_nilclass
+}
```
k0kubun, line 2512:
```suggestion
} unless rjit_enabled? 
```

PR #9505, opened 2024-01-12
[Bug #20001] Fix GC issues with ASAN fake stacks (v2)

KJTsanaktsidis:
(this is https://github.com/ruby/ruby/pull/8903 again - I just accidently merged it too soon before)

----

This fixes some of the issues with ASAN I outlined in https://bugs.ruby-lang.org/issues/20001

With this merged, `make` now works on my machine with the following:

```
CC=clang ./configure cppflags="-fsanitize=address -fno-omit-frame-pointer" optflags=-O2 LDFLAGS="-fsanitize=address -fno-omit-frame-pointer" --prefix="$(realpath installed)"
ASAN_OPTIONS=use_sigaltstack=0:detect_leaks=0:abort_on_error=1 make
```

Without these fixes, I get crashes in various steps which use the built miniruby (e.g. parsing rdoc or runing extconf files) because `VALUE`'s on the machine stack are not getting properly marked (because they are instead living in ASAN fake stacks).

This seems to build fine with ASAN both enabled and disabled on linux/mac, and builds without asan on windows (i wasn't brave enough to try turning asan on with msvc heh).

[Bug #20001]

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -70,8 +70,6 @@ ruby_setup(void)
     if (GET_VM())
         return 0;
 
-    ruby_init_stack((void *)&state);
-
     /*
      * Disable THP early before mallocs happen because we want this to
      * affect as many future pages as possible for CoW-friendliness
@@ -115,7 +113,6 @@ ruby_options(int argc, char **argv)
     enum ruby_tag_type state;
     void *volatile iseq = 0;
 
-    ruby_init_stack((void *)&iseq);
     EC_PUSH_TAG(ec);
     if ((state = EC_EXEC_TAG()) == TAG_NONE) {
         SAVE_ROOT_JMPBUF(GET_THREAD(), iseq = ruby_process_options(argc, argv));
@@ -205,7 +202,6 @@ rb_ec_cleanup(rb_execution_context_t *ec, enum ruby_tag_type ex)
       step_0: step++;
         save_error = ec->errinfo;
         if (THROW_DATA_P(ec->errinfo)) ec->errinfo = Qnil;
-        ruby_init_stack(&message);
 
         /* exits with failure but silently when an exception raised
          * here */
@@ -324,14 +320,12 @@ ruby_run_node(void *n)
         rb_ec_cleanup(ec, (NIL_P(ec->errinfo) ? TAG_NONE : TAG_RAISE));
         return status;
     }
-    ruby_init_stack((void *)&status);
     return rb_ec_cleanup(ec, rb_ec_exec_node(ec, n));
 }
 
 int
 ruby_exec_node(void *n)
 {
-    ruby_init_stack((void *)&n);
     return rb_ec_exec_node(GET_EC(), n);
 }
 
--- a/gc.c
+++ b/gc.c
@@ -953,6 +953,11 @@ typedef struct rb_objspace {
 
     rb_darray(VALUE *) weak_references;
     rb_postponed_job_handle_t finalize_deferred_pjob;
+
+#ifdef RUBY_ASAN_ENABLED
+    rb_execution_context_t *marking_machine_context_ec;
+#endif
+
 } rb_objspace_t;
 
 
@@ -6821,6 +6826,26 @@ mark_const_tbl(rb_objspace_t *objspace, struct rb_id_table *tbl)
 static void each_stack_location(rb_objspace_t *objspace, const rb_execution_context_t *ec,
                                  const VALUE *stack_start, const VALUE *stack_end, void (*cb)(rb_objspace_t *, VALUE));
 
+static void
+gc_mark_machine_stack_location_maybe(rb_objspace_t *objspace, VALUE obj)
+{
+    gc_mark_maybe(objspace, obj);
+
+#ifdef RUBY_ASAN_ENABLED
+    rb_execution_context_t *ec = objspace->marking_machine_context_ec;
+    void *fake_frame_start;
+    void *fake_frame_end;
+    bool is_fake_frame = asan_get_fake_stack_extents(
+        ec->thread_ptr->asan_fake_stack_handle, obj,
+        ec->machine.stack_start, ec->machine.stack_end,
+        &fake_frame_start, &fake_frame_end
+    );
+    if (is_fake_frame) {
+        each_stack_location(objspace, ec, fake_frame_start, fake_frame_end, gc_mark_maybe);
+    }
+#endif
+}
+
 #if defined(__wasm__)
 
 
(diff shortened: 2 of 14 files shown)


KJTsanaktsidis:
@eightbitraptor to pick you more or less at random amongst people who know about the GC ... :) Would you mind giving a quick once-over of this PR? I would like a second opinion on the changes to `each_location` & stack marking in it. Thanks!

eval.c
```diff
@@ -79,7 +79,7 @@ ruby_setup(void)
 #if defined(__linux__) && defined(PR_SET_THP_DISABLE)
     prctl(PR_SET_THP_DISABLE, 1, 0, 0, 0);
 #endif
-    Init_BareVM();
+    Init_BareVM(&state);
```
eightbitraptor, line 82: Similar here. Do we have to wind the address all the way down to `native_thread_init_stack`?

gc.c
```diff
@@ -6819,7 +6825,27 @@ mark_const_tbl(rb_objspace_t *objspace, struct rb_id_table *tbl)
...
-                                 const VALUE *stack_start, const VALUE *stack_end, void (*cb)(rb_objspace_t *, VALUE));
+                                 const VALUE *stack_start, const VALUE *stack_end, void *ctx,
+                                 void (*cb)(rb_objspace_t *, void *, VALUE));
+
+static void
+gc_mark_machine_stack_location_maybe(rb_objspace_t *objspace, void *ctx, VALUE obj)
+{
+    gc_mark_maybe(objspace, obj);
+#ifdef RUBY_ASAN_ENABLED
+    rb_execution_context_t *ec = ctx;
```
eightbitraptor, line 6836: The changes to `each_location` and the callback signature make sense to me. However, It seems a shame to introduce quite a lot of extra machinery in order to support drilling down a pointer that will only ever be used when `RUBY_ASAN_ENABLED`.

Have you looked at whether this has any impact on marking performance? Is there a way we could make this zero overhead if asan is not enabled?

gc.c
```diff
@@ -6819,7 +6825,27 @@ mark_const_tbl(rb_objspace_t *objspace, struct rb_id_table *tbl)
 #endif
 
 static void each_stack_location(rb_objspace_t *objspace, const rb_execution_context_t *ec,
-                                 const VALUE *stack_start, const VALUE *stack_end, void (*cb)(rb_objspace_t *, VALUE));
+                                 const VALUE *stack_start, const VALUE *stack_end, void *ctx,
+                                 void (*cb)(rb_objspace_t *, void *, VALUE));
+
+static void
+gc_mark_machine_stack_location_maybe(rb_objspace_t *objspace, void *ctx, VALUE obj)
```
eightbitraptor, line 6832: It'd be helpful for me if the intentional uses for `ctx` were documented in the sites where they were used. It's not immediately obvious what the intention of it is given that in the majority of callbacks it's unused.

thread_pthread.c
```diff
@@ -1964,9 +1965,10 @@ reserve_stack(volatile char *limit, size_t size)
 
 #undef ruby_init_stack
 void
-ruby_init_stack(volatile VALUE *addr)
+ruby_init_stack(volatile void *addr)
 {
     native_main_thread.id = pthread_self();
+    addr = asan_get_real_stack_addr((void *)addr);
```
eightbitraptor, line 1971: It's not obvious to me why we're using functions directly referencing asan in places where we don't know whether asan is available or enabled or not.

thread_pthread.c
```diff
@@ -2185,7 +2187,16 @@ native_thread_create_dedicated(rb_thread_t *th)
 static void
 call_thread_start_func_2(rb_thread_t *th)
 {
-    native_thread_init_stack(th);
+    /* Capture the address of a local in this stack frame to mark the beginning of the
+       machine stack for this thread. This is required even if we can tell the real
+       stack beginning from the pthread API in native_thread_init_stack, because
+       glibc stores some of its own data on the stack before calling into user code
+       on a new thread, and replacing that data on fiber-switch would break it (see
+       bug #13887) */
```
eightbitraptor, line 2195: Nice explanation! Thanks 👍

eval.c
```diff
@@ -70,7 +70,7 @@ ruby_setup(void)
     if (GET_VM())
         return 0;
 
-    ruby_init_stack((void *)&state);
+    ruby_init_stack(&state);
```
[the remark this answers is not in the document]
KJTsanaktsidis: The rationale here for the main thread is the same as for non-main threads:

https://github.com/ruby/ruby/pull/9505/files/09c83f44d2c66e9e432742d76df95d60afa5ebef#diff-28c449b5ef066b9e7b6e6f149211ba67bf320a633d32acac4117184c02f429e8R2190-R2195

I agree the type and variable name is a little strange. I renamed all of this punching through as `local_in_parent_frame` for the non-main thread case, perhaps I should do the same thing here as well (and duplicate the contents of that comment).

gc.c
KJTsanaktsidis: I think it would be safe to just stash a pointer to the execution context we're currently marking on `objspace`. That would be zero-cost if ASAN is not enabled, and avoid a lot of the messy diff here. I'll do that instead.

thread_pthread.c
KJTsanaktsidis: the pattern for most of the existing ASAN macros is for them to be no-ops if ASAN is disabled, and to elide the checks where they're used. I guess the motivation for this is to avoid spraying a lot of conditional compilation checks all throughout the codebase, like:

```
#if RUBY_ASAN_ENABLED
    void *poisoned = asan_unpoison_object_temporary(val)
#endif
   /* use val */
#if RUBY_ASAN_ENABLED
  if (poisoned) { asan_poison_object(val); }
#endif
```

For these more specialised macros only used in a couple of places though, perhaps it's better to be explicit and put the check here.

KJTsanaktsidis:
Thanks for the review 🙏  I'll make some adjustments.

KJTsanaktsidis:
@tenderlove or @peterzhu2118 - would either of you mind having a look at this as well as matt suggested? Thank you!

thread_pthread.c
```diff
@@ -1962,11 +1963,13 @@ reserve_stack(volatile char *limit, size_t size)
...
 #endif
 
-#undef ruby_init_stack
-void
-ruby_init_stack(volatile VALUE *addr)
+static void
+native_thread_init_main_thread_stack(void *addr)
 {
     native_main_thread.id = pthread_self();
+#ifdef RUBYASAN_ENABLED
```
peterzhu2118, line 1970: Typo:

```suggestion
#ifdef RUBY_ASAN_ENABLED
```

peterzhu2118 (approved):
I think this makes sense 👍

KJTsanaktsidis:
Awesome, thanks so much for your reviews.

PR #9506, opened 2024-01-12
Use native Windows commands instead of msys2

hsbt:
cat, sleep, stat is only provided msys, not native Windows. We should replace them for `mswin` environment.


hsbt:
How about `(Get-Item "C:\Path\To\File.txt").LastAccessTime` ?

PR #9526, opened 2024-01-13
Fix TODO in `hash.c` concerning `pair` values.

gazayas:
[st_update_callback_func](https://github.com/ruby/ruby/blob/55335eab80d763fb11d621c041d23aaf8f4857c6/include/ruby/st.h#L133) takes pointers for `key` and `val` for their first arguments:

```h
typedef int st_update_callback_func(st_data_t *key, st_data_t *value, st_data_t arg, int existing);
```

Although these were being passed here, this invocation was altering the values:

```c
VALUE key = pair->key;
VALUE val = pair->val;
retval = (*replace)(&key, &val, arg, TRUE);
```

Passing the pointer for the struct member directly ensures the value doesn't get altered.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -856,14 +856,7 @@ ar_general_foreach(VALUE hash, st_foreach_check_callback_func *func, st_update_c
                 return 0;
               case ST_REPLACE:
                 if (replace) {
-                    VALUE key = pair->key;
-                    VALUE val = pair->val;
-                    retval = (*replace)(&key, &val, arg, TRUE);
-
-                    // TODO: pair should be same as pair before.
-                    ar_table_pair *pair = RHASH_AR_TABLE_REF(hash, i);
-                    pair->key = key;
-                    pair->val = val;
+                    retval = (*replace)(&pair->key, &pair->val, arg, TRUE);
                 }
                 break;
               case ST_DELETE:


nobu:
Passing the pointers to members doesn't mean the struct pointer will not be clobbered.

gazayas:
Thank you @nobu, I'll try to remember that about pointers in the future!

In case it helps and if anyone wants to look into the same issue in the future, here is one of the test failures after removing the following lines and running `make test-spec`:

```c
// After removing the temporary fix from the TODO...
ar_table_pair *pair = RHASH_AR_TABLE_REF(hash, i);
pair->key = key;
pair->val = val;
```

Test error:
```
1)
Enumerable#tally counts values as gathered array when yielded with multiple arguments FAILED
Expected {nil=>2, 0=>1, [0, 1]=>2, [0, 1, 2]=>3, :default_arg=>1, []=>2, [0]=>1} == {nil=>[nil, nil],
 0=>[0],
 [0, 1]=>[[0, 1], [0, 1]],
 [0, 1, 2]=>[[0, 1, 2], [0, 1, 2], [0, 1, 2]],
 :default_arg=>[:default_arg],
 []=>[[], []],
 [0]=>[[0]]}
to be truthy but was false
/Users/gabriel_zayas/workdir/ruby/spec/ruby/core/enumerable/tally_spec.rb:25:in `block (2 levels) in <top (required)>'
/Users/gabriel_zayas/workdir/ruby/spec/ruby/core/enumerable/tally_spec.rb:4:in `<top (required)>'
```

With this PR `make check` passed for me locally, but there must be something I was missing, so I'll leave it be. Thanks again!

PR #9528, opened 2024-01-13
Correctly handle `MAP_FAILED` in `Init_default_shapes`.

ioquatix:
(Still waiting for bugs.ruby-lang.org issue to show up).

On low-memory systems (< ~400MB RAM+Swap), `Init_default_shapes` will fail and cause Ruby to segfault later.

Even something as simple as `ruby -e "puts 123"` will fail.

The latter `mmap` tries to allocate 384MB as a cache. This fails with `MAP_FAILED` and later this pointer is derefenced causing a segfault.

A better strategy should probably be used, e.g.

- Prefer dynamic allocation instead of mmap, OR
- If `mmap` fails, half the size of the cache and try again.

The latter approach might be problematic if the fixed cache size causes other (later) allocations to fail. I think that we should be extremely careful about allocating a large chunk of memory like this.


ioquatix:
Closing in favour of https://github.com/ruby/ruby/commit/8b8dcc7af174def5216044019c1d3e42edfdc7cf

PR #6687, opened 2022-11-08
Rewrite Array#each in Ruby

k0kubun:
Similarly to `Integer#times` https://github.com/ruby/ruby/pull/8388, this PR rewrites `Array#each` in Ruby.

### Interpreter
It's a bit slower than the original C method.

```
$ benchmark-driver benchmark/loop_each.yml -v --chruby 'before;after'
before: ruby 3.4.0dev (2024-01-13T00:28:26Z master f7178045bb) [x86_64-linux]
after: ruby 3.4.0dev (2024-01-13T04:31:50Z builtin-array-each 18c9a45314) [x86_64-linux]
Warming up --------------------------------------
           loop_each      2.129 i/s -       3.000 times in 1.408852s (469.62ms/i)
Calculating -------------------------------------
                         before       after
           loop_each      2.103       1.545 i/s -       6.000 times in 2.852494s 3.882698s

Comparison:
                        loop_each
              before:         2.1 i/s
               after:         1.5 i/s - 1.36x  slower
```

### YJIT

It's 5.3x faster.

```
$ benchmark-driver benchmark/loop_each.yml -v --chruby 'before::before --yjit-call-threshold=1;after::after --yjit-call-threshold=1'
before: ruby 3.4.0dev (2024-01-13T00:28:26Z master f7178045bb) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-01-13T04:31:50Z builtin-array-each 18c9a45314) +YJIT [x86_64-linux]
Warming up --------------------------------------
           loop_each      2.451 i/s -       3.000 times in 1.223915s (407.97ms/i)
Calculating -------------------------------------
                         before       after
           loop_each      2.456      13.074 i/s -       7.000 times in 2.850521s 0.535404s

Comparison:
                        loop_each
               after:        13.1 i/s
              before:         2.5 i/s - 5.32x  slower
```
```

PR #9533, opened 2024-01-14
Rewrite Array#each in Ruby using Primitive

k0kubun:
Same as https://github.com/ruby/ruby/pull/6687, but race condition-free.

### Interpreter
Thanks to `Primitive`, the interpreter doesn't slow down.

```
$ benchmark-driver benchmark/loop_each.yml -v --chruby 'before;after'
before: ruby 3.4.0dev (2024-01-20T01:12:07Z master 99d6e2f1ee) [x86_64-linux]
after: ruby 3.4.0dev (2024-01-20T01:36:08Z primitive-array-each 3aaa6de3ec) [x86_64-linux]
           loop_each      1.539 i/s -       2.000 times in 1.299363s (649.68ms/i)
                         before       after
           loop_each      1.551       1.796 i/s -       4.000 times in 2.578413s 2.227167s

Comparison:
                        loop_each
               after:         1.8 i/s
              before:         1.6 i/s - 1.16x  slower
```

### YJIT
Even faster than https://github.com/ruby/ruby/pull/6687.

```
$ benchmark-driver benchmark/loop_each.yml -v --chruby 'before::before --yjit-call-threshold=1;after::after --yjit-call-threshold=1'
before: ruby 3.4.0dev (2024-01-20T01:12:07Z master 99d6e2f1ee) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-01-20T01:36:08Z primitive-array-each 3aaa6de3ec) +YJIT [x86_64-linux]
           loop_each      1.789 i/s -       2.000 times in 1.118068s (559.03ms/i)
                         before       after
           loop_each      1.790      12.986 i/s -       5.000 times in 2.793497s 0.385044s

Comparison:
                        loop_each
               after:        13.0 i/s
              before:         1.8 i/s - 7.26x  slower
```


maximecb:
Nice. Happy to see this working. Surprised it's so far in YJIT as well!

Ideally, for YJIT, we'd like to be able to avoid doing a function call at all, so we can generate tight inlined code though, so we might need a specialized instruction instead of a cexpr? What do you think?

array.rb
```diff
@@ -1,4 +1,58 @@
...
+  #   Integer 2
+  #
+  # Related: #each_index, #reverse_each.
+  def each
+    unless block_given?
+      return to_enum(:each) { self.length }
+    end
+    i = 0
+    value = nil
+    while Primitive.cexpr!(%q{ ary_fetch_next(self, LOCAL_PTR(i), LOCAL_PTR(value)) })
```
nobu, line 50: https://github.com/ruby/ruby/actions/runs/7524693591/job/20479896887#step:8:21
> <internal:array>:48: warning: assigned but unused variable - i

```suggestion
    _i = 0
    value = nil
    while Primitive.cexpr!(%q{ ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value)) })
```

array.rb
```diff
@@ -1,4 +1,58 @@
...
+  #
+  #   Symbol foo
+  #   String bar
+  #   Integer 2
+  #
+  # Related: #each_index, #reverse_each.
+  def each
+    unless block_given?
+      return to_enum(:each) { self.length }
+    end
```
nobu, line 47:
```suggestion
    unless defined?(yield)
      return Primitive.cexpr!(%q{ SIZED_ENUMERATOR(self, 0, 0, ary_enum_length) })
    end
```

tool/mk_builtin_loader.rb
```diff
@@ -275,6 +275,7 @@ def generate_cexpr(ofile, lineno, line_file, body_lineno, text, locals, func_nam
     next unless Symbol === param
     next unless local_candidates.include?(param.to_s)
     f.puts "MAYBE_UNUSED(const VALUE) #{param} = rb_vm_lvar(ec, #{-3 - i});"
+    f.puts "MAYBE_UNUSED(VALUE *) #{param}__ptr = (VALUE *)&ec->cfp->ep[#{-3 - i}];"
```
nobu, line 278: - constify `#{param}__ptr` itself.
- dereference `*#{param}__ptr` and no `MAYBE_UNUSED` on it.
- adding `rb_vm_lvar_ptr` or something would be better.

```suggestion
    f.puts "VALUE *const #{param}__ptr = (VALUE *)&ec->cfp->ep[#{-3 - i}];"
    f.puts "MAYBE_UNUSED(const VALUE) #{param} = *#{param}__ptr;"
```

array.rb
eregon: IMO it's cleaner to use the idiomatic plain Ruby code for this (i.e. before this suggestion).
It's also good because it encourages JITs and the interpreter to optimize patterns used in Ruby gems & apps.

As more of the core library becomes written in Ruby I think it's valuable it looks like regular Ruby code whenever possible vs patching pieces of C code together.

maximecb: I agree with Benoit.

array.rb
```diff
@@ -1,4 +1,58 @@
...
+  # Related: #each_index, #reverse_each.
+  def each
+    unless defined?(yield)
+      return to_enum(:each) { self.length }
+    end
+    _i = 0
+    value = nil
+    while Primitive.cexpr!(%q{ ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value)) })
+      yield value
+    end
```
maximecb, line 52: I'm fully in support of what is being done with this PR.

The part that I don't like so much is that `Primitive.cexpr` seems very mysterious to me. It's not really the case that a C expression is being compiled by a C compiler here, is it? I guess the syntax of a C function call is being parsed somewhere in the Ruby parser? Unless this whole thing is being compiled into a function call as a pre-processing step? 

The other concern I have is, passing pointers to CRuby stack slots. I'm not sure how we can handle that in YJIT. Could we come up with another solution? Originally you had proposed having an annotation that simply says a function can't be interrupted?

k0kubun: > It's not really the case that a C expression is being compiled by a C compiler here, is it?

It is, if I understand your question correctly.

> I guess the syntax of a C function call is being parsed somewhere in the Ruby parser?

In Ruby's build process, `tool/mk_builtin_loader.rb` converts files like `array.rb` to serialized ISEQ binaries and a C source files like `array.rbinc`. `array.rbinc` contains something like this:

```c
#define LOCAL_PTR(local) local ## __ptr

static VALUE builtin_inline_class_50(struct rb_execution_context_struct *ec, const VALUE self)
{
    VALUE *const value__ptr = (VALUE *)&ec->cfp->ep[-3];
    MAYBE_UNUSED(const VALUE) value = *value__ptr;
    VALUE *const _i__ptr = (VALUE *)&ec->cfp->ep[-4];
    MAYBE_UNUSED(const VALUE) _i = *_i__ptr;

    return  ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value));
}
```

In the ISEQ binary for `Array#each`, `Primitive.cexpr!` is not a method call but `invokebuiltin` instruction that calls this C function `builtin_inline_class_50`.

So, `ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value))` is a valid C code, and is being compiled by a C compiler.

> The other concern I have is, passing pointers to CRuby stack slots. I'm not sure how we can handle that in YJIT. 

YJIT only needs to know an offset from EP for each local variable. ISEQ has a local table (`iseq->body->local_table`), and we can convert local variable names (`"_i"` and `"value"` in Rust) to EP offsets based on it.

> Could we come up with another solution? Originally you had proposed having an annotation that simply says a function can't be interrupted?

Right now, we have only ISEQ-level (i.e. not per-insn) annotations, and it's probably not a good idea to make a whole loop uninterruptible.

So we'd like to make a subset of instructions uninterruptible. But there's no easy way to encode that information in the instructions right now. `Primitive` (`invokebuiltin`) exists for doing something not possible in Ruby, so it's natural to use that for having a special behavior in part of the ISEQ.

maximecb: Ok, so the cexpr expressions get precompiled into C functions as part of a preprocessing step, which can then be called by CRuby.

> YJIT only needs to know an offset from EP for each local variable. ISEQ has a local table (iseq->body->local_table), and we can convert local variable names ("_i" and "value" in Rust) to EP offsets based on it.

That seems pretty hacky though. Not only do we need to parse and pattern-match on the cexpr primitives in YJIT, but we also need some way to expose EP offsets to a new primitive. The main issue that I see with this is that, as we get a more sophisticated register allocator, we want to be able to move away from the constraint that Ruby stack slots map to actual memory addresses. Local variables can live in a CPU register. If we have to force spill local variable slots so that we can refer to EP offsets, it adds more complexity to the JIT as well.

I think that your current solution works well in the context of the CRuby interpreter as it's implemented right now, but for the sake of the JIT, it would almost be better if we could introduce a new primitive that directly maps to a special YARV bytecode instruction. Or think of some other way to solve this problem.

test/ruby/test_gc_compact.rb
```diff
@@ -373,7 +373,7 @@ def add_ivars
 
       Fiber.new {
         $ary = OBJ_COUNT.times.map { Foo.new }
-        $ary.each(&:add_ivars)
+        $ary.reverse_each(&:add_ivars)
```
XrXr, line 376: Seems unnecessary to change this?

XrXr (approved):

test/ruby/test_gc_compact.rb
k0kubun: You're right. I remember it was failing at one point, but I can't reproduce that anymore. So I removed it https://github.com/ruby/ruby/pull/9533/commits/2e2cd3f60c5ff8a4b72414e96275b687045342f3.

array.rb
k0kubun: As we saw in a pairing session, I sketched out what we could do for this `Primitive` in YJIT https://github.com/k0kubun/ruby/pull/27. We can specialize `Primitive` of `Array#each` as if it was a C method. Since we can statically know the local variable offset for `_i` and `value`, it's no different from `getlocal` and `setlocal` to YJIT as long as we can specialize it that way.

PR #9540, opened 2024-01-15
[PRISM] Fix keyword splat inside of array

peterzhu2118:
Fixes ruby/prism#2155.


kddnewton (approved):
Is this the same thing that happens with call arguments or is it different? If it's the same maybe we should log a ticket to come back and share this.

PR #9543, opened 2024-01-15
YJIT: optimized codegen for String#length

maximecb:
Minor optimization for programs doing string processing.

@XrXr feel free to tell me if my assumptions wrt this function not needing special setup and always returning a fixnum on 64-bit machines are wrong.


XrXr (approved):
Looks good. It handles broken encoding without raising exceptions (see `spec/ruby/core/string/shared/length.rb`). And it's <= bytesize so we carry over the fixnum assumption.

PR #9548, opened 2024-01-16
Always enabled leakchecker by ruby/mspec

hsbt:
We should enable it to find https://github.com/ruby/ruby/commit/be7c91db44d6b8dba8fa11ff782965b4bfa0b3c8 on our dev box.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -43,6 +43,9 @@ RUBYLIB       = $(PATH_SEPARATOR)
 RUBYOPT       = -
 RUN_OPTS      = --disable-gems
 
+# Enabld leakcheckers by ruby/mspec
+CHECK_LEAKS = true
+
 # GITPULLOPTIONS = --no-tags
 
 PRISM_SRCDIR = $(srcdir)/prism


eregon:
@ko1 told me leakchecker is causing some problems, http://ci.rvm.jp/search search for `Leaked file descriptor:` and it's clear it started happening on Jan 16, with this commit + https://github.com/ruby/ruby/commit/01459f1644602ab7dfede46b07a56fc93ffa0165.
Also related: https://github.com/ruby/ruby/commit/cf8fb9429511068c69ce70da7f7818c7441c280b

While always-enabled CHECK_LEAKS=true is good to catch issues early, it's also significantly slower.
```
(with MAKEFLAGS=-j8)
With CHECK_LEAKS=true:
$ make test-spec
Finished in 15.357182 seconds
Finished in 15.079134 seconds

Without:
$ make test-spec
Finished in 11.669046 seconds
Finished in 11.702741 seconds
```

So I think the trade-off is not worth it, because it is fairly rare to have it fail, and enabling always is significantly slower for running test-spec locally.

So, what should we do?
I guess either:
* Stop setting `CHECK_LEAKS` in `spec/default.mspec` and only set it for some workflows, as before
* Add a new env var to skip file descriptor leak checks (and maybe other slow checks)
* Find out how to fix the failures. The Net::HTTP ones might be because of using `localhost` instead of `127.0.0.1` and that might trigger name resolution which might leak fds in libc.

Discussed in https://ruby.slack.com/archives/CR2QGFCAE/p1707447714985609

hsbt:
We discussed about this after DevMeeting. We should only enabled `ConstantsLeakCheckerAction`. 

In background, I enabled this because we couldn't find [this constant](https://github.com/ruby/ruby/commit/be7c91db44d6b8dba8fa11ff782965b4bfa0b3c8) leak at Ruby 3.3. release day.

eregon:
@hsbt Thanks, makes sense. I will make a PR for it.

eregon:
https://github.com/ruby/ruby/pull/9961

PR #9549, opened 2024-01-16
Make `const_source_location` return the real constant as soon as defined

casperisfine:
[[Bug #20188]](https://bugs.ruby-lang.org/issues/20188)

Ref: https://github.com/fxn/zeitwerk/issues/281#issuecomment-1893228355

Previously, it would only return the real constant location once the
autoload was fully completed.

### DIFF
--- a/spec/ruby/core/module/const_source_location_spec.rb
+++ b/spec/ruby/core/module/const_source_location_spec.rb
@@ -233,5 +233,17 @@ describe "Module#const_source_location" do
       line = ConstantSpecs::CONST_LOCATION
       ConstantSpecs.const_source_location('CONST_LOCATION').should == [file, line]
     end
+
+    ruby_bug("#20188", ""..."3.4") do
+      it 'returns the real constant location as soon as it is defined' do
+        file = fixture(__FILE__, 'autoload_const_source_location.rb')
+        ConstantSpecs.autoload :ConstSource, file
+        autoload_location = [__FILE__, __LINE__ - 1]
+
+        ConstantSpecs.const_source_location(:ConstSource).should == autoload_location
+        ConstantSpecs::ConstSource::LOCATION.should == ConstantSpecs.const_source_location(:ConstSource)
+        ConstantSpecs::BEFORE_DEFINE_LOCATION.should == autoload_location
+      end
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


spec/ruby/core/module/const_source_location_spec.rb
```diff
@@ -233,5 +233,17 @@
       line = ConstantSpecs::CONST_LOCATION
       ConstantSpecs.const_source_location('CONST_LOCATION').should == [file, line]
     end
+
+    ruby_bug("#20188", ""..."3.4") do
```
eregon, line 237:
```suggestion
    ruby_version_is "3.4" do # https://bugs.ruby-lang.org/issues/20188
```
`ruby_bug` is too much here, i.e. other Rubies shouldn't necessarily change this before they target 3.4, in fact it could maybe even be incompatible for some programs (likely not, but who knows).
In short `ruby_bug` is almost only to avoid CRuby segfaults.

casperisfine: My interpretation is that this is a bug, hence will be backported, but yes, perhaps it's not one, and perhaps it won't even be accepted as a feature. 

I'll make this change depending on how it goes with the issue.

eregon: It's true that if this would be decided to be backported then using `ruby_bug` is appropriate.

eregon:
To clarify this changes the behavior of const_source_location in the thread doing the autoload (during the autoload) but not for other threads, correct?
i.e. the constant's location is only visible to other threads after the autoload finished, i.e., at the same time the constant is published to these other threads.

casperisfine:
> this changes the behavior of const_source_location in the thread doing the autoload (during the autoload) but not for other threads, correct?

No. I made it to be consistent with the current behavior of `defined?` and `constants.include?` which both consider the constant defined from the other thread as soon as the constant was defined, not once the autoload clear.

The only thing that wait for the autoload to be cleared, is `.autoload?`, which makes sense.

eregon:
> No. I made it to be consistent with the current behavior of `defined?` and `constants.include?` which both consider the constant defined from the other thread as soon as the constant was defined, not once the autoload clear.

Mmh but that's incorrect, both `defined?` and `constants.include?` are always true for other threads whether only the autoload is registered or the constant is set to an actual value.

I really think `const_source_location` should be consistent with when the actual value of the constant is visible.
If not, it's likely to be broken for the assign+raise case and it lets other threads see the oscillation between two locations which seems quite wrong given autoload is supposed to be as much as possible transparent and thread-safe.

casperisfine:
Alright, I think I see what you mean now. I updated the code to check if the current fiber owns the autoload mutex.

nobu (approved):

PR #9550, opened 2024-01-16
Extract base64 as bundled gems

hsbt:
https://bugs.ruby-lang.org/issues/20187


SuperSandro2000:
Will the actual implementation in pack.c also move to the new repository?

PR #9552, opened 2024-01-16
Clear coderange when rb_str_resize change size

tompng:
Fixes https://bugs.ruby-lang.org/issues/20189

`rb_str_resize` was clearing coderange only when size shrinks
```c
if (slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) {
  ENC_CODERANGE_CLEAR(str);
}
```
But coderange should be cleared also when size is extended.

```
Example of coderange change with extend
"\x7f" (invalid utf16) → "\x7f\x00" (valid utf16)
"\x7f\x7f" (valid utf16) → "\x7f\x7f\x00" (invalid utf16)
```

Every existing encoding except utf16 utf32 does not end with `"\0"`.
```ruby
Encoding.list.select do |encoding|
  (0..0x10ffff).any? do |codepoint|
    bytes = codepoint.chr(encoding).bytes
    bytes != [0] && bytes.last == 0
  rescue
    false
  end
end
#=> [#<Encoding:UTF-16BE>, #<Encoding:UTF-16LE>, #<Encoding:UTF-32BE>, #<Encoding:UTF-32LE>]
```

Extending with \0 will change coderange only on these four encodings. These are the only encodings which is `TERM_LEN > 1`.

### DIFF
--- a/string.c
+++ b/string.c
@@ -3111,8 +3111,7 @@ rb_str_resize(VALUE str, long len)
     long slen = RSTRING_LEN(str);
     const int termlen = TERM_LEN(str);
 
-    if ((slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) ||
-        (termlen > 1 && (slen % termlen == 0) != (len % termlen == 0))) {
+    if (slen > len || (termlen != 1 && slen < len)) {
         ENC_CODERANGE_CLEAR(str);
     }
 
--- a/string.c
+++ b/string.c
@@ -3109,14 +3109,15 @@ rb_str_resize(VALUE str, long len)
 
     int independent = str_independent(str);
     long slen = RSTRING_LEN(str);
+    const int termlen = TERM_LEN(str);
 
-    if (slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) {
+    if ((slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) ||
+        (termlen > 1 && (slen % termlen == 0) != (len % termlen == 0))) {
         ENC_CODERANGE_CLEAR(str);
     }
 
     {
         long capa;
-        const int termlen = TERM_LEN(str);
         if (STR_EMBED_P(str)) {
             if (len == slen) return str;
             if (str_embed_capa(str) >= len + termlen) {
--- a/test/-ext-/string/test_set_len.rb
+++ b/test/-ext-/string/test_set_len.rb
@@ -63,4 +63,22 @@ class Test_StrSetLen < Test::Unit::TestCase
     assert_not_predicate str, :ascii_only?
     assert_equal u, str
   end
+
+  def test_valid_encoding_after_resized
+    s = "\0\0".force_encoding(Encoding::UTF_16BE)
+    str = Bug::String.new(s)
+    assert_predicate str, :valid_encoding?
+    str.resize(1)
+    assert_not_predicate str, :valid_encoding?
+    str.resize(2)
+    assert_predicate str, :valid_encoding?
+    str.resize(3)
+    assert_not_predicate str, :valid_encoding?
+
+    s = "\xDB\x00\xDC\x00".force_encoding(Encoding::UTF_16BE)
+    str = Bug::String.new(s)
+    assert_predicate str, :valid_encoding?
+    str.resize(2)
+    assert_not_predicate str, :valid_encoding?
+  end
 end
--- a/ext/-test-/string/set_len.c
+++ b/ext/-test-/string/set_len.c
@@ -16,9 +16,17 @@ bug_str_append(VALUE str, VALUE addendum)
     return str;
 }
 
+static VALUE
+bug_str_resize(VALUE str, VALUE len)
+{
+    rb_str_resize(str, NUM2LONG(len));
+    return str;
+}
+
 void
 Init_string_set_len(VALUE klass)
 {
     rb_define_method(klass, "set_len", bug_str_set_len, 1);
     rb_define_method(klass, "append", bug_str_append, 1);
+    rb_define_method(klass, "resize", bug_str_resize, 1);
 }


string.c
```diff
@@ -3110,7 +3110,7 @@ rb_str_resize(VALUE str, long len)
     int independent = str_independent(str);
     long slen = RSTRING_LEN(str);
 
-    if (slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) {
+    if (slen != len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) {
```
nobu, line 3113: With moving `termlen` from the following block to here, and:
```suggestion
    if ((slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) ||
        (termlen > 1 && (slen % termlen || len % termlen))) {
```

nobu:
```diff
diff --git i/ext/-test-/string/set_len.c w/ext/-test-/string/set_len.c
index 049da2cdb5c..b55ef6f4691 100644
--- i/ext/-test-/string/set_len.c
+++ w/ext/-test-/string/set_len.c
@@ -16,9 +16,17 @@ bug_str_append(VALUE str, VALUE addendum)
     return str;
 }
 
+static VALUE
+bug_str_resize(VALUE str, VALUE len)
+{
+    rb_str_resize(str, NUM2LONG(len));
+    return str;
+}
+
 void
 Init_string_set_len(VALUE klass)
 {
     rb_define_method(klass, "set_len", bug_str_set_len, 1);
     rb_define_method(klass, "append", bug_str_append, 1);
+    rb_define_method(klass, "resize", bug_str_resize, 1);
 }
diff --git i/test/-ext-/string/test_set_len.rb w/test/-ext-/string/test_set_len.rb
index e3eff75d9b9..ecd50083650 100644
--- i/test/-ext-/string/test_set_len.rb
+++ w/test/-ext-/string/test_set_len.rb
@@ -63,4 +63,13 @@
     assert_not_predicate str, :ascii_only?
     assert_equal u, str
   end
+
+  def test_valid_encoding_after_resized
+    s = "\0\0".force_encoding(Encoding::UTF_16BE)
+    str = Bug::String.new(s)
+    str.resize(1)
+    assert_not_predicate str, :valid_encoding?
+    str.resize(2)
+    assert_predicate str, :valid_encoding?
+  end
 end
```

string.c
tompng: Changed it to `(slen % termlen == 0) != (len % termlen == 0)` because I think it's more precise

test/-ext-/string/test_set_len.rb
```diff
@@ -63,4 +63,22 @@ def test_coderange_after_trunc
...
+    str.resize(2)
+    assert_predicate str, :valid_encoding?
+    str.resize(3)
+    assert_not_predicate str, :valid_encoding?
+
+    s = "\xDB\x00\xDC\x00".force_encoding(Encoding::UTF_16BE)
+    str = Bug::String.new(s)
+    assert_predicate str, :valid_encoding?
+    str.resize(2)
+    assert_not_predicate str, :valid_encoding?
```
tompng, line 82: This is a test case for shrinking surrogate pair that `len % termlen` is not changed but `valid_encoding?` is changed.
This is covered by `slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT` branch.

string.c
```diff
@@ -3109,14 +3109,15 @@ rb_str_resize(VALUE str, long len)
 
     int independent = str_independent(str);
     long slen = RSTRING_LEN(str);
+    const int termlen = TERM_LEN(str);
 
-    if (slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) {
+    if ((slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) ||
+        (termlen > 1 && (slen % termlen == 0) != (len % termlen == 0))) {
```
byroot, line 3115: I think you need to clear if `slen < len` (truncation) regardless.

I'd advocate for the simplest condition that catch the most common cases. I don't think the goal should be to never clear the coderange when it's theoretically avoidable, just to use fast and robust heuristics to avoid clearing it in the overwhelming majority of cases.

tompng: > I'd advocate for the simplest condition

I'd like to use `slen != len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT` because it's simple and respects previous optimization for 7bit coderange. (not updated the pull request to this condition yet)

I think there are about four options.

```c
// 1
slen != len
// Most robust

// 2
slen != len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT
// ENC_CODERANGE_7BIT means it's ascii compatible encoding and all bytes are <=127,
// so truncate and extend will never change coderange.

// 3
(slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) || (slen < len && termlen > 1)
// Need non obvious knowledge that every character(codepoint = 1..0x10ffff) in
every termlen==1 encodings currently defined in `Encoding.list` does not end with "\0".

// 4
(slen > len && ENC_CODERANGE(str) != ENC_CODERANGE_7BIT) || (termlen > 1 && (slen % termlen == 0) != (len % termlen == 0))
// Most efficient
// Need additional knowledge that every termlen > 1 encodings (utf16 and utf32) are fixed-width,
// and the latter part of surrogate pair(utf16) is not "\0\0" so extending by 2bytes does not change broken to valid.
```

What do you think?

byroot: I think the cases that are worth optimizing for are the very common encodings, namely: `ASCII`, `ASCII-BIT` and `UTF-8`, and in the increase case, which is common when appending strings.

Truncation is much rarer in my opinion, so not worth the complexity.

So I would advocate for:

```c
bool truncating = slen > len;
if (truncating || termlen != 1) {
  CLEAR_CODERANGE(...)
}
```

We lose the optimization for truncating ASCII-only strings, but I really doubt it's common enough to justify the complexity.

Of course I don't pretend to be an expert in this area, so I'm happy to defer to @nobu's opinion.

byroot: 👍

byroot:
@tompng anything blocking this PR?

tompng:
@byroot
IMO I don't think there is a blocker in the implementation.
Can you merge this if the change looks good?

composerinteralia:
This change ended up causing some test failures for us at GitHub. I'm still trying to wrap my head around how this all works, but here's a reproduction in the meantime in case anybody has ideas about whether this is an expected change in behavior.

```
ruby -rtempfile -e 'Encoding.default_internal = Encoding::UTF_8; Tempfile.open { _1.printf "%s", "\xE2".b }'
```

This passed on Ruby 3.3, but now it's an UndefinedConversionError.

tompng:
Looks like there is a bug in Ruby 3.3 fixed in 3.4. (I'm not sure if this pull request fixed it or another one did)
```ruby
Encoding.default_external = Encoding::US_ASCII
Encoding.default_internal = Encoding::UTF_8
s1 = "\xE2".b
s2 = '%s' % "\xE2".b
p s1 == s2 #=> true
p s1.ascii_only? == s2.ascii_only? #=> false, bug

Tempfile.open { _1.binmode.print s1 } # OK
Tempfile.open { _1.print s1 } #=> Raises Encoding::UndefinedConversionError
Tempfile.open { _1.print s2 } #=> Succeeds because s2.ascii_only? wrongly returns true
Tempfile.open { _1.printf '%s', s1 } # Succeeds
```
Therefore, I think `_1.printf "%s", "\xE2".b` and `_1.print "%s" % "\xE2".b` should fail just like `_1.print "\xE2".b` fails.

PR #9558, opened 2024-01-16
[DOC] RDoc for ARGF

BurdetteLamar:


io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *     ```
+ *     p ['ARGV', ARGV]
+ *     p ['ARGF.read', ARGF.read]
+ *     ```
+ *
+ * - Commands and outputs
+ *   (see below for the content of files `foo.txt` and `bar.txt`):
+ *
+ *     ```sh
+ *     echo "Open the pod bay doors, Hal." | ruby t.rb
```
peterzhu2118, line 14743: Nit: we should add a leading `$` to differentiate the command and the output.

```suggestion
 *     $ echo "Open the pod bay doors, Hal." | ruby t.rb
```

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *
+ * For any \ARGF read _except_ the {simplest case}[rdoc-ref:ARGF@Simplest+Case]
+ * (that is, _except_ for the <i>very first</i> \ARGF read with an empty `ARGV`),
+ * the sources are found in `ARGV`.
+ *
+ * \ARGF assumes that each element in array `ARGV` is a potential source,
+ * and is one of:
+ *
+ * - The string path to a file that may be opened as a stream.
+ * - The character `'-'`, meaning stream`$stdin`.
```
peterzhu2118, line 14779:
```suggestion
 * - The character `'-'`, meaning stream `$stdin`.
```

io.c
BurdetteLamar: I'd love to do this throughout, but the colorizer does not do even the first line.  Lemme no what you think.

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
- *     option = ARGV.shift #=> "--verbose"
- *     ARGV  #=> ["file1", "file2"]
+ * - **ARGV** may be thought of as the <b>argument vector</b> array.
  *
- * You can now use ARGF to work with a concatenation of each of these named
- * files. For instance, ARGF.read will return the contents of _file1_
- * followed by the contents of _file2_.
+ *     Initially, it contains the arguments and options
+ *     that are passed to it from the command line;
+ *     the running Ruby program can modify that array as it likes.
```
nobu, line 14713: What do "options" mean here?
It sounds like options to ruby interpreter itself.

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
- * After a file in +ARGV+ has been read ARGF removes it from the Array.
- * Thus, after all files have been read +ARGV+ will be empty.
+ * - **ARGF** may be thought of as the <b>argument files</b> object.
  *
- * You can manipulate +ARGV+ yourself to control what ARGF operates on. If
- * you remove a file from +ARGV+, it is ignored by ARGF; if you add files to
- * +ARGV+, they are treated as if they were named on the command line. For
- * example:
+ *     It can access file streams and/or the `$stdin` stream,
+ *     based on what it finds in `ARGV`.
```
nobu, line 14718:
```suggestion
 *     It can access file streams and/or the <tt>$stdin</tt> stream,
 *     based on what it finds in +ARGV+.
```

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
- * script. For example:
+ * \ARGF may read from _source_ streams,
+ * which at any particular time are determined by the content of `ARGV`.
  *
- *     $ echo "glark" | ruby -e 'p ARGF.read'
- *     "glark\n"
+ * ### Simplest Case
+ *
+ * When the <i>very first</i> \ARGF read occurs with an empty `ARGV` (`[]`),
+ * the source is `$stdin`:
```
nobu, line 14730:
```suggestion
 * When the <i>very first</i> \ARGF read occurs with an empty +ARGV+ (<tt>[]</tt>),
 * the source is <tt>$stdin</tt>:
```

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *
+ * - \File `t.rb`:
```
nobu, line 14732:
```suggestion
 * - \File +t.rb+:
```

io.c
```diff
@@ -14700,47 +14700,286 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *
+ *     ```
+ *     p ['ARGV', ARGV]
+ *     p ['ARGF.read', ARGF.read]
+ *     ```
+ *
+ * - Commands and outputs
+ *   (see below for the content of files `foo.txt` and `bar.txt`):
```
nobu, line 14740:
```suggestion
 *   (see below for the content of files +foo.txt+ and +bar.txt+):
```

io.c
nobu: RDoc is not markdown.

BurdetteLamar: The document begins with this:
```
 * Document-class:  ARGF
 *
 * :markup: markdown
 *
 * ## \ARGF and `ARGV`
```
So it is markdown.  Do you mean that it should not be markdown?

nobu: Ok, I missed it.

io.c
BurdetteLamar: I note that in this markdown, the syntax highlighting occurs only for the simplest of command lines (no options or pipe), and in one spot, not even then.  This is very disappointing, b/c my main reason for writing this revision in markdown was to get the highlighting.

I could do one of these:
- Add `$ ` prompts to the example command lines, in which case we might consider switching this document from markdown to rdoc.
- Separate each "Command and Output" section into two sections, "Command" and "Output."

peterzhu2118: I see. Maybe it's not too important to have syntax highlighting for shell commands and we can just switch back to RDoc again? Using RDoc over markdown would also make the documentation style more consistent with the rest of the code base.

BurdetteLamar: Prompts added to example shell commands;  format changed from markdown to rdoc.

io.c
BurdetteLamar: Rephrased.

peterzhu2118 (approved):

PR #9560, opened 2024-01-16
[Prism] Fix more method call argumnents

eileencodes:
In #2087 it was noted that there was a bug in the number of arguments in `SplatNode` and `KeywordHashNode`. I looked into this with Aaron before the linked PR was merged and we found a bunch of cases that weren't working quite right. This PR aims to fix some of those cases, but there may be more.

A splat argument followed by a positional argument will concat the array until the end or unless the argument is a kwarg or splat kwarg. For example

```
foo(a, *b, c, *d, e)
```

Will have an `argc` of 2, because `b`, `c`, `d`, and `e` will be concatenated together.

```
foo(a, *b, c, *d, **e)
```

Will have an `argc` of 3, because `b`, `c`, and `d` will be concatenated together and `e` is a separate argument.


eileencodes:
cc/ @tenderlove @kddnewton

PR #9562, opened 2024-01-16
Fix determing if option inside magic comments requires an arguement

mikhailnov:
Test:
```
$ echo '#!ruby -0 --disable' > file.rb
$ ruby file.rb
```
Ruby crashed with segmentation fault, because `is_option_with_arg("disable", Qtrue, Qtrue)` in ruby.c:1399 (per tree of commit 7b6731b1bb7c8fab72580f92450eea6e4cc3d943) returned true, while it must return false.

But
```
$ ruby -0 --disable
```
does not crash, `is_option_with_arg("disable", Qtrue, Qtrue)` in ruby.c:1399 correctly returns false.

I suspect that a good fix must be done somewhere else, but I cannot understand the difference between two cases described above, so just adding a check.

Not sure that this patch does not break anything.


mikhailnov:
Thanks!

PR #9564, opened 2024-01-16
YJIT: specialized codegen for integer right shift

maximecb:
Used in optcarrot. May also be used to write pure-Ruby gems. No overflow check or fixnum untagging required.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4754,10 +4754,6 @@ fn jit_rb_int_rshift(
 
     let comptime_shift = jit.peek_at_stack(&asm.ctx, 0);
 
-    if !comptime_shift.fixnum_p() {
-        return false;
-    }
-
     // Untag the fixnum shift amount
     let shift_amt = comptime_shift.as_isize() >> 1;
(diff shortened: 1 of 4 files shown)


XrXr (approved):

yjit/src/codegen.rs
```diff
@@ -4737,6 +4737,61 @@ fn jit_rb_int_lshift(
...
+    if asm.ctx.two_fixnums_on_stack(jit) != Some(true) {
+        return false;
+    }
+    guard_two_fixnums(jit, asm, ocb);
+
+    let comptime_shift = jit.peek_at_stack(&asm.ctx, 0);
+
+    if !comptime_shift.fixnum_p() {
+        return false;
+    }
```
k0kubun, line 4759:
```suggestion
```

`two_fixnums_on_stack` seems to have checked this already.

k0kubun, line 4759:
```suggestion
```

I think `return false` is unreachable because of the `two_fixnums_on_stack` check.

k0kubun (approved):

PR #9565, opened 2024-01-16
YJIT: Finer breakdown for block_arg counters

XrXr:
For example, on `lobsters` the block_arg reason was at ~10%. With this
change it shows that 6% of that is `send_cfunc_block_arg`.


k0kubun (approved):
:+1:

PR #9568, opened 2024-01-17
Hide public implementation of `rb_io`.

ioquatix:
https://bugs.ruby-lang.org/issues/19057


ioquatix:
We decided to try merging this.

https://github.com/ruby/dev-meeting-log/blob/master/2024/DevMeeting-2024-01-17.md#feature-19057-hide-implementation-of-rb_io_t-ioquatix

PR #9571, opened 2024-01-17
Don't zero out existing coverage of eval'ed code.

ioquatix:
https://bugs.ruby-lang.org/issues/19857


hsbt:
https://bugs.ruby-lang.org/issues/19857 is rejected.

PR #9576, opened 2024-01-17
Rewrite Numeric#times in Ruby using Primitive

nobu:


k0kubun:
Can you clarify the motivation? I don't think Matz asked for it https://bugs.ruby-lang.org/issues/20182#note-5:

> If the user intentionally redefines the fundamental method, it should be his/her own responsibility.

`Integer#<` and `Integer#succ` are two things that are already optimized well on the interpreter and YJIT, and replacing them with a random `cexpr!` would only complicate YJIT so much to maintain the current performance. This seems to hurt maintainability for no good reason.

nobu:
This was just an experiment, but I found that builtin-indexes were not working with `while`.
#9533 worked because it moved `i.succ` into `ary_fetch_next` by chance.
In that sense, this might be better to focus that bug and be earlier 2 commits only.

k0kubun:
> This was just an experiment, but I found that builtin-indexes were not working with while.
> In that sense, this might be better to focus that bug and be earlier 2 commits only.

It'd be nice to have those builtin index fixes :+1:

nobu:
> It'd be nice to have those builtin index fixes 👍

Are you OK with the restriction, `cexpr!` must be up to one per line?

k0kubun:
Yeah, I'm fine with it as long as it gives an error on conflict. RJIT is an exception, but it'd usually mean that you should merge such primitives.

PR #5179, opened 2021-11-25
YJIT: Introduce jit_putobject

jhawthorn:
This extracts the logic previously inside `gen_putobject` to a more reusable helper method `jit_putobject`.

The motivation for this is that it both simplifies the implementation of other instructions, and other instructions can reuse the optimized behaviour for 32-bit special constants (most importantly `opt_getinlinecache`).

This commit also expands the optimization to use a `mov` directly to memory when we encounter a 32-bit immediate constant. Previously it covered fixnums and `Qtrue`/`Qfalse`, now it will cover any `SPECIAL_CONST_P` value which can be represented as a 32-bit immediate. Notably, this includes static symbols, and Qnil.

### putobject with static symbol
```
:foo
```

**Before:**
```
  ; putobject
  55976964005d:  mov    eax, 0x84710c
  559769640062:  mov    qword ptr [rbx], rax
```

**After:**
```
  ; putobject
  55e4bb62105d:  mov    qword ptr [rbx], 0x84710c
```

### getinlinecache

```
FOO = 123

def get_foo
  FOO
end

10.times { get_foo }
```

**Before:**
```
  ; opt_getinlinecache
  55bc6656b11a:  mov    eax, 0xf7
  55bc6656b11f:  mov    qword ptr [rbx], rax
```

**After:**
```
  ; opt_getinlinecache
  55ab1a1f111a:  mov    qword ptr [rbx], 0xf7
```

PR #9581, opened 2024-01-17
Handle zero-like imaginary part as zero in to_r

mrkn:
Use imag.to_r when imag is not 0 or 0r.

Fixes [Bug #5179]

### DIFF
--- a/complex.c
+++ b/complex.c
@@ -1841,9 +1841,11 @@ nucomp_to_f(VALUE self)
  *
  *   Complex.rect(1, 0).to_r              # => (1/1)
  *   Complex.rect(1, Rational(0, 1)).to_r # => (1/1)
+ *   Complex.rect(1, 0.0).to_r            # => (1/1)
  *
  * Raises RangeError if <tt>self.imag</tt> is not exactly zero
- * (either <tt>Integer(0)</tt> or <tt>Rational(0, _n_)</tt>).
+ * (either <tt>Integer(0)</tt> or <tt>Rational(0, _n_)</tt>)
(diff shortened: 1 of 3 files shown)


complex.c
```diff
@@ -1851,8 +1851,11 @@ nucomp_to_r(VALUE self)
     get_dat1(self);
 
     if (!k_exact_zero_p(dat->imag)) {
-        rb_raise(rb_eRangeError, "can't convert %"PRIsVALUE" into Rational",
-                 self);
+        VALUE imag = rb_check_convert_type_with_id(dat->imag, T_RATIONAL, "Rational", idTo_r);
```
KJTsanaktsidis, line 1854: why convert `imag` to rational here just to pass it into `k_exact_zero_p()`? Wouldn't it be easier to just change the check `k_exact_zero_p(dat->imag)` to `f_zero_p(dat->imag)` instead?

mrkn: @KJTsanaktsidis This approach doesn’t impose the responsibility of deciding what exactly constitutes zero on the Complex class. Instead, it assigns the responsibility to each individual class.

mrkn: And this implementation is for proof of concept. Before merging this to the master branch, I’ll add the optimization for Float class later.

KJTsanaktsidis: Ah that makes sense - thanks for the explanation!

mrkn:
Waiting for https://github.com/ruby/prism/issues/2468.

kddnewton:
Sorry about the delay @mrkn, you should be able to rebase and it should pass. Please let me know if you run into any more trouble!

PR #9582, opened 2024-01-17
[PRISM] Fix case splat with no predicate

eightbitraptor:
Fixes https://github.com/ruby/prism/issues/2199

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -3514,10 +3514,11 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
                 pm_node_t *condition_node = when_node->conditions.nodes[i];
 
                 if (PM_NODE_TYPE_P(condition_node, PM_SPLAT_NODE)) {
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
This looks right for this example, but I'm surprised that this was in here in the first place. Is there some other case we're not considering where the splatarray is necessary?

eightbitraptor:
> This looks right for this example, but I'm surprised that this was in here in the first place. Is there some other case we're not considering where the splatarray is necessary?

I've only been routinely running `make btest` and `make test/ruby/test_compile_prism.rb` but I haven't seen anything yet that relies on the `splatarray`. That being said, it does look like switching `VM_CHECKMATCH_TYPE_CASE` isn't as simple as it seems.

Marking this as draft again for now while I investigate.

eightbitraptor:
After more investigation, it looks like the value of VM_CHECKMATCH_TYPE varies depending on whether a `predicate` is present. This is represented in the CRuby AST as 2 seperate node types: `NODE_CASE`, and `NODE_CASE2`.

And with this current patch `make -s btest BTESTS="bootstraptest/test_syntax.rb" RUN_OPTS=--parser=prism` is showing only 2 remaining failures, neither of which are `case` related.

kddnewton (approved):

kddnewton:
Nice investigation!

PR #9586, opened 2024-01-17
[PRISM] Refactor keyword hash nodes 

peterzhu2118:
Follow up to https://github.com/ruby/ruby/pull/9540.


kddnewton (approved):
Yeah looks great

PR #9587, opened 2024-01-17
Remove null checks for xfree

peterzhu2118:
xfree can handle null values, so we don't need to check it.

### DIFF
--- a/regparse.c
+++ b/regparse.c
@@ -763,10 +763,9 @@ names_clear(regex_t* reg)
 	e->back_refs = (int* )NULL;
       }
     }
-    if (IS_NOT_NULL(t->e)) {
-      xfree(t->e);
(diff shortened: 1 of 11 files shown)


tenderlove (approved):
Seems fine to me 🤷‍♀️

regparse.c
```diff
@@ -763,9 +763,9 @@ names_clear(regex_t* reg)
 	e->back_refs = (int* )NULL;
       }
     }
-    if (IS_NOT_NULL(t->e)) {
-      xfree(t->e);
-      t->e = NULL;
+
+    xfree(t->e);
+    t->e = NULL;
     }
```
XrXr, line 769:
```suggestion
```

We don't build this code, but you need to remove this brace to remove the if wrapping.

PR #9588, opened 2024-01-17
[PRISM] Implement remaining nodes for `defined?` that use `expression`

eileencodes:
This PR implements the remaining nodes for `defined?`. All of them emit the `expression` instruction. Note that `MatchWriteNode` is slightly different as it also has a local table (see commit message for https://github.com/ruby/ruby/commit/b1473db423c695a47055b1d4eb937f928cba8598).

There are 4 nodes left in ruby/prism#2188 after this PR but I have determined all of them are not possible to call `defined?` on since they throw a `SyntaxError`. So this is the final PR for `defined?` to close ruby/prism#2188.

This PR implements the following nodes:

- `PM_UNTIL_NODE`
- `PM_WHILE_NODE`
- `PM_SINGLETON_CLASS_NODE`
- `PM_RATIONAL_NODE`
- `PM_MODULE_NODE`
- `PM_MATCH_REQUIRED_NODE`
- `PM_MATCH_WRITE_NODE`
- `PM_IF_NODE`
- `PM_FOR_NODE`
- `PM_DEF_NODE`
- `PM_CLASS_NODE`
- `PM_CASE_MATCH_NODE`
- `PM_CASE_NODE`

cc/ @kddnewton

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -2546,6 +2546,7 @@ pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *co
       case PM_KEYWORD_HASH_NODE:
       case PM_LAMBDA_NODE:
       case PM_MATCH_PREDICATE_NODE:
+      case PM_MATCH_WRITE_NODE:
       case PM_NEXT_NODE:
       case PM_OR_NODE:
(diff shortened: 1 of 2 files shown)


kddnewton (approved):
Nice!

PR #9595, opened 2024-01-18
[PRISM] Add function to free scope node

peterzhu2118:
pm_scope_node_destroy frees the scope node after we're done using it to make sure that the index_lookup_table is not leaked.

For example:

```ruby
10.times do
  100_000.times do
    RubyVM::InstructionSequence.compile_prism("begin; 1; rescue; 2; end")
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

    33056
    50304
    67776
    84544
    101520
    118448
    135712
    152352
    169136
    186656

After:

    15264
    15296
    15408
    17040
    17152
    17152
    18320
    18352
    18400
    18608

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -2476,6 +2476,14 @@ pm_scope_node_init(const pm_node_t *node, pm_scope_node_t *scope, pm_scope_node_
     }
 }
 
+void
+pm_scope_node_destroy(pm_scope_node_t *scope_node)
+{
+    if (scope_node->index_lookup_table) {
+        st_free_table(scope_node->index_lookup_table);
+    }
+}
+
 static void pm_compile_call(rb_iseq_t *iseq, const pm_call_node_t *call_node, LINK_ANCHOR *const ret, const uint8_t *src, bool popped, pm_scope_node_t *scope_node, ID method_id, LABEL *start);
 
 void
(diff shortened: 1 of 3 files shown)


prism_compile.c
```diff
@@ -6410,6 +6432,9 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
 
         // We set the index_lookup_table on the scope node so we can
         // refer to the parameters correctly
+        if (scope_node->index_lookup_table) {
+            st_free_table(scope_node->index_lookup_table);
+        }
```
kddnewton, line 6437:
```suggestion
        pm_scope_node_destroy(scope_node);
```

prism_compile.c
```diff
@@ -6898,12 +6924,14 @@ rb_translate_prism(pm_parser_t *parser, rb_iseq_t *iseq, pm_scope_node_t *scope_
     for (size_t i = 0; i < locals->size; i++) {
         st_insert(index_lookup_table, locals->ids[i], i);
     }
+
+    if (scope_node->index_lookup_table) {
+        st_free_table(scope_node->index_lookup_table);
+    }
```
kddnewton, line 6930:
```suggestion
    pm_scope_node_destroy(scope_node);
```

kddnewton (approved):
This looks good, but I added a couple of suggestions that I _think_ are correct?

prism_compile.c
peterzhu2118: I specifically did not use `pm_scope_node_destroy` here because we're not destroying the scope node, but rather only overwriting the `index_lookup_table`. If we freed other things in `pm_scope_node_destroy`, we don't want those to get freed here.

kddnewton: Got it, that makes sense!

PR #9596, opened 2024-01-18
YJIT: Drop extra arguments passed by yield

XrXr:
Support dropping extra arguments passed by `yield` in blocks. For
example `10.times { work }` drops the count argument. This is common
enough that it's about 3% of fallback reasons in `lobsters`.

Only support simple cases where the surplus arguments are at the top of
the stack, that way they just need to be popped, which take no work.


k0kubun (approved):
Nice, I've wanted this feature.

maximecb (approved):

XrXr:
The SEGV with `pp` is legit and uncovered #9601 (separate since it needs a backport). Should go away once that fix lands.

XrXr:
Had to amend the logic for autosplat a bit to match up with `setup_parameters_complex()` to pass the test I added, but it should be good now.

PR #9597, opened 2024-01-18
YJIT: Stop incrementing chain_depth on defer_compilation

k0kubun:
This PR proposes to isolate a bit for double-defer checks as discussed in https://github.com/ruby/ruby/pull/9426#discussion_r1445028453. 

Currently, `defer_compilation` increments `chain_depth` to check if the block is doubly deferred. It results in confusion like this https://github.com/ruby/ruby/pull/9426#discussion_r1445013315, and unnecessarily decrements the number of chains allowed for the instruction from the max depth constant, which is surprising. As I'm working on megamorphic callsites, this has been too confusing for me.

We use up to 20 chains in a single instruction, so reducing the max chain_depth from 127 to 63 seems fine.

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -464,7 +464,7 @@ pub struct Context {
     /// - 1st bit from the left: Whether this code is the target of a JIT-to-JIT Ruby return ([Self::is_return_landing])
     /// - 2nd bit from the left: Whether the compilation of this code has been deferred ([Self::is_deferred])
     /// - Last 6 bits (max: 63): Depth of this block in the sidechain (eg: inline-cache chain)
-    chain_depth_return_landing_defer: u8,
+    chain_depth_and_flags: u8,
 
     // Type we track for self
     self_type: Type,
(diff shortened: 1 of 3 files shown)


yjit/src/core.rs
```diff
@@ -1704,30 +1712,39 @@ impl Context {
...
-        self.chain_depth_return_landing &= 0x7f;
+        self.chain_depth_return_landing_defer &= !RETURN_LANDING_BIT;
     }
 
     pub fn is_return_landing(&self) -> bool {
-        self.chain_depth_return_landing & 0x80 > 0
+        self.chain_depth_return_landing_defer & RETURN_LANDING_BIT != 0
+    }
+
+    pub fn set_as_deferred(&mut self) {
```
maximecb, line 1742:
```suggestion
    pub fn mark_as_deferred(&mut self) {
```

yjit/src/core.rs
```diff
@@ -456,10 +461,10 @@ pub struct Context {
...
 
     /// Fields packed into u8
-    /// - Lower 7 bits: Depth of this block in the sidechain (eg: inline-cache chain)
-    /// - Top bit: Whether this code is the target of a JIT-to-JIT Ruby return
-    ///   ([Self::is_return_landing])
-    chain_depth_return_landing: u8,
+    /// - 1st bit from the left: Whether this code is the target of a JIT-to-JIT Ruby return ([Self::is_return_landing])
+    /// - 2nd bit from the left: Whether the compilation of this code has been deferred ([Self::is_deferred])
+    /// - Last 6 bits (max: 63): Depth of this block in the sidechain (eg: inline-cache chain)
+    chain_depth_return_landing_defer: u8,
```
maximecb, line 467: May want to rename this field to just flags or something like that?

maximecb (approved):

yjit/src/core.rs
k0kubun: How about `chain_depth_and_flags`? https://github.com/ruby/ruby/pull/9597/commits/f2113c4df37ed69d86a6f72c4f71e3d63a93673e

maximecb: 👌

PR #9599, opened 2024-01-18
YJIT: Optimize defined?(yield)

k0kubun:
This PR optimizes `defined` instruction for the `DEFINED_YIELD` case, which is used by `defined?(yield)`.

It has been raised as an idea to use `defined?(yield)` instead of `block_given?` when we rewrite a C method in Ruby so that it will not be impacted by method redefinition.

Matz didn't ask for tolerating such redefinition, so we don't need to do it for compatibility. However, it could be still useful for performance when we rewrite C methods in Ruby since `defined?(yield)` is not a method call unlike `block_given?`, so it's faster on the interpreter.

Even for YJIT, `defined` instruction is more preferable because you don't need to split blocks for `defer_compilation` (smaller memory overhead) or add a guard for the receiver (faster).

Since we're interested in removing C -> Ruby calls, we would naturally rewrite methods that check `block_given?`. To make it easier for us to merge such changes, being able to use this optimization for the interpreter performance seems useful.

### block_given? 

```asm
  # Insn: 0001 opt_send_without_block (stack_size: 1)
  # call to Object#block_given?
  # guard known object with singleton class
  0x559640a680bf: movabs rax, 0x7f9f7f6ace00
  0x559640a680c9: cmp rsi, rax
  0x559640a680cc: jne 0x559640a6a0aa
  # block_given?
  0x559640a680d2: mov rax, qword ptr [r13 + 0x20]
  0x559640a680d6: mov rax, qword ptr [rax - 8]
  0x559640a680da: test rax, rax
  0x559640a680dd: mov eax, 0x14
  0x559640a680e2: mov ecx, 0
  0x559640a680e7: cmove rax, rcx
  0x559640a680eb: mov rsi, rax
```

### defined?(yield)

```asm
  # Insn: 0001 defined (stack_size: 1)
  # block_given?
  0x559640a6805d: mov rax, qword ptr [r13 + 0x20]
  0x559640a68061: mov rax, qword ptr [rax - 8]
  0x559640a68065: test rax, rax
  0x559640a68068: movabs rax, 0x7f9f7f671940
  0x559640a68072: mov rax, rax
  0x559640a68075: mov ecx, 4
  0x559640a6807a: cmove rax, rcx
  0x559640a6807e: mov rsi, rax
```

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4357,3 +4357,9 @@ assert_equal '[2, 4611686018427387904]', %q{
 assert_equal '[0, 1, -4]', %q{
   [0 >> 1, 2 >> 1, -7 >> 1]
 }
+
+# Integer right shift
+assert_equal '[nil, "yield"]', %q{
+  def defined_yield = defined?(yield)
+  [defined_yield, defined_yield {}]
+}
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4358,7 +4358,6 @@ assert_equal '[0, 1, -4]', %q{
   [0 >> 1, 2 >> 1, -7 >> 1]
 }
 
-# Integer right shift
 assert_equal '[nil, "yield"]', %q{
   def defined_yield = defined?(yield)
   [defined_yield, defined_yield {}]
--- a/yjit/bindgen/src/main.rs
+++ b/yjit/bindgen/src/main.rs
@@ -348,6 +348,7 @@ fn main() {
         .allowlist_function("rb_iseqw_to_iseq")
         .allowlist_function("rb_iseq_label")
         .allowlist_function("rb_iseq_line_no")
+        .allowlist_type("defined_type")
 
         // From builtin.h
         .allowlist_type("rb_builtin_function.*")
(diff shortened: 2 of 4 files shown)


XrXr (approved):
Neat method-call free way to do block_given? without taking a block param!

jeremyevans:
Sequel and Roda have both been using `defined?(yield)` instead of `block_given?` for a couple years now, so I'm definitely in favor of further optimizing it.

luke-gru:
Is it not possible to replace block_given? with defined(yield) in the peephole optimizer? I guess a user could have redefined the method, but why would they want to?

XrXr:
> Is it not possible to replace block_given? with defined(yield) in the peephole optimizer? I guess a user could have redefined the method, but why would they want to?

We can't make sneaky language semantics changes like that, it's too surprising for users and too hard to discover. Also, there is already many redefinitions: https://github.com/search?q=%22def+block_given%3F%22&type=code

maximecb (approved):
I agree with Alan. This is neat. Good thinking :)

byroot:
For people stumbling on this and thinking to write a `rubocop-performance` Cop for it, I beg you not to.

PR #9613, opened 2024-01-19
Add an autoload spec for the behavior inside autoload but after the constant is defined

casperisfine:
From discussion on https://bugs.ruby-lang.org/issues/20188

cc @fxn @eregon

### DIFF
--- a/spec/ruby/core/module/autoload_spec.rb
+++ b/spec/ruby/core/module/autoload_spec.rb
@@ -343,6 +343,29 @@ describe "Module#autoload" do
     end
   end
 
+  def check_before_during_thread_after(const, &check)
+    before = check.call
+    to_autoload_thread, from_autoload_thread = Queue.new, Queue.new
+    ScratchPad.record -> {
+      from_autoload_thread.push check.call
+      to_autoload_thread.pop
+    }
+    t = Thread.new {
+      in_loading_thread = from_autoload_thread.pop
+      in_other_thread = check.call
+      to_autoload_thread.push :done
+      [in_loading_thread, in_other_thread]
+    }
+    in_loading_thread, in_other_thread = nil
+    begin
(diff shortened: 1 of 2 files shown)


spec/ruby/core/module/autoload_spec.rb
```diff
@@ -403,6 +403,66 @@ def check_before_during_thread_after(&check)
...
+      begin
+        ModuleSpecs::Autoload::DuringAutoload
+      ensure
+        in_loading_thread, in_other_thread = t.value
+      end
+      after = check.call
+      [before, in_loading_thread, in_other_thread, after]
+    end
+
+    it "returns nil in autoload thread and 'constant' otherwise for defined?" do
```
eregon, line 437: This spec description needs to be adapted to reflect the results in this case

spec/ruby/core/module/autoload_spec.rb
```diff
@@ -403,6 +403,66 @@ def check_before_during_thread_after(&check)
...
+    end
+
+    it "keeps the constant in Module#constants" do
+      results = check_before_during_thread_after {
+        ModuleSpecs::Autoload.constants(false).include?(:DuringAutoload)
+      }
+      results.should == [true, true, true, true]
+    end
+
+    it "returns false in autoload thread and true otherwise for Module#const_defined?" do
```
eregon, line 451: This spec description needs to be adapted to reflect the results in this case

eregon:
BTW it would be great to also test the case of the constant is set and then an exception happens in the autoload.

casperisfine:
I fixed the naming issues.

I'll look into the error thing but no promises.

eregon (approved):

fxn:
❤️ I'll have a stab at the spec for the error.

PR #9619, opened 2024-01-19
[PRISM] Fix ensure code running twice

eileencodes:
Fixes: ruby/prism#2212


kddnewton (approved):
Looks good, but @tenderlove can you take a look?

PR #9621, opened 2024-01-19
YJIT: Fix ruby2_keywords edge case and drop bogus checks

XrXr:
YJIT didn't guard for ruby2_keywords hash in case of splat calls that
land in methods with a rest parameter, creating incorrect results.

The compile checks didn't correspond to any actual effects of
ruby2_keywords, so it was masking this bug and YJIT was needlessly
refusing to compile some code. About 16% of fallback reasons in
`lobsters` was due to the ISeq check.

We already handle the tagging part with
exit_if_supplying_kw_and_has_no_kw() and should now have a dynamic guard
for all splat cases.

Note for backporting: You also need https://github.com/ruby/ruby/commit/7f51959ff14fbe06bc1afd283d1af17b26161cf4.

https://bugs.ruby-lang.org/issues/20195


XrXr:
The PRISM failure is unrelated and should be fixed by #9662

maximecb (approved):

yjit/src/codegen.rs
```diff
@@ -6014,6 +6002,23 @@ fn push_splat_args(required_args: u32, asm: &mut Assembler) {
...
+c_callable! {
+    /// Return non-zero when `obj` is an array and its last item is a
+    /// `ruby2_keywords` hash. We don't support this kind of splat.
+    fn splat_array_ends_in_ruby2_keywords(obj: VALUE) -> usize {
+        unsafe {
+            if !RB_TYPE_P(obj, RUBY_T_ARRAY) {
+                return 0;
+            }
+            let last = rb_ary_entry_internal(obj, -1);
+            if !RB_TYPE_P(last, RUBY_T_HASH) {
```
k0kubun, line 6014: This `RB_TYPE_P` is defined in yjit.c, so it's a couple of C function calls. Since it's used for rest & splat calls at runtime, can we either import the `RB_TYPE_P` definition to Rust or just write this function in C?

k0kubun: Same applies to `rb_ary_entry_internal`.

yjit.c
```diff
@@ -871,6 +871,19 @@ rb_yjit_fix_mod_fix(VALUE recv, VALUE obj)
     return rb_fix_mod_fix(recv, obj);
 }
 
+// Return non-zero when `obj` is an array and its last item is a
+// `ruby2_keywords` hash. We don't support this kind of splat.
+size_t
```
k0kubun, line 876: It shouldn't matter, but any reason why this is `size_t`? `FL_TEST_RAW` seems to return `VALUE`.

k0kubun (approved):

yjit.c
XrXr: I just wanted to make sure it zeros out the full register. VALUE indicates returning an object to me so I went with `size_t`. I could change it if you want.

PR #9622, opened 2024-01-20
YJIT: Allow inlining ISEQ calls with a block

k0kubun:
This PR introduces an annotation `Primitive.attr! :inline_block` to allow monomorphizing `yield` in methods like `Array#each` https://github.com/ruby/ruby/pull/9533.

## yjit-bench
This PR or `Array#each` in Ruby (https://github.com/ruby/ruby/pull/9533) doesn't have a significant impact on benchmarks by itself.

However, combining this PR and `Array#each` in Ruby with the annotation ([branch](https://github.com/k0kubun/ruby/tree/yjit-inline-block-array-each)), it gives the following speedups.

* `hexapdf` is 7% faster
* `chunky-png` and `ruby-lsp` are 2% faster
* `activerecord`, `liquid-c`, and `railsbench` are 1% faster

```
before: ruby 3.4.0dev (2024-01-20T01:12:07Z master 99d6e2f1ee) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-01-20T04:37:44Z yjit-inline-block-.. ae09d8f7a5) +YJIT [x86_64-linux]

activerecord    54.1         0.6         53.8        0.7         1.01           1.01
chunky-png      558.9        0.2         550.4       0.6         1.02           1.02
erubi-rails     1610.9       0.9         1611.4      0.9         1.01           1.00
hexapdf         2035.2       1.0         1895.8      2.0         0.99           1.07
liquid-c        71.3         1.3         70.7        0.9         1.00           1.01
liquid-compile  72.2         0.3         72.2        0.4         0.99           1.00
liquid-render   96.2         0.4         96.4        1.1         1.00           1.00
lobsters        945.4        1.1         941.4       1.0         0.99           1.00
mail            145.3        0.4         145.4       1.6         1.00           1.00
psych-load      1912.1       0.2         1921.0      0.2         0.99           1.00
railsbench      2066.7       0.2         2043.3      0.1         0.99           1.01
ruby-lsp        57.5         26.2        56.5        27.2        1.00           1.02
sequel          86.3         0.6         86.1        1.3         1.00           1.00
```

See also: the number of Ruby block calls from `Array#each` https://github.com/Shopify/yjit-bench/pull/168

## Memory usage
Running 25 itrs on `lobsters` with this PR and `Array#each` using the annotation ([branch](https://github.com/k0kubun/ruby/tree/yjit-inline-block-array-each)), `code_region_size` was increased by 1.2%, and `yjit_alloc_size` was increased by 4.5%.

For optimizing the size of `Context`, I'm thinking of having a global `HashMap` like `{ iseq1 => 1, iseq2 => 2, ... }` and store an `u16` integer instead of an `u64` ISEQ pointer as a next step, which would likely decrease `yjit_alloc_size`.

### Before

```
code_region_size:         15,024,128
yjit_alloc_size:          27,201,208
```

### After

```
code_region_size:         15,204,352
yjit_alloc_size:          28,414,571
max_inline_versions:             121
```

## microbenchmark
This PR makes the synthetic benchmark of megamorphic `yield` in `Integer#times` 4.5x faster.

```
$ benchmark-driver benchmark/loop_times_megamorphic.yml --chruby 'before::before --yjit-call-threshold=1;after::after --yjit-call-threshold=1'
loop_times_megamorphic    18.810k i/s -     20.004k times in 1.063473s (53.16μs/i)
                           before       after
loop_times_megamorphic    17.820k     79.926k i/s -     56.430k times in 3.166588s 0.706025s

Comparison:
             loop_times_megamorphic
                 after:     79926.3 i/s
                before:     17820.4 i/s - 4.49x  slower
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1810,7 +1810,7 @@ fn gen_get_ep(asm: &mut Assembler, level: u32) -> Opnd {
 
 // Gets the EP of the ISeq of the containing method, or "local level".
 // Equivalent of GET_LEP() macro.
-fn gen_get_lep(jit: &mut JITState, asm: &mut Assembler) -> Opnd {
+fn gen_get_lep(jit: &JITState, asm: &mut Assembler) -> Opnd {
     // Equivalent of get_lvar_level() in compile.c
     fn get_lvar_level(iseq: IseqPtr) -> u32 {
         if iseq == unsafe { rb_get_iseq_body_local_iseq(iseq) } {
@@ -6876,6 +6876,12 @@ fn gen_send_iseq(
     // Create a context for the callee
     let mut callee_ctx = Context::default();
 
+    // If the callee has :inline_block annotation and the callsite has a block ISEQ,
+    // duplicate a callee block for each block ISEQ to make its `yield` monomorphic.
+    if let (Some(BlockHandler::BlockISeq(iseq)), true) = (block, builtin_attrs & BUILTIN_ATTR_INLINE_BLOCK != 0) {
+        callee_ctx.set_inline_block(iseq);
+    }
+
     // Set the argument types in the callee's context
     for arg_idx in 0..argc {
         let stack_offs: u8 = (argc - arg_idx - 1).try_into().unwrap();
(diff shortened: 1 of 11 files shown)


maximecb:
I don't quite understand how this PR works. Please bear with me. The code for send is so complex that I find it hard to follow sometimes.

You say it's duplicating the ISEQ for each block. Conceptually, I understand what that means. However, I don't understand why this has an impact on the maximum number of versions in the JIT.

Are multiple entry blocks being generated for the ISEQs based on the block given in the call?

I guess I'm wondering: would it be possible to somehow duplicate the ISEQ directly at the interpreter level? That might be slightly less complex on the JIT side, and it would avoid having to add extra state to the context?

maximecb (approved):

yjit/src/codegen.rs
```diff
@@ -6876,6 +6876,12 @@ fn gen_send_iseq(
     // Create a context for the callee
     let mut callee_ctx = Context::default();
 
+    // If the callee has :inline_block annotation and the callsite has a block ISEQ,
+    // duplicate a callee block for each block ISEQ to make its `yield` monomorphic.
+    if let (Some(BlockHandler::BlockISeq(iseq)), true) = (block, builtin_attrs & BUILTIN_ATTR_INLINE_BLOCK != 0) {
+        callee_ctx.set_inline_block(iseq);
```
XrXr, line 6882: I guess this is the magic sauce. Because the callee_ctx is probably doesn't match anything that currently exists, it's going to generate a new block directly following the call.

XrXr (approved):

PR #9623, opened 2024-01-20
Add ARGF native methods

BurdetteLamar:
Adds native code for methods #putc, #puts, #print, #printf, which formerly called corresponding methods in class IO;  this meant that the doc for the methods in ARGF came from IO.

This change will mean that the methods in ARGF can have their own documentation (not yet added).

This is a draft b/c I don't know how to run the test with the modified C code.  @peterzhu2118, if you can test this and it's ok, I will add the documentation.

### DIFF
--- a/io.c
+++ b/io.c
@@ -8557,6 +8557,12 @@ rb_io_printf(int argc, const VALUE *argv, VALUE out)
     return Qnil;
 }
 
+VALUE
+argf_printf(int argc, const VALUE *argv, VALUE out)
+{
+  rb_io_printf(argc, argv, out)
+}
+
 /*
  *  call-seq:
  *    printf(format_string, *objects)               -> nil
@@ -8704,6 +8710,12 @@ rb_io_print(int argc, const VALUE *argv, VALUE out)
     return Qnil;
 }
 
+VALUE
+argf_print(int argc, const VALUE *argv, VALUE out)
+{
+  rb_io_print(argc, argv, out)
+}
+
 /*
  *  call-seq:
  *    print(*objects) -> nil
@@ -8794,6 +8806,12 @@ rb_io_putc(VALUE io, VALUE ch)
     }
     rb_io_write(io, str);
     return ch;
+}
+
+static VALUE
+argf_putc(VALUE io, VALUE ch)
+{
+  rb_io_putc(io, ch)
 }
 
 #define forward(obj, id, argc, argv) \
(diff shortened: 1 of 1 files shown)


io.c
```diff
@@ -8557,6 +8557,12 @@ rb_io_printf(int argc, const VALUE *argv, VALUE out)
     return Qnil;
 }
 
+VALUE
```
peterzhu2118, line 8560: We should add a `static` here to make sure it's not publicly visible.

```suggestion
static VALUE
```

io.c
```diff
@@ -8557,6 +8557,12 @@ rb_io_printf(int argc, const VALUE *argv, VALUE out)
...
+argf_printf(int argc, const VALUE *argv, VALUE out)
+{
+  rb_io_printf(argc, argv, out)
```
peterzhu2118: Missing semicolon:

```suggestion
  rb_io_printf(argc, argv, out);
```

io.c
```diff
@@ -8557,6 +8557,12 @@ rb_io_printf(int argc, const VALUE *argv, VALUE out)
     return Qnil;
 }
 
+VALUE
+argf_printf(int argc, const VALUE *argv, VALUE out)
+{
+  return rb_io_printf(argc, argv, out)
```
peterzhu2118, line 8563: Missing semicolon at the end:

```suggestion
  return rb_io_printf(argc, argv, out);
```

peterzhu2118 (commented):
I think this approach makes sense.

BurdetteLamar:
@peterzhu2118, more problems here than I know what to do with.  (It's been decades since I was a C coder;  plus, I'm not sure I've ever pushed C code to ruby/ruby.)  If you can help, I'll appreciate it;  if not, I can just close this (as a swing and a miss).

peterzhu2118:
@BurdetteLamar You just need to prefix every function with `static` so that it's not publicly visible.

nobu:
You can put them in `doc/argf.rb`, I think.

BurdetteLamar:
Closed by mistake; reopened as https://github.com/ruby/ruby/pull/9762.

BurdetteLamar:
> You can put them in `doc/argf.rb`, I think.

I didn't understand this at first.  Maybe I do now.

Define the four native ARGF methods in argf.rb, where each calls the corresponding method in IO?  Then the local testing works and the doc can be put there too?

PR #9624, opened 2024-01-20
Remove empty keyword splat in array after positional splat

jeremyevans:
Previously, this wasn't removed, because compile_array_1 was missing the newarraykwsplat instruction that compile_array uses.

Fixes [Bug #20180]


jeremyevans:
@nobu I rebased this after the changes in 223910b329751fbee36efe66ccd544e66dbe90f8, to fix the conflict.

jeremyevans:
Closing as this was fixed in a different way but not backportable way in #9766.

PR #9628, opened 2024-01-22
[Bug #20198] Make sure the correct error is raised for EAI_SYSTEM resolver fail

KJTsanaktsidis:
In case of EAI_SYSTEM, getaddrinfo is supposed to set more detail in errno; however, because we call getaddrinfo on a thread now, and errno is threadlocal, that information is being lost. Instead, we just raise whatever errno happens to be on the calling thread (which can be something very confusing, like `ECHILD`).

Fix it by explicitly propagating errno back to the calling thread through the getaddrinfo_arg structure.

[Bug #20198]

### DIFF
--- a/ext/socket/raddrinfo.c
+++ b/ext/socket/raddrinfo.c
@@ -345,7 +345,7 @@ struct getaddrinfo_arg
     char *node, *service;
     struct addrinfo hints;
     struct addrinfo *ai;
-    int err, refcount, done, cancelled;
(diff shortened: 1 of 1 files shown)


mame:
@KJTsanaktsidis Thanks! It looks good to me. I guess maybe we need to fix `rb_getnameinfo` too. Could you update it?

KJTsanaktsidis:
Oh good catch, I didn't notice that method. I pushed up the fix to `rb_getnameinfo` as well.

PR #9637, opened 2024-01-22
[PRISM] Add TP call/return events to method ISEQs

eightbitraptor:
Fixes https://github.com/ruby/prism/issues/2227


kddnewton (approved):
Nice!

PR #9641, opened 2024-01-22
[PRISM] Fix keywords arguments in IndexAndWriteNode

peterzhu2118:
Fixes ruby/prism#2233.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -931,10 +931,10 @@ pm_compile_call_and_or_write_node(bool and_node, pm_node_t *receiver, pm_node_t 
 }
 
 static void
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -951,6 +951,11 @@ pm_compile_index_write_nodes_add_send(bool popped, LINK_ANCHOR *const ret, rb_is
         }
         ADD_SEND_WITH_FLAG(ret, &dummy_line_node, idASET, argc, INT2FIX(flag));
     }
+    else if (keywords && keywords->keyword_len) {
+        ADD_INSN1(ret, &dummy_line_node, opt_reverse, INT2FIX(keywords->keyword_len + block_offset + 1));
```
kddnewton, line 955: Will this fall back if optimizations are turned off?

kddnewton (approved):

prism_compile.c
peterzhu2118: `opt_reverse` is used to reverse the stack elements and it's used even when optimizations are turned off.

PR #9657, opened 2024-01-22
YJIT: Move guard up for a case of splat+rest

XrXr:
Previously, YJIT put the guard for having enough items to extract from
splat array at a place where the side exit is invalid, so if the guard
fails, YJIT could raise something other than ArgumentError. Move the
guard up to a place before any stack manipulation.

[[Bug #20204]](https://bugs.ruby-lang.org/issues/20204)


maximecb (approved):
Thank you for the fix and the extra tests 🙏

PR #9662, opened 2024-01-23
[PRISM] Support block parameters with no name

peterzhu2118:
Fixes ruby/prism#2249.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -6632,8 +6632,10 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
                 body->param.flags.has_block = true;
 
                 pm_constant_id_t name = ((pm_block_parameter_node_t *)parameters_node->block)->name;
-                pm_insert_local_index(name, local_index, index_lookup_table, local_table_for_iseq, scope_node);
-                local_index++;
+                if (name != 0) {
+                    pm_insert_local_index(name, local_index, index_lookup_table, local_table_for_iseq, scope_node);
+                    local_index++;
+                }
             }
         }
 
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -57,18 +57,8 @@
 #define PM_NOP \
     ADD_INSN(ret, &dummy_line_node, nop)
 
-/**
- * We're using the top most bit of a pm_constant_id_t as a tag to represent an
- * anonymous local. When a child iseq is created and needs access to a value
- * that has yet to be defined, or is defined by the parent node's iseq. This can
- * be added to it's local table and then handled accordingly when compiling the
- * scope node associated with the child iseq.
- *
- * See the compilation process for PM_FOR_NODE: as an example, where the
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -2790,7 +2790,7 @@ pm_add_ensure_iseq(LINK_ANCHOR *const ret, rb_iseq_t *iseq, int is_return, const
 static void
 pm_insert_local_index(pm_constant_id_t constant_id, int local_index, st_table *index_lookup_table, rb_ast_id_table_t *local_table_for_iseq, pm_scope_node_t *scope_node)
 {
-    ID local = pm_constant_id_lookup(scope_node, constant_id);
+    ID local = constant_id == 0 ? 0 : pm_constant_id_lookup(scope_node, constant_id);
```
kddnewton, line 2793: This is going to add an invalid ID into the local table. Is this necessary at all? It seems like we should probably skip this step entirely if there is no name on the block parameter.

kddnewton (changes requested):

prism_compile.c
peterzhu2118: Good point! I changed it to skip writing to the local table at all if the `constant_id` is 0.

prism_compile.c
```diff
@@ -2790,8 +2790,9 @@ pm_add_ensure_iseq(LINK_ANCHOR *const ret, rb_iseq_t *iseq, int is_return, const
 static void
 pm_insert_local_index(pm_constant_id_t constant_id, int local_index, st_table *index_lookup_table, rb_ast_id_table_t *local_table_for_iseq, pm_scope_node_t *scope_node)
 {
-    ID local = pm_constant_id_lookup(scope_node, constant_id);
-    local_table_for_iseq->ids[local_index] = local;
+    if (constant_id != 0) {
+        local_table_for_iseq->ids[local_index] = pm_constant_id_lookup(scope_node, constant_id);
+    }
     st_insert(index_lookup_table, constant_id, local_index);
```
kddnewton, line 2796: This is going to insert `0` into the index lookup table for every local that is anonymous. I _think_ this means that code like:

```ruby
def foo(*, &); end
```

will insert them both into the lookup table. Is there a way we could skip the `pm_insert_local_index` call entirely?

prism_compile.c
```diff
@@ -6632,8 +6632,10 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
                 body->param.flags.has_block = true;
 
                 pm_constant_id_t name = ((pm_block_parameter_node_t *)parameters_node->block)->name;
-                pm_insert_local_index(name, local_index, index_lookup_table, local_table_for_iseq, scope_node);
-                local_index++;
+                if (name != 0) {
+                    pm_insert_local_index(name, local_index, index_lookup_table, local_table_for_iseq, scope_node);
+                    local_index++;
```
XrXr, line 6637: I think you need to increment regardless of whether there is a name since getblockparamproxy gets the right index and you have the right local count.

kddnewton (approved):
Looks great

PR #9666, opened 2024-01-23
[DOC] Correction for Process.spawn doc

BurdetteLamar:
Response to https://github.com/ruby/ruby/pull/8342#issuecomment-1902090485.

The restored paragraph may need further adjustments, but I'll begin with just restoring the literal original.

### DIFF
--- a/process.c
+++ b/process.c
@@ -4991,7 +4991,9 @@ rb_f_system(int argc, VALUE *argv, VALUE _)
  *
  *      "Hello! 1\n"
  *
- *  Ruby invokes the executable directly, with no shell and no shell expansion.
+ *  Ruby invokes the executable directly.
+ *  This form does not use the shell;
+ *  see below for caveats.
  *
  *  If one or more +args+ is given, each is an argument or option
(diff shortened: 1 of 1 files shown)


nobu:
It should be the same for `Kernel.exec` and `Kernel.system`.

process.c
```diff
@@ -5006,6 +5008,14 @@ rb_f_system(int argc, VALUE *argv, VALUE _)
  *    C*
  *    hello world
  *
+ *  The 'cmdname, arg1, ...' form does not use the shell. However,
+ *  on different OSes, different things are provided as built-in
+ *  commands. An example of this is 'echo', which is a built-in
+ *  on Windows, but is a normal program on Linux and Mac OS X.
+ *  This means that `Process.spawn 'echo', '%Path%'` will display
+ *  the contents of the `%Path%` environment variable on Windows,
+ *  but `Process.spawn 'echo', '$PATH'` prints the literal '$PATH'.
```
nobu, line 5017: This comment is not in markdown.

PR #9671, opened 2024-01-24
Use exit 0 instead of true on windows platform

hsbt:


test/fiber/test_process.rb
```diff
@@ -3,13 +3,15 @@
 require_relative 'scheduler'
 
 class TestFiberProcess < Test::Unit::TestCase
+  TRUE_CMD = RUBY_PLATFORM =~ /mswin|mingw/ ? "exit 0" : "true"
```
nobu, line 6: `TRUE_CMD = "exit"`

PR #9673, opened 2024-01-24
Fix default gemspec file list for extension gems

deivid-rodriguez:
This is work to fix:

* https://github.com/rubygems/rubygems/issues/6207

and

* https://github.com/rubygems/rubygems/issues/7160

### DIFF
--- a/tool/rbinstall.rb
+++ b/tool/rbinstall.rb
@@ -573,7 +573,12 @@ module RbInstall
           return [] if name.empty?
 
           feature = makefile[/^DLLIB[ \t]*=[ \t]*((?:.*\\\n)*.*)/, 1]
-          Array(feature.sub("$(TARGET)", name))
+          feature = feature.sub("$(TARGET)", name)
+
+          target_prefix = makefile[/^target_prefix[ \t]*=[ \t]*((?:.*\\\n)*.*)/, 1]
+          feature = File.join(target_prefix.delete_prefix("/"), feature) unless target_prefix.empty?
+
+          Array(feature)
         end
 
         def makefile_path
--- a/tool/rbinstall.rb
+++ b/tool/rbinstall.rb
@@ -549,18 +549,36 @@ module RbInstall
       end
 
       def collect
-        ruby_libraries.sort
+        libraries.sort
       end
 
       class Ext < self
-        def ruby_libraries
+        def libraries
           # install ext only when it's configured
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
I also added an extra change to generate requirable features directly in this file list (without `lib/` in front) consistently for all default gems. This way, we'll be able to eventually stop supporting the other format in RubyGems (see https://github.com/rubygems/rubygems/blob/53604cb26e23ad9b6b81b8c7eb909f5ac5d4e5fe/lib/rubygems.rb#L1182-L1215).

hsbt:
Sorry to late response. I'll look tomorrow.

hsbt (approved):

hsbt:
I confirmed this change to add `.so` files into `spec.files` and remove `lib/` prefix to .rb files for default gems. But I have some concerns. Sorry after my approval.

I checked all of gemspec with this PR and without one.

https://gist.github.com/hsbt/520371620754e4be7ce01d358598b6c4

I found two things:
* `cgi/escape.bundle` is not located on `cgi.gemspec`. Is it intentional?
* `io/console.bundle` is located with `console.bundle` without `io/` prefix. Is it also intentional?

deivid-rodriguez:
Thanks for checking! No, it's not expected. I will have a look and try to fix those :+1:.

deivid-rodriguez:
@hsbt I looked into both issues, I think it should be correct now!

hsbt:
@deivid-rodriguez Thanks. I'll look this at next week.

hsbt:
@deivid-rodriguez I confirmed to fix my comment.

https://gist.github.com/hsbt/91aa79a2bfdf14ba95cb83e262ab41b8

Let's try this.

hsbt:
This change leads to fail with cross-build. https://rubyci.s3.amazonaws.com/crossruby/crossruby-master-arm/log/20240325T034328Z.fail.html.gz

I'll look that.

deivid-rodriguez:
Thanks! Let me know if help is needed with that failure.

hsbt:
@MSP-Greg Thanks. I fixed it with https://github.com/ruby/ruby/commit/b39057f32cdcf15d8586bd1629d5fc78038320b7

MSP-Greg:
@hsbt Thank you for the quick fix.  As you've probably seen, fixed the CI here.  All three ruby-loco builds also passed...

deivid-rodriguez:
Thank you both!

PR #9682, opened 2024-01-24
Add an internal event hook for JIT side_exits

jhawthorn:
This adds the ability for an extension to trace side exits from YJIT/RJIT on demand. This allows, for example, gathering the same type of data as ``.

I believe this will be very low overhead as we only need to check in the case we are already side-exiting (the C compiler should be able to combine the `val == Qundef` check with an existing one) and when not hooked into `EXEC_EVENT_HOOK` is a simple bit test.

I've prototyped this in vernier https://github.com/jhawthorn/vernier/pull/24 but would happily add the same functionality to `stackprof` or a standlone gem as well so that it's widely available.

(TODO: I'll run benchmarks to verify no increase in overhead)

### DIFF
--- a/include/ruby/internal/event.h
+++ b/include/ruby/internal/event.h
@@ -89,7 +89,7 @@
  */
 #define RUBY_INTERNAL_EVENT_SWITCH          0x040000 /**< Thread switched. */
 #define RUBY_EVENT_SWITCH                   0x040000 /**< @old{RUBY_INTERNAL_EVENT_SWITCH} */
-                                         /* 0x080000 */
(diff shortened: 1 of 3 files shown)


k0kubun:
JIT entry/exit could be fairly rare today, so I don’t expect a significant slowdown. 

However, have you considered calling the hook from generated code? That would remove any overhead to normal exits, I don’t think it’s too complicated to implement, and we could do that in a trampoline to minimize the code size for it.

maximecb:
I don't think that entry/exit is that rare. We definitely want it to perform as well as possible.

I agree with what Kokubun said. If you could generate a call in the machine code, from YJIT, and only if the hook is registered, then this could be effectively zero-cost if the hook isn't registered. That way you could have your cake and eat it took 🍰 :)

PR #9685, opened 2024-01-24
YJIT: reduce default exec mem size to 48MiB

maximecb:
Based on user feedback from @jhawthorn and others. Better for small and memory-constrained deployments.

NOTE: This commit should be included in the next Ruby 3.3.x point release. @xrxr should we tag someone specific?


XrXr:
To follow the usual workflow, I think you need to have a closed ticket on https://bugs.ruby-lang.org/ and tag it with this commit by mentioning the ticket number in the message, like `[Bug #XXXX]`. Otherwise you could make a PR against the branch.

k0kubun:
a small fix: https://github.com/ruby/ruby/pull/9687

k0kubun:
> what is the branch?

It's [ruby_3_3](https://github.com/ruby/ruby/tree/ruby_3_3) branch.

PR #9692, opened 2024-01-24
YJIT: reduce default exec mem size to 48MiB

maximecb:
Reduce memory overhead for small deployments (e.g. Heroku) and memory-constrained deployments.


maximecb:
@nurse 😃

PR #9694, opened 2024-01-24
Omit TestCompilePrism on s390x for now

k0kubun:


nobu:
Isn’t it better to skip by `-x` option in `.travis.yml`?

k0kubun:
Ruby CI s390x-linux would keep failing with that. I'd be fine with that too, so feel free to push that change though.

k0kubun:
Hmm, but from the gem sync perspective, maybe it's better?

k0kubun:
like this https://github.com/ruby/ruby/pull/9694/commits/417a63fda443b3959f598e42607c896f98a7b8c1? 🤔

PR #9697, opened 2024-01-25
Annotate some commonly used methods as leaf

k0kubun:
This PR adds `:leaf` annotation to methods that appear in `--yjit-stats`'s "Top-20 most frequent C calls" on yjit-bench.

It doesn't have a significant impact on those benchmarks, but I think it's nice to remove them from the TODO list to optimize the frame push overhead.


class.c
```diff
@@ -2644,6 +2644,4 @@ rb_scan_args_kw(int kw_flag, int argc, const VALUE *argv, const char *fmt, ...)
     return rb_scan_args_result(&arg, argc);
 }
 
-/*!
- * \}
- */
```
nobu, line 2649: This comment is probably for some documentation tools.

k0kubun:
More methods hit the problem on CI than I thought, so closing this PR for now. will revisit this later.

PR #9698, opened 2024-01-25
Make Net::HTTP tests ensure local & remote hosts are same addr family

KJTsanaktsidis:
* When making connections wiht TCPSocket.new and friends, we resolve the passed-in hostname with AI_ADDRCONFIG.
* AI_ADDRCONFIG means (in practice) "do not make AAAA queries or return any IPv6 addresses unless this machine has a network interface with an IPv6 address".
* The loopback adapter does _NOT_ count as "a network interface with an IPv6 address" for this check; your machine must have some _other_, non loopback address.

This causes issues with the Net::HTTP tests which set the local_host to connect _from_. This obviously needs to have the same address family as the host to connect _to_. However, we call Addrinfo.tcp('localhost')
 _without_ AI_ADDRCONFIG to set this, which means it might be `::1` even
in circumstances where the lookup done on the hostname to connect _to_ (also 'localhost') returns `127.0.0.1`.

Fix this by passing a hostname into `local_host=`, not an address; the hostname will be resolved with the same address family used for connecting.

https://bugs.ruby-lang.org/issues/20208

### DIFF
--- a/test/net/http/test_http.rb
+++ b/test/net/http/test_http.rb
@@ -1270,21 +1270,30 @@ class TestNetHTTPLocalBind < Test::Unit::TestCase
     @server.mount_proc('/show_ip') { |req, res| res.body = req.remote_ip }
 
     http = Net::HTTP.new(config('host'), config('port'))
-    http.local_host = Addrinfo.tcp(config('host'), config('port')).ip_address
+    http.local_host = config('host')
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
We decided to just revert the whole AI_ADDRCONFIG thing instead.

PR #9702, opened 2024-01-25
Bundle rbs-3.4.3

soutaro:
Updating bundled rbs to 3.4.3 introduced a test failure.

* `test_loading_from_rbs_collection__gem_version_mismatch(RBS::EnvironmentLoaderTest)` is failing because it requires `rbs-amber` gem, which is generated during tests into Bundler context. Since the test is running without Bundler, it should be skipped. (This is a test case added in rbs-3.4.3, so it's a bug of rbs-3.4.3...)
* Another failure is `test_argument_forwarding(RBS::RbPrototypeTest)`, which is introduced during recent parser changes. So, I'm just moving the ignore from `TEST_BUNDLED_GEMS_ALLOW_FAILURES` to more fine grained `rbs_skip_tests`. We will enable the tests before Ruby 3.4 release, once the parser got stable enough.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -85,7 +85,7 @@ jobs:
         env:
           RUBY_TESTOPTS: '-q --tty=no'
           TESTS: ${{ matrix.test_task == 'check' && matrix.skipped_tests || '' }}
-          TEST_BUNDLED_GEMS_ALLOW_FAILURES: 'rbs,typeprof'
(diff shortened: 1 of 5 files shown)


soutaro:
typeprof is still failing:

```
RuntimeError: Unknown insn: #<struct TypeProf::ISeq::Insn insn=:pushtoarray, operands=[], lineno=5, code_range=(5,4)-(5,27), definitions=nil>
```

It looks like unrelated to rbs update. So leaving it for @mame. 🙏

PR #9708, opened 2024-01-25
YJIT: Support concattoarray and pushtoarray

k0kubun:
This PR adds support for `concatarray` and `pushtoarray` instructions added in https://github.com/ruby/ruby/pull/9247 so that `railsbench` will not exit on those instructions.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4459,3 +4459,17 @@ assert_equal '[[{:a=>1}], {}]', %q{
 
   body
 }
+
+# concatarray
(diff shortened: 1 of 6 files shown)


XrXr:
The ARM backtrace is more helpful in case you haven't caught it.

k0kubun:
Interesting, thanks for the info. I got that from https://github.com/ruby/ruby/blob/2a9635c9060f750c849f89b93d6ecfc014274c9f/yjit/src/codegen.rs#L2056, but maybe that path is unreachable? I'll use `sp_offset` instead.

maximecb (approved):

PR #9711, opened 2024-01-25
YJIT: Fix exits on splatkw instruction

k0kubun:
This PR adds support for `splatkw` instruction added in https://github.com/ruby/ruby/pull/8877. It was merged on Dec 9th last year. Although we did have that instruction in rc1, it was so late in Ruby 3.3 development that it slipped from our mind.

It's 70-80% of exits on `railsbench` and `lobsters`, and 40% of exits on SFR. It might be better to backport this to 3.3.1.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1436,6 +1436,7 @@ fn gen_splatarray(
 }
 
 // call to_hash on hash to keyword splat before converting block
+// e.g. foo(**object, &block)
 fn gen_splatkw(
     jit: &mut JITState,
     asm: &mut Assembler,
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1435,6 +1435,32 @@ fn gen_splatarray(
     Some(KeepCompiling)
 }
 
+// call to_hash on hash to keyword splat before converting block
+fn gen_splatkw(
+    jit: &mut JITState,
+    asm: &mut Assembler,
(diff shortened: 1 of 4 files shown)


yjit/src/codegen.rs
```diff
@@ -1435,6 +1435,32 @@ fn gen_splatarray(
     Some(KeepCompiling)
 }
 
+// call to_hash on hash to keyword splat before converting block
```
maximecb, line 1438: Can you include an example in the comment of a call that would generate this?

maximecb (approved):

yjit/src/codegen.rs
k0kubun: sure https://github.com/ruby/ruby/pull/9711/commits/6a43c3efff0a0284c8b8bc7b8fbc79026020cc98

XrXr:
Oh no it's exposing what looks like a code GC bug 

```
  thread '<unnamed>' panicked at src/core.rs:2751:9:
  assertion `left == right` failed: each stub expects a particular iseq
    left: 0x7f1f32fa9850
   right: 0x7f1f29d06820
```

jeremyevans:
I received a similar failure to the one @XrXr mentioned (without this PR) on my repository earlier today: https://github.com/jeremyevans/ruby/actions/runs/7660291088/job/20877275810#step:12:107

So it doesn't look related to this PR.

PR #9716, opened 2024-01-26
Reduce array allocations during method calling in some additional cases

jeremyevans:
This reduces unnecessary array allocations for the following types of calls:

```ruby
    super(arg, *ary)
    super(arg, *ary, &block)
    super(*ary, **kw)
    super(*ary, **kw, &block)
    super(*ary, kw: 1)
    super(*ary, kw: 1, &block)
    super(...)

    f(*ary, **kw, &block)
    f(...)

    def f(*ary, post_arg)
      super
    end
```

The ` f(*ary, **kw, &block)` case was previously optimized before Ruby 3.3 was released, but the optimization no longer took effect after the introduction of the splatkw VM instruction.  This also resulted in `f(...)` not being fully optimized, after the recent switch to operate as `f(*, **, &)`.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -3192,6 +3192,10 @@ optimize_args_splat_no_copy(rb_iseq_t *iseq, INSN *insn, LINK_ELEMENT *niobj,
                                  unsigned int set_flags, unsigned int unset_flags)
 {
     LINK_ELEMENT *iobj = (LINK_ELEMENT *)insn;
+    if ((set_flags & VM_CALL_ARGS_BLOCKARG) && (set_flags & VM_CALL_KW_SPLAT) &&
+            IS_NEXT_INSN_ID(niobj, splatkw)) {
+        niobj = niobj->next;
+    }
     if (!IS_NEXT_INSN_ID(niobj, send) && !IS_NEXT_INSN_ID(niobj, invokesuper)) {
         return false;
     }
--- a/compile.c
+++ b/compile.c
@@ -3192,7 +3192,7 @@ optimize_args_splat_no_copy(rb_iseq_t *iseq, INSN *insn, LINK_ELEMENT *niobj,
                                  unsigned int set_flags, unsigned int unset_flags)
 {
     LINK_ELEMENT *iobj = (LINK_ELEMENT *)insn;
-    if (!IS_NEXT_INSN_ID(niobj, send)) {
+    if (!IS_NEXT_INSN_ID(niobj, send) && !IS_NEXT_INSN_ID(niobj, invokesuper)) {
         return false;
     }
     niobj = niobj->next;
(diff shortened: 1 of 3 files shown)


ko1:
Sorry I couldn't catch what kind of object allocation was reduced? Could you show us some example code with allocation comments?

jeremyevans:
@ko1 For all of these cases:

```
    super(arg, *ary)
    super(arg, *ary, &block)
    super(*ary, **kw)
    super(*ary, **kw, &block)
    super(*ary, kw: 1)
    super(*ary, kw: 1, &block)
    f(*ary, **kw, &block)
```

Previously, the code would use `splatarray true` for `ary`, which duplicates `ary` caller side.  This changes it to use `splatarray false`, which does not duplicate the `ary` caller side.  This saves 1 array allocation per method call for the types of calls above.

For these cases:

```
    super(...)
    f(...)
```

The same change is made, but it affects the implicit positional array splat argument, saving 1 array allocation per method call.

For this case:

```
    def f(*ary, post_arg)
      super
    end
```

instead of `splatarray false`, `newarray` and `concatarray` (array allocations for both `newarray` and `concatarray`), `splatarray true` and `pushtoarray` is used to append `post_arg` to the copy of `ary`, saving 1 allocation (only `splatarray true` allocates).  This also sets `VM_CALL_ARGS_SPLAT_MUT`, so if the super method has a splat parameter, this saves an additional array.

compile.c
```diff
@@ -9495,7 +9499,7 @@ compile_super(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, i
             /* rest argument */
             int idx = local_body->local_table_size - local_body->param.rest_start;
             ADD_GETLOCAL(args, node, idx, lvar_level);
-            ADD_INSN1(args, node, splatarray, Qfalse);
+            ADD_INSN1(args, node, splatarray, local_body->param.flags.has_post ? Qtrue : Qfalse);
```
nobu, line 9502: You can use `RBOOL` macro.

compile.c
```diff
@@ -9504,15 +9508,15 @@ compile_super(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, i
             /* post arguments */
             int post_len = local_body->param.post_num;
             int post_start = local_body->param.post_start;
+            flag |= VM_CALL_ARGS_SPLAT_MUT;
```
nobu, line 9511: Is it OK even if `!has_rest`?

jeremyevans: No, I pushed a commit to move this into the has_rest block.

compile.c
jeremyevans: Thanks. I pushed that change.

jeremyevans:
@nobu is this OK to merge after the changes?

PR #9722, opened 2024-01-26
[PRISM] Fix for s390x

peterzhu2118:
On s390x, a long is 8 bytes. st_data_t is an unsigned long but pm_constant_id_t is a 4 byte integer. We need to cast it to st_data_t when passing it to ST functions.


kddnewton (approved):

junaruga:
I just would like to say thank you for caring the s390x case!

PR #9725, opened 2024-01-26
Eliminate 1-2 array allocations for each splat used in a op_asgn method

jeremyevans:
Given code such as:

```ruby
  h[*a, 1] += 1
  h[*b] += 2
```

Ruby would previously allocate 5 arrays:

* splatarray true for a
* newarray for 1
* concatarray for [*a, 1] and [1]
* newarray for 2
* concatarray for b and [2]

This optimizes it to only allocate 2 arrays:

* splatarray true for a
* splatarray true for b

Instead of the newarray/concatarray combination, pushtoarray is used.

Note above that there was no splatarray true for b originally. The normal compilation uses splatarray false for b.  Instead of trying to find and modify the splatarray false to splatarray true, this adds splatarray true for b, which requires a couple of swap instructions, before the pushtoarray.  This could be further optimized to remove the need for those three instructions, but I'm not sure if the complexity is worth it.

Additionally, this sets VM_CALL_ARGS_SPLAT_MUT on the call to []= in the h[*b] case, so that if []= has a splat parameter, the new array can be used directly, without additional duplication.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -9100,20 +9100,28 @@ compile_op_asgn1(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node
         if (flag & VM_CALL_ARGS_SPLAT) {
             if (flag & VM_CALL_KW_SPLAT) {
                 ADD_INSN1(ret, node, topn, INT2FIX(2 + boff));
+                if (!(flag & VM_CALL_ARGS_SPLAT_MUT)) {
+                    ADD_INSN1(ret, node, splatarray, Qtrue);
+                    flag |= VM_CALL_ARGS_SPLAT_MUT;
+                }
                 ADD_INSN(ret, node, swap);
-                ADD_INSN1(ret, node, newarray, INT2FIX(1));
-                ADD_INSN(ret, node, concatarray);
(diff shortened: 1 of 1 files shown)


nobu (approved):
Seems repeating the patterns very similar to #9723.
Any idea to deduplicate?

jeremyevans:
> Seems repeating the patterns very similar to #9723. Any idea to deduplicate?

Similar in idea, but fairly different in terms of code:

Flag modification:

* masgn: generate new ci using `ci_flag_set`
* op_asgn: modify `flag` directly using `flag |=`

Instruction addition:

* masgn: Uses INSERT_BEFORE_INSN{,1}
* op_asgn: Uses ADD_INSN{,1}

I couldn't think of a good way to reuse the logic.  It may be best to keep them separate.

For this PR (op_asgn), we could use a couple macros to deduplicate some of the code.  Let me know if you would like me to do that.

nobu:
> For this PR (op_asgn), we could use a couple macros to deduplicate some of the code. Let me know if you would like me to do that.

Yeah, I think it might be possible by macros probably too but would not be needed for now.

PR #9734, opened 2024-01-28
[Bug #20220] ASAN fixes for M:N threading & objspace

KJTsanaktsidis:
This PR contains two commits for fixing ASAN.

---

1:  Notify ASAN about M:N threading stack switches

In a similar way to how we do it with fibers in cont.c, we need to call __sanitize_start_switch_fiber and __sanitize_finish_switch_fiber around the call to coroutine_transfer to let ASAN save & restore the fake stack pointer.

When a M:N thread is exiting, we pass `to_dead` to the new coroutine_transfer0 function, so that we can pass NULL for saving the stack pointer. This signals to ASAN that the fake stack can be freed (otherwise it would be leaked).

----

2. Don't check __asan_region_is_poisoned in objspace_each_objects

This returns whether or not _any_ piece of memory in the range is poisoned, not if _all_ of it is. That means that currently, with ASAN enabled, pages which contain a single poisoned object are skipped entirely from being iterated with objspace_each* family of functions

----

With these two patches, `make btest` now works on my machine with asan! Progress!


nobu:
The commit message of 98123cebfda9c08f50fbaaa81c26ddb0620589e0:
```
1s/Don't chekc/Don't check/
```

KJTsanaktsidis:
@ko1 could you please have a look at this :eyes:  since it touches M:N threading?

KJTsanaktsidis:
> The commit message

Ah, oops. Thanks, fixed. I should put a spell checker in as a pre-commit hook or something...

KJTsanaktsidis:
Worked when I built with `-DVM_CHECK_MODE=1` too.

PR #9741, opened 2024-01-29
[PRISM] Fix InterpolatedStringNode

eightbitraptor:
If the first element of an interpolated string node is an embedded statement, CRuby "pre-initializes" the interpolation with a string of known encoding to concat into.

This patch replicates thate behaviour in Prism

Fixes: https://github.com/ruby/prism/issues/2271

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -5351,7 +5351,12 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
       }
       case PM_INTERPOLATED_STRING_NODE: {
         pm_interpolated_string_node_t *interp_string_node = (pm_interpolated_string_node_t *) node;
-        int number_of_items_pushed = pm_interpolated_node_compile(&interp_string_node->parts, iseq, dummy_line_node, ret, src, popped, scope_node, parser);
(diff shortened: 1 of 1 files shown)


prism_compile.c
```diff
@@ -5351,7 +5351,12 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
       }
       case PM_INTERPOLATED_STRING_NODE: {
         pm_interpolated_string_node_t *interp_string_node = (pm_interpolated_string_node_t *) node;
-        int number_of_items_pushed = pm_interpolated_node_compile(&interp_string_node->parts, iseq, dummy_line_node, ret, src, popped, scope_node, parser);
+        int number_of_items_pushed = 0;
+        if (PM_NODE_TYPE_P(interp_string_node->parts.nodes[0], PM_EMBEDDED_STATEMENTS_NODE)) {
```
kddnewton, line 5355:
```suggestion
        if (!PM_NODE_TYPE_P(interp_string_node->parts.nodes[0], PM_STRING_NODE)) {
```

It could also be an embedded variable node

kddnewton (changes requested):

kddnewton (approved):

PR #9742, opened 2024-01-29
[PRISM] Use opt_str_freeze instruction

peterzhu2118:
Fixes ruby/prism#2290.


prism_compile.c
```diff
@@ -3956,11 +3956,19 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
 
         if (call_node->receiver == NULL) {
             PM_PUTSELF;
-        } else {
+            pm_compile_call(iseq, call_node, ret, src, popped, scope_node, method_id, start);
+        }
+        else if (PM_NODE_TYPE_P(call_node->receiver, PM_STRING_NODE) &&
+                call_node->arguments == NULL &&
+                method_id == idFreeze) {
+            VALUE str = rb_fstring(parse_string(&((pm_string_node_t *)call_node->receiver)->unescaped, parser));
```
kddnewton, line 3964: I think we might lose the encoding of the string here?

prism_compile.c
```diff
@@ -3956,11 +3956,19 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
 
         if (call_node->receiver == NULL) {
             PM_PUTSELF;
-        } else {
+            pm_compile_call(iseq, call_node, ret, src, popped, scope_node, method_id, start);
+        }
+        else if (PM_NODE_TYPE_P(call_node->receiver, PM_STRING_NODE) &&
+                call_node->arguments == NULL &&
```
kddnewton, line 3962: We should also check there's no block on the method

kddnewton (changes requested):

kddnewton (approved):

PR #9749, opened 2024-01-29
Fix RegExp warning causing flaky Ripper failure

XrXr:
Sometimes this file get picked up and break Ripper tests:

    TestRipper::Generic#test_parse_files:test/ruby
    assert_separately failed with error message
    pid 63392 exit 0
    | test_regexp.rb:2025: warning: character class has duplicated range

https://github.com/ruby/ruby/actions/runs/7699956651/job/20982702553#step:12:103


makenowjust:
@XrXr Thank you!

PR #9752, opened 2024-01-30
Avoid unnecessary array allocation for ARGSCAT with LIST body

jeremyevans:
Previously, this would use newarray followed by concattoarray. This now uses pushtoarray instead, avoiding the unnecessary array allocation.

This is implemented by making compile_array take a first_chunk argument, passing in 1 in the normal array case, and 0 in the ARGSCAT with LIST body case.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -4786,7 +4786,7 @@ static_literal_value(const NODE *node, rb_iseq_t *iseq)
 }
 
 static int
-compile_array(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *node, int popped, int first_chunk)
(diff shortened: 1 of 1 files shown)


compile.c
```diff
@@ -10174,7 +10173,7 @@ iseq_compile_each0(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const no
         CHECK(compile_super(iseq, ret, node, popped, type));
         break;
       case NODE_LIST:{
-        CHECK(compile_array(iseq, ret, node, popped) >= 0);
+        CHECK(compile_array(iseq, ret, node, popped, 1) >= 0);
```
nobu, line 10176: Maybe `TRUE`/`FALSE` are better, including assignments to this variable.

nobu (approved):

PR #9762, opened 2024-01-30
Add ARGF native methods

BurdetteLamar:
Mistakenly closed earlier PR (https://github.com/ruby/ruby/pull/9623).

Adds native code for methods #putc, #puts, #print, #printf, which formerly called corresponding methods in class IO; this meant that the doc for the methods in ARGF came from IO.

This change will mean that the methods in ARGF can have their own documentation (not yet added).

This is a draft b/c I want to get the new code approved before adding the documentation.

### DIFF
--- a/io.c
+++ b/io.c
@@ -8556,6 +8556,12 @@ rb_io_printf(int argc, const VALUE *argv, VALUE out)
     return Qnil;
 }
 
+static VALUE
+argf_printf(int argc, const VALUE *argv, VALUE out)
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@peterzhu2118 and @nobu, I'm thinking of close this as not being the best use of our time.  Objections?

BurdetteLamar:
May need a different approach.  Maybe later.

PR #9764, opened 2024-01-30
YJIT: Specialize splatkw on T_HASH

k0kubun:
Follow-up on https://github.com/ruby/ruby/pull/9711.

`splatkw` converts the operand only when it's not already `T_HASH`. People don't really use `**` for non-Hash objects, so we could guard that it's `T_HASH` so that we can skip `jit_prepare_routine_call`, etc.

The instruction is used in Rails-based benchmarks, and it seems to help `erubi-rails` and `railsbench` a little:

```
before: ruby 3.4.0dev (2024-01-30T17:54:04Z master 6b350bc6e4) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-01-30T18:30:09Z yjit-splat cd9b20761f) +YJIT [x86_64-linux]

erubi-rails  1264.6       0.1         1255.7      0.1         1.01           1.01
railsbench   2489.0       1.0         2459.1      1.2         1.01           1.01
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1658,7 +1658,7 @@ fn guard_object_is_hash(
     let flags_opnd = Opnd::mem(VALUE_BITS, object_reg, RUBY_OFFSET_RBASIC_FLAGS);
     let flags_opnd = asm.and(flags_opnd, (RUBY_T_MASK as u64).into());
 
-    // Compare the result with T_ARRAY
+    // Compare the result with T_HASH
     asm.cmp(flags_opnd, (RUBY_T_HASH as u64).into());
     asm.jne(Target::side_exit(counter));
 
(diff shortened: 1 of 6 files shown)


yjit/src/codegen.rs
```diff
@@ -1620,6 +1635,38 @@ fn guard_object_is_array(
...
+    };
+    guard_object_is_heap(asm, object_reg, object_opnd, counter);
+
+    asm_comment!(asm, "guard object is hash");
+
+    // Pull out the type mask
+    let flags_opnd = Opnd::mem(VALUE_BITS, object_reg, RUBY_OFFSET_RBASIC_FLAGS);
+    let flags_opnd = asm.and(flags_opnd, (RUBY_T_MASK as u64).into());
+
+    // Compare the result with T_ARRAY
```
XrXr, line 1661:
```suggestion
    // Compare the result with T_HASH
```

XrXr (approved):

PR #9767, opened 2024-01-30
YJIT: add codegen for String#setbyte

maximecb:
Before:
```
--------  -----------  ----------  ---------  ----------  ------------  -----------
bench     interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
ruby-xor  987.2        0.1         109.3      0.3         8.38          9.04       
--------  -----------  ----------  ---------  ----------  ------------  -----------
```

After:
```
--------  -----------  ----------  ---------  ----------  ------------  -----------
bench     interp (ms)  stddev (%)  yjit (ms)  stddev (%)  yjit 1st itr  interp/yjit
ruby-xor  991.8        0.6         99.6       0.4         9.25          9.96       
--------  -----------  ----------  ---------  ----------  ------------  -----------
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5113,6 +5113,37 @@ fn jit_rb_str_getbyte(
     true
 }
 
+fn jit_rb_str_setbyte(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    _ocb: &mut OutlinedCb,
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    asm_comment!(asm, "String#getbyte");
+    extern "C" {
+        fn rb_str_setbyte(str: VALUE, index: VALUE, value: VALUE) -> VALUE;
+    }
+
+    // Raises when index is out of range
+    jit_prepare_routine_call(jit, asm);
+
+    let value = asm.stack_opnd(0);
+    let index = asm.stack_opnd(1);
+    let recv = asm.stack_opnd(2);
+
+    let ret_opnd = asm.ccall(rb_str_setbyte as *const u8, vec![recv, index, value]);
+    asm.stack_pop(3); // Keep them on stack during ccall for GC
+
+    let out_opnd = asm.stack_push(Type::UnknownImm);
+    asm.mov(out_opnd, ret_opnd);
+
+    true
+}
+
 // Codegen for rb_str_to_s()
 // When String#to_s is called on a String instance, the method returns self and
 // most of the overhead comes from setting up the method call. We observed that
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -5113,6 +5113,37 @@ fn jit_rb_str_getbyte(
...
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: *const VALUE,
+) -> bool {
+    asm_comment!(asm, "String#getbyte");
+    extern "C" {
+        fn rb_str_setbyte(str: VALUE, index: VALUE, value: VALUE) -> VALUE;
+    }
+
+    // Raises when index is out of range
```
k0kubun, line 5131: Are we okay with breaking the backtrace when YJIT is enabled? We discussed the exact problem on Thursday and you suggested to put a guard on the index.

k0kubun: On the flip side, you could probably remove `jit_prepare_routine_call` if you guard it, so it would allow you to keep local types.

XrXr: We'd also need to guard for frozen strings.

maximecb: The method can still raise though? You're saying it won't report the method raising correctly? Why is that?

k0kubun: I didn't say it can't raise. I said the backtrace will change.

master:

```
$ ruby -v --yjit-call-threshold=1 -e '"".setbyte(1, 0)'
ruby 3.4.0dev (2024-01-30T19:59:53Z master c1f8d974a8) +YJIT [x86_64-linux]
-e:1:in `setbyte': index 1 out of string (IndexError)
        from -e:1:in `<main>'
```

this branch:

```
$ ruby -v --yjit-call-threshold=1 -e '"".setbyte(1, 0)'
ruby 3.4.0dev (2024-01-30T20:39:42Z yjit_setbyte 247ee663df) +YJIT [x86_64-linux]
-e:1:in `<main>': index 1 out of string (IndexError)
```

With this change, you can't know which method is raising the exception. If a line has a chain of method calls, it'd be harder to tell whether it's from `setbyte` or not.

XrXr: It raises but without mentioning that the exception comes from `setbyte`, because there is no C frame to indicate that. You can compare the backtrace with the interpreter.

maximecb: Ideally we'd preserve the backtrace, but at the same time, this can really get in the way of performance optimizations. How do you guys feel about it?

In theory, it may be possible to add JIT support for recovering a correct backtrace based on the PC (we do set up the PC in `jit_prepare_routine_call`) ?

maximecb: Ok hmmm.

I feel like if we have to guard that the index is fixnum and in range, and that the string is also frozen, it really nerfs this PR because we're adding a fair bit of extra work on top of a C call that will duplicate the guards we just did.

So this opens up at least 2 discussion threads:
1. Can we find a way to recover the backtrace for C calls based on the PC only, without needing to push a Ruby frame?
2. How easy it it to guard that a string is not frozen, and that the string has binary encoding?

k0kubun: For this PR, Alan's idea on (2) seems reasonable to me.

> 1. Can we find a way to recover the backtrace for C calls based on the PC only, without needing to push a Ruby frame?

For a long term, we should aim for having this capability. We have so many almost-leaf methods that are not leaf only because they raise. Many C methods need to be inlining-safe when we start inlining non-trivial ISEQs.

We could have a YJIT-global hash map `{ PC => C method }`, make sure it's currently the only callee from the ISEQ on YJIT (otherwise invalidate the existing block), and resurrect the frame on `rb_raise` if the top-most frame's PC is in the hash map.

If the same ISEQ is called from the interpreter for any reason, another frame (`setbyte` or another method, not the caller we're compiling) would be the top-most frame, so it wouldn't doubly push the frame.

PR #9768, opened 2024-01-30
[PRISM] Fix StackOverflow for IfNode with logical predicate

eightbitraptor:
Fixes: https://github.com/ruby/prism/issues/2294


kddnewton:
Nice cleanup!

PR #9770, opened 2024-01-30
[PRISM] Fix encoding of interpolated strings

peterzhu2118:
Fixes ruby/prism#2313.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -751,18 +751,33 @@ pm_interpolated_node_compile(pm_node_list_t *parts, rb_iseq_t *iseq, NODE dummy_
                     current_string = string_value;
                 }
             }
-            else {
+            else if (PM_NODE_TYPE_P(part, PM_EMBEDDED_STATEMENTS_NODE) &&
+                    ((pm_embedded_statements_node_t *)part)->statements->body.size == 1 &&
+                    PM_NODE_TYPE_P(((pm_embedded_statements_node_t *)part)->statements->body.nodes[0], PM_STRING_NODE)) {
+                pm_string_node_t *string_node = (pm_string_node_t *)((pm_embedded_statements_node_t *)part)->statements->body.nodes[0];
+                VALUE string_value = parse_string_encoded((pm_node_t *)string_node, &string_node->unescaped, parser);
                 if (RTEST(current_string)) {
-                    current_string = rb_fstring(current_string);
-                    if (parser->frozen_string_literal) {
-                        ADD_INSN1(ret, &dummy_line_node, putobject, current_string);
-                    }
-                    else {
-                        ADD_INSN1(ret, &dummy_line_node, putstring, current_string);
-                    }
-                    current_string = Qnil;
-                    number_of_items_pushed++;
+                    current_string = rb_str_concat(current_string, string_value);
                 }
+                else {
+                    current_string = string_value;
+                }
+            }
+            else {
+                if (!RTEST(current_string)) {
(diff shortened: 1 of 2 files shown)


kddnewton (changes requested):
The optimization you've done here will work for some cases, but I think the bigger problem is that we're pushing an empty `""` at the beginning of the string parts that is ASCII-8BIT, which is making concat-ed strings also ASCII-8BIT. That first empty string (which is currently `rb_str_new(0, 0)`) should be in the same encoding as the file.

peterzhu2118:
After this change, we're no longer pushing the empty string at the beginning. These lines were removed:

```
        if (!PM_NODE_TYPE_P(interp_string_node->parts.nodes[0], PM_STRING_NODE)) {
            ADD_INSN1(ret, &dummy_line_node, putstring, rb_str_new(0, 0));
            number_of_items_pushed++;
        }
```

kddnewton:
I know, but we need that thing to be pushed. Otherwise this PR breaks:

```ruby
a = 'Foo'
b = "#{a}" << 'Bar'
p [a, b]
```

because the `a` object becomes `"FooBar"`.

prism_compile.c
```diff
@@ -751,18 +751,33 @@ pm_interpolated_node_compile(pm_node_list_t *parts, rb_iseq_t *iseq, NODE dummy_
...
+                }
+            }
+            else {
+                if (!RTEST(current_string)) {
+                    rb_encoding *enc = rb_enc_from_index(rb_enc_find_index(parser->encoding->name));
+                    current_string = rb_enc_str_new(NULL, 0, enc);
                 }
+
+                if (parser->frozen_string_literal) {
+                    ADD_INSN1(ret, &dummy_line_node, putobject, rb_str_freeze(current_string));
```
kddnewton, line 773: Should this be an fstring?

prism_compile.c
```diff
@@ -7232,13 +7242,14 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
                   case PM_INTERPOLATED_REGULAR_EXPRESSION_NODE: {
                     pm_interpolated_regular_expression_node_t *cast = (pm_interpolated_regular_expression_node_t *) scope_node->ast_node;
 
-                    int parts_size = (int)cast->parts.size;
+                    int parts_size = 0;
                     if (parts_size > 0 && !PM_NODE_TYPE_P(cast->parts.nodes[0], PM_STRING_NODE)) {
```
kddnewton, line 7246: Should this be checking the parts size? Looks like the first check here will never be true.

prism_compile.c
```diff
@@ -7232,13 +7242,14 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
                         ADD_INSN1(ret, &dummy_line_node, putobject, rb_str_new(0, 0));
```
kddnewton, line 7247: Can you fix this encoding as well?

kddnewton (changes requested):

prism_compile.c
peterzhu2118: No, I don't think so. There are individual parts of the string rather than the resulting string. There is a `rb_fstring` call later on when we finish building the string.

prism_compile.c
peterzhu2118: Yeah I think this is dead code. We don't need to push the initial empty string anymore because `pm_interpolated_node_compile` handles it.

kddnewton (approved):
Looks great, thanks!

PR #9774, opened 2024-01-31
YJIT: Add codegen for Float arithmetics

k0kubun:
This speeds up `blurhash`, `matmul`, and `nbody`.

```
blurhash  240.8        0.4         233.7       0.4         1.03           1.03
matmul    1026.4       0.1         946.7       0.1         1.07           1.08
nbody     62.6         0.4         55.7        0.3         1.11           1.13
```

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -2484,6 +2484,11 @@ assert_equal '[true, false, true, false]', %q{
   [is_odd(123), is_odd(456), is_odd(bignum), is_odd(bignum+1)]
 }
 
+# Float calls
+assert_equal '[2.0, 0.0, 1.0, 4.0]', %q{
+  [1.0 + 1.0, 1.0 - 1.0, 1.0 * 1.0, 8.0 / 2.0]
+}
+
 # Call to static and dynamic symbol
 assert_equal 'bar', %q{
   def to_string(obj)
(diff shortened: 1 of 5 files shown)


bootstraptest/test_yjit.rb
```diff
@@ -2484,6 +2484,11 @@ def is_odd(obj)
   [is_odd(123), is_odd(456), is_odd(bignum), is_odd(bignum+1)]
 }
 
+# Float calls
+assert_equal '[2.0, 0.0, 1.0, 4.0]', %q{
+  [1.0 + 1.0, 1.0 - 1.0, 1.0 * 1.0, 8.0 / 2.0]
+}
```
maximecb, line 2490: Your test is perfectly good and valid, but I will point out in case you don't know: numbers having binary fractions are also guaranteed to be representable exactly using floating-point numbers :)

e.g. `1.50`, `1.25`, `1.125`, `2.5` all have exact IEEE FP representations.

maximecb (approved):
This is a nice first step in terms of adding FP support to YJIT 👌 

Reading the code, I think it would be good to also add tests that mix fixnum and flonum, e.g. `1.5 * 2`

maximecb (approved):

bootstraptest/test_yjit.rb
k0kubun: I wasn't clever enough to figure out `1.125` is `2^0 + 2^(-3)` right away, but thanks for a nice refresher 🙂

PR #9775, opened 2024-01-31
YJIT: Add jit_prepare_call_with_gc function

k0kubun:
This PR introduces `jit_prepare_for_gc` function to do `jit_prepare_routine_call` without `asm.clear_local_types()`.

When a C call may allocate but not call arbitrary methods, we should be able to skip `asm.clear_local_types()`. I applied this change to some instructions where I think it makes sense.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -341,22 +341,34 @@ fn gen_save_sp_with_offset(asm: &mut Assembler, offset: i8) {
     }
 }
 
-/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that
-/// could:
+/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that could:
 ///  - Perform GC allocation
 ///  - Take the VM lock through RB_VM_LOCK_ENTER()
 ///  - Perform Ruby method call
+///
+/// If the routine doesn't call arbitrary methods, use jit_prepare_allocation() instead.
 fn jit_prepare_routine_call(
     jit: &mut JITState,
     asm: &mut Assembler
 ) {
-    jit.record_boundary_patch_point = true;
-    jit_save_pc(jit, asm);
-    gen_save_sp(asm);
+    // Prepare for GC. This also sets PC, which prepares for showing a backtrace.
+    jit_prepare_for_gc(jit, asm);
 
     // In case the routine calls Ruby methods, it can set local variables
-    // through Kernel#binding and other means.
+    // through Kernel#binding, rb_debug_inspector API, and other means.
     asm.clear_local_types();
+}
+
+/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that could:
+///  - Perform GC allocation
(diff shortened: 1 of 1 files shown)


XrXr (approved):
I think we can also change the one for leaf builtins.

yjit/src/codegen.rs
```diff
@@ -341,24 +341,36 @@ fn gen_save_sp_with_offset(asm: &mut Assembler, offset: i8) {
...
 
-/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that
-/// could:
+/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that could:
 ///  - Perform GC allocation
 ///  - Take the VM lock through RB_VM_LOCK_ENTER()
 ///  - Perform Ruby method call
+///
+/// If the routine doesn't call arbitrary methods, use jit_prepare_allocation() instead.
 fn jit_prepare_routine_call(
```
maximecb, line 350: Just to improve the clarity about what this function does, we could rename it as follows?

```suggestion
fn jit_prepare_call_non_leaf(
```

That's the crux of it, right, that this can indirectly call binding because the C function can call back into Ruby code?

yjit/src/codegen.rs
```diff
@@ -341,24 +341,36 @@ fn gen_save_sp_with_offset(asm: &mut Assembler, offset: i8) {
...
     // In case the routine calls Ruby methods, it can set local variables
-    // through Kernel#binding and other means.
+    // through Kernel#binding, rb_debug_inspector API, and other means.
     asm.clear_local_types();
 }
 
+/// jit_save_pc() + gen_save_sp(). Should be used before calling a routine that could:
+///  - Perform GC allocation
+///  - Take the VM lock through RB_VM_LOCK_ENTER()
+fn jit_prepare_for_gc(
```
maximecb, line 365:
```suggestion
fn jit_prepare_call_may_gc(
```

yjit/src/codegen.rs
XrXr: Right. If the C function calls methods, all bets are off. Calling into blocks is probably more common than using binding, though.

k0kubun: I changed it to `jit_prepare_non_leaf_call` https://github.com/ruby/ruby/pull/9775/commits/a2ddbf3b354b7d16e709e0e963cd7445772c5dd3 since "prepare xxx call" like `jit_prepare_routine_call` seemed more natural to me. WDYT?

yjit/src/codegen.rs
k0kubun: If you want to keep `call`, WDYT about `jit_prepare_call_with_gc`? https://github.com/ruby/ruby/pull/9775/commits/753631921a6ea98c69cd18cea120a030d7e6b91c

maximecb (approved):

k0kubun:
> I think we can also change the one for leaf builtins.

Good call :+1: https://github.com/ruby/ruby/pull/9775/commits/402884dfd9a733686ece3a691f49237e316ab8fb

PR #9777, opened 2024-01-31
Add launchable into CI

ono-max:

PR #9778, opened 2024-01-31
Test for #9777

ono-max:
See https://github.com/ruby/ruby/pull/9777#issuecomment-1919116859


junaruga:
I meant you can push the commits only to the branch on your forked repository. I didn't mean creating another PR to this repository. Then you can check GitHub Actions on your forked repository: https://github.com/ono-max/ruby/actions The GitHub Actions are just triggered when only pushing the commits to a branch.

ono-max:
Thank you so much for your valuable suggestion. I truly appreciate your input. I understand your perspective, and I want to express that conducting tests on a forked repository poses a challenge due to authentication issues with Launchable. The system verifies the origin of the requests, and if it's not from the original repository, it returns a 401 error. Therefore, I find it necessary to conduct experiments on the original repository. I apologize for any inconvenience this may cause, and I appreciate your understanding.

junaruga:
> I understand your perspective, and I want to express that conducting tests on a forked repository poses a challenge due to authentication issues with Launchable. The system verifies the origin of the requests, and if it's not from the original repository, it returns a 401 error.

I haven't used the Launchable. So, does the forked repository's Launchable's client code have to access to the original repository's Launchable service?

Each user cannot create a Launchable service for their forked repository?

For example, Travis CI has the following page.
https://app.travis-ci.com/github/ruby/ruby?serverType=git

Then when people fork the original repository, people can create their forked Travis CI page as well.
https://app.travis-ci.com/github/junaruga/ruby?serverType=git

Launchable is not like this?

ono-max:
> I haven't used the Launchable. So, does the forked repository's Launchable's client code have to access to the original repository's Launchable service?

We can't use Launchable within the forked repository, but it is accessible in the original repository due to the presence of the 'ruby/ruby' account. Therefore, if the developer creates the PR within the original repository, they can use Launchable without any issues, even if their PR originates from the forked repository. However, if they create the PR within the forked repository and do not have a different account, they cannot utilize Launchable in that case.

PR #9780, opened 2024-01-31
[PRISM] Cleanup iseqw_s_compile_prism_compile

eightbitraptor:
Let's make this return an iseq instead of passing a reference to one, seeing as we always call it in the same way.


kddnewton:
This function is going away as part of https://github.com/ruby/ruby/pull/9781, so I'm going to close this one in favor of that.

PR #9782, opened 2024-01-31
Check un-leafness of String#setbyte

k0kubun:
Checking what CI says


k0kubun:
I didn't realize you can't even use local variables correctly, which makes sense. Closing

PR #9788, opened 2024-01-31
[PRISM] Fix multiple return with splat and splat+kwsplat

XrXr:
Previously, `return *array, 1` didn't behave like `return [*array, 1]`
properly. Also, it crashed when splat and kwsplat is combined like in
`array = [*things, **hash]`.

Fix this by grouping `PM_ARGUMENTS_NODE` with `PM_ARRAY_NODE` handling and
combining splat and kwsplat handling.


XrXr:
Note, this is also fixing https://bugs.ruby-lang.org/issues/20229 and https://bugs.ruby-lang.org/issues/20229 in the case of empty keyword splat, assuming #9766 and #9624 will be merged.

PR #9799, opened 2024-02-01
Support NODE_ONCE for pattern matching

S-H-GAMELINKS:
I saw this Issue and found that `NODE_ONCE` such as `/#{var}/o` are not supported by pattern matching.
https://github.com/rubocop/rubocop-performance/issues/438

So, add a simple test and modifying that `NODE_ONCE` can also be handled by pattern matching.


nobu (approved):

k-tsj (approved):
LGTM. Thank you!

PR #9803, opened 2024-02-01
[PRISM] Use block opening line as `source_location` line of lambda

nvasilevski:
Fixes https://github.com/ruby/prism/issues/2312

There are several prism tests failing related to the `source_location` for lambda returning line of the operator (`->`) while original parser execution results in `source_location` line pointing to the block opening location (`{` or `do`)

This commit changes `PM_LAMBDA_NODE` compilation case to use block opening location instead of the whole node (operator) opening location to get the line number to build block iseq


kddnewton (approved):
Looks great!

PR #9806, opened 2024-02-01
[PRISM] dedup hash string keys

jenshenny:
Fixes ruby/prism#2321


kddnewton (approved):
Wow great work!

jenshenny:
@kddnewton I think the tests are failing because there's missing parentheses in a macro arg https://github.com/ruby/prism/pull/2339

PR #9810, opened 2024-02-01
Correctly set anon_kwrest flag for def f(b: 1, **)

jeremyevans:
In cases where a method accepts both keywords and an anonymous keyword splat, the method was not marked as taking an anonymous keyword splat.  Fix that in the compiler.

This currently doesn't break anything, but with the changes to optimize such methods to avoid an extra hash allocation, we need to disable the anonymous keyword splat optimization in this case. Add a test to detect future breakage.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -1976,8 +1976,11 @@ iseq_set_arguments_keywords(rb_iseq_t *iseq, LINK_ANCHOR *const optargs,
     keyword->num = kw;
 
     if (RNODE_DVAR(args->kw_rest_arg)->nd_vid != 0) {
+        ID kw_id = iseq->body->local_table[arg_size];
         keyword->rest_start = arg_size++;
         body->param.flags.has_kwrest = TRUE;
+
+        if (kw_id == idPow) body->param.flags.anon_kwrest = TRUE;
     }
     keyword->required_num = rkw;
     keyword->table = &body->local_table[keyword->bits_start - keyword->num];
--- a/test/ruby/test_keyword.rb
+++ b/test/ruby/test_keyword.rb
@@ -182,6 +182,13 @@ class TestKeywordArguments < Test::Unit::TestCase
                   [:keyrest, :kw], [:block, :b]], method(:f9).parameters)
   end
 
+  def test_keyword_with_anonymous_keyword_splat
+    def self.a(b: 1, **) [b, **] end
+    kw = {b: 2, c: 3}
+    assert_equal([2, {c: 3}], a(**kw))
+    assert_equal({b: 2, c: 3}, kw)
+  end
+
   def test_keyword_splat_nil
     # cfunc call
     assert_equal(nil, p(**nil))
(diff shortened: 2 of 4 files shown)


compile.c
```diff
@@ -1974,8 +1974,13 @@ iseq_set_arguments_keywords(rb_iseq_t *iseq, LINK_ANCHOR *const optargs,
     keyword->num = kw;
 
     if (RNODE_DVAR(args->kw_rest_arg)->nd_vid != 0) {
+        ID kw_id = iseq->body->local_table[arg_size];
         keyword->rest_start = arg_size++;
         body->param.flags.has_kwrest = TRUE;
+
+        static ID anon_kwrest = 0;
+        if (!anon_kwrest) anon_kwrest = rb_intern("**");
+        if (kw_id == anon_kwrest) body->param.flags.anon_kwrest = TRUE;
```
nobu, line 1983:
```suggestion
        if (kw_id == idPow) body->param.flags.anon_kwrest = TRUE;
```

test/ruby/test_keyword.rb
```diff
@@ -182,6 +182,13 @@ def test_method_parameters
                   [:keyrest, :kw], [:block, :b]], method(:f9).parameters)
   end
 
+  def test_keyword_with_anonymous_keyword_splat
+    def self.a(b: 1, **) [b, **] end
+    kw = {b: 2, c: 3}
+    assert_equal([2, {c: 3}], a(**kw))
+    assert_equal({b: 2, c: 3}, kw)
+  end
```
nobu, line 190: 3af47f14d4e seems to pass this.

jeremyevans: Yes.  However, if you commit the `compile.c` change without the `vm_args.c` change, and you merge #9776 to avoid the unnecessary array allocation, I think it fails.

jeremyevans:
@XrXr This contains a change to yjit, because I was getting TypeErrors in CI when running with yjit without the change (https://github.com/ruby/ruby/actions/runs/8115221234/job/22182614930?pr=9810#step:13:112).

I'm not sure if the yjit change is best, as I think it disables the related optimization.  Do you have any concerns with me merging this?

XrXr (approved):
The YJIT part is fine, and the optimization is still working for `...` methods, which we care the most about.

This is actually mirroring the same optimization in the interpreter more closely, since it only passes `nil` for `annon_kwrest` when there is no keyword arguments.

The `TypeError` is from compile.c assuming `**` can only be a hash inside `hash_merge_ptr()`:
```
0005 opt_send_without_block                 <calldata!mid:dup, argc:0, ARGS_SIMPLE>
0007 putobject                              :chomp
0009 getlocal_WC_0                          chomp@0
0011 opt_send_without_block                 <calldata!mid:core#hash_merge_ptr, argc:3, ARGS_SIMPLE>
0013 invokesuper                            <calldata!argc:1, FCALL|SUPER|ZSUPER|KW_SPLAT|KW_SPLAT_MUT>, nil
```
We could expand the optimization to support kw+kwrest cases in the future.

PR #9812, opened 2024-02-01
[PRISM] Fix numbered parameters stealing local names

XrXr:
Previously, the local index of numbered parameters were assigned to
names of regular locals, making it hard to read both of them. Use proper
`_[1-9]` numbered parameters. This fixes `test_shapes.rb`.

Also, properly mark the iseq as having lead parameters.


kddnewton (changes requested):
You're right, this is a little weird. I've added an API for this now — can you use it? https://github.com/ruby/prism/pull/2357

kddnewton (approved):

PR #9814, opened 2024-02-02
Add an assertion that `%x` literals call `` ` `` method

nobu:


nobu:
Off topic: It's surprising that ` `` ` works as expected in the description and comments but not in the title.

PR #9815, opened 2024-02-02
[Feature #20233] Support pkgconf on windows

jmarrec:
1. Store the `PKG_CONFIG` variable in Makefile.sub (or try to get it from the ENV var PKG_CONFIG in mkmf.rb)
2. Try to use --msvc-syntax, with a fallback to replacing -Lxxx with -libpath:xxx. --msvc-syntax has been in pkgconf since 1.4.0 (released 7 years ago). pkg-config (freedesktop), does not support it, hence the fallback.
3. The `try_ldflags` passes these `ldflags` as the `opt` parameter to the `link_command`, not as `ldflags`. Unix systems are forgiving in that regard, MSVC is not: as a result as passing them as `opt`, they (specifically the `/libpath:xxx` ones) end up passed before the `-link` command to `cl.exe` and it throws because it ignores it and therefore can't find the lib.

```
cl : Command line warning D9002 : ignoring unknown option '-libpath:C:/Users/julien/.conan2/p/libff3726d89a6255c/p/lib'
```

Context: I am trying to make a [conan](https://conan.io/) recipe (the C++ package manager) for ruby, and realized pkgconf was not working on windows and that's why I couldn't link fiddle to libffi.

### DIFF
--- a/lib/mkmf.rb
+++ b/lib/mkmf.rb
@@ -1980,32 +1980,22 @@ SRC
         pkgconfig = nil
       end
       if pkgconfig
-        has_ms_win_syntax = false
-        if $mswin
-          has_ms_win_syntax = xpopen([pkgconfig, "--help"]).read.include?('msvc-syntax')
-          if has_ms_win_syntax
-            args << "--msvc-syntax"
-          else
-            Logging.message("WARNING: #{pkgconfig} does not support the --msvc-syntax. Try using a recent pkgconf instead")
-          end
-        end
         get = proc {|opts|
           opts = Array(opts).map { |o| "--#{o}" }
           opts = xpopen([*envs, pkgconfig, *opts, *args], err:[:child, :out], &:read)
           Logging.open {puts opts.each_line.map{|s|"=> #{s.inspect}"}}
           if $?.success?
             opts = opts.strip
-            if $mswin and not has_ms_win_syntax
-              opts = Shellwords.shellwords(opts).map { |s|
-                if s.start_with?('-l')
-                  "#{s[2..]}.lib"
-                elsif s.start_with?('-L')
-                  "/libpath:#{s[2..]}"
-                else
-                  s
-                end
-              }.quote.join(" ")
-            end
+            libarg, libpath = LIBARG, LIBPATHFLAG.strip
+            opts = opts.shellsplit.map { |s|
+              if s.start_with?('-l')
+                libarg % s[2..]
+              elsif s.start_with?('-L')
+                libpath % s[2..]
+              else
+                s
+              end
+            }.quote.join(" ")
             opts
           end
         }
--- a/lib/mkmf.rb
+++ b/lib/mkmf.rb
@@ -604,9 +604,9 @@ MSG
     yield(opt, opts)
   end
 
-  def try_link0(src, opt = "", **opts, &b) # :nodoc:
+  def try_link0(src, opt = "", ldflags: "", **opts, &b) # :nodoc:
     exe = CONFTEST+$EXEEXT
-    cmd = link_command("", opt)
+    cmd = link_command(ldflags, opt)
     if $universal
       require 'tmpdir'
       Dir.mktmpdir("mkmf_", oldtmpdir = ENV["TMPDIR"]) do |tmpdir|
@@ -750,7 +750,7 @@ MSG
 
   # :nodoc:
   def try_ldflags(flags, werror: $mswin, **opts)
-    try_link(MAIN_DOES_NOTHING, flags, werror: werror, **opts)
+    try_link(MAIN_DOES_NOTHING, "", ldflags: flags, werror: werror, **opts)
   end
 
   # :startdoc:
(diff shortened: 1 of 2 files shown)


win32/Makefile.sub
```diff
@@ -1129,6 +1129,7 @@ s,@top_srcdir@,$(srcdir),;t t
 s,@try_header@,try_compile,;t t
 s,@ruby_pc@,$(ruby_pc),;t t
 s,@RJIT_SUPPORT@,$(RJIT_SUPPORT),;t t
+s,@PKG_CONFIG@,$(PKG_CONFIG),;t t
```
jmarrec, line 1132: Store PKG_CONFIG in makefile.sub

lib/mkmf.rb
```diff
@@ -1957,11 +1957,34 @@ def fmt.%(x)
...
       if pkgconfig
+        has_ms_win_syntax = false
+        if $mswin
+          has_ms_win_syntax = xpopen([pkgconfig, "--help"]).read.include?('msvc-syntax')
+          if has_ms_win_syntax
+            args << "--msvc-syntax"
+          else
+            Logging.message("WARNING: #{pkgconfig} does not support the --msvc-syntax. Try using a recent pkgconf instead")
+          end
+        end
```
jmarrec, line 1968: Try to detect if pkgconf supports the "--msvc-syntax" flag. If so, use it.

jmarrec (commented):
Added walkthrough of the changes for completeness

lib/mkmf.rb
```diff
@@ -727,7 +727,7 @@ def with_ldflags(flags)
 
   # :nodoc:
   def try_ldflags(flags, werror: $mswin, **opts)
-    try_link(MAIN_DOES_NOTHING, flags, werror: werror, **opts)
+    try_link(MAIN_DOES_NOTHING, "", ldflags: flags, werror: werror, **opts)
```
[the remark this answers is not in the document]
jmarrec: (Not that I am not very used to kwargs in ruby, so it's very possible this isn't a good implementation, but the bottom line is that we do want them passed as ldflags, and I'm sure the maintainers will know what to do given their experience!)

lib/mkmf.rb
```diff
@@ -1957,11 +1957,34 @@ def fmt.%(x)
...
+              opts = Shellwords.shellwords(opts).map { |s|
+                if s.start_with?('-l')
+                  "#{s[2..]}.lib"
+                elsif s.start_with?('-L')
+                  "/libpath:#{s[2..]}"
+                else
+                  s
+                end
+              }.quote.join(" ")
+            end
```
nobu, line 1985: Maybe msvc-syntax is not needed?

```suggestion
            libarg, libpath = LIBARG, LIBPATHFLAG.strip
            opts = opts.shellsplit.map { |s|
              if s.start_with?('-l')
                libarg % s[2..]
              elsif s.start_with?('-L')
                libpath % s[2..]
              else
                s
              end
            }.quote.join(" ")
```

jmarrec: Yeah I don't know. I kinda like the idea of defering to the pkgconf since that's the tool's job. But maybe there isn't much to do to switch the syntax from Nix to MSVC.

Up to you.

The MSVC renderer from pkgconf is here, doesn't appear to do a lot of complicated things: https://github.com/pkgconf/pkgconf/blob/b6e04e2d47b4ca093632b9efde6303a137ea0634/cli/renderer-msvc.c#L133

lib/mkmf.rb
```diff
@@ -1957,11 +1957,33 @@ def fmt.%(x)
...
       if pkgconfig
+        has_ms_win_syntax = false
+        if $mswin
+          has_ms_win_syntax = xpopen([pkgconfig, "--help"]).read.include?('msvc-syntax')
+          if has_ms_win_syntax
+            args << "--msvc-syntax"
+          else
+            Logging.message("WARNING: #{pkgconfig} does not support the --msvc-syntax. Try using a recent pkgconf instead")
+          end
+        end
```
nobu, line 1968: This seems no longer needed.

```suggestion
```

jmarrec: Could you explain why/how this is not needed anymore please? Thank you

nobu: https://github.com/ruby/ruby/pull/9815/files#diff-3e65e2e886eec8afd3714a64d5a673de4e276d0461299c052562892d7176d9f2R1995-R2006
Now arguments of `-l` and `-L` are always built using `LIBARG` and `LIBPATHFLAG`, even if they are not in msvc-syntax.

lib/mkmf.rb
```diff
@@ -1945,7 +1945,7 @@ def fmt.%(x)
       if pkgconfig = with_config("#{pkg}-config") and find_executable0(pkgconfig)
       # if and only if package specific config command is given
       elsif ($PKGCONFIG ||=
-             (pkgconfig = with_config("pkg-config") {config_string("PKG_CONFIG") || "pkg-config"}) &&
+             (pkgconfig = with_config("pkg-config") {config_string("PKG_CONFIG") || ENV["PKG_CONFIG"] || "pkg-config"}) &&
```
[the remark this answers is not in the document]
nobu: If taking the `PKG_CONFIG` environment variable here, `PKG_CONFIG` constant in test/mkmf/test_pkg_config.rb also should take it.

nobu:
Needs to update test/mkmf/test_pkg_config.rb.

PR #9817, opened 2024-02-02
[Bug #20169] Make io_fwrite safe for compaction

peterzhu2118:
Embedded strings are not safe for system calls without the GVL because compaction can cause pages to be locked causing the operation to fail with EFAULT. This commit changes io_fwrite to use rb_str_tmp_frozen_no_embed_acquire, which guarantees that the return string is not embedded.

### DIFF
--- a/io.c
+++ b/io.c
@@ -1975,7 +1975,7 @@ io_fwrite(VALUE str, rb_io_t *fptr, int nosync)
     if (converted)
         OBJ_FREEZE(str);
 
-    tmp = rb_str_tmp_frozen_acquire(str);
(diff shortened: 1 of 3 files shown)


KJTsanaktsidis (approved):
This looks correct to me - thank you!

io.c
```diff
@@ -1975,7 +1975,7 @@ io_fwrite(VALUE str, rb_io_t *fptr, int nosync)
     if (converted)
         OBJ_FREEZE(str);
 
-    tmp = rb_str_tmp_frozen_acquire(str);
```
ioquatix, line 1978: Would it make sense just to change `rb_str_tmp_frozen_acquire` rather than introducing a new method?

PR #9820, opened 2024-02-02
[PRISM] Implement opt_aset_with

jenshenny:
Part of https://github.com/ruby/prism/issues/2231

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -4142,17 +4142,8 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         }
 
         ID method_id = pm_constant_id_lookup(scope_node, call_node->name);
-        if (node->flags & PM_CALL_NODE_FLAGS_ATTRIBUTE_WRITE) {
-            if (!popped) {
-                PM_PUTNIL;
-            }
-        }
 
-        if (call_node->receiver == NULL) {
-            PM_PUTSELF;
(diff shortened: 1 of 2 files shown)


kddnewton (changes requested):
This looks good!

Question: I think `"foo"&.freeze` might be incorrectly using optimizations now, because it's not checked unless I'm misreading?

jenshenny:
> Question: I think "foo"&.freeze might be incorrectly using optimizations now, because it's not checked unless I'm misreading?

Yeah good catch, when checking Ruby's instruction sequence with freeze, -@, aref and aset, they aren't using optimizations. Added the check here and opened a PR to fix the other cases as well https://github.com/ruby/ruby/pull/9826

prism_compile.c
```diff
@@ -4142,17 +4142,8 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
...
-                PM_PUTNIL;
-            }
-        }
 
-        if (call_node->receiver == NULL) {
-            PM_PUTSELF;
-            pm_compile_call(iseq, call_node, ret, popped, scope_node, method_id, start);
-        }
-        else if ((method_id == idUMinus || method_id == idFreeze) &&
+        if ((method_id == idUMinus || method_id == idFreeze) &&
```
kddnewton, line 4146: I think we need the safe navigation check here as well.

kddnewton (changes requested):

kddnewton (approved):

PR #9824, opened 2024-02-03
Fix test session reuse but expire

nurse:
https://github.com/ruby/ruby/actions/runs/7765423659/job/21180054255


voxik:
Is there by a chance more details what is the root of the issue? Was it reported upstream? I am asking because we are facing the same issue on Fedora, therefore I have found this patch, but it does not really explain much :/

voxik:
Just looking at #9036 which originally disabled the test case, there is even less details :/

PR #9832, opened 2024-02-05
Do not truncate strings inefficiently

s7nfo:
Hey folks,
as I demonstrated in [this blog post](http://blog.mattstuchlik.com/2024/01/31/sneaky-one-liner.html), truncating string with `s[0, x] = ''` seems to lead to unnecessary memory copying. I've identified couple places in Ruby where this pattern is being used and this PR replaces it with the more efficient pattern.

If this look like something you might be interested in merging let me know if I need to do anything else here, it's my first PR against Ruby.

Related PR against this same issue in openssl: https://github.com/ruby/openssl/pull/706

### DIFF
--- a/lib/cgi/core.rb
+++ b/lib/cgi/core.rb
@@ -530,7 +530,7 @@ class CGI
             if head && buf.size > boundary_size
               len = buf.size - boundary_size
               body.print(buf[0, len])
-              buf[0, len] = ''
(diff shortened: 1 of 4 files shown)


s7nfo:
I just read https://github.com/ruby/openssl/pull/706#issuecomment-1926603063, I think this PR might be moot!

PR #9842, opened 2024-02-05
More on IO doc

BurdetteLamar:
Improve documentation for line-oriented IO (which I suspect is of greatest interest to many users).

### DIFF
--- a/io.c
+++ b/io.c
@@ -15172,16 +15172,10 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
  *
  *    f = File.new('t.txt')
  *    $/ = 'l'
- *    p f.gets
- *    p f.gets
- *    p f.gets
+ *    f.gets # => "First l"
+ *    f.gets # => "ine\nSecond l"
+ *    f.gets # => "ine\n\nFourth l"
  *    f.close
- *
- *  Output:
- *
- *    "First l"
- *    "ine\nSecond l"
- *    "ine\n\nFourth l"
  *
  *  ===== Special Line Separator Values
  *
--- a/io.c
+++ b/io.c
@@ -15094,56 +15094,64 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
  *
  *  == Line \IO
  *
- *  You can read an \IO stream line-by-line using these methods:
+ *  \Class \IO supports line-oriented
+ *  {input}[rdoc-ref:IO@Line+Input] and {output}[rdoc-ref:IO@Line+Output]
+ *
+ *  === Line Input
+ *
+ *  \Class \IO supports line-oriented input for
+ *  {files}[rdoc-ref:IO@File+Line+Input] and {IO streams}[rdoc-ref:IO@Stream+Line+Input]
+ *
+ *  ==== \File Line Input
+ *
+ *  You can read lines from a file using these methods:
+ *
+ *  - IO.foreach: Reads each line and passes it to the given block.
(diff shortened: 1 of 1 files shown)


io.c
```diff
@@ -15094,64 +15094,72 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *  \Class \IO supports line-oriented
+ *  {input}[rdoc-ref:IO@Line+Input] and {output}[rdoc-ref:IO@Line+Output]
  *
- *  - IO#each_line: Reads each remaining line, passing it to the given block.
- *  - IO#gets: Returns the next line.
- *  - IO#readline: Like #gets, but raises an exception at end-of-stream.
- *  - IO#readlines: Returns all remaining lines in an array.
+ *  === Line Input
+ *
+ *  \Class \IO supports line-oriented input for
```
nobu, line 15102: At [Auto-Linking](https://github.com/ruby/ruby/blob/master/doc/contributing/documentation_guide.md#auto-linking):
> If not, or if you suppress autolinking, consider forcing monofont.

```suggestion
 *  \Class +IO+ supports line-oriented input for
```

"Class" is "the word in question does not refer to a Ruby entity", so it should be rendered as a normal text.
But "IO" is "the reference is to the current class document", it should be rendered as code, I think.

BurdetteLamar: @nobu, point taken.  There are many of these in this class overview.  Most should be changed as you've indicated;  some may be better rendered as `I/O`.  If okay with you, I'd like to handle in the next PR instead of complicating matters here.

io.c
```diff
@@ -15160,39 +15168,52 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
  *    f.close
  *
- *  There are two special line separators:
+ *  Or by setting global variable <tt>$/</tt>:
+ *
+ *    f = File.new('t.txt')
+ *    $/ = 'l'
+ *    p f.gets
+ *    p f.gets
+ *    p f.gets
```
peterzhu2118, line 15177: I think this is more consistent with the rest of the docs:

```suggestion
 *    f.gets # => "First l"
 *    f.gets # => "ine\nSecond l"
 *    f.gets # => "ine\n\nFourth l"
```

io.c
```diff
@@ -15160,39 +15168,52 @@ set_LAST_READ_LINE(VALUE val, ID _x, VALUE *_y)
...
+ *    "ine\nSecond l"
+ *    "ine\n\nFourth l"
+ *
+ *  ===== Special Line Separator Values
  *
- *  - +nil+: The entire stream is read into a single string:
+ *  Each of the {line input methods}[rdoc-ref:IO@Line+Input]
+ *  accepts two special values for parameter +sep+:
+ *
+ *  - +nil+: The entire stream is to be read ("slurped") into a single string:
```
peterzhu2118, line 15191: Nit: why are we calling it "slurp"? It doesn't seem like a term used anywhere in the docs.

BurdetteLamar: The term is well established in other interpreted languages (Perl and Python in particular), where its meaning is "read it all."

io.c
BurdetteLamar: FIxed (and "Output" below removed).

BurdetteLamar:
@peterzhu2118, if you approve this, please don't merge it;  I can then go ahead and take care of the comment from @nobu (re \IO).

BurdetteLamar:
@peterzhu2118, can we go ahead and get this merged?

peterzhu2118 (approved):

PR #9858, opened 2024-02-06
YJIT: Allow popping before falling back

XrXr:
Popping but not generating any code before returning `None` was allowed
before fallbacks were introduced so this is restoring that support in
the same way. The included test used to trip an assert due to popping
too much.


k0kubun (approved):

maximecb:
Thanks for the nice PRs Alan 👌 :)

PR #9859, opened 2024-02-06
skip parsing with parse.y before dumping prism parse tree

HParker:
Prevously the prism parse tree was being dumped after parsing with tree with parse.y. prism_parsetree doesn't require the results of that parse, so we can skip it in this case

On invalid syntax,

Before:

```
$ ./miniruby --dump=prism_parsetree -e "puts :2hi"
-e:1: syntax error, unexpected integer literal, expecting literal content or terminator or tSTRING_DBEG or tSTRING_DVAR
puts :2hi
-e: compile error (SyntaxError)
```

After:

```
$ ./miniruby --dump=prism_parsetree -e "puts :2hi"
./miniruby: syntax errors found (SyntaxError)
> 1 | puts :2hi
    |       ^ expected a newline or semicolon after the statement
    |       ^ invalid symbol
    |        ^ expected a newline or semicolon after the statement
```

Prism issue: https://github.com/ruby/prism/issues/2322

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -2260,7 +2260,45 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
     ruby_set_argv(argc, argv);
     opt->sflag = process_sflag(opt->sflag);
 
-    if (!(*rb_ruby_prism_ptr())) {
+    if (dump & (DUMP_BIT(prism_parsetree))) {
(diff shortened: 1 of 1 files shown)


kddnewton:
Fixed on `master` since https://github.com/ruby/ruby/pull/9991

PR #9862, opened 2024-02-06
[PRISM] Fix error handling in `pm_parse_prism`

eileencodes:
Following changes made in ruby/prism#2365 this implements error handling for when `pm_string_mapped_init` returns `false`.

Related: ruby/prism#2207

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -7811,7 +7811,13 @@ VALUE
 pm_parse_file(pm_parse_result_t *result, VALUE filepath)
 {
     if (!pm_string_mapped_init(&result->input, RSTRING_PTR(filepath))) {
-        return rb_exc_new3(rb_eRuntimeError, rb_sprintf("Failed to map file: %s", RSTRING_PTR(filepath)));
+#ifdef _WIN32
+        int e = rb_w32_map_errno(GetLastError());
+#else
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -7811,6 +7811,14 @@ VALUE
 pm_parse_file(pm_parse_result_t *result, VALUE filepath)
 {
     if (!pm_string_mapped_init(&result->input, RSTRING_PTR(filepath))) {
+#ifdef _WIN32
+        int e = rb_w32_map_errno(GetLastError());
+#else
+        int e = errno;
+#endif
+
+        rb_syserr_fail(e, RSTRING_PTR(filepath));
```
kddnewton, line 7820: We want to return an error from this function so that callers can do cleanup (free anything they allocated). Could you create the error from this function and return it instead?

kddnewton (changes requested):

prism_compile.c
eileencodes: Oh no, I didn't even remove the following line. OK I think this is fixed now.

kddnewton (approved):

kddnewton:
Nice looks great! Thanks @eileencodes!

PR #9870, opened 2024-02-07
[Feature #20244] Show the conflicting another chdir block

nobu:
https://bugs.ruby-lang.org/issues/20244

Show the conflicting another chdir block.

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -1082,9 +1082,13 @@ chdir_alone_block_p(void)
         if (rb_thread_current() != chdir_lock.thread)
             rb_raise(rb_eRuntimeError, "conflicting chdir during another chdir block");
         if (!block_given) {
-            rb_warn("conflicting chdir during another chdir block");
             if (!NIL_P(chdir_lock.path)) {
-                rb_compile_warn(RSTRING_PTR(chdir_lock.path), chdir_lock.line, "here");
+                rb_warn("conflicting chdir during another chdir block\n"
+                        "%" PRIsVALUE ":%d: note: previous chdir was here",
+                        chdir_lock.path, chdir_lock.line);
+            }
+            else {
+                rb_warn("conflicting chdir during another chdir block");
             }
         }
     }
--- a/dir.c
+++ b/dir.c
@@ -3652,6 +3652,7 @@ void
 Init_Dir(void)
 {
     rb_gc_register_address(&chdir_lock.path);
+    rb_gc_register_address(&chdir_lock.thread);
 
     rb_cDir = rb_define_class("Dir", rb_cObject);
 
--- a/dir.c
+++ b/dir.c
@@ -1043,14 +1043,20 @@ dir_chdir0(VALUE path)
 
 static struct {
     VALUE thread;
+    VALUE path;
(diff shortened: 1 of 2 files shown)


dir.c
```diff
@@ -3626,6 +3651,8 @@ rb_dir_s_empty_p(VALUE obj, VALUE dirname)
 void
 Init_Dir(void)
 {
+    rb_gc_register_address(&chdir_lock.path);
```
XrXr, line 3654:
```suggestion
    rb_gc_register_address(&chdir_lock.path);
    rb_gc_register_address(&chdir_lock.thread);
```

Without this, if GC compaction moves the thread object, comparisons with the current thread would fail. I guess it's also a problem currently but no one is compacting inside chdir 😅

dir.c
```diff
@@ -1041,8 +1041,55 @@ dir_chdir0(VALUE path)
...
+{
+    int block_given = rb_block_given_p();
+    if (chdir_lock.blocking > 0) {
+        if (rb_thread_current() != chdir_lock.thread)
+            rb_raise(rb_eRuntimeError, "conflicting chdir during another chdir block");
+        if (!block_given) {
+            rb_warn("conflicting chdir during another chdir block");
+            if (!NIL_P(chdir_lock.path)) {
+                rb_compile_warn(RSTRING_PTR(chdir_lock.path), chdir_lock.line, "here");
+            }
```
casperisfine, line 1088: Would it be possible to make it a single warning?

Warnings that are split in two are a bit painful for libraries and applications that hook into `Warning.warn`. e.g. https://github.com/jeremyevans/ruby-warning or https://github.com/Shopify/deprecation_toolkit

nobu: It is for editors to jump to each lines.

casperisfine: Sorry I wasn't clear. I didn't mean not to include both locations, but to include them both in a single warning / string.

casperisfine: e.g.:

```ruby
rb_warn("conflicting chdir during another chdir block\nPrevious chdir was here: #{location}")
```

Instead of 

```ruby
rb_warn("conflicting chdir during another chdir block")
rb_warn("Previous chdir was here: #{location}")
```

Otherwise we need to do things like this: https://github.com/Shopify/deprecation_toolkit/blob/main/lib/deprecation_toolkit/warning.rb#L9-L23

nobu: I wanted to jump to both inner `chdir` line and caller `chdir` block line.
To using the messages for tag jump, each needs to be separate lines.

casperisfine: What I'm trying to say it that it can be two lines, and look identical to two `rb_warn` calls, but be a single `rb_warn` call (with a multiline string).

```ruby
rb_warn(
  "conflicting chdir during another chdir block\n"
  "#{location}: warning: previous chdir was here"
);
```

That would work with your editor, no?

But I can also try to make this work later if you want.

nobu: I think I fixed it.
Is it OK?

casperisfine: Looks great. Thank you very much!

PR #9874, opened 2024-02-07
YJIT: Use jit_prepare_call_with_gc as much as possible

k0kubun:
Follow-up on https://github.com/ruby/ruby/pull/9775

I checked the remaining `jit_prepare_non_leaf_call` calls and replaced ones that don't seem to make any call.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2797,9 +2797,9 @@ fn gen_defined(
             gen_block_given(jit, asm, out_opnd, pushval.into(), Qnil.into());
         }
         _ => {
-            // Save the PC and SP because the callee may allocate
+            // Save the PC and SP because the callee may allocate or call #respond_to?
             // Note that this modifies REG_SP, which is why we do it first
-            jit_prepare_call_with_gc(jit, asm);
+            jit_prepare_non_leaf_call(jit, asm);
 
             // Get the operands from the stack
             let v_opnd = asm.stack_opnd(0);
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -2799,7 +2799,7 @@ fn gen_defined(
         _ => {
             // Save the PC and SP because the callee may allocate
             // Note that this modifies REG_SP, which is why we do it first
-            jit_prepare_non_leaf_call(jit, asm);
+            jit_prepare_call_with_gc(jit, asm);
```
XrXr, line 2802: Hopefully this is issue causing assert_ctx to trip?

`verify_ctx: ctx type (Fixnum) incompatible with actual value of local: T_NIL (type Nil)`

You can call methods with side effects inside `defined?`.

k0kubun: Yeah, that was it https://github.com/ruby/ruby/pull/9874/commits/3567148a38d7efd96e4e58af1f10ca959fc13031. 

I knew expressions in `defined?` can make a call, but ISEQs have method calls before `defined` instruction, so I still thought `defined` instruction itself is leaf. It turns out it could still make a `respond_to?` call inside `defined` instructions, so it was not leaf.

Maybe we could still switch this if `defined_type` is not `DEFINED_FUNC` or `DEFINED_METHOD`, but that's too specific to `defined` instruction, so I'll leave it to another PR.

PR #9884, opened 2024-02-08
Replace assert with RUBY_ASSERT

peterzhu2118:
assert does not print the bug report, only the file and line number of the assertion that failed. RUBY_ASSERT prints the full bug report, which makes it much easier to debug.

### DIFF
--- a/array.c
+++ b/array.c
@@ -77,47 +77,47 @@ should_be_T_ARRAY(VALUE ary)
     return RB_TYPE_P(ary, T_ARRAY);
 }
 
-#define ARY_HEAP_PTR(a) (assert(!ARY_EMBED_P(a)), RARRAY(a)->as.heap.ptr)
-#define ARY_HEAP_LEN(a) (assert(!ARY_EMBED_P(a)), RARRAY(a)->as.heap.len)
-#define ARY_HEAP_CAPA(a) (assert(!ARY_EMBED_P(a)), assert(!ARY_SHARED_ROOT_P(a)), \
+#define ARY_HEAP_PTR(a) (RUBY_ASSERT(!ARY_EMBED_P(a)), RARRAY(a)->as.heap.ptr)
+#define ARY_HEAP_LEN(a) (RUBY_ASSERT(!ARY_EMBED_P(a)), RARRAY(a)->as.heap.len)
+#define ARY_HEAP_CAPA(a) (RUBY_ASSERT(!ARY_EMBED_P(a)), RUBY_ASSERT(!ARY_SHARED_ROOT_P(a)), \
                           RARRAY(a)->as.heap.aux.capa)
 
-#define ARY_EMBED_PTR(a) (assert(ARY_EMBED_P(a)), RARRAY(a)->as.ary)
+#define ARY_EMBED_PTR(a) (RUBY_ASSERT(ARY_EMBED_P(a)), RARRAY(a)->as.ary)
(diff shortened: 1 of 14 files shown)


peterzhu2118:
But when I use a `assert(false)`, I get:

```
Assertion failed: (false), function ary_new, file array.c, line 700.
```

Whereas, `RUBY_ASSERT(false)` is:

```
Assertion Failed: ../array.c:700:ary_new:false
ruby 3.4.0dev (2024-02-08T15:22:07Z master b74c8abd11) [arm64-darwin23]

-- Crash Report log information --------------------------------------------
   See Crash Report log file in one of the following locations:             
     * ~/Library/Logs/DiagnosticReports                                     
     * /Library/Logs/DiagnosticReports                                      
   for more details.                                                        
Don't forget to include the above Crash Report log file in bug reports.     

c:0001 p:---- s:0003 e:000002 DUMMY  [FINISH]


Total ractor count: 0
Ruby thread count for this ractor: 0

/Users/peter/src/ruby/build/miniruby(rb_vm_bugreport+0xb60) [0x1051f4154] ../vm_dump.c:1151
/Users/peter/src/ruby/build/miniruby(rb_vm_bugreport) (null):0
/Users/peter/src/ruby/build/miniruby(rb_assert_failure_detail+0xa4) [0x1052fdb40] ../error.c:1146
/Users/peter/src/ruby/build/miniruby(rb_assert_failure_detail+0x0) [0x1052fda9c] ../error.c:1126
/Users/peter/src/ruby/build/miniruby(rb_assert_failure) (null):0
/Users/peter/src/ruby/build/miniruby(rb_ary_hidden_new_fill+0x0) [0x104f7001c] ../array.c:700
/Users/peter/src/ruby/build/miniruby(ary_new) (null):0
/Users/peter/src/ruby/build/miniruby(rb_ary_hidden_new) ../array.c:847
/Users/peter/src/ruby/build/miniruby(Init_vm_objects+0x24) [0x1051da7e8] ../vm.c:4240
/Users/peter/src/ruby/build/miniruby(rb_current_execution_context+0x0) [0x1050188cc] ../eval.c:83
/Users/peter/src/ruby/build/miniruby(ruby_setup) ../eval.c:85
/Users/peter/src/ruby/build/miniruby(ruby_init+0x10) [0x105018ab0] ../eval.c:99
/Users/peter/src/ruby/build/miniruby(rb_main+0xc) [0x104f6e66c] ../main.c:38
/Users/peter/src/ruby/build/miniruby(main) ../main.c:58
```

peterzhu2118:
It looks like the [implementation of `assert.h`](https://codebrowser.dev/glibc/glibc/assert/assert.h.html) doesn't have header guards, so any `#include <assert.h>` after the first `#include "ruby/assert.h"` will override the `#define assert RUBY_ASSERT_NDEBUG`.

nobu:
`assert.h` doesn’t  have the guard because the standard `assert` should be able to redefined according to `NDEBUG` every time the file is included.

peterzhu2118:
> assert.h doesn’t have the guard because the standard assert should be able to redefined according to NDEBUG every time the file is included.

I'm not sure what the optimal solution is then. Do we replace all `assert` with `RUBY_ASSERT`? Do we forbid using `assert.h` and only allow using `ruby_assert.h`?

peterzhu2118:
@nobu can I merge this or do you prefer this solved using another way?

PR #9891, opened 2024-02-08
YJIT: Add support for `**kwrest` parameters

XrXr:
Now that `...` uses `**kwrest` instead of regular splat and
ruby2keywords, we need to support these type of methods to
support `...` well.

On `lobsters`, this cuts `num_send_dynamic` by ~1%:
```
-num_send_dynamic:          1,089,894 ( 4.1%)
+num_send_dynamic:            874,933 ( 3.3%)
```


maximecb (approved):

yjit/src/codegen.rs
```diff
@@ -6961,114 +6958,189 @@ fn gen_send_iseq(
...
+        for _ in caller_keyword_len..callee_kw_count {
+            argc += 1;
+            asm.stack_push(Type::Unknown);
+        }
+        // Now this is the stack_opnd() index to the 0th keyword argument.
+        let kwargs_stack_base = kwargs_order.len() as i32 - 1;
+
+        // Build the keyword rest parameter hash before we make any changes to the order of
+        // the supplied keyword arguments
+        if has_kwrest {
```
k0kubun, line 6988: Do you think it makes sense to carve out the kwrest logic as a separate function? It's generally hard to follow the sequence of operations in `gen_send_iseq`, so I wonder if it makes it easier to understand what happens after what in that function.

k0kubun (approved):

yjit/src/codegen.rs
XrXr: Make sense. I think we should move all the keywords handling out, though. I'll make a PR.

PR #9897, opened 2024-02-09
Fix [Bug #20246]: Don't set next_head_exact if a capture is called

makenowjust:


bestwebua:
@makenowjust Thanks! 🔥

PR #9898, opened 2024-02-09
Split line_no and node_id before new_insn_body

kddnewton:
Before this commit, there were many places where we had to generate dummy line nodes to hold both the line number and the node id that would then immediately get pulled out from the created node. Now we pass them explicitly so that we don't have to generate these nodes.

This makes a clearer line between the parser and compiler, and also makes it easier to generate instructions when we don't have a specific node to tie them to. As such, it removes almost every single place where we needed to previously generate dummy nodes.

This also makes it easier for the prism compiler, because now we can pass in line number and node id instead of trying to generate dummy nodes for every instruction that we compile.

@ko1 would this be okay? This would really help out the prism compiler, and I think it's good for `compile.c` (you can see it simplifies a lot of the callsites without having to generate dummy line nodes).

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -222,30 +222,34 @@ const ID rb_iseq_shared_exc_local_tbl[] = {idERROR_INFO};
 
 /* add an instruction */
 #define ADD_INSN(seq, line_node, insn) \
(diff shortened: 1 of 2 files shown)


ko1:
I have no objection. (I'm not a author of `node_id` related code though)

PR #9899, opened 2024-02-09
rb_profile_thread_frames: Immediately return if passed thread is dead

osyoyu:
Accept completed Threads as rb_profile_thread_frames' first argument.

Currently, the caller needs to check if the Thread is complete before calling rb_profile_thread_frames(). This requires an extra rb_funcall to `Thread#status` and also has a possibility of a race condition (the Thread could quit between the Thread#status check and the rb_profile_thread_frames() call).


osyoyu:
There was no need for this, since `!(th->ec->cfp)` is checked inside thread_profile_frames().

PR #9900, opened 2024-02-09
Introduce and default to "minimal" crash report mode

osyoyu:
A PoC for https://bugs.ruby-lang.org/issues/20249.

### DIFF
--- a/vm_dump.c
+++ b/vm_dump.c
@@ -1107,6 +1107,19 @@ rb_vm_bugreport(const void *ctx, FILE *errout)
         }
     }
 
+    // Print a "minimal" crash report when errout is a TTY
+    int full_crash_report = !(isatty(fileno(errout)));
+    // RUBY_FULL_CRASH_REPORT takes precedence over the default
+    const char *ruby_full_crash_report_env = getenv("RUBY_FULL_CRASH_REPORT");
+    if (ruby_full_crash_report_env != NULL) {
+        if (strcmp(ruby_full_crash_report_env, "0") == 0) {
+            full_crash_report = false;
+        }
+        else {
+            full_crash_report = true;
+        }
+    }
+
     // Thread unsafe best effort attempt to stop printing the bug report in an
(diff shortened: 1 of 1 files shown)


vm_dump.c
```diff
@@ -1107,6 +1107,19 @@ rb_vm_bugreport(const void *ctx, FILE *errout)
         }
     }
 
+    // Print a "minimal" crash report when errout is a TTY
+    int full_crash_report = !(isatty(fileno(errout)));
```
nobu, line 1111:
```suggestion
    bool full_crash_report = !(isatty(fileno(errout)));
```

vm_dump.c
```diff
@@ -1107,6 +1107,19 @@ rb_vm_bugreport(const void *ctx, FILE *errout)
...
+    // RUBY_FULL_CRASH_REPORT takes precedence over the default
+    const char *ruby_full_crash_report_env = getenv("RUBY_FULL_CRASH_REPORT");
+    if (ruby_full_crash_report_env != NULL) {
+        if (strcmp(ruby_full_crash_report_env, "0") == 0) {
+            full_crash_report = false;
+        }
+        else {
+            full_crash_report = true;
+        }
```
nobu, line 1120: This makes `RUBY_FULL_CRASH_REPORT= ruby crash.rb` full dump.
I'd expect that empty value would disable the feature.
```suggestion
        full_crash_report = strtoul(ruby_full_crash_report_env, 0, 0) != 0;
```

Or even:
```suggestion
    bool full_crash_report = ruby_full_crash_report_env  ?
        strtoul(ruby_full_crash_report_env, 0, 0) != 0 :
        !isatty(fileno(errout));
```

osyoyu: Thank you! I agree that `RUBY_FULL_CRASH_REPORT= ruby crash.rb` should disable the feature.
Fixed in 4a3ebd4a4af7574a88bce7db2d344c2edfd0cc68.

PR #9901, opened 2024-02-09
[ruby/optparse] Respect default values in block parameters

nobu:
Fix https://github.com/ruby/optparse/pull/55

https://github.com/ruby/optparse/commit/e6be9b71d0


nobu:
Rake uses [lambdas as callbacks](https://github.com/ruby/rake/blob/master/lib/rake/application.rb#L543).
Calling it without omitted argument raises an `ArgumentError`.

PR #9903, opened 2024-02-09
rb_obj_setup: do not copy RUBY_FL_SEEN_OBJ_ID

casperisfine:
[[Bug #20250]](https://bugs.ruby-lang.org/issues/20250)

We're seting up a new instance, so it never had an associated object_id.

### DIFF
--- a/object.c
+++ b/object.c
@@ -120,7 +120,7 @@ rb_obj_reveal(VALUE obj, VALUE klass)
 VALUE
 rb_obj_setup(VALUE obj, VALUE klass, VALUE type)
 {
-    VALUE ignored_flags = RUBY_FL_PROMOTED;
+    VALUE ignored_flags = RUBY_FL_PROMOTED | RUBY_FL_SEEN_OBJ_ID;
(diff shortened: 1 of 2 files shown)


object.c
```diff
@@ -120,7 +120,7 @@ rb_obj_reveal(VALUE obj, VALUE klass)
 VALUE
 rb_obj_setup(VALUE obj, VALUE klass, VALUE type)
 {
-    VALUE ignored_flags = RUBY_FL_PROMOTED;
+    VALUE ignored_flags = RUBY_FL_PROMOTED | RUBY_FL_SEEN_OBJ_ID;
```
casperisfine, line 123: I think it would be valuable to define some mask such as `RUBY_FL_INHERITED_ON_CLONE` in `fl_type.h` to make it easier to reason about. But since this is meant as a bugfix, I'd rather make the change as small as possible to make it easier to backport.

Also this line changed in 3.3, so the backport to 3.2 and 3.1 will conflict, but it should be easy to resolve.

peterzhu2118 (approved):

PR #9907, opened 2024-02-09
More on auto-link

BurdetteLamar:
@nobu [suggests](https://github.com/ruby/ruby/pull/9842#discussion_r1481348817) that a reference to a class such as `IO` (even the local one) should be `+IO+` (to force monofont), not `\IO` (plain text).  In my view, if it's not plain text, it may as well be auto-linked (`IO`), since both forced monofont and auto-linking cause both the font and the background color to be changed.

Therefore I propose allowing auto-linking in almost all cases.

This proposed change does not imply that current docs should be revised for this purpose, but only that additions or revisions should consider this.

### DIFF
--- a/doc/contributing/documentation_guide.md
+++ b/doc/contributing/documentation_guide.md
@@ -173,14 +173,10 @@ Code that is a simple string should include the quote marks.
 In general, \RDoc's auto-linking should not be suppressed.
 For example, we should write `Array`, not `\Array`.
 
-We might consider whether to suppress when:
+Suppress when:
 
 - The word in question does not refer to a Ruby entity
   (e.g., some uses of _Class_ or _English_).
-- The reference is to the current class document
-  (e.g., _Array_ in the documentation for class `Array`).
-- The same reference is repeated many times
-  (e.g., _RDoc_ on this page).
 - The reference is to a class or module that users
   usually don't deal with, including these:
 
--- a/doc/contributing/documentation_guide.md
+++ b/doc/contributing/documentation_guide.md
@@ -178,7 +178,9 @@ However, suppress when the word in question:
 - Does not refer to a Ruby entity
   (e.g., some uses of _Class_ or _English_).
 - Refers to the current document
-  (e.g., _Array_ in the documentation for class `Array`).
+  (e.g., _Array_ in the documentation for class `Array`);
+  in that case, the word should be forced to
+  [monofont](rdoc-ref:RDoc::MarkupReference@Monofont).
 
 Most often, the name of a class, module, or method
 will be auto-linked:
--- a/doc/contributing/documentation_guide.md
+++ b/doc/contributing/documentation_guide.md
@@ -173,19 +173,11 @@ Code that is a simple string should include the quote marks.
 In general, \RDoc's auto-linking should not be suppressed.
 For example, we should write `Array`, not `\Array`.
 
-Suppress when:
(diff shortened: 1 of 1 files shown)


doc/contributing/documentation_guide.md
```diff
@@ -173,14 +173,10 @@ Code that is a simple string should include the quote marks.
...
+Suppress when:
 
 - The word in question does not refer to a Ruby entity
   (e.g., some uses of _Class_ or _English_).
-- The reference is to the current class document
-  (e.g., _Array_ in the documentation for class `Array`).
-- The same reference is repeated many times
-  (e.g., _RDoc_ on this page).
 - The reference is to a class or module that users
   usually don't deal with, including these:
```
jeremyevans, line 181: I recommend removing the last bullet as well.  It mentions `Class`, `Module`, and `Method`.  First, I'm not sure users don't usually deal with them.  Second, even if they don't usually deal with them, there is no problem auto-linking when referring to the Ruby class/entity.

I think the only time auto-linking should be supressed is if the usage does not refer to the Ruby class/entity.  Maybe:

```
In general, \RDoc's auto-linking should not be suppressed.
For example, we should write `Array`, not `\Array`.
However, suppress when the word in question does not refer to the Ruby entity
(e.g., some uses of _Class_ or _English_).
```

jeremyevans (changes requested):

doc/contributing/documentation_guide.md
BurdetteLamar: Done.

jeremyevans (approved):
You may want to remove the blank line above the new paragraph, but that's up to you.

peterzhu2118:
> In my view, if it's not plain text, it may as well be auto-linked (IO), since both forced monofont and auto-linking cause both the font and the background color to be changed.

I don't agree. Using monofont signals that it's code, but linking to the same page (except linking to a particular section) is pointless and distracts the user from which links are useful to click on.

BurdetteLamar:
> > In my view, if it's not plain text, it may as well be auto-linked (IO), since both forced monofont and auto-linking cause both the font and the background color to be changed.
> 
> I don't agree. Using monofont signals that it's code, but linking to the same page (except linking to a particular section) is pointless and distracts the user from which links are useful to click on.

You'd be comfortable with code for page self-reference, and auto-links for others?

doc/contributing/documentation_guide.md
```diff
@@ -171,9 +171,14 @@ Code that is a simple string should include the quote marks.
 ### Auto-Linking
 
 In general, \RDoc's auto-linking should not be suppressed.
-For example, we should write `Array`, not `\Array`.
-However, suppress when the word in question does not refer to a Ruby entity
-(e.g., some uses of _Class_ or _English_).
+For example, we should write `Array`, not `\Array`
```
nobu, line 174: No period at the end?

nobu (approved):

doc/contributing/documentation_guide.md
BurdetteLamar: Fixed.

BurdetteLamar:
Once we have consensus/approval for the specific reviewers here, I'd like to leave this open for a few days for any others to comment.

jeremyevans (approved):

peterzhu2118 (approved):

PR #9908, opened 2024-02-09
Get #object_id working across ractors for moved objects

luke-gru:


ko1:
I think there is no discussion about equality of moved objects and I think it can be different object ID, because original objects were prohibited to touch.

luke-gru:
Okay that makes sense. In that case I think the object_id should be removed from the global table if `FL_SEEN_OBJ_ID` is set for the original object.

PR #9911, opened 2024-02-09
YJIT: Fix insufficient type guards

k0kubun:
I think we copied `Type::UnknownHeap.diff(object_type)` from `guard_object_is_heap`, but the guarded type should match the one it's being upgraded to.


maximecb (approved):
Good catch 👍

PR #9918, opened 2024-02-10
Fix moving embedded arrays across ractors

luke-gru:
The array values must be copied to the new embedded array's space.

[Bug #20255]

### DIFF
--- a/include/ruby/internal/intern/array.h
+++ b/include/ruby/internal/intern/array.h
@@ -89,6 +89,11 @@ VALUE rb_ary_new(void);
 VALUE rb_ary_new_capa(long capa);
 
 /**
+ * Identical to rb_ary_new_capa(), except it additionally specifies a klass
+ */
+VALUE rb_ary_new_capa_with_klass(VALUE klass, long capa);
+
+/**
  * Constructs an array from the passed objects.
  *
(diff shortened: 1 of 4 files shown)


luke-gru:
I added a new allocation function for arrays `rb_ary_new_capa_with_klass` because if the old allocation function is used here (`rb_obj_alloc`), then a 0-sized array is created. Then if you push into it during each iteration it might not be embedded anymore because capa doubles each time it reaches capacity. Instead I create a same size array, and if that array is also embedded then I just write into its space.

include/ruby/internal/intern/array.h
```diff
@@ -88,6 +88,11 @@ VALUE rb_ary_new(void);
  */
 VALUE rb_ary_new_capa(long capa);
 
+/**
+ * Identical to rb_ary_new_capa(), except it additionally specifies a klass
+ */
+VALUE rb_ary_new_capa_with_klass(VALUE klass, long capa);
```
ko1, line 94: sorry for long absent.
we shouldn't expose new C-API even if Ruby internal requires it.
it should be in internal/array.h.

byroot:
This bug was solved by another PR.

PR #9920, opened 2024-02-11
Burn default ASAN options into the built Ruby

KJTsanaktsidis:
* We always need use_sigaltstack=0 because Ruby registers sigaltstack handlers
* We also need to disable leak detection (unless RUBY_FREE_AT_EXIT is set - I might experiment later with automatically enabling leak detection if RUBY_FREE_AT_EXIT is set).

Burning it into the built ruby binary in this way avoids people needing to remember to start their Ruby program with these flags all the time.

We also need a small fix in mkmf to make sure that test programs also don't have leak detection enabled (this is never desirable)

[Bug #20256]

### DIFF
--- a/main.c
+++ b/main.c
@@ -20,6 +20,7 @@
 #undef RUBY_EXPORT
 #include "ruby.h"
 #include "vm_debug.h"
+#include "internal/sanitizers.h"
 #ifdef HAVE_LOCALE_H
 #include <locale.h>
 #endif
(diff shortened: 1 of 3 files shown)


main.c
```diff
@@ -57,3 +58,12 @@ main(int argc, char **argv)
...
 }
+
+#ifdef RUBY_ASAN_ENABLED
+/* Compile in the ASAN options Ruby needs, rather than relying on environment variables, so
+ * that even tests which fork ruby with a clean environment will run ASAN with the right
+ * settings */
+const char *__asan_default_options(void) {
+  return "use_sigaltstack=0:detect_leaks=0";
+}
+#endif
```
nobu, line 69: Does this need to be placed in the main executable file?

PR #9923, opened 2024-02-12
[Feature #20257] Rearchitect Ripper

yui-knk:
https://bugs.ruby-lang.org/issues/20257

Introduce another semantic value stack for Ripper so that Ripper can manage both Node and Ruby Object separately.
This rearchitectutre of Ripper solves these issues.
Therefore adding test cases for them.

* [Bug 10436] https://bugs.ruby-lang.org/issues/10436
* [Bug 18988] https://bugs.ruby-lang.org/issues/18988
* [Bug 20055] https://bugs.ruby-lang.org/issues/20055


yui-knk:
@kddnewton 

This PR fixes ripper bugs and makes lex state transition of parser and ripper to be aligned. However it makes one of the Prism test failed. I wrote the detail on [the second commit](https://github.com/ruby/ruby/pull/9923/commits/97ca269b89a0c049b0ee865cd12c59ce1d1591ed). What do you think about it?

kddnewton:
> This PR fixes ripper bugs and makes lex state transition of parser and ripper to be aligned. However it makes one of the Prism test failed. I wrote the detail on [the second commit](https://github.com/ruby/ruby/pull/9923/commits/97ca269b89a0c049b0ee865cd12c59ce1d1591ed). What do you think about it?

In this case the parser state of prism is incorrect, as it should be in the ARG state after `bar`. The test failure is legitimate in that case, so we shouldn't skip the test, we should fix the problem.

yui-knk:
Then I want to merge this PR with the second commit, is it ok for you? I you have another option, please let me know.

kddnewton:
Please implement a fix yourself or wait until I implement one. I don't think it's a good idea to just comment out tests that break when they are legitimate issues.

yui-knk:
@kddnewton I implemented workaround and change to not modify "test/prism/parse_test.rb". What do you think about it?

kddnewton:
That seems to mask the problem. I'll take care of this first thing Monday, thank you for your patience.

yui-knk:
This PR changes a lot of line of parse.y and my other commits are created based on this change.
I will merge this PR.

PR #9926, opened 2024-02-12
proc.c: get rid of `CLONESETUP`

casperisfine:
[[Bug #20253]](https://bugs.ruby-lang.org/issues/20253)

All the way down to Ruby 1.9, `Proc`, `Method`, `UnboundMethod` and `Binding` always had their own specific clone and dup routine.

This caused various discrepancies with how other objects behave on `dup` and `clone. [Bug #20250], [Bug #20253].

This commit get rid of `CLONESETUP` and use the the same codepath as all other types, so ensure consistency.

NB: It's still not accepting the `freeze` keyword argument on `clone`.

cc @peterzhu2118

### DIFF
--- a/object.c
+++ b/object.c
@@ -454,15 +454,12 @@ immutable_obj_clone(VALUE obj, VALUE kwfreeze)
     return obj;
 }
 
-static VALUE
-mutable_obj_clone(VALUE obj, VALUE kwfreeze)
+VALUE
+rb_obj_clone_setup(VALUE obj, VALUE clone, VALUE kwfreeze)
 {
-    VALUE clone, singleton;
     VALUE argv[2];
 
-    clone = rb_obj_alloc(rb_obj_class(obj));
-
-    singleton = rb_singleton_class_clone_and_attach(obj, clone);
+    VALUE singleton = rb_singleton_class_clone_and_attach(obj, clone);
     RBASIC_SET_CLASS(clone, singleton);
     if (FL_TEST(singleton, FL_SINGLETON)) {
         rb_singleton_class_attached(singleton, clone);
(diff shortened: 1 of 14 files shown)


eregon:
Could we use the clone/dup from Kernel and avoid defining custom dup/clone for these types?
Custom dup/clone seem error-prone.

casperisfine:
> Could we use the clone/dup from Kernel and avoid defining custom dup/clone for these types?

I'm afraid not, for multiple reasons:

  - `Kernel#clone` and `Kernel#dup` call `rb_obj_alloc` but these classes have `rb_undef_alloc_func`.
  - For `Proc`, the `->block` is also copied and mutated.
  - For `Method`, the `->me` (method entry) is cloned too.

So they do need some for of specialization. One possibility could be to implement all that in `initialize_copy`, but then it's unclear what the consequences would be if it was redefined without calling super (potential crash or corruption?).

So this refactoring seemed like a reasonable enough middleground.

eregon:
Re `rb_obj_alloc` right one would need a way to deal with special cases like that one.
TruffleRuby already does this by abstracting that behind some "basic copy" functionality which is basically `rb_obj_alloc` + handling special cases.

Regarding the rest I think it makes sense to move that to `initialize_copy` since that's where this stuff lives in other classes.

Indeed it feels a bit unsafe that we don't have the guarantee e.g. that initialize/initialize_copy is called after allocate/the alloc function. But it's already the case for all other core classes.

Reusing more of the Kernel#dup/clone implementation like this PR does feels like a good step forward :+1:

object.c
```diff
@@ -529,13 +526,28 @@ mutable_obj_clone(VALUE obj, VALUE kwfreeze)
...
+}
+
 VALUE
 rb_obj_clone(VALUE obj)
 {
     if (special_object_p(obj)) return obj;
     return mutable_obj_clone(obj, Qnil);
 }
 
+VALUE rb_obj_dup_setup(VALUE obj, VALUE dup)
```
peterzhu2118, line 543: Nit:

```suggestion
VALUE
rb_obj_dup_setup(VALUE obj, VALUE dup)
```

peterzhu2118 (approved):

PR #9929, opened 2024-02-12
Try `nmake install` before `nmake test-spec`

eregon:


eregon:
This [works](https://github.com/ruby/ruby/actions/runs/7872142078/job/21476753507?pr=9929) and fixes CI, so I'll merge it.
@nobu If there is a better solution, please commit it.

PR #9931, opened 2024-02-12
Use RJIT for make test-spec

kddnewton:
Previously only the test runner was using RJIT. This switches the Ruby that is being tested to use RJIT. cc @k0kubun 

I just noticed this because of prism trying to run the specs.

### DIFF
--- a/.github/workflows/rjit.yml
+++ b/.github/workflows/rjit.yml
@@ -95,11 +95,12 @@ jobs:
 
       - name: make test-spec
         run: |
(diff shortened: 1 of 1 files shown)


k0kubun:
Thanks for the PR. Since it's failing, we'd need to find which test causes it to fail/timeout, and skip them for now.

Though it's probably more important to do this for YJIT than making effort for RJIT. None of our CIs seems to test `SPECOPTS=--yjit-call-threshold=1` on `make test-spec` (as part of `make check`).

PR #9934, opened 2024-02-12
[PRISM] Provide runtime flag for prism in iseq

kddnewton:
The reason we want this is that we want to store prism-specific node ids in the iseq in order to support error highlight. So we need to be able to map from the compiler that produced the iseq to the node ids that are contained in the iseq.

This adds a single bool to `rb_iseq_constant_body` and adds `:prism` to the `misc` hash in the serialized array form of an iseq.

Note that this PR includes https://github.com/ruby/error_highlight/pull/41.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -253,6 +253,11 @@ ast_s_of(rb_execution_context_t *ec, VALUE module, VALUE body, VALUE keep_script
     if (!iseq) {
         return Qnil;
     }
+
+    if (ISEQ_BODY(iseq)->prism) {
+        rb_raise(rb_eRuntimeError, "cannot get AST for ISEQ compiled by prism");
+    }
+
     lines = ISEQ_BODY(iseq)->variable.script_lines;
 
     VALUE path = rb_iseq_path(iseq);
--- a/compile.c
+++ b/compile.c
@@ -12740,6 +12740,7 @@ ibf_dump_iseq_each(struct ibf_dump *dump, const rb_iseq_t *iseq)
     ibf_dump_write_small_value(dump, body->ci_size);
     ibf_dump_write_small_value(dump, body->stack_max);
     ibf_dump_write_small_value(dump, body->builtin_attrs);
+    ibf_dump_write_small_value(dump, body->prism ? 1 : 0);
 
 #undef IBF_BODY_OFFSET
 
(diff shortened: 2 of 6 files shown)


mame (approved):
Looks good to me, but I would like @ko1 to review this too.

ko1:
Two comments:

* There is a comment about `#to_a` format, which should not contain `prism: true` flag but `parser: :prism`. This allows us to increase further parsers. Now we only have two parsers so it is okay to have `rb_iseq_t::prism` (I prefer `bool prism_parser` but not big issue).
* This PR contains `BacktraceLocatioin#node_id` but I'm not sure it is approved by Matz

kddnewton:
@ko1 no problem, I've asked for matz's approval on the redmine ticket. I'll wait on merging this until we get that.

kddnewton:
@ko1 @mame we could also make that `RubyVM::InstructionSequence.node_id_for(location)`. Then it might be more clear that it's restricted to `RubyVM` and it wouldn't be such a public API. Would that be better?

mame:
@kddnewton I just remembered! There is already `RubyVM::AbstractSyntaxTree.node_id_for_backtrace_location(loc)`. I think you can proceed with this for now.

https://bugs.ruby-lang.org/issues/19083

It would be good to create a separate ticket to introduce an alias `Thread::Backtrace::Location#node_id` or `RubyVM::InstructionSequence.node_id_for_backtrace_location(location)` if you mind that the belonging class is `RubyVM::AbstractSyntaxTree`.

kddnewton:
Ahh that's great @mame thanks. I don't mind it being on `RubyVM::AbstractSyntaxTree` for now.

ko1:
How about that?

> There is a comment about #to_a format, which should not contain prism: true flag but parser: :prism. This allows us to increase further parsers. Now we only have two parsers so it is okay to have rb_iseq_t::prism (I prefer bool prism_parser but not big issue).

kddnewton:
I'm sorry if I misunderstood, I thought you were saying you preferred parser_prism but were okay with just prism. I'm happy to change it if you'd like.

The reason I wanted to go with this was because it's coming from the prism compiler, not the prism parser. In the future even if prism isn't chosen matz has said the AST will be prism's, so there won't be a need for multiple compilers. So I thought having a parser flag would be confusing, since it's more about the compiler that generated the ISEQ.

That being said, I am happy to change it!

mame:
@kddnewton I do not blame you for misinterpreting his explanation, but what ko1 wanted to say is that `parser: :prism` is better for the external representation (return value of `ISeq#to_a`), while bool is good enough for the internal representation. 

But I am unsure what `ISeq#to_binary` should return. Currently, 0 represents for the traditional compiler, 1 represents for prism compiler. @ko1 Is it ok? Or should it include a string "prism"?

kddnewton:
Oh @mame thank you I understand now! I will change that first thing tomorrow.

I am also happy to change the binary representation, whatever you think makes sense, just let me know.

PR #9935, opened 2024-02-12
Revert "YJIT: Add support for `**kwrest` parameters"

XrXr:
Reverts GH-9891
Seems to be causing a crash during GC: https://github.com/ruby/ruby/actions/runs/7877298133/job/21493179294


maximecb (approved):

XrXr:
Actually, I think I have a simple fix, so I'll make a PR to fix forward instead.

PR #9939, opened 2024-02-12
Specialize String#byteslice(a, b)

tenderlove:
This adds a specialization for String#byteslice when there are two parameters.

This makes our protobuf parser go from 5.84x slower to 5.33x slower

```
Comparison:
decode upstream (53738 bytes):     7228.5 i/s
decode protobuff (53738 bytes):     1236.8 i/s - 5.84x  slower

Comparison:
decode upstream (53738 bytes):     7024.8 i/s
decode protobuff (53738 bytes):     1318.5 i/s - 5.33x  slower
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5304,6 +5304,38 @@ fn jit_rb_str_bytesize(
     true
 }
 
+fn jit_rb_str_byteslice(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    _ocb: &mut OutlinedCb,
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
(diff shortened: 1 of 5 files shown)


yjit/src/codegen.rs
```diff
@@ -5304,6 +5304,38 @@ fn jit_rb_str_bytesize(
...
+    // Raises when non-integers are passed in
+    jit_prepare_non_leaf_call(jit, asm);
+
+    let len = asm.stack_opnd(0);
+    let beg = asm.stack_opnd(1);
+    let recv = asm.stack_opnd(2);
+    let ret_opnd = asm.ccall(rb_str_byte_substr as *const u8, vec![recv, beg, len]);
+    asm.stack_pop(3);
+
+    // Can either return a FIXNUM or nil
```
maximecb, line 5332:
```suggestion
```

PR #9942, opened 2024-02-13
Move rbinc files into separate directory

nobu:


k0kubun:
If the motivation is to eventually move everything to a sub-directory and this is part of the effort, could we just call it `src`? `rblib` seems so unique to this PR that it's probably harder for newcomers to discover files in that directory.

PR #9948, opened 2024-02-13
[Bug #20234] Fix segv when parsing begin statement in method definition

eileencodes:
In a method definition, the `begin` may not have an `nd_body`. When that happens we get a null expr back from `last_expr_node` which causes a segv for the following examples:

```ruby
def (begin;end).foo; end
def (begin;else;end).foo; end
def (begin;ensure;else;end).foo; end
```

In addition, I've added tests for other cases that weren't causing a segv but appeared untested.`

Fixes https://bugs.ruby-lang.org/issues/20234

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -1395,7 +1395,7 @@ last_expr_node(NODE *expr)
         if (nd_type_p(expr, NODE_BLOCK)) {
             expr = RNODE_BLOCK(RNODE_BLOCK(expr)->nd_end)->nd_head;
         }
-        else if (nd_type_p(expr, NODE_BEGIN)) {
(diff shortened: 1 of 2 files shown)


eileencodes:
Oh lol. Ok so I had an `eval` on a `foo` method and that broke error highlight because it was checking for error highlighting on foo. The error I was seeing could be reproduced with the following:

```ruby
eval("def (begin;end).foo; end")
v = nil 
(v).foo += 42
```

I've updated the method name to be `bug_20234` instead so it won't conflict with anything else. Not sure if there's a better recommended way of testing these. Happy to update the tests if we want to do this differently.

nobu (approved):

PR #9953, opened 2024-02-13
Update PKCS7.write_smime to guard against null-dereference

pkuzco:


jeremyevans:
This would need to be filed against upstream (ruby/openssl repository), except that there it an existing pull request for it (https://github.com/ruby/openssl/pull/690, specific commit: https://github.com/ruby/openssl/commit/5325c06018e298b781edbae00449024f8eb2357a).

PR #9955, opened 2024-02-13
[DOC] Doc compliance

BurdetteLamar:
Monofont for each reference to the local class.


string.c
```diff
@@ -11511,17 +11511,17 @@ rb_str_unicode_normalized_p(int argc, VALUE *argv, VALUE str)
 /**********************************************************************
  * Document-class: Symbol
  *
- * \Symbol objects represent named identifiers inside the Ruby interpreter.
+ * A +Symbol+ objects represents a named identifier inside the Ruby interpreter.
```
peterzhu2118, line 11514:
```suggestion
 * A +Symbol+ object represents a named identifier inside the Ruby interpreter.
```

peterzhu2118 (approved):

PR #9956, opened 2024-02-13
YJIT: Simplify Kernel#send guards and admit more cases

XrXr:
Previously, our compile time check rejected dynamic symbols (e.g. what
String#to_sym could return) even though we could handle them just fine.
The runtime guards for the type of method name was also overly
restrictive and didn't accept dynamic symbols.

Fold the type check into the rb_get_symbol_id() and take advantage of
the guard already checking for 0. This also avoids generating the same
call twice in case the same method name is presented as different
types.


maximecb (approved):
Nice 👌

PR #9960, opened 2024-02-14
[Feature #20187] Skip examples for bundled gems at Ruby 3.4

hsbt:
I added version guard to `ruby/spec` for Ruby 3.4 again.

https://github.com/ruby/spec/pull/1137 reverts all of commits and enable to run them again.

In background, maintaining bundled gems is out of scope ruby repository. We shouldn't keep their test code because they are not our responsibility. We only test them with `test-bundled-gems` for ruby Interpreter keep their support.

https://github.com/ruby/ruby/pull/6130 makes to load bundled gems on `test-spec`. But it's accidentally changes for `test-spec`.

Also see https://github.com/ruby/ruby/commit/5bf75c20a2098e626d0dd65708a35c46039c5310

### DIFF
--- a/spec/ruby/core/integer/coerce_spec.rb
+++ b/spec/ruby/core/integer/coerce_spec.rb
@@ -1,104 +1,107 @@
 require_relative '../../spec_helper'
 
-require 'bigdecimal'
+ruby_version_is ""..."3.4" do
 
-describe "Integer#coerce" do
-  context "fixnum" do
-    describe "when given a Fixnum" do
-      it "returns an array containing two Fixnums" do
-        1.coerce(2).should == [2, 1]
-        1.coerce(2).map { |i| i.class }.should == [Integer, Integer]
+  require 'bigdecimal'
+
+  describe "Integer#coerce" do
+    context "fixnum" do
+      describe "when given a Fixnum" do
+        it "returns an array containing two Fixnums" do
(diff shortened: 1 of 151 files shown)


spec/ruby/core/integer/coerce_spec.rb
```diff
@@ -1,104 +1,107 @@
...
-
-describe "Integer#coerce" do
-  context "fixnum" do
-    describe "when given a Fixnum" do
-      it "returns an array containing two Fixnums" do
-        1.coerce(2).should == [2, 1]
-        1.coerce(2).map { |i| i.class }.should == [Integer, Integer]
-      end
-    end
+ruby_version_is ""..."3.4" do
```
eregon, line 3: This disables all Integer#coerce specs, it's not OK.
Same for Rational#coerce specs below.

hsbt:
I discussed this with @nurse. We will separate `ruby/spec` to two groups.

1. `core` and `library` for default gems
2. `library` for bundled gems

eregon:
That sounds good.

eregon:
I would still like to understand what is a concrete problem with `extract-gems` as a dependency of `test-spec` (as mentioned in https://bugs.ruby-lang.org/issues/20187#note-7).
It seems simplest as on current master.

PR #9964, opened 2024-02-14
[Feature #20265] Deprecate and remove rb_newobj and rb_newobj_of

peterzhu2118:


nobu:
After applying this PR, still there are some comments and `#undef`s.

PR #9966, opened 2024-02-14
Consider rb_str_getbyte as leaf sometimes

tenderlove:
If YJIT knows the parameter to rb_str_getbyte is a fixnum, then I think we can consider the function to be a leaf

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5399,10 +5399,8 @@ fn jit_rb_str_getbyte(
     let index = asm.stack_opnd(0);
     let recv = asm.stack_opnd(1);
 
-    let arg0_type = asm.ctx.get_opnd_type(index.into());
-
     // rb_str_getbyte should be leaf if the index is a fixnum
-    if arg0_type != Type::Fixnum {
+    if asm.ctx.get_opnd_type(index.into()) != Type::Fixnum {
         // Raises when non-integers are passed in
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -5395,11 +5395,18 @@ fn jit_rb_str_getbyte(
...
     let index = asm.stack_opnd(0);
     let recv = asm.stack_opnd(1);
+
+    let arg0_type = asm.ctx.get_opnd_type(index.into());
+
+    // rb_str_getbyte should be leaf if the index is a fixnum
+    if arg0_type != Type::Fixnum {
+        // Raises when non-integers are passed in
+        jit_prepare_non_leaf_call(jit, asm);
+    }
```
k0kubun, line 5408: The name "arg0" seems kind of cryptic. I think this is enough?

```suggestion
    // rb_str_getbyte should be leaf if the index is a fixnum
    if asm.ctx.get_opnd_type(index.into()) != Type::Fixnum {
        // Raises when non-integers are passed in
        jit_prepare_non_leaf_call(jit, asm);
    }
```

k0kubun (approved):

yjit/src/codegen.rs
tenderlove: ya, seems fine to me, I was just copy / pasting from somewhere else 😅

PR #9967, opened 2024-02-14
YJIT: Extract keyword handling to make gen_send_iseq() shorter

XrXr:
This should make it easier to grok the order of operations.

as requested in https://github.com/ruby/ruby/pull/9891#discussion_r1484924829


maximecb (approved):

k0kubun (approved):
Thanks!

PR #9970, opened 2024-02-14
[DOC] Fix links in doc

BurdetteLamar:


dir.c
```diff
@@ -3689,7 +3689,7 @@ Init_Dir(void)
      * {File::FNM_SYSCASE}[rdoc-ref:File::Constants@File-3A-3AFNM_SYSCASE] */
     rb_file_const("FNM_SYSCASE", INT2FIX(FNM_SYSCASE));
     /* Document-const: FNM_SHORTNAME
-     * {File::FNM_SHORTNAME}[rdoc-ref:File::Constants@File-3A-3AFNM_SHORTNAME] */
+     * {File::FNM_SHORTNAME}[rdoc-ref:File::Constants@File-3A-3AFNM_SHORTNAME+-28Windows+Only-29] */
```
nobu, line 3692: I'd rather propose moving "(Windows Only)" from the heading to its content.

```
     * ==== File::FNM_SHORTNAME
     *
     * Flag File::FNM_SHORTNAME Allows patterns to match short names if they exist.
     * (Windows Only).
```

BurdetteLamar: Heading and link fixed, along with one other similar header and its link.

io.c
```diff
@@ -10473,9 +10473,6 @@ static VALUE argf_readlines(int, VALUE *, VALUE);
  *    $cat t.txt | ruby -e "p readlines 12"
  *    ["First line\n", "Second line\n", "\n", "Fourth line\n", "Fifth line\n"]
  *
- *  With arguments +sep+ and +limit+ given, combines the two behaviors;
- *  see {Line Separator and Line Limit}[rdoc-ref:IO@Line+Separator+and+Line+Limit].
```
peterzhu2118, line 10477: Why was this removed?

peterzhu2118 (commented):
Why is there an empty file `doc/field_processing.md` created?

io.c
BurdetteLamar: Multiple similar passages removed in previous PRs.  In my view, not needed;  plus, the cited section no longer exists (just the two separate ones).

peterzhu2118: Why do some methods have a section about `sep` and `limit` and some methods don't?

BurdetteLamar: Probably an error.  Let's get the link fixes merged, then deal with that?

peterzhu2118: I think this one should be fixed before it can be merged.

BurdetteLamar: Not sure what the fix should be.

peterzhu2118: Well after this change, it seems like arguments `sep` and `limit` are mutually exclusive (especially with the wording "With only [...] argument [...] given"). The documentation needs to be obvious that the two can be combined and should be clear what the behaviour will be.

BurdetteLamar: Done (I think).

io.c
```diff
@@ -10473,8 +10466,9 @@ static VALUE argf_readlines(int, VALUE *, VALUE);
...
+ *  With arguments +sep+ and +limit+ given,
+ *  combines the two behaviors
+ *  (see {Line Separator and Line Limit}[rdoc-ref:IO@Line+Separator+and+Line+Limit]).
```
peterzhu2118, line 10471: I don't think this link exists.

BurdetteLamar: I have to figure this out.  I have that section in `io.c` on my branch `links_repair`;  the file is not showing as modified or staged, and committing and/or pushing does nothing.

BurdetteLamar: I see the section in https://github.com/ruby/ruby/blob/2f29d42f1c7eb2a9294b9c818745d78b0b18da13/io.c, line 15221.  On my machine, RDoc builds valid links to this heading.  Link does not exist or does not work for you?

peterzhu2118: Your branch is out-of-date. You removed that section in 80490acfb601557497d5c10841fc5493e691e6da.

peterzhu2118: Ah I see you added it back in 2f29d42f1c7eb2a9294b9c818745d78b0b18da13.

peterzhu2118 (approved):

PR #9971, opened 2024-02-14
YJIT: Replace perf_fn! with a simpler macro

k0kubun:
Follow-up on https://github.com/ruby/ruby/pull/9957

The `perf_fn!` macro was useful to make sure every such call is profiled. However, VSCode (rust-analyzer) doesn't support showing the function name when you're a inside a macro (`perf_fn!`), and that makes it harder to read code for me.

This PR adds `perf_call!` macro that needs to be used by the caller. It has the risk of leaving un-profiled calls, but we don't add calls of these functions that often, so I think it's acceptable. This macro is useful when you want to add a prefix too.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -319,17 +319,20 @@ macro_rules! jit_perf_symbol_pop {
     };
 }
 
-/// Macro to push and pop perf symbols around a function definition.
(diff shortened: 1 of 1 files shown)


XrXr (approved):
Nice. I was a bit worried about breaking IDE support too.

PR #9973, opened 2024-02-15
De-dup callinfo objects

jhawthorn:
https://bugs.ruby-lang.org/issues/19907

Previously every call to vm_ci_new (when the CI was not packable) would
result in a different callinfo being returned this meant that every
kwarg callsite had its own CI.

When calling, different CIs result in different CCs. These CIs and CCs
both end up persisted on the T_CLASS inside cc_tbl. So in an eval loop
this resulted in a memory leak of both types of object. This also likely
resulted in extra memory used, and extra time searching, in non-eval
cases.

For simplicity in this commit I always allocate a CI object inside
rb_vm_ci_lookup, but ideally we would lazily allocate it only when
needed. I hope to do that as a follow up in the future.

```ruby
10.times do
  1000.times do
    eval("Hash.new(foo: 456)")
  end
  GC.start
  puts(GC.stat(:heap_live_slots))
end
```

**Before**

```
$ ruby --disable-gems test_ci_leak.rb
9029
11037
13037
15037
17037
19037
21037
23037
25037
27037
```

**After**

```
$ ./miniruby --disable-gems test_ci_leak.rb
8446
8460
8466
8472
8478
8484
8490
8496
8502
8508
```

### DIFF
--- a/vm_callinfo.h
+++ b/vm_callinfo.h
@@ -199,12 +199,13 @@ vm_ci_dump(const struct rb_callinfo *ci)
       (((VALUE)(argc)) << CI_EMBED_ARGC_SHFT) |  \
       RUBY_FIXNUM_FLAG))
 
+// vm_method.c
+const struct rb_callinfo *rb_vm_ci_lookup(ID mid, unsigned int flag, unsigned int argc, const struct rb_callinfo_kwarg *kwarg);
+void rb_vm_ci_free(const struct rb_callinfo *);
+
 static inline const struct rb_callinfo *
 vm_ci_new_(ID mid, unsigned int flag, unsigned int argc, const struct rb_callinfo_kwarg *kwarg, const char *file, int line)
 {
-    if (kwarg) {
-        ((struct rb_callinfo_kwarg *)kwarg)->references++;
-    }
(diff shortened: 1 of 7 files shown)


vm_callinfo.h
```diff
@@ -213,13 +214,10 @@ vm_ci_new_(ID mid, unsigned int flag, unsigned int argc, const struct rb_callinf
...
 
-    // TODO: dedup
-    const struct rb_callinfo *ci = (const struct rb_callinfo *)
-      rb_imemo_new(imemo_callinfo,
-                   (VALUE)mid,
-                   (VALUE)flag,
-                   (VALUE)argc,
-                   (VALUE)kwarg);
+    if (kwarg) {
+        ((struct rb_callinfo_kwarg *)kwarg)->references++;
```
jhawthorn, line 218: Reminder to me: need to make sure this doesn't leak

jhawthorn:
Booting a mastodon production server, this reduced `ObjectSpace.count_imemo_objects[:imemo_callinfo]` from 15709 to 3798 🎉. Which can largely be explained by files [like this](https://github.com/ruby-rdf/json-ld-preloaded/blob/develop/lib/json/ld/preloaded/prefix.rb) which is why I started looking into this.

PR #9975, opened 2024-02-15
Show the invalid source encoding in messages

nobu:
Related to [Feature #20210], just display the line where the invalid encoding pragma is present without changing the exception.


nobu:
Talked about this idea in the dev-meeting 2024-02-14, and approved by matz.

PR #9986, opened 2024-02-16
YJIT: Optimize attr_writer

k0kubun:
This PR optimizes `attr_writer` by reusing `setinstancevariable`'s shape-based implementation `gen_set_ivar`.

This seems to speed up railsbench by 2%.

```
before: ruby 3.4.0dev (2024-02-21T23:09:17Z master 35f68e7dff) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-02-22T01:26:29Z yjit-attrset f27198fac3) +YJIT [x86_64-linux]

railsbench  2439.5       1.0         2382.9      0.8         1.01           1.02
```


yjit/src/codegen.rs
```diff
@@ -2759,33 +2737,44 @@ fn gen_setinstancevariable(
...
+                rb_vm_setinstancevariable as *const u8,
+                vec![
+                    Opnd::const_ptr(jit.iseq as *const u8),
+                    Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SELF),
+                    ivar_name.into(),
+                    val_opnd,
+                    Opnd::const_ptr(ic.unwrap() as *const u8),
+                ],
+            );
+        }
```
maximecb, line 2770: It seems like a good idea to reuse code and it's nice to see a 2% speedup! 👍 

However, this PR introduces multiple if statements to use different operands and change the behavior, which seems a bit weird.

Just to make sure I understand, attr_writers do not have ICs? That's why you have to call a different function here?

yjit/src/codegen.rs
```diff
@@ -2880,7 +2873,14 @@ fn gen_setinstancevariable(
...
     }
-    asm.stack_pop(1); // Keep it on stack during ccall for GC
+    let val = asm.stack_pop(1); // Keep it on stack during ccall for GC
+
+    if recv_opnd != SelfOpnd {
+        asm.stack_pop(1); // Pop receiver
+
+        let out_opnd = asm.stack_push(Type::Unknown); // Push a return value
+        asm.mov(out_opnd, val);
+    }
```
maximecb, line 2883: This confused me for a bit, as in "wait, why are we not producing an output value if the operand is self? why are we producing a value if the operand is not self?"

Probably deserves a comment.

yjit/src/codegen.rs
k0kubun: You're right. attr_writers don't have ICs we need here, so attr_writers can't use `rb_vm_setinstancevariable`. An attr_writer is a method call, so it has an `cd->cc` (method call cache), but it doesn't have an instance-variable cache `ic` (`IVC`), which is used by setinstancevariable.

Similarly, setinstancevariable can't use `rb_vm_set_ivar_id`. attr_writer fallbacks to a dynamic dispatch when the receiver is not `T_OBJECT`, but setinstancevariable wants to use this general case for non-`T_OBJECT` cases.

We could use `rb_ivar_set` for both cases, which supports non-`T_OBJECT` receivers as well, but I saw a slowdown when I tried that, so I kept this logic. But let me benchmark it again 🤔

yjit/src/codegen.rs
k0kubun: wrote comments to clarify that https://github.com/ruby/ruby/pull/9986/commits/2b73548263dd9acd903f52f68cfc6ba566d0a5bf

yjit/src/codegen.rs
```diff
@@ -2759,33 +2737,44 @@ fn gen_setinstancevariable(
...
     } else {
         // Get the receiver
-        let mut recv = asm.load(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SELF));
-
-        let recv_opnd = SelfOpnd;
+        let mut recv = asm.load(if let StackOpnd(index) = recv_opnd {
+            asm.stack_opnd(index as i32)
+        } else {
+            Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SELF)
+        });
```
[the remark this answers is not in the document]
k0kubun: The blocker here is that `Opnd::Stack` operands may become invalid once you call `gen_save_sp`. 

We could automatically correct all `Opnd::Stack` operands on `push_insn`, but we sometimes need to interact with `Opnd::Stack` before it reaches `push_insn` (e.g. `asm.ccall`), so there's a risk of having inconsistent `Opnd::Stack` values for a single instruction. So I only introduced an assertion in https://github.com/ruby/ruby/pull/10061.

yjit/src/codegen.rs
k0kubun: > We could use rb_ivar_set for both cases, which supports non-T_OBJECT receivers as well, but I saw a slowdown when I tried that, so I kept this logic. But let me benchmark it again 🤔

It does seem to give a slowdown (these benchmarks consistently give a 0.99x change). It seems to hurt the performance if you handle both cases using `rb_ivar_set`. Another thing we could try is to replace `rb_vm_setinstancevariable` with `rb_ivar_set`, but it doesn't remove this `if` branch, so it doesn't simplify code.

```
before: ruby 3.4.0dev (2024-02-22T01:26:29Z yjit-attrset f27198fac3) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-02-22T17:12:46Z yjit-attrset 2b73548263) +YJIT [x86_64-linux]

hexapdf     1862.5       3.0         1874.3      1.9         1.02           0.99
railsbench  2423.5       0.9         2436.3      0.8         0.98           0.99
```

k0kubun: > Another thing we could try is to replace rb_vm_setinstancevariable with rb_ivar_set, but it still leaves this if branch, so it doesn't simplify code.

I also tried this, but it also gives a slowdown (0.99x) on railsbench. So it seems more efficient to keep using `rb_vm_setinstancevariable` for setinstancevariable and `rb_vm_set_ivar_id` for attr_writer.

maximecb: Oh yeah, I wasn't suggesting that we use the less efficient function for both. The code is a bit harder to follow though with the special casing, but I understand why it is there.

PR #9994, opened 2024-02-16
Test setup-ruby 3.1 on macos-arm-oss

k0kubun:
It's not supposed to have cached Ruby 3.1, so I want to see what happens there.


k0kubun:
https://github.com/ruby/ruby/actions/runs/7934684203/job/21666179685?pr=9994

Screenshot 2024-02-16 at 10 30 12

It seems fast enough.

PR #9998, opened 2024-02-16
Switch mk_builtin_loader.rb to use prism to parse

kddnewton:
https://bugs.ruby-lang.org/issues/20238

Short overview:

* Create a new `tool/collect_builtins.c` script that is compiled with prism source files into `tool/collect_builtins`
* Call out to `tool/collect_builtins` in `tool/mk_builtin_loader.rb` to collect builtins, inlines, and requires
* Limit syntax that is allowed to be used to Ruby 3.3.0 released syntax

Longer explanation:

We would like to be able to use the latest released syntax in Ruby core files. Fortunately, we have the prism source files in the build system already and have access to them, and they don't depend on anything in Ruby. `mk_builtin_loader.rb` before this commit used `Ripper` to parse through the source files and find (1) `require`/`require_relative` calls for sublibraries (2) calls to `Primitive`/`__builtin`/method calls to `__builtin_*` methods.

This commit instead builds a small script that can be compiled using libprism that has a very simple format. It accepts a filepath on the command line and prints out one line per require, builtin, or inline. `mk_builtin_loader.rb` will invoke the script, read the output, and use that instead of ripper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -238,6 +238,7 @@ lcov*.info
 # /tool/
 /tool/config.guess
 /tool/config.sub
+/tool/collect_builtins
 
 # /win32/
 /win32/*.ico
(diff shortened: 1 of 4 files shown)


eregon:
Looks like this is replaced by https://github.com/ruby/ruby/pull/10005

PR #10001, opened 2024-02-16
YJIT: Remove unused counters

XrXr:


maximecb (approved):

yjit/src/stats.rs
```diff
@@ -575,8 +553,6 @@ make_counters! {
     invalidate_constant_state_bump,
     invalidate_constant_ic_fill,
 
-    constant_state_bumps,
```
k0kubun, line 578: Let's remove that from `yjit.rb` too.

PR #10002, opened 2024-02-17
YJIT: Verify the assumption of leaf C calls

k0kubun:


XrXr (approved):

yjit/src/codegen.rs
```diff
@@ -5507,10 +5524,11 @@ fn jit_rb_str_concat(
...
     guard_object_is_string(asm, asm.stack_opnd(0), StackOpnd(0), Counter::guard_send_not_string);
 
-    // Guard buffers from GC since rb_str_buf_append may allocate. During the VM lock on GC,
-    // other Ractors may trigger global invalidation, so we need ctx.clear_local_types().
-    // PC is used on errors like Encoding::CompatibilityError raised by rb_str_buf_append.
+    // Guard buffers from GC since rb_str_buf_append may allocate.
     jit_prepare_non_leaf_call(jit, asm);
+    // rb_str_buf_append may raise Encoding::CompatibilityError, but we accept compromised
+    // backtraces on this method since the interpreter does the same thing on opt_ltlt.
+    asm.leaf_ccall = false;
```
maximecb, line 5531: A bit confused as to why `asm.leaf_call` is being manually set in `codegen.rs` here, and for some very specific C functions such as `jit_rb_str_concat`. I think that maybe the "API" for verifying leafness could be polished up a bit?

That being said I'm in favor of the general idea of this PR. It seems like a good idea to verify leafness 👍 It's a good idea.

yjit/src/codegen.rs
```diff
@@ -6068,7 +6086,9 @@ fn gen_send_cfunc(
     if kw_arg.is_null() && flags & VM_CALL_OPT_SEND == 0 && flags & VM_CALL_ARGS_SPLAT == 0 && (cfunc_argc == -1 || argc == cfunc_argc) {
         let expected_stack_after = asm.ctx.get_stack_size() as i32 - argc;
         if let Some(known_cfunc_codegen) = lookup_cfunc_codegen(unsafe { (*cme).def }) {
-            if perf_call!("gen_send_cfunc: ", known_cfunc_codegen(jit, asm, ocb, ci, cme, block, argc, recv_known_class)) {
+            if asm.with_leaf_ccall(|asm|
+                perf_call!("gen_send_cfunc: ", known_cfunc_codegen(jit, asm, ocb, ci, cme, block, argc, recv_known_class))
+            ) {
```
maximecb, line 6091: Here, you are assuming that by default, all known cfunc codegen will be leaf? I guess this tracks current usage?

yjit/src/cruby.rs
```diff
@@ -260,6 +261,18 @@ pub fn iseq_opcode_at_idx(iseq: IseqPtr, insn_idx: u32) -> u32 {
     unsafe { rb_iseq_opcode_at_pc(iseq, pc) as u32 }
 }
 
+/// Return a poison value to be set above the stack top to verify leafness.
+#[cfg(not(test))]
+pub fn vm_stack_canary() -> u64 {
+    unsafe { rb_vm_stack_canary() }.as_u64()
+}
+
+/// Avoid linking the C function on `cargo test``
```
maximecb, line 270:
```suggestion
/// Avoid linking the C function in `cargo test`
```

yjit/src/codegen.rs
k0kubun: Correct. The interpreter skips pushing a frame for `opt_*` insns (e.g. `String#<<` implemented by `jit_rb_str_concat`), and I want to allow YJIT to do the same thing in that case, but they should be leaf by default.

yjit/src/codegen.rs
k0kubun: relevant context: https://github.com/ruby/ruby/pull/10002#discussion_r1496263381

For methods specialized by the interpreter as `opt_*` insns, I want to allow YJIT to break this rule because the interpreter does the same thing (it's even more "compatible" with the interpreter).

We could clarify that if we move the implementation to `opt_ltlt`, or maybe propagate a boolean flag from `opt_ltlt` to allow this exception.

maximecb: Ok but why are you manually setting `asm.leaf_ccall = false;` in this method? 😅

Also, directly setting this boolean in `codegen.rs` is kind of "breaking" the abstraction of `with_leaf_ccall` ? It feels like this variable should not be public.

yjit/src/codegen.rs
maximecb: Could maybe add a comment to explicitly say "We assume that specialized cfunc codegen will only be done for methods that are leaf".

yjit/src/codegen.rs
k0kubun: > Ok but why are you manually setting asm.leaf_ccall = false; in this method? 😅

We want to make it `true` inside specialized C codegen "by default", and make it `false` for these `opt_*` exceptions. Again, we can obviate this if we move `String#<<` specialization to `gen_opt_ltlt`, and I can make a follow-up PR. If you want me to remove this code in this PR, I would do that change (remove `jit_rb_str_concat` and implement the same thing in `gen_opt_ltlt`) in this PR.

> Also, directly setting this boolean in codegen.rs is kind of "breaking" the abstraction of with_leaf_ccall ? It feels like this variable should not be public.

For the assertion (`ccall() wasn't used after leaf_ccall was set in {}`), we'd need `asm.get_leaf_ccall()`, and we'd also need to be able to set both `false` and `true` to `asm.leaf_ccall`. In that case, making that private and providing those public methods doesn't seem to hide anything to me, which is why I didn't do that. If you prefer it to be that way, I can make that change too.

k0kubun: I guess, instead of `asm.set_leaf_ccall()`, we could add something like `asm.allow_non_leaf_call_as_an_exception()` to indicate it's an exception. We could do that in this PR, and then remove it once it's moved to `gen_opt_ltlt`?

maximecb: I understand what you mean and I think your idea of moving the specialization to ltlt (in this PR or in another) makes sense 👍

yjit/src/codegen.rs
k0kubun: added a comment here https://github.com/ruby/ruby/pull/10002/commits/60f635961dbd425d45b01a5647c515fd4edca84e

yjit/src/codegen.rs
k0kubun: I just made it private in this PR https://github.com/ruby/ruby/pull/10002/commits/d887bef7dd30a3f367c08b85c8c0873395445a42. I'll make another PR to move it to `opt_ltlt`, and I'll remove `allow_non_leaf_ccall()` in it.

k0kubun: https://github.com/ruby/ruby/pull/10033

PR #10004, opened 2024-02-17
[]Introduce rb_parser_ary_t

hasumikin:
- Introduce `rb_parser_ary_t` struct to partly eliminate RArray from parse.y
- In this PR, `parser_params->tokens` and `parser_params->ast->node_buffer->tokens` becomes `rb_parser_ary_t *`
- Instead, `ast_node_all_tokens()` internally creates a Ruby Array object from the rb_parser_ary_t


node.c
```diff
@@ -483,7 +511,7 @@ rb_ast_delete_mark_object(rb_ast_t *ast, VALUE obj)
 VALUE
 rb_ast_tokens(rb_ast_t *ast)
 {
-    return ast->node_buffer->tokens;
+    return (VALUE)ast->node_buffer->tokens;
```
nobu, line 514: Is `rb_parser_ary_t` compatible with `VALUE`?
This function is used in the implementation of `RubyVM::AbstractSyntaxTree::Node#all_tokens`, and the return value will be seen in Ruby level.

parse.y
```diff
@@ -7564,35 +7631,20 @@ parser_has_token(struct parser_params *p)
...
+    token->type_id = parser_token2id(p, t);
+    token->str = str;
+    token->loc.beg_pos = p->yylloc->beg_pos;
+    token->loc.end_pos = p->yylloc->end_pos;
+    rb_parser_ary_push(p, p->tokens, (VALUE)token);
     p->token_id++;
 
     if (p->debug) {
-        rb_parser_printf(p, "Append tokens (line: %d) %"PRIsVALUE"\n", line, ary);
+        rb_parser_printf(p, "Append tokens (line: %d) %"PRIsVALUE"\n", line, (VALUE)token);
```
nobu, line 7647: `PRsVALUE` calls `to_s` method on the corresponding argument.
`rb_parser_ast_token_t` does not seem to compatible with `VALUE`.

nobu (changes requested):

parse.y
hasumikin: Thank you for the review!
I implemented `rb_parser_str_escape()` which does not depend on VALUE
https://github.com/ruby/ruby/pull/10004/commits/15e42bcc75f65a7ccd7dc21f569c8f567c31ee35

node.c
hasumikin: I reimplemented `rb_ast_tokens()` that constructs RArray from rb_parser_ary_t https://github.com/ruby/ruby/commit/8a986a2ba1e7956ce804c77032e8d4da1e6357db

parse.y
```diff
@@ -2613,6 +2614,64 @@ rb_str_to_parser_string(rb_parser_t *p, VALUE str)
...
+    long i;
+    if (ary->capa < len) {
+        ary->capa = len;
+        if (ary->data == NULL) {
+            ary->data = xcalloc(len, sizeof(void *));
+        } else {
+            ary->data = xrealloc(ary->data, sizeof(void *) * len);
+        }
+        for (i = ary->len; i < len; i++) {
+            ary->data[i] = Qnil;
```
yui-knk, line 2631: Not `Qnil` but `0`?

parse.y
```diff
@@ -2613,6 +2614,64 @@ rb_str_to_parser_string(rb_parser_t *p, VALUE str)
...
+    ary->len = 0;
+    ary->capa = len;
+    if (0 < len) {
+        ary->data = xcalloc(len, sizeof(void *));
+    }
+    else {
+        ary->data = NULL;
+    }
+    for (i = 0; i < len; i++) {
+        ary->data[i] = Qnil;
```
yui-knk, line 2653: Not `Qnil` but `0` (or `NULL`)?

parse.y
```diff
@@ -2613,6 +2614,64 @@ rb_str_to_parser_string(rb_parser_t *p, VALUE str)
...
+rb_parser_ary_new(rb_parser_t *p)
+{
+    return rb_parser_ary_new_capa(p, 0);
+}
+
+static rb_parser_ary_t *
+rb_parser_ary_push(rb_parser_t *p, rb_parser_ary_t *ary, VALUE val)
+{
+    if (ary->len == ary->capa) {
+        rb_parser_ary_extend(p, ary, ary->len + 1);
```
yui-knk, line 2669: Expanding the array with `ary->len + 1` is no efficient especially for ast token array because it's expected to be expanded again and again. How about to double the capacity?

node.c
```diff
@@ -177,6 +185,29 @@ parser_string_free(rb_ast_t *ast, rb_parser_string_t *str)
     xfree(str);
 }
 
+static void
+parser_ast_token_free(rb_ast_t *ast, rb_parser_ast_token_t *token)
+{
+    if (!token) return;
+    parser_string_free(ast, (rb_parser_string_t *)token->str);
```
yui-knk, line 192: Is this cast for `rb_parser_string_t *` needed?

node.c
```diff
@@ -483,7 +523,32 @@ rb_ast_delete_mark_object(rb_ast_t *ast, VALUE obj)
 VALUE
 rb_ast_tokens(rb_ast_t *ast)
 {
-    return ast->node_buffer->tokens;
+    long i;
```
yui-knk, line 526: How about to move this ruby object construction logic to `ast.c`? By the change, node.c no need to handle VALUE.

parse.y
```diff
@@ -8140,16 +8263,16 @@ yycompile0(VALUE arg)
...
         NODE *body = parser_append_options(p, RNODE_SCOPE(tree)->nd_body);
         prelude = block_append(p, p->eval_tree_begin, body);
         RNODE_SCOPE(tree)->nd_body = prelude;
         p->ast->body.frozen_string_literal = p->frozen_string_literal;
         p->ast->body.coverage_enabled = cov;
         if (p->keep_tokens) {
-            rb_obj_freeze(tokens);
-            rb_ast_set_tokens(p->ast, tokens);
+            p->ast->node_buffer->tokens = tokens;
+            p->tokens = NULL;
```
yui-knk, line 8275: Could you add free for `rb_ruby_parser_free` just in case, if an exception happens in the middle of parsing, this line can be skipped.

parse.y
```diff
@@ -7564,35 +7623,98 @@ parser_has_token(struct parser_params *p)
...
+    token->str = str;
+    token->loc.beg_pos = p->yylloc->beg_pos;
+    token->loc.end_pos = p->yylloc->end_pos;
+    rb_parser_ary_push(p, p->tokens, (VALUE)token);
     p->token_id++;
 
     if (p->debug) {
-        rb_parser_printf(p, "Append tokens (line: %d) %"PRIsVALUE"\n", line, ary);
+        rb_parser_printf(p, "Append tokens (line: %d) [%d, :%s, \"%s\", [%d, %d, %d, %d]]\n",
+                            line, token->id, token->type_name, rb_parser_str_escape(p, token->str)->ptr,
```
yui-knk, line 7715: I guess `rb_parser_string_t` allocated in `rb_parser_str_escape` is not freed.

node.c
hasumikin: Yeah, I thought that too but,
If we move the logic from node.c to ast.c, do you also think we should redesign `ast->node_buffer->tokens` into `ast->tokens` because we can say the `tokens` member no longer belongs to a Node?
(Of course, we cannot change the Ruby level design: `AbstractSyntaxTree::Node` though)

parse.y
hasumikin: Fixed in https://github.com/hasumikin/ruby/commit/9c64383a8a112f2b4c7f9c74bead2e416529160d

parse.y
hasumikin: Thanks. Fixed in https://github.com/hasumikin/ruby/commit/edd7c6d563523d6fe99efaeda800ad69d60948c7

parse.y
hasumikin: Is this OK? https://github.com/hasumikin/ruby/commit/c2d3fc8ec4b536f122ddc5a3748145481b2d3385 (I amend a function name `parser_ary_free()` to `parser_token_free()` for clarity)

parse.y
hasumikin: Fixed: https://github.com/ruby/ruby/pull/10004/commits/d1fd5bae7fa1768ab5f179bd3c679996636c2b3e
BTW, what does the comment `// TODO` in `rb_ruby_parser_keep_tokens()` mean?

node.c
yui-knk: In a long term, it's correct however It's impossible right now. `rb_ast_t` is imemo (internal ruby object) now therefore it's size should be 5 words. It's already uses 5 words like below. We can refactor it once `rb_ast_t` is not managed as `imemo`.

```c
typedef struct rb_ast_body_struct {
    const NODE *root;
    VALUE script_lines;
    // script_lines is either:
    // - a Fixnum that represents the line count of the original source, or
    // - an Array that contains the lines of the original source
    signed int frozen_string_literal:2; /* -1: not specified, 0: false, 1: true */
    signed int coverage_enabled:2; /* -1: not specified, 0: false, 1: true */
} rb_ast_body_t;
typedef struct rb_ast_struct {
    VALUE flags;
    node_buffer_t *node_buffer;
    rb_ast_body_t body;
} rb_ast_t;
```

parse.y
yui-knk: `parser_token_free` is preferred, BTW the existence of `ast` and `ast->node_buffer` is confirmed by `rb_ast_free`. Therefore it's more simple to call `parser_tokens_free` directly in `rb_node_buffer_free`.

node.c
hasumikin: Ah, I see

hasumikin: I did
https://github.com/hasumikin/ruby/commit/5ef62bbdb1c1a7dd01e5730246b4f2f2ff8cad94

parse.y
hasumikin: Remove the line https://github.com/hasumikin/ruby/commit/64d18710129f1652a1ee5ea9d13bb168a4c5922d

parse.y
hasumikin: You are right. Fixed https://github.com/hasumikin/ruby/commit/e9e3ac6436adbbdd50ed39315fc1406508f8f3bd

parse.y
hasumikin: @nobu Could you please review them again?

parse.y
```diff
@@ -2587,7 +2587,81 @@ rb_parser_string_hash_cmp(rb_parser_string_t *str1, rb_parser_string_t *str2)
...
+rb_parser_ary_extend(rb_parser_t *p, rb_parser_ary_t *ary, long len)
+{
+    long i;
+    if (ary->capa < len) {
+        ary->capa = len;
+        if (ary->data == NULL) {
+            ary->data = xcalloc(len, sizeof(void *));
+        } else {
+            ary->data = xrealloc(ary->data, sizeof(void *) * len);
+        }
```
nobu, line 2601: `xrealloc` allocates a new memory when the pointer is NULL, as well as the standard `realloc`.

```suggestion
        ary->data = xrealloc(ary->data, sizeof(void *) * len);
```

parse.y
```diff
@@ -2587,7 +2587,81 @@ rb_parser_string_hash_cmp(rb_parser_string_t *str1, rb_parser_string_t *str2)
...
+    ary->capa = len;
+    if (0 < len) {
+        ary->data = xcalloc(len, sizeof(void *));
+    }
+    else {
+        ary->data = NULL;
+    }
+    for (i = 0; i < len; i++) {
+        ary->data[i] = 0;
+    }
```
nobu, line 2626: `xcalloc` zero-fills as well as the standard `calloc`.

rubyparser.h
```diff
@@ -54,6 +54,15 @@ typedef struct rb_parser_string {
     char *ptr;
 } rb_parser_string_t;
 
+/*
+ * Array-like object for parser
+ */
+typedef struct rb_parser_ary {
+    uintptr_t *data;
```
nobu, line 61: Why this `data` is a pointer to `uintptr_t` instead of  `rb_parser_ast_token_t`?
Also `val` of  `rb_parser_ary_push`.

parse.y
hasumikin: Thank you. Fixed

rubyparser.h
hasumikin: We plan to have rb_parser_ary store elements of arbitrary types. However, after talking to Kaneko-san, we decided to limit it to the `rb_parser_ast_token_t*` type for now.
https://github.com/ruby/ruby/pull/10004/commits/85ff09cb381da4be277823aba41de8596c0ecc1c
Thank you for the review!

nobu (approved):

yui-knk (approved):

PR #10005, opened 2024-02-17
Switch mk_builtin_loader.rb to use prism to parse

kddnewton:
Creates a new C file (tool/dump_ast.c) that is compiled using prism and dumps Ruby files to JSON. Then switch `mk_builtin_loader.c` to invoke that script to get the AST and walk it using Ruby.


kddnewton:
@nobu I am trying to get this to work, and it appears to work on every system except nmake. The issue appears to be that:

```
{$(srcdir)}.rb.rbinc:
	$(ECHO) making $@
	$(Q) $(BASERUBY) $(tooldir)/mk_builtin_loader.rb $(tooldir)/dump_ast$(EXEEXT) $<

$(BUILTIN_RB_INCS): $(tooldir)/mk_builtin_loader.rb $(tooldir)/dump_ast$(EXEEXT)
```

does not pick up on the fact that the `.rbinc` files need `$(tooldir)/dump_ast$(EXEEXT)`. Do you know how to fix this with nmake? I will keep looking for a solution, but I was hoping you might know how.

k0kubun:
@kateinoigakukun Could you help us fix the failure on the WebAssembly CI? In this PR, we're trying to let BASERUBY use a Prism binary compiled from source as `$(srcdir)/tool/dump_ast` so that we can use the latest Ruby syntax in builtin ISEQs.

So when you cross-compile Ruby into WebAssembly from source, `dump_ast` would be cross-compiled and cannot be executed during the build process. What do you think we should do to address this? Since we build Ruby for the host in the workflow first, I attempted a workaround https://github.com/ruby/ruby/pull/10005/commits/89b83e8332b6d5919a85a3697c1d5a37ef76f89a to skip re-building it during the cross-compile step, but it didn't work.

(Also, if you have an idea to fix the failure on Windows jobs, that'd be appreciated too 🥹)

kateinoigakukun:
Thank you for taking care of Wasm CI 😄 

I think the root issue here is that we don't have a good convention for running just-built host tools other than the Ruby executable via `--with-baseruby`.

Fortunately, we currently require building baseruby from the same source and pass it to `--with-baseruby` when cross-compiling ruby, so I think we have two options:

1. Add `--with-dump-ast=<path>` configure option and mandate it when cross-compiling. We already require building baseruby from source, so dump_ast for host should always exist somewhere. It's very simple but one drawback is we need to ask crossruby builders to update their build scripts to add with-dump-ast.

2. Provide a Ruby script that mirrors tool/dump_ast.c functionalities and run it with baseruby when cross-compiling. We can assume baseruby is built from the same source, so we should be able to use the latest syntax. We need to maintain two dump scripts in .c and .rb but we don't need to introduce a breaking change in cross build interface.

In the long term, it would be ideal to unify host/target build graph into a single task graph, then we can naturally address this kind of challenges, but not feasible at this moment anyway 🤷

k0kubun:
Thank you for your reply!

> 1. Add --with-dump-ast=<path> configure option and mandate it when cross-compiling.

I'll try this one. Please note that we need to generate source code only for source build or creating release tarballs, so this change shouldn't bother "crossruby builders" who build released Ruby versions.

> 2. Provide a Ruby script that mirrors tool/dump_ast.c functionalities and run it with baseruby when cross-compiling. We can assume baseruby is built from the same source, so we should be able to use the latest syntax.

I'll think about it when we give up on (1) 🙂

PR #10009, opened 2024-02-18
[DOC] rewrite inject doc

pragdave:
I was reading the RDoc for `Enumerable#inject` and I felt it needed some love.

It was just plain wrong in places. But it also (I think) missed the whole point of `inject`/`reduce`. The examples were almost all `[1,2,3].inject(:+)` style, which is only a trivial subset of the real-world power of `reduce`.

Feel free to ignore this, or to change anything you don't like. I feel better for having done it :)

### DIFF
--- a/enum.c
+++ b/enum.c
@@ -956,7 +956,7 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
  *
  *  <b>Second Shortcut: a Reducer function</b>
  *
- *  A _reducer function_ is a function that takes a partial result and the next value,
+ *  A <i>reducer function</i> is a function that takes a partial result and the next value,
  *  returning the next partial result. The block that is given to +inject+ is a reducer.
  *
  *  You can also write a reducer as a function and pass the name of that function
--- a/enum.c
+++ b/enum.c
@@ -886,7 +886,7 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
  *
  *  as being
  *
- *      fn(d, fn(c, fn(b, fn(i, a))))
+ *      fn(fn(fn(fn(i, a), b), c), d)
  *
  *  In a way the +inject+ function _injects_ the function
  *  between the elements of the enumerable.
--- a/enum.c
+++ b/enum.c
@@ -891,7 +891,7 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
  *  In a way the +inject+ function _injects_ the function
  *  between the elements of the enumerable.
  *
- *  `inject` is aliased as `reduce`. You use it when you want to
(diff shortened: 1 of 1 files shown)


enum.c
```diff
@@ -870,134 +870,151 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
...
  *
- *    # Integer.
- *    (1..4).inject(:+)                # => 10
- *    # Float.
- *    [1.0, 2, 3, 4].inject(:+)        # => 10.0
- *    # Character.
- *    ('a'..'d').inject(:+)            # => "abcd"
- *    # Complex.
- *    [Complex(1, 2), 3, 4].inject(:+) # => (8+2i)
+ *  `inject` is aliased as `reduce`. You use it when you want to
```
nobu, line 894:
```suggestion
 *  +inject+ is aliased as +reduce+. You use it when you want to
```

batate:
This description makes more sense to me than the old one. It's the way I teach reject and reduce.

dskecse:
Thank you for the changes, @pragdave! ❤️

enum.c
```diff
@@ -870,134 +870,151 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
...
- *  If argument +initial_operand+ is not given,
- *  the operands for +inject+ are simply the elements of +self+.
- *  Example calls and their operands:
+ *  as being
  *
- *  - <tt>(1..4).inject(:+)</tt>:: <tt>[1, 2, 3, 4]</tt>.
- *  - <tt>(1...4).inject(:+)</tt>:: <tt>[1, 2, 3]</tt>.
- *  - <tt>('a'..'d').inject(:+)</tt>:: <tt>['a', 'b', 'c', 'd']</tt>.
- *  - <tt>('a'...'d').inject(:+)</tt>:: <tt>['a', 'b', 'c']</tt>.
+ *      fn(d, fn(c, fn(b, fn(i, a))))
```
nobu, line 889:
```suggestion
 *      fn(fn(fn(fn(i, a), b), c), d)
```

enum.c
```diff
@@ -870,134 +870,151 @@ ary_inject_op(VALUE ary, VALUE init, VALUE op)
...
- *  - The third call passes the result of the second call,
- *    along with the fourth operand.
- *  - And so on.
+ *  <b>Second Shortcut: a Reducer function</b>
  *
- *  The return value from +inject+ is the return value from the last block call.
- *
- *  This call to +inject+ gives a block
- *  that writes the memo and element, and also sums the elements:
+ *  A _reducer function_ is a function that takes a partial result and the next value,
```
nobu, line 959:
```suggestion
 *  A <i>reducer function</i> is a function that takes a partial result and the next value,
```

PR #10012, opened 2024-02-18
Disable callcc when ASAN is enabled

KJTsanaktsidis:
callcc's implementation is fundamentally incompatible with ASAN. Since callcc is deprecated and almost never used, it's probably OK to disable callcc when ruby is compiled with ASAN.

This PR adds an autoconf test to detect when we're building with ASAN, and uses this to either define `RUBY_CALLCC_ENABLED` or not. Obviously the define defaults to on.

I also moved a lot of the `cont.c` code around so that all of the callcc implementation could be guarded by a single ifdef.

[Bug #20273]

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -1775,6 +1775,13 @@ rb_callcc(VALUE self)
         return rb_yield(val);
     }
 }
+#ifdef RUBY_ASAN_ENABLED
+/* callcc can't possibly work with ASAN; see bug #20273. Also this function
+ * definition below avoids a "defined and not used" warning. */
+MAYBE_UNUSED(static void notusing_callcc(void)) { rb_callcc(Qnil); }
+# define rb_callcc rb_f_notimplement
+#endif
+
 
 static VALUE
 make_passing_arg(int argc, const VALUE *argv)
--- a/test/ruby/test_array.rb
+++ b/test/ruby/test_array.rb
@@ -3561,6 +3561,7 @@ class TestArray < Test::Unit::TestCase
     unless respond_to?(:callcc, true)
       EnvUtil.suppress_warning {require 'continuation'}
     end
+    omit 'requires callcc support' unless respond_to?(:callcc, true)
   end
 end
 
--- a/test/ruby/test_beginendblock.rb
+++ b/test/ruby/test_beginendblock.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: false
 require 'test/unit'
+EnvUtil.suppress_warning {require 'continuation'}
 
 class TestBeginEndBlock < Test::Unit::TestCase
   DIR = File.dirname(File.expand_path(__FILE__))
@@ -131,6 +132,8 @@ class TestBeginEndBlock < Test::Unit::TestCase
   end
 
   def test_callcc_at_exit
+    omit 'requires callcc support' unless respond_to?(:callcc)
+
     bug9110 = '[ruby-core:58329][Bug #9110]'
     assert_ruby_status([], "#{<<~"begin;"}\n#{<<~'end;'}", bug9110)
     begin;
--- a/test/ruby/test_continuation.rb
+++ b/test/ruby/test_continuation.rb
@@ -4,6 +4,10 @@ EnvUtil.suppress_warning {require 'continuation'}
 require 'fiber'
 
 class TestContinuation < Test::Unit::TestCase
+  def setup
+    omit 'requires callcc support' unless respond_to?(:callcc)
+  end
+
   def test_create
     assert_equal(:ok, callcc{:ok})
     assert_equal(:ok, callcc{|c| c.call :ok})
--- a/test/ruby/test_enum.rb
+++ b/test/ruby/test_enum.rb
@@ -843,6 +843,8 @@ class TestEnumerable < Test::Unit::TestCase
   end
 
   def test_callcc
+    omit 'requires callcc support' unless respond_to?(:callcc)
+
     assert_raise(RuntimeError) do
       c = nil
       @obj.sort_by {|x| callcc {|c2| c ||= c2 }; x }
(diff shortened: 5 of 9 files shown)


nobu (changes requested):
Seems many unnecessary changes make this patch to read hard.

configure.ac
```diff
@@ -4175,6 +4175,32 @@ AS_IF([test "${universal_binary-no}" = yes ], [
 
 [end]_group
 
+AC_MSG_CHECKING([whether ASAN is enabled])
+AC_RUN_IFELSE([AC_LANG_PROGRAM([[]], [[
+#if (defined(__has_feature) && __has_feature(address_sanitizer)) || defined(__SANITIZE_ADDRES__)
+        return 0;
+#else
+        return 1;
+#endif
```
nobu, line 4184: Don't use `AC_RUN_IFELSE`.

```suggestion
AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[]], [[
#if (defined(__has_feature) && __has_feature(address_sanitizer)) || defined(__SANITIZE_ADDRES__)
        return 0;
#else
# error no address sanitizer
#endif
```

I don't think this is not needed to do in configure.

cont.c
```diff
@@ -1547,55 +1610,15 @@ cont_restore_thread(rb_context_t *cont)
     }
 }
 
-NOINLINE(static void fiber_setcontext(rb_fiber_t *new_fiber, rb_fiber_t *old_fiber));
 
-static void
-fiber_setcontext(rb_fiber_t *new_fiber, rb_fiber_t *old_fiber)
+static rb_context_t *
+cont_ptr(VALUE obj)
```
nobu, line 1615: Could you place before `cont_save_machine_stack` in the same order?
Shuffling functions makes this harder to read.

internal/sanitizers.h
```diff
@@ -16,17 +16,11 @@
...
-# if __has_feature(address_sanitizer)
-#  define RUBY_ASAN_ENABLED
-#  include <sanitizer/asan_interface.h>
-# endif
+# include <sanitizer/asan_interface.h>
 #endif
 
 #ifdef HAVE_SANITIZER_MSAN_INTERFACE_H
-# if __has_feature(memory_sanitizer)
-#  define RUBY_MSAN_ENABLED
```
nobu, line 27: These macros, `RUBY_ASAN_ENABLED` and `RUBY_MSAN_ENABLED`, are used in many places.
They are all unnecessary?

cont.c
```diff
@@ -1427,6 +1369,127 @@ show_vm_pcs(const rb_control_frame_t *cfp,
...
+      case 0:
+        return Qnil;
+      case 1:
+        return argv[0];
+      default:
+        return rb_ary_new4(argc, argv);
+    }
+}
+
+#ifdef RUBY_CALLCC_ENABLED
```
nobu, line 1438: This macro should be defined/used only in this file.

nobu, line 1438:
```suggestion
#if defined(RUBY_ASAN_ENABLED) || defined(RUBY_MSAN_ENABLED)
# define RUBY_CALLCC_ENABLED 1
#else
# undef RUBY_CALLCC_ENABLED
#endif
```

The other files should not define/use it.

KJTsanaktsidis: So I can do this, but there are parts of `vm_core.h`, `hash.c`, and `vm.c` which are there for callcc support. And I thought if I was going to have a switch which disabled callcc, it really should remove all of the bits of it throughout the codebase.

For example, `hash.c` calls `ruby_register_rollback_func_for_ensure`, and if i don't ifdef it out there, then I need to keep that function in here. And so some parts of callcc support would be compiled out, but some would be kept.

I think the best alternative to this kind of invasive patch would be to simply do

```
#ifdef RUBY_ASAN_ENABLED
    rb_define_global_function("callcc", rb_callcc, 0);
#else
    rb_define_global_function("callcc", rb_f_notimplement, 0);
#end
```

and then go and mark `rb_callcc` with `RBIMPL_ATTR_MAYBE_UNUSED`. But I wanted to check first if we should avoid skipping the warning like that, hence this PR.

configure.ac
KJTsanaktsidis: Ah of course I didn't realise it actually didn't need to run.

I put this in autoconf because it affected multiple files, but we're already having a discussion about whether that's a good idea or not in the other thread. But that's _why_ I put it in autoconf.

cont.c
KJTsanaktsidis: Apologies, I didn't mean to re-order these - just got a bit mixed up.

internal/sanitizers.h
KJTsanaktsidis: These are also defined in autoconf by this patch, which is why I deleted them from here.

KJTsanaktsidis:
@nobu thank you for the review, I very much appreciate it.

Let me know if you want me to #ifdef out _just_ the call to `rb_define_global_function` and ignore the warnings, or if you want me to #ifdef out as much of callcc as possible, but only inside cont.c?

cont.c
nobu: ASAN is for debugging purpose; it should not change behaviors as possible.

```C
#ifdef RUBY_ASAN_ENABLED
# define rb_callcc ((void)rb_callcc, rb_f_notimplement)
#endif
```

KJTsanaktsidis:
OK, I changed my PR do do roughly what you suggested @nobu . What do you think?

I couldn't _actually_ do `# define rb_callcc ((void)rb_callcc, rb_f_notimplement)` because the `RBIMPL_ANYARGS_DISPATCH` machinery special-cases `rb_f_notimplement` in a way that's somehow not compatible with the comma operator like this. I don't understand why it didn't work, but using `STATIC_ASSERT` achieves the same thing in the same place so I think let's go with that?

KJTsanaktsidis:
@nobu 🙏 would love if you could have a quick look at this again when you get a moment - thanks!

PR #10017, opened 2024-02-19
`Exception#set_backtrace` accept arrays of `Backtrace::Location`

casperisfine:
[[Feature #13557]](https://bugs.ruby-lang.org/issues/13557)

Setting the backtrace with an array of strings is lossy. The resulting exception will return nil on `#backtrace_locations`.

By accepting an array of `Backtrace::Location` instance, we can rebuild a `Backtrace` instance and have a fully functioning Exception.

~~NOTE: Exception's instance variables `id_bt` and `id_bt_locations` are a bit inconsistent right now and would deserve a cleanup.~~

### DIFF
--- a/vm_backtrace.c
+++ b/vm_backtrace.c
@@ -531,6 +531,16 @@ backtrace_alloc(VALUE klass)
     return obj;
 }
 
+static VALUE
+backtrace_alloc_capa(long num_frames, rb_backtrace_t **backtrace)
+{
+    size_t memsize = offsetof(rb_backtrace_t, backtrace) + num_frames * sizeof(rb_backtrace_location_t);
+    VALUE btobj = rb_data_typed_object_zalloc(rb_cBacktrace, memsize, &backtrace_data_type);
+    TypedData_Get_Struct(btobj, rb_backtrace_t, &backtrace_data_type, *backtrace);
+    return btobj;
+}
+
+
 static long
 backtrace_size(const rb_execution_context_t *ec)
 {
@@ -617,9 +627,7 @@ rb_ec_partial_backtrace_object(const rb_execution_context_t *ec, long start_fram
         }
     }
 
-    size_t memsize = offsetof(rb_backtrace_t, backtrace) + num_frames * sizeof(rb_backtrace_location_t);
-    btobj = rb_data_typed_object_zalloc(rb_cBacktrace, memsize, &backtrace_data_type);
-    TypedData_Get_Struct(btobj, rb_backtrace_t, &backtrace_data_type, bt);
+    btobj = backtrace_alloc_capa(num_frames, &bt);
 
     bt->backtrace_size = 0;
     if (num_frames == 0) {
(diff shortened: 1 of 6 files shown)


vm_backtrace.c
```diff
@@ -783,6 +791,36 @@ rb_backtrace_to_location_ary(VALUE self)
...
+        RB_OBJ_WRITE(btobj, &dst_location->cme, src_vloc->loc->cme);
+        RB_OBJ_WRITE(btobj, &dst_location->iseq, src_vloc->loc->iseq);
+        dst_location->pc = src_vloc->loc->pc;
+
+        new_backtrace->backtrace_size++;
+
+        RB_GC_GUARD(locobj);
+    }
+
+    RB_GC_GUARD(btobj);
```
KJTsanaktsidis, line 820: I think it's unnecessary to guard this if you're returning it since `btobj` obviously must either be in the stack or register state for the entirety of this frame since it's being returned?

KJTsanaktsidis (approved):
Looks reasonable to me.

> NOTE: Exception's instance variables id_bt and id_bt_locations are a bit inconsistent right now and would deserve a cleanup.

Agree this could use some love. Why do we even have both given that `exc_backtrace` seems happy enough for either strings or `Thread::Backtrace::Location`s to be in `id_bt`?

vm_backtrace.c
casperisfine: Yeah, I started to be a bit liberal with `GC_GUARD`...

casperisfine:
> Why do we even have both given that `exc_backtrace` seems happy enough for either strings or `Thread::Backtrace::Location`s to be in `id_bt`?

Yeah, that surprised me as well, I'd need to do some spelunking to see when it was introduced.

vm_backtrace.c
KJTsanaktsidis: The more I learn about how compilers work the more scared I am that on day some compiler will do something really clever like store a pointer to the middle of a struct in a register, rather than the actual VALUE, and break everything :/

I wonder if anybody has tried compiling Ruby with LTO, maybe aggressive enough cross-TU inlining could make this happen…

ko1:
```ruby
def foo
  raise
end

begin
  foo
rescue => e
  e.set_backtrace(caller_locations + [""])
end

=begin
-e:8:in 'Exception#set_backtrace': wrong argument type String (expected frame_info) (TypeError)

e.set_backtrace(caller_locations + [""])
^^^^^^^^^^^^^^^^^^^^^^^
from -e:8:in '<main>'
from -e:5:in '<main>'
-e:2:in 'Object#foo': unhandled exception
from -e:6:in '<main>'
=end

----

def foo
  raise
end

begin
  foo
rescue => e
  e.set_backtrace([""] + caller_locations)
end

=begin
-e:8:in 'Exception#set_backtrace': backtrace must be an Array of String or an Array of Thread::Backtrace::Location (TypeError)

e.set_backtrace([""] + caller_locations)
^^^^^^^^^^^^^^^^^^^^^^^
from -e:8:in '<main>'
from -e:5:in '<main>'
-e:2:in 'Object#foo': unhandled exception
from -e:6:in '<main>'
=end
```

can we make them same error?

> 'Exception#set_backtrace': backtrace must be an Array of String or an Array of Thread::Backtrace::Location (TypeError)

casperisfine:
> can we make them same error?

I think so, I'll try.

casperisfine:
Alright, the error message is now consistent. I'll merge when CI pass.

ko1:
Could you modify docs and maybe NEWS?

casperisfine:
> Could you modify docs and maybe NEWS?

Of course: https://github.com/ruby/ruby/pull/10255

Let me know what you think.

PR #10018, opened 2024-02-19
Add a hint of `ASCII-8BIT` being `BINARY`

casperisfine:
[[Feature #18576]](https://bugs.ruby-lang.org/issues/18576#note-37)

Since outright renaming `ASCII-8BIT` is deemed to backward incompatible, the next best thing would be to only change its `#inspect`, particularly in exception messages.

### DIFF
--- a/encoding.c
+++ b/encoding.c
@@ -1015,13 +1015,22 @@ rb_enc_get(VALUE obj)
     return rb_enc_from_index(rb_enc_get_index(obj));
 }
 
+const char *
+rb_enc_inspect_name(rb_encoding *enc)
+{
+    if (enc == global_enc_ascii) {
+        return "BINARY (ASCII-8BIT)";
+    }
+    return enc->name;
+}
+
 static rb_encoding*
 rb_encoding_check(rb_encoding* enc, VALUE str1, VALUE str2)
 {
     if (!enc)
         rb_raise(rb_eEncCompatError, "incompatible character encodings: %s and %s",
-                 rb_enc_name(rb_enc_get(str1)),
-                 rb_enc_name(rb_enc_get(str2)));
+                 rb_enc_inspect_name(rb_enc_get(str1)),
+                 rb_enc_inspect_name(rb_enc_get(str2)));
     return enc;
 }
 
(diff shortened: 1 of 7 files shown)


encoding.c
```diff
@@ -1015,13 +1015,22 @@ rb_enc_get(VALUE obj)
     return rb_enc_from_index(rb_enc_get_index(obj));
 }
 
+const char *
+rb_enc_inspect_name(rb_encoding *enc)
+{
+    if (enc == global_enc_ascii) {
+        return "ASCII-8BIT (BINARY)";
```
eregon, line 1022: Why not just `"BINARY"`?
It's an existing alias of it, so it seems fair to use that name in error messages.
```
(irb):8:in `+': incompatible character encodings: UTF-8 and BINARY (Encoding::CompatibilityError)
```
seems much clearer than
```
(irb):8:in `+': incompatible character encodings: UTF-8 and ASCII-8BIT (BINARY) (Encoding::CompatibilityError)
```

casperisfine: Because that's what was accepted on the ticket.

eregon: @nurse Would you be OK with `incompatible character encodings: UTF-8 and BINARY (Encoding::CompatibilityError)`?

For Encoding#inspect, I think `"#<Encoding:ASCII-8BIT (BINARY)>"` as fine, as already the case in the PR.
I'm just concerned the exception message `incompatible character encodings: UTF-8 and ASCII-8BIT (BINARY)` does not really solve the confusion of "incompatible UTF-8 and ASCII ???".

jpcamara:
As someone who was recently thrown off by this and had to research that ASCII-8BIT is binary, I'm happy to see this change 🎉

test/ruby/test_yjit.rb
```diff
@@ -1458,7 +1458,7 @@ def foo = bar
   end
 
   def test_str_concat_encoding_mismatch
-    assert_compiles(<<~'RUBY', result: "incompatible character encodings: ASCII-8BIT and EUC-JP")
+    assert_compiles(<<~'RUBY', result: "incompatible character encodings: BINARY (ASCII-8BIT) and EUC-JP")
```
eregon, line 1461: `BINARY (ASCII-8BIT)` seems a good compromise.

eregon (approved):

PR #10030, opened 2024-02-20
Fix string value in hash literal being forced frozen

XrXr:
We should pass `false` for `hash_key` for value nodes. Credits to
`@kddnewton` for noticing and bisecting.


yui-knk (approved):

yui-knk:
Thank you!

PR #10033, opened 2024-02-20
YJIT: Allow non-leaf calls on opt_* insns

k0kubun:
Follow-up on https://github.com/ruby/ruby/pull/10002

We allowed `jit_rb_str_concat` to make a non-leaf call as an exception because the interpreter skips pushing a frame too. This PR refactors https://github.com/ruby/ruby/pull/10002 to encode the rule in the `with_leaf_ccall()` call instead of reversing it in cfunc codegen.

I was going to clarify that by moving the implementation to `gen_opt_ltlt` using `BOP`, which would limit the optimization to `rb_cString`, but I want the optimization to be applied to all `T_STRING`s because Rails defines subclasses of `String`.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6098,10 +6098,16 @@ fn gen_send_cfunc(
         if let Some(known_cfunc_codegen) = lookup_cfunc_codegen(unsafe { (*cme).def }) {
             // We don't push a frame for specialized cfunc codegen, so the generated code must be leaf.
             // However, the interpreter doesn't push a frame on opt_* instruction either, so we allow
-            // non-send instructions to break this rule as an exception.
-            if asm.with_leaf_ccall(jit.is_sendish(), |asm|
+            // non-sendish instructions to break this rule as an exception.
+            let cfunc_codegen = if jit.is_sendish() {
+                asm.with_leaf_ccall(|asm|
+                    perf_call!("gen_send_cfunc: ", known_cfunc_codegen(jit, asm, ocb, ci, cme, block, argc, recv_known_class))
+                )
+            } else {
                 perf_call!("gen_send_cfunc: ", known_cfunc_codegen(jit, asm, ocb, ci, cme, block, argc, recv_known_class))
-            ) {
+            };
+
+            if cfunc_codegen {
                 assert_eq!(expected_stack_after, asm.ctx.get_stack_size() as i32);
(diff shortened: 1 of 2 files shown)


yjit/src/codegen.rs
```diff
@@ -298,6 +298,16 @@ impl JITState {
             }
         }
     }
+
+    /// Return true if we're compiling a send-like instruction, not an opt_* instruction.
+    pub fn on_send_insn(&self) -> bool {
```
maximecb, line 303:
```suggestion
    pub fn is_send_insn(&self) -> bool {
```

Maybe `is_send` or `is_sendish` ? 😅

yjit/src/codegen.rs
```diff
@@ -6088,7 +6097,9 @@ fn gen_send_cfunc(
         let expected_stack_after = asm.ctx.get_stack_size() as i32 - argc;
         if let Some(known_cfunc_codegen) = lookup_cfunc_codegen(unsafe { (*cme).def }) {
             // We don't push a frame for specialized cfunc codegen, so the generated code must be leaf.
-            if asm.with_leaf_ccall(|asm|
+            // However, the interpreter doesn't push a frame on opt_* instruction either, so we allow
+            // non-send instructions to break this rule as an exception.
+            if asm.with_leaf_ccall(jit.on_send_insn(), |asm|
```
maximecb, line 6102: I feel like this could be refactored to either call or not call `asm.with_leaf_call` instead of passing a boolean parameter which makes the function not do what the name implies 😅

yjit/src/codegen.rs
k0kubun: chose `is_sendish` https://github.com/ruby/ruby/pull/10033/commits/761fa5e22856cff1f1bb8f60ecf46b7ea13e3c87

yjit/src/codegen.rs
maximecb: I was thinking the latter. I think it's ok sometimes to repeat a bit of code (one expression here really). It just seems weird to have a function called `with_leaf_call` and then passing a boolean that says "well, actually not" 😅

maximecb (approved):

yjit/src/codegen.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/10033/commits/5955a44f2373db12dbf85d6bf871a4238cea7734

PR #10043, opened 2024-02-21
Explicitly removed tmpdir with Windows platform

hsbt:


hsbt:
We will revert related commits https://github.com/ruby/ruby/pull/10044 or https://github.com/ruby/ruby/pull/10046

PR #10044, opened 2024-02-21
Revert "[rubygems/rubygems] Reset `Gem.state_file` along with `Gem.state_home`"

hsbt:
6ecb14df2cda223e2280891150b3f41c1eb2edf0 is broken with Windows platform

https://github.com/ruby/ruby/actions/runs/7986321662/job/21806489509


hsbt:
Use https://github.com/ruby/ruby/pull/10046 instead of this.

PR #10051, opened 2024-02-21
Use `'\n'?` instead of `opt_nl`

ydah:
Change to use Parameterizing rules.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -3234,7 +3234,7 @@ endless_command : command
                         $$ = rescued_expr(p, $1, $4, &@1, &@2, &@4);
                     /*% ripper: rescue_mod!($:1, $:4) %*/
                     }
-                | keyword_not opt_nl endless_command
+                | keyword_not '\n'? endless_command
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -6942,21 +6942,17 @@ call_op2	: call_op
...
                 ;
 
-rbrace		: opt_nl '}'
+rbrace		: '\n'? '}'
                 ;
 
-trailer		: opt_nl
-                | ','
+trailer		: '\n'?
+            | ','
```
nobu, line 6955: This line too.

ydah:
> Maybe we should wait for lrama to implement `'\n'?`?

I thought `option (...) `might be easier to understand, but '\n'? sounds good. I changed it.

> BTW, why changing indents?

Oh, sorry. I put it back.

parse.y
ydah: Oops, I put it back.

nobu (approved):

PR #10055, opened 2024-02-21
[Feature #20290] Add API for C extensions to free memory

peterzhu2118:
[Ticket #19993](https://bugs.ruby-lang.org/issues/19993) added the new feature RUBY_FREE_AT_EXIT, which frees memory in Ruby at shutdown. This allowed tools like Valgrind, ASAN, and macOS leaks to find memory leaks in Ruby without a large number of false-positives outputted. However, this feature is not complete for C extensions, as they may also need to free their memory and there was no way to do so. This means that C extensions might not be able to directly use tools like Valgrind, ASAN, or macOS leaks to find memory leaks.

This ticket proposes an API for C extensions to free memory by defining a function called `Destruct_<extension name>` that is called during shutdown when RUBY_FREE_AT_EXIT is enabled. This name mirrors the `Init_<extension name>` API that already exists for extension initialization. However, unlike the `Init_<extension name>` function, `Destruct_<extension name>` is NOT mandatory for the C extension to implement so that we can preserve backwards compatibility.

### DIFF
--- a/dln.c
+++ b/dln.c
@@ -540,3 +540,26 @@ dln_load(const char *file)
 
     return 0;			/* dummy return */
 }
+
+#define DESTRUCT_FUNC_PREFIX EXTERNAL_PREFIX"Destruct_"
+
+void
+dln_unload(const char *file, void *handle)
+{
+#if defined(_WIN32) || defined(USE_DLN_DLOPEN)
+    struct string_part extname = init_funcname_len(file);
+    const size_t plen = sizeof(DESTRUCT_FUNC_PREFIX) - 1;
+    char *const destruct_func_name = ALLOCA_N(char, plen + extname.len + 1);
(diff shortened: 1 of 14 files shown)


dln.c
```diff
@@ -531,3 +531,27 @@ dln_load(const char *file)
 
     return 0;			/* dummy return */
 }
+
+#define DESTRUCT_FUNC_PREFIX "Destruct_"
```
nobu, line 535:
```suggestion
#define DESTRUCT_FUNC_PREFIX EXTERNAL_PREFIX"Destruct_"
```

nobu: And on many Unix-like systems, the opposite of `init` is `fini` historically.

peterzhu2118: I'm open to calling it Fini as well.

PR #10060, opened 2024-02-21
YJIT: Grab stack operands after SP change in String#byteslice

XrXr:
Previously, `StackOperand`s caching `sp_offset` was held across a
jit_prepare_call_with_gc(), which invalidates the offsets. With the
right register allocation state, the canary overlapped with the old
address of the receiver and YJIT clobbered the receiver writing the
canary.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1594,6 +1594,12 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_byteslice_sp_invalidation
+    assert_compiles(<<~'RUBY', result: 'ok', no_send_fallbacks: true)
+      "okng".itself.byteslice(0, 2)
+    RUBY
+  end
+
   private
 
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
I was writing the exact same thing 😄 Looks good.

k0kubun:
Here's the rest of the change I wanted to file as part of the fix: https://github.com/ruby/ruby/pull/10061

XrXr:
> We are killing the stack canary set by leave, at <RubyVM::InstructionSequence:webauthn_verification_poll_response@/home/runner/work/ruby/ruby/src/lib/rubygems/gemcutter_utilities/webauthn_poller.rb:67>@pc=29

Seems to be revealing some other problem 🤔

maximecb (approved):

k0kubun:
> EDIT: it's happening on master as well, so unrelated to this.

It seems like the case. Let's merge this first and fix it separately.

PR #10079, opened 2024-02-23
Remove ruby objects from numeric node comparison

yui-knk:
Parser needs to compare nodes to warn duplicate keys in a hash, for example:

```
{1 => :a, 1 => :b}
# => test.rb:1: warning: key 1 is duplicated and overwritten on line 1
```

This commit stop using ruby objects for hash value calculation and node comparison.
Bignum is needed when comparing different radix numeric nodes.

### DIFF
--- a/node.c
+++ b/node.c
@@ -177,6 +177,12 @@ parser_string_free(rb_ast_t *ast, rb_parser_string_t *str)
 }
 
 static void
+parser_bigfree(rb_ast_t *ast, rb_parser_bignum_t *big)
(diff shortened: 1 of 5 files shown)


nobu:
🤔 …

yui-knk:
We discussed what should be warned in https://bugs.ruby-lang.org/issues/20331.
I created and applied the PR (https://github.com/ruby/ruby/pull/10420) based on the discussion then close this ticket.

PR #10080, opened 2024-02-23
YJIT: Lazily push a frame for specialized C funcs

k0kubun:
This PR adds the capability to lazily push a frame for specialized C methods. It saves the PC before a C call, remembers `{ PC => cme }`, and lazily pushes a frame using it when it calls non-leaf functions like `rb_raise` or `rb_to_int`.

To demonstrate the capability, this PR also merges `String#setbyte` specialization https://github.com/ruby/ruby/pull/9767. While the JIT code never pushes a frame for `String#setbyte`, it gets a `String#setbyte` frame when a backtrace is obtained inside the C call.

```
before: ruby 3.4.0dev (2024-02-23T04:55:43Z master f403660805) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-02-23T05:38:59Z yjit-lazy-frame 0845cbddb6) +YJIT [x86_64-linux]

ruby-xor  88.4         0.3         82.6        0.3         1.07           1.07
```


maximecb:
Looking at this PR 👀 

Nice results on the `ruby-xor` microbenchmark 👍

yjit/src/codegen.rs
```diff
@@ -9769,6 +9839,9 @@ pub struct CodegenGlobals {
 
     /// Page indexes for outlined code that are not associated to any ISEQ.
     ocb_pages: Vec<usize>,
+
+    /// Lazily push a frame when it calls rb_yjit_check_pc() with a PC in this HashMap
```
maximecb, line 9843:
```suggestion
    /// Map of cfunc YARC PCs to CMEs, used to
    /// lazily push a frame when rb_yjit_check_pc() is called with a PC in this HashMap
```

yjit/src/yjit.rs
```diff
@@ -207,3 +207,19 @@ pub extern "C" fn rb_yjit_simulate_oom_bang(_ec: EcPtr, _ruby_self: VALUE) -> VA
 
     return Qnil;
 }
+
+/// Push a C method frame if the given PC is supposed to lazily push one.
+/// This is called from rb_raise() (at rb_exc_new_str()) and other functions
+/// that may make a method call (e.g. rb_to_int()).
+#[no_mangle]
+pub extern "C" fn rb_yjit_check_pc(pc: *mut VALUE) {
```
maximecb, line 215: Small nitpick: I feel like this method name could be more descriptive, maybe `rb_yjit_lazy_push_frame` or something like that? The current name makes it sound like it's an error or safety check.

yjit/src/codegen.rs
k0kubun: applied at https://github.com/ruby/ruby/pull/10080/commits/f80b64a74043282b4f2dc3ee0923ad8563c215a8 (and https://github.com/ruby/ruby/pull/10080/commits/9372f4f93cd8128d320c5cb7f558fd3492e5e5d8)

yjit/src/yjit.rs
k0kubun: renamed it to `rb_yjit_lazy_push_frame` https://github.com/ruby/ruby/pull/10080/commits/b153483379036187cadbf8c24c8449d862261290

k0kubun:
> I can apply this to String#getbyte and String#byteslice to get rid of all existing "give up because it's not leaf" scenarios

Done https://github.com/ruby/ruby/pull/10080/commits/340cfb51354e3a11375e9d823520852cdc584b86 https://github.com/ruby/ruby/pull/10080/commits/c4bd35e0d4758a27c66d36005a2afe74bd2a293b. For `String#byteslice`, I added `-1` argc support too.

maximecb (approved):
Well done 👌

PR #10081, opened 2024-02-23
[ruby/fileutils] Fix optional argument descriptions

akihikodaki:
Several optional positional arguments were incorrectly denoted as keyword arguments so correct them.


nobu:
Could you send this to https://github.com/ruby/fileutils?

PR #10085, opened 2024-02-23
Don't lookup finalizers if FL_FINALIZE flag not set

peterzhu2118:
The FL_FINALIZE flag is set when there are finalizers for the object. We can improver performance by not looking up in the table if the flag is not set.

Using the following C extension:

```c
#include "ruby/ruby.h"

static void data_free(void *_ptr) {}

static const rb_data_type_t data_type = {
    "my_type",
    {
        NULL,
        data_free,
    },
    0, 0, 0
};

static VALUE data_alloc(VALUE klass) {
    return TypedData_Wrap_Struct(klass, &data_type, (void *)1);
}

void Init_myext(void) {
    VALUE my_klass = rb_define_class("MyClass", rb_cObject);
    rb_define_alloc_func(my_klass, data_alloc);
}
```

And the following benchmark:

```ruby
require "benchmark"

final_objs = 1_000_000.times.map do
  o = Object.new
  ObjectSpace.define_finalizer(o, proc {})
  o
end

puts(Benchmark.measure do
  100_000_000.times do
    MyClass.new
  end
end)
```

Before:

    10.974190   0.355037  11.329227 ( 11.416772)

After:

    7.664310   0.347598   8.011908 (  8.268969)

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -3143,7 +3143,7 @@ static inline void
 make_zombie(rb_objspace_t *objspace, VALUE obj, void (*dfree)(void *), void *data)
 {
     struct RZombie *zombie = RZOMBIE(obj);
-    zombie->basic.flags = T_ZOMBIE | (zombie->basic.flags & FL_SEEN_OBJ_ID);
+    zombie->basic.flags = T_ZOMBIE | (zombie->basic.flags & (FL_SEEN_OBJ_ID | FL_FINALIZE));
     zombie->dfree = dfree;
     zombie->data = data;
     VALUE prev, next = heap_pages_deferred_final;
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -4115,15 +4115,19 @@ run_finalizer(rb_objspace_t *objspace, VALUE obj, VALUE table)
...
 
-    key = (st_data_t)zombie;
-    if (st_delete(finalizer_table, &key, &table)) {
-        run_finalizer(objspace, zombie, (VALUE)table);
+    st_data_t key = (st_data_t)zombie;
+    if (FL_TEST_RAW(zombie, FL_FINALIZE)) {
+        st_data_t table;
+        if (st_delete(finalizer_table, &key, &table)) {
+            run_finalizer(objspace, zombie, (VALUE)table);
+        }
```
ko1, line 4127: `rb_bug()` for `else`?

ko1 (approved):

PR #10087, opened 2024-02-24
Skip assert_no_memory_leak when ASAN is enabled

KJTsanaktsidis:
ASAN greatly increases the memory footprint of Ruby, so these static thresholds are not appropriate. There's no real need to run these tests under ASAN.

[Bug #20274]

### DIFF
--- a/ext/-test-/asan/asan.c
+++ b/ext/-test-/asan/asan.c
@@ -1,0 +1,24 @@
+#include "ruby/ruby.h"
+
+static VALUE
+asan_enabled_p(VALUE self)
+{
+#if defined(__has_feature)
+    /* clang uses __has_feature for determining asan */
+    return __has_feature(address_sanitizer) ? Qtrue : Qfalse;
+#elif defined(__SANITIZE_ADDRESS__)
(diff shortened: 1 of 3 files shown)


ext/-test-/asan/asan.c
```diff
@@ -0,0 +1,28 @@
...
+#if defined(__SANITIZE_ADDRESS)
+    /* GCC defines this when asan is enabled */
+    return Qtrue;
+#elif defined(__has_feature)
+    /* clang uses __has_feature for determining asan */
+# if __has_feature(address_sanitizer)
+    return Qtrue;
+# else
+    return Qfalse;
+# endif
```
nobu, line 15: `__has_feature` is expanded to a compile-time constant, not a preprocessor constant.
So you can write as:

```suggestion
    return __has_feature(address_sanitizer) ? Qtrue : Qfalse;
```

KJTsanaktsidis: Oh that's a good idea, I've done that.

nobu (approved):

PR #10092, opened 2024-02-26
Add `Hash#update_value`

amomchilov:


hash.c
```diff
@@ -2906,8 +2923,37 @@ hash_aset_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existi
...
+// Like +hash_aset_str+, but stashes away the pre-existing value (if any) into +arg->arg+
+static int
+hash_update_value_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existing)
+{
+    if (!existing && !RB_OBJ_FROZEN(*key)) {
+        *key = rb_hash_key_str(*key);
+    }
+    return hash_update_value(key, val, arg, existing);
+}
+
```
nobu, line 2952:
```suggestion
```

hash.c
```diff
@@ -2906,8 +2923,37 @@ hash_aset_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existi
...
 NOINSERT_UPDATE_CALLBACK(hash_aset)
 NOINSERT_UPDATE_CALLBACK(hash_aset_str)
+NOINSERT_UPDATE_CALLBACK(hash_update_value)
+NOINSERT_UPDATE_CALLBACK(hash_update_value_str)
```
nobu, line 2956:
```suggestion
NOINSERT_UPDATE_CALLBACK(hash_exchange_value)
```

hash.c
```diff
@@ -2936,19 +2982,44 @@ NOINSERT_UPDATE_CALLBACK(hash_aset_str)
...
+        .hash = hash,
+        .key  = key,
+        .value = val,
+    };
+
+    tbl_update2(&update_info);
+
+    const VALUE pre_existing_value = update_info.arg;
+
+    return pre_existing_value != Qundef ? pre_existing_value : Qnil;
```
nobu, line 3020:
```suggestion
    rb_hash_modify(hash);
    RHASH_UPDATE(hash, key, hash_exchange_value, val);
    return update_info.arg;
```

hash.c
```diff
@@ -2936,19 +2982,44 @@ NOINSERT_UPDATE_CALLBACK(hash_aset_str)
...
     else {
-        RHASH_UPDATE_ITER(hash, iter_p, key, hash_aset_str, val);
+        RHASH_UPDATE(hash, key, hash_aset_str, val);
     }
+
     return val;
 }
 
+VALUE
+rb_hash_update_value(VALUE hash, VALUE key, VALUE val)
```
nobu, line 2998:
```suggestion
rb_hash_exchange_value(VALUE hash, VALUE key, VALUE val)
```

hash.c
```diff
@@ -7155,6 +7226,7 @@ Init_Hash(void)
     rb_define_method(rb_cHash, "fetch", rb_hash_fetch_m, -1);
     rb_define_method(rb_cHash, "[]=", rb_hash_aset, 2);
     rb_define_method(rb_cHash, "store", rb_hash_aset, 2);
+    rb_define_method(rb_cHash, "update_value", rb_hash_update_value, 2);
```
nobu, line 7229:
```suggestion
    rb_define_method(rb_cHash, "exchange_value", rb_hash_exchange_value, 2);
```

nobu (commented):
- Why is `tbl_update2` needed?
- `update_value` doesn't feel to return the previous value, what about `exchange_value`?

hash.c
```diff
@@ -2906,8 +2923,37 @@ hash_aset_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existi
...
+{
+    if (existing) {
+        st_data_t pre_existing_value = *val;
+        *val = arg->arg;
+        arg->arg = pre_existing_value;
+    }
+    else {
+        *val = arg->arg;
+        arg->arg = Qundef;
+    }
```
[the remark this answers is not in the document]
amomchilov: `Qnil`, or `Qundef`?

hash.c
```diff
@@ -2906,8 +2923,37 @@ hash_aset_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existi
     return hash_aset(key, val, arg, existing);
 }
 
+// Like +hash_aset+, but stashes away the pre-existing value (if any) into +arg->arg+
+static int
+hash_update_value(st_data_t *key, st_data_t *val, struct update_arg *arg, int existing)
```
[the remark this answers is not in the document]
amomchilov: Ooo I like this.

hash.c
amomchilov: Why should this be deleted?

(I'm not sure why it should exist, either, I just added it copying from the example of `hash_aset`/`hash_aset_str` 😅 )

amomchilov:
@nobu Wow that was a fast review! I left some comments to ask questions and explain why I added `tbl_update2`, but you reviewed before I could even post them! :D

I think your use of `RHASH_UPDATE(hash, key, hash_exchange_value, val);` achieves what I was trying to do by extracting out `tbl_update()`'s guts into the shared `tbl_update2()` function. That looks really promising! I'll look at it in the morning with fresh eyes.

hash.c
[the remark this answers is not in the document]
amomchilov: OOOOhhhh this is nice. If we set `Qnil` here, then we no longer need this snippet (which just replaced `Qundef` with `Qnil`):

https://github.com/amomchilov/ruby/blob/d19bedffc0a5060245b6c663ff63e5c765509294/hash.c#L3018-L3020

And then, because we no longer access the value of `update_info` after the `tbl_update2()` call, we no longer need the `tbl_update2()`!

hash.c
amomchilov: I don't think this works, because you don't have access to the `update_info.arg` that you're trying to return.

nobu:
```C
static int
hash_exchange_value(st_data_t *key, st_data_t *val, struct update_arg *arg, int existing)
{
    st_data_t *arg_value = (st_data_t *)arg->arg;
    st_data_t pre_existing_value = existing ? *val : Qnil;
    *val = *arg_value;
    *arg_value = pre_existing_value;

    return ST_CONTINUE;
}

NOINSERT_UPDATE_CALLBACK(hash_exchange_value)

static VALUE
rb_hash_exchange_value(VALUE hash, VALUE key, VALUE val)
{
    st_data_t value = (st_data_t)val;
    rb_hash_modify(hash);

    RHASH_UPDATE(hash, key, hash_exchange_value, &value);
    return (VALUE)value;
}
```

amomchilov: Neat! This is great.

PR #10093, opened 2024-02-26
Optimize `Set#add?` by only doing 1 hash look-up

amomchilov:
Builds on top of #10092


hash.c
```diff
@@ -1698,22 +1699,38 @@ tbl_update_modify(st_data_t *key, st_data_t *val, st_data_t arg, int existing)
     return ret;
 }
 
+static int
+tbl_update2(struct update_arg *update_info);
```
amomchilov, line 1703: Rearranging `tbl2_update` before `tbl_update` caused a lot of diff churn, which made the PR difficult to read.

I added this prototype for now, just to make reviewing easier.

hash.c
```diff
@@ -1698,22 +1699,38 @@ tbl_update_modify(st_data_t *key, st_data_t *val, st_data_t arg, int existing)
...
+        .value = (VALUE)optional_arg, // This is the object that will be protected by the write barrier
     };
 
-    int ret = rb_hash_stlike_update(hash, key, tbl_update_modify, (st_data_t)&arg);
+    return tbl_update2(&update_info);
+}
+
+// TODO: What's a good name for this
+static int
+tbl_update2(struct update_arg *update_info)
```
amomchilov, line 1725: What's a good name for this?

hash.c
```diff
@@ -2936,19 +2982,44 @@ NOINSERT_UPDATE_CALLBACK(hash_aset_str)
...
+        ? UPDATE_CALLBACK(iter_p, hash_update_value)
+        : UPDATE_CALLBACK(iter_p, hash_update_value_str);
+
+    struct update_arg update_info = {
+        .arg = val,
+        .func = func,
+        .hash = hash,
+        .key  = key,
+        .value = val,
+    };
```
amomchilov, line 3014: This is basically the same as the `update_info` in `tbl_update()`, except we need to access it, so we can fish out the `.arg` after `tbl_update()` finished.

That's why I extracted out `tbl2_update()`.

lib/set.rb
```diff
@@ -521,9 +521,12 @@ def add(o)
   #     Set[1, 2].add?([3, 4])               #=> #<Set: {1, 2, [3, 4]}>
   #     Set[1, 2].add?(2)                    #=> nil
   def add?(o)
-    add(o) unless include?(o)
+    self unless @hash.update_value(o, true)
```
amomchilov, line 524: This alternative implementation had promising benchmark results. We should investigate it, too.

```ruby
def add?(o)
  (@hash.update_value(o, true) || nil) && self
end
```

It's really unreadable, though :D

nobu:
Please put a benchmark YAML file under `benchmark` directory, not particular file at the top.
And `set` has its separated upstream repository, https://github.com/ruby/set.

amomchilov:
@nobu This is just a rough proof-of-concept. I'll extract it to the correct repo, polish off the benchmarks, etc. when I get more feedback on the bug tracker indicating that people actually want this

PR #10096, opened 2024-02-26
Fix compilation for platforms without pthread

pbo-linaro:
Found when compiling ruby for windows-arm64 using msys2

Missing return type for function Init_lock_native_thread

```
lock_native_thread.c:45:1: error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int [-Wimplicit-int]
   45 | Init_lock_native_thread(void)
      | ^
      | int
```

### DIFF
--- a/ext/-test-/thread/lock_native_thread/lock_native_thread.c
+++ b/ext/-test-/thread/lock_native_thread/lock_native_thread.c
@@ -42,6 +42,7 @@ Init_lock_native_thread(void)
 }
 
 #else // HAVE_PTHREAD_H
+void
 Init_lock_native_thread(void)
 {
     // do nothing


ext/-test-/thread/lock_native_thread/lock_native_thread.c
```diff
@@ -42,7 +42,7 @@ Init_lock_native_thread(void)
 }
 
 #else // HAVE_PTHREAD_H
-Init_lock_native_thread(void)
+void Init_lock_native_thread(void)
```
nobu, line 45:
```suggestion
void
Init_lock_native_thread(void)
```

PR #10097, opened 2024-02-26
Resolve a warning in `launchable record build` command

ono-max:
When executing the `launchable record build` command, a warning is issued, as shown here: [link to the warning](https://github.com/ruby/ruby/actions/runs/8029119718/job/21935044266#step:10:179). This warning occurs because Blob objects cannot be retrieved in the cloned repository when using sparse checkout, resulting in a "blob-less" clone. You can find more information about this behavior [here](https://github.com/actions/checkout/blob/b32f140b0c872d58512e0a66172253c302617b90/src/git-source-provider.ts#L166).

In the ruby/ruby repository, the source code is cloned into the `src/` directory without employing sparse checkout. Therefore, by executing the `launchable record build` command in the `src/` directory, we can resolve the warning.

### DIFF
--- a/.github/actions/setup/directories/action.yml
+++ b/.github/actions/setup/directories/action.yml
@@ -44,6 +44,11 @@ inputs:
     description: >-
       If set to true, creates dummy files in build dir.
 
+  fetch-depth:
+    required: false
+    default: '1'
+    description: The depth of commit history fetched from the remote repository
+
 outputs: {} # nothing?
 
 runs:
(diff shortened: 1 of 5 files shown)


.github/actions/setup/directories/action.yml
```diff
@@ -44,6 +44,11 @@ inputs:
     description: >-
       If set to true, creates dummy files in build dir.
 
+  fetch-depth:
+    required: false
+    default: ''
```
hsbt, line 49: Should we set this default value to `1` same as https://github.com/actions/checkout/blob/main/action.yml#L72?

ono-max: Setting the default value `1` is not mandatory because I confirmed the default value is configured correctly when `fetch-depth` is not configured at https://github.com/ruby/ruby/actions/runs/8047076163/job/21975601001?pr=10097#step:4:92. However, I think it would be more readable if we set the default value `1` through this review. Let me change it. Thanks!

ono-max: Updated

hsbt: I see. It's not strong opinion because I'm not sure what happend to pass empty value into `actions/checkout`.

PR #10108, opened 2024-02-26
YJIT: Support splat with C methods with -1 arity

XrXr:
Usually we deal with splats by speculating that they're of a specific
size. In this case, the C method takes a pointer and a length, so
we can support changing sizes just fine.

On lobsters:

```
num_send_dynamic:            727,218 ( 2.6%)
num_send_dynamic:            684,529 ( 2.5%)
```

### DIFF
--- a/yjit.c
+++ b/yjit.c
@@ -894,7 +894,7 @@ rb_yjit_ruby2_keywords_splat_p(VALUE obj)
 VALUE
 rb_yjit_splat_varg_checks(VALUE *sp, VALUE splat_array, rb_control_frame_t *cfp)
 {
-    if (!RB_TYPE_P(splat_array, T_ARRAY)) return Qfalse;
+    // We inserted a T_ARRAY guard before this call
     long len = RARRAY_LEN(splat_array);
 
     // Large splat arrays need a separate allocation
--- a/yjit.c
+++ b/yjit.c
@@ -890,6 +890,51 @@ rb_yjit_ruby2_keywords_splat_p(VALUE obj)
     return FL_TEST_RAW(last, RHASH_PASS_AS_KEYWORDS);
 }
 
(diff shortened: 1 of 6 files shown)


maximecb (approved):
Open to merging it, but my question would be, is this faster than the dynamic fallback since it's adding two C calls?

XrXr:
> is this faster than the dynamic fallback?

```
                     yjit-pre --yjit  yjit-post --yjit 
       var_len_splat         33.948M           35.412M i/s -     10.000M times in 0.294571s 0.282392s

Comparison:
                    var_len_splat
    yjit-post --yjit:  35411768.0 i/s 
     yjit-pre --yjit:  33947673.1 i/s - 1.04x  slower

$ cat ../../splat_neg.yml 
prelude:
  splat = [0, 1]
benchmark:
  var_len_splat: |
    splat.fetch(*splat)
loop_count: 10_000_000
```

Yes, even with the C calls we still win out by having static calls and no overhead for falling back.

yjit.c
```diff
@@ -890,6 +890,51 @@ rb_yjit_ruby2_keywords_splat_p(VALUE obj)
     return FL_TEST_RAW(last, RHASH_PASS_AS_KEYWORDS);
 }
 
+// Checks to establish preconditions for rb_yjit_splat_varg_cfunc()
+VALUE
+rb_yjit_splat_varg_checks(VALUE *sp, VALUE splat_array, rb_control_frame_t *cfp)
+{
+    if (!RB_TYPE_P(splat_array, T_ARRAY)) return Qfalse;
```
k0kubun, line 897: Is this something we could optimize away if we do this using `guard_object_is_array`, or is the information often wiped out on benchmarks?

k0kubun (approved):

yjit.c
XrXr: It's optimizable, yeah. I'll toss it in.

PR #10112, opened 2024-02-27
Switch `{prism: true}` to `{parser: :prism}` in ISeq.to_a

kddnewton:
A follow-up to https://github.com/ruby/ruby/pull/9934#issuecomment-1958401060.

@ko1 could you review? Sorry for missing the comment the first time!


kddnewton:
cc @mame also tagging you just because you reviewed the other one

nobu (approved):

PR #10118, opened 2024-02-27
Don't assume that Qfalse is 0 in rb_special_const_p

peterzhu2118:


nobu:
I'm against changing `Qfalse` value.

peterzhu2118:
I'm not planning on changing the value of `Qfalse`, but I don't think we should assume that `Qfalse` is 0 in case we want to experiment with other values in the future.

nobu:
No reason to use other values, I think.

XrXr:
Yeah, it's just too hard/incompatible to change at this point. And not much to gain from it.

include/ruby/internal/special_consts.h
```diff
@@ -346,7 +346,7 @@ RBIMPL_ATTR_CONSTEXPR(CXX11)
 static inline VALUE
 rb_special_const_p(VALUE obj)
 {
-    return RB_SPECIAL_CONST_P(obj) * RUBY_Qtrue;
```
amomchilov, line 349: Was this doing... multiplication?! 🧐

byroot:
IMO this change is justified simply by readability. `bool * RUBY_Qtrue` is a nice hack but is absolutely not obvious. I don't see any argument not to change it.

PR #10121, opened 2024-02-28
Update bundled_gems

soutaro:
https://github.com/ruby/rbs/pull/1758


soutaro:
This is an experiment to see if the `EBADF` disappears in ruby repository.

PR #10133, opened 2024-02-28
Migrate auto_request_review back to CODEOWNERS

k0kubun:
We switched from CODEOWNERS to `auto_request_review` for sending review requests automatically because the YJIT team used to have members who don't have write access and CODEOWNERS doesn't support teams or people without write access on the repository.

`auto_request_review` has a couple of minor problems:

* The notification is sent from @matzbot, so you can't know who filed the pull request until you open it.
* The `last_files_match_only` feature has not been working as expected. So we currently can't exclude `yjit/src/cruby_bindings.inc.rs` from the review target.

At the moment, @ruby/yjit has only committers, so we don't need to use `auto_request_review`. I gave write access to @ruby/yjit so that we could use CODEOWNERS with it.

Until we get a new team member, I'd like to switch this back to CODEOWNERS.


XrXr (approved):
👍🏼

PR #10136, opened 2024-02-28
YJIT: Squash canary before falling back

XrXr:
Recent flaky canary-related CI failures have all happened while trying
to fall back. It's unclear what is leaving the canary on the stack and
causing gen_send_dynamic() to falsely assume that it should be leaf,
and this patch isn't going to help us find the source. One source I
found is Array#<< with a frozen array, but it's unclear if that's what's
causing the CI failures. I'm somewhat afraid to add a canary check to
rb_longjmp() since that might introduce more flaky failures, and maybe
ones unrelated to YJIT.

See: https://github.com/ruby/ruby/actions/runs/8083502532/job/22086714152
See: https://github.com/ruby/ruby/actions/runs/8066858522/job/22035963315

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -8006,6 +8006,12 @@ fn gen_send_dynamic<F: Fn(&mut Assembler) -> Opnd>(
     // Save PC and SP to prepare for dynamic dispatch
     jit_prepare_non_leaf_call(jit, asm);
 
+    // Squash stack canary that might be left over from elsewhere
+    assert_eq!(false, asm.get_leaf_ccall());
+    if cfg!(debug_assertions) {
+        asm.store(asm.ctx.sp_opnd(0), 0.into());
+    }
+
     // Dispatch a method
     let ret = vm_sendish(asm);
 


yjit/src/codegen.rs
```diff
@@ -8006,6 +8006,12 @@ fn gen_send_dynamic<F: Fn(&mut Assembler) -> Opnd>(
     // Save PC and SP to prepare for dynamic dispatch
     jit_prepare_non_leaf_call(jit, asm);
 
+    // Squash stack canary that might be left over from elsewhere
+    assert_eq!(false, asm.get_leaf_ccall());
```
k0kubun, line 8010: I was wondering if we can do this in `jit_prepare_non_leaf_call()`. I think https://github.com/ruby/ruby/pull/10033 satisfied the assumption.

yjit/src/codegen.rs
```diff
@@ -8006,6 +8006,12 @@ fn gen_send_dynamic<F: Fn(&mut Assembler) -> Opnd>(
...
+    if cfg!(debug_assertions) {
+        asm.store(asm.ctx.sp_opnd(0), 0.into());
```
k0kubun, line 8012: Looking at past backtraces, fallbacks are indeed suspicious sources. But I'm not sure if this helps. If we set a canary on `asm.ccall`, it should have been cleared if `longjmp` did not happen. If `longjmp` did happen, however, it wouldn't reach this fallback code. WDYT?

k0kubun: Though, if the canary is not about the one set by a ccall prior to this fallback but something set by a completely irrelevant block (that was left after a longjmp), that might make sense 🤔 So maybe this is necessary.

To give us more information, could we insert a C call here to trigger `rb_bug` right away if the canary is set, instead of suppressing the assertion in the fallback?

XrXr: The situation I'm imaging is some canary setting `asm.ccall` raises, which leaves the canary on the stack. Later, the stack grows to be just the same size again, and fallback checks the canary by accident.

XrXr:
Looks like moving the assert doesn't work because jit_prepare_non_leaf_call() uses it. Maybe we should just clear the expectation in jit_prepare_non_leaf_call()?

k0kubun:
As I commented at https://github.com/ruby/ruby/pull/10136#discussion_r1506705411, I think the failure comes only from `jit_prepare_lazy_frame_call`. We indeed utilizes the leaf verification on `jit_prepare_lazy_frame_call`, so please don't clear it for `jit_prepare_lazy_frame_call`.

Feel free to resurrect the original patch, or we could assert that in `jit_prepare_non_leaf_call` only when it's not called from `jit_prepare_lazy_frame_call`.

yjit/src/codegen.rs
k0kubun: Yeah, so I think you thought the same thing as https://github.com/ruby/ruby/pull/10136#discussion_r1506710266.

Before just clearing it, could we add a runtime assertion here (using a C call that may call `rb_bug`) so that we'll know we indeed need to clear it?

XrXr: > To give us more information, could we insert a C call here to trigger rb_bug right away if the canary is set, instead of suppressing the assertion in the fallback?

I think that would give us mostly the same backtrace, because the fallback functions pretty much all start by pushing a frame, and that's when they check for the canary.

To find who left the canary I think we'd need to check in rb_longjmp()

k0kubun: I agree it's the case for cases like [this](https://github.com/ruby/ruby/actions/runs/8083502532/job/22086714152) where it's coming from a fallback function `rb_vm_opt_send_without_block`. However, failures on `webauthn_poller` like [this](https://github.com/ruby/ruby/actions/runs/7992014651/job/21824580101#step:13:101) don't have a fallback function in the backtrace, so the theory doesn't necessarily apply to all cases.

But, hmm, I guess it doesn't really help both cases 🤷 So I guess it's not worth an assertion unless we do it earlier at longjmp.

k0kubun (approved):

XrXr:
I guess if the theory about rb_raise() inside checked ccall holds, this won't fix all the cases. The ones that don't involve fallbacks like https://github.com/ruby/ruby/pull/10136#discussion_r1506735809 might still fail. This PR should reduce the frequency and give us some new info when it fails, though.

XrXr:
https://github.com/ruby/ruby/pull/10267 should fix the cases that are left 🤞🏼

PR #10141, opened 2024-02-29
Use configured `LD` for linking enc and ext libraries

kateinoigakukun:
"AR" was well propagated to the enc.mk and mkmf, but "LD" was not. This caused the dynamic libraries to be linked with a linker found in the PATH, which could be different from the one used in the Ruby build process. This is especially important for cross-compilation, where the host linker may not be compatible with the target system. (e.g. WebAssembly)


nobu (commented):
`LD` is to link executable files, not shared libraries.
What do you want to make with `LD` there?

PR #10143, opened 2024-02-29
Module#remove_const: cleanup vm->defined_module_hash

casperisfine:
[[Bug #20311]](https://bugs.ruby-lang.org/issues/20311)

All module and classes defined from C are stored in this table to ensure they are pinned and won't be moved by the compactor.

However this means we need to remove them from it whenever a constant referencing them is removed.

But since a same class or module may be referenced under multiple names, we have to do some ref counting.

### DIFF
--- a/array.c
+++ b/array.c
@@ -8577,7 +8577,8 @@ rb_ary_deconstruct(VALUE ary)
 void
 Init_Array(void)
 {
-    rb_cArray  = rb_define_class("Array", rb_cObject);
+    rb_global_variable(&rb_cArray);
+    rb_cArray = rb_define_class("Array", rb_cObject);
     rb_include_module(rb_cArray, rb_mEnumerable);
 
     rb_define_alloc_func(rb_cArray, empty_ary_alloc);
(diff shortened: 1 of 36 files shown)


casperisfine:
looks like I introduced a bug that cause some classes not to be pinned. Trying to figure it out.

casperisfine:
Alright, so the initial patch would break because some code in Ruby rely on this bug. e.g. `rb_cNilClass` ins't declared and some test unreference `::NilClass` causing it to be GCed. That isn't to hard to fix.

But that also means removing modules from the root set can't be done when they are de-referenced, it must be done when they are GCed. Currently my patch does that, but now the problem is that a lot of core, including inside Ruby itself rely on classes and module define from C being immortal.

For the most part it's fine, but it means if their const is removed it then become trivial to crash the VM. So I'd need to add a whole lot of `rb_global_variable` in tons of places.

casperisfine:
Closing in favor of https://github.com/ruby/ruby/pull/10144

PR #10144, opened 2024-02-29
Don't root named structs defined in Ruby

casperisfine:
[[Bug #20311]](https://bugs.ruby-lang.org/issues/20311)

`rb_define_class_under` assumes it's called from C and that the reference might be held in a C global variable, so it adds the class to the VM root.

In the case of `Struct.new('Name')` it's wasteful and make the struct immortal.

### DIFF
--- a/class.c
+++ b/class.c
@@ -1007,7 +1007,7 @@ rb_define_class_under(VALUE outer, const char *name, VALUE super)
 }
 
 VALUE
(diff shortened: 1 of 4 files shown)


casperisfine:
Oh...

```
Object.__send__(:remove_const, :NilClass)
                            GC.start
                            nil.inspect
```

casperisfine:
Actually, a better patch could just be to unpin when GCed.

PR #10146, opened 2024-02-29
Clarify C API documentation about pinned classes

casperisfine:
They are not only pinned, but also immortal. Even if the constant referencing them is removed, they will remain alive.

It's a precision worth noting.

cc @peterzhu2118

### DIFF
--- a/include/ruby/internal/intern/class.h
+++ b/include/ruby/internal/intern/class.h
@@ -88,8 +88,8 @@ VALUE rb_define_class_id(ID id, VALUE super);
  * @post        `outer::id` refers the returned class.
  * @note        If a class named `id` is  already defined and its superclass is
  *              `super`, the function just returns the defined class.
- * @note        The  compaction  GC does  not  move  classes returned  by  this
- *              function.
+ * @note        The GC does not collect nor move classes returned by this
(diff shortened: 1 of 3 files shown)


include/ruby/internal/intern/class.h
```diff
@@ -88,8 +88,8 @@ VALUE rb_define_class_id(ID id, VALUE super);
  * @post        `outer::id` refers the returned class.
  * @note        If a class named `id` is  already defined and its superclass is
  *              `super`, the function just returns the defined class.
- * @note        The  compaction  GC does  not  move  classes returned  by  this
- *              function.
+ * @note        The GC does not collect nor move move classes returned by this
```
peterzhu2118, line 91: There's a duplicated "move" everywhere.

```suggestion
 * @note        The GC does not collect nor move classes returned by this
```

casperisfine: 🤦 .

Corrected.

peterzhu2118: I don't see it corrected in this file.

casperisfine: Yeah, I'm an idiot, should be good now.

PR #10149, opened 2024-02-29
Add POSTLINK step to Makefile generated by mkmf.rb

kateinoigakukun:
`POSTLINK` must be applied to dynamic extension library as well as how `libruby.so` is built in the main build process. This is especially crucial for Wasm/WASI, where `POSTLINK` is used to run `wasm-opt` to apply mandatory Asyncify pass.

### DIFF
--- a/lib/mkmf.rb
+++ b/lib/mkmf.rb
@@ -2148,6 +2148,7 @@ ARCH_FLAG = #{$ARCH_FLAG}
 DLDFLAGS = $(ldflags) $(dldflags) $(ARCH_FLAG)
 LDSHARED = #{CONFIG['LDSHARED']}
 LDSHAREDXX = #{config_string('LDSHAREDXX') || '$(LDSHARED)'}
+POSTLINK = #{config_string('POSTLINK', RbConfig::CONFIG)}
 AR = #{CONFIG['AR']}
 LD = #{CONFIG['LD']}
 EXEEXT = #{CONFIG['EXEEXT']}
--- a/configure.ac
+++ b/configure.ac
@@ -1107,10 +1107,6 @@ main()
 		AS_IF([test -n "$dsymutil"], [
 		    POSTLINK="$dsymutil \$@ 2>/dev/null${POSTLINK:+; $POSTLINK}"
 		])
-		AS_IF([test -n "${POSTLINK}"], [
-		    LINK_SO="$LINK_SO
-\$(POSTLINK)"
-		])
 		AC_CHECK_HEADERS(crt_externs.h, [], [], [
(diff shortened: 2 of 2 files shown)


lib/mkmf.rb
```diff
@@ -2148,6 +2148,7 @@ def configuration(srcdir)
 DLDFLAGS = $(ldflags) $(dldflags) $(ARCH_FLAG)
 LDSHARED = #{CONFIG['LDSHARED']}
 LDSHAREDXX = #{config_string('LDSHAREDXX') || '$(LDSHARED)'}
+POSTLINK = #{RbConfig.expand(CONFIG['POSTLINK'], CONFIG)}
```
kateinoigakukun, line 2151: I used `RbConfig.expand` here to avoid carrying all possible referenced variables like `WASMOPT`, `wasmoptflags`, and so on.

lib/mkmf.rb
```diff
@@ -2148,6 +2148,7 @@ def configuration(srcdir)
 DLDFLAGS = $(ldflags) $(dldflags) $(ARCH_FLAG)
 LDSHARED = #{CONFIG['LDSHARED']}
 LDSHAREDXX = #{config_string('LDSHAREDXX') || '$(LDSHARED)'}
+POSTLINK = #{RbConfig.expand('$(POSTLINK)', CONFIG)}
```
nobu, line 2151: You should be able to use `RbConfig::CONFIG` instead of `CONFIG` (= `RbConfig::MAKEFILE_CONFIG`).

```suggestion
POSTLINK = #{config_string('POSTLINK', RbConfig::CONFIG)}
```

lib/mkmf.rb
```diff
@@ -2633,7 +2634,8 @@ def create_makefile(target, srcprefix = nil)
     if srcs.any?(&%r"\.(?:#{CXX_EXT.join('|')})\z".method(:===))
       link_so = link_so.sub(/\bLDSHARED\b/, '\&XX')
     end
-    mfile.print link_so, "\n\n"
+    mfile.print link_so, "\n"
+    mfile.print "\t$(Q) $(POSTLINK)\n\n"
```
nobu, line 2638: Also you can append `$(POSTLINK)` to `LINK_SO` in configure.ac.

kateinoigakukun: Ah, I didn't notice `LINK_SO` can be multiple commands. Looks like we should generalize the way darwin platforms do here. https://github.com/ruby/ruby/blob/a465393cb9b6012f01a98f3da861c92be54445e7/configure.ac#L1110-L1113

nobu (approved):

PR #10152, opened 2024-03-01
YJIT: Support opt_invokebuiltin_delegate for leaf builtin

k0kubun:
Methods like `Kernel#class` have an ISEQ like `opt_invokebuiltin_delegate_leave` + `leave` in local environments. On SFR and Core, however, it turns into `opt_invokebuiltin_delegate` + `leave` for some reason. While we're still investigating why it happens, these ISEQs work in the same way whether we fix it or not.

This PR lets `rb_yjit_builtin_function` extract a builtin function whether the first instruction is `opt_invokebuiltin_delegate_leave` or `opt_invokebuiltin_delegate`.

On SFR, `Kernel#class` is 2.0% of all ISEQ calls, `Symbol#to_s` is 1.1%, `Integer#zero?` is 0.6%, and `Array#last` is 0.5%. So this PR should inline 4.2% or more ISEQ calls on SFR.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1600,6 +1600,19 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_leaf_builtin
+    assert_compiles(code_gc_helpers + <<~'RUBY', exits: :any, result: 1)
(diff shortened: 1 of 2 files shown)


XrXr (approved):
By the way, it's from tracepoint and 8070cb56db1 (instead of being changed to a `trace_` variant, it becomes `opt_invokebuiltin_delegate`)

PR #10169, opened 2024-03-02
[Bug #20322] Fix rb_enc_interned_str_cstr null encoding

thomasmarshall:
https://bugs.ruby-lang.org/issues/20322

While adding support for `rb_enc_interned_str_cstr` [in TruffleRuby](https://github.com/oracle/truffleruby/pull/3427), we noticed the documentation says that `enc` can be a null pointer: https://github.com/ruby/ruby/blob/93556d46203545bc2364b1c0dd1281ba098f3cc9/include/ruby/internal/encoding/string.h#L120

However, this currently causes a segmentation fault when it tries to [autoload the encoding](https://github.com/ruby/ruby/blob/93556d46203545bc2364b1c0dd1281ba098f3cc9/string.c#L12125-L12127) because it calls `rb_enc_mbmaxlen` which [expects a non-null encoding](https://github.com/ruby/ruby/blob/93556d46203545bc2364b1c0dd1281ba098f3cc9/include/ruby/internal/encoding/encoding.h#L448). This commit fixes the issue by checking for NULL before calling `rb_enc_autoload`.

I'm not sure how necessary the fix is, perhaps nobody has encountered this issue. Alternatively we could just remove the note about null pointers from the docs.

### DIFF
--- a/spec/ruby/optional/capi/string_spec.rb
+++ b/spec/ruby/optional/capi/string_spec.rb
@@ -1236,6 +1236,14 @@ end
     it "returns the same string as String#-@" do
       @s.rb_enc_interned_str_cstr("hello", Encoding::UTF_8).should.equal?(-"hello")
     end
+
+    ruby_version_is "3.4" do
+      it "uses the default encoding if encoding is null" do
+        str = "hello"
+        val = @s.rb_enc_interned_str_cstr(str, nil)
+        val.encoding.should == Encoding::ASCII_8BIT
+      end
+    end
   end
 
   describe "rb_str_to_interned_str" do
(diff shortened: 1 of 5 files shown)


spec/ruby/optional/capi/string_spec.rb
```diff
@@ -1236,6 +1236,14 @@ def inspect
     it "returns the same string as String#-@" do
       @s.rb_enc_interned_str_cstr("hello", Encoding::UTF_8).should.equal?(-"hello")
     end
+
+    ruby_version_is "3.4" do
```
nobu, line 1240: Since https://bugs.ruby-lang.org/issues/18772 has been backported, this fix should be backported too.
```suggestion
    ruby_bug "#20322", ""..."3.4" do
```

nobu (changes requested):

nobu (approved):

PR #10171, opened 2024-03-03
Refactor VM root modules out

casperisfine:
This `st_table` is used to both mark and pin classes defined from the C API. But `vm->mark_object_ary` already does both much more efficiently.

Currently a Ruby process starts with 252 rooted classes, which uses `7224B` in an `st_table` or `2016B` in an `RArray`.

So a baseline of 5kB saved, but since `mark_object_ary` is preallocated with `1024` slots but only use `405` of them, it's a net `7kB` save.

### DIFF
--- a/class.c
+++ b/class.c
@@ -810,7 +810,7 @@ boot_defclass(const char *name, VALUE super)
     ID id = rb_intern(name);
 
     rb_const_set((rb_cObject ? rb_cObject : obj), id, obj);
-    rb_vm_add_root_module(obj);
+    rb_gc_register_mark_object(obj);
     return obj;
 }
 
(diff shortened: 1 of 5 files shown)


casperisfine:
cc @ko1 is there a reason I'm overlooking why `vm->defined_module_hash` exist alongside `vm->mark_object_ary`?

The only difference I can see if that with `vm->defined_module_hash` you can root the same Module/Class multiple times and it won't lead to duplicated entries. But I can't really imagine extensions calling `rb_define_class*` more than a couple times for the same class.

casperisfine:
A followup I'd like to make is to change `mark_object_ary` to no longer be an `Array` but a custom type, so that we don't have to mark all of its content on every minor.

PR #10178, opened 2024-03-05
Some edits to variable.c

fxn:
Most edits are self-explanatory.

The elimination of the unnecessary `else` branch follows existing style for early returns elsewhere in the file. For example:

```c
if (p >= pend || !*p) {
    return false;
}
```

in `is_constant_path`.

In `const_missing`, I added a paragraph to explicitly explain the callback is not required to define the constant. This was something I wondered when I started studying this topic back in the day.

The autoloading example was not right, because if `const_get` does not find the constant, what happens is that you get a recursive call (that is halted by the hash table).


nobu:
> The elimination of the unnecessary `else` branch follows existing style for early returns elsewhere in the file.

I think it's better if that change, other than document, was a separate commit.

> The autoloading example was not right, because if `const_get` does not find the constant, what happens is that you get a recursive call (that is halted by the hash table).

"Not right" means that the code after `const_get(name)` will be unreachable?

fxn:
@nobu sure!

I am reading _variable.c_ from top to bottom and was just editing things as I saw them. Since some projects squash on merge and it was just a bunch of small edits, I did not bother to split the changes. I have done it now.

> "Not right" means that the code after const_get(name) will be unreachable?

Kind of.

That code is written as if the value returned by `const_get` acted as a flag to indicate the success of the call. But the returned value is irrelevant, really:

1. If the constant could not be loaded, an exception is raised, so not even the assignment is executed.
2. If the constant could be loaded, the user should get its stored value regardless of its boolean meaning.

There's still an additional fine point in that the `const_get` call should prevent looking the ancestor chain up, because you could argue the file is expected to define the constant in the receiver (not in an ancestor of the receiver). Though this a debatrable detail, because it is not specified. I only find it be what you'd expect from that example. The main points are the ones above, anyway.

PR #10179, opened 2024-03-05
Refactor `vm->mark_object_ary` to be a linked list of WB protected objects

casperisfine:
Fix: https://github.com/ruby/ruby/pull/10171

This `st_table` is used to both mark and pin classes defined from the C API. But `vm->mark_object_ary` already does both much more efficiently.

Currently a Ruby process starts with 252 rooted classes, which uses `7224B` in an `st_table` or `2016B` in an `RArray`.

So a baseline of 5kB saved, but since `mark_object_ary` is preallocated with `1024` slots but only use `405` of them, it's a net `7kB` save.

`vm->mark_object_ary` is also being refactored.

Prior to this changes, `mark_object_ary` was a regular `RArray`, but since this allows for references to be moved, it was marked a second time from `rb_vm_mark()` to pin these objects.

This has the detrimental effect of marking these references on every minors even though it's a mostly append only list.

But using a custom TypedData we can save from having to mark all the references on minor GC runs.

Addtionally, immediate values are now ignored and not appended to `vm->mark_object_ary` as it's just wasted space.

cc @peterzhu2118

### DIFF
--- a/bignum.c
+++ b/bignum.c
@@ -4728,7 +4728,7 @@ power_cache_get_power(int base, int power_level, size_t *numdigits_ret)
         rb_obj_hide(power);
         base36_power_cache[base - 2][power_level] = power;
         base36_numdigits_cache[base - 2][power_level] = numdigits;
-        rb_gc_register_mark_object(power);
+        rb_vm_register_global_object(power);
     }
     if (numdigits_ret)
         *numdigits_ret = base36_numdigits_cache[base - 2][power_level];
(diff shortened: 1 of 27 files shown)


casperisfine:
> So a baseline of 5kB saved, but since mark_object_ary is preallocated with 1024 slots but only use 405 of them, it's a net 7kB save.

I checked the number on a large application.`defined_module_hash` has `1k` entires for a memsize of `28728B`, and `mark_object_ary` `3242` entries. So there the saving is rougthly `20KB`, and saves marking `4.2k` objects on every minors.

peterzhu2118 (approved):

PR #10180, opened 2024-03-05
.travis.yml: Allow failures for arm64 and arm32 cases temporarily.

junaruga:
It's to avoid the failures by the infra issues in the term of the following maintenance.

There is a planned maintenance on March 6th 2024 between 8:00 and 12:00 UTC+0. The Travis service may be temporarily unavailable. https://bugs.ruby-lang.org/issues/20013#note-28

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,11 +113,11 @@ matrix:
     - <<: *arm32-linux
   allow_failures:
     # Allow failures for the unstable jobs.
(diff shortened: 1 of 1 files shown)


junaruga:
I confirmed Travis was running with the allow_failure for all the CPU cases.
https://app.travis-ci.com/github/ruby/ruby/builds/269289300

PR #10181, opened 2024-03-06
Load iseq hidden variables as 0

jeremyevans:
This fixes test failures when running tests with
RUBY_ISEQ_DUMP_DEBUG=to_binary.


compile.c
```diff
@@ -11863,9 +11863,9 @@ ibf_load_id(const struct ibf_load *load, const ID id_index)
         return 0;
     }
     VALUE sym = ibf_load_object(load, id_index);
-    if (rb_type_p(sym, T_FIXNUM)) {
+    if (FIXNUM_P(sym) || RB_BIGNUM_TYPE_P(sym)) {
```
nobu, line 11866:
```suggestion
    if (rb_integer_type_p(sym)) {
```

PR #10184, opened 2024-03-06
WIP: Run parallel testing with split-subset in Launchable

ono-max:


.github/workflows/macos.yml
```diff
@@ -39,9 +51,56 @@ jobs:
...
+        if: ${{ matrix.test_task == 'check' || matrix.test_task == 'test-all' }}
+      - name: Set up Launchable
+        id: enable-launchable
+        uses: ./.github/actions/launchable/setup
+        with:
+          launchable-token: ${{ secrets.LAUNCHABLE_TOKEN }}
+          srcdir: src
+        if: ${{ matrix.test_task == 'check' || matrix.test_task == 'test-all' }}
+      - run: |
+          ruby -e 'File.write("launchable-subset-input.txt", Dir["test/**/*_test.rb", "test/**/test_*.rb"].map {|file| "file=#{file}"}.join("\n"))'
```
nobu, line 79: `find` command is simpler for a simple task like this.

```suggestion
          find test -name "*_test.rb" -o -name "test_*.rb" -printf "file=%P\n" > launchable-subset-input.txt
```

.github/workflows/macos.yml
```diff
@@ -86,12 +143,116 @@ jobs:
...
+        if: ${{ always() && (matrix.test_task == 'check' || matrix.test_task == 'test-all') }}
+      - uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+        if: ${{ always() && (matrix.test_task == 'check' || matrix.test_task == 'test-all') }}
+      - name: Install launchable command
+        run: |
+          pip install launchable
+          launchable split-subset --subset-id $(cat ../launchable-subset-id.txt) --bin 1/4 raw > launchable-subset.txt
+          echo "TESTS=$(cat launchable-subset.txt | tr '\n' ' ' | sed 's/file=/..\/src\//g')" >> $GITHUB_ENV
```
nobu, line 158: `sed` applies the command for each line, and `$(...)` replaces `\n` with a space like `tr` does here.
```suggestion
          echo "TESTS=$(cat launchable-subset.txt | sed 's:file=:../src/:')" >> $GITHUB_ENV
```

nobu (commented):
It doesn't feel nice to repeat all things four times.
And `skipped_tests` won't be repeated every time?

ono-max:
@nobu 

Thank you for reviewing this PR. However, I'm going to close this PR because I don't feel that this solution is optimal. I apologize for any confusion.

PR #10185, opened 2024-03-06
Document exception thrown by `Time.strptime`

artur-intech:


nobu:
https://github.com/ruby/time is the upstream of `lib/time.rb`.

PR #10196, opened 2024-03-07
Use ubuntu-latest

nobu:
Ubuntu 20.04 is nearly out of date.


hsbt:
Fixed at https://github.com/ruby/ruby/pull/10202

PR #10199, opened 2024-03-07
[DOC] Array doc

BurdetteLamar:
Adds section 'In brief' for Array#[]=.

### DIFF
--- a/array.c
+++ b/array.c
@@ -1733,6 +1733,16 @@ static VALUE rb_ary_aref2(VALUE ary, VALUE b, VALUE e);
  *
  *  Returns elements from +self+; does not modify +self+.
  *
+ *  In brief:
+ *
+ *    a = [:foo, 'bar', 2]
+ *    a[0]     # => :foo
+ *    a[-1]    # => 2
+ *    a[1, 2]  # => ["bar", 2]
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
There may be too many examples in the new section;  suggest deletions please.

BurdetteLamar:
I think the doc for Array#[]= may be missing the cases where the operand is an array;  if so, that will be for a different PR.

array.c
```diff
@@ -2284,6 +2294,73 @@ ary_aset_by_rb_ary_splice(VALUE ary, long beg, long len, VALUE val)
  *
  *  Assigns elements in +self+; returns the given +object+.
  *
+ *  In brief:
```
peterzhu2118, line 2297: IMO the "in brief" section stops being useful when it tries to exhaustively cover all the cases. It becomes harder to read than the regular documentation.

BurdetteLamar: Agree;  pared.

BurdetteLamar:
@peterzhu2118, will you be reviewing this?  I've pared down the examples (as I suggested in the OP might be necessary).

peterzhu2118 (approved):

peterzhu2118:
Sorry, I must have missed this.

PR #10201, opened 2024-03-07
Disregard si_addr for fatal signals sent from other processes

peterzhu2118:
Previously, when another process sends a fatal signals such as SIGBUS to Ruby, we could mis-interpret it as a stack overflow Ruby itself generated. When the si_pid field is set on the signal, we shouldn't check the si_addr field to check for stack overflow.

> Signals sent with kill(2) and sigqueue(3) fill in si_pid and si_uid.

### DIFF
--- a/signal.c
+++ b/signal.c
@@ -872,9 +872,9 @@ check_stack_overflow(int sig, const void *addr)
 # else
 #   define FAULT_ADDRESS info->si_addr
 #   ifdef USE_UCONTEXT_REG
-#     define CHECK_STACK_OVERFLOW() check_stack_overflow(sig, (uintptr_t)FAULT_ADDRESS, ctx)
+#     define CHECK_STACK_OVERFLOW() (info->si_pid ? (void)0 : check_stack_overflow(sig, (uintptr_t)FAULT_ADDRESS, ctx))
 #   else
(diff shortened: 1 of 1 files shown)


signal.c
```diff
@@ -872,9 +872,9 @@ check_stack_overflow(int sig, const void *addr)
 # else
 #   define FAULT_ADDRESS info->si_addr
 #   ifdef USE_UCONTEXT_REG
-#     define CHECK_STACK_OVERFLOW() check_stack_overflow(sig, (uintptr_t)FAULT_ADDRESS, ctx)
+#     define CHECK_STACK_OVERFLOW() if (!info->si_pid) check_stack_overflow(sig, (uintptr_t)FAULT_ADDRESS, ctx)
```
nobu, line 875: Could you keep it an expression?
```suggestion
#     define CHECK_STACK_OVERFLOW() (info->si_pid ? (void)0 : check_stack_overflow(sig, (uintptr_t)FAULT_ADDRESS, ctx))
```

nobu:
I guess you are using macOS or Linux because you mentioned SIGBUS.

However, on Linux, `siginfo_t` uses a union and the field corresponding to `si_pid` does not belong to the `_sigfault` field for SIGSEGV and SIGBUS.
It actually overlaps the `si_addr` field, which is usually non-zero on stack overflow.

In short, the stack overflow check does not work on Linux since this PR.

PR #10207, opened 2024-03-10
Report to Launchable at post-run

nobu:
Separate two actions don't feel nice to me.
It seems fine to just make the result report at last.


.github/workflows/macos.yml
```diff
@@ -91,7 +91,10 @@ jobs:
         uses: ./.github/actions/launchable/setup
         with:
           launchable-token: ${{ secrets.LAUNCHABLE_TOKEN }}
+          os: ${{ matrix.os }}
+          test-opts: ${{ matrix.test_opts }}
           srcdir: src
+          builddir: build
```
junaruga, line 97: The order of the options is os, test-opts, launchable-token, builddir, srcdir. We can align the order here, also to align with the one in the `ubuntu.yml`.

.github/workflows/ubuntu.yml
```diff
@@ -107,8 +107,11 @@ jobs:
         id: enable-launchable
         uses: ./.github/actions/launchable/setup
         with:
+          os: ubuntu-20.04
+          test-opts: ${{ matrix.configure }}
           launchable-token: ${{ secrets.LAUNCHABLE_TOKEN }}
           srcdir: src
+          builddir: build
```
junaruga, line 114: The order of the options is os, test-opts, launchable-token, builddir, srcdir. We can swap the builddir with srcdir here.

.github/actions/launchable/setup/action.yml
```diff
@@ -90,3 +108,42 @@ runs:
...
+        report_path="${dir}${builddir:+${builddir}/}${report_path}"
+        echo report-path="${report_path}" >> $GITHUB_OUTPUT
+      if: steps.enable-launchable.outputs.enable-launchable
+      env:
+        srcdir: ${{ inputs.srcdir }}
+        builddir: ${{ inputs.builddir }}
+        report_path: ${{ inputs.report-path }}
+
+    - name: Record test results in Launchable
+      uses: gacts/run-and-post-run@674528335da98a7afc80915ff2b4b860a0b3553a
```
ono-max, line 138: nit
```suggestion
      uses: gacts/run-and-post-run@674528335da98a7afc80915ff2b4b860a0b3553a # v1.4.0
```

PR #10235, opened 2024-03-13
Make the test suite is compatible with --enable-frozen-string-literal

casperisfine:
As preparation for https://bugs.ruby-lang.org/issues/20205 making sure the test suite is compatible with frozen string literals is making things easier.

Note that `spec/` is being handled upstream: https://github.com/ruby/spec/pull/1142

### DIFF
--- a/basictest/test.rb
+++ b/basictest/test.rb
@@ -879,7 +879,7 @@ $x.sort!{|a,b| b-a}		# reverse sort
 test_ok($x == [7,5,3,2,1])
 
 # split test
-$x = "The Book of Mormon"
+$x = +"The Book of Mormon"
 test_ok($x.split(//).reverse!.join == $x.reverse)
 test_ok($x.reverse == $x.reverse!)
 test_ok("1 byte string".split(//).reverse.join(":") == "g:n:i:r:t:s: :e:t:y:b: :1")
@@ -1643,7 +1643,7 @@ test_ok(/^(?:ab+)+/ =~ "ababb" && $& == "ababb")
 test_ok(/(\s+\d+){2}/ =~ " 1 2" && $& == " 1 2")
 test_ok(/(?:\s+\d+){2}/ =~ " 1 2" && $& == " 1 2")
 
-$x = <<END;
+$x = +<<END;
 ABCD
 ABCD
 END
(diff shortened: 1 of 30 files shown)


spec/mspec/lib/mspec/helpers/ruby_exe.rb
```diff
@@ -132,7 +132,7 @@ def ruby_exe(code = :not_given, opts = {})
 
   escape = opts.delete(:escape)
   if code and !File.exist?(code) and escape != false
-    tmpfile = tmp("rubyexe.rb")
+    tmpfile = tmp(+"rubyexe.rb")
```
eregon, line 135: This feels like the `tmp` helper should use `+` or `dup`, so not every caller needs to change.

spec/ruby/core/complex/inspect_spec.rb
```diff
@@ -17,18 +17,18 @@
 
   it "calls #inspect on real and imaginary" do
     real = NumericSpecs::Subclass.new
-    real.should_receive(:inspect).and_return("1")
+    real.should_receive(:inspect).and_return(+"1")
```
eregon, line 20: It feels unexpected this String is mutated, would you know where it's mutated?

spec/mspec/lib/mspec/helpers/ruby_exe.rb
casperisfine: Problem is it mutates the argument, and some callers expect that. So raising on frozen makes it easier to find the broken ones.

spec/ruby/core/complex/inspect_spec.rb
casperisfine: Sorry, I have thousands of these to update so I'm not digging too much in every single one.

spec/mspec/lib/mspec/helpers/ruby_exe.rb
eregon: > Problem is it mutates the argument, and some callers expect that

We should fix those callers then.
I can try to take a look this week.
It feels definitely wrong that a helper method mutates its argument.

The practical problem is it's impossible to review the diff with all the `tmp(+` changes.

casperisfine: As discussed on slack, in the end there was mostly one impacted callsite. So I reverted all that and just changed that one callsite a bit.

maximecb (approved):

PR #10244, opened 2024-03-13
Skip Bootsnap and Zeitwerk in bundled gems warning

fxn:
Bootsnap and Zeitwerk decorate `Kernel#require`.

The current logic in `bundled_gems.rb` produces warnings like this:

> Also contact author of zeitwerk-2.6.12 to add csv into its gemspec

which are not useful, since that `require` is just a proxy, the real one to be warned about is upwards. There have been several issues created in the Zeitwerk project due to this.

In this patch I propose to iterate over `caller_locations`, skipping references to `bootsnap` and `zeitwerk`. Hopefully, with this change, the warnings will point to the actual gem issuing the `require` call.

DISCLAIMER: I cannot build Ruby right now, I get

```
% ./autogen.sh      
Argument "/opt/homebrew/Cellar/autoconf/2.72/share/autoconf/build-..." isn't numeric in numeric ne (!=) at /opt/homebrew/bin/autoreconf line 383.
```

and was not able to test the patch, submitting optimistically!

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -109,12 +109,7 @@ module Gem::BUNDLED_GEMS
     else
       return
     end
-    # Warning feature is not working correctly with Bootsnap.
-    # caller_locations returns:
-    #   lib/ruby/3.3.0+0/bundled_gems.rb:65:in `block (2 levels) in replace_require'
-    #   $GEM_HOME/gems/bootsnap-1.17.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'"
-    #   ...
-    return if caller_locations(2).find {|c| c&.path.match?(/bootsnap/) }
+
     return if WARNED[name]
     WARNED[name] = true
     if gem == true
@@ -134,11 +129,29 @@ module Gem::BUNDLED_GEMS
 
     if defined?(Bundler)
       msg += " Add #{gem} to your Gemfile or gemspec."
+
       # We detect the gem name from caller_locations. We need to skip 2 frames like:
       # lib/ruby/3.3.0+0/bundled_gems.rb:90:in `warning?'",
       # lib/ruby/3.3.0+0/bundler/rubygems_integration.rb:247:in `block (2 levels) in replace_require'",
-      location = caller_locations(3,1)[0]&.path
-      if File.file?(location) && !location.start_with?(Gem::BUNDLED_GEMS::LIBDIR)
+      #
+      # Additionally, we need to skip Bootsnap and Zeitwerk if present, these
+      # gems decorate Kernel#require, so they are not really the ones issuing
+      # the require call users should be warned about. Those are upwards.
+      frames_to_skip = 2
+      location = nil
+      Thread.each_caller_location do |cl|
+        if frames_to_skip >= 1
+          frames_to_skip -= 1
+          next
+        end
+
+        unless cl.path.match?(/bootsnap|zeitwerk/)
+          location = cl.path
+          break
+        end
+      end
+
+      if location && File.file?(location) && !location.start_with?(Gem::BUNDLED_GEMS::LIBDIR)
         caller_gem = nil
         Gem.path.each do |path|
           if location =~ %r{#{path}/gems/([\w\-\.]+)}


lib/bundled_gems.rb
```diff
@@ -134,11 +129,29 @@ def self.build_message(gem)
...
+      # the require call users should be warned about. Those are upwards.
+      frames_to_skip = 2
+      location = nil
+      Thread.each_caller_location do |cl|
+        if frames_to_skip >= 1
+          frames_to_skip -= 1
+          next
+        end
+
+        unless cl.path.match?(/bootsnap|zeitwerk/)
```
casperisfine, line 148:
```suggestion
        unless cl.path.match?(%r{/(bootsnap|zeitwerk)/})
```

casperisfine:
cc @hsbt would you be OK with this? I'm watching the Zeitwerk repository, and there was at least 5 issues filed blaming Zeitwerk for that warning in the last couple weeks. If that's OK too, it would be best to backport this to 3.3.

jeremyevans:
We should not be adding exceptions for external libraries to core Ruby, regardless of how popular those libraries may be.  If we want to support something like this, we should provide a way so that any library can register themselves as opting out of this, and both bootsnap and zeitwerk can use that (as well as any other libraries that do similar things).

fxn:
@jeremyevans note that Bootsnap was already special-cased.

casperisfine:
> We should not be adding exceptions for external libraries to core Ruby

It does indeed feels wrong, but given `bootsnap` was already there it seems logical. 

> we should provide a way so that any library can register themselves as opting out of this

Make sense. But I wonder if instead we could ignore all the frames until we find one that isn't for the `require` method. Not 100% sure it would cover both bootsnap and zeitwerk, but worth a try.

jeremyevans:
> @jeremyevans note that Bootsnap was already special-cased.

Thanks for pointing that out. I'm strongly against that as it leads to a "some libraries are more equal than others" situation.  If we are considering backporting this to 3.3, which should consider instead backporting an API so that any library can get the same treatment.

casperisfine:
> I wonder if instead we could ignore all the frames until we find one that isn't for the `require` method.

Looking at both methods:

  - https://github.com/fxn/zeitwerk/blob/0b6477fafda528f01e88a7529128e1f6f955d6ab/lib/zeitwerk/kernel.rb#L23-L43
  - https://github.com/Shopify/bootsnap/blob/5d37046710f782cd1cc4bf6ac3d92c76f6d41fa4/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb#L8-L34

I think it would work. Neither call any methods that take a block, so all the frame should have `"require"` as `#base_label`. So we could walk up the stack until we find one frame not called `require`.

How does that sound?

casperisfine:
Actually, I'm having doubts now because the methods are aliased, It's late so can't test right now, but I suspect it might not work 😞

fxn:
Yeah, I understand the switch from hard-coding to an API.

But let's imagine, we are proposing to have a method so that gems can say: In that warning that you issue because people are loading this directly, please skip me while walking caller locations up.

I don't know, there is something off for me in that feature. I mean, from a design perspective, I understand. But I look at that sentence, and it just does not feel right to me. It is like plugging into something very implementation-dependent and kind of internal, to me.

casperisfine:
> Actually, I'm having doubts now because the methods are aliased, It's late so can't test right now, but I suspect it might not work 😞

Seems like it would work:

```ruby
module M
  extend self

  def print_callers
    p caller_locations.map(&:base_label)
  end

  def foo
    print_callers
  end

  alias_method :original_foo, :foo

  def foo
    original_foo
  end

  def bar
    foo
  end
end

M.bar
```

```ruby
["foo", "foo", "bar", "<main>"]
```

hsbt:
@fxn Thank you for submitting this PR and handling incorrect warning by my rough implementation like https://github.com/fxn/zeitwerk/issues/289.

I merged your commit at https://github.com/ruby/ruby/pull/10347.

@jeremyevans I agreed your opinion. Do you have an idea to avoid to specify gem name directly?

I would like to fix this for Ruby 3.3.1. We can improve this 3.3.2 and 3.4.0.

jeremyevans:
@hsbt If the approach given by @casperisfine would not work, then an API that allows gems to register their name as one to skip, and having bootsnap and zeitwerk call it if it exists.  That avoids hard coding the names, and allows other gems the ability to get the same behavior.

PR #10245, opened 2024-03-14
Freeze singleton class chain

jeremyevans:
The following code:

```ruby
x = Object.new
sc1 = x.singleton_class
sc2 = sc1.singleton_class
x.freeze
```

Would freeze sc1 but not sc2, even though sc1 would be frozen.

Not freezing the entire singleton class chain is expected, because at a certain point, the singleton class is shared.  However, you can walk the the attached object chain for each singleton class, and only freeze if it results in the original object.  This way, only non-shared singleton classes are frozen.

Fixes [Bug #20319]


jeremyevans:
I came up with a revised algorithm for this, which is O(N) instead of O(N^2), where N is the depth of the singleton class chain. Instead of going all the way down the chain back to the original frozen object in every level, we only need to check at each level that the object's class pointer is the singleton class for the current object, by checking the attached object for that class matches.

Previously, I wasn't sure about whether the code was correct in all cases. I'm now fairly confident that it is. I've expanded the tests as well.

@nobu when you have time, could you please rereview?

PR #10246, opened 2024-03-14
YJIT: Let yjit_perf.py support perf with Python disabled

k0kubun:
Many Linux distributions don't enable Python scripting in `perf` by default. `perf` in our production environment didn't have that either. We could change our production image to manually build `perf` to fix the problem too, but it's a lot of work to do it everywhere. 

This PR proposes to add an alternative way to run our perf script. It adds an extra interface to parse the output of `perf script` to obviate Python scripting support in production environments. With this PR, we can use this script like this:

```bash
production$ perf script > /tmp/perf.txt
local$ scp production:/tmp/perf.txt /tmp/perf.txt
local$ misc/yjit_perf.py /tmp/perf.txt
```


maximecb:
Can you explain a bit more how this works? With this script, you still need the `perf` tool to be available? This Python script parses the output of the `perf` program?

Can you update the documentation in `yjit.md` to show the complete perf workflow, including invoking YJIT to gather perf data?

doc/yjit/yjit.md
```diff
@@ -499,3 +504,28 @@ PERF=record ruby --yjit-perf=codegen -Iharness-perf benchmarks/lobsters/benchmar
...
+echo 0 | sudo tee /proc/sys/kernel/kptr_restrict
+echo -1 | sudo tee /proc/sys/kernel/perf_event_paranoid
+
+# Profile Ruby with --yjit-perf=codegen
+cd ../yjit-bench
+PERF=record ruby --yjit-perf=codegen -Iharness-perf benchmarks/lobsters/benchmark.rb
+
+# Aggregate results
+perf script > /tmp/perf.txt
+../ruby/misc/yjit_perf.py /tmp/perf.txt
```
maximecb, line 530: Assuming that this works just as well, maybe we should only give the instructions assuming people are using perf without Python support. Less confusing if we only provide one workflow that works everywhere?

maximecb (approved):

k0kubun:
> Can you explain a bit more how this works? With this script, you still need the perf tool to be available? 

The point is that we can separate the machine to be profiled and the one to aggregate the results can be separated. The profiling target machine still needs perf tool. But it no longer requires perf with Python scripting (because Python is used only for aggregation). 

> This Python script parses the output of the perf program?

Correct. The example output of `perf script` command is left as a comment in `misc/yjit_perf.py`. We only need a symbol name, DSO name, and the number of cycles, and the Python scripts parses the information from each line.

> Can you update the documentation in yjit.md to show the complete perf workflow, including invoking YJIT to gather perf data?

done https://github.com/ruby/ruby/pull/10246/commits/1d5c8518712e34859e6adcb142aae91480fcaf0e

doc/yjit/yjit.md
k0kubun: Sure, I updated the document that way in https://github.com/ruby/ruby/pull/10246/commits/2abcd19a2456a9bcdefba59daa49977a9cac1471. I still left the instructions to build perf from source in case we want to develop a new feature using something we currently don't parse but is available in the original perf events.

PR #10247, opened 2024-03-14
Namespace on read

tagomoris:
https://bugs.ruby-lang.org/issues/19744

### Features
* [x] Namespaces to require/load .rb files
* [x] Namespaces to require extensions
* [x] Supporting `autoload`
* [x] Reopening existing classes/modules
* [x] Adding hooks to `end` of class/module definitions
* [x] Defining methods on existing classes/modules
* [x] Eval code string in namespaces
* [x] Top-level methods in namespaces
* [x] Creating Procs in namespaces
* [x] Set the global namespace before running scripts
* [ ] Defining singleton methods
* [ ] Calling include on existing classes/modules
* [x] Newly added constant separation
* [x] Global variable separation
* [x] Implicit reopen with `class/module` in namespaces on built-in classes/modules
* [ ] Loading C libraries as dependencies of extensions in namespaces

### Nothing to do
* Defining singleton methods on existing objects
* Calling extend on global objects with namespaced modules
* Class variables
* Class instance variables

### TODOs

* [ ] Apply refined definitions on methods defined previously
* [ ] Implement `basename` in C instead of calling `rb_funcall` in `rb_namespace_local_extension`
* [ ] Update `Namespace` to contain `rb_namespace_t` directly instead of having `Namespace::Entry`
* [ ] Initialize `refiner` module lazily to not make overhead if monkey patches are not done
* [ ] Support calling `require` in methods defined in namespace on non-namespaced threads
* [ ] Support `remove_const`, `const_defined`, `const_added`, `rb_define_global_const`, etc


tagomoris:
I replaced this by #13226

PR #10249, opened 2024-03-14
merge revision(s) bbd249e351af7e4929b518a5de73a832b5617273: [Backport #20192]

nurse:
YJIT: Properly reject keyword splat with `yield`

We don't have support for keyword splat anywhere, but we tried to
compile these anyways in case of `invokeblock`. This led to bad things
happening such as passing the wrong value and passing a hash into
rb_yjit_array_len(), which raised in the middle of compilation.

[Bug #20192]


bootstraptest/test_yjit.rb
```diff
@@ -11,6 +11,13 @@ def call_foo
   call_foo
 }
 
+# regression test for keyword splat with yield
+assert_equal 'nil', %q{
+  def splat_kw(kwargs) = yield(**kwargs)
+
+  splat_kw({}) { _1 }.inspect
+}
```
k0kubun, line 19:
```suggestion
} unless rjit_enabled? # Not yet working on RJIT
```

PR #10262, opened 2024-03-15
Warn on modified future frozen string literal

casperisfine:
    
As a path toward enabling frozen string literals by default in the future, this commit introduce "chilled strings". From a user perspective chilled strings pretend to be frozen, but on the first attempt to mutate them, they lose their frozen status and emit a warning rather than to raise a `FrozenError`.

Implementation wise, `rb_compile_option_struct.frozen_string_literal` is no longer a boolean but a tri-state of `enabled/disabled/unset`.

When code is compiled with frozen string literals neither explictly enabled or disabled, string literals are compiled with a new `putchilledstring` instruction. This instruction is identical to `putstring` except it marks the String with the `STR_CHILLED (FL_USER3)` and `FL_FREEZE` flags.

Chilled strings have the `FL_FREEZE` flag as to minimize the need to check for chilled strings across the codebase, and to improve compatibility with C extensions.

Notes:
  - `String#freeze`: clears the chilled flag.
  - `String#-@`: acts as if the string was mutable.
  - `String#+@`: acts as if the string was mutable.
  - `String#clone`: copies the chilled flag.

cc @etiennebarrie


casperisfine:
Alright, we're down to a single failure in `rbs` which I don't quite understand:

```
  Error: test_record_with_optional_key(RBS::TypeParsingTest): RBS::ParsingError: a.rbs:1:2...1:3: Syntax error: unexpected record key token, token=`?` (pQUESTION)
  /Users/runner/work/ruby/ruby/src/gems/src/rbs/lib/rbs/parser_aux.rb:7:in 'RBS::Parser._parse_type'
  /Users/runner/work/ruby/ruby/src/gems/src/rbs/lib/rbs/parser_aux.rb:7:in 'RBS::Parser.parse_type'
  /Users/runner/work/ruby/ruby/src/gems/src/rbs/test/rbs/type_parsing_test.rb:673:in 'RBS::TypeParsingTest#test_record_with_optional_key'
```

```
    Parser.parse_type("{ ?foo: untyped }").yield_self do |type|
      assert_instance_of Types::Record, type
      assert_equal({}, type.fields)
      assert_equal({
                     foo: Types::Bases::Any.new(location: nil),
                   }, type.optional_fields)
      assert_equal "{ ?foo: untyped }", type.location.source
    end
```

Seems like there might be a regression in the parser? I'll dig a bit more.

test/ruby/test_compile_prism.rb
```diff
@@ -724,6 +718,16 @@ def test_InterpolatedStringNode
       RUBY
     end
 
+    def test_concatenated_StringNode
+      pend("PRISM doesn't currently allow to compile this correctly")
```
casperisfine, line 722: cc @kddnewton, as mentioned on Slack, I don't think we can currently make this test pass with the current parser. It only passed by accident previously.

Dunno if `pend` is the best way, perhaps you'd like an issue open somewhere?

spec/ruby/core/string/chilled_string_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      input = "chilled"
+      Ractor.shareable?(input).should == false
+    end
+
+    it "can be made ractor shareable" do
+      input = "chilled"
+      shareable = Ractor.make_shareable(input)
+      shareable.should == input
+      Ractor.shareable?(shareable).should == true
+    end
```
eregon, line 56: Could you move the Ractor-related tests to test-all?
We avoid testing Ractor in ruby/spec for reasons explained [here](https://github.com/ruby/spec/pull/1043#issuecomment-1930313473), in short Ractor is at least currently a CRuby-specific API.

casperisfine: Thanks. Done.

tenderlove (approved):

spec/ruby/core/string/chilled_string_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+      it "returns a different instance" do
+        input = "chilled"
+        duped = (+input)
+        duped.frozen?.should == false
+        duped.object_id.should_not == input.object_id
+      end
+    end
+
+    describe "mutation" do
+      it "emits a warning" do
```
KJTsanaktsidis, line 30: I think it would be good if there was also a spec which asserted that the mutation actually sticks (i.e. something like:)

```
input = "chilled"
input << "mutated"
input.should eql("chilledmutated")
```

I opened the spec first when reviewing this to try and understand precicely what the behaviour was supposed to be and it took me a few goes to figure out that chilled strings are `frozen?` but still actually mutable (with a warning).

error.c
```diff
@@ -3860,6 +3860,12 @@ void
 rb_error_frozen_object(VALUE frozen_obj)
 {
     rb_yjit_lazy_push_frame(GET_EC()->cfp->pc);
+
+    if (CHILLED_STRING_P(frozen_obj)) {
+        CHILLED_STRING_MUTATED(frozen_obj);
```
KJTsanaktsidis, line 3865: I first thought it was strange that this was done here (and after `rb_yjit_lazy_push_frame`), and not as part of the implementation of `rb_check_frozen_inline` (which calls this).  _however_, I see it really actually is required. The documentation for `rb_yjit_lazy_push_frame` says:

> Push a C method frame if the given PC is supposed to lazily push one.
> This is called from rb_raise() (at rb_exc_new_str()) and other functions
> that may make a method call (e.g. rb_to_int()).

And CHILLED_STRING_MUTATED _can_ wind up making a method call because warnings are implemented by actually calling `::Warning.warn`.

I guess this isn't news to anybody but me though :P

include/ruby/internal/intern/error.h
```diff
@@ -190,7 +190,6 @@ RBIMPL_ATTR_NONNULL(())
  */
 void rb_error_frozen(const char *what);
 
-RBIMPL_ATTR_NORETURN()
```
KJTsanaktsidis, line 193: (Picking on this line because it was what made me start investigating)

The only other caller of this is in `vm_setinstancevariable`, which does this:

```
    if (RB_SPECIAL_CONST_P(obj)) {
        rb_error_frozen_object(obj);
        return;
    }
```

I _think_ we want to get rid of that `return` (it was previously unreachable before this PR). Otherwise, I think setting an instance variable on a chilled string will emit the warning but not actually set the ivar, if I'm reading this correctly.

(I guess there should be a test case for setting ivars on chilled strings - the yjit bootstraptest has one but it's not actually checking that the ivar stuck)

iseq.c
```diff
@@ -720,18 +720,20 @@ finish_iseq_build(rb_iseq_t *iseq)
...
-    OPT_INLINE_CONST_CACHE, /* int inline_const_cache; */
-    OPT_PEEPHOLE_OPTIMIZATION, /* int peephole_optimization; */
-    OPT_TAILCALL_OPTIMIZATION, /* int tailcall_optimization */
-    OPT_SPECIALISED_INSTRUCTION, /* int specialized_instruction; */
-    OPT_OPERANDS_UNIFICATION, /* int operands_unification; */
-    OPT_INSTRUCTIONS_UNIFICATION, /* int instructions_unification; */
-    OPT_FROZEN_STRING_LITERAL,
-    OPT_DEBUG_FROZEN_STRING_LITERAL,
-    TRUE,			/* coverage_enabled */
+    .inline_const_cache = OPT_INLINE_CONST_CACHE,
```
KJTsanaktsidis, line 723: +1 I much prefer this style especially for a struct this big.

prism_compile.c
```diff
@@ -787,6 +787,44 @@ pm_interpolated_node_compile(pm_node_list_t *parts, rb_iseq_t *iseq, NODE dummy_
...
+            const pm_node_t *part = parts->nodes[index];
+
+            if (!PM_NODE_TYPE_P(part, PM_STRING_NODE)) {
+                literal = false;
+                break;
+            }
+        }
+
+        if (literal) {
+            for (size_t index = 0; index < parts_size; index++) {
```
KJTsanaktsidis, line 801: I'm not sure I understand why this string part gluing-together logic is duplicated in the literal and the section below?

KJTsanaktsidis (approved):
I had a couple of questions but overall it looks right to me 👍

prism_compile.c
casperisfine: This is change is not strictly related to this PR, it's just to fix a prism issue that happens now that we feed `--enable-frozen-literal / disable-frozen-literal` to prism.

Prism parses `a = "foo""bar"` as string interpolation nodes, even though it's just a static string. Interpolated strings are always mutable, never frozen. So we need to do all this to figure out if we need to produce a frozen/mutabe/chilled string.

I could refactor to share this code, but I think @kddnewton wants to change how it's parsed, so not too sure if it's worth it.

include/ruby/internal/intern/error.h
casperisfine: > Otherwise, I think setting an instance variable on a chilled string will emit the warning but not actually set the ivar, if I'm reading this correctly.

`RB_SPECIAL_CONST_P` is all immediates (`nil/false/true/Integer/Float`), so chilled strings wouldn't enter this branch. But it's a good catch regardless.

spec/ruby/core/string/chilled_string_spec.rb
casperisfine: Added.

include/ruby/internal/intern/error.h
casperisfine: I added a test for that.

prism_compile.c
eregon: One thought about this is maybe `PM_NODE_FLAG_STATIC_LITERAL` could be used instead of the loop above?
Wouldn't simplify that much though.

casperisfine: Ah, didn't notice this. But yeah, probably not worth bothering.

Earlopain:
Hey @casperisfine , I don't have an ruby issue tracker account so this will have to do:

This works in Ruby 3.3 but raises on head and results in some test failures for unicorn:

```rb
require "stringio"

StringIO.new("").write("foo")
#=> 'StringIO#write': not opened for writing (IOError)
```

This is because the chilld string is not considered writeable in this codepath: https://github.com/ruby/stringio/blob/779f71359d4760df49b70cfbcf908bc012e7e4ba/ext/stringio/stringio.c#L156-L165

casperisfine:
For the short term, you can set `RUBYOPT="--disable-frozen-string-literal"`

Earlopain:
Hm, I guess this is something like `raise if frozen?`. Since they pretent to be, this now fails. I'll try to open a PR so that this prints the deprecation instead.

casperisfine:
Yes, it's right there: https://github.com/ruby/stringio/blob/779f71359d4760df49b70cfbcf908bc012e7e4ba/ext/stringio/stringio.c#L290-L294

PR #10268, opened 2024-03-15
YJIT: Support arity=-2 cfuncs

XrXr:
This type of cfuncs shows up as consume a lot of cycles in profiles of
the lobsters benchmark, even though in the stats they don't happen that
frequently. Might be a bug in the profiling, but these calls are not
too bad to support, so might as well do it.


maximecb:
There was a CI failure related to the stack canary, updating branch and restarting CI 🤞 

Thank you for diving into the profiles and tackling things like this. Will be awesome if we can speed up rails even more 🙏

maximecb (approved):

PR #10282, opened 2024-03-19
[PoC] Extract RDoc as bundled gems

hsbt:
This pull-request is PoC for Ruby 3.5.

I have a plan to migrate rdoc to bundled gems. I need to investigate that we can do `make rdoc` with rdoc as bundled gems.


hsbt:
The following failures are intentional. I can fix it after Ruby 3.5.

* https://github.com/ruby/ruby/actions/runs/8340211835/job/22823743558?pr=10282#step:11:467
* https://github.com/ruby/ruby/actions/runs/8340211833/job/22823747229?pr=10282#step:11:467

PR #10283, opened 2024-03-19
Revert "Hide public implementation of `rb_io`. (#9568)"

ioquatix:
Unfortunately this breaks `unicorn` so we will revert it.

This reverts commit 9ab1fa3bf570bf19b0d6808adf12e965aacc6d83.

https://bugs.ruby-lang.org/issues/19057#change-107293


mame:
Is there any obstacle to merge this? If not, I will merge it.

mame:
Thanks!

PR #10294, opened 2024-03-20
Fix `SSL_read: unexpected eof while reading` issues

ahrechushkin:
Ignore unexpected shutdown of TLS connections. After updating to ssl 3.0 some requests may fail with unexpected eof while reading issue.


ahrechushkin:
@junaruga Hello!
Sorry, thx

junaruga:
Thanks for sending another PR https://github.com/ruby/openssl/pull/730!

PR #10306, opened 2024-03-20
Remove excess allocation for kwsplat to kw call

XrXr:
Previously, calls like the following duplicated the kwsplat hash
unnecessarily:

```ruby
def foo(a:) = a
hash = {a: 10}
foo(**hash)
```

This is due to the fix in ca204a20231. Since it targets when the callee
has no keyword parameters, skip duplicating when the method takes
keywords.

### DIFF
--- a/vm_args.c
+++ b/vm_args.c
@@ -725,8 +725,9 @@ setup_parameters_complex(rb_execution_context_t * const ec, const rb_iseq_t * co
                 kw_flag &= ~(VM_CALL_KW_SPLAT | VM_CALL_KW_SPLAT_MUT);
             }
             else {
-                if (!(kw_flag & VM_CALL_KW_SPLAT_MUT)) {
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
If the method accepts both keyword parameters and a keyword splat parameter, the value needs to be duplicated.  However, that appears handled in `args_setup_kw_rest_parameter` .

I think when the branch is taken it should do `kw_flag |= VM_CALL_KW_SPLAT_MUT;`, since the value is now mutable.  What do you think?

XrXr:
> If the method accepts both keyword parameters and a keyword splat parameter, the value needs to be duplicated.

Debugger showed that it was duplicated in ignore_keyword_hash_p(). Anyways, it's handled.

> I think when the branch is taken it should do kw_flag |= VM_CALL_KW_SPLAT_MUT;, since the value is now mutable.

Makes sense, since we do this everywhere else.

jeremyevans (approved):

PR #10323, opened 2024-03-21
YJIT: Propagate Array, Hash, and String classes

k0kubun:
This PR allows YJIT to upgrade types in `Context` to `Array`, `Hash`, and `String` even if the object is not frozen. YJIT can now skip class checks when it's already known.

Because `RBASIC_CLASS` of an object can change when the object gets a singleton class (or calls `IO#reopen`), blocks skipping a class guard will be invalidated when any object of the class gets a singleton class, disabling the optimization for the class going forward.

This PR seems to impact the following benchmarks:

```
before: ruby 3.4.0dev (2024-03-21T18:14:25Z master 75d9019e6f) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-03-21T21:26:30Z yjit-singleton 982653a4fe) +YJIT [x86_64-linux]

------------  -----------  ----------  ----------  ----------  -------------  ------------
bench         before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
activerecord  46.4         0.7         45.1        0.7         1.00           1.03
lobsters      980.6        1.3         971.3       1.5         0.99           1.01
railsbench    2318.6       0.8         2289.7      0.8         1.01           1.01
rubocop       128.3        4.3         125.7       4.5         1.03           1.02
ruby-lsp      119.8        7.4         114.0       0.9         0.95           1.05
protoboeuf    44.0         6.0         42.5        6.5         1.04           1.04
------------  -----------  ----------  ----------  ----------  -------------  ------------
```

### DIFF
--- a/class.c
+++ b/class.c
@@ -29,6 +29,7 @@
 #include "internal/variable.h"
 #include "ruby/st.h"
 #include "vm_core.h"
+#include "yjit.h"
 
 /* Flags of T_CLASS
  *
@@ -805,6 +806,7 @@ make_singleton_class(VALUE obj)
     FL_SET(klass, FL_SINGLETON);
     RBASIC_SET_CLASS(obj, klass);
     rb_singleton_class_attached(klass, obj);
+    rb_yjit_invalidate_no_singleton_class(orig_class);
 
     SET_METACLASS_OF(klass, METACLASS_OF(rb_class_real(orig_class)));
     return klass;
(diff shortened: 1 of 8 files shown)


XrXr (approved):
Nice. It looks like once we see a singleton class, we insert guards like usual as a fallback. So everything is covered.

class.c
```diff
@@ -805,6 +806,7 @@ make_singleton_class(VALUE obj)
     FL_SET(klass, FL_SINGLETON);
     RBASIC_SET_CLASS(obj, klass);
     rb_singleton_class_attached(klass, obj);
+    rb_yjit_invalidate_no_singleton_class(orig_class);
```
maximecb, line 809: Nice descriptive function name :)

maximecb (approved):
Nicely done. Clean PR 👍  Great performance results 👌

PR #10334, opened 2024-03-22
[DOC] Small edits in rbasic.h

fxn:
I leveraged a pass reading `rbasic.h` to fix a few minor things. Mostly English usage, and ruby -> Ruby when referring to the language.

For example, in the sentence

> Every  single ruby objects have  them in common.

"objects" and "have" do not agree with the "Every single".

In this file, comments are justified. Also, there are two spaces after periods. The patch preserves this.

### DIFF
--- a/include/ruby/internal/core/rbasic.h
+++ b/include/ruby/internal/core/rbasic.h
@@ -63,8 +63,8 @@ RUBY_ALIGNAS(SIZEOF_VALUE)
 RBasic {
 
     /**
-     * Per-object  flags.  Each  Ruby  object has its own  characteristics apart
-     * from its class.  For instance, whether an object  is frozen or not is not
+     * Per-object flags.   Each Ruby object  has its own  characteristics apart
+     * from its class.  For instance, whether an object is frozen or not is not
      * controlled by its class.  This is where such properties are stored.
      *
      * @see enum ::ruby_fl_type
--- a/include/ruby/internal/core/rbasic.h
+++ b/include/ruby/internal/core/rbasic.h
@@ -56,22 +56,20 @@ enum ruby_rvalue_flags {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


include/ruby/internal/core/rbasic.h
```diff
@@ -56,33 +56,31 @@ enum ruby_rvalue_flags {
...
 RBasic {
 
     /**
-     * Per-object  flags.  Each  ruby  objects have  their own  characteristics
-     * apart from their  classes.  For instance whether an object  is frozen or
-     * not is not  controlled by its class.  This is  where such properties are
-     * stored.
+     * Per-object  flags.  Each  Ruby  object has its own  characteristics apart
+     * from its class.  For instance, whether an object  is frozen or not is not
+     * controlled by its class.  This is where such properties are stored.
```
nobu, line 68:
```suggestion
     * Per-object flags.   Each Ruby object  has its own  characteristics apart
     * from its class.  For instance, whether an object is frozen or not is not
     * controlled by its class.  This is where such properties are stored.
```

include/ruby/internal/core/rbasic.h
```diff
@@ -56,33 +56,31 @@ enum ruby_rvalue_flags {
...
      *
      * @see enum ::ruby_fl_type
      *
-     * @note  This is ::VALUE rather than  an enum for alignment purpose.  Back
+     * @note  This is ::VALUE rather than  an enum for alignment purposes.  Back
```
nobu, line 72:
```suggestion
     * @note  This is ::VALUE rather than an enum for alignment purposes.  Back
```

nobu (approved):
Justify paragraphs.

fxn:
@nobu is that paragraph justification done by Emacs?

PR #10335, opened 2024-03-22
Mark frame info structs with rb_gc_mark_movable

gmcgibbon:
Using `rb_gc_mark_movable` and a reference update function, we can make frame infos movable in memory, and avoid pinning frame info backtraces.

This demonstrates what I'm trying to fix:
```rb
require "objspace"
exceptions = []

GC.disable
50_000.times do
  begin
    raise "some exception"
  rescue => exception
    exception.backtrace_locations
    exceptions << exception
  end
end
GC.enable
GC.compact

p ObjectSpace.dump_all(output: :string).lines.grep(/"pinned":true/).count
```

On master, it seems this creates ~50k pinned objects, and with this patch it doesn't.

### DIFF
--- a/vm_backtrace.c
+++ b/vm_backtrace.c
@@ -522,7 +522,10 @@ static const rb_data_type_t backtrace_data_type = {
         NULL, // No external memory to report,
         backtrace_update,
     },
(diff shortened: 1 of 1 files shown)


gmcgibbon:
Hm, looks like something might not be working quite right. Peter and I will have a look next week.

nobu:
Hmmm, sorry, is this different from #10336?

PR #10336, opened 2024-03-22
Mark iseq structs with rb_gc_mark_movable

gmcgibbon:
Using `rb_gc_mark_movable` and a reference update function, we can make instruction sequences movable in memory, and avoid pinning compiled iseqs.

This demonstrates what I'm trying to fix:
```rb
require "objspace"
iseqs = []

GC.disable
50_000.times do
  iseqs << RubyVM::InstructionSequence.compile("")
end
GC.enable
GC.compact

p ObjectSpace.dump_all(output: :string).lines.grep(/"pinned":true/).count
```

On master, it seems this creates ~50k pinned objects, and with this patch it doesn't.

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -1386,18 +1386,30 @@ rb_iseq_remove_coverage_all(void)
 static void
 iseqw_mark(void *ptr)
 {
-    rb_gc_mark((VALUE)ptr);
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):

nobu:
This broke debug inspector API.

PR #10338, opened 2024-03-22
Do not apply anon_rest optimization when passed array uses keyword-flagged hash

jeremyevans:
The optimization sets args->rest_dupped to avoid allocating an array, but this is not safe if the splat array ends in a keyword flagged hash.  Unset args->rest_dupped in this case.

Fixes [Bug #20388]


tenderlove:
Thanks Jeremy!!

PR #10339, opened 2024-03-22
Add array/hash implicit allocation tests

jeremyevans:
These are designed to prevent allocation regressions (commits that increase the number of implicitly allocated arrays and hashes). We have already had three commits in the last couple weeks to fix allocation regressions:

* 15dc3aaa311b32203d8ffb414bcf9b8e55ce5691
* aceee71c35e0b387691836e756b4e008efd84cf1
* c38878494377c94f2425a81e598260ea944ef7f3

This test suite should hopefully allow us to find such regressions in CI before commit, to avoid committing future allocation regressions.

This uses assert_separately to test each case, which is very slow. Hopefully there is a better way that can isolate the entire test suite, instead of isolating each individual check.  This does need some isolation as it disables garbage collection to get accurate allocation counts, and it cannot be run concurrently with other threads in the same process.

### DIFF
--- a/test/ruby/test_allocation.rb
+++ b/test/ruby/test_allocation.rb
@@ -1,0 +1,656 @@
+# frozen_string_literal: false
+require 'test/unit'
+
+class TestAllocation < Test::Unit::TestCase
+  def check_allocations(checks)
+    assert_separately([], <<~RUBY)
+      $allocations = [0, 0]
+      $counts = {}
+      failures = []
+
+      def self.num_allocations
+        ObjectSpace.count_objects($counts)
+        arrays = $counts[:T_ARRAY]
+        hashes = $counts[:T_HASH]
+        yield
+        ObjectSpace.count_objects($counts)
+        arrays -= $counts[:T_ARRAY]
+        hashes -= $counts[:T_HASH]
+        $allocations[0] = -arrays
+        $allocations[1] = -hashes
+      end
+
+      define_singleton_method(:check_allocations) do |num_arrays, num_hashes, check_code|
+        instance_eval <<~RB
+          empty_array = empty_array = []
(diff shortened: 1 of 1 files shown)


test/ruby/test_allocation.rb
```diff
@@ -0,0 +1,608 @@
...
+      GC.disable
+
+      allocations = [0, 0]
+      counts = {}
+      empty_array = empty_array = []
+      empty_hash = empty_hash = {}
+      array1 = array1 = [1]
+      hash1 = hash1 = {a: 2}
+      nill = nill = nil
+      block = block = lambda{}
```
XrXr, line 30: Not sure why these are assigned to twice.

XrXr (approved):
You could group together the cases in each test method into one assert_separately if the current approach is too slow.

Thanks for putting these test in place!

test/ruby/test_allocation.rb
jeremyevans: To avoid verbose warnings for the calls that do not use them.

jeremyevans:
> You could group together the cases in each test method into one assert_separately if the current approach is too slow.

The main issue with that approach is that it stops at the first failure.  I suppose I could work around that, building an array of all cases that fail, and then report it at the end if the array isn't empty.

One area this doesn't currently test is use of ruby2_keywords.  I'd like to add tests for that, not only as a method modifier, but more importantly, when passing the flagged hash in a splat to all of the other method types.

jeremyevans:
I pushed a rebased commit with a much faster approach, using one `assert_separately` per test.  Assuming it passes CI, I plan to merge it.   I'll try to add the `ruby2_keywords` tests later when I find time.

PR #10340, opened 2024-03-23
[DOC] Comment fixes in prism/prism.c

francktrouillez:
I was browsing the `prism/prism.c` file, and saw some typos.

- Some English fixes (missing words, etc.)
  i.e.: `// Indicate that this node an exclusive range if the operator is '...'.` misses a verb and becomes `// Indicate that this node is an exclusive range if the operator is '...'.`
- Use infinitive verbs instead of 3rd person present simple to describe functions
  i.e.: `* Returns a string content token at a particular location that is empty.` is using 3rd person present simple and becomes `* Return a string content token at a particular location that is empty.`
- Sentences in comments ends with `.`
  i.e.: `This sets the end location on all rescues once we know it` adds `.` and becomes `This sets the end location on all rescues once we know it.`
- Some other minor fixes, etc.

It also adds a missing operator in the descriptive comment of operators for `||=`.


hsbt:
Can you submit this to [upstream](https://github.com/ruby/prism)?

/cc @kddnewton

francktrouillez:
Sure! Here it is:
- https://github.com/ruby/prism/pull/2623

PR #10341, opened 2024-03-23
Avoid using `else` for `peek_at` function in prism/prism.c

francktrouillez:
I was reading `prism/prism.c` and found out this situation.

This allows to avoid using `else` since the `if` branch has a return statement. This case occurred in the `peek_at` function in `prism/prism.c`.


hsbt:
Can you submit this to [upstream](https://github.com/ruby/prism)?

francktrouillez:
Sure! Here it is:
- https://github.com/ruby/prism/pull/2624

PR #10347, opened 2024-03-25
Support warning feature for bundled gems with bootsnap and zeitwerk

hsbt:
I added some fixes from https://github.com/ruby/ruby/pull/10244

PR #10351, opened 2024-03-25
lib/bundled_gems.rb: dynamically ignore Kernel.require decorators

casperisfine:
Followup: https://github.com/ruby/ruby/pull/10347

This avoid directly referencing bootsnap and zeitwerk, and also handle other gems that may decorate `require`.

@hsbt @fxn @jeremyevans

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -147,7 +147,7 @@ module Gem::BUNDLED_GEMS
           next
         end
 
-        unless cl.path.match?(/bootsnap|zeitwerk/)
+        if cl.base_label != "require"
           location = cl.path
           break
         end


fxn:
I think this could be a good compromise. It removes the hard-coded gem names, and does not go all the way to defining an API.

From a logical point of view, if I am not mistaken, it assumes the decoration invokes the previous version of `require` without indirections. That is, I have not tested it, by I guess this decoration would trigger the `break` too early (in `foo`):

```ruby
module Kernel
  alias_method :original_require, :require

  def require(feature)
    foo(feature)
  end

  def foo(feature)
    original_require(feature)
  end
end
```

That is a synthetic example, in a real scenario that `foo` would exist to extract some code that makes sense.

However, while that may be logically possible, in practice you normally do not toss new methods into `Kernel` like that, I suppose, since `Kernel` is such an important module share by everyone.

To me, it is a good approach to be tried first for its simplicity. Only complicate things if practice demonstrate its need.

👍

casperisfine:
> That is, I have not tested it, by I guess this decoration would trigger the `break` too early (in `foo`):

Yes indeed. It's not entirely fool proof for sure. But I think it's good enough?

hsbt:
Thanks. I confirmed to work this with `bootsnap` and `zeitwerk`.

mattbrictson:
This is not working for me in ruby master @ e0f4cbd612. Here is a simple reproduction script:

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "bootsnap", require: false
  gem "childprocess", "5.0.0", require: false # Has undeclared logger dependency
end

ENV["BOOTSNAP_CACHE_DIR"] ||= "tmp/cache/bootsnap"
require "bootsnap/setup"
require "childprocess"
```

The warning for `logger` is still being attributed to `bootsnap`:

> /Users/mbrictson/.rbenv/versions/ruby-dev/lib/ruby/gems/3.4.0+0/gems/bootsnap-1.18.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30: warning: /Users/mbrictson/.rbenv/versions/ruby-dev/lib/ruby/3.4.0+0/logger.rb was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.5.0. Add logger to your Gemfile or gemspec.

Whereas if I disable bootsnap with `DISABLE_BOOTSNAP=1`, the warning is correctly attributed to `childprocess`

> /Users/mbrictson/.rbenv/versions/ruby-dev/lib/ruby/gems/3.4.0+0/gems/childprocess-5.0.0/lib/childprocess.rb:7: warning: logger was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.5.0. Add logger to your Gemfile or gemspec.

I'm using:

```
$ ruby -v
ruby 3.4.0dev (2024-08-01T08:49:32Z master e0f4cbd612) [x86_64-darwin23]
```

Should I report a bug at https://bugs.ruby-lang.org ?

casperisfine:
Looking.

casperisfine:
Alright, I see two issues:

  - The `frames_to_skip = 2` is outdated, should be `3` now. I'll try to find a way that we don't need to keep this in sync.
  - Then the main problem is that we'd also need to compute the `uplevel` when calling `warn`.

I'll try to improve that.

PR #10352, opened 2024-03-25
Update minitest and power_assert bundled gems

casperisfine:
They were pointing to branches to be chilled string compatible. Both patches have been merged now.


hsbt (approved):
I also confirmed that commit hash.

PR #10353, opened 2024-03-25
Add an example to base_label API

fxn:
Easy one.

### DIFF
--- a/vm_backtrace.c
+++ b/vm_backtrace.c
@@ -287,15 +287,14 @@ location_label(rb_backtrace_location_t *loc)
  *	    1.times do
  *	      puts caller_locations(0).first.label
  *	    end
- *
  *	  end
  *	end
  *
  * The result of calling +foo+ is this:
  *
- *	label: foo
- *	label: block in foo
- *	label: block (2 levels) in foo
+ *	foo
+ *	block in foo
+ *	block (2 levels) in foo
  *
  */
 static VALUE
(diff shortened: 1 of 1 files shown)


vm_backtrace.c
```diff
@@ -315,9 +315,10 @@ location_base_label(rb_backtrace_location_t *loc)
 }
 
 /*
- * Returns the base label of this frame.
+ * Returns the label of this frame without decoration.
  *
- * Usually same as #label, without decoration.
+ * For example, if the label is `foo`, this method returns `foo`, same, but if
```
byroot, line 320:
```suggestion
 * For example, if the label is `foo`, this method returns `foo` as well, but if
```

byroot: Maybe it would be worth to take the same example from `#label`?

fxn:
@byroot I am looking at the implementation of `location_label` and it does a lot of things, I am not confident "Returns the label of this frame without decoration" is accurate. But "usually" as it was originally written is not good documentation either, you have to tell me what does the method return.

fxn:
@byroot In order to be more accurate with confidence, I should dig into this. By now, I have restored the "usually", so the patch does not change the documented relationship between `label` and `base_label`.

byroot:
> I should dig into this

No problem. Your patch look fine to me, but just let me know when your are done.

fxn:
@byroot yes, let's keep this patch as it is, and if in the future I dig into the implementation of `label` and may submit a followup.

PR #10355, opened 2024-03-25
Expose rb_str_chilled_p

etiennebarrie:
Some extensions (like stringio) may need to differentiate between chilled strings and frozen strings.

They can now use rb_str_chilled_p but must check for its presence since the function will be removed when chilled strings are removed.

[[Bug #20389]](https://bugs.ruby-lang.org/issues/20389)

cc @eregon @nobu @byroot


ext/-test-/string/chilled.c
```diff
@@ -0,0 +1,13 @@
+#include "ruby.h"
+
+static VALUE
+bug_s_rb_str_chilled_p(VALUE self, VALUE str)
+{
+    return INT2FIX(rb_str_chilled_p(str));
```
eregon, line 6:
```suggestion
    return rb_str_chilled_p(str) ? Qtrue : Qfalse;
```
seems nicer because then it returns a boolean to Ruby and not an Integer.

include/ruby/internal/intern/string.h
```diff
@@ -601,6 +601,19 @@ VALUE rb_str_dup(VALUE str);
...
+ *
+ * This function is temporary and users must check for its presence using
+ * mkmf's have_func helper. If this function is not implemented, then strings
+ * can't be chilled.
+ *
+ * @param[in]  str  A string.
+ * @retval     1    The string is chilled.
+ * @retval     0    Otherwise.
+ */
+int rb_str_chilled_p(VALUE str);
```
eregon, line 615: We could use `bool` as a return type these days, e.g. there is already `include/ruby/vm.h:57:bool ruby_free_at_exit_p(void);` and more.

ext/-test-/string/chilled.c
casperisfine: Why thought? It's meant meant to be used from C. I don't see the value of returning `Qtrue / Qfalse` over `1/0` as it requires an extra `RTEST()`.

include/ruby/internal/intern/string.h
casperisfine: Ah, I wasn't sure. Updated.

ext/-test-/string/chilled.c
eregon: I was commenting about the test, not the impl of rb_str_chilled_p, but looks like you saw it as well after.

eregon (approved):

include/ruby/internal/intern/string.h
```diff
@@ -601,6 +601,19 @@ VALUE rb_str_dup(VALUE str);
...
+ * This function is temporary and users must check for its presence using
+ * mkmf's have_func helper. If this function is not implemented, then strings
+ * can't be chilled.
+ *
+ * @param[in]  str  A string.
+ * @retval     1    The string is chilled.
+ * @retval     0    Otherwise.
+ */
+bool rb_str_chilled_p(VALUE str);
+
```
nobu, line 616: For convenience.
```suggestion

#define HAVE_RB_STR_CHILLED_P 1

```

eregon: I was thinking the same, but then thought probably very few extensions need this so then it seems OK to use `have_func` for those extensions (`have_func` being rather slow).
If it's added the documentation above should be updated to mention no need to use `have_func`.

casperisfine: I updated the PR with @nobu's suggestion.

nobu (approved):

include/ruby/internal/intern/string.h
```diff
@@ -601,6 +601,21 @@ VALUE rb_str_dup(VALUE str);
  */
 VALUE rb_str_resurrect(VALUE str);
 
+/**
+ * Returns whether a string is chilled or not.
+ *
+ * This function is temporary and users must check for its presence using
+ * #ifdef HAVE_RB_STR_CHILLED_P. If this function is not implemented, then strings
```
eregon, line 608:
```suggestion
 * #ifdef HAVE_RB_STR_CHILLED_P. If HAVE_RB_STR_CHILLED_P is not defined, then strings
```

PR #10360, opened 2024-03-25
YJIT: Inline simple getlocal+leave iseqs

XrXr:
This mainly targets things like `T.let()` from Sorbet, which is just an
identity function at runtime and only a hint for the static checker.
Only deal with simple caller and callees (no keywords and splat etc.).

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6754,11 +6754,16 @@ fn gen_send_bmethod(
 /// The kind of a value an ISEQ returns
 enum IseqReturn {
     Value(VALUE),
+    LocalVariable(u32),
     Receiver,
 }
 
-/// Return the ISEQ's return value if it consists of only putnil/putobject and leave.
-fn iseq_get_return_value(iseq: IseqPtr, captured_opnd: Option<Opnd>) -> Option<IseqReturn> {
+extern {
+    fn rb_simple_iseq_p(iseq: IseqPtr) -> bool;
+}
+
+/// Return the ISEQ's return value if it consists of one simple instruction and leave.
+fn iseq_get_return_value(iseq: IseqPtr, captured_opnd: Option<Opnd>, ci_flags: u32) -> Option<IseqReturn> {
     // Expect only two instructions and one possible operand
     let iseq_size = unsafe { get_iseq_encoded_size(iseq) };
     if !(2..=3).contains(&iseq_size) {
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -7055,11 +7071,24 @@ fn gen_send_iseq(
...
-    if let (Some(value), None, false) = (iseq_get_return_value(iseq, captured_opnd), block_arg_type, opt_send_call) {
+    if let (Some(value), None, false) = (iseq_get_return_value(iseq, captured_opnd, flags), block_arg_type, opt_send_call) {
         asm_comment!(asm, "inlined simple ISEQ");
         gen_counter_incr(asm, Counter::num_send_iseq_inline);
 
         match value {
+            IseqReturn::LocalVariable(local_idx) => {
+                // Put the local variable at the return slot
+                let stack_local = asm.stack_opnd(argc - 1 - local_idx as i32);
+                let stack_return = asm.stack_opnd(argc - i32::from(captured_opnd.is_some()));
```
k0kubun, line 7082: Now that we're rejecting the `captured_opnd.is_some()` case, this check shouldn't be necessary.

```suggestion
                let stack_return = asm.stack_opnd(argc);
```

yjit/src/codegen.rs
```diff
@@ -7055,11 +7071,24 @@ fn gen_send_iseq(
...
+                asm.mov(stack_return, stack_local);
+
+                // Update the mapping for the return value
+                let mapping = asm.ctx.get_opnd_mapping(stack_local.into());
+                asm.ctx.set_opnd_mapping(stack_return.into(), mapping);
+
+                // Pop everything but the return value
+                asm.stack_pop(argc as usize - usize::from(captured_opnd.is_some()));
```
k0kubun, line 7090: ditto

```suggestion
                asm.stack_pop(argc);
```

yjit/src/codegen.rs
```diff
@@ -6774,6 +6779,17 @@ fn iseq_get_return_value(iseq: IseqPtr, captured_opnd: Option<Opnd>) -> Option<I
         return None;
     }
     match first_insn {
+        YARVINSN_getlocal_WC_0  => {
+            // Only accept simple positional only cases for both the caller and the callee.
+            // Reject block ISEQs to avoid autosplat and other block parameter complications.
```
k0kubun, line 6784: Rejecting the `captured_opnd.is_some()` case only impacts `yield`. Is this not a problem for `block.call`?

maximecb (approved):

jez:
> This mainly targets things like T.let() from Sorbet

I think possibly you mean `T.unsafe`, unless you're also working with a monkey patch of `T.let` defined like this:

```ruby
def self.let(x, type)
  x
end
```

anyways, neat PR!

yjit/src/codegen.rs
XrXr: It is a problem for `block.call`, but we essentially do a manual fallback with rb_optimized_call() for those cases 😅 All the block setup code check `captured_opnd `.

casperisfine:
> you're also working with a monkey patch of `T.let` defined like this:

We do yes.

PR #10365, opened 2024-03-26
Don't clear pending interrupts in the parent process.

ioquatix:
While experimenting with `Thread.handle_interrupt`, I found this bug with `Process.fork`.

Essentially, `Process.fork` will clear any pending interrupts... in the child process AND the parent process. I believe it should only clear pending interrupts in the child process and NOT the parent process.

See <https://bugs.ruby-lang.org/issues/20393> for more context.

### DIFF
--- a/process.c
+++ b/process.c
@@ -1682,7 +1682,6 @@ before_fork_ruby(void)
 static void
 after_fork_ruby(rb_pid_t pid)
 {
-    rb_threadptr_pending_interrupt_clear(GET_THREAD());
     if (pid == 0) {
         // child
         clear_pid_cache();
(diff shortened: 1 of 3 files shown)


eregon:
Yes, that'd be fine, you'll need `guard -> { Process.respond_to? :fork } do` around then.
I suppose it only happens on Process.fork and not on system() or spawn()?

eregon:
FYI it's also OK to only add in ruby/spec and not test-all.
Might makes sense here, `fork clears pending interrupts in the child but not in the parent` seems much clearer than `test_handle_interrupt_with_fork`.

nobu (approved):

ioquatix:
@eregon as we expect to backport this commit, I'll add a ruby-spec in a follow-up PR.

KJTsanaktsidis:
I'm a bit late to the party on this :P but i agree that this is correct for what it's worth.

PR #10367, opened 2024-03-26
Sync power_assert version field in `gems/bundled_gems`

kateinoigakukun:
`power_assert` gem suffixes the version number with `dev` when it is not released yet. Out of sync version number in `gems/bundled_gems` confuses `tool/rbinstall.rb` and results in not installing the gem.


kateinoigakukun:
Huh, it broke `make up`...

nobu:
#10381  may work for this.

PR #10401, opened 2024-03-28
YJIT: Optimize putobject+opt_ltlt for integers

XrXr:
In `jit_rb_int_lshift()`, we guard against the right hand side changing
since we want to avoid generating variable length shifts. When control
reaches a putobject and `opt_ltlt` pair, though, we know that the right
hand side never changes.

This commit detects this situation and substitutes an implementation
that does not guard against the right hand side changing, saving that
work.

Deleted some putobject Rust tests since they aren't that valuable and
cause linking issues.

Nice boost to `optcarrot` and `protoboeuf`:
```
yjit-pre: ruby 3.4.0dev (2024-03-28T19:30:38Z master 8780059c38) +YJIT [arm64-darwin23]
yjit-post: ruby 3.4.0dev (2024-03-28T19:46:08Z yjit-fused-ltlt 811aa3eb6b) +YJIT [arm64-darwin23]

----------  -------------  ----------  --------------  ----------  -----------------  ------------------
bench       yjit-pre (ms)  stddev (%)  yjit-post (ms)  stddev (%)  yjit-post 1st itr  yjit-pre/yjit-post
optcarrot   915.8          0.7         838.2           0.8         1.09               1.09              
protoboeuf  29.6           12.6        26.5            2.5         1.05               1.12              
----------  -------------  ----------  --------------  ----------  -----------------  ------------------
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -30,6 +30,7 @@ pub use crate::virtualmem::CodePtr;
 /// Status returned by code generation functions
 #[derive(PartialEq, Debug)]
 enum CodegenStatus {
+    SkipNextInsn,
     KeepCompiling,
     EndBlock,
 }
@@ -1197,6 +1198,13 @@ pub fn gen_single_block(
         // Move to the next instruction to compile
         insn_idx += insn_len(opcode) as u16;
 
+        // Move past next instruction when instructed
+        if status == Some(SkipNextInsn) {
+            let next_pc = unsafe { rb_iseq_pc_at_idx(iseq, insn_idx.into()) };
+            let next_opcode: usize = unsafe { rb_iseq_opcode_at_pc(iseq, next_pc) }.try_into().unwrap();
+            insn_idx += insn_len(next_opcode) as u16;
+        }
+
         // If the instruction terminates this block
         if status == Some(EndBlock) {
             break;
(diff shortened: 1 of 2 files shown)


XrXr:
Come to think of it, we might even want to remove the other implementation that don't know that rhs is constant for sure and only keep this one 🤔

yjit/src/codegen.rs
```diff
@@ -1343,30 +1351,92 @@ fn jit_putobject(asm: &mut Assembler, arg: VALUE) {
...
 
+    if let Some(result) = fused_putobject_opt_ltlt(jit, asm, arg, ocb) {
+        return Some(result);
+    }
+
     jit_putobject(asm, arg);
     Some(KeepCompiling)
 }
 
+fn fused_putobject_opt_ltlt(
```
maximecb, line 1387: Would be good to comment explaining what this does more specifically

maximecb:
This PR is pretty cool. I'm amazed that it makes that much of a performance difference on optcarrot. Waaaaaaaaat!?

> Come to think of it, we might even want to remove the other implementation that don't know that rhs is constant for sure and only keep this one 🤔

There are some benchmarks that use a variable rhs in the shift. I think that ideally, we should implement variable-shift support in our backend/cross-platform-assembler.

Now that you have support for generating code for multiple YARV instructions at once, it could make sense to use this to tackle comparison + branch insns, especially for the most common cases. Could be a significant win on most benchmarks.

maximecb (approved):
Clever 👌 :)

k0kubun (approved):
In the future, it'd be nice to propagate (small) integer constants in Context or JITState and somehow generate the same code without introducing a fused codegen, but it seems hard to optimize away `mov` instruction in putobject in the backend for now.

PR #10403, opened 2024-03-29
show warning for unused block

ko1:
With verbopse mode (-w), the interpreter shows a warning if a block is passed to a method which does not use the given block.

Warning on:

* the invoked method is written in C
* the invoked method is not `initialize`
* not invoked with `super`
* the first time on the call-site with the invoked method (`obj.foo{}` will be warned once if `foo` is same method)

[Feature #15554]

`Primitive.attr! :use_block` is introduced to declare that primitive functions (written in C) will use passed block.

### DIFF
--- a/lib/optparse.rb
+++ b/lib/optparse.rb
@@ -461,7 +461,7 @@ class OptionParser
       candidates
     end
 
-    def candidate(key, icase = false, pat = nil)
+    def candidate(key, icase = false, pat = nil, &_)
       Completion.candidate(key, icase, pat, &method(:each))
     end
 
@@ -739,7 +739,7 @@ class OptionParser
       #
       # Raises an exception if argument is not present.
       #
-      def parse(arg, argv)
+      def parse(arg, argv, &_)
         unless arg
           raise MissingArgument if argv.empty?
           arg = argv.shift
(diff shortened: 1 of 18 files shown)


lib/optparse.rb
```diff
@@ -461,7 +461,7 @@ def self.candidate(key, icase = false, pat = nil, &block)
       candidates
     end
 
-    def candidate(key, icase = false, pat = nil)
+    def candidate(key, icase = false, pat = nil, &)
```
nobu, line 464: This is a syntax error in 3.0 or earlier, and unacceptable for optparse that still supports older versions.

```suggestion
    def candidate(key, icase = false, pat = nil, &_)
```

lib/optparse.rb
```diff
@@ -739,7 +739,7 @@ class RequiredArgument < self
       #
       # Raises an exception if argument is not present.
       #
-      def parse(arg, argv)
+      def parse(arg, argv, &)
```
nobu, line 742:
```suggestion
      def parse(arg, argv, &_)
```

lib/rdoc/context.rb
```diff
@@ -710,7 +710,7 @@ def display(method_attr) # :nodoc:
   # This method exists to make it easy to work with Context subclasses that
   # aren't part of RDoc.
 
-  def each_ancestor # :nodoc:
+  def each_ancestor(&) # :nodoc:
```
nobu, line 713: Ditto, for RDoc.

```suggestion
  def each_ancestor(&_) # :nodoc:
```

PR #10408, opened 2024-03-30
Add missing RB_GC_GUARDs related to DATA_PTR

KJTsanaktsidis:
I discovered the problem in `compile.c` from a failing TestIseqLoad#test_stressful_roundtrip test with ASAN enabled. The other two changes in array.c and string.c I found by auditing similar usages of DATA_PTR in the codebase.

[Bug #20402]


nobu (approved):
Good catch.

PR #10413, opened 2024-03-31
[Bug #20403] Add an explicit with_udp_and_tcp helper to test_dns.rb

KJTsanaktsidis:
https://stackoverflow.com/questions/48478869/cannot-bind-to-some-ports-due-to-permission-deniedThis helper tries to bind UDP and TCP sockets to the same port, by retrying the bind if the randomly-assinged UDP port is already taken in TCP. This fixes a flaky test.

[Bug #20403]

### DIFF
--- a/test/resolv/test_dns.rb
+++ b/test/resolv/test_dns.rb
@@ -64,6 +64,50 @@ class TestResolvDNS < Test::Unit::TestCase
     end
   end
 
+  def with_udp_and_tcp(host, port)
+    if port == 0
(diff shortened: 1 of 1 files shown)


jeremyevans:
I like it, but it should be submitted upstream to https://github.com/ruby/resolv

KJTsanaktsidis:
Oh, I forgot this was mirrored, thanks. I opened https://github.com/ruby/resolv/pull/49 instead.

PR #10422, opened 2024-04-01
Update ruby/spec

andrykonchin:


eregon:
FYI: @andrykonchin is helping me with synchronizing specs between ruby/spec, ruby/ruby, jruby/jruby and oracle/truffleruby.

PR #10432, opened 2024-04-02
Fix wheather -> whether typos in configure.

eightbitraptor:


nobu:
Seems duplicate of #10431.

PR #10435, opened 2024-04-03
Use Rake 13.2.1 at Ruby 3.3

hsbt:
Fixed https://github.com/ruby/actions/actions/runs/8527754153/job/23360014476#step:16:2936


hsbt:
This change is available on https://github.com/ruby/ruby/pull/10348. We need to merge it instead of this.

PR #10442, opened 2024-04-03
YJIT: Let sp_opnd take the number of slots

k0kubun:
This PR simplifies the interface of `ctx.sp_opnd()`.

For most of the time, `ctx.sp_opnd` takes `something * SIZEOF_VALUE`. It seems tedious to repeat `* SIZEOF_VALUE` everywhere. It needs to take the number of bytes when `RUBY_SIZEOF_CONTROL_FRAME` is used for stack overflow checks, but it's a multiple of `SIZEOF_VALUE` too. 

Statically asserting `RUBY_SIZEOF_CONTROL_FRAME % SIZEOF_VALUE == 0`, which seems like a fair assumption, we should be able to safely change the interface to take the number of `VALUE` slots.


XrXr (approved):

maximecb (approved):
Good change 👍

PR #10456, opened 2024-04-04
Enable GC loading from a shared library

eightbitraptor:
This PR replaces https://github.com/ruby/ruby/pull/10302 as a method for dynamically loading a shared GC library. This new approach has been guided by feedback from @kateinoigakukun and @nobu 

The approach taken here is as follows:

- Enable this feature by configuring with `--with-shared-gc`
- When enabled, Ruby will check for the presence of the environment variable `RUBY_GC_LIBRARY_PATH`
- If that path exists and points to a valid shared object
  - Open the shared object (using `dln_open`), and map the `Init_GC` function to a function map
  - call `Init_GC` as part of `rb_objspace_alloc` when initialising the GC
- If that path does not point to a valid shared object
  - Degrade gracefully to initialising the existing Ruby objspace and GC
  
When Ruby is configured without the `--with-shared-gc` flag, no extra code is compiled and so existing behaviour is maintained without any performance penalty.

Enabling this feature, but not populating `RUBY_GC_LIBRARY_PATH` will incur a small performance penalty due to the overhead of using `getenv` to check for the environment variable.

The use of `dln.c` rather than using `dlopen` directly, should enable this feature to be supported on all platforms that support loading shared libraries, rather than the initial implementation, which explicitly only supported Linux and MacOS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,7 @@ m4_include([tool/m4/ruby_replace_type.m4])dnl
 m4_include([tool/m4/ruby_require_funcs.m4])dnl
 m4_include([tool/m4/ruby_rm_recursive.m4])dnl
 m4_include([tool/m4/ruby_setjmp_type.m4])dnl
+m4_include([tool/m4/ruby_shared_gc.m4])dnl
 m4_include([tool/m4/ruby_stack_grow_direction.m4])dnl
 m4_include([tool/m4/ruby_thread.m4])dnl
 m4_include([tool/m4/ruby_try_cflags.m4])dnl
@@ -3750,6 +3751,7 @@ AS_IF([test x"$gcov" = xyes], [
 ])
 
 RUBY_SETJMP_TYPE
+RUBY_SHARED_GC
 }
 
 [begin]_group "installation section" && {
@@ -4656,6 +4658,7 @@ config_summary "target OS"           "$target_os"
 config_summary "compiler"            "$CC"
 config_summary "with thread"         "$THREAD_MODEL"
 config_summary "with coroutine"      "$coroutine_type"
+config_summary "with shared GC"      "$with_shared_gc"
 config_summary "enable shared libs"  "$ENABLE_SHARED"
 config_summary "dynamic library ext" "$DLEXT"
 config_summary "CFLAGS"              "$cflags"
--- a/dln.h
+++ b/dln.h
@@ -25,6 +25,7 @@ RUBY_SYMBOL_EXPORT_BEGIN
 char *dln_find_exe_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 char *dln_find_file_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 void *dln_load(const char*);
+void *dln_open(const char *file);
 void *dln_symbol(void*,const char*);
 
 RUBY_SYMBOL_EXPORT_END
--- a/gc.c
+++ b/gc.c
@@ -1884,6 +1884,55 @@ rb_gc_initial_stress_set(VALUE flag)
     initial_stress = flag;
 }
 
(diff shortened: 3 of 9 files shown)


gc.c
```diff
@@ -1884,6 +1884,52 @@ rb_gc_initial_stress_set(VALUE flag)
...
+    char *gc_so_path = getenv("RUBY_GC_LIBRARY_PATH");
+    if (!gc_so_path) {
+        map->init = Alloc_GC_impl;
+        return;
+    }
+
+    void *h = dln_open(gc_so_path);
+    if (!h) {
+        rb_bug(
+            "ruby_external_gc_init: Shared library %s cannot be opened.\n",
```
nobu, line 1908: `rb_bug` appends a newline.
```suggestion
            "ruby_external_gc_init: Shared library %s cannot be opened.",
```

gc.c
```diff
@@ -1884,6 +1884,52 @@ rb_gc_initial_stress_set(VALUE flag)
...
+            gc_so_path
+        );
+    }
+
+    void *gc_init_func = dln_symbol(h, "Init_GC");
+    if (!gc_init_func) {
+        rb_bug("ruby_external_gc_init: Init_GC func not exported by library");
```
nobu, line 1915: The failing shared library path is not needed?

```suggestion
        rb_bug("ruby_external_gc_init: Init_GC func not exported by library %s.",
            gc_so_path);
```

dln.h
```diff
@@ -25,6 +25,7 @@ RUBY_SYMBOL_EXPORT_BEGIN
 char *dln_find_exe_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 char *dln_find_file_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 void *dln_load(const char*);
+void * dln_open(const char *file);
```
nobu, line 28:
```suggestion
void *dln_open(const char *file);
```

configure.ac
```diff
@@ -4657,6 +4659,7 @@ config_summary "compiler"            "$CC"
 config_summary "with thread"         "$THREAD_MODEL"
 config_summary "with coroutine"      "$coroutine_type"
 config_summary "enable shared libs"  "$ENABLE_SHARED"
+config_summary "enable shared GC"    "$gc_support"
```
nobu, line 4662: Why not `$with_shared_gc`?

kateinoigakukun (approved):

tool/m4/ruby_shared_gc.m4
```diff
@@ -0,0 +1,22 @@
...
+AS_IF([test "$with_shared_gc" = "yes"], [
+    AC_MSG_RESULT([yes])
+    gc_support="yes"
+
+    AC_DEFINE([USE_SHARED_GC], [1])
+], [
+    AC_MSG_RESULT([no])
+    gc_support="no"
+
+    AC_DEFINE([USE_SHARED_GC], [0])
```
nobu, line 20: This `gc_support` word feels like GC is not opted out if it is "no".

```suggestion
    AC_MSG_RESULT([yes])
    AC_DEFINE([USE_SHARED_GC], [1])
], [
    AC_MSG_RESULT([no])
    with_shared_gc="no"
    AC_DEFINE([USE_SHARED_GC], [0])
```

configure.ac
```diff
@@ -4656,6 +4658,7 @@ config_summary "target OS"           "$target_os"
 config_summary "compiler"            "$CC"
 config_summary "with thread"         "$THREAD_MODEL"
 config_summary "with coroutine"      "$coroutine_type"
+config_summary "with shared GC"      "$gc_support"
```
nobu, line 4661:
```suggestion
config_summary "with shared GC"      "$with_shared_gc"
```

nobu (changes requested):

PR #10457, opened 2024-04-04
YJIT: Fuse `opt_eq`, `opt_lt`, and `branchunless` for fixnums

XrXr:
This targets the instruction pair that usually result from code like
`if a == b` or `if a < b`.

Fusing the two makes it so we don't need to use csel and then test the
result:

```diff
-# Insn: 0007 opt_eq
+# Insn: 0007 opt_eq (stack_size: 3)
 cmp x9, x10
-mov x11, #0x14
-mov x12, #0
-csel x11, x11, x12, eq
-mov x9, x11
-# Insn: 0009 branchunless
-tst x9, #-5
-b.eq #0x1092ee174
+b.ne #0x10457a174
 nop
-b #0x1092ee1a8
+# gen_direct_jmp: fallthrough
```

By using gen_direct_jmp() for when the branch is not taken, the fusion
also reduces one stub for each fused branch, saving some code size.

```
yjit-pre: ruby 3.4.0dev (2024-04-04T20:59:08Z master 440c63df31) +YJIT [x86_64-linux]
yjit-post: ruby 3.4.0dev (2024-04-09T21:33:39Z yjit-fused-branch e92320519a) +YJIT [x86_64-linux]

----------  -------------  ----------  --------------  ----------  -----------------  ------------------
bench       yjit-pre (ms)  stddev (%)  yjit-post (ms)  stddev (%)  yjit-post 1st itr  yjit-pre/yjit-post
optcarrot   1954.5         0.6         1953.4          0.7         1.00               1.00              
protoboeuf  50.5           1.1         49.2            1.1         1.07               1.03              
activerecord    36.0           1.2         36.3            1.3         0.98               0.99              
chunky-png      615.0          0.4         639.9           0.6         0.96               0.96              
erubi-rails     931.1          0.3         937.2           0.2         0.98               0.99              
hexapdf         1913.2         1.3         1933.0          2.6         0.98               0.99              
liquid-c        56.4           1.4         56.3            1.4         1.00               1.00              
liquid-compile  57.7           1.9         57.2            1.8         0.99               1.01              
liquid-render   76.8           1.3         76.8            1.3         0.99               1.00              
lobsters        860.5          1.3         859.8           1.1         1.06               1.00              
mail            120.4          0.8         120.4           0.7         1.00               1.00              
psych-load      1616.7         0.1         1617.0          0.1         1.00               1.00              
railsbench      1959.9         0.1         1952.9          0.1         1.00               1.00              
rubocop         109.6          5.0         109.5           5.0         1.00               1.00              
ruby-lsp        95.6           2.4         93.9            2.4         1.01               1.02              
sequel          66.5           0.8         66.3            0.7         0.99               1.00              
--------------  -------------  ----------  --------------  ----------  -----------------  ------------------

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -364,6 +364,7 @@ module RubyVM::YJIT
 
       out.puts "branch_insn_count:     " + format_number(13, stats[:branch_insn_count])
       out.puts "branch_known_count:    " + format_number_pct(13, stats[:branch_known_count], stats[:branch_insn_count])
+      out.puts "branch_fused_count:    " + format_number_pct(13, stats[:branch_fused_count], stats[:branch_insn_count])
 
       out.puts "freed_iseq_count:      " + format_number(13, stats[:freed_iseq_count])
       out.puts "invalidation_count:    " + format_number(13, stats[:invalidation_count])
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -171,9 +171,26 @@ impl JITState {
         unsafe { *(self.pc.offset(arg_idx + 1)) }
     }
 
+    // Get an argument of the next instruction
+    pub fn next_insn_arg(&self, arg_idx: usize) -> VALUE {
+        let current_len = insn_len(self.get_opcode()).as_usize();
+        unsafe {
+            let next_pc = self.pc.add(current_len);
+            let next_opcode = rb_iseq_opcode_at_pc(self.iseq, next_pc);
+            assert!(insn_len(next_opcode as usize) as usize > (arg_idx + 1));
+            next_pc.add(arg_idx + 1).read()
+        }
+    }
+
     // Get the index of the next instruction
     fn next_insn_idx(&self) -> u16 {
         self.insn_idx + insn_len(self.get_opcode()) as u16
+    }
+
+    // Get the opcode of the next instruction
+    pub fn next_insn_opcode(&self) -> ruby_vminsn_type {
+        let next_idx = self.next_insn_idx();
+        iseq_opcode_at_idx(self.iseq, next_idx.into())
     }
 
     // Check if we are compiling the instruction at the stub PC
@@ -1392,8 +1409,7 @@ fn fuse_putobject_opt_ltlt(
     constant_object: VALUE,
     ocb: &mut OutlinedCb,
 ) -> Option<CodegenStatus> {
-    let next_opcode = unsafe { rb_vm_insn_addr2opcode(jit.pc.add(insn_len(jit.opcode).as_usize()).read().as_ptr()) };
-    if next_opcode == YARVINSN_opt_ltlt as i32 && constant_object.fixnum_p() {
+    if jit.next_insn_opcode() == YARVINSN_opt_ltlt && constant_object.fixnum_p() {
         // Untag the fixnum shift amount
         let shift_amt = constant_object.as_isize() >> 1;
         if shift_amt > 63 || shift_amt < 0 {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by maximecb** [comment]
Thanks Alan for taking the time to implement this.

Given that there is no measurable performance difference, I'm hesitant to merge it.

Might make sense to inventory the kinds of branches uses on optcarrot and protoboeuf at some point. I would guess protoboeuf uses a lot of less than and greater than comparisons.

**Comment by XrXr** [comment]
Sounds good, I'll close this for now.

> Might make sense to inventory the kinds of branches uses on optcarrot and protoboeuf at some point. I would guess protoboeuf uses a lot of less than and greater than comparisons.

I added counting on top of this branch for fusable fixnum comparisons by type. `branch_fused_count` is for `==`. In `protoboeuf`:

```
branch_insn_count:               267
branch_known_count:                0 ( 0.0%)
branch_fused_count:               97 (36.3%)
branch_fuse_lt:               94 (35.2%)
branch_fuse_le:                0 ( 0.0%)
branch_fuse_gt:                0 ( 0.0%)
branch_fuse_ge:               65 (24.3%)
```

In `optcarrot`:

```
branch_insn_count:               311
branch_known_count:                9 ( 2.9%)
branch_fused_count:               40 (12.9%)
branch_fuse_lt:               31 (10.0%)
branch_fuse_le:               11 ( 3.5%)
branch_fuse_gt:               13 ( 4.2%)
branch_fuse_ge:               15 ( 4.8%)
```

This is by sites, not by how many times they run, though. Kind of interesting that `==` is the most common.

XrXr:
With a tweak to the code layout and also fusion for less than, there is now a measurable boost to `protoboeuf`. I've updated the top description, please take another look. There is maybe a slowdown for `chunky-png`, I guess due to always having code for the if branch in the fallthrough path.

I'll check if there is a bug with generating fallthrough branches later.

maximecb:
Thanks for continuing the investigation Alan 🙏 

> I guess due to always having code for the if branch in the fallthrough path.

Can you explain what you mean by having code for the if branch in the fallthrough path?

XrXr:
> Can you explain what you mean by having code for the if branch in the fallthrough path?

Normally for branch instructions we generate two stubs, one for when the branch is taken and one for the not taken path (fallthrough). Once the branch runs for the first time, we then put code for the fallthrough or the taken path directly after the branch depending how it went. With this PR, though, I always put the code for the fallthrough path after the branch. With how `branchunless` is used that means code from the truthy side of fused ifs get put right after the branch.

maximecb:
> With this PR, though, I always put the code for the fallthrough path after the branch. With how branchunless is used that means code from the truthy side of fused ifs get put right after the branch.

Oh, that does seem very suboptimal. The linearization of code is one of the biggest strengths of BBV. AFAIK branch predictors love it because we largely do their job for them. We essentially structure the machine code so that the fallthrough path is (almost always) the most likely path... So if there is no more space in the branch predictor caches, BTB, whatever system is used, the default behavior is so much more likely to be the right prediction.

Do you think there is a way to implement this patch whilst not automatically putting the truthy side after, but continuing to use the previous stub behavior?

XrXr:
> Do you think there is a way to implement this patch whilst not automatically putting the truthy side after, but continuing to use the previous stub behavior?

That's how I had the patch originally, but that introduces a performance regression on optcarrot because that adds mispredictions for some reason. 😅 

I've double checked that the normal layout strategy is indeed working as intended.

I guess while the fusion does give a boost to protoboeuf, it just triggers too much uarch weirdness to be worth it. Even with the biased way I have it at the moment it doesn't work well for all cases (chunkypng). The gains from the fusion is not enough to offset the branch prediction weirdness unfortunately. Will give up on this for now.

PR #10467, opened 2024-04-05
Enable RUBY_DEBUG on --enable-debug-env

peterzhu2118:
The documentation for --enable-debug-env says:

    enable RUBY_DEBUG environment variable

However, it only seems to enable RUBY_DEBUG_ENV for main.c, and RUBY_DEBUG_ENV seems to do nothing. This commit changes it to set -DRUBY_DEBUG=1 in CPPFLAGS.


nobu:
> The documentation for --enable-debug-env says:
> 
> ```
> enable RUBY_DEBUG 
> ```
> 
> However, it only seems to enable RUBY_DEBUG_ENV for main.c, and RUBY_DEBUG_ENV seems to do nothing. This commit changes it to set -DRUBY_DEBUG=1 in CPPFLAGS.

Yes, `RUBY_DEBUG_ENV` is used only in main.c and the environment variable is passed from `main`, so it doesn't make sense for the other files.

Isn't `RUBY_DEBUG` macro what you want to use?

peterzhu2118:
Ah I see, I mistook RUBY_DEBUG environment variable for RUBY_DEBUG macro.

PR #10469, opened 2024-04-05
Fix nil error with KNOWNBUGS.rb

XrXr:
Previously, `make test-knownbugs` crashed with `NoMethodError` due to the failed regex match if there is a test case in it.

The note about 1.8 compatibility is probably bogus as we require a way more recent BASERUBY now.

### DIFF
--- a/bootstraptest/runner.rb
+++ b/bootstraptest/runner.rb
@@ -6,7 +6,6 @@
 # Never use optparse in this file.
 # Never use test/unit in this file.
 # Never use Ruby extensions in this file.
-# Maintain Ruby 1.8 compatibility for now
 
 $start_time = Time.now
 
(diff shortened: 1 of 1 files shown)


nobu:
> Previously, `make test-knownbugs` crashed with `NoMethodError` due to the failed regex match if there is a test case in it.

What file did you add?

bootstraptest/runner.rb
```diff
@@ -428,7 +427,7 @@ def add as
   def initialize(*args)
     super
     self.class.add self
-    @category = self.path.match(/test_(.+)\.rb/)[1]
+    @category = self.path.match(/test_(.+)\.rb/)&.[](1)
```
nobu, line 430:
```suggestion
    @category = self.path[/\Atest_(.+)\.rb\z/, 1]
```

XrXr:
> What file did you add?

I didn't add a file, just added a test case like `assert_equal "1", "1"` to the end of KNOWNBUGS.rb.

PR #10474, opened 2024-04-06
Don't set T_TYPES of NODE

yui-knk:
T_TYPES was needed once Ripper jumbled NODEs and other type objects. However such hack was already removed.
Therefore don't need to set T_TYPES of NODE.


nobu (approved):

yui-knk:
Thanks for the review.

PR #2002, opened 2018-11-03
Enumerator fiber yield

ioquatix:
This makes it so that it is possible to call `Fiber.yield` within a Enumerator block.

This is breaking `async` when non-blocking IO is used in an Enumerator: https://github.com/socketry/async/issues/23

I'm not sure if this is the best solution, but it feels like the right approach.

The general idea is that user should not worry about how Enumerator is implemented, `Fiber.yield` should work as expected.

PR #10478, opened 2024-04-07
Enumerator should use a non-blocking fiber.

ioquatix:
Some usage of `Enumerator` can be problematic in the event loop, e.g. calling `sleep`.

See <https://github.com/ruby/ruby/pull/2002> for background context.

This PR allows the fiber scheduler to transfer on blocking operations within an Enumerator's block.


ioquatix:
I have no idea how the parser change slipped in, I'm going to revert this and re-apply it.

PR #10481, opened 2024-04-07
Enumerator should use a non-blocking fiber, change `rb_fiber_new` to be non-blocking by default.

ioquatix:
This re-applies <https://github.com/ruby/ruby/pull/10478> sans the `parse.h` change.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2382,7 +2382,7 @@ rb_fiber_initialize(int argc, VALUE* argv, VALUE self)
 VALUE
 rb_fiber_new_storage(rb_block_call_func_t func, VALUE obj, VALUE storage)
 {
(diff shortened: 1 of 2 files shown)


eregon:
This changes `rb_fiber_new` which [is public C API](https://github.com/ruby/ruby/blob/b473d304d449b520a967e45b7d5d5ecb8556b4c3/include/ruby/internal/intern/cont.h#L39) to always use a non-blocking fiber.
That seems incompatible.

eregon (dismissed):
See my comment

eregon:
I disabled the auto-merge since this incompatibility seems worth thinking about before merging.

ioquatix:
Can you show where it's incompatible?

eregon:
If it's already the behavior of Fiber.new then it sounds fair to change rb_fiber_new() too (which should have been changed at the same time ideally).

ioquatix:
Appreciate the review.

PR #10482, opened 2024-04-07
`Fiber#raise` recursively raises on nested resuming_fiber.

ioquatix:
https://bugs.ruby-lang.org/issues/20414

### DIFF
--- a/spec/ruby/core/fiber/raise_spec.rb
+++ b/spec/ruby/core/fiber/raise_spec.rb
@@ -99,13 +99,9 @@ describe "Fiber#raise" do
       f2 = Fiber.new { f1.resume }
       f2.transfer
 
-      error = nil
-      begin
-        f2.raise(RuntimeError)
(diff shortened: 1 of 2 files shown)


spec/ruby/core/fiber/raise_spec.rb
```diff
@@ -91,6 +91,23 @@
...
+      f2 = Fiber.new { f1.resume }
+      f2.transfer
+
+      error = nil
+      begin
+        f2.raise(RuntimeError)
+      rescue RuntimeError => error
+      end
+
+      error.should be_kind_of(RuntimeError)
```
eregon, line 108: `-> { ... }.should raise_error(RuntimeError, "message")`.
Would also be good to give a message to `Fiber#raise`  to not accidentally catch another RuntimeError.

ioquatix: Thanks, that's a good improvement.

PR #10493, opened 2024-04-10
put empty `rb_gc_force_recycle()`

ko1:
and declare it will be removed soon.

ddtrace is still referes the API and build was failed. See https://github.com/DataDog/dd-trace-rb/pull/3578

Maybe threre are only few users of this C-API now so we can remove it soon.

### DIFF
--- a/include/ruby/internal/gc.h
+++ b/include/ruby/internal/gc.h
@@ -823,4 +823,7 @@ rb_obj_write(
     return a;
 }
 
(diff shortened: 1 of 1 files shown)


ivoanjo:
Hey! https://github.com/DataDog/dd-trace-rb/pull/3578 has been released in version 1.22.0 so all good from ddtrace to remove it :)

PR #10503, opened 2024-04-11
prism_compile.c: X_STRING should be frozen

casperisfine:
The backtick method receives a frozen string unless it is interpolated.

Otherwise the string held in the ISeq could be mutated by a custom backtick method.

cc @kddnewton

cc @eregon (not 100% sure how best to spec this)


eregon:
The specs look fine. Ideally there would be a way to use `should_receive(:`).with('string').and { |arg,| arg.frozen? }` or so but that `and {}` doesn't exist currently.

kddnewton (approved):
Thanks @casperisfine!

PR #10504, opened 2024-04-11
prism_compile.c: use rb_enc_interned_str to reduce allocations

casperisfine:
The `rb_fstring(rb_enc_str_new())` pattern is inefficient because:

- It passes a mutable string to `rb_fstring` so if it has to be interned it will first be duped.
- It an equivalent interned string already exists, we allocated the string for nothing.

With `rb_enc_interned_str` we either directly get the pre-existing string with 0 allocations, or efficiently directly intern the one we create without first duping it.

Prism version of https://github.com/ruby/ruby/pull/10494

cc @kddnewton @etiennebarrie

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -289,6 +289,35 @@ parse_string_encoded(const pm_scope_node_t *scope_node, const pm_node_t *node, c
     return rb_enc_str_new((const char *) pm_string_source(string), pm_string_length(string), encoding);
 }
 
+static inline VALUE
+parse_static_literal_string(rb_iseq_t *iseq, const pm_scope_node_t *scope_node, const pm_node_t *node, const pm_string_t *string)
+{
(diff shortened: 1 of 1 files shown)


kddnewton (approved):
Thanks!

kddnewton:
It's interesting that the objspace test is failing now because the coderange isn't known. Is there a chance these codepaths are wiping out the code ranges?

casperisfine:
> Is there a chance these codepaths are wiping out the code ranges?

No, they're not wiping it, they just never need it. I had to force scanning the code range in the other compiler to fix that test: https://github.com/ruby/ruby/pull/10494/files#diff-3ed41748262cd887644971349a59bea1e993d7207594638f44d8e6ca2d751ae2R746

I'll need to do the same here.

PR #10514, opened 2024-04-12
[Bug #20423] Disallow anonymous block within argument forwarding

nobu:
https://bugs.ruby-lang.org/issues/20423


tenderlove:
Thank you!

PR #10527, opened 2024-04-15
[Universal parser] DeVALUE of p->debug_lines and ast->body.script_lines

hasumikin:
This patch is part of universal parser work.

## Summary
- Decouple VALUE from the members below:
  - `(struct parser_params *)->debug_lines`
  - `(rb_ast_t *)->body.script_lines`
- Instead, they are now `rb_parser_ary_t *`
  - They can also be a `(VALUE)FIXNUM` as before to hold line count
- `ISEQ_BODY(iseq)->variable.script_lines` remains VALUE
  - To do this,
  - Add `VALUE script_lines` param to `rb_iseq_new_with_opt()`
  - Introduce `rb_parser_build_script_lines_from()` to convert `rb_parser_ary_t *` into `VALUE` (`RArray`)

## Other points
- Extend `rb_parser_ary_t *`. It previously could only store `rb_parser_ast_token *`, now can store script_lines, too
- Change tactics of building the top-level `SCRIPT_LINES__` in `yycompile0()`
  - Before: While parsing, each line of the script is added to `SCRIPT_LINES__[path]`
  - After: After `yyparse(p)`, `SCRIPT_LINES__[path]` will be built from `p->debug_lines`
- Remove the second parameter of `rb_parser_set_script_lines()` to make it simple
- Introduce `script_lines_free()` to be called from `rb_ast_free()` because the GC no longer takes care of the script_lines
- Introduce `rb_parser_string_deep_copy()` in parse.y to maintain script_lines when `rb_ruby_parser_free()` called
  - Concerning this, please see *Future tasks* below

## Future tasks
- Decouple IMEMO from `rb_ast_t *`
  - This lifts the five-words-restriction of Ruby object,
  - So we will be able to move the ownership of the `lex.string_buffer` from parser to AST
  - Then we remove `rb_parser_string_deep_copy()` to avoid making duplicated objects

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -834,20 +834,21 @@ rb_iseq_new(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE realpath,
             const rb_iseq_t *parent, enum rb_iseq_type type)
 {
     return rb_iseq_new_with_opt(ast, name, path, realpath, 0, parent,
-                                0, type, &COMPILE_OPTION_DEFAULT);
+                                0, type, &COMPILE_OPTION_DEFAULT,
+                                Qnil);
 }
 
 static int
 ast_line_count(const rb_ast_body_t *ast)
 {
-    if (ast->script_lines == Qfalse) {
+    if (ast->script_lines == NULL) {
         // this occurs when failed to parse the source code with a syntax error
         return 0;
     }
-    if (RB_TYPE_P(ast->script_lines, T_ARRAY)){
-        return (int)RARRAY_LEN(ast->script_lines);
+    if (!FIXNUM_P((VALUE)ast->script_lines)) {
+        return (int)ast->script_lines->len;
     }
-    return FIX2INT(ast->script_lines);
+    return FIX2INT((VALUE)ast->script_lines);
 }
 
 static VALUE
@@ -883,7 +884,8 @@ rb_iseq_new_top(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE realpath
     iseq_new_setup_coverage(path, ast, 0);
 
     return rb_iseq_new_with_opt(ast, name, path, realpath, 0, parent, 0,
-                                ISEQ_TYPE_TOP, &COMPILE_OPTION_DEFAULT);
+                                ISEQ_TYPE_TOP, &COMPILE_OPTION_DEFAULT,
+                                Qnil);
 }
 
 /**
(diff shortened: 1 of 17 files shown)


rubyparser.h
```diff
@@ -219,8 +219,14 @@ typedef struct rb_parser_ast_token {
...
  */
+enum rb_parser_ary_entry_type {
+    PARSER_ARY_ENTRY_AST_TOKEN,
+    PARSER_ARY_ENTRY_SCRIPT_LINE
+};
+
 typedef struct rb_parser_ary {
-    rb_parser_ast_token_t **data;
+    enum rb_parser_ary_entry_type entry_type;
+    void **data;
```
yui-knk, line 229: It might be better to give a name to `void *` like `typedef void* rb_parser_ary_data`.

iseq.c
```diff
@@ -974,13 +979,8 @@ rb_iseq_new_with_opt(const rb_ast_body_t *ast, VALUE name, VALUE path, VALUE rea
...
     }
 
-    VALUE script_lines = Qnil;
-
-    if (ast && !FIXNUM_P(ast->script_lines) && ast->script_lines) {
-        script_lines = ast->script_lines;
-    }
-    else if (parent) {
-        script_lines = ISEQ_BODY(parent)->variable.script_lines;
+    if (NIL_P(script_lines) && ast && ast->script_lines && !FIXNUM_P((VALUE)ast->script_lines)) {
```
yui-knk, line 982: Do we need to check `!FIXNUM_P(script_lines)` too?
By the way, is it possible to make the call of this function always pass meaningful value for `VALUE script_lines`? If so, no need to check

rubyparser.h
hasumikin: Also, I fixed ambiguity of `entry` and `data` by using `data`

iseq.c
hasumikin: I reverted the possible over-corrected code

node.c
```diff
@@ -344,18 +345,26 @@ iterate_node_values(rb_ast_t *ast, node_buffer_list_t *nb, node_itr_t * func, vo
...
-void
-rb_ast_mark_and_move(rb_ast_t *ast, bool reference_updating)
+static void
+script_lines_free(rb_ast_t *ast, rb_parser_ary_t *script_lines)
 {
-    if (ast->node_buffer) {
-        if (ast->body.script_lines) rb_gc_mark_and_move(&ast->body.script_lines);
+    for (long i = 0; i < script_lines->len; i++) {
+        rb_parser_string_t *str = (rb_parser_string_t *)script_lines->data[i];
+        xfree(str->ptr);
```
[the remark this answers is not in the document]
hasumikin: Got it, thanks!

yui-knk:
Thank you!

PR #10530, opened 2024-04-15
Specify Kernel#autoload? uses current namespace

gmcgibbon:
Related to https://bugs.ruby-lang.org/issues/20411

Because `Kernel#autoload?` uses the current namespace, it can lead to potentially confusing results. We should make it clearer that modules count as separate namespaces to lookup in.

cc @jeremyevans

### DIFF
--- a/load.c
+++ b/load.c
@@ -1532,7 +1532,7 @@ rb_f_autoload(VALUE obj, VALUE sym, VALUE file)
  *     autoload?(name, inherit=true)   -> String or nil
  *
  *  Returns _filename_ to be loaded if _name_ is registered as
- *  +autoload+ in the current namespae or one of its ancestors.
+ *  +autoload+ in the current namespace or one of its ancestors.
  *
  *     autoload(:B, "b")
  *     autoload?(:B)            #=> "b"
--- a/load.c
+++ b/load.c
@@ -1532,10 +1532,22 @@ rb_f_autoload(VALUE obj, VALUE sym, VALUE file)
  *     autoload?(name, inherit=true)   -> String or nil
  *
  *  Returns _filename_ to be loaded if _name_ is registered as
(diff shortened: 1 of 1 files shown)


load.c
```diff
@@ -1532,10 +1532,16 @@ rb_f_autoload(VALUE obj, VALUE sym, VALUE file)
  *     autoload?(name, inherit=true)   -> String or nil
  *
  *  Returns _filename_ to be loaded if _name_ is registered as
- *  +autoload+.
+ *  +autoload+ in the current namespae or one of its ancestors.
```
nobu, line 1535:
```suggestion
 *  +autoload+ in the current namespace or one of its ancestors.
```

load.c
```diff
@@ -1532,10 +1532,16 @@ rb_f_autoload(VALUE obj, VALUE sym, VALUE file)
...
  *  Returns _filename_ to be loaded if _name_ is registered as
- *  +autoload+.
+ *  +autoload+ in the current namespace or one of its ancestors.
  *
  *     autoload(:B, "b")
  *     autoload?(:B)            #=> "b"
+ *
+ *     module C
+ *       autoload(:D, "d")
+ *       autoload?(:D)          #=> "c"
```
jeremyevans, line 1542: The `"c"` should be `"d"`.

Additionally, this doesn't show an example for classes, only for modules.  I recommend adding a class example after the module example:

```ruby
class E
  autoload(:F, "f")
  autoload?(:F)      #=> "f"
  autoload?(:B)      #=> "b"
end
```

gmcgibbon:
Oops, I think I overwrote a commit. Let me restore it

load.c
gmcgibbon: Good idea. Added!

jeremyevans (approved):

PR #10532, opened 2024-04-15
Emit a performance warning when redefining specially optimized methods

casperisfine:
This makes it easier to notice a dependency is causing interpreter or JIT deoptimization.

```ruby
Warning[:performance] = true

class String
  def freeze
    super
  end
end
```

```
./test.rb:4: warning: Redefining 'String#freeze' disable multiple interpreter and JIT optimizations
```

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -2132,6 +2132,12 @@ rb_vm_check_redefinition_opt_method(const rb_method_entry_t *me, VALUE klass)
         if (st_lookup(vm_opt_method_def_table, (st_data_t)me->def, &bop)) {
             int flag = vm_redefinition_check_flag(klass);
             if (flag != 0) {
+                rb_category_warn(
+                    RB_WARN_CATEGORY_PERFORMANCE,
+                    "Redefining '%s#%s' disables interpreter and JIT optimizations",
+                    rb_class2name(me->owner),
+                    rb_id2name(me->called_id)
+                );
                 rb_yjit_bop_redefined(flag, (enum ruby_basic_operators)bop);
                 rb_rjit_bop_redefined(flag, (enum ruby_basic_operators)bop);
                 ruby_vm_redefined_flag[bop] |= flag;
--- a/test/ruby/test_optimization.rb
+++ b/test/ruby/test_optimization.rb
@@ -16,6 +16,18 @@ class TestRubyOptimization < Test::Unit::TestCase
     end;
   end
 
+  def assert_performance_warning(klass, method)
+    assert_in_out_err([], "#{<<-"begin;"}\n#{<<~"end;"}", [], ["-:4: warning: Redefining '#{klass}##{method}' disables interpreter and JIT optimizations"])
+    begin;
+      Warning[:performance] = true
+      class #{klass}
+        undef #{method}
+        def #{method}
+        end
+      end
+    end;
+  end
+
   def disasm(name)
     RubyVM::InstructionSequence.of(method(name)).disasm
   end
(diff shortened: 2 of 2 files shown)


eregon:
> ./test.rb:4: warning: Redefining 'String#freeze' disable multiple interpreter and JIT optimizations

I would suggest to replace `multiple` by <nothing> or `several` or `some` to avoid confusion.
On my first read I thought it disabled "multiple interpreter" (i.e. MVM) optimizations.

Also `disable` -> `disables`

casperisfine:
@eregon done.

eregon (approved):

vm.c
```diff
@@ -2132,6 +2132,14 @@ rb_vm_check_redefinition_opt_method(const rb_method_entry_t *me, VALUE klass)
...
             int flag = vm_redefinition_check_flag(klass);
             if (flag != 0) {
+                if (rb_warning_category_enabled_p(RB_WARN_CATEGORY_PERFORMANCE)) {
+                    rb_category_warn(
+                        RB_WARN_CATEGORY_PERFORMANCE,
+                        "Redefining '%s#%s' disables interpreter and JIT optimizations",
+                        rb_class2name(me->owner),
+                        rb_id2name(me->called_id)
+                    );
+                }
```
nobu, line 2142: Checking `rb_warning_category_enabled_p` looks redundant.

```suggestion
                rb_category_warn(
                    RB_WARN_CATEGORY_PERFORMANCE,
                    "Redefining '%s#%s' disables interpreter and JIT optimizations",
                    me->owner, ID2SYM(me->called_id));
```

k0kubun (approved):

ko1:
@byroot 
I don't have strong opinion but could you make a ticket about it and NEWS entry?
I'll confirm it tomorrow's devmeeting by asking matz if the ticket is avaiable.

casperisfine:
> could you make a ticket about it and NEWS entry?

Ah sorry, I went a bit too fast indeed. Here's the ticket and `NEWS` update:

https://bugs.ruby-lang.org/issues/20429

https://github.com/ruby/ruby/pull/10542

ko1:
Thank you!

headius:
These should not be tested as part of normal behavioral tests for these methods. This performance warning is specific to CRuby's way of optimizing and other implementations may not have the same limitation. Having these tests be done alongside normal behavioral testing means all implementations must warn for redefinition regardless of whether performance optimizations are affected.

These tests are also not a core behavior of any of these methods, and are a side effect of one implementation. They should be tested separately.

eregon:
These are CRuby tests, it shouldn't be surprising they test CRuby-specific details.
It's good these new warnings are tested, and where else than `test/ruby/test_optimization.rb` which already has a list of such intrinsified methods?
It should be very easy for JRuby to define `assert_performance_warning` as no-op if it doesn't want to emit such warnings or test that there.

FWIW TruffleRuby has very similar warnings although for a slightly different set of methods:
https://github.com/oracle/truffleruby/blob/master/spec/truffle/redefining_optimized_core_methods_spec.rb

headius:
> It should be very easy for JRuby to define assert_performance_warning

JRuby does not patch tests from MRI. We run them as is without modification and submit patches back when there are MRI specific expectations such as using RubyVM or testing for MRI specific behavior.

Is this warning now considered specified Ruby behavior? If so, should it not be tested in ruby/spec?

If it is not specified behavior, why is it tested alongside the ability to redefine these operators, which *is* specified behavior?

This test has been running in JRuby's test suite for many years. Adding this MRI-specific expectation caused it to start failing.

PR #10535, opened 2024-04-15
Introduce rb_parser_dedent_string function

S-H-GAMELINKS:
For reduce rb_fatal function and MEMOVE macro dependency to Universal Parser


yui-knk:
Thanks for PR however this change introduces the dependency from parser to `rb_parser_dedent_string`. This is not heading towards the goal.

S-H-GAMELINKS:
Thanks for your comment.
I changed it to use rb_parser_dedent_string in ruby_parser.c or parse.y depending on whether the `UNIVERSAL_PARSER` is present or not

PR #10551, opened 2024-04-16
Use unsigned long long for object ID

peterzhu2118:
Since unsigned long long are minumum 64 bits, we have at least 10**17 object IDs available, so there is no chance it will overflow.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -818,7 +818,7 @@ typedef struct rb_objspace {
     } flags;
 
     rb_event_flag_t hook_events;
-    VALUE next_object_id;
+    unsigned long long next_object_id;
 
     rb_size_pool_t size_pools[SIZE_POOL_COUNT];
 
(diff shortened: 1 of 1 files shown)


nobu:
It is very unlikely, but not "no chance".

gc.c
```diff
@@ -4546,8 +4546,8 @@ cached_object_id(VALUE obj)
     else {
         GC_ASSERT(!FL_TEST(obj, FL_SEEN_OBJ_ID));
 
-        id = objspace->next_object_id;
-        objspace->next_object_id = rb_int_plus(id, INT2FIX(OBJ_ID_INCREMENT));
+        id = ULL2NUM(objspace->next_object_id);
+        objspace->next_object_id += OBJ_ID_INCREMENT;
```
nobu, line 4550: Needs overflow check?

peterzhu2118: I don't think it's necessary. If we allocated 1000 object IDs every millisecond (i.e. 1,000,000 per second), it would take 15,000 years for it to overflow.

```
(((2**64) / 40) / ((1000 * 60 * 60 * 24) * 1000)) / 365
```

PR #10556, opened 2024-04-17
Fix ignored block warning in weakref

casperisfine:
```
delegate.rb:84: warning: the passed block for 'WeakRef#__getobj__' defined at weakref.rb:44 may be ignored
```

FYI: @ko1


casperisfine:
Apparently @nobu is already on it. Closing.

PR #10558, opened 2024-04-17
Separate `docs` task from `make all`

hsbt:
[Feature #20347]

### DIFF
--- a/doc/contributing/building_ruby.md
+++ b/doc/contributing/building_ruby.md
@@ -96,10 +96,19 @@
 5. Build Ruby:
 
     ``` shell
-    make install
+    make
     ```
 
 6. [Run tests](testing_ruby.md) to confirm your build succeeded.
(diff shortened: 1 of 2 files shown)


doc/contributing/building_ruby.md
```diff
@@ -96,11 +96,20 @@
...
 6. [Run tests](testing_ruby.md) to confirm your build succeeded.
 
+7. Install Ruby:
+
+    ``` shell
+    make install
+    ```
+
+    - If you need to run `make install` with `sudo` and want to avoid document generation with different permissions, you can use
+    `make docs` before `make install` to generate the documentation.
```
nobu, line 111: `make SUDO=sudo install` would not generate not-owned files.

hsbt: I don't know that!

PR #10559, opened 2024-04-17
relax unused block warning for duck typing

ko1:
if a method `foo` uses a block, other (unrelated) method `foo` can receives a block. So try to relax the unused block warning condition.

```ruby
      class C0
        def f = yield
      end

      class C1 < C0
        def f = nil
      end

      [C0, C1].f{ block } # do not warn
```

### DIFF
--- a/test/ruby/test_method.rb
+++ b/test/ruby/test_method.rb
@@ -1692,5 +1692,19 @@ class TestMethod < Test::Unit::TestCase
       assert_match(/-:23: warning.+f5/, err.join)
       assert_match(/-:24: warning.+f6/, err.join)
     end
+
+    assert_in_out_err '-w', <<-'RUBY' do |_out, err, _status|
+      class C0
+        def f = yield
+      end
+
+      class C1 < C0
+        def f = nil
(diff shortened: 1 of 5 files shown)


test/ruby/test_method.rb
```diff
@@ -1692,5 +1692,19 @@ def f6 = super{}       # zsuper / unuse
...
+    assert_in_out_err '-w', <<-'RUBY' do |_out, err, _status|
+      class C0
+        def f = yield
+      end
+
+      class C1 < C0
+        def f = nil
+      end
+
+      C1.new.f{} # do not warn on duck typing
```
eregon, line 1705: Shouldn't this warn?
Because this call site never uses C0#f, and even more C0#f has never been called.
So I think is warning too little.

From your example at https://bugs.ruby-lang.org/issues/15554#note-29 I thought the logic is "if at a call site we noticed one of the called methods accepts a block, then we don't warn if later methods don't accept a block". I think that's fair, but it is of course order-dependent and would still warn if the method not accepting a block is called first.

PR #10568, opened 2024-04-17
YJIT: Fix canary crash with Array#<<

XrXr:
Previously, we got "We are killing the stack canary set by opt_ltlt"
from `$./miniruby --yjit-call-threshold=1 -e 'a = [].freeze; a << 1'`

Found by running ruby-spec with yjit-call-threshold=1.


k0kubun (approved):
:+1:

PR #10572, opened 2024-04-18
Refer autoconfigured endian macro

nobu:
Remove the case `RB_IO_BUFFER_HOST_ENDIAN` is not defined.

`REG_DWORD` on Windows is unconditionally equal to `REG_DWORD_LITTLE_ENDIAN`, which probably would have been added for symmetry with `REG_DWORD_BIG_ENDIAN` not for portability.


ioquatix (approved):
LGTM, thanks.

Just to confirm, the compiler provides `WORDS_BIGENDIAN`?

nobu:
`WORDS_BIGENDIAN` is defined by `AC_C_BIGENDIAN`, a macro of autoconf.

ioquatix:
Does every windows build use autoconf?

nobu:
Windows builds do not use autoconf, then `WORDS_BIGENDIAN` is kept undefined.
I don't think big-endian Windows has ever existed.

nobu:
> I don't think big-endian Windows has ever existed.

So `REG_DWORD` is equal to `REG_DWORD_LITTLE_ENDIAN`.

PR #10575, opened 2024-04-19
YJIT: Fix String#setbyte crashing for converted arguments

XrXr:
Previously, passing objects that respond to #to_int to `String#setbyte`
resulted in a crash when compiled by YJIT. This was due to the lazily
pushed frame from rb_yjit_lazy_push_frame() lingering and not being
popped by an exception as expected.

Since the lazy frame push infrastructure doesn't work when the C
function calls a method, we need to restore fixnum guards to avoid
crashing.

Found running `ruby/spec`.


k0kubun:
> This was due to the lazily pushed frame from rb_yjit_lazy_push_frame() lingering and not being popped by an exception as expected.

The added test doesn't raise an exception, does it? 

In the original PR https://github.com/ruby/ruby/pull/10080, the intention was to leave the un-popped frame during the C call and let JIT code resurrect the correct frame using the CFP register. But I guess `ec->cfp` could be used again before JIT code fixes it on `leave`?

A potentially more efficient way to fix this might be to just pop a frame after a lazy frame push. I used that trick in https://github.com/ruby/ruby/pull/10165 (see `WITH_YJIT_LAZY_FRAME`) while I ended up not merging it. Popping it adds overhead, but if the slow path is never called on our benchmarks, maybe we can do that for `setbyte`? Then we won't have the overhead of `guard_two_fixnums`.

maximecb:
Does this mean we need to be more careful about lazy push frame usage?

> pop a frame after a lazy frame push.

It would be nice to have a more efficient solution. How would this work? Would only the C code have to care about this?

XrXr:
> But I guess ec->cfp could be used again before JIT code fixes it on leave?

Yes. YJIT ends up looking it at in a branch stub 😅 

> How would this work? Would only the C code have to care about this?

Yes. Basically we need to pop the lazy frame after the function finishes making a call. I'll try poking at that later today.

XrXr:
Okay, restoring the CFP for the non-rising path seems to work (at least locally). Moving forward, we just need to be careful when lazily pushing frames in the C side.

object.c
```diff
@@ -3236,15 +3236,22 @@ ALWAYS_INLINE(static VALUE rb_to_integer_with_id_exception(VALUE val, const char
 static inline VALUE
 rb_to_integer_with_id_exception(VALUE val, const char *method, ID mid, int raise)
 {
+    // when !raise, we need to pop the lazily pushed frame.
```
k0kubun, line 3239: It seems to be popped on L3254 when `raise` too. But is the case unreachable because `try_to_int` raises?

k0kubun (approved):
Thank you!

object.c
XrXr: I edited the comment to clarify that it's not really related to the parameter, but whether an exception is raised (we need to pop if the function returns normally)

PR #10581, opened 2024-04-19
Consolidate redefined the method warning

casperisfine:
Currently redefining a method doesn't emit one but two warnings. One at the location of the new method, and one at the location of the old method.

I believe this is not ideal because when collecting warnings via a custom `Warning.warn`, it has to be pieced together. It's even more tricky because the second part may or may not be emitted depending on whether the original method has an associated ISeq.

I think it's much better to emit a single warning with all the information in one go.

### DIFF
--- a/vm_method.c
+++ b/vm_method.c
@@ -1015,7 +1015,6 @@ rb_method_entry_make(VALUE klass, ID mid, VALUE defined_class, rb_method_visibil
             old_def->type != VM_METHOD_TYPE_ALIAS) {
             const rb_iseq_t *iseq = 0;
 
-            rb_warning("method redefined; discarding old %"PRIsVALUE, rb_id2str(mid));
             switch (old_def->type) {
               case VM_METHOD_TYPE_ISEQ:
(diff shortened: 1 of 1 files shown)


vm_method.c
```diff
@@ -1027,10 +1026,16 @@ rb_method_entry_make(VALUE klass, ID mid, VALUE defined_class, rb_method_visibil
                 break;
             }
             if (iseq) {
-                rb_compile_warning(RSTRING_PTR(rb_iseq_path(iseq)),
-                                   ISEQ_BODY(iseq)->location.first_lineno,
-                                   "previous definition of %"PRIsVALUE" was here",
-                                   rb_id2str(old_def->original_id));
+                rb_warning(
+                    "method redefined; discarding old %"PRIsVALUE", previous definition of %"PRIsVALUE" was at %s:%d",
```
nobu, line 1030: This format can't tag-jump to the previous definition.
Keep two lines with one `Warning.warn` call.

nobu (changes requested):

PR #10582, opened 2024-04-19
Don't warn about missing libyaml if error was a version conflict

adam12:
The error triggered when loading psych may not be due to a missing library but an already activated version of psych which is now conflicting.

Hide the missing dependency warning to prevent any sort of confusion.

### DIFF
--- a/lib/yaml.rb
+++ b/lib/yaml.rb
@@ -2,7 +2,12 @@
 
 begin
   require 'psych'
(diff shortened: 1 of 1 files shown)


nobu:
Please send the PR to https://github.com/ruby/yaml.
It is the upstream and this file will be imported automatically once merged there.

PR #10583, opened 2024-04-19
Use `' '+` instead of words_sep

ydah:
Use parameterizing rules.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -6112,11 +6112,7 @@ regexp		: tREGEXP_BEG regexp_contents tREGEXP_END
                     }
                 ;
 
-words_sep	: ' ' {}
-                | words_sep ' '
-                ;
-
-words		: tWORDS_BEG words_sep word_list tSTRING_END
+words		: tWORDS_BEG ' '+ word_list tSTRING_END
                     {
                         $$ = make_list($3, &@$);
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -6112,11 +6112,7 @@ regexp		: tREGEXP_BEG regexp_contents tREGEXP_END
                     }
                 ;
 
-words_sep	: ' ' {}
```
yui-knk, line 6115: Another option is defining `words_sep` like below. @nobu Is there any preference?

```c
words_sep : ' '+
          ;
```

yui-knk: JFYI: Generated rules are bit different. The difference is that `words_sep` rule exists or not.

## 1. Keep `words_sep` rule

```
nonempty_list_' ': ' '
                 | nonempty_list_' ' ' '

words_sep: nonempty_list_' '

words: "word list" words_sep word_list "terminator"
```

## 2. Remove `words_sep` rule

```
nonempty_list_' ': ' '
                 | nonempty_list_' ' ' '

words: "word list" nonempty_list_' ' word_list "terminator"
```

nobu: Thank you, it seems fine to expand `words_sep` to `' '+`.

nobu (approved):

PR #10594, opened 2024-04-22
Ignore the result of pthread_kill in ubf_wakeup_thread

jeremyevans:
After an upgrade to Ruby 3.3.0, I experienced reproducible production crashes of the form:

[BUG] pthread_kill: No such process (ESRCH)

This is the only pthread_kill call in Ruby. The result of pthread_kill was previously ignored in Ruby 3.2 and below. Checking the result was added in be1bbd5b7d40ad863ab35097765d3754726bbd54 (MaNy).

I have not yet been able to create a minimal self-contained example, but it should be safe to remove the checks.

### DIFF
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -2569,10 +2569,7 @@ ubf_wakeup_thread(rb_thread_t *th)
 {
     RUBY_DEBUG_LOG("th:%u thread_id:%p", rb_th_serial(th), (void *)th->nt->thread_id);
 
-    int r = pthread_kill(th->nt->thread_id, SIGVTALRM);
-    if (r != 0) {
-        rb_bug_errno("pthread_kill", r);
-    }
+    pthread_kill(th->nt->thread_id, SIGVTALRM);
 }
 
 static void


nobu:
No clue by enabling `USE_RUBY_DEBUG_LOG`?

kosaki:
LGTM

jeremyevans:
> No clue by enabling `USE_RUBY_DEBUG_LOG`?

I haven't had a chance to do that.  `ESRCH` means the thread id given is invalid. Maybe it already exited and just needs to be joined, or maybe this an an OpenBSD-specific issue.  In discussion with @ko1, he said the result of `pthread_kill` can be ignored.

kosaki:
> > No clue by enabling `USE_RUBY_DEBUG_LOG`?
> 
> I haven't had a chance to do that. `ESRCH` means the thread id given is invalid. Maybe it already exited and just needs to be joined, or maybe this an an OpenBSD-specific issue. In discussion with @ko1, he said the result of `pthread_kill` can be ignored.

I don't think this is OpenBSD-specific. But you know, OpenBSD's userland thread scheduler often find some thread timing issue. That's just their intention. I really respect them.

PR #10595, opened 2024-04-22
Use ppc64le coroutines on powerpc64-freebsd*

pkubaj:
Only one ractor-related test fails, but it also fails with ucontext.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -2731,6 +2731,9 @@ AS_CASE([$coroutine_type], [yes|''], [
         [aarch64-freebsd*], [
             coroutine_type=arm64
         ],
+        [powerpc64-freebsd*], [
+            coroutine_type=ppc64le
+        ],
         [powerpc64le-freebsd*], [
             coroutine_type=ppc64le
         ],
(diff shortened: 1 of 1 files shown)


pkubaj:
FreeBSD uses ELFv2 ABI on both powerpc64 and powerpc64le. On powerpc64 all the tests pass on both ucontext and ppc64le. On powerpc64le both ucontext and ppc64le fail at one ractor test.

pkubaj:
Also look at e.g. Chimera Linux, which also forces ppc64le coroutines on ppc64: https://github.com/chimera-linux/cports/blob/master/main/ruby/template.py#L47

They also use ELFv2.

barracuda156:
@pkubaj I cannot verify, but I trust you here.

What is the status of ppc (32-bit)? If ABI is the same, then offsets and insns can be set via macros in the same file. (Or which ABI does it use, SysV?)

P. S. On a side note, does FreeBSD use IEEE or IBM long double?

pkubaj:
32-bit ABI is not the same. We use SysV ABI.

FreeBSD uses 64-bit long double.

PR #10596, opened 2024-04-22
Precompute embedded string literals hash code

etiennebarrie:
[[Feature #20415]](https://bugs.ruby-lang.org/issues/20415)

With embedded strings we often have some space left in the slot, which we can use to store the string Hash code.

It's probably only worth it for string literals, as they are the ones likely to be used as hash keys.

We chose to store the Hash code right after the string terminator as to make it easy/fast to compute, and not require one more union in RString.

```
compare-ruby: ruby 3.4.0dev (2024-04-22T06:32:21Z main f77618c1fa) [arm64-darwin23]
built-ruby: ruby 3.4.0dev (2024-04-22T10:13:03Z interned-string-ha.. 8a1a32331b) [arm64-darwin23]
last_commit=Precompute embedded string literals hash code

|            |compare-ruby|built-ruby|
|:-----------|-----------:|---------:|
|symbol      |     39.275M|   39.753M|
|            |           -|     1.01x|
|dyn_symbol  |     37.348M|   37.704M|
|            |           -|     1.01x|
|small_lit   |     29.514M|   33.948M|
|            |           -|     1.15x|
|frozen_lit  |     27.180M|   33.056M|
|            |           -|     1.22x|
|iseq_lit    |     27.391M|   32.242M|
|            |           -|     1.18x|
```

### DIFF
--- a/string.c
+++ b/string.c
@@ -89,6 +89,9 @@ VALUE rb_cSymbol;
  *            another string (the shared root).
  * 3:     STR_CHILLED (will be frozen in a future version)
  *            The string appears frozen but can be mutated with a warning.
+ * 4:     STR_PRECOMPUTED_HASH
+ *            The string is embedded and has its precomputed hascode stored
+ *            after the terminator.
  * 5:     STR_SHARED_ROOT
  *            Other strings may point to the contents of this string. When this
  *            flag is set, STR_SHARED must not be set.
@@ -116,6 +119,7 @@ VALUE rb_cSymbol;
  */
 
 #define RUBY_MAX_CHAR_LEN 16
+#define STR_PRECOMPUTED_HASH FL_USER4
 #define STR_SHARED_ROOT FL_USER5
 #define STR_BORROWED FL_USER6
 #define STR_TMPLOCK FL_USER7
@@ -240,6 +244,11 @@ rb_str_size_as_embedded(VALUE str)
     else {
         real_size = sizeof(struct RString);
     }
+
+    if (FL_TEST_RAW(str, STR_PRECOMPUTED_HASH)) {
+        real_size += sizeof(st_index_t);
+    }
+
     return real_size;
 }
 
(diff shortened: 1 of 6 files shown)


string.c
```diff
@@ -343,9 +348,28 @@ const struct st_hash_type rb_fstring_hash_type = {
...
+    RUBY_ASSERT(!FL_TEST_RAW(str, STR_PRECOMPUTED_HASH));
+    RUBY_ASSERT(STR_EMBED_P(str));
+
+#if RUBY_DEBUG
+    size_t used_bytes = (RSTRING_LEN(str) + TERM_LEN(str));
+    size_t free_bytes = str_embed_capa(str) - used_bytes;
+    RUBY_ASSERT(free_bytes >= sizeof(st_index_t));
+#endif
+
+    *(st_index_t *)(RSTRING_END(str) + TERM_LEN(str)) = rb_str_hash(str);
```
mame, line 363: I am afraid if this may cause a bus error on some platform. Is it safe?

casperisfine: > this may cause a bus error on some platform

Because of alignment? If it's a concern, I think we could ensure it's always aligned, the overhead would be very slightly larger.

mame: Yes > because of alignment

It would be safer to use aligned address if possible.

casperisfine: Thanks, I'll look into it.

nobu:
```suggestion
    typedef struct {char bytes[sizeof(st_index_t)];} unaligned_index;
    union {st_index_t i; unaligned_index b;} u = {.i = rb_str_hash(str)};
    *(unaligned_index *)(RSTRING_END(str) + TERM_LEN(str)) = u.b;
```

casperisfine:
Looks like we broke something after the refactor, I'll investigate.

string.c
```diff
@@ -240,6 +244,11 @@ rb_str_size_as_embedded(VALUE str)
     else {
         real_size = sizeof(struct RString);
     }
+
+    if (FL_TEST_RAW(str, STR_PRECOMPUTED_HASH)) {
+        real_size += sizeof(st_index_t);
+    }
```
casperisfine, line 250: This was the bug. The compactor thought the string could fit in a smaller slot, corrupting the precomputed hash.

string.c
casperisfine: Thank @nobu, interesting trick. I suppose we need it on read as well?

casperisfine:
Alright, I think the bugs have been squashed now.

PR #10600, opened 2024-04-23
Reuse slow path method search for gccct

tenderlove:
This way all code paths use the same search code for finding call caches for a particular method.

### DIFF
--- a/vm_insnhelper.h
+++ b/vm_insnhelper.h
@@ -58,6 +58,14 @@ RUBY_EXTERN rb_serial_t ruby_vm_global_cvar_state;
     VM_REG_CFP = ec->cfp; \
 } while (0)
 
+typedef enum call_type {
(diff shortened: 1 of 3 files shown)


vm_insnhelper.h
```diff
@@ -58,6 +58,15 @@ RUBY_EXTERN rb_serial_t ruby_vm_global_cvar_state;
     VM_REG_CFP = ec->cfp; \
 } while (0)
 
+typedef enum call_type {
+    CALL_PUBLIC,
+    CALL_FCALL,
+    CALL_VCALL,
+    CALL_PUBLIC_KW,
+    CALL_FCALL_KW,
+    CALL_TYPE_MAX
```
ko1, line 67: we can remove it.

ko1 (approved):

PR #10602, opened 2024-04-23
Use user defined parameterizing rules

ydah:
- Use user defined parameterizing rules
- Add DSL support for ripper with `%rule` actions in preproc.rb


nobu:
Not only moving the regexp to a constant, how about this?
https://github.com/nobu/ruby/tree/ripper-dsl-line

ydah:
> Not only moving the regexp to a constant, how about this? https://github.com/nobu/ruby/tree/ripper-dsl-line

It looks good to me. @yui-knk WDYT?

yui-knk:
LGTM!

PR #10608, opened 2024-04-23
Get error from dln_open when USE_SHARED_GC

peterzhu2118:
Before, if dln_open failed to open RUBY_GC_LIBRARY_PATH, it would segfault because it would try to raise an error, which cannot happen because the GC has not been initialized yet.

This commit changes dln_open to return the error that occurred so the caller can handle the error.

### DIFF
--- a/dln.c
+++ b/dln.c
@@ -340,10 +340,9 @@ dln_disable_dlclose(void)
 
 #if defined(_WIN32) || defined(USE_DLN_DLOPEN)
 void *
-dln_open(const char *file)
+dln_open(const char *file, const char **error)
 {
     static const char incompatible[] = "incompatible library version";
-    const char *error = NULL;
     void *handle;
 
 #if defined(_WIN32)
@@ -360,15 +359,15 @@ dln_open(const char *file)
     free(winfile);
 
     if (!handle) {
-        error = dln_strerror();
-        goto failed;
+        *error = dln_strerror();
+        return NULL;
     }
 
 # if defined(RUBY_EXPORT)
     if (!rb_w32_check_imported(handle, rb_libruby_handle())) {
         FreeLibrary(handle);
-        error = incompatible;
-        goto failed;
+        *error = incompatible;
+        return NULL;
     }
 # endif
 
@@ -387,8 +386,8 @@ dln_open(const char *file)
     /* Load file */
     handle = dlopen(file, RTLD_LAZY|RTLD_GLOBAL);
     if (handle == NULL) {
-        error = dln_strerror();
-        goto failed;
+        *error = dln_strerror();
+        return NULL;
     }
 
 # if defined(RUBY_EXPORT)
@@ -413,8 +412,8 @@ dln_open(const char *file)
                 if (libruby_name) {
                     dln_loaderror("linked to incompatible %s - %s", libruby_name, file);
                 }
-                error = incompatible;
-                goto failed;
+                *error = incompatible;
+                return NULL;
             }
         }
     }
@@ -422,9 +421,6 @@ dln_open(const char *file)
 #endif
 
     return handle;
-
-  failed:
-    dln_loaderror("%s - %s", error, file);
 }
 
 void *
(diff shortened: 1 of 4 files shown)


dln.c
```diff
@@ -340,10 +340,9 @@ dln_disable_dlclose(void)
 
 #if defined(_WIN32) || defined(USE_DLN_DLOPEN)
 void *
-dln_open(const char *file)
+dln_open(const char *file, const char **error)
```
nobu, line 343:
```suggestion
dln_open(const char *file, char *message, size_t size)
```

dln.c
```diff
@@ -360,15 +359,15 @@ dln_open(const char *file)
...
+        *error = dln_strerror();
+        return NULL;
     }
 
 # if defined(RUBY_EXPORT)
     if (!rb_w32_check_imported(handle, rb_libruby_handle())) {
         FreeLibrary(handle);
-        error = incompatible;
-        goto failed;
+        *error = incompatible;
```
nobu, line 369:
```suggestion
        strlcpy(message, incompatible, size);
```

dln.c
```diff
@@ -387,8 +386,8 @@ dln_open(const char *file)
     /* Load file */
     handle = dlopen(file, RTLD_LAZY|RTLD_GLOBAL);
     if (handle == NULL) {
-        error = dln_strerror();
-        goto failed;
+        *error = dln_strerror();
```
nobu, line 389: Ditto.

dln.c
```diff
@@ -413,18 +412,15 @@ dln_open(const char *file)
                 if (libruby_name) {
                     dln_loaderror("linked to incompatible %s - %s", libruby_name, file);
                 }
-                error = incompatible;
-                goto failed;
+                *error = incompatible;
```
nobu, line 415: Ditto.

dln.c
```diff
@@ -501,7 +497,12 @@ void *
 dln_load(const char *file)
 {
 #if defined(_WIN32) || defined(USE_DLN_DLOPEN)
-    void *handle = dln_open(file);
+    const char *error = NULL;
+    void *handle = dln_open(file, &error);
```
nobu, line 501:
```suggestion
    char error[1024];
    void *handle = dln_open(file, error, sizeof(error));
```

dln.h
```diff
@@ -25,7 +25,7 @@ RUBY_SYMBOL_EXPORT_BEGIN
 char *dln_find_exe_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 char *dln_find_file_r(const char*,const char*,char*,size_t DLN_FIND_EXTRA_ARG_DECL);
 void *dln_load(const char*);
-void *dln_open(const char *file);
+void *dln_open(const char *file, const char **error);
```
nobu, line 28:
```suggestion
void *dln_open(const char *file, char *error, size_t size);
```

dmydln.c
```diff
@@ -21,9 +21,9 @@ dln_symbol(void *handle, const char *symbol)
     UNREACHABLE_RETURN(NULL);
 }
 
-NORETURN(void *dln_open(const char*));
+NORETURN(void *dln_open(const char *library, const char **error));
 void*
-dln_open(const char *library)
+dln_open(const char *library, const char **error)
```
nobu, line 26:
```suggestion
NORETURN(void *dln_open(const char *library, char *error, size_t size));
void*
dln_open(const char *library, char *error, size_t size)
```

gc.c
```diff
@@ -1900,20 +1900,15 @@ ruby_external_gc_init()
         return;
     }
 
-    void *h = dln_open(gc_so_path);
+    const char *error = NULL;
+    void *h = dln_open(gc_so_path, &error);
```
nobu, line 1904:
```suggestion
    char *error[1024];
    void *h = dln_open(gc_so_path, error, sizeof(error));
```

dln.c
```diff
@@ -413,18 +412,15 @@ dln_open(const char *file)
                 if (libruby_name) {
                     dln_loaderror("linked to incompatible %s - %s", libruby_name, file);
```
nobu, line 413: What about this, and `if (dln_disable_dlclose())` block?

nobu (commented):
And what about `dln_memerror`?

peterzhu2118:
@nobu I opened https://github.com/ruby/ruby/pull/10620 to address the comments.

> And what about dln_memerror?

I only want to replace the common errors in `dln_loaderror`. I think the `dln_memerror` and `dln_fatalerror` are corner-case enough that it's ok for it to crash if it happens with `RUBY_GC_LIBRARY_PATH` enabled.

PR #10610, opened 2024-04-23
Fixes [Bug #20449]

rdimartino:
Bisecting between 3.1.4 and 3.1.5 showed that https://bugs.ruby-lang.org/issues/20449 was introduced here https://github.com/ruby/ruby/commit/4153e807c55ec0fc234969ae08a01c1e0c8ed167

As far as I can tell, this issue only affect Ruby 3.1.5


hsbt:
Thank for file this. Unfortunately, Ruby 3.1.x is security maintenance phase. This means I don't merge normal bugfix.

PR #10612, opened 2024-04-23
Add ruby_mimcalloc

peterzhu2118:
Many places call ruby_mimmalloc then MEMZERO. This can be reduced by using ruby_mimcalloc instead.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -2569,11 +2569,10 @@ rb_fiber_start(rb_fiber_t *fiber)
 void
 rb_threadptr_root_fiber_setup(rb_thread_t *th)
 {
-    rb_fiber_t *fiber = ruby_mimmalloc(sizeof(rb_fiber_t));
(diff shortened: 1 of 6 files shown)


nobu:
If it's called `calloc`, shouldn't it have two arguments?

peterzhu2118:
I changed it to accept two arguments.

cont.c
```diff
@@ -2569,11 +2569,10 @@ rb_fiber_start(rb_fiber_t *fiber)
 void
 rb_threadptr_root_fiber_setup(rb_thread_t *th)
 {
-    rb_fiber_t *fiber = ruby_mimmalloc(sizeof(rb_fiber_t));
+    rb_fiber_t *fiber = ruby_mimcalloc(sizeof(rb_fiber_t), 1);
```
nobu, line 2572:
```suggestion
    rb_fiber_t *fiber = ruby_mimcalloc(1, sizeof(rb_fiber_t));
```

PR #10618, opened 2024-04-24
[Universal parser] Decouple IMEMO from rb_ast_t

hasumikin:
This patch removes the `VALUE flags` member from the `rb_ast_t` structure making `rb_ast_t` no longer an IMEMO object.

## Background

We are trying to make the Ruby parser generated from parse.y a universal parser that can be used by other implementations such as mruby. To achieve this, it is necessary to exclude VALUE and IMEMO from parse.y, AST, and NODE.

## Summary (file by file)

- `rubyparser.h`
  - Remove the `VALUE flags` member from `rb_ast_t`
- `ruby_parser.c` and `internal/ruby_parser.h`
  - Introduce a TypedData_Wrap_Struct VALUE which wraps `rb_ast_t` so that GC can manage it (see ``) - You can retrieve `rb_ast_t` from the VALUE by `rb_ruby_ast_data_get()`
  - Change the return type of `rb_parser_compile_XXXX()` functions from `rb_ast_t *` to `VALUE`
- `iseq.c` and `vm_core.h`
  - Amend the first parameter of `rb_iseq_new_XXXX()` functions from `rb_ast_body_t *` to `VALUE`
  - This keeps the VALUE of AST on the machine stack to prevent being removed by GC
- `ast.c`
  - Almost all change is replacement `rb_ast_t *ast` with `VALUE vast` (sorry for the big diff)
  - Fix `node_memsize()` to improve memory profiling
    - Now it includes `rb_ast_local_table_link`, `tokens` and `script_lines`
- `compile.c`, `load.c`, `node.c`, `parse.y`, `proc.c`, `ruby.c`, `template/prelude.c.tmpl`, `vm.c` and `vm_eval.c`
  - Follow-up due to the above changes
- `imemo.{c|h}`
  - If an object with `imemo_ast` appears, consider it a bug
  - 👀 If you think the related code should be removed in this PR, please tell me

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -16,7 +16,7 @@ static VALUE rb_mAST;
 static VALUE rb_cNode;
 
 struct ASTNodeData {
-    rb_ast_t *ast;
+    VALUE vast;
     const NODE *node;
 };
 
@@ -24,14 +24,16 @@ static void
 node_gc_mark(void *ptr)
 {
     struct ASTNodeData *data = (struct ASTNodeData *)ptr;
-    rb_gc_mark((VALUE)data->ast);
+    rb_gc_mark(data->vast);
 }
 
 static size_t
 node_memsize(const void *ptr)
 {
     struct ASTNodeData *data = (struct ASTNodeData *)ptr;
-    return rb_ast_memsize(data->ast);
+    rb_ast_t *ast = rb_ruby_ast_data_get(data->vast);
+
+    return sizeof(struct ASTNodeData) + rb_ast_memsize(ast);
 }
 
 static const rb_data_type_t rb_node_type = {
@@ -44,22 +46,22 @@ static const rb_data_type_t rb_node_type = {
 static VALUE rb_ast_node_alloc(VALUE klass);
 
 static void
-setup_node(VALUE obj, rb_ast_t *ast, const NODE *node)
+setup_node(VALUE obj, VALUE vast, const NODE *node)
 {
     struct ASTNodeData *data;
 
     TypedData_Get_Struct(obj, struct ASTNodeData, &rb_node_type, data);
-    data->ast = ast;
+    data->vast = vast;
     data->node = node;
 }
 
 static VALUE
-ast_new_internal(rb_ast_t *ast, const NODE *node)
+ast_new_internal(VALUE vast, const NODE *node)
 {
     VALUE obj;
 
     obj = rb_ast_node_alloc(rb_cNode);
-    setup_node(obj, ast, node);
+    setup_node(obj, vast, node);
 
     return obj;
 }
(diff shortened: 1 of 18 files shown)


ast.c
```diff
@@ -93,15 +97,15 @@ ast_s_parse(rb_execution_context_t *ec, VALUE module, VALUE str, VALUE keep_scri
 static VALUE
 rb_ast_parse_str(VALUE str, VALUE keep_script_lines, VALUE error_tolerant, VALUE keep_tokens)
 {
-    rb_ast_t *ast = 0;
+    VALUE vast = Qnil;
```
nobu, line 100: Does this need to be initialized?

ast.c
```diff
@@ -277,10 +281,10 @@ ast_node_node_id(rb_execution_context_t *ec, VALUE self)
     return INT2FIX(nd_node_id(data->node));
 }
 
-#define NEW_CHILD(ast, node) node ? ast_new_internal(ast, node) : Qnil
+#define NEW_CHILD(vast, node) node ? ast_new_internal(vast, node) : Qnil
```
nobu, line 284: I think a conditional expression macro should be parenthesized.

```suggestion
#define NEW_CHILD(vast, node) (node ? ast_new_internal(vast, node) : Qnil)
```

compile.c
```diff
@@ -1478,16 +1478,20 @@ new_child_iseq(rb_iseq_t *iseq, const NODE *const node,
...
+    rb_ast_t *ast = ruby_xcalloc(1, sizeof(rb_ast_t));
+    VALUE vast;
+    rb_ast_body_t body = {
+        .root = node,
+        .frozen_string_literal = -1,
+        .coverage_enabled = -1,
+        .script_lines = NULL
+    };
+    ast->body = body;
+    vast = rb_ruby_ast_new(ast);
```
nobu, line 1490: If `rb_ruby_ast_new` raises an exception, `ast` will be leaked.

nobu (changes requested):
"Allocate then wrap" is bad, because the "wrapping" itself can fail.
Allocate a wrapped pointer, then fill it.

nobu (approved):
Looks good.

One more thing.
The usages of `rb_ruby_ast_new` in ruby_parser.c and the outside seem different.
It may be better to separate allocation function (`rb_ruby_ast_alloc` for example) for ruby_parser.c internal usage,  and to pass `root` and `script_lines` arguments to `rb_ruby_ast_new` outside.

hasumikin:
@nobu 
How about this https://github.com/ruby/ruby/pull/10618/commits/9c3002a3a94e93a2c97f01c34a845858a71b55f9 ?
I wrote `ast_alloc()` instead of `rb_ruby_ast_alloc()` as it's static in ruby_parser.c

nobu:
Looks good.

yui-knk (approved):
LGTM!

PR #10619, opened 2024-04-24
[Bug #20450] Remove rubyarchdir from bootsnap paths

eugeneius:
https://bugs.ruby-lang.org/issues/20450

See also https://github.com/ruby/ruby/pull/10347#issuecomment-2073591775.

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -101,8 +101,11 @@ module Gem::BUNDLED_GEMS
   def self.warning?(name, specs: nil)
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
-    # bootsnap expand `require "csv"` to `require "#{LIBDIR}/csv.rb"`
-    name = feature.delete_prefix(LIBDIR).chomp(".rb").tr("/", "-")
+    # bootsnap expands `require "csv"` to `require "#{LIBDIR}/csv.rb"`,
+    # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
+    name = feature.delete_prefix(ARCHDIR)
+    name.delete_prefix!(LIBDIR)
+    name.tr!("/", "-")
     name.sub!(LIBEXT, "")
     return if specs.include?(name)
     _t, path = $:.resolve_feature_path(feature)
--- a/tool/test_for_warn_bundled_gems/test.sh
+++ b/tool/test_for_warn_bundled_gems/test.sh
@@ -32,6 +32,10 @@ echo "* Show warning with bootsnap"
 ruby test_warn_bootsnap.rb
 echo
 
+echo "* Show warning with bootsnap for gem with native extension"
+ruby test_warn_bootsnap_rubyarchdir_gem.rb
+echo
+
 echo "* Show warning with zeitwerk"
 ruby test_warn_zeitwerk.rb
 echo
--- a/tool/test_for_warn_bundled_gems/test_warn_bootsnap_rubyarchdir_gem.rb
+++ b/tool/test_for_warn_bundled_gems/test_warn_bootsnap_rubyarchdir_gem.rb
@@ -1,0 +1,11 @@
+require "bundler/inline"
+
+gemfile do
+  source "https://rubygems.org"
+  gem "bootsnap", require: false
+end
+
+require 'bootsnap'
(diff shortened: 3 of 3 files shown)


professor:
Thank you for fixing this. I confirmed that making this change fixed Ruby 3.3.1 for us with respect to https://bugs.ruby-lang.org/issues/20450

lib/bundled_gems.rb
```diff
@@ -101,8 +101,9 @@ def self.find_gem(path)
   def self.warning?(name, specs: nil)
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
-    # bootsnap expand `require "csv"` to `require "#{LIBDIR}/csv.rb"`
-    name = feature.delete_prefix(LIBDIR).chomp(".rb").tr("/", "-")
+    # bootsnap expands `require "csv"` to `require "#{LIBDIR}/csv.rb"`,
+    # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
+    name = feature.delete_prefix(ARCHDIR).delete_prefix(LIBDIR).chomp(".rb").tr("/", "-")
```
segiddins, line 106: can the latter methods on this chain be changed to the `!` varieties, so we only make one new string, rather than 4?

mscrivo:
Is there a decent way to patch this fix into existing 3.3.1 release? or are we stuck waiting for the next release?

lib/bundled_gems.rb
eugeneius: Sure, I pushed an update.

In the process I realised that `.chomp(".rb")` is redundant with `.sub!(LIBEXT, "")`, so I removed it.

eugeneius:
Here's the patch we used (assumes a Rails app, no warranty implied etc etc):

<details>

```patch
diff --git a/config/application.rb b/config/application.rb
index e18211a..0017ef6 100644
--- a/config/application.rb
+++ b/config/application.rb
@@ -2,6 +2,8 @@ require_relative "boot"
 
 require "rails/all"
 
+require_relative "../lib/monkey_patches/rubygems_bundled_gems_warning"
+
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
diff --git a/lib/monkey_patches/rubygems_bundled_gems_warning.rb b/lib/monkey_patches/rubygems_bundled_gems_warning.rb
new file mode 100644
index 0000000..ef992ca
--- /dev/null
+++ b/lib/monkey_patches/rubygems_bundled_gems_warning.rb
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+mod = Gem.send(:remove_const, :BUNDLED_GEMS).dup
+Gem::BUNDLED_GEMS = mod
+def mod.warning?(name, specs: nil)
+  # name can be a feature name or a file path with String or Pathname
+  feature = File.path(name)
+  # bootsnap expands `require "csv"` to `require "#{LIBDIR}/csv.rb"`,
+  # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
+  name = feature.delete_prefix(Gem::BUNDLED_GEMS::ARCHDIR)
+  name.delete_prefix!(Gem::BUNDLED_GEMS::LIBDIR)
+  name.tr!("/", "-")
+  name.sub!(Gem::BUNDLED_GEMS::LIBEXT, "")
+  return if specs.include?(name)
+  _t, path = $:.resolve_feature_path(feature)
+  if gem = find_gem(path)
+    return if specs.include?(gem)
+    caller = caller_locations(3, 3)&.find {|c| c&.absolute_path}
+    return if find_gem(caller&.absolute_path)
+  elsif Gem::BUNDLED_GEMS::SINCE[name]
+    gem = true
+  else
+    return
+  end
+
+  return if Gem::BUNDLED_GEMS::WARNED[name]
+  Gem::BUNDLED_GEMS::WARNED[name] = true
+  if gem == true
+    gem = name
+    "#{feature} was loaded from the standard library, but"
+  elsif gem
+    return if Gem::BUNDLED_GEMS::WARNED[gem]
+    Gem::BUNDLED_GEMS::WARNED[gem] = true
+    "#{feature} is found in #{gem}, which"
+  else
+    return
+  end + build_message(gem)
+end
```

</details>

hsbt:
@eugeneius Thanks!

PR #10630, opened 2024-04-25
[Bug #20453] Pointer being freed was not allocated in Regexp timeout

composerinteralia:
https://bugs.ruby-lang.org/issues/20228 started freeing `stk_base` to avoid a memory leak. But `stk_base` is sometimes stack allocated (using `xalloca`), so the free only works if the regex stack has grown enough to hit `stack_double` (which uses `xmalloc` and `xrealloc`).

To reproduce the problem on master and 3.3.1:

```ruby
Regexp.timeout = 0.001
/^(a*)x$/ =~ "a" * 1000000 + "x"'
#=> pointer being freed was not allocated / segfault
```

Some details about this potential fix:

`stk_base == stk_alloc` on [init](https://github.com/ruby/ruby/blob/dde99215f2bc60c22a00fc941ff7f714f011e920/regexec.c#L1153), so if `stk_base != stk_alloc` we can be sure we called [`stack_double`](https://github.com/ruby/ruby/blob/dde99215f2bc60c22a00fc941ff7f714f011e920/regexec.c#L1210) and it's safe to free. It's also safe to free if we've [saved](https://github.com/ruby/ruby/blob/dde99215f2bc60c22a00fc941ff7f714f011e920/regexec.c#L1187-L1189) the stack to `msa->stack_p`, since we do the `stk_base != stk_alloc` check before saving.

This matches the check we do inside [`stack_double`](https://github.com/ruby/ruby/blob/dde99215f2bc60c22a00fc941ff7f714f011e920/regexec.c#L1221)

### DIFF
--- a/regexec.c
+++ b/regexec.c
@@ -4218,7 +4218,8 @@ match_at(regex_t* reg, const UChar* str, const UChar* end,
 
  timeout:
   xfree(xmalloc_base);
-  xfree(stk_base);
+  if (stk_base != stk_alloc || IS_NOT_NULL(msa->stack_p))
+      xfree(stk_base);
   HANDLE_REG_TIMEOUT_IN_MATCH_AT;
 }
 
--- a/test/ruby/test_regexp.rb
+++ b/test/ruby/test_regexp.rb
@@ -1827,6 +1827,17 @@ class TestRegexp < Test::Unit::TestCase
     end;
   end
 
+  def test_bug_20453
+    assert_separately([], "#{<<-"begin;"}\n#{<<-'end;'}")
+    begin;
+      Regexp.timeout = 0.001
+
+      assert_raise(Regexp::TimeoutError) do
+        /^(a*)x$/ =~ "a" * 1000000 + "x"
+      end
+    end;
+  end
+
   def per_instance_redos_test(global_timeout, per_instance_timeout, expected_timeout)
     assert_separately([], "#{<<-"begin;"}\n#{<<-'end;'}")
       global_timeout = #{ EnvUtil.apply_timeout_scale(global_timeout).inspect }


regexec.c
```diff
@@ -4218,7 +4218,8 @@ match_at(regex_t* reg, const UChar* str, const UChar* end,
 
  timeout:
   xfree(xmalloc_base);
-  xfree(stk_base);
+  if (stk_base != stk_alloc || IS_NOT_NULL(msa->stack_p))
```
peterzhu2118, line 4221: Why do we need to check that `stk_base != stk_alloc` here?

composerinteralia: https://github.com/ruby/ruby/blob/dde99215f2bc60c22a00fc941ff7f714f011e920/regexec.c#L1177-L1181

It looks to me like if `stk_base == stk_alloc` then it could be stack allocated?

But maybe we don't need this check as long as we do a `STACK_SAVE` first? I'm not sure if we do that though. Still getting my bearings in this code.

test/ruby/test_regexp.rb
```diff
@@ -1827,6 +1827,17 @@ def test_s_timeout_memory_leak
...
   end
 
+  def test_bug_20453
+    assert_separately([], "#{<<-"begin;"}\n#{<<-'end;'}")
+    begin;
+      Regexp.timeout = 0.001
+
+      assert_raise(Regexp::TimeoutError) do
+        /^(a*)x$/ =~ "a" * 1000000 + "x"
+      end
```
nobu, line 1837: It is possible without changing the global setting.
And `assert_separately` may not be necessary, if you use it for this purpose.

```suggestion
      re = Regexp.new("^(a*)x$", timeout: 0.001)

      assert_raise(Regexp::TimeoutError) do
        re =~ "a" * 1000000 + "x"
      end
```

regexec.c
composerinteralia: Confirmed that if we only check `IS_NOT_NULL(msa->stack_p)` then we end up with a memory leak again.

But if we call `STACK_SAVE` first then it works. Maybe we could even do:

```
  STACK_SAVE;
  xfree(xmalloc_base);
  xfree(msa->stack_p);
```

What do you think?

test/ruby/test_regexp.rb
composerinteralia: Yeah, that makes good sense to me. Thank you!

peterzhu2118: Oops, I merged it before you were able to address this comment. Can you open a followup PR that addresses this?

PR #10633, opened 2024-04-25
[Bug #20452] Obtain the main thread address and size only with glibc

nobu:


thesamesam:
We're also using this patch in Gentoo. Is there any chance it could be merged?

PR #10634, opened 2024-04-25
Do not enable ReadlineInputMethod for dumb terminal

zonuexe:
resolve https://bugs.ruby-lang.org/issues/20454


hsbt:
Can you submit this to https://github.com/ruby/irb instead of this?

PR #10648, opened 2024-04-26
YJIT: Try splitting getlocal/setlocal blocks

k0kubun:
After the take-2 PR https://github.com/ruby/ruby/pull/10607, the same symptom ([timeout](https://github.com/ruby/ruby/actions/runs/8839441222/job/24272812243), [nil on stack](https://github.com/ruby/ruby/actions/runs/8851651227/job/24308678802)) has unfortunately happened.

I have a suspicion that the block needs to start with getlocal/setlocal to invalidate them right away. While we isolate send-ish instructions to invalidate them, some unexpected instructions might trigger the invalidation in the same block.

I'll keep investigating this, but I'd like to leave this code in master so that the CI can tell if this theory is relevant.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2265,6 +2265,12 @@ fn gen_getlocal_generic(
     ep_offset: u32,
     level: u32,
 ) -> Option<CodegenStatus> {
+    // Start the block with this instruction for EP-escape invalidation
+    if level == 0 && !jit.at_current_insn() {
+        defer_compilation(jit, asm, ocb);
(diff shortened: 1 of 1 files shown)


XrXr (approved):

maximecb:
Just some caution that the more we use defer_compilation and split blocks, the more invalidation assumptions we store, the more memory we use. We do want to be somewhat careful with memory usage growth.

k0kubun:
Yeah, I was aware of that constraint. This is more for debugging purposes, and this has been reverted since it turned out to be not relevant.

PR #10656, opened 2024-04-27
YJIT: Fix reference update for `Invariants::no_ep_escape_iseqs`

XrXr:
- **YJIT: Fix reference update for `Invariants::no_ep_escape_iseqs`**
- **YJIT: Stop asserting rb_objspace_markable_object_p()**
- **YJIT: Correct signature of rb_yjit_root_mark()**

The nil-on-stack issue on CI was still reproducible locally by running `make test-all TESTS='-j1 ../test/ruby/test_enumerator.rb' RUN_OPTS='--yjit-call-threshold=1 --yjit-verify-ctx'` in a loop locally. After apply this changeset the crash seem to have gone away.

### DIFF
--- a/yjit.c
+++ b/yjit.c
@@ -1164,20 +1164,14 @@ yjit_root_memsize(const void *ptr)
     return 0; // TODO: more accurate accounting
 }
 
-// GC callback during compaction
-static void
-yjit_root_update_references(void *ptr)
-{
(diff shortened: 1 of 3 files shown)


k0kubun (approved):
> That effectively never updated anything because the callback itself is given an intact reference, so it could update its content, and `rb_gc_location(iseq)` never returned a new address.

I see. Thanks for fixing this :bow:

PR #10660, opened 2024-04-27
[Universal parser] Improve AST structure

hasumikin:
This patch moves `ast->node_buffer->config` to `ast->config` aiming to improve readability of the source.

## Background

We could not add the `config` field to the `rb_ast_t *` before due to the five-word restriction of the IMEMO object. But it is now doable by merging https://github.com/ruby/ruby/pull/10618


yui-knk:
IIRC, `ifdef UNIVERSAL_PARSER` branch in `rb_ast_free` and `rb_ast_dispose` can be simpler by this change. Will you include the simplification into this PR?

PR #10661, opened 2024-04-27
Add DSL support for ripper with `%rule` actions

ydah:


yui-knk:
I think it might be better to include this change into https://github.com/ruby/ruby/pull/10602 so that we can confirm the DSL extension works well or not.

ydah:
Indeed it is. I cherry-pick into #10602 so I close this PR. Thank you!

PR #10667, opened 2024-04-29
Use user defined parameterizing rules `opt_args_tail(tail)`

ydah:


nobu (approved):

yui-knk:
@ydah JFYI:
I get feedback from @nobu.  nobu suggests it's good if we can reduce redundant `<node_args>` specification on caller side. I think it might be possible to introduce new Lrama grammar file syntax to support default LHS type specifier on `%rule` definition like below, which will be overridden caller side tag specification.
As a fact, all these 3 PRs use same tag 

* https://github.com/ruby/ruby/pull/10667 uses only `<node_args>`
* https://github.com/ruby/ruby/pull/10669 uses only `<node_kw_arg>`
* https://github.com/ruby/ruby/pull/10670 uses only `<node_opt_arg>`

```
%rule opt_args_tail(tail) <node_args>: ',' tail { ... } ...
```

yui-knk:
https://github.com/ruby/ruby/pull/10706 removes redundant `<tag>` specifications on caller side.

PR #10677, opened 2024-04-29
Memory leak in parser

peterzhu2118:
For example:

```ruby
10.times do
  100_000.times do
    eval("")
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

    19872
    26480
    32848
    39504
    45904
    52672
    59200
    65760
    72128
    78496

After:

    14320
    14320
    14320
    14320
    14320
    14320
    14320
    14336
    14336
    14336


kddnewton (approved):

nobu:
This introduced "allocate then wrap" strategy, which can cause another memory leak.

PR #10679, opened 2024-04-29
YJIT: Expand codegen for `TrueClass#===` to `FalseClass` and `NilClass`

rwstauner:
They are also high on the list of C functions from the lobsters benchmarks.


maximecb (approved):
I like that you refactored it to use the same codegen function, and added several tests. Well done! 🤘

PR #10680, opened 2024-04-29
YJIT: Take VM lock when invalidating

XrXr:
We need the lock to patch code safely.


k0kubun (approved):
I was wondering if they need that too for fixing Ractor failures. So :+1:

PR #10688, opened 2024-04-30
YJIT: print msg to stderr when RubyVM::YJIT.disasm not available

maximecb:
Print a more useful error message when people try to use this feature without YJIT dev.

Previously, this cryptic error message would be printed when trying to use disasm without dev mode:
```
irb(main):001> RubyVM::YJIT.disasm(MyThing.instance_method(:call))
<internal:yjit>:203:in `+': no implicit conversion of nil into String (TypeError)
	from <internal:yjit>:203:in `disasm'
	from /home/samuel/Projects/socketry/console/test.rb(irb):1:in `<main>'
	from <internal:kernel>:187:in `loop'
	from <internal:prelude>:5:in `irb'
	from ./test.rb:16:in `<main>'
```

Also fix an issue with .gitignore file on macOS

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -201,20 +201,27 @@ module RubyVM::YJIT
     # If a method or proc is passed in, get its iseq
     iseq = RubyVM::InstructionSequence.of(iseq)
 
-    if self.enabled?
-      disasm_str = Primitive.rb_yjit_disasm_iseq(iseq)
-
-      if !disasm_str
-        $stderr.puts("YJIT disasm is only available when YJIT is built in dev mode, i.e.:\n")
-        $stderr.puts("./configure --enable-yjit=dev (see doc/yjit/yjit.md)\n")
-      else
-        # Produce the disassembly string
-        # Include the YARV iseq disasm in the string for additional context
-        iseq.disasm + "\n" + disasm_str
-      end
-    else
-      iseq.disasm
+    if !self.enabled?
+      warn(
+        "YJIT needs to be enabled to produce disasm output, e.g.\n" +
+        "ruby --yjit-call-threshold=1 my_script.rb (see doc/yjit/yjit.md)"
+      )
+      return nil
     end
+
+    disasm_str = Primitive.rb_yjit_disasm_iseq(iseq)
+
+    if !disasm_str
+      warn(
+        "YJIT disasm is only available when YJIT is built in dev mode, i.e.\n" +
+        "./configure --enable-yjit=dev (see doc/yjit/yjit.md)\n"
+      )
+      return nil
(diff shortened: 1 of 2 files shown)


yjit.rb
```diff
@@ -202,9 +202,16 @@ def self.disasm(iseq) # :nodoc:
...
 
     if self.enabled?
-      # Produce the disassembly string
-      # Include the YARV iseq disasm in the string for additional context
-      iseq.disasm + "\n" + Primitive.rb_yjit_disasm_iseq(iseq)
+      disasm_str = Primitive.rb_yjit_disasm_iseq(iseq)
+
+      if !disasm_str
+        $stderr.puts("YJIT disasm is only available when YJIT is built in dev mode, i.e.:\n")
+        $stderr.puts("./configure --enable-yjit (see doc/yjit/yjit.md)\n")
```
rwstauner, line 209: should this say `=dev` ?
```suggestion
        $stderr.puts("./configure --enable-yjit=dev (see doc/yjit/yjit.md)\n")
```

maximecb: 🤦‍♀️ yes it should

yjit.rb
```diff
@@ -202,9 +202,16 @@ def self.disasm(iseq) # :nodoc:
...
+      disasm_str = Primitive.rb_yjit_disasm_iseq(iseq)
+
+      if !disasm_str
+        $stderr.puts("YJIT disasm is only available when YJIT is built in dev mode, i.e.:\n")
+        $stderr.puts("./configure --enable-yjit=dev (see doc/yjit/yjit.md)\n")
+      else
+        # Produce the disassembly string
+        # Include the YARV iseq disasm in the string for additional context
+        iseq.disasm + "\n" + disasm_str
+      end
```
rwstauner, line 214: If enabled but not a dev build this will return nil.
If not enabled it returns `iseq.disaam`.

Would it be better to consistently return `iseq.disasm` and just warn if we can't add the yjit disasm to it?  Or do you see value in returning the `nil` here?

Also, is there a reason to (or not to) use `warn` here?

maximecb: I didn't know about `warn`. I'm assuming this also prints to standard error?

I think returning `nil` might be useful because otherwise errors like this can go unnoticed, but maybe we should always return `nil` instead.

rwstauner: Returning `nil` more often might be the least confusing route

rwstauner: Yes, warn prints to stderr and includes stacktrace. Warnings can be also disabled
https://docs.ruby-lang.org/en/3.3/Kernel.html#method-i-warn

PR #10690, opened 2024-04-30
YJIT: Fix `Struct` accessors not firing tracing events

XrXr:
- **YJIT: Fix `Struct` accessors not firing tracing events**
- **RJIT: YJIT: Fix `Struct` readers not firing tracing events**

Reading and writing to structs should fire `c_call` and `c_return`, but
YJIT wasn't correctly dropping those calls when tracing.
This has been missing since this functionality was added in 3081c83169c55ef7eead6222e49248e09232c22c,
but the added test only fails when ran in isolation with
`--yjit-call-threshold=1`. The test sometimes failed on CI.


maximecb (approved):
👌

PR #10695, opened 2024-05-01
[Bug 20541] Revert 84f2aabd272a54e79979795d2d405090704a1d07 to build fiddle with libffi < 3.2

ylecuyer:
Bug: https://bugs.ruby-lang.org/issues/20451

I reverted 84f2aabd272a54e79979795d2d405090704a1d07 which prevents compiling ruby 3.1.5 on a system with libffi < 3.2 (oracle linux 8 for example)

I checked the repro https://github.com/ruby/fiddle/issues/102#issuecomment-1241763091 within fedora 36 and everything is ok.

If people are still having an issue as described inhttps://github.com/ruby/fiddle/issues/102 they can bump to ruby 3.2 or 3.3 where it should be fixed


ylecuyer:
Closing in favor of https://github.com/ruby/ruby/pull/10696

PR #10697, opened 2024-05-01
Skip Warning.[] spec if mspec is passing warning options

kddnewton:
@eregon is this alright?

In order to run the specs we run with `-T -W:no-experimental`, which causes this spec to fail. So I wanted to skip that spec if warning options were passed on the command line. Eventually the warning should go away with the prism compiler, so we'll go back to passing this without skipping (and can remove this line).


eregon (commented):
Yes that makes sense.

PR #10701, opened 2024-05-01
Move rvalue_overhead out of RVALUE

peterzhu2118:


gc.c
```diff
@@ -689,29 +689,27 @@ typedef struct RVALUE {
...
-#endif
-#if GC_DEBUG
+# endif
+# if GC_DEBUG
     const char *file;
     int line;
-#endif
-} RVALUE;
+# endif
+} __attribute__((aligned(sizeof(VALUE))));
```
nobu, line 704: May not be portable.

peterzhu2118: I changed it to use a union in a struct instead. That should be portable.

PR #10704, opened 2024-05-01
[PRISM] Enable test_ast.rb

kddnewton:


nobu:
Why not add them to test/.excludes-prism?

kddnewton:
I've been using `.excludes-prism` to track unintentional failures, but these are intentional — these APIs will never work with prism (on purpose). So I wanted to put this into the test suite so that it's more obvious why these are being skipped.

PR #10720, opened 2024-05-03
Add a JSON addition for `Enumerator::ArithmeticSequence`

havenwood:
See https://bugs.ruby-lang.org/issues/20469


havenwood:
I just realized I opened this PR in the wrong place since JSON is a gem. Sorry for the noise!

PR #10723, opened 2024-05-04
[Bug #20468] More strictly check loop variable

nobu:
Prohibit a constant and safe navigation call as a loop variable of `for`.


spec/ruby/language/for_spec.rb
```diff
@@ -67,17 +67,16 @@ class OFor
...
-      m = [1,2,3]
-      n = 0
-      -> {
-        for CONST in m
-          n += 1
-        end
-      }.should complain(/already initialized constant/)
-      CONST.should == 3
-      n.should == 3
+  ruby_bug("#20468", ""..."3.4") do
```
eregon, line 70: Do you think this PR will be backported?
I would think not as-is since it could emit SyntaxError where it previously worked OK (although weird code).
If not backported, I think we should keep the old spec under `ruby_version_is ""..."3.4" do` and add the new spec under `ruby_version_is  "3.4" do`

nobu: I'm uncertain yet whether it should be backported.
I'll ask matz on the next Tuesday.

IMO, the current behavior is not expected and not a specification, so `""..."3.4"` part does is unnecessary.

PR #10724, opened 2024-05-04
Create jekyll-gh-pages.yml

mpmcservice111:
.github/workflows/jekyll-gh-pages.yml


zzak:
I don't know what the point of this PR is, please [open a ticket](https://bugs.ruby-lang.org/) if you'd like to propose this change with your rationale.

PR #10735, opened 2024-05-07
[PRISM] Enable more passing tests

kddnewton:


tool/lib/test/unit/assertions.rb
```diff
@@ -768,7 +768,14 @@ def assert_syntax_error(code, error, *args, **opt)
           e = assert_raise(SyntaxError, mesg) do
             syntax_check(src, fname, line)
           end
-          assert_match(error, e.message, mesg)
+
+          # Prism adds ANSI escape sequences to syntax error messages to
+          # colorize and format them. We strip them out here to make them easier
+          # to match against in tests.
+          message = e.message
+          message.gsub!(/\e\[.*?m/, "")
```
nobu, line 776: Why adding ANSI escape sequences even not tty?

PR #10737, opened 2024-05-07
Use spaces around => in Hash#inspect for symbol keys that would generate invalid syntax

jeremyevans:
Also use spaces around => in PP.pp for hashes with symbol keys that would generate invalid syntax.

Fixes [Bug #20433]

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -3440,6 +3440,7 @@ static int
 inspect_i(VALUE key, VALUE value, VALUE str)
 {
     VALUE str2;
+    bool spaces = false;
 
     str2 = rb_inspect(key);
     if (RSTRING_LEN(str) > 1) {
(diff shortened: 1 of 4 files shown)


hash.c
```diff
@@ -3449,7 +3450,18 @@ inspect_i(VALUE key, VALUE value, VALUE str)
         rb_enc_copy(str, str2);
     }
     rb_str_buf_append(str, str2);
-    rb_str_buf_cat_ascii(str, "=>");
+    if (RB_SYMBOL_P(key)) {
+        switch (*(RSTRING_PTR(str) + RSTRING_LEN(str) - 1)) {
```
KJTsanaktsidis, line 3454: This is a bit nitpicky, but are symbols guaranteed to be in any particular encoding? This check might incorrectly not-space a symbol that does need it?

Maybe the way to go is to just change the separator to `" => "` unconditionally, then we don't have to play this game.

jeremyevans: I'm not opposed to always having spaces, but that is a much less compatible change that probably needs approval from Matz.  I'm not sure whether there are other failing cases besides the ones reported in the bug report (this one fixes the problematic cases specified in the bug report).  Symbols can be in any encoding, but inspect does not try to dump the encoding information.

KJTsanaktsidis (approved):
👍 should work in most cases at least and won’t do any harm in the cases where the symbol is in a strange encoding where !/?/etc don’t match ascii values

jeremyevans:
Closing as Matz decided on a different approach.

PR #10747, opened 2024-05-09
Skip test-bundled-gems on mswin

hsbt:
test-syntax-suggest is not working with mswin environment. We should skip it on mswin.


nobu:
Is it only on mswin, but ok on mingw?

hsbt:
Yes, I'm now sure why `rspec-3.x` is not working with mswin yet.

PR #10750, opened 2024-05-09
Add a ruby_bug spec for String#index clearing $~

casperisfine:
[[Bug #20421]](https://bugs.ruby-lang.org/issues/20421)

The bug was fixed in Ruby 3.3 via 9dcdffb8bf8a3654fd78bf1a58b30c8e13888a7a

@eregon


eregon (approved):
Thanks!

PR #10751, opened 2024-05-09
Fix `String#index` to clear MatchData when a regexp is passed

casperisfine:
[[Bug #20421]](https://bugs.ruby-lang.org/issues/20421)

The bug was fixed in Ruby 3.3 via 9dcdffb8bf8a3654fd78bf1a58b30c8e13888a7a


eregon (approved):

eregon:
Ah oops I should not have merged this, but have waited for @nagachika, sorry.
@nagachika Is it OK this is merged? It fixes https://bugs.ruby-lang.org/issues/20421

nagachika:
I checked the patch afterward. OK I updated the redmine ticket with Backport field.

PR #10770, opened 2024-05-15
No need to specify tags anymore

yui-knk:
In the past, these codes were used by both parser and ripper. On ripper, the type of LHS is `<val>` then type cast was needed. However currently these are only used by parser then no need to cast.


yui-knk:
Thanks, updated the commit message.

PR #10771, opened 2024-05-15
Allow waitpid(-1, Process::WNOHANG) to be woken if a waitpid(pid) is pending (3.2 backport)

stanhu:
If two threads are running, with one calling waitpid(-1, Process::WNOHANG), and another calling waitpid($some_pid), and then $some_other_pid exits, we would expect the waitpid(-1, Process::WNOHANG) call to retrieve that exit status. However, it cannot actually do so until $some_pid _also_ exits.

This patch fixes the issue by ensuring that if -1 is specified with the WNOHANG flag, -1 gets added to the process group waits on SIGCHLD.

This is a similar issue to https://bugs.ruby-lang.org/issues/19837, except the WNOHANG flag is used.

Fixes [#20490](https://bugs.ruby-lang.org/issues/20490)

### DIFF
--- a/test/ruby/test_process.rb
+++ b/test/ruby/test_process.rb
@@ -2659,7 +2659,7 @@ EOS
     end;
   end if Process.respond_to?(:_fork)
 
-  def test_concurrent_group_and_pid_wait
+  def _test_concurrent_group_and_pid_wait(nohang)
     # Use a pair of pipes that will make long_pid exit when this test exits, to avoid
     # leaking temp processes.
     long_rpipe, long_wpipe = IO.pipe
@@ -2683,8 +2683,23 @@ EOS
       Thread.pass until t1.stop?
       short_wpipe.close # Make short_pid exit
 
-      # The short pid has exited, so -1 should pick that up.
-      assert_equal short_pid, Process.waitpid(-1)
+      # The short pid has exited, so waitpid(-1) should pick that up.
+      exited_pid =
+        unless nohang
+          Process.waitpid(-1)
+        else
+          EnvUtil.timeout(5) do
+            loop do
+              pid = Process.waitpid(-1, Process::WNOHANG)
+
+              break pid if pid
+
+              sleep 0.1
+            end
+          end
+        end
+
+      assert_equal short_pid, exited_pid
 
       # Terminate t1 for the next phase of the test.
       t1.kill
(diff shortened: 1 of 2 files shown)


test/ruby/test_process.rb
```diff
@@ -2684,7 +2684,7 @@ def test_concurrent_group_and_pid_wait
       short_wpipe.close # Make short_pid exit
 
       # The short pid has exited, so -1 should pick that up.
-      assert_equal short_pid, Process.waitpid(-1)
+      assert_equal short_pid, Process.waitpid(-1, waitpid_flags)
```
stanhu, line 2687: This might be a little racy since `short_pid` may not have exited by the time `waitpid` is called.

stanhu:
@nagachika @KJTsanaktsidis Since you have context in https://bugs.ruby-lang.org/issues/19837, could you review this and https://github.com/ruby/ruby/pull/10772?

See bug report in https://bugs.ruby-lang.org/issues/20490.

nagachika:
Seems good to me. @KJTsanaktsidis, Do you have any comment?

nagachika (approved):

KJTsanaktsidis:
OK, I think I understand the problem:

* A thread is blocked in `Process.waitpid $PID1`
* Another process `$PID2` exits
* A call to `Process.waitpid -1, Process::WNOHANG` returns nil, even though it should be able to reap and return `$PID2`.

I don't believe this patch is correct though - it will cause `Process::WNOHANG` to actually hang when a process is _not_ ready to reap. You can see this if you put this assertion in the test, before the call to `short_pipe.wclose`.

```
      # Wait for us to be blocking in a call to waitpid2
      Thread.pass until t1.stop?
      assert_nil Process.waitpid(-1, Process::WNOHANG) ### Add this assertion here
      short_wpipe.close # Make short_pid exit
```

It passes on the current `ruby_3_2`, but fails with this patch.

Let me have a think about how we can fix this.... (or if you have any ideas, please shout out!)

KJTsanaktsidis:
@stanhu WDYT about this? (🤞 let's see if it passes the tests ) https://github.com/ruby/ruby/pull/10787

stanhu:
@KJTsanaktsidis Thanks! I'll close this in favor of #10787.

PR #10772, opened 2024-05-15
Allow waitpid(-1, Process::WNOHANG) to be woken if a waitpid(pid) is pending (3.1 backport)

stanhu:
If two threads are running, with one calling waitpid(-1, Process::WNOHANG), and another calling waitpid($some_pid), and then $some_other_pid exits, we would expect the waitpid(-1, Process::WNOHANG) call to retrieve that exit status. However, it cannot actually do so until $some_pid _also_ exits.

This patch fixes the issue by ensuring that if -1 is specified with the WNOHANG flag, -1 gets added to the process group waits on SIGCHLD.

This is a similar issue to https://bugs.ruby-lang.org/issues/19837, except the WNOHANG flag is used.

Fixes [#20490](https://bugs.ruby-lang.org/issues/20490)

### DIFF
--- a/process.c
+++ b/process.c
@@ -1292,7 +1292,7 @@ waitpid_wait(struct waitpid_state *w)
     if (w->ret) {
         if (w->ret == -1) w->errnum = errno;
     }
-    else if (w->options & WNOHANG) {
+    else if (w->options & WNOHANG && w->pid > 0) {
     }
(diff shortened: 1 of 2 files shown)


hsbt:
It seems bugfix. I won't backport this.

PR #10773, opened 2024-05-15
Introduce `iniline_operation` rule

S-H-GAMELINKS:
`lrama` allow `%inline` directive, which may be add `inline_operation`.

```yacc
/*
 *	inlining rules
 */
%rule %inline inline_operation : tIDENTIFIER
                               | tCONSTANT
                               | tFID
                               ;
```

And it looks like can replace the same BNF pattern of `fname` and `operation` with `inline_operation`.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -2908,6 +2908,9 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
                 | keyword_variable
                 ;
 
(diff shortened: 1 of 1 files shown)


S-H-GAMELINKS:
Thank you for the rebase this pull request.
But, this pull request was conflict that seems be merged #10941.
So I will rebase it again on my end.

PR #10774, opened 2024-05-15
[DOC] Adjust doc for rb_ary_new_capa

zzak:
* "Designed capacity of the generating array" becomes:
  * "Desired capacity of the array"
* "this could be effective than" becomes:
  * "this could be more efficient than"
* I'm not sure what "how many rooms of objects it allocates" means.
  * Instead use "the `capa` parameter to specify the length of the array."
  * This is not to be confused with the `Array#length`...
* Removed excess spaces mid-sentence
  * /cc @shyouhei who added these, I'm not sure if it was intentional?

### DIFF
--- a/include/ruby/internal/intern/array.h
+++ b/include/ruby/internal/intern/array.h
@@ -77,13 +77,14 @@ VALUE rb_check_array_type(VALUE obj);
 VALUE rb_ary_new(void);
 
 /**
- * Identical to rb_ary_new(),  except it additionally specifies  how many rooms
- * of  objects it  should allocate.   This way  you can  create an  array whose
- * capacity is  bigger than the  length of  it.  If you  can say that  an array
- * grows to a  specific amount, this could be effective  than resizing an array
- * over and over again and again.
+ * Identical to rb_ary_new(), except it accepts the `capa` parameter to specify
+ * the length of the array.
  *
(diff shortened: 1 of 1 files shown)


zzak:
Ahh, I realized the spaces are because the text is meant to be justified to 80 characters. :thinking:

include/ruby/internal/intern/array.h
```diff
@@ -77,13 +77,14 @@ VALUE rb_check_array_type(VALUE obj);
...
- * of  objects it  should allocate.   This way  you can  create an  array whose
- * capacity is  bigger than the  length of  it.  If you  can say that  an array
- * grows to a  specific amount, this could be effective  than resizing an array
- * over and over again and again.
+ * Identical to rb_ary_new(), except it accepts the `capa` parameter to specify
+ * the length of the array.
  *
- * @param[in]  capa  Designed capacity of the generating array.
+ * This way you can create an array whose capacity is bigger than the length of
+ * it. If you can say that an array grows to a specific amount, this could be
```
shyouhei, line 84: "The length" that appear twice in this part seems something different each other.  It is confusing to non-native English speakers like me.

shyouhei (commented):
The spaces were done using Emacs for aesthetics reasons only.  You can safely remove them.

PR #10776, opened 2024-05-15
Update duplicated 'when' clause warning message

kddnewton:
Previously for duplicated when clauses like:

1: case foo
2: when :bar
3: when :baz
4: when :bar
5: end

You would get a warning like:

```
warning: duplicated 'when' clause with line 2 is ignored
```

which is supposed to read as in:

"The 'when' clause at line 4 duplicates (conflicts) with (a 'when' clause at) line 2. This is ignored."

The issue is that it can be read as the clause on line 2 is ignored. This commit updates the warning message to instead say:

```
warning: 'when' clause on line 4 duplicates 'when' clause on line 2 and is ignored
```

This makes it clear which clause is begin duplicated and which clause is being ignored.

This commit fixes both parsers.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -13334,8 +13334,8 @@ check_literal_when(struct parser_params *p, NODE *arg, const YYLTYPE *loc)
     else {
         st_data_t line;
         if (st_lookup(p->case_labels, (st_data_t)arg, &line)) {
-            rb_warning1("duplicated 'when' clause with line %d is ignored",
-                        WARN_I((int)line));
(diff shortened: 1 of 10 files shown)


Edwing123:
At first I thought the one on line 4 was being ignored, I think something like the following is an option, too:

> warning: 'when' clause on line 2  is ignored due to duplicated 'when' clause on line 4.

kddnewton:
The one on line 4 _is_ ignored, the first one wins.

Edwing123:
> The one on line 4 _is_ ignored, the first one wins.

I see.

PR #10787, opened 2024-05-17
Allow waitpid(-1, Process::WNOHANG) to be woken if a waitpid(pid) is

KJTsanaktsidis:
...pending

If two threads are running, with one calling waitpid(-1, Process::WNOHANG), and another calling waitpid($some_pid), and then
$some_other_pid exits, we would expect the waitpid(-1, Process::WNOHANG) call to retrieve that exit status. However, it cannot actually do so until $some_pid _also_ exits.

This patch fixes the issue by calling do_waitpid unconditionally in waitpid_wait; this will ensure that a waitpid -1 actually reaps something (after first checking that no PID-directed call wants the process).

### DIFF
--- a/process.c
+++ b/process.c
@@ -1270,6 +1270,7 @@ waitpid_cleanup(VALUE x)
     return Qfalse;
 }
 
+#if RUBY_SIGCHLD
 static void
 waitpid_wait(struct waitpid_state *w)
 {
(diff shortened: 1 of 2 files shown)


KJTsanaktsidis:
argh, I have to guard `waitpid_wait` with conditional compilation now because `waitpid_each` is too. Fixing...

nagachika (approved):

PR #10803, opened 2024-05-19
Tempfile.create_io implemented.

akr:
Tempfile.create_io creates an unnamed file.

It is implemented using Tempfile.create and unlink. It try to use O_TMPFILE available since Linux 3.11.

### DIFF
--- a/lib/tempfile.rb
+++ b/lib/tempfile.rb
@@ -513,7 +513,7 @@ end
 # this method fallbacks to create-and-unlink as on POSIX.
 #
 # Related: Tempfile.create.
-def Tempfile.create_io(basename="", tmpdir=nil, **options, &block)
+def Tempfile.create_io(basename="", tmpdir=nil, mode: 0, **options, &block)
   tmpio = nil
   if defined? File::TMPFILE # O_TMPFILE since Linux 3.11
     tmpfile_supported = true
@@ -528,7 +528,8 @@ def Tempfile.create_io(basename="", tmpdir=nil, **options, &block)
     end
   end
   if tmpio.nil?
-    tmpfile = Tempfile.create(basename, tmpdir, **options)
+    mode |= File::SHARE_DELETE
+    tmpfile = Tempfile.create(basename, tmpdir, mode: mode, **options)
     File.unlink(tmpfile.path)
     tmpfile.autoclose = false
     tmpio = IO.new(tmpfile.fileno, File::RDWR) # change File to IO to drop path.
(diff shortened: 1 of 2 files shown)


nobu:
Unfortunately, create-and-unlink can't work on Windows.

akr:
> Unfortunately, create-and-unlink can't work on Windows.

I heard that it is possible on Windows.
(I talked about this feature with kosaki-san and naruse-san at RubyKaigi 2024 after party.)
I found FILE_SHARE_DELETE.
Is it really impossible?

nobu:
Hmm, I've forgot it.
`File::SHARE_DELETE` would be necessary to do it.

akr:
I added `File::SHARE_DELETE`. (not tested)

I searched FILE_SHARE_DELETE.
It seems it makes subsequent delete operation succeeds but the actual file is not deleted until the closing.
I think it is acceptable for this feature because a user of Tempfile.create_io doesn't need to remove the file anyway.

I also found FILE_FLAG_DELETE_ON_CLOSE and FILE_ATTRIBUTE_TEMPORARY.
I'm not sure if they are useful here or not.

test/test_tempfile.rb
```diff
@@ -425,4 +426,38 @@ def assert_mktmpdir_traversal
...
+      tmpio.rewind
+      assert_equal("foo\n", tmpio.read)
+      :result
+    }
+    assert_equal(:result, result)
+  end
+
+  def test_create_io_removes_file
+    Dir.mktmpdir {|d|
+      tmpio = Tempfile.create_io(nil, d)
```
nobu, line 458:
```suggestion
      tmpio = Tempfile.create_io("", d)
```

lib/tempfile.rb
```diff
@@ -464,3 +464,89 @@ def Tempfile.create(basename="", tmpdir=nil, mode: 0, **options)
...
+      fd = IO.sysopen(tmpdir, File::RDWR | File::TMPFILE, 0600)
+    rescue Errno::EISDIR, Errno::ENOENT, Errno::EOPNOTSUPP # kernel or the filesystem does not support O_TMPFILE
+      tmpfile_supported = false
+    end
+    if tmpfile_supported
+      tmpio = IO.new(fd, File::RDWR)
+    end
+  end
+  if tmpio.nil?
+    mode |= File::SHARE_DELETE
```
nobu, line 537: In `w32_wopen`, `_wopen` is used unless binary mode, and it does not have share mode argument.
```suggestion
    mode |= File::SHARE_DELETE | File::BINARY
```

ioquatix:
I believe we used the same feature for unnamed unix sockets on Windows.

PR #10819, opened 2024-05-21
readline: Make arm32 platform regex broader

matoro:
This needs to be broader to match all arm32 variants.  E.g. `armv7a-linux-eabihf`, `armv6j-linux-eabihf`

Fixes: aefc98891c42024039f19ef45bdfe93fbc590b7c

### DIFF
--- a/test/readline/test_readline.rb
+++ b/test/readline/test_readline.rb
@@ -497,7 +497,7 @@ module BasetestReadline
     omit if /i[3-6]86-linux/ =~ RUBY_PLATFORM
 
     # Skip arm32-linux (Travis CI).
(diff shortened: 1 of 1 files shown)


hsbt:
Can you submit this to https://github.com/ruby/readline-ext instead of this?

matoro:
> Can you submit this to https://github.com/ruby/readline-ext instead of this?

I can send it there in addition, but I believe that external gem is only used for Ruby 3.3 and up right?  Vs Ruby 3.2 uses the vendored code?  This is still needed to make Ruby 3.2 test suite pass on arm32.

hsbt:
>but I believe that external gem is only used for Ruby 3.3 and up right?

No. We need to fix that `ruby/readline-ext` at first. After that, I or branch maintainer backport `ruby_3_2` or others.

PR #10823, opened 2024-05-22
Prefer `/usr/bin/clang` as default CC on darwin

kateinoigakukun:
`/usr/bin/clang` is always available as long as Command Line Tools or Xcode is installed. Prefer the absolute path to avoid PATH lookup, which may lead to unaligned toolset selection.

This change makes toolset selection more robust by:
1. Use `/usr/bin/clang` as default CC on darwin.
2. Prefer other tools next to `CC` if selected CC is absolute and the tool is not explicitly specified.


configure.ac
```diff
@@ -202,7 +220,25 @@ set rb_dummy ${CC}
...
+            AS_IF([test -x "${rb_prog}"], [
+                rb_prog_found=yes
+                $1=${rb_prog}
+            ])
+            AC_MSG_RESULT([$rb_prog_found])
+        ])
+    ])
+    # Otherwise, try to find it in PATH with given program name
+    rb_prog=`echo "${rb_prog}" | sed 's/.*\///'`
+    AS_IF([test "${rb_prog_found}" = no], [
```
nobu, line 239: The basename of `rb_prog` is needed only if `rb_prog_found` is no.

```suggestion
    AS_IF([test "${rb_prog_found}" = no], [
        # Otherwise, try to find it in PATH with given program name
        rb_prog=`echo "${rb_prog}" | sed 's/.*\///'`
```

configure.ac
```diff
@@ -180,6 +180,24 @@ AS_CASE(["${build_os}"],
...
+            AC_MSG_RESULT([yes])
+            CC=$rb_darwin_default_CC
+        ],
+        [
+            AC_MSG_RESULT([no])
+            # Fall back to PATH lookup just in case the /usr/bin/clang is not
+            # available for some reason.
+            AC_CHECK_TOOLS([CC], [clang cc gcc c99])
+        ]
+    )
```
nobu, line 200: What about expand `rb_CC` absolute path, only if the found `CC` is Xcode?

```
    AS_IF([$rb_CC -E -dM -xc - < /dev/null | grep -F __apple_build_version__ > /dev/null],
          [llvm_prefix= rb_CC=`command -v $rb_CC`], [llvm_prefix=llvm-])
```

rubyFeedback: Not disagreeing with you as such, but it should be pointed out that at the least on Linux, the default assumption is that the entries in $PATH coming first, have higher precedence. So, for instance, if a user would have a PATH that first has /opt/bin/, and /usr/bin/ as second entry, then /opt/bin/ should be searched first. This will probably happen only very, very rarely, so it does not invalidate your statement, but I wanted to mention it in regards to (possible) user's expectations.

nobu: I think we should abort in that case, if not `[test -x "${rb_prog}"]`, rather than searching from `$PATH`.

configure.ac
```diff
@@ -202,7 +220,25 @@ set rb_dummy ${CC}
...
+                $1=${rb_prog}
+            ])
+            AC_MSG_RESULT([$rb_prog_found])
+        ])
+    ])
+    AS_IF([test "${rb_prog_found}" = no], [
+        # Otherwise, try to find it in PATH with given program name
+        rb_prog=`echo "${rb_prog}" | sed 's,.*/,,'`
+        AC_CHECK_PROG([$1], [$rb_prog], [$rb_prog])
+    ])
```
nobu, line 241:
```suggestion
    # If the tool is not explicitly specified and the compiler
    # is absolute, check the one next to the compiler.
    AS_IF([test -z "$$1"], [
        AS_CASE(["${rb_prog}"],
        [/*], [
            AS_IF([test -x "${rb_prog}"], [
                $1="${rb_prog}"
            ], [
                AC_MSG_ERROR("executable ${rb_prog} is not found")
            ])
        ])
    ])
```

configure.ac
kateinoigakukun: It makes sense to me. It sounds well consistent behavior.

nobu: I think `rb_darwin_default_CC` is not needed anymore.

kateinoigakukun: Sorry, do you mean we should find CC from PATH instead of having the default value?

nobu: Yes, only for Xcode clang, as commented above.

PR #10848, opened 2024-05-27
[Bug #20511] Update reline-0.5.7

hsbt:
https://bugs.ruby-lang.org/issues/20511


hsbt:
```
   1) Failure:
  TestIRB::ColorTest#test_colorize_code [/Users/runner/work/ruby/ruby/src/test/irb/test_color.rb:147]:
  Case: colorize_code("\t", complete: true)
  Result: "  ".
  <"\t"> expected but was
  <"  ">.
```

We need to update compatible version of IRB.

PR #10849, opened 2024-05-27
Stop marking chilled strings as frozen

etiennebarrie:
They were initially made frozen to avoid false positives for cases such as:

```ruby
str = str.dup if str.frozen?
```

But this may cause bugs and is generally confusing for users.

### DIFF
--- a/include/ruby/internal/intern/error.h
+++ b/include/ruby/internal/intern/error.h
@@ -190,6 +190,7 @@ RBIMPL_ATTR_NONNULL(())
  */
 void rb_error_frozen(const char *what);
 
+RBIMPL_ATTR_NORETURN()
 /**
  * Identical  to  rb_error_frozen(),  except  it takes  arbitrary  Ruby  object
  * instead of C's string.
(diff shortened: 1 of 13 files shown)


include/ruby/internal/intern/error.h
```diff
@@ -248,12 +249,27 @@ RBIMPL_SYMBOL_EXPORT_END()
...
     } while (0)
 
+# ifdef __cplusplus
+extern "C" {
+# endif
+bool rb_str_chilled_p(VALUE str);
+void rb_str_modify(VALUE str);
+# ifdef __cplusplus
+}
+# endif
```
nobu, line 264:
```suggestion
RBIMPL_SYMBOL_EXPORT_BEGIN()

bool rb_str_chilled_p(VALUE str);
void rb_str_modify(VALUE str);

RBIMPL_SYMBOL_EXPORT_END()
```

Or you can move these declarations before `RBIMPL_SYMBOL_EXPORT_END()`.

casperisfine: Thank you, I was looking for a clean way to do this.

casperisfine:
Alright, I figured why CI is still failing. The YJIT `str_uplus` implementation need to be updated to dup chilled strings.

casperisfine:
Should go green now.

PR #10851, opened 2024-05-27
Ensure that multiple Timeouts call Thread#raise at most once

mame:
https://bugs.ruby-lang.org/issues/20314


mame:
@ko1 reviewed and found many problems with this PR. I withdraw this for now.

* Since `request.finished` is called outside of the timer thread, it seemed better to have a mutex for each request.
* An entry of `@@request_stacks` should be deleted when it was empty. Also, using Fiber can break invariant. It might be better to use Fiber.storage.
* `prev_request.inner_request` should be cleared when timeout returns.
* `ensure; stack.pop; request.finished` can be interrupted by `Thread#raise`, which breaks consistency.

PR #10843, opened 2024-05-24
[Feature #20507] Allow compilation of C extensions with `-Wconversion`

flavorjones:
https://bugs.ruby-lang.org/issues/20507

As a maintainer of several C extensions, I like to compile my code with some of the common [Warning Options](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html) to help maintain code quality and safety.

One of the options that is currently very difficult to use is `-Wsign-conversion`, because some of the Ruby public header files contain code that will generate warnings. So many warnings are printed when compiling against Ruby's header files that it's hard to see any warnings from my code.

As an example, compiling just one file in Nokogiri with this flag enabled will print:

```
.../include/ruby-3.4.0+0/ruby/internal/special_consts.h: In function ‘RB_TEST’:
.../include/ruby-3.4.0+0/ruby/internal/special_consts.h:159:16: warning: unsigned conversion from ‘int’ to ‘VALUE’ {aka ‘long unsigned int’} changes value from ‘-5’ to ‘18446744073709551611’ [-Wsign-conversion]
  159 |     return obj & ~RUBY_Qnil;
      |                ^
.../include/ruby-3.4.0+0/ruby/internal/special_consts.h: In function ‘RB_NIL_OR_UNDEF_P’:
.../include/ruby-3.4.0+0/ruby/internal/special_consts.h:229:24: warning: unsigned conversion from ‘int’ to ‘VALUE’ {aka ‘const long unsigned int’} changes value from ‘-33’ to ‘18446744073709551583’ [-Wsign-conversion]
  229 |     const VALUE mask = ~(RUBY_Qundef ^ RUBY_Qnil);
      |                        ^
.../include/ruby-3.4.0+0/ruby/internal/arithmetic/long.h: In function ‘RB_INT2FIX’:
.../include/ruby-3.4.0+0/ruby/internal/arithmetic/long.h:117:29: warning: conversion to ‘long unsigned int’ from ‘long int’ may change the sign of the result [-Wsign-conversion]
  117 |     const unsigned long j = i;
      |                             ^
.../include/ruby-3.4.0+0/ruby/internal/arithmetic/long.h:119:29: warning: conversion to ‘long int’ from ‘long unsigned int’ may change the sign of the result [-Wsign-conversion]
  119 |     const long          l = k;
      |                             ^
[...]

PR #10854, opened 2024-05-28
Make value_type.h compatible with -Wconversion

casperisfine:
Followup: https://github.com/ruby/ruby/pull/10843

[[Feature #20507]](https://bugs.ruby-lang.org/issues/20507)

This was missed from the initial commit.

```
../../.././include/ruby/internal/value_type.h:446:27: error: implicit conversion changes signedness: 'enum ruby_value_type' to 'int' [-Werror,-Wsign-conversion]
    rb_unexpected_type(v, t);
    ~~~~~~~~~~~~~~~~~~    ^
```

@flavorjones @nobu @shyouhei

### DIFF
--- a/include/ruby/internal/value_type.h
+++ b/include/ruby/internal/value_type.h
@@ -443,7 +443,7 @@ Check_Type(VALUE v, enum ruby_value_type t)
     }
 
   unexpected_type:
-    rb_unexpected_type(v, t);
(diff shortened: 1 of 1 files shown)


shyouhei (approved):
OK.  TBH an `enum` needs explicit cast to `int` sounds esoteric to me though.

byroot:
I found this one because it was failing the regression test when running make on my machine. Not too sure why CI didn't catch it.

PR #10858, opened 2024-05-28
YJIT: limit size of call count stats dict

maximecb:
Someone reported two weeks ago that logs were getting bloated because the ISEQ and C call count dicts were huge, since they include all of the call sites. I wrote code on the Rust size to limit the size of the dict to avoid this problem. The size limit is hardcoded at 20, but I figure this is probably fine?

### DIFF
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -821,7 +821,7 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
 
                 // Cap the number of counts reported to avoid
                 // bloating log files, etc.
-                pairs = pairs[0..20].to_vec();
+                pairs.truncate(20);
 
                 // Add the pairs to the dict
                 for (name, call_count) in pairs {
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -802,16 +802,31 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
             rb_hash_aset(hash, key, value);
         }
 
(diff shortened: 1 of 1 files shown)


yjit/src/stats.rs
```diff
@@ -802,16 +802,31 @@ fn rb_yjit_gen_stats_dict(context: bool) -> VALUE {
...
                     let count = call_counts[*idx];
+                    pairs.push((name, count));
+                }
+
+                // Sort the vectors by decreasing call counts
+                pairs.sort_by_key(|e| -(e.1 as i64));
+
+                // Cap the number of counts reported to avoid
+                // bloating log files, etc.
+                pairs = pairs[0..20].to_vec();
```
k0kubun, line 824: This seems to crash when it has less than 20 elements. I guess you can use `pairs.iter().take(20)`?

k0kubun:
This would break the calculation of `top_n_pct` https://github.com/ruby/ruby/blob/1eff5a98f1c83f37ca2daac1c52f3eb84f395acd/yjit.rb#L420-L421. Is that intentional?

maximecb:
> This would break the calculation of `top_n_pct`.
> 
> https://github.com/ruby/ruby/blob/1eff5a98f1c83f37ca2daac1c52f3eb84f395acd/yjit.rb#L420-L421
> 
> Is that intentional?

AFAIK it doesn't break `top_n_pct`, because `num_calls` is not calculated based on the dict, we're gathering that count separately?

k0kubun:
🤦 You're right, I was reading it wrong. I'm sorry.

PR #10864, opened 2024-05-29
Prevent fork from happening in known fork unsafe API

casperisfine:
[[Feature #20590]](https://bugs.ruby-lang.org/issues/20590)

For better of for worse, fork(2) remain the primary provider of parallelism in Ruby programs. Even though it's frowned uppon in many circles, and a lot of literature will simply state that only async-signal safe APIs are safe to use after `fork()`, in practice most APIs work well as long as you are careful about not forking while another thread is holding a pthread mutex.

One of the APIs that is known cause fork safety issues is `getaddrinfo`. If you fork while another thread is inside `getaddrinfo`, a mutex may be left locked in the child, with no way to unlock it.

I think we could reduce the impact of these problem by preventing in for the most notorious and common cases, by locking around `fork(2)` and known unsafe APIs with a read-write lock.

FYI: @mame @beauraF

### DIFF
--- a/ext/socket/raddrinfo.c
+++ b/ext/socket/raddrinfo.c
@@ -327,6 +327,12 @@ nogvl_getaddrinfo(void *arg)
     return (void *)(VALUE)ret;
 }
 
+static void *
+fork_safe_getaddrinfo(void *arg)
+{
+    return rb_thread_prevent_fork(nogvl_getaddrinfo, arg);
+}
+
 static int
 rb_getaddrinfo(const char *hostp, const char *portp, const struct addrinfo *hints, struct addrinfo **ai)
 {
@@ -336,7 +342,7 @@ rb_getaddrinfo(const char *hostp, const char *portp, const struct addrinfo *hint
     arg.service = portp;
     arg.hints = hints;
     arg.res = ai;
-    return (int)(VALUE)rb_thread_call_without_gvl(nogvl_getaddrinfo, &arg, RUBY_UBF_IO, 0);
+    return (int)(VALUE)rb_thread_call_without_gvl(fork_safe_getaddrinfo, &arg, RUBY_UBF_IO, 0);
 }
 
 #elif GETADDRINFO_IMPL == 2
@@ -477,6 +483,12 @@ do_pthread_create(pthread_t *th, void *(*start_routine) (void *), void *arg)
     return ret;
 }
 
+static void *
+fork_safe_do_getaddrinfo(void *ptr)
+{
+    return rb_thread_prevent_fork(do_getaddrinfo, ptr);
+}
+
 static int
 rb_getaddrinfo(const char *hostp, const char *portp, const struct addrinfo *hints, struct addrinfo **ai)
 {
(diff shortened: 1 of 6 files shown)


ext/socket/raddrinfo.c
```diff
@@ -498,9 +524,6 @@ rb_getaddrinfo(const char *hostp, const char *portp, const struct addrinfo *hint
         errno = err;
         return EAI_SYSTEM;
     }
-    pthread_detach(th);
-
-    rb_thread_call_without_gvl2(wait_getaddrinfo, arg, cancel_getaddrinfo, arg);
```
KJTsanaktsidis, line 503: Was this deleted on purpose? it seems important to me! (it's the bit that actually waits on the condvar for the lookup to be done, right?)

casperisfine: Nah, just a bad rebase I think, let me check.

casperisfine: Ok, fixed it. It should have been replaced by `fork_safe_async_getaddrinfo(arg);`

ext/socket/raddrinfo.c
```diff
@@ -477,6 +483,26 @@ do_pthread_create(pthread_t *th, void *(*start_routine) (void *), void *arg)
...
+    pthread_detach(th);
+
+    rb_thread_call_without_gvl2(wait_getaddrinfo, arg, cancel_getaddrinfo, arg);
+    return NULL;
+}
+
+static void *
+fork_safe_async_getaddrinfo(void *ptr)
+{
+    return rb_thread_prevent_fork(async_getaddrinfo, ptr);
```
mame, line 503: @akr pointed out that it is wrong to call `rb_thread_prevent_fork` here.

If Ruby's getaddrinfo is interrupted, such as by Ctrl+C, the pthread in question calling getaddrinfo(3) may remain even after `async_getaddrinfo` returns. So this does not satisfy your purpose (to avoid `fork` is called during getaddrinfo(3) execution).
Instead, you should guard the call of getaddrinfo(3) (and freeaddrinfo(3) too?), in do_getaddrinfo.

Note that `rb_thread_prevent_fork` would be called with no GVL in this case. Because `rb_ensure` requires GVL, `rb_thread_prevent_fork` must not use it. I think `rb_thread_prevent_fork` doesn't have to guard longjmp.

casperisfine: Thank you @mame. I finally figured my linking error, and made changes based on your review.

I think it's much simpler now, how does it look?

mame: @casperisfine Thanks! It looks good to me. @akr Could you review it too?

mame: @akr said it looks okay

casperisfine: That's great. Does that means you are good with merging it, or do you wish to add it at a dev meeting first? https://bugs.ruby-lang.org/issues/20590

mame: Let's merge it!

casperisfine: 🎉 I rebased it and enabled auto-merge. Thanks again!

PR #10865, opened 2024-05-29
[DOC] Exceptions doc

BurdetteLamar:
Free-standing document about exceptions.

### DIFF
--- a/doc/exceptions.md
+++ b/doc/exceptions.md
@@ -1,0 +1,298 @@
+# Exceptions
+
+Any Ruby code can raise an exception.
+
+Most often, a raised exception is meant to alert the running program
+that an unusual (i.e., _exceptional_) situation has arisen,
+and may need to be handled.
+
+Code throughout the Ruby core, Ruby standard library, and Ruby gems generates exceptions
+in certain circumstances:
+
+```
+File.open('nope.txt') # Raises Errno::ENOENT: "No such file or directory"
+```
+
+# Raised Exceptions
+
+A raised exception transfers program execution, one way or another:
+
+- If the exception is _handled_ (see below),
+  execution transfers to an exception handler.
+- Otherwise,  execution transfers to code in the Ruby interpreter
+  that prints a message and exits the program (or thread):
+
+    ```
+    $ ruby -e "raise"
+    -e:1:in `<main>': unhandled exception
+    ```
+
+# \Exception Handlers
+
+An exception handler may determine what is to happen
+when an exception is raised.
+
+A simple example:
+
+```
+begin
+  raise 'Boom!'                # Raises an exception, transfers control.
+  puts 'Will not get here.'
+rescue
+  puts 'Rescued an exception.' # Control tranferred to here.
+end
+puts 'Got here.'
+```
+
+Output:
+
+```
+Rescued an exception.
+Got here.
+```
+
+An exception handler has several elements:
+
+| Element                     | Use                                                                                      |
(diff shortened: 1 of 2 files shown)


BurdetteLamar:
@peterzhu2118 (and anyone else):  I've marked this as a draft while we discuss scope.  I'm planning work on class Exception (class doc and methods doc), and on Kernel#raise.  It may be best to add those to this PR and review all at once;  on the other hand, that will make a biggish PR.  Thoughts?

doc/exceptions.md
```diff
@@ -0,0 +1,365 @@
...
+The `else` clause:
+
+- Starts with an `else` statement.
+- Contains code that is to be executed if no exception is raised in the begin clause.
+- Ends with the first following `ensure` or `end` statement.
+
+```
+begin
+  puts 'Begin.'
+rescue NoSuch
```
peterzhu2118, line 192: I think it's confusing and not relevant to use a nonexistent exception here. We could just use an empty rescue clause I think

```suggestion
rescue
```

doc/exceptions.md
```diff
@@ -0,0 +1,365 @@
...
+  # Do something with the retrieved data.
+rescue OpenURI::HTTPError
+  retry if (retries += 1) < 5
+else
+  puts "Got it, with #{retries} retries."
+end
+```
+
+```
+Got it, with 0 retries.
```
peterzhu2118, line 331: This example shows a real-life example of how to use retries but doesn't really show how retries work because it succeeded on the first try. We should use a more artificial example that demonstrates how retries work.

doc/exceptions.md
```diff
@@ -0,0 +1,365 @@
...
+To provide additional or alternate information,
+you may create custom exception classes;
+each should be a subclass of one of the built-in exception classes.
+
+If you are building a library or gem (or even if you're not),
+it's good practice to start with a single “generic” exception class
+(commonly an immediate subclass of StandardError or RuntimeError),
+and have its other exception classes derive from that class.
+This allows an exception handler to rescue the generic exception,
+thus also rescuing all its derived exceptions.
```
peterzhu2118, line 352: I'm not sure if this is really a common or necessarily encouraged practice. Do have an example of where this is done?

peterzhu2118 (commented):
Do you have any plans to link this document from any existing documentation?

BurdetteLamar:
> Do you have any plans to link this document from any existing documentation?

Yes; at a minimum, from: the class doc for Exception; method doc for Kernel#raise.

doc/exceptions.md
BurdetteLamar: Net::HTTP has most of its exceptions subclassed from `HTTPClientError` and `HTTPServerError`.

doc/exceptions.md
BurdetteLamar: Done.

doc/exceptions.md
BurdetteLamar: Done.

doc/exceptions.md
peterzhu2118: I feel like Net::HTTP is a special case, since all HTTP errors are of the 4xx class or 5xx class (for client and server errors, respectively). I don't think this this statement is generally true though.

BurdetteLamar: Removed.

BurdetteLamar:
@peterzhu2118, ready for you.

PR #10875, opened 2024-05-30
ci: Test whether GMP is working in compilers.yml

sorah:
Add CI scenario to avoid reoccurence of https://bugs.ruby-lang.org/issues/20515 https://github.com/ruby/ruby/pull/10876

(This was a test pull-request just to trigger CI to verify the issue along with https://github.com/ruby/ruby/pull/10874)


sorah:
oops, forgot that the gmp check is not mandatory to merge! Needs #10876 to pass.

PR #10876, opened 2024-05-30
RUBY_CHECK_HEADER didn't define HAVE_{header-file}

sorah:
`--with-gmp` is not working at all because HAVE_GMP_H was missing since 18eaf0be90. [Bug #20515]

bug: https://bugs.ruby-lang.org/issues/20515
follow-up: https://bugs.ruby-lang.org/issues/20494
follow-up: 18eaf0be905e3e251423b42d6f4e56b7cae1bc3b
follow-up: https://github.com/ruby/ruby/pull/10805


sorah:
✅  verified `--with-opt-dir=/opt/homebrew` works well (original bug reported at https://bugs.ruby-lang.org/issues/20494)

PR #10877, opened 2024-05-30
Make error messages clear blocks/keywords are disallowed in index assignment

jeremyevans:
Blocks and keywords are allowed in regular index.

Also update NEWS to make this more clear.


byroot (approved):

NEWS.md
```diff
@@ -19,9 +19,11 @@ Note that each entry is kept to a minimum, see links for details.
   `**nil` is treated similarly to `**{}`, passing no keywords,
   and not calling any conversion methods.  [[Bug #20064]]
 
-* Block passing is no longer allowed in index.  [[Bug #19918]]
+* Block passing is no longer allowed in index assignment
```
nobu, line 22: A trailing space is intended?

```suggestion
* Block passing is no longer allowed in index assignment
```

nobu (approved):

NEWS.md
jeremyevans: Not intended.  I pushed a patch to remove that.

PR #10888, opened 2024-05-31
YJIT: implement variable-length context encoding scheme

maximecb:
This PR implements a scheme where `Context` objects are stored into a flat vector of bits, using a variable-length encoding scheme. This is essentially a simple ad-hoc compression scheme.

I wrote a method `Context::encode()` that encodes the information from the current context into a vector of bits. There is also a corresponding `Context::decode()` method. The compressed contexts are referred to using a `u32` which is a bit index into the flat `BitVector` that stores contexts.

The encoding is not particularly clever or anything. It could probably be optimized to be more compact. The main space saving feature is that if we don't know the type of a stack temp or local, we don't write anything. We can also avoid writing the inline block pointer when it's null. This alone seems to be potentially very effective in terms of space savings. Some values are currently always written at the beginning (i.e. stack size, sp offset). Type information is written using a system of 4-bit opcodes and variable-length instructions, with a special "end of encoding" opcode. 

There are two further space-saving optimizations. Offset 0 in the bit vector is reserved for the default context, and there is also a simple cache of the last context encoded. These save a few percent more memory.

I added some stats as well to compute the average number of bytes per context. Those values look quite good:

```
lobsters:
  context_data_bytes:          648,388
  num_contexts_encoded:        134,125
  bytes_per_context:              4.83

railsbench:
  context_data_bytes:          385,321
  num_contexts_encoded:         78,865
  bytes_per_context:              4.89

liquid-render:
  context_data_bytes:           86,032
  num_contexts_encoded:         17,311
  bytes_per_context:              4.97
```

The impact on `yjit_alloc_size` is not as clear cut and amazing as I would like it to be:
```
masters:
  lobsters:
    yjit_alloc_size:          20,651,810

  railsbench:
    yjit_alloc_size:           9,887,827

  liquid-render:
    yjit_alloc_size:           2,595,099

var-len:
  lobsters:
    yjit_alloc_size:          20,164,574

  railsbench:
    yjit_alloc_size:           9,640,691

  liquid-render:
    yjit_alloc_size:           2,624,275
```

We seem to save a little bit of space on the larger benchmarks but no space on others. One challenge is that the Rust allocator may have size classes like Ruby's allocator. So shrinking objects does not immediately translate into a visible reduction on memory usage. Another potential issue it that apparently Rust's `Vec` uses a 2x growth rate, which could waste significant space. I think I may be able to fix that issue by using a manually allocated slice of bytes and a 1.2x growth rate instead.

A third possible source of overhead is that it's possible that I'm encoding-decoding and re-encoding contexts in places where it's not really necessary. I would appreciate your help in reviewing this PR to identify such cases if there are any. Ideally, if we've encoded a context, we want to avoid decoding an re-encoding it again, because that could allocate more memory when it isn't necessary.

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -390,6 +390,12 @@ module RubyVM::YJIT
       out.puts "yjit_alloc_size:       " + format_number(13, stats[:yjit_alloc_size]) if stats.key?(:yjit_alloc_size)
       out.puts "live_context_size:     " + format_number(13, stats[:live_context_size])
       out.puts "live_context_count:    " + format_number(13, stats[:live_context_count])
+
+      bytes_per_context = stats[:context_data_bytes].fdiv(stats[:num_contexts_encoded])
+      out.puts "context_data_bytes:    " + format_number(13, stats[:context_data_bytes])
+      out.puts "num_contexts_encoded:  " + format_number(13, stats[:num_contexts_encoded])
+      out.puts "bytes_per_context:     " + ("%13.2f" % bytes_per_context)
+
       out.puts "live_page_count:       " + format_number(13, stats[:live_page_count])
       out.puts "freed_page_count:      " + format_number(13, stats[:freed_page_count])
       out.puts "code_gc_count:         " + format_number(13, stats[:code_gc_count])
--- a/yjit.rb
+++ b/yjit.rb
@@ -155,8 +155,8 @@ module RubyVM::YJIT
 
   # Return a hash for statistics generated for the `--yjit-stats` command line option.
   # Return `nil` when option is not passed or unavailable.
-  def self.runtime_stats(context: false)
-    stats = Primitive.rb_yjit_get_stats(context)
+  def self.runtime_stats()
+    stats = Primitive.rb_yjit_get_stats()
     return stats if stats.nil?
 
     stats[:object_shape_count] = Primitive.object_shape_count
@@ -313,7 +313,7 @@ module RubyVM::YJIT
 
     # Format and print out counters
     def _print_stats(out: $stderr) # :nodoc:
-      stats = runtime_stats(context: true)
+      stats = runtime_stats()
       return unless Primitive.rb_yjit_stats_enabled_p
 
       out.puts("***YJIT: Printing YJIT statistics on exit***")
@@ -388,8 +388,6 @@ module RubyVM::YJIT
 
       out.puts "freed_code_size:       " + format_number(13, stats[:freed_code_size])
       out.puts "yjit_alloc_size:       " + format_number(13, stats[:yjit_alloc_size]) if stats.key?(:yjit_alloc_size)
-      out.puts "live_context_size:     " + format_number(13, stats[:live_context_size])
-      out.puts "live_context_count:    " + format_number(13, stats[:live_context_count])
 
       bytes_per_context = stats[:context_data_bytes].fdiv(stats[:num_contexts_encoded])
       out.puts "context_data_bytes:    " + format_number(13, stats[:context_data_bytes])
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -458,8 +458,11 @@ const CHAIN_DEPTH_MASK: u8   = 0b00111111; // 63
 /// There are a lot of context objects so we try to keep the size small.
 #[derive(Copy, Clone, Default, Eq, Hash, PartialEq, Debug)]
 pub struct Context {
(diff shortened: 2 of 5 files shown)


XrXr:
I've only looked at this roughly, but I've always wanted some interactive tools for exploring binary formats. I tried [Kaitai Struct](https://kaitai.io/) for the encoding scheme here, and it somewhat worked, but unfortunately, it [doesn't have great support for bit-oriented things](https://github.com/kaitai-io/kaitai_struct/issues/576), which this scheme wants.

I ended up making each bit into a byte to make the following demo:

[yjitkaitai.webm](https://github.com/ruby/ruby/assets/6457510/3a8dd6cb-0a20-4e40-bb30-661f89c1eafb)

A WireShark dissector is probably more flexible than Kaitai's declarative DSL and I could look into that next. 😅

yjit.rb
```diff
@@ -390,6 +390,12 @@ def _print_stats(out: $stderr) # :nodoc:
       out.puts "yjit_alloc_size:       " + format_number(13, stats[:yjit_alloc_size]) if stats.key?(:yjit_alloc_size)
       out.puts "live_context_size:     " + format_number(13, stats[:live_context_size])
       out.puts "live_context_count:    " + format_number(13, stats[:live_context_count])
+
+      bytes_per_context = stats[:context_data_bytes].fdiv(stats[:num_contexts_encoded])
+      out.puts "context_data_bytes:    " + format_number(13, stats[:context_data_bytes])
+      out.puts "num_contexts_encoded:  " + format_number(13, stats[:num_contexts_encoded])
```
k0kubun, line 396: Do we need both `live_context_size`/`live_context_count` and `context_data_bytes`/`num_contexts_encoded`?

yjit/src/core.rs
```diff
@@ -1425,13 +2008,17 @@ pub fn take_version_list(blockid: BlockId) -> VersionList {
 fn get_num_versions(blockid: BlockId, inlined: bool) -> usize {
     let insn_idx = blockid.idx.as_usize();
     match get_iseq_payload(blockid.iseq) {
+
+        // FIXME: this counting logic is going to be expensive.
+        // We should avoid it if possible
```
k0kubun, line 2013: like always encoding `SetInlineBlock` as a first instruction? 🤔

yjit/src/core.rs
```diff
@@ -1561,7 +2149,7 @@ unsafe fn add_block_version(blockref: BlockRef, cb: &CodeBlock) {
     let block = unsafe { blockref.as_ref() };
 
     // Function entry blocks must have stack size 0
-    assert!(!(block.iseq_range.start == 0 && block.ctx.stack_size > 0));
+    //assert!(!(block.iseq_range.start == 0 && block.ctx.stack_size > 0));
```
k0kubun, line 2152: Is it possible to keep the assertion?
```suggestion
    assert!(!(block.iseq_range.start == 0 && block_ctx.stack_size > 0));
```

k0kubun (approved):

yjit.rb
maximecb: No we don't. The live_context_size / live_context_count are the old stats. I kept them around for comparison purposes but I can get rid of them since they will no longer be relevant.

yjit/src/core.rs
maximecb: I commented it because executing the assertion will require decoding the context, but I could turn it into a `debug_assert` ?

k0kubun: Oh sorry, I thought L2052 was in the same function but it wasn't. Then it makes sense. But yeah, making it a `debug_assert` would be nice too.

yjit/src/core.rs
maximecb: not sure, but it seemed expensive-ish to decode all the contexts to count this?

k0kubun: right, I just raised a potential idea. feel free to leave it in this PR.

XrXr:
```
thread '<unnamed>' panicked at src/core.rs:881:9:
assertion failed: Self::decode(idx) == *self
```
I guess two otherwise equivalent contexts can have different `decoded_from`.

maximecb:
> ```
> ruby: YJIT has panicked. More info to follow...
> thread '<unnamed>' panicked at src/core.rs:881:9:
> assertion failed: Self::decode(idx) == *self
> ```
> 
> I guess two otherwise equivalent contexts can have different `decoded_from`.

Thanks for pointing that out 😅

PR #10889, opened 2024-05-31
reinitialize the sched lock on `thread_atfork`

froydnj:
We are seeing intermittent hangs in various tests at Stripe when using Ruby 3.3.1 that involve subprocesses: one process is stuck waiting for a child process to complete, while the child process is stuck waiting for `vm->ractor.sched.lock`.  Inspecting the lock on x86-64 Linux says that the lock is being held by the parent process -- but the parent process's threads are all nowhere near the critical section that this lock is protecting, i.e. the parent process is not actually holding the lock.  The tests in question are all using the `subprocess` gem, but we believe the below scenario is applicable to any use of `fork`, including much of `Open3` in the standard library.

We think what must have happened is:

1. Some thread T2 in the parent process calls `fork` from Ruby.
1. Some thread T1 in the parent process locks `vm->ractor.sched.lock`
1. T2 calls `fork(2)`.
1. The child process starts, but only with a copy of T2; T1 only exists in the parent process.
1. The child process eventually tries to lock `vm->ractor.sched.lock` and discovers that the parent process is holding onto the lock.
1. The parent process eventually resumes T1 to unlock `vm->ractor.sched.lock`.
1. The child process is stuck waiting forever, because nothing in it exists to unlock `vm->ractor.sched.lock`.

To address the above problem, we re-initialize the mutex in `thread_atfork` so the child process starts with a clean slate.  This is undefined behavior according to pthreads, but so is re-initializing the condition variables and barrier here.  I don't think this makes things any worse. 😅 

We've done some limited testing with this patch on our internal CI and it appears to make the hangs go away (none of the affected tests hang in a handful of runs, whereas at least one of them would fail with high probability on any given run before).  We plan on doing wider testing with it starting next week.

### DIFF
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -1555,6 +1555,7 @@ thread_sched_atfork(struct rb_thread_sched *sched)
     }
     vm->ractor.sched.running_cnt = 0;
 
+    rb_native_mutex_initialize(&vm->ractor.sched.lock);
     // rb_native_cond_destroy(&vm->ractor.sched.cond);
     rb_native_cond_initialize(&vm->ractor.sched.cond);
     rb_native_cond_initialize(&vm->ractor.sched.barrier_complete_cond);


luke-gru:
Oddly enough I opened an issue regarding this behavior too, but could not reproduce it well enough to have confidence in my change. Here's my ticket, now closed: https://bugs.ruby-lang.org/issues/19395

Bo98:
> We plan on doing wider testing with it starting next week.

Any update on this?

I also have been debugging an intermittent hang on x86_64 macOS and have also deduced it to be a hang waiting on `vm->ractor.sched.lock` on a `fork`.

froydnj:
> > We plan on doing wider testing with it starting next week.
> 
> Any update on this?

We have been running with this patch in our CI for two months (hundreds of millions of tests) and haven't had any problems with it.

Bo98:
Thanks, that's useful info! Hopefully this PR will be reviewed soon and in the meantime I will apply this to downstream to Homebrew's internal Ruby at least.

byroot:
> Hopefully this PR will be reviewed soon

PRs aren't generally monitored by devs, hence why this has been sitting here for so long, it wasn't reported where devs look at bugs, see https://docs.ruby-lang.org/en/master/contributing/reporting_issues_md.html

Someone ran into the same bug and opened an issue at https://bugs.ruby-lang.org/issues/20670, so now it will be merged soon.

Bo98:
It did seem to be seen above but the tricky bit was a way to reproduce it and write a test case. I deployed this and saw that it fixed the issues as the hang rate reduced to zero, but couldn't figure out a precise test case. Thankfully someone did and they were better positioned to file https://bugs.ruby-lang.org/issues/20670.

PR #10890, opened 2024-05-31
YJIT: Fix out of bounds access when splatting empty array

XrXr:
Previously, we read the last element array even when the array was empty, doing an out-of-bounds access. This sometimes caused a SEGV.

[Bug #20496]


XrXr:
I'm setting auto-merge for this one and skipping review to reduce noise for people. It's a small straight forward patch. Hopefully that's okay with y'all 🙏🏼

k0kubun (approved):

PR #10901, opened 2024-06-03
One location for each message line

nobu:
The current message shows the location passing the block clearly, but the block passed method location is in the middle of that line.

```
-:2: warning: the block passed to 'Object#foo' defined at -:1 may be ignored
```

It is easy to jump to the caller side but tiresome to jump to the method side.

The solution to this is to split the message onto two lines and show the method being called as `info`, like this:

```
-:2: warning: the block passed to 'Object#foo' may be ignored
-:1: info: 'Object#foo' does not use a block
```


ko1:
`-:1: info: 'Object#foo' does not use a block` is not clear that `Object#foo` is defined at `-:1`, isn't it?

PR #10902, opened 2024-06-03
More on exceptions

BurdetteLamar:
I think we should be more tolerant (if not actually encouraging) about documenting the exceptions that a method may raise.


jeremyevans (commented):
I don't think we should be encourage documenting exceptions for methods in most cases.  The previous wording is preferable to me.

PR #10903, opened 2024-06-03
YJIT: Fix getconstant exits after opt_ltlt fusion

k0kubun:
Follow-up to https://github.com/ruby/ruby/pull/10401

This PR fixes a crash on SFR. When `opt_ltlt` fusion returns `SkipNextInsn`, it updates `insn_idx` variable to skip the next instruction. When the next instruction is `opt_getconstant_path`, it calls `jump_to_next_insn` before `jit.insn_idx = insn_idx`. For that reason, `opt_getconstant_path` ends up exiting with a wrong PC, using an old `insn_idx`.

It also ends the block at `opt_ltlt` like other method calls. It's probably nice to share the successor block by ending the block like in case the receiver class varies. So we removed `SkipNextInsn` and just made a jump.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -4813,6 +4813,15 @@ assert_equal [0x80000000000, 'a+', :ok].inspect, %q{
   tests
 }
 
+# test integer left shift fusion followed by opt_getconstant_path
+assert_equal '33', %q{
+  def test(a)
+    (a << 5) | (Object; a)
+  end
+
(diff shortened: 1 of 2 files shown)


maximecb:
Good work catching this bug.

One thing that seems slightly annoying is that this would seem to create more blocks, and so increase memory usage? I will merge the PR because correctness is the most important, but I guess in the medium to long term we really want to propagate constants through the context, and/or come up with a better way to do op fusion 🤔

PR #10904, opened 2024-06-03
YJIT: Add another regression test for empty splat

XrXr:
Follow-up for 6c8ae44a388e5c03b7db90376af3652007b574e8 ("YJIT: Fix out
of bounds access when splatting empty array"). This test crashes Ruby
3.3.2.


k0kubun (approved):
Nice, thanks!

PR #10905, opened 2024-06-03
YJIT: Fix out of bounds access when splatting empty array

XrXr:
This is a backport of 6c8ae44a388e5c03b7db90376af3652007b574e8 with a
test tailored to crash the 3.3.x branch (from GH-10904).

    Previously, we read the last element array even when the array was
    empty, doing an out-of-bounds access. This sometimes caused a SEGV.

    [Bug #20496]


k0kubun (approved):
Thank you!

PR #10910, opened 2024-06-04
Do not emit shape transition warnings when YJIT is compiling

casperisfine:
[[Bug #20522]](https://bugs.ruby-lang.org/issues/20522)

If `Warning.warn` is redefined in Ruby, emitting a warning would invoke Ruby code, which can't safely be done when YJIT is compiling.

cc @XrXr @k0kubun (I'll create a backport PR for `ruby_3_3` once this one is merged).

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -5013,3 +5013,37 @@ assert_normal_exit %q{
     test_body(array)
   end
 }
+
+# compiling code shouldn't emit warnings as it may call into more Ruby code
+assert_normal_exit %q{
+  # [Bug #20522]
+  Warning[:performance] = true
+
+  module StrictWarnings
+    def warn(msg, category: nil, **)
+      raise msg
+    end
+  end
+  Warning.singleton_class.prepend(StrictWarnings)
+
+  class A
+    def compiled_method(is_private)
+      @some_ivar = is_private
+    end
+  end
+
+  100.times do |i|
+    klass = Class.new(A)
+    7.times do |j|
+      obj = klass.new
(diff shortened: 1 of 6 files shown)


maximecb (approved):

k0kubun (approved):
At first I thought this was about the generated code's behavior, but it seems to be about a helper function called at compile time. Then there should be no problem with this fix 👌

casperisfine:
Yes, my description wasn't super clear. `rb_shape_get_next` is called by the YJIT compiler, not by the generated code.

bootstraptest/test_yjit.rb
```diff
@@ -5013,3 +5013,37 @@ def test_body(args) = Array(1, *args)
...
+
+  class A
+    def compiled_method(is_private)
+      @some_ivar = is_private
+    end
+  end
+
+  100.times do |i|
+    klass = Class.new(A)
+    7.times do |j|
```
XrXr, line 5037:
```suggestion
  setup_count = if defined?(RubyVM::Shape::SHAPE_MAX_VARIATIONS)
      RubyVM::Shape::SHAPE_MAX_VARIATIONS - 1
    else
     7
    end

  100.times do |i|
    klass = Class.new(A)
    setup_count.times do |j|
```

XrXr (approved):
Thanks! Nit: I thought it might be nice to use the constant from RubyVM::Shape to give more hints about the pathology, but it's whatever.

bootstraptest/test_yjit.rb
byroot: Good point.

PR #10911, opened 2024-06-04
[3.3 backport] Do not emit shape transition warnings when YJIT is compiling

casperisfine:
[[Bug #20522]](https://bugs.ruby-lang.org/issues/20522)

Ref: https://github.com/ruby/ruby/pull/10910

If `Warning.warn` is redefined in Ruby, emitting a warning would invoke Ruby code, which can't safely be done when YJIT is compiling.

cc @k0kubun


k0kubun (approved):
Thanks!

PR #10913, opened 2024-06-04
[DOC] Doc for module Errno

BurdetteLamar:
Revision for the documentation for module Errno.

### DIFF
--- a/error.c
+++ b/error.c
@@ -2768,7 +2768,7 @@ syntax_error_with_path(VALUE exc, VALUE path, VALUE *mesg, rb_encoding *enc)
  *    Errno::ENOENT.superclass # => SystemCallError
  *    Errno::ENOENT::Errno     # => 2
  *
- *  The names of nested classes are returned by method `Errno.constants`:
+ *  The names of nested classes are returned by method +Errno.constants+:
  *
  *    Errno.constants.size         # => 158
  *    Errno.constants.sort.take(5) # => [:E2BIG, :EACCES, :EADDRINUSE, :EADDRNOTAVAIL, :EADV]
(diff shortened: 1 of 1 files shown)


peterzhu2118 (commented):
I think this looks good, but do we need to use markdown for this document? I prefer if we stay consistent with RDoc in Class/Module/Method documentation.

BurdetteLamar:
I had originally intended to build a lovely GFMD table of Name/Description cells for the Errno nested classes, which would be far more compact and far prettier than the constants as autodocked.  In fact, I did build the table, but then deleted it b/c of the possible objection around maintainability, and b/c unless the constants themselves were nodocked it would be completely redundant.

If, @peterzhu2118, you think the table would be useful, I'll put it in and leave the doc as markdown;  else, I'll change the doc to rdoc.  Your call.

peterzhu2118:
I can see the difficulty of maintaining such a table. It would also be difficult to use that table because there are many errno that are system specific and may not apply to the system of the reader. It would be far too difficult to determine which errno apply to which systems, so I think it should be up to the user to determine which ones apply to them.

BurdetteLamar:
Changed from markdown to RDoc.

error.c
```diff
@@ -2741,32 +2741,47 @@ syntax_error_with_path(VALUE exc, VALUE path, VALUE *mesg, rb_encoding *enc)
...
- *  is created in module Errno, its name will start
- *  <code>Errno::</code>.
+ *    Errno::ENOENT.superclass # => SystemCallError
+ *    Errno::ENOENT::Errno     # => 2
  *
- *  The names of the <code>Errno::</code> classes depend on the
- *  environment in which Ruby runs. On a typical Unix or Windows
- *  platform, there are Errno classes such as Errno::EACCES,
- *  Errno::EAGAIN, Errno::EINTR, and so on.
+ *  The names of nested classes are returned by method `Errno.constants`:
```
peterzhu2118, line 2771:
```suggestion
 *  The names of nested classes are returned by method +Errno.constants+:
```

peterzhu2118 (approved):

PR #10916, opened 2024-06-05
Don't add `+YJIT` to `RUBY_DESCRIPTION` until it's actually enabled

casperisfine:
If you start Ruby with `--yjit-disable`, the `+YJIT` shouldn't be added until `RubyVM::YJIT.enable` is actually called. Otherwise it's confusing in crash reports etc.

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -2302,7 +2302,8 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
 #endif
 #if USE_YJIT
     if (FEATURE_SET_P(opt->features, yjit)) {
-        opt->yjit = true; // set opt->yjit for Init_ruby_description() and calling rb_yjit_init()
+        bool rb_yjit_option_disable(void);
+        opt->yjit = !rb_yjit_option_disable(); // set opt->yjit for Init_ruby_description() and calling rb_yjit_init()
     }
 #endif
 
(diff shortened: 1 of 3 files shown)


test/ruby/test_yjit.rb
```diff
@@ -67,6 +67,18 @@ def test_yjit_enable
     RUBY
   end
 
+  def test_yjit_disable
+    assert_separately(["--yjit-disable"], <<~'RUBY')
+      assert_false RubyVM::YJIT.enabled?
+      assert_false RUBY_DESCRIPTION.include?("+YJIT")
```
nobu, line 73: `assert_predicate` and `assert_include` are possible.

ruby.c
```diff
@@ -2302,7 +2302,8 @@ process_options(int argc, char **argv, ruby_cmdline_options_t *opt)
 #endif
 #if USE_YJIT
     if (FEATURE_SET_P(opt->features, yjit)) {
-        opt->yjit = true; // set opt->yjit for Init_ruby_description() and calling rb_yjit_init()
+        bool rb_yjit_option_disable(void);
+        opt->yjit = !rb_yjit_option_disable(); // set opt->yjit for Init_ruby_description() and calling rb_yjit_init()
```
k0kubun, line 2306: Now that you have this, could you change https://github.com/ruby/ruby/blob/f7b53a75b648e7156f49c1d5c266e2d85f159fc6/yjit/src/yjit.rs#L37 to just `if yjit_enabled {`? The fact that we still have the duplicated check in `rb_yjit_init()` confuses me.

k0kubun (approved):
This is a nice change. We could ship this fix in 3.3.3 too.

ruby.c
maximecb: Seconded

casperisfine: Done.

maximecb (approved):

PR #10920, opened 2024-06-05
Don't add `+YJIT` to `RUBY_DESCRIPTION` until it's actually enabled

casperisfine:
If you start Ruby with `--yjit-disable`, the `+YJIT` shouldn't be added until `RubyVM::YJIT.enable` is actually called. Otherwise it's confusing in crash reports etc.

Backport of https://github.com/ruby/ruby/pull/10916 as suggested by @k0kubun


k0kubun (approved):
Thanks!

PR #10924, opened 2024-06-06
Hash#inspect with colon style

tompng:
Implements https://bugs.ruby-lang.org/issues/20433#note-7
Change Hash#inspect to use colon style if the key is a symbol.
When key is not a symbol, space is added around `=>`
```ruby
puts({a!: 1, "<=>" => 1, "a=": 1, 42 => 1}.inspect)
# => {a!: 1, "<=>" => 1, "a=": 1, 42 => 1}
# it was {:a!=>1, "<=>"=>1, :a==>1, 42=>1}
```

Symbol key that needs quote can be determined by this
```ruby
chars = (32..127).map(&:chr)+["☆"]-[*'b'..'z',*'B'..'Z',*'1'..'9']
(1..3).each do |n|
  symbols = chars.repeated_permutation(n).map(&:join).map(&:to_sym)
  # Found symbols that needs quote
  a = symbols.reject do |sym|
    k = sym.inspect[1..]
    begin;{sym=>1} == eval("{#{k}: 1}");rescue SyntaxError;false;end
  end
  # Determine logic
  b = symbols.select do |sym|
    s = sym.inspect
    s[1]=='@'||s[1]=='$'||s[1]=='!'||"%&*+-/<=>@]^`|~".include?(s[-1])
  end
  p a == b
end
```

### DIFF
--- a/bootstraptest/test_literal.rb
+++ b/bootstraptest/test_literal.rb
@@ -119,14 +119,14 @@ assert_equal '5',               'a = [1,2,3]; a[1] = 5; a[1]'
 assert_equal 'bar',             '[*:foo];:bar'
 assert_equal '[1, 2]',          'def nil.to_a; [2]; end; [1, *nil]'
 assert_equal '[1, 2]',          'def nil.to_a; [1, 2]; end; [*nil]'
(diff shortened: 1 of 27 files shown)


tompng:
test-bundled-gems (debug) still fails but it is because debug(tag=v1.9.2) does not contain fixed test code yet.

PR #10925, opened 2024-06-06
Remove circular parameter syntax error

kddnewton:
As per https://bugs.ruby-lang.org/issues/20478

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -539,8 +539,6 @@ struct parser_params {
         int end_col;
     } delayed;
 
-    ID cur_arg;
-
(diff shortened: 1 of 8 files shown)


nobu:
Files under "spec/ruby" as well.
[spec/ruby/language/block_spec.rb](https://github.com/ruby/ruby/compare/master...nobu:ruby:cur_arg-error#diff-20e15cf326bdc3225d6813c65b1dadfffefdcfed254454cb8edf0d720fb52559)

kddnewton:
Ahh thanks @nobu!

PR #10753, opened 2024-05-09
Update ruby/spec

andrykonchin:
Import recent changes in ruby/spec.

The previous PR #10422

PR #10937, opened 2024-06-07
Update ruby/spec

andrykonchin:
Import recent changes in ruby/spec.

The previous PR - #10753


eregon:
```
1)
  RUBY_DESCRIPTION contains version FAILED
  Expected "description_with_yjit".include? "3.4.0"
  to be truthy but was false
```

@nobu @k0kubun `description_with_yjit` is a weird value for `RUBY_DESCRIPTION`.
Is it a bug it's visible inside tests/specs?
If not, how can we detect that we are running from built-but-not-installed CRuby, so we can skip the spec in that case? (IMO this is an example why we should test on installed CRuby, it does not have such problems).

eregon:
I found:
https://github.com/ruby/spec/blob/517f06f66d533175d8586d222156323b89a43909/library/rubygems/gem/load_path_insert_index_spec.rb#L5
and
https://github.com/ruby/spec/blob/517f06f66d533175d8586d222156323b89a43909/library/rbconfig/rbconfig_spec.rb#L117-L125
So we can use that to detect this case.

It still seems messy that RUBY_DESCRIPTION is wrong while running tests.

k0kubun:
I think `template/fake.rb.in` got broken. I don't know much about that file or why it even exists, but @nobu would know.

andrykonchin:
All the failures seem fixed so the PR is ready for reviewing/merging

PR #10939, opened 2024-06-08
Ensure prism is testing everything

kddnewton:


hsbt:
Is missing `make test-syntax-suggest` intentional?

kddnewton:
Ahh that was a mistake, I've added it now. It passes locally, so hopefully there's no issue on CI.

PR #10941, opened 2024-06-09
Introduce `ident_or_const` inline rule

S-H-GAMELINKS:
Introduce `ident_or_const` inline rule.
And, it replace to `tIDENTIFIER` and `tCONSTANT` match pattern replaced.


ydah:
Of the parameterization rules, inlines are difficult to read if the definition position is far away, so we are changing them so that they can be defined where the generation rules are defined.
- https://github.com/ruby/lrama/pull/420
So, please wait a little longer use Inline rules.

nobu (approved):

PR #10944, opened 2024-06-09
[Bug #20154] Extract hardening CFLAGS to a special $hardenflags variable

KJTsanaktsidis:
This changes the automatic detection of -fstack-protector, -D_FORTIFY_SOURCE, and -mbranch-protection to write to $hardenflags instead of $XCFLAGS. The definition of $cflags is changed to "$hardenflags $orig_cflags $optflags $debugflags $warnflags" to match.

Furthermore, these flags are _prepended_ to $hardenflags, rather than appended.

The implications of doing this are as follows:

* If a CRuby builder specifies cflags="-mbranch-protection=foobar" at the ./configure script, and the configure script detects that -mbranch-protection=pac-ret is accepted, then GCC will be invoked as "gcc -mbranch-protection=pac-ret -mbranch-protection=foobar". Since the last flags take precedence, that means that user-supplied values of these flags in $cflags will take priority.
* Likewise, if a CRuby builder explicitly specifies "hardenflags=-mbranch-protection=foobar", because we _prepend_ to $hardenflags in our autoconf script, we will still invoke GCC as "gcc -mbranch-protection=pac-ret -mbranch-protection=foobar".
* If a CRuby builder specifies CFLAGS="..." at the configure line, automatic detection of hardening flags is ignored as before.
* C extensions will _also_ be built with hardening flags now as well (this was not the case by default before because the detected flags went into $XCFLAGS).

Additionally, as part of this work, I changed how the detection of PAC/BTI in Context.S works. Rather than appending the autodetected option to ASFLAGS, we simply compile a set of test programs with the actual CFLAGS in use to determine what PAC/BTI settings were actually chosen by the builder. Context.S is made aware of these choices through some custom macros.

The result of this work is that:

* Ruby will continue to choose some sensible defaults for hardening options for the C compiler
* Distributors are able to specify CFLAGS that are consistent with their distribution and override these defaults
* Context.S will react to whatever -mbranch-protection is actually in use, not what was autodetected
* Extensions get built with hardening flags too.

[Bug #20154]
[Bug #20520]

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -360,7 +360,7 @@ test -z "$warnflags" ||
 AS_IF([test -z "${CFLAGS+set}"], [
     cflags=`echo " $cflags " | sed "$cflagspat;s/^ *//;s/ *$//"`
     orig_cflags="$cflags"
-    cflags="$cflags "'${optflags} ${debugflags} ${warnflags}'
+    cflags='${hardenflags} '"$cflags "'${optflags} ${debugflags} ${warnflags}'
 ])
 dnl AS_IF([test -z "${CXXFLAGS+set}"], [
 dnl     cxxflags=`echo " $cxxflags " | sed "$cflagspat;s/^ *//;s/ *$//"`
@@ -813,7 +813,7 @@ AS_IF([test "$GCC" = yes], [
 		  [fortify_source=$enableval])
     AS_IF([test "x$fortify_source" != xno], [
         RUBY_TRY_CFLAGS([$optflags -D_FORTIFY_SOURCE=2],
-                        [RUBY_APPEND_OPTION(XCFLAGS, -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2)], [],
+                        [RUBY_PREPEND_OPTION(hardenflags, -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2)], [],
                         [@%:@include <stdio.h>])
     ])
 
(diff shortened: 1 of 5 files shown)


KJTsanaktsidis:
@nobu - You're the expert on Autoconf - does this look good to you? 🙏 
@ioquatix - the Context.S changes - are you OK with this too?

Thanks!

ioquatix (approved):
LGTM!

configure.ac
```diff
@@ -995,6 +999,52 @@ AC_SUBST(incflags, "$INCFLAGS")
 test -z "${ac_env_CFLAGS_set}" -a -n "${cflags+set}" && eval CFLAGS="\"$cflags $ARCH_FLAG\""
 test -z "${ac_env_CXXFLAGS_set}" -a -n "${cxxflags+set}" && eval CXXFLAGS="\"$cxxflags $ARCH_FLAG\""
 
+# The lines above expand out the $cflags/$optflags/$debugflags/$hardenflags variables into the
+# CFLAGS variable. So, at this point, we have a $CFLAGS var with the actual compiler flags we're
+# going to use.
+# That means this is the right time to check what branch protection flags are going to be in use
+# and define appropriate macros for use in Context.S based on this
+AS_CASE(["$target_cpu"], [aarch64], [
+    AC_MSG_CHECKING(whether __ARM_FEATURE_BTI_DEFAULT is defined)
```
nobu, line 1008: Check every time, not using `AC_CACHE_CHECK`?

KJTsanaktsidis: I actually didn't think about caching. but, thinking about it now, it should be uncached - these defines will change depending on the cflags used. (Or does the autoconf cache invalidate if cflags are changed? I’ll try it tomorrow)

KJTsanaktsidis: Good news, looks like autoconf is smart enough to bust the cache here.

```
ktsanaktsidis@lima-fedora build % ../configure  --prefix=$(realpath ~/rubies/3.4.0-master) --disable-install-doc --disable-yjit cflags="-mbranch-protection=standard" --config-cache
configure: loading cache config.cache
configure: error: `cflags' has changed since the previous run:
configure:   former value:  `-mbranch-protection=none'
configure:   current value: `-mbranch-protection=standard'
configure: error: in `/home/ktsanaktsidis.linux/ruby/build':
configure: error: changes in the environment can compromise the build
configure: error: run `make distclean' and/or `rm config.cache'
	    and start over
```

So i'll wrap it in `AC_CACHE_CHECK`.

ioquatix:
Nice work, the simplifications to `Context.S` are meaningful.

PR #10946, opened 2024-06-10
respect `RUBY_TEST_SUBPROCESS_TIMEOUT_SCALE`

ko1:
GC benchmarks will spend long time with assertions so we need to respect `RUBY_TEST_SUBPROCESS_TIMEOUT_SCALE` environment variable.


nobu:
Not `RUBY_TEST_TIMEOUT_SCALE`?

PR #10947, opened 2024-06-10
Check current file on TracePoint

ko1:
```
1)
TracePoint#inspect returns a String showing the event, method, path and line for a :call event FAILED
Expected "#<TracePoint:call 'call' /tmp/ruby/src/trunk/spec/ruby/core/objectspace/define_finalizer_spec.rb:33>" =~ /\A#<TracePoint:call [`']trace_point_spec_test_call' \/tmp\/ruby\/src\/trunk\/spec\/ruby\/core\/tracepoint\/inspect_spec.rb:43>\z/
to be truthy but was nil
```

This kind of failures comes because of finaizers.
So check the current file or not.

### DIFF
--- a/spec/ruby/core/tracepoint/inspect_spec.rb
+++ b/spec/ruby/core/tracepoint/inspect_spec.rb
@@ -24,6 +24,8 @@ describe 'TracePoint#inspect' do
     line = nil
     TracePoint.new(:line) { |tp|
       next unless TracePointSpec.target_thread?
+      next unless tp.path == __FILE__
+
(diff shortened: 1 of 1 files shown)


eregon:
Thanks.

If finalizers were run in a separate thread it wouldn't be a problem :)
(I think we should really do that)

We could also undefine the finalizer for https://github.com/ruby/ruby/blob/master/spec/ruby/core/objectspace/define_finalizer_spec.rb#L33 but that seems more brittle so not a good solution.

PR #10951, opened 2024-06-10
compile.c: support RubyVM::FrozenCore special class

casperisfine:
When using `shareable_constant_value: literal`, the instruction sequence might include `putobject RubyVM::FrozenCore`, and frozen core being a special hidden ICLASS, it needs special handling in `ibf_dump` and `ibf_load`.


nobu (approved):
It was because moved from parser.

PR #10953, opened 2024-06-10
Don't skip test_inspect_under_gc_compact_stress

peterzhu2118:
Commit 1471a16 seems to have fixed this flaky test, so we don't need to skip it for YJIT or RJIT anymore.


mame (approved):

k0kubun:
> 1471a16

Which commit was this? https://github.com/ruby/ruby/commit/1471a16 doesn't seem to exist, at least not on GitHub as of the time of this writing.

PR #10956, opened 2024-06-10
YJIT: add context cache size stat, increase cache size

maximecb:
It seems easy to justify increasing the cache size for now, given that large apps will allocate a lot more context data.

Before:
```
context_cache_bytes:          16,384
context_data_bytes:          237,988
num_contexts_encoded:        133,925
bytes_per_context:              1.78
```

After:
```
context_cache_bytes:          32,768
context_data_bytes:          210,618
num_contexts_encoded:        133,903
bytes_per_context:              1.57
```

Curious what these stats will look like on SFR 🤔


yjit/src/core.rs
```diff
@@ -919,7 +924,7 @@ impl Context {
     {
         unsafe {
             if CTX_CACHE == None {
-                CTX_CACHE = Some( [(Context::default(), 0); CTX_CACHE_SIZE] );
+                CTX_CACHE = CtxCache::default();
```
k0kubun, line 927: Since `CtxCache` is an `Option`, is this `if CTX_CACHE == None { CTX_CACHE = None; }`?

maximecb: Aah I see what you mean 😆

k0kubun:
To make the stats more accessible on SFR, I made `num_contexts_encoded` a default counter https://github.com/ruby/ruby/commit/ec1ea2c5b941d5ff94013772a0dba905f8fe5c78. I'll ship those metrics to SFR shortly.

maximecb:
> To make the stats more accessible on SFR, I made `num_contexts_encoded` a default counter [ec1ea2c](https://github.com/ruby/ruby/commit/ec1ea2c5b941d5ff94013772a0dba905f8fe5c78). I'll ship those metrics to SFR shortly.

Thanks Kokubun

I should note that I encountered a bug when I tried to increase the context cache size to 1024. I added an extra `debug_assert!`, but I haven't been able to figure out where it comes from yet. I've left the cache size at 512 for now, but I suspect there is still a bug hiding somewhere in the context encoding/decoding.

PR #10969, opened 2024-06-11
YJIT: Delete otherwise-empty defer_compilation() blocks

XrXr:
Calls to defer_compilation() leave behind a stub and a `struct Block`
that we retain. If the block is empty, it's only reason for existing is
to hold the `struct Branch` that the stub needs.

This patch transplants the branch out of the empty block into the newly
generated block when the defer_compilation() stub is hit, and deletes
the empty block to save memory.

To assist the transplantation, `Block::outgoing` is now a
`MutableBranchList`, and `Branch::Block` now in a `Cell`. These types
don't incur a size cost.

On the `lobsters` benchmark, `yjit_alloc_size` is roughly 98% of what
it was before the change.

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -1293,7 +1293,7 @@ struct BranchStub {
 /// Note: care must be taken to minimize the size of branch objects
 pub struct Branch {
     // Block this is attached to
-    block: BlockRef,
+    block: Cell<BlockRef>,
 
     // Positions where the generated code starts and ends
     start_addr: CodePtr,
@@ -1433,7 +1433,7 @@ impl PendingBranch {
     fn into_branch(mut self, uninit_block: BlockRef) -> BranchRef {
         // Make the branch
         let branch = Branch {
-            block: uninit_block,
+            block: Cell::new(uninit_block),
             start_addr: self.start_addr.get().unwrap(),
             end_addr: Cell::new(self.end_addr.get().unwrap()),
             targets: self.targets,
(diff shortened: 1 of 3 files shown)


yjit/src/stats.rs
```diff
@@ -553,6 +553,7 @@ make_counters! {
     block_next_count,
     defer_count,
     defer_empty_count,
+    deleted_defer_block_count,
```
k0kubun, line 556: WDYT about adding this to `DEFAULT_COUNTERS`? I want to monitor the value whether `--yjit-stats` is given or not.

XrXr:
crash in `--enable-yjit=dev, --yjit-call-threshold=1` job at test_array.rb:979 with

```
  /home/runner/work/ruby/ruby/build/ruby(rb_unexpected_type) (null):0
  /home/runner/work/ruby/ruby/build/ruby(RB_FL_TEST_RAW+0x0) [0x55f2b4ff8bd2] ../src/include/ruby/internal/value_type.h:446
  /home/runner/work/ruby/ruby/build/ruby(RB_FL_ANY_RAW+0x0) [0x55f2b55e0c49] ../src/include/ruby/internal/core/rarray.h:257
  /home/runner/work/ruby/ruby/build/ruby(rb_array_len) ../src/include/ruby/internal/core/rarray.h:259
  /home/runner/work/ruby/ruby/build/ruby(rb_ary_resurrect+0xd) [0x55f2b55eebcd] ../src/array.c:2726
```
Probably relevant. Will need to debug that.

yjit/src/core.rs
```diff
@@ -3455,6 +3472,44 @@ fn branch_stub_hit_body(branch_ptr: *const c_void, target_idx: u32, ec: EcPtr) -
...
+                    mem::drop(housing_block.outgoing.0.take());
+                    new_block.outgoing.push(branch_ref);
+                    let housing_block: BlockRef = branch.block.replace(new_block.into());
+                    // Free the old housing block. There should now be no live &Block.
+                    remove_block_version(&housing_block);
+                    unsafe { free_block(housing_block, false) };
+
+                    incr_counter!(deleted_defer_block_count);
+                }
+            }
```
maximecb, line 3511: `branch_stub_hit_body` is already a huge function. Do you think this logic could be moved into its own function to make the code more readable?

yjit/src/core.rs
```diff
@@ -3457,6 +3477,8 @@ fn branch_stub_hit_body(branch_ptr: *const c_void, target_idx: u32, ec: EcPtr) -
             // Branch shape should reflect layout
             assert!(!(branch.gen_fn.get_shape() == target_branch_shape && new_block.start_addr != branch.end_addr.get()));
 
+            delete_empty_defer_block(branch, new_block, target_ctx, target_blockid);
```
maximecb, line 3480: Can you add a comment above here, just a one-sentence explanation of what this call does?

maximecb (approved):

PR #3508, opened 2020-09-03
Improve support for warning categories

jeremyevans:
This adds support for the `category` keyword for `Kernel#warn` (for Ruby level warnings), and adds `rb_category_warn` and `rb_category_warning` C-API functions (for C level warnings).  It changes `Warning.warn` to only accept a `category` keyword, instead of accepting arbitrary keywords.  It modifies all of the warnings emitted by core Ruby to use a category.

This also removes some deprecated functions/methods that were scheduled to be removed in Ruby 3.0 that I found while looking for warnings to modify.

PR #10978, opened 2024-06-12
Kernel#warn: don't call `Warning.warn` unless the category is enabled

casperisfine:
[[Bug #20573]](https://bugs.ruby-lang.org/issues/20573)

Followup: https://github.com/ruby/ruby/pull/10960

I believe `Kernel#warn` should behave in the same way than internal `rb_warning_* APIs

cc @tenderlove @jeremyevans

### DIFF
--- a/spec/ruby/core/warning/warn_spec.rb
+++ b/spec/ruby/core/warning/warn_spec.rb
@@ -121,6 +121,30 @@ describe "Warning.warn" do
     end
   end
 
+  ruby_bug '#19530', ''...'3.4' do
+    it "isn't called by Kernel.warn when category is :deprecated but Warning[:deprecated] is false" do
+      warn_deprecated = Warning[:deprecated]
+      begin
+        Warning[:deprecated] = false
+        Warning.should_not_receive(:warn)
+        Kernel.warn("foo", category: :deprecated)
+      ensure
+        Warning[:deprecated] = warn_deprecated
+      end
+    end
+
+    it "isn't called by Kernel.warn when category is :experimental but Warning[:experimental] is false" do
+      warn_experimental = Warning[:experimental]
+      begin
+        Warning[:experimental] = false
+        Warning.should_not_receive(:warn)
+        Kernel.warn("foo", category: :experimental)
+      ensure
+        Warning[:experimental] = warn_experimental
+      end
+    end
+  end
+
   it "prints the message when VERBOSE is false" do
     -> { Warning.warn("foo") }.should complain("foo")
   end
--- a/warning.rb
+++ b/warning.rb
@@ -47,7 +47,13 @@ module Kernel
   #                be removed in the future.
   # :experimental :: Used for experimental features that may change in
   #                  future releases.
+  # :performance  :: Used for warning about APIs or pattern that have
+  #                  negative performance impact
   def warn(*msgs, uplevel: nil, category: nil)
-    Primitive.rb_warn_m(msgs, uplevel, category)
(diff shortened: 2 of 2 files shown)


warning.rb
```diff
@@ -44,10 +44,16 @@ module Kernel
...
+  #                  be removed in the future.
   # :experimental :: Used for experimental features that may change in
   #                  future releases.
+  # :performance  :: Used for warning about APIs or pattern that have
+  #                  negative performance impact
   def warn(*msgs, uplevel: nil, category: nil)
-    Primitive.rb_warn_m(msgs, uplevel, category)
+    if category.nil?
+      Primitive.rb_warn_m(msgs, uplevel, nil)
+    elsif Warning[Primitive.cexpr!("rb_to_symbol_type(category)")]
```
nobu, line 55: > I believe `Kernel#warn` should behave in the same way than internal `rb_warning_`* APIs

These APIs accept `Symbol` only, not object with `to_sym`, I think.

```suggestion
    elsif Warning[category]
```

warning.rb
```diff
@@ -44,10 +44,16 @@ module Kernel
...
   # :deprecated :: Used for warning for deprecated functionality that may
-  #                be removed in the future.
+  #                  be removed in the future.
   # :experimental :: Used for experimental features that may change in
   #                  future releases.
+  # :performance  :: Used for warning about APIs or pattern that have
+  #                  negative performance impact
   def warn(*msgs, uplevel: nil, category: nil)
-    Primitive.rb_warn_m(msgs, uplevel, category)
+    if category.nil?
```
nobu, line 53: This works unexpectedly if `category.nil?` is overridden.

```suggestion
    if Primitive.cexpr!("NIL_P(category)")
```

warning.rb
casperisfine: I thought so as well, but @jeremyevans added a rubyspec to allow `warn msg, category: "deprecated"` in https://github.com/ruby/ruby/pull/3508

nobu: I see.
The APIs do not accept other than `Symbol`s but `Kernel#warn` method coerces `category` to a `Symbol`.

Another small topic, to get rid of the coercing it twice:

```suggestion
    elsif Warning[category = Primitive.cexpr!("rb_to_symbol_type(category)")]
```

warning.rb
```diff
@@ -44,10 +44,16 @@ module Kernel
   # following categories:
   #
   # :deprecated :: Used for warning for deprecated functionality that may
-  #                be removed in the future.
+  #                  be removed in the future.
```
jeremyevans, line 47: This looks like it would result in misformatted documentation.

spec/ruby/core/warning/warn_spec.rb
```diff
@@ -121,6 +121,30 @@ def Warning.warn(msg)
...
+      begin
+        Warning[:deprecated] = false
+        Warning.should_not_receive(:warn)
+        Kernel.warn("foo", category: :deprecated)
+      ensure
+        Warning[:deprecated] = warn_deprecated
+      end
+    end
+
+    it "isn't called by Kernel.warn when category is :experimental but Warning[:deprecated] is false" do
```
jeremyevans, line 136: `s/Warning[:deprecated]/Warning[:experimental]/`

jeremyevans (changes requested):
I agree with the idea.  I found a couple of small issues that I think should be fixed first.

warning.rb
casperisfine: Thanks. corrected.

PR #10979, opened 2024-06-12
YJIT: add context cache hits stat

maximecb:
This stat should make more sense when it comes to interpreting the effectiveness of the cache on large deployed apps.


k0kubun (approved):
:+1:

PR #10984, opened 2024-06-12
Fix freeing symbols when RUBY_FREE_AT_EXIT

peterzhu2118:
Dynamic symbols point to a fstring. When we free the symbol, we hash the fstring to remove it from the table. However, the fstring could have already been freed, which can cause a crash.

This commit changes it to remove the reference to the fstring before freeing the symbol so we can avoid this crash.


flavorjones:
This branch fixes the reported issue on my machine. :+1:

PR #10987, opened 2024-06-13
Use ENDBR instruction in amd64 coroutine on OpenBSD

jeremyevans:
When running on newer Intel processors supporting the feature, OpenBSD enforces indirect branch tracking.  Without this endbr64 instruction, jumps to the coroutine_transfer function result in SIGILL on OpenBSD/amd64 when using such processors.

The OpenBSD Ruby ports have been using a patch similar to this for the past two months.

From some research, cet.h has been supported by GCC for about 6 years and LLVM for about 4 years.

### DIFF
--- a/coroutine/amd64/Context.S
+++ b/coroutine/amd64/Context.S
@@ -5,6 +5,10 @@
 ##  Copyright, 2018, by Samuel Williams.
 ##
 
+#if defined(__OpenBSD__)
+#include <cet.h>
+#endif
+
 #define TOKEN_PASTE(x,y) x##y
 #define PREFIXED_SYMBOL(prefix,name) TOKEN_PASTE(prefix,name)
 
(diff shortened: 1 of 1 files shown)


ioquatix (approved):
Thanks this looks good to me.

My only concern is whether we should feature detect rather than assuming `__OpenBSD__`, e.g. `have_header(cet.h)` in configure etc.

jeremyevans:
> My only concern is whether we should feature detect rather than assuming `__OpenBSD__`, e.g. `have_header(cet.h)` in configure etc. Will other BSDs adopt (or already support) similar mechanisms?

I'm not sure.  I'd like to start out with this to solve the immediate problem.  We can modify it later if necessary.

> Also, will we need to support other architectures in the future?

The arm64 coroutine has BTI  support already (arm counterpart of Intel IBT). I think currently it's only needed on amd64.  If OpenBSD adds IBT support on i386, we could update the x86 coroutine to support it.

PR #10996, opened 2024-06-13
Fix yjit bootstrap tests

eightbitraptor:
This PR fixes some guards in the yjit bootstrap tests that cause errors on platforms that don't implement GC compaction.


XrXr (commented):
Thanks! I think we can `s/return/skip/g` instead for a cleaner diff. Ractor tests use the skip method a lot.

eightbitraptor:
Thanks @XrXr - That's a much better approach!

PR #10997, opened 2024-06-13
rb_str_hash(): Avoid UB with making misaligned pointer

XrXr:
Previously, on common platforms, this code made a pointer to a union of
8 byte alignment out of a char pointer that is not guaranteed to satisfy
the alignment requirement. That is undefined behavior according
to [C99 6.3.2.3p7](https://port70.net/~nsz/c/c99/n1256.html#6.3.2.3p7).

Use memcpy() to do the unaligned read instead.

cc @byroot

### DIFF
--- a/string.c
+++ b/string.c
@@ -376,9 +376,8 @@ str_precompute_hash(VALUE str)
     RUBY_ASSERT(free_bytes >= sizeof(st_index_t));
 #endif
 
-    typedef struct {char bytes[sizeof(st_index_t)];} unaligned_index;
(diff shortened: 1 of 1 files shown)


byroot:
Should the other callsite be modifed too? https://github.com/ruby/ruby/blob/e8bb729ac26ef97efc02d4c4e9309453149aedb8/string.c#L379-L381

XrXr:
The other one is found AFAICT. `_Align_of(unaligned_index) == 1` so it's always aligned. memcpy can make the code read better, though.

byroot:
Yeah you are the expert, but intuitively I'd use the same code both both call sites.

PR #10998, opened 2024-06-13
Enable LeakChecker for RJIT previously disabled for MJIT

XrXr:
RJIT doesn't spawn subprocesses so there should now be no need to
special case it.


k0kubun (approved):
:+1:

PR #11001, opened 2024-06-14
Fixed indents in recently RJIT patch

ukolovda:
@k0kubun  I fix indents in previous patch. 
Please, apply it, sorry.


k0kubun (approved):
:ok_hand:

PR #11004, opened 2024-06-14
YJIT: `--yjit-dump-disasm=dir`: Hold descriptor for dump file

XrXr:
This mainly aims to make `--yjit-dump-disasm=<relative_path>` more
usable. Previously, it crashed if the program did chdir(2), since it
opened the dump file every time when appending.

Tested with:

    ./miniruby --yjit-dump-disasm=. --yjit-call-threshold=1 -e 'Dir.chdir("/") {}'

And the `lobsters` benchmark.


rwstauner (approved):
Confirmed this used to (try to) write two files and now writes all the data into one.

k0kubun (approved):
nice :+1:

PR #11005, opened 2024-06-16
.github/workflows/windows.yml - fix use of vcpkg,json

MSP-Greg:
The `vcpkg.json` file must exist before `vcpkg install` is run, and the command must run in the file's directory.


MSP-Greg:
Currently the windows.yml workflow has a 'Install libraries with vcpkg' step.  This can be verbose.

Added a 'List vcpkg libraries' step that logs the library packages and their versions.  There is a concern that library package upgrades may cause CI failures, and placing the info in a separate step makes it easy to compare package versions.

MSP-Greg:
Closing in favor of #11014

PR #11006, opened 2024-06-17
[Bug #20581][3.3] Fix unintentional truncation for dependencies of bundled gems

hsbt:
https://bugs.ruby-lang.org/issues/20581

### DIFF
--- a/tool/rbinstall.rb
+++ b/tool/rbinstall.rb
@@ -1097,12 +1097,12 @@ install?(:ext, :comm, :gem, :'bundled-gems') do
       # This gemspec keep the original dependencies
       path = "#{srcdir}/.bundle/gems/#{gem_name}/#{gem_name}.gemspec"
       unless File.exist?(path)
-          # Try to find the gemspec file for gems that hasn't own gemspec
-          path = "#{srcdir}/.bundle/specifications/#{gem_name}.gemspec"
-          unless File.exist?(path)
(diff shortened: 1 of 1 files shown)


crevete:
@hsbt Hi, this fix will be included in the next ruby maintenance release (3.3.4) ? If so, when it will be cut? thanks

schneems:
Thank you @hsbt for your work on this. As a workaround, anyone seeing this failure:

```
Downloading net-pop-0.1.2 revealed dependencies not in the API or the lockfile
(net-protocol (>= 0)).
Running `bundle update net-pop` should fix the problem.
```

Can add this to their `Gemfile` and it will fix the issue until Ruby 3.3.4 is released:

```ruby
# Needed until Ruby 3.3.4 is released https://github.com/ruby/ruby/pull/11006
gem 'net-pop', github: 'ruby/net-pop'
```

I've tested and confirmed that this method works on Heroku, (where people cannot manually add gems outside of `bundle install`).

k0kubun:
> If so, when it will be cut?

I'm planning to release 3.3.4 between August 12th and September 12th. Going forward, I intend to make a regular release of Ruby 3.3 every 2 or 3 months (which could become shorter when there's a CVE), and Ruby 3.3.3 was released on June 12th.

k0kubun (approved):
Thanks for fixing this!

PR #11014, opened 2024-06-18
windows.yml - fixup vcpkg code, list packages

MSP-Greg:
Correctly updates the vcpkg library packages, adds a step 'List vcpkg libraries' that lists all packages and their versions.  This can be helpful if a library package update is suspected of causing a CI failure.

Note that the file `vcpkg.json` was added to the repo, but the workflow was never updated to work with it.  So, it only works with a local installation when used properly.

Note that Windows Rubies 3.2.4 and 3.3.3 are using OpenSSL 3.3.z and zlib 1.3.1.  They are both MSYS2 ucrt64 builds.

Packages will be updated as required, based on the vcpkg version included in the GHA runner image.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -115,7 +115,12 @@ jobs:
 
       - name: Install libraries with vcpkg
         run: |
-          vcpkg --triplet x64-windows install gmp libffi libyaml openssl zlib
+          set pkgs=gmp libffi libyaml openssl zlib
+          vcpkg --triplet x64-windows install %pkgs%
+          vcpkg --triplet x64-windows upgrade %pkgs% --no-dry-run
+
+      - name: List vcpkg libraries
+        run: vcpkg list
 
       - name: Install libraries with scoop
         run: |


nobu:
> Note that the file `vcpkg.json` was added to the repo, but the workflow was never updated to work with it. So, it only works with a local installation when used properly.

Tried to install using manifest mode.
https://github.com/ruby/ruby/commit/463ebb8f3f665a184958933a1186dbf7df0e323f
This takes 20min to install packages.

MSP-Greg:
When one switches to 'manifest' mode, at least one of the vcpkg caches should be moved, as the packages are installed in the same folder as the json file.  That may be part of the problem.

For GHA, an archive file (7z?) is generated once a day from vcpkg master.  That could be used, but that's a 'rolling release'.

All of the default and bundled gems that test with mswin use it.

There's two issues here...

1. Should the repo lock library package versions?

2. Fix the CI.

Re the first issue, I think Ruby head should use the most current library packages available.  I believe both MSYS2 and MSFT/vcpkg recommend that.


> This takes 20min to install packages.

Without the cache, that's probably a full build.

MSP-Greg:
@nobu

The code behind setup-ruby and its installation of WIndow's build tools/packages seems to be reliable.  The only recent issue I can recall is when MSYS2 decided to split a package into two packages, a dev and runtime, similar to some packages on *nix systems.

So, I replaced the vcpkg code here with code to download the mswin package files.  The commit is [here](https://github.com/ruby/ruby/commit/bd88fbefda0a31a883e9d3fb64c5e3c04e69bc17) and the Actions run is [here](https://github.com/MSP-Greg/ruby/actions/runs/9586103860).  It is using the same archive file for all jobs, it's compiled with VS 2022.

Any thoughts?  It's quite a bit quicker.

hsbt:
I don't want to upgrade with rolling releases like homebrwe, vcpkg. It's fragile on old branches like `ruby_3_1`.

I'm okay to use manifest mode with `builtin-baseline`.

MSP-Greg:
@hsbt 

You mentioned Homebrew & vcpkg, MSYS2 is also a rolling release.

These two PRs for Windows workflows are for master branch.   Are you ok with using rolling releases for master?  'stable' branches are another issue.

An issue with some of the workflows is that some cache keys are not job specific.   Or, if workflow changes are backported, a master cache could be incorrectly used on a stable branch job.

In the 'bug tracker' you mentioned a problem.  Do you recall the exact details?  Both of the PRs added code to list dependent package versions.

Also, re 3.1 mingw, it is using an incorrect base Ruby, which loads ucrt tools with OpenSSL 3.  I think I have a patch for the mingw workflow, haven't looked at the mswin.  I haven't looked at the 3.2 or 3.3 workflows.

hsbt:
>These two PRs for Windows workflows are for master branch. Are you ok with using rolling releases for master? 'stable' branches are another issue.

No. master will become stable branch every year.

hsbt:
I know you work to support mswin and mingw platform. Thank you for that.

But It's your work, not me. I spent a lot of time to keep stable status for `ruby_3_1` like:

* https://github.com/ruby/ruby/pull/10840
* https://github.com/ruby/ruby/pull/10838
* https://github.com/ruby/ruby/pull/10814
* https://github.com/ruby/ruby/pull/10784
* https://github.com/ruby/ruby/pull/10512
* https://github.com/ruby/ruby/pull/7935
* https://github.com/ruby/ruby/pull/6468

The above PR is all for differences with old branch and modern platform. I don't wont to expamnd these differences for mswin platform.

PR #11019, opened 2024-06-19
Avoid a hash allocation when keyword splatting empty hash when calling ruby2_keywords method

jeremyevans:
Treat this similar to keyword splatting nil, using goto ignore. However, keep previous behavior if the method accepts a keyword splat, to avoid double hash allocation.

This also can avoid an array allocation when calling a method that doesn't have any splat parameters but supports literal keyword parameters, because ignore_keyword_hash_p was not ignoring the keyword hash in that case.

This change doesn't remove the empty ruby2_keywords hash from the array, which caused an assertion failure if the method being called accepted keywords in some cases.  Modify the assertion to handle this case.  An alternative approach would add a flag to the args struct so the args_argc calculation could handle this case and report the correct argc, but such an approach would likely be slower.

### DIFF
--- a/test/ruby/test_allocation.rb
+++ b/test/ruby/test_allocation.rb
@@ -273,7 +273,7 @@ class TestAllocation < Test::Unit::TestCase
         check_allocations(0, 1, "keyword(**hash1, **empty_hash#{block})")
         check_allocations(1, 0, "keyword(*empty_array, *empty_array, **empty_hash#{block})")
 
-        check_allocations(1, 0, "keyword(*r2k_empty_array#{block})")
+        check_allocations(0, 0, "keyword(*r2k_empty_array#{block})")
         check_allocations(1, 1, "keyword(*r2k_array#{block})")
 
         check_allocations(0, 1, "keyword(*empty_array, a: 2, **empty_hash#{block})")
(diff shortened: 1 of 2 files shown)


vm_args.c
```diff
@@ -856,7 +860,15 @@ setup_parameters_complex(rb_execution_context_t * const ec, const rb_iseq_t * co
             args_setup_kw_parameters_from_kwsplat(ec, iseq, keyword_hash, klocals, remove_hash_value);
         }
         else {
-            VM_ASSERT(args_argc(args) == 0);
+            VM_ASSERT(args_argc(args) == 0 ||
```
ko1, line 863: It is better to use

```C
  if (args_argc(args) != 0) {
    VM_ASSERT(cond1);
    VM_ASSERT(cond2);
    ...
  }
```

because it is hard to read which assumption violates.

test/ruby/test_allocation.rb
```diff
@@ -277,7 +277,7 @@ def self.keyword(a: nil#{block}); end
         check_allocations(0, 1, "keyword(**hash1, **empty_hash#{block})")
```
ko1, line 277: Just curious is it robust for the finalizer which allocates objects? 
I don't see such failure though.

vm_args.c
jeremyevans: I'll make that change.

test/ruby/test_allocation.rb
jeremyevans: It uses `assert_separately`, so hopefully finalizers that allocate shouldn't be an issue.

PR #11020, opened 2024-06-19
Eliminate array allocations for single splat followed by mutable keywords

jeremyevans:
For calls such as:

  m(*ary, a: 2, **h)
  m(*ary, **h, **h, **h)

Where m does not take a positional argument splat, there was previously an array allocation (splatarray true) to dup ary, even though it was not necessary to do so.  This is because the elimination of the array allocation (splatarray false) was performed in the optimizer, and the optimizer didn't handle this case, because the instructions for the keywords can be of arbitrary length.

Move part of the optimization from the optimizer to the compiler, detecting parse trees of the form:

  ARGS_PUSH:
    head: SPLAT
    tail: HASH (without brace)

And using splatarray false instead of splatarray true for them.

Unfortunately, moving part of the optimization to the compiler broke the hash allocation elimination optimization for calls of the form:

  m(*ary, a: 2)

That's because the compiler had already set splatarray false, and the optimizer code was looking for splatarray true.

Split the array allocation elimination and hash allocation elimination in the optimizer so that the hash allocation elimination will still apply if the compiler performs the splatarray false optimization.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -4019,21 +4019,18 @@ iseq_peephole_optimize(rb_iseq_t *iseq, LINK_ELEMENT *list, const int do_tailcal
             niobj = niobj->next;
 
             /*
-            * Eliminate array and hash allocation for f(*a, kw: 1)
+            * Eliminate array allocation for f(*a, kw: 1)
             *
             *  splatarray true
             *  duphash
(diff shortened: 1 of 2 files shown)


ko1:
I'm not sure how it is important because we will drop the compiler code if Prism is default, but if it is important and there is no breakage, please merge it.

PR #11021, opened 2024-06-19
Skip 3 specs when $HOME is not a valid directory

jeremyevans:
If $HOME isn't a valid directory, these specs fail, but I think it is better to skip in this case, as it does not indicate a bug.

This fixes specs when run in the OpenBSD port, which sets $HOME to an invalid directory to ensure that software's build system is not accidentally relying on it.


hsbt (approved):

HASNAT204 (commented):
Looks good 👍

PR #11029, opened 2024-06-20
Introduce `inline_var_lhs` rule

S-H-GAMELINKS:
And, it replace to `user_variable` and `keyword_variable` rhs for `var_lhs` pattern.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -6337,12 +6337,11 @@ var_ref		: user_variable
                     }
                 ;
 
-var_lhs		: user_variable
-                    {
-                    /*% ripper: var_field!($:1) %*/
-                        $$ = assignable(p, $1, 0, &@$);
-                    }
-                | keyword_variable
+%rule %inline inline_var_lhs: user_variable
+                            | keyword_variable
+                            ;
+
+var_lhs		: inline_var_lhs
                     {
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -6299,12 +6303,7 @@ var_ref		: user_variable
                     }
                 ;
 
-var_lhs		: user_variable
-                    {
-                    /*% ripper: var_field!($:1) %*/
-                        $$ = assignable(p, $1, 0, &@$);
-                    }
-                | keyword_variable
+var_lhs		: inline_var_lhs
```
ydah, line 6306: I think we should wait for the release of the next change:
https://github.com/ruby/lrama/pull/420

Since this inline rule only referenced it from here, I thought it would be better if the definitions were closer together.

S-H-GAMELINKS: Thanks your commnet.

Lrama v0.6.10 has been released that include https://github.com/ruby/lrama/pull/420, and Ruby has use Lrama v0.6.10 since in https://github.com/ruby/ruby/pull/11590.

So, I rebased and modify inline rules definitions position near tha used.

S-H-GAMELINKS:
Similar change commited in https://github.com/ruby/ruby/pull/11752
So, I'll close thi pull request

PR #11032, opened 2024-06-20
YJIT: Enhance the `String#<<` method substitution to handle integer codepoint values.

nirvdrum:
This PR extends YJIT's method substitution for `String#<<` to handle integer codepoints as well. If the string is `ASCII-8BIT` and the codepoint is a byte value, YJIT will dispatch to `rb_str_buf_cat_byte` as a fast path for working with binary strings. Otherwise, it'll dispatch to the general `rb_str_concat` just as `vm_opt_ltlt` would.

`rb_str_buf_cat_byte` currently works with both `ASCII-8BIT` and `US-ASCII`, but this YJIT side only optimizes for `ASCII-8BIT`. It could be extended easily enough with an additional comparison. The encoding indices for a handful of encodings, including both `ASCII-8BIT` and `US-ASCII` are fixed and sequential, so we could also do a range check. For the time being, I've omitted the handling of `US-ASCII`. I'd like to get feedback on the simplified PR and extend it with `US-ASCII` handling if needed (which I'm also not convinced we do).

Please advise if the mechanism I'm using to handle polymorphic dispatch is incorrect. We already have `jit_rb_str_concat` as a method substitution for `String#<<`, but it deliberately only handled string arguments. I could have merged the two, but that struck me as being complicated. However, I also don't know if it's fine to call between methods like this. And, it ends up duplicating some of the type checks to ensure we dispatch to the correct type depending on what we see at compile time. I was unsure on how to handle the runtime guards, so please pay extra attention to that.

### DIFF
--- a/string.c
+++ b/string.c
@@ -12296,6 +12296,23 @@ rb_enc_interned_str_cstr(const char *ptr, rb_encoding *enc)
     return rb_enc_interned_str(ptr, strlen(ptr), enc);
 }
 
+#if USE_YJIT
+void
+rb_yjit_str_concat_codepoint(VALUE str, VALUE codepoint)
+{
+    if (RB_LIKELY(FIXNUM_P(codepoint)) && RB_LIKELY(ENCODING_GET_INLINED(str) == rb_ascii8bit_encindex())) {
+        ssize_t code = RB_NUM2SSIZE(codepoint);
+
+        if (RB_LIKELY(code >= 0 && code < 0xff)) {
+            rb_str_buf_cat_byte(str, (char) code);
+            return;
+        }
+    }
+
+    rb_str_concat(str, codepoint);
+}
+#endif
+
 void
 Init_String(void)
 {
--- a/string.c
+++ b/string.c
@@ -3375,8 +3375,7 @@ rb_str_buf_cat_byte(VALUE str, unsigned char byte)
     }
     else {
         // If there's not enough string_capacity, make a call into the general string concatenation function.
-        char buf[1] = {byte};
-        str_buf_cat(str, buf, 1);
+        str_buf_cat(str, (char *)&byte, 1);
     }
 
     // If the code range is already known, we can derive the resulting code range cheaply by looking at the byte we
--- a/string.c
+++ b/string.c
@@ -12300,7 +12300,7 @@ rb_enc_interned_str_cstr(const char *ptr, rb_encoding *enc)
 void
 rb_yjit_str_concat_codepoint(VALUE str, VALUE codepoint)
 {
-    if (RB_LIKELY(FIXNUM_P(codepoint)) && RB_LIKELY(ENCODING_GET_INLINED(str) == rb_ascii8bit_encindex())) {
+    if (RB_LIKELY(ENCODING_GET_INLINED(str) == rb_ascii8bit_encindex())) {
         ssize_t code = RB_NUM2SSIZE(codepoint);
 
         if (RB_LIKELY(code >= 0 && code < 0xff)) {
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1999,6 +1999,46 @@ fn guard_object_is_hash(
     }
 }
 
+fn guard_object_is_fixnum(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    object: Opnd,
+    object_opnd: YARVOpnd
+) {
+    let object_type = asm.ctx.get_opnd_type(object_opnd);
+    if object_type.is_heap() {
+        asm_comment!(asm, "arg is heap object");
+        asm.jmp(Target::side_exit(Counter::guard_send_not_fixnum));
+        return;
+    }
+
+    if object_type != Type::Fixnum && object_type.is_specific() {
+        asm_comment!(asm, "arg is not fixnum");
+        asm.jmp(Target::side_exit(Counter::guard_send_not_fixnum));
+        return;
+    }
+
+    assert!(!object_type.is_heap());
+    assert!(object_type == Type::Fixnum || object_type.is_unknown());
+
+    // If not fixnums at run-time, fall back
+    if object_type != Type::Fixnum {
+        asm_comment!(asm, "guard object fixnum");
+        asm.test(object, Opnd::UImm(RUBY_FIXNUM_FLAG as u64));
+
+        jit_chain_guard(
+            JCC_JZ,
+            jit,
+            asm,
+            SEND_MAX_DEPTH,
+            Counter::guard_send_not_fixnum,
+        );
+    }
+
+    // Set the stack type in the context.
+    asm.ctx.upgrade_opnd_type(object.into(), Type::Fixnum);
+}
+
 fn guard_object_is_string(
     asm: &mut Assembler,
     object: Opnd,
@@ -5790,6 +5830,8 @@ fn jit_rb_str_concat_codepoint(
 
     let codepoint = asm.stack_opnd(0);
     let recv = asm.stack_opnd(1);
+
+    guard_object_is_fixnum(jit, asm, codepoint, StackOpnd(0));
 
     asm.ccall(rb_yjit_str_concat_codepoint as *const u8, vec![recv, codepoint]);
 
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5798,7 +5798,14 @@ fn jit_rb_str_concat(
     // rb_str_buf_append may raise Encoding::CompatibilityError, but we accept compromised
     // backtraces on this method since the interpreter does the same thing on opt_ltlt.
     jit_prepare_non_leaf_call(jit, asm);
-    asm.spill_regs(); // For ccall. Unconditionally spill them for RegMappings consistency.
+
+    // Explicitly spill temps before making any C calls. `ccall` will spill temps, but it does a
+    // check to only spill if it thinks it's necessary. That logic can't see through the runtime
+    // branching occurring in the code generated for this function. Consequently, the branch for
+    // the first `ccall` will spill registers but the second one will not. At run time, we may
+    // jump over that spill code when executing the second branch, leading situations that are
+    // quite hard to debug. If we spill up front we avoid diverging behavior.
+    asm.spill_regs();
 
     let concat_arg = asm.stack_pop(1);
     let recv = asm.stack_pop(1);
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5770,10 +5770,9 @@ fn jit_rb_str_empty_p(
     return true;
 }
 
-// Codegen for rb_str_concat() -- *not* String#concat
-// Frequently strings are concatenated using "out_str << next_str".
-// This is common in Erb and similar templating languages.
-fn jit_rb_str_concat(
+// Codegen for rb_str_concat() with an integer argument -- *not* String#concat
+// Using strings as a byte buffer often includes appending byte values to the end of the string.
+fn jit_rb_str_concat_codepoint(
     jit: &mut JITState,
     asm: &mut Assembler,
     _ci: *const rb_callinfo,
@@ -5782,11 +5781,46 @@ fn jit_rb_str_concat(
     _argc: i32,
     _known_recv_class: Option<VALUE>,
 ) -> bool {
+    asm_comment!(asm, "String#<< with codepoint argument");
+
+    // Either of the string concatenation functions we call will reallocate the string to grow its
+    // capacity if necessary. In extremely rare cases (i.e., string exceeds `LONG_MAX` bytes),
+    // either of the called functions will raise an exception.
+    jit_prepare_non_leaf_call(jit, asm);
+
+    let codepoint = asm.stack_opnd(0);
+    let recv = asm.stack_opnd(1);
+
+    asm.ccall(rb_yjit_str_concat_codepoint as *const u8, vec![recv, codepoint]);
+
+    // The receiver is the return value, so we only need to pop the codepoint argument off the stack.
+    // We can reuse the receiver slot in the stack as the return value.
+    asm.stack_pop(1);
+
+    true
+}
+
+// Codegen for rb_str_concat() -- *not* String#concat
+// Frequently strings are concatenated using "out_str << next_str".
+// This is common in Erb and similar templating languages.
+fn jit_rb_str_concat(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    ci: *const rb_callinfo,
+    cme: *const rb_callable_method_entry_t,
+    block: Option<BlockHandler>,
+    argc: i32,
+    known_recv_class: Option<VALUE>,
+) -> bool {
     // The << operator can accept integer codepoints for characters
     // as the argument. We only specially optimise string arguments.
     // If the peeked-at compile time argument is something other than
     // a string, assume it won't be a string later either.
     let comptime_arg = jit.peek_at_stack(&asm.ctx, 0);
+    if unsafe { RB_TYPE_P(comptime_arg, RUBY_T_FIXNUM) } {
+        return jit_rb_str_concat_codepoint(jit, asm, ci, cme, block, argc, known_recv_class);
+    }
+
     if ! unsafe { RB_TYPE_P(comptime_arg, RUBY_T_STRING) } {
         return false;
     }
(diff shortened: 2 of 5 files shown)


yjit/src/codegen.rs
```diff
@@ -5911,6 +6000,10 @@ fn jit_rb_str_concat(
     // If the peeked-at compile time argument is something other than
     // a string, assume it won't be a string later either.
     let comptime_arg = jit.peek_at_stack(&asm.ctx, 0);
+    if unsafe { RB_TYPE_P(comptime_arg, RUBY_T_FIXNUM) } {
+        return jit_rb_str_concat_codepoint(jit, asm, _ocb, _ci, _cme, _block, _argc, _known_recv_class);
+    }
```
maximecb, line 6005: Minor nitpick that the underscore by convention e.g. `_ocb` means that an argument won't be used, so you shouldn't be passing or using arguments named `_something`. If you do use them, you should remove the underscore from the name.

yjit/src/codegen.rs
```diff
@@ -5893,6 +5893,95 @@ fn jit_rb_str_empty_p(
...
+            comptime_arg.class_of(),
+            arg,
+            arg.into(),
+            comptime_arg,
+            SEND_MAX_DEPTH,
+            Counter::guard_send_not_fixnums,
+        );
+    } else if ! unsafe { RB_TYPE_P(comptime_arg, RUBY_T_STRING) } {
+        return jit_rb_str_concat(jit, asm, ocb, _ci, _cme, _block, _argc, _known_recv_class);
+    } else {
```
maximecb, line 5927: Feels a bit weird to see this check here since you already checked in `jit_rb_str_concat`. You could just go straight to return false.

yjit/src/codegen.rs
```diff
@@ -5893,6 +5893,95 @@ fn jit_rb_str_empty_p(
...
+    // either of the called functions will raise an exception.
+    jit_prepare_non_leaf_call(jit, asm);
+
+    let codepoint = asm.stack_pop(1);
+    let recv = asm.stack_pop(1);
+
+    // In order to use the fast path (rb_str_buf_cat_byte), the string encoding must be ASCII-8BIT
+    // and the codepoint must be in the byte range (0x00 - 0xff).
+    // If either of those conditions are not met we must use the general string concat (str_buf_cat)
+    // function with the original codepoint argument.
```
maximecb, line 5942: As a general comment. I like to avoid generating branches inside the inline block because it can be taxing for branch prediction. This is another discussion, but I think it's useful to think of branch prediction as a finite resource. It works well when you don't have a lot of branches, but the code we generate is going to have (tens of) thousands of them, so the CPU can't always remember which direction a given branch went the last time.

Branches inside the inline block also tends to result in bigger inline code size (worse for instruction cache). I like to make it so that failing guards side-exit to the interpreter (outlined code) as much as possible. One thing you could potentially do is peek at the encoding at compilation time... This would involve speculating that the string being appended to will always have the same encoding. We can measure how often this is the case in practice.

We can continue this discussion in #yjit-internal though. Still open to merging this PR as is.

yjit/src/codegen.rs
```diff
@@ -5893,6 +5893,95 @@ fn jit_rb_str_empty_p(
...
+    let codepoint_untag = asm.rshift(codepoint, Opnd::UImm(1));
+    asm.cmp(codepoint_untag, Opnd::Imm(0xff));
+    asm.jg(generic_str_concat_codepoint);
+
+    // If we've made it this far, we must be appending a single byte value to a binary string.
+    asm_comment!(asm, "Optimized (string, codepoint) concatenation");
+    let ret_opnd = asm.ccall(rb_str_buf_cat_byte as *const u8, vec![recv, codepoint_untag]);
+    let stack_ret = asm.stack_push(Type::TString);
+    asm.mov(stack_ret, ret_opnd);
+    asm.stack_pop(1);
```
maximecb, line 5969: Noticing the `stack_pop` here. One thing to know is that `asm.stack_pop` is not going to care whether the conditional branches are taken or not.

maximecb:
Hi Kevin.

Commenting to give you as much feedback as possible.

> This PR extends YJIT's method substitution for `String#<<` to handle integer codepoints as well. If the string is `ASCII-8BIT` and the codepoint is a byte value, YJIT will dispatch to `rb_str_buf_cat_byte` as a fast path for working with binary strings.

I think that the idea of this PR is a good one. It seems like a sensible specialization to implement.

I wrote some comments on the code itself.

Two things I would like to see in a PR like this:
- Benchmark results to know how it affects performance on protoboeuf and our headline benchmarks (if any impact)
- Maybe a dump of machine code generated when concatenating a few bytes in a loop just to see if we're doing anything obviously inefficient

> Please advise if the mechanism I'm using to handle polymorphic dispatch is incorrect. We already have `jit_rb_str_concat` as a method substitution for `String#<<`, but it deliberately only handled string arguments. I could have merged the two, but that struck me as being complicated. However, I also don't know if it's fine to call between methods like this. And, it ends up duplicating some of the type checks to ensure we dispatch to the correct type depending on what we see at compile time. I was unsure on how to handle the runtime guards, so please pay extra attention to that.

Seems sensible. Duplicating checks that occur at run-time would not be good but the checks you duplicated occur at compilation time. It makes sense to write the logic for this in a separate function to avoid ending up with a megafunction that is hard to follow.

yjit/src/codegen.rs
nirvdrum: Thanks. I wasn't sure if passing arguments forward constituted being used. It's an easy enough fix. Thanks.

yjit/src/codegen.rs
nirvdrum: Thanks. I was hoping someone would weigh in on this one. I wasn't sure to what extent we can make assumptions given we know all the current callers vs needing to be safe in the presence of a new caller. I'll remove this.

yjit/src/codegen.rs
maximecb: I guess just be aware that the stack_pop is always executed, not conditionally executed.

tenderlove:
Before this commit Protoboeuf+YJIT is about 5.28x slower than Google's protobuf implementation, but after this commit it is 4.33x slower.

Before:

```
ruby 3.4.0dev (2024-07-09T17:22:29Z master 6f6aff56b1) +YJIT [arm64-darwin23]
     encode upstream    14.000 i/100ms
   encode protoboeuf     2.000 i/100ms
     encode upstream    140.854 (± 0.7%) i/s -    714.000 in   5.069352s
   encode protoboeuf     26.672 (± 3.7%) i/s -    134.000 in   5.028086s

Comparison:
     encode upstream:      140.9 i/s
   encode protoboeuf:       26.7 i/s - 5.28x  slower
```

After:

```
ruby 3.4.0dev (2024-07-09T19:35:29Z yjit-optimize-stri.. 00cc8e4429) +YJIT [arm64-darwin23]
     encode upstream    13.000 i/100ms
   encode protoboeuf     3.000 i/100ms
     encode upstream    137.078 (± 0.7%) i/s -    689.000 in   5.026818s
   encode protoboeuf     31.678 (± 3.2%) i/s -    159.000 in   5.023590s

Comparison:
     encode upstream:      137.1 i/s
   encode protoboeuf:       31.7 i/s - 4.33x  slower
```

XrXr (commented):
An alternative to what you have is to make a new Rust/C function that essentially does the logic that you currently inline into every site (`if codepoint.is_fixnum() { rb_str_buf_cat_byte } else { rb_str_concat }`). That way you only generate one ccall at each site.

It should be easier to read than checking in assembler, perf should be about the same, and it's a code size win.

maximecb:
I like Alan's idea of embedding some of the checks in a C function to save on code size and maybe make the code a bit more readable.

Also thanks Kevin for persisting. You picked a hard problem for your first big PR! 😉

nirvdrum:
> An alternative to what you have is to make a new Rust/C function that essentially does the logic that you currently inline into every site (`if codepoint.is_fixnum() { rb_str_buf_cat_byte } else { rb_str_concat }`). That way you only generate one ccall at each site.

Okay. I think you were simplifying, but for completeness the check is really (in pseudo-code) `if codepoint.is_fixnum() && receiver.is_binary_string() && codepoint.is_byte()`. `rb_str_buf_cat_byte` does not have that logic. The caller is expected to call it only when those conditions hold (actually, `rb_str_buf_cat_byte` also works on US-ASCII strings, but I omitted that in YJIT) and there are assertions at the start of `rb_str_buf_cat_byte` that checks those conditions are held.

> If you write it in C, it could go into string.c and you can additionally avoid removing `static` from rb_str_buf_cat_byte().

I'll have to introduce a new function. I don't think updating `rb_str_buf_cat_byte` to do those checks is the right way to go. It would mean duplicating checks in the interpreter to derive information we already have.

XrXr:
Yes, you should make clear that the new function you add is a YJIT helper and by not putting it in headers, express that it has many preconditions that are hard to meet and is not for general usage.

yjit/src/codegen.rs
```diff
@@ -5754,23 +5754,118 @@ fn jit_rb_str_empty_p(
...
-fn jit_rb_str_concat(
+// Codegen for rb_str_concat() with an integer argument -- *not* String#concat
+// Using strings as a byte buffer often includes appending byte values to the end of the string.
+fn jit_rb_str_concat_codepoint(
     jit: &mut JITState,
     asm: &mut Assembler,
     _ci: *const rb_callinfo,
     _cme: *const rb_callable_method_entry_t,
-    _block: Option<BlockHandler>,
+    _block: Option<crate::codegen::BlockHandler>,
```
maximecb, line 5764: nitpick but why do you need to specify `crate::codegen::` here? Isn't that the module we're currently in?

yjit/src/codegen.rs
```diff
@@ -5754,23 +5754,118 @@ fn jit_rb_str_empty_p(
...
+    asm_comment!(asm, "Optimized (string, codepoint) concatenation");
+    asm.ccall(rb_str_buf_cat_byte as *const u8, vec![recv, codepoint_untag]);
+    asm.jmp(ret_label);
+
+    // Either the string isn't binary or the codepoint is too large.
+    asm_comment!(asm, "Fallback to generic (string, codepoint) concatenation");
+    asm.write_label(generic_concat_label);
+    asm.ccall(rb_str_concat as *const u8, vec![recv, codepoint]);
+
+    asm.write_label(ret_label);
```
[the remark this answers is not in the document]
maximecb: The benefit of not duplicating the branches, and instead having the check inside of a C function, is that it's friendlier to the CPU's branch predictor (less branches to keep track of).

yjit/src/codegen.rs
nirvdrum: I think the IDE inserted that. I'll fix it. Thanks for catching it.

yjit/src/codegen.rs
```diff
@@ -5762,23 +5762,75 @@ fn jit_rb_str_empty_p(
...
+            comptime_arg.class_of(),
+            arg,
+            arg.into(),
+            comptime_arg,
+            SEND_MAX_DEPTH,
+            Counter::guard_send_not_fixnums,
+        );
+    } else {
+        return false;
+    }
```
maximecb, line 5794: Here you already checked that `comptime_arg` is a fixnum in the caller.

Maybe we can also fold the guard that the value is fixnum in the caller? Because presumably the fallback `rb_str_concat(str, codepoint)` can handle any kind of input type? @XrXr would this be valid?

nirvdrum: The original version with the assembler required the type to be a `Fixnum`, but that's no longer the case. I can move this guard easily enough. I just want to confirm that if we see another type at the call site we don't want to try to cut over to `jit_rb_str_concat`?

nirvdrum: If we're going to allow non-codepoint arguments into `rb_yjit_str_concat_codepoint`, should I rename that function? Or is naming it after its primary intention okay?

XrXr: > I just want to confirm that if we see another type at the call site we don't want to try to cut over to jit_rb_str_concat?

Seems like if it's already behind a `FIXNUM_P` check in the caller (checking in a different way), so you could turn this into an assert.

Renaming the C function sounds good. Maybe `rb_yjit_str_concat_likely_byte`? Also a short comment on the C side would be nice.

nirvdrum: I was looking at this with @paracycle, too. Since we're talking about turning `rb_yjit_str_concat_codepoint` into a generic fallback, should we collapse the `jit_rb_str_concat` and `jit_rb_str_concat_codepoint` codegen methods into a single handler that calls to this C function?

I don't know the history of `jit_rb_str_concat` to know whether it's doing more in assembly for speed or because of necessity. That's what I used as the template for my original implementation of `jit_rb_str_concat_codepoint` and we decided to move things out to C to keep the inline code block smaller and reduce branching. If we applied those same principals here, I suppose we could simplify the codegen a fair bit. But, if that's out of scope, that's fine with me.

k0kubun: Collapsing them into a single C call sounds good to me. Most of today's `jit_rb_str_concat` seems like something that could be done in the C function.

nirvdrum: I've consolidate the two methods and pushed that up. It's ready for review if anyone is interested.

maximecb:
I see that you modified the regular string concat to fold it into the C call as well. The C code looks well structured, but the downside of that is that we don't necessarily benefit from type information that YJIT might have.

I benchmarked this PR locally:
```
master: ruby 3.4.0dev (2024-07-29T15:07:53Z master 86a762ce56) [arm64-darwin23]
cat: ruby 3.4.0dev (2024-07-26T19:34:15Z yjit-optimize-stri.. 3c3aba4119) [arm64-darwin23]

----------  -----------  ----------  --------  ----------  -----------  ----------
bench       master (ms)  stddev (%)  cat (ms)  stddev (%)  cat 1st itr  master/cat
protoboeuf  78.7         0.9         81.8      1.1         0.97         0.96      
str_concat  39.7         6.4         42.7      4.0         0.92         0.93      
----------  -----------  ----------  --------  ----------  -----------  ----------
```

At the moment it seems like it's slightly slower, so not a win at the moment. Initially Aaron had reported a big speedup on protoboeuf with your original patch. 

I'm sorry about all the back and forth. I don't fully understand why the original patch had such a big speedup in the first place compared to this. It might be worth digging more into where the overhead is coming from and doing more experiments. Maybe we could organize a pairing session with @XrXr ?

XrXr:
>  Initially Aaron had reported a big speedup on protoboeuf with your original patch. ...  I don't fully understand why the original patch had such a big speedup in the first place compared to this.


It looks like the speed-up was on the encoding benchmark and you ran the decoding benchmark.

maximecb:
> > Initially Aaron had reported a big speedup on protoboeuf with your original patch. ...  I don't fully understand why the original patch had such a big speedup in the first place compared to this.
> 
> It looks like the speed-up was on the encoding benchmark and you ran the decoding benchmark. [#11032 (comment)](https://github.com/ruby/ruby/pull/11032#issuecomment-2218808415)

Mildly embarrassed 😅 Nevertheless, not ideal that it would slow down decode. 

It does seem slower on encode, too but we should probably run these benchmarks on an AWS machine as well:

```
master: ruby 3.4.0dev (2024-07-29T15:07:53Z master 86a762ce56) [arm64-darwin23]
cat: ruby 3.4.0dev (2024-07-26T19:34:15Z yjit-optimize-stri.. 3c3aba4119) [arm64-darwin23]

-----------------  -----------  ----------  --------  ----------  -----------  ----------
bench              master (ms)  stddev (%)  cat (ms)  stddev (%)  cat 1st itr  master/cat
protoboeuf         76.2         2.9         78.6      0.8         1.00         0.97      
protoboeuf-encode  89.8         1.2         92.6      1.4         0.96         0.97      
str_concat         43.8         4.8         44.0      4.0         0.96         0.99      
-----------------  -----------  ----------  --------  ----------  -----------  ----------
```

Also, @nirvdrum, have you rebased this PR on master recently? Just to make sure it's a fair comparison.

nirvdrum:
I put the code that moves the older `jit_rb_str_concat` into a separate commit. It should be easy to compare performance with and without that change.

I've been rebasing the branch each time I edit it (although, it's possible I've forgotten to pull first). It's currently based on 86a762ce56cd44db69ccff2e02587ed748a3ad04, which matched the version of _master_ you were using. But, I pushed that about 7 minutes after your comment (sorry, I didn't see it before our I would have mentioned it). It looks like the version you used was based off 83b0cedffe5a9bd6c5288fa0d9614d6ca39e63e8.

string.c
```diff
@@ -12297,6 +12296,23 @@ rb_enc_interned_str_cstr(const char *ptr, rb_encoding *enc)
     return rb_enc_interned_str(ptr, strlen(ptr), enc);
 }
 
+#if USE_YJIT
+void
+rb_yjit_str_concat_codepoint(VALUE str, VALUE codepoint)
+{
+    if (ENCODING_GET_INLINED(str) == rb_ascii8bit_encindex()) {
```
maximecb, line 12303: You should probably mark this as likely as well

PR #11041, opened 2024-06-21
[YJIT] Fix block and splat handling when forwarding

tenderlove:
This commit fixes splat and block handling when calling in to a forwarding iseq.  In the case of a splat we need to avoid expanding the array to the stack.  We need to also ensure the CI write is flushed to the SP, otherwise it's possible for a block handler to clobber the CI

[ruby-core:118360]

This also fixes the problem in lobsters benchmark.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -7116,9 +7116,18 @@ fn gen_send_iseq(
     let iseq_has_block_param = unsafe { get_iseq_flags_has_block(iseq) };
     let arg_setup_block = captured_opnd.is_some(); // arg_setup_type: arg_setup_block (invokeblock)
     let kw_splat = flags & VM_CALL_KW_SPLAT != 0;
-    let splat_call = flags & VM_CALL_ARGS_SPLAT != 0;
 
-    let forwarding_call = unsafe { rb_get_iseq_flags_forwardable(iseq) };
+    // Is this iseq tagged as "forwardable"? Iseqs that take `...` as a
+    // parameter are tagged as forwardable (e.g. `def foo(...); end`)
+    let forwarding = unsafe { rb_get_iseq_flags_forwardable(iseq) };
+
+    // If a "forwardable" iseq has been called with a splat, then we _do not_
+    // want to expand the splat to the stack. So we'll only consider this
+    // a splat call if the callee iseq is not forwardable.  For example,
+    // we do not want to handle the following code:
+    //
+    // `def foo(...); end; foo(*blah)`
+    let splat_call = (flags & VM_CALL_ARGS_SPLAT != 0) && !forwarding;
 
     // For computing offsets to callee locals
     let num_params = unsafe { get_iseq_body_param_size(iseq) as i32 };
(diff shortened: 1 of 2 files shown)


k0kubun (approved):

rwstauner (approved):
Confirmed this fixes things locally for me.

yjit/src/codegen.rs
```diff
@@ -7711,6 +7716,7 @@ fn gen_send_iseq(
 
     if forwarding_call {
         assert_eq!(1, num_params);
+        asm.ctx.dealloc_temp_reg(asm.ctx.get_stack_size());
```
XrXr, line 7719: Is this load-bearing? Existing uses are close to where the store actually is, and this seems to have no relation to what follows.
If it is deallocating for `(-1)`, can you extract `asm.stack_opnd(-1)` as a variable and use `asm.ctx.dealloc_temp_reg(ci_opnd.stack_idx())` instead?

TBH I'm not too sure how reg alloc for negative slots work and the sign to unsigned cast is confusing.

k0kubun: This is what I suggested during pairing since `asm.spill_temps()` ignores negative indexes. Without this, it would be written only to a register and the interpreter can't see it on the fallback.

casperisfine:
This also seems to fix a crash in Rails test suite:

```
Assertion Failed: /usr/src/ruby/include/ruby/internal/core/rarray.h:257:rb_array_len:RB_TYPE_P(a, RUBY_T_ARRAY)
```

Which I reduced to:

```ruby
class Model
  class << self
    def scope(name, body, &block)
      singleton_class.define_method(name) do |*args|
        new._exec_scope(*args, &body)
      end
      singleton_class.send(:ruby2_keywords, name)
    end
  end

  def _exec_scope(...)
    _scoping(nil, Object.new) { instance_exec(...) || self }
  end

  def _scoping(a, b)
    yield
  end
end

Model.scope(:some_query, -> { p 1 })

Model.some_query
```

tenderworks:
> This also seems to fix a crash in Rails test suite:

Ya, I golfed that down to one of the tests I introduced in this PR (specifically the `*args, &block` calls).  I need some help getting the book keeping on the context correct before I can merge this.

yjit/src/codegen.rs
```diff
@@ -7116,10 +7116,11 @@ fn gen_send_iseq(
     let iseq_has_block_param = unsafe { get_iseq_flags_has_block(iseq) };
     let arg_setup_block = captured_opnd.is_some(); // arg_setup_type: arg_setup_block (invokeblock)
     let kw_splat = flags & VM_CALL_KW_SPLAT != 0;
-    let splat_call = flags & VM_CALL_ARGS_SPLAT != 0;
 
     let forwarding_call = unsafe { rb_get_iseq_flags_forwardable(iseq) };
 
+    let splat_call = (flags & VM_CALL_ARGS_SPLAT != 0) && !forwarding_call;
```
maximecb, line 7122: This is because this call site (this send instruction) is a forwarding call?

Pls add more comments for us laypeople

yjit/src/codegen.rs
```diff
@@ -7116,10 +7116,19 @@ fn gen_send_iseq(
     let iseq_has_block_param = unsafe { get_iseq_flags_has_block(iseq) };
     let arg_setup_block = captured_opnd.is_some(); // arg_setup_type: arg_setup_block (invokeblock)
     let kw_splat = flags & VM_CALL_KW_SPLAT != 0;
-    let splat_call = flags & VM_CALL_ARGS_SPLAT != 0;
 
+    // Is this iseq tagged as "forwardable"? Iseqs that take `...` as a
+    // parameter are tagged as forwardable (e.g. `def foo(...); end`)
     let forwarding_call = unsafe { rb_get_iseq_flags_forwardable(iseq) };
```
maximecb, line 7122: Can we rename this variable to forwardable?

PR #11043, opened 2024-06-22
 Integrate Launchable into mingw

ono-max:
Recently, Launchable has been used for analyzing flaky tests. Kokubun-san asked me to integrate Launchable into mingw for analyzing flay tests in mingw, so I've created this PR to address it.

Examples for using Launchable:

https://github.com/ruby/ruby/pull/11025
https://github.com/ruby/ruby/pull/10749

### DIFF
--- a/.github/actions/launchable/setup/action.yml
+++ b/.github/actions/launchable/setup/action.yml
@@ -88,8 +88,7 @@ runs:
       working-directory: ${{ inputs.srcdir }}
       run: |
         set -x
-        PATH=$PATH:$(python -msite --user-base)/bin
-        echo "PATH=$PATH" >> $GITHUB_ENV
+        echo "$(python -msite --user-base)/bin" >> $GITHUB_PATH
         pip install --user launchable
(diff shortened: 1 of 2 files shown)


.github/actions/launchable/setup/action.yml
```diff
@@ -88,8 +88,7 @@ runs:
       working-directory: ${{ inputs.srcdir }}
       run: |
         set -x
-        PATH=$PATH:$(python -msite --user-base)/bin
-        echo "PATH=$PATH" >> $GITHUB_ENV
+        echo "$(python -msite --user-base)/bin" >> $GITHUB_PATH
```
ono-max, line 91: To prevent this step from replacing all backslashes with forward slashes, `$GITHUB_PATH` is used. Based on my investigation, it seems to be that all paths are expressed as Unix path style when referring them in Bash.

https://github.com/ruby/ruby/actions/runs/9626581444/job/26576319914?pr=11043#step:10:271

PR #11048, opened 2024-06-24
Extend the timeout of Fiddle::TestFunction#test_nogvl_poll

mame:


kou:
Should we do this in upstream https://github.com/ruby/fiddle/blob/master/test/fiddle/test_function.rb ?

PR #11053, opened 2024-06-24
Do not warn about extra `format` argument when passed as keywords

casperisfine:
[[Bug #20593]](https://bugs.ruby-lang.org/issues/20593)

It's fairly common to use `format` to interpolate a number of values into a user provided strings.

The arguments not matching are a problem when they are positional, but when they are named, it's absolutely fine and we shouldn't emit a warning.

### DIFF
--- a/sprintf.c
+++ b/sprintf.c
@@ -937,7 +937,7 @@ rb_str_format(int argc, const VALUE *argv, VALUE fmt)
     rb_str_tmp_frozen_release(orig, fmt);
     /* XXX - We cannot validate the number of arguments if (digit)$ style used.
      */
-    if (posarg >= 0 && nextarg < argc) {
+    if (posarg >= 0 && nextarg < argc && !(argc == 2 && RB_TYPE_P(argv[1], T_HASH))) {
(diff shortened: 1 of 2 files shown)


sprintf.c
```diff
@@ -937,7 +937,7 @@ rb_str_format(int argc, const VALUE *argv, VALUE fmt)
     rb_str_tmp_frozen_release(orig, fmt);
     /* XXX - We cannot validate the number of arguments if (digit)$ style used.
      */
-    if (posarg >= 0 && nextarg < argc) {
+    if (posarg >= 0 && nextarg < argc && !(argc == 2 && RB_TYPE_P(argv[1], T_HASH))) {
```
nobu, line 940: Doesn't this make `sprintf "test", {}` no warning?

casperisfine: You think it should be a warning? I don't think it should be.

nobu: It's OK if you don't mean keyword arguments by "passed as keywords".

casperisfine: Right, I meant passing only a single hash as argument.

PR #11064, opened 2024-06-26
YJIT: Move `ocb` parameters into JITState

XrXr:
Many functions take an outlined code block but do nothing more than
passing it along; only a couple of functions actually make use of it.
So, in most cases the `ocb` parameter is just boilerplate.

Most functions that take `ocb` already also take a `JITState`, so this
commit moves `ocb` into `JITState` to remove the visual noise of the
`ocb` parameter.


yjit/src/core.rs
```diff
@@ -3817,7 +3816,7 @@ pub fn defer_compilation(
     };
 
     // Likely a stub since the context is marked as deferred().
-    let target0_address = branch.set_target(0, blockid, &next_ctx, ocb);
+    let target0_address = branch.set_target(0, blockid, &next_ctx, jit.get_outlined_cb());
```
maximecb, line 3819: Maybe we could just make it `jit.get_ocb()` since it seems to get called somewhat often

maximecb (approved):
Good simplification 👍 :)

nirvdrum:
Thanks. I like the simplification. It cleans up the code and should make it easier to track down real uses of the the OCB.

PR #11065, opened 2024-06-26
[Feature #19979] Method definition with `&nil`

paracycle:
This PR pushes the original work done by @nobu in #10020 (thank you 🙏) and reworks it slightly along two directions:

1. This PR introduces a new iseq flag named `accepts_no_block` (similar to `accepts_no_kwarg`) and decouples the handling of `&nil` from the `use_block` flag which is used for detection for methods that don't make use of the blocks that are passed to them. `use_block` should NOT mean the same thing as `&nil`, since the latter is an explicit declaration that the method definitely does NOT accept a block, so passing a block should be an error. On the other hand, passing a block to an iseq marked as `use_block == FALSE` will be a warning.
2. This PR also adds support for exposing the `accepts_no_block` flag as `:noblock` in the `Method#parameters`, `UnboundMethod#parameters` and `Proc#parameters` result arrays. This is inline with `**nil` declaration surfacing in `parameters` as `:nokey`, and allows method introspection to detect if a method declares that it definitely does NOT accept a block.

### DIFF
--- a/spec/ruby/core/method/parameters_spec.rb
+++ b/spec/ruby/core/method/parameters_spec.rb
@@ -22,6 +22,8 @@ describe "Method#parameters" do
       local_is_not_parameter = {}
     end
 
+    def one_noblock(&nil); end
+
     def forward_parameters(...) end
 
     def underscore_parameters(_, _, _ = 1, *_, _:, _: 2, **_, &_); end
@@ -185,6 +187,11 @@ describe "Method#parameters" do
   it "returns [[:nokey]] for a method with a single **nil parameter" do
     m = MethodSpecs::Methods.instance_method(:one_nokey)
     m.parameters.should == [[:nokey]]
+  end
+
+  it "returns [[:noblock]] for a method with a single &nil parameter" do
+    m = MethodSpecs::Methods.instance_method(:one_noblock)
+    m.parameters.should == [[:noblock]]
   end
 
   it "works with ->(){} as the value of an optional argument" do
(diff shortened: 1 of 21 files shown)


spec/ruby/core/method/parameters_spec.rb
```diff
@@ -22,6 +22,8 @@ def one_splat_one_block(*args, &block)
       local_is_not_parameter = {}
     end
 
+    def one_noblock(&nil); end
```
jeremyevans, line 25: I think this needs to be guarded in a `ruby_version_is '3.4'` block.  Same for the two other spec changes (no test changes needed, as tests always assume the master branch is used).

paracycle: @eregon I had to use `eval` for this, even after wrapping the new methods inside a `ruby_version_is` block, since the new syntax is (understandably) a syntax error on old Rubies.

Is this the right way to do this, or am I missing something? I also couldn't find any guidance about syntax changes in the README or CONTRIBUTING docs inside the `spec` folder.

eregon: Yes, that's fine and already used in some other places, notably language/ specs.

trinistr:
Feature was accepted by Matz in [a dev meeting](https://github.com/ruby/dev-meeting-log/blob/master/2025/DevMeeting-2025-12-11.md#feature-19979-reconsider-adding-nil-to-method-declarations-to-signal-it-wont-accept-a-block-pabloh), so this PR probably should be updated.
(Have been waiting for this feature for a while now :sweat:)

paracycle:
@kddnewton I've rebased this and made the changes you asked for. It should be ready for your review. 🙏

kddnewton:
@paracycle  looks like there's a zjit thing that needs fixing, then it's good

paracycle:
> @paracycle looks like there's a zjit thing that needs fixing, then it's good

Thanks! I applied the patch from the failing job and force-pushed the changes.

PR #11066, opened 2024-06-27
YJIT: Inline simple ISEQs with unused kwarg params

Gadiguibou:
This is mostly relevant for a monkey-patched version of Sorbet in use at Shopify.

For example, the following function is not currently inlined because its signature contains a kwarg (even though the kwarg is unused).

```ruby
def let(value, type, checked: true)
    value
end
```


Gadiguibou:
One test appears to failing, but that test fails on master as well it seems.

maximecb (approved):

yjit/src/codegen.rs
```diff
@@ -7061,15 +7062,38 @@ fn iseq_get_return_value(iseq: IseqPtr, captured_opnd: Option<Opnd>, ci_flags: u
...
-                None
+            if captured_opnd.is_some()
+                // Reject if block ISEQ is present
+                || block.is_some()
+                // Equivalent to `VM_CALL_ARGS_SIMPLE - VM_CALL_KWARG - has_block_iseq`
+                || ci_flags & (
+                      VM_CALL_ARGS_SPLAT
+                    | VM_CALL_KW_SPLAT
+                    | VM_CALL_ARGS_BLOCKARG
+                    | VM_CALL_FORWARDING
```
XrXr, line 7076: Since overall we're targeting unused keyword parameters, allowing the call site to specify keywords seems odd. I don't see checks for passing an unknown keyword, and I wonder if they are rejected as they should be.

yjit/src/codegen.rs
```diff
@@ -7061,15 +7062,38 @@ fn iseq_get_return_value(iseq: IseqPtr, captured_opnd: Option<Opnd>, ci_flags: u
...
+            }
+
+            let ep_offset = unsafe { *rb_iseq_pc_at_idx(iseq, 1) }.as_u32();
+            let local_idx = ep_offset_to_local_idx(iseq, ep_offset);
+
+            if unsafe { rb_simple_iseq_p(iseq) } {
+                return Some(IseqReturn::LocalVariable(local_idx));
+            } else if unsafe {  rb_iseq_only_kwparam_p(iseq) } {
+                // Inline if only positional parameters are used
+                if let Ok(i) = i32::try_from(local_idx) {
```
XrXr, line 7089: There is an implicit check going on here that should be made explicit in comment. For callees that have expressions in the optional keywords like: `def foo(kw: itself) = nil`, we're relying on the instruction count check to filter them out. Note that depending on the nature of the expression the iseq you get for the callee is different. Compare to `def foo(kw: 3) = nil`, for example.

XrXr (commented):
Thanks! Just minor issues around commenting and testing for myself when I do a more in-depth second pass. It's probably alright to merge since it looks like this is passing existing tests.

yjit/src/codegen.rs
Gadiguibou: Good catch! I tested this locally but I forgot to add a test case for it. Added one in 41276418b4073e4d01cbe2da051e2c2758495668.

Gadiguibou: The reason I chose to allow keyword arguments at the call site is to allow calls like the following to be inlined:

```ruby
def let(value, type, checked: true)
    value
end

# ...

let(5, Integer, checked: false)
```

Gadiguibou: I have to leave now but I will add this later this afternoon.

yjit/src/codegen.rs
Gadiguibou: I have to leave now but I will add this later this afternoon.

Gadiguibou: I added a comment [here](https://github.com/Shopify/ruby/blob/9df72f8ec0137a0539506d06e1503563783fd9d5/yjit/src/codegen.rs#L7050-L7051). Let me know if it's clear enough.

maximecb:
@XrXr are you satisfied with the changes?

XrXr (approved):

PR #11067, opened 2024-06-27
Fix comment for VM_CALL_ARGS_SIMPLE

Gadiguibou:
The comment for that field did not match its actual value. This corrects it. In the process, it changes the "set-site" slightly so it matches that new comment (_no change to functionality, just readability_).

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -1439,14 +1439,14 @@ new_callinfo(rb_iseq_t *iseq, ID mid, int argc, unsigned int flag, struct rb_cal
 {
     VM_ASSERT(argc >= 0);
 
-    if (!(flag & (VM_CALL_ARGS_SPLAT | VM_CALL_ARGS_BLOCKARG | VM_CALL_KW_SPLAT | VM_CALL_FORWARDING)) &&
-        kw_arg == NULL && !has_blockiseq) {
-        flag |= VM_CALL_ARGS_SIMPLE;
-    }
-
     if (kw_arg) {
         flag |= VM_CALL_KWARG;
         argc += kw_arg->keyword_len;
+    }
+
+    if (!(flag & (VM_CALL_ARGS_SPLAT | VM_CALL_ARGS_BLOCKARG | VM_CALL_KWARG | VM_CALL_KW_SPLAT | VM_CALL_FORWARDING))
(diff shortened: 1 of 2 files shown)


vm_callinfo.h
```diff
@@ -12,12 +12,13 @@
 #include "internal/class.h"
 #include "shape.h"
 
-enum vm_call_flag_bits {
+enum vm_call_flag_bits
+{
```
k0kubun, line 16: It seems like we don't do that style for things like structs/enums. Let's keep unrelated stuff as is.

```suggestion
enum vm_call_flag_bits {
```

compile.c
```diff
@@ -1439,16 +1439,16 @@ new_callinfo(rb_iseq_t *iseq, ID mid, int argc, unsigned int flag, struct rb_cal
...
     if (kw_arg) {
         flag |= VM_CALL_KWARG;
         argc += kw_arg->keyword_len;
     }
 
+    if (!(flag & (VM_CALL_ARGS_SPLAT | VM_CALL_ARGS_BLOCKARG | VM_CALL_KWARG | VM_CALL_KW_SPLAT | VM_CALL_FORWARDING))
+        && !has_blockiseq) {
+        flag |= VM_CALL_ARGS_SIMPLE;
+    }
+
```
k0kubun, line 1451:
```suggestion
    if (!(flag & (VM_CALL_ARGS_SPLAT | VM_CALL_ARGS_BLOCKARG | VM_CALL_KWARG | VM_CALL_KW_SPLAT | VM_CALL_FORWARDING))
        && !has_blockiseq) {
        flag |= VM_CALL_ARGS_SIMPLE;
    }

    if (kw_arg) {
        flag |= VM_CALL_KWARG;
        argc += kw_arg->keyword_len;
    }

```

To me, `VM_CALL_ARGS_SIMPLE` (the most common case) and `VM_CALL_ARGS_SPLAT` (appears earlier than keyword arguments) showing up first makes more sense. Let's only fix the comment and not squeeze in unrelated changes that don't necessarily make sense.

k0kubun: Oh, seeing the result of the reordering, now it makes more sense. I'm actually tempted to take your change in that case.

PR #11071, opened 2024-06-28
Synchronize Bundler & RubyGems

deivid-rodriguez:
I was informed that CI for Bundler is failing.

I thinking that's most likely because some changes failed to be synchronized automatically, but then other changes depending on the previous commit did manage to sneak in.

This PR reconciles the code with upstream.


k0kubun (approved):
thanks 👍

PR #11077, opened 2024-06-29
[DOC] Doc for Integer#floor

BurdetteLamar:


numeric.c
```diff
@@ -5739,24 +5739,60 @@ int_round(int argc, VALUE* argv, VALUE num)
...
+ *      |--------:|------------:|---------------------:|
+ *      | -1      | 10          | -1240                |
+ *      | -2      | 100         | -1300                |
+ *      | -3      | 1000        | -2000                |
+ *      | -4      | 10000       | -10000               |
+ *      | -5      | 100000      | -100000              |
  *
- *    555.floor     # => 555
- *    555.floor(50) # => 555
+ *      <br>
```
adam12, line 5795: Do you think we'd be able to avoid these? They pollute the markup displayed inside `ri`.

Markdown supports a line break using two or more white-space characters at the end of the line. Perhaps we could try that?

BurdetteLamar: Thanks, @adam12.  I'll look into it.  (There are others, elsewhere.)

BurdetteLamar: @adam12, I've removed the offending raw HTML.  The trailing whitespace did not work for me in rdoc 6.6.2 -- did not generate a line break -- so I have not added it in the recent commit.

adam12: Thanks for trying. I wonder if this is a limitation in the markdown parser that ships inside RDoc. I know the feature itself isn't well documented.

adam12: Looks like Markdown parser handles it correctly but maybe ignored when transforming the AST.

```
$ ruby -rrdoc -e '
pp RDoc::Markdown.parse(<<EOM)
First line with two spaces after.  
And the next line.
EOM'
[doc: [para: "First line with two spaces after.", [break], "And the next line."]]
```

BurdetteLamar: So what does that tell us (relative to this PR)?

adam12: You've already removed the HTML markup so I'm presuming the extra linebreak was just for aesthetics and not critical. If that's the case, I'm just musing in our existing discussion and don't think it's holding anything up.

BurdetteLamar:
@peterzhu2118, ready to go?

numeric.c
BurdetteLamar: Thanks, @adam12.

numeric.c
```diff
@@ -5739,24 +5739,56 @@ int_round(int argc, VALUE* argv, VALUE num)
...
+ *  :markup: markdown
+ *
  *  call-seq:
  *    floor(ndigits = 0) -> integer
  *
- *  Returns the largest number less than or equal to +self+ with
- *  a precision of +ndigits+ decimal digits.
+ *  Returns an integer that is a "floor" value for `self`,
+ *  as specified by the given `ndigits`,
+ *  which must be an
```
adam12, line 5749: Looks good to me, but minor nit: the linebreaks on this line look unnecessarily short.

BurdetteLamar: As an old-time documentor, I've found that future edits (and the diffs thereof) go better if each clause has its own line;  that's why I broke after each of the two commas.   The link is on its own line (l. 5750) because it's 90+ characters long by itself.

peterzhu2118 (approved):

numeric.c
adam12: Closing the loop on this, RDoc + Markdown where there are two spaces trailing a line will indeed insert a linebreak, but only if the line has non-whitespace characters. Empty lines or lines with just whitespace are collapsed.

BurdetteLamar: I thought I tried putting two spaces after the vertical bar in the table, but maybe I didn't.   Will try again when I remove the remaining `<br>`s from other markdown passages.

PR #11081, opened 2024-07-01
Use `__CET__` macro to enable IBT support.

voxik:
According to the GCC documentation [[1]], the macro `__CET__` is dfined when `-fcf-protection` compiler option is used. Therefore use this macro to enable IBT support instead of special casing for `__OpenBSD__`.

[1]: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fcf-protection

### DIFF
--- a/coroutine/amd64/Context.S
+++ b/coroutine/amd64/Context.S
@@ -5,7 +5,7 @@
 ##  Copyright, 2018, by Samuel Williams.
 ##
 
-#if defined(__OpenBSD__)
+#if defined(__CET__)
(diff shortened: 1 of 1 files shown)


voxik:
This builds on top of #10987 and is discussed as part of https://bugs.ruby-lang.org/issues/18061

jeremyevans (commented):
From an OpenBSD perspective, this is fine.

ioquatix:
Thanks, this is a good improvement.

PR #11082, opened 2024-07-01
[PRISM] Support node id in Prism

kddnewton:
This commit adds support for reproducable parses with node ids, like parse.y supports. Previously we attempted to replicate this functionality using line/column pairs, but this proved impossible to accomplish when attempting to support Rails' use of ErrorHighlight.spot because line numbers can differ between uncompiled and compiled versions of the source.

Unfortunately at the moment this is a CRuby-only API (Prism.of) but this could be extended to support other runtimes/tools if we begin exposing pm_node_t#id through some other mechanism.

Fortunately this doesn't actually end up taking up any more space on most 64-bit systems because we currently have a hole in `pm_node_t` on those systems, due to 32 bits worth of padding between the flags and the locations.

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -4218,6 +4218,91 @@ iseqw_script_lines(VALUE self)
     return ISEQ_BODY(iseq)->variable.script_lines;
 }
 
+// We need a handle on the script passed to -e in order to parse it again when
+// calling Prism.of.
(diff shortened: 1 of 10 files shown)


kddnewton:
@mame tagging you just to let you know of this change, no action necessary!

PR #11083, opened 2024-07-01
YJIT: Use a special breakpoint address if one isn't explicitly supplied in order to support natural line stepping.

nirvdrum:
ARM64 will not increment the program counter (PC) upon hitting a breakpoint instruction. Consequently, stepping through code with a debugger ends up looping back to the breakpoint instruction. LLDB has a special breakpoint address of 0xf000 that will increment the PC and allow the debugger to work as expected. This change makes it possible to debug YJIT generated code on ARM64.

More details at: https://discourse.llvm.org/t/stepping-over-a-brk-instruction-on-arm64/69766/8

### DIFF
--- a/yjit/src/asm/arm64/mod.rs
+++ b/yjit/src/asm/arm64/mod.rs
@@ -257,7 +257,7 @@ pub fn br(cb: &mut CodeBlock, rn: A64Opnd) {
 /// BRK - create a breakpoint
 pub fn brk(cb: &mut CodeBlock, imm16: A64Opnd) {
     let bytes: [u8; 4] = match imm16 {
-        A64Opnd::None => Breakpoint::brk(0).into(),
+        A64Opnd::None => Breakpoint::brk(0xf000).into(),
         A64Opnd::UImm(imm16) => {
             assert!(uimm_fits_bits(imm16, 16), "The immediate operand must be 16 bits or less.");
             Breakpoint::brk(imm16 as u16).into()


tenderlove (changes requested):
Looks good, but breaks some encoding tests. Can you update those please?

XrXr:
> a special breakpoint address

The argument to BRK is not an address, it's just a number without any predefined meaning: https://developer.arm.com/documentation/dui0801/l/A64-General-Instructions/BRK--A64-

I would `s/address/immediate/`.

Also, can you configure your editor to wrap commit messages with newlines? At the moment the whole thing is one very long line.

New suggested title: _YJIT: A64: Use a special BRK immediate to work better with LLDB_

maximecb (approved):

nirvdrum:
> The argument to BRK is not an address, it's just a number without any predefined meaning: https://developer.arm.com/documentation/dui0801/l/A64-General-Instructions/BRK--A64-
> 
> I would `s/address/immediate/`.

Fair enough. I read the docs and felt that would not really clarify anything. I was waiting to see if anyone would object to a more layman's definition. I thought being imprecise was useful to concisely describe things.

> Also, can you configure your editor to wrap commit messages with newlines? At the moment the whole thing is one very long line.

Is this a problem? I'd expect a tool to soft-wrap if the reader has a preference about line length. I'd rather not have to contort sentences to meet a cutoff. But, I'll do it if that's the expectation.

> New suggested title: YJIT: A64: Use a special BRK immediate to work better with LLDB

This is shorter, but also far less meaningful to me.

XrXr:
> Is this a problem? I'd expect a tool to soft-wrap if the reader has a preference about line length. I'd rather not have to contort sentences to meet a cutoff. But, I'll do it if that's the expectation.

It's a curtesy for `git log` users, of course it's not a hard requirement by any means. Also, I'm not sure if `git log` supports soft-wrapping just the message. Maybe it's possible through some settings for the pager.

Screenshot 2024-07-02 at 16 58 44

PR #11087, opened 2024-07-02
Speed up chunkypng benchmark

tenderlove:
Since d037c5196a14c03e72746ccdf0437b5dd4f80a69 we're seeing a slowdown in ChunkyPNG benchmarks in YJIT bench. This patch addresses the slowdown. Making the thread volatile speeds up the benchmark by 2 or 3% on my machine.

```
before: ruby 3.4.0dev (2024-07-02T18:48:43Z master b2b8306b46) [x86_64-linux]
after: ruby 3.4.0dev (2024-07-02T20:07:44Z speed-chunkypng 418334dba9) [x86_64-linux]

chunky-png  1000.2       0.1         980.6       0.1         1.02           1.02
Legend:

Output:
./data/output_015.csv
```

cc @k0kubun @maximecb

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -1771,7 +1771,7 @@ VALUE
 rb_thread_io_blocking_call(rb_blocking_function_t *func, void *data1, int fd, int events)
 {
     rb_execution_context_t *volatile ec = GET_EC();
-    rb_thread_t * volatile th = rb_ec_thread_ptr(ec);
+    rb_thread_t *volatile th = rb_ec_thread_ptr(ec);
 
     RUBY_DEBUG_LOG("th:%u fd:%d ev:%d", rb_th_serial(th), fd, events);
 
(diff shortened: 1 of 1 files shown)


k0kubun (approved):

thread.c
```diff
@@ -1771,7 +1771,7 @@ VALUE
 rb_thread_io_blocking_call(rb_blocking_function_t *func, void *data1, int fd, int events)
 {
     rb_execution_context_t *volatile ec = GET_EC();
-    rb_thread_t *th = rb_ec_thread_ptr(ec);
+    rb_thread_t * volatile th = rb_ec_thread_ptr(ec);
```
XrXr, line 1774:
```suggestion
    rb_thread_t *volatile th = rb_ec_thread_ptr(ec);
```

sorry to be annoying, but there is no space between the two on just the line above 😆

maximecb (approved):

PR #3296, opened 2020-07-06
Tidy up vm_push_frame

shyouhei:
For instance its return value is used from nowhere so why not return `void`.

It seems compilers are smarter than me.  Neither speedup nor slowdown can be observed.  Must harm no one, though.

```
master: ruby 2.8.0dev (2020-07-01T13:41:16Z master 49029811d9) [aarch64-linux]
ours: ruby 2.8.0dev (2020-07-06T06:41:57Z origin/vm_push_frame 37bac56b10) [aarch64-linux]
Calculating -------------------------------------
                                       master                  ours
Optcarrot Lan_Master.nes   24.991672635908220    24.674321099849632 fps
                           25.049511891086389    24.751949030361153
                           25.107730926780810    24.832866393560870
                           25.193115588149983    25.084869359715185
                           25.227626632615966    25.130355271340051
                           25.247251028021367    25.223713670923104
                           25.323752498520427    25.294004296822507
```

PR #11091, opened 2024-07-03
[Backport #11036 to 3.2] Add explicit compiler fence when pushing frames to ensure safe profiling

ivoanjo:
**What does this PR do?**

This PR tweaks the `vm_push_frame` function to add an explicit compiler fence (`atomic_signal_fence`) to ensure profilers that use signals to interrupt applications (stackprof, vernier, pf2, Datadog profiler) can safely sample from the signal handler.

This is a backport of #11036 to Ruby 3.2 .

**Motivation:**

The `vm_push_frame` was specifically tweaked in https://github.com/ruby/ruby/pull/3296 to initialize the a frame before updating the `cfp` pointer.

But since there's nothing stopping the compiler from reordering the initialization of a frame (`*cfp =`) with the update of the cfp pointer (`ec->cfp = cfp`) we've been hesitant to rely on this on the Datadog profiler.

In practice, after some experimentation + talking to folks, this reordering does not seem to happen.

But since modern compilers have a way for us to exactly tell them not to do the reordering (`atomic_signal_fence`), this seems even better.

I've actually extracted `vm_push_frame` into the "Compiler Explorer" website, which you can use to see the assembly output of this function across many compilers and architectures: https://godbolt.org/z/3oxd1446K

On that link you can observe two things across many compilers:
1. The compilers are not reordering the writes
2. The barrier does not change the generated assembly output (== has no cost in practice)

**Additional Notes:**

The checks added in `configure.ac` define two new macros:
* `HAVE_STDATOMIC_H`
* `HAVE_DECL_ATOMIC_SIGNAL_FENCE`

Since Ruby generates an arch-specific `config.h` header with these macros upon installation, this can be used by profilers and other libraries to test if Ruby was compiled with the fence enabled.

**How to test the change?**

As I mentioned above, you can check https://godbolt.org/z/3oxd1446K to confirm the compiled output of `vm_push_frame` does not change in most compilers (at least all that I've checked on that site).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1347,6 +1347,8 @@ AC_CHECK_HEADERS(syscall.h)
 AC_CHECK_HEADERS(time.h)
 AC_CHECK_HEADERS(ucontext.h)
 AC_CHECK_HEADERS(utime.h)
+AC_CHECK_HEADERS(stdatomic.h)
+
 AS_CASE("$target_cpu", [x64|x86_64|i[3-6]86*], [
   AC_CHECK_HEADERS(x86intrin.h)
 ])
@@ -2004,6 +2006,7 @@ AC_CHECK_FUNCS(_longjmp)		# used for AC_ARG_WITH(setjmp-type)
 test x$ac_cv_func__longjmp = xno && ac_cv_func__setjmp=no
 AC_CHECK_FUNCS(arc4random_buf)
 AC_CHECK_FUNCS(atan2l atan2f)
+AC_CHECK_DECLS(atomic_signal_fence, [], [], [#include <stdatomic.h>])
 AC_CHECK_FUNCS(chmod)
 AC_CHECK_FUNCS(chown)
 AC_CHECK_FUNCS(chroot)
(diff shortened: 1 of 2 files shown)


ivoanjo:
Huh... I see one CI job "pending" and another one "failed" but they both seem to be flaky CI rather than this change.

Pending job:

[image
](https://github.com/ruby/ruby/actions/runs/9779016853/job/26997197402?pr=11091)

Travis failure: 

[image
](https://app.travis-ci.com/github/ruby/ruby/jobs/623660272)

I'll re-push the commit to trigger CI again and see if it the flaky stuff goes away.

ivoanjo:
CI is almost green -- I think github seems be struggling with macos 11 runners? :eyes: 

image

hsbt:
@ivoanjo I update our CI configuration at https://github.com/ruby/ruby/pull/11108

Can you rebase with that?

ivoanjo:
Rebase pushed! :+1: Fingers crossed it'll finally be green :)

ivoanjo:
Success! CI is now green, no more flaky failures ^^

PR #11092, opened 2024-07-03
array.c: Remove outdated assertions

casperisfine:
Following [Feature #20589] it can happen that we change the capacity of a frozen array, so these assertions no longer make sense.

Normally we don't hit them because `Array#freeze` shrinks the array, but if somehow the Array was frozen using `Object#freeze` then we may shrink it after it was frozen.

cc @eileencodes


mame:
@byroot As far as I see, [Feature #20589] wanted to ensure the capacity of a frozen array *unchanged*. In fact, the assertion in question was added by [Feature #20589] itself. I think it was never outdated.

I think what we need here is for rb_obj_freeze to delegate to rb_ary_freeze when the object is T_ARRAY. I have created a PR #11103.

PR #11099, opened 2024-07-03
Raise a compile error for break/next/redo inside eval in cases where it is optimized away

jeremyevans:
In cases where break/next/redo are not valid syntax, they should raise a SyntaxError even if inside a conditional block that is optimized away.

@nobu This appears to be the simplest change, and passes tests, but I would appreciate your review in case I am missing a corner case that is not tested.

@kddnewton Can you update the prism parser for this?  I made changes to the prism_compiler, but they don't pass the specs I added.  I'm guessing the prism compiler changes are for dead code if you make changes to the prism parser (I intend to clean up the parse.y compiler to remove the related dead code in a later commit).

Fixes [Bug #20597]


kddnewton:
Hey @jeremyevans yup I'll have a look

kddnewton:
Hey @jeremyevans — I pushed a fix for Prism in a separate commit. Let me know if there's any other issues!

PR #11103, opened 2024-07-04
Make rb_obj_free use rb_ary_freeze when the target is an Array

mame:
... and restore the assertions removed by 7542ae458a214ef7e160935555643265b6304ffd

@byroot @eileencodes

### DIFF
--- a/array.c
+++ b/array.c
@@ -652,7 +652,8 @@ rb_ary_freeze(VALUE ary)
         ary_shrink_capa(ary);
     }
 
-    return rb_obj_freeze(ary);
+    OBJ_FREEZE(ary);
+    return ary;
 }
 
 /* This can be used to take a snapshot of an array (with
--- a/object.c
+++ b/object.c
@@ -1317,6 +1317,11 @@ VALUE
 rb_obj_freeze(VALUE obj)
 {
     if (!OBJ_FROZEN(obj)) {
+        // If the target object is an array, we want to shrink it
+        if (RB_TYPE_P(obj, T_ARRAY)) {
+            return rb_ary_freeze(obj);
+        }
+
         OBJ_FREEZE(obj);
         if (SPECIAL_CONST_P(obj)) {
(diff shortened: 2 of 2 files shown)


mame:
We might want to remove the definition of `Array#freeze` which would be no longer needed.

byroot:
So I thought about doing this, but the reason I didn't is that I wasn't certain there would be no way left for C extensions to freeze arrays without shrinking them.

If you are confident that's the case 👍

mame:
Hmm. I actually don't like my patch very much. It is not beautiful to have special handling only for Array in rb_obj_freeze. 

The reason why I did write this nevertheless is because I think this is what #11030 set out to do. Defining `Array#freeze` and preferring `rb_ary_freeze` to `rb_obj_freeze` is an incomplete way to do it.

If we don't have to complete #11030, then it makes more sense to revert #11030 rather than erasing the assertion in ARY_SET_CAPA.

TBH I do not understand the motivation behind #11030. was there something wrong with ary_heap_realloc for a frozen array?

@eileencodes @tenderlove What do you think?

tenderlove:
> TBH I do not understand the motivation behind #11030. was there something wrong with ary_heap_realloc for a frozen array?

Nothing wrong in particular, but since the array is frozen I would think we consider it "immutable".  It's possible realloc could return a different pointer, so the underlying buffer could change even though we consider the object immutable.  Certainly we could allow the buffer pointer to change even though it's frozen, it just seems very strange (to me).

jeremyevans:
> Certainly we could allow the buffer pointer to change even though it's frozen, it just seems very strange (to me).

Isn't the buffer pointer already changed for embedded arrays whenever they are moved by `GC.compact` (by buffer pointer, I mean the pointer that would be set in `RARRAY_PTR_USE` before and after compaction)?

mame:
The assertion in question was restored by #11124 (thanks!) so I'd like to close this PR.

PR #11105, opened 2024-07-05
Use macos-14 instead of macos-arm-oss

hsbt:
Unfortunately, we can't use `macos-arm-oss` with enterprise account


ono-max (commented):
Can you please update https://github.com/ruby/ruby/pull/11105/files#diff-7b74d73de878e0e53f6768bbf3a3d4b50c3c6f1f16f9b64518690028b8206b1cR101 and  https://github.com/ruby/ruby/pull/11105/files#diff-72335a32d9627809751d1cd12751e4997ae9caafd11cbc86925ee43410464128R122, too?

PR #11109, opened 2024-07-05
Raise a TypeError for Thread#thread_variable{?,_get} for non-symbol

jeremyevans:
Previously, a TypeError was not raised if there were no thread variables, because the conversion to symbol was done after that check.  Convert to symbol before checking for whether thread variables are set to make the behavior consistent.

Fixes [Bug #20606]


eregon (approved):
Thank you!

PR #11110, opened 2024-07-05
Delay the array allocation of multiple arguments for Enumerable#find, #any?, etc.

mame:
```
def allocated_now = GC.stat(:total_allocated_objects)

dummy_hash = 100_000.times.each_with_index.to_h
before_allocs = allocated_now
dummy_hash.find {|k, v| false }
puts "Allocated #{allocated_now - before_allocs}"
# before this patch: Allocated 100005
# after this patch:  Allocated 5
```

[[Bug #20608]](https://bugs.ruby-lang.org/issues/20608)

### DIFF
--- a/enum.c
+++ b/enum.c
@@ -322,16 +322,32 @@ enum_count(int argc, VALUE *argv, VALUE obj)
     return imemo_count_value(memo);
 }
 
+NORETURN(static void found(VALUE i, VALUE memop));
+static void
+found(VALUE i, VALUE memop) {
+    struct MEMO *memo = MEMO_CAST(memop);
+    MEMO_V1_SET(memo, i);
+    memo->u3.cnt = 1;
+    rb_iter_break();
+}
+
+static VALUE
+find_i_fast(RB_BLOCK_CALL_FUNC_ARGLIST(i, memop))
+{
+    if (RTEST(rb_yield_values2(argc, argv))) {
+        ENUM_WANT_SVALUE();
+        found(i, memop);
+    }
+    return Qnil;
+}
+
 static VALUE
 find_i(RB_BLOCK_CALL_FUNC_ARGLIST(i, memop))
 {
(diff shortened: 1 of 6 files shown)


ko1 (approved):

ko1:
Could you check the performance on Hash (this patch will improve the speed) and Range (this patch doesn't improve the speed)?

mame:
I evaluated `all?` on Hash and Range with the following scripts.

```
# hash-all.rb (this patch will improve the speed)
big_hash = 1_000_000.times.each_with_index.to_h
t = Time.now
o = GC.stat(:total_allocated_objects)
30.times do
  big_hash.all? {|k, v| true }
end
p [Time.now - t, GC.stat(:total_allocated_objects) - o]
```

```
# range-all.rb (this patch doesn't improve the speed)
t = Time.now
o = GC.stat(:total_allocated_objects)
(0...30_000_000).all? {|k, v| true }
p [Time.now - t, GC.stat(:total_allocated_objects) - o]
```

On one machine, the speedup was almost twice as fast for Hash, and only slightly faster for Range.

```
$ ./miniruby.orig hash-all.rb
[3.864568559, 30000069]
$ ./miniruby.patched hash-all.rb
[1.980329461, 69]
$ ./miniruby.orig range-all.rb
[3.811854031, 10]
$ ./miniruby.patched range-all.rb
[3.640517268, 10]
```

On the other machine, speedup of about 25% was observed for Hash, and speed *degradation* of about 10% was observed for Range.

```
$ ./miniruby.orig hash-all.rb
[1.612158798, 30000069]
$ ./miniruby hash-all.rb
[1.204641842, 69]
$ ./miniruby.orig range-all.rb
[1.649715745, 10]
$ ./miniruby range-all.rb
[1.819054961, 10]
```

I guess the performance is very sensitive to the compiler threshold or something.

jeremyevans (approved):

PR #11111, opened 2024-07-05
Add make target shared-gc

peterzhu2118:
Allows building shared GC using `make shared-gc SHARED_GC=gc_impl`


common.mk
```diff
@@ -1905,6 +1905,20 @@ rewindable:
 
 HELP_EXTRA_TASKS = ""
 
+shared-gc: probes.h
+	$(Q) @if test -z $(shared_gc_dir); then \
+		echo "You must configure with --with-shared-gc to use shared GC"; \
+		exit 1; \
+	else \
+		if test -z $(SHARED_GC); then \
```
nobu, line 1913: Why not `elif`?

common.mk
```diff
@@ -1905,6 +1905,20 @@ rewindable:
 
 HELP_EXTRA_TASKS = ""
 
+shared-gc: probes.h
+	$(Q) @if test -z $(shared_gc_dir); then \
```
nobu, line 1909: Why `@` after `$(Q)`?

PR #11113, opened 2024-07-06
Integrate Launchable into make btest

ono-max:
Recently, Launchable has been used for analyzing flaky tests. Like https://github.com/ruby/ruby/pull/11043, Kokubun-san asked me to integrate Launchable into `make btest` for analyzing flay tests in `make btest`, so I've created this PR to address it.

RJIT is a good place as a first step for `make btest` since RJIT is flaky now.


hsbt:
I suggest to move `tool/lib/test/unit/launchable.rb` to `tool/lib/launchable.rb` same as `envutil.rb`, `leakchecker.rb`.

We should move them to under the `tool/lib/test`. But it has side effect for LOAD_PATH of the default gems.

PR #11123, opened 2024-07-08
Run annocheck for libruby.so

voxik:
When building with `--shared` option, most functionality is kept in `libruby.so`. Therefore also run annocheck for `libruby.so`. Testing `ruby` executable is almost pointless.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1668,7 +1668,7 @@ no-test-bundler-parallel:
 # architecture. It is designed to be independent of the host OS and the
 # architecture. The test-annocheck.sh requires docker or podman.
 test-annocheck: $(PROGRAM)
-	$(tooldir)/test-annocheck.sh $(PROGRAM)
+	$(tooldir)/test-annocheck.sh $(PROGRAM) $(LIBRUBY_SO)
 
 GEM = up
 sync-default-gems:
--- a/tool/test-annocheck.sh
+++ b/tool/test-annocheck.sh
@@ -27,7 +27,8 @@ else
   # volume in container in container on GitHub Actions
   # <.github/workflows/compilers.yml>.
   TAG="${TAG}-copy"
-  "${DOCKER}" build --rm -t "${TAG}" --build-arg=FILES="${*}" -f ${TOOL_DIR}/annocheck/Dockerfile-copy .
+  sed -r "s/\\$\{FILES\}/${*}/" ${TOOL_DIR}/annocheck/Dockerfile-copy | \
+    "${DOCKER}" build --rm -t "${TAG}" --build-arg=FILES="${*}" -f - .
 fi
 
 "${DOCKER}" run --rm -t ${DOCKER_RUN_VOLUME_OPTS} "${TAG}" annocheck --verbose ${TEST_ANNOCHECK_OPTS-} "${@}"
--- a/tool/annocheck/Dockerfile-copy
+++ b/tool/annocheck/Dockerfile-copy
@@ -2,6 +2,5 @@ FROM docker.io/fedora:latest
 ARG FILES
 
 RUN dnf -y install annobin-annocheck
-RUN mkdir /work
-COPY ${FILES} /work
+COPY ${FILES} /work/
 WORKDIR /work


voxik:
A few thing to consider:

1) Not sure if the target should not depend also on existence of `$(LIBRUBY_SO)`. IOW if this version would not be better:

   ~~~patch
   $ git diff
   index 05145313c6..cfc474ca19 100644
   --- a/common.mk
   +++ b/common.mk
   @@ -1667,8 +1667,8 @@ no-test-bundler-parallel:
    # The annocheck supports ELF format binaries compiled for any OS and for any
    # architecture. It is designed to be independent of the host OS and the
    # architecture. The test-annocheck.sh requires docker or podman.
   -test-annocheck: $(PROGRAM)
   -       $(tooldir)/test-annocheck.sh $(PROGRAM)
   +test-annocheck: $(PROGRAM) $(LIBRUBY_SO)
   +       $(tooldir)/test-annocheck.sh $(PROGRAM) $(LIBRUBY_SO)
    
    GEM = up
    sync-default-gems:
   ~~~

2) Not sure if it would not be good idea to test all *.so files
3) It could even make sense to test the *.o files, because that would provide more details, should there be some issue.

voxik:
Apparently `COPY $FILES /work/` treats `ruby libruby.so3.4.0` env variable as a single file (discussed [here](https://stackoverflow.com/questions/61599384/specify-multiple-files-in-arg-to-copy-in-dockerfile). So this won't be as easy as it could.

voxik:
> Apparently `COPY $FILES /work/` treats `ruby libruby.so.3.4.0` env variable as a single file (discussed [here](https://stackoverflow.com/questions/61599384/specify-multiple-files-in-arg-to-copy-in-dockerfile). So this won't be as easy as it could.

I decided to expand the `${FILES}` variable by `sed` instead.

voxik:
> Not sure if the build should not be configured with `--disable-yjit`, because Rust hides possible fortification issues:
> 
> ```
> Hardened: libruby.so.3.4.0: skip: cf-protection test because mixed Rust and C code - control flow protection is needed but not yet supported by Rust 
> ```

And just FTR, this is the Rust tracker:

https://github.com/rust-lang/rust/issues/93754

tool/test-annocheck.sh
```diff
@@ -27,7 +27,8 @@ else
   # volume in container in container on GitHub Actions
   # <.github/workflows/compilers.yml>.
   TAG="${TAG}-copy"
-  "${DOCKER}" build --rm -t "${TAG}" --build-arg=FILES="${*}" -f ${TOOL_DIR}/annocheck/Dockerfile-copy .
+  sed -r "s/\\$\{FILES\}/${*}/" ${TOOL_DIR}/annocheck/Dockerfile-copy | \
```
nobu, line 30: Isn't `/` as the separator unsafe, as it is also a file separator.

common.mk
```diff
@@ -1668,7 +1668,7 @@ no-test-bundler-parallel:
 # architecture. It is designed to be independent of the host OS and the
 # architecture. The test-annocheck.sh requires docker or podman.
 test-annocheck: $(PROGRAM)
```
junaruga, line 1670: The `$(LIBRUBY_SO)` is missed at this line.

tool/test-annocheck.sh
junaruga: @nobu I would agree with your opinion. the `/` is unsafe when the `${*}` includes the file separator `/`.

However, I don't think we need to use the `sed` command. Because this hack just ignores the functionality of the `--build-arg=FILES="${*}"` as an argument. And I found alternative ideas to deal with this limitation in a way to understand the logic easily in my opinion. I will share it soon.

junaruga (changes requested):
For someone who is interested in the annocheck test on CI. The test is executed in the `.github/workflows/annocheck.yml` calling the `make test-annocheck` calling `tool/test-annocheck.sh`.

I tested and debugged this PR on my following branch in my forked repository. The latest commit on the branch is my change proposal for this PR. 

https://github.com/junaruga/ruby/tree/wip/v_annocheck-libruby-so-debug

Below is the diff. I think this way is better in terms of understanding the logic easily without using the `sed` hack making the `--build-arg=FILES="${*}"` in the `docker build` command meaningless.

What do you think?

```diff
diff --git a/tool/annocheck/Dockerfile-copy b/tool/annocheck/Dockerfile-copy
index 132febc86c..0a79f3a50a 100644
--- a/tool/annocheck/Dockerfile-copy
+++ b/tool/annocheck/Dockerfile-copy
@@ -1,6 +1,6 @@
 FROM docker.io/fedora:latest
-ARG FILES
+ARG IN_DIR

 RUN dnf -y install annobin-annocheck
-COPY ${FILES} /work/
+COPY ${IN_DIR} /work
 WORKDIR /work
diff --git a/tool/test-annocheck.sh b/tool/test-annocheck.sh
index f6a6bd8223..cbe833d97c 100755
--- a/tool/test-annocheck.sh
+++ b/tool/test-annocheck.sh
@@ -15,6 +15,7 @@ set -x
 DOCKER="$(command -v docker || command -v podman)"
 TAG=ruby-fedora-annocheck
 TOOL_DIR=$(dirname "${0}")
+TMP_DIR="tmp/annocheck"
 DOCKER_RUN_VOLUME_OPTS=

 if [ -z "${CI-}" ]; then
@@ -27,8 +28,13 @@ else
   # volume in container in container on GitHub Actions
   # <.github/workflows/compilers.yml>.
   TAG="${TAG}-copy"
-  sed -r "s/\\$\{FILES\}/${*}/" ${TOOL_DIR}/annocheck/Dockerfile-copy | \
-    "${DOCKER}" build --rm -t "${TAG}" --build-arg=FILES="${*}" -f - .
+  rm -rf "${TMP_DIR}"
+  mkdir -p "${TMP_DIR}"
+  for file in ${@}; do
+    cp -p "${file}" "${TMP_DIR}"
+  done
+  "${DOCKER}" build --rm -t "${TAG}" --build-arg=IN_DIR="${TMP_DIR}" -f ${TOOL_DIR}/annocheck/Dockerfile-copy .
+  rm -rf "${TMP_DIR}"
 fi

 "${DOCKER}" run --rm -t ${DOCKER_RUN_VOLUME_OPTS} "${TAG}" annocheck --verbose ${TEST_ANNOCHECK_OPTS-} "${@}"
```

I confirmed the `make test-annocheck` passed on GitHub Actions on my forked repository.
https://github.com/junaruga/ruby/actions/runs/10073821672/job/27848650745#step:10:106

junaruga:
By the way, below is an article about `${@}` vs `${*}`.

https://www.baeldung.com/linux/dollar-star-at
> In most cases, using $@ or “$@” instead of $* is safer and more flexible when working with command-line arguments because they preserve the original argument structure and make it easier to work with individual arguments in our bash script.

junaruga:
I did cherry-pick this PR, adding my change, and opened another PR https://github.com/ruby/ruby/pull/11324. So, I would close this PR. Thanks for your PR!

PR #11125, opened 2024-07-08
[DOC] Doc for Float#ceil

BurdetteLamar:


numeric.c
```diff
@@ -2213,36 +2213,78 @@ flo_floor(int argc, VALUE *argv, VALUE num)
...
+ *  f.ceil(30) # => -12345.6789
+ *  f = 0.0
+ *  f.ceil(1)   # => 0.0
+ *  f.ceil(100) # => 0.0
+ *  ```
+ *
+ *  When `ndigits` is non-positive,
+ *  returns an Integer based on a computed granularity:
+ *
+ *  - The granularity is ``ndigits.abs * 10`.
```
peterzhu2118, line 2247: I think this formula is incorrect, I opened #11129 to fix the existing ones.

```suggestion
 *  - The granularity is `10 ** ndigits.abs`.
```

PR #11129, opened 2024-07-08
[DOC] Fix granularity calculation

peterzhu2118:
The granularity is calculated as `10 ** ndigits.abs` rather than `ndigits.abs * 10`. For example, if `ndigits` is `-2`, it should be `10 ** 2 == 100` rather than `2 * 10 == 20`.


BurdetteLamar (approved):
Thanks for noticing this blunder.

PR #11130, opened 2024-07-09
ruby/spec changes to future proof for 64-bit fixnums on 64-bit Windows

XrXr:
See discussions in [[Bug #20614]](https://bugs.ruby-lang.org/issues/20614) https://bugs.ruby-lang.org/issues/20614#note-9

See commentary in each commit.

### DIFF
--- a/spec/mspec/lib/mspec/guards/platform.rb
+++ b/spec/mspec/lib/mspec/guards/platform.rb
@@ -53,21 +53,25 @@ class PlatformGuard < SpecGuard
     end
   end
 
-  WORD_SIZE = 1.size * 8
-
   POINTER_SIZE = begin
     require 'rbconfig/sizeof'
     RbConfig::SIZEOF["void*"] * 8
   rescue LoadError
-    WORD_SIZE
+    [0].pack('j').size
   end
 
-  def self.wordsize?(size)
-    size == WORD_SIZE
+  C_LONG_SIZE = if defined?(RbConfig::SIZEOF[])
+    RbConfig::SIZEOF["long"] * 8
+  else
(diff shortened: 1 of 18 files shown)


XrXr:
Looks like this breaks test-spec on Windows a lot. Not so easy!

spec/mspec/lib/mspec/guards/platform.rb
```diff
@@ -62,6 +62,12 @@ def self.wsl?
     WORD_SIZE
   end
 
+  C_LONG_SIZE = if defined?(RbConfig::SIZEOF[])
```
eregon, line 65: I wonder what `WORD_SIZE` actually means given there is already `POINTER_SIZE`.
Maybe `WORD_SIZE` is meant to be `sizeof(long)` already but with a bad name?
Do you think it would make sense to basically rename :wordsize to :c_long_size (or make it an alias of it and still update all usages).

XrXr: Nice catch! In fact, it looks like the only usage of `wordsize?` left is `platform_spec`. Will try for a rename.

eregon (approved):
Thank you, this is great!

eregon:
FWIW I think one reason ruby/spec or mspec didn't use much `RbConfig::SIZEOF["type"]` is it didn't exist in older versions and it's a C extension, which can be annoying e.g. when adding support for a new platform for TruffleRuby and it boots but loading C extensions doesn't work yet; or e.g. when working on a new Ruby impl which likely does not have RbConfig::SIZEOF yet. But since it's all guarded with `LoadError` that should work fine or be easy to work around.

PR #11133, opened 2024-07-09
Shrink capa from OBJ_FREEZE for arrays

eileencodes:
When we freeze an array ensure that the capacity is shrunk first. We're already doing type checking in `OBJ_FREEZE` for `T_STRING` so this shouldn't impact performance.

Some gems like JSON call `OBJ_FREEZE` so they don't go through `rb_obj_freeze` or `rb_ary_freeze`, and instead use `rb_obj_freeze_inline`.

Followup to #11124

cc/ @tenderlove @byroot @mame

### DIFF
--- a/array.c
+++ b/array.c
@@ -450,6 +450,14 @@ ary_shrink_capa(VALUE ary)
     ary_verify(ary);
 }
 
+void
+rb_ary_shrink_capa(VALUE ary)
+{
+    if (!ARY_EMBED_P(ary) && !ARY_SHARED_P(ary) && !ARY_SHARED_ROOT_P(ary)) {
+        ary_shrink_capa(ary);
+    }
+}
+
 static void
 ary_double_capa(VALUE ary, long min)
 {
@@ -647,9 +655,7 @@ rb_ary_freeze(VALUE ary)
 
     if (OBJ_FROZEN(ary)) return ary;
 
-    if (!ARY_EMBED_P(ary) && !ARY_SHARED_P(ary) && !ARY_SHARED_ROOT_P(ary)) {
-        ary_shrink_capa(ary);
-    }
+    rb_ary_shrink_capa(ary);
 
     return rb_obj_freeze(ary);
 }
--- a/internal/array.h
+++ b/internal/array.h
@@ -45,6 +45,7 @@ VALUE rb_ary_tmp_new_from_values(VALUE, long, const VALUE *);
 VALUE rb_check_to_array(VALUE ary);
 VALUE rb_ary_behead(VALUE, long);
 VALUE rb_ary_aref1(VALUE ary, VALUE i);
+void rb_ary_shrink_capa(VALUE ary);
 
 struct rb_execution_context_struct;
 VALUE rb_ec_ary_new_from_values(struct rb_execution_context_struct *ec, long n, const VALUE *elts);
--- a/variable.c
+++ b/variable.c
@@ -1812,13 +1812,19 @@ rb_shape_set_shape_id(VALUE obj, shape_id_t shape_id)
  *
  * @param[out]  x  Object in question.
  */
-void rb_obj_freeze_inline(VALUE x)
+void
+rb_obj_freeze_inline(VALUE x)
 {
     if (RB_FL_ABLE(x)) {
-        RB_FL_SET_RAW(x, RUBY_FL_FREEZE);
-        if (TYPE(x) == T_STRING) {
+        switch (BUILTIN_TYPE(x)) {
+          case T_STRING:
             RB_FL_UNSET_RAW(x, FL_USER3); // STR_CHILLED
+            break;
+          case T_ARRAY:
+            rb_ary_shrink_capa(x);
+            break;
         }
+        RB_FL_SET_RAW(x, RUBY_FL_FREEZE);
 
         rb_shape_t * next_shape = rb_shape_transition_shape_frozen(x);
 


byroot (approved):
I'm OK with this.

rhenium:
Wouldn't it be better that json gem is updated to use `rb_ary_freeze()` etc. appropriately?

I think those all-uppercase functions are meant to be very low level in general, so I'm afraid making `OBJ_FREEZE()` potentially raise `NoMemoryError` or invalidate pointers obtained with `RARRAY_PTR()` may break existing C extensions. I can't immediately think of an example, though.

byroot:
> Wouldn't it be better that json gem is updated to use `rb_ary_freeze()` etc. appropriately?

The issue is that in many case, and in the case of `json`, `OBJ_FREEZE` is used on an object of unknown type.

So it would need to be replaced by a `switch(RB_TYPE(obj))...` which is a bit of a hard sell.

rhenium:
This PR completely removes ways to freeze an array cheaply without doing the shrinkage. I hope low-level functions will continue to be provided and I'd expect `OBJ_FREEZE()` to be the function, given the all-uppercase name and the existence of other `rb_*_freeze()` functions.

We happen to have `rb_obj_freeze()` exposed to C extensions, which is currently mostly equivalent to `OBJ_FREEZE()`. Is it something that could be considered to repurpose it instead, leaving `OBJ_FREEZE()` as is?

That may be a safer approach while not requiring too much changes in C extensions. The name sounds less low-level or internal, and hopefully users who absolutely need low-level operation would have chosen `OBJ_FREEZE()` according to the convention.

XrXr:
I don't like forcing everyone to shrink-on-freeze either. It's expectation defying and a subtle, possibly breaking change.

See also: https://github.com/ruby/ruby/pull/11103#issuecomment-2208349628

tenderlove:
> The issue is that in many case, and in the case of `json`, `OBJ_FREEZE` is used on an object of unknown type.

The JSON gem does know the type, I think the upstream code just didn't want to repeat "freezing" logic.  For example, in [this production](https://github.com/flori/json/blob/80b58650065b2e937dc0ab0b6708e8afae117acd/ext/json/ext/parser/parser.rl#L404) it knows it's parsing an array, and could use the appropriate freeze function.  The annoying thing is that we have to go around and patch all C extensions that _could_ be context aware, but aren't.

> This PR completely removes ways to freeze an array cheaply without doing the shrinkage.

What is the use case for not shrinking?  Besides this, I can't imagine reallocing to a _smaller_ buffer is that expensive.  Of course it depends on the allocator though.

> It's expectation defying and a subtle, possibly breaking change.

Is it?  Before 6f6aff56b11d7f3df9b446b9399cc546875e07f5, we would auto-shrink arrays when they became shared.  That seems even more expectation defying.  I can easily explain to someone "when you freeze an array, it gets right sized", versus "your frozen array will get right sized if something happens to dup it".

eileencodes:
> We happen to have rb_obj_freeze() exposed to C extensions, which is currently mostly equivalent to OBJ_FREEZE(). Is it something that could be considered to repurpose it instead, leaving OBJ_FREEZE() as is?

We can certainly do this change in `rb_obj_freeze()` but then we're ignoring the C extensions and will have to change if they want shrinking to happen first. I chose `OBJ_FREEZE()` because it is already doing type checking so it seemed like a natural place to check for and shrink arrays too.

XrXr:
> Is it? That seems even more expectation defying. I can easily explain to someone "when you freeze an array, it gets right sized", versus "your frozen array will get right sized if something happens to dup it".

False dichotomy. I wouldn't expect `OBJ_FREEZE()` to be a pointer invalidation hazard, and there is no way to tell from the C API docs. [Feature #20589] was "Resize array in `rb_ary_freeze` and use `rb_ary_freeze` internally for arrays", but now we're touching something else. 

What rubs me the wrong way is that this is clearly a change to the public interface (c.f. pointer invalidation problem mentioned earlier), and it's being pushed through without discussion on the tracker and with no update to the C API documentation.

Further, putting in logic without replacement feels paternalistic. Sure, in many cases shrinking is what C extensions want. But not always. Forcing the logic upon them without replacement feels like saying you know better than what the C extension authors want, and you don't care about their intention.

eileencodes:
> What rubs me the wrong way is that this is clearly a change to the public interface, and it's being pushed through without discussion on the tracker and with no update to the C API documentation.

That was entirely not the intention. I opened an issue on the tracker [21 days ago](https://bugs.ruby-lang.org/issues/20589) and there wasn't any feedback for 14 days. But this was also discussed previously a [few years ago](https://bugs.ruby-lang.org/issues/16291) and there were no objections. We ended up opening followup PRs because JSON broke, and no one has merged this change without discussion. I can certainly open a new issue or we can discuss this on the one I opened already. I'm not trying to push anything through without feedback. I was just trying to fix a bug I introduced in a prior change.

luke-gru:
I agree that this could break C extensions due to pointer invalidation and it could be very confusing to some authors. At the very least I think it needs to be documented if it's changed.

XrXr:
Please make a ticket about changing `OBJ_FREEZE` if you want to proceed. This is a breaking change that deserves to be broadcasted on the mailing list. Your ticket doesn't mention changing `OBJ_FREEZE`, 
and the older [`#16291`](https://bugs.ruby-lang.org/issues/16291) doesn't proposal touching the public interface either.

Concerns about compatibility might be small enough that in practice it's not going to break too many people, but we need to run the risk consciously, and with proper documentation and consensus.

PR #11139, opened 2024-07-10
[DOC] Doc for BasicObject

BurdetteLamar:


peterzhu2118:
I think there are some changes in your PR that's irrelevant, can you undo them?

BurdetteLamar:
@peterzhu2118, how now?  (Think I got it cleaned up.)

vm_eval.c
```diff
@@ -885,41 +885,47 @@ NORETURN(static void raise_method_missing(rb_execution_context_t *ec, int argc,
...
+ *
+ *    'foo'.nosuch
+ *    # Raises NoMethodError: undefined method `nosuch' for an instance of String
+ *
+ *  The method may be overridden to provide dynamic handling for the call;
+ *  the overriding method should call +super+ if it can't handle the call.
+ *
+ *  This class overrides +method_missing+:
+ *
+ *    require 'roman-numerals' # Use Ruby gem roman-numerals.
```
peterzhu2118, line 901: I'm not a fan of having external dependencies in the example. Maybe sometime simple like this is a good enough example of now `method_missing` works:

```ruby
class Greeter
  def method_missing(method_name, *args, &block)
    if method_name.to_s =~ /^greet_(.*)$/
      puts "Hello #{$1.capitalize}"
    else
      super
    end
  end
end

greeter = Greeter.new
greeter.greet_world # => Hello world
greeter.greet_ruby # => Hello ruby
greeter.foo # Raises NoMethodError: undefined method 'foo' for an instance of Greeter
```

vm_eval.c
```diff
@@ -2169,35 +2179,55 @@ specific_eval(int argc, const VALUE *argv, VALUE self, int singleton, int kw_spl
...
 /*
  *  call-seq:
- *     obj.instance_eval(string [, filename [, lineno]] )   -> obj
- *     obj.instance_eval {|obj| block }                     -> obj
+ *    instance_eval {|self| ... } -> object
+ *    instance_eval(string, filename = nil, lineno = nil) -> string
+ *
+ *  With a block given, passes +self+ to the block
+ *  and returns the block's return value;
+ *  this gives the block access to instance variables and private methods in +self+:
```
peterzhu2118, line 2187: I don't think "access to instance variables and private methods" is a completely accurate description. For example, you can also define methods for the object:

```ruby
o = Object.new
o.instance_eval do
  def hello = puts "hello world"
end
o.hello # => "hello world"
```

I think the original description is more accurate:

> Evaluates a string containing Ruby source code, or the given block, within the context of the receiver.

vm_eval.c
```diff
@@ -2214,20 +2244,29 @@ rb_obj_instance_eval(int argc, const VALUE *argv, VALUE self)
...
  *  call-seq:
- *     obj.instance_exec(arg...) {|var...| block }                       -> obj
+ *    instance_exec(*args) {|args| ... } -> object
  *
- *  Executes the given block within the context of the receiver
- *  (_obj_). In order to set the context, the variable +self+ is set
- *  to _obj_ while the code is executing, giving the code access to
- *  _obj_'s instance variables.  Arguments are passed as block parameters.
+ *  Executes the given block,
+ *  giving access to the instance variables and private methods in +self+:
```
peterzhu2118, line 2250: Again, you can also define methods and do other things just like if you were inside of the object:

```ruby
o = Object.new
o.instance_exec do
  def hello = puts "hello world"
end
o.hello
```

object.c
```diff
@@ -203,44 +203,20 @@ rb_eql(VALUE obj1, VALUE obj2)
...
- *  Numeric types, for example, perform type conversion across #==,
- *  but not across #eql?, so:
- *
- *     1 == 1.0     #=> true
- *     1.eql? 1.0   #=> false
- *--
- * \private
- *++
+ *    self == object -> true or false
+ *    self.equal?(object) -> true or false
```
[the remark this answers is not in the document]
BurdetteLamar: b = BasicObject.new
# => #<BasicObject:0x0000023c01e58e88>
b.eql?
(irb):2:in `<main>': undefined method `eql?' for an instance of BasicObject (NoMethodError)

object.c
```diff
@@ -203,44 +203,20 @@ rb_eql(VALUE obj1, VALUE obj2)
...
- *  object as <code>b</code>):
- *
- *    obj = "a"
- *    other = obj.dup
- *
- *    obj == other      #=> true
- *    obj.equal? other  #=> false
- *    obj.equal? obj    #=> true
- *
- *  The #eql? method returns <code>true</code> if +obj+ and +other+
```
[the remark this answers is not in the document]
BurdetteLamar: Not sure exactly what to say here and above.

vm_eval.c
```diff
@@ -2169,35 +2176,55 @@ specific_eval(int argc, const VALUE *argv, VALUE self, int singleton, int kw_spl
 
 /*
  *  call-seq:
- *     obj.instance_eval(string [, filename [, lineno]] )   -> obj
- *     obj.instance_eval {|obj| block }                     -> obj
+ *    instance_eval {|self| ... } -> object
```
peterzhu2118, line 2179: `self` is a keyword so it can't actually use it as a variable name:

```ruby
1.times { |self| }
# SyntaxError: syntax error, unexpected `self', expecting '|'
```

Can we go back to using `obj`?

vm_eval.c
```diff
@@ -2169,35 +2176,55 @@ specific_eval(int argc, const VALUE *argv, VALUE self, int singleton, int kw_spl
...
+ *    instance_eval(string, filename = nil, lineno = nil) -> string
+ *
+ *  With a block given, passes +self+ to the block
+ *  and returns the block's return value;
+ *  this gives the block full access to +self+:
```
peterzhu2118, line 2184: Can we use the term "context" here? It's a term used to describe executing as if inside of another object.

```suggestion
 *  this executes within the context of +self+:
```

object.c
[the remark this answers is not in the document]
BurdetteLamar: I don't get it.  If `BasicObject#eql?` is undefined, how does `#eql?` belong on the `BasicObject` page?

BurdetteLamar:
@peterzhu2118, would you be open to the idea of my restoring the previous write-ups of all the methods, and retainiing only the new write-up of the class itself?

peterzhu2118:
Sure, you can break this PR down into several smaller PRs.

In the future, it might be easier and more organized if you opened more PRs (perhaps in parallel with each other), each with a smaller set of changes. That way, each PR will have fewer comments and I can merge the ones that pass reviews.

BurdetteLamar:
> Sure, you can break this PR down into several smaller PRs.
> 
> In the future, it might be easier and more organized if you opened more PRs (perhaps in parallel with each other), each with a smaller set of changes. That way, each PR will have fewer comments and I can merge the ones that pass reviews.

Done.  No changes now except to the class doc itself.

object.c
```diff
@@ -4065,72 +4065,77 @@ rb_f_loop_size(VALUE self, VALUE args, VALUE eobj)
...
+ *        def method_missing(name, *args, &block)
+ *          return super unless DELEGATE.include? name
+ *          ::Kernel.send(name, *args, &block)
+ *        end
+ *        def respond_to_missing?(name, include_private = false)
+ *          DELEGATE.include?(name)
+ *        end
  *      end
- *    end
+ *      ```
```
peterzhu2118, line 4118: Can we add the newlines back between the methods? I think it improves readability.

```suggestion
 *      ```
 *      class MyObjectSystem < BasicObject
 *        DELEGATE = [:puts, :p]
 *
 *        def method_missing(name, *args, &block)
 *          return super unless DELEGATE.include? name
 *          ::Kernel.send(name, *args, &block)
 *        end
 *
 *        def respond_to_missing?(name, include_private = false)
 *          DELEGATE.include?(name)
 *        end
 *      end
 *      ```
 ```

File: object.c [hunk 9]
```diff
@@ -4065,72 +4065,77 @@ rb_f_loop_size(VALUE self, VALUE args, VALUE eobj)
...
- *  - #!: Returns the boolean negation of +self+: +true+ or +false+.
- *  - #!=: Returns whether +self+ and the given object are _not_ equal.
- *  - #==: Returns whether +self+ and the given object are equivalent.
- *  - #__id__: Returns the integer object identifier for +self+.
- *  - #__send__: Calls the method identified by the given symbol.
- *  - #equal?: Returns whether +self+ and the given object are the same object.
- *  - #instance_eval: Evaluates the given string or block in the context of +self+.
- *  - #instance_exec: Executes the given block in the context of +self+,
- *    passing the given arguments.
+ *  | \\Method                    | Effect                                                                          |
```
peterzhu2118, line 4124: Why are we switching from bullet points to tables? Are we planning to change all "What's Here" sections from bullet points to tables?

object.c
```diff
@@ -4065,72 +4065,77 @@ rb_f_loop_size(VALUE self, VALUE args, VALUE eobj)
...
+ *  | #==                         | Returns whether +self+ and the given object are equivalent.                     |
+ *  | #__id__                     | Returns the integer object identifier for +self+.                               |
+ *  | #__send__                   | Calls the method identified by the given symbol.                                |
+ *  | #equal?                     | Returns whether +self+ and the given object are the same object.                |
+ *  | #instance_eval              | Evaluates the given string or block in the context of +self+.                   |
+ *  | #instance_exec              | Executes the given block in the context of +self+, passing the given arguments. |
+ *  | #method_missing             | Called when +self+ is called with a method it does not define.                  |
+ *  | #singleton_method_added     | Called when a singleton method is added to +self+.                              |
+ *  | #singleton_method_removed   | Called when a singleton method is removed from +self+.                          |
+ *  | #singleton_method_undefined | Called when a singleton method is undefined in +self+.                          |
```
peterzhu2118, line 4138: I think these all need to change from RDoc's `+` to Markdown's <code>`</code>.

object.c
BurdetteLamar: Done.

object.c
BurdetteLamar: Done.

object.c
BurdetteLamar: The GFM table is, IMO, *much* more readable.  Whether all such should be changed is a matter for discussion.  (Plus several other issues around "What's Here" sections.)  Where's a good place to start such a discussion?

Issues:
- Should the section have a GFM instead of bullets?
- Should attributes be included (in addition to methods defined with `def`)?  Why or why not?
- Should the entire section be off-page (in a separate linked file)?  That would make it simple to do a GFM table in a separate `.md` file (instead of changing a whole RDoc section from markup to markdown).  If so, where would those files live?
- Should there be a discussion and example of an included or inherited method that may behave in an unusual way?  (I'm thinking of one that calls into a local method in the local class, which may make behavior differ from class to class.)
- Should constants be included?  I think, *emphatically*, No;  but if not, should the title be changed from "What's Here" to something about methods?

peterzhu2118: > Where's a good place to start such a discussion?

I don't think there's a formal place to start a discussion, but you can try bugs.ruby-lang.org if you want. We can have some discussions here for now.

> Should the section have a GFM instead of bullets?

As-is, I don't think the table is any more readable than the bullet points. Do you plan on adding additional columns to the table? If so, what do you plan on adding? We should be cautious in adding too much information in the "What's Here" section in order to not bloat it and turn it from a quick reference into method documentation.

> Should attributes be included (in addition to methods defined with def)? Why or why not?

I'm not sure what you mean by attributes? Do you mean constants or method parameters? 

> Should the entire section be off-page (in a separate linked file)? That would make it simple to do a GFM table in a separate .md file (instead of changing a whole RDoc section from markup to markdown). If so, where would those files live?

I don't think it should live in a separate file. Documentation that live in a separate file should either be shared information that is used in multiple places or information that is not critical. I don't think the "What's Here" section will ever be shared across multiple files. If we place the "What's Here" section in another file and link it, then it makes it harder to use because the user can no longer search on the page and it's another level of indirection the user needs to do to access the information.

> Should there be a discussion and example of an included or inherited method that may behave in an unusual way? (I'm thinking of one that calls into a local method in the local class, which may make behavior differ from class to class.)

IMO, the "What's Here" section should be simple. I use it to find a method that I think exists but am unsure. I quickly skim the description to find candidates that are promising. The more complicated it is, the less useful it is for me. If I need examples, I expect to find it in the method documentation.

> Should constants be included? I think, emphatically, No; but if not, should the title be changed from "What's Here" to something about methods?

We could consider adding a section for just constants, or just putting the constants at the top of the list. But IMO, most constants are supplementary to the methods, meaning that users use the method and use constants to help them use the methods. In this case, I don't think the constant should be referenced in the "What's Here" section and instead only be referenced in the method documentation.

BurdetteLamar: Mostly agree with you, @peterzhu2118.

- Exception:  as I said, I think the table is much more readable; maybe others will have opinions?
- By "attributes," I mean methods defined by `attr_*`.  These appear to the user as full-fledged methods, though some say that internally they're different from those defined by `def`.
- Separate file is not a good idea.
- Discussion outside of the table complicates things, and should be avoided.
- Constants should not be included.

peterzhu2118: > By "attributes," I mean methods defined by attr_*. These appear to the user as full-fledged methods, though some say that internally they're different from those defined by def.

Ah attribute accessors. Yes these should definitely be included since they (for the most part) behave identically to methods (and the user should treat it like a method anyways).

BurdetteLamar:
@peterzhu2118, what about merging this with the table in, and asking (at reddit, dev?) which Ruby users like?  I'd link to both old and new in the querying post.

object.c
nobu: > As-is, I don't think the table is any more readable than the bullet points.

I'm for @peterzhu2118.

> No plan to add columns.

If you just want to align columns, why not definition list?

BurdetteLamar:
Converting to draft (while I switch back from table to bullets in "What's Here").

BurdetteLamar:
Having reformatted the table, there's no need for this to be markdown (rather than the default rdoc), so changed that also.

BurdetteLamar:
@peterzhu2118, ready to merge?

peterzhu2118 (approved):

PR #11140, opened 2024-07-10
YJIT: increase context cache size to 1024 redux

maximecb:
Recreating this PR with only the cache size change

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -847,10 +847,11 @@ const CTX_CACHE_SIZE: usize = 1024;
 // Cache of the last contexts encoded
 // Empirically this saves a few percent of memory
 // We can experiment with varying the size of this cache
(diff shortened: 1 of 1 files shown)


XrXr:
If this is indeed https://github.com/rust-lang/rust/issues/53827
we can keep the old types and golf down the diff to just changing the allocation site:

```
let boxed_array: Box<[u8; 100_000_000]> = vec![0; 100_000_000].try_into().unwrap();
```

(It's using [array::TryFrom<Vec<T>>](https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3CVec%3CT%3E%3E-for-Box%3C%5BT;+N%5D%3E))

> This can be used with [vec!](https://doc.rust-lang.org/std/macro.vec.html) to create an array on the heap:

PR #11146, opened 2024-07-11
Give up `USE_GC_MALLOC_OBJ_INFO_DETAILS`

shyouhei:
@peterzhu2118 feel free to revive the functionality if possible, but the current situation that it's simply broken.  It doesn't even compile.


peterzhu2118 (approved):
I've never used this feature. I don't think anyone else does either? So I think it should be fine to remove.

PR #11147, opened 2024-07-11
Release GVL around system calls in dir.c

jeremyevans:
This attempts to address the system calls mentioned in Bug #20587, with the exception of `getattrlist` and `fgetattrlist` (which seem to be Mac OS specific).

Releasing the GVL for readdir and stat caused problems on Windows, because these system calls are emulated on Windows, so this patch does not release the GVL on Windows.  I think any GVL releasing for Windows would need to happen in `win32/win32.c`.

While working on this, I noticed that `getpwnam` is called without releasing the GVL in `process.c` and `ext/etc/etc.c`.  If it makes sense to release the GVL for the `getpwnam` call in `Dir.home`, I think it makes sense to handle `getpwnam`, but that can be addressed later. Also, if `getpwnam` should release the GVL, `getpwuid`/`getgrgid`/`getgrnam` and similar methods should probably also release it.

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -1264,7 +1264,7 @@ nogvl_fchdir(void *ptr)
 static void
 dir_fchdir(int fd)
 {
-    if (fchdir(fd) < 0)
+    if (IO_WITHOUT_GVL_INT(nogvl_fchdir, (void *)&fd) < 0)
         rb_sys_fail("fchdir");
 }
 
--- a/dir.c
+++ b/dir.c
@@ -1724,13 +1724,30 @@ do_stat(int fd, size_t baselen, const char *path, struct stat *pst, int flags, r
 }
 
 #if defined HAVE_LSTAT || defined lstat || USE_OPENDIR_AT
+#if !USE_OPENDIR_AT
+struct lstat_args {
+    const char *path;
+    struct stat *pst;
+};
+
+static void *
+nogvl_lstat(void *args)
+{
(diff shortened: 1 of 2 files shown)


dir.c
```diff
@@ -1653,11 +1673,11 @@ to_be_ignored(int e)
...
 #ifdef _WIN32
-#define STAT(p, s)	rb_w32_ustati128((p), (s))
-#undef lstat
-#define lstat(p, s)	rb_w32_ulstati128((p), (s))
+#define STAT(args)	(int)(VALUE)nogvl_stat((void *)&(args))
+#define LSTAT(args)	(int)(VALUE)nogvl_lstat((void *)&(args))
 #else
-#define STAT(p, s)	stat((p), (s))
+#define STAT(args)	IO_WITHOUT_GVL_INT(nogvl_stat, (void *)&(args))
+#define LSTAT(args)	IO_WITHOUT_GVL_INT(nogvl_lstat, (void *)&(args))
```
nobu, line 1680: These `(void *)` casts are needed?

nobu (approved):

dir.c
jeremyevans: After some research, I do not think so.  Implicit casts to and from `void *` are allowed, unless they are part of a function pointer.  I found other unnecessary casts to and from `void *` in `dir.c`, and tried to fix those as well (let's see if CI likes it).

PR #11152, opened 2024-07-11
Fix Makefile.in for non GNUmake

peterzhu2118:
On OpenBSD the Makefile errors with:

    Using $< in a non-suffix rule context is a GNUmake idiom


jeremyevans (approved):
I tested this on OpenBSD and it does fix the issue.

PR #11156, opened 2024-07-12
Don't crash if madvise(MADV_FREE or MADV_DONTNEED) fails

KJTsanaktsidis:
The M:N threading stack cleanup machinery tries to call MADV_FREE on the native thread's stack, and calls rb_bug if it fails. Unfortunately, there's no way to distinguish between "You passed bad parameters to madvise" and "MADV_FREE is not supported on the kernel you are running on"; both cases just return EINVAL. This means that if you have a Ruby on a system that was built on a system with MADV_FREE and run it on a system without it, you get a crash in nt_free_stack.

I ran into this because rr actually emulates MADV_FREE by just returning EINVAL and pretending it's not supported (since it can otherwise introduce nondeterministic behaviour). So if you run bootstraptest/test_ractor.rb under rr, you get this crash.

I think we should just get rid of the error handling here; freeing memory like this is strictly optional anyway.

[Bug #20632]

### DIFF
--- a/thread_pthread_mn.c
+++ b/thread_pthread_mn.c
@@ -338,6 +338,30 @@ nt_alloc_stack(rb_vm_t *vm, void **vm_stack, void **machine_stack)
 }
 
 static void
+nt_madvise_free_or_dontneed(void *addr, size_t len)
+{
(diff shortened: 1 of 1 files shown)


ko1:
:+1:

PR #11157, opened 2024-07-12
YJIT: Local variable register allocation

k0kubun:
This PR introduces a register allocator for local variables. This PR contains the following changes for it:

* Migrate `Context::reg_temps` from `u8` to `[Option<RegTemp>; 5]`
    * `RegTemp` is either `RegTemp::Stack(u8)` or `RegTemp::Local(u8)`.
    * The encoder uses 1-5 bits for each of the 5 registers. 1 bit for `Option`, 1 bit for enum variant, and 3 bits for index.
* `Opnd::Stack` can now be assigned to any register while it still defaults to `stack_idx % 5` (register bias)
    * When the register at `stack_idx % 5` is already used, it uses an unused register with the smallest index.
* Allocate a register for a local variable on setivar
    * Convert local variable access from `Opnd::Mem` to `Opnd::Stack`
    * It defaults to the register at `4 - (local_idx % 5)`. If it's taken, it uses an unused register with the largest index.

## Out of Scope

Method arguments are local variables but they aren't set by `setivar`, so it's not supported in this PR. 

I'm thinking of working on that next, but I think the current optimization should be merged separately to keep the patches simple and to allow benchmarking them independently.

## Benchmark

It gives 1-2% speedup to some benchmarks. `railsbench` specifically seems to consistently show a 2% speedup.

```
before: ruby 3.4.0dev (2024-07-09T17:22:29Z master 6f6aff56b1) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-07-12T09:13:23Z yjit-local-regs 1ad043bd3e) +YJIT [x86_64-linux]

activerecord    302.0        0.2         302.2       0.4         1.00           1.00
chunky-png      487.5        0.1         486.1       0.2         0.99           1.00
erubi-rails     1275.5       0.1         1259.9      0.1         1.01           1.01
hexapdf         1876.7       2.5         1869.0      1.3         0.98           1.00
liquid-c        76.5         0.7         75.4        0.5         0.96           1.01
liquid-compile  75.7         0.5         74.7        0.9         0.90           1.01
liquid-render   98.8         0.3         97.8        0.5         0.96           1.01
lobsters        985.5        1.6         979.9       1.4         0.97           1.01
mail            155.9        0.4         155.6       0.5         0.95           1.00
psych-load      1803.7       0.1         1826.4      0.1         0.99           0.99
railsbench      2505.3       0.7         2457.8      0.8         0.99           1.02
rubocop         130.5        6.6         131.1       7.1         0.94           1.00
ruby-lsp        145.3        0.9         143.6       1.0         0.93           1.01
sequel          92.6         0.6         92.7        0.4         0.99           1.00
Legend:

Output:
./data/output_1343.csv
```

## Memory Usage

The current master always uses 8 bits for a `reg_temps`. This PR uses 5-25 bits. So, on average, `bytes_per_context` is increased. However, it doesn't seem to have a significant impact on overall `yjit_alloc_size`, thanks to context caching.

With 25 itrs on lobsters, 

### Before

```
yjit_alloc_size:          27,728,430
context_data_bytes:          341,320
context_cache_bytes:          16,384
num_contexts_encoded:        192,683
bytes_per_context:              1.77
context_cache_hit_rate:      141,224 (73.3%)
```

### After

```
yjit_alloc_size:          27,608,015
context_data_bytes:          390,787
context_cache_bytes:          20,480
num_contexts_encoded:        191,803
bytes_per_context:              2.04
context_cache_hit_rate:      136,528 (71.2%)
```

### DIFF
--- a/yjit/src/backend/ir.rs
+++ b/yjit/src/backend/ir.rs
@@ -6,7 +6,7 @@ use crate::codegen::{gen_counted_exit, gen_outlined_exit};
 use crate::cruby::{vm_stack_canary, SIZEOF_VALUE_I32, VALUE, VM_ENV_DATA_SIZE};
 use crate::virtualmem::CodePtr;
 use crate::asm::{CodeBlock, OutlinedCb};
-use crate::core::{Context, RegTemp, RegTemps, MAX_REG_TEMPS};
+use crate::core::{Context, RegMapping, RegMappings, MAX_REG_MAPPINGS};
 use crate::options::*;
 use crate::stats::*;
 
@@ -81,8 +81,8 @@ pub enum Opnd
         local_size: Option<u32>,
         /// ctx.sp_offset when this operand is made. Used with idx for Opnd::Mem.
         sp_offset: i8,
-        /// ctx.reg_temps when this operand is read. Used for register allocation.
-        reg_temps: Option<RegTemps>
+        /// ctx.reg_mappings when this operand is read. Used for register allocation.
+        reg_mappings: Option<RegMappings>
     },
 
     // Low-level operands, for lowering
@@ -174,7 +174,7 @@ impl Opnd
             Opnd::Reg(reg) => Some(Opnd::Reg(reg.with_num_bits(num_bits))),
             Opnd::Mem(Mem { base, disp, .. }) => Some(Opnd::Mem(Mem { base, disp, num_bits })),
             Opnd::InsnOut { idx, .. } => Some(Opnd::InsnOut { idx, num_bits }),
-            Opnd::Stack { idx, stack_size, local_size, sp_offset, reg_temps, .. } => Some(Opnd::Stack { idx, num_bits, stack_size, local_size, sp_offset, reg_temps }),
+            Opnd::Stack { idx, stack_size, local_size, sp_offset, reg_mappings, .. } => Some(Opnd::Stack { idx, num_bits, stack_size, local_size, sp_offset, reg_mappings }),
             _ => None,
         }
     }
@@ -229,23 +229,23 @@ impl Opnd
         Self::match_num_bits_iter(opnds.iter())
     }
 
-    /// Convert Opnd::Stack into RegTemp
-    pub fn reg_temp(&self) -> RegTemp {
-        self.get_reg_temp().unwrap()
+    /// Convert Opnd::Stack into RegMapping
+    pub fn reg_mapping(&self) -> RegMapping {
+        self.get_reg_mapping().unwrap()
     }
 
-    /// Convert an operand into RegTemp if it's Opnd::Stack
-    pub fn get_reg_temp(&self) -> Option<RegTemp> {
+    /// Convert an operand into RegMapping if it's Opnd::Stack
+    pub fn get_reg_mapping(&self) -> Option<RegMapping> {
         match *self {
             Opnd::Stack { idx, stack_size, local_size, .. } => Some(
                 if let Some(local_size) = local_size {
                     let last_idx = stack_size as i32 + VM_ENV_DATA_SIZE as i32 - 1;
                     assert!(last_idx <= idx, "Local index {} must be >= last local index {}", idx, last_idx);
                     assert!(idx <= last_idx + local_size as i32, "Local index {} must be < last local index {} + local size {}", idx, last_idx, local_size);
-                    RegTemp::Local((last_idx + local_size as i32 - idx) as u8)
+                    RegMapping::Local((last_idx + local_size as i32 - idx) as u8)
                 } else {
                     assert!(idx < stack_size as i32);
-                    RegTemp::Stack((stack_size as i32 - idx - 1) as u8)
+                    RegMapping::Stack((stack_size as i32 - idx - 1) as u8)
                 }
             ),
             _ => None,
@@ -974,7 +974,7 @@ pub struct SideExitContext {
     /// Context fields used by get_generic_ctx()
     pub stack_size: u8,
     pub sp_offset: i8,
-    pub reg_temps: RegTemps,
+    pub reg_mappings: RegMappings,
     pub is_return_landing: bool,
     pub is_deferred: bool,
 }
@@ -986,7 +986,7 @@ impl SideExitContext {
             pc,
             stack_size: ctx.get_stack_size(),
             sp_offset: ctx.get_sp_offset(),
-            reg_temps: ctx.get_reg_temps(),
+            reg_mappings: ctx.get_reg_mappings(),
             is_return_landing: ctx.is_return_landing(),
             is_deferred: ctx.is_deferred(),
         };
@@ -1002,7 +1002,7 @@ impl SideExitContext {
         let mut ctx = Context::default();
         ctx.set_stack_size(self.stack_size);
         ctx.set_sp_offset(self.sp_offset);
-        ctx.set_reg_temps(self.reg_temps);
+        ctx.set_reg_mappings(self.reg_mappings);
         if self.is_return_landing {
             ctx.set_as_return_landing();
         }
@@ -1110,8 +1110,8 @@ impl Assembler
                     assert!(idx < self.insns.len());
                     self.live_ranges[idx] = insn_idx;
                 }
-                // Set current ctx.reg_temps to Opnd::Stack.
-                Opnd::Stack { idx, num_bits, stack_size, local_size, sp_offset, reg_temps: None } => {
+                // Set current ctx.reg_mappings to Opnd::Stack.
+                Opnd::Stack { idx, num_bits, stack_size, local_size, sp_offset, reg_mappings: None } => {
                     assert_eq!(
                         self.ctx.get_stack_size() as i16 - self.ctx.get_sp_offset() as i16,
                         stack_size as i16 - sp_offset as i16,
@@ -1124,7 +1124,7 @@ impl Assembler
                         stack_size,
                         local_size,
                         sp_offset,
-                        reg_temps: Some(self.ctx.get_reg_temps()),
+                        reg_mappings: Some(self.ctx.get_reg_mappings()),
                     };
                 }
                 _ => {}
@@ -1196,8 +1196,8 @@ impl Assembler
         }
 
         match opnd {
-            Opnd::Stack { reg_temps, .. } => {
-                if let Some(reg_idx) = reg_temps.unwrap().get_reg(opnd.reg_temp()) {
+            Opnd::Stack { reg_mappings, .. } => {
+                if let Some(reg_idx) = reg_mappings.unwrap().get_reg(opnd.reg_mapping()) {
                     reg_opnd(opnd, reg_idx)
                 } else {
                     mem_opnd(opnd)
@@ -1208,11 +1208,11 @@ impl Assembler
     }
 
     /// Allocate a register to a stack temp if available.
-    pub fn alloc_temp_reg(&mut self, reg_temp: RegTemp) {
+    pub fn alloc_reg(&mut self, mapping: RegMapping) {
         // Allocate a register if there's no conflict.
-        let mut reg_temps = self.ctx.get_reg_temps();
-        if reg_temps.alloc_reg(reg_temp) {
-            self.set_reg_temps(reg_temps);
+        let mut reg_mappings = self.ctx.get_reg_mappings();
+        if reg_mappings.alloc_reg(mapping) {
+            self.set_reg_mappings(reg_mappings);
         }
     }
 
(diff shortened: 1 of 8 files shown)


yjit/src/backend/ir.rs
```diff
@@ -77,6 +77,8 @@ pub enum Opnd
         num_bits: u8,
         /// ctx.stack_size when this operand is made. Used with idx for Opnd::Reg.
         stack_size: u8,
+        /// The number of local variables in the current ISEQ. Used only for locals.
+        local_size: Option<u32>,
```
maximecb, line 81: Would `num_locals` be a better name here?

Also, why is it an option? Can it just be 0 when there are no locals?

yjit/src/core.rs
```diff
@@ -411,43 +411,101 @@ impl From<Opnd> for YARVOpnd {
...
 
-/// Bitmap of which stack temps are in a register
-#[derive(Copy, Clone, Default, Eq, Hash, PartialEq, Debug)]
-pub struct RegTemps(u8);
+/// A stack slot or a local variable. u8 represents the index of it (<= 8).
+#[derive(Copy, Clone, Eq, Hash, PartialEq, Debug)]
+pub enum RegTemp {
+    Stack(u8),
+    Local(u8),
+}
```
maximecb, line 425: I find the name `RegTemp` confusing here.

Usually temp means a temporary (like a value on the stack), so a local is not a temp.

Does `RegTemp` represent something a register can be mapped to? If so, can we rename it to `RegMapping` ?

yjit/src/core.rs
```diff
@@ -411,43 +411,101 @@ impl From<Opnd> for YARVOpnd {
...
+/// Each element of the array represents each of the registers.
+/// If an element is Some, the temporary value uses a register.
+#[derive(Copy, Clone, Default, Eq, Hash, PartialEq)]
+pub struct RegTemps([Option<RegTemp>; MAX_TEMP_REGS]);
 
 impl RegTemps {
-    pub fn get(&self, index: u8) -> bool {
-        assert!(index < MAX_REG_TEMPS);
-        (self.0 >> index) & 1 == 1
+    /// Return the index of the register for a given stack value if allocated.
```
maximecb, line 434: The comment says stack value, but RegTemp suggest it could also be a local?

yjit/src/core.rs
maximecb: Or looking at the code, I guess `RegTemp` is meant to represent an operand that is register-allocatable?

yjit/src/core.rs
```diff
@@ -961,9 +1019,24 @@ impl Context {
             bits.push_u8(self.sp_offset as u8);
         }
 
-        // Bitmap of which stack temps are in a register
-        let RegTemps(reg_temps) = self.reg_temps;
-        bits.push_u8(reg_temps);
+        // Which stack temps are in a register
```
maximecb, line 1022:
```suggestion
        // Which stack temps or locals are in a register
```

maximecb:
Hi Kokubun!

I took a first look through this PR just now. Props for getting the changes passing all the tests 🙏 

I think your encoding is pretty good in terms of compactness 👍 Happy that it doesn't increase YJIT alloc size significantly.

Can you report on the inline/outlined code size before and after?

I'm a bit surprised that we don't get more of a speedup on benchmarks. Do we get much of a speedup on any of the smaller benchmarks? How does protoboeuf look specifically? Also curious if you've inspected the generated code to see that it's not doing anything obviously suboptimal.

I think performance should not ultimately be the criteria by which we judge this change because it seems like we can probably make more iterative improvements to code quality once we have a better unified register allocator, but it would still be good to know how code size is affected and if there are obvious things that could be improved.

yjit/src/core.rs
k0kubun: My intention was that both temps and locals live on the stack so both of them are a "stack value". But I'll think of something else for clarity.

yjit/src/backend/ir.rs
k0kubun: > Would num_locals be a better name here?

Can you elaborate on why it's a better name? `local_size` is a variable name used throughout the CRuby codebase for this purpose. Since there are stack temps that have `stack_size`, it also seems consistent across YJIT to use `local_size` for locals.

> Also, why is it an option? Can it just be 0 when there are no locals?

At first, I thought there are `Assembler` usages where locals and `spill_temps()` are not used, so in that sense it could be optional and I wanted to distinguish ISEQs with no locals and the number not being set.

I ended up using it in outlined code like exits and stubs as well, so maybe I ended up using it everywhere. I'll see if I can always set it and remove `Option`.

yjit/src/core.rs
k0kubun: https://github.com/ruby/ruby/pull/11157/commits/775adcbb2a4ba6957040650d5165546f9a0eb625

yjit/src/core.rs
k0kubun: > Does RegTemp represent something a register can be mapped to? If so, can we rename it to RegMapping ?
>
> Or looking at the code, I guess RegTemp is meant to represent an operand that is register-allocatable?

I'd say `RegTemps` are a mapping as a whole (the former) but a `RegTemp` is an operand (the latter). I just renamed both of them to `RegMapping`(s) https://github.com/ruby/ruby/pull/11157/commits/a7155c2b0bdaefd87b5ebc0a7802f427ee7296ec, but maybe `RegMapping` (former `RegTemp`) could be `RegOperand`? 🤔

yjit/src/backend/ir.rs
maximecb: I guess I feel that `num_locals` is more descriptive given that it's the number of locals. `local_size` makes me think of a size in bytes or something like that. I wish we could go change the name in the rest of the Ruby codebase though I understand this is probably not possible. If you want to keep the same name for consistency I understand.

yjit/src/core.rs
maximecb: Maybe `RegOpnd` for consistency with other `Opnd` in the codebase 🤔 ?

k0kubun:
> Can you report on the inline/outlined code size before and after?

With 25 itrs on lobsters (the same run as the one in the PR description), the code size seems to be reduced.

### Before
```
inline_code_size:          7,561,968
outlined_code_size:        6,312,621
code_region_size:         15,466,496
code_region_overhead:      1,591,907 (10.3%)
```

### After
```
inline_code_size:          7,513,927
outlined_code_size:        6,423,682
code_region_size:         15,433,728
code_region_overhead:      1,496,119 ( 9.7%)
```

> Also curious if you've inspected the generated code to see that it's not doing anything obviously suboptimal.

I have. This is the generated code for `def test; a = 1; a; end; test`:

```asm
  # Block: test@-e:1
  # reg_mappings: [None, None, None, None, None]
  # Insn: 0000 putobject_INT2FIX_1_ (stack_size: 0)
  # reg_mappings: [None, None, None, None, None] -> [Some(Stack(0)), None, None, None, None]
  0x5f84e6dfd058: mov esi, 3
  # Insn: 0001 setlocal_WC_0 (stack_size: 1)
  # reg_mappings: [Some(Stack(0)), None, None, None, None] -> [Some(Stack(0)), None, None, None, Some(Local(0))]
  0x5f84e6dfd05d: mov r10, rsi
  # Insn: 0003 getlocal_WC_0 (stack_size: 0)
  # reg_mappings: [None, None, None, None, Some(Local(0))] -> [Some(Stack(0)), None, None, None, Some(Local(0))]
  0x5f84e6dfd060: mov rsi, r10
```

which is the intended outcome as of this PR, and for `def test; a = 1; puts; a; end; test` with a spill on a C call:

```asm
  # Block: test@-e:1
  # reg_mappings: [None, None, None, None, None]
  # Insn: 0000 putobject_INT2FIX_1_ (stack_size: 0)
  # reg_mappings: [None, None, None, None, None] -> [Some(Stack(0)), None, None, None, None]
  0x5f28bbb48058: mov esi, 3
  # Insn: 0001 setlocal_WC_0 (stack_size: 1)
  # reg_mappings: [Some(Stack(0)), None, None, None, None] -> [Some(Stack(0)), None, None, None, Some(Local(0))]
  0x5f28bbb4805d: mov r10, rsi
  # Insn: 0003 putself (stack_size: 0)
  # reg_mappings: [None, None, None, None, Some(Local(0))] -> [Some(Stack(0)), None, None, None, Some(Local(0))]
  0x5f28bbb48060: mov rsi, qword ptr [r13 + 0x18]

  # Insn: 0004 opt_send_without_block (stack_size: 1)
  # call to Object#puts
  ...
  # call C function
  # spill_temps: [Some(Stack(0)), None, None, None, Some(Local(0))] -> [None, None, None, None, None]
  0x5f28bbb480f9: mov qword ptr [rbx], rsi
  0x5f28bbb480fc: mov qword ptr [rbx - 0x20], r10
  ...

  # reg_mappings: [Some(Stack(0)), None, None, None, None]
  # Insn: 0006 pop (stack_size: 1)
  # Insn: 0007 getlocal_WC_0 (stack_size: 0)
  # reg_mappings: [None, None, None, None, None] -> [Some(Stack(0)), None, None, None, None]
  0x5f28bbb4811d: mov rsi, qword ptr [rbx - 0x20]
```

which is also what I intended.

> Do we get much of a speedup on any of the smaller benchmarks? How does protoboeuf look specifically?

Not much on smaller benchmarks either. It doesn't seem to speed up protoboeuf 🤔 

```
before: ruby 3.4.0dev (2024-07-09T17:22:29Z master 6f6aff56b1) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-07-12T09:13:23Z yjit-local-regs 1ad043bd3e) +YJIT [x86_64-linux]

protoboeuf  54.2         0.8         54.4        0.9         0.94           1.00
```

I haven't looked into the generated code of protoboeuf, so I can take a look at that and see if anything can be easily improved. Looking at the source code, `buff` and `index` are not locals set by `setlocal` but method arguments, which are not yet optimized in this PR, so it might partly explain why it doesn't have a significant impact.

yjit/src/core.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/11157/commits/35119a841462aede6748565bbb2d33469f20cccb

yjit/src/backend/ir.rs
k0kubun: The same argument applies to `stack_size`. Does `stack_size` also seem like the number of bytes to you?

Anyway, I agree `num_locals` is more descriptive about what it is, and we seem to already call it `num_locals` in a few places in YJIT, so I just renamed `local_size` to `num_locals`. https://github.com/ruby/ruby/pull/11157/commits/ba1f92609e34a5f1f21d96e708c2c2d607ac6600

yjit/src/core.rs
k0kubun: Now that we renamed `RegTemp` to `RegOpnd`, I just called it "operand" https://github.com/ruby/ruby/pull/11157/commits/e4fe66dc1f9df5b79636edb4400be56c05b7954e

maximecb:
It's nice to see a reduction in inline code size even if it's small. The reg mapping is a lot more readable in the disassembly output which is also nice.

> method arguments, which are not yet optimized in this PR, so it might partly explain why it doesn't have a significant impact.

Supporting method arguments would probably have a big impact as they are the most used locals. Would it be difficult to add to this design? Not asking you to do it in this PR.

I won't push the performance question more for this PR. I think we need to do more in terms of improving the quality of generated code to maximize performance, but I assume we can do it as part of followup PRs.

yjit/src/backend/ir.rs
```diff
@@ -1031,6 +1031,10 @@ pub struct Assembler {
     /// Context for generating the current insn
     pub ctx: Context,
 
+    /// The current ISEQ's local table size. asm.local_opnd() uses this, and it's
+    /// sometimes hard to pass this value, e.g. asm.spill_temps() in asm.ccall().
+    pub num_locals: Option<u32>,
```
maximecb, line 1036: Can you document in the comment in which cases this would be `None` ?

yjit/src/core.rs
```diff
@@ -411,43 +411,101 @@ impl From<Opnd> for YARVOpnd {
...
-/// Maximum index of stack temps that could be in a register
-pub const MAX_REG_TEMPS: u8 = 8;
+/// Number of registers that can be used for stack temps or locals
+pub const MAX_MAPPED_REGS: usize = 5;
 
-/// Bitmap of which stack temps are in a register
-#[derive(Copy, Clone, Default, Eq, Hash, PartialEq, Debug)]
-pub struct RegTemps(u8);
+/// Maximum index of stack temps or locals that could be in a register
+pub const MAX_REG_OPNDS: u8 = 8;
```
maximecb, line 418: Don't hate me but I find this constant name a bit confusing. The names gives the impression that there is a limit on the number of registers or operands that could be passed into registers somewhere.

However, this constant has the same value and is used in the same way as `MAX_TEMP_TYPES` and `MAX_LOCAL_TYPES`. So maybe we should just rename those constants `MAX_CTX_TEMPS` and `MAX_CTX_LOCALS` and use those constants instead?

yjit/src/backend/ir.rs
```diff
@@ -227,28 +229,26 @@ impl Opnd
...
-
-    /// Get the index for stack temp registers.
-    pub fn reg_idx(&self) -> usize {
-        match self {
-            Opnd::Stack { .. } => {
-                self.stack_idx() as usize % get_option!(num_temp_regs)
-            },
-            _ => unreachable!(),
+    /// Convert an operand into RegMapping if it's Opnd::Stack
+    pub fn get_reg_mapping(&self) -> Option<RegOpnd> {
```
maximecb, line 238: I think this method should be renamed `get_reg_opnd` to avoid confusion with `RegMapping`.

maximecb:
The PR seems good from a technical standpoint, almost ready to merge.

I wrote a few more comments on the PR. They are all relatively trivial things with respect to naming, to help make the code easier to understand.

XrXr (commented):
On first impression, the patch looks pretty straight forward. Congrats on scoping it right and getting it done!

> I'm a bit surprised that we don't get more of a speedup on benchmarks. 

I'm pretty happy with the boost. I assume when we get to hold some values in registers across method calls (explicitly out of scope here) we'd see a bigger boost. I'm of the impression that people love to keep their ruby methods short have their logic compartmentalized across methods.

As for protoboeuf, I think there were some tweaks in there to specifically to avoid emitting `setlocals`, so it makes sense that it doesn't benefit as much from this patch.

yjit/src/backend/ir.rs
k0kubun: After https://github.com/ruby/ruby/pull/11157#discussion_r1676236891, I checked if we can remove `Option`, but there were too many places and it didn't make sense to me to just use `0` for them as a compromise.

I prepared `Assembler::new_without_iseq` separately from `Assembler::new` for clarity https://github.com/ruby/ruby/pull/11157/commits/11ac40a3df6d044a4a06912bd147d7a03673a91d. The commit also added a comment https://github.com/ruby/ruby/pull/11157#discussion_r1676386533 that explains when it's `None`.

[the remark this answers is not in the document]
k0kubun: Thanks, that makes sense. I merged your patch and made a few more clarifying improvements https://github.com/ruby/ruby/pull/11157/commits/11ac40a3df6d044a4a06912bd147d7a03673a91d.

yjit/src/core.rs
```diff
@@ -411,43 +411,101 @@ impl From<Opnd> for YARVOpnd {
...
+pub enum RegOpnd {
+    Stack(u8),
+    Local(u8),
+}
+
+/// RegMappings manages a set of registers used for temporary values on the stack.
+/// Each element of the array represents each of the registers.
+/// If an element is Some, the temporary value uses a register.
+#[derive(Copy, Clone, Default, Eq, Hash, PartialEq)]
+pub struct RegMapping([Option<RegOpnd>; MAX_MAPPED_REGS]);
```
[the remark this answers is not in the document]
k0kubun: Correct. Mentioned it in the comment https://github.com/ruby/ruby/pull/11157/commits/170f3cad11b47d66f9fdc8b2ab5f82a9de9e13c4.

yjit/src/backend/ir.rs
k0kubun: Thanks, that was an oversight. https://github.com/ruby/ruby/pull/11157/commits/e52cca0d3398da1d2200ee7a589f2bb44f2c7b10

yjit/src/core.rs
k0kubun: Makes sense. Renamed `MAX_TEMP_TYPES`/`MAX_LOCAL_TYPES` to `MAX_CTX_TEMPS`/`MAX_CTX_LOCALS` and used them https://github.com/ruby/ruby/pull/11157/commits/6bae029f9b8eea1b439ef03a7c08874dbac9d6de.

maximecb (approved):
Well done Kokubun!

PR #11161, opened 2024-07-12
Multiple array and hash allocation reduction optimizations

jeremyevans:
Please see individual commits for more details. In brief, this has four separate optimizations (of varying complexity), and a couple of commits to remove unneeded code.

1. Avoid unnecessary array allocations for f(arg, *arg, **arg, **arg), f(*arg, a: lvar), and other calls

   This modifies the compiler to handle all `splatarray true` -> `splatarray false` optimizations, handling numerous cases not previously handled, such as:

   * Calls with positional arguments, positional splats, and multiple keyword splats or literal keywords + keyword splats
   * Calls with positional splats and non-static literal keywords
   * Calls with positional splats and keyword splatted constants
   * Calls with positional splats and passing constants as blocks

   This requires an API change to `setup_args_core` (a static function) to make `dup_rest` a pointer to int instead of an int, to allow to tracking `dup_rest` state across recursive calls.  

2. Remove splatarray true -> splatarray false peephole optimization

   Now that the compiler handles every splatarray optimization previously handled by the peephole optimizer (and implements additional splatarray optimizations), this removes the related code from the peephole optimizer.

3. Avoid array allocation for f(*empty_ary, **hash) when def f(x)

   Ruby previously allocated in this case to push the keyword splat onto the array before flattening the array to the VM stack.  Instead, flatten the array to the VM stack, and push the keyword splat to the VM stack.

4. Check for and remove duplicate checks in test_allocation

   I noticed I had many duplicate checks in test_allocation.  I added code to detect duplicate checks, and treat them as errors.  Then I fixed all issues.

5. Avoid hash allocation for f(*r2k_ary) when def f(kw: 1)

   This is a simple one line change to not duplicate the ruby2_keywords flagged hash, just use it directly to set the keyword values.

6. Avoid array allocation for f(*r2k_ary) when def f(x)

   This takes a similar approach to commit 3, flattening the ruby2_keywords array to the VM stack and adjusting the VM stack instead of duping the array, modifying it, and then flattening it.


ko1 (approved):
One concern is this approach highly depends NODE and it seems hard to port to prism compiler

PR #11163, opened 2024-07-14
[Prism] Remove redundant return flag

nobu:
Apply https://bugs.ruby-lang.org/issues/20457 to Prism as well.


kddnewton:
This PR does 2 things — makes it so that redundant returns can have line events, and also changes the way they are compiled.

Right now, when you have:

```ruby
def foo
  return 1
end
```

you get:

```
== disasm: #<ISeq:foo@test.rb:1 (1,0)-(3,3)>
0000 putobject_INT2FIX_1_                                             (   2)[LiCa]
0001 leave                                                            (   3)[Re]
```

but with this PR you get:

```
== disasm: #<ISeq:foo@test.rb:1 (1,0)-(3,3)>
0000 putobject_INT2FIX_1_                                             (   2)[LiCa]
0001 leave                                  [Re]
0002 putnil
0003 leave                                                            (   3)[Re]
```

Should we keep that part of the functionality and just do the line tracepoint part? Or should we do the whole thing?

nobu:
It think that the correct line number is 2, the line where `return` exists.
OTOH, Dead code elimination is a part of optimization, and optional.
The reason originally `reduce_nodes` in the parser did that DCE was there was no separated optimization phase before 1.9.

kddnewton:
Okay! I think this is good to merge then.

PR #11169, opened 2024-07-15
YJIT: split chain_depth and flag booleans in context

maximecb:
Split these values to avoid using a bit mask in the context Use variable length encoding to save a few bits on chain depth.

On lobsters:
```
Before:
  context_data_bytes:          254,269
  context_cache_bytes:          40,960
  num_contexts_encoded:        136,477
  bytes_per_context:              1.86
After:
  context_data_bytes:          241,214
  context_cache_bytes:          49,152
  num_contexts_encoded:        136,505
  bytes_per_context:              1.77
```

Not a big saving, especially since refactoring the context increases the context cache size. However, we should see a bigger saving on large production apps, and every little `bit` counts 🌈 . The main idea is that we no longer have to have bit masks and encodings that are hard to read inside the context itself.

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -510,22 +510,13 @@ impl fmt::Debug for RegMapping {
     }
 }
 
-/// Bits for chain_depth_return_landing_defer
-const RETURN_LANDING_BIT: u8 = 0b10000000;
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
nice :+1:

PR #11172, opened 2024-07-16
Bump up REXML-3.3.2

hsbt:
https://www.ruby-lang.org/en/news/2024/07/16/dos-rexml-cve-2024-39908/

### DIFF
--- a/.github/workflows/compilers.yml
+++ b/.github/workflows/compilers.yml
@@ -66,9 +66,9 @@ jobs:
           - { key: default_cc, name: gcc-9,     value: gcc-9,     container: gcc-9 }
           - { key: default_cc, name: gcc-8,     value: gcc-8,     container: gcc-8 }
           - { key: default_cc, name: gcc-7,     value: gcc-7,     container: gcc-7 }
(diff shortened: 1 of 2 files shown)


hsbt:
@kou @Watson1978 The following test of REXML-3.3.2 is failing with Ruby 3.1.x. 

```
Failure: test_linear_performance_attribute_value_gt(REXMLTests::TestParseElement):
    [150000]: in 0.015s.
    <0.04155644600001551> was expected to be
    <=
    <0.015>.
```

Do you have an idea to fix that?

kou:
Hmm. It's not happen on ruby/rexml: https://github.com/ruby/rexml/actions/runs/9949888216/job/27486869388

I re-run these jobs on this PR for now.
We may need to relax the expected range.

hsbt:
Thanks, I also re-run another failure: https://github.com/ruby/ruby/actions/runs/9950962925/job/27500262330?pr=11172

hsbt:
@kou I merged REXML-3.3.2 for Ruby 3.1.x. I will suggest to relax this condition if I encounter same failure in the next time. Thanks!

PR #11181, opened 2024-07-17
Do not retry running tests in `make test-all`

ono-max:


mame:
Instead of adding `--no-retry` in the CI configuration, why not default the testing framework to no retry?
You can leave the `--retry` option to retry when added, but I personally don't think we need the `--retry` option either.

PR #11185, opened 2024-07-17
Lanunchable: Add missing condition statement

ono-max:
Addresses https://github.com/ruby/ruby/pull/11183/files#r1680617485.


ono-max:
@nobu 

Can you take a look at this PR? Thanks!

nobu (approved):

PR #11190, opened 2024-07-17
Fix empty hash instruction

eileencodes:
When we have an empty hash the iseq should have a `newhash` but instead had a `duphash`. To fix, return `Qnil` if the `hash` elements are empty, use `newhash` in that case, otherwise `duphash`.

Before:

```
== disasm: #<ISeq:<main>@test2.rb:1 (1,0)-(1,2)>
0000 duphash                                {}                        (   1)[Li]
0002 leave
```

After:

```
== disasm: #<ISeq:<main>@test2.rb:1 (1,0)-(1,2)>
0000 newhash                                0                         (   1)[Li]
0002 leave
```

Fixes the test `TestYJIT#test_compile_newhash`. Related to ruby/prism#2935

cc/ @kddnewton

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -7164,9 +7164,16 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
             // is popped, then we know we don't need to do anything since it's
             // statically known.
             if (!popped) {
-                VALUE value = pm_static_literal_value(iseq, node, scope_node);
-                PUSH_INSN1(ret, location, duphash, value);
-                RB_OBJ_WRITTEN(iseq, Qundef, value);
+                const pm_hash_node_t *cast = (const pm_hash_node_t *) node;
+
+                if (cast->elements.size == 0) {
+                    PUSH_INSN1(ret, location, newhash, INT2FIX(0));
+                }
+                else {
+                    VALUE value = pm_static_literal_value(iseq, node, scope_node);
+                    PUSH_INSN1(ret, location, duphash, value);
+                    RB_OBJ_WRITTEN(iseq, Qundef, value);
+                }
             }
         }
         else {


prism_compile.c
```diff
@@ -755,6 +755,11 @@ pm_static_literal_value(rb_iseq_t *iseq, const pm_node_t *node, const pm_scope_n
       case PM_HASH_NODE: {
         const pm_hash_node_t *cast = (const pm_hash_node_t *) node;
         const pm_node_list_t *elements = &cast->elements;
+        // If there are 0 elements, we have an empty array, so we want to
+        // return nil as the value in order to use the newhash instruction.
+        if (elements->size == 0) {
+            return Qnil;
+        }
```
eileencodes, line 762: I'm definitely not feeling confident this is the right way to handle this. But if we have no elements there's no reason to go through the rest, we know it's an empty hash and should use `newhash` instead of `duphash`.

kddnewton: Yeah, I'd be nervous about this, because it's not really returning the value that it represents. Instead, can you check the number of elements where we compile `PM_HASH_NODE` and call into this function?

eileencodes: Ah ok, yea I can do that!

prism_compile.c
```diff
@@ -7164,9 +7164,16 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
             // is popped, then we know we don't need to do anything since it's
             // statically known.
             if (!popped) {
-                VALUE value = pm_static_literal_value(iseq, node, scope_node);
-                PUSH_INSN1(ret, location, duphash, value);
-                RB_OBJ_WRITTEN(iseq, Qundef, value);
+                const pm_hash_node_t *cast = (const pm_hash_node_t *) node;
+
+                if (&cast->elements.size == 0) {
```
kddnewton, line 7169: I think this is almost it, but this is checking if the address for the `size` member is 0, which I think effectively is checking if it's null. I think instead we want:

```suggestion
                if (cast->elements.size == 0) {
```

kddnewton (approved):

PR #11191, opened 2024-07-17
[PRISM] Fix interpolated sybmol node instructions

eileencodes:
If the symbol node is interpolated like this `:"#{foo}"` the instruction sequence should be `putstring` followed by `intern`. In this case it was a `putobject` causing the `test_yjit` tests to fail. Note that yjit is not required to reproduce - the instructions are `putstring` and `intern` for yjit and non-yjit with the original parser.

Before:

```
== disasm: #<ISeq:<main>@test2.rb:1 (1,0)-(1,11)>
0000 putobject                              :foo                      (   1)[Li]
0002 leave
```

After:

```
== disasm: #<ISeq:<main>@test2.rb:1 (1,0)-(1,11)>
0000 putstring                              "foo"                     (   1)[Li]
0002 intern
0003 leave
```

Fixes the test `TestYJIT#test_compile_dynamic_symbol`. Related to ruby/prism#2935

cc/ @kddnewton

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -7445,25 +7445,17 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         // :"foo #{bar}"
         // ^^^^^^^^^^^^^
         const pm_interpolated_symbol_node_t *cast = (const pm_interpolated_symbol_node_t *) node;
+        int length = pm_interpolated_node_compile(iseq, &cast->parts, &location, ret, popped, scope_node, NULL, NULL);
 
-        if (PM_NODE_FLAG_P(node, PM_NODE_FLAG_STATIC_LITERAL)) {
-            if (!popped) {
-                VALUE symbol = pm_static_literal_value(iseq, node, scope_node);
-                PUSH_INSN1(ret, location, putobject, symbol);
-            }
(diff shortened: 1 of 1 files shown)


prism_compile.c
```diff
@@ -7442,7 +7442,15 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret,
         if (PM_NODE_FLAG_P(node, PM_NODE_FLAG_STATIC_LITERAL)) {
             if (!popped) {
                 VALUE symbol = pm_static_literal_value(iseq, node, scope_node);
-                PUSH_INSN1(ret, location, putobject, symbol);
+
+                if (PM_NODE_FLAG_P(node, PM_INTERPOLATED_STRING_NODE)) {
```
kddnewton, line 7446: This one I'm not so sure about. `PM_NODE_FLAG_P` boils down to:

```c
node->flags & PM_INTERPOLATED_STRING_NODE
```

`PM_INTERPOLATED_STRING_NODE` is part of the node type enum. It happens to be 85, which is `1010101`. So I think this is effectively checking if the newline flag (the first common flag) is set. 

All of that is to say, I don't think this is quite the solution we're going for.

kddnewton (approved):

PR #11198, opened 2024-07-18
YJIT: Allow dev_nodebug to disasm release-mode code

k0kubun:
* Add `runtime_assertions` feature to the cargo config
    * To check the quality of the generated code that would run in the release mode, I sometimes want to see the disasm of generated code without runtime assertions like "set stack canary".
    * Having this feature (separately from `cfg!(debug_assertions)`) allows `--enable-yjit=dev_nodebug` to disasm code without runtime assertions while enabling other compile-time assertions.
* Drop `stats` feature from cargo
    * This is no longer used. All stats implemented in YJIT's code are available in the release mode as well. I still left `--enable-yjit=stats` for the `YJIT_STATS` macro in C code.

### DIFF
--- a/yjit/Cargo.toml
+++ b/yjit/Cargo.toml
@@ -22,10 +22,11 @@ capstone = { version = "0.12.0", optional = true }
 [features]
 # Support --yjit-dump-disasm and RubyVM::YJIT.disasm using libcapstone.
 disasm = ["capstone"]
-# Modify generated code for assertions. This is managed separately
+# Modify generated code for assertions, e.g. and poison value in PC
+# for C method calls and stack canary. This is managed separately
 # from cfg!(debug_assertions) so that we can see disasm of the code
 # that would run in the release mode.
-runtime_assertions = []
+runtime_checks = []
 
 [profile.dev]
 opt-level = 0
--- a/yjit/Cargo.toml
+++ b/yjit/Cargo.toml
@@ -17,11 +17,15 @@ crate-type = ["staticlib"]
 # written rationale. Optional For development and testing purposes
 capstone = { version = "0.12.0", optional = true }
 
-[features]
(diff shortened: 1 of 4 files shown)


XrXr (commented):
This is nice. There seems to be a few codegen decisions still made with `debug_assertions` that probably should be changed, though. `git grep debug_assertions`

XrXr:
I'm gonna push a commit onto this PR that removes one of the places that codegens using `debug_assertions` still, hopefully you don't mind.

k0kubun:
Yes please!

yjit/Cargo.toml
```diff
@@ -17,11 +17,15 @@ crate-type = ["staticlib"]
...
 # NOTE: Development builds select a set of these via configure.ac
 # For debugging, `make V=1` shows exact cargo invocation.
+[features]
+# Support --yjit-dump-disasm and RubyVM::YJIT.disasm using libcapstone.
 disasm = ["capstone"]
-stats = []
+# Modify generated code for assertions. This is managed separately
+# from cfg!(debug_assertions) so that we can see disasm of the code
+# that would run in the release mode.
+runtime_assertions = []
```
maximecb, line 28: Just to clarify, `runtime_assertions` here would be things like the stack canary check? Anything else? Can you mention those specifically in the comment.

Also, could we rename `runtime_assertions` to `runtime_checks`, just to make it clear that this has nothing to do with `assert!` in C/Rust code?

k0kubun: Right now it's only in two places: PC poison value for C method calls and stack canary. I mentioned them in the comment and renamed it to `runtime_checks` https://github.com/ruby/ruby/pull/11198/commits/b9529a187cdbf0e5b19b38a82849e1baf4b9a28a.

maximecb (approved):

PR #11205, opened 2024-07-19
Launchable: Aggregate test results based on file level

ono-max:
In https://github.com/ruby/ruby/pull/11113, we started to send test results of bootstraptest to Launchable. Now, we send test results based on id level, but it's hard to identify them. So, aggregating test results based on file level will be better. This PR addresses it.

https://app.launchableinc.com/organizations/ruby/workspaces/ruby-make-btest/insights/unhealthy-tests

### DIFF
--- a/bootstraptest/runner.rb
+++ b/bootstraptest/runner.rb
@@ -364,7 +364,7 @@ end
 # In bootstraptest, we aggregate the test results based on file level.
 module Launchable
   @@last_test_name = nil
-  @@stderr = ''
+  @@failure_log = ''
   @@duration = 0
 
   def show_progress(message = '')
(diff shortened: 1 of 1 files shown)


bootstraptest/runner.rb
```diff
@@ -359,36 +359,50 @@ def concurrent_exec_test
   end
 end
 
+##
+# Module for writing a test file for uploading test results into Launchable.
+# In bootstraptest, we aggregate the test results based on file level.
 module Launchable
+  @@last_test_name = nil
+  @@stderr = ''
```
ko1, line 367: usually `stderr` means `$stderr` (IO).

ono-max: [ddc3eef](https://github.com/ruby/ruby/pull/11205/commits/ddc3eef78cda402a3fc6b3e9f02f6b6c21736b56)

PR #11208, opened 2024-07-19
bundled_gems.rb: Add a fast path

casperisfine:
[[Bug #20641]](https://bugs.ruby-lang.org/issues/20641)

`Gem::BUNDLED_GEMS.warning?` adds a lot of extra work on top of `require`. When the call end up atually loading code the overhead is somewhat marginal.

However it's not uncommon for code to go some late `require` in some paths, so it's expected that calling `require` with something already required is somewhat fast, and `bundled_gems.rb` breaks this assumption.

To avoid this, we can have a fast path that in most case allow to short-circuit all the heavy computations. If we extract the feature basename and it doesn't match any of the bundled gems we care about we can return very early.

With this change `require 'date'` is now only 1.33x slower on Ruby 3.3.3, than it was on Ruby 3.2.2, whereas before this change it was at least 100x slower.

cc @hsbt

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -34,6 +34,8 @@ module Gem::BUNDLED_GEMS
     "logger" => "3.5.0",
   }.freeze
 
+  SINCE_FAST_PATH = SINCE.transform_keys { |g| g.sub(/\A.*\-/, "") }.freeze
+
   EXACT = {
     "kconv" => "nkf",
   }.freeze
@@ -92,6 +94,13 @@ module Gem::BUNDLED_GEMS
   def self.warning?(name, specs: nil)
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
+
+    if feature.include?("/")
+      return unless SINCE_FAST_PATH[File.basename(feature, ".*")]
+    else
+      return unless SINCE_FAST_PATH[feature]
+    end
+
     # bootsnap expands `require "csv"` to `require "#{LIBDIR}/csv.rb"`,
     # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
     name = feature.delete_prefix(ARCHDIR)
--- a/test/test_bundled_gems.rb
+++ b/test/test_bundled_gems.rb
@@ -1,0 +1,35 @@
+require_relative "rubygems/helper"
+require "rubygems"
+require "bundled_gems"
+
+class TestBundlerGem < Gem::TestCase
+  def setup
+    Gem::BUNDLED_GEMS::WARNED.clear
+  end
+
+  def teardown
+    Gem::BUNDLED_GEMS::WARNED.clear
+  end
+
+  def test_warning
+    assert Gem::BUNDLED_GEMS.warning?("rss", specs: {})
+    assert_nil Gem::BUNDLED_GEMS.warning?("rss", specs: {})
+  end
+
+  def test_no_warning_warning
+    assert_nil Gem::BUNDLED_GEMS.warning?("some_gem", specs: {})
+    assert_nil Gem::BUNDLED_GEMS.warning?("/path/to/some_gem.rb", specs: {})
+  end
+
+  def test_warning_libdir
+    path = File.join(::RbConfig::CONFIG.fetch("rubylibdir"), "rss.rb")
+    assert Gem::BUNDLED_GEMS.warning?(path, specs: {})
+    assert_nil Gem::BUNDLED_GEMS.warning?(path, specs: {})
+  end
+
+  def test_warning_archdir
+    path = File.join(::RbConfig::CONFIG.fetch("rubyarchdir"), "syslog.so")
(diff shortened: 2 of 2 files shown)


casperisfine:
Thanks, but your patch does break the functionality if bootsnap is involved.

deivid-rodriguez:
Since `lib/bundled_gems.rb` is not vendored, I think it's best to put the test directly at `test/test_bundled_gems.rb` to make sure it doesn't interfere with synchronizing with `rubygems/rubygems` repo.

I'm not sure it will cause any issues but I think it makes more sense that the whole `test/rubygems` directory is vendored from rubygems/rubygems.

Thanks for fixing this by the way!

casperisfine:
@deivid-rodriguez done!

jd-stripe:
I applied both of our patches against 3.3.3 (slightly changed your patch so it applies cleanly) and tested how long some services take to boot within Stripe. They improve performance vs 3.3.3 by almost the exact same amount.

Thanks for putting up this change and next time I'll raise these issues sooner

lib/bundled_gems.rb
```diff
@@ -92,6 +94,13 @@ def self.find_gem(path)
   def self.warning?(name, specs: nil)
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
+
+    if feature.include?("/")
```
hsbt, line 98: Is it for full path by bootsnap?

casperisfine: Yes, but even without bootsnap if you are requiring `some/thing`.

We could just always check `SINCE_FAST_PATH[File.basename(feature).sub(/(\.[^\.]+)\z/, "")]`, but this `include?` allow to skip the `File.basename` and `.sub`.

casperisfine: I realized we could do:

```diff
diff --git a/lib/bundled_gems.rb b/lib/bundled_gems.rb
index f8f2ef5b27..245ff9b036 100644
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -95,11 +95,7 @@ def self.warning?(name, specs: nil)
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
 
-    if feature.include?("/")
-      return unless SINCE_FAST_PATH[File.basename(feature).sub(/(\.[^\.]+)\z/, "")]
-    else
-      return unless SINCE_FAST_PATH[feature]
-    end
+    return unless SINCE_FAST_PATH[File.basename(feature, ".*")]
 
     # bootsnap expands `require "csv"` to `require "#{LIBDIR}/csv.rb"`,
     # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
```

Which might be simpler.

hsbt:
@byroot Thank you for this. In my understanding, this change makes `bundled_gems.rb` only check items from `SINCE` list.

We can speed up to load time with `psych` or other stdlibs without `SINCE` list?

casperisfine:
> this change makes `bundled_gems.rb` only check items from `SINCE` list.

Yes, it use a fast heuristic to return quickly if the required feature has no chance to be in the `SINCE` constant.

> We can speed up to load time with psych or other stdlibs without SINCE list?

That's the idea.

hsbt (approved):

hsbt:
@casperisfine Thanks for your explanation 👍

hsbt:
📝 I confirmed to work this patch and `ruby_3_3` HEAD with `tool/test_for_warn_bundled_gems/test.sh`

lib/bundled_gems.rb
eregon: Is `File.basename(feature, ".*")` correct?
What if `feature` is e.g. `'rexml/node'` (or more realistic e.g. `digest/md5` but that's not extracted/warned)? Then `File.basename(feature, ".*")` is `node` and it will miss the warning, no?
I think we'd need the "root name" considering rubylibdir/rubyarchdir, i.e. similar to the logic below.

casperisfine: It's an heuristic based on a few assumptions, the main one being that you'll require the entry point at some point. 

e.g. you can indeed `require "logger/formatter"` and that won't trigger a warnings, but the idea is that you'll very likely end up requiring `"logger"`.

In the `digest/md5` case it end up requiring `"digest"`:

- https://github.com/ruby/digest/blob/d152f5a043ea49ea7968bfebf7cd6b9318b4d557/ext/digest/md5/md5init.c#L55
- https://github.com/ruby/digest/blob/d152f5a043ea49ea7968bfebf7cd6b9318b4d557/ext/digest/digest.h#L57

But yes, it's possible there are a few corner cases not covered, but as explained in the issue, I think we need to be careful not to do too much work to identify the gem being called, because the performance of `require` is important.

casperisfine: But if you find another heuristic that's more robust and in the same performance ballpark I'll be happy to review it.

eregon: Could you document these assumptions as a code comment?
I don't see any mention of them before these 2 replies.

PR #11210, opened 2024-07-19
Restrict const checks to current namespace to avoid conflicts with globals

rwstauner:
Getting this error on latest master when a top-level Parser constant is defined before URI is loaded:

```
$ ./ruby -Ilib -e 'module Parser; end; require "uri/common"'
/Users/rwstauner/src/github.com/ruby/ruby/lib/uri/common.rb:26:in 'Module#remove_const': constant URI::Parser not defined (NameError)

    remove_const(:Parser) if defined?(Parser)
    ^^^^^^^^^^^^
        from /Users/rwstauner/src/github.com/ruby/ruby/lib/uri/common.rb:26:in 'URI.parser='
 ```

upstream PR https://github.com/ruby/uri/pull/108

@hsbt

### DIFF
--- a/lib/uri/common.rb
+++ b/lib/uri/common.rb
@@ -23,11 +23,11 @@ module URI
   Ractor.make_shareable(DEFAULT_PARSER) if defined?(Ractor)
 
   def self.parser=(parser = RFC3986_PARSER)
-    remove_const(:Parser) if defined?(Parser)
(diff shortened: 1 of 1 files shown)


rwstauner:
synced from uri via da49bee04b1ea664199c14318dbf594464974636

PR #11215, opened 2024-07-20
Fix SEGV on method call with empty args and brace block for do block command call

yui-knk:


parse.y
```diff
@@ -5286,7 +5286,7 @@ block_call	: command do_block
                 | block_call call_op2 operation2 opt_paren_args brace_block
                     {
                         bool has_args = $5 != 0;
-                        if (NODE_EMPTY_ARGS_P($5)) $5 = 0;
+                        if (NODE_EMPTY_ARGS_P($4)) $4 = 0;
```
nobu, line 5289: Is `has_args` `$4 != 0` as well?

nobu (approved):

parse.y
nobu: Hmmm, probably no, `has_args` is misleading here.

yui-knk: On current implementation, `has_args` has no meaning because `brace_block` (`$5`) is always evaluated to true. However I guess the ripper events for this rule might be wrong. Should it to be `call!`, `method_add_arg!` and `method_add_block!`?

Anyway I will merge this fix.

PR #11219, opened 2024-07-22
Document the reasoning behind the fix for [Bug #20641]

casperisfine:
@eregon as requested in https://github.com/ruby/ruby/pull/11208#discussion_r1686458864

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -95,7 +95,16 @@ module Gem::BUNDLED_GEMS
     # name can be a feature name or a file path with String or Pathname
     feature = File.path(name)
 
+    # The actual checks needed to properly identify the gem being required
+    # are costly (see [Bug #20641]), so we first do a much cheaper check
+    # to exclude the vast majority of candidates.
     if feature.include?("/")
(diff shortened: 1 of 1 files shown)


lib/bundled_gems.rb
```diff
@@ -95,7 +95,16 @@ def self.warning?(name, specs: nil)
...
 
+    # The actual checks needed to properly identify the gem being required
+    # are costly (see [Bug #20641]), so we first do a much cheaper check
+    # to exclude the vast majority of candidates.
     if feature.include?("/")
+      # If requiring $LIBDIR/mutex_m.rb, we check SINCE_FAST_PATH["mutex_m"]
+      # We'll fail to warn requires for files that are not the entry point
+      # of the gem, e.g. require "logger/formatter.rb" won't warn.
+      # But that's acceptable because this warning is best effort,
+      # and in the overwhelming majority of case logger.rb will end
```
eregon, line 106:
```suggestion
      # and in the overwhelming majority of cases logger.rb will end
```

eregon (approved):

PR #11222, opened 2024-07-22
[PRISM] Fix memory leak in constants

peterzhu2118:
For example, the following code leaks:

```ruby
code = 1000.times.map { |i| "var#{i} = 1" }.join("\n")

10.times do
  1000.times do
    RubyVM::InstructionSequence.compile_prism(code)
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

    70384
    88032
    103856
    115712
    125584
    132768
    144784
    152624
    165296
    180608

After:

    62368
    78784
    74512
    87712
    85072
    77728
    69424
    74992
    71264
    81440

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -10289,7 +10289,7 @@ pm_parse_process(pm_parse_result_t *result, pm_node_t *node)
     // Now set up the constant pool and intern all of the various constants into
     // their corresponding IDs.
     scope_node->parser = parser;
-    scope_node->constants = calloc(parser->constant_pool.size, sizeof(ID));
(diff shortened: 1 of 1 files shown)


kddnewton (changes requested):
Could you call this in `pm_parse_result_free` instead? That lines up with the lifetime of the overall scope node/parse result.

peterzhu2118:
I have confirmed that moving it to `pm_parse_result_free` also fixes it. I also changed the `calloc` to `xcalloc` to be consistent.

kddnewton (approved):

PR #11233, opened 2024-07-23
Don't check live slot count when multi-Ractor

peterzhu2118:


maximecb (approved):

XrXr:
🙏🏼 I would merge with context that this is for improving allocation throughput. Compared to the start of the year it has regressed a bit, and this returns some of it.

PR #11238, opened 2024-07-24
[Bug #20650] Fix memory leak in Regexp capture group when timeout

peterzhu2118:
The capture group allocates memory that is leaked when it times out.

For example:

```ruby
re = Regexp.new("^#{"(a*)" * 10_000}x$", timeout: 0.000001)
str = "a" * 1000000 + "x"

10.times do
  100.times do
    re =~ str
  rescue Regexp::TimeoutError
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

```
34688
56416
78288
100368
120784
140704
161904
183568
204320
224800
```

After:

```
16288
16288
16880
16896
16912
16928
16944
17184
17184
17200
```


mame (approved):
Thanks!

PR #11245, opened 2024-07-25
[DOC] Doc for some #ceil and #floor

BurdetteLamar:
Completes #ceil and #floor for Integer, Float, and Numeric.

### DIFF
--- a/numeric.c
+++ b/numeric.c
@@ -2170,36 +2170,82 @@ flo_ndigits(int argc, VALUE *argv)
 }
 
 /*
+ *  :markup: markdown
+ *
  *  call-seq:
  *    floor(ndigits = 0) -> float or integer
  *
- *  Returns the largest number less than or equal to +self+ with
- *  a precision of +ndigits+ decimal digits.
+ *  Returns a float or integer that is a "floor" value for `self`,
+ *  as specified by `ndigits`,
+ *  which must be an
+ *  [integer-convertible object](rdoc-ref:implicit_conversion.rdoc@Integer-Convertible+Objects).
  *
- *  When +ndigits+ is positive, returns a float with +ndigits+
+ *  When `self` is zero,
+ *  returns a zero value:
+ *  a float if `ndigits` is positive,
+ *  an integer otherwise:
+ *
+ *  ```
+ *  f = 0.0      # => 0.0
+ *  f.floor(20)  # => 0.0
+ *  f.floor(0)   # => 0
+ *  f.floor(-20) # => 0
+ *  ```
+ *
+ *  When `self` is non-zero and `ndigits` is positive, returns a float with `ndigits`
  *  digits after the decimal point (as available):
  *
- *    f = 12345.6789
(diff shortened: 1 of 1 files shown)


numeric.c
```diff
@@ -2170,36 +2170,82 @@ flo_ndigits(int argc, VALUE *argv)
...
+ *  |      -1 |          10 |                     -12350 |
+ *  |      -2 |         100 |                     -12400 |
+ *  |      -3 |        1000 |                     -13000 |
+ *  |      -4 |       10000 |                     -20000 |
+ *  |      -5 |      100000 |                    -100000 |
+ *  |      -6 |     1000000 |                   -1000000 |
+ *  |      -7 |    10000000 |                  -10000000 |
+ *  |      -8 |   100000000 |                 -100000000 |
+ *  |      -9 |  1000000000 |                -1000000000 |
+ *  |     -10 | 10000000000 |                          0 |
```
peterzhu2118, line 2241: I'm not sure why the result is 0 here, is it not `-10000000000`?

```ruby
-12345.6789.floor(-10) # => -10000000000
```

BurdetteLamar: Not on my machine:
```
-12345.6789.floor(-10)
# => 0
-12345.6789.floor(-20)
# => 0
```

```
irb --version
irb 1.11.0 (2023-12-19)
```

peterzhu2118: That's odd. What's your Ruby version (can you paste the output of `ruby -v`)? Also, what's the output of your `-12345.floor(-10)`? It should be `-10000000000`.

BurdetteLamar:
```
ruby -v
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x64-mingw-ucrt]
```

```
irb
-12345.floor(-10)
# => 0
```

BurdetteLamar: I also see the difference between the two.  I'll be happy to confirm and put up the Linux results if that seems best.

peterzhu2118: Oh I see what's going on. When the rounding will overflow, it returns 0 instead. Windows' long is 4 bytes whereas on macOS/Linux it's 8 bytes. I can reproduce the result with a larger magnitude `ndigits`.

```ruby
-12345.6789.floor(-100) # => 0
-12345.floor(-100) # => 0
12344.ceil(-100) # => 0
```

I'm not sure if this is intended behaviour. If it's not intended behaviour, then we shouldn't document it. If it is, then we should have a separate section to describe it.

@nobu do you know if this is intended behaviour?

BurdetteLamar: I have removed the questionable lines from the table.

BurdetteLamar:
Had an incomplete commit for Numeric#floor, now fixed.

jeremyevans (approved):

numeric.c
peterzhu2118: I opened a PR that will fix this issue in https://github.com/ruby/ruby/pull/11257.

peterzhu2118 (approved):

PR #11249, opened 2024-07-25
Expand opt_newarray_send to support Array#pack with buffer keyword arg

rwstauner:
This way a call to `[v].pack(spec, buffer: b)` doesn't have to allocate for the array _or_ the string buffer it appends to. 

This eliminates 99% of the remaining allocations in the [`protoboeuf-encode` benchmark](https://github.com/Shopify/yjit-bench/blob/main/benchmarks/protoboeuf-encode/benchmark.rb)

The disasm for `$ ruby --dump=insns -e 'b = "x"; [v].pack("E*", buffer: b)'`:

before:

```
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,34)>
[ 1] b@0
0000 putchilledstring                       "x"                       (   1)[Li]
0002 setlocal_WC_0                          b@0
0004 putself
0005 opt_send_without_block                 <calldata!mid:v, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0007 newarray                               1
0009 putchilledstring                       "E*"
0011 getlocal_WC_0                          b@0
0013 opt_send_without_block                 <calldata!mid:pack, argc:2, kw:[#<Symbol:0x000000000023110c>], KWARG>
0015 leave
```

after:

```
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,34)>
[ 1] b@0
0000 putchilledstring                       "x"                       (   1)[Li]
0002 setlocal_WC_0                          b@0
0004 putself
0005 opt_send_without_block                 <calldata!mid:v, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0007 putchilledstring                       "E*"
0009 getlocal                               b@0, 0
0012 opt_newarray_send                      3, 5
0015 leave
```

For reference support for `:pack` in `opt_newarray_send` was added in https://github.com/ruby/ruby/pull/8734


insns.def
```diff
@@ -983,7 +983,7 @@ opt_str_uminus
 
 DEFINE_INSN
 opt_newarray_send
-(rb_num_t num, ID method)
+(rb_num_t num, ID method, VALUE arg)
```
rwstauner, line 986: My original implementation instead setup a new id for `pack__buffer` and used that instead of `pack`.
I thought the double underscore would avoid ever confusing it with an actual method name.

While it was a little simpler, I wondered if maybe adding `idPack__buffer` was too weird so I changed this instruction to take an additional argument.

It could be useful in the future if we found another case to support.
I do see that `min` and `max` take an arg, so if we wanted to optimize that case we could use this argument to do that.
Until then `min`, `max`, `hash` (and the single-arg case of `pack`) are just sending a `Qnil` in its place.

I'm not sure if "separate id" had any advantages over "additional arg" so I thought I'd mention it.

rwstauner: From the benchmarks I've tried performance seems the same between the two solutions.

rwstauner: Also open to other suggestions for the type of the third arg.
The `VALUE` is versatile and the symbol is clear but we could use an int as a enum and avoid all the conversions 🤷

k0kubun: If I understand correctly, `rb_id2sym` seems like an unneeded overhead. Why not pass `ID` for now, and consider converting it to VALUE when it really needs to be?

rwstauner: Thanks, I'll do that.  As it is it looks like the other version might be a little faster so maybe that will help bring them in line.

rwstauner: Now I realize that the `operands` array is typed `VALUE`
https://github.com/ruby/ruby/blob/f2728c3393d7571f681accd9c887dcd5bad65950/compile.c#L95

Trying to change it to just the id is giving me
`-e: wrong argument type Integer (expected Symbol) (TypeError)`

rwstauner: If I can't change the assignment to the operands it looks like I might be able to change the type in the insn def and then drop the rb_id2sym/sym2id calls when doing the comparisons.
I'll try that and see if it measures any differently.

k0kubun: We place insns and operands every 8 bytes and access it using a `VALUE *` pointer. But `ID` doesn't take more than 8 bytes, and remember, the previous operand `ID method` is already an `ID`. I don't know where the `TypeError` comes from, but it might not necessarily be a blocker.

rwstauner: Interesting, thanks!

Yeah, the `ID method` is what I was referring to... I realized that was there, so I could change the type of this new arg at that same level.
The `TypeError` seems to be coming from that conversion... the `operand` is a `VALUE` and if I set the arg type on the insn to anything else I get the TypeError when the arg doesn't convert.

If I make it `ID` it does simplify the receiving code and it does work, but it then requires me to pass a symbol in the other two code paths ("min/max/hash" and "pack without buffer").

I had `Qnil` and that produces
`wrong argument type nil (expected Symbol) (TypeError)`.
If I remove the assignment to the last element I get (as you can probably guess):
`wrong argument type false (expected Symbol) (TypeError)`.

So in order to make it an `ID` I have to make up a symbol to pass to the other ones (or duplicate the `mid` arg).
This seems to be true for any type more specific than `VALUE` (since 0 works correctly for that and probably nothing else).

One thing we could do to simplify the conditions it just treat the VALUE as a boolean:
For the other code paths the arg is false, and if it's truthy we consider the buffer to be passed.
We don't need to do the symbol/ID conversion and comparison if there aren't any other possible values.

It's a little simpler, but still not great.
It makes me like the original implementation (of just using an alternate symbol) a bit more 🤔 .

XrXr: The indirection through a Symbol before it gets turned into an ID in `encoded_iseq` could be avoided if we have the instruction take a `rb_num_t` instead. The `method` operand is already acting like an enum, using ID as the underlying representation.
Changing away from ID also gets rid of the weirdness of having to make a bespoke ID with an entry in the ID table with the string content left consumed nowhere.

rwstauner: Thanks!  I had considered `rb_num_t` because I thought it would be simple to just use it like an enum and the other code paths could assume 0... but I discovered they cannot, since they need to be a VALUE they have to set `operands[2] = FIX2INT(0)` or something similar, so it's not quite as clean as I was hoping.

If `rb_num_t` seems the clearest or most consistent I'm happy to switch to that.

XrXr: I'm suggesting we change the `method` operand to `rb_num_t` and add no new operand. That way you don't need to set a dummy value for the other paths as they'd each be a variation of the enum.

rwstauner: Oh, ok.  Sure, I'll try that.
The code was simpler when I was just adding another case.  If we can do it that way without the extra id that sounds good to me.

maximecb (approved):
This is a very useful patch. Thanks for adding all the extra tests as well. Good work 🙏

vm_insnhelper.c
```diff
@@ -6204,18 +6204,38 @@ VALUE rb_setup_fake_ary(struct RArray *fake_ary, const VALUE *list, long len, bo
...
+            kw_splat = RB_PASS_KEYWORDS;
+            argc++;
+        }
+
+        return rb_vm_call_with_refinements(ec, rb_ary_new4(num, ptr), idPack, argc, args, kw_splat);
     }
 }
 
+VALUE
+rb_vm_opt_newarray_pack(rb_execution_context_t *ec, rb_num_t num, const VALUE *ptr, VALUE fmt)
```
k0kubun, line 6234:
```suggestion
static VALUE
vm_opt_newarray_pack(rb_execution_context_t *ec, rb_num_t num, const VALUE *ptr, VALUE fmt)
```

We export a symbol and use an `rb_` prefix to avoid collision when other object files like YJIT use it. Now that YJIT stopped using `rb_vm_opt_newarray_pack`, it seems confusing that it's still exported.

Nobu added a `rb_vm_opt_newarray_pack` bindgen to RJIT but he didn't seem to use it, so you can just remove  [this line](https://github.com/ruby/ruby/blob/86a762ce56cd44db69ccff2e02587ed748a3ad04/tool/rjit/bindgen.rb#L544) and run `make rjit-bindgen` again (or use its output on the CI). If you're unsure, I can handle this myself too.

k0kubun (approved):
looks good to me :+1:

vm_insnhelper.c
```diff
@@ -6204,18 +6204,38 @@ VALUE rb_setup_fake_ary(struct RArray *fake_ary, const VALUE *list, long len, bo
...
     else {
-        return rb_vm_call_with_refinements(ec, rb_ary_new4(num, ptr), idPack, 1, &fmt, RB_PASS_CALLED_KEYWORDS);
+        // The opt_newarray_send insn drops the keyword args so we need to rebuild them.
+        // Setup an array with room for keyword hash.
+        VALUE args[2];
+        args[0] = fmt;
+        int kw_splat = RB_NO_KEYWORDS;
+        int argc = 1;
+
+        if (!NIL_P(buffer)) {
```
XrXr, line 6222: This have it not being able to distinguish between passing `buffer: nil` and not passing a keyword at all. The following raises when it should not:

```
class Array
  def pack(_fmt, buffer:) = buffer
end

v = 1
b = nil;
p [v].pack("E*", buffer: b)
```

rwstauner: Indeed, thank you!
At one point I was trying to figure out if it would be odd for the method to receive the keyword arg when it wasn't provided (something like `def pack(_fmt, **kwargs); kwargs.empty?`) and then apparently I didn't add enough tests.

rwstauner: Do you think the difference is important enough to warrant having duplicate logic in `rb_vm_opt_newarray_pack`
(rather than having it just be a wrapper around `rb_vm_opt_newarray_pack_buffer`)?

vm_insnhelper.c
rwstauner: I was going to add the function references to the rjit code like #8734 did until I realized rjit wasn't using the insn:
https://github.com/ruby/ruby/blob/e798f45cc2417ae778e4790e3159654c90ba91fe/lib/ruby_vm/rjit/insn_compiler.rb#L1526-L1537

When I tried `make rjit-bindgen` I was getting this error:
```
bindgen.rb:249:in `block (2 levels) in generate_node': non-leading bit fields are not supported. consider including 'is_from_method' in skip_fields. (RuntimeError)
```
so I thought I'd leave that for another PR.

vm_insnhelper.c
XrXr: Maybe you can get it to work by passing `Qundef` to indicate that the callsite passes no keywords?

rwstauner: Oh, good point.  I forgot we had that one.  Thanks!

vm_insnhelper.c
rwstauner: I changed to use a static `vm_opt_newarray_pack_buffer` with the `rb_*` ones just being non-static wrappers for consistency with the other `vm_opt_newarray_*` functions.

vm_insnhelper.c
rwstauner: Switched to `Qundef` and added tests.

XrXr (commented):
Looks good! Please update the disassembly in the commit message, though.

rwstauner:
Oops.  I had updated the PR description but not the commit message.  Fixed!

PR #11254, opened 2024-07-26
BUG 20655: cont.call triggers the ensure function specified by rb_ensure

tagomoris:
The `cont.call` in the required ruby script triggers the ensure function specified on `rb_ensure` when the ruby script is required as the result of `func1` of `rb_ensure`.

https://bugs.ruby-lang.org/issues/20655

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -226,7 +226,6 @@ typedef struct rb_context_struct {
     } machine;
     rb_execution_context_t saved_ec;
     rb_jmpbuf_t jmpbuf;
-    rb_ensure_entry_t *ensure_array;
     struct rb_jit_cont *jit_cont; // Continuation contexts for JITs
 } rb_context_t;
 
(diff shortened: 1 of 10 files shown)


ko1:
confirmed with pure ruby script:

```ruby
require 'continuation'

Dir.chdir('/tmp') do
  p [__LINE__, tmp = Dir.pwd]
  cont = nil
  callcc{|c| cont = c}
  p [__LINE__, Dir.pwd]
  exit! if Dir.pwd != tmp
  cont.call
end
```

```
[4, "/tmp"]
[7, "/tmp"]
[7, "/tmp"]
[7, "/home/ko1/ruby/build/trunk"]
```

`Dir.chdir` uses `rb_ensure` and ensure func will restore the pwd (`/home/ko1/ruby/build/trunk` in this case).

tagomoris:
I created the bug ticket: https://bugs.ruby-lang.org/issues/20655

tagomoris:
The patch (by @mame ) was cherry-picked from https://github.com/ruby/ruby/pull/11264

tagomoris:
And I also included the test case inspired by the @ko1 's comment.

PR #11256, opened 2024-07-26
[DOC] rb_ary_s_create

BurdetteLamar:
Tune up Array::[].  Removes two examples that don't actually call Array::[].

### DIFF
--- a/array.c
+++ b/array.c
@@ -1159,11 +1159,11 @@ rb_ary_initialize(int argc, VALUE *argv, VALUE ary)
 }
 
 /*
- * Returns a new array populated with the given objects.
+ * Returns a new +Array+ object, populated with the given objects:
  *
- *   Array.[]( 1, 'a', /^A/)  # => [1, "a", /^A/]
- *   Array[ 1, 'a', /^A/ ]    # => [1, "a", /^A/]
- *   [ 1, 'a', /^A/ ]         # => [1, "a", /^A/]
+ *   Array.[]( 1, 'a', /^A/) # => [1, "a", /^A/]
+ *   Array.[]                # => []
+ *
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -1159,11 +1159,11 @@ rb_ary_initialize(int argc, VALUE *argv, VALUE ary)
 }
 
 /*
- * Returns a new array populated with the given objects.
+ * Returns a new +Array+ object, populated with the given objects:
+ *
+ *   Array.[]( 1, 'a', /^A/) # => [1, "a", /^A/]
+ *   Array.[]                # => []
```
peterzhu2118, line 1165: Can we also keep the example using `Array[1, 'a', /^A/]`? It's the same method call but using different syntax (the most idiomatic syntax).

array.c
```diff
@@ -1159,11 +1159,11 @@ rb_ary_initialize(int argc, VALUE *argv, VALUE ary)
 }
 
 /*
- * Returns a new array populated with the given objects.
+ * Returns a new +Array+ object, populated with the given objects:
+ *
+ *   Array.[]( 1, 'a', /^A/) # => [1, "a", /^A/]
```
peterzhu2118, line 1164: Nit: extra space:

```suggestion
 *   Array.[](1, 'a', /^A/) # => [1, "a", /^A/]
```

array.c
BurdetteLamar: Restored (and moved the idiomatic syntax to top).

array.c
BurdetteLamar: Fixed.  (There are no nits.)

peterzhu2118 (approved):

PR #11259, opened 2024-07-27
[DOC] Tweaks for Array.new

BurdetteLamar:
- Rationalizes call-seq schemas.
- Reorganizes discussion into no-block/block.

### DIFF
--- a/array.c
+++ b/array.c
@@ -1060,48 +1060,41 @@ rb_ary_s_new(int argc, VALUE *argv, VALUE klass)
  *  call-seq:
  *    Array.new -> new_empty_array
  *    Array.new(array) -> new_array
- *    Array.new(size) -> new_array
- *    Array.new(size, default_value) -> new_array
- *    Array.new(size) {|index| ... } -> new_array
+ *    Array.new(size, default_value = nil) -> new_array
+ *    Array.new(size = 0) {|index| ... } -> new_array
  *
- *  Returns a new +Array+.
+ *  Returns a new +Array+ object.
  *
- *  With no block and no arguments, returns a new empty +Array+ object.
+ *  With no block given:
  *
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -1060,48 +1060,41 @@ rb_ary_s_new(int argc, VALUE *argv, VALUE klass)
...
- *  the element for that +index+ is the return value from the block:
+ *  With a block given,
+ *  returns an +Array+ of the given +size+;
+ *  calls the block with each +index+ in the range <tt>(0..size-1)</tt>;
+ *  the element at that +index+ is the return value from the block:
  *
  *    a = Array.new(3) {|index| "Element #{index}" }
  *    a # => ["Element 0", "Element 1", "Element 2"]
  *
- *  Raises ArgumentError if +size+ is negative.
```
peterzhu2118, line 1100: Can we keep this one? I think we should describe what happens when `size` is negative.

BurdetteLamar: Restored.

peterzhu2118: Can we put this under the bullet point for `size`? I think it would make the most sense there.

BurdetteLamar: Done.

BurdetteLamar:
I've made some more changes here (hope not too burdensome to the review):
- Use outer bullets to emphasize the two cases:  no-block/block.
- Say 'array' instead of '+Array+ object', to keep it simple and casual (avoids the clutter of font-change).
- Move integer-convertible object remark to the end (applies to both no-block and block cases).
- Cite exceptions in the examples when feasible.
- Cite trivial case (size = 0) in examples.

Some of these ideas may belong as "considers" in the guidelines document?

peterzhu2118 (approved):

PR #11260, opened 2024-07-27
[DOC] Tweaks to Array.try_convert

BurdetteLamar:
- Adds an introductory overview statement.
- Simplifies "case" cascade (if, otherwise, otherwise).

### DIFF
--- a/array.c
+++ b/array.c
@@ -1023,7 +1023,7 @@ rb_to_array(VALUE ary)
  *    returns <tt>object.to_ary</tt>.
  *  - Otherwise returns +nil+.
  *
- *  Raises an exception unless <tt>object.to_ary</tt> returns an +Array+ object.
+ *  Raises an exception unless <tt>object.to_ary</tt> returns an +Array+ object or +nil+.
  */
 
 static VALUE
--- a/array.c
+++ b/array.c
@@ -1019,11 +1019,14 @@ rb_to_array(VALUE ary)
  *  Attempts to convert the given +object+ to an +Array+ object:
  *
  *  - If +object+ is an +Array+ object, returns +object+.
- *  - Otherwise if +object+ responds to <tt>:to_ary</tt>,
- *    returns <tt>object.to_ary</tt>.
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -1016,12 +1016,12 @@ rb_to_array(VALUE ary)
...
- *  If +object+ is an +Array+ object, returns +object+.
+ *  Attempts to convert the given +object+ to an +Array+ object:
  *
- *  Otherwise if +object+ responds to <tt>:to_ary</tt>,
- *  calls <tt>object.to_ary</tt> and returns the result.
- *
- *  Returns +nil+ if +object+ does not respond to <tt>:to_ary</tt>
+ *  - If +object+ is an +Array+ object, returns +object+.
+ *  - Otherwise if +object+ responds to <tt>:to_ary</tt>,
+ *    returns <tt>object.to_ary</tt>.
```
nobu, line 1023: This may be worth to mention, that a `TypeError` is raised if the result of `object.to_ary` is neither an `Array` nor `nil`.

BurdetteLamar: Done.

array.c
```diff
@@ -1016,14 +1016,14 @@ rb_to_array(VALUE ary)
...
- *  calls <tt>object.to_ary</tt> and returns the result.
+ *  - If +object+ is an +Array+ object, returns +object+.
+ *  - Otherwise if +object+ responds to <tt>:to_ary</tt>,
+ *    returns <tt>object.to_ary</tt>.
+ *  - Otherwise returns +nil+.
  *
- *  Returns +nil+ if +object+ does not respond to <tt>:to_ary</tt>
- *
- *  Raises an exception unless <tt>object.to_ary</tt> returns an +Array+ object.
+ *  Raises TypeError unless <tt>object.to_ary</tt> returns an +Array+ object or +nil+.
```
peterzhu2118, line 1026: I think we should move this under the second bullet point since `object` could be an array or nil, then the return value of `object.to_ary` is never checked.

BurdetteLamar: Modified.

peterzhu2118 (approved):

PR #11261, opened 2024-07-27
[DOC] Add remark about call-seq

BurdetteLamar:


BurdetteLamar:
The more I think about this, the less I like it.  `array & other_array` has a naturalness, while `self & other_array` seems clunky.

But I think we should retain the guideline, one way or the other.

I'd be happy to see more opinions.

jeremyevans (approved):

BurdetteLamar:
@jeremyevans, I'm not clear:  you approve `array & other_array`?  Or `self & other_array`?  (I know I've clouded the issue by changing my mind mid-stream.)

jeremyevans:
I was looking at the patch when I approved it (which has `self &`).  I don't have strong feelings either way.  I think `self &` is more similar to the documentation of regular methods, which use implicit `self` (e.g. `method(arg)`, not `array.method(arg)`).

peterzhu2118 (approved):
I agree with @jeremyevans. I also don't have a strong opinion, but `self & obj` is more consistent with the rest of the documentation.

PR #11266, opened 2024-07-28
Add spec for keywords, and their usage as Hash keys

amomchilov:
This documents behaviour like:

```ruby
def do_something(if:)
  # Behaves like `binding.local_variable_get(:if)`, but faster.
  condition = { if: }[:if]
end
```

cc @eregon, because I saw you contributed the Spec :)

### DIFF
--- a/spec/ruby/language/reserved_keywords.rb
+++ b/spec/ruby/language/reserved_keywords.rb
@@ -1,0 +1,113 @@
+require_relative '../spec_helper'
+
+describe "Ruby's reserved keywords" do
+  # Copied from Prism::Translation::Ripper
+  keywords = [
+    "alias",
+    "and",
+    "begin",
+    "BEGIN",
+    "break",
+    "case",
+    "class",
+    "def",
+    "defined?",
+    "do",
+    "else",
+    "elsif",
+    "end",
+    "END",
+    "ensure",
+    "false",
+    "for",
+    "if",
+    "in",
+    "module",
+    "next",
+    "nil",
+    "not",
+    "or",
+    "redo",
+    "rescue",
+    "retry",
+    "return",
+    "self",
+    "super",
+    "then",
+    "true",
+    "undef",
+    "unless",
+    "until",
+    "when",
+    "while",
+    "yield",
+    "__ENCODING__",
+    "__FILE__",
+    "__LINE__"
+  ]
+
+  invalid_kw_param_names = [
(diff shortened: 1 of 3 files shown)


spec/ruby/language/reserved_keywords.rb
```diff
@@ -0,0 +1,113 @@
+require_relative '../spec_helper'
+
+describe "Ruby's reserved keywords" do
+  # Copied from Prism::Translation::Ripper
+  keywords = [
```
amomchilov, line 5: TODO: It would be nice to have one canonical list that can be used in all these places:

1. `spec/ruby/language/hash_spec.rb`
2. `spec/ruby/language/reserved_keywords.rb`
3. `lib/prism/translation/ripper.rb`
4. Ideally, it would also be nice to generate `doc/keywords.rdoc` from that list, to ensure that all keywords are documented.

Where should such a canonical list live?

spec/ruby/language/reserved_keywords.rb
```diff
@@ -0,0 +1,113 @@
...
+    "defined?",
+  ]
+
+  invalid_method_names = [
+    "BEGIN",
+    "END",
+    "defined?",
+  ]
+
+  def expect_syntax_error(ruby_src)
```
amomchilov, line 61: TODO: move to `spec_helper`?

spec/ruby/language/reserved_keywords.rb
```diff
@@ -0,0 +1,113 @@
...
+    -> { eval(ruby_src) }.should raise_error(SyntaxError)
+  end
+
+  # Evaluates the given Ruby source in a temporary Module, to prevent
+  # the surrounding context from being polluted with the new methods.
+  def sandboxed_eval(ruby_src)
```
amomchilov, line 67: TODO: fix code duplication

spec/ruby/language/reserved_keywords.rb
nobu: I don't think definition of global methods in example files have good effects.

nobu (commented):
Please send this PR to its own upstream, https://github.com/ruby/spec.

amomchilov:
Reopened in:

1. https://github.com/ruby/spec/pull/1187
2. https://github.com/ruby/spec/pull/1188
3. https://github.com/ruby/spec/pull/1189

PR #11274, opened 2024-07-29
Fix memory leak reported in ripper tests

peterzhu2118:
There's memory leaks reported in ripper tests because it returns out of the Ripper#compile_error, Ripper#warn, Ripper#warning methods, so the parsing is abruptly terminated and cannot clean up.


nobu:
It shouldn't leak even if `return`ed from the block, I think.

peterzhu2118:
I opened #11287, #11288, and #11289 to fix these issues in another way.

PR #11280, opened 2024-07-30
YJIT: Pass method arguments using registers

k0kubun:
This PR allows `gen_send_iseq` to convert a caller's stack temp registers to a callee's local variable registers. It manipulates `callee_ctx` to reuse the same registers, so no `mov` instructions are generated to transfer registers.

It gives a 2% speedup to `liquid-c`, `liquid-render`, and `mail`.

```
before: ruby 3.4.0dev (2024-08-15T20:00:09Z master abc04e898b) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-08-15T20:19:50Z yjit-local-args 7bd011cc8b) +YJIT [x86_64-linux]

activerecord    117.3        0.4         118.6       0.2         0.98           0.99
chunky-png      292.3        0.3         293.4       0.3         1.00           1.00
erubi-rails     509.7        0.4         502.9       0.4         0.99           1.01
hexapdf         994.6        1.0         997.3       1.1         1.00           1.00
liquid-c        31.6         0.8         31.1        0.7         0.97           1.02
liquid-compile  34.1         0.6         34.2        1.0         0.93           1.00
liquid-render   43.3         0.5         42.6        0.6         0.96           1.02
lobsters        504.9        1.0         508.3       2.2         0.97           0.99
mail            69.6         0.3         68.3        0.5         1.01           1.02
psych-load      954.3        0.1         958.1       0.1         0.99           1.00
railsbench      1207.3       0.9         1210.5      1.3         0.99           1.00
rubocop         62.1         5.6         62.3        5.7         0.98           1.00
ruby-lsp        68.9         0.8         72.2        1.2         0.97           0.96
sequel          33.9         0.5         33.6        0.3         1.01           1.01
```


yjit/src/core.rs
```diff
@@ -2094,9 +2099,14 @@ fn get_num_versions(blockid: BlockId, inlined: bool) -> usize {
...
-                        Context::decode(unsafe { version.as_ref() }.ctx).inline() == inlined
-                    ).count()
+                    versions.iter().filter(|&&version| {
+                        let version_ctx = Context::decode(unsafe { version.as_ref() }.ctx);
+                        // Inline versions are counted separately towards MAX_INLINE_VERSIONS.
+                        version_ctx.inline() == ctx.inline() &&
+                            // find_block_versions() finds only blocks with compatible reg_mapping,
+                            // so count only versions with compatible reg_mapping.
+                            version_ctx.reg_mapping == ctx.reg_mapping
+                    }).count()
```
maximecb, line 2109: Important to look at code size since we are increasing the number of versions generated

yjit/src/codegen.rs
```diff
@@ -2345,7 +2357,16 @@ fn gen_getlocal_generic(
...
         let offs = -(SIZEOF_VALUE_I32 * ep_offset as i32);
-        Opnd::mem(64, ep_opnd, offs)
+        let local_opnd = Opnd::mem(64, ep_opnd, offs);
+
+        // Write back an argument register to the stack. If the local variable
+        // is an argument, it might have an allocated register, but if this ISEQ
+        // is known to escape EP, the register shouldn't be used after this getlocal.
+        if level == 0 && asm.ctx.get_reg_mapping().get_reg(asm.local_opnd(ep_offset).reg_opnd()).is_some() {
+            asm.mov(local_opnd, asm.local_opnd(ep_offset));
+        }
```
maximecb, line 2367: I'm a bit confused here.

So we have a situation where some ISEQs escape their environment. We allow getlocal to use a register once and then not again after?

yjit/src/codegen.rs
```diff
@@ -114,10 +114,13 @@ pub struct JITState<'a> {
 
     /// Stack of symbol names for --yjit-perf
     perf_stack: Vec<String>,
+
+    /// When true, this block is the first block compiled by gen_block_series().
+    first_block: bool,
```
maximecb, line 119: Maybe we could make this an unsigned integer and then we could know how long block series can get (could be interesting)

yjit/src/core.rs
maximecb: Can you explain, are you making block versions during calls as inline? Are new versions being marked as inline that were not being marked as inline before?

maximecb:
Hi Kokubun.

Congratulations on getting all the tests passing on this. I'm sure it was not easy. At first glance I think the PR makes sense. I think I understand most of it, but I did ask some clarifying questions.

My main concern would be code size, if we are generating more versions to match register allocation. Can you check the code size before and after on lobsters? I'm also curious to see how performance changes on `fib` and `30k_methods` (because those benchmarks exercise calls heavily and may amplify the difference).

I wish we had gotten more of a speedup from this change. It would seem like being able to access arguments through registers should be a decent win. It may be worth looking at the machine code to see how efficient things are. Not opposed to merging if we think that code size is not an issue and if we think we have a path forward for improving code generation more.

yjit/src/codegen.rs
```diff
@@ -215,6 +219,13 @@ impl<'a> JITState<'a> {
     // Check if we are compiling the instruction at the stub PC
     // Meaning we are compiling the instruction that is next to execute
     pub fn at_current_insn(&self) -> bool {
+        // If this is not the first block compiled by gen_block_series(),
+        // it might be compiling the same block again with a different Context.
+        // In that case, it should defer_compilation() and inspect the stack there.
+        if !self.first_block {
```
XrXr, line 225: This is a bit weird, because this is not about the PC anymore. Maybe we should improve the name of this method. Not a big deal though.

yjit/src/core.rs
XrXr: I would also like to get a better picture of how this is interacting with the version number limit. It's clearly under counting versions now, but I'm not sure how big an impact that has.

yjit/src/codegen.rs
maximecb: Also kind of curious to have an example of how the same block would get compiled again with a different context? Does this mean twice the same block in the same sequence? Like a loop getting unrolled?

yjit/src/core.rs
k0kubun: > Important to look at code size since we are increasing the number of versions generated

Before (master):

```
compiled_block_count:         70,803
versions_per_block:            1.501
inline_code_size:          7,503,452
outlined_code_size:        6,436,718
code_region_size:         15,413,248
```

After (this PR):

```
compiled_block_count:         75,454
versions_per_block:            1.600
inline_code_size:          8,359,786
outlined_code_size:        6,892,488
code_region_size:         17,068,032
```

For reference, here's the stats without this `&& version_ctx.reg_mapping == ctx.reg_mapping` change but with the rest of the change in this PR:

```
compiled_block_count:         73,872
versions_per_block:            1.588
inline_code_size:          8,158,683
outlined_code_size:        6,744,236
code_region_size:         16,642,048
```

> Can you explain, are you making block versions during calls as inline?

Yes I am when the callee has `Primitive.attr! :inline_block`, e.g. `Array#each`. No for most callees, which are user-defined or don't have `:inline_block` annotation.

> Are new versions being marked as inline that were not being marked as inline before?

No, nothing about inline was changed in this PR. I only added a comment about it since I touched this place.

> I would also like to get a better picture of how this is interacting with the version number limit. It's clearly under counting versions now, but I'm not sure how big an impact that has.

So the reason why I made this change is because, without this change (i.e. removing `&& version_ctx.reg_mapping == ctx.reg_mapping`), `activerecord` would slow down by 6-8%. I looked at the generated code closely, and one of the frequently-called BlockIds had an increased number of versions than before, which reached 5 (or more), due to a different `reg_mapping`. Thus it erased type information from Context, which was `BlockParamProxy`. Erasing `BlockParamProxy` resulted in always falling back to a generic method call in that block.

Since it's above the max versions limit, it creates `ctx.generic_ctx()` to share a block with the generic Context. However, because the `reg_mapping` was incompatible with exiting blocks, it ended up creating a new block that had a new reg_mapping and no type information. Since generating a generic Context doesn't help in this specific situation and `activerecord` slows down too much, I thought generating more `reg_mapping` variations shouldn't reduce the amount of type information.

yjit/src/codegen.rs
k0kubun: > We allow getlocal to use a register once and then not again after?

Yes. We could skip using a register and spill the stack slot early before making the call when YJIT already knows it would escape EP, you would still need this exact same logic in case YJIT doesn't know about it yet, and spilling it early doesn't reduce the number of instructions executed, so I kept the implementation simple by not doing it.

yjit/src/codegen.rs
k0kubun: We already have the max number of `get_num_versions(blockid, ctx)` in YJIT stats, but I guess you want to have the max count of all versions for a `BlockId` regardless of `inline` and `reg_mapping`?

yjit/src/codegen.rs
k0kubun: >  Maybe we should improve the name of this method.

Sure, I'll come up with something after having a lunch break.

> Also kind of curious to have an example of how the same block would get compiled again with a different context? Does this mean twice the same block in the same sequence?

I included an example in `test_yjit`, which originally lived in rubyspec. The `Foo#==` calls itself with a different `reg_mapping` at `opt_eq`.

k0kubun:
> My main concern would be code size, if we are generating more versions to match register allocation. Can you check the code size before and after on lobsters?

Copying from https://github.com/ruby/ruby/pull/11280#discussion_r1714115804, 

Before (master):

```
compiled_block_count:         70,803
versions_per_block:            1.501
inline_code_size:          7,503,452
outlined_code_size:        6,436,718
code_region_size:         15,413,248
```

After (this PR):

```
compiled_block_count:         75,454
versions_per_block:            1.600
inline_code_size:          8,359,786
outlined_code_size:        6,892,488
code_region_size:         17,068,032
```

> I'm also curious to see how performance changes on fib and 30k_methods (because those benchmarks exercise calls heavily and may amplify the difference).

It seems to generate code for `fib` as I intended, but the speedup is just 2%. With so many instructions for a frame push and removing or optimizing only a few instructions in the block, I'm not that surprised by the result though.

On the contrary, `30k_methods` seems to have a weird slowdown. I'll look into it. 

```
before: ruby 3.4.0dev (2024-08-08T23:13:16Z master 77ffdfe79f) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-08-09T03:18:25Z yjit-local-args a459c77cae) +YJIT [x86_64-linux]

30k_methods  352.1        1.2         363.7       0.1         0.98           0.97
fib          20.2         0.7         19.8        0.3         0.99           1.02
```

> I wish we had gotten more of a speedup from this change. It would seem like being able to access arguments through registers should be a decent win. It may be worth looking at the machine code to see how efficient things are.

I also feel like there might be some unnoticed trick that would realize its full potential across the board, but so far, nothing in generated code seems wrong.

It seems like I confused you in the weekly meeting that I discovered a bug that I haven't fixed yet, but this PR already has the fix for it. The `&& version_ctx.reg_mapping == ctx.reg_mapping` trick https://github.com/ruby/ruby/pull/11280#discussion_r1714115804 indeed sped up `activerecord` by a lot, compared to before fixing it, but it unfortunately didn't have much impact on other benchmarks.

I'll try to keep improving the implementation, but I have a suspicion that we have too many C calls/spills to benefit from this, or that the current bottleneck is not here and thus it seems insignificant as of today.

Nevertheless, I still want to move forward with this patch merged so that we can stop thinking about "our JIT code is slow because we haven't even optimized locals yet".

---

Now I thought we can use our time more efficiently if we pair over this first, so I sent an invite.

yjit/src/codegen.rs
k0kubun: > Maybe we should improve the name of this method.

Given that it's essentially a guard for `defer_compilation`, how about `jit.at_compile_target()` https://github.com/ruby/ruby/pull/11280/commits/acd8cd72de740503ddfb7ec634212ab19456b728?

maximecb:
> Nevertheless, I still want to move forward with this patch merged so that we can stop thinking about "our JIT code is slow because we haven't even optimized locals yet".

To me the code size is the main worry right now. It's about an 11% increase in inline code size. This is big. Might be worth looking at whether instruction cache misses increase.

It would also make sense to check if there are too many spills as you suggest. Ideally code using more registers should be smaller (less loads/stores to/from memory), so the question is, does the increase just come from more versions, or is it because there are many more loads and spills.

> I also feel like there might be some unnoticed trick that would realize its full potential across the board, but so far, nothing in generated code seems wrong.

I think it's likely that with some targeted tweaks you might be able to tip the balance such that it speeds up most benchmarks.

maximecb:
Something occurred to me yesterday.

I think it would be worth trying to get more insight into where extra blocks gets created.  We have a 6.5% increased in compiled block count. What worries me is that this increase could be very unevenly distributed. What I mean is that most methods may not see any increase in block version count, but you could get into a situation where some loop or some method gets duplicated many times over, which would be potentially very bad in terms of instruction cache performance. It's worth digging a bit to make sure that is not the case.

I apologize for making your life difficult Kokubun 😅 . Code size is really my only concern here, the only obstacle to merging. So I think it would make sense to try to look for pathological cases in our benchmarks and try to convince ourselves that there are none, or find ways to mitigate them if there are. Might be worth looking at protoboeuf since it has very big generated methods.

Back when I implemented BBV for my masters, I had set it up so that register allocation was not accounted for when looking for a compatible version. The downside is that this requires generating moves to go from versions with incompatible register allocation. The upside is that there is no increase in the number of versions (though I can't speak about the effect on code size). Alternatively, it may be possible to use heuristics to try to make the register allocation at the input of blocks match up better so that less versions are created.

k0kubun:
> generating moves to go from versions with incompatible register allocation

Pairing with Alan and Randy yesterday, Alan suggested the same thing. I've just finished what we left off yesterday, and pushed a commit to generate such moves https://github.com/ruby/ruby/pull/11280/commits/974f465cab5fc86e6533c8a2a5058d802d64aa60.

It decreased the number of versions generated, and it didn't slow down. Here's the code size comparison on lobsters (Note: I used `YJIT.runtime_stats` in the harness with `--yjit`, instead of using `--yjit-stats` which counts stats code too):

master:

```
compiled_block_count: 68127
inline_code_size: 4575579
outlined_code_size: 3957747
code_region_size: 11010048
```

Before (this PR without register moves):

```
compiled_block_count: 72293
inline_code_size: 4817514
outlined_code_size: 4510070
code_region_size: 11759616
```

After (this PR with register moves):

```
compiled_block_count: 69805
inline_code_size: 4669535
outlined_code_size: 4232636
code_region_size: 11366400
```

Compared to master, `inline_code_size` is increased only by 2%, which was 5.3% without register moves. `inline_code_size / compiled_block_count` is down from 67.16 to 66.89, so I think we're generating more compact code for each block.

maximecb:
> Compared to master, inline_code_size is increased only by 2%, which was 5.3% without register moves. inline_code_size / compiled_block_count is down from 67.16 to 66.89, so I think we're generating more compact code for each block.

That is great news 👌  Can you post the performance difference on headline as well as `30k_methods`, `fib` and `protoboeuf`?

Are we able to remove the changes to `fn get_num_versions` now that you are generating moves that can map between different register mappings? That is the only bit that seems maybe a bit hack-y. Hopefully should not be necessary now?

I think that with the changes you made the code size increase seems very reasonable. LGTM. Assuming that the performance change is generally in the positive direction I would say it is mostly ready to merge. I think there are probably other things that could be done to improve the register allocation further, but they can probably wait until subsequent PRs. Worth looking at the generated code on protoboeuf and other large methods at a future time to make sense of where efficiency could be gained.

k0kubun:
> Are we able to remove the changes to `fn get_num_versions` now that you are generating moves that can map between different register mappings?
> Hopefully should not be necessary now?

Removing `&& version_ctx.reg_mapping == ctx.reg_mapping` still significantly slows down `activerecord`, so it seems necessary.

In `ActiveRecord::AttributeMethods::Read#_read_attribute`, which was the place that slowed down, a `send` instruction takes a block argument that is either `Type::Nil` or `Type::BlockParamProxy`. The receiver type is either `Type::Unknown` or `Type::UnknownHeap`. Thus the block had 4 versions (the max) prior to this PR. If you want to use a register for the argument, it inevitably goes over the limit. 

Erasing the type information of the block argument is too bad for `activerecord`'s performance. If we want to allocate a register for arguments and still maintain `activerecord`'s performance, something has to change with regard to `get_num_versions` or the limit.

>  That is the only bit that seems maybe a bit hack-y. 

To me, counting versions for each `reg_mapping` seems very natural. Since blocks with different `reg_mapping`s have to generate different blocks whether you attempt to limit versions or not, it feels like the current `master` is counting different blocks as the same block. I don't find it hack-y to fix the discrepancy.

With the argument shuffling, the number of `reg_mapping` variations per block should be very minimal, like 2 in this case. One thing we could do on top of the current PR is that we limit the number of `reg_mapping` variations to 2 or 3, and let the caller spill locals to limit them if necessary. But looking at `lobsters` increasing code size only by 2%, I think it's already in a good shape.

k0kubun:
> Can you post the performance difference on headline as well as 30k_methods, fib and protoboeuf?

Here's the latest result:

```
before: ruby 3.4.0dev (2024-08-15T20:00:09Z master abc04e898b) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-08-15T20:19:50Z yjit-local-args 7bd011cc8b) +YJIT [x86_64-linux]

activerecord    117.3        0.4         118.6       0.2         0.98           0.99
chunky-png      292.3        0.3         293.4       0.3         1.00           1.00
erubi-rails     509.7        0.4         502.9       0.4         0.99           1.01
hexapdf         994.6        1.0         997.3       1.1         1.00           1.00
liquid-c        31.6         0.8         31.1        0.7         0.97           1.02
liquid-compile  34.1         0.6         34.2        1.0         0.93           1.00
liquid-render   43.3         0.5         42.6        0.6         0.96           1.02
lobsters        504.9        1.0         508.3       2.2         0.97           0.99
mail            69.6         0.3         68.3        0.5         1.01           1.02
psych-load      954.3        0.1         958.1       0.1         0.99           1.00
railsbench      1207.3       0.9         1210.5      1.3         0.99           1.00
rubocop         62.1         5.6         62.3        5.7         0.98           1.00
ruby-lsp        68.9         0.8         72.2        1.2         0.97           0.96
sequel          33.9         0.5         33.6        0.3         1.01           1.01
protoboeuf      26.2         0.2         26.2        0.1         0.97           1.00
30k_methods     352.8        0.2         360.6       0.2         0.98           0.98
fib             20.2         0.6         19.8        0.2         1.02           1.02
```

Three benchmarks are sped up by 2%, so I think the result is still positive. However, `ruby-lsp` slowed down a lot this time, so I'll look into the generated code.

maximecb:
Sidenote: I feel like we should add a row for the geomean of perf results at the bottom of the yjit-bench output, just to give a better idea of aggregate perf difference :)

The results do look generally positive 👍 

In terms of things to potentially improve: I find it surprising that protoboeuf is not sped up given that it has big synthetic methods. Also a bit unfortunate about 30k_methods since it has lots of relatively simple calls.

One thing that may help perf and reduce the number of versions is to see if it might be possible to add better biasing heuristics so that the same variables generally end up in the same registers. Sometimes may be possible to have things "line up" better in different versions. For that it would be worth looking at where things don't line up and why.

I don't want to make your life difficult. If you want to merge this PR you have my approval. I just don't want us to give up on register allocation improvements. I believe it will be possible to improve the perf.

maximecb:
One additional note tho: ruby-lsp and lobsters are some of our bigger benchmarks and they are slowed down a bit. @XrXr could you check if there is an increase in instruction cache misses and if that has anything to do with the slowdown?

k0kubun:
So I've just rebased this PR from the latest master, and I can no longer reproduce the slowdown in those benchmarks. ruby-lsp had a speedup before a past rebase, so it's not surprising that it's getting a speedup again. 

```
before: ruby 3.4.0dev (2024-08-27T23:15:37Z master f2ac013009) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-08-27T23:51:25Z yjit-local-args bffbbd124f) +YJIT [x86_64-linux]

lobsters  641.6        2.5         643.5       1.6         0.97           1.00
ruby-lsp  88.5         2.5         84.8        2.8         0.98           1.04
```

> If you want to merge this PR you have my approval. I just don't want us to give up on register allocation improvements.

I'll merge this and profile benchmarks with Ruby master again to find next optimization targets.

PR #11283, opened 2024-07-31
YJIT: Decouple Context from encoding details

k0kubun:
This PR refactors `local_types`, `temp_mapping_kind`, `temp_payload`, and `inline_block` of `Context` into rich types.

Those fields are currently using types with their encoding in mind, but since it's now handled by `encode_into`/`decode_from`, we could use richer types there without compromising the heap memory usage. 

Debugging my patches in the last few weeks, I find it so inconvenient that we can't just dump `Context` to see what's supposed to be in stack temps and locals. I'd like to merge this change so that we can easily use the `Debug` trait to inspect a `Context` when necessary.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -660,8 +660,8 @@ fn verify_ctx(jit: &JITState, ctx: &Context) {
         let stack_val = jit.peek_at_stack(ctx, i as isize);
         let val_type = Type::from(stack_val);
 
-        match learned_mapping.get_kind() {
-            TempMappingKind::MapToSelf => {
+        match learned_mapping {
+            TempMapping::MapToSelf => {
                 if self_val != stack_val {
                     panic!(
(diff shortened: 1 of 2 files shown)


maximecb:
> Debugging my patches in the last few weeks, I find it so inconvenient that we can't just dump Context to see what's supposed to be in stack temps and locals. I'd like to merge this change so that we can easily use the Debug trait to inspect a Context when necessary.

That is a fair concern. Thank you for taking the time to do this refactoring. It is a nice simplification of the code 🙏 

I took the time to get the YJIT stats on lobsters for the context encoding with/without this PR out of an abundance of caution:
```
master:
  context_data_bytes:          232,663
  context_cache_bytes:          49,152
  num_contexts_encoded:        134,237
  bytes_per_context:              1.73

This PR:
  context_data_bytes:          230,343
  context_cache_bytes:          65,536
  num_contexts_encoded:        134,183
  bytes_per_context:              1.72
```

Small increase in the context cache size but this should be negligible on production deployments. Looks like everything is working as it should 👍 🙂

PR #11286, opened 2024-07-31
Automated release package publisher

hsbt:
We should reduce amount of time for release workflow.

I would like to connect some of current release workflow. This PR makes the following chain:

1. The maintainer tagged like `v3_3_5` in this repo.
2. `publish.yml` (new) notify `draft-release.yml` at `ruby/actions`
3. <del>https://github.com/ruby/actions/pull/73 notify `release.yml` (new) at this repo when the all of tests are passed.</del>
4.  The maintainer confirm draft packages. After that, they can invoke `release.yml` with `workflow_dispatch`.
5. `release.yml` trigger `tool/release.sh` with modified name like `3.3.5` from 1st step.
6. `release.yml` trigger `update_index` at `ruby/actions`.
7. `release.yml` also trigger `ruby/docker-images` with `3.3.5`

@k0kubun @nagachika How about this? I have bit of concerns about reproducible build. If we invoke `draft-release.yml` after release work, It will update the current release package. I'm not sure `draft-release.yml` always build same package anytime.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,8 +15,9 @@ jobs:
 
       - name: Convert tag name to dot from underscore
         run: |
-            # vX_Y_Z to X.Y.Z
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
:heart:

PR #11288, opened 2024-07-31
Fix leak in warning of duplicate keys when Ripper#warn jumps

peterzhu2118:
Based on #11287.

For example, the following code leaks:

```ruby
class MyRipper < Ripper
  def initialize(src, &blk)
    super(src)
    @blk = blk
  end

  def warn(msg, *args) = @blk.call(msg)
end

$VERBOSE = true
def call_parse = MyRipper.new("if true\n  end\n") { |msg| return msg }.parse

10.times do
  500_000.times do
    call_parse
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

    34832
    51952
    69760
    88048
    105344
    123040
    141152
    159152
    176656
    194272

After:

    18400
    20256
    20272
    20272
    20272
    20304
    20368
    20368
    20368
    20400

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -543,6 +543,8 @@ struct parser_params {
     rb_ast_t *ast;
     int node_id;
 
+    st_table *warn_duplicate_keys_table;
+
     int max_numparam;
     ID it_id;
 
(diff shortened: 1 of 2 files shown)


parse.y
```diff
@@ -14717,16 +14719,17 @@ warn_duplicate_keys(struct parser_params *p, NODE *hash)
...
                           "key %+"PRIsWARN" is duplicated and overwritten on line %d",
                           nd_value(p, head), WARN_I(nd_line(head)));
             }
-            st_insert(literal_keys, (st_data_t)key, (st_data_t)hash);
+            st_insert(p->warn_duplicate_keys_table, (st_data_t)key, (st_data_t)hash);
         }
         hash = next;
     }
-    st_free_table(literal_keys);
+    st_free_table(p->warn_duplicate_keys_table);
```
nobu, line 14731: Isn't the local variable better here?

nobu (approved):

parse.y
peterzhu2118: I'm not sure what you mean by this, do you mean setting a local variable `literal_keys` and then setting `warn_duplicate_keys_table` to be `literal_keys` and then using `literal_keys`?

```c
st_table *literal_keys = st_init_table_with_size(&literal_type, RNODE_LIST(hash)->as.nd_alen / 2);
p->warn_duplicate_keys_table = literal_keys;
```

PR #11293, opened 2024-08-01
Implement String#append_bytes(String)

casperisfine:
[Feature #20594]

A handy method to construct a string out of multiple chunks.

Contrary to `String#concat`, it only takes a single argument and doesn't do any encoding negociation. The string content is appended, and the receiver encoding is preserved.

It's the caller responsibility to check for `String#valid_encoding?` in cases where it's needed.

cc @nirvdrum

### DIFF
--- a/spec/ruby/core/string/append_bytes_spec.rb
+++ b/spec/ruby/core/string/append_bytes_spec.rb
@@ -1,0 +1,33 @@
+require_relative '../../spec_helper'
+
+describe "String#append_bytes" do
+  ruby_version_is "3.4" do
+    it "allows creating broken strings" do
+      str = +"hello"
+      str.append_bytes("\xE2\x82")
+      str.valid_encoding?.should == false
+
+      str.append_bytes("\xAC")
+      str.valid_encoding?.should == true
+
+      str = "abc".encode(Encoding::UTF_32LE)
+      str.append_bytes("def")
+      str.encoding.should == Encoding::UTF_32LE
+      str.valid_encoding?.should == false
+    end
+
+    it "never changes the receiver encoding" do
+      str = "".b
+      str.append_bytes("€")
+      str.encoding.should == Encoding::BINARY
+    end
+
+    it "only accepts strings, and doesn't attempt to cast with #to_str" do
+      to_str = mock("to_str")
+      to_str.should_not_receive(:to_str)
+
+      str = +"hello"
+      -> { str.append_bytes(to_str) }.should raise_error(TypeError, "wrong argument type MockObject (expected String)")
+    end
+  end
+end
--- a/string.c
+++ b/string.c
@@ -3658,6 +3658,10 @@ static inline bool
 str_combine_coderange_fastpath(VALUE str, VALUE str2)
 {
     int str_cr = ENC_CODERANGE(str);
+    if (RB_LIKELY(str_cr == ENC_CODERANGE_UNKNOWN)) {
+        return true;
+    }
+
     int str2_cr = ENC_CODERANGE(str2);
     int str_enc = ENCODING_GET_INLINED(str);
 
@@ -3665,7 +3669,6 @@ str_combine_coderange_fastpath(VALUE str, VALUE str2)
     if (RB_LIKELY(str_enc == ENCINDEX_ASCII_8BIT)) {
         switch (str_cr) {
           case ENC_CODERANGE_VALID:
-          case ENC_CODERANGE_UNKNOWN:
             return true;
           case ENC_CODERANGE_7BIT:
             return str2_cr == ENC_CODERANGE_7BIT;
@@ -3685,12 +3688,6 @@ str_combine_coderange_fastpath(VALUE str, VALUE str2)
         }
       case ENC_CODERANGE_7BIT:
         return str2_cr == ENC_CODERANGE_7BIT;
-      case ENC_CODERANGE_UNKNOWN:
-        return true;
-      case ENC_CODERANGE_BROKEN:
-        // rb_str_coderange_scan_restartable isn't capable of repairing a coderange.
-        ENC_CODERANGE_CLEAR(str);
-        return true;
     }
     return false;
 }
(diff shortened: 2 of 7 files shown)


nirvdrum (commented):
This looks a good implementation of the proposal from https://bugs.ruby-lang.org/issues/20594.

test/ruby/test_string.rb
```diff
@@ -3630,6 +3630,46 @@ def test_bytesplice
...
+    assert_equal S("helloこんにちは\xE2\x82"), buf
+    assert_equal Encoding::UTF_8, buf.encoding
+    refute_predicate buf, :valid_encoding?
+
+    buf.append_bytes(S("\xAC".b))
+    assert_equal S("helloこんにちは€"), buf
+    assert_equal Encoding::UTF_8, buf.encoding
+    assert_predicate buf, :valid_encoding?
+  end
+
```
nirvdrum, line 3672: I think it'd be good to have a test for at least one of the UTF-16 or UTF-32 variants. Based on my reading of the code, I think something like the following should fail:

```ruby
buf = S("abc".encode("UTF-16LE"))
buf.append_bytes("def")
refute_predicate buf, :valid_encoding?
```

spec/ruby/core/string/append_bytes_spec.rb
```diff
@@ -0,0 +1,28 @@
+require_relative '../../spec_helper'
+
+describe "String#append_bytes" do
+  ruby_version_is "3.4" do
+    it "allow creating broken strings" do
```
nirvdrum, line 5:
```suggestion
    it "allows creating broken strings" do
```

spec/ruby/core/string/append_bytes_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+      str.append_bytes("\xAC")
+      str.valid_encoding?.should == true
+    end
+
+    it "never changes the receiver encoding" do
+      str = "".b
+      str.append_bytes("€")
+      str.encoding.should == Encoding::BINARY
+    end
+
```
nirvdrum, line 19: It would probably be good to have a UTF-16 or UTF-32 spec as well.

```ruby
it "handles non-ASCII strings" do
  str = "abc".encode("UTF-16LE")
  str.append_bytes("def")
  str.encoding.should == Encoding::UTF_16LE
  str.valid_encoding?.should == false
end
```

string.c
```diff
@@ -3630,6 +3630,55 @@ rb_str_concat_multi(int argc, VALUE *argv, VALUE str)
...
+
+    // Fast path for common cases
+    switch (str_cr) {
+      case ENC_CODERANGE_VALID:
+        if (str2_cr == ENC_CODERANGE_VALID) {
+            if (ENCODING_GET_INLINED(str) == ENCODING_GET_INLINED(str2)) {
+                return str;
+            }
+        }
+        else if (str2_cr == ENC_CODERANGE_7BIT) {
```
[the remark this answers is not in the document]
casperisfine: Good catch.

`str_enc_fastpath` is made for that, I'll add it.

string.c
```diff
@@ -3630,6 +3630,55 @@ rb_str_concat_multi(int argc, VALUE *argv, VALUE str)
...
+static VALUE
+rb_str_append_bytes(VALUE str, VALUE str2)
+{
+    Check_Type(str2, T_STRING);
+
+    long str2_len = RSTRING_LEN(str2);
+
+    if (RB_UNLIKELY(str2_len == 0)) {
+        return str;
+    }
```
maximecb, line 3642: Nitpick, but is this test needed for correctness? If not, might as well remove it, because why would people do this?

string.c
```diff
@@ -3630,6 +3630,55 @@ rb_str_concat_multi(int argc, VALUE *argv, VALUE str)
     return str;
 }
 
+static VALUE
+rb_str_append_bytes(VALUE str, VALUE str2)
```
maximecb, line 3634: May want to add a comment explaining what this function does.

string.c
```diff
@@ -3630,6 +3630,55 @@ rb_str_concat_multi(int argc, VALUE *argv, VALUE str)
...
+            return str;
+        }
+        break;
+      case ENC_CODERANGE_UNKNOWN:
+        return str;
+      case ENC_CODERANGE_BROKEN:
+        // rb_str_coderange_scan_restartable isn't capable of repairing a coderange.
+        ENC_CODERANGE_CLEAR(str);
+        return str;
+    }
```
maximecb, line 3672: Sorry if I'm being ignorant here, but why do we need to have any coderange logic in this function? I thought this function would just preserve/ignore the first operand's encoding?

casperisfine: No worries.

Encoding and coderange are two related, but different things. If you append or remove bytes from a string, you MUST update it's coderage. The cheapest way of course it to always reset it to `ENC_CODERANGE_UNKNOWN`, but if you know the coderange of both operands, in many cases you can cheaply compute the resulting coderange without having to scan the string.

We could totally decide to always set it to unknown, it's simple and fast, however if the coderange has to be scanned latter, it will be more costly. So it's a tradeoff.

Also I probably went a bit too far here, maybe it's only work to check for the very common cases and just fallback to either scaning or clearing in other cases.

string.c
casperisfine: Yeah, it's a public method, I'm planning to add doc.

string.c
casperisfine: It's not required no. As for why people do this, well, generally when you combine strings together you are not checking if any of them is empty:

```ruby
user.city + user.address_line_1 + user.address_line_2 + etc
```

But yeah, we could decide it's the rare case and remove that one check.

string.c
casperisfine: I wrote an alternative version that uses a lookup table for the common and simple/fast cases and fallback to just unknown for others: https://github.com/Shopify/ruby/commit/bfc6eae390fc78c17563f75cac65a8c47214b406

maximecb (approved):

maximecb (approved):

casperisfine:
Closing in favor of https://github.com/ruby/ruby/pull/11552

PR #11296, opened 2024-08-02
Make sure to always use the right `warn` in `bundled_gems.rb`

deivid-rodriguez:
Depending on scope, we may end up not calling the right `warn` method. I guess an explicit receiver can fix such case?

Fixes https://github.com/rubygems/rubygems/issues/7914.


KostyaSha:
Could this be backported to 3.3.x?

hsbt:
@KostyaSha I consider to backport this and https://github.com/ruby/ruby/pull/11322 for Ruby 3.3

PR #11302, opened 2024-08-02
[DOC] Tweaks to Array#+

BurdetteLamar:


array.c
```diff
@@ -4904,15 +4904,19 @@ rb_ary_fill(int argc, VALUE *argv, VALUE ary)
...
  *  followed by all elements of +other_array+:
  *
  *    a = [0, 1] + [2, 3]
  *    a # => [0, 1, 2, 3]
  *
- *  Related: #concat.
+ *  Related:
+ *
+ *  - Array#concat: returns the concatenation of +self+ and multiple other arrays.
+ *  - Array#*: returns the concatenation of multiple copies of +self+.
```
peterzhu2118, line 4918: Are we changing the "related" section to have descriptions now rather than just listing similar methods? Since this is supposed to be for quick reference, I find that more information here is distracting and makes it difficult to use.

I find that most of the methods listed here is self-explanatory. If the user wants more information, they should refer to the documentation of the method instead.

For example, Java's documentation only lists related methods without descriptions:

Screenshot 2024-08-06 at 3 42 07 PM

BurdetteLamar: I don't feel strongly about this, but it's not always obvious how the related methods differ.  In particular, sometimes here in Array an argument is one array, sometimes multiple arrays.  So I've thought it helpful (sometimes) to explicate a little.

I'll simplify if you think it best.

BurdetteLamar: Simplified.

peterzhu2118 (approved):

PR #11321, opened 2024-08-07
Tweak codegen for rb_yield() calls landing in ISeqs

XrXr:
Unlike in older revisions in the year, GCC 11 isn't inlining the call
to vm_push_frame() inside invoke_iseq_block_from_c() anymore. We do
want it to be inlined since rb_yield() speed is fairly important.
Logs from -fopt-info-optimized-inline reveal that GCC was blowing its
code size budget inlining invoke_block_from_c_bh() into its various
callers, leaving suboptimal code for its body.

Take away some uses of the `inline` keyword and merge a common tail
call to vm_exec() for overall better code.

This tweak gives about 18% on a micro benchmark and 1% on the
chunky-png benchmark from yjit-bench. I tested on a Skylake server.

```
$ cat c-to-ruby-call.yml
benchmark:
  - 0.upto(10_000_000) {}

$ benchmark-driver --chruby '+patch;master' c-to-ruby-call.yml
0.upto(10_000_000) {}      2.299 i/s -       3.000 times in 1.304689s (434.90ms/i)
                          +patch      master
0.upto(10_000_000) {}      2.299       1.943 i/s -       6.000 times in 2.609393s 3.088353s

Comparison:
             0.upto(10_000_000) {}
               +patch:         2.3 i/s
               master:         1.9 i/s - 1.18x  slower

$ ruby run_benchmarks.rb --chruby 'master;+patch' chunky-png
<snip>

----------  -----------  ----------  -----------  ----------  --------------  -------------
bench       master (ms)  stddev (%)  +patch (ms)  stddev (%)  +patch 1st itr  master/+patch
chunky-png  1156.1       0.1         1142.2       0.2         1.01            1.01
----------  -----------  ----------  -----------  ----------  --------------  -------------
```

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -1509,7 +1509,7 @@ rb_binding_add_dynavars(VALUE bindval, rb_binding_t *bind, int dyncount, const I
 
 /* C -> Ruby: block */
 
-static inline VALUE
+static inline void
 invoke_block(rb_execution_context_t *ec, const rb_iseq_t *iseq, VALUE self, const struct rb_captured_block *captured, const rb_cref_t *cref, VALUE type, int opt_pc)
 {
     int arg_size = ISEQ_BODY(iseq)->param.size;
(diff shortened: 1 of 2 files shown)


maximecb:
Unsure why the compilations workflow failed but seems like a sensible PR 👍

XrXr:
> linking miniruby
gcc-13: fatal error: Killed signal terminated program lto1
compilation terminated.
lto-wrapper: fatal error: gcc-13 returned 1 exit status

It's probably getting OOM killed because of the flatten attribute I added. Will need to find another way. Already tried `always_inline` on vm_push_frame, but then that triggers a different error:

```
error: inlining failed in call to ‘always_inline’ ‘vm_push_frame_inlined’: --param inline-unit-growth limit reached
  441 |         vm_push_frame_inlined(rb_execution_context_t *ec, const rb_iseq_t *iseq,
```

PR #11322, opened 2024-08-07
lib/bundled_gems.rb: more reliable caller detection

casperisfine:
The `2` skipped frames went out of sync and now it should be `3`.

Rather than just update the offset, we can implement a way that is adaptative as long as all require decorators are also called require.

Also we should compute the corresponding `uplevel` otherwise the warning will still point decorators.

Using the script reported at: https://github.com/ruby/ruby/pull/10351#issuecomment-2271725423

I also improved the message slightly:

```
$ ruby /tmp/bundled.rb 
/Users/byroot/.gem/ruby/head/gems/childprocess-5.0.0/lib/childprocess.rb:7: warning: /opt/rubies/head/lib/ruby/3.4.0+0/logger.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
Also please contact the author of childprocess-5.0.0 to request adding logger into its gemspec.
```

I'd really like to add a test for this, but I really don't see how I could...

@mattbrictson @hsbt

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -66,7 +66,7 @@ module Gem::BUNDLED_GEMS
       kernel_class.send(:alias_method, :no_warning_require, :require)
       kernel_class.send(:define_method, :require) do |name|
         if message = ::Gem::BUNDLED_GEMS.warning?(name, specs: spec_names) # rubocop:disable Style/HashSyntax
-          Kernel.warn message, :uplevel => 1
+          Kernel.warn message, uplevel: ::Gem::BUNDLED_GEMS.uplevel
         end
         kernel_class.send(:no_warning_require, name)
       end
@@ -76,6 +76,24 @@ module Gem::BUNDLED_GEMS
         kernel_class.send(:public, :require)
       end
     end
+  end
+
+  def self.uplevel
+    uplevel = 0
+    require_found = false
+    Thread.each_caller_location(2) do |cl|
+      uplevel += 1
+      if require_found
+        if cl.base_label != "require"
+          return uplevel
+        end
+      else
+        if cl.base_label == "require"
+          require_found = true
+        end
+      end
+    end
+    1
   end
 
   def self.find_gem(path)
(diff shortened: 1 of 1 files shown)


lib/bundled_gems.rb
```diff
@@ -168,7 +194,7 @@ def self.build_message(gem)
           end
         end
         if caller_gem
-          msg += " Also contact author of #{caller_gem} to add #{gem} into its gemspec."
+          msg += "\nAlso please contact the author of #{caller_gem} to request adding #{gem} into its gemspec."
```
mattbrictson, line 197: For some reason, when using this branch I am not seeing the "please contact the author" message when I run the reproduction script from <https://github.com/ruby/ruby/pull/10351#issuecomment-2271725423>.

```
$ ruby -v
ruby 3.4.0dev (2024-08-07T09:05:43Z improve-bundled-gems e4e2ab05f0) [x86_64-darwin23]
$ ruby bundled.rb
/Users/mbrictson/.rbenv/versions/ruby-dev/lib/ruby/gems/3.4.0+0/gems/childprocess-5.0.0/lib/childprocess.rb:7: warning: /Users/mbrictson/.rbenv/versions/ruby-dev/lib/ruby/3.4.0+0/logger.rb was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
```

mattbrictson (commented):
I confirmed on my machine that this PR fixes the problem I reported in #10351, but I am not sure if the solution is working entirely as intended; see comment below.

lib/bundled_gems.rb
```diff
@@ -78,6 +78,24 @@ def self.replace_require(specs)
     end
   end
 
+  def self.uplevel
+    uplevel = 0
+    require_found = false
+    Thread.each_caller_location(2) do |cl|
```
hsbt, line 84:
```suggestion
    frame_to_skip = 3
    Thread.each_caller_location do |cl|
      if frames_to_skip >= 1
        frames_to_skip -= 1
        next
      end
```

hsbt:
@casperisfine Thanks.

I merged this at https://github.com/ruby/ruby/commit/7594a292b19cb3bde971da294930f6bb2b4e4bc1 and https://github.com/ruby/ruby/commit/7e0910a82c81137731b5fd553bc312177d69d6a4.

I update you patch with Ruby 3.3 support because I want to backport `bundled_gems.rb` to `ruby_3_3`. Please let me know, my change is wrong.

casperisfine:
Thank you @hsbt

PR #11324, opened 2024-08-07
CI: Run annocheck for libruby.so

junaruga:
This PR is a rework for the https://github.com/ruby/ruby/pull/11123. I tested in my forked repository, and confirmed the test-annocheck case passing properly testing both `ruby` and `libruby.so`. Below is the CI log.

https://github.com/junaruga/ruby/actions/runs/10284090286/job/28459376922#step:10:106

---

When building with the `--shared` option, most functionality is kept in `libruby.so`. Therefore also run annocheck for `libruby.so`.

Use `ARG IN_DIR` to propagate multiple files into the container instead of `ARG FILES` in the `Dockerfile-copy`. Because the `COPY` syntax in Dockerfile doesn't support copying the multiple files.
https://stackoverflow.com/questions/61599384/specify-multiple-files-in-arg-to-copy-in-dockerfile

Co-authored-by: Vít Ondruch <vondruch@redhat.com>

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1666,8 +1666,8 @@ no-test-bundler-parallel:
 # The annocheck supports ELF format binaries compiled for any OS and for any
 # architecture. It is designed to be independent of the host OS and the
 # architecture. The test-annocheck.sh requires docker or podman.
-test-annocheck: $(PROGRAM)
-	$(tooldir)/test-annocheck.sh $(PROGRAM)
+test-annocheck: $(PROGRAM) $(LIBRUBY_SO)
+	$(tooldir)/test-annocheck.sh $(PROGRAM) $(LIBRUBY_SO)
 
 GEM = up
 sync-default-gems:
--- a/tool/test-annocheck.sh
+++ b/tool/test-annocheck.sh
@@ -14,7 +14,7 @@ set -x
 
 DOCKER="$(command -v docker || command -v podman)"
 TAG=ruby-fedora-annocheck
-TOOL_DIR=$(dirname "${0}")
+TOOL_DIR="$(dirname "${0}")"
 TMP_DIR="tmp/annocheck"
 DOCKER_RUN_VOLUME_OPTS=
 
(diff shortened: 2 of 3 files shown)


tool/test-annocheck.sh
```diff
@@ -27,7 +28,13 @@ else
   # volume in container in container on GitHub Actions
   # <.github/workflows/compilers.yml>.
   TAG="${TAG}-copy"
-  "${DOCKER}" build --rm -t "${TAG}" --build-arg=FILES="${*}" -f ${TOOL_DIR}/annocheck/Dockerfile-copy .
+  rm -rf "${TMP_DIR}"
+  mkdir -p "${TMP_DIR}"
+  for file in ${@}; do
```
nobu, line 33:
```suggestion
  for file in "${@}"; do
```

junaruga: I fixed as the suggested change on the latest rebase.

common.mk
```diff
@@ -1667,7 +1667,7 @@ no-test-bundler-parallel:
 # architecture. It is designed to be independent of the host OS and the
 # architecture. The test-annocheck.sh requires docker or podman.
 test-annocheck: $(PROGRAM)
```
junaruga, line 1669: I forgot to add the `$(LIBRUBY_SO)` at this line. I will add it.

junaruga: I added the `$(LIBRUBY_SO)` on the latest rebase.

junaruga:
I confirmed that the test-annocheck passed, checking the `ruby` and `libruby.so` on this PR's CI.

https://github.com/ruby/ruby/actions/runs/10285685378/job/28464546087?pr=11324#step:10:100

voxik:
Just FTR, I would love to see if this was resolved:

~~~
Hardened: libruby.so.3.4.0: skip: cf-protection test because mixed Rust and C code - control flow protection is needed but not yet supported by Rust 
~~~

Because that actually means Ruby is not properly hardened. But that has separate tracker:

https://bugs.ruby-lang.org/issues/20642

PR #11330, opened 2024-08-07
[DOC] Add remark about in-brief for method doc

BurdetteLamar:


jeremyevans (approved):
I'm OK with this.  I think you should have empty lines between different types of examples in an "in-brief" section.  The way the `Array#[]=` "in-brief" section is documented now is a bit hard to read.

BurdetteLamar:
> I'm OK with this. I think you should have empty lines between different types of examples in an "in-brief" section. The way the `Array#[]=` "in-brief" section is documented now is a bit hard to read.

Good idea; thanks.

peterzhu2118 (approved):

PR #11332, opened 2024-08-08
Call  method when a timeout occurs while waiting for the worker's response

ono-max:
Currently, Launchable cannot capture the timeout error of waiting for worker's response (Link: https://github.com/ruby/ruby/actions/runs/10240923674/job/28328470385?pr=11271#step:30:4). It's because `record` method is not called in this case. I've added the process of calling `record` method before killing the worker in this PR.


ono-max:
https://github.com/ruby/ruby/pull/11338 is an alternative solution.

PR #11337, opened 2024-08-08
string.c: avoid costly encoding checks in str_ensure_byte_pos when possible

casperisfine:
Profiling shows byteslice spend quite a bit of time ensuring character boundaries. The costly part is to fetch the `rb_encoding`, which we can avoid most of the time when the encoding and coderange combinaison indicate that we're only dealing with single bytes.

Before:
Capture d’écran 2024-08-08 à 11 15 47

After:
Capture d’écran 2024-08-08 à 11 16 00

There is likely a few other places where this fast path could be used.

### DIFF
--- a/string.c
+++ b/string.c
@@ -150,6 +150,12 @@ str_enc_fastpath(VALUE str)
     }
 }
 
+static inline bool
+str_enc_single_byte_fastpath(VALUE str)
+{
+    return ENCODING_GET_INLINED(str) == ENCINDEX_ASCII_8BIT || ENC_CODERANGE(str) == ENC_CODERANGE_7BIT;
+}
+
 #define TERM_LEN(str) (str_enc_fastpath(str) ? 1 : rb_enc_mbminlen(rb_enc_from_index(ENCODING_GET(str))))
 #define TERM_FILL(ptr, termlen) do {\
     char *const term_fill_ptr = (ptr);\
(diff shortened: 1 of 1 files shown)


casperisfine:
I found a way to eliminate the remaining `rb_enc_get_index` calls in `bytesplice` https://github.com/ruby/ruby/pull/11337/commits/0d4a9ea81441329afc73dc4ed1219ab67713d0a7

Capture d’écran 2024-08-08 à 11 51 04

string.c
```diff
@@ -150,6 +150,12 @@ str_enc_fastpath(VALUE str)
     }
 }
 
+static inline bool
+str_enc_single_byte_fastpath(VALUE str)
```
nirvdrum, line 154: This looks very similar to `single_byte_optimizable`, which already fast-paths `ENC_CODERANGE_7BIT`. Would adding a fast-path check for `ENCINDEX_ASCII_8BIT` there be sufficient? That could have wide reaching positive impact.

string.c
```diff
@@ -6595,10 +6607,13 @@ rb_str_bytesplice(int argc, VALUE *argv, VALUE str)
     }
     str_check_beg_len(str, &beg, &len);
     str_check_beg_len(val, &vbeg, &vlen);
-    enc = rb_enc_check(str, val);
     str_modify_keep_cr(str);
+
+    if (RB_UNLIKELY(ENCODING_GET_INLINED(str) != ENCODING_GET_INLINED(val))) {
```
nirvdrum, line 6612: There's a semantic difference here, although I'm not sure if it really matters. Previously, if the encodings were incompatible the receiver was untouched. Now `str_modify_keep_cr` is being called in all cases which would reify shared strings.

nirvdrum (commented):
I like the change overall. Encoding negotiation is an expensive operation that I did a lot of work to optimize in TruffleRuby. Ultimately, there, we're able to use an inline cache to treat the encoding as if it were constant and that allows subsequent code to optimize based on that assumption. I don't see an opportunity like that for the interpreter other than reducing the number of times having to hit the encoding lookup table, which is what this PR does. It'd be interesting if we could do something in YJIT longer-term.

string.c
byroot: Right. The worst case scenario is the reciever  is made independent but is unmodified from a user point of view. I think it's acceptable.

string.c
byroot: I like that idea. It means that on a miss we fetch the encoding twice, but that's probably OK.

PR #11338, opened 2024-08-08
Launchable: Capture Timeout error

ono-max:


KJTsanaktsidis:
This makes sense, but I wonder how hard it would be to actually make it call `#record` instead in this case? That would make _more_ sense, I think...

PR #11339, opened 2024-08-08
str_independent: add a fastpath with a single flag check

casperisfine:
If we assume that most strings we modify are not frozen and are independent, then we can optimize this case by replacing multiple flag checks by a single mask check.

Profiled on top of: https://github.com/ruby/ruby/pull/11337

Before:

Capture d’écran 2024-08-08 à 12 19 12

After:

Capture d’écran 2024-08-08 à 12 20 00

We can see:

  - `str_modify_keep_cr` `1.3%` -> `0.5%`
  - `rb_str_modify` `1.2%` -> `0.2%` (outside screenshot)

### DIFF
--- a/string.c
+++ b/string.c
@@ -2500,32 +2500,39 @@ rb_check_lockedtmp(VALUE str)
     }
 }
 
+#define STR_UNMODIFIABLE_MASK (FL_FREEZE | STR_TMPLOCK | STR_CHILLED)
 static inline void
 str_modifiable(VALUE str)
 {
-    if (CHILLED_STRING_P(str)) {
-        CHILLED_STRING_MUTATED(str);
+    if (RB_UNLIKELY(FL_ANY_RAW(str, STR_UNMODIFIABLE_MASK))) {
+        if (CHILLED_STRING_P(str)) {
+            CHILLED_STRING_MUTATED(str);
+        }
+        rb_check_lockedtmp(str);
+        rb_check_frozen(str);
     }
-    rb_check_lockedtmp(str);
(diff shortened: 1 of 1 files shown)


string.c
```diff
@@ -2500,32 +2500,39 @@ rb_check_lockedtmp(VALUE str)
...
     }
 }
 
+#define STR_DEPENDANT_MASK (STR_UNMODIFIABLE_MASK | STR_SHARED | STR_NOFREE)
 static inline int
 str_independent(VALUE str)
 {
-    str_modifiable(str);
-    return !str_dependent_p(str);
+    if (RB_UNLIKELY(FL_ANY_RAW(str, STR_DEPENDANT_MASK))) {
```
XrXr, line 2531: This is missing the STR_EMBED_P check.

XrXr:
It looks like both clang and gcc have trouble lowering these down to simple flag checks, so this is definitely the right direction.

string.c
byroot: Yeah, this is meant as a fast path, so we try to cheaply check the common case to bail out early.

And I don't think we can combine `STR_EMBED_P` in a single bitmask like this, so I just let it go in the "slowpath".  I'll add a comment.

XrXr: Makes sense to me now writing out the truth table:

```
original !str_dependent_p
embed shared nofree
0     0      0        = 1
0     0      1        = 0
0     1      0        = 0
0     1      1        = 0
1     0      0        = 1
1     0      1        = 1
1     1      1        = 1
```

when !shared && !nofree, it's always true.

PR #11341, opened 2024-08-08
rb_setup_fake_ary: use precomputed flags

casperisfine:
Setting up the fake array is a bit more expensive than would be expected because `rb_ary_freeze` does a lot of checks and lookup a shape transition.

If we assume fake arrays will always be frozen, we can precompute the flags state and just assign it.

Profiled on top of https://github.com/ruby/ruby/pull/11339

Before:

Capture d’écran 2024-08-08 à 14 22 59

After:

Capture d’écran 2024-08-08 à 14 23 14

### DIFF
--- a/array.c
+++ b/array.c
@@ -892,16 +892,27 @@ rb_ary_free(VALUE ary)
     }
 }
 
-VALUE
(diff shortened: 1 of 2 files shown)


byroot:
Closing in favor of https://github.com/ruby/ruby/pull/11344, because reasons.

PR #11347, opened 2024-08-08
YJIT: Allow tracing fallback counters

k0kubun:
This PR adds the capability to trace a specific fallback counter with `--yjit-trace-exits=counter`. This is useful when you want to know which code has megamorphic callsites, using `--yjit-trace-exits=send_megamorphic`, for example.

In the previous version https://github.com/ruby/ruby/pull/9890, the feature was limited to counters given to `gen_counted_exit`. Now it's expanded to all counters given to `gen_counter_incr`. Some code, e.g. shared trampolines, doesn't have PC, so such code uses `gen_counter_incr_without_pc` and isn't traced by the option.


XrXr (approved):

maximecb (approved):
Clever 👌 

Might be good to update `yjit.md` to explain that this is now possible.

k0kubun:
> Might be good to update yjit.md to explain that this is now possible.

:+1: https://github.com/ruby/ruby/pull/11347/commits/2ae89f988be5e9d2611255ddf06b8d693333666b

PR #11357, opened 2024-08-10
Delete unnecessary rubocop disable comment

obregonia1:
Delete rubocop disable comments that are no longer needed due to the changes made in below commit.
https://github.com/ruby/ruby/commit/83bdf127b8ddbb64c941bb9c467c686140b1696b


nobu:
Before the commit, 214f6d6598be moved that code to lib/bundled_gems.rb that does not use rubocop.

PR #11362, opened 2024-08-11
[MinGW] Revert soname changes

raedrizqie:
- only i386-ucrt soname is changed to fix building on x86 clang
- fix detection of x86intrin.h on x86 system
- mingw does not have LIBRUBY_SONAME


raedrizqie:
thank you

PR #11363, opened 2024-08-11
prism_compiler: re-use the iseq local table for keyword->table

byroot:
That's what the regular compiler does.

Would be surprised if it works, but if it does we could reclaim that table pointer in all keywords.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -8828,9 +8828,8 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
             const VALUE default_values = rb_ary_hidden_new(1);
             const VALUE complex_mark = rb_str_tmp_new(0);
 
-            ID *ids = xcalloc(keywords_list->size, sizeof(ID));
-
-            size_t kw_index = 0;
+            // ID *ids = xcalloc(keywords_list->size, sizeof(ID));
+            // size_t kw_index = 0;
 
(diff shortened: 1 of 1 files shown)


byroot:
cc @kddnewton nothing urgent, but perhaps this is something you understand better than me.

With the old compiler, `keyword->table` seem to always points inside the `iseq->body->local_table`:

  - https://github.com/ruby/ruby/blob/436d23f80da20ab56ebe42da4f37fc5c3e276f69/compile.c#L1986
  - https://github.com/ruby/ruby/blob/436d23f80da20ab56ebe42da4f37fc5c3e276f69/compile.c#L11794

But with `prism_compile.c` it's always an independently allocated region: https://github.com/ruby/ruby/blob/436d23f80da20ab56ebe42da4f37fc5c3e276f69/prism_compile.c#L8831

This draft tries to make the behavior match, but I'm struggling to make it work. If it worked I think we could reclaim that pointer.

kddnewton:
@byroot so I addressed this in https://github.com/ruby/ruby/pull/11912. I _mostly_ fixed it. I tried to change `iseq.c` in the free function to `rb_bug` where it checks `->table` but there was still some kind of place where the pointer wasn't set properly. It's not coming from Prism anymore at least. @tenderlove and I also paired on this and fixed it for loading from binary. I _think_ it's coming from the `rb_*` family of functions that load the iseqs, but I couldn't figure out where. I can keep looking at it, but I wanted to let you know the status.

byroot:
thank you ❤️

PR #11358, opened 2024-08-10
[MinGW] Fix building on i386-ucrt

raedrizqie:

PR #11364, opened 2024-08-12
Revert change of "mingw-ucrt" platform string

larskanis:
... of commit 00176cd40fe9f385231e9c20b956fc4a84d240b9 .

The reverted change was made only for constistency, as discussed in
  https://github.com/ruby/ruby/pull/11358#issuecomment-2282222369

But the platform string "mingw-ucrt" should not be changed. It is used as RUBY_PLATFORM and as the rubygems platform, so that there should be a good reason to change the name of an established platform. "mingw-ucrt" was introduced intentionally in commit 576b2e64cdc5ea42ad345dd3c1c215e006c06fca .

Related to #11358

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -4424,7 +4424,7 @@ AS_IF([test "${universal_binary-no}" = yes ], [
     AC_DEFINE_UNQUOTED(RUBY_PLATFORM, "universal." RUBY_PLATFORM_CPU "-" RUBY_PLATFORM_OS)
 ], [
     AS_IF([test "${target_os}-${rb_cv_msvcrt}" = "mingw32-ucrt" ], [
-        arch="${target_cpu}-${target_os}-ucrt"
+        arch="${target_cpu}-mingw-ucrt"
     ], [
         arch="${target_cpu}-${target_os}"
     ])


raedrizqie:
ah.. looking at your commit https://github.com/ruby/ruby/commit/576b2e64cdc5ea42ad345dd3c1c215e006c06fca i think i now understand why i386-ucrt fails to build the test gems

raedrizqie:
looking at the MinGW (UCRT64) CI job https://github.com/ruby/ruby/actions/runs/10350961146/job/28648444330?pr=11364
```
Configuration summary for ruby version 3.4.0

   * Installation prefix: /.
   * exec prefix:         ${prefix}
   * arch:                x64-mingw32
   * site arch:           x64-msvcrt
   * RUBY_BASE_NAME:      ruby
   * enable shared:       yes
   * ruby lib prefix:     ${libdir}/${RUBY_BASE_NAME}
   * site libraries path: ${rubylibprefix}/${sitearch}
   * vendor path:         ${rubylibprefix}/vendor_ruby
   * target OS:           mingw32
   * compiler:            x86_64-w64-mingw32-cc
   * with thread:         win32
   * with coroutine:      win64
   * with shared GC:      no
   * enable shared libs:  yes
   * dynamic library ext: so
```
shouldn't
```
   * arch:                x64-mingw32
   * site arch:           x64-msvcrt
```
supposed to be:
```
   * arch:                x64-mingw-ucrt
   * site arch:           x64-ucrt
```

MSP-Greg:
@raedrizqie

Currently, the CI that is labelled 'MinGW (UCRT64)' is not actually building a UCRT Ruby.  It's relying on setup-ruby to setup it's build environment, but it's loading a mingw Ruby.  See https://github.com/ruby/ruby/pull/11015.

Both Lar's [head build](https://github.com/oneclick/rubyinstaller2-packages/actions/runs/10254913348/job/28370696691#step:8:819) and the ruby-loco [ucrt build](https://github.com/MSP-Greg/ruby-loco/actions/runs/10330428222/job/28599456850#step:9:671) correctly show:
```text
   * Installation prefix: D:/a/ruby-loco/ruby-loco/ruby-ucrt
   * exec prefix:         ${prefix}
   * arch:                x64-mingw-ucrt
   * site arch:           x64-ucrt
```

MSP-Greg:
@raedrizqie

JFYI, long ago (2.3?), Windows Rubies were built with MSYS, and that kind of fell apart.  Lars and I were active in converting to MSYS2, improving Windows testing, etc.

Lars' code builds all the 'standard' Windows Ruby releases since Ruby 2.4 (https://rubyinstaller.org/), and we've both had repos that generate stand-alone builds of Ruby master for years.

All these builds are available in setup-ruby in GitHub Actions when running Windows jobs.  The Ubuntu & macOS builds are done by repos in the Ruby organization.

PR #11370, opened 2024-08-13
Implement Process.argv

byroot:
[Feature #6648]

Return a array of strings repesenting the original argv received by ruby.

Need to add a spec and documention for the method, but this is meant to get the discussion going first and foremost.


process.c
```diff
@@ -3011,6 +3012,31 @@ rb_execarg_fail(VALUE execarg_obj, int err, const char *errmsg)
...
+ *
+ *  Returns the list of original command line arguments passed to the
+ *  Ruby executable, while +ARGV+ are the arguments for the Ruby program.
+ *
+ *  Example:
+ *
+ *      $ ruby -e 'p Process.argv'
+ *      ["ruby", "-e", "p Process.argv"]
+ *
+ *  The arguments are encoded using the filesytem encoding.
```
eregon, line 3027: I think it should be the locale encoding instead since we basically want the terminal's encoding.

process.c
```diff
@@ -3011,6 +3012,31 @@ rb_execarg_fail(VALUE execarg_obj, int err, const char *errmsg)
...
+ *  call-seq:
+ *     Process.argv    -> array
+ *
+ *  Returns the list of original command line arguments passed to the
+ *  Ruby executable, while +ARGV+ are the arguments for the Ruby program.
+ *
+ *  Example:
+ *
+ *      $ ruby -e 'p Process.argv'
+ *      ["ruby", "-e", "p Process.argv"]
```
eregon, line 3025: I wonder if we should include `argv[0]`.
My main concern is it's a mistake to use that to reexecute, `argv[0]` can be set to anything, `RbConfig.ruby`/`/proc/self/exe`/etc is more reliable.
So I think we should exclude it to avoid that mistake.

I feel it's also more useful to separate Ruby flags from program arguments (there are none in this example, would be clearer if there are some what's the behavior about them), because one might want to use the same flags but run a different script/file.

process.c
byroot: I suppose it's debatable. Command line arguments are pretty intimately tied to file system paths, hence why I did it this way. Another option is just BINARY.

eregon: It should be the same encoding that `ARGV` uses, which I'd think is the locale encoding, but I haven't checked:
```
$ ruby -e 'p ARGV.first.encoding' é
#<Encoding:UTF-8>
```
On Linux/macOS IIRC the filesystem encoding is just the same as the locale encoding for CRuby, but might differ on other platforms.

byroot: `ARGV` uses `external_str_new_cstr`.

eregon: Alright, seems fine then (sorry for the very late reply).

process.c
eregon: In other words, for `ruby -e 'p Process.argv'` I think the method should return `[]`
and for e.g `ruby -v --yjit -e 'p Process.argv'` it should return `["-v", "--yjit"]`.
So like `vm_args` in the original issue description.

PR #11379, opened 2024-08-14
Simplify how finalizers are ran at shutdown

peterzhu2118:
We don't need to build a linked list from the finalizer table and instead we can just run the finalizers by iterating the ST table.

This also improves the performance at shutdown, for example:

```ruby
1_000_000.times.map do
  o = Object.new
  ObjectSpace.define_finalizer(o, proc { })
  o
end
```

Before:

    Time (mean ± σ):      1.722 s ±  0.056 s    [User: 1.597 s, System: 0.113 s]
    Range (min … max):    1.676 s …  1.863 s    10 runs

After:

    Time (mean ± σ):      1.538 s ±  0.025 s    [User: 1.437 s, System: 0.093 s]
    Range (min … max):    1.510 s …  1.586 s    10 runs

### DIFF
--- a/gc/default.c
+++ b/gc/default.c
@@ -3195,24 +3195,6 @@ gc_abort(void *objspace_ptr)
     gc_mode_set(objspace, gc_mode_none);
 }
 
-struct force_finalize_list {
-    VALUE obj;
-    VALUE table;
-    struct force_finalize_list *next;
-};
-
-static int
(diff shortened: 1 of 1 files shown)


peterzhu2118:
This requires a fix in Tempfile in https://github.com/ruby/tempfile/pull/31

PR #11384, opened 2024-08-15
[DOC] Tweaks to Array#assoc

BurdetteLamar:
@Anyone:  What is the motivation for having this method?  (It seems pretty specialized.)

### DIFF
--- a/array.c
+++ b/array.c
@@ -5041,17 +5041,17 @@ rb_ary_times(VALUE ary, VALUE times)
 
 /*
  *  call-seq:
- *    array.assoc(obj) -> found_array or nil
+ *    assoc(object) -> found_array or nil
  *
- *  Returns the first element in +self+ that is an +Array+
- *  whose first element <tt>==</tt> +obj+:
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -5041,17 +5041,17 @@ rb_ary_times(VALUE ary, VALUE times)
 
 /*
  *  call-seq:
- *    array.assoc(obj) -> found_array or nil
+ *    assoc(object) -> found_array or nil
  *
- *  Returns the first element in +self+ that is an +Array+
- *  whose first element <tt>==</tt> +obj+:
+ *  Returns the first element in +self+ that is an array
+ *  whose first element <tt>==</tt> +object+:
```
peterzhu2118, line 5047: I think this is more consistent with how the documentation of the other methods are written:

```suggestion
 *  Returns the first element +ele+ in +self+ such that +ele+ is an array
 *  and <tt>ele[0] == object</tt>:
```

peterzhu2118 (approved):

PR #11388, opened 2024-08-16
YJIT: Encode doubles to VALUE objects and move stat generation to rust

rwstauner:
This adds a flonum_from_double to match the fixnum_from_usize function.
Stats that can now be generated from rust were moved there and
the ruby code is now a test to ensure we're getting the same values.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1623,6 +1623,21 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  def test_runtime_stats_types
+    assert_compiles(<<~'RUBY', exits: :any, result: true)
+      def test = :ok
+      3.times { test }
+
+      stats = RubyVM::YJIT.runtime_stats
+      return true unless stats[:all_stats]
+
+      [
+        stats[:object_shape_count].is_a?(Integer),
+        stats[:ratio_in_yjit].is_a?(Float),
+      ].all?
+    RUBY
+  end
+
   private
 
   def code_gc_helpers
--- a/yjit.c
+++ b/yjit.c
@@ -1165,8 +1165,8 @@ rb_yjit_invalidate_all_method_lookup_assumptions(void)
 }
 
 // Number of object shapes, which might be useful for investigating YJIT exit reasons.
-static VALUE
-object_shape_count(rb_execution_context_t *ec, VALUE self)
+VALUE
+rb_object_shape_count(void)
 {
     // next_shape_id starts from 0, so it's the same as the count
     return ULONG2NUM((unsigned long)GET_SHAPE_TREE()->next_shape_id);
--- a/yjit.rb
+++ b/yjit.rb
@@ -162,27 +162,6 @@ module RubyVM::YJIT
     stats[:object_shape_count] = Primitive.object_shape_count
     return stats unless Primitive.rb_yjit_stats_enabled_p
 
-    side_exits = total_exit_count(stats)
-    total_exits = side_exits + stats[:leave_interp_return]
-
-    # Number of instructions that finish executing in YJIT.
-    # See :count-placement: about the subtraction.
-    retired_in_yjit = stats[:yjit_insns_count] - side_exits
-
-    # Average length of instruction sequences executed by YJIT
-    avg_len_in_yjit = total_exits > 0 ? retired_in_yjit.to_f / total_exits : 0
-
-    # Proportion of instructions that retire in YJIT
-    total_insns_count = retired_in_yjit + stats[:vm_insns_count]
-    yjit_ratio_pct = 100.0 * retired_in_yjit.to_f / total_insns_count
-    stats[:total_insns_count] = total_insns_count
-    stats[:ratio_in_yjit] = yjit_ratio_pct
-
-    # Make those stats available in RubyVM::YJIT.runtime_stats as well
-    stats[:side_exit_count]  = side_exits
-    stats[:total_exit_count] = total_exits
-    stats[:avg_len_in_yjit]  = avg_len_in_yjit
-
     stats
   end
 
@@ -437,7 +416,7 @@ module RubyVM::YJIT
     end
 
     def print_sorted_exit_counts(stats, out:, prefix:, how_many: 20, left_pad: 4) # :nodoc:
-      total_exits = total_exit_count(stats)
+      total_exits = stats[:side_exit_count]
 
       if total_exits > 0
         exits = []
@@ -462,14 +441,6 @@ module RubyVM::YJIT
       else
         out.puts "total_exits:           " + format_number(13, total_exits)
       end
-    end
-
-    def total_exit_count(stats, prefix: "exit_") # :nodoc:
-      total = 0
-      stats.each do |k,v|
-        total += v if k.start_with?(prefix)
-      end
-      total
     end
 
     def print_counters(counters, out:, prefix:, prompt:, optional: false) # :nodoc:
--- a/yjit.rb
+++ b/yjit.rb
@@ -156,13 +156,7 @@ module RubyVM::YJIT
   # Return a hash for statistics generated for the `--yjit-stats` command line option.
   # Return `nil` when option is not passed or unavailable.
   def self.runtime_stats()
-    stats = Primitive.rb_yjit_get_stats()
-    return stats if stats.nil?
-
-    stats[:object_shape_count] = Primitive.object_shape_count
-    return stats unless Primitive.rb_yjit_stats_enabled_p
-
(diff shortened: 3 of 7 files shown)


yjit.rb
```diff
@@ -162,27 +162,6 @@ def self.runtime_stats()
     stats[:object_shape_count] = Primitive.object_shape_count
```
k0kubun, line 162: If the motivation of this change was to generate the stats Hash in a single place, we probably want to move this to the Rust function as well.

yjit/src/cruby.rs
```diff
@@ -544,6 +545,75 @@ impl VALUE {
         let k: usize = item.wrapping_add(item.wrapping_add(1));
         VALUE(k)
     }
+
+    // Logic from internal/numeric.h rb_float_new_inline.
+    pub fn flonum_from_double(item: f64) -> Self {
```
k0kubun, line 550: This seems much more involved than `fixnum_from_usize` to maintain a duplicated implementation. How about making a bindgen to `rb_float_new` (which is defined in `numeric.c`. In yjit.c, you could `#undef rb_float_value` and declare `double rb_float_value(VALUE v);` to reuse it from the bindgen) and call it?

test/ruby/test_yjit.rb
```diff
@@ -1623,6 +1623,40 @@ def entry = self.class
     RUBY
   end
 
+  def test_runtime_stat_floats
```
k0kubun, line 1626: This is something useful to test this PR, but I personally feel this kind of test ends up being a burden in the future, given the fact that it's essentially duplicating the implementation and you'll be forced to replicate the implementation whenever you touch it.

I would test only that those keys are encoded as a Float object by `is_a?(Float)`, or just omit this test.

yjit/src/cruby.rs
maximecb: One thing is that ideally we probably want to avoid allocating on the Ruby heap while inside YJIT, just to reduce potential failure modes (eg triggering GC while YJIT is running).

I'd be OK with reusing the C implementation but ideally we would be able to encode to a flonum rather than allocating if possible.

yjit/src/stats.rs
```diff
@@ -724,6 +724,14 @@ fn rb_yjit_gen_stats_dict() -> VALUE {
         }
     }
 
+    macro_rules! hash_aset_flonum {
+        ($hash:ident, $key:expr, $value:expr) => {
+            let key = rust_str_to_sym($key);
+            let value = VALUE::flonum_from_double($value);
+            rb_hash_aset($hash, key, value);
+        }
+    }
```
maximecb, line 733: Seems like this could be a function rather than a macro? but it's a good idea to encapsulate this functionality 👍

yjit/src/cruby.rs
rwstauner: I mentioned in the meeting this week that I had it working with `rb_float_new`.
It was discussed that doing it without the C function would be best.
Maybe it would look less complicated if I removed some of the comments I added. 😄 
Or maybe feelings are different after seeing the algorithm. 😅 

The way CRuby does it is to encode these if they are within the MIN/MAX range, and anything outside of that needs to be on the heap.
It seems likely that YJIT won't need to use anything outside those bounds, so one option would be (like the `fixnum_from_usize` above) to just assert that we are within range.  Then we wouldn't need the C function and there would be no path for allocating.

If we don't want to duplicate this logic then yes we could use `rb_float_new` which has the same logic as this function: Encode values within range and create an object on the heap for anything else.

test/ruby/test_yjit.rb
rwstauner: Indeed, that seems wise.  I also compared before and after diffs of the stat output, so I'm confident that the code change has produced no unexpected changes in the result.
Once we merge this, we might decide to change something about the result and wouldn't need this test hindering us at that point 👍

yjit/src/stats.rs
rwstauner: I did it that way because that's what was done for fixnums right above it. 🤷 
That does seem sensible to me though. I'm happy to convert both of them to functions.

yjit/src/cruby.rs
k0kubun: > I'd be OK with reusing the C implementation but ideally we would be able to encode to a flonum rather than allocating if possible.
>
> I mentioned in the meeting this week that I had it working with rb_float_new.
> It was discussed that doing it without the C function would be best.

I'm not sure if I follow your logic. It seems like you assume that using the C implementation `rb_float_new` doesn't use flonum encoding, but it encodes a flonum whenever this Rust `flonum_from_double` would use the flonum encoding.

Similarly, it seems like you're saying this Rust implementation `flonum_from_double` doesn't allocate a Float object in the heap, but it does whenever `rb_float_new` would.

So, as long as you exactly replicate what the interpreter does, I don't think rewriting it in Rust changes anything with regard to "allocating on the Ruby heap" or "potential failure modes". If you're OK with reusing the C implementation, what is the motivation to duplicate the logic?

maximecb: I did not look at the implementation of `rb_float_new`. I assumed that it allocated because of `_new` in the name. If it does use flonum encoding when possible, then I agree it makes sense to use the C function. It's very unlikely a float will ever be allocated.

rwstauner: This rust function is a direct port of `rb_float_new`: it does the transmute and encodes it if it's within range.  Out of range it falls back to `rb_float_new_in_heap`.

https://github.com/ruby/ruby/blob/002e785e7756e5105910434586a5752c8772bb2f/internal/numeric.h#L244-L273

When I saw the other stuff in this file about "not linking against CRuby in tests" I thought that was another reason we wanted it this way.

Is there any value in dropping the `rb_float_new_in_heap` call from here and just asserting that any floats yjit is using are in range?
Or is it best to just use `rb_float_new` and not duplicate any logic?

k0kubun: > When I saw the other stuff in this file about "not linking against CRuby in tests" I thought that was another reason we wanted it this way.

We can't test Ruby's Float values in `cargo test`, so you'd need to test the result in `make test-all` anyway (if you want). So "not linking against CRuby in tests" doesn't help in that case. 

> Is there any value in dropping the rb_float_new_in_heap call from here and just asserting that any floats yjit is using are in range?

Leaving a `panic` for `#[cfg(test)]` is fine, but I don't think it's a good idea to potentially crash `RubyVM::YJIT.runtime_stats`. Remember, for the stats implementation, we are already allocating a Hash object using a C function. Nothing really changes even if you make sure there's no GC or C call for making a Float value. I still don't get what you're trying to solve with it.

> Or is it best to just use rb_float_new and not duplicate any logic?

I think so.

yjit.c
```diff
@@ -1165,8 +1165,8 @@ rb_yjit_invalidate_all_method_lookup_assumptions(void)
 }
 
 // Number of object shapes, which might be useful for investigating YJIT exit reasons.
-static VALUE
-object_shape_count(rb_execution_context_t *ec, VALUE self)
+VALUE
+object_shape_count(void)
```
rwstauner, line 1169: Does this seem like the right change?  We used to use it as a `Primitive`, now we are just calling it from rust.

yjit.c
```diff
@@ -1165,8 +1165,8 @@ rb_yjit_invalidate_all_method_lookup_assumptions(void)
 }
 
 // Number of object shapes, which might be useful for investigating YJIT exit reasons.
-static VALUE
-object_shape_count(rb_execution_context_t *ec, VALUE self)
+VALUE
+rb_object_shape_count(void)
```
rwstauner, line 1169: Does this seem like the right change? We used to use it as a Primitive, now we are just calling it from rust.

yjit/src/cruby.rs
rwstauner: I removed the implementation and linked to `rb_float_new` which simplifies a lot.

Regarding

> In yjit.c, you could #undef rb_float_value and declare double rb_float_value(VALUE v); to reuse it from the bindgen

I didn't seem to need this, if you think it's valuable, can you explain that a bit more?

k0kubun: :+1:

Since bindgen finds `rb_float_new` successfully, it seems like you don't need to undef it.

yjit.c
k0kubun: Yes, it does.

I personally prefer having an `rb_yjit_` prefix for non-static C functions in yjit.c, but existing functions don't necessarily follow the convention and I'm not sure if other files like shape.c would use this name, so the current function name seems also fine.

k0kubun (approved):

PR #11390, opened 2024-08-16
Fix assertion error when TracePoint has incompatible events

peterzhu2118:
TracePoints with incompatible events (i.e. events not in ISEQ_TRACE_EVENTS) with a method target will fail an assertion error because it does not filter for the supported events. For example, the following lines will cause an assertion error:

    def foo; end
    # No arguments passed into TracePoint.new enables all ISEQ_TRACE_EVENTS
    TracePoint.new {}.enable(target: method(:foo))
    # Raise is not supported with a target
    TracePoint.new(:raise, :return) {}.enable(target: method(:foo))
    foo

Crashes with:

    Assertion Failed: vm_insnhelper.c:7026:vm_trace:(iseq_local_events & ~(0x0001 | 0x0002 | 0x0004 | 0x0008 | 0x0010| 0x0020| 0x0040 | 0x0100 | 0x0200 | 0x4000 | 0x010000| 0x020000)) == 0

### DIFF
--- a/vm_trace.c
+++ b/vm_trace.c
@@ -1342,7 +1342,7 @@ void
 rb_hook_list_connect_tracepoint(VALUE target, rb_hook_list_t *list, VALUE tpval, unsigned int target_line)
 {
     rb_tp_t *tp = tpptr(tpval);
-    rb_event_hook_t *hook = alloc_event_hook((rb_event_hook_func_t)tp_call_trace, tp->events, tpval,
+    rb_event_hook_t *hook = alloc_event_hook((rb_event_hook_func_t)tp_call_trace, tp->events & ISEQ_TRACE_EVENTS, tpval,
                                              RUBY_EVENT_HOOK_FLAG_SAFE | RUBY_EVENT_HOOK_FLAG_RAW_ARG);
     hook->filter.target_line = target_line;
     hook_list_connect(target, list, hook, FALSE);
--- a/test/ruby/test_settracefunc.rb
+++ b/test/ruby/test_settracefunc.rb
@@ -2304,7 +2304,7 @@ CODE
     _c = a + b
   end
 
-  def check_with_events *trace_events
+  def check_with_events(trace_point_events, expected_events = trace_point_events)
     all_events = [[:call, :method_for_enable_target1],
                   [:line, :method_for_enable_target1],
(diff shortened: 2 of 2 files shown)


vm_trace.c
```diff
@@ -1342,7 +1342,7 @@ void
 rb_hook_list_connect_tracepoint(VALUE target, rb_hook_list_t *list, VALUE tpval, unsigned int target_line)
 {
     rb_tp_t *tp = tpptr(tpval);
-    rb_event_hook_t *hook = alloc_event_hook((rb_event_hook_func_t)tp_call_trace, tp->events, tpval,
+    rb_event_hook_t *hook = alloc_event_hook((rb_event_hook_func_t)tp_call_trace, tp->events & ISEQ_TRACE_EVENTS, tpval,
```
XrXr, line 1345: Since this is can be used by the public rb_tracepoint_new() in the future, could you leave a comment about how the filtering is done here because rb_tracepoint_new() doesn't support targeting stuff yet and how rb_hook_list_connect_tracepoint() is only used when targeting stuff? Alternatively, maybe there is more specific place to do this filtering.

peterzhu2118: I not sure I follow what you mean.

`rb_hook_list_connect_tracepoint` requires a iseq or bmethod, so it cannot target anything except the events in `ISEQ_TRACE_EVENTS`. The events created in `rb_tracepoint_new` could be either internal or external events, but not both. In the case where the TracePoint created in `rb_tracepoint_new` is enabled with a target, this code applies. Otherwise, this code doesn't apply.

XrXr: I'm a bit surprised that hook list implies iseq or bmethod. Without looking at other code, I expected a global hook list to be a possibility. Which why I suggest leaving a comment.

peterzhu2118: A global hook list is created using `rb_add_event_hook2` when the TracePoint is enabled without a target specified. The function `rb_hook_list_connect_tracepoint` is used only for when there is a target iseq/bmethod. Perhaps the function names could be improved....

XrXr: Okay, I don't want to block this fix over a clarity concern. I can rename and/or add docs separately later. Thanks!

PR #11393, opened 2024-08-16
Add error checking to readdir, telldir, and closedir calls in dir.c

jeremyevans:
Raise SystemCallError exception when these functions return an error.

This changes behavior for the following case (found by the tests):

```ruby
dir1 = Dir.new('..')
dir2 = Dir.for_fd(dir1.fileno)
dir1.close
dir2.close
```

The above code is basically broken, as `dir1.close` closed the file
descriptor.  The subsequent `dir2.close` call is undefined behavior.
When run in isolation, it raises Errno::EBADF after the change, but
if another thread opens a file descriptor between the `dir1.close`
and `dir2.close` calls, the `dir2.close` call could close the file
descriptor opened by the other thread.  Raising an exception is much
better in this case as it makes it obvious there is a bug in the code.

For the readdir check, since the GVL has already been released,
reacquire it rb_thread_call_with_gvl if an exception needs to be
raised.

Due to the number of closedir calls, this adds static close_dir_data
and check_closedir functions.  The close_dir_data takes a
struct dir_data * and handles setting the dir entry to NULL regardless
of failure.

Fixes [Bug #20586]

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -571,6 +571,25 @@ opendir_without_gvl(const char *path)
         return opendir(path);
 }
 
+static void
+close_dir_data(struct dir_data *dp)
+{
+    if (dp->dir) {
+        if (closedir(dp->dir) < 0) {
+            dp->dir = NULL;
+            rb_sys_fail("closedir");
+        }
+        dp->dir = NULL;
+    }
+}
+
+static void
+check_closedir(DIR *dirp)
+{
+    if (closedir(dirp) < 0)
+        rb_sys_fail("closedir");
+}
+
 static VALUE
 dir_initialize(rb_execution_context_t *ec, VALUE dir, VALUE dirname, VALUE enc)
 {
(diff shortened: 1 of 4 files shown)


dir.c
```diff
@@ -810,10 +828,22 @@ fundamental_encoding_p(rb_encoding *enc)
...
+
 static void *
 nogvl_readdir(void *dir)
 {
-    return readdir(dir);
+    errno = 0;
+    if ((dir = readdir(dir)) == NULL) {
+        if (errno)
+            rb_thread_call_with_gvl(sys_failure, (void *)"readdir");
+    }
```
nobu, line 845: This looks to make the check in `dir_read` unneeded.

And maybe this needs `rb_errno_set` and `rb_errno` for M:N thread scheduler.

```suggestion
    rb_errno_set(0);
    if ((dir = readdir(dir)) == NULL) {
        if (rb_errno())
            rb_thread_call_with_gvl(sys_failure, (void *)"readdir");
    }
```

PR #11397, opened 2024-08-17
[DOC] Tweaks for Array#clear

BurdetteLamar:


BurdetteLamar:
@peterzhu2118, there's some conditional code here that I don't understand;  are the examples adequate?

Also 

```
if (!ARY_EMBED_P(ary)) {
            rb_ary_unshare(ary);
            FL_SET_EMBED(ary);
            ARY_SET_EMBED_LEN(ary, 0);
        }
```
has no `else` clause;  is it possible that the array is not cleared?

peterzhu2118:
> has no else clause; is it possible that the array is not cleared?

Yeah, it's not possible for an array to be both embedded and shared (it would be a really bad bug if it was!). We already assert that the array cannot be shared and embedded:

https://github.com/ruby/ruby/blob/124f1754dfd7a53ad1f80d92b8c369f7d1cafd1e/internal/array.h#L86

I opened https://github.com/ruby/ruby/pull/11422 to fix this.

peterzhu2118 (approved):

PR #11398, opened 2024-08-17
[DOC] Tweaks for Array#collect

BurdetteLamar:


BurdetteLamar:
@peterzhu2118, I have used `map` in the example (as being the more common usage), but that's the alias.  Should we say `collect` instead?

peterzhu2118 (approved):

PR #11402, opened 2024-08-18
Skip some tests which don't work under permissionless containers

KJTsanaktsidis:
When running as UID 0 but without CAP_DAC_OVERRIDE (for example, in a docker container running with --uid 0 but --cap-drop=all), these tests won't work because of hard-coded assumptions about what uid 0 can and can't do.

### DIFF
--- a/test/fileutils/test_fileutils.rb
+++ b/test/fileutils/test_fileutils.rb
@@ -93,12 +93,24 @@ class TestFileUtils < Test::Unit::TestCase
       @@no_broken_symlink
     end
 
+    def has_capsh?
+      !!system('capsh', '--print', out: File::NULL, err: File::NULL)
+    end
+
+    def has_root_file_capabilities?
+      !!system(
+        'capsh', '--has-p=CAP_DAC_OVERRIDE', '--has-p=CAP_CHOWN', '--has-p=CAP_FOWNER',
+        out: File::NULL, err: File::NULL
+      )
+    end
+
     def root_in_posix?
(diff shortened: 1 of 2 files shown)


test/fileutils/test_fileutils.rb
```diff
@@ -97,8 +97,12 @@ def root_in_posix?
...
         # FIXME: privilege if groups include root user?
         return Process.groups.include?(0)
-      end
-      if Process.respond_to?('uid')
+      elsif system('command', '-v', 'capsh', out: File::NULL)
+        return system(
+          'capsh', '--has-p=CAP_DAC_OVERRIDE', '--has-p=CAP_CHOWN', '--has-p=CAP_FOWNER',
+          out: File::NULL, err: File::NULL
+        )
+      elsif Process.respond_to?('uid')
```
nobu, line 105: I don't think you should assume `command -v` to work as expected on non-POSIX systems.

```suggestion
      end
      begin
        return system(
          'capsh', '--has-p=CAP_DAC_OVERRIDE', '--has-p=CAP_CHOWN', '--has-p=CAP_FOWNER',
          out: File::NULL, err: File::NULL, exception: true
        )
      rescue Errno::ENOENT
      end
      if Process.respond_to?('uid')
```

KJTsanaktsidis: 👍  - I refactored it a bit differently (so we could keep the same if... else if... structure in `root_in_posix?`). The cost is that it means `capsh` is executed twice if it's present but.... seems like not a big deal for a test.

PR #11403, opened 2024-08-18
[Bug #20693] Perform an actual access check in Dir.tmpdir for writability

KJTsanaktsidis:
At the moment, this code is looking at the stat output to determine if a temp directory can be written to. However, just because the stat bits say that a directory is writable, does not make it so; and, likewise, the operating system may in fact grant access to paths that the stat bits and process UID say should be inaccessible.

These systems include:

* Posix ACL's
* Linux's capabilities like CAP_DAC_OVERRIDE
* Linux Security Modules like SELinux or AppArmor
* Syscall filters like Linux's seccomp
* Granular capability systems like FreeBSD's Capsicum
* OpenBSD's pledge and unveil
* Windows too has a rich ACL system for controlling filesystem access

The best thing to do is simply to try and access the path with `File.writable?` and let the operating system tell us if the path can be accessed.

https://bugs.ruby-lang.org/issues/20693

### DIFF
--- a/lib/tmpdir.rb
+++ b/lib/tmpdir.rb
@@ -36,7 +36,9 @@ class Dir
       case
       when !stat.directory?
         warn "#{name} is not a directory: #{dir}"
-      when !stat.writable?
+      when !File.writable?(dir)
+        # We call File.writable?, not stat.writable?, because you can't tell if a dir is actually
+        # writable just from stat; OS mechanisms other than user/group/world bits can affect this.
         warn "#{name} is not writable: #{dir}"
       when stat.world_writable? && !stat.sticky?
         warn "#{name} is world-writable: #{dir}"


KJTsanaktsidis:
Hm, I'll have to make a PR in bundler to change how `File.writable?` is being stubbed.

KJTsanaktsidis:
That… seems like an openbsd bug? Surely `access(2)` should be checking whether the path is veiled? Reckon I should investigate & open an issue with them?

The pledge bit I guess makes a little more sense because it operates at a level where all it can do is kill the process…

Should we just attempt to write a file in the temp directory rather than doing an access check? I don’t know what else we can do, really

jeremyevans:
I'll check with other OpenBSD developers regarding whether `access(2)` should respect `unveil(2)` limitations.

jeremyevans:
@KJTsanaktsidis I checked and the unveil/access issue was considered a bug.  A fix has just been committed for it.  Thanks for asking about it.

mame (approved):

PR #11405, opened 2024-08-19
[Bug #20677] Don't emit ELF notes for coroutines on non-ELF platforms

KJTsanaktsidis:
Three commits in this one:

* Stop emitting ELF notes on non-ELF platforms, like MacOS. They don't actually _do_ anything and only really have a platform-defined meaning on Linux.
* There was a typo in one of the macros for generating the ELF notes on aarch64.
* The configure check for pac-ret was only looking for `aarch64`, not `arm64`, so it wasn't auto-enabling on MacOS. Fixed that.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -839,7 +839,7 @@ AS_IF([test "$GCC" = yes], [
     ])
 
     # aarch64 branch protection
-    AS_CASE(["$target_cpu"], [aarch64], [
+    AS_CASE(["$target_cpu"], [aarch64|arm64], [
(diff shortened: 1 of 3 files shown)


nobu (approved):
Although I can't old Xcode, this seems reasonable.

KJTsanaktsidis:
Yeah I couldn't actually make the build fail because of this either, but I don't have access to an intel mac. But yeah in any case MacOS does nothing with these sections so it seems sensible enough not to emit them.

peterzhu2118:
@KJTsanaktsidis This PR breaks the C level backtrace on macOS ARM machines, I've opened a revert in #11548.

PR #11414, opened 2024-08-20
[DOC] Include keywords.rdoc in doc/syntax/

BurdetteLamar:
I think this doc belongs in `doc/syntax/`.  I chose to include, rather than move, so as not to disturb any links out there in the web.


peterzhu2118 (commented):
>  I chose to include, rather than move, so as not to disturb any links out there in the web.

I would prefer to move rather than include, so we don't have inclusions and duplicate pages. The documentation in the master branch is not stable and links are not guaranteed. If a user wants a stable link, then they should point to a released version of Ruby (e.g. Ruby 3.3).

BurdetteLamar:
@peterzhu2118, I fouled my repo wrt this PR.  I think the PR is now correct (rename vs. include); if not, I'll rework somehow.

peterzhu2118 (approved):

PR #11415, opened 2024-08-20
[PRISM] Implement unused block warning

eileencodes:
Related: ruby/prism#2935

cc/ @kddnewton


kddnewton (changes requested):
Can you enable `TestISeq#test_unused_param` as a part of this PR?

Also, I'm a bit confused about the super node one. In the upstream compiler it looks like it checks if there's a block on the parent ISEQ. I'm not sure how that translates here. (Not saying your implementation is wrong, I just don't understand.)

kddnewton (approved):

PR #11429, opened 2024-08-22
automerge needs windows results

hsbt:
https://github.com/ruby/ruby/pull/11419 is merged with Windows failure. We should prevent that.


nobu:
👍 
The mswin platform is special for both the compiler, runtime, and build system.
Also it is not the bottleneck of the CI.

PR #11441, opened 2024-08-23
Enable native SOCKS support by default [Feature #20673]

znz:
Before:
Enable SOCKS support when `configure` with `--enable-socks` or `make` with `SOCKS_SERVER` environment variable.

After:
Enable SOCKS support when `configure` without `--disable-socks`.

Both:
Enabling SOCKS support needs socks library when `make`. (socks5, socksd, or socks)

### DIFF
--- a/ext/socket/extconf.rb
+++ b/ext/socket/extconf.rb
@@ -661,7 +661,7 @@ EOS
 
   $distcleanfiles << "constants.h" << "constdefs.*"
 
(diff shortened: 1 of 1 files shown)


hsbt:
From https://bugs.ruby-lang.org/issues/20673

We would like to extract `socks` support from Ruby core in the future. We shouldn't enable this by default.

PR #11445, opened 2024-08-23
Fix how Prism parses encoding from shebang args (`#!`)

amomchilov:
Fix the first of the two assertion failures here:

```ruby
TestRubyOptions#test_shebang [/Users/alex/src/github.com/Ruby/ruby/test/ruby/test_rubyoptions.rb:514]:
pid 49518 exit 0.

1. [1/2] Assertion for "stdout"
   | <["\"あ\""]> expected but was
   | <["\"\\u3042\""]>.

2. [2/2] Assertion for "stderr"
   | Expected /shebang line ending with \\r/ to match "".

Finished tests in 0.030572s, 32.7097 tests/s, 359.8064 assertions/s.
1 tests, 11 assertions, 1 failures, 0 errors, 0 skips

ruby -v: ruby 3.4.0dev (2024-08-22T14:01:55Z master 56a34b5af5) +PRISM [arm64-darwin23]
make: *** [yes-test-all] Error 1
```

(The latter was fixed in https://github.com/ruby/prism/pull/3007)

Fixes https://github.com/ruby/prism/issues/2996

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -2121,12 +2121,24 @@ prism_script_shebang_callback(pm_options_t *options, const uint8_t *source, size
     memcpy(switches, source, length);
     switches[length] = '\0';
 
+    int no_src_enc = !opt->src.enc.name;
+    int no_ext_enc = !opt->ext.enc.name;
+    int no_int_enc = !opt->intern.enc.name;
+
     moreswitches(switches, opt, 0);
     free(switches);
 
     pm_options_command_line_set(options, prism_script_command_line(opt));
-    if (opt->ext.enc.name != 0) {
+
+    if (no_src_enc && opt->src.enc.name) {
+        opt->src.enc.index = opt_enc_index(opt->src.enc.name);
         pm_options_encoding_set(options, StringValueCStr(opt->ext.enc.name));
+    }
+    if (no_ext_enc && opt->ext.enc.name) {
+        opt->ext.enc.index = opt_enc_index(opt->ext.enc.name);
+    }
+    if (no_int_enc && opt->intern.enc.name) {
+        opt->intern.enc.index = opt_enc_index(opt->intern.enc.name);
     }
 }
 


ruby.c
```diff
@@ -2127,6 +2127,7 @@ prism_script_shebang_callback(pm_options_t *options, const uint8_t *source, size
     pm_options_command_line_set(options, prism_script_command_line(opt));
     if (opt->ext.enc.name != 0) {
         pm_options_encoding_set(options, StringValueCStr(opt->ext.enc.name));
+        opt->ext.enc.index = opt_enc_index(opt->ext.enc.name);
```
amomchilov, line 2130: This is the wrong place to set this, but just a proof-of-concept

prism/prism.c
```diff
@@ -21711,7 +21711,7 @@ pm_strnstr(const char *big, const char *little, size_t big_length) {
  */
 static void
 pm_parser_warn_shebang_carriage_return(pm_parser_t *parser, const uint8_t *start, size_t length) {
-    if (length > 2 && start[length - 1] == '\n' && start[length - 2] == '\r') {
+    if (length > 2 && start[length - 1] == '\r' && start[length] == '\n') {
```
amomchilov, line 21714: FIXME: out-of-bounds access https://github.com/ruby/prism/pull/3007#pullrequestreview-2257164857

ruby.c
```diff
@@ -1298,6 +1298,9 @@ proc_e_option(ruby_cmdline_options_t *opt, const char *s, long argc, char **argv
     return n;
 }
 
+static int
+opt_enc_index(VALUE enc_name);
+
```
amomchilov, line 1303: Put this prototype here (immediately above the `proc_K_option` that uses it) just to make it compile.

Where should I actually put it? Should I just move the whole `opt_enc_index` function itself instead?

ruby.c
amomchilov: okok so this is a weird one.

When not using Prism, `opt->ext.enc.name` is set by `proc_K_option()`, but not `index`:

https://github.com/ruby/ruby/blob/efa4ec0f981ed51af82d32146f24158552eb2bd7/ruby.c#L1322-L1324

Then, later on in `load_file_internal()`, the `index` is updated from the `name`, but only if `no_ext_enc` is true:

https://github.com/ruby/ruby/blob/efa4ec0f981ed51af82d32146f24158552eb2bd7/ruby.c#L2703-L2705

I tried modifying `proc_K_option()` to set both of these, but that doesn't work because it can't access the `no_ext_enc` variable to only set it conditionally.

This split source of truth (and that they aren't always perfectly in sync) is pretty confusing!

amomchilov: Copied the `no_src_enc`/`no_ext_enc`/`no_int_enc` logic exactly as-is from `load_file_internal()`. Would be nice to eventually clean up this split source of truth.

PR #11453, opened 2024-08-24
Remove outdated documentation information from `Set.to_a`

formigarafa:
Just removing the note _"The order of elements is uncertain."_  from the documentation.

This information is no longer true since ruby 1.9, when hashes started having its elements iterated in order of their respective insertion.

The methods `delete_if` and `keep if` in this same class even go on the length of ensure the *subclasses won't break the enumeration order*.

### DIFF
--- a/lib/set.rb
+++ b/lib/set.rb
@@ -335,7 +335,7 @@ class Set
     end
   end
 
-  # Converts the set to an array.  The order of elements is uncertain.
+  # Converts the set to an array.
   #
   #     Set[1, 2].to_a                    #=> [1, 2]
   #     Set[1, 'c', :s].to_a              #=> [1, "c", :s]


nobu:
Could you report it to the upstream https://github.com/ruby/set?

formigarafa:
Yup.

mame:
I don't think this statement should be removed. Set is currently the library required by default, but may be replaced by a C implementation in future. At that time, it may be hard to continue to guarantee the order.

formigarafa:
@mame, Please correct me if I am wrong. I wondered that too, but I saw these other comments on `delete_if` and `keep_if` which made me wonder if this guarantee wasn't already been made.

> (@hash.delete_if | @hash.keep_if) should be faster, but using it breaks the order of enumeration in subclasses.
source: https://github.com/ruby/ruby/blob/2040d6b38d46f41a5185efe0fffca250d63af4e6/lib/set.rb#L540-L560

hsbt:
@knu What do you think this? I agreed @mame 's opinion.

knu:
I've updated the document and clarified the implementation notes.  See the reference above.  Thanks!

PR #11454, opened 2024-08-25
Range#step: restore legacy behavior for String ranges

zverok:
See discussion in https://bugs.ruby-lang.org/issues/18368#note-34

Of options I’ve named I implemented support for String ranges for both old and new behavior:

```ruby
# Old behavior with integer steps and #succ
('a'..).step(2).take(3)
#=> ["a", "c", "e"]
('a'..).step.take(3)
#=> ["a", "b", "c"]

# New behavior with string step and #+
('a'..).step('b').take(3)
#=> ["a", "ab", "abb"]
```

### DIFF
--- a/range.c
+++ b/range.c
@@ -309,6 +309,34 @@ range_each_func(VALUE range, int (*func)(VALUE, VALUE), VALUE arg)
     }
 }
 
+// NB: Two functions below (step_i_iter and step_i) are used only to maintain the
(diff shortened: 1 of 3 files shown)


zverok:
> You could do the same for symbols. What do you think?

@knu, sorry, it took me some time to find a minute to work on this.

Here we go: #11573

PR #11458, opened 2024-08-26
Fix test failure of daily package CI

hsbt:
Fix https://github.com/ruby/actions/actions/runs/10513692214/job/29130044376#step:19:3550

```
       Git error: command `git rev-parse --is-bare-repository` in directory /home/runner/work/actions/actions/snapshot-master/tmp/1.7/bundled_app/vendor/cache/foo-1.0-f18431e82ed6 has failed.
       fatal: not a git repository (or any of the parent directories): .git
```

This failure is only happened under that tar package.


deivid-rodriguez:
Makes sense, thank you @hsbt.

PR #11459, opened 2024-08-26
[DOC] Tweaks for Array#cycle

BurdetteLamar:


array.c
```diff
@@ -6782,13 +6782,18 @@ rb_ary_cycle_size(VALUE self, VALUE args, VALUE eobj)
...
 /*
  *  call-seq:
- *    array.cycle {|element| ... } -> nil
- *    array.cycle(count) {|element| ... } -> nil
- *    array.cycle -> new_enumerator
- *    array.cycle(count) -> new_enumerator
- *
- *  When called with positive Integer argument +count+ and a block,
- *  calls the block with each element, then does so again,
+ *    cycle {|element| ... } -> nil
```
peterzhu2118, line 6785: I don't think we need this case since the case below encompasses this one.

array.c
```diff
@@ -6782,13 +6782,18 @@ rb_ary_cycle_size(VALUE self, VALUE args, VALUE eobj)
...
+ *    cycle(count = nil) {|element| ... } -> nil
+ *    cycle -> new_enumerator
```
peterzhu2118, line 6787: Ditto

array.c

array.c
BurdetteLamar: Ditto.

peterzhu2118 (approved):

PR #11465, opened 2024-08-27
Merge URI-0.12.3 for Ruby 3.2

hsbt:
For https://github.com/ruby/uri/issues/118.

Ruby 3.4 will deprecate some methods like `RFC3986_PARSER.escape`. We should add `RFC2396_PARSER` constant for that migration with old Ruby version.


nagachika:
I will revert https://github.com/ruby/ruby/commit/3a3784a197383046537e66a9c567b96a52f0f86f (bump rexml to 3.3.6) for a while to pass test-bundled-gems.

PR #11467, opened 2024-08-27
[Bug #20701] Add tests for mutable kwargs bug

pabloh:
Adds tests for for the case a splatted Hash is mutated inside a method, after is called with alongside a splatted array followed by a positional argument.

See: https://bugs.ruby-lang.org/issues/20701


jeremyevans:
Merged as part of a3562c2a0abf1c2bdd1d50377b4f929580782594

PR #11474, opened 2024-08-27
Remove docs for Timeout::Error example in handle_interrupt

jpcamara:
The Timeout::Error example no longer works consistently

- This PR from the timeout gem (https://github.com/ruby/timeout/pull/30) made it so you have to handle_interrupt on Timeout::ExitException instead of Timeout::Error

- Efficiency changes to the gem (one shared thread) mean you can't consistently handle timeout errors using handle_timeout: https://github.com/ruby/timeout/issues/41

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -2215,30 +2215,6 @@ handle_interrupt_arg_check_i(VALUE key, VALUE val, VALUE args)
  * resource allocation code. Then, the ensure block is where we can safely
  * deallocate your resources.
  *
- * ==== Guarding from Timeout::Error
- *
- * In the next example, we will guard from the Timeout::Error exception. This
- * will help prevent from leaking resources when Timeout::Error exceptions occur
- * during normal ensure clause. For this example we use the help of the
- * standard library Timeout, from lib/timeout.rb
- *
- *   require 'timeout'
- *   Thread.handle_interrupt(Timeout::Error => :never) {
- *     timeout(10){
- *       # Timeout::Error doesn't occur here
- *       Thread.handle_interrupt(Timeout::Error => :on_blocking) {
- *         # possible to be killed by Timeout::Error
- *         # while blocking operation
- *       }
- *       # Timeout::Error doesn't occur here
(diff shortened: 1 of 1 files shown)


thread.c
```diff
@@ -2223,10 +2223,10 @@ handle_interrupt_arg_check_i(VALUE key, VALUE val, VALUE args)
  * standard library Timeout, from lib/timeout.rb
  *
  *   require 'timeout'
- *   Thread.handle_interrupt(Timeout::Error => :never) {
- *     timeout(10){
+ *   Thread.handle_interrupt(Timeout::ExitException => :never) {
+ *     Timeout.timeout(10){
  *       # Timeout::Error doesn't occur here
```
olleolleolle, line 2228: The specific error message on L2228, should it _also_ be ExitException?

jpcamara: It's a good question - I left it that way only because it still seems you still should _catch_ Timeout::Error, not ExitException.

@jeremyevans I think you made that ExitException change - what do you think is clearer here?

eregon: Mmh, I don't understand thsi example well, why is it even using `handle_interrupt` here?
It looks useless to me.
Or maybe the example needs to be more complete and actually include an `ensure` to illustrate what it wants to do.
In the current state I think it would be better to remove this confusing example.

jeremyevans: I agree with @eregon, we should remove the entire "Guarding from Timeout::Error" section.

jpcamara: Ok, I'll remove it with another commit

jpcamara:
@olleolleolle @eregon I removed the docs instead of updating them and squashed the commit

jpcamara:
@jeremyevans @eregon any chance of getting another look at this? I removed the doc entirely. Would be nice to reduce the confusion around it (I got caught up for awhile trying to make it work for my next concurrency article until I realized it was busted)

jeremyevans (approved):

PR #11476, opened 2024-08-28
[DOC] Remove Array#abbrev

BurdetteLamar:
Doesn't seem to be in master any more. (Don't know what happened to it.)

### DIFF
--- a/array.c
+++ b/array.c
@@ -8692,7 +8692,6 @@ rb_ary_deconstruct(VALUE ary)
  *    - With string argument +field_separator+, a new string that is equivalent to
  *      <tt>join(field_separator)</tt>.
  *
- *  - #abbrev: Returns a hash of unambiguous abbreviations for elements.
  *  - #pack: Packs the elements into a binary sequence.
  *  - #sum: Returns a sum of elements according to either <tt>+</tt> or a given block.
  */


adam12:
It needs to be required first (unfortunately).

```ruby
require "abbrev"

["foo", "bar"].abbrev # => {"foo"=>"foo", "fo"=>"foo", "f"=>"foo", "bar"=>"bar", "ba"=>"bar", "b"=>"bar"}
```

BurdetteLamar:
> It needs to be required first (unfortunately).
> 
> ```ruby
> require "abbrev"
> 
> ["foo", "bar"].abbrev # => {"foo"=>"foo", "fo"=>"foo", "f"=>"foo", "bar"=>"bar", "ba"=>"bar", "b"=>"bar"}
> ```

I see it documented at docs.ruby-lang.org in 3.1, 3.2, and 3.3, but not in master.  So will it be documented in 3.x, or not?

adam12:
Oh, I see! I think it's [moved from default to bundled gem](https://github.com/ruby/ruby/commit/e28bdff94140bd9884e27ff1c42c31e121e65827) so presumably it's not included in the doc generation. Good catch.

BurdetteLamar:
> Oh, I see! I think it's [moved from default to bundled gem](https://github.com/ruby/ruby/commit/e28bdff94140bd9884e27ff1c42c31e121e65827) so presumably it's not included in the doc generation. Good catch.

Thanks, @adam12.

peterzhu2118 (approved):

PR #11487, opened 2024-08-28
[DOC] Related for Array#[]

BurdetteLamar:


array.c
```diff
@@ -8567,7 +8568,7 @@ rb_ary_deconstruct(VALUE ary)
  *
  *  These methods do not modify +self+.
  *
- *  - #[]: Returns one or more elements.
+ *  - #[] (aliased as #slice): Returns consecutive elements as determined by a given argument.
```
peterzhu2118, line 8571: The other aliases are displayed as a list (e.g. `#select` and `#filter`), but this one is different. I don't have a strong opinion on how we display aliases, but can we make it consistent?

BurdetteLamar: I like this way better.  Will add something in the doc guide.

If I try to resolve these here, I will almost certainly foul my repo and branches.  Can we merge this and let me fix afterwards?

peterzhu2118: Ok, but can we be consistent in this PR and you can change it back in a future PR?

PR #11492, opened 2024-08-29
Make benchmark to bundled gems

hsbt:
https://bugs.ruby-lang.org/issues/20309

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -32,6 +32,7 @@ module Gem::BUNDLED_GEMS
     "win32ole" => "3.5.0",
     "fiddle" => "3.5.0",
     "logger" => "3.5.0",
+    "benchmark" => "3.5.0",
   }.freeze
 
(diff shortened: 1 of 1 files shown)


eregon:
https://bugs.ruby-lang.org/issues/20309 doesn't mention moving `benchmark` to bundled gems.
Was it discussed anywhere?

hsbt:
Do not rush to me. I will discuss this at next meeting.

eregon:
I didn't mean to rush you, but yes I think it would be good to have some discussion about it on a Redmine ticket.
Updating the existing ticket without a comment doesn't notify watchers of the issue or the mailing list (unfortunately).

FWIW I found out about this PR from @Earlopain who started making PRs assuming `benchmark` will become a bundled gem. I think it's a bit too early for that, before the final decision has been taken to make it a bundled gem. (but doing that work is very much appreciated, thanks!)

hsbt:
I merged this change to https://github.com/ruby/ruby/pull/11560

PR #11500, opened 2024-08-29
prism_compile.c: Fix read_entire_file() for platforms without file mmap

kateinoigakukun:
Apply similar fix to https://github.com/ruby/prism/pull/2944


XrXr:
Thank you! I was wondering what the cyrptic illegal WASM instruction is coming from, and I didn't catch the `-Wreturn-type`. Now I know to make sure preprocessor if-elsif trees always end with an `else`. 😅

PR #11508, opened 2024-08-30
Experiment: Optimize `strftime`

casperisfine:
`Time#strftime` is a hotspot for applications that do a lot of JSON serialization, typical API endpoint etc. They generally contains quite a bit of dates and `Time#iso8601` and its variant are just a `strftime` call. 

image

image

Looking at profiles, I found a few ideas, e.g. the logic to grow the string capacity seem to try to minimize the final string capacity, by using a more agressive growing strategy we can make the method roughtly 20% faster.

Then it calls a lot into `rb_str_set_len` after having written past the string terminator. That function does a lot of checks, so by using a simplified function we can optimize a bit further.

But then I can't really see a path to really get more:

```
compare-ruby: ruby 3.4.0dev (2024-08-29T13:11:40Z master 6b08a50a62) +YJIT [arm64-darwin23]
built-ruby: ruby 3.4.0dev (2024-08-30T08:52:31Z opt-strftime 32bfb1cb7f) +YJIT [arm64-darwin23]
warming up...
# Iteration per second (i/s)

|                            |compare-ruby|built-ruby|
|:---------------------------|-----------:|---------:|
|time.strftime("%FT%T")      |      1.757M|    2.165M|
|                            |           -|     1.23x|
|time.strftime("%FT%T.%3N")  |      1.529M|    2.007M|
|                            |           -|     1.31x|
|time.strftime("%FT%T.%6N")  |      1.471M|    2.004M|
|                            |           -|     1.36x|
```

Now most of the time is in `ruby_vfprintf`, which perhaps could be improved too as it repeatedly check if the string is modifiable and have enough capacity, but it's becoming hairy:

Capture d’écran 2024-08-30 à 11 16 12

Capture d’écran 2024-08-30 à 11 17 44

I think one avenue could be to open a feature request to make `xmlschema/iso8601` a core method, which would allow to implement it without using any `printf` like implementation, which would allow to know upfront the necessary string size.

I'm tempted to experiment with a gem that does a native implementation of `iso8601`, so see that the upper performance cap is like. 

FYI @ianks

### DIFF
--- a/benchmark/time_strftime.yml
+++ b/benchmark/time_strftime.yml
@@ -1,0 +1,5 @@
+prelude: |
+  # frozen_string_literal: true
+  time = Time.now
+benchmark:
+  - time.strftime("%FT%T.%6N")
--- a/benchmark/time_strftime.yml
+++ b/benchmark/time_strftime.yml
@@ -2,4 +2,6 @@ prelude: |
   # frozen_string_literal: true
   time = Time.now
 benchmark:
-  - time.strftime("%FT%T.%6N")
+  - time.strftime("%FT%T")     # 19B
+  - time.strftime("%FT%T.%3N") # 23B
+  - time.strftime("%FT%T.%6N") # 26B
--- a/internal/string.h
+++ b/internal/string.h
@@ -51,6 +51,7 @@ int rb_enc_str_coderange_scan(VALUE str, rb_encoding *enc);
 int rb_ascii8bit_appendable_encoding_index(rb_encoding *enc, unsigned int code);
 VALUE rb_str_include(VALUE str, VALUE arg);
 VALUE rb_str_byte_substr(VALUE str, VALUE beg, VALUE len);
+long rb_str_ensure_capa_for(VALUE str, long len);
 
 static inline bool STR_EMBED_P(VALUE str);
 static inline bool STR_SHARED_P(VALUE str);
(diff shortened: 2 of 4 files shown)


casperisfine:
> I'm tempted to experiment with a gem that does a native implementation of iso8601, so see that the upper performance cap is like.

So actually it's tricky because as a C ext you just get a `timespec` or `timeval`, then you need ot rely on C's `strftime` etc. So you don't have support for arbitrary large year etc.

So maybe I'll just experiment in ruby-core directly.

As for `Time#strftime`, removing more duplicated length and capacity checks I can squeeze some more performance.

```
compare-ruby: ruby 3.4.0dev (2024-08-29T13:11:40Z master 6b08a50a62) +YJIT [arm64-darwin23]
built-ruby: ruby 3.4.0dev (2024-08-30T10:00:52Z opt-strftime 895abe6fce) +YJIT [arm64-darwin23]
warming up...
# Iteration per second (i/s)

|                            |compare-ruby|built-ruby|
|:---------------------------|-----------:|---------:|
|time.strftime("%FT%T")      |      1.752M|    2.614M|
|                            |           -|     1.49x|
|time.strftime("%FT%T.%3N")  |      1.509M|    2.402M|
|                            |           -|     1.59x|
|time.strftime("%FT%T.%6N")  |      1.474M|    2.393M|
|                            |           -|     1.62x|
```

But it's a slog and very risky. I think there's a potential to speedup `strftime` by a whole lot (probably 2X if not 3X) but it requires a big refactoring to rely less on string primitive functions that constantly check the capacity, length etc. It's quite a big project and likely to introduce bugs.

casperisfine:
Arf, I meant to open that on the Shopify fork -_-.

Anyway, I experimented with bringing `xmlschema` in core, and was able to make it 4-5x faster: https://github.com/ruby/ruby/pull/11510, and that's still using `snprintf`. It's probably possible to be even faster by manually doing some integer to char conversions, given we know many of the components are validated upfront.

So closing this as I think it would be worth optimizing `strftime` but it's too hairy for me.

PR #11513, opened 2024-08-30
[DOC] Rationalize aliases in "What's Here"

BurdetteLamar:
Make compliant with the documentation guide.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -7019,10 +7019,9 @@ static const rb_data_type_t env_data_type = {
  *  - #empty?: Returns whether there are no entries.
  *  - #eql?: Returns whether a given object is equal to +self+.
  *  - #hash: Returns the integer hash code.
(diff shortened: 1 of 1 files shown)


peterzhu2118 (commented):
Many of the methods listed here are not aliased to the other, but rather the two have the same implementation. Considering that, how do you determine which one is the real name and which one is the alias?

BurdetteLamar:
> Many of the methods listed here are not aliased to the other, but rather the two have the same implementation. Considering that, how do you determine which one is the real name and which one is the alias?

I keyed off the RDoc-generated remark "Also aliased as: ...."

peterzhu2118 (approved):

peterzhu2118:
> I keyed off the RDoc-generated remark "Also aliased as: ...."

That makes sense.

PR #11515, opened 2024-08-31
Fix function name

infracaninophile:
There is an EVP_aes_128_cbc() function in openssl, but no EVP_EVP_aes_128_cbc().

To trigger this, you need an OpenSSL library compiled without DES/3DES support, or various other obsolescent crypto.


nobu:
Please report it to the upstream, https://github.com/ruby/openssl.

rhenium:
While this patch is a correct fix, I think removing these lines would be a better idea considering that it has never worked since initially written in 2003.

As @nobu says, https://github.com/ruby/openssl is the upstream for files in ext/openssl. I'm closing this PR here.

PR #11517, opened 2024-08-31
[DOC] Rationalize aliases in "What's Here"

BurdetteLamar:


string.rb
```diff
@@ -163,7 +163,7 @@
 #
 #  These instance methods make use of slicing:
 #
-#  - String#[] (also aliased as String#slice) returns a slice copied from +self+.
+#  - String#[] (aliased as String#slice) returns a slice copied from +self+.
```
peterzhu2118, line 166:
```suggestion
#  - String#[] (aliased as String#slice): returns a slice copied from +self+.
```

string.rb
```diff
@@ -488,13 +488,13 @@
...
 #  - #squeeze: Returns a copy of +self+ with contiguous duplicate characters removed.
-#  - #[], #slice: Returns a substring determined by a given index, start/length, or range, or string.
+#  - #[] (aliased as #slice): Returns a substring determined by a given index, start/length, or range, or string.
 #  - #byteslice: Returns a substring determined by a given index, start/length, or range.
 #  - #chr: Returns the first character.
 #
 #  _Duplication_
 #
-#  - #to_s, $to_str: If +self+ is a subclass of +String+, returns +self+ copied into a +String+;
+#  - #to_s (aliased as $to_str): If +self+ is a subclass of +String+, returns +self+ copied into a +String+;
```
peterzhu2118, line 497: I think the original was a typo.

```suggestion
#  - #to_s (aliased as #to_str): If +self+ is a subclass of +String+, returns +self+ copied into a +String+;
```

peterzhu2118 (approved):

PR #11519, opened 2024-08-31
[Feature #20630] Improve `String#rindex` Performance on OSX

fresh-eggs:
**Ruby Issue Tracker: [Feature [#20630](https://bugs.ruby-lang.org/issues/20709)]**

String#rindex is much slower on OSX than on Linux hosts. This appears due to the lack of a `memrchr` implementation on OSX. The fallback solution currently performs a `memcmp` on each character in the search string looking for a substring match.

https://github.com/ruby/ruby/blob/e69945fc57efba5c8895c21235e109145865952d/string.c#L4392-L4398

For hosts with no native `memrchr` implementation, this change introduces a simple `memrchr` implementation (`rb_memrchr`) that can be used instead. An example benchmark below demonstrates results on an 8000 char search string with a 10 char substring near the end:

```
ruby-master | substring near the end | osx

UTF-8
       user     system      total        real
index  0.000111   0.000000   0.000111 (  0.000110)
rindex  0.000446   0.000005   0.000451 (  0.000454)
```

```
ruby-patched | substring near the end | osx

UTF-8
       user     system      total        real
index  0.000112   0.000000   0.000112 (  0.000111)
rindex  0.000057   0.000001   0.000058 (  0.000057)
```

For additional benchmark results please see [Feature [#20630](https://bugs.ruby-lang.org/issues/20709)]

### DIFF
--- a/string.c
+++ b/string.c
@@ -4345,7 +4345,19 @@ rb_str_byteindex_m(int argc, VALUE *argv, VALUE str)
     return Qnil;
 }
 
-#ifdef HAVE_MEMRCHR
+#ifndef HAVE_MEMRCHR
+static void*
+memrchr(const char *search_str, int chr, long search_len)
+{
+    const char *ptr = search_str + search_len;
+    do {
+        if ((unsigned char)*(--ptr) == chr) return (void *)ptr;
+    } while (ptr >= search_str);
+
+    return ((void *)0);
+}
+#endif
+
 static long
 str_rindex(VALUE str, VALUE sub, const char *s, rb_encoding *enc)
 {
@@ -4362,6 +4374,10 @@ str_rindex(VALUE str, VALUE sub, const char *s, rb_encoding *enc)
     c = *t & 0xff;
     searchlen = s - sbeg + 1;
 
+    if (memcmp(s, t, slen) == 0) {
+        return s - sbeg;
+    }
+
     do {
         hit = memrchr(sbeg, c, searchlen);
         if (!hit) break;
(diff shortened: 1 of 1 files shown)


string.c
```diff
@@ -4362,8 +4376,12 @@ str_rindex(VALUE str, VALUE sub, const char *s, rb_encoding *enc)
     c = *t & 0xff;
     searchlen = s - sbeg + 1;
 
+    if (memcmp(s, t, slen) == 0) {
+      return s - sbeg;
```
nobu, line 4380:
```suggestion
        return s - sbeg;
```

string.c
```diff
@@ -4346,11 +4346,25 @@ rb_str_byteindex_m(int argc, VALUE *argv, VALUE str)
...
 
 #ifdef HAVE_MEMRCHR
+#define rb_memrchr memrchr
+#else
+static void*
+rb_memrchr(const char *search_str, int chr, long search_len)
+{
+    const char *ptr = search_str + search_len;
+    do {
+       if (*(--ptr) == chr) return (void *)ptr;
```
nobu, line 4356:
```suggestion
        if ((unsigned char)*(--ptr) == chr) return (void *)ptr;
```

string.c
```diff
@@ -4346,11 +4346,25 @@ rb_str_byteindex_m(int argc, VALUE *argv, VALUE str)
...
+    return ((void *)0);
+}
+#endif
+
 static long
 str_rindex(VALUE str, VALUE sub, const char *s, rb_encoding *enc)
 {
     char *hit, *adjusted;
-    int c;
+    char c;
```
nobu, line 4367:
```suggestion
    int c;
```

string.c
```diff
@@ -4346,11 +4346,25 @@ rb_str_byteindex_m(int argc, VALUE *argv, VALUE str)
 }
 
 #ifdef HAVE_MEMRCHR
+#define rb_memrchr memrchr
+#else
+static void*
+rb_memrchr(const char *search_str, int chr, long search_len)
```
nobu, line 4352: I guess we can define a function with the same name.

```suggestion
#ifndef HAVE_MEMRCHR
static void*
memrchr(const char *search_str, int chr, long search_len)
```

string.c
fresh-eggs: I like this approach better, thank you for the suggestion.

string.c
fresh-eggs: Agreed, cleans it up a little. Pushed this and the other changes.

string.c
```diff
@@ -4345,7 +4345,19 @@ rb_str_byteindex_m(int argc, VALUE *argv, VALUE str)
...
 
-#ifdef HAVE_MEMRCHR
+#ifndef HAVE_MEMRCHR
+static void*
+memrchr(const char *search_str, int chr, long search_len)
+{
+    const char *ptr = search_str + search_len;
+    do {
+        if ((unsigned char)*(--ptr) == chr) return (void *)ptr;
+    } while (ptr >= search_str);
```
eregon, line 4355: FYI: https://bugs.ruby-lang.org/issues/20796#note-6 https://github.com/ruby/ruby/pull/11923

PR #11536, opened 2024-09-03
YJIT: Accept key for runtime_stats to return only that stat

rwstauner:
This effectively reverts the "use functions instead of macros" commit but now there's a good reason.

I had to pull in `rb_raise` and `rb_eTypeError` to get it to work.
I wondered if `rb_yjit_get_stats` would be better implemented in `yjit.c` but then the locking would have to be done differently.
I'm happy to hear suggestions for a better way to implement and/or organize this.

### DIFF
--- a/yjit.rb
+++ b/yjit.rb
@@ -155,8 +155,12 @@ module RubyVM::YJIT
 
   # Return a hash for statistics generated for the `--yjit-stats` command line option.
   # Return `nil` when option is not passed or unavailable.
-  def self.runtime_stats()
-    Primitive.rb_yjit_get_stats
+  # If a symbol argument is provided, return only the value for the named stat.
+  # If any other type is provided, raises TypeError.
+  def self.runtime_stats(key = nil)
+    raise TypeError, "non-symbol given" unless key.nil? || Symbol === key
+
+    Primitive.rb_yjit_get_stats(key)
   end
 
   # Format and print out counters as a String. This returns a non-empty
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -650,8 +650,8 @@ pub extern "C" fn rb_yjit_print_stats_p(_ec: EcPtr, _ruby_self: VALUE) -> VALUE 
 /// Primitive called in yjit.rb.
 /// Export all YJIT statistics as a Ruby hash.
 #[no_mangle]
-pub extern "C" fn rb_yjit_get_stats(_ec: EcPtr, _ruby_self: VALUE) -> VALUE {
-    with_vm_lock(src_loc!(), || rb_yjit_gen_stats_dict())
+pub extern "C" fn rb_yjit_get_stats(_ec: EcPtr, _ruby_self: VALUE, key: VALUE) -> VALUE {
+    with_vm_lock(src_loc!(), || rb_yjit_gen_stats_dict(key))
 }
 
 /// Primitive called in yjit.rb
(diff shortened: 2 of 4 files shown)


yjit/src/stats.rs
```diff
@@ -650,8 +652,21 @@ pub extern "C" fn rb_yjit_print_stats_p(_ec: EcPtr, _ruby_self: VALUE) -> VALUE
...
 #[no_mangle]
-pub extern "C" fn rb_yjit_get_stats(_ec: EcPtr, _ruby_self: VALUE) -> VALUE {
-    with_vm_lock(src_loc!(), || rb_yjit_gen_stats_dict())
+pub extern "C" fn rb_yjit_get_stats(_ec: EcPtr, _ruby_self: VALUE, hash_or_key: VALUE) -> VALUE {
+    // Do this before taking vm lock.
+    let (hash, key): (VALUE, VALUE) = if unsafe { RB_TYPE_P(hash_or_key, RUBY_T_HASH) } {
+        (hash_or_key, Qnil)
+    } else if unsafe { RB_TYPE_P(hash_or_key, RUBY_T_SYMBOL) } {
+        (Qnil, hash_or_key)
+    } else if hash_or_key.nil_p() {
```
rwstauner, line 661: Are there better methods to use for these type tests?

yjit/src/stats.rs
```diff
@@ -650,8 +652,21 @@ pub extern "C" fn rb_yjit_print_stats_p(_ec: EcPtr, _ruby_self: VALUE) -> VALUE
...
+        (unsafe { rb_hash_new() }, Qnil)
+    } else {
+        unsafe {
+            rb_raise(rb_eTypeError, b"non-hash or symbol given\0".as_ref().as_ptr() as *const raw::c_char)
```
rwstauner, line 665: I got this from https://github.com/ruby/ruby/blob/e7fbdf8187ab62bfbe8b2660eda01de94341412e/yjit/src/yjit.rs#L104
but it probably makes sense to build some sort of wrapper to do this.

yjit.rb
```diff
@@ -155,8 +155,10 @@ def self.dump_exit_locations(filename)
 
   # Return a hash for statistics generated for the `--yjit-stats` command line option.
   # Return `nil` when option is not passed or unavailable.
-  def self.runtime_stats()
-    Primitive.rb_yjit_get_stats
+  # If arg is a hash, populate stats into it.
```
k0kubun, line 158: I get that you followed the `GC.stat` pattern, but do we really need this "If arg is a hash" interface now? As long as it populates all stats, it'd be so slow that removing a single allocation might not really matter. This seems like a relatively complicated API design, and it also seems to complicate the implementation.

yjit/src/stats.rs
k0kubun: Have you considered splitting this `Primitive` into multiple different `Primitive`s and choosing one in the Ruby method? It might be easier in Ruby than in Rust to match Ruby arguments and raise a Ruby exception while leaving other YJIT-specific stuff to Rust.

yjit.rb
rwstauner: Great, thanks! As you noted I did this for consistency but I agree just taking a key would make it a lot simpler.

yjit/src/stats.rs
```diff
@@ -872,13 +897,13 @@ fn rb_yjit_gen_stats_dict() -> VALUE {
 
         // Create a hash for the cfunc call counts
         let cfunc_calls = rb_hash_new();
-        rb_hash_aset(hash, rust_str_to_sym("cfunc_calls"), cfunc_calls);
         set_call_counts(cfunc_calls, &mut *addr_of_mut!(CFUNC_NAME_TO_IDX), &mut *addr_of_mut!(CFUNC_CALL_COUNT));
+        set_stat!(hash, "cfunc_calls", cfunc_calls);
```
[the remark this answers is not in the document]
rwstauner: Indeed, thanks! So far it just quits early if it finds the key.  This will be a good addition.

yjit/src/stats.rs
rwstauner: I was curious about that as I do think it would be simpler.
I wasn't sure if there would be any notable difference between using `is_a?` or `===` and using the `RB_TYPE_P` but if that's not important I will move that there.  Then we don't need to bring in `rb_raise`.

yjit/src/stats.rs
rwstauner: Will do, thanks!

yjit/src/stats.rs
k0kubun: Even if you don't split it, I think it would help to reject unexpected cases in Ruby. `YJIT.enable` normalizes the `stats` argument in Ruby first, for example. If we reject all not-Symbol-or-nil cases in Ruby and give `key_or_nil` to the `Primitive`, we could just use a single Primitive.

PR #11547, opened 2024-09-04
Time#strftime: don't resize the produced String

casperisfine:
Followup: https://github.com/ruby/ruby/pull/11544 and https://github.com/ruby/ruby/pull/11542

`strftime` spend a significant amount of time right sizing the produced string at the end. The benefits are debatable, on one hand it does free a bit of memory upfront, but on the other it's a lot of work and malloc churn for what's generally very minimal gains.

Especially since we can make the assumtion that such formatted string is unlikely to be held onto for very long, as the main use of formatting is to output.

It also use `rb_str_set_len` a lot, and that function does a tons of checks that we can assert are useless. So this PR also add an internal only fast version that only performs the check in debug mode.

```
compare-ruby: ruby 3.4.0dev (2024-09-04T13:23:20Z opt-strftime-coder.. fd6bac3205) +YJIT [arm64-darwin23]
built-ruby: ruby 3.4.0dev (2024-09-04T13:23:20Z opt-strftime-coder.. fd6bac3205) +YJIT [arm64-darwin23]
warming up...

|                            |compare-ruby|built-ruby|
|:---------------------------|-----------:|---------:|
|time.strftime("%FT%T")      |      2.267M|    2.732M|
|                            |           -|     1.21x|
|time.strftime("%FT%T.%3N")  |      2.026M|    2.479M|
|                            |           -|     1.22x|
|time.strftime("%FT%T.%6N")  |      2.105M|    2.491M|
|                            |           -|     1.18x|
```

Cumulative with a3f589640fd443deea230c27efd6bdfc92f5817f and 294dad22d7ac5f307d567cff897507b9fd9d05e9:

```
compare-ruby: ruby 3.4.0dev (2024-09-04T11:55:26Z opt-strftime-coder.. 29a5153cba) +YJIT [arm64-darwin23]
built-ruby: ruby 3.4.0dev (2024-09-04T15:41:42Z opt-strftime-no-re.. c711f4f004) +YJIT [arm64-darwin23]
warming up...

|                            |compare-ruby|built-ruby|
|:---------------------------|-----------:|---------:|
|time.strftime("%FT%T")      |      1.713M|    2.673M|
|                            |           -|     1.56x|
|time.strftime("%FT%T.%3N")  |      1.498M|    2.591M|
|                            |           -|     1.73x|
|time.strftime("%FT%T.%6N")  |      1.428M|    2.436M|
|                            |           -|     1.71x|
```


casperisfine:
> `strftime` spend a significant amount of time right sizing the produced string at the end.

I misread the code a bit. It's only doing this if it allows the string to be embeddable, but that's a costly check and unlikely to happen in `strftime`.

The calls inside `vsprintf` though, were to grow the string, so I need to keep those.

PR #11548, opened 2024-09-04
Fix C level backtrace for macOS ARM

peterzhu2118:
After 6a746e1bc902d4245aac58db4e9ffc2f72d79629, the C level backtrace is empty on macOS ARM machines.

```

-- Other runtime information -----------------------------------------------
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -839,7 +839,7 @@ AS_IF([test "$GCC" = yes], [
     ])
 
     # aarch64 branch protection
-    AS_CASE(["$target_cpu"], [aarch64|arm64], [
+    AS_CASE(["$target_cpu"], [aarch64], [
 	AS_FOR(option, opt, [-mbranch-protection=pac-ret -msign-return-address=all], [
             # Try these flags in the _prepended_ position - i.e. we want to try building a program
             # with CFLAGS="-mbranch-protection=pac-ret $CFLAGS". If the builder has provided different
--- a/configure.ac
+++ b/configure.ac
@@ -839,19 +839,39 @@ AS_IF([test "$GCC" = yes], [
     ])
 
     # aarch64 branch protection
-    AS_CASE(["$target_cpu"], [aarch64], [
-	AS_FOR(option, opt, [-mbranch-protection=pac-ret -msign-return-address=all], [
-            # Try these flags in the _prepended_ position - i.e. we want to try building a program
-            # with CFLAGS="-mbranch-protection=pac-ret $CFLAGS". If the builder has provided different
-            # branch protection flags in CFLAGS, we don't want to overwrite those. We just want to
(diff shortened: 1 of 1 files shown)


KJTsanaktsidis:
@peterzhu2118 Thanks for catching this!

I did a bit of digging today. The short version is that the unwinder in LLVM just flat-out does not support unwinding with pointer authentication if it's been compiled with `LIBUNWIND_ENABLE_CROSS_UNWINDING`, and it seems that MacOS's toolchain has been built that way. The code in question is here: https://github.com/llvm/llvm-project/blob/8e35c86977ce5529a9387657321ac9fefcdae5b5/libunwind/src/DwarfInstructions.hpp#L294. I was able to confirm that `unw_step` was in fact actually returning this `UNW_ECROSSRASIGNING` code.

I was able to make this work by compiling libunwind separately with these instructions - https://github.com/llvm/llvm-project/blob/main/libunwind/docs/BuildingLibunwind.rst and then compiling Ruby with `./configure --disable-install-doc --enable-yjit LDFLAGS="-Wl,-rpath /Users/ktsanaktsidis/llvm_libunwind/lib  -L/Users/ktsanaktsidis/llvm_libunwind/lib -lunwind" CFLAGS="-D_LIBUNWIND_IS_NATIVE_ONLY -I/Users/ktsanaktsidis/llvm_libunwind/include"` (yes, you have to specify the `D_LIBUNWIND_IS_NATIVE_ONLY` flag because nothing in the libunwind build system records what value of this libunwind _itself_ was compiled with, and the headers care about this value too).

I pushed a commit to your branch with a modification to the configure script to explicitly check for this brokenness. It has the same effect as your patch anyway (macOS gets pac-ret disabled by default unless asked for explicitly), but I guess doing it not-by-accident is good.

I'm not 100% sure if we should commit this test though. Getting unwinding to work can be finicky and toolchain specific... for example my machine for some reason isn't managing to unwind past the signal handler (even without pac-ret)

```
| -- C level backtrace information -------------------------------------------
| /Users/ktsanaktsidis/Code/zendesk/ruby/build/ruby(rb_vm_bugreport+0xb4c) [0x10436c97c] ../vm_dump.c:1151
| /Users/ktsanaktsidis/Code/zendesk/ruby/build/ruby(rb_vm_bugreport) (null):0
| /Users/ktsanaktsidis/Code/zendesk/ruby/build/ruby(rb_bug_for_fatal_signal+0x104) [0x1041a2250] ../error.c:1108
```

peterzhu2118:
Thank you for looking into this.

> I'm not 100% sure if we should commit this test though. Getting unwinding to work can be finicky and toolchain specific... for example my machine for some reason isn't managing to unwind past the signal handler (even without pac-ret)

I've removed the test.

PR #11549, opened 2024-09-04
[ruby/reline] Fix redisplay/insert_text called from pre_input_hook (backport)

yogo1212:
@tompng was kind enough to fix [a bug](https://bugs.ruby-lang.org/issues/20711) in master.
in hope of reducing work for the maintainers, i'm submitting his patch for the 3.3 branch, which is also affected.

commit description:

(https://github.com/ruby/reline/pull/742)

* Fix redisplay/insert_text called from pre_input_hook

* Rename insert_pasted_text to insert_multiline_text

It is now used from Reline.insert_text which is not inserting pasted text

https://github.com/ruby/reline/commit/694a540939

### DIFF
--- a/lib/reline.rb
+++ b/lib/reline.rb
@@ -335,7 +335,11 @@ module Reline
       line_editor.prompt_proc = prompt_proc
       line_editor.auto_indent_proc = auto_indent_proc
       line_editor.dig_perfect_match_proc = dig_perfect_match_proc
+
+      # Readline calls pre_input_hook just after printing the first prompt.
+      line_editor.print_nomultiline_prompt
       pre_input_hook&.call
+
       unless Reline::IOGate == Reline::GeneralIO
         @dialog_proc_list.each_pair do |name_sym, d|
           line_editor.add_dialog_proc(name_sym, d.dialog_proc, d.context)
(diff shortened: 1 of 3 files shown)


k0kubun:
When this is released as part of Ruby 3.3.6, the default gem implementation should have a new version which is published on rubygems.org. So default gem backports should be handled differently from the rest of ruby/ruby. If we just merge this PR as is, it could (not only break CI but also) complicate the release process.

@tompng As a maintainer of Reline, how do you want to proceed with releasing Reline for Ruby 3.3? Example options:

1. We cut a stable branch for Ruby 3.3 (e.g. `v0.5` or `ruby_3_3`) in Reline, we make this PR's change to that branch in ruby/reline, and then:
    1. @tompng releases it to rubygems.org as v0.5.7.1 first, and we sync the version to ruby/ruby `ruby_3_3`.
    2. @hsbt makes me a gem owner of Reline, we sync the branch to ruby/ruby `ruby_3_3`, and k0kubun releases the Ruby 3.3 branch of ruby/reline as v0.5.7.1 before the 3.3.6 release.
2. Cut a v0.5.10 release in ruby/reline master now, and sync v0.5.10 to ruby/ruby `ruby_3_3`. This is acceptable if v0.5.7~master has only backward-compatible minor changes.

tompng:
@k0kubun I'd like option 2, cutting v0.5.10 compared to option 1. The internal structure has changed but there is no backward-incompatible public API change.
I'm also fine with not including the fix to 3.3.6 just like other normal bug fixes. There are workarounds: install latest reline, install readline-ext, add it to Gemfile.

Eather way, irb/reline team will release reline-0.5.10 soon. https://github.com/ruby/reline/pull/745

k0kubun:
Got it, thanks! We've merged v0.5.10 https://github.com/ruby/ruby/pull/11558.

PR #11550, opened 2024-09-04
Only `warn` about bundled gems when require succeeds

deivid-rodriguez:


hsbt (approved):

hsbt:
Sorry, my late review. It's reasonable for optional dependencies. But I have another concern.

In my understanding, this PR makes to remove warning with bundled gems like `drb`, `csv` etc under the bundler environment. I'm not sure how notice that gems are no longer loaded without Gemfile to users.

deivid-rodriguez:
Hi @hsbt, no problem!

> In my understanding, this PR makes to remove warning with bundled gems like drb, csv etc under the bundler environment. I'm not sure how notice that gems are no longer loaded without Gemfile to users.

This PR keeps the warning when the behavior has not yet changed, i.e., it still warns when using `csv` without including it in the Gemfile in Ruby 3.3. But indeed, once the breaking change happens, it will no longer show the warning reading "csv is not part of the default gems starting from Ruby 3.4.0" after the require fails in Ruby 3.4.

I also don't have a good solution for that.

hsbt:
Thanks for explanation.

>I also don't have a good solution for that.

I also don't have that yet. I understood this works correctly during Ruby 3.3.x.

If Ruby 3.2 users upgrade their version to Ruby 3.4 directly in the future. There is no chance to know `csv` is already bundled gems, not default gems. I concerned that situation.

deivid-rodriguez:
Yep, that's the concern I also have. It may be acceptable, since jumping two versions in a row is not a great practice, but I'm not sure.

PR #11554, opened 2024-09-05
Fix issue in compile.c with a short-circuit if/unless `defined?` condition

luke-gru:
This caused an issue when `defined?` was in the `if` condition. Its instructions weren't appended to the instruction sequence even though it was compiled if a known logical short-circuit happened before the `defined?`. The catch table (`defined?` compilation produces a catch table) was still on the iseq even though the instructions weren't there. This caused faulty exception handling in the method. The solution is to no longer add the catch table entry when compiling `defined?` in this case.

This shouldn't touch much code, it's only for cases like the following, which can occur during debugging:

  if false && defined?(Some::CONSTANT)
    "more code..."
  end

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -495,7 +495,7 @@ static int iseq_set_sequence(rb_iseq_t *iseq, LINK_ANCHOR *const anchor);
 static int iseq_set_exception_table(rb_iseq_t *iseq);
 static int iseq_set_optargs_table(rb_iseq_t *iseq);
 
-static int compile_defined_expr(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, VALUE needstr);
+static int compile_defined_expr(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, VALUE needstr, bool ignore);
 static int compile_hash(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *node, int method_call_keywords, int popped);
 
 /*
@@ -1353,6 +1353,7 @@ new_label_body(rb_iseq_t *iseq, long line)
     labelobj->set = 0;
     labelobj->rescued = LABEL_RESCUE_NONE;
     labelobj->unremovable = 0;
+    labelobj->position = -1;
     return labelobj;
 }
 
@@ -2878,11 +2879,16 @@ iseq_set_exception_table(rb_iseq_t *iseq)
         table->size = tlen;
 
         for (i = 0; i < table->size; i++) {
+            int pos;
             ptr = RARRAY_CONST_PTR(tptr[i]);
             entry = UNALIGNED_MEMBER_PTR(table, entries[i]);
             entry->type = (enum rb_catch_type)(ptr[0] & 0xffff);
-            entry->start = label_get_position((LABEL *)(ptr[1] & ~1));
-            entry->end = label_get_position((LABEL *)(ptr[2] & ~1));
+            pos = label_get_position((LABEL *)(ptr[1] & ~1));
+            RUBY_ASSERT(pos >= 0);
+            entry->start = (unsigned int)pos;
+            pos = label_get_position((LABEL *)(ptr[2] & ~1));
+            RUBY_ASSERT(pos >= 0);
+            entry->end = (unsigned int)pos;
             entry->iseq = (rb_iseq_t *)ptr[3];
             RB_OBJ_WRITTEN(iseq, Qundef, entry->iseq);
 
@@ -4640,7 +4646,7 @@ compile_branch_condition(rb_iseq_t *iseq, LINK_ANCHOR *ret, const NODE *cond,
         CHECK(compile_flip_flop(iseq, ret, cond, FALSE, then_label, else_label));
         return COMPILE_OK;
       case NODE_DEFINED:
-        CHECK(compile_defined_expr(iseq, ret, cond, Qfalse));
+        CHECK(compile_defined_expr(iseq, ret, cond, Qfalse, ret == ignore));
         break;
       default:
         {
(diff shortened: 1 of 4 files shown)


luke-gru:
I'm not sure if this is an okay fix because of language servers.

luke-gru:
I think this change could lead to backward incompatibility with language servers, in that ruby doesn't check compilation of short-circuited logical operators now. I don't know if that's an issue or not, but if it is then there needs to be a different solution to this problem.

test/ruby/test_iseq.rb
```diff
@@ -809,7 +809,6 @@ def test_ever_condition_loop
   def test_unreachable_syntax_error
     mesg = /Invalid break/
     assert_syntax_error("false and break", mesg)
-    assert_syntax_error("if false and break; end", mesg)
```
nobu, line 812: Why removing this assertion?

bootstraptest/test_exception.rb
```diff
@@ -430,3 +430,31 @@ def ===(o)
...
+      raise
+    end
+    begin
+      my_method
+    rescue
+      :ok
+    else
+      :ng
+    end
+  }, '[ruby-core:20501]'
```
nobu, line 459: Could you move this test under test/?

compile.c
```diff
@@ -4458,16 +4458,23 @@ static int
...
 static int
 compile_logical(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *cond,
                 LABEL *then_label, LABEL *else_label)
 {
     DECL_ANCHOR(seq);
     INIT_ANCHOR(seq);
     LABEL *label = NEW_LABEL(nd_line(cond));
-    if (!then_label) then_label = label;
-    else if (!else_label) else_label = label;
+    bool and_cond = false; bool or_cond = false;
```
nobu, line 4469:
```suggestion
    bool and_cond = false, or_cond = false;
```

test/ruby/test_iseq.rb
luke-gru: Since the next part of the conditional is no longer compiled, it won't trigger that syntax error anymore. For the line just above it, it enters a different code path since there is no `if` and I didn't write this short-circuit optimization in that part of the code.

bootstraptest/test_exception.rb
luke-gru: done 👍

compile.c
luke-gru: 👍

test/ruby/test_iseq.rb
nobu: The `break` exists syntactically, and should be a subject of checks.

luke-gru: If that's the case I'll try to come up with a different solution to the problem.

test/ruby/test_exception.rb
```diff
@@ -149,6 +149,41 @@ def self.meow
...
+        rescue
+          print 'ok'
+        else
+          print 'ng'
+        end
+      }
+      # invoke in subprocess because the bug caused a segfault
+      out, _err, status = EnvUtil.invoke_ruby('--disable-gems', code, true, false)
+      assert_predicate(status, :success?, bug)
+      assert_equal 'ok', out
```
nobu, line 183: We have the dedicated method to run assertions in a separated process.

```suggestion
      assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
      begin;
        def my_method
          var = nil
          if #{cond}
            var = var = "here"
          end
          raise
        end
        assert_raise(RuntimeError) do
          my_method
        end
      end;
```

And this doesn't seem related to behaviors of exception, maybe test_syntax.rb?

nobu (changes requested):

test/ruby/test_exception.rb
```diff
@@ -149,6 +149,41 @@ def self.meow
     end
   end
 
+  def test_exception_with_defined_in_short_circuit_if_condition
+    bug = '[ruby-core:20501]'
+    conds = [
+      "false && defined?(Some::CONSTANT)",
+      "true || defined?(Some::CONSTANT)",
+      "(false && defined?(Some::CONSTANT))",
+      "(true || defined?(Some::CONSTANT))",
```
nobu, line 158: These lines are same as previous lines, except for parentheses.
Necessary?

luke-gru: It exercises a different code path, I put a comment about that in my most recent commit on this PR

test/ruby/test_exception.rb
luke-gru: Ok I moved to test_syntax, but I just tried using assert_separately and couldn't reproduce the issue. I could reproduce it only in a clean environment with no test environment required, so that's why I didn't use it. I've added a comment about that in my most recent commit.

tool/lib/envutil.rb
```diff
@@ -165,6 +165,11 @@ def invoke_ruby(args, stdin_data = "", capture_stdout = false, capture_stderr =
     }
 
     args = [args] if args.kind_of?(String)
+    # use the same parser as current ruby
+    if args.none? { |arg| arg.to_s =~ /--parser=/ }
+      current_parser = RUBY_DESCRIPTION =~ /prism/i ? "prism" : "parse.y"
+      args << "--parser=#{current_parser}"
+    end
```
luke-gru, line 172: I need these lines because prism is the current parser now (it wasn't when I made my initial commit). So now it only used prism parser and compiler in a subprocess without this. Now, when parse.y is tested it will use parse.y parser/compiler in subprocess too.

tool/lib/envutil.rb
```diff
@@ -165,6 +165,11 @@ def invoke_ruby(args, stdin_data = "", capture_stdout = false, capture_stderr =
     }
 
     args = [args] if args.kind_of?(String)
+    # use the same parser as current ruby
+    if args.none? { |arg| arg.to_s =~ /--parser=/ }
```
nobu, line 169:
```suggestion
    if args.none? { |arg| arg.start_with?("--parser=") }
```

nobu (approved):

PR #11555, opened 2024-09-05
Implement `Array#fetch_values`

byroot:
[[Feature #20702]](https://bugs.ruby-lang.org/issues/20702)

Works the same way than `Hash#fetch_values` for for array.

### DIFF
--- a/array.c
+++ b/array.c
@@ -2030,7 +2030,7 @@ rb_ary_last(int argc, const VALUE *argv, VALUE ary) // used by parse.y
  */
 
 static VALUE
-rb_ary_fetch(int argc, VALUE *argv, VALUE ary)
+rb_ary_fetch_m(int argc, VALUE *argv, VALUE ary)
 {
     VALUE pos, ifnone;
     long block_given;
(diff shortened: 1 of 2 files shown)


tenderlove (approved):
Looks good, but should we implement it in Ruby?

byroot:
I though about that way.

I suppose it's just:

```ruby
def fetch_values(*indexes, &block)
  indexes.map { |i| fetch(i, &block) }
end
```

tenderlove:
Since it's a new method, clearly it's not a bottleneck in anyone's app. We could make a primitive call for `fetch` if we don't want to push an extra frame. I don't have a strong opinion though, implementing in C is fine too (though we'd probably want to convert to Ruby again in the future for YJIT? 🤷🏻‍♀️)

byroot:
https://github.com/ruby/ruby/pull/11557

byroot:
Yeah, my concern isn't so much performance but more about potentially yielding at an unexpected point or some other very subtle behavior issues. And also just consistency with `Hash#fetch_values` which it's modeled on.

But yeah, this method doesn't exactly cry hotspot.

tenderlove:
> But yeah, this method doesn't exactly cry hotspot.

The good part about implementing it from the beginning in Ruby is that if people complain about speed we can rewrite in C (or tell them to use YJIT). 😂

nobu:
The commit message is wrong, `s/-/_/`

byroot:
Corrected.

byroot:
Merged https://github.com/ruby/ruby/pull/11557 instead

PR #11556, opened 2024-09-05
YJIT: Speed up block_assumptions_free

k0kubun:
This PR speeds up `block_assumptions_free()` when there are too many ISEQs with local variable access. Prior to this PR, even when it's invalidating a block that doesn't have local variables, it ends up scanning all these unrelated ISEQs, so that's why it was slow.

This PR changes `block_assumptions_free()` to first extract the ISEQ of a block to be invalidated, and use it as a key to remove the block, instead of scanning the entire hash table.

The following repro (credit: @byroot for profiling our app and @tenderlove for writing this repro) is sped up from 1666ms to 8ms. 

```rb
class Foo
end

N = 1000
M = 100000

# define many methods that depend on Foo.new
eval N.times.map { |i|
  "def call_new_#{i} = Foo.new"
}.join(";")

# define a method to call all of those methods
eval "def test = (#{N.times.map { |i| "call_new_#{i}" }.join(";")})"

test

# define a bunch of methods with locals
eval M.times.map { |i|
  "def has_local_#{i} = (_x = 1)"
}.join(";")

# compile all methods that access a local
eval M.times.map { |i| "has_local_#{i}" }.join(";")

def time
  now = Time.now
  yield
  Time.now - now
end

# Invalidate the new method so that call_new_* get invalidated
p time { Foo.send(:define_singleton_method, :new) {} }
```

### DIFF
--- a/yjit/src/invariants.rs
+++ b/yjit/src/invariants.rs
@@ -30,7 +30,6 @@ pub struct Invariants {
     /// quick access to all of the blocks that are making this assumption when
     /// the operator is redefined.
     basic_operator_blocks: HashMap<(RedefinitionFlag, ruby_basic_operators), HashSet<BlockRef>>,
-
     /// A map from a block to a set of classes and their associated basic
     /// operators that the block is assuming are not redefined. This is used for
     /// quick access to all of the assumptions that a block is making when it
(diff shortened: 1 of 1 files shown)


tenderlove (approved):
Thank you!

PR #11559, opened 2024-09-06
Accept floating point number strings without digit in fractional part

kou:
[Feature #20705]

String#to_f accepts "0." but didn't accept "0.E-9".

Kernel.#Float didn't accept both of "0." and "0.E-9".

Other systems such as Python, Node.js, PostgreSQL and MySQL accept "0." and "0.E-9" as valid floating point number strings. So Ruby also accepts them.

### DIFF
--- a/missing/dtoa.c
+++ b/missing/dtoa.c
@@ -1647,8 +1647,12 @@ break2:
     }
 #endif
     if (c == '.') {
-        if (!ISDIGIT(s[1]))
+        if (!ISDIGIT(s[1])) {
+            if (s[1] == 'e' || s[1] == 'E' || s[1] == '\0') {
+                c = *++s;
+            }
             goto dig_done;
+        }
         c = *++s;
         if (!nd) {
             for (; c == '0'; c = *++s)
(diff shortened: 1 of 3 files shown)


missing/dtoa.c
```diff
@@ -1647,8 +1647,12 @@ strtod(const char *s00, char **se)
     }
 #endif
     if (c == '.') {
-        if (!ISDIGIT(s[1]))
+        if (!ISDIGIT(s[1])) {
+            if (s[1] == 'e' || s[1] == 'E' || s[1] == '\0') {
```
kou, line 1651: We can disallow `0.` by removing `|| s[1] == '\0'`.

nobu:
Needs to update for hexadecimal floats.

```ruby
Float("0xf.p0") #=> 15.0
```
https://github.com/nobu/ruby/tree/float-dtoa

kou:
What is "hexadecimal float"!?

I can withdraw this PR in favor of your implementation.

nobu:
It is similar to hexadecimal floating-point numbers in C99.
But cannot be literals, e.g. `0xdead.beafp+00`.

kou:
Thanks.
Could you open a PR from your branch?

kou:
I close this in favor of #11807.

PR #11565, opened 2024-09-07
Fix issue with super and forwarding arguments in prism_compile.c

luke-gru:
Fixes [Bug #20720]

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -15132,11 +15132,11 @@ new_args_forward_call(struct parser_params *p, NODE *leading, const YYLTYPE *loc
 #ifndef FORWARD_ARGS_WITH_RUBY2_KEYWORDS
     NODE *kwrest = list_append(p, NEW_LIST(0, loc), NEW_LVAR(idFWD_KWREST, loc));
 #endif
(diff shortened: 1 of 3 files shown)


luke-gru:
There's an issue with failing tests now after my change. It's happening in ruby with and without prism enabled, and it's to do with RubyVM::InstructionSequence and code positions. There's a validation of code position ranges that is failing with this code:
 ```ruby
class A
  def m(*args)
    p args
  end
end
class B < A
  def m(...)
    a = nil
    super(a, ...)
  end
end
B.new.m()
```
I could look into it but it would be nice if someone with more experience in the parser could help.

**Review by kddnewton — CHANGES_REQUESTED** [review]
Thanks @luke-gru!

I've added a flag to the arguments node that should hopefully simplify a lot of the code in this PR. That PR is here: https://github.com/ruby/prism/pull/3051.

Where you currently have to loop through the elements in the argument list, you should now be able to do something like:

```c
bool is_forwardable = (cast->arguments != NULL) && PM_NODE_FLAG_P(cast->arguments, PM_ARGUMENTS_NODE_FLAGS_CONTAINS_FORWARDING);
```

Could you rebase and try that out?

luke-gru:
Thanks for the flag, that makes it easier. I've rebased and it seems to be working well.

kddnewton (approved):

PR #11570, opened 2024-09-07
[DOC] Related for Array#compact!

BurdetteLamar:


peterzhu2118:
Ah looks like #11569 was for Array#compact!, so this is a duplicate change.

peterzhu2118:
I opened #11587 instead.

PR #11573, opened 2024-09-08
Return back legacy Range#step behavior for symbol ranges

zverok:
As requested here: https://github.com/ruby/ruby/pull/11454#pullrequestreview-2260830043


knu (approved):
Thank you!

PR #11576, opened 2024-09-09
[Backport 3.3] Fixes [Bug #20718]

jcalvert:
Allow objects that are not of type `RTypedData` to use the default free function, as `RTYPEDDATA_EMBEDDED_P` can return a false positive when casting non-`RTypedData` objects

Backport of fix from #11563

@k0kubun


k0kubun (approved):
Thanks!

PR #11593, opened 2024-09-11
[DOC] Tweaks for Array#difference

BurdetteLamar:


array.c
```diff
@@ -5518,19 +5518,19 @@ rb_ary_diff(VALUE ary1, VALUE ary2)
...
  *  items are compared using <tt>eql?</tt>;  order from +self+ is preserved:
  *
  *    [0, 1, 1, 2, 1, 1, 3, 1, 1].difference([1]) # => [0, 2, 3]
- *    [0, 1, 2, 3].difference([3, 0], [1, 3]) # => [2]
- *    [0, 1, 2].difference([4]) # => [0, 1, 2]
+ *    [0, 1, 2, 3].difference([3, 0], [1, 3])     # => [2]
+ *    [0, 1, 2].difference([4])                   # => [0, 1, 2]
+ *    [0, 1, 2].difference                        # => [0, 1, 2]
  *
- *  Returns a copy of +self+ if no arguments given.
```
peterzhu2118, line 5531: I think we should keep this line. It's not clear if `self` or a copy is returned in this case.

BurdetteLamar: Done.

PR #11602, opened 2024-09-12
[DOC] Tweaks for Array#fetch

BurdetteLamar:


array.c
```diff
@@ -1995,30 +1995,32 @@ rb_ary_last(int argc, const VALUE *argv, VALUE ary) // used by parse.y
...
- *    array.fetch(index) -> element
- *    array.fetch(index, default_value) -> element
- *    array.fetch(index) {|index| ... } -> element
+ *    fetch(index) -> element
+ *    fetch(index, default_value) -> element or default_value
+ *    fetch(index) {|index| ... } -> element or block_return_value
  *
- *  Returns the element at offset  +index+.
+ *  If +index+ is in range, returns the element of +self+ at offset +index+, which must be an
+ *  {integer-convertible object}[rdoc-ref:implicit_conversion.rdoc@Integer-Convertible+Objects].
```
peterzhu2118, line 2003: I find the "which must be" part to be ambiguous. In my first read I thought the return value must be an integer-convertible object but then I realized it meant `index` instead. Maybe something like this?

```suggestion
 *  Returns the element of +self+ at offset +index+ if +index+ is in range; +index+ must be an
 *  {integer-convertible object}[rdoc-ref:implicit_conversion.rdoc@Integer-Convertible+Objects].
```

peterzhu2118 (approved):

PR #11610, opened 2024-09-13
[DOC] More on Array#fill

BurdetteLamar:
- Uses idiom `('a'..'d').to_a` to create example arrays.
- Revises examples to (more often) vertically align elements before/after call:
  ```
  a = ('a'..'d').to_a # => ["a", "b", "c", "d"]
  a.fill('-', 1, 2)   # => ["a", "-", "-", "d"]
  ```
- Revises examples to make changed elements more obvious: `'a'` replaced by `'-'` instead of by `:X`.
- Makes emphatic separation (via pseudo-headings) between no-block and block discussions.
- Adds "In brief" section;  it's long, but only about 20% as long as the rest of the discussion.  With any luck, an experienced Rubyist's question will be answered there.
- Notes that numeric arguments `start` and `length` must be integer-convertible objects.

### DIFF
--- a/array.c
+++ b/array.c
@@ -4711,7 +4711,7 @@ rb_ary_clear(VALUE ary)
  *    ['a', 'b', 'c', 'd'].fill('-', 3, 2)          # => ["a", "b", "c", "-", "-"]
  *    ['a', 'b', 'c', 'd'].fill(3, 2) {|e| e.to_s } # => ["a", "b", "c", "3", "4"]
  *
- *    # Extends with nils and specified values if necessary.
+ *    # Fills with nils if necessary.
  *    ['a', 'b', 'c', 'd'].fill('-', 6, 2)          # => ["a", "b", "c", "d", nil, nil, "-", "-"]
  *    ['a', 'b', 'c', 'd'].fill(6, 2) {|e| e.to_s } # => ["a", "b", "c", "d", nil, nil, "6", "7"]
  *
@@ -4735,7 +4735,7 @@ rb_ary_clear(VALUE ary)
  *    +nil+ means "all the rest."
  *
  *  With argument +object+ given,
- *  that one object is the sole replacement value:
+ *  that one object is used for all replacements:
  *
  *    o = Object.new           # => #<Object:0x0000014e7bff7600>
  *    a = ['a', 'b', 'c', 'd'] # => ["a", "b", "c", "d"]
@@ -4745,7 +4745,7 @@ rb_ary_clear(VALUE ary)
  *  With a block given, the block is called once for each element to be replaced;
  *  the value passed to the block is the _index_ of the element to be replaced
  *  (not the element itself);
- *  the block's return value is the element's replacement:
+ *  the block's return value replaces the element:
  *
  *    a = ['a', 'b', 'c', 'd']               # => ["a", "b", "c", "d"]
  *    a.fill(1, 2) {|element| element.to_s } # => ["a", "1", "2", "d"]
@@ -4753,7 +4753,7 @@ rb_ary_clear(VALUE ary)
  *  For arguments +start+ and +count+:
  *
  *  - If +start+ is non-negative,
- *    replaces +count+ elements from offset +start+ through the end:
+ *    replaces +count+ elements beginning at offset +start+:
  *
  *      ['a', 'b', 'c', 'd'].fill('-', 0, 2) # => ["-", "-", "c", "d"]
  *      ['a', 'b', 'c', 'd'].fill('-', 1, 2) # => ["a", "-", "-", "d"]
@@ -4832,7 +4832,7 @@ rb_ary_clear(VALUE ary)
  *      ['a', 'b', 'c', 'd'].fill('-', 1..2)          # => ["a", "-", "-", "d"]
  *      ['a', 'b', 'c', 'd'].fill(1..2) {|e| e.to_s } # => ["a", "1", "2", "d"]
  *
- *    If +end+ is smaller than +begin+, no elements are replaced:
+ *    If +end+ is smaller than +begin+, replaces no elements:
  *
  *      ['a', 'b', 'c', 'd'].fill('-', 2..1)          # => ["a", "b", "c", "d"]
  *      ['a', 'b', 'c', 'd'].fill(2..1) {|e| e.to_s } # => ["a", "b", "c", "d"]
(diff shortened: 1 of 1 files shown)


array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
  *
- *    a = ['a', 'b', 'c', 'd']
- *    a.fill(:X, 1, 0) # => ["a", "b", "c", "d"]
- *    a.fill(:X, 1, -1) # => ["a", "b", "c", "d"]
+ *      a = ('a'..'d').to_a # => ["a", "b", "c", "d"]
+ *      a.fill('-', 1, 2)   # => ["a", "-", "-", "d"]
  *
- *  With arguments +obj+ and Range +range+, and no block given,
- *  replaces elements based on the given range.
+ *  - If +start+ is negative, counts from the end:
```
peterzhu2118, line 4804: This is missing the case when `start <= -self.size`.

array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
  *
- *  If the range is positive and ascending (<tt>0 < range.begin <= range.end</tt>),
- *  replaces elements from <tt>range.begin</tt> to <tt>range.end</tt>:
+ *      a = ('a'..'d').to_a # => ["a", "b", "c", "d"]
+ *      a.fill('-', -3, 2)  # => ["a", "-", "-", "d"]
  *
- *    a = ['a', 'b', 'c', 'd']
- *    a.fill(:X, (1..1)) # => ["a", :X, "c", "d"]
+ *  - If +start+ is large (<tt>start >= array.size</tt>), extends +self+ with +nil+:
```
peterzhu2118, line 4809: Ditto:

```suggestion
 *  - If +start+ is large (<tt>start >= self.size</tt>), extends +self+ with +nil+:
```

array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
- *    a.fill(:X, (0..-2)) # => [:X, :X, :X, "d"]
- *    a = ['a', 'b', 'c', 'd']
- *    a.fill(:X, (1..-2)) # => ["a", :X, :X, "d"]
+ *  With argument +object+ and Range argument +range+,
+ *  replaces elements based on the given range:
  *
- *  If <tt>range.last</tt> and <tt>range.last</tt> are both negative,
- *  both count from the end of the array:
+ *  - If the range is positive and ascending (<tt>0 < range.begin <= range.end</tt>),
+ *    replaces elements from <tt>range.begin</tt> to <tt>range.end</tt>:
```
peterzhu2118, line 4827: This is not always true, it depends on if the end is inclusive or exclusive, for example:

```ruby
["a", "b", "c", "d"].fill("-", 1..2) # => ["a", "-", "-", "d"]
["a", "b", "c", "d"].fill("-", 1...2) # => ["a", "-", "c", "d"]
(1..2).begin # => 1
(1..2).end # => 2
(1...2).begin # => 1
(1...2).end # => 2
```

array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
+ *      a = ('a'..'d').to_a  # => ["a", "b", "c", "d"]
+ *      a.fill('-', (1..-2)) # => ["a", "-", "-", "d"]
+ *
+ *  - If <tt>range.last</tt> and <tt>range.last</tt> are both negative,
+ *    both count from the end of the array:
+ *
+ *      a = ('a'..'d').to_a   # => ["a", "b", "c", "d"]
+ *      a.fill('-', (-3..-2)) # => ["a", "-", "-", "d"]
+ *
+ *  <b>With a Block Given</b>
```
peterzhu2118, line 4848: It seems like with a block works the same as without a block except it calls the block for each replacement instead of using the predefined `object`? In that case, I don't think we need to duplicate the whole documentation for this.

array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
+ *  Replaces specified elements in +self+ with specified objects;
+ *  always returns +self+ (never a new array).
  *
- *  With argument +obj+ and no block given, replaces all elements with that one object:
+ *  Note: many examples here use this array, derived from a Range of characters:
  *
- *    a = ['a', 'b', 'c', 'd']
- *    a # => ["a", "b", "c", "d"]
- *    a.fill(:X) # => [:X, :X, :X, :X]
+ *    ('a'..'d').to_a # => ["a", "b", "c", "d"]
```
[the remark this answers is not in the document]
BurdetteLamar: How about just `[0, 0, 0, 0]`, which avoids the clutter of the single-quotes?  Then fill with `9`, to show the changes compactly and effectively?
```
a = [0, 0, 0, 0] # => [0, 0, 0, 0]
a.fill(9, 1, 2)  # => [0, 9, 9, 0]
```

array.c
BurdetteLamar: Agreed;  with-block part should be much briefer.

array.c
```diff
@@ -4692,198 +4692,251 @@ rb_ary_clear(VALUE ary)
...
  *
- *  With arguments +obj+ and Integer +start+, and no block given,
- *  replaces elements based on the given start.
+ *  In all cases, numeric arguments +start+ and +length+ must be
+ *  {integer-convertible objects}[rdoc-ref:implicit_conversion.rdoc@Integer-Convertible+Objects].
+ *
+ *  In brief:
```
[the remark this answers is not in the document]
BurdetteLamar: Agreed;  once other comments are resolved, pare down.

array.c
BurdetteLamar: Agreed;  needs revision.

array.c
BurdetteLamar: Should be added.

array.c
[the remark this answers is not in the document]
peterzhu2118: Or `[0, 1, 2, 3]` to make it easier to see which indexes were replaced?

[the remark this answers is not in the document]
BurdetteLamar: Yes!

BurdetteLamar:
@peterzhu2118, I'm not sure what to say about range here.  [Doc for 2.7.0](https://docs.ruby-lang.org/en/2.7.0/Array.html#method-i-fill) says almost nothing about it.  [Doc for 3.0.0](https://docs.ruby-lang.org/en/3.0/Array.html#method-i-fill) (which I think I wrote) is inadequate (does not cover all details).

It's tempting (tempting enough?) to just document simplest range, and say the rest is subtle, sometimes non-intuitive, and not recommended.

peterzhu2118:
> say the rest is subtle, sometimes non-intuitive, and not recommended.

No please don't do that.

The range case is intuitive, because it depends on whether the end is inclusive or exclusive.

peterzhu2118:
> I say not intuitive, b/c:
>
> ```ruby
> r = 0..-2 # => 0..-2
> r.size # => 0
> ['a', 'b', 'c', 'd'].fill('-', r) # => ["-", "-", "-", "d"]
> ```

Well that range means "from the first element to the second last, both inclusive". If you think about it like that, obviously `Range#size` doesn't know how many elements there are because it depends on the array that it's applied on.

BurdetteLamar:
> Change call-seq to:
> 
> ```
> fill(object, start = nil, length = nil) -> new_array
> fill(object, range) -> new_array
> fill(start = nil, length = nil) {|element| ... } -> new_array
> fill(range) {|element| ... } -> new_array
> ```
> 
> Then explain that nil start means 0, nil-length means all.

Done.

BurdetteLamar:
I've changed call-seq `length` to `count`, to avoid any confusion with Array#length.

peterzhu2118 (approved):

PR #11624, opened 2024-09-13
[PRISM] Remove snapshot testing from Prism sync

kddnewton:
Per @hsbt's request, removing snapshot testing from ruby/ruby.

Fixes https://github.com/ruby/prism/issues/3060


hsbt:
Thanks.

PR #11631, opened 2024-09-16
[PRISM] allow retry in BEGIN_ELSE node

luke-gru:


luke-gru:
Ah yes okay, I'll leave you to fix this issue. There's also this code, which is valid ruby but is an issue with retry in an ensure node not walking up the context stack:

```ruby
  begin
    yield
  rescue StandardError => e
    begin
      puts "hi"
      retry
    rescue
      retry unless e
      raise e
    else
      retry
    ensure
      retry
    end
  end

```

I got this from `test_invalid_retry` in `test_ast.rb`.

PR #11619, opened 2024-09-13
Expose ruby_thread_has_gvl_p in thread.h

kbrock:
Hello great ruby people,

I noticed a pattern:

```c
int ruby_thread_has_gvl_p(void);

[...]

if (ruby_thread_has_gvl_p()) {
  method_call(&context);
}
else {
  rb_thread_call_with_gvl(method_call, &context);
}
```

Methods of interest:

* `rb_thread_call_with_gvl`:
  - Introduced in Dec 30 2008 via 9c04a0647f
  - Made public on Jul 18 2012 via e9a91d2c95
* `ruby_thread_has_gvl_p`
  - Introduced in Jan 12 2009 via 6f09fc2e
  - Not public yet. **<== my concern**

Is now a good time to make `ruby_thread_has_gvl_p` public?

Background
==========

I stumbled upon this issue while trying to compile `ovirt-engine-sdk` and reading through their PRs. But I noticed `ruby_thread_has_gvl_p()` is referenced by 400 [non-forked projects](https://github.com/search?q=ruby_thread_has_gvl_p+language%3AC++NOT+is%3Afork++NOT+is%3Aarchived&type=code) like [ffi](https://github.com/ffi/ffi/blob/85d0fabce6ab5ceb3848f7e09a265341672a272d/ext/ffi_c/Function.c#L104).

ffi and a few others used to detect `ruby_thread_has_gvl_p` in extconf.rb, but now they have removed this code and assume it is available. Not saying this is correct, just sharing what I have observed.

Please advise if I can do anything to make this easier to evaluate and if I should be raising this question in another forum.

Thank you for the wonderful world we live in and making all of our tools nice.

Keenan

PR #11649, opened 2024-09-19
Lenient call with gvl

kbrock:
Alternative to #11619

This is part of https://bugs.ruby-lang.org/issues/20750

Before
----

Calling `rb_thread_call_with_gvl` when the GVL is acquired throws an error.

After
---

Calling `rb_thread_call_with_gvl` when the GVL is acquired graciously continues.

Sorry, I do not know if we want a larger or smaller change.
So I kept as 2 commits.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -342,6 +342,12 @@ The following bundled gems are updated.
       `IO` objects share the same file descriptor, closing one does not affect
       the other. [[Feature #18455]]
 
+* GVL
+
+    * `rb_thread_call_with_gvl` now works with or without the GVL.
+      This allows gems to avoid checking `ruby_thread_has_gvl_p`.
+      Please still be diligent about the GVL. [[Feature #20750]]
+
 * Set
 
     * A C API for `Set` has been added. The following methods are supported:
@@ -402,6 +408,7 @@ A lot of work has gone into making Ractors more stable, performant, and usable. 
 [Feature #19908]: https://bugs.ruby-lang.org/issues/19908
 [Feature #20610]: https://bugs.ruby-lang.org/issues/20610
 [Feature #20724]: https://bugs.ruby-lang.org/issues/20724
+[Feature #20750]: https://bugs.ruby-lang.org/issues/20750
 [Feature #20884]: https://bugs.ruby-lang.org/issues/20884
 [Feature #20925]: https://bugs.ruby-lang.org/issues/20925
 [Feature #20971]: https://bugs.ruby-lang.org/issues/20971
--- a/thread.c
+++ b/thread.c
@@ -2042,6 +2042,9 @@ rb_thread_io_blocking_region(struct rb_io *io, rb_blocking_function_t *func, voi
  *       created as Ruby thread (created by Thread.new or so).  In other
  *       words, this function *DOES NOT* associate or convert a NON-Ruby
  *       thread to a Ruby thread.
+ *
+ * NOTE: If this thread has already acquired the GVL, then the method call
+ *       is performed without acquiring or releasing the GVL (from Ruby 4.0).
  */
 void *
 rb_thread_call_with_gvl(void *(*func)(void *), void *data1)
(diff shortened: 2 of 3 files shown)


nobu:
Since fiddle is an external gem, so you have to send that part to https://github.com/ruby/fiddle.
Also, it needs to be workable with previous ruby versions, and can't apply this blindly.

A possible way is to define a new macro so that extension libraries can know if it is save to call `rb_thread_call_with_gvl` always.

kbrock:
Thank you @nobu 

`fiddle` checks `ruby_thread_has_gvl_p` before calling `rb_thread_call_with_gvl`.
I've seen other gems do this, too.

Since `ruby_thread_has_gvl_p` is experimental and private, this makes me nervous.
The proposed solution is to make the call to `ruby_thread_has_gvl_p` unnecessary.

@eregon explains his thoughts in https://bugs.ruby-lang.org/issues/20750

thread.c
```diff
@@ -1873,6 +1873,9 @@ rb_thread_io_blocking_region(rb_blocking_function_t *func, void *data1, int fd)
  *       created as Ruby thread (created by Thread.new or so).  In other
  *       words, this function *DOES NOT* associate or convert a NON-Ruby
  *       thread to a Ruby thread.
+ *
+ * NOTE: If this thread has already aquired the GVL, then the method call
```
alexcameron89, line 1877: 👋 Hello! I noticed a minor typo: `aquired` -> `acquired`

kbrock: thank you @alexcameron89

nobu:
> The proposed solution is to make the call to `ruby_thread_has_gvl_p` unnecessary.

Consider the case fiddle with this PR running with previous versions.
The new fiddle may call `rb_thread_call_with_gvl` while GVL is acquired, but that old ruby will crash.

kbrock:
> > The proposed solution is to make the call to `ruby_thread_has_gvl_p` unnecessary.
> 
> Consider the case fiddle with this PR running with previous versions. The new fiddle may call `rb_thread_call_with_gvl` while GVL is acquired, but that old ruby will crash.

Yes, the code will need to continue using the syntax with the `ruby_thread_has_gvl_p()` check for a number of years. But in 2028, we will be able to remove the extra check (depending upon when they drop support for 3.4).

kbrock:
@nobu are you suggesting we introduce `rb_thread_call`? And then we merge `rb_thread_call_with_gvl` and `rb_thread_call_without_gvl`?

nobu:
> Yes, the code will need to continue using the syntax with the `ruby_thread_has_gvl_p()` check for a number of years. But in 2028, we will be able to remove the extra check (depending upon when they drop support for 3.4).

Actually `fiddle` requires 2.5, and I'm not sure when `fiddle` drop old versions.

> @nobu are you suggesting we introduce `rb_thread_call`? And then we merge `rb_thread_call_with_gvl` and `rb_thread_call_without_gvl`?

What I suggested is adding a macro:
> A possible way is to define a new macro so that extension libraries can know if it is save to call `rb_thread_call_with_gvl` always.

kbrock:
update:
- rebased
- pruned out fiddle which is no longer part of ruby base

I think this has a 🍏 per https://bugs.ruby-lang.org/issues/20750 
I will look into make-ibm to see the issue on s390x.

It looks like it is applying a patch to some code I modified, but I'm stumpted as to how that patch came into play.

kbrock:
update:
- rebased to latest (now all tests pass)
- Added NEWS item

NEWS.md
```diff
@@ -154,6 +154,11 @@ Note: We're only listing outstanding class updates.
       waiting on a blocking IO operation when the IO operation is closed.
       [[Feature #21166]]
 
+* GVL
+
+    * `rb_thread_call_with_gvl` now works with or without the GVL.
+      This allows gems to avoid checking `ruby_native_thread_p`.
```
eregon:
```suggestion
      This allows gems to avoid checking `ruby_thread_has_gvl_p`.
```

NEWS.md
```diff
@@ -154,6 +154,11 @@ Note: We're only listing outstanding class updates.
       waiting on a blocking IO operation when the IO operation is closed.
       [[Feature #21166]]
 
+* GVL
```
eregon: This should be under `C API updates` I'll move it there

eregon (approved):

PR #11653, opened 2024-09-20
[Feature #120782] Introduction of Happy Eyeballs Version 2 (RFC8305) in TCPSocket.new

shioimm:
This is an implementation of Happy Eyeballs version 2 (RFC 8305) in `TCPSocket.new`.
See https://bugs.ruby-lang.org/issues/20782

### Background
Prior to this implementation, I implemented Happy Eyeballs Version 2 (HEv2) for `Socket.tcp` in https://github.com/ruby/ruby/pull/9374. 
HEv2 is an algorithm defined in [RFC 8305](https://datatracker.ietf.org/doc/html/rfc8305), aimed at improving network connectivity. 
For more details on the specific cases that HEv2 helps, please refer to https://bugs.ruby-lang.org/issues/20108.

### Proposal & Outcome
This proposal implements the same HEv2 algorithm in `TCPSocket.new`.
Since `TCPSocket.new` is used more widely than `Socket.tcp`, this change is expected to broaden the impact of HEv2's benefits. 
Like `Socket.tcp`, I have also added `fast_fallback` keyword argument to `TCPSocket.new`. 
This option is set to true by default, enabling the HEv2 functionality. 
However, users can explicitly set it to false to disable HEv2 and use the previous behavior of `TCPSocket.new`.

It should be noted that HEv2 is enabled only in environments where pthreads are available.
This specification follows the approach taken in https://bugs.ruby-lang.org/issues/19965, where name resolution can be interrupted. 
(In environments where pthreads are not available, the `fast_fallback` option is ignored.)

### Performance
Below is the benchmark of 100 requests to `www.ruby-lang.org` with the fast_fallback option set to true and false, respectively. 
While there is a slight performance degradation when HEv2 is enabled, the degradation is smaller compared to that seen in `Socket.tcp` .

```ruby
require 'socket'
require 'benchmark'

hostname = "www.ruby-lang.org"
port = 80
n = 100

Benchmark.bmbm do |x|
  x.report("fast_fallback: true") do
    n.times { TCPSocket.new(hostname, port).close }
  end

  x.report("fast_fallback: false") do
    n.times { TCPSocket.new(hostname, port, fast_fallback: false).close }
  end
end
```

```
~/s/build ❯❯❯ ../install/bin/ruby ../ruby/test.rb
Rehearsal --------------------------------------------------------
fast_fallback: true    0.017588   0.097045   0.114633 (  1.460664)
fast_fallback: false   0.014033   0.078984   0.093017 (  1.413951)
----------------------------------------------- total: 0.207650sec

                           user     system      total        real
fast_fallback: true    0.020891   0.124054   0.144945 (  1.473816)
fast_fallback: false   0.018392   0.110852   0.129244 (  1.466014)
```

### DIFF
--- a/ext/socket/ipsocket.c
+++ b/ext/socket/ipsocket.c
@@ -195,6 +195,11 @@ rsock_init_inetsock(VALUE sock, VALUE remote_host, VALUE remote_serv,
 
 #elif FAST_FALLBACK_INIT_INETSOCK_IMPL == 1
 
+#define IPV6_ENTRY_POS 0
+#define IPV4_ENTRY_POS 1
+#define RESOLUTION_ERROR 0
+#define SYSCALL_ERROR 1
+
 static int
 is_specified_ip_address(const char *hostname)
 {
@@ -474,6 +479,12 @@ in_progress_fds(const int *fds, int fds_size)
     return false;
 }
 
+struct fast_fallback_error
+{
+    int type;
+    int ecode;
+};
+
 static VALUE
 init_fast_fallback_inetsock_internal(VALUE v)
 {
@@ -484,8 +495,9 @@ init_fast_fallback_inetsock_internal(VALUE v)
     struct addrinfo *remote_ai = NULL, *local_ai = NULL;
     int fd = -1, connected_fd = -1, status = 0, local_status = 0;
     int remote_addrinfo_hints = 0;
-    int last_error = 0;
+    struct fast_fallback_error last_error = { 0, 0 };
     const char *syscall = 0;
+    VALUE host, serv;
 
     #ifdef HAVE_CONST_AI_ADDRCONFIG
     remote_addrinfo_hints |= AI_ADDRCONFIG;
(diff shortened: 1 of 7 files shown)


ext/socket/ipsocket.c
```diff
@@ -167,10 +167,930 @@ init_inetsock_internal(VALUE v)
...
+        ends_at = select_expires_at(
+            &resolution_store,
+            resolution_delay_expires_at,
+            connection_attempt_delay_expires_at,
+            user_specified_resolv_timeout_at,
+            user_specified_connect_timeout_at
+        );
+        if (ends_at) {
+            if (debug) printf("[DEBUG] %d: ends_at->tv_sec %ld\n", count, ends_at->tv_sec);
+            if (debug) printf("[DEBUG] %d: ends_at->tv_usec %d\n", count, ends_at->tv_usec);
```
nobu, line 814: 32bit is enough for 1,000,000, and ruby assumes that `int` has 32bit at least.

```suggestion
            if (debug) printf("[DEBUG] %d: ends_at->tv_sec %"PRI_TIMET_PREFIX"d\n", count, ends_at->tv_sec);
            if (debug) printf("[DEBUG] %d: ends_at->tv_usec %d\n", count, (int)ends_at->tv_usec);
```

shioimm: Thank you for your comment. `If (debug)` will be removed after CI passes.

ext/socket/ipsocket.c
```diff
@@ -188,6 +190,1081 @@ rsock_init_inetsock(VALUE self, VALUE remote_host, VALUE remote_serv, VALUE loca
...
+                            } else {
+                                resolution_store.v6.ai = arg->getaddrinfo_entries[IPV6_ENTRY_POS]->ai;
+                            }
+                            if (resolution_store.v4.finished) {
+                                resolution_store.is_all_finised = true;
+                                resolution_delay_expires_at = NULL;
+                                user_specified_resolv_timeout_at = NULL;
+                                break;
+                            }
+                        } else if (strcmp(resolved_type, IPV4_HOSTNAME_RESOLVED) == 0) {
```
nurse, line 1004: strcmpは過去に何度も脆弱性の原因となったので、strncmpやmemcmpで代替できる場合は変えた方が良いです。そもそも今回は1 byteの比較なので、 `resolved_type[0]` と比較したほうがいいです

ext/socket/ipsocket.c
```diff
@@ -188,6 +190,1081 @@ rsock_init_inetsock(VALUE self, VALUE remote_host, VALUE remote_serv, VALUE loca
...
+    if ((flags & O_NONBLOCK) != 0) return;
+
+    flags |= O_NONBLOCK;
+
+    if (fcntl(fd, F_SETFL, flags) < 0) rb_syserr_fail(errno, "fcntl(2)");
+    return;
+}
+
+static int
+in_progress_fds(const int *fds, int fds_size)
```
nurse, line 494: この関数はconnection_attempt_fdsにしか使わないのですから、もっとそちらとあわせてもうちょっと抽象度を上げたほうが読みやすい気がします。
また、実装でループして判定していますが、capacityとsizeを分ければ、sizeを返すだけですみますね

ext/socket/ipsocket.c
```diff
@@ -188,6 +190,1081 @@ rsock_init_inetsock(VALUE self, VALUE remote_host, VALUE remote_serv, VALUE loca
...
+    arg.remote.res = 0;
+    arg.local.host = local_host;
+    arg.local.serv = local_serv;
+    arg.local.res = 0;
+    arg.type = type;
+    arg.resolv_timeout = resolv_timeout;
+    arg.connect_timeout = connect_timeout;
+
+    if (type == INET_CLIENT && FAST_FALLBACK_INIT_INETSOCK_IMPL == 1 && RTEST(fast_fallback)) {
+        char *hostp, *portp;
```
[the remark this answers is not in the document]
shioimm: 一旦`struct inetsock_arg arg`にセットした値を使ったほうが抜け漏れがないかなと思っていたのですが、確かに不要ですね。直します!

ext/socket/ipsocket.c
shioimm: すみません、こちらのコメントをうまく読み解けず…
connection_attempt_fdsの中には接続に失敗したfd(値が`-1`になっているもの)も含まれており、それらも含めてsizeの値になっているため、中身を走査して実際に接続中のものがあるかを確認する必要があると考えています。

shioimm: なるほど!
- connection_attempt_fdsのチェック済みのfdの位置に-1を入れるのではなく、connection_attempt_fdsを前方に詰め、connection_attempt_fds_sizeは-1する
- そうするとconnection_attempt_fds_size > 0ならin_progress_fdsはtrue
- この辺りのコードを整理してカプセル化する

と理解しました。ありがとうございます。修正してみます。

ext/socket/raddrinfo.c
```diff
@@ -469,7 +469,7 @@ cancel_getaddrinfo(void *ptr)
     rb_nativethread_lock_unlock(&arg->lock);
 }
 
-static int
+int
 do_pthread_create(pthread_t *th, void *(*start_routine) (void *), void *arg)
```
kou, line 473: Can we add a prefix for non-static functions to avoid conflicting as much as possible?

shioimm: Thanks for pointing that out.I have remedied this with [27faf7f](https://github.com/ruby/ruby/pull/11653/commits/27faf7fcaeee92e0a664132ceb87843823301fec).

kou: Thanks. But this adds more non-prefixed functions...
For example: https://github.com/ruby/ruby/pull/11653/files#diff-443562a7b162dc0d001a053a745f194c4dbbae5a68d94e9d385d86d9e670431bR826

mperham:
Will this go in Ruby 3.4 or 3.5?

PR #11655, opened 2024-09-20
Add RB_GC_GUARD for operands in PUSH_INSN

peterzhu2118:
In the ASAN CI, there is sometimes a use-after-poison in the objtostring instruction. This suggests that there is a T_NONE in the stack.

http://ci.rvm.jp/logfiles/brlog.trunk_asan.20240920-082802

My hypothesis is that the in the parse.y compiler, there are unnecessary RB_OBJ_WRITTEN write barriers after the putobject instructions, which ensures that the objects are on the stack. These write barriers are unnecessary because new_insn_core guarantees that all objects in the operands have write barriers applied.

When we call new_insn_core, it allocates an INSN, which could trigger GC. If the objects are not on the stack, the objects could be swept before the INSN is attached to the iseq.

This commit changes the PUSH_INSN to ensure that the objects remains on the stack until the INSN is attached to the iseq.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -2785,8 +2785,8 @@ pm_compile_pattern(rb_iseq_t *iseq, pm_scope_node_t *scope_node, const pm_node_t
             PUSH_INSN(ret, location, putnil);
         }
         else {
+            rb_obj_hide(keys);
             PUSH_INSN1(ret, location, duparray, keys);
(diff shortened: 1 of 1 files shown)


peterzhu2118:
Closing because commit 69f28ab715a02692fb2a9128bed46044963cbb50 didn't fix the issue, so this wouldn't fix the issue either.

PR #11660, opened 2024-09-21
Unpoison page->freelist before trying to assert on it

KJTsanaktsidis:
Otherwise trying to deref the pointer can cause an ASAN crash, even though the only reason we're dereferencing it is so that we can assert on it.

This causes ASAN crashes when building with both VM_CHECK_MODE and ASAN - e.g. https://ruby-rr-ci.kjtsanaktsidis.id.au/jenkins/job/ruby_ci-rr-asan/789/testReport/test-all/TestGCCompact/TestGCCompact_test_moving_arrays_down_size_pools/

CC @peterzhu2118 does this look reasonable to you? I didn't want to call asan_unlock_freelist/asan_lock_freelist because I wasn't entirely sure if it'd be already poisoned in this code path always.


peterzhu2118 (approved):
Looks good to me!

PR #11664, opened 2024-09-23
Ensure fiber scheduler is woken up when close interrupts read

KJTsanaktsidis:
Backports https://bugs.ruby-lang.org/issues/20723 to Ruby 3.3


k0kubun (approved):
Thanks!

PR #11666, opened 2024-09-23
Changed to use rb_parser_string_t * instead of VALUE to check if a string ends with a newline

S-H-GAMELINKS:
Remove on RSTRING_END dependency from parser

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -708,15 +708,6 @@ after_pop_stack(int len, struct parser_params *p)
 #define TOK_INTERN() intern_cstr(tok(p), toklen(p), p->enc)
 #define VALID_SYMNAME_P(s, l, enc, type) (rb_enc_symname_type(s, l, enc, (1U<<(type))) == (int)(type))
 
-#ifndef RIPPER
-static inline int
-char_at_end(struct parser_params *p, VALUE str, int when_empty)
-{
-    long len = RSTRING_LEN(str);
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -7347,7 +7339,8 @@ ruby_show_error_line(struct parser_params *p, VALUE errbuf, const YYLTYPE *yyllo
     }
     if (RTEST(errbuf)) {
         mesg = rb_attr_get(errbuf, idMesg);
-        if (RSTRING_LEN(mesg) > 0 && *(RSTRING_END(mesg)-1) != '\n')
+        rb_parser_string_t *str = rb_str_to_parser_string(p, mesg);
```
nobu, line 7342: Doesn't this `str` leak the memory?

parse.y
```diff
@@ -15943,7 +15936,7 @@ rb_parser_printf(struct parser_params *p, const char *fmt, ...)
     va_start(ap, fmt);
     rb_str_vcatf(mesg, fmt, ap);
     va_end(ap);
-    if (end_with_newline_p(p, mesg)) {
+    if (parser_string_end_with_newline_p(p, rb_str_to_parser_string(p, mesg))) {
```
nobu, line 15939: Ditto.

S-H-GAMELINKS:
Already removed RSTRING_END dependency since in https://github.com/ruby/ruby/pull/11713

Changed to use rb_parser_string_t * instead of VALUE.  And , use parser_string_char_at_end function instead of char_at_end function.

PR #11674, opened 2024-09-24
Use user defined parameterizing rules `f_kw(value)`

ydah:
Apply the parameterizing rule, a newly added feature in Lrama.


ydah:
Conflicts were resolved.

PR #11683, opened 2024-09-25
[DOC] Layout fix for Float#to_s

herwinw:
The parser gets confused with the enumeration followed by a code snippet, which uses the sam indentation as a continuation of the enumeration. The chosen solution has been copied from the docs of Float#floor.

Example of current layout: https://ruby-doc.org/3.3.5/Float.html#method-i-to_s

### DIFF
--- a/numeric.c
+++ b/numeric.c
@@ -1039,6 +1039,8 @@ rb_float_new_in_heap(double d)
 }
 
 /*
+ *  :markup: markdown
+ *
  *  call-seq:
  *    to_s -> string
  *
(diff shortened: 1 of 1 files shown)


nobu:
What about moving each example item by item?

```c
 *  - A fixed-point number.
 *      3.14.to_s         # => "3.14"
 *  - A number in "scientific notation" (containing an exponent).
 *      (10.1**50).to_s   # => "1.644631821843879e+50"
 *  - 'Infinity'.
 *      (10.1**500).to_s  # => "Infinity"
 *  - '-Infinity'.
 *      (-10.1**500).to_s # => "-Infinity"
 *  - 'NaN' (indicating not-a-number).
 *      (0.0/0.0).to_s    # => "NaN"
```

nobu:
`Float#floor` was changed because of tables, as far as I understand.

nobu:
Another way is to indent the body of the list items more deeply.

```c
 *  -   A fixed-point number.
 *  -   A number in "scientific notation" (containing an exponent).
 *  -   'Infinity'.
 *  -   '-Infinity'.
 *  -   'NaN' (indicating not-a-number).
 *
 *    3.14.to_s         # => "3.14"
 *    (10.1**50).to_s   # => "1.644631821843879e+50"
 *    (10.1**500).to_s  # => "Infinity"
 *    (-10.1**500).to_s # => "-Infinity"
 *    (0.0/0.0).to_s    # => "NaN"
```

PR #11689, opened 2024-09-25
[DOC] Tweaks for Array#flatten!

BurdetteLamar:


array.c
```diff
@@ -6495,33 +6495,37 @@ flatten(VALUE ary, int level)
...
+ *  +depth+ must be an
+ *  {integer-convertible object}[rdoc-ref:implicit_conversion.rdoc@Integer-Convertible+Objects],
+ *  or +nil+.
+ *  At each level of recursion:
  *
- *  With non-negative Integer argument +level+, flattens recursively through +level+ levels:
+ *  - Each element that is an array is "flattened"
+ *    (that is, replaced by its individual array elements).
+ *  - Each element that is not an array is unchanged
+ *    (even if the element is an object that has instance method flatten).
```
peterzhu2118, line 6509: Nit: I think it would be easier to understand if we added some formatting to make it more obvious that it's code?

```suggestion
 *    (even if the element is an object that has instance method +flatten+).
```

PR #11699, opened 2024-09-26
Extract `setup_branch`

ydah:
From duplicate code in `decl_branch_base` and `add_trace_branch_coverage`.


nobu:
Why is `setup_branch` `void` instead of returning `branch`?

ydah:
Indeed it is. I change `setup_branch` returning `branch` instead of `void`.
Thank you.

PR #11704, opened 2024-09-26
[DOC] Tweaks for Array#hash

BurdetteLamar:


array.c
```diff
@@ -5270,15 +5270,18 @@ rb_ary_hash_values(long len, const VALUE *elements)
 
 /*
  *  call-seq:
- *    array.hash -> integer
+ *    hash -> integer
  *
  *  Returns the integer hash value for +self+.
  *
- *  Two arrays with the same content will have the same hash code (and will compare using #eql?):
+ *  Two arrays with the same content have the same hash value (and are #eql?):
```
peterzhu2118, line 5277: This is not true. Having the same hash value does not guarantee that they are `#eql?`. However, if they are `#eql?`, then they must have the same hash value.

The original documentation meant that `#hash` is used for the hash and `#eql?` is used for comparison in operations that hash.

array.c
```diff
@@ -5270,15 +5270,18 @@ rb_ary_hash_values(long len, const VALUE *elements)
...
  *
- *    [0, 1, 2].hash == [0, 1, 2].hash # => true
- *    [0, 1, 2].hash == [0, 1, 3].hash # => false
+ *    ['a', 'b'].hash == ['a', 'b'].hash # => true
+ *    ['a', 'b'].hash == ['a', 'c'].hash # => false
+ *    ['a', 'b'].hash == ['a'].hash      # => false
  *
+ *  Related: Object#__id__;
```
peterzhu2118, line 5283: I'm not sure why `BasicObject#__id__` or `Object#object_id` is a related method. It does fundamentally different things. `#hash` is based on the contents of the object whereas `BasicObject#__id__` or `Object#object_id` is based on the exact object. For example:

```ruby
a = [1, 2]
b = [1, 2]
a.__id__ # => 47200
b.__id__ # => 54260
a.hash # => 2062042495941918302
b.hash # => 2062042495941918302
```

array.c

array.c

peterzhu2118 (approved):

PR #11711, opened 2024-09-27
Fix RGENGC_FORCE_MAJOR_GC CI workflow

eightbitraptor:
Addresses https://bugs.ruby-lang.org/issues/20762

If we run with RGENGC_FORCE_MAJOR_GC, then all GC's will be major GC's. This means that any pages added to pooled pages will be unable to be allocated into until the next major GC. This frequently leaves the size pools with no allocatable pages, as there are no minor collections happening to sweep the pooled pages and add them back to the heap. 

We should take these "held" slots into account when deciding whether to allocate another page after a major or not.

### DIFF
--- a/gc/default.c
+++ b/gc/default.c
@@ -428,6 +428,7 @@ typedef struct rb_size_pool_struct {
     size_t total_allocated_objects;
     size_t total_freed_objects;
     size_t final_slots_count;
+    size_t pooled_slots_count;
 
     /* Sweeping statistics */
     size_t freed_slots;
@@ -595,7 +596,6 @@ typedef struct rb_objspace {
     } rcompactor;
 
     struct {
-        size_t pooled_slots;
         size_t step_slots;
     } rincgc;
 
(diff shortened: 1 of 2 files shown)


gc/default.c
```diff
@@ -2164,9 +2164,11 @@ heap_prepare(rb_objspace_t *objspace, rb_size_pool_t *size_pool, rb_heap_t *heap
                         /* Do steps of incremental marking or lazy sweeping. */
                         gc_continue(objspace, size_pool, heap);
 
-                        if (heap->free_pages == NULL &&
-                                !heap_page_allocate_and_initialize(objspace, size_pool, heap)) {
-                            rb_bug("cannot create a new page after major GC");
+                        if (heap->free_pages == NULL) {
+                            heap_page_allocate_and_initialize_force(objspace, size_pool, heap);
```
peterzhu2118, line 2168: We definitely don't want to use `heap_page_allocate_and_initialize_force` here because it creates a new page without changing the `allocatable_slots` counter. This could cause the heap to grow larger than expected.

Instead, I think `gc_sweep_finish_size_pool` should guarantee that the size pool has at least one empty slot or set `allocatable_slots > 0`.

eightbitraptor:
Closing this now that RGENGC_FORCE_MAJOR_GC has been removed.

PR #11715, opened 2024-09-28
[BUG:20767] VM_ASSERT(cme->owner == T_CLASS) could be invalid

tagomoris:
See https://bugs.ruby-lang.org/issues/20767

### DIFF
--- a/test/ruby/test_require.rb
+++ b/test/ruby/test_require.rb
@@ -1014,7 +1014,7 @@ class TestRequire < Test::Unit::TestCase
 
   def test_require_refined_and_redefined
     bug20767 = '[ruby-core:20767]'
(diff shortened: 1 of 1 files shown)


test/ruby/test_require.rb
```diff
@@ -1011,4 +1011,45 @@ class Object
       end
     RUBY
   end
+
+  def test_require_refined_and_redefined
+    bug20767 = '[ruby-core:20767]'
+    expectedn = <<~EOE.split("\n")
```
nobu, line 1017:
```suggestion
    expected = <<~EOE.split("\n")
```

XrXr:
Looks like this was fixed by a838f980f599d95ccf344157f7074e997af31f48.

PR #11730, opened 2024-09-30
-l option of 7z is unknown switch with the `ubuntu-latest` for Ruby 3.2

hsbt:
https://github.com/ruby/actions/actions/runs/11095032727/job/30823174026#step:3:349


hsbt:
@nagachika I removed Travis CI because Travis is not working today.

PR #11743, opened 2024-09-30
[DOC] Tweaks for Array#inspect

BurdetteLamar:


array.c
```diff
@@ -2969,14 +2969,14 @@ inspect_ary(VALUE ary, VALUE dummy, int recur)
 
 /*
  *  call-seq:
- *    array.inspect -> new_string
+ *    inspect -> new_string
  *
- *  Returns the new String formed by calling method <tt>#inspect</tt>
- *  on each array element:
```
peterzhu2118, line 2975: This old documentation holds more information, especially the part where it calls `#inspect` on each element. We should keep that.

peterzhu2118 (approved):

PR #11745, opened 2024-09-30
[DOC] Tweaks for Array#intersection

BurdetteLamar:


array.c
```diff
@@ -5624,23 +5624,21 @@ rb_ary_and(VALUE ary1, VALUE ary2)
 
 /*
  *  call-seq:
- *    array.intersection(*other_arrays) -> new_array
+ *    intersection(*other_arrays) -> new_array
  *
- *  Returns a new +Array+ containing each element found both in +self+
- *  and in all of the given Arrays +other_arrays+;
- *  duplicates are omitted; items are compared using <tt>eql?</tt>
- *  (items must also implement +hash+ correctly):
```
peterzhu2118, line 5632: I think we should keep the part here about implementing `#hash`.

peterzhu2118 (approved):

PR #11752, opened 2024-10-01
Use user defined inline rules `user_or_keyword_variable`

ydah:
Apply the Inline rule, a newly added feature in Lrama.
https://github.com/ruby/lrama/blob/master/NEWS.md#inlining

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -2925,6 +2925,11 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
                 | tCONSTANT
                 ;
 
+%rule %inline user_or_keyword_variable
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -2925,6 +2925,11 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
                 | tCONSTANT
                 ;
 
+%rule %inline user_or_keyword_variable
+                : user_variable
```
yui-knk, line 2929: @ydah It seems that `ripper: var_field!` and `assignable` call is omitted from the rule whose RHS is `user_variable`, is it intentional?

PR #11760, opened 2024-10-02
Alias Float#magnitude to abs

nithinbekal:
The `Float#magnitude` method has no docs, but has the same implementation as `Float#abs`. By aliasing, we get the benefit of linking to the other method in the API docs.

We already alias the method to `abs` for `Integer` in the same file.

### Screenshots

Before:

Screenshot 2024-10-01 at 11 07 28 PM

After:

Screenshot 2024-10-01 at 11 07 07 PM

### DIFF
--- a/numeric.rb
+++ b/numeric.rb
@@ -328,10 +328,7 @@ class Float
     Primitive.cexpr! 'rb_float_abs(self)'
   end
 
-  def magnitude
-    Primitive.attr! :leaf
-    Primitive.cexpr! 'rb_float_abs(self)'
-  end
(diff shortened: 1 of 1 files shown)


hsbt:
@nithinbekal Thanks. This and https://github.com/ruby/ruby/pull/11757 are not document changes. 

I and @nobu updated titles and test them for merging.

nithinbekal:
@hsbt Thank you so much! I was looking at them mostly as a way to clarify the docs, but you're right - they do have code changes. Will keep that in mind in future PRs. 🙂

PR #11762, opened 2024-10-02
Add test for RbConfig Ractor safety

mohamedhafez:
In planning ahead to the near future when TruffleRuby can run C-extensions marked with `rb_ext_ractor_safe(true)` in parallel, and for when Ractors are no longer just experimental, I'm trying to get all commonly used C-extensions marked as `rb_ext_ractor_safe(true)`, would really appreciate it if you could take a look!

### DIFF
--- a/test/test_rbconfig.rb
+++ b/test/test_rbconfig.rb
@@ -51,4 +51,21 @@ class TestRbConfig < Test::Unit::TestCase
       assert_match(/\$\(sitearch|\$\(rubysitearchprefix\)/, val, "#{key} #{bug7823}")
     end
   end
+
+  def test_limits_and_sizeof_access_in_ractor
+    if defined?(Ractor)
+      assert_separately(["-W0"], <<~'RUBY')
+        r = Ractor.new do
+          sizeof_int = RbConfig::SIZEOF["int"]
+          fixnum_max = RbConfig::LIMITS["FIXNUM_MAX"]
+          [sizeof_int, fixnum_max]
+        end
+
+        sizeof_int, fixnum_max = r.take
+
+        assert_kind_of Numeric, sizeof_int, "RbConfig::SIZEOF['int'] should be a Numeric"
+        assert_kind_of Numeric, fixnum_max, "RbConfig::LIMITS['FIXNUM_MAX'] should be a Numeric"
+      RUBY
+    end
+  end
 end
--- a/test/test_rbconfig.rb
+++ b/test/test_rbconfig.rb
@@ -53,19 +53,17 @@ class TestRbConfig < Test::Unit::TestCase
   end
 
   def test_limits_and_sizeof_access_in_ractor
-    if defined?(Ractor)
-      assert_separately(["-W0"], <<~'RUBY')
-        r = Ractor.new do
-          sizeof_int = RbConfig::SIZEOF["int"]
-          fixnum_max = RbConfig::LIMITS["FIXNUM_MAX"]
-          [sizeof_int, fixnum_max]
-        end
+    assert_separately(["-W0"], <<~'RUBY')
+      r = Ractor.new do
+        sizeof_int = RbConfig::SIZEOF["int"]
+        fixnum_max = RbConfig::LIMITS["FIXNUM_MAX"]
+        [sizeof_int, fixnum_max]
+      end
 
-        sizeof_int, fixnum_max = r.take
+      sizeof_int, fixnum_max = r.take
 
-        assert_kind_of Numeric, sizeof_int, "RbConfig::SIZEOF['int'] should be a Numeric"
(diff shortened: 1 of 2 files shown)


eregon:
@ko1 Could you review this?

The `rbconfig/sizeof` extension defines two frozen Hashes, so it should be fine to access it from multiple Ractors.
Do they need to be made `Ractor.make_shareable` explicitly or it already works as-is since they are immutable/deeply frozen?

@mohamedhafez I think it would be helpful to add a test accessing `RbConfig::LIMITS["int"]` and `RbConfig::LIMITS["FIXNUM_MAX"]` in https://github.com/ruby/ruby/blob/master/test/test_rbconfig.rb.
It should use `assert_separately` and `if defined?(Ractor)` for consistency with other Ractor tests.

mohamedhafez:
@eregon so I don't have much experience with this, just taking my best stab at it (with some help from chatgpt i must admit 😅) I was thinking I just add the following method to https://github.com/ruby/ruby/blob/master/test/test_rbconfig.rb?

```ruby
def test_limits_access_in_ractor
  if defined?(Ractor)
    assert_separately([], <<~'RUBY')
      require 'rbconfig'
      require 'test/unit'
      include Test::Unit::Assertions

      r = Ractor.new do
        int_limit = RbConfig::LIMITS["int"]
        fixnum_max = RbConfig::LIMITS["FIXNUM_MAX"]
        [int_limit, fixnum_max]
      end

      int_limit, fixnum_max = r.take

      assert_kind_of Numeric, int_limit, "RbConfig::LIMITS['int'] should be a Numeric"
      assert_kind_of Numeric, fixnum_max, "RbConfig::LIMITS['FIXNUM_MAX'] should be a Numeric"
    RUBY
  end
end
```

If that's way off, maybe if you could point me to an existing test to pattern it off of?

eregon:
It looks OK except the first 3 lines are unnecessary.
See https://github.com/search?q=repo%3Aruby%2Fruby%20assert_separately&type=code for examples

mohamedhafez:
@eregon I'm guessing you meant to test `RbConfig::SIZEOF["int"]` instead of `RbConfig::LIMITS["int"]`, which doesn't seem to exist? Added the "-W0" option to the `assert_separately` call because it was failing because of a warning about Ractors being experimental, and tests should be passing now!

test/test_rbconfig.rb
```diff
@@ -51,4 +51,21 @@ def test_vendorarchdirs
...
+        r = Ractor.new do
+          sizeof_int = RbConfig::SIZEOF["int"]
+          fixnum_max = RbConfig::LIMITS["FIXNUM_MAX"]
+          [sizeof_int, fixnum_max]
+        end
+
+        sizeof_int, fixnum_max = r.take
+
+        assert_kind_of Numeric, sizeof_int, "RbConfig::SIZEOF['int'] should be a Numeric"
+        assert_kind_of Numeric, fixnum_max, "RbConfig::LIMITS['FIXNUM_MAX'] should be a Numeric"
```
eregon, line 67: It makes sense to check they are Integer not just Numeric

test/test_rbconfig.rb
```diff
@@ -51,4 +51,21 @@ def test_vendorarchdirs
       assert_match(/\$\(sitearch|\$\(rubysitearchprefix\)/, val, "#{key} #{bug7823}")
     end
   end
+
+  def test_limits_and_sizeof_access_in_ractor
+    if defined?(Ractor)
```
eregon, line 56: This `if` would be better after the `end` of the method (even though it may look weird it's standard practice in CRuby tests), that way the method isn't even defined if Ractor is not defined, which is good because otherwise in the output it would look like this test succeeds even though it didn't do anything, when Ractor is not defined,

mohamedhafez: Done! Tested, and the updated test is passing locally. 

(The previous push passed all tests in CI, so we should see CI report all is well in an hour or so)

test/test_rbconfig.rb
mohamedhafez: Done!

nobu:
It is a common rule that a class/module that has no ractor-unshareable variables/constants is ractor-shareable, IIRC.
`RbConfig::SIZEOF` and `RbConfig::LIMITS` are just the cases.

eregon:
Right, I guess rb_ext_ractor_safe() only affects methods defined in native extensions and not modules/classes/constants?

Would it be OK to still add to mark in the source code to ensure this extension remains Ractor-shareable?

eregon:
I think we should add this test or similar because e.g. it wouldn't work with Ractor if the Hash were not shareable/deeply-frozen.

There is no need to mark the extension with `rb_ext_ractor_safe(true);` since it defines no methods so that doesn't really have any effect.

eregon (approved):

PR #11768, opened 2024-10-02
[DOC] Tweaks for Array#max

BurdetteLamar:


array.c
```diff
@@ -5913,42 +5913,51 @@ ary_max_opt_string(VALUE ary, long i, VALUE vmax)
...
  *
  *  - The maximum-valued element from +self+.
- *  - A new +Array+ of maximum-valued elements selected from +self+.
+ *  - A new array of maximum-valued elements from +self+.
  *
- *  When no block is given, each element in +self+ must respond to method <tt><=></tt>
- *  with an Integer.
+ *  Does not modify +self+.
+ *
+ *  With no block given, each element in +self+ must respond to method <tt><=></tt>
```
peterzhu2118, line 5928: The way we refer methods is not consistent, in #11745 you referred to the method `#hash` with a pound sign but not using the pound sign here for `<=>`. I don't have strong opinions on whether to use the pound sign or not, but we should be consistent.

BurdetteLamar: Pound sign added.

PR #11772, opened 2024-10-02
Introduce `Kernel#not_nil!`

Morriar:
This PR introduces 2 methods:

* `Kernel#not_nil!` which raises if `self` is nil and returns `self` otherwise
* `Kernel#not_nil` which allows the user to specify the return value with a block in case `self` is nil

This feature was discussed https://bugs.ruby-lang.org/issues/17326 and the general consensus seems to be that the method is useful and should be named `not_nil!`.

cc. @jez, @mame, @soutaro

### DIFF
--- a/kernel.rb
+++ b/kernel.rb
@@ -71,6 +71,19 @@ module Kernel
 
   #
   #  call-seq:
+  #     obj.not_nil!
+  #
+  #  Raises a TypeError if <i>obj</i> is nil, returns <i>self</i> otherwise.
+  #
+  #     1.not_nil!   #=> 1
+  #     nil.not_nil!  #=> TypeError: Receiver of not_nil! is nil
+  #
+  def not_nil!
+    self
+  end
+
+  #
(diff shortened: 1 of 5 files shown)


halostatue:
I do not believe that there is consensus on the name or functionality *except* that `must!` is unacceptable to matz. The last communication from matz on this said:

> With a different name, there may be a chance for the method.

This should *probably* be reopened for discussion in the original ticket and put on the dev meeting prior to adding these methods. There is substantial opposition to this from multiple observers and substantial support from multiple observers with what I consider to be tepid support from Matz.

(I am *personally* against this in any form, whether `must!` or `not_nil!` and will reject its use from any codebase where I have such control. This can be *better* provided by a gem such that it is wholly opt-in for people who want to have this functionality.)

XrXr:
Closing since the proposal was rejected. https://bugs.ruby-lang.org/issues/17326#note-21

PR #11777, opened 2024-10-03
Fix size calcuation when offset is given + updated documentation.

ioquatix:
Unfortunately during the last backport I missed this important detail.

cc @nagachika


ioquatix:
I've added some tests to confirm the fix (as there was nothing before) and also created a PR to merge them to master: https://github.com/ruby/ruby/pull/11779

nagachika:
Looks good to me.
Currently annocheck is broken in the 3.2 branch (probably) due to the environment, so don't worry about it.

PR #11786, opened 2024-10-03
[DOC] Tweaks for Array#min

BurdetteLamar:


array.c
```diff
@@ -6081,42 +6081,51 @@ ary_min_opt_string(VALUE ary, long i, VALUE vmin)
...
+      [1, 0, 3, 2].min(9)   # => [0, 1, 2, 3]
+      [1, 0, 3, 2].min(0)   # => []
  *
- *  When a block is given, the block must return an Integer.
+ *  With a block given, the block must return a numeric.
  *
- *  With a block and no argument, calls the block <tt>self.size-1</tt> times to compare elements;
- *  returns the element having the minimum value per the block:
+ *  With a block and no argument, calls the block <tt>self.size - 1</tt> times to compare elements;
+ *  returns the element having the minimum return value from the block:
```
peterzhu2118, line 6116: The return value isn't the smallest return value of the block, but rather the smallest value based on the comparisons done using the block.

```suggestion
 *  returns the element having the minimum return value per the block:
```

peterzhu2118 (approved):

PR #11787, opened 2024-10-03
[DOC] Tweaks for Array#minmax

BurdetteLamar:


array.c
```diff
@@ -6160,26 +6160,25 @@ rb_ary_min(int argc, VALUE *argv, VALUE ary)
...
 /*
  *  call-seq:
- *    array.minmax -> [min_val, max_val]
- *    array.minmax {|a, b| ... } -> [min_val, max_val]
+ *    minmax -> array
+ *    minmax {|a, b| ... } -> array
  *
- *  Returns a new 2-element +Array+ containing the minimum and maximum values
- *  from +self+, either per method <tt><=></tt> or per a given block:.
+ *  Returns a 2-element array containing the minimum-valued and maximum values
```
peterzhu2118, line 6166:
```suggestion
 *  Returns a 2-element array containing the minimum-valued and maximum-valued
```

array.c
```diff
@@ -6160,26 +6160,25 @@ rb_ary_min(int argc, VALUE *argv, VALUE ary)
...
- *    [0, 1, 2].minmax # => [0, 2]
+ *    [1, 0, 3, 2].minmax # => [0, 3]
  *
- *  When a block is given, the block must return an Integer;
- *  the block is called <tt>self.size-1</tt> times to compare elements;
- *  returns a new 2-element +Array+ containing the minimum and maximum values
- *  from +self+, per the block:
+ *  With a block given, the block must return a numeric;
+ *  the block is called <tt>self.size - 1</tt> times to compare elements;
+ *  returns the elements having the minimum and maximum return values from the block:
```
peterzhu2118, line 6176: I think the "minimum and maximum return values" here makes it sound like it's based on the return values of the block, but the block just aids comparison:

```suggestion
 *  returns the elements having the minimum and maximum values per the block:
```

PR #11788, opened 2024-10-03
[DOC] Tweaks for Array#none?

BurdetteLamar:


array.c
```diff
@@ -7892,34 +7892,34 @@ rb_ary_all_p(int argc, VALUE *argv, VALUE ary)
...
  *  +false+ otherwise:
  *
- *    [nil, false].none? # => true
+ *    [nil, false].none?    # => true
  *    [nil, 0, false].none? # => false
- *    [].none? # => true
+ *    [].none?              # => true
  *
- *  With a block given and no argument, calls the block with each element in +self+;
+ *  With argument +object+ given, returns +true+ if <tt>object.===</tt> no element, +false+ otherwise:
```
peterzhu2118, line 7908: I found this sentence a bit confusing because the negative comes first, how about:

```suggestion
 *  With argument +object+ given, returns +false+ if for any element +element+,
 *  <tt>object === element</tt>; +true+ otherwise:
```

array.c
```diff
@@ -7892,34 +7892,34 @@ rb_ary_all_p(int argc, VALUE *argv, VALUE ary)
...
+ *
+ *    ['food', 'drink'].none?(/bar/) # => true
+ *    ['food', 'drink'].none?(/foo/) # => false
+ *    [].none?(/foo/)                # => true
+ *    [0, 1, 2].none?(3)             # => true
+ *    [0, 1, 2].none?(1)             # => false
+ *
+ *  With a block given , calls the block with each element in +self+;
```
peterzhu2118, line 7916:
```suggestion
 *  With a block given, calls the block with each element in +self+;
```

peterzhu2118 (approved):

PR #11795, opened 2024-10-04
Introduce `rb_io_blocking_region` which takes `struct rb_io` argument.

ioquatix:
This is in preparation for changes to `struct rb_io`. Instead of passing `fd`, we should pass the `struct rb_io`, so that we have access to all the internals. This does not change any visible behaviour, but funnels all calls to `rb_thread_io_blocking_region` via this interface, which can be later modified to handle `IO#close` more effectively.

I will do a few parts of the core conversion in this PR, and then follow up with separate PRs for `ext/socket` etc.

See the previous work <https://github.com/ruby/ruby/pull/5384> for more details.


ioquatix:
See <https://github.com/ruby/ruby/pull/11796> for `ext/socket`.

PR #11798, opened 2024-10-04
Reduce rb_enc_str_new function dependency for parser

S-H-GAMELINKS:
Introduce `rb_str_new_encoding_parser_string` function and Reduce `rb_enc_str_new` function dependency for parser


nobu:
This seems to increase copies between `VALUE` and `rb_parser_string_t`.
Right?

S-H-GAMELINKS:
That's right.
Should I fix it so that rb_node_sym_new receives rb_parser_string_t* so that no copies are made?

PR #11802, opened 2024-10-04
[DOC] Tweaks for Array#permutation

BurdetteLamar:


array.c
```diff
@@ -7000,82 +7000,41 @@ rb_ary_permutation_size(VALUE ary, VALUE args, VALUE eobj)
...
- *    [1, 0, 2]
- *    [1, 2, 0]
- *    [2, 0, 1]
- *    [2, 1, 0]
+ *  With a block and an in-range positive integer argument +n+ (<tt>0 < n <= self.size</tt>) given,
+ *  calls the block with each +n+-tuple permutations of +self+;
+ *  returns +self+:
  *
- *  When +n+ is zero, calls the block once with a new empty +Array+:
+ *    get_perms(a, 1) # => [[0], [1], [2]]
```
peterzhu2118, line 7021: I don't think you ever introduced what `a` is here.

array.c
```diff
@@ -7000,82 +7000,41 @@ rb_ary_permutation_size(VALUE ary, VALUE args, VALUE eobj)
...
+ *    get_perms(a, 2) # => [[0, 1], [0, 2], [1, 0], [1, 2], [2, 0], [2, 1]]
+ *    get_perms(a, 3) # => [[0, 1, 2], [0, 2, 1], [1, 0, 2], [1, 2, 0], [2, 0, 1], [2, 1, 0]]
```
peterzhu2118, line 7023: Alternatively, you can use `Enumerator#to_a` without needing to define a `get_perms` method:

```ruby
[1, 2, 3].permutation.to_a # => [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
```

BurdetteLamar: Yes, but that call does not have a block (as the examples require).

Is this better?

```
perms = []
[1, 2, 3].permutation(3) {|perm| perms.push(perm) }
perms # => [[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]]
```

array.c
BurdetteLamar: Fixed.

array.c
peterzhu2118: It's true that it does not call the block, but IMO, it's a lot more readable than calling `get_perms`.

BurdetteLamar: IMO, the examples must call the block (which is the primary form for the call-seq).  I've removed `get_perms`, which I agree is sub-optimal.

PR #11818, opened 2024-10-07
YJIT: Add compilation log

nirvdrum:
This PR introduces a new compilation log to help gain insights into how YJIT is optimizing an application. Finding the correct amount information to display for a given context is a bit hard. Some folks may want an entire log while others are only interested in the most recent events. This PR tries to strike a balance by offering a few different operating modes.

# Command-Line Flag

This PR introduces the `--yjit-compilation-log` flag, which functions like a hybrid of `--yjit-stats` and `--yjit-dump-disasm`. The presence of an optional argument will dictate where the compilation log is written.

## No Argument

~`--yjit-compilation-log` without an argument will print the compilation log to `$stderr` when the application exits. Events are buffered until the application exit in order to avoid mixing messages up with other output from the application. To constrain the total amount of memory used, a circular buffer is used to store the logged events. Only the entries present in the buffer at the time the application exits will be printed.~

`--yjit-compilation-log` without an argument will print the compilation log to `$stderr` as events occur. It will be logically equivalent to running with `--yjit-compilation-log=/dev/stderr`, just without file truncation. Events will still be collected in the internal circular buffer for programmatic access.

## Directory Argument

The `--yjit-compilation-log` flag can take an optional argument pointing at a directory (e.g., `--yjit-compilation-log=$PWD`). In this case, it functions very similarly to `--yjit-dump-disasm`: a new file named `yjit_compilation_$PID.log` will be created in the specified directory and all compilation events will be logged to that file. In this way, one directory can capture output from multiple processes. NB: the directory must exist when starting Ruby, otherwise the argument will be treated as if it were a file name (see below).

Unlike the case where output is sent to `$stderr`, all events will be written to the file as they occur. The circular buffer will still be populated in the event you wish to work with the log in memory (accessible via `RubyVM::YJIT.compilation_log`).

## File Argument

The `--yjit-compilation-log` flag can take an optional argument pointing at a file (e.g., `--yjit-compilation-log=my_compilation.log`). This works very similarly to the directory argument case, except instead of a file name being generated, it will use the one supplied.

The file will be created when the process starts. If the file already exists, it will be truncated. This supports a nice workflow whereby you can re-run the same `ruby` invocation repeatedly and avoid mixing results for multiple runs. Moreover, since the file is truncated rather than re-created, you can `tail -f my_compilation.log` without having to restart the `tail` command each time you re-run `ruby`.

# Programmatically Control

In addition to the command-line flag, you can enable or disable the compilation log via `RubyVM::YJIT.enable` and the `compilation_log` keyword argument. It is designed to function similarly to the programmatic control of YJIT's runtime stats. Principally, it accepts three values:

* `false`: Don't enable the compilation log (more of a placeholder since you can't call this method again to disable)
* `true`: Enable the compilation log and print it to `$stderr` at exit
* `:quiet`: Enable the compilation log, but do not write or print it

NB: There is not an option to enable the log and write to a file. If you want to operate in that mode you should supply the `--yjit-compilation-log` argument.

# Programmatic Access

## Check if Enabled

You can check if the YJIT compilation log is currently enabled with `RubyVM::YJIT.compilation_log_enabled?`.

## Reading the Log

The last _N=1024_ compilation log entries are accessible via `RubyVM::YJIT.compilation_log`. The output has the type `Array[Tuple[Time, String]]`, providing the log is enabled. If it is not, the value will be `nil` to help differentiate from a log where no compilation has occurred yet (i.e., an empty array) . The log is automatically cleared when accessed so subsequent reads will be guaranteed to contain new events.

### DIFF
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -166,10 +166,10 @@ You can dump statistics about compilation and execution by running YJIT with the
 ./miniruby --yjit-stats myscript.rb
 ```
 
-You can see what YJIT has compiled by running YJIT with the `--yjit-compilation-log` command-line option:
+You can see what YJIT has compiled by running YJIT with the `--yjit-log` command-line option:
 
 ```sh
-./miniruby --yjit-compilation-log myscript.rb
+./miniruby --yjit-log myscript.rb
 ```
 
 The machine code generated for a given method can be printed by adding `puts RubyVM::YJIT.disasm(method(:method_name))` to a Ruby script. Note that no code will be generated if the method is not compiled.
@@ -187,8 +187,8 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
   compiled, lower values mean less code is compiled (default 200K)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
 - `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost)
-- `--yjit-compilation-log[=file|dir]`: log all compilation events to the specified file or directory. If no name is supplied, the last 1024 log entries will be printed to stderr when the application exits.
-- `--yjit-compilation-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.compilation_log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
+- `--yjit-log[=file|dir]`: log all compilation events to the specified file or directory. If no name is supplied, the last 1024 log entries will be printed to stderr when the application exits.
+- `--yjit-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
 - `--yjit-disable`: disable YJIT despite other `--yjit*` flags for lazily enabling it with `RubyVM::YJIT.enable`
 - `--yjit-code-gc`: enable code GC (disabled by default as of Ruby 3.3).
   It will cause all machine code to be discarded when the executable memory size limit is hit, meaning JIT compilation will then start over.
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -166,6 +166,12 @@ You can dump statistics about compilation and execution by running YJIT with the
 ./miniruby --yjit-stats myscript.rb
 ```
 
+You can see what YJIT has compiled by running YJIT with the `--yjit-compilation-log` command-line option:
+
+```sh
+./miniruby --yjit-compilation-log myscript.rb
+```
+
 The machine code generated for a given method can be printed by adding `puts RubyVM::YJIT.disasm(method(:method_name))` to a Ruby script. Note that no code will be generated if the method is not compiled.
 
 <h3 id="command-line-options">Command-Line Options</h3>
@@ -181,6 +187,8 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
   compiled, lower values mean less code is compiled (default 200K)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
 - `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost)
+- `--yjit-compilation-log`: print a log of all compiled methods to the console.
+- `--yjit-compilation-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.compilation_log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
 - `--yjit-disable`: disable YJIT despite other `--yjit*` flags for lazily enabling it with `RubyVM::YJIT.enable`
 - `--yjit-code-gc`: enable code GC (disabled by default as of Ruby 3.3).
   It will cause all machine code to be discarded when the executable memory size limit is hit, meaning JIT compilation will then start over.
--- a/doc/yjit/yjit.md
+++ b/doc/yjit/yjit.md
@@ -187,7 +187,7 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
   compiled, lower values mean less code is compiled (default 200K)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
 - `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost)
-- `--yjit-compilation-log`: print a log of all compiled methods to the console.
+- `--yjit-compilation-log[=file|dir]`: log all compilation events to the specified file or directory. If no name is supplied, the last 1024 log entries will be printed to stderr when the application exits.
 - `--yjit-compilation-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.compilation_log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
 - `--yjit-disable`: disable YJIT despite other `--yjit*` flags for lazily enabling it with `RubyVM::YJIT.enable`
 - `--yjit-code-gc`: enable code GC (disabled by default as of Ruby 3.3).
--- a/yjit.rb
+++ b/yjit.rb
@@ -18,9 +18,9 @@ module RubyVM::YJIT
     Primitive.rb_yjit_stats_enabled_p
   end
 
-  # Check if `--yjit-compilation-log` is used.
-  def self.compilation_log_enabled?
-    Primitive.rb_yjit_compilation_log_enabled_p
+  # Check if `--yjit-log` is used.
+  def self.log_enabled?
+    Primitive.rb_yjit_log_enabled_p
   end
 
   # Check if rb_yjit_trace_exit_locations_enabled_p is enabled.
@@ -41,14 +41,14 @@ module RubyVM::YJIT
   # * `true`: Enable stats. Print stats at exit.
   # * `:quiet`: Enable stats. Do not print stats at exit.
   #
-  # `compilation_log`:
-  # * `false`: Don't enable the compilation log.
-  # * `true`: Enable the compilation log. Print compilation log at exit.
-  # * `:quiet`: Enable the compilation log. Do not print compilation log at exit.
-  def self.enable(stats: false, compilation_log: false)
+  # `log`:
+  # * `false`: Don't enable the log.
+  # * `true`: Enable the log. Print log at exit.
+  # * `:quiet`: Enable the log. Do not print log at exit.
+  def self.enable(stats: false, log: false)
     return false if enabled?
     at_exit { print_and_dump_stats } if stats
-    Primitive.rb_yjit_enable(stats, stats != :quiet, compilation_log, compilation_log != :quiet)
+    Primitive.rb_yjit_enable(stats, stats != :quiet, log, log != :quiet)
   end
 
   # If --yjit-trace-exits is enabled parse the hashes from
@@ -185,12 +185,12 @@ module RubyVM::YJIT
     strio.string
   end
 
-  # Return an array of compilation log entries.
+  # Return an array of log entries.
   # Return `nil` when option is not passed or unavailable.
-  def self.compilation_log
-    return nil unless compilation_log_enabled?
+  def self.log
+    return nil unless log_enabled?
 
-    Primitive.rb_yjit_get_compilation_log.map do |timestamp, path|
+    Primitive.rb_yjit_get_log.map do |timestamp, path|
       [Time.at(timestamp), path]
     end
   end
--- a/yjit.rb
+++ b/yjit.rb
@@ -186,6 +186,12 @@ module RubyVM::YJIT
     strio.string
   end
 
+  # Return an array of compilation log entries.
+  # Return `nil` when option is not passed or unavailable.
+  def self.compilation_log
+    Primitive.rb_yjit_get_compilation_log
+  end
+
   # Produce disassembly for an iseq. This requires a `--enable-yjit=dev` build.
   def self.disasm(iseq) # :nodoc:
     # If a method or proc is passed in, get its iseq
--- a/yjit.rb
+++ b/yjit.rb
@@ -191,8 +191,8 @@ module RubyVM::YJIT
   def self.compilation_log
     return nil unless compilation_log_enabled?
 
-    Primitive.rb_yjit_get_compilation_log.map do |path, timestamp|
-      [path, Time.at(timestamp)]
+    Primitive.rb_yjit_get_compilation_log.map do |timestamp, path|
+      [Time.at(timestamp), path]
     end
   end
 
--- a/yjit.rb
+++ b/yjit.rb
@@ -18,6 +18,11 @@ module RubyVM::YJIT
     Primitive.rb_yjit_stats_enabled_p
   end
 
+  # Check if `--yjit-compilation-log` is used.
+  def self.compilation_log_enabled?
+    Primitive.rb_yjit_compilation_log_enabled_p
+  end
+
   # Check if rb_yjit_trace_exit_locations_enabled_p is enabled.
   def self.trace_exit_locations_enabled? # :nodoc:
     Primitive.rb_yjit_trace_exit_locations_enabled_p
@@ -223,6 +228,10 @@ module RubyVM::YJIT
   # Avoid calling a Ruby method here to not interfere with compilation tests
   if Primitive.rb_yjit_stats_enabled_p
     at_exit { print_and_dump_stats }
+  end
+
+  if Primitive.rb_yjit_compilation_log_enabled_p
+    at_exit { $stderr.puts("***YJIT: Printing YJIT compilation log on exit***") if Primitive.rb_yjit_print_compilation_log_p }
   end
 
   class << self
--- a/yjit.rb
+++ b/yjit.rb
@@ -263,7 +263,10 @@ module RubyVM::YJIT
     # Print the compilation log
     def print_compilation_log # :nodoc:
       if Primitive.rb_yjit_print_compilation_log_p
-        $stderr.puts("***YJIT: Printing YJIT compilation log on exit***")
+        RubyVM::YJIT.compilation_log.each do |iseq_path, timestamp|
+          t = Time.at(timestamp)
+          $stderr.puts "%15.6f: %s" % [t.to_f, iseq_path]
+        end
       end
     end
 
--- a/yjit.rb
+++ b/yjit.rb
@@ -189,7 +189,11 @@ module RubyVM::YJIT
   # Return an array of compilation log entries.
   # Return `nil` when option is not passed or unavailable.
   def self.compilation_log
-    Primitive.rb_yjit_get_compilation_log
+    return nil unless compilation_log_enabled?
+
+    Primitive.rb_yjit_get_compilation_log.map do |path, timestamp|
+      [path, Time.at(timestamp)]
+    end
   end
 
   # Produce disassembly for an iseq. This requires a `--enable-yjit=dev` build.
@@ -263,9 +267,9 @@ module RubyVM::YJIT
     # Print the compilation log
     def print_compilation_log # :nodoc:
       if Primitive.rb_yjit_print_compilation_log_p
+        $stderr.puts "***YJIT: Printing YJIT compilation log on exit***"
         RubyVM::YJIT.compilation_log.each do |iseq_path, timestamp|
-          t = Time.at(timestamp)
-          $stderr.puts "%15.6f: %s" % [t.to_f, iseq_path]
+          $stderr.puts "%15.6f: %s" % [timestamp.to_f, iseq_path]
         end
       end
     end
--- a/yjit.rb
+++ b/yjit.rb
@@ -33,15 +33,23 @@ module RubyVM::YJIT
     Primitive.rb_yjit_reset_stats_bang
   end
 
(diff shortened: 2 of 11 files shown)


yjit/src/options.rs
```diff
@@ -128,6 +137,14 @@ pub enum TraceExits {
     Counter(Counter),
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum CompilationLogOutput {
+    // Dump to stdout when the process exits
+    Stdout,
+    // Dump to the log file as compilation events occur.
```
k0kubun, line 144: I expected the compilation log to be printed as it compiles. When you use the `Stdout` mode and compilation logs are printed in between yjit-bench iterations, you can know which methods are long-tail problems. But if it's only printed at the end, it's harder to understand when things are compiled. I would also like `File` to print logs as it compiles so that I can tail the file.

Basically, I don't get why you need to buffer outputs and print all of them at the end. The feature seems closer to `--yjit-dump-disasm` (which doesn't print stuff at exit) than `--yjit-stats` (which prints stats at exit).

yjit/src/core.rs
```diff
@@ -3229,6 +3234,8 @@ pub fn gen_entry_point(iseq: IseqPtr, ec: EcPtr, jit_exception: bool) -> Option<
     // Count the number of entry points we compile
     incr_counter!(compiled_iseq_entry);
 
+    CompilationLog::add_iseq(blockid);
```
k0kubun, line 3237: This place only hooks interpreter -> JIT calls, so it misses JIT -> JIT calls.

Since YJIT is not a method JIT and a compilation unit is a basic block, I think it should log blocks instead of methods. Otherwise, it would hide some long-tail compilations we probably don't want to miss.

yjit/src/codegen.rs
```diff
@@ -6567,6 +6568,9 @@ fn gen_send_cfunc(
                 perf_call!("gen_send_cfunc: ", known_cfunc_codegen(jit, asm, ci, cme, block, argc, recv_known_class))
             };
 
+            let method_id = unsafe { (*cme).called_id };
+            CompilationLog::add_cfunc(recv_known_class, method_id);
```
k0kubun, line 6572: I still don't understand the motivation to include the logs of C method calls, especially when you don't provide a convenient (instead of letting the user match "\<cfunc\>") to filter them out.

There's an issue related to this project in our team repository, which is titled "Tooling to detect dynamic code generation". We want to find Ruby code that is dynamically generated and compiled by YJIT. C method calls are just details of what's compiled in such Ruby code, so it doesn't seem to help you "detect" dynamic code generation.

yjit/src/options.rs
```diff
@@ -82,6 +82,12 @@ pub struct Options {
 
     /// Enable writing /tmp/perf-{pid}.map for Linux perf
     pub perf_map: Option<PerfMap>,
+
+    // Log compilations
+    pub gen_compilation_log: bool,
+
+    // Print compilation log on exit (when gen_compilation_log is also true)
+    pub print_compilation_log: Option<CompilationLogOutput>,
```
maximecb, line 90: Nitpick but it seems like these two variables could be one thing.

CompilationLog could have 4 values:
Disabled
Quiet,
Stdout,
File,

yjit/src/compilation_log.rs
```diff
@@ -0,0 +1,242 @@
...
+
+    /// The compilation event payload.
+    pub payload: CompilationLogPayload,
+}
+
+#[derive(Copy, Clone, Debug)]
+pub enum CompilationLogPayload {
+    ISeq(BlockId),
+    CFunc(Option<VALUE>, ID)
+}
```
maximecb, line 25: Seems potentially risky to store BlockIds and `VALUE` in the compilation log. Some of these things can get GC'd and potentially move... Imo it would be safer to convert compilation log entries to strings as soon as they are added.

WDYT @k0kubun ?

k0kubun: Yeah, the ISEQ of `BlockId` and `VALUE` need to be marked/updated on GC. It's possible but you often miss edge cases whenever you store `BlockId`.

Conversion to strings may bring allocation overhead. I think it's probably fine since we would run it only in the few servers that run `--yjit-stats`. But if we're interested in speeding it up, we could accept only limited characters for each entry and reuse the same string buffers, truncating entries if too long.

yjit/src/options.rs
nirvdrum: I'm happy to do it that way. I structurally mirrored what we're doing with stats, where we have `gen_stats` and `print_stats`. I didn't really love having two separate values anyway.

yjit/src/compilation_log.rs
nirvdrum: If `BlockId` can be GC'd, that'll be a problem. I thought `CFunc` was only storing references to the core classes we substitute. But, I'll need to double-check that's accurate.

Eagerly resolving the string would avoid any of that headache. We'd just have to accept it'll consume more memory and there may be wasted resolutions if an entry is overwritten before retrieved from the circular buffer. If that's not a concern, I'll make the change.

k0kubun: > If `BlockId` can be GC'd

`BlockId` is a pair of an ISEQ pointer and a u16 integer. The ISEQ address could be freed or moved.

maximecb: imo the simplest and safest option is to eagerly convert to a `String`. Since we only store 1024 entries by default, memory usage should not be a concern. We could be talking about ~128KiB when the compilation log is enabled.

k0kubun: I was concerned not about the size but about the speed. With a String, I assume updating the String may require heap allocation every time and be somewhat slow. But yeah, it's probably fine too.

nobu:
In the commit log, at the beginning of "Directory Argument" and "File Argument", 

> The `--yjit-compilation-log` falg can take an optional argument 

"falg" looks like a typo.

yjit/src/options.rs
nirvdrum: It now prints to `$stderr` as compilation occurs. That obviated the need for this enum entirely, so I've also removed that.

nirvdrum: Actually, we still have the enum. It turns out that `std::os:fd::AsRawFd` wasn't stabilized until Rust 1.66.0. I've changed my local Rust to 1.58.0 to avoid surprises like that going forward.

yjit/src/core.rs
nirvdrum: That was my misunderstanding. I thought we wanted to keep the volume down. I'm happy to add it.

yjit/src/codegen.rs
nirvdrum: I guess this was my misunderstanding as well. I thought we'd want to include it in the log because it generates code at runtime. If that's not useful it's easy to remove.

k0kubun: > I thought we'd want to include it in the log because it generates code at runtime.

You just need to log `gen_single_block` and drop this `gen_send_cfunc` logging. Every code generation goes through `gen_single_block`, and `gen_send_cfunc` is just one of the functions that `gen_single_block` may call. If the purpose is to cover everything that "generates code at runtime", you only need to log `gen_single_block`.

yjit/src/core.rs
nirvdrum: I've removed logging of method entry points and only log in `gen_single_block`, mostly duplicating the log message format already in use by `dump_disasm` feature. We can tweak this as needed.

yjit/src/codegen.rs
nirvdrum: There's more to what I had in mind, but it doesn't particularly matter. I've followed your guidance and dropped logging this event.

yjit/src/codegen.rs
```diff
@@ -9075,7 +9079,7 @@ fn get_class_name(class: Option<VALUE>) -> String {
 }
 
 /// Assemble "{class_name}#{method_name}" from a class pointer and a method ID
-fn get_method_name(class: Option<VALUE>, mid: u64) -> String {
+pub fn get_method_name(class: Option<VALUE>, mid: u64) -> String {
     let class_name = get_class_name(class);
     let method_name = if mid != 0 {
         unsafe { cstr_to_rust_string(rb_id2name(mid)) }
```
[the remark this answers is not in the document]
nirvdrum: I've removed these usages and now use `iseq_get_location`, which is already in _utils.rs_. I didn't know about this method prior to the code review, so it was a good discovery.

yjit/src/compilation_log.rs
nirvdrum: I've updated the code to eagerly construct the `String` instances. However, I couldn't find a way to use a fixed-length buffer without resorting to `MaybeUninit`. Since memory optimization doesn't seem to be required here, I've moved it over to a `Vec` to keep the code simple.

nirvdrum:
> "falg" looks like a typo.

Thank you, @nobu. I've updated the PR description. I grepped the code and the commit log and couldn't find that particular typo anywhere else in this PR.

yjit/src/options.rs
nirvdrum: I've consolidated the options. I left the runtime stats one as is. I don't know if we want to take a similar pass on those as well, but it's out of scope for this PR in any event.

yjit.rb
```diff
@@ -18,6 +18,11 @@ def self.stats_enabled?
     Primitive.rb_yjit_stats_enabled_p
   end
 
+  # Check if `--yjit-compilation-log` is used.
+  def self.compilation_log_enabled?
```
maximecb, line 22: Don't hate me but, could we simplify all occurrences of `compilation_log` (including `compilation_log.rs` filename) and just rename them to log? We could talk about this as the YJIT log? I don't think we'll want to have multiple separate log streams in YJIT (because that would be inherently repulsive).

maximecb:
Testing out the compilation log output:

```
1728499309.025884: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:50
1728499309.025957: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:50
1728499309.026126: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:50
1728499309.026158: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:52
1728499309.026211: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:52
1728499309.026278: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:52
1728499309.026352: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:52
1728499309.026460: assign_score_of@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:52
```

What does it mean when a method shows up multiple times in a row? Are we compiling the same entry point several times in a row?

In this case:
```
1728499309.052179: block (2 levels) in render_segment@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/for.rb:157
1728499309.052258: block (2 levels) in render_segment@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/for.rb:158
1728499309.052362: block (2 levels) in render_segment@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/for.rb:158
1728499309.052473: block (2 levels) in render_segment@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/for.rb:158
1728499309.052775: block (2 levels) in render_segment@/Users/maximecb/.gem/ruby/3.4.0/bundler/gems/liquid-4d4a565e98c1/lib/liquid/tags/for.rb:158
```

Does 2 levels mean chain depth 2? In which case, could we print `block (depth 2)`?

k0kubun:
> Does 2 levels mean chain depth 2? In which case, could we print block (depth 2)?

It's the depth of the Ruby block. It's just how CRuby labels block ISEQs.

> What does it mean when a method shows up multiple times in a row? Are we compiling the same entry point several times in a row?

That Ruby code seems to have multiple basic blocks in the same line. If we want to clarify that, we could print the index to the instruction of an ISEQ.

maximecb:
Ok I got confused because I thought block (2 levels) was a reference to a YJIT block with chain depth 2.

Might be good to print the iseq index as Kokubun suggests.

yjit/src/compilation_log.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    pub fn clear() {
+        unsafe {
+            COMPILATION_LOG.as_mut().unwrap().reset()
+        }
+    }
+}
+
+pub struct CircularBuffer<T, const N: usize> {
```
XrXr, line 97: Can you explain why you chose to implement this instead of using [std::collections::VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html)?

nirvdrum:
> That Ruby code seems to have multiple basic blocks in the same line. If we want to clarify that, we could print the index to the instruction of an ISEQ.

I took the log format from what we [we're using in diassembly dumps](https://github.com/ruby/ruby/blob/a838f980f599d95ccf344157f7074e997af31f48/yjit/src/codegen.rs#L1222). It doesn't need to be in this PR, but should we update as well?

k0kubun:
I don't think it's worth modifying CRuby (or deviating from what it uses) for it. It's CRuby's common format, which is also used in backtraces.

yjit/src/options.rs
```diff
@@ -128,6 +133,18 @@ pub enum TraceExits {
     Counter(Counter),
 }
 
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum CompilationLogOutput {
+    // Disable the log entirely
+    Disabled,
```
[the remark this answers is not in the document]
nirvdrum: I've restored the `Option<T>`.

yjit/src/compilation_log.rs
nirvdrum: Up until I made adjustments from review feedback the circular buffer was a fixed size. I was under the impression that's what we wanted. We can change to a standard collection easily enough.

yjit/src/options.rs
```diff
@@ -310,6 +328,28 @@ pub fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
...
+                    format!("{arg_value}/yjit_compilation_{}.log", std::process::id())
+                } else {
+                    arg_value.to_string()
+                };
+
+                match File::options().create(true).write(true).truncate(true).open(&log_file_path) {
+                    Ok(file) => {
+                        use std::os::unix::io::IntoRawFd;
+                        eprintln!("YJIT compilation log: {log_file_path}");
+                        unsafe { OPTIONS.compilation_log = CompilationLogOutput::File(file.into_raw_fd()) }
```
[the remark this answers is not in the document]
nirvdrum: Good catch. It was a bad rebase. I'll fix it.

nirvdrum:
> That Ruby code seems to have multiple basic blocks in the same line. If we want to clarify that, we could print the index to the instruction of an ISEQ.

> I don't think it's worth modifying CRuby (or deviating from what it uses) for it. It's CRuby's common format, which is also used in backtraces.

I'm understanding this to mean that we don't want to modify the format in `iseq_get_location`. How do you want the index to appear in the log line? It impacts whether I just append text to the output to `iseq_get_location` or make a more invasive change to the formatting. A sample line of what you'd like to see in the log would help.

k0kubun:
> How do you want the index to appear in the log line?
> A sample line of what you'd like to see in the log would help.

Since I'm personally used to adding stuff inside parens like `(foo: xxx, bar: yyy)` locally for debugging, I would like:

```
assign_score_of@/path/to/liquid-4d4a565e98c1/lib/liquid/tags/assign.rb:50 (iseq_idx: 0, chain_depth: 1)
```

yjit/src/compilation_log.rs
nirvdrum: A requirement for this project was to bound the size of the collection, with the collection always containing the N most recent log entries (specifically, a circular buffer). Looking at the `VecDeque` docs, I don't see a way to use a `VecDeque` in such a way. I could layer the semantics on top of that, but I think we're largely back to the current implementation.

I'd prefer not rolling my own, but we can't add dependencies either. If the concern is the `clone()`, that really shouldn't have been there. I was just trying to get the iterator working when I switched the payload out to `String`. I can remove that easily enough. I suppose the easiest thing is to just use `pop()`. The buffer is already being cleared once fetched by Ruby. The most straightforward would be:

```rust
 while log.len() > 0 {
        let entry = log.pop().unwrap();

        unsafe {
            let entry_array = rb_ary_new_capa(2);
            rb_ary_push(entry_array, rb_float_new(entry.timestamp));
            rb_ary_push(entry_array, entry.message.into());
            rb_ary_push(array, entry_array);
        }
    }
```

I could update the iterator's `next()` to call `pop()`. That'd let us use a `for` loop, which may look a bit more idiomatic. But, it may be unconventional for the iterator to mutate the backing store. In this case, the store isn't being physically resized so we don't have to deal with moving memory locations.

Do you have a preferred path?

nirvdrum: I've gone with the `while` loop for now, but it's a separate commit that can be dropped easily. If that's the way we go I can squash it back into the original and remove the iterator entirely.

XrXr: I'm more okay with hand-rolling this with the iterator gone. 👍🏼

yjit/src/compilation_log.rs
```diff
@@ -0,0 +1,180 @@
...
+        self.buffer[self.head] = Some(value);
+        self.head = (self.head + 1) % N;
+        if self.size == N {
+            self.tail = (self.tail + 1) % N;
+        } else {
+            self.size += 1;
+        }
+    }
+
+    #[allow(dead_code)]
```
XrXr, line 122:
```suggestion
```
It's used now.

yjit/src/compilation_log.rs
```diff
@@ -0,0 +1,180 @@
...
+                CompilationLogOutput::MemoryOnly => () // Don't print or write anything
+            }
+        }
+
+        Self::get_instance().push(entry);
+    }
+}
+
+pub struct CircularBuffer<T, const N: usize> {
+    buffer: Vec<Option<T>>,
```
XrXr, line 96:
```suggestion
    buffer: [T; N],
```

An array here should work just fine.

nirvdrum: Thanks. I think the iterator was getting in the way because `String` isn't `Copy`. I like this better.

Why do you suggest getting rid of the `Option`? I've been using `Option<T>#take` to clear the value once read. I think it makes `pop()` look nicer.

XrXr: Oh sorry, I meant `[Option<T>; N]`. That should make it easy to initialize too right?

XrXr: nvm, this is not a big issue, and in the past we've ran into footguns with accidentally creating big arrays on the stack. Let's just leave it as is. https://github.com/rust-lang/rust/issues/53827

yjit.rb
nirvdrum: @maximecb I've gone ahead and renamed everything to simplify to "log". I've isolated all of those changes to [5f81f1b](https://github.com/ruby/ruby/pull/11818/commits/5f81f1b31d7919c1ff9505ead8119c2a591c0ae6) for easy review.

doc/yjit/yjit.md
```diff
@@ -166,6 +166,12 @@ You can dump statistics about compilation and execution by running YJIT with the
 ./miniruby --yjit-stats myscript.rb
 ```
 
+You can see what YJIT has compiled by running YJIT with the `--yjit-compilation-log` command-line option:
+
+```sh
+./miniruby --yjit-compilation-log myscript.rb
+```
```
maximecb, line 173:
```suggestion
```sh
./miniruby --yjit-log myscript.rb
```
```

File: doc/yjit/yjit.md [hunk 14]
```diff
@@ -181,6 +187,8 @@ YJIT supports all command-line options supported by upstream CRuby, but also add
   compiled, lower values mean less code is compiled (default 200K)
 - `--yjit-stats`: print statistics after the execution of a program (incurs a run-time cost)
 - `--yjit-stats=quiet`: gather statistics while running a program but don't print them. Stats are accessible through `RubyVM::YJIT.runtime_stats`. (incurs a run-time cost)
+- `--yjit-compilation-log[=file|dir]`: log all compilation events to the specified file or directory. If no name is supplied, the last 1024 log entries will be printed to stderr when the application exits.
+- `--yjit-compilation-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
```
maximecb, line 191:
```suggestion
- `--yjit-log[=file|dir]`: log all compilation events to the specified file or directory. If no name is supplied, the last 1024 log entries will be printed to stderr when the application exits.
- `--yjit-log=quiet`: gather a circular buffer of recent YJIT compilations. The compilation log entries are accessible through `RubyVM::YJIT.log` and old entries will be discarded if the buffer is not drained quickly. (incurs a run-time cost)
```

yjit.rb
maximecb: Thank you Kevin and also thanks for your patience 🙏

doc/yjit/yjit.md
nirvdrum: Thanks. I missed a grep pattern. I've searched through the rest of the code for "compilation-log" and amended the last commit.

PR #11819, opened 2024-10-07
YJIT: Fastpath for Module#name

XrXr:
Module#name shows up as a top C method callee in lobsters so probably
common enough. It's also easy to substitute thanks to rb_mod_name()
already having no GC yield points.

    klass = BasicObject
    50_000_000.times { klass.name }

    Benchmark 1: /.rubies/post/bin/ruby --yjit mod_name.rb
      Time (mean ± σ):      1.433 s ±  0.010 s    [User: 1.410 s, System: 0.010 s]
      Range (min … max):    1.421 s …  1.449 s    10 runs

    Benchmark 2: /.rubies/mstr/bin/ruby --yjit mod_name.rb
      Time (mean ± σ):      1.491 s ±  0.012 s    [User: 1.468 s, System: 0.010 s]
      Range (min … max):    1.470 s …  1.511 s    10 runs

    Summary
      /.rubies/post/bin/ruby --yjit mod_name.rb ran
        1.04 ± 0.01 times faster than /.rubies/mstr/bin/ruby --yjit mod_name.rb

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -97,6 +97,8 @@ rb_namespace_p(VALUE obj)
  * to not be anonymous. <code>*permanent</code> is set to 1
  * if +classpath+ has no anonymous components. There is no builtin
  * Ruby level APIs that can change a permanent +classpath+.
+ *
+ * YJIT needs this function to not allocate.
  */
 static VALUE
 classname(VALUE klass, bool *permanent)
@@ -127,6 +129,7 @@ rb_mod_name0(VALUE klass, bool *permanent)
 VALUE
 rb_mod_name(VALUE mod)
 {
+    // YJIT needs this function to not allocate.
     bool permanent;
     return classname(mod, &permanent);
 }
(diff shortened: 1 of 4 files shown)


k0kubun (approved):

maximecb:
Thanks Alan

I'm inclined to merge this PR, but wondering if we can do even better performance-wise?

Can the name of a class be treated as a constant? Presumably we can know the class name in many cases? We do have `_known_recv_class: Option<VALUE>`? Could we return a frozen string if the class is known?

XrXr:
It's possible to return a frozen string inline when the class/module name is permanent, but it doesn't feel like it'd be a total win, since we'd need to guard on the identity of the receiver and take on the costs that come with that (`_known_recv_class` will mostly be exactly `::Module` or `::Class`, which has no bearing on the name of the receiver). Since `rb_mod_name()` is very simple already, there doesn't seem to be much to gain.

maximecb:
Oh, so this doesn't get called for something like `String#name` ?

XrXr:
I'm not sure what `String#name` would do, but yeah I'd imagine most calls to `Module#name` are direct calls.

PR #11820, opened 2024-10-07
Update references to flori/json

byroot:
Now that the repository was transfered, these links will become dead in a few months.


nobu:
Old change logs should not be subjects to update, I think.

byroot:
Why not? These links will end up broken, if someone is looking into the history to understand why a change was done, it is helpful to have valid links.

But if you feel strongly against it, I can revert that part.

PR #11839, opened 2024-10-08
[DOC] Tweaks for Array#repeated_combination

BurdetteLamar:


array.c
```diff
@@ -7369,65 +7369,56 @@ rb_ary_repeated_combination_size(VALUE ary, VALUE args, VALUE eobj)
...
- *    e = a.repeated_combination(1)
+ *
+ *    e = [0, 1, 2].repeated_combination(1)
  *    e.size # => 3
  *    e.to_a # => [[0], [1], [2]]
- *    e = a.repeated_combination(2)
+ *
+ *    e = [0, 1, 2].repeated_combination(2)
  *    e.size # => 6
  *    e.to_a # => [[0, 0], [0, 1], [0, 2], [1, 1], [1, 2], [2, 2]]
```
peterzhu2118, line 7417: I'm not sure what the purpose of these examples are, only the 0 case is new, otherwise the other two are already shown above.

PR #11845, opened 2024-10-09
Debug LIBRUBY_RELATIVE on Windows

deivid-rodriguez:
I noticed that `RbConfig::CONFIG` does not have the `LIBRUBY_RELATIVE` key on `mswin`. The `rbinstall` script relies on that for deciding whether it needs to include a bash prolog in binstubs: https://github.com/ruby/ruby/blob/823f29a36edd25136999d5a649eaff5e4a41481c/tool/rbinstall.rb#L418-L426.

So I believe that's no working correctly on mswin because of that.

I wonder if this special code I'm removing here is what's breaking that.

### DIFF
--- a/tool/mkconfig.rb
+++ b/tool/mkconfig.rb
@@ -20,7 +20,6 @@ mkconfig = File.basename($0)
 
 fast = {'prefix'=>true, 'ruby_install_name'=>true, 'INSTALL'=>true, 'EXEEXT'=>true}
 
-win32 = /mswin/ =~ arch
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@nobu fixed this with https://github.com/ruby/ruby/commit/3aa044f10c1db1608974ba4bc282c7dd1c9db3f5, so closing. Maybe it'd be good to find out why this special case is needed, since it does not make any tests fail, but not this time :)

PR #11851, opened 2024-10-09
String#forcible_encoding?

kddnewton:
Introduce String#forcible_encoding?(encoding) to check if it is possible to cleanly force an encoding onto a string without it breaking.

### DIFF
--- a/string.c
+++ b/string.c
@@ -11372,6 +11372,41 @@ rb_str_valid_encoding_p(VALUE str)
 
 /*
  *  call-seq:
+ *    forcible_encoding?(encoding) -> true or false
+ *
+ *  Returns +true+ if when +self+ is forced into the given encoding it will be
+ *  encoded correctly, +false+ otherwise:
+ *
+ *    "\xc2\xa1".forcible_encoding?("UTF-8") # => true
+ *    "\xc2".forcible_encoding?("UTF-8")     # => false
+ *    "\x80".forcible_encoding?("UTF-8")     # => false
+ */
+
+static VALUE
+rb_str_forcible_encoding_p(VALUE str, VALUE enc)
+{
+    rb_encoding *encoding = rb_to_encoding(enc);
+    int idx = rb_enc_to_index(encoding);
+
+    // If the encoding is the same as the receiver's encoding then this is
+    // forcible by identity.
+    if (ENCODING_GET(str) == idx) {
+        return rb_str_valid_encoding_p(str);
+    }
+
+    // If the coderange was 7bit and the new encoding is ASCII-compatible then
+    // this is forcible by definition.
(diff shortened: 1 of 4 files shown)


string.c
```diff
@@ -11310,6 +11310,51 @@ rb_str_force_encoding(VALUE str, VALUE enc)
...
+
+    rb_enc_associate_index(str, idx);
+
+    ENC_CODERANGE_CLEAR(str);
+    VALUE result = RBOOL(rb_enc_str_coderange(str) != ENC_CODERANGE_BROKEN);
+
+    rb_enc_associate_index(str, original_encoding);
+    ENC_CODERANGE_SET(str, original_coderange);
+
+    return result;
```
nobu, line 11355:
```suggestion
    if (ENC_CODERANGE(str) == ENC_CODERANGE_7BIT && encoding && rb_enc_asciicompat(encoding)) {
        return Qtrue;
    }

    return RBOOL(rb_enc_str_coderange_scan(str, encoding) != ENC_CODERANGE_BROKEN);
```

string.c
```diff
@@ -11310,6 +11310,51 @@ rb_str_force_encoding(VALUE str, VALUE enc)
...
+ *    "\xc2\xa1".forcible_encoding?("UTF-8") # => true
+ *    "\xc2".forcible_encoding?("UTF-8")     # => false
+ *    "\x80".forcible_encoding?("UTF-8")     # => false
+ */
+
+static VALUE
+rb_str_forcible_encoding_p(VALUE str, VALUE enc)
+{
+    str_modifiable(str);
+
```
nobu, line 11329:
```suggestion
```

test/ruby/test_m17n.rb
```diff
@@ -1579,6 +1579,17 @@ def test_force_encoding
...
 
+  def test_forcible_encoding
+    a = "\xC3"
+    refute(a.forcible_encoding?(Encoding::UTF_8))
+
+    a = "\xC3\xB6"
+    assert(a.forcible_encoding?(Encoding::UTF_8))
+
+    a = "\x80\x00"
+    assert(a.forcible_encoding?(Encoding::UTF_16LE))
```
nobu, line 1590:
```suggestion
    a = -"\xC3"
    assert_not_operator(a, :forcible_encoding?, Encoding::UTF_8)

    a = -"\xC3\xB6"
    assert_operator(a, :forcible_encoding?, Encoding::UTF_8)

    a = -"\x80\x00"
    assert_operator(a, :forcible_encoding?, Encoding::UTF_16LE)
```

nobu:
`"\xc2A".valid_encoding?` is false, but  should `"\xc2A".forcible_encoding?("UTF-8")` be true?

kddnewton:
> `"\xc2A".valid_encoding?` is false, but should `"\xc2A".forcible_encoding?("UTF-8")` be true?

That's a good point! I think it should probably return `false`. I will update the PR.

PR #11859, opened 2024-10-09
`me->defined_class` should be T_CLASS/T_ICLASS

ko1:
`me->defined_class` will be used for ancestor searching so that it should be T_CLASS or T_ICLASS.

This patch will resoleve https://github.com/ruby/ruby/pull/11715


ko1:
memo: the following should run without an error.

```ruby
class C0
  def foo
    p :C0_foo
  end
end

module M
  def foo
    p :M_foo
    super
  end
end

class C < C0
  include M
end

module R
  refine M do
    def foo
      p :refined
      super
    end
  end
end

using R

class C
  alias aliased_foo foo
end

C.new.aliased_foo```

PR #11865, opened 2024-10-10
Introduce backref_with parameterizing rules

S-H-GAMELINKS:
`backref tOP_ASGN lex_ctxt command_rhs` and `backref tOP_ASGN lex_ctxt arg_rhs` has similar pattern.
So, I thought better to cut out these pattern, and introduce backref_with_rhs parameterizing rules.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -2907,6 +2907,15 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
 /*
  *	parameterizing rules
  */
+%rule backref_with_rhs(value) <node>
+                : backref tOP_ASGN lex_ctxt value
+                    {
+                        VALUE MAYBE_UNUSED(e) = rb_backref_error(p, $1);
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -2911,6 +2911,15 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
 /*
  *	parameterizing rules
  */
+%rule backref_with_rhs(value) <node>
```
nobu, line 2914: There are other `backref` errors in `mlhs_node` and `lhs`, and these have also RHS as they are LHS.
This name doesn't look appropriate for op-assign rules only.

S-H-GAMELINKS: Certainly.
I'm thinking of renaming it backref_with, but how do you think?

S-H-GAMELINKS: It was renamed to `backref_with`.

PR #11866, opened 2024-10-10
Improve RUBY_GC_LIBRARY

peterzhu2118:
Instead of passing the full GC SO file name to RUBY_GC_LIBRARY, we now only need to pass the GC name.

For example, before we needed to pass `RUBY_GC_LIBRARY=librubygc.default.so` but now we only need to pass `RUBY_GC_LIBRARY=default`.


gc.c
```diff
@@ -659,25 +659,33 @@ ruby_external_gc_init(void)
...
+        {
+            char *curr_gc_so_path = gc_so_path;
+#define GC_SO_PATH_APPEND(str) do { \
+    strcpy(curr_gc_so_path, str); \
+    curr_gc_so_path += strlen(str); \
+} while (0)
+           GC_SO_PATH_APPEND(SHARED_GC_DIR);
+           GC_SO_PATH_APPEND("librubygc.");
+           GC_SO_PATH_APPEND(gc_so_file);
+           GC_SO_PATH_APPEND(SOEXT);
```
nobu, line 687:
```suggestion
            GC_SO_PATH_APPEND(SHARED_GC_DIR);
            GC_SO_PATH_APPEND("librubygc.");
            GC_SO_PATH_APPEND(gc_so_file);
            GC_SO_PATH_APPEND(SOEXT);
```

configure.ac
```diff
@@ -3403,6 +3403,8 @@ AC_DEFINE_UNQUOTED(DLEXT_MAXLEN, `expr $len + 1`)
 test ".$DLEXT"  = "." || AC_DEFINE_UNQUOTED(DLEXT,  ".$DLEXT")
 AC_SUBST(DLEXT)
 
+AC_DEFINE_UNQUOTED(SOEXT, ".$SOEXT")
```
nobu, line 3406: Including a dot, only in config.h?

peterzhu2118: I'm not sure what you mean. This is similar to DLEXT, which includes a dot at the beginning.

nobu: Sorry, indeed.

PR #11868, opened 2024-10-10
Remove 1 allocation in Enumerable#each_with_index

XrXr:
Previously, each call to Enumerable#each_with_index allocates 2 objects, one for the counting index, the other an imemo_ifunc passed to `self.each` as a block.

Use `struct vm_ifunc::data` to hold the counting index directly to remove 1 allocation.

### DIFF
--- a/enum.c
+++ b/enum.c
@@ -2984,13 +2984,12 @@ enum_member(VALUE obj, VALUE val)
 }
 
 static VALUE
-each_with_index_i(RB_BLOCK_CALL_FUNC_ARGLIST(i, memo))
+each_with_index_i(RB_BLOCK_CALL_FUNC_ARGLIST(_, index))
 {
-    struct MEMO *m = MEMO_CAST(memo);
-    VALUE n = imemo_count_value(m);
+    struct vm_ifunc *ifunc = rb_current_ifunc();
+    ifunc->data = (const void *)rb_int_succ(index);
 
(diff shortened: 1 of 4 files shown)


enum.c
```diff
@@ -2984,13 +2984,12 @@ enum_member(VALUE obj, VALUE val)
 }
 
 static VALUE
-each_with_index_i(RB_BLOCK_CALL_FUNC_ARGLIST(i, memo))
+each_with_index_i(RB_BLOCK_CALL_FUNC_ARGLIST(_, index))
 {
-    struct MEMO *m = MEMO_CAST(memo);
-    VALUE n = imemo_count_value(m);
+    struct vm_ifunc *ifunc = rb_current_ifunc();
+    ifunc->data = (const void *)rb_int_succ(index);
```
maximecb, line 2990: Would you mind adding a comment explaining what the ifunc does, what is its purpose etc?

This code can be fairly opaque to those not as familiar with CRuby internals.

maximecb (approved):
Nicely done Alan

enum.c
XrXr: [c018336](https://github.com/ruby/ruby/pull/11868/commits/c018336e794403406820cd6af09e3728b972f873)
Sure! I think this kind of commentary sits better on the struct, though.

PR #11872, opened 2024-10-10
Query the actively loaded GC module

eightbitraptor:
This PR adds information about the currently loaded GC module to `RUBY_DESCRIPTION` and hence the output of `ruby -v` when `--with-shared-gc` is enabled at compile time.

Additionally, when a shared GC module is loaded with `RUBY_GC_LIBRARY` the reported name of the GC library will also be displayed in `RUBY_DESCRIPTION`.

This PR also introduces a readonly `GC.config` key called `implementation` to expose this information to Ruby. It can be in three states:

1. Ruby compiled without shared GC support. `GC.config[:implementation] => "default"`
2. Ruby compiled with shared GC support but no GC module loaded. `GC.config[:implementation] => "default"`
3. Ruby compiled with shared GC support and an active GC module loaded using `RUBY_GC_LIBRARY`: `GC.config[:implementation] => "gc_name"`

The intention of this PR is to give users of the modular GC system, GC developers, and Ruby users better tooling to introspect the state of the currently CG. It will also make writing tests that work accurately with different GC configurations easier.

### DIFF
--- a/test/ruby/test_gc.rb
+++ b/test/ruby/test_gc.rb
@@ -127,9 +127,7 @@ class TestGc < Test::Unit::TestCase
   def test_gc_config_implementation_is_readonly
     omit unless /darwin|linux/.match(RUBY_PLATFORM)
 
-    impl = GC.config[:implementation]
-    GC.config(implementation: "somethingelse")
-    assert_equal(impl, GC.config[:implementation])
+    assert_raise(ArgumentError) { GC.config(implementation: "somethingelse") }
   end
 
   def test_start_full_mark
--- a/gc.rb
+++ b/gc.rb
@@ -306,7 +306,15 @@ module GC
   def self.config hash = nil
     return Primitive.gc_config_get unless hash
 
-    Primitive.gc_config_set hash
+    if(Primitive.cexpr!("RBOOL(RB_TYPE_P(hash, T_HASH))"))
+      if hash.include?(:implementation)
+        raise ArgumentError, 'Attempting to set read-only key "Implementation"'
+      end
+
+      Primitive.gc_config_set hash
+    else
+      raise ArgumentError
+    end
   end
 
   # call-seq:


gc.c
```diff
@@ -2761,6 +2767,23 @@ rb_gc_copy_attributes(VALUE dest, VALUE obj)
     rb_gc_impl_copy_attributes(rb_gc_get_objspace(), dest, obj);
 }
 
+const char *
+rb_gc_active_gc_name(void)
+{
+    const char *gc_name = rb_gc_impl_active_gc_name();
```
peterzhu2118, line 2773: Can we make `rb_gc_impl_active_gc_name` return a string literal or a malloc'd string? That way we don't need to allocate a new string every time.

gc/gc.h
```diff
@@ -88,6 +91,7 @@ hash_replace_ref_value(st_data_t *key, st_data_t *value, st_data_t argp, int exi
     return ST_CONTINUE;
 }
 
+RBIMPL_ATTR_MAYBE_UNUSED()
```
peterzhu2118, line 94: Shouldn't the `RBIMPL_WARNING_IGNORED(-Wunused-function)` above suppress warnings if this is unused?

test/ruby/test_gc.rb
```diff
@@ -117,6 +117,21 @@ def test_gc_config_disable_major_gc_start_always_works
...
+
+    gc_name = (ENV['RUBY_GC_LIBRARY'] || "default")
+    assert_equal gc_name, GC.config[:implementation]
+  end
+
+  def test_gc_config_implementation_is_readonly
+    omit unless /darwin|linux/.match(RUBY_PLATFORM)
+
+    impl = GC.config[:implementation]
+    GC.config(implementation: "somethingelse")
```
peterzhu2118, line 131: IMO we should raise if the user tries to change a read-only configuration at runtime.

version.c
```diff
@@ -176,6 +190,14 @@ define_ruby_description(const char *const jit_opt)
     RUBY_ASSERT(n <= ruby_description_opt_point + (int)rb_strlen_lit(YJIT_DESCRIPTION));
     if (ruby_mn_threads_enabled) append(" +MN");
     if (rb_ruby_prism_p()) append(" +PRISM");
+#if USE_SHARED_GC
+    append(GC_DESCRIPTION);
+    if (getenv("RUBY_GC_LIBRARY")) {
+        append("[");
+        append(rb_gc_active_gc_name());
```
peterzhu2118, line 197: Since `rb_gc_active_gc_name` returns a malloc'd buffer, I think this will leak memory.

version.c
```diff
@@ -176,6 +190,14 @@ define_ruby_description(const char *const jit_opt)
     RUBY_ASSERT(n <= ruby_description_opt_point + (int)rb_strlen_lit(YJIT_DESCRIPTION));
     if (ruby_mn_threads_enabled) append(" +MN");
     if (rb_ruby_prism_p()) append(" +PRISM");
+#if USE_SHARED_GC
+    append(GC_DESCRIPTION);
+    if (getenv("RUBY_GC_LIBRARY")) {
```
peterzhu2118, line 195: IMO we shouldn't depend on the environment variable. `gc.c` should define a function to query whether an external GC has been loaded.

common.mk
```diff
@@ -19555,6 +19555,7 @@ version.$(OBJEXT): $(CCAN_DIR)/str/str.h
 version.$(OBJEXT): $(hdrdir)/ruby.h
 version.$(OBJEXT): $(hdrdir)/ruby/ruby.h
 version.$(OBJEXT): $(hdrdir)/ruby/version.h
+version.$(OBJEXT): $(top_srcdir)/gc/gc.h
```
peterzhu2118, line 19558: I don't think `version.c` needs to include `gc/gc.h` anymore.

gc/gc.h
```diff
@@ -134,6 +134,7 @@ hash_replace_ref(st_data_t *key, st_data_t *value, st_data_t argp, int existing)
...
```
peterzhu2118, line 137: We don't need this change if we don't include `gc/gc.h` in`version.c`.

test/ruby/test_rubyoptions.rb
```diff
@@ -307,6 +307,16 @@ def test_rjit_version
     end
   end
 
+  def test_enabled_gc
+    omit unless /linux|darwin/ =~ RUBY_PLATFORM
+
+    if RbConfig::CONFIG['shared_gc_dir'].length > 0
+      assert_match(/\+GC/, RUBY_DESCRIPTION)
+    else
+      assert_match(/^((?!\+GC(\[\w+\])?).)*$/, RUBY_DESCRIPTION)
```
peterzhu2118, line 316: Can we just do this here?

```suggestion
      assert_no_match(/\+GC/, RUBY_DESCRIPTION)
```

version.c
```diff
@@ -167,6 +173,14 @@ define_ruby_description(const char *const jit_opt)
         + rb_strlen_lit(YJIT_DESCRIPTION)
         + rb_strlen_lit(" +MN")
         + rb_strlen_lit(" +PRISM")
+#if USE_SHARED_GC
+        + rb_strlen_lit(GC_DESCRIPTION)
+        // Assume the active GC name can not be longer than 20 chars
+        // so that we don't have to use strlen and remove the static
+        // qualifier from desc.
+        + RB_GC_MAX_NAME_LEN + 3
```
nobu, line 181: Why is 3, `[]` and what?

PR #11875, opened 2024-10-10
[DOC] Tweaks for Array#rotate!

BurdetteLamar:


array.c
```diff
@@ -3186,48 +3186,34 @@ rb_ary_rotate(VALUE ary, long cnt)
...
- *    a # => [2, :foo, "bar"]
+ *    [0, 1, 2, 3].rotate!(2)   # => [2, 3, 0, 1]
+      [0, 1, 2, 3].rotate!(2.1) # => [2, 3, 0, 1]
  *
  *  If +count+ is large, uses <tt>count % array.size</tt> as the count:
  *
- *    a = [:foo, 'bar', 2]
- *    a.rotate!(20)
- *    a # => [2, :foo, "bar"]
+ *    [0, 1, 2, 3].rotate!(20) # => [0, 1, 2, 3]
```
peterzhu2118, line 3201: 20 is a multiple of 4, so it ends up not demonstrating any rotation at all. Let's use something that rotates, like 19 or 21.

peterzhu2118 (approved):

PR #11876, opened 2024-10-10
[DOC] Tweaks for Array#sample

BurdetteLamar:


array.rb
```diff
@@ -73,35 +73,52 @@ def shuffle(random: Random)
   end
 
   # call-seq:
-  #    array.sample(random: Random) -> object
-  #    array.sample(n, random: Random) -> new_ary
+  #   sample(random: Random.new) -> object
+  #   sample(count, random: Random.new) -> new_ary
```
peterzhu2118, line 77: The default value is the `Random` class, not an instance of `Random`. The `Random` class implements the random methods as instance methods.

```suggestion
  #   sample(random: Random) -> object
  #   sample(count, random: Random) -> new_ary
```

peterzhu2118 (approved):

PR #11877, opened 2024-10-10
gems/bundled_gems - fix typeprof so it installs

MSP-Greg:
See https://github.com/ruby/ruby-dev-builder/actions/runs/11280388443/job/31373282203#step:19:357

Switch to most recent commit and current version


hsbt:
There is no release or pre-release version of `typeprof-0.30.0`. Please wait to stable release of typeprof.

PR #11880, opened 2024-10-11
Support `IO#timeout` for `rsock_connect`.

ioquatix:
Introduce support for general `IO#timeout` to `rsock_connect`:

```ruby
# Simple connect example:

require 'socket'

socket = Socket.new(:INET, :STREAM)
socket.timeout = 1
remote_addr = Socket.pack_sockaddr_in(80, '192.168.1.200') # Doesn't exist.
socket.connect(remote_addr) # raises IO::TimeoutError
```

See <https://github.com/socketry/async-http/issues/184> for more details/context. For reference, this builds on the previous changes: https://github.com/ruby/ruby/commits/master/?author=ioquatix&since=2024-10-04&until=2024-10-11

### DIFF
--- a/ext/socket/init.c
+++ b/ext/socket/init.c
@@ -473,10 +473,11 @@ rsock_socket(int domain, int type, int proto)
 
 /* emulate blocking connect behavior on EINTR or non-blocking socket */
 static int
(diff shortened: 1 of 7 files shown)


nobu:
Why isn't that example in tests?

PR #11881, opened 2024-10-11
Introduce inline_primary rule

S-H-GAMELINKS:
Introduce inline_primary rule to merge the same BNF pattern

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -4374,14 +4374,17 @@ mrhs		: args ',' arg_value
                     }
                 ;
 
-primary		: literal
-                | strings
-                | xstring
-                | regexp
-                | words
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -4374,14 +4374,17 @@ mrhs		: args ',' arg_value
...
 
-primary		: literal
-                | strings
-                | xstring
-                | regexp
-                | words
-                | qwords
-                | symbols
-                | qsymbols
+%rule %inline inline_primary : literal
```
nobu, line 4377: I don't think it's necessary to indent each rule that deeply.

```suggestion
%rule %inline inline_primary
                : literal
```

PR #11886, opened 2024-10-11
[DOC] Tweaks for Array#shift

BurdetteLamar:


array.c
```diff
@@ -1512,35 +1512,38 @@ rb_ary_shift(VALUE ary)
 
 /*
  *  call-seq:
- *     array.shift -> object or nil
- *     array.shift(n) -> new_array
+ *    shift -> new_array or nil
```
peterzhu2118, line 1515:
```suggestion
 *    shift -> object or nil
```

array.c
```diff
@@ -1512,35 +1512,38 @@ rb_ary_shift(VALUE ary)
...
+ *    shift(count) -> new_array or nil
  *
- *  Removes and returns leading elements.
+ *  Removes and returns leading elements from +self+ in a new array.
```
peterzhu2118, line 1518: It only returns a new array when `count` is given.

array.c
```diff
@@ -1512,35 +1512,38 @@ rb_ary_shift(VALUE ary)
...
+ *  If +count+ is large,
+ *  removes and returns all elements:
  *
- *    a = [:foo, 'bar', 2]
- *    a.shift(3) # => [:foo, 'bar', 2]
+ *    a = [0, 1, 2, 3]
+ *    a.shift(50) # => [0, 1, 2, 3]
+ *    a           # => []
  *
- *  If +n+ is zero, returns a new empty +Array+; +self+ is unmodified.
```
peterzhu2118, line 1541: The new documentation has lost the case where `count` is zero.

BurdetteLamar: Example added.

array.c

peterzhu2118 (approved):

PR #11898, opened 2024-10-15
Use GC.respond_to?(:compact) in bootstraptest/test_yjit.rb

peterzhu2118:
defined?(GC.compact) will always return true even when compaction is not supported. We should use GC.respond_to?(:compact) instead.


XrXr (approved):

XrXr:
I confirmed that defined? on a rb_f_notimplement method returns `"method"`.

PR #11911, opened 2024-10-17
[ruby/json] Sync changes

peterzhu2118:
Some changes were missed in the automatic sync.


byroot (approved):
I assumed it would happen eventually?

I'm not familiar with the sync thing, first time I work on a synced gem.

peterzhu2118:
> I assumed it would happen eventually?

I don't think it happens when it gets missed.

hsbt:
Thanks @peterzhu2118 

The sync is sometimes failure caused by git conflict or non-git files at `ruby/ruby` repository like `Gemfile`, `Rakefile` and others. I manually fixed them every 2-3 days.

PR #11913, opened 2024-10-17
YJIT: In stats, group by resolved C method name

XrXr:
Previously, in the "Top-N most frequent C calls"
section of --yjit-stats output, we printed the class
name of the receiver, not the method owner. This meant
that calls on subclass instances that land on the same
method showed up as different entires.

Similarly, method called using an alias showed up as
different entries from other aliases.

Group by the resolved method instead.

Test program:

    1.itself; [].itself; true.inspect; true.to_s

Before:

    Top-4 most frequent C calls (80.0% of C calls):
      1 (20.0%): Integer#itself
      1 (20.0%): TrueClass#to_s
      1 (20.0%): TrueClass#inspect
      1 (20.0%): Array#itself

After:

    Top-2 most frequent C calls (80.0% of C calls):
      2 (40.0%): Kernel#itself
      2 (40.0%): TrueClass#to_s

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6868,8 +6868,8 @@ fn gen_send_cfunc(
     // We also do this after the C call to minimize the impact of spill_temps() on asm.ccall().
     if get_option!(gen_stats) {
         // Assemble the method name string
-        let mid = unsafe { vm_ci_mid(ci) };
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
:+1:

maximecb (approved):
Thanks for the fix 🙏

PR #11917, opened 2024-10-18
YJIT: Allow shareable consts in multi-ractor mode

jhawthorn:
Constant entries have a flag to indicate whether or not the stored constant value is shareable across Ractors. When we assume the constant is stable we can assume the shareability is stable as well.

In the event that the constant isn't sharaeble, we can reuse the medium-speed path for crefs as that already calls `rb_vm_ic_hit_p` which checks both for cref changes and that we are on the main ractor.

I needed to update bindgen as the older version was crashing on my system (arch user btw).

``` ruby
def measure(text)
  before = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
  yield
  after = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_millisecond)
  puts "%s: %.2fms" % [text, after - before]
end

module Fib
  def self.fib(n)
    if n < 2
      return n
    end

    return Fib.fib(n-1) + Fib.fib(n-2)
  end
end

# warm
100.times { Fib.fib(4) }

measure("single threaded") do
  Fib.fib(32)
end

Ractor.new{}.take

measure("single thread after ractor spawn") do
  Fib.fib(32)
end

measure("4x in ractors") do
  4.times.map do
    Ractor.new do
      Fib.fib(32)
    end
  end.each(&:take)
end
```

**Before**
```
❯ ./miniruby --yjit test.rb
single threaded: 29.20ms
test.rb:25: warning: Ractor is experimental, and the behavior may change in future versions of Ruby! Also there are many implementation issues.
single thread after ractor spawn: 935.27ms
4x in ractors: 3824.14ms
```

**After**
```
❯ ./miniruby --yjit test.rb
single threaded: 28.91ms
test.rb:25: warning: Ractor is experimental, and the behavior may change in future versions of Ruby! Also there are many implementation issues.
single thread after ractor spawn: 28.91ms
4x in ractors: 32.90ms
```


vm_core.h
```diff
@@ -230,6 +230,8 @@ union ic_serial_entry {
     VALUE data[2];
 };
 
+#define IMEMO_CONST_CACHE_SHAREABLE IMEMO_FL_USER0
```
maximecb, line 233: Could you add a comment to explain what this flag means? (single constant is ractor-shareable?)

maximecb (commented):
Very cool. I would still like @XrXr to review it as he understands this code better than me, but it's nice to see better ractor support in YJIT 👌 

Also happy to see that this is a fairly simple PR.

yjit/src/codegen.rs
```diff
@@ -9915,11 +9915,16 @@ fn gen_opt_getconstant_path(
         return Some(EndBlock);
     }
 
-    if !unsafe { (*ice).ic_cref }.is_null() {
+    let is_cref = !unsafe { (*ice).ic_cref }.is_null();
+    let is_shareable = unsafe { rb_yjit_constcache_shareable(ice) };
+    let needs_checks = is_cref || (!is_shareable && !assume_single_ractor_mode(jit, asm));
```
XrXr, line 9920:
```suggestion
    let cref_sensitive = !unsafe { (*ice).ic_cref }.is_null();
    let is_shareable = unsafe { rb_yjit_constcache_shareable(ice) };
    let needs_checks = cref_sensitive || (!is_shareable && !assume_single_ractor_mode(jit, asm));
```

I think this is a slightly better name.

XrXr (approved):
Thread safety logic checks out to me. Thanks!

---
Rambling:

`IMEMO_CONST_CACHE_SHAREABLE` is essentially an assertion about the cached value being shareable (fine with concurrent access among multiple threads, in contrast to most objects which can be safely accessed in only one ractor). This bit of code shows it:

```C
static inline bool
vm_inlined_ic_hit_p(VALUE flags, VALUE value, const rb_cref_t *ic_cref, const VALUE *reg_ep)
{
    if ((flags & IMEMO_CONST_CACHE_SHAREABLE) || rb_ractor_main_p()) {
        VM_ASSERT(ractor_incidental_shareable_p(flags & IMEMO_CONST_CACHE_SHAREABLE, value));
```

PR #11919, opened 2024-10-18
YJIT: Rename method substitution functions and improve docs (+1)

XrXr:
... and "YJIT: Fill in commented-out assertion"

I always fail to recall this pair of functions I thought it'd be nice to give them a better name and make them link to each other in the rustdoc so Analyzer can help with jumping between them.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6347,8 +6347,9 @@ fn jit_thread_s_current(
     true
 }
 
-// Check if we know how to codegen for a particular cfunc method
-fn lookup_cfunc_codegen(def: *const rb_method_definition_t) -> Option<MethodGenFn> {
+/// Check if we know how to codegen for a particular cfunc method
+/// See also: [reg_method_codegen].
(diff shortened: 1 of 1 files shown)


yjit/src/codegen.rs
```diff
@@ -6347,8 +6347,9 @@ fn jit_thread_s_current(
     true
 }
 
-// Check if we know how to codegen for a particular cfunc method
-fn lookup_cfunc_codegen(def: *const rb_method_definition_t) -> Option<MethodGenFn> {
```
k0kubun, line 6351: What is wrong with calling it cfunc? It's hooked into only `gen_send_cfunc` at the moment, so we probably shouldn't imply it handles ISEQs as well.

XrXr: Yes, I agree lookup_cfunc_codegen() is a better name currently.

k0kubun (approved):

maximecb (approved):
Thanks Alan! :)

PR #11920, opened 2024-10-20
[DOC] Improve `Socket::tcp` with Happy Eyeballs Version 2

shioimm:
With the introduction of Happy Eyeballs Version 2 to `Socket::tcp`, the following areas have been mainly enhanced:

- How the value specified for `connect_timeout` works
- How Socket.tcp operates with Happy Eyeballs Version 2

A description for the new option `fast_fallback` has been added in https://github.com/ruby/ruby/pull/11813.

### DIFF
--- a/ext/socket/lib/socket.rb
+++ b/ext/socket/lib/socket.rb
@@ -629,15 +629,22 @@ class Socket < BasicSocket
   #
   # creates a new socket object connected to host:port using TCP/IP.
   #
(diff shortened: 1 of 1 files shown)


nobu:
Note that  the `[DOC]` tag in the PR title is only for that PR and does not work for other than it, i.e., master after merge and your branch; include `[DOC]` or `Document` in the commit log instead.

PR #11921, opened 2024-10-20
Make reserve_stack aware of ASAN fake stacks

KJTsanaktsidis:
Like other stack manipulation routines in thread_pthread.c, this also needs to be made aware of ASAN fake stacks when trying to use local variables to compute stack extents.

---

I'm actually not sure if we should do this at all - I'm fairly sure in 2024 we can just delete `reserve_stack` entirely. But i'm going to open an issue on bugs.ruby-lang.org about that.


KJTsanaktsidis:
We should consider https://bugs.ruby-lang.org/issues/20804 instead of merging this.

PR #11926, opened 2024-10-21
YJIT: Implement specialization for no-op `{Kernel,Numeric}#dup`

XrXr:
Type information in the context for no additional work!

This is the `if (special_object_p(obj)) return obj;` path in
rb_obj_dup() and for Numeric#dup, it's always the identity function.

---

I did some quick benching with this and no visible improvements on optcarrot and lobsters. But we do see our production app using `dup` a lot, so this still seems useful, especially since the implementation is trivial.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6347,6 +6347,42 @@ fn jit_thread_s_current(
     true
 }
 
+fn jit_numeric_dup(
+    _jit: &mut JITState,
+    _asm: &mut Assembler,
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: Option<VALUE>,
+) -> bool {
+    // Numeric#dup has arity=0 and is the identity function.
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -6347,6 +6347,42 @@ fn jit_thread_s_current(
...
+    _ci: *const rb_callinfo,
+    _cme: *const rb_callable_method_entry_t,
+    _block: Option<BlockHandler>,
+    _argc: i32,
+    _known_recv_class: Option<VALUE>,
+) -> bool {
+    // Numeric#dup has arity=0 and is the identity function.
+    // Our caller already did argument count check, so this is
+    // no-op to return the receiver that is already on the stack.
+    true
```
maximecb, line 6362: Could we propagate the type here if known/fixnum?

maximecb (approved):
Thanks Alan :)

yjit/src/codegen.rs
XrXr: Because of the guard on the receiver type info is already in the context and we don't need to do anything to propagate it here.

PR #11928, opened 2024-10-22
gcc 14 on macOS

nobu:


XrXr:
Seems like a GCC bug, no? Might be worth reporting.

nobu:
As for `-pipe`, https://github.com/Homebrew/homebrew-core/issues/195299

nobu:
I can't make a short code to reproduce `-Warray-bounds` warning.

nobu:
At least, the warning seems to need:
1. gcc-14 on aarch64 macOS
2. `-O3`
3. successive `RB_TYPE_P()`

PR #11930, opened 2024-10-22
[DOC] Tweaks for Array#take_while

BurdetteLamar:


array.c
```diff
@@ -7536,25 +7536,21 @@ rb_ary_take(VALUE obj, VALUE n)
...
- *  Returns a new +Array+ containing zero or more leading elements of +self+;
- *  does not modify +self+.
+ *    take_while {|element| ... } -> new_array
+ *    take_while -> new_enumerator
  *
  *  With a block given, calls the block with each successive element of +self+;
- *  stops if the block returns +false+ or +nil+;
- *  returns a new +Array+ containing those elements for which the block returned a truthy value:
+ *  stops iterating if the block returns +false+ or +nil+;
+ *  returns a new array containing those elements for which the block returned a truthy value:
```
peterzhu2118, line 7544: Like `Array#take`, we should specify that the original array is not modified.

BurdetteLamar: Added.

peterzhu2118 (approved):

PR #11943, opened 2024-10-23
Travis reenable

nobu:


junaruga:
Thank you for testing Travis! A few days ago I was informed by Travis customer support that they fixed the long-term infra issue on the non-x86_64 pipelines, I was testing Travis to reenable it on my forked repository.

junaruga:
In this PR, and in my testing, the Travis arm32-linux case always fails. I think we can drop the case at least for now.
https://app.travis-ci.com/github/ruby/ruby/jobs/627413572#L691

junaruga:
I would close this PR, as the issue was resolved by https://github.com/ruby/ruby/pull/11948.

PR #11945, opened 2024-10-23
Fix shared GC with -DRUBY_DEBUG

eightbitraptor:
RUBY_DEBUG enables ractor assertions, which sets up some space at the end of each RVALUE to store the associated ractor ID. We need to make sure the function that does this is visible to shared GC libraries.

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -12,6 +12,7 @@
 #include "internal/error.h"
 #include "internal/gc.h"
 #include "internal/hash.h"
+#include "internal/ractor.h"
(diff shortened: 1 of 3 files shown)


ractor.c
```diff
@@ -12,6 +12,7 @@
 #include "internal/error.h"
 #include "internal/gc.h"
 #include "internal/hash.h"
+#include "internal/ractor.h"
```
nobu, line 15: :rofl:

PR #11947, opened 2024-10-24
Make StringScanner compactable, WB-protected and embedded

wks:
Modify strscan_mark and implement strscan_compact so that the reference fields of StringScanner, i.e. "str" and "regex", can be updated in a moving GC.

Also apply write barriers when updating the two fields.  That makes StringScanner WB-protected.

Also make StringScanner an embedded RTypedData to reduce one unnecessary malloc and free.

The motivation of this PR is that some workloads, especially the Liquid framework, use StringScanner intensively, and StringScanner becomes the majority of WB-unprotected objects.  This change will improve the performance by reducing the number of WB-unprotected objects and the number of objects unnecessarily pinned.


kou:
The StringScanner's repository is https://github.com/ruby/strscan . Could you use it instead of here?

PR #11951, opened 2024-10-25
Convert spaces to tabs in yjit.mk

aidenfoxivey:
I noticed that there were spaces put in lieu of tabs here. Since the rest of the file uses tabs, I replaced said spaces with tabs.


aidenfoxivey:
It's rather unfortunate, but it seems that the Github UI doesn't distinguish between whitespace characters.

nobu:
Commands must be indented with tabs, but continued line indentations don’t need to be tabs.
It works even with no indents.

aidenfoxivey:
That makes sense. I just noticed that it didn’t fit stylistically - I’m absolutely fine with closing the PR if it’s not deemed a useful change.

I just figured that consistency might be good there - not that it changes functionality.

XrXr:
Thanks, but the inconsistency itself is a stylistic choice so I'd like to keep it the way it is. If you have tabs render differently from spaces it shows the continuation visually. Also, this is not the only place we go for this style.

Screenshot 2024-10-25 at 10 08 22

aidenfoxivey:
Ah okay, interesting. Sorry to add to the noise then!

I hope you folks have a good rest of your day. :)

PR #11955, opened 2024-10-26
YJIT: Replace Array#each only when YJIT is enabled

k0kubun:
This PR resurrects the C implementation of `Array#each` for the interpreter and lets YJIT use the Ruby one. To minimize incompatibilities between the interpreter and YJIT, it also introduces `Primitive.attr! :c_trace` to let `Array#each` act like a C method in backtraces.

It speeds up the interpreter on lobsters a little. While the rewrite of `Array#each` in Ruby https://github.com/ruby/ruby/pull/9533 did not slow down `Array#each` on the interpreter with microbenchmarks, it had a small impact on larger benchmarks.

```
before: ruby 3.4.0dev (2024-10-26T13:20:34Z master 484ea00d2e) +PRISM [x86_64-linux]
after: ruby 3.4.0dev (2024-10-28T10:43:42Z yjit-ruby-mode 192faf1882) +PRISM [x86_64-linux]

lobsters  843.4        1.1         838.8       0.6         0.95           1.01
```

### DIFF
--- a/array.rb
+++ b/array.rb
@@ -215,7 +215,7 @@ class Array
   end
 
   with_yjit do
-    if Array.instance_method(:each).source_location == nil # preserve monkey patches
+    if Primitive.rb_builtin_basic_definition_p(:each)
       undef :each
 
       def each # :nodoc:
--- a/array.rb
+++ b/array.rb
@@ -1,50 +1,5 @@
 class Array
   # call-seq:
-  #   each {|element| ... } -> self
-  #   each -> new_enumerator
-  #
-  # With a block given, iterates over the elements of +self+,
-  # passing each element to the block;
-  # returns +self+:
-  #
-  #   a = [:foo, 'bar', 2]
-  #   a.each {|element|  puts "#{element.class} #{element}" }
-  #
-  # Output:
-  #
-  #   Symbol foo
-  #   String bar
-  #   Integer 2
-  #
-  # Allows the array to be modified during iteration:
-  #
-  #   a = [:foo, 'bar', 2]
-  #   a.each {|element| puts element; a.clear if element.to_s.start_with?('b') }
-  #
-  # Output:
-  #
-  #   foo
-  #   bar
-  #
-  # With no block given, returns a new Enumerator.
-  #
-  # Related: see {Methods for Iterating}[rdoc-ref:Array@Methods+for+Iterating].
-
-  def each
-    Primitive.attr! :inline_block
-
-    unless defined?(yield)
-      return Primitive.cexpr! 'SIZED_ENUMERATOR(self, 0, 0, ary_enum_length)'
-    end
-    _i = 0
-    value = nil
-    while Primitive.cexpr!(%q{ ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value)) })
-      yield value
-    end
-    self
-  end
-
-  # call-seq:
   #   shuffle!(random: Random) -> self
   #
   # Shuffles all elements in +self+ into a random order,
@@ -257,5 +212,25 @@ class Array
   def fetch_values(*indexes, &block)
     indexes.map! { |i| fetch(i, &block) }
     indexes
+  end
+
+  with_yjit do
+    if Array.instance_method(:each).source_location == nil # preserve monkey patches
+      undef :each
+
+      def each # :nodoc:
+        Primitive.attr! :inline_block, :c_trace
+
+        unless defined?(yield)
+          return Primitive.cexpr! 'SIZED_ENUMERATOR(self, 0, 0, ary_enum_length)'
+        end
+        _i = 0
+        value = nil
+        while Primitive.cexpr!(%q{ ary_fetch_next(self, LOCAL_PTR(_i), LOCAL_PTR(value)) })
+          yield value
+        end
+        self
+      end
+    end
   end
 end
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -3387,6 +3387,9 @@ pm_compile_builtin_attr(rb_iseq_t *iseq, const pm_scope_node_t *scope_node, cons
         else if (strcmp(RSTRING_PTR(string), "use_block") == 0) {
             iseq_set_use_block(iseq);
         }
+        else if (strcmp(RSTRING_PTR(string), "c_trace") == 0) {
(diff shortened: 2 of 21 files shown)


prism_compile.c
```diff
@@ -3387,6 +3387,9 @@ pm_compile_builtin_attr(rb_iseq_t *iseq, const pm_scope_node_t *scope_node, cons
         else if (strcmp(RSTRING_PTR(string), "use_block") == 0) {
             iseq_set_use_block(iseq);
         }
+        else if (strcmp(RSTRING_PTR(string), "c_trace") == 0) {
+            ISEQ_BODY(iseq)->builtin_attrs |= BUILTIN_ATTR_C_TRACE;
+        }
```
maximecb, line 3392: What does this do? Can you add a short comment?

array.rb
```diff
@@ -258,4 +213,24 @@ def fetch_values(*indexes, &block)
     indexes.map! { |i| fetch(i, &block) }
     indexes
   end
+
+  with_yjit do
+    if Array.instance_method(:each).source_location == nil # preserve monkey patches
```
XrXr, line 218: This technically isn't comprehensive enough of a check since the money patch could be a C method:

```
$ ruby -e 'its = Kernel.instance_method(:itself); \
      Array.define_method(:each, its);\
      p Array.instance_method(:each).source_location'
nil
```
Maybe check with rb_method_basic_definition_p()?

XrXr (commented):
We should make sure that once the Ruby implementation is subbed in, it's still a basic method, because we have code like the following for example in Enumerable#inject:

```
    if (iter == inject_op_i &&
        SYMBOL_P(op) &&
        RB_TYPE_P(obj, T_ARRAY) &&
        rb_method_basic_definition_p(CLASS_OF(obj), id_each)) {
        return ary_inject_op(obj, init, op);
    }
```

maximecb (commented):
Given that we are likely to port more runtime methods to Ruby (bring back Aaron's ports and add some more?), and that this may increase generated code size, I think it would be good to have a command line flag (or an environment variable?) so we can turn this on or off. This would allow us to test the impact on benchmarks and in production at various stages.

An environment variable might be easier for production?

prism_compile.c
k0kubun: I thought documenting it above the `BUILTIN_C_TRACE` definition was enough, but I copied it here for you. https://github.com/ruby/ruby/pull/11955/commits/a378fb0727781a6130aed5a3cadc856c66e578f4

vm_core.h
```diff
@@ -372,6 +372,8 @@ enum rb_builtin_attr {
     BUILTIN_ATTR_SINGLE_NOARG_LEAF = 0x02,
     // This attribute signals JIT to duplicate the iseq for each block iseq so that its `yield` will be monomorphic.
     BUILTIN_ATTR_INLINE_BLOCK = 0x04,
+    // The iseq acts like a C method in backtraces.
+    BUILTIN_ATTR_C_TRACE = 0x08,
```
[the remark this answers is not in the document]
k0kubun: Yes, we do.

For some reason, people like to test backtraces. In fact, when Ruby 3.4 introduced an incompatibility in the format of backtraces, quite a few tests failed on Core and SFR.

If we leave the backtrace behavior, people will get test failures when they enable YJIT. To maximize the adoption of YJIT, I don't want to make YJIT migration harder.

Does that make sense?

yjit.rb
```diff
@@ -247,10 +248,24 @@ def self.simulate_oom! # :nodoc:
...
   end
 
+  # Blocks that are called when YJIT is enabled
+  @yjit_hooks = []
+
   class << self
     # :stopdoc:
     private
 
+    # Lazily call a given block when YJIT is enabled
```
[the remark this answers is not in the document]
k0kubun: Updated the comment https://github.com/ruby/ruby/pull/11955/commits/0fdd26b8ae50a969d2cb621ec9eacaa07ac75d1e

array.rb
k0kubun: Fixed in https://github.com/ruby/ruby/pull/11955/commits/218776d8aa1a651151c88b449036ef0d23b12387

k0kubun:
I was so confused by the `test-all` failure that it took a while to fix it, but I finally fixed the `test-all` failure https://github.com/ruby/ruby/pull/11955/commits/fd9a3064de1a706cc6553f36354e7ec2b7f9f4f2.

> We should make sure that once the Ruby implementation is subbed in, it's still a basic method, because we have code like the following for example in Enumerable#inject:

Fixed in https://github.com/ruby/ruby/pull/11955/commits/d06ac2b85cbd44d574ebf5e4e048e70bb085085a

> I think it would be good to have a command line flag (or an environment variable?)

Added `--yjit-c-builtin` https://github.com/ruby/ruby/pull/11955/commits/fdade1e0bfca8184527d67616f0bddae2e725e48. A command line flag would be the most useful thing on yjit-bench, but a build flag is more useful for comparison on SFR. So I allowed `YJIT_C_BUILTIN` macro to switch it too.

yjit.rb
```diff
@@ -247,10 +248,30 @@ def self.simulate_oom! # :nodoc:
...
+      @yjit_hooks << hook
+    end
+
+    # Run YJIT hooks registered by RubyVM::YJIT.with_yjit
+    def call_yjit_hooks
+      # Skip using builtin methods in Ruby if --yjit-c-builtin is given
+      return if Primitive.yjit_c_builtin_p
+
+      # Unset GET_VM()->running to give METHOD_ENTRY_BASIC flag to methods defined by the hooks
+      Primitive.cexpr! 'GET_VM()->running = 0'
```
XrXr, line 269: This seems no thread safe. It's probably fine without ractors, but theoretically we could get a timed thread switch in the middle of this region with plain `Thread`s too, then this would at least allow potentially disguising methods as basic. I don't know what other special privilege running the this flag cleared grants.

k0kubun: How about this? https://github.com/ruby/ruby/pull/11955/commits/d066d0173ae0cd268c33c1a7355aa0e08700351d

XrXr: Thanks! Sorry for bugging you about it.

maximecb (approved):

maximecb:
Good work Kokubun! 👏

PR #11956, opened 2024-10-28
Allow passing a block to Hash#store (to update current value)

furunkel:
This is a proposal to allow passing a block to `Hash#store` to obtain the value to be stored.
The block is called with the current value or, if unset, the hash's default value; the block's return value will be the value that is stored. This is similar to e.g., `computeIfPresent`/`computeIfAbsent` in Java (I think).

I can think of several situations where this is useful, in particular for caches and counters.
For instance:
```ruby
counts = {}
elements.each do |element|
  counts.store(element){ (_1 || 0) + element.weight}
end
```

or even more elegant with a default value:
```ruby
counts = {}
counts.default = 0
elements.each do |element|
  counts.store(element){ _1 + element.weight}
end
```

Moreover, using the block form we should be able to do operations such as  `h[k] ||= x`, `h[k] -= x`, `h[k] += x`, or more generally `h[k] = f(h[k])`, with a single "hashing round-trip". If I'm not mistaken, currently these involve two separate calls to `#[]` and `#[]=` (with two calls to `#hash`?).

Finally, this makes `#store` a proper dual of `#fetch` which, similarly, can be passed a block.

I'm not entirely sure whether it is safe to call `rb_yield` from inside the hash update callback (e.g., because the block may raise).
At any rate, this change will likely need further discussion.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -2897,10 +2897,39 @@ hash_aset_str(st_data_t *key, st_data_t *val, struct update_arg *arg, int existi
 NOINSERT_UPDATE_CALLBACK(hash_aset)
 NOINSERT_UPDATE_CALLBACK(hash_aset_str)
 
+static int
+hash_store_yield(st_data_t *key, st_data_t *val, struct update_arg *arg, int existing)
+{
+    VALUE current_value;
+    if(existing) {
+        current_value = *val;
+    } else {
+        if (FL_TEST(arg->hash, RHASH_PROC_DEFAULT)) {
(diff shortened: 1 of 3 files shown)


nobu:
Please file a [new feature](https://bugs.ruby-lang.org/projects/ruby-master/issues/new).

furunkel:
@nobu, thanks for your reply, corresponding feature at https://bugs.ruby-lang.org/issues/20818.

PR #11963, opened 2024-10-30
Introduce Fiber Scheduler `blocking_region` hook.

ioquatix:
This allows us to hoist nogvl code out of the event loop to prevent blocking.

https://bugs.ruby-lang.org/issues/20855

### DIFF
--- a/scheduler.c
+++ b/scheduler.c
@@ -13,6 +13,9 @@
 #include "ruby/io.h"
 #include "ruby/io/buffer.h"
 
+#include "ruby/thread.h"
+
+// For `ruby_thread_has_gvl_p`.
 #include "internal/thread.h"
 
 static ID id_close;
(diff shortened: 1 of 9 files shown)


scheduler.c
```diff
@@ -693,6 +700,49 @@ rb_fiber_scheduler_address_resolve(VALUE scheduler, VALUE hostname)
...
+    void *(*function)(void *);
+    void *data;
+    rb_unblock_function_t *unblock_function;
+    void *data2;
+    int flags;
+
+    struct rb_fiber_scheduler_blocking_region_state *state;
+};
+
+VALUE rb_fiber_scheduler_blocking_region_proc(RB_BLOCK_CALL_FUNC_ARGLIST(value, _arguments))
```
nobu, line 713: Why is this global?

```suggestion
static VALUE
rb_fiber_scheduler_blocking_region_proc(RB_BLOCK_CALL_FUNC_ARGLIST(value, _arguments))
```

PR #11966, opened 2024-10-30
[Bug #20853] Fix Proc#hash to not change after compaction

peterzhu2118:
The hash value of a Proc must remain constant after a compaction, otherwise it may not work as the key in a hash table.

### DIFF
--- a/proc.c
+++ b/proc.c
@@ -15,6 +15,7 @@
 #include "internal/error.h"
 #include "internal/eval.h"
 #include "internal/gc.h"
+#include "internal/hash.h"
 #include "internal/object.h"
 #include "internal/proc.h"
 #include "internal/symbol.h"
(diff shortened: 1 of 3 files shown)


proc.c
```diff
@@ -1437,8 +1438,24 @@ rb_hash_proc(st_index_t hash, VALUE prc)
 {
     rb_proc_t *proc;
     GetProcPtr(prc, proc);
-    hash = rb_hash_uint(hash, (st_index_t)proc->block.as.captured.code.val);
-    hash = rb_hash_uint(hash, (st_index_t)proc->block.as.captured.self);
+
+    switch (vm_block_type(&proc->block)) {
```
casperisfine, line 1442: Any reason to have a `rb_hash_proc ` instead of just falling back to `rb_obj_hash`? I guess it saves generating an object id?

casperisfine: TIL, carry on then :)

peterzhu2118:
The assertion errors are a separate problem and should be fixed in #11970.

apotonick:
Hi @peterzhu2118 thanks for taking care of this issue :green_heart: as this is also a problem in 3.1.6, are there any plans for fixes in the 3.1 line? Thanks again :beers:

apotonick:
And one more question: this bug persists if you use `method(:a)` as a hash key, and then call GC.

```ruby
def a; end
hsh = {method(:a) => 1}

hsh[method(:a)]
 => 1 

GC.verify_compaction_references(expand_heap: true, toward: :empty)

hsh[method(:a)]
 => nil

peterzhu2118:
Thank you for reporting this bug. I opened a [ticket](https://bugs.ruby-lang.org/issues/20868) and have a fix in #12004.

apotonick:
Wrote a short blog post about this https://dev.to/trailblazer/gc-compaction-behold-your-hash-keys-2ii1

PR #11965, opened 2024-10-30
use `instance_method` to get method object

ko1:
instead of `method()`.

There is a bug around `define_method`, so this patch is workaround.

PR #11968, opened 2024-10-31
`alias` should care complemented flag

ko1:
`me->defined_class` should be 0 for method entries of Modules. The current implementation violates this condition when the aliased object was complemented cme.

This patch checks this condition
and fix https://github.com/ruby/ruby/pull/11965#issuecomment-2448291790


ko1:
complemented flag is not good condition. See the latest commit.

PR #11969, opened 2024-10-31
Make gemspec files for default gems with extensions

nobu:
Make gemspec files for default gems with extensions, so that rubygems can find them as gems.
However, the `--install-dir` option of `gem install` seems to exclude prerelease gems, even already installed in that directory, from the dependencies for some reasons; use the `GEM_HOME` environment variable instead.

Now net-imap 0.5.0 depends on json gem.


hsbt:
👍

PR #11975, opened 2024-11-01
Introduce `RUBY_ASSERT(ruby_thread_has_gvl_p())` to detect invalid interface usage.

ioquatix:
https://bugs.ruby-lang.org/issues/20877

This was originally to demonstrate/show the problems with zlib. That has now been fixed: https://github.com/ruby/zlib/pull/88

This change helps us find other (similar) issues and prevent regressions.

### DIFF
--- a/string.c
+++ b/string.c
@@ -42,6 +42,7 @@
 #include "probes.h"
 #include "ruby/encoding.h"
 #include "ruby/re.h"
+#include "ruby/thread.h"
 #include "ruby/util.h"
 #include "ruby_assert.h"
 #include "vm_sync.h"
@@ -2586,6 +2587,8 @@ rb_check_lockedtmp(VALUE str)
 static inline void
 str_modifiable(VALUE str)
 {
+    RUBY_ASSERT(ruby_thread_has_gvl_p());
+
     if (RB_UNLIKELY(FL_ANY_RAW(str, STR_UNMODIFIABLE_MASK))) {
         if (CHILLED_STRING_P(str)) {
             CHILLED_STRING_MUTATED(str);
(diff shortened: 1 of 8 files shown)


jeremyevans (changes requested):
The zlib changes should be sent upstream to ruby/zlib.  string.c changes look fine.

jeremyevans:
Never mind, looks like you did submit the changes ro ruby/zlib already: https://github.com/ruby/zlib/pull/88

jeremyevans (approved):
I recommend updating the commit message to describe that this is being done to debug issues in the zlib extension.  Code changes look fine.

ioquatix:
The reason why this PR is failing is because `zlib.c` is using the C functions without holding the GVL. Once https://github.com/ruby/zlib/pull/88 is merged, this PR will become green.

string.c
```diff
@@ -41,6 +41,7 @@
 #include "probes.h"
 #include "ruby/encoding.h"
 #include "ruby/re.h"
+#include "ruby/thread.h"
```
ko1, line 44: why does it needed?

ko1:
Do you plan to introduce assertions to all of `rb_` prefix C API?

string.c
ioquatix: For `ruby_thread_has_gvl_p`.

ioquatix:
> Do you plan to introduce assertions to all of rb_ prefix C API?

I feel like we should, but it's also quite a lot of work. So for now, my main goal is to find any bugs. Therefore, adding it to methods like `rb_xxx_modify` and so on may be sufficient.

If you have a different idea, please let me know.

string.c
ioquatix: Ah yes, I'll fix it.

ioquatix: Okay, it is fixed.

samuel-williams-shopify:
If you see an error like this:

```
string.c:3660: Assertion Failed: rb_str_set_len:ruby_thread_has_gvl_p()
```

It might mean you are (1) using a build of Ruby that includes this PR, but (2) haven't updated zlib gem to the latest version which does not trigger this assertion.

PR #11981, opened 2024-11-01
Add an environment variable for controlling the default Thread quantum

tenderlove:
This commit adds an environment variable `RUBY_THREAD_TIMESLICE` for
specifying the default thread quantum in milliseconds.  You can adjust
this variable to tune throughput, which is especially useful on
multithreaded systems that are mixing CPU bound work and IO bound work.

The default quantum remains 100ms.

[Feature #20861]

### DIFF
--- a/man/ruby.1
+++ b/man/ruby.1
@@ -537,6 +537,10 @@ Introduced in Ruby 3.3, default: unset.
 .Pp
 .It Ev RUBY_PAGER
 The pager command that will be used for
+.Pp
+.It Ev RUBY_THREAD_TIMESLICE
+Sets the default thread time slice (thread quantum) in milliseconds.
+Introduced in Ruby 3.4, default: 100ms.
 .Fl -help
 option.
 Introduced in Ruby 3.0, default:
--- a/thread.c
+++ b/thread.c
@@ -116,6 +116,8 @@ static VALUE sym_immediate;
 static VALUE sym_on_blocking;
 static VALUE sym_never;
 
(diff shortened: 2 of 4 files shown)


nobu:
`TIME_QUANTUM_USEC` is also in thread_win32.c.

thread_win32.c
```diff
@@ -768,7 +767,7 @@ timer_thread_func(void *dummy)
     RUBY_DEBUG_LOG("start");
     rb_w32_set_thread_description(GetCurrentThread(), L"ruby-timer-thread");
     while (WaitForSingleObject(timer_thread.lock,
-                               TIME_QUANTUM_USEC/1000) == WAIT_TIMEOUT) {
+                               thread_default_quantum_ms) == WAIT_TIMEOUT) {
```
jhawthorn, line 770: I'm not 100% sure this is the same as the other purposes we've used the ENV var. This looks a bit more like the "minimum quantum" the pthread code uses on the poll (it is also 10ms).

man/ruby.1
```diff
@@ -537,6 +537,10 @@ Introduced in Ruby 3.3, default: unset.
 .Pp
 .It Ev RUBY_PAGER
 The pager command that will be used for
+.Pp
+.It Ev RUBY_THREAD_DEFAULT_QUANTUM_MS
+Sets the default thread quantum in milliseconds.
+Introduced in Ruby 3.4, default: 100ms.
```
nateberkopec, line 543: 10ms default on Windows

jhawthorn (approved):

PR #11990, opened 2024-11-04
Precompute hash when there is enough capacity

etiennebarrie:
Follow-up to #11989 and #10596 

Context: While working on ruby/json perf, we noticed significant amount of time spent hashing the same strings over and over again.

This extends the heuristic to store the precomputed hash for interned strings if the capacity allows without bumping to the next slot size.

### DIFF
--- a/string.c
+++ b/string.c
@@ -350,9 +350,14 @@ mustnot_wchar(VALUE str)
 
 static int fstring_cmp(VALUE a, VALUE b);
 
-static VALUE register_fstring(VALUE str, bool copy, bool precompute_hash);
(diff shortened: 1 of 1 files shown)


casperisfine:
Profiling `ruby/json` with this comparing the `twitter.json` payload:

Before: https://share.firefox.dev/4fbcbVc

Capture d’écran 2024-11-04 à 14 57 55

After: https://share.firefox.dev/3Aso5L4

Capture d’écran 2024-11-04 à 14 58 38

The time spent in `ruby_sip_hash13` is clearly cut in half.

PR #11991, opened 2024-11-04
Mingw: Exclude failing tests due to the crt change

nobu:
Mingw crt-git 12.0.0.r369.g0d4221712-1 now prohibits "command line contains characters that are not supported in the active code page".
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/0d42217123d3aec0341b79f6d959c76e09648a1e/

Provisionally exclude tests that fail by passing such characters.

### DIFF
--- a/test/.excludes/TestArgf.rb
+++ b/test/.excludes/TestArgf.rb
@@ -1,0 +1,9 @@
+if RUBY_PLATFORM.include?("mingw")
+  reason = <<~EOS
+  Mingw crt-git 12.0.0.r369.g0d4221712-1 now prohibits "command line
+  contains characters that are not supported in the active code page".
+  https://sourceforge.net/p/mingw-w64/mingw-w64/ci/0d42217123d3aec0341b79f6d959c76e09648a1e/
+  EOS
+
+  exclude(:test_inplace_nonascii, reason)
+end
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
This also broke [one Bundler spec](https://github.com/rubygems/rubygems/pull/8221), which I'll be skipping similarly.

Is this related to https://github.com/ruby/ruby/pull/11799 in any way, @larskanis?

deivid-rodriguez:
I don't understand what's different between Ruby 3.3.6 and Ruby 3.3.5 that makes the former use the broken version, though.

larskanis:
Unfortunately this slipped through the release tests on RubyInstaller-3.3.6-1 since they are running on codepage 65001 (UTF-8) on Github so that this issue wasn't triggered. I'm working on a solution now, probably by using `wmain()` as the entry point.

> Is this related to #11799 in any way, @larskanis?

Yes, somewhat - command line handling is inconsistent without #11799.

larskanis:
I tried to repro this issue on ruby master to no avail, until I noticed that nobu already fixed it in https://github.com/ruby/ruby/commit/74fcf437676f0019c55368753252fbe4afdf52ac . I patched RubyInstaller and released 3.3.6-2 fixing this regression.

PR #11995, opened 2024-11-05
Make rb_objspace_garbage_object_p()=true on non-heap pointers

XrXr:
In rare situations, by the time vm_ccs_free() runs, the stale cc pointer
might point to a heap page that has been freed a long time ago and is no
longer part of the active GC heap anymore. Previously we'd try to read
and write through the stale pointer anyways, which caused use-after-free
crashes. Surely other callers are at least fine with this change, if not
also avoid bugs in similar situations.

    if (!rb_objspace_garbage_object_p((VALUE)cc) &&
        IMEMO_TYPE_P(cc, imemo_callcache) && // <-- Previously UAF
        cc->klass == klass) {

Not changing rb_gc_impl_garbage_object_p(), since the GC always work
within the confines of its heap already, so no need to check there.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1595,7 +1595,11 @@ rb_objspace_free_objects(void *objspace)
 int
 rb_objspace_garbage_object_p(VALUE obj)
 {
-    return rb_gc_impl_garbage_object_p(rb_gc_get_objspace(), obj);
+    void *objspace = rb_gc_get_objspace();
+    if (!rb_gc_impl_pointer_to_heap_p(objspace, (void *)obj)) {
+        return TRUE;
+    }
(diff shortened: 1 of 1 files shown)


peterzhu2118 (commented):
If we're doing this, we might as well combine `rb_gc_impl_pointer_to_heap_p` with this since T_NONE/T_MOVED/T_ZOMBIE objects are internal to the GC anyways so they shouldn't be considered live objects.

Also on another note, `rb_objspace_garbage_object_p` is used by method caches, frozen strings, and dynamic symbols. This will significantly impact the performance of those things. Maybe we want to transition those things to native weak references so that this performance impact is mitigated.

PR #12001, opened 2024-11-05
Use rb_fiber_scheduler_get() instead of rb_fiber_scheduler_current()

mame:
`rb_fiber_scheduler_current()` may return `nil` depending on whether the scheduler is being prevented for some reason, e.g., `Fiber.blocking{}`.

Using `rb_fiber_scheduler_get()` returns the scheduler, if one was set, even in a blocking context.


ioquatix (approved):
Thanks, LGTM!

PR #12027, opened 2024-11-07
YJIT: Pass panic message to rb_bug()

XrXr:
So that the Rust panic message is forwarded to the RUBY_CRASH_REPORT
system, instead of only the static "YJIT panicked" message done so
previously. This helps with triaging crashes since it's easier than
trying to parse stderr output.

Sample:

    <internal:yjit_hook>:2: [BUG] YJIT: panicked at src/codegen.rs:1197:5:
    explicit panic
    ...


maximecb (approved):
Small QoL improvements ftw 👌

PR #12035, opened 2024-11-08
YJIT: Always abandon the block when gen_branch() or defer_compilation…

XrXr:
…() fails

In [1], we started checking for gen_branch failures, but I made two crucial mistakes. One, defer_compilation(), which has the same issue as gen_branch(), still weren't checked. Two, returning None from a codegen function does not throw away the block. Checking how gen_single_block() handles codegen functions, you can see that None terminates the block with an exit, but does not overall return an Err. This handling is fine for unimplemented instructions, for example, but incorrect in case gen_branch() fails, since the missing branch essentially corrupts the block; adding more code after a missing branch doesn't correct the code.

Always abandon the block when defer_compilation() or gen_branch() fails.

[1]: https://github.com/ruby/ruby/commit/cb661d7d82984cdb54485ea3f4af01ac21960882
Fixup: [1]

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -195,6 +195,34 @@ impl<'a> JITState<'a> {
         self.outlined_code_block
     }
 
+    /// Leave a code stub to re-enter the compiler at runtime when the compiling program point is
+    /// reached. Should always be used in tail position like `return jit.defer_compilation(asm);`.
+    #[must_use]
+    fn defer_compilation(&mut self, asm: &mut Assembler) -> Option<CodegenStatus> {
+        if crate::core::defer_compilation(self, asm).is_err() {
+            // If we can't leave a stub, the block isn't usable and we have to bail.
+            self.block_abandoned = true;
+        }
+        Some(EndBlock)
+    }
+
+    /// Generate a branch with either end possibly stubbed out
+    fn gen_branch(
+        &mut self,
+        asm: &mut Assembler,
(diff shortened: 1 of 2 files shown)


yjit/src/codegen.rs
```diff
@@ -4677,15 +4677,18 @@ fn gen_branchnil(
...
             asm,
             jump_block,
             &ctx,
             Some(next_block),
             Some(&ctx),
             BranchGenFn::BranchNil(Cell::new(BranchShape::Default)),
-        )?;
+        ).is_none() {
+            // Block unusable without the branch.
+            jit.block_abandoned = true;
```
k0kubun, line 4690: It seems like we always set this flag when `gen_branch()` returns `None`. Similarly to `jit.defer_compilation()`, could we introduce `jit.gen_branch()` that sets `jit.block_abandoned = true;` on failure (or let `gen_branch()` do that)?

k0kubun (approved):

maximecb (approved):

maximecb:
> By the way, I was essentially taken in by the convenience of `?`. From this experience, maybe in the future we want to change `Option<CodegenStatus>` to be `Result<CodegenStatus, ???>` to be more explicit about the different kinds of errors. It's probably fine as-it for now, though, and my incorrect fix isn't strong enough of a motivation just yet.

Yes that could make sense. Some of these design decisions were made when we did not have as much experience with the language.

PR #12048, opened 2024-11-11
improve error message

soda92:
1. Why does this happen?
2. The error message  is not friendly.


nobu:
Please submit to the upstream https://github.com/ruby/uri instead of the vendored copy of rubygems.

PR #12055, opened 2024-11-11
[don't merge] windows console unicode input study

YO4:
This code is not intended to merge.
In response to a part of #11799, this is a case study and POC code to show a different way to input unicode via console.

This PR allows unicode input in the Windows console.
Receive from the console in UTF-16 and use ruby's transcode infrastructure to obtain the necessary encoding.

Device-side encoding is fixed to UTF-16LE and converted to internal_encoding or external_encoding.
Here, when internal_encoding is specified, external_encoding is ignored.

If no transcoding mechanism is involved, such as when using the binary input method, ANSI input is used as before.


hsbt:
This is study code that is not meant to merge and it no longer applies to master. Closing. The console Unicode work stays on https://bugs.ruby-lang.org/issues/19191.

PR #12065, opened 2024-11-12
Mark strings returned by Symbol#to_s as chilled 

casperisfine:
[[Feature #20350]](https://bugs.ruby-lang.org/issues/20350)

`STR_CHILLED` now spans on two user flags. If one bit is set it marks a chilled string literal, if it's the other it marks a `Symbol#to_s` chilled string.

Since it's not possible, and doesn't make much sense to include debug info when `--debug-frozen-string-literal` is set, we can't include allocation source, but we can safely include the symbol name in the warning message, making it much easier to find the source of the issue.

cc @jhawthorn @eregon

@peterzhu2118 do you know if I should bump the ABI version given some flags were moved around?


eregon:
> @peterzhu2118 do you know if I should bump the ABI version given some flags were moved around?

Yes, changing the values of macros is definitely ABI-incompatible, so it should be bumped.

casperisfine:
@eregon would you happen to remember where it's set.

eregon:
https://github.com/ruby/ruby/blob/master/include/ruby/internal/abi.h

eregon (approved):
LGTM

maximecb (approved):

PR #12069, opened 2024-11-12
YJIT: Specialize `String#[]` (`String#slice`) with fixnum arguments

rwstauner:
`String#[]` is in the top few C calls of several YJIT benchmarks:
liquid-compile rubocop mail sudoku

While it has 5 different argument scenarios the most common by far are one fixnum and two fixnums
so we match those argument types and write the ccall directly to the corresponding C function.

This speeds up these benchmarks by 1-5%.

This is similar to the `String#byteslice` specialization https://github.com/ruby/ruby/pull/9939
but it works harder to find the type info which was necessary for the specialization to apply in the large generated code of the `mail` gem.

```
bench           ruby-master (ms)  stddev (%)  ruby-yjit-ss3 (ms)  stddev (%)  ruby-yjit-ss3 1st itr  ruby-master/ruby-yjit-ss3
liquid-compile  495.7             0.1         493.4               0.1         1.03                   1.00
mail            211.3             0.1         199.5               0.3         1.00                   1.06
rubocop         902.2             0.0         897.1               0.1         0.99                   1.01
sudoku          1420.0            0.0         1409.4              0.0         1.00                   1.01
```

### DIFF
--- a/string.c
+++ b/string.c
@@ -3152,6 +3152,12 @@ rb_str_substr(VALUE str, long beg, long len)
     return str_substr(str, beg, len, TRUE);
 }
 
+VALUE
+rb_str_substr_two_fixnums(VALUE str, VALUE beg, VALUE len, int empty)
+{
+    return str_substr(str, NUM2LONG(beg), NUM2LONG(len), empty);
+}
+
 static VALUE
 str_substr(VALUE str, long beg, long len, int empty)
 {
@@ -5680,9 +5686,7 @@ rb_str_aref_m(int argc, VALUE *argv, VALUE str)
             return rb_str_subpat(str, argv[0], argv[1]);
         }
         else {
-            long beg = NUM2LONG(argv[0]);
-            long len = NUM2LONG(argv[1]);
-            return rb_str_substr(str, beg, len);
+            return rb_str_substr_two_fixnums(str, argv[0], argv[1], TRUE);
         }
     }
     rb_check_arity(argc, 1, 2);
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -5815,7 +5815,24 @@ fn jit_rb_str_aref_m(
             Type::Fixnum => {},
             // Besides Fixnum this could also be a Range or a RegExp which are handled by separate c funcs.
             // Other types will raise.
-            _ => { return false },
+            _ => {
+                // If the context doesn't have the type info we try a little harder.
+                let comptime_arg = jit.peek_at_stack(&asm.ctx, 0);
+                let arg0 = asm.stack_opnd(0);
+                if comptime_arg.fixnum_p() {
+                    asm.test(arg0, Opnd::UImm(RUBY_FIXNUM_FLAG as u64));
+
+                    jit_chain_guard(
+                        JCC_JZ,
+                        jit,
+                        asm,
+                        SEND_MAX_DEPTH,
+                        Counter::guard_send_str_aref_not_fixnum,
+                    );
+                } else {
+                    return false
+                }
+            },
         }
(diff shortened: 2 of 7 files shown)


string.c
```diff
@@ -3154,6 +3154,12 @@ rb_str_substr(VALUE str, long beg, long len)
     return str_substr(str, beg, len, TRUE);
 }
 
+VALUE
+rb_str_substr_with_value_nums(VALUE str, VALUE beg, VALUE len, int empty)
```
maximecb, line 3158: `value_nums` as in fixnums?

rwstauner: Yes?  I originally called it `...with_values`, but then I added the `empty` param which is _not_ a value, so I felt like I needed to the change the name 🤷 
I wish there wasn't already a `rb_str_substr` that took `long`s, this seems like what that function should have been.

rwstauner: so the "nums" being the "begin" and "length" args as opposed to the "empty" which is technically a number but used as a bool.

rwstauner: Open to suggestions

maximecb: Maybe just `rb_str_substr_two_fixnums` ?

maximecb (approved):
Well done Randy 👏

string.c
rwstauner: I like it

yjit/src/codegen.rs
```diff
@@ -5792,6 +5792,82 @@ fn jit_rb_str_byteslice(
...
+                let arg0 = asm.stack_opnd(0);
+                if comptime_arg.fixnum_p() {
+                    asm.test(arg0, Opnd::UImm(RUBY_FIXNUM_FLAG as u64));
+
+                    jit_chain_guard(
+                        JCC_JZ,
+                        jit,
+                        asm,
+                        SEND_MAX_DEPTH,
+                        Counter::guard_send_not_fixnums,
```
k0kubun, line 5830: We overloaded it while pairing, but let's add a separate counter for this. Something like `gurad_send_str_aref_not_fixnum`?

yjit/src/codegen.rs
```diff
@@ -5792,6 +5792,82 @@ fn jit_rb_str_byteslice(
...
+
+    // The "empty" arg distinguishes between the normal "one arg" behavior
+    // and the "two arg" special case that returns an empty string
+    // when the begin index is the length of the string.
+    // See the usages of rb_str_substr in string.c for more information.
+    let (beg_idx, empty, len) = if argc == 2 {
+        (1, Opnd::Imm(1), asm.stack_opnd(0))
+    } else {
+        // If there is only one arg, the length will be 1 (which we encode as a VALUE).
+        (0, Opnd::Imm(0), Opnd::Imm(1 * 2 + 1))
```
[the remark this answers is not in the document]
rwstauner: Thanks, that is nicer :)

yjit/src/codegen.rs
rwstauner: Sounds good.  I added a line for it to `stats.rs`, is that all that is needed?

maximecb: You also need to print the value in `yjit.rb`.

We should probably have some automated/default formatting there 😅

rwstauner: I think it would be covered by this one:
https://github.com/ruby/ruby/blob/ee1cd1656fc667840282a4e9c3f2cc5e3154e50e/yjit.rb#L293
but I'm not sure what code would prove it.

rwstauner:
```
method call exit reasons:
    str_aref_not_fixnum:          6 (100.0%)
```

PR #11094, opened 2024-07-03
Move Array#select to Ruby

tenderlove:
This speeds up the mail benchmark by about 7% on the interpreter:

```
before: ruby 3.4.0dev (2024-07-03T17:01:41Z master f4b313f733) [arm64-darwin23]
after: ruby 3.4.0dev (2024-07-03T17:45:50Z ruby-select de282cacd5) [arm64-darwin23]

-----  -----------  ----------  ----------  ----------  -------------  ------------
bench  before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
mail   72.9         0.8         68.2        1.0         1.02           1.07
-----  -----------  ----------  ----------  ----------  -------------  ------------
Legend:
- after 1st itr: ratio of before/after time for the first benchmarking iteration.
- before/after: ratio of before/after time. Higher is better for after. Above 1 represents a speedup.
```

YJIT is about 13% faster:

```
before: ruby 3.4.0dev (2024-07-03T17:01:41Z master f4b313f733) +YJIT [arm64-darwin23]
after: ruby 3.4.0dev (2024-07-03T17:45:50Z ruby-select de282cacd5) +YJIT [arm64-darwin23]

-----  -----------  ----------  ----------  ----------  -------------  ------------
bench  before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
mail   51.0         0.8         45.2        0.6         1.00           1.13
-----  -----------  ----------  ----------  ----------  -------------  ------------
Legend:
- after 1st itr: ratio of before/after time for the first benchmarking iteration.
- before/after: ratio of before/after time. Higher is better for after. Above 1 represents a speedup.
```

PR #11093, opened 2024-07-03
move Integer#downto to Ruby

tenderlove:
Speeds up ChunkyPNG.

The interpreter is about 70% faster:

```
before: ruby 3.4.0dev (2024-07-03T15:16:17Z master 786cf9db48) [arm64-darwin23]
after: ruby 3.4.0dev (2024-07-03T15:32:25Z ruby-downto 0b8b744ce2) [arm64-darwin23]

----------  -----------  ----------  ----------  ----------  -------------  ------------
bench       before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
chunky-png  892.2        0.1         526.3       1.0         1.65           1.70
----------  -----------  ----------  ----------  ----------  -------------  ------------
Legend:
- after 1st itr: ratio of before/after time for the first benchmarking iteration.
- before/after: ratio of before/after time. Higher is better for after. Above 1 represents a speedup.
```

YJIT is 2.5x faster:

```
before: ruby 3.4.0dev (2024-07-03T15:16:17Z master 786cf9db48) +YJIT [arm64-darwin23]
after: ruby 3.4.0dev (2024-07-03T15:32:25Z ruby-downto 0b8b744ce2) +YJIT [arm64-darwin23]

----------  -----------  ----------  ----------  ----------  -------------  ------------
bench       before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
chunky-png  709.4        0.1         278.8       0.3         2.35           2.54
----------  -----------  ----------  ----------  ----------  -------------  ------------
Legend:
- after 1st itr: ratio of before/after time for the first benchmarking iteration.
- before/after: ratio of before/after time. Higher is better for after. Above 1 represents a speedup.
```

PR #12074, opened 2024-11-13
YJIT: Move Integer#downto, Array#select, and Array#map to Ruby

k0kubun:
This PR reverts https://github.com/ruby/ruby/pull/11275, which reverted https://github.com/ruby/ruby/pull/11093 https://github.com/ruby/ruby/pull/11094 https://github.com/ruby/ruby/pull/11095, using https://github.com/ruby/ruby/pull/11955.


numeric.rb
```diff
@@ -321,6 +321,27 @@ def numerator
...
+  with_yjit do
+    if Primitive.rb_builtin_basic_definition_p(:downto)
+      undef :downto
+
+      def downto(to) # :nodoc:
+        Primitive.attr! :inline_block, :c_trace
+
+        unless defined?(yield)
+          return Primitive.cexpr! 'SIZED_ENUMERATOR(self, 1, &to, int_downto_size)'
+        end
```
maximecb, line 334: Can you explain what this bit does? May be good to add a comment

maximecb (approved):
Thanks for looking at the code size.

I would assume that SFR/Core are going to be relatively similar to lobsters, and so the impact on code size seems acceptable 👍

numeric.rb
k0kubun: added a comment https://github.com/ruby/ruby/pull/12074/commits/300891fce49752130cfc57c7fded7356e0ee2ccd

PR #12093, opened 2024-11-14
Always run rb_thread_schedule_limits when Ractors

jhawthorn:
rb_thread_alone only checks the current ractor for other threads. If we are in multi-ractor mode we are guaranteed to have multiple threads.

cc @ivoanjo

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -1461,7 +1461,7 @@ rb_thread_sleep(int sec)
 static void
 rb_thread_schedule_limits(uint32_t limits_us)
 {
-    if (!rb_thread_alone()) {
+    if (!rb_thread_alone() || rb_multi_ractor_p()) {
         rb_thread_t *th = GET_THREAD();
         RUBY_DEBUG_LOG("us:%u", (unsigned int)limits_us);
 


ivoanjo:
For context, this is a fix for this issue (<https://ivoanjo.me/rubyconf2024/#:~:text=take)%0Aend-,Observing%20the%20GVL%20%237,-(Ruby%203.3%2B)>) -- specifically that threads in separate ractors weren't being preempted on Ruby 3.3+ and thus unless they would run for as long as they needed on CPU even though there's more threads waiting to run.

ivoanjo:
Interesting... I just re-tested it with my example and I still don't see the Ractors being preempted. So I think we're missing something still...

ivoanjo:
I spent some more time on it, and I think the issue goes a bit deeper. Specifically, I think the timer thread is only looking at the main Ractor, whereas it should be looking across all Ractors.

So I think the fix will involve make the timer thread aware of what all ractors are doing, perhaps? There may not be a quick fix here :cry:.

Do you think it's worth creating an entry on bugs.ruby-lang.org for this one?

PR #12094, opened 2024-11-14
Force reset running time in timer interrupt

jhawthorn:
cc @ivoanjo

### DIFF
--- a/ractor_core.h
+++ b/ractor_core.h
@@ -277,11 +277,15 @@ rb_ractor_sleeper_thread_num(rb_ractor_t *r)
 }
 
 static inline void
-rb_ractor_thread_switch(rb_ractor_t *cr, rb_thread_t *th)
+rb_ractor_thread_switch(rb_ractor_t *cr, rb_thread_t *th, bool always_reset)
 {
     RUBY_DEBUG_LOG("th:%d->%u%s",
                    cr->threads.running_ec ? (int)rb_th_serial(cr->threads.running_ec->thread_ptr) : -1,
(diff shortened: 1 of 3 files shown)


nobu:
What does this mean?

ivoanjo:
@nobu this is a fix for [this issue I recently spotted](https://ivoanjo.me/rubyconf2024/#:~:text=Observing%20the%20GVL%20%238).

Specifically, when running this example:

```ruby
require "gvl-tracing"

def counter_loop
  counter = 0
  counter += 1 while counter < 100_000_000
end

GvlTracing.start("rc-example2.json", os_threads_view_enabled: true) do
  threads = 9.times.map { Thread.new { counter_loop } }
  counter_loop
  threads.map(&:join)
end
```

With `RUBY_MN_THREADS=1`, the MN switching code switches the thread before `rb_ractor_thread_switch` gets called, and thus this condition

```c
    if (cr->threads.running_ec != th->ec) {
        th->running_time_us = 0;
    }
```

was not being entered, and thus the `running_time_us` never got reset.

The result is that instead of being given the usual 100ms to run, whenever Ruby checked if the `running_time_us` was above 100ms, this was true, so Ruby would immediately switch the thread out.

In the pic below you can see how the main thread is able to run 100ms period, but the other threads only get 100ms the first time, and then afterwards only get 10ms at a time.

>

nobu (approved):

ivoanjo:
Re-ran the example with this fix, as expected it looks good! The extra threads are still all sharing the underlying native thread, and they all get their 100ms.
image

ivoanjo:
Not that I know of! :)

PR #12097, opened 2024-11-15
Disable HEv2 tests temporarily

shioimm:
To suppress error log output in CI.
They should have been DISABLE in PR #12070.


hsbt:
`pend` is better than `return` while our test suite.

PR #12100, opened 2024-11-15
[Backport 3.2] wasm: align fiber stack pointer to 16 bytes

kateinoigakukun:
In WebAssembly C ABI, the linear stack pointer must be always aligned to 16 bytes like other archs.
The misaligned stack pointer causes some weird memory corruption since compiler assumes the aligned stack pointer.

----

Backport ticket: https://bugs.ruby-lang.org/issues/20898

### DIFF
--- a/coroutine/asyncify/Context.h
+++ b/coroutine/asyncify/Context.h
@@ -13,6 +13,7 @@
 
 #include <stddef.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


XrXr:
Looks to be merged already in 1fc0895971812d5287b23d8cbb00529e425041c0.

PR #12109, opened 2024-11-18
[DOC] Fix the default `limit` of `String#split`

kyanagi:
We can't pass `nil` as the second parameter of `String#split`.
Therefore, descriptions like "if limit is nil, ..." are not appropriate.

```
% ruby -e '"".split("", nil)'
-e:1:in `split': no implicit conversion from nil to integer (TypeError)

"".split("", nil)
         ^^^^^^^
        from -e:1:in `<main>'
```

### DIFF
--- a/doc/string/split.rdoc
+++ b/doc/string/split.rdoc
@@ -12,7 +12,7 @@ When +field_sep+ is <tt>$;</tt>:
   the split occurs just as if +field_sep+ were given as that string
   (see below).
 
-When +field_sep+ is <tt>' '</tt> and +limit+ is +nil+,
+When +field_sep+ is <tt>' '</tt> and +limit+ is +0+ (its default value),
 the split occurs at each sequence of whitespace:
 
   'abc def ghi'.split(' ')         => ["abc", "def", "ghi"]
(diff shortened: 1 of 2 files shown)


doc/string/split.rdoc
```diff
@@ -12,7 +12,7 @@ When +field_sep+ is <tt>$;</tt>:
   the split occurs just as if +field_sep+ were given as that string
   (see below).
 
-When +field_sep+ is <tt>' '</tt> and +limit+ is +nil+,
+When +field_sep+ is <tt>' '</tt> and +limit+ is omitted,
```
jeremyevans, line 15: I think we should change `omitted` to `0`, both here and in other cases.  When `limit` is omitted, it has the default value of `0`, which is why it works. We should only mention behavior with omission of arguments if the behavior is different than when passing the default argument values explicitly.

jeremyevans (changes requested):
Thank you for working on this. I agree we should update the documentation, but I think we should change the wording.

doc/string/split.rdoc
kyanagi: Thank you for your review. PR fixed.

jeremyevans (approved):
Changes look good, thank you!

PR #12111, opened 2024-11-18
Extract bundled_gems_spec.rb from `spec/bundler`.

hsbt:
That example is not for Bundler. We should move that to under the `spec`.

And I make that test at `make test-bundled-gems`, not `make test-bundler`.


deivid-rodriguez:
I'm seeing some failures in the master branch now that may be caused by this PR? I'm seeing the same failures in `rubygems/rubygems` too when running tests in the same way `ruby/ruby` runs them.

deivid-rodriguez:
Actually the failures in `rubygems/rubygems` are using an older revision of `ruby/ruby`, so it must be something else 🤔.

deivid-rodriguez:
In the end it was a change in sinatra, completely unrelated to this pull request. Sorry for the false alarm!

PR #12112, opened 2024-11-18
Fix CI

deivid-rodriguez:
Bundler specs broke because of a change in Sinatra 4.1.0. In my opinion, a change in an external test dependency should not break ruby-core's CI.

This is how Bundler specs were run until
426434556f43e597da8ddfa33c659321971f1b83, but I'm not sure what the motivation for that commit was.

I'm working on adapting Bundler specs to sinatra changes but for now I wanted to propose this since it encourages a more stable CI both here and in rubygems/rubygems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -235,6 +235,7 @@ lcov*.info
 # for `make test-bundler`
 /.rspec_status
 /tool/bundler/*.lock
+!tool/bundler/test_gems.rb.lock
 
 # /tool/
 /tool/config.guess
--- a/tool/bundler/test_gems.rb.lock
+++ b/tool/bundler/test_gems.rb.lock
@@ -1,0 +1,74 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    base64 (0.2.0)
+    builder (3.3.0)
+    compact_index (0.15.0)
+    mustermann (3.0.3)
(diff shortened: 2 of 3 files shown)


deivid-rodriguez:
PR to fix Bundler specs with Sinatra 4.1.0 is https://github.com/rubygems/rubygems/pull/8260, so an alternative fix to get CI here back to green would be to backport that PR once merged.

hsbt:
Thanks @deivid-rodriguez 

https://github.com/ruby/ruby/commit/519b23369514f72e89ea8bed2e053685c4060ac7 resolved this issue. I'm closing this.

FYI: I ignored lockfile because they are updated while test suite of bundler in the past. It may be resolve today. I will look after Ruby 3.4 release.

deivid-rodriguez:
> FYI: I ignored lockfile because they are updated while test suite of bundler in the past. It may be resolve today. I will look after Ruby 3.4 release.

Thanks, that's good to know! Once you try it, let me know if you find any issues 👍.

deivid-rodriguez:
@hsbt I think I run into the same issue you were running when platform specific versions of nokogiri would be removed from the lockfile during the test suite in Ruby 3.4. I think https://github.com/rubygems/rubygems/pull/8251 should fix that, so we can retry syncing the lockfiles too after that is merged.

PR #12114, opened 2024-11-19
[Bug #20900] Warn deprecated constant when removing

nobu:
[[Bug #20900]](https://bugs.ruby-lang.org/issues/20900)

### DIFF
--- a/test/ruby/test_module.rb
+++ b/test/ruby/test_module.rb
@@ -2148,9 +2148,8 @@ class TestModule < Test::Unit::TestCase
       Warning[:deprecated] = false
       Class.new(c)::FOO
     end
-    assert_warn('') do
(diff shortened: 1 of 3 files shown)


test/ruby/test_module.rb
```diff
@@ -2148,9 +2148,8 @@ def test_deprecate_constant
       Warning[:deprecated] = false
       Class.new(c)::FOO
     end
-    assert_warn('') do
-      Warning[:deprecated] = false
-      c.class_eval "FOO"
```
deivid-rodriguez, line 2153: Shouldn't this test be kept? Or is this commit changing the behavior?

deivid-rodriguez (commented):
This is a very nice change. I just noticed one test being removed and I don't fully understand why.

PR #12116, opened 2024-11-19
Fix test failure for Ruby 3.3

hsbt:
Fixed http://rubyci.s3.amazonaws.com/arch/ruby-3.3/log/20241119T033259Z.fail.html.gz and https://github.com/ruby/actions/actions/runs/11898909121/job/33156608325


hsbt:
@k0kubun I merged this we should suppress failure alert of our daily CI.

PR #12123, opened 2024-11-19
Optimize instructions when creating an array just to call `include?`

rwstauner:
This creates a new instruction `opt_duparray_send` to avoid the array allocation for static arrays that are created just for calling `.include?(value)`.
This also adds the same functionality to the existing `opt_newarray_send` instruction for dynamic arrays.

Together this supports several simple forms and avoids allocating a new array for them:

```
[:a, :b].include?(@ivar)
[1, 2].include?(local_var.to_f)

[:a, something_dynamic].include?(:b)
["1", DYNAMIC].include?(@x.to_s)
[DYNAMIC, true, "true", 1].include?(method1.method2)
```

This adds a BOP for `include?` and tracks redefinition for it on Array.

Doing some simple code searches this pattern seems to be used quite often.

Using the `yjit-bench` suite most benchmarks waffle around half a percent either way.
However the `fluentd` and `optcarrot` benchmarks appear consistently 3-6% faster with the interpreter:

```
bench           master (ms)  stddev (%)  opt (ms)  stddev (%)  opt 1st itr  master/opt
fluentd         458.2        6.2         429.0     3.1         1.019        1.068               
optcarrot       5687.2       0.2         5498.3    0.1         1.036        1.034               
bench           master (ms)  stddev (%)  opt (ms)  stddev (%)  opt 1st itr  master/opt
fluentd         465.3        4.1         440.9     5.0         0.991        1.055               
optcarrot       5815.2       0.2         5490.3    0.1         1.061        1.059               
bench           master (ms)  stddev (%)  opt (ms)  stddev (%)  opt 1st itr  master/opt
fluentd         459.7        6.1         442.2     5.0         0.988        1.040               
optcarrot       5713.3       0.2         5505.0    0.1         1.037        1.038
```

With YJIT `optcarrot` goes back to `1` but `fluentd` gets even faster and `mail` seems to improve by 2-3%:
```
bench           master (ms)  stddev (%)  opt (ms)  stddev (%)  opt 1st itr  master/opt
fluentd         448.8        0.2         389.0     5.8         0.937        1.154
mail            105.1        0.2         102.6     0.4         1.004        1.024
bench           master (ms)  stddev (%)  opt (ms)  stddev (%)  opt 1st itr  master/opt
fluentd         415.5        0.1         382.5     7.4         0.936        1.086
mail            105.0        0.2         102.0     0.4         1.029        1.030      
```


bootstraptest/test_insns.rb
```diff
@@ -214,8 +214,21 @@ def freeze
...
 
+  [ 'opt_duparray_send', %q{ x = :a; [:a, :b].include?(x) }, ],
+  [ 'opt_duparray_send', <<-'},', ], # {
+    class Array
+      def include?(i)
+        i == 1
+      end
+    end
+    x = 3
+    [:a, :b].include?(1)
```
nirvdrum, line 225: Should this be `include?(x)`? If not, what's the assignment in the line above needed for?

vm_insnhelper.c
```diff
@@ -6152,6 +6152,29 @@ vm_opt_str_freeze(VALUE str, int bop, ID id)
 /* this macro is mandatory to use OPTIMIZED_CMP. What a design! */
 #define id_cmp idCmp
 
+static VALUE
+vm_opt_duparray_include_p(rb_execution_context_t *ec, const VALUE ary, VALUE target)
+{
+    if (BASIC_OP_UNREDEFINED_P(BOP_INCLUDE_P, ARRAY_REDEFINED_OP_FLAG)) {
```
nirvdrum, line 6158: I think it's fairly unlikely this will get redefined. You should be able `UNLIKELY` here to help the compiler order code more efficiently.

vm_insnhelper.c
```diff
@@ -6233,6 +6256,26 @@ rb_vm_opt_newarray_hash(rb_execution_context_t *ec, rb_num_t num, const VALUE *p
 VALUE rb_setup_fake_ary(struct RArray *fake_ary, const VALUE *list, long len);
 VALUE rb_ec_pack_ary(rb_execution_context_t *ec, VALUE ary, VALUE fmt, VALUE buffer);
 
+static VALUE
+vm_opt_newarray_include_p(rb_execution_context_t *ec, rb_num_t num, const VALUE *ptr, VALUE target)
+{
+    if (BASIC_OP_UNREDEFINED_P(BOP_INCLUDE_P, ARRAY_REDEFINED_OP_FLAG)) {
```
nirvdrum, line 6262: I think it's fairly unlikely this will get redefined. You should be able `UNLIKELY` here to help the compiler order code more efficiently.

vm_insnhelper.c
maximecb: Nice catch Kevin

yjit/src/codegen.rs
```diff
@@ -4298,6 +4298,53 @@ fn gen_opt_newarray_max(
...
+
+    let ary = jit.get_arg(0);
+    let argc = jit.get_arg(2).as_usize();
+
+    // Save the PC and SP because we may call #include?
+    jit_prepare_non_leaf_call(jit, asm);
+
+    extern "C" {
+        fn rb_vm_opt_duparray_include_p(ec: EcPtr, ary: VALUE, target: VALUE) -> VALUE;
+    }
```
maximecb, line 4328: Here I think it's cool that you added a YJIT specialization (thank you).

However I wish YJIT could do more. Presumably we can iterate over the values in the YJIT and make this much faster than calling into a C function that sets up a fake array. Though I guess maybe this is harder than I imagine because you need to match Ruby equality semantics, which can also be redefined?

yjit/src/codegen.rs
```diff
@@ -4310,6 +4357,8 @@ fn gen_opt_newarray_send(
         gen_opt_newarray_max(jit, asm)
     } else if method == VM_OPT_NEWARRAY_SEND_HASH {
         gen_opt_newarray_hash(jit, asm)
+    } else if method == VM_OPT_NEWARRAY_SEND_INCLUDE_P {
+        gen_opt_newarray_include_p(jit, asm)
     } else if method == VM_OPT_NEWARRAY_SEND_PACK {
         gen_opt_newarray_pack_buffer(jit, asm, 1, None)
     } else if method == VM_OPT_NEWARRAY_SEND_PACK_BUFFER {
```
maximecb, line 4364: Are all these functions mostly doing the same thing, and if so, could we make them generic on the function being called? I know this is not your doing. Just wondering if we can avoid doing boilerplate for every newarray send type.

maximecb:
Seems like a sensible optimization to do for a pattern that people really do use in Ruby 👍 

> It seems to offer a 1-2% speedup on several benchmarks, both using the interpreter and using YJIT.

I believe you it's faster, but could you share some of these numbers on the PR? Just pasting from yjit-bench?

vm_insnhelper.c
maximecb: Yes

maximecb:
Do you know what's with the compilations CI failure?

yjit/src/codegen.rs
maximecb: Sounds good 👍

maximecb (approved):

maximecb (approved):

rwstauner:
See also: https://github.com/ruby/ruby/pull/12142

ko1:
```
[1, 2].include?(3)
#=>
== disasm: #<ISeq:<main>@/home/ko1/ruby/src/master/test.rb:2 (2,0)-(2,18)>
0000 duparray                               [1, 2]                    (   2)[Li]
0002 putobject                              3
0004 opt_send_without_block                 <calldata!mid:include?, argc:1, ARGS_SIMPLE>
0006 leave
```

is it intentional?

rwstauner:
It was.
If the array is static, and the argument is a static value, this is likely not performance critical code (if it was you'd just replace it with `true` or `false`).

We could support the pattern with something like this:

```
index 7de791a033..dcc950956d 100644
--- a/compile.c
+++ b/compile.c
@@ -4218,6 +4218,7 @@ iseq_specialized_instruction(rb_iseq_t *iseq, INSN *iobj)
         INSN *niobj = (INSN *)iobj->link.next;
         if ((IS_INSN_ID(niobj, getlocal) ||
              IS_INSN_ID(niobj, getinstancevariable) ||
+             IS_INSN_ID(niobj, putobject)) &&
              IS_INSN_ID(niobj, putself)) &&
             IS_NEXT_INSN_ID(&niobj->link, send)) {
```

It's quite a small change which I would be happy to add if you think it's valuable,
but I wasn't sure where to draw the line with how many patterns to support since the optimizer code has to look for specific insns.

For duparray it should optimize if you pass it a variable.
For newarray (non static arrays) you can pass a static value as an arg:

https://github.com/ruby/ruby/pull/12123/files#diff-a5ba41b51e3655f9f244362a616282b5119d3e15dd6c52ee999bbdfcc5b86a77R4178

PR #12126, opened 2024-11-20
[DOC] Refine `String#split` documentation

kyanagi:
The current documentation for `String#split` is missing some details about its behavior.

* Removing leading whitespace by `split(' ')`
* Splitting into indivisual characters by `split('')`

Additionally, the first argument `field_sep` and the second argument `limit` operate independently. Explaining them separately would enhance readability.


doc/string/split.rdoc
```diff
@@ -12,47 +12,59 @@ When +field_sep+ is <tt>$;</tt>:
...
-
-  'abracadabra'.split('ab')  => ["", "racad", "ra"]
-  'aaabcdaaa'.split('a')     => ["", "", "", "bcd"]
-  ''.split('a')              => []
-  '3.14159'.split('1')       => ["3.", "4", "59"]
+  'abc  def   ghi'.split(' ')       # => ["abc", "def", "ghi"]
+  ''.split(' ')                     # => []
+  '  abc  def'.split(' ')           # => ["abc", "def"]
+
+When +field_sep+ is <tt>''</tt> (an empty string), the string is split into indivisual characters.
```
jeremyevans, line 24:
```suggestion
When +field_sep+ is <tt>''</tt> (an empty string), the string is split into individual characters.
```

doc/string/split.rdoc
```diff
@@ -12,47 +12,59 @@ When +field_sep+ is <tt>$;</tt>:
...
-trailing empty substrings are not returned:
+When +field_sep+ is a Regexp,
+the split occurs at each occurrence of a match:
 
   'abracadabra'.split(/ab/) # => ["", "racad", "ra"]
-  'aaabcdaaa'.split(/a/)   => ["", "", "", "bcd"]
-  'aaabcdaaa'.split(//)    => ["a", "a", "a", "b", "c", "d", "a", "a", "a"]
+  'aaabcdaaa'.split(/[bcd]+/)   # => ["aaa", "aaa"]
+  'aaabcdaaa'.split(//)     # => ["a", "a", "a", "b", "c", "d", "a", "a", "a"]
   '1 + 1 == 2'.split(/\W+/) # => ["1", "1", "2"]
```
jeremyevans, line 44: It may not be obvious to the user how to split on `' '` without special handling.  Do you think it's worth adding an example telling the user to split on `/ /` instead of `' '` if they want to split on each single space?

jeremyevans (approved):
This looks good, thank you for working on this.  I think it would good for @BurdetteLamar to review as well.

I found one spelling issue, which should be fixed before merging.  The other comment is just a suggestion, you can merge without it if you don't think it would be helpful.

doc/string/split.rdoc
kyanagi: Thank you! I also found another typo and fixed them together.

doc/string/split.rdoc
kyanagi: Indeed, it can be somewhat confusing that splitting by a single space requires `/ /` instead of `' '`.
I believe it would be helpful to include this as an example.

BurdetteLamar:
Thanks, @everyone, for calling this to my attention.  I'd like to address the comments here about examples, but in a new PR, after this is merged.  (I like to think I'm good with the docs, but I'm definitely not good with the details of reviewing on GH.)

PR #12130, opened 2024-11-20
[prism/compiler] end_cursor should never be NULL

eightbitraptor:
This fixes a failed Prism assertion [reported to SimpleCov](https://github.com/simplecov-ruby/simplecov/issues/1113
)

This can be repro'd as follows:

1. Create a file `test.rb` containing the following code

```
@foo&.(@bar)
```

2. require it with branch coverage enabled

```
ruby -rcoverage -e "Coverage.start(branches: true); require_relative 'test.rb'"
```

The assertion is failing because the Prism compiler is incorrectly detecting the start and end cursor position of the call site for the implicit call .()

This patch replicates the parse.y behaviour of setting the default end_cursor to be the final closing location of the call node.

This behaviour can be verified against `parse.y` by modifying the test command as follows:

```
ruby --parser=parse.y -rcoverage -e "Coverage.start(branches: true); require_relative 'test.rb'"
```

[Bug #20866]


kddnewton (approved):
Thanks!

PR #12131, opened 2024-11-20
Remove unused arg in parse.y Github CI

eightbitraptor:
It contained a typo (with-praser instead of with-parser), but regardless of this, it is unnecessary anyway, as we RUN_OPTS and SPECOPTS to configure the parser on Github Actions.


hsbt:
It seems test for `configure` and `make` commands.

PR #12134, opened 2024-11-21
Fix the usage of realloc

mame:
http://ci.rvm.jp/results/trunk-repeat50@ruby-sp2-noble-docker/5420911
```
/tmp/ruby/src/trunk-repeat50/ext/socket/ipsocket.c: In function ‘reallocate_connection_attempt_fds’:
/tmp/ruby/src/trunk-repeat50/ext/socket/ipsocket.c:292:62: warning: pointer ‘fds’ may be used after ‘realloc’ [-Wuse-after-free]
  292 |     for (int i = current_capacity; i < new_capacity; i++) fds[i] = -1;
      |                                                              ^
/tmp/ruby/src/trunk-repeat50/ext/socket/ipsocket.c:288:9: note: call to ‘realloc’ here
  288 |     if (realloc(fds, new_capacity * sizeof(int)) == NULL) {
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/ext/socket/ipsocket.c
+++ b/ext/socket/ipsocket.c
@@ -281,15 +281,18 @@ allocate_connection_attempt_fds(int additional_capacity)
 }
 
 static int
-reallocate_connection_attempt_fds(int *fds, int current_capacity, int additional_capacity)
+reallocate_connection_attempt_fds(int **fds, int current_capacity, int additional_capacity)
 {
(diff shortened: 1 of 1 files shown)


shioimm (approved):
👍

PR #12138, opened 2024-11-21
allow malloc outside of GVL

shyouhei:
We currently do not support calling ruby_xmalloc out of GVL.  OTOH oftentimes people do `#define malloc ruby_xmalloc`.  Under GVL that has no practical problem but without one it must be a nightmare.

Because whether a runtime context has GVL or not is not fixed when the macro is expanded, why don't we actively check that property on the fly and (instead of abnormal end of the process) return NULL.

Better safe than sorry I guess.

One problem here is the GC API.  It seems `rb_gc_impl_malloc()` currently assumes GVL because it runs GC inside.  If we want this non-GVL malloc condition to also be pluggable, there need be something more to consider.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -4208,6 +4208,19 @@ rb_malloc_info_show_results(void)
 void *
 ruby_xmalloc(size_t size)
 {
+    if (! ruby_thread_has_gvl_p()) {
+        /* if (ruby_native_thread_p()) {
+         *     // We could reacquire the GVL here using rb_thread_call_with_gvl
+         *     // but what if that raised an  exception?  There would be no way
+         *     // to recover.
+         *     //
+         *     // This saves no one from nothing.
+         *     return rb_thread_call_with_gvl(ruby_xmalloc, (void*)size);
+         * }
+         */
+        return ruby_mimmalloc(size);
+    }
+
     if ((ssize_t)size < 0) {
         negative_size_allocation_error("too large allocation size");
     }
@@ -4226,12 +4239,38 @@ ruby_malloc_size_overflow(size_t count, size_t elsize)
 void *
 ruby_xmalloc2(size_t n, size_t size)
 {
+    if (! ruby_thread_has_gvl_p()) {
+        /* Out of GVL.  No GC, and no way to inform the integer overflow. */
+        struct rbimpl_size_mul_overflow_tag of =
+            rbimpl_size_mul_overflow(n, size);
+
+        if (of.left) {
+            return NULL;
+        }
+        else {
+            return ruby_mimmalloc(of.right);
+        }
+    }
+
     return rb_gc_impl_malloc(rb_gc_get_objspace(), xmalloc2_size(n, size));
 }
 
 void *
 ruby_xcalloc(size_t n, size_t size)
 {
+    if (! ruby_thread_has_gvl_p()) {
+        /* Out of GVL.  No GC, and no way to inform the integer overflow. */
+        struct rbimpl_size_mul_overflow_tag of =
+            rbimpl_size_mul_overflow(n, size);
+
+        if (of.left) {
+            return NULL;
+        }
+        else {
+            return ruby_mimcalloc(of.right, 1);
+        }
+    }
+
     return rb_gc_impl_calloc(rb_gc_get_objspace(), xmalloc2_size(n, size));
 }
 
@@ -4241,6 +4280,12 @@ ruby_xcalloc(size_t n, size_t size)
 void *
 ruby_sized_xrealloc(void *ptr, size_t new_size, size_t old_size)
 {
+    if (! ruby_thread_has_gvl_p()) {
+        /* Out of GVL, no GC. */
+        /* Also there is no such thing like ruby_mimrealloc. */
+        return realloc(ptr, new_size);
+    }
+
     if ((ssize_t)new_size < 0) {
         negative_size_allocation_error("too large allocation size");
     }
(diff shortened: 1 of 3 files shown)


shyouhei:
(CI failure seems unrelated)

ko1:
Current `ruby_xmalloc()` also be allowed called without GVL, but it `exit(1)` if `malloc()` returns NULL.

For performance perspective, acquiring GVL should be checked only if `malloc()` returns NULL. Otherwise, returning result of `malloc()` is enough.

shyouhei:
> Current `ruby_xmalloc()` also be allowed called without GVL, but it `exit(1)` if `malloc()` returns NULL.

Yes I meant that behaviour being "not supported".

"There _must_ be no way for `malloc` to survive an error" is a super strong opinion that I honestly have a bit of difficulty to agree with.

> For performance perspective, acquiring GVL should be checked only if `malloc()` returns NULL. Otherwise, returning result of `malloc()` is enough.

1. Do we need performance here?
1. In order to achieve this `rb_gc_impl_malloc()` needs a redesign.  I don't think it returns NULL.

shyouhei:
@peterzhu2118 OK, I know the currently proposed modification is suboptimal.  You want your implementation to have more control.  Understandable.  Then what do you think we should do for the problem (that there is no established way to handle malloc failure except raising exceptions, which is not possible without GVL)?

peterzhu2118:
> Understandable. Then what do you think we should do for the problem (that there is no established way to handle malloc failure except raising exceptions, which is not possible without GVL)?

Maybe an alternative solution is to change `rb_gc_impl_malloc` to return NULL when it absolutely cannot allocate any memory (e.g. after it runs a GC), and then Ruby's `ruby_xmalloc` can acquire the GVL and raise an exception (or maybe we can consider `rb_bug` when running out of memory if there is no GVL 🤔 ).

KJTsanaktsidis:
> or maybe we can consider rb_bug when running out of memory if there is no GVL

Honestly I would just go with this. Unless you're running your system with overcommit disabled (which I _think_ is pretty uncommon) or you do something crazy and `mlockall(2)` all your memory allocations into physical RAM, your process exists at the mercy of the OOM killer anyway and any page you try and access for the first time might generate a fault and invoke the OOM killer on you. Or, your process might become a victim of the OOM killer even if some _other_ process consumed the last bit of RAM.

tl;dr - IMO trying to do sensible things in out-of-memory conditions on normal, generic linux systems (and i assume many other systems work the same way) is pretty pointless and crashing immediately makes sense.

shyouhei:
On linux systems, glibc's allocator can fail for instance under `ulimit(2)` influence.  This is not that esoteric.  For instance ruby_calloc can allocate terabytes of regions because of its arguments being multiplied, but failure in doing so doesn't necessarily mean we should die immediately.

Out-of-GVL threads cannot raise exceptions even if it reacquires one.  Ruby level exceptions unwinds machine stack.  Everything on the stack has to be under ruby's influence then.  Otherwise the stack gets smashed.  So no, we cannot raise.

shyouhei:
@KJTsanaktsidis 

```console
$ $ /bin/sh -c 'ulimit -v 65535; ruby -ve0'
```

And you can see something other than OOM.  (65536 is too low for practical use of course, just to demonstrate the situation)

KJTsanaktsidis:
`ulimit -v` actually sets `RLIMIT_AS`, which is limiting the amount of _virtual_ memory address space that can be mapped, rather than how many actual pages of memory are resident in RAM or swap. When `RLIMIT_AS` is set, calls to `mmap(2)`, `brk(2)` etc do fail, so glib can see `mmap(2)` return `MAP_FAILED` with `ENOMEM`, and so return `NULL` out of `malloc(3)`.

However, I don't think anybody should be using `ulimit -v` for anything, at all, ever. It's not useful to limit the amount of mapped virtual memory a program can use, really; it's not a finite resource. If a program maps a few terrabytes of file-backed page, the system can manage that just fine by reading the file from disk as needed and discarding pages when they're not.

The normal way to limit an application's memory usage on Linux is with cgroups, and cgroups limit actual resident memory in RAM or swap. And because of this, large anonymous mappings bigger than the cgroup limit with `mmap(2)` etc are totally allowed, and it's only when the application tries to write to them that a page fault occurs and the kernel realises that actually allocating a new memory page would blow the cgroup limit. At that point, all the kernel can do is kill the program.

So, IMHO, because of this, "in reality" out-of-memory conditions usually just result in a SIGKILL to the application anyway, so I don't think it's worth doing anything more elaborate than just hard crashing in the rare instance we get `NULL` out of `malloc(3)`.

shyouhei:
I'm not going to debate what is the idiomatic way to restrict memory usage.  I just say NULL is a thing that a malloc returns, and ruby must handle it.  It is nothing special.

XrXr:
@KJTsanaktsidis, [fly.io](https://fly.io/) widely deploys virtual memory limits, IIUC due to their virtualization topology:

- https://community.fly.io/t/fatal-process-oom-in-failed-to-reserve-virtual-memory-for-coderange/3212
- https://community.fly.io/t/next-js-and-graphql-ran-out-of-memory/14787
- https://bugs.ruby-lang.org/issues/20184#note-4

> ... it's not a finite resource

It very much is with virtualization, where all the guests need to share the however many level page table the host has.

I think xmalloc outside of GVL is a pretty advanced, and maybe it's fine for advanced usecases to have a harder-to-use interface (i.e. outside the GVL, caller is responsible for NULL handling). But that does complicate the interface for everyone since people now need to know whether they have the GVL before calling. Mostly mitigable with good documentation?

ko1:
> Yes I meant that behaviour being "not supported".

In the past, there was an argument that this was a good idea, and it is the current behaviour.
I'm not sure it is documented though.

KJTsanaktsidis:
I actually had a bit of a look at https://bugs.ruby-lang.org/issues/20184 and I think I've been missing a bit of nuance in my statements that "overcommit means malloc should never return NULL".

Actually the default value for `vm.overcommit_memory` in Linux is 0, which means "heuristic overcommit". That means that the system can and will allow you to map more memory in than exists, and that the memory might not be available when you use it, in which case the OOM killer is invoked 

BUT - it _also_ means that "Obvious overcommits of memory are refused" and that "a seriously wild allocation fails". That means if you try and mmap a 384MB chunk of anonymous memory on a system with only 256MB of RAM+swap, the kernel will reject that at `mmap` time (and thus it's possible for malloc to return zero if an allocation is just way way too big). The issue in https://bugs.ruby-lang.org/issues/20184 was that we were mmaping a buffer way way way too big for the available memory, and so the overcommit heuristics refused to map it.

(see: https://www.kernel.org/doc/Documentation/vm/overcommit-accounting)

So, I was wrong to say "a sanely configured Linux system should never return NULL from malloc", because there _are_ circumstances where the default overcommit handling will fail page mappings due to a lack of physical memory.

PR #12144, opened 2024-11-21
YJIT: introduce branch timeout to avoid compiling cold branches

maximecb:
Here we're using CPU time for the timeout. This is not perfect, but we need some kind of proxy for how much time, how much code has been executed since the ISEQ started being compiled. In the past I used the amount of blocks compiled as a proxy, but this has its own issue, which is that as compilation slows down, we hit the cold threshold less and less.

I set the default value of the branch timeout to a fairly generous 30s. Experimentation in production will be needed to determine a good value and if this is helpful in stopping compilation after warmup or not.


yjit/src/core.rs
```diff
@@ -1774,9 +1773,25 @@ pub struct IseqPayload {
     // Blocks that are invalidated but are not yet deallocated.
     // The code GC will free them later.
     pub dead_blocks: Vec<BlockRef>,
+
+    // Process CPU time when compilation started for this ISEQ
+    pub comp_time_start: f64,
```
k0kubun, line 1778: Since the option specifies a timeout in seconds, we probably don't need all the 64 bits to maintain the accuracy. Could we use something like `u32` that stores seconds elapsed since the process started?

k0kubun:
How much impact did it have on `yjit_alloc_size`?

Also, do we need both `--yjit-cold-threshold` and `--yjit-branch-timeout`? WDYT about removing or deprecating `--yjit-cold-threshold` once we merge this?

maximecb:
> How much impact did it have on `yjit_alloc_size`?

That's going to be very dependent on the value set for the branch timeout. Currently, I set it high enough that it has no real impact on the benchmarks, but I anticipate that in production it will cut down on generated code after the app has started serving requests.
 
> Also, do we need both `--yjit-cold-threshold` and `--yjit-branch-timeout`? WDYT about removing or deprecating `--yjit-cold-threshold` once we merge this?

They do different things. This is to cut out on branches being compiled. The cold threshold was to try and estimate how frequently a method is called to determine if it's worth compiling. That being said I would like to revisit the cold threshold (in another PR) to convert it to use CPU seconds as well, instead of number of blocks compiled.

byroot:
> I set the default value of the branch timeout to a fairly generous 45s.

Experimentation is king, so I can't say much, but my expectation reading this is that 45s is very low. Currently we're warming up over a way longer amount of time.

There's also the question of whether just lowering the `exec-memsize` wouldn't have a very similar effect in a much more deterministic fashion. I think we discussed in the past how I don't like time based heuristic. CPU time is better than WALL time, but still too unpredictable IMHO.

maximecb:
> I think we discussed in the past how I don't like time based heuristic. CPU time is better than WALL time, but still too unpredictable IMHO.

Yes, this is partly why I tagged you. It's hard because we don't really have a better metric for "how much computation has happened in this process". It has its flaws. We could try to increment some counter during execution, but that would slow execution down, which is also not great.

> There's also the question of whether just lowering the exec-memsize wouldn't have a very similar effect in a much more deterministic fashion.

Well, so what this is getting at is that it's not just a question of how much code we compile, but what code we compile. How can we make use of our memory (and compilation time) budget more effectively. Ideally we want to avoid compiling infrequent code paths.

The heuristic here is that if you started to compile a method 45 seconds ago, and you only hit a branch in that method for the first time CPU 45 second later, then probably this branch is not very frequent, and so not worth compiling. I can be more conservative and increase it to 60s or 80s. I'll probably make it 80s so that when we deploy this in prod we don't get a big negative impact by accident. Mitigate the. risks.

How this came about is that I was experimenting with the lobsters benchmark and looking at its compilation log. What I found is that lobsters never fully stops compiling code. Even after 500 benchmark iterations, it is still slowly hitting more infrequent code paths. Presumably the same thing is happening in our deployments. Ideally, we would have better heuristics that would make it so that we eventually fully stop compiling new code.

XrXr:
Oh, storing clock_gettime(CLOCK_PROCESS_CPUTIME_ID) is not fork safe because CPU time gets reset on fork, but we keep all the old timestamps. This was probably why CPU time looked like it was decreasing. I guess one way to make it work is to keep a fork generation (or the pid, but it's racy due to pid reuse) with each timestamp so we could disregard old ones after fork. Or use std::time::Instant (monotonic clock) which isn't tied to the process.

byroot:
> We would take a small performance hit if we add a call counter. Probably not huge, but I don't love adding more instructions to the hot path 😢

Maybe I'm totally off, but I remember needing to do this: https://github.com/ruby/ruby/pull/8259 to stop the current counter from being incremented forever. Presumably we could just let it increment more and flip some boolean to disable cold branch compilation once a second threshold it hit?

> Would you be ok with us trying it in prod with a long-ish timeout like 80 to 120 seconds?

It's not my call. You do you. I'd recommend trying on experimental pods though, so that the blast radius is limited.

maximecb:
> Maybe I'm totally off, but I remember needing to do this: https://github.com/ruby/ruby/pull/8259 to stop the current counter from being incremented forever. Presumably we could just let it increment more and flip some boolean to disable cold branch compilation once a second threshold it hit?

The main issue there is that this counter only gets incremented when we call from the interpreter into YJIT. It doesn't track JIT-to-JIT calls. Incrementing it would also make copy-on-write perf worse.

There is unfortunately no perfect proxy. I could go ahead and just use the `vm_insn_count` counter that we already have, but that also has its own issues. CPU time, thread time or wall clock time seems like it's probably the least bad option? Unless there is some other counter that gets regularly incremented in CRuby.

Something else we could do, maybe, is just use the minor GC count? Maybe that's a good proxy for how much activity/computation has happened so far? Would need to have some idea of how often we run minor GCs.

> It's not my call. You do you. I'd recommend trying on experimental pods though, so that the blast radius is limited.

Makes sense.

> Or use `std::time::Instant` (monotonic clock) which isn't tied to the process.

I guess I would probably go for that just because it's simpler. Though I just suggested another alternative which would be to use the minor GC count, if that makes sense. Curious what both you @XrXr and Jean think?

byroot:
> this counter only gets incremented when we call from the interpreter into YJIT. It doesn't track JIT-to-JIT calls. 

Is this a fact of life or a shortcut we're taking?

What I had in mind was a two-step:

  - 0 to call-threshold -> increment and use interpreter
  - call-threshold reached -> compile but keep incrementing
  - other threshold reached -> no longer compile cold branches and stop incrementing.

> Incrementing it would also make copy-on-write perf worse.

Not if we stop incrementing it once we reached the second threshold.

maximecb:
> Is this a fact of life or a shortcut we're taking?

The original purpose of that counter was to decide when YJIT would start compiling, rather than to keep counting once YJIT is running code.

> Not if we stop incrementing it once we reached the second threshold.

This is not a bad idea, however it would require code patching. It would mean we have one or more instructions to do the incrementation whenever we enter a function in YJIT, and later we need to come back and overwrite those instructions with no-ops. The no-ops are faster than incrementing but they are not completely free because they also take space in the instruction cache.

There's also another mild issue which is that I would also like to change the way we do the "cold threshold". This also requires some kind of proxy for time, but this has to be some kind of a global value, it can't be local to the function.

So again this kind of comes back to, could we use the minor GC count?

Otherwise, is there a way to know from the Ruby side how many "requests" we have served. Obviously we can't use counters from prod here, but is there some way we could detect and count how many times we are called to serve a request somehow, from the Ruby side? In JavaScript, the browser calls your JS code in an event loop, so you could count how many times you are entering said event loop, how many events you've handled. May not be possible because I assume Rails will just block and wait for network traffic?

Otherwise we could just use monotonic time as Alan suggested. It has flaws, but it's simple. It's probably that or something like minor GC count?

byroot:
> however it would require code patching. It would mean we have one or more instructions to do the incrementation whenever we enter a function in YJIT,

That's where I'm a bit at the limit of my comprehension of YJIT plumbing. Right now this counter is incremented on method dispatch right? So when YJIT code call some other YJIT code, it bypass that `jit_compile` function? But if so, how come I had to do that PR to stop the incrementation?

> could we use the minor GC count?

The downside here is that code might be more or less allocation heavy, hence may trigger GC more or less, so not sure it's better than CPU time.

> how many times we are called to serve a request somehow, from the Ruby side?

Nope. Ruby isn't architectured around a specific lifetime like in-browser JS or PHP. So there's no such concept baked into the language.

k0kubun:
> But if so, how come I had to do that PR to stop the incrementation?

When the interpreter calls an ISEQ so many times that it hits the threshold, YJIT compiles that callee ISEQ as an entry to JIT code. It converts the interpreter -> interpreter call into interpreter -> JIT call. It compiles the callee ISEQ but not the caller ISEQ. So it would keep incrementing the counter for the callee ISEQ (if we didn't merge your PR) until the caller ISEQ also hits the threshold and gets compiled. Does that make sense?

byroot:
> Does that make sense?

I think?

So Interpreter -> JIT calls still go thought that routine, but JIT -> JIT calls don't. That's what I understood.

maximecb:
> Nope. Ruby isn't architectured around a specific lifetime like in-browser JS or PHP. So there's no such concept baked into the language.

Yeah that's what I thought. In theory we could still count the number of times Ruby reads from a TCP socket or something (unblocking after the read call), but the problem there is that it's really not a general solution, we'd be hacking specifically to detect Rails usage patterns.

> The downside here is that code might be more or less allocation heavy, hence may trigger GC more or less, so not sure it's better than CPU time.

Yeah there really is no perfect solution here 😅 My best solution so far is to just set the timeout to something very generous. It will help cut the long tail of cold code being compiled.

k0kubun:
I think we're not actively considering this idea, so let me close this for now.

PR #12120, opened 2024-11-19
Configure ruby/mmtk sync

eightbitraptor:
I think that after discussing some proposed file layout changes with @peterzhu2118 these are the changes to the sync script that will be needed to sync `ruby/mmtk` into `ruby/ruby` as per https://bugs.ruby-lang.org/issues/20860

Discussed in the November dev meeting recorded here: https://github.com/ruby/dev-meeting-log/blob/master/2024/DevMeeting-2024-11-07.md#feature-20860-merge-optional-experimental-feature-mmtk-into-ruby-peterzhu2118

PR #12149, opened 2024-11-22
GC API changes required to use MMTk

eightbitraptor:
This PR contains the GC API and Ruby build system changes proposed in https://github.com/ruby/ruby/pull/11979

Notably:

* We're now using `extconf` to build shared GC modules, and each shared GC module resides in an isolated directory (eg. `gc/default`, `gc/mmtk` etc), with only shared code being directly in `gc/`. This makes building new GC's easier. And allows us to keep implementation specific build configuration out of Ruby's build system.
* All parts of the GC API that are used by GC's other than Ruby's default GC are now hidden behind a pre-processor flag `USE_SHARED_GC`. This means that when Ruby is built without the `--with-shared-gc` flag, none of this code will be included in the final binaries.

https://github.com/ruby/ruby/pull/11979 also contained a concrete GC implementation and binding layer to MMTk. This code has been extracted into the [`ruby/mmtk`](https://github.com/ruby/mmtk) repository, and included in this tree via the default gem syncing mechanism. This work was merged in https://github.com/ruby/ruby/pull/12120

This PR supersedes https://github.com/ruby/ruby/pull/11979 and when merged, will fully implement  [[Feature 20860]](https://bugs.ruby-lang.org/issues/20860)


gc.c
```diff
@@ -179,10 +179,67 @@ rb_gc_vm_barrier(void)
...
+
+#ifdef RB_THREAD_LOCAL_SPECIFIER
+# ifdef __APPLE__
+    rb_current_ec_set(context->ec);
+# else
+    ruby_current_ec = context->ec;
+# endif
+#else
+    native_tls_set(ruby_current_ec_key, context->ec);
+#endif
```
nobu, line 211: We should define `rb_current_ec_set` in all cases, instead of copying.

PR #12151, opened 2024-11-22
[DOC] Fix grammar errors, typos, and improve readability of string.rb

st0012:
Because this file's comments have one extra space after the `#` sign, almost every line is updated. IMO, it's better to address this issue in one go.

To ignore the whitespace changes, please use [this link](https://github.com/ruby/ruby/pull/12151/files?diff=split&w=1) for review.

### DIFF
--- a/string.rb
+++ b/string.rb
@@ -166,7 +166,7 @@
 # These instance methods utilize slicing:
 #
 # - String#[] (aliased as String#slice): Returns a slice copied from +self+.
-# - String#[]=: Mutates +self+ with the slice replaced and returns the given slice.
+# - String#[]=: Mutates +self+ with the slice replaced.
 # - String#slice!: Mutates +self+ with the slice removed and returns the removed slice.
 #
 # Each of the above methods takes arguments that determine the slice
--- a/string.rb
+++ b/string.rb
@@ -1,556 +1,556 @@
-#  A +String+ object has an arbitrary sequence of bytes,
-#  typically representing text or binary data.
-#  A +String+ object may be created using String::new or as literals.
+# A +String+ object has an arbitrary sequence of bytes,
+# typically representing text or binary data.
+# A +String+ object may be created using String::new or as literals.
 #
-#  String objects differ from Symbol objects in that Symbol objects are
-#  designed to be used as identifiers, instead of text or data.
+# String objects differ from Symbol objects in that Symbol objects are
+# designed to be used as identifiers, instead of text or data.
 #
-#  You can create a +String+ object explicitly with:
+# You can create a +String+ object explicitly with:
 #
-#  - A {string literal}[rdoc-ref:syntax/literals.rdoc@String+Literals].
-#  - A {heredoc literal}[rdoc-ref:syntax/literals.rdoc@Here+Document+Literals].
+# - A {string literal}[rdoc-ref:syntax/literals.rdoc@String+Literals].
+# - A {heredoc literal}[rdoc-ref:syntax/literals.rdoc@Here+Document+Literals].
 #
-#  You can convert certain objects to Strings with:
+# You can convert certain objects to Strings with:
 #
-#  - \Method #String.
+# - \Method #String.
 #
-#  Some +String+ methods modify +self+.
-#  Typically, a method whose name ends with <tt>!</tt> modifies +self+
-#  and returns +self+;
-#  often a similarly named method (without the <tt>!</tt>)
-#  returns a new string.
+# Some +String+ methods modify +self+.
+# Typically, a method whose name ends with <tt>!</tt> modifies +self+
+# and returns +self+;
+# often, a similarly named method (without the <tt>!</tt>)
+# returns a new string.
 #
-#  In general, if there exist both bang and non-bang version of method,
-#  the bang! mutates and the non-bang! does not.
-#  However, a method without a bang can also mutate, such as String#replace.
+# In general, if both bang and non-bang versions of a method exist,
+# the bang method mutates and the non-bang method does not.
+# However, a method without a bang can also mutate, such as String#replace.
 #
-#  == Substitution Methods
+# == Substitution Methods
 #
-#  These methods perform substitutions:
+# These methods perform substitutions:
 #
-#  - String#sub: One substitution (or none); returns a new string.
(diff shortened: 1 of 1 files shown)


string.rb
```diff
@@ -1,556 +1,556 @@
...
+# - String#gsub: Zero or more substitutions; returns a new string.
+# - String#gsub!: Zero or more substitutions; returns +self+ if any changes,
+#   +nil+ otherwise.
 #
-#  Each of these methods takes:
+# Each of these methods takes:
 #
-#  - A first argument, +pattern+ (string or regexp),
-#    that specifies the substring(s) to be replaced.
+# - A first argument, +pattern+ (String or Regexp),
```
st0012, line 40: Capitalizing changes like this will make them link to the relevant classes, which is an improvement IMO.

string.rb
```diff
@@ -1,556 +1,556 @@
...
-#  == What's Here
+# == What's Here
 #
-#  First, what's elsewhere. \Class +String+:
+# First, what's elsewhere. \Class +String+:
 #
-#  - Inherits from {class Object}[rdoc-ref:Object@What-27s+Here].
-#  - Includes {module Comparable}[rdoc-ref:Comparable@What-27s+Here].
+# - Inherits the {Object class}[rdoc-ref:Object@What-27s+Here].
+# - Includes the {Comparable module}[rdoc-ref:Comparable@What-27s+Here].
```
st0012, line 306: FWIW, if in the future we can enable the [`--embed-mixins`](https://github.com/ruby/rdoc/pull/842) flag added by @flavorjones in `ruby/ruby`, these methods will be added to the `String` class's doc directly.

string.rb
```diff
@@ -1,556 +1,556 @@
...
+# - #delete: Returns a copy of +self+ with characters removed.
+# - #delete_prefix: Returns a copy of +self+ with a given prefix removed.
+# - #delete_suffix: Returns a copy of +self+ with a given suffix removed.
+# - #lstrip: Returns a copy of +self+ with leading whitespace removed.
+# - #rstrip: Returns a copy of +self+ with trailing whitespace removed.
+# - #strip: Returns a copy of +self+ with leading and trailing whitespace removed.
+# - #chomp: Returns a copy of +self+ with a trailing record separator removed, if found.
+# - #chop: Returns a copy of +self+ with trailing newline characters or the last character removed.
+# - #squeeze: Returns a copy of +self+ with contiguous duplicate characters removed.
+# - #[] (aliased as #slice): Returns a substring determined by a given index, start/length, range, regexp, or string.
```
st0012, line 495: I think it takes `regexp` too so I just added it:

```
irb(main):001> s = "foobar"
=> "foobar"
irb(main):002> s[/foo/]
=> "foo"
```

st0012:
@jeremyevans @peterzhu2118 would you mind giving this a look? Thx

string.rb
```diff
@@ -1,556 +1,556 @@
...
+# A _slice_ of a string is a substring selected by certain criteria.
 #
-#  These instance methods make use of slicing:
+# These instance methods utilize slicing:
 #
-#  - String#[] (aliased as String#slice): returns a slice copied from +self+.
-#  - String#[]=: returns a copy of +self+ with a slice replaced.
-#  - String#slice!: returns +self+ with a slice removed.
+# - String#[] (aliased as String#slice): Returns a slice copied from +self+.
+# - String#[]=: Returns a copy of +self+ with a slice replaced.
```
jeremyevans, line 169: This was incorrect before and is still incorrect, just better formatted.  `String#[]=` mutates the receiver:

```ruby
s = "s"
s[0] = "t"
s
# => "t"
```

As aset assignment in Ruby returns RHS regardless of what the aset method returns, having the method return a modified copy and not mutating the receiver would not make sense.

string.rb
```diff
@@ -1,556 +1,556 @@
...
 #
-#  == What's Here
+# == What's Here
 #
-#  First, what's elsewhere. \Class +String+:
+# First, what's elsewhere. \Class +String+:
 #
-#  - Inherits from {class Object}[rdoc-ref:Object@What-27s+Here].
-#  - Includes {module Comparable}[rdoc-ref:Comparable@What-27s+Here].
+# - Inherits the {Object class}[rdoc-ref:Object@What-27s+Here].
```
jeremyevans, line 305:
```suggestion
# - Inherits from the {Object class}[rdoc-ref:Object@What-27s+Here].
```

I believe that you inherit methods/constants from a class, not that you inherit the class itself.  Inheriting from an class represents an "is a" relationship, will inheriting an object itself represents a "has a" relationship.

string.rb
```diff
@@ -1,556 +1,556 @@
...
+# === Methods for Converting to Non-+String+
+#
+# Each of these methods converts the contents of +self+ to a non-+String+.
+#
+# <em>Characters, Bytes, and Clusters</em>
+#
+# - #bytes: Returns an array of the bytes in +self+.
+# - #chars: Returns an array of the characters in +self+.
+# - #codepoints: Returns an array of the integer ordinals in +self+.
+# - #getbyte: Returns an integer byte determined by a given index.
```
jeremyevans, line 513:
```suggestion
# - #getbyte: Returns the integer byte at the given index in +self+.
```

jeremyevans (approved):
Looks good.  I do have some suggestions for minor improvements, but I don't need to review again before merging.

string.rb
```diff
@@ -1,556 +1,556 @@
...
+# A _slice_ of a string is a substring selected by certain criteria.
 #
-#  These instance methods make use of slicing:
+# These instance methods utilize slicing:
 #
-#  - String#[] (aliased as String#slice): returns a slice copied from +self+.
-#  - String#[]=: returns a copy of +self+ with a slice replaced.
-#  - String#slice!: returns +self+ with a slice removed.
+# - String#[] (aliased as String#slice): Returns a slice copied from +self+.
+# - String#[]=: Mutates +self+ with the slice replaced and returns the given slice.
```
jeremyevans, line 169:
```suggestion
# - String#[]=: Mutates +self+ with the slice replaced.
```

I don't think we should discuss return values of assignment methods, since when used in assignment, return values are ignored and the RHS is returned.

jeremyevans (approved):
Looks good.  Only one change I would make.

PR #12152, opened 2024-11-23
Remove Cargo.lock files

nobu:
Libraries should not lock.


XrXr:
We should keep the lock files for YJIT. bindgen is a binary, not a library, and YJIT proper is effectively a binary, too. The lock files have been important to keep consistent between platforms.

maximecb:
Yeah seems like removing the lock files increases risks of failed builds and random bugs. We've tested with specific library versions. We can't guarantee that things will work and work well with any other versions our dependencies.

k0kubun:
> Libraries should not lock.

Could you elaborate more on why? I don't know much about mspec and mmtk, but for YJIT, YJIT CI jobs like `--enable-yjit=dev` and bindgen may suddenly fail if you don't lock dependencies.

Updates to these YJIT lock files have been fairly rare too. I don't see any motivation to remove them.
https://github.com/ruby/ruby/commits/master/yjit/Cargo.lock
https://github.com/ruby/ruby/commits/master/yjit/bindgen/Cargo.lock

hsbt:
We keep these lockfile for reproducible tests. 

BTW, some gemfiles and lockfiles like `tool/bundler/vendor_gems.rb{,lock}` is unnecessary on ruby/ruby repo. I will remove them.

PR #12157, opened 2024-11-23
[DOC] Improve maintainers list

st0012:
1. Add myself and @colby-swandale to the RDoc's maintainers list.
2. Add new lines after headers so the source file is more readable.
3. Escape `Module` in the header so it doesn't link to the `Module` class.
4. (Optional) I assume the handles after each maintainer's name are GitHub handles. If that's the case, I think we can make them up-to-date (few are not), and then link them to the actual GH profiles.
    - I only updated the handles that I personally known or can be verified through git history.


jeremyevans (approved):

kou (approved):

nobu:
> 4. (Optional) I assume the handles after each maintainer's name are GitHub handles. If that's the case, I think we can make them up-to-date (few are not), and then link them to the actual GH profiles.

Good job.

And there are same links of course, why not using footnotes?
    
>    * I only updated the handles that I personally known or can be verified through git history.

Another I could find was the wasm maintainer, [katei](https://github.com/kateinoigakukun).

As for the other three, (keiju, suke, kanemoto), I suspect they might not be using github accounts (for ruby at least).

st0012:
@nobu Good call 👍 Updated

doc/maintainers.md
```diff
@@ -27,480 +29,636 @@ consensus on ruby-core/ruby-dev before changing/adding. Some of submaintainers
...
+[hasumikin]: https://github.com/hasumikin
+[hsbt]: https://github.com/hsbt
+[ima1zumi]: https://github.com/ima1zumi
+[jeremyevans]: https://github.com/jeremyevans
+[k-tsj]: https://github.com/k-tsj
+[k0kubun]: https://github.com/k0kubun
+[kanemoto]: https://github.com/kanemoto
+[kateinoigakukun]: https://github.com/kateinoigakukun
+[kddnewton]: https://github.com/kddnewton
+[keiju]: https://github.com/keiju
```
nobu, line 645: You don't use this link.

doc/maintainers.md
```diff
@@ -27,480 +29,636 @@ consensus on ruby-core/ruby-dev before changing/adding. Some of submaintainers
...
+[drbrain]: https://github.com/drbrain
+[eban]: https://github.com/eban
+[eileencodes]: https://github.com/eileencodes
+[hasumikin]: https://github.com/hasumikin
+[hsbt]: https://github.com/hsbt
+[ima1zumi]: https://github.com/ima1zumi
+[jeremyevans]: https://github.com/jeremyevans
+[k-tsj]: https://github.com/k-tsj
+[k0kubun]: https://github.com/k0kubun
+[kanemoto]: https://github.com/kanemoto
```
nobu, line 642: You don't use this link.

kou:
suke's GitHub account is https://github.com/suketa .

st0012:
@kou Thanks, I've updated the handle with a link.

doc/maintainers.md
nobu: Are these updated?

doc/maintainers.md
```diff
@@ -27,480 +29,635 @@ consensus on ruby-core/ruby-dev before changing/adding. Some of submaintainers
...
-* Eric Hodel (drbrain)
-* Hiroshi SHIBATA (hsbt)
+
+* Eric Hodel ([drbrain])
+* Hiroshi SHIBATA ([hsbt])
 * https://github.com/rubygems/rubygems
 
 #### lib/unicode_normalize.rb, lib/unicode_normalize/*
+
 * Martin J. Dürst
```
nobu, line 59:
```suggestion
* Martin J. Dürst ([duerst])
```
and
```
[duerst]: https://github.com/duerst
```

doc/maintainers.md
st0012: Yeah I think it's removed?

doc/maintainers.md
st0012: Added 👍

peterzhu2118 (approved):

PR #12161, opened 2024-11-25
Fix initialization of `struct wait_fast_fallback_arg::cancelled`

nobu:
Fix up of #12156

PR #12162, opened 2024-11-25
Avoid SEGV with test-bundled-gems-drb

shioimm:
To fix https://github.com/ruby/ruby/actions/runs/12003690124/job/33457535316


shioimm:
fix with https://github.com/ruby/ruby/pull/12161

PR #12164, opened 2024-11-25
error.c: call `va_end` before jumping

byroot:
The man page is clear that every `va_start` call MUST be succeeded by the corresponding `va_end` call.

So `rb_raise` can't call `rb_exc_raise` before `va_end`, otherwise `va_end` is never called.

NB: I'm not sure how bad this really is, as I'm unfamiliar with the various `va_end` implementations. From reading online it seems this could cause memory leaks on some platforms, but I assume it's not the case on the most common ones.

### DIFF
--- a/error.c
+++ b/error.c
@@ -3632,12 +3632,13 @@ rb_vraise(VALUE exc, const char *fmt, va_list ap)
 }
 
 void
-rb_raise(VALUE exc, const char *fmt, ...)
+rb_raise(VALUE exc_class, const char *fmt, ...)
 {
     va_list args;
     va_start(args, fmt);
-    rb_vraise(exc, fmt, args);
+    VALUE exc = rb_exc_new3(exc_class, rb_vsprintf(fmt, args));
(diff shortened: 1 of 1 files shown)


XrXr (commented):
Patch is fine because it's essentially a rescheduling of rb_vraise()

```c
void
rb_vraise(VALUE exc, const char *fmt, va_list ap)
{
    rb_exc_raise(rb_exc_new3(exc, rb_vsprintf(fmt, ap)));
}
```

> I'm not sure how bad this really is ...

For efficiency, I think all the sensible platforms implement vargs in a way that doesn't use malloc, so the current code probably don't leak anywhere. Technically, even with the patch there are cases va_end is unreachable through e.g. NoMemoryError and the `initialize` of the error class raising, but it's not addressing an immediate issue in any case.

byroot:
> I think all the sensible platforms implement vargs in a way that doesn't use malloc

I figured as much.

PR #12175, opened 2024-11-26
YJIT: Implement opt_reverse insn

rwstauner:
I noticed that with the Prism parser YJIT's side exit count increased because Prism creates an `opt_reverse` insn that `parse.y` does not.


maximecb (approved):
Good job Randy 👍

PR #12178, opened 2024-11-27
Move `Kernel#with_yjit` to `Module#with_yjit`

nobu:
`Kernel#with_yjit` seems used only in `class`es, not the top level.
And `Kernel.with_yjit` is not used at all but still defined.


k0kubun (approved):
Makes sense :+1:

PR #12179, opened 2024-11-27
Improve test-bundled-gems

hsbt:
We can test only csv and drb with `make test-bundled-gems BUNDLED_GEMS=csv,drb`.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -1646,8 +1646,9 @@ test-bundled-gems-spec: $(TEST_RUNNABLE)-test-bundled-gems-spec
 yes-test-bundled-gems-spec: yes-test-spec-precheck $(PREPARE_BUNDLED_GEMS)
 	$(ACTIONS_GROUP)
 	$(gnumake_recursive)$(Q) \
-	BUNDLED_GEMS=$(BUNDLED_GEMS) $(RUNRUBY) -r./$(arch)-fake -r$(tooldir)/lib/_tmpdir \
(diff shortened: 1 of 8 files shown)


common.mk
```diff
@@ -1646,7 +1646,7 @@ test-bundled-gems-spec: $(TEST_RUNNABLE)-test-bundled-gems-spec
 yes-test-bundled-gems-spec: yes-test-spec-precheck $(PREPARE_BUNDLED_GEMS)
 	$(ACTIONS_GROUP)
 	$(gnumake_recursive)$(Q) \
-	$(RUNRUBY) -r./$(arch)-fake -r$(tooldir)/lib/_tmpdir \
+	BUNDLED_GEMS=$(BUNDLED_GEMS) $(RUNRUBY) -r./$(arch)-fake -r$(tooldir)/lib/_tmpdir \
```
nobu, line 1649: `evar=value command` works only in `sh`, but not in `cmd.exe`.

spec/bundled_gems.mspec
```diff
@@ -1,6 +1,13 @@
 load File.dirname(__FILE__) + '/default.mspec'
 
 class MSpecScript
-  set :library, get(:stdlibs).to_a & get(:bundled_gems).to_a
+  test_bundled_gems = get(:stdlibs).to_a & get(:bundled_gems).to_a
+  unless ENV["BUNDLED_GEMS"].empty?
```
nobu, line 5:
```suggestion
  if ENV["BUNDLED_GEMS"]&.then {!_1.empty?}
```

PR #12182, opened 2024-11-27
Fix C level backtrace info on macOS clang 16

peterzhu2118:
macOS clang 16 generates DWARF5, which have Mach-O section names that are limited to 16 characters, which causes sections with long names to be truncated and not match above.
    
See: https://wiki.dwarfstd.org/Best_Practices.md#Mach-2d-O

### DIFF
--- a/addr2line.c
+++ b/addr2line.c
@@ -2426,7 +2426,17 @@ found_mach_header:
                         for (int j=0; j < DWARF_SECTION_COUNT; j++) {
                             struct dwarf_section *s = obj_dwarf_section_at(obj, j);
 
-                            if (strcmp(sect->sectname, debug_section_names[j]) != 0)
+                            if (strcmp(sect->sectname, debug_section_names[j]) != 0
+#ifdef __APPLE__
+                                    /* macOS clang 16 generates DWARF5, which have Mach-O
+                                     * section names that are limited to 16 characters,
+                                     * which causes sections with long names to be truncated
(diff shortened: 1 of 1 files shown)


XrXr (commented):
Looks like the 16 char limit has been there since time immemorial, so it's something else that changed in macOS 15 that caused this issue. Maybe the section names themselves got longer. Can you rephrase your commit message?

https://github.com/llvm/llvm-project/blame/f67ba5855278401728413431216dda5d370ac2e0/llvm/include/llvm/BinaryFormat/MachO.h#L581
https://wiki.dwarfstd.org/Best_Practices.md

peterzhu2118:
But it doesn't seem like it was an issue before macOS 15? So I think it wasn't truncated before and it's truncated since macOS 15.

peterzhu2118:
Most of the fields work fine, it crashes on `__debug_str_offsets` because it's longer than 16 characters. [`__debug_str_offsets` field was introduced 2 years ago](https://github.com/ruby/ruby/commit/303e11697dab49a3522712cdc0b4d34af1dc6bfc), so it's nothing new.

XrXr:
Ah __debug_str_offsets is from DWARF 5, so maybe it's because the newer toolchain just started generating DWARF 5 on macOS. I'll try building with dwarf 4 on macOS 15 to see if that also evades the bug.

XrXr (approved):

PR #6094, opened 2022-07-06
Expand tabs

k0kubun:
[Misc #18891]

<details>

<summary>The code I used to generate this diff</summary>

Because ruby-commit-hook's bin/auto-style.rb knows what kind of files should be skipped, I wrote this code based off of auto-style.rb.

```rb
DEFAULT_GEM_LIBS = %w[
  bundler
  cmath
  csv
  e2mmap
  fileutils
  forwardable
  ipaddr
  irb
  logger
  matrix
  mutex_m
  ostruct
  prime
  rdoc
  rexml
  rss
  scanf
  shell
  sync
  thwait
  tracer
  webrick
]

DEFAULT_GEM_EXTS = %w[
  bigdecimal
  date
  dbm
  etc
  fcntl
  fiddle
  gdbm
  io/console
  json
  openssl
  psych
  racc
  sdbm
  stringio
  strscan
  zlib
]

IGNORED_FILES = [
  # default gems whose master is GitHub
  %r{\Abin/(?!erb)\w+\z},
  *(DEFAULT_GEM_LIBS + DEFAULT_GEM_EXTS).flat_map { |lib|
    [
      %r{\Alib/#{lib}/},
      %r{\Alib/#{lib}\.gemspec\z},
      %r{\Alib/#{lib}\.rb\z},
      %r{\Atest/#{lib}/},
    ]
  },
  *DEFAULT_GEM_EXTS.flat_map { |ext|
    [
      %r{\Aext/#{ext}/},
      %r{\Atest/#{ext}/},
    ]
  },

  # vendoring (ccan)
  %r{\Accan/},

  # vendoring (nkf)
  %r{\Aext/nkf/nkf-utf8/},

  # vendoring (onigmo)
  %r{\Aenc/},
  %r{\Ainclude/ruby/onigmo\.h\z},
  %r{\Areg.+\.(c|h)\z},

  # explicit or implicit `c-file-style: "linux"`
  %r{\Aaddr2line\.c\z},
  %r{\Amissing/},
  %r{\Astrftime\.c\z},
  %r{\Avsnprintf\.c\z},

  # to respect the original statements of licenses
  %r{\ALEGAL\z},
]

paths = `git ls-files`.split(/\n/)
paths.select! do |l|
  /^\d/ !~ l and /\.bat\z/ !~ l and
  (/\A(?:config|[Mm]akefile|GNUmakefile|README)/ =~ File.basename(l) or
[...]

PR #12184, opened 2024-11-27
Expand tabs in addr2line.c

peterzhu2118:
.git-blame-ignore-revs will be updated after this commit is merged.

### DIFF
--- a/addr2line.c
+++ b/addr2line.c
@@ -56,11 +56,11 @@
 #  ifdef _AIX
 #pragma alloca
 #  else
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
✅

We previously didn't do this at https://github.com/ruby/ruby/pull/6094 because the file used to have the imported code of `kprintf`, but it's no longer the case today https://github.com/ruby/ruby/commit/15b2e912300f43fb6fdf9f663d0979efb464efba.

PR #12185, opened 2024-11-27
Fix global-buffer-overflow when outputting C backtrace

peterzhu2118:
fill_lines is passed -1 for offset, which causes it to read the -1 index of traces. This is not valid memory as -1 is reading before the trace global variable in rb_print_backtrace. This code comes from commit 99d1f5f88b9b6de3124e31019902f91e030d334f, where there used to be special handling for the -1 index.

We can see this error in ASAN:

    ==71037==ERROR: AddressSanitizer: global-buffer-overflow on address 0x00010157abf8 at pc 0x00010116f3b8 bp 0x00016f92c3b0 sp 0x00016f92c3a8
    READ of size 8 at 0x00010157abf8 thread T0
        #0 0x10116f3b4 in debug_info_read addr2line.c:1945
        #1 0x10116cc90 in fill_lines addr2line.c:2497
        #2 0x101169dbc in rb_dump_backtrace_with_lines addr2line.c:2635
        #3 0x100e56788 in rb_print_backtrace vm_dump.c:825
        #4 0x100e56db4 in rb_vm_bugreport vm_dump.c:1155
        #5 0x100734dc4 in rb_bug_without_die error.c:1085
        #6 0x100734ae4 in rb_bug error.c:109

### DIFF
--- a/addr2line.c
+++ b/addr2line.c
@@ -2632,7 +2632,7 @@ rb_dump_backtrace_with_lines(int num_traces, void **traces, FILE *errout)
 	    memcpy(main_path, binary_filename, len+1);
 	    append_obj(&obj);
 	    obj->path = main_path;
-	    addr = fill_lines(num_traces, traces, 1, &obj, lines, -1, errout);
+	    addr = fill_lines(num_traces, traces, 1, &obj, lines, 0, errout);
 	    if (addr != (uintptr_t)-1) {
 		dladdr_fbases[0] = (void *)addr;
 	    }


composerinteralia:
@peterzhu2118 I've just noticed that after including this commit at GitHub we lost our C level backtrace information for crashes. I think the [special -1 case](https://github.com/ruby/ruby/blob/9aed9e94867654b94d88be141026110d1f247bb5/addr2line.c#L2178-L2209) was somehow working for us, although I don't understand this code particularly well yet. (GitHub runs Ruby via a script that puts our custom build into the path and then calls `exec ruby`. That seems to have something to do with it. I can get the C level backtrace information again by running Ruby directly or exec-ing with the full path.)

It looks like the `#ifdef USE_ELF` version of `fill_lines` should already be safe from ASAN errors without this change, since it sets offset back to 0 if it is -1. Do you think it would make sense to switch this back to -1, at least for `USE_ELF`? (Or, alternatively, add an `if (offset == -1) offset = 0` to the non ELF version of `fill_lines`.)

I can open something on the bug tracker if this seems worth exploring further. Thank you!

PR #12189, opened 2024-11-28
Avoid uninitialized variable access

mame:
With parse.y, the following pattern caused an access to an uninitialized variable `key_val`.

```
C = { **{ a: 1 } }
```

Just for the case, `value_val` is also initialized.


XrXr:
Looks similar to https://github.com/ruby/ruby/pull/12275. Not sure which is better between Qnil and Qfalse.

mame:
It is already fixed.

PR #12190, opened 2024-11-28
Avoid accessing to a local variable that is out of scope

mame:
`PUSH_SEQ(ret, cond_seq)` stores to `ret` a pointer to `cond_seq`. After `cond_seq` is out of scope, `PUSH_INSNL(ret, ...)` attempts to dereference of the pointer.

This change moves `DECL_ANCHOR(cond_seq)` to the toplevel.


mame:
Sorry, @nobu pointed out that they are false positives.

PR #12191, opened 2024-11-28
Avoid accessing to a local variable that is out of scope

mame:
The same issue as https://github.com/ruby/ruby/pull/12190


mame:
Sorry, @nobu pointed out that they are false positives.

PR #12204, opened 2024-11-28
RUBY_DEBUG: Verify PC correctness every GC allocation

jhawthorn:
In RUBY_DEBUG mode this adds a check that on every GC allocation that cfp->pc is within the bounds of the current iseq.

This aims to make a common problem (particularly with YJIT) more obvious and easier to catch. This is essentially the code I used to reproduce the bug fixed in #12202 (which I had to narrow down from a failure on GitHub's CI)

Here's a build with this and #12202 reverted where we can see how this will reliably fail: https://github.com/jhawthorn/ruby/actions/runs/12075928651/job/33676531616

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -969,10 +969,24 @@ rb_gc_obj_slot_size(VALUE obj)
     return rb_gc_impl_obj_slot_size(obj);
 }
 
+static inline void
+gc_validate_pc(void) {
+#if RUBY_DEBUG
+    rb_execution_context_t *ec = GET_EC();
+    const rb_control_frame_t *cfp = ec->cfp;
+    if (cfp && VM_FRAME_RUBYFRAME_P(cfp) && cfp->pc) {
+        RUBY_ASSERT(cfp->pc >= ISEQ_BODY(cfp->iseq)->iseq_encoded);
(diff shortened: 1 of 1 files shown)


XrXr:
I'm a bit worried about the perf cost here, but it's probably fine to try. Could also revert it if it's too slow.

jhawthorn:
Performance impact, at least from this hacky benchmark, seems negligible or at least in line with the other slowdowns introduced by debug mode.

```
Benchmark 1: ruby --disable-gems -e '10_000_000.times { Object.new }'
  Time (mean ± σ):      1.099 s ±  0.007 s    [User: 1.091 s, System: 0.006 s]
  Range (min … max):    1.090 s …  1.113 s    10 runs

Benchmark 2: ./miniruby_with_check -e '10_000_000.times { Object.new }'
  Time (mean ± σ):      3.348 s ±  0.058 s    [User: 3.333 s, System: 0.011 s]
  Range (min … max):    3.230 s …  3.428 s    10 runs

Benchmark 3: ./miniruby_without_check -e '10_000_000.times { Object.new }'
  Time (mean ± σ):      3.362 s ±  0.046 s    [User: 3.346 s, System: 0.011 s]
  Range (min … max):    3.296 s …  3.442 s    10 runs
```

maximecb:
Wrt performance cost, this only runs in debug builds, not release builds, correct? If so then no need to worry about perf.

PR #12209, opened 2024-11-29
Do not depend on the evaluation order of C arguments

mame:
The evaluation order of C arguments is unspecified. `RSTRING_LEN(str)` would fails if the conversion to a String by `StringValuePtr` is not done yet.


ioquatix:
Great!

PR #12219, opened 2024-11-29
Fix ATTRIBUTE_NO_ADDRESS_SAFETY_ANALYSIS for MSAN

peterzhu2118:
There's no case for when RUBY_MSAN_ENABLED, so the macro ends up doing nothing when it should instead have `__attribute__((__no_sanitize__("memory")))`.

### DIFF
--- a/internal/sanitizers.h
+++ b/internal/sanitizers.h
@@ -39,6 +39,9 @@
 #elif defined(RUBY_ASAN_ENABLED)
 # define ATTRIBUTE_NO_ADDRESS_SAFETY_ANALYSIS(x) \
     __attribute__((__no_sanitize__("address"), __noinline__)) x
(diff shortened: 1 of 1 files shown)


internal/sanitizers.h
```diff
@@ -32,8 +32,7 @@
 #include "ruby/internal/stdbool.h"     /* for bool */
 #include "ruby/ruby.h"          /* for VALUE */
 
-#if 0
-#elif defined(RUBY_ASAN_ENABLED) && defined(RUBY_MSAN_ENABLED)
+#if defined(RUBY_ASAN_ENABLED) && defined(RUBY_MSAN_ENABLED)
```
nobu, line 35: It was to align the conditions.

peterzhu2118: I dropped that change.

PR #12222, opened 2024-11-30
[win32] fix arm64 instruction decoding

jeremyd2019:
Two minor fixes to arm64 instruction decoding when looking for __pioinfo:
1. add_mask was shifted by one bit, it was intended to be 0x7f800000.  However, since the mask was already excluding matching the 'sh' bit, and since the purpose of the add following the adrp is to add in the lower 12 bits, I opted to set the mask to 0x7fc00000 and simply remove the handling for the 12 bit shift option which is now required to be disabled in order to match.
2. adrp's immediate was supposed to be sign extended.  So far, I have not seen cases where the global variable ends up before the code in memory, but it's a possibility, so handle the sign extension.

See the recent commentary on #8995 between me and @pbo-linaro

### DIFF
--- a/win32/win32.c
+++ b/win32/win32.c
@@ -2580,7 +2580,7 @@ set_pioinfo_extra(void)
     const uint32_t adrp_id = 0x90000000;
     const uint32_t adrp_mask = 0x9f000000;
     const uint32_t add_id = 0x11000000;
-    const uint32_t add_mask = 0x3fc00000;
+    const uint32_t add_mask = 0x7fc00000;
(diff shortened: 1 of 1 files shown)


jeremyd2019:
As an aside, I might be able to contribute code to make this work on i686/x86_64 emulation on arm64, but it would require this code to be built unconditionally and detection of running on arm64 hosts be done at runtime.

pbo-linaro:
I reviewed this change and confirm it's correct.
Thanks for the fix @jeremyd2019 :+1: .

PR #12225, opened 2024-11-30
Update mswin build

nobu:


MSP-Greg:
@nobu

Good day.  This (or the last few Win32 commits) seems to have broken ruby-loco's mswin build.  I could also verify locally.  Both are using `--with-opt-dir=C:/vcpkg/installed/x64-windows`.

Is there anything I can check?

JFYI, comments in GH commits are 'quirky' lately....

Thanks,  Greg

PR #12227, opened 2024-11-30
[DOC] New homepage for docs.ruby-lang.org/en

st0012:
The homepage of a language's documentation website should provide developers with quick access to important components, topics, and resources. Here are some good examples:

- [Ruby's Japanese Documentation](https://docs.ruby-lang.org/ja/master/doc/index.html)
- [Python Documentation](https://docs.python.org/)
- [Go Documentation](https://go.dev/doc/)
- [TypeScript Documentation](https://www.typescriptlang.org/docs/)
- [Ruby API](https://rubyapi.org/)

Currently, [docs.ruby-lang.org/en](https://docs.ruby-lang.org/en) uses the project README as its homepage, which offers less value to most users. I propose creating a dedicated index page to serve as the homepage for [docs.ruby-lang.org/en](https://docs.ruby-lang.org/en). This would improve the user experience by providing easy navigation to essential documentation topics.

### Result

Screenshot 2024-11-30 at 18 32 31

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -20,10 +20,8 @@ Explore the essential classes and modules:
 - [File](File.html) - \File operations and handling.
 - [IO](IO.html) - Input/output functionality.
 - [Time](Time.html) - \Time representation.
-- [Date](Date.html) - \Date representation.
 - [Regexp](Regexp.html) - Regular expressions for pattern matching.
 - [Range](Range.html) - Representing a range of values.
-- [JSON](JSON.html) - \JSON encoding and decoding.
 - [Exception](Exception.html) - Base class for all exceptions.
 - [Thread](Thread.html) - Multithreading and concurrency.
 
@@ -35,9 +33,16 @@ Deep dive into Ruby's syntax and features:
 - [Exceptions](exceptions_md.html)
 - [Implicit Conversions](implicit_conversions_md.html)
 
-## Standard Library
+## Standard Libraries
 
-Access the comprehensive set of libraries included with Ruby:
+There are some standard libraries included in Ruby that are also commonly used, such as:
+
+- [Date](Date.html) - \Date representation.
+- [JSON](JSON.html) - \JSON encoding and decoding.
+- [ERB](ERB.html) - Embedded Ruby for templating.
+- [Net::HTTP](Net/HTTP.html) - HTTP client library.
+
+Use the following links to access the comprehensive set of libraries included with Ruby:
 
 - [Standard Library Documentation](standard_library_rdoc.html)
 - [Maintainers](maintainers_md.html)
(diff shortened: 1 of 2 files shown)


doc/index.md
```diff
@@ -0,0 +1,60 @@
...
+- [Documentation Guide](contributing/documentation_guide_md.html)
+- [Reporting Issues](contributing/reporting_issues_md.html)
+- [Building Ruby](contributing/building_ruby_md.html)
+- [Testing Ruby](contributing/testing_ruby_md.html)
+
+## Additional Resources
+
+- [Ruby Homepage](https://www.ruby-lang.org/)
+- [RubyGems](https://rubygems.org/)
+- [Ruby on Rails](https://rubyonrails.org/)
```
jeremyevans, line 59: I don't think Ruby's documentation should link to external libraries/frameworks.
```suggestion
```

doc/index.md
```diff
@@ -0,0 +1,60 @@
...
+- [Symbol](Symbol.html) - Named identifiers inside the Ruby interpreter.
+- [Array](Array.html) - Ordered collections of objects.
+- [Hash](Hash.html) - Key-value pairs for efficient data retrieval.
+- [Integer](Integer.html) - \Integer number class.
+- [Float](Float.html) - Floating-point number class.
+- [Enumerable](Enumerable.html) - Collection traversal and searching.
+- [File](File.html) - \File operations and handling.
+- [IO](IO.html) - Input/output functionality.
+- [Time](Time.html) - \Time representation.
+- [Date](Date.html) - \Date representation.
```
jeremyevans, line 23:
```suggestion
```

Date is a standard library and not a core class.

doc/index.md
```diff
@@ -0,0 +1,60 @@
...
+- [Regexp](Regexp.html) - Regular expressions for pattern matching.
+- [Range](Range.html) - Representing a range of values.
+- [JSON](JSON.html) - \JSON encoding and decoding.
```
jeremyevans, line 26:
```suggestion
```

JSON is a standard library and not a core class.

jeremyevans (changes requested):

st0012:
@jeremyevans I moved `Date` and `JSON` to the `Standard Libraries` section, along with `ERB` and `Net::HTTP`, which are a few libraries that I think are commonly used. Also, listing that that way can help readers understand the difference between them and the core classes. WDYT?

jeremyevans (approved):

doc/index.md
```diff
@@ -0,0 +1,65 @@
...
+- [Regexp](Regexp.html) - Regular expressions for pattern matching.
+- [Range](Range.html) - Representing a range of values.
+- [Exception](Exception.html) - Base class for all exceptions.
+- [Thread](Thread.html) - Multithreading and concurrency.
+
+## Language Reference
+
+Deep dive into Ruby's syntax and features:
+
+- [Ruby Syntax](syntax_rdoc.html)
```
peterzhu2118, line 32: Can we use `rdoc-ref` instead? E.g.
```suggestion
- [Ruby Syntax](rdoc-ref:syntax.rdoc)
```

doc/index.md
```diff
@@ -0,0 +1,65 @@
...
+- [Exceptions](exceptions_md.html)
+- [Implicit Conversions](implicit_conversions_md.html)
```
peterzhu2118, line 34: This file doesn't exist for me.

```suggestion
- [Implicit Conversion](implicit_conversion_rdoc.html)
```

doc/index.md
st0012: I applied `rdoc-ref` for page links, but when used on classes, they'll change the links' style. So I decided to leave class links as they are now.

doc/index.md
st0012: Good catch. Updated.

peterzhu2118 (approved):

PR #12228, opened 2024-12-01
[DOC] Adjust argument unpacking docs and document **nil

zverok:
Motivation: I looked where to put the new `**nil` functionality (as implemented in [Bug #20064](https://bugs.ruby-lang.org/issues/20064)) and found out that the perspective of the unpacking documentation seems to be slightly off. So I adjusted that a bit and added the `#to_a`/`#to_hash` convention, as well as the new `**nil`.

### DIFF
--- a/doc/syntax/calling_methods.rdoc
+++ b/doc/syntax/calling_methods.rdoc
@@ -300,7 +300,7 @@ gives this result:
   hello main this is block
   place is block
 
-=== Array to Arguments Conversion
+=== Unpacking Positional Arguments
 
 Given the following method:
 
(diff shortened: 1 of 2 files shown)


doc/syntax/calling_methods.rdoc
```diff
@@ -322,19 +322,74 @@ Both are equivalent to:
...
 You may also use the <code>**</code> (described next) to convert a Hash into
 keyword arguments.
 
 If the number of objects in the Array do not match the number of arguments for
 the method, an ArgumentError will be raised.
 
 If the splat operator comes first in the call, parentheses must be used to
-avoid a warning:
+avoid an ambiguity of interpretation as an unpacking operator or multiplication
+operator. In this case, Ruby issues a warning:
```
jeremyevans, line 387:
```suggestion
operator. In this case, Ruby issues a warning in verbose mode:
```

doc/syntax/calling_methods.rdoc
```diff
@@ -356,6 +411,35 @@ Both are equivalent to:
 
   my_method(first: 3, second: 4, third: 5)
 
+The <code>**</code> unpacking operator can be applied to any object, not only
+arrays. If the object responds to a <code>#to_hash</code> method, this method
+is called, and is expected to return an Hash, and elements of this array are passed
+as separate positional arguments:
```
jeremyevans, line 417:
```suggestion
hashes. If the object responds to a <code>#to_hash</code> method, this method
is called, and is expected to return an Hash, and elements of this hash are passed
as keyword arguments:
```

doc/syntax/calling_methods.rdoc
```diff
@@ -356,6 +411,35 @@ Both are equivalent to:
...
+  end
+
+  name = Name.new('Jane Doe')
+  p(**name)
+  # Prints: {name: "Jane", last: "Doe"}
+
+Unlike <code>*</code> operator, <code>**</code> raises an error when used on an
+object that doesn't respond to <code>#to_hash</code>. The one exception is
++nil+, which doesn't explicitly define this method, but is still allowed to
+be used in <code>**</code> unpacking, producing an empty hash of keyword arguments.
```
jeremyevans, line 434:
```suggestion
be used in <code>**</code> unpacking, not adding any keyword arguments.
```

object.c
```diff
@@ -1357,6 +1357,10 @@ rb_obj_frozen_p(VALUE obj)
  * - #to_r
  * - #to_s
  *
+ * While +nil+ doesn't have an explicitly defined #to_hash method,
+ * it can be used in <code>**</code> unpacking, producing an
+ * empty Hash.
```
jeremyevans, line 1362:
```suggestion
 * it can be used in <code>**</code> unpacking, not adding any
 * keyword arguments.
```

jeremyevans (changes requested):
This looks mostly good.  There are some typos that need to be fixed, and an important semantic difference that `**nil` does not produce an empty hash, it is just ignored and does not add keywords (it's been optimized to avoid the allocation of an unnecessary hash).

doc/syntax/calling_methods.rdoc
```diff
@@ -322,19 +322,74 @@ Both are equivalent to:
...
 If the number of objects in the Array do not match the number of arguments for
 the method, an ArgumentError will be raised.
 
 If the splat operator comes first in the call, parentheses must be used to
-avoid a warning:
+avoid an ambiguity of interpretation as an unpacking operator or multiplication
+operator. In this case, Ruby issues a warning:
 
-  my_method *arguments  # warning
+  my_method *arguments  # warning: `*' interpreted as argument prefix
```
nobu, line 389:
```suggestion
  my_method *arguments  # warning: '*' interpreted as argument prefix
```

zverok:
Thank you @jeremyevans @nobu, suggestions applied :pray:

jeremyevans (approved):

PR #12232, opened 2024-12-02
Lock json-schema-5.1.0 and fix build failure iwth prism

hsbt:
We can't use C extension as json-schema dependency.


hsbt:
@kddnewton @tenderlove https://github.com/ruby/ruby/commit/c8dd0479553d4c872b41d667c8fc7a9340594a08 is broken with Windows platform. see https://github.com/ruby/ruby/actions/runs/12129962811/job/33819370301#step:20:278

I fixed that with https://github.com/ruby/ruby/pull/12232/commits/19e67bf4c6beb32b91e748f032796e43c9b48903

kddnewton:
Thanks @hsbt!!

PR #12241, opened 2024-12-03
Re-introduce HAVE_MALLOC_H load guard for malloc.h

eightbitraptor:


gc.c
```diff
@@ -48,7 +48,9 @@
 #  define malloc_usable_size(a) malloc_size(a)
 # endif
 #else
```
peterzhu2118, line 50: I think we should check for `HAVE_MALLOC_H` rather than `HAVE_MALLOC_TRIM`. We only check for `HAVE_MALLOC_TRIM` in `default.c` because we don't need to include `malloc.h` if we don't have `malloc_trim`.

```suggestion
#elif defined(HAVE_MALLOC_H)
# include <malloc.h>
#endif
```

peterzhu2118 (approved):

byroot:
Fixed by @nobu in https://github.com/ruby/ruby/commit/86c01b6aa0833e1bddda6ba8315c9d414aef0f76

PR #12243, opened 2024-12-03
Use BUILDING_SHARED_GC instead of RB_AMALGAMATED_DEFAULT_GC

peterzhu2118:
We can use the BUILDING_SHARED_GC flag to check if we're building gc_impl.h as a shared GC or building the default GC.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -615,7 +615,6 @@ rb_gc_guarded_ptr_val(volatile VALUE *ptr, VALUE val)
 #endif
 
 static const char *obj_type_name(VALUE obj);
-#define RB_AMALGAMATED_DEFAULT_GC
 #include "gc/default/default.c"
 static int external_gc_loaded = FALSE;
 
--- a/gc/gc_impl.h
+++ b/gc/gc_impl.h
@@ -10,6 +10,9 @@
  */
 #include "ruby/ruby.h"
 
+#ifdef BUILDING_SHARED_GC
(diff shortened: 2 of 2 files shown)


XrXr (commented):
This is fine for `!BUILDING_SHARED_GC`, since things will continue to be static, but makes it so that functions are not static when `BUILDING_SHARED_GC`. If you're fine with that then go ahead.

(also, typo in the commit message.)

peterzhu2118:
> makes it so that functions are not static when BUILDING_SHARED_GC

Yeah, we need these to not be static when `BUILDING_SHARED_GC` is set (i.e. only static when building `default.c` as not a shared GC).

> (also, typo in the commit message.)

🙈

XrXr:
There is also a third situation where you have `BUILDING_SHARED_GC`, but the default GC is still built as a fallback in case DSO loading fails. In that case these will become non-static. RB_AMALGAMATED_DEFAULT_GC had the fallback use `static`, too.

peterzhu2118:
> In that case these will become non-static.

I don't think that's true.

`BUILDING_SHARED_GC` is only set when building using the `make shared-gc` target (it's added as a compiler flag:

https://github.com/ruby/ruby/blob/3a90663776f94e4a617a4069d1d073cc1eca0526/gc/extconf_base.rb#L8

So the default GC/fallback inside of Ruby will be static.

XrXr (approved):

PR #12248, opened 2024-12-04
Extend `include?` optimization to work when arg is a method with 1 arg

rwstauner:
Support common cases like:

```
%i[a b].include?(hash[:B])
[:a, :b].include?(array[0])
%w[100 200 400 600 700].include? response['status_code'].to_s
%w[1 true].include? ENV['VAR']
```

Original implementation in https://github.com/ruby/ruby/pull/12123


test/ruby/test_optimization.rb
```diff
@@ -1113,6 +1113,11 @@ def test_opt_duparray_send_include_p
       'x = :b; [:a, :b].include?(x)',
       '@c = :b; [:a, :b].include?(@c)',
       '@c = "b"; %i[a b].include?(@c.to_sym)',
+      'hash = {B: :b}; %i[a b].include?(hash[:B])',
+      'arr = Array(:b); %i[a b].include?(arr[0])',
+      'response = {"status_code" => 200}; [100, 200].include? response["status_code"].to_i',
```
nirvdrum, line 1118: In the implementation you noted that arbitrarily chained simple calls could be optimized, but your test case only has one explicit call with dot notation (you have another implicit call with `[]`). It could be good to add a test matching your comment: `[...].include?(arg.method1.method2)`.

rwstauner: Quite right, thank you!

PR #12254, opened 2024-12-04
[DOC] Mention `Object#initialize_clone` and `#initialize_dup`

kyanagi:
It seems there is no mention in the documentation about what `Object#initialize_clone` and `#initialize_dup` are.
Adding a brief explanation about them might be helpful.

### DIFF
--- a/object.c
+++ b/object.c
@@ -604,6 +604,10 @@ rb_obj_dup_setup(VALUE obj, VALUE dup)
  *  When using #dup, any modules that the object has been extended with will not
  *  be copied.
  *
+ *  If you need #clone and #dup to behave differently, you can use
+ *  #+initialize_clone+ and #+initialize_dup+.
+ *  By default, both methods call #+initialize_copy+.
+ *
  *	class Klass
  *	  attr_accessor :str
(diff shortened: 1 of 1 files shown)


object.c
```diff
@@ -604,6 +604,10 @@ rb_obj_dup_setup(VALUE obj, VALUE dup)
  *  When using #dup, any modules that the object has been extended with will not
  *  be copied.
  *
+ *  If you need #clone and #dup to behave differently, you can use
+ *  #+initialize_clone+ and #+initialize_dup+.
+ *  By default, both methods call #+initialize_copy+.
```
st0012, line 609: I think the phrasing can be clearer and the `#` sign should be inside the `+` quote:

```suggestion
 *  If you need #clone and #dup to behave differently, you can override
 *  +#initialize_clone+ and +#initialize_dup+ respectively.
 *  By default, both methods invoke +#initialize_copy+.
```

(There are other cases of incorrect quoting above this section but we can address them separately)

kyanagi:
@st0012 Thank you for your review. I updated the pull request.

PR #12261, opened 2024-12-04
Standardize on the name "modular GC"

peterzhu2118:
We have name fragmentation for this feature, including "shared GC", "modular GC", and "external GC". This commit standardizes the feature name to "modular GC" and the implementation to "GC library".


eileencodes (commented):
I agree going with modular for the pluggable makes the most sense since it's not really a shared object.

eightbitraptor:
I agree on the implementation being called a "GC Library", as it is a library that's being loaded, but I'm not entirely sure about  "modular GC" as a concept - to me that implies that the GC is composed of modules, which I don't think is entirely accurate in this case. it's more like there's an abstract GC interface with concrete implementations.

I agree though that "shared GC" and "external GC" are also not entirely accurate.

PR #12269, opened 2024-12-05
[Bug #20929] Fix timezone encoding

nobu:
- The default internal encoding is not taken into account to encode timezone name.
- Encode timezone name in UTF-8 on Windows.


unak (approved):
LGTM!

PR #12274, opened 2024-12-06
[DOC] Show `NONE` within `Measure` namespace

havenwood:
A couple times in code review I've seen constants inadvertently leak to top level from within a `Struct` or `Data` do block. I think it would be nice to show reopening the `Data` class when a constant is defined, so the constant is defined within the namespace. In this case, `Measure::NONE` instead of top level `Object::NONE`. It would also show readers that it's okay to reopen a `Data` class, which seems nice since some folk might not realize. Thanks for considering!

### DIFF
--- a/struct.c
+++ b/struct.c
@@ -1759,7 +1759,8 @@ rb_data_define(VALUE super, ...)
  *  important for redefining initialize in order to convert arguments or provide
  *  defaults:
  *
- *     Measure = Data.define(:amount, :unit) do
+ *     Measure = Data.define(:amount, :unit)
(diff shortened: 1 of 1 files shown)


havenwood:
cc @zverok (Please let me know if you'd prefer a bugs.ruby-lang.org issue discussion instead. I wasn't sure what you'd prefer for a nit docs suggestion. Thanks!)

nobu:
That `NONE` probably might be intended to be defined under `Measure`.
We would need a discussion for it, I guess.

PR #12277, opened 2024-12-06
Define constants in `Data` scope

nobu:
#12274


jeremyevans (commented):
We do not do this in other cases currently.  `Class.new` and `Struct.new` do not work this way, so I don't think `Data.define` should.  I suppose we could consider changing all of them at once, but that would be quite a significant change, which I think would need to be discussed during a developer meeting.

PR #12283, opened 2024-12-06
YJIT: Only enable disassembly colors for tty

tekknolagi:
This reduces noise when piping to a file:

```console
$ ruby --yjit --yjit-dump-disasm test.rb > log
```

or

```console
$ ruby --yjit --yjit-dump-disasm=log test.rb
```

### DIFF
--- a/yjit/src/disasm.rs
+++ b/yjit/src/disasm.rs
@@ -7,6 +7,37 @@ use crate::options::DumpDisasm;
 
 use std::fmt::Write;
 
+#[derive(Copy, Clone, Debug)]
+pub struct TerminalColor {
+    pub blue_begin: &'static str,
+    pub blue_end: &'static str,
+    pub bold_begin: &'static str,
+    pub bold_end: &'static str,
+}
+
+pub static TTY_TERMINAL_COLOR: TerminalColor = TerminalColor {
+    blue_begin: "\x1b[34m",
+    blue_end: "\x1b[0m",
+    bold_begin: "\x1b[1m",
+    bold_end: "\x1b[22m",
+};
+
+pub static NON_TTY_TERMINAL_COLOR: TerminalColor = TerminalColor {
+    blue_begin: "",
+    blue_end: "",
+    bold_begin: "",
+    bold_end: "",
+};
+
+/// Terminal escape codes for colors, font weight, etc. Only enabled if stdout is a TTY.
+pub fn get_colors() -> &'static TerminalColor {
+    if crate::utils::stdout_supports_colors() {
+        &TTY_TERMINAL_COLOR
+    } else {
+        &NON_TTY_TERMINAL_COLOR
+    }
+}
+
 /// Primitive called in yjit.rb
 /// Produce a string representing the disassembly for an ISEQ
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


yjit/src/options.rs
```diff
@@ -111,6 +136,7 @@ pub static mut OPTIONS: Options = Options {
     code_gc: false,
     perf_map: None,
     log: None,
+    color: NON_TTY_TERMINAL_COLOR,
```
maximecb, line 139: You're going to hate me because I can be very nitpicky, but `color` doesn't fit here because it's not a command-line option.

yjit/src/options.rs
```diff
@@ -300,7 +326,15 @@ pub fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
             }
 
             match opt_val {
-                "" => unsafe { OPTIONS.dump_disasm = Some(DumpDisasm::Stdout) },
+                "" => {
+                    unsafe { OPTIONS.dump_disasm = Some(DumpDisasm::Stdout) };
+                    extern "C" { fn isatty(fd: c_int) -> c_int; }
+                    let stdout = 1;
+                    let is_terminal = unsafe { isatty(stdout) } != 0;
+                    if is_terminal {
```
maximecb, line 334: Here I would suggest wrapping this logic into a function in `yjit/src/utils.rs`. Maybe `stdout_is_tty() -> bool`.

I also think that the rigorous way to do this would be to also check if the terminal supports color, but this is probably more rigorous than what we are currently doing (not checking at all).

yjit/src/options.rs
```diff
@@ -88,8 +96,25 @@ pub struct Options {
...
+    bold_begin: "\x1b[1m",
+    bold_end: "\x1b[22m",
+};
+
+pub static NON_TTY_TERMINAL_COLOR: TerminalColor = TerminalColor {
+    blue_begin: "",
+    blue_end: "",
+    bold_begin: "",
+    bold_end: "",
+};
```
maximecb, line 116: I would suggest moving these definitions to `disasm.rs` or `utils.rs`

yjit/src/disasm.rs
```diff
@@ -158,24 +158,25 @@ pub fn disasm_addr_range(cb: &CodeBlock, start_addr: usize, end_addr: usize) ->
...
     for insn in insns.as_ref() {
         // Comments for this block
         if let Some(comment_list) = cb.comments_at(insn.address() as usize) {
             for comment in comment_list {
                 if cb.outlined {
-                    write!(&mut out, "\x1b[34m").unwrap(); // Make outlined code blue
+                    write!(&mut out, "{}", color.blue_begin).unwrap(); // Make outlined code blue
                 }
-                writeln!(&mut out, "  \x1b[1m# {comment}\x1b[22m").unwrap(); // Make comments bold
+                writeln!(&mut out, "  {}# {comment}{}", color.bold_begin, color.bold_end).unwrap(); // Make comments bold
```
k0kubun, line 171:
```suggestion
                writeln!(&mut out, "  {color.bold_begin}# {comment}{color.bold_end}").unwrap(); // Make comments bold
```

It seems weird to me to mix the use of different interpolation styles (`"{comment}"` vs `"{}", color.bold_begin`) in the same line. The same applies to other uses of `color`.

XrXr: Interpolation only works for simple reference to locals like `"{comment}"` unfortunately, so "{color.bold_end}" doesn't build. Could switch to all "{}", though.

k0kubun: Oh sorry, I tested it using a release build, so it wasn't tested correctly. Never mind then :pray:

maximecb (approved):
Looks good! :)

PR #12287, opened 2024-12-08
YJIT: Spill/load argument registers to reuse blocks

k0kubun:
This PR improves the argument register allocator in `gen_send_iseq` to reuse existing blocks with different registers.  It previously allowed only moving operands already in a register, but this PR also allows spilling and loading registers. It allows YJIT to avoid compiling duplicated blocks to handle different `RegMapping` variations in most cases.

## YJIT stats
On lobsters, it reduces compilation time and inline_code_size by 3.3% (48ms and 154KB).

### Before
```
compiled_block_count:   71,284
inline_code_size:    4,728,964
outlined_code_size:  4,404,724
code_region_size:   11,628,544
yjit_alloc_size:    26,731,381
yjit_compile_time:   1464.60ms
```

### After
```
compiled_block_count:     68,339
inline_code_size:      4,574,334
outlined_code_size:    4,060,763
code_region_size:     11,157,504
yjit_alloc_size:      25,974,041
yjit_compile_time:     1416.86ms
```

## Benchmark
It seems to have a positive impact on multiple benchmarks. `chunky-png`, `erubi-rails`, `liquid-compile`, and `ruby-lsp` show 4-5% improvements. `activerecord`, `hexapdf`, and `psych-load` are 1% better too.

```
before: ruby 3.4.0dev (2024-12-07T10:02:17Z master bd831bcca5) +YJIT +PRISM [x86_64-linux]
after: ruby 3.4.0dev (2024-12-08T09:06:52Z yjit-reg-reuse 91c1d394f7) +YJIT +PRISM [x86_64-linux]

activerecord    97.3         0.3         95.8        0.2         1.034          1.016
chunky-png      250.0        0.4         239.1       0.3         1.043          1.046
erubi-rails     486.8        0.2         466.6       0.2         1.042          1.043
hexapdf         1040.9       2.1         1022.1      2.0         1.092          1.018
liquid-c        30.4         0.9         30.5        0.9         1.026          0.997
liquid-compile  34.1         0.8         32.6        2.7         1.021          1.047
liquid-render   40.8         0.6         41.1        0.6         1.021          0.991
lobsters        510.8        2.0         507.2       1.5         1.016          1.007
mail            70.4         0.3         70.4        0.3         1.022          1.000
psych-load      972.9        0.1         957.1       0.1         1.023          1.017
railsbench      1200.5       1.5         1197.6      2.7         1.037          1.002
rubocop         61.9         4.4         61.4        4.1         1.031          1.007
ruby-lsp        72.4         0.9         68.6        0.7         1.018          1.056
sequel          34.0         0.4         33.8        0.4         0.999          1.005
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -8136,53 +8136,16 @@ fn gen_send_iseq(
         pc: None, // We are calling into jitted code, which will set the PC as necessary
     }));
 
-    // Create a context for the callee
-    let mut callee_ctx = Context::default();
-
-    // Transfer some stack temp registers to the callee's locals for arguments.
-    let mapped_temps = if !forwarding {
-        asm.map_temp_regs_to_args(&mut callee_ctx, argc)
-    } else {
-        // When forwarding, the callee's local table has only a callinfo,
-        // so we can't map the actual arguments to the callee's locals.
-        vec![]
-    };
-
-    // Spill stack temps and locals that are not used by the callee.
-    // This must be done before changing the SP register.
-    asm.spill_regs_except(&mapped_temps);
-
-    // Saving SP before calculating ep avoids a dependency on a register
-    // However this must be done after referencing frame.recv, which may be SP-relative
-    asm.mov(SP, callee_sp);
-
-    // Log the name of the method we're calling to. We intentionally don't do this for inlined ISEQs.
-    // We also do this after gen_push_frame() to minimize the impact of spill_temps() on asm.ccall().
-    if get_option!(gen_stats) {
-        // Protect caller-saved registers in case they're used for arguments
-        asm.cpush_all();
-
-        // Assemble the ISEQ name string
-        let name_str = get_iseq_name(iseq);
-
-        // Get an index for this ISEQ name
-        let iseq_idx = get_iseq_idx(&name_str);
-
-        // Increment the counter for this cfunc
-        asm.ccall(incr_iseq_counter as *const u8, vec![iseq_idx.into()]);
-        asm.cpop_all();
-    }
-
     // No need to set cfp->pc since the callee sets it whenever calling into routines
     // that could look at it through jit_save_pc().
     // mov(cb, REG0, const_ptr_opnd(start_pc));
     // mov(cb, member_opnd(REG_CFP, rb_control_frame_t, pc), REG0);
 
-    // Stub so we can return to JITted code
-    let return_block = BlockId {
-        iseq: jit.iseq,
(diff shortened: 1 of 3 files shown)


yjit/src/codegen.rs
```diff
@@ -8211,29 +8174,90 @@ fn gen_send_iseq(
...
+                            asm.spill_reg(spilled_temp);
+                            callee_ctx.dealloc_reg(reg_opnd);
+                        }
+                        RegOpnd::Stack(_) => unreachable!("callee {:?} should have been spilled", reg_opnd),
+                    }
+                }
+            }
+            assert!(callee_ctx.get_reg_mapping().get_reg_opnds().len() <= existing_reg_mapping.get_reg_opnds().len());
+
+            // Load registers that are currently not used but used by the existing block
```
maximecb, line 8210: Just a small detail: could you clarify the above comment? This code is challenging to follow (not your fault, it's a non-trivial problem).

Load registers not used by the existing block. This would be the existing entry block? What is being loaded into registers here?

XrXr (approved):

yjit/src/codegen.rs
k0kubun: Updated the comment at https://github.com/ruby/ruby/pull/12287/commits/4ae7dcb095909d22a0289f4299c72bc0678068dc.

> Load registers not used by the existing block. This would be the existing entry block? What is being loaded into registers here?

An existing entry block would have no registers, so it would "Spill registers" (in L8197-8208) when the callee is an existing entry block.

"Load registers" (in L8213-8225) is used when there are other callsites that have compiled the same callee as a non-entry block. Imagine the previous instruction used `asm.ccall` in this block, the block would have no registers allocated. But other callsites may be using `putobject` to populate arguments, which would be in registers.

XrXr:
> It allows YJIT to avoid compiling duplicated blocks to handle different RegMapping variations in most cases.

Curious in what cases the attempt to reuse fail?

k0kubun:
> Curious in what cases the attempt to reuse fail?

First, when an ISEQ has been compiled as a JIT entry and then compiled as a JIT callsite, it fails. The entry block would have no registers allocated after the prologue code, but the JIT callsite may have some operands in registers. They use a different set of registers, so just shuffling registers wasn't enough. An easy example is `fib`.

Otherwise, it also easily fails when an ISEQ is called from multiple blocks. It can be multiple different caller ISEQs, which may spill at different times, and it can also be the above (entry vs non-entry) problem propagated to callees (when the caller block is simple enough to have no spill prior to the call).

PR #12294, opened 2024-12-09
Fix Cntrl, Graph and Print character properties of Cf characters.

raxoft:
Currently the invisible characters like BOM and other formatting characters from the `Cf `category are part of the `Graph` and `Print` group, which I believe is wrong, as they do not match the "Non-blank character (excludes spaces, control characters, and similar)" definition at all.  This change makes them part of `Cntrl` group instead.

Alternatively, if this is not accepted, at least the documentation of the `Graph` group should mention that it actually includes these characters, otherwise people might get burned by assuming they are not.

### DIFF
--- a/tool/enc-unicode.rb
+++ b/tool/enc-unicode.rb
@@ -142,7 +142,7 @@ def define_posix_props(data)
   data['Alnum'] = data['Alpha'] + data['Digit']
   data['Space'] = data['White_Space']
   data['Blank'] = data['Space_Separator'] + [0x0009]
-  data['Cntrl'] = data['Cc']
+  data['Cntrl'] = data['Cc'] + data['Cf']
   data['Word'] = data['Alpha'] + data['Mark'] + data['Digit'] + data['Connector_Punctuation']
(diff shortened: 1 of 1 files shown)


raxoft:
BTW, in case it is desirable to keep the `Cntrl` group intact, I don't mind. All I would expect is that they are not part of the `Graph` and `Print` group. So just subtracting `data['Cf']` from the `data['Graph']` few lines below would do the trick, too. However, I have chosen to make them part of the `Cntrl` group as I believe the original intention of these groups was that each character is in the end part of one group or another.

XrXr:
Thanks for looking into this. I've pushed some documentation fixes to point out this pitfall.

While I agree with you that this behavior is confusing, it's also long-standing behavior adopted intentionally -- `Graph` [behaves][1] this way in Oniguruma, the regex engine on which Ruby's is based. If you'd like to push for making a breaking change to make it more intuitive, please consider filing a report at https://bugs.ruby-lang.org/projects/ruby-master/issues

[1]: https://github.com/kkos/oniguruma/blob/5eaee9f5f8f674aff4875c2b35db00758fa349d6/doc/RE#L246

PR #12295, opened 2024-12-09
Fix cleaning modular GC

peterzhu2118:


common.mk
```diff
@@ -1958,10 +1954,12 @@ modular-gc: probes.h modular-gc-precheck
 	$(CP) gc/$(MODULAR_GC)/librubygc.$(MODULAR_GC).$(DLEXT) $(modular_gc_dir)
 
 clean-modular-gc:
-	- $(CHDIR) gc/$(MODULAR_GC) && $(exec) $(MAKE) TARGET_SO_DIR=./ clean || $(NULLCMD)
+	- find gc -type d -mindepth 1 -maxdepth 1 -exec sh -c '$(CHDIR) "{}" && $(MAKE) TARGET_SO_DIR=./ clean || $(NULLCMD)' \; || $(NULLCMD)
```
nobu, line 1957: This command should be in template/Makefile.in, not in common.mk.
`find.exe` on Windows differs from POSIX `find`, rather it is close to `fgrep`.

common.mk
```diff
@@ -1958,10 +1954,12 @@ modular-gc: probes.h modular-gc-precheck
...
+	-$(Q) $(RMDIR) gc
```
nobu, line 1958: `clean` targets remove just compiled results, object files, shared objects and so on.
Makefiles and headers will be kept, so this `rmdir` will fail.

PR #12317, opened 2024-12-12
[DOC] Fix `LoadError`'s linking issue

st0012:
Original issue: https://github.com/ruby/rdoc/issues/1128

The problem is caused by the `# :stopdoc:` directive in `bundled_gems.rb`, which's scope covers the redefinition of `LoadError`.

Since the goal of `# :stopdoc:` is to hide the documentation of `Gem::BUNDLED_GEMS`, we can use `# :nodoc:` on it instead.

### Result

`LoadError` is now auto-linked like other classes/modules:

Screenshot 2024-12-12 at 12 57 29

The patched `message` method is still not included in the documentation:

Screenshot 2024-12-12 at 12 58 19

And the `Gem::BUNDLED_GEMS` module is now properly hidden from the documentation:

**Before**

Screenshot 2024-12-12 at 12 59 22

**After**

Screenshot 2024-12-12 at 12 59 04

### DIFF
--- a/error.c
+++ b/error.c
@@ -3392,7 +3392,7 @@ syserr_eqq(VALUE self, VALUE exc)
  *
  *  * NoMemoryError
  *  * ScriptError
- *    * {LoadError}[https://docs.ruby-lang.org/en/master/LoadError.html]
+ *    * LoadError
  *    * NotImplementedError
  *    * SyntaxError
  *  * SecurityError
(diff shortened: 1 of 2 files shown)


peterzhu2118 (approved):
Thanks for looking into this!

hsbt:
👍

PR #12318, opened 2024-12-12
[DOC] Small fix and tweak to string literal syntax documentation

Earlopain:
Since https://github.com/ruby/ruby/commit/5e1001f754b34e1f0cc67563512c6036b6eb75ab, the "String Literals" link is broken on https://docs.ruby-lang.org/en/master/syntax/literals_rdoc.html

Instead of changing the anchor reference, I changed the header directly, to keep in line with other titles. I checked if there are references so `label-String` to not break other references and didn't find any, so this is probably fine?

In a separate commit, I moved the `Escape Sequence` documentation further down. I don't believe that should be the first (or even second) thing someone sees when looking up string syntax. Instead I've moved it after `Character Literal` and before `Here Document Literals`

### DIFF
--- a/doc/syntax/literals.rdoc
+++ b/doc/syntax/literals.rdoc
@@ -138,46 +138,6 @@ Also \Rational numbers may be imaginary numbers.
 
 == \String Literals
 
-=== Escape Sequences
-
-Some characters can be represented as escape sequences in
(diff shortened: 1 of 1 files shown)


Earlopain:
No problem, I rebased

PR #12319, opened 2024-12-12
Fix use-after-free in ep in Proc#dup for ifunc procs

peterzhu2118:
[Bug #20950]

ifunc proc has the ep allocated in the data of the TypedData object. If an ifunc proc is duplicated, the ep points to the ep of the source object. If the source object is freed, then the ep of the duplicated object now points to a freed memory region. If we try to use the ep we could crash.

For example, the following script crashes:

```ruby
p = { a: 1 }.to_proc
100.times do
  p = p.dup
  GC.start
  p.call
rescue ArgumentError
end
```

This commit changes ifunc proc to also duplicate the ep when it is duplicated.


proc.c
```diff
@@ -680,6 +680,30 @@ cfunc_proc_new(VALUE klass, VALUE ifunc)
     return procval;
 }
 
+VALUE
+rb_func_proc_dup(VALUE src_obj)
+{
+    RUBY_ASSERT(BUILTIN_TYPE(src_obj) == T_DATA);
+    RUBY_ASSERT(RTYPEDDATA_TYPE(src_obj) == &proc_data_type);
```
XrXr, line 687:
```suggestion
    RUBY_ASSERT(rb_typeddata_is_instance_of(src_obj, &proc_data_type));
```

proc.c
```diff
@@ -680,6 +680,29 @@ cfunc_proc_new(VALUE klass, VALUE ifunc)
...
+    RUBY_ASSERT(rb_typeddata_is_instance_of(src_obj, &proc_data_type));
+
+    rb_proc_t *src_proc;
+    GetProcPtr(src_obj, src_proc);
+    RUBY_ASSERT(vm_block_type(&src_proc->block) == block_type_ifunc);
+
+    cfunc_proc_t *proc;
+    VALUE proc_obj = TypedData_Make_Struct(rb_obj_class(src_obj), cfunc_proc_t, &proc_data_type, proc);
+
+    memcpy(&proc->basic, src_proc, sizeof(rb_proc_t));
```
XrXr, line 695:
```suggestion
    proc->basic = src_proc->basic;
```

Tricky memcpy assuming layout doesn't seem necessary here.

peterzhu2118: We can't do that because the `block` in `rb_proc_t` is defined as a const.

XrXr:
```suggestion
    memcpy(&proc->basic, &src_proc->basic, sizeof(proc->basic));
```
Ah okay, let's just make lhs the same as rhs then.

peterzhu2118: 😅 Unfortunately the LHS and RHS are different types. LHS is a `cfunc_proc_t` while RHS is a `rb_proc_t`.

proc.c
```diff
@@ -1456,7 +1485,14 @@ rb_hash_proc(st_index_t hash, VALUE prc)
         rb_bug("rb_hash_proc: unknown block type %d", vm_block_type(&proc->block));
     }
 
-    return rb_hash_uint(hash, (st_index_t)proc->block.as.captured.ep);
+    /* ifunc procs have their own allocated ep. If an ifunc is duplicated, they
+     * will point to different ep but they should return the same hash code, so
+     * we cannot include the ep in the hash. */
+    if (vm_block_type(&proc->block) != block_type_ifunc) {
+        hash = rb_hash_uint(hash, (st_index_t)proc->block.as.captured.ep);
+    }
```
XrXr, line 1493: This seems like it'd introduce a hash collision for all the ifunc procs that have the same func and data. Maybe we need to hash the contents of ep (VM_ENV_DATA_SIZE)?

peterzhu2118: I don't think there's much benefits to it. All `ifunc` have the same values for the four ep attributes except for ones created from `Method#to_proc`, and we have to make sure we don't use `VM_ENV_DATA_INDEX_ENV` since it's an object that can change in value during compaction.

XrXr (approved):

PR #12320, opened 2024-12-12
Improve "Building Ruby" docs 

amomchilov:
Currently, several of the commands listed in the "Testing Ruby" commands won't work, because they're expecting to run from the repository root ([Example 1](https://github.com/ruby/ruby/blob/79d90e73511dc9dfda7926802b624d6fec22fe12/doc/contributing/testing_ruby.md?plain=1#L26-L27), [2](https://github.com/ruby/ruby/blob/79d90e73511dc9dfda7926802b624d6fec22fe12/doc/contributing/testing_ruby.md?plain=1#L45-L46), [3](https://github.com/ruby/ruby/blob/79d90e73511dc9dfda7926802b624d6fec22fe12/doc/contributing/testing_ruby.md?plain=1#L60), [4](https://github.com/ruby/ruby/blob/79d90e73511dc9dfda7926802b624d6fec22fe12/doc/contributing/testing_ruby.md?plain=1#L111)). This contradicts the "Building Ruby" docs, which suggest [building Ruby inside a `build/` subdirectory](https://github.com/ruby/ruby/blob/79d90e73511dc9dfda7926802b624d6fec22fe12/doc/contributing/building_ruby.md?plain=1#L94-L101). This PR corrects all those commands, to make them work from within the `build/` dir.

This PR also makes several other improvements to the general clarity and readability of this documentation, especially for new readers. See the individual commits for more details.

### DIFF
--- a/doc/contributing/testing_ruby.md
+++ b/doc/contributing/testing_ruby.md
@@ -2,6 +2,8 @@
 
 All the commands below assume that you're running them from the `build/` directory made during [Building Ruby](building_ruby.md).
 
+Most commands below should work with [GNU make](https://www.gnu.org/software/make/) (the default on Linux and macOS), [BSD make](https://man.freebsd.org/cgi/man.cgi?make(1)) and [NMAKE](https://learn.microsoft.com/en-us/cpp/build/reference/nmake-reference), except where indicated otherwise.
+
 ## Test suites
 
 There are several test suites in the Ruby codebase:
@@ -41,7 +43,7 @@ We can run any of the make scripts [in parallel](building_ruby.md#label-Running+
     make test OPTS=-v
     ```
 
-    To run a specific file with GNU make, we can use:
+    (GNU make only) To run a specific file, we can use:
 
     ```sh
     make ../test/ruby/test_string.rb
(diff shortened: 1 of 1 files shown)


NEWS.md
```diff
@@ -242,13 +242,13 @@ details of the default gems or bundled gems.
     * Kernel#caller, Thread::Backtrace::Location’s methods, etc. are also changed accordingly.
 
         Old:
-        ```
+        ```text
```
k0kubun, line 245: You seem to have mixed up doc improvements and the diff from https://github.com/ruby/ruby/pull/12322.

k0kubun:
Duplicated diffs are harder to review and sometimes become a source of conflicts. If you're proposing a diff in other PRs, please remove it from this PR.

NEWS.md
k0kubun: When you're "stacking" PRs, you usually point the stacked PR to the base branch you're stacking on (which doesn't work if you do that from a fork, so you have to give up in this case). What you're doing is filing PRs with duplicated changes to the same branch. Since you're actually not stacking PRs, you should rebase your branch and remove duplication. 

As Stan said on Slack, your changes don't need to be stacked because markdown files don't have dependencies unlike code and these changes work correctly without each other.

k0kubun (approved):

PR #12322, opened 2024-12-12
[DOC] Improve formatting in Markdown files

amomchilov:
Best reviewed commit-by-commit.

This PR:

1. Corrects issues with the indentation in the WASM docs
2. Uses the `console` highlight language in some code blocks, which better handles the `$` at the start of lines when viewed on GitHub (and behaves the same in RDoc)
3. Migrates the the feature request template to MarkDown
4. Sets the highlight language for any all code blocks that didn't have it, and corrects it in some places.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -259,21 +259,24 @@ details of the default gems or bundled gems.
     * Other keys now have spaces around `=>`: `'{"user" => 1}'`, while previously they didn't: `'{"user"=>1}'`
 
 * Kernel#Float() now accepts a decimal string with decimal part omitted. [[Feature #20705]]
-  ```
+
+  ```rb
   Float("1.")    #=> 1.0 (previously, an ArgumentError was raised)
   Float("1.E-1") #=> 0.1 (previously, an ArgumentError was raised)
   ```
 
 * String#to_f now accepts a decimal string with decimal part omitted. [[Feature #20705]]
   Note that the result changes when an exponent is specified.
-  ```
+
+  ```rb
   "1.".to_f    #=> 1.0
   "1.E-1".to_f #=> 0.1 (previously, 1.0 was returned)
   ```
 
 * Object#singleton_method now returns methods in modules prepended to or included in the
   receiver's singleton class. [[Bug #20620]]
-  ```
+
+  ```rb
   o = Object.new
   o.extend(Module.new{def a = 1})
   o.singleton_method(:a).call #=> 1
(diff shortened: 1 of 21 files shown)


NEWS.md
```diff
@@ -242,13 +242,13 @@ details of the default gems or bundled gems.
     * Kernel#caller, Thread::Backtrace::Location’s methods, etc. are also changed accordingly.
 
         Old:
-        ```
+        ```text
```
k0kubun, line 245: What difference does this change make?

amomchilov: It communicates that the syntax highlighting language was intentionally set to none, rather than having been forgotten.

That way, when you search for opening bare triple-back ticks, this won't surface any more.

hsbt (changes requested):
Can you exclude `doc/strscan`, `spec/ruby` and `tool/lrama`? We should fix upstream, not this repository.

amomchilov:
@hsbt Ah yes, I always forget about that when I'm doing bulk changes across this repo. Will update

NEWS.md
k0kubun: I see. I was so used to <code>```</code> used for plain-text code blocks that I got rather confused by looking at this style as markdown code. It might help some people like you, but it could be rather hard to understand for people like me.

So I don't find it's necessarily a positive change. We probably also don't want to spend time reviewing diffs that change <code>``\`</code> to <code>```text</code>, which does nothing visually and doesn't clarify the intention by itself (at least it didn't to me), going forward. I appreciate your effort to improve the doc but I would like this diff to be removed from the PR.

amomchilov:
@hsbt Done. I split off 3 PRs:

1. https://github.com/ruby/strscan/pull/126
2. https://github.com/ruby/spec/pull/1227
3. https://github.com/ruby/lrama/pull/482

k0kubun:
> [amomchilov](https://github.com/amomchilov) requested review from [hsbt](https://github.com/hsbt) and [k0kubun](https://github.com/k0kubun)

Did you see https://github.com/ruby/ruby/pull/12322#discussion_r1882952859?

NEWS.md
amomchilov: Fair enough, I've pulled it out.

It already served its purpose for me, which was to mark this code block as "fixed" to the regexp I was using to find the triple-back-tick pairs that didn't have a language set.

This could be worth reconsidering in the future, if there was a lint step added to remind people to set the language on new code blocks they write. It's like the diff between `void f()` and `void f(void)`.

amomchilov (commented):
Missed it somehow @k0kubun. Addressed.

k0kubun (approved):

hsbt:
👌

PR #12325, opened 2024-12-12
[DOC] Exclude irb from documentation generation

st0012:
## Motivation

- IRB has its own documentation in the project readme and https://ruby.github.io/irb/ already.
- As a tool, its internal implementation details are not relevant to the user.
- Excluding it from the documentation generation reduces the size of the generated documentation and makes search results more relevant.

## Implementation

The `.rdoc_options` change will exclude IRB from `docs.ruby-lang.org/en`'s documentation generation. And since IRB is not included anymore, the `lib/irb/.document` is obsolete so I deleted it (which was manually added in the past, not synced from `ruby/irb`).

I also checked the there are no references to IRB so it should not break any links.


tompng:
It depends on the policy of ruby/ruby's document, but I think this change makes sense.

I think having only a single document URL is less confusing. Some default gem's document will be removed from docs.ruby-lang.org when it become a bundled gem, so the standalone document deployed in each gem will last longer.

jeremyevans (approved):

PR #12334, opened 2024-12-13
Update NEWS.md with `Time#xmlschema`, `Array#fetch_values` and `String#append_as_bytes`

byroot:
cc @ko1


byroot:
Also added an entry for `Symbol#to_s` returning a chilled string.

PR #12335, opened 2024-12-13
sprintf.c: avoid excessive rb_str_resize() calls

rhenium:
Calling rb_str_resize() with the same length repeatedly is pointless. This is currently done every single time something is about to be appended to the result String. This makes a difference in this micro-benchmark:

    $ cat t-sprintf.yml
    prelude: |
      fmt1 = "%d"
      args1 = [1]
      fmt2 = "%d"*10000
      args2 = [1]*10000

    benchmark:
      rep1: fmt1 % args1
      rep10000: fmt2 % args2
    $ benchmark-driver -e ./miniruby-654404a9c266 -e ./miniruby ./t-sprintf.yml
    [...]
    Comparison:
                                 rep1
              ./miniruby:   7261451.0 i/s
    ./miniruby-654404a9c266:   6704623.2 i/s - 1.08x  slower

                             rep10000
              ./miniruby:      7825.9 i/s
    ./miniruby-654404a9c266:      5592.9 i/s - 1.40x  slower

### DIFF
--- a/sprintf.c
+++ b/sprintf.c
@@ -66,15 +66,17 @@ sign_bits(int base, const char *p)
 #define FPREC0 128
 
 #define CHECK(l) do {\
-    int cr = ENC_CODERANGE(result);\
     RUBY_ASSERT(bsiz >= blen); \
-    while ((l) > bsiz - blen) {\
-        bsiz*=2;\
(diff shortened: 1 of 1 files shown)


rhenium:
I broke something. This is not ready for review, closing for now.

PR #12342, opened 2024-12-13
[PRISM] using []= to set kwargs is a syntax error

eightbitraptor:
Fixes [Bug #20952]


kddnewton (approved):
Looks great! Do you mind adding a test? You can generate it with bin/prism error in the prism repo.

PR #12355, opened 2024-12-16
YJIT: Load registers on JIT entry to reuse blocks

k0kubun:
There are two ways of compiling the first block of an ISEQ:

1. JIT entry block: This happens on interpreter -> JIT calls.
2. Callee block: This happens on JIT -> JIT calls.

To reuse blocks with different `RegMapping`, this PR loads the registers of an existing block when compiling (1) it if it has been compiled as (2) before. The previous PR https://github.com/ruby/ruby/pull/12287 covered the most common case that (2) is compiled after (1), and this PR covers the other case that (1) is compiled after (2).

## YJIT stats
On lobsters, this PR reduces the number of compiled blocks by 0.5%. No significant impact on compilation performance.

### Before
```
compiled_block_count:     47,814
inline_code_size:      4,561,009
outlined_code_size:    4,164,701
code_region_size:     11,108,352
yjit_alloc_size:      25,887,989
yjit_compile_time:     1055.95ms
```

### YJIT
```
compiled_block_count:     47,557
inline_code_size:      4,541,408
outlined_code_size:    4,103,075
code_region_size:     11,087,872
yjit_alloc_size:      25,798,021
yjit_compile_time:     1056.78ms
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -1061,14 +1061,13 @@ fn gen_leave_exception(ocb: &mut OutlinedCb) -> Option<CodePtr> {
 pub fn gen_entry_chain_guard(
     asm: &mut Assembler,
     ocb: &mut OutlinedCb,
-    iseq: IseqPtr,
-    insn_idx: u16,
+    blockid: BlockId,
 ) -> Option<PendingEntryRef> {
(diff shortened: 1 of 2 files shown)


maximecb (approved):
Not a huge impact, but ok sure :)

PR #12357, opened 2024-12-16
Fix constant documents

soutaro:
This PR fixes two constant document issues:

### RDoc doesn't generate docs for constants registered by `rb_define_global_const`

You can find the docs for `ARGF`, `ARGV`, ... are missing in [Object docs](https://docs.ruby-lang.org/en/master/Object.html). 

スクリーンショット 2024-12-16 20 52 10

This is because `rb_define_global_const` is not supported in RDoc. This PR adds `Document-const` directive.

### The generated docs have odd constant name duplication

Some constants defined in [File::Constants](https://docs.ruby-lang.org/en/master/File/Constants.html) has wired constant name duplication.

スクリーンショット 2024-12-16 20 53 25

I don't know the detail of the RDoc documentation generation, but I found the constant definition function calls (`rb_file_const` and `rb_define_const`) having `Document-class:` causes this issue. This PR drops the unnecessary `Document-class:` directives.


nobu:
> ### RDoc doesn't generate docs for constants registered by `rb_define_global_const`

I think it is fixed now.

soutaro:
Thanks @nobu. Dropped the commit for `rb_define_global_const`. I hope the new version of RDoc will ship soon...

PR #12361, opened 2024-12-16
Bundle repl_type_completor 0.1.9

tompng:
It was updated to 0.1.8 in #12351, then Rolled back to 0.1.7 because of test-bundled-gems ci failure.
0.1.9 should fix(omit) the failed test.

The failed test is omitted (omit only when run from test-bundled-gems ci) because the environment did not meet the prerequisites of the test.


hsbt:
👍

PR #12362, opened 2024-12-16
Sync net-http commits

st0012:
Syncing the below commits from https://github.com/ruby/net-http/pull/199 failed:
- https://github.com/ruby/net-http/commit/9bcf818fd009eafb11107c7457aa56d533d16d94 
- https://github.com/ruby/net-http/commit/5e34e74261f40f4f10c93d7700761c437117f494

### DIFF
--- a/lib/net/http.rb
+++ b/lib/net/http.rb
@@ -475,8 +475,7 @@ module Net   #:nodoc:
   #
   # - {::start}[rdoc-ref:Net::HTTP.start]:
   #   Begins a new session in a new \Net::HTTP object.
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):

jeremyevans:
This might need to wait until after 3.4 release, as @hsbt is freezing gem versions for the release.

jeremyevans:
> This might need to wait until after 3.4 release, as @hsbt is freezing gem versions for the release.

Looks like the auto merging isn't supposed to be stopped until December 18, so this should be OK to merge now.

PR #12368, opened 2024-12-17
[Bug #20958] fix ENV.keys encoding on windows

YO4:
Fix the encoding of ENV.keys elements to UTF-8 on windows. They contain UTF-8 byte sequences with locale encoding.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -5731,7 +5731,7 @@ env_aset(VALUE nm, VALUE val)
 static VALUE
 env_keys(int raw)
 {
-    rb_encoding *enc = raw ? 0 : rb_locale_encoding();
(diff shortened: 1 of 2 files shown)


nobu:
Good catch!
Preferably, couldn't you file the bug to the [BTS](https://bugs.ruby-lang.org), and add a test?

YO4:
This is filed as [[Bug #20958]](https://bugs.ruby-lang.org/issues/20958)
Also I pushed a test for this. That covers other platforms too.
Possibly a test with default_internal would be more appropriate in test_m17n.rb, though.

hsbt (approved):

PR #12370, opened 2024-12-17
[PRISM] Recurse use_deconstructed_cache in Alternation Nodes

eightbitraptor:
This fixes the behavioural difference between Prism and parse.y when evaluating the following code

```ruby
1 in [1 | [1]]
```

Fixes [Bug #20956]


kddnewton (approved):
Knew it was going to be something small...

PR #12373, opened 2024-12-17
[DOC] Tweaks for array.rb

BurdetteLamar:
I have remarked/emphasized/exemplified that certain arguments may be Numeric (not just Integer).

Another method (#fetch) in this same file describes the arguments as integer-convertible (not just Numeric).  There are many such descriptions throughout the core and standard library, Should these (for simplicity) describe as Numeric (instead of integer-convertible)?  Or should the ones that say Numeric be changed to integer-convertible (whenever that's true)?  Or does it even matter?

### DIFF
--- a/array.rb
+++ b/array.rb
@@ -66,8 +66,9 @@ class Array
   # With a non-negative numeric argument +count+ given,
   # returns a new array containing +count+ random elements from +self+:
   #
-  #   a.sample(3) # => [8, 9, 2]
-  #   a.sample(6) # => [9, 6, 0, 3, 1, 4]
+  #   a.sample(3)   # => [8, 9, 2]
+  #   a.sample(3.1) # => [4, 6, 9]
+  #   a.sample(6)   # => [9, 6, 0, 3, 1, 4]
   #
   # The order of the result array is unrelated to the order of +self+.
   #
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
The fact that non-integers work in these methods seems like a mistake.  Maybe we don't want to fix it for backwards compatibility reasons, but we should definitely not be encouraging additional use by documenting that.

BurdetteLamar:
So mention neither integer-convertible nor numeric?

    On Tuesday, December 17, 2024 at 11:05:11 AM CST, Jeremy Evans ***@***.***> wrote:  
 
 


@jeremyevans commented on this pull request.

The fact that non-integers work in these methods seems like a mistake. Maybe we don't want to fix it for backwards compatibility reasons, but we should definitely not be encouraging additional use by documenting that.

jeremyevans:
> So mention neither integer-convertible nor numeric?

Correct.  I don't think it was intended for these to be passed non-integer.  It probably just works due to implementation details.

BurdetteLamar:
> > So mention neither integer-convertible nor numeric?
> 
> Correct. I don't think it was intended for these to be passed non-integer. It probably just works due to implementation details.

@peterzhu2118, do you agree that this PR should be closed (and not merged)?

peterzhu2118:
I agree with @jeremyevans. I think while they work (perhaps briefly mentioned in the documentation), this should not be encouraged, so perhaps we don't want to include examples of it.

BurdetteLamar:
> I agree with @jeremyevans. I think while they work (perhaps briefly mentioned in the documentation), this should not be encouraged, so perhaps we don't want to include examples of it.

So the merge should be undone?

peterzhu2118:
> So the merge should be undone?

Sorry, I'm not quite sure what you mean by this?

BurdetteLamar:
> > So the merge should be undone?
> 
> Sorry, I'm not quite sure what you mean by this?

This PR mentions numeric and provides examples;  I thought we were agreeing that that's not a good idea.

jeremyevans:
@BurdetteLamar I think the problem is "the merge should be undone" doesn't make sense, as this hasn't been merged.  The words you are looking for is "This pull request can be closed".

BurdetteLamar:
> @BurdetteLamar I think the problem is "the merge should be undone" doesn't make sense, as this hasn't been merged. The words you are looking for is "This pull request can be closed".

Duh.  Somehow I got the idea that this PR had been merged.  So, may I close it?

jeremyevans:
Sure, you can always close a PR you created.

PR #12376, opened 2024-12-17
[DOC] Add note to rb_id2name about GC compaction

peterzhu2118:


XrXr (approved):
Technically the string could go away during GC too (for dynamic IDs), but this is good enough of a warning.

PR #12384, opened 2024-12-18
[DOC] Fix string output example that contains old Hash#inspect style

tompng:
https://bugs.ruby-lang.org/issues/20962

This pull request fixes document of Hash#inspect and some more string output examples.
But does not fix these kind of documents:
```ruby
h # => {:x=>0, :y=>100}

h.slice(:baz, :foo) # => {:baz=>2, :foo=>0}
```
because it's not clearly wrong, and there are too many of them.

### DIFF
--- a/array.c
+++ b/array.c
@@ -5084,7 +5084,7 @@ rb_ary_concat(VALUE x, VALUE y)
  *  When string argument +string_separator+ is given,
  *  equivalent to <tt>self.join(string_separator)</tt>:
  *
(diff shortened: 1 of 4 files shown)


st0012:
I think having one `Hash#inspect` entry under the `Hash` item in [`Core classes updates`](https://docs.ruby-lang.org/en/master/NEWS_md.html#label-Core+classes+updates) should help developers discover the change too. I know it's already mentioned [here](https://docs.ruby-lang.org/en/master/NEWS_md.html#label-Compatibility+issues) but IMO we can mention it twice if it helps developers understand the changes.

PR #12386, opened 2024-12-18
[DOC] Don't document gemspec files

Earlopain:
Since https://github.com/ruby/ruby/commit/cf711863cbe0b81be5ff0adfb64936137995cc4e (https://github.com/ruby/ruby/pull/12325), gemspecs are included because the default exclude is overwritten.

I'm not sure if this change is the best way to solve it but it does remove these files from the generated documentation.

https://docs.ruby-lang.org/en/master/lib/English_gemspec.html

image

### DIFF
--- a/.rdoc_options
+++ b/.rdoc_options
@@ -10,3 +10,4 @@ rdoc_include:
 
 exclude:
 - lib/irb
+- .gemspec


st0012 (approved):
Good catch!
I think `.rdoc_options`' overriding behaviour is a bit surprising, especially when the same flag only appends value 🤔 But at the same time, there's no way to override the [defaults](https://github.com/ruby/rdoc/blob/2b79892ab436a728d5ddebdb5663d7915c08e755/lib/rdoc/options.rb#L366-L369) if the user doesn't want them.

nobu:
I think it would be better add an option to remove arguments from that list, or to clear it entirely.

Earlopain:
I can make an issue over at rdoc to discuss this at, finishing that before next release doesn't seem feasable. Would that be OK?

st0012:
Yeah I agree this should be merged first as I've released the final rdoc version for Ruby 3.4. @nobu can you help us merge it?

jeremyevans (approved):

PR #12387, opened 2024-12-18
Don't use xmalloc for newobj_i

peterzhu2118:
newobj_i is a GC hook that runs when a new object is allocated. We should not use xmalloc because it can potentially trigger GC and a lot of code is unsafe to trigger a GC right after an object has been allocated because they perform initialization for the object and assume that the GC does not trigger before then.

### DIFF
--- a/ext/objspace/object_tracing.c
+++ b/ext/objspace/object_tracing.c
@@ -99,7 +99,12 @@ newobj_i(VALUE tpval, void *data)
         delete_unique_str(arg->str_table, info->class_path);
     }
     else {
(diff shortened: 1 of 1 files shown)


jhawthorn:
I think `malloc` is probably desirable here, but the `st_insert` lower is going to `xmalloc` if it needs to resize, no?

peterzhu2118:
I opened https://github.com/ruby/ruby/pull/12419 which fixes this in an alternate method that allows `xmalloc` in the `RUBY_INTERNAL_EVENT_NEWOBJ` hooks.

PR #12389, opened 2024-12-18
Properly document Ractor#require

zverok:
[Currently](https://docs.ruby-lang.org/en/master/Ractor.html), the docs have:
* internal methods `Ractor._activated` and `Ractor._require` included;
* (metaprogrammed) `Ractor.require` wrongly recognized by RDoc as class method (with no documentation).

This PR just adds `:nodoc: for those methods.

### DIFF
--- a/ractor.rb
+++ b/ractor.rb
@@ -889,7 +889,7 @@ class Ractor
   end
 
   # internal method
-  def self._require feature
+  def self._require feature # :nodoc:
     if main?
       super feature
     else
@@ -901,11 +901,11 @@ class Ractor
     private
 
     # internal method that is called when the first "Ractor.new" is called
-    def _activated
+    def _activated # :nodoc:
       Kernel.prepend Module.new{|m|
         m.set_temporary_name '<RactorRequire>'
 
-        def require feature
+        def require feature # :nodoc: -- otherwise RDoc outputs it as a class method
           if Ractor.main?
             super
           else


ractor.rb
```diff
@@ -888,8 +888,22 @@ def self.main?
...
 
+  ##
+  # :method: require
+  # :args: feature
+  #
+  # A redefinition of Kernel#require, available inside ractor body. In non-main ractor,
+  # it delegates to the main one to do the real +require+.
+  #
+  #    r = Ractor.new {
+  #      require 'json' # it is a call to Ractor#require
```
ko1, line 899: `it is a call to Ractor#require` is misleading.

* There is no `Ractor#require`
* `Ractor#_require` is called, but user doesn't need to know the internal API

So,

```suggestion
  #      require 'json' # require on the main ractor and wait for the result on this ractor.
```

ko1: I'm not sure it is in RDoc, but doc/ractor.md is better to write?

zverok: Yeah, my understanding was imperfect (to put it mildly). I adjusted my mental model via a bit of experimenting, to understand it better. These docs aren’t needed, just hiding internal methods with `:nodoc:` is.

I’ll see about updating `ractor.md` a bit later (with proper explanations).

Sorry for the confusion :pray:

ko1 (approved):

jeremyevans (approved):

PR #12390, opened 2024-12-18
[DOC] Improve the NEWS page's rendering on docs.ruby-lang.org

st0012:
1. RDoc's markdown parser expects 4 spaces for nested lists. So currently default/bundled gems' releases are not rendered correctly on [docs.ruby-lang.org/en](https://docs.ruby-lang.org/en/master/NEWS_md.html)
    Screenshot 2024-12-18 at 20 18 42

2. I think it'd be very helpful to link to default/bundled gems' GitHub repos.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -182,96 +182,97 @@ The following default gem is added.
 The following default gems are updated.
 
 * RubyGems 3.6.1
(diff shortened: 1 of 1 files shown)


st0012:
cc @hsbt

hsbt:
👍

PR #12398, opened 2024-12-19
[Bug #20965] Define `it` like an ordinary argument

nobu:
[[Bug #20965]](https://bugs.ruby-lang.org/issues/20965)


k0kubun (approved):

k0kubun:
This fixes https://bugs.ruby-lang.org/issues/20955 too.

PR #12412, opened 2024-12-20
Fix bundled gem warnings locations

deivid-rodriguez:
I tried to fix some warnings bundled gems warnings when using Ruby 3.4.0-rc1 and it was inconvenient to find where `require`'s were actually happening because the uplevel seemed one level higher than I was expecting.

I'm trying to fix it here and add some specs to prove the fix.

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -72,8 +72,9 @@ module Gem::BUNDLED_GEMS # :nodoc:
       kernel_class.send(:alias_method, :no_warning_require, :require)
       kernel_class.send(:define_method, :require) do |name|
         if message = ::Gem::BUNDLED_GEMS.warning?(name, specs: spec_names)
-          if ::Gem::BUNDLED_GEMS.uplevel > 0
-            Kernel.warn message, uplevel: ::Gem::BUNDLED_GEMS.uplevel
+          uplevel = ::Gem::BUNDLED_GEMS.uplevel
+          if uplevel > 0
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
Specs seem to pass but I should probably check bootsnap manually because the bootsnap specs seem not yet enabled.

deivid-rodriguez:
I checked manually that this also fixes the location when bootsnap is present 👍.

hsbt:
@deivid-rodriguez Thank you. I also confirmed to fix location for dependencies.

But this changed sub-feature location like:

```
-e:7: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add fiddle to your Gemfile or gemspec to silence this warning.
```

to


```
/Users/hsbt/Documents/github.com/ruby/ruby/.ext/common/fiddle/import.rb:2: warning: fiddle was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add fiddle to your Gemfile or gemspec to silence this warning.
```

My expectation of sub-feature examples, it will show caller location of `fiddle/import`, not `fiddle`. We should fix wrong location for dependencies over sub-feature location. It's good to fix sub-feature example until Ruby 3.4.0 or 3.4.1.

deivid-rodriguez:
@hsbt You're totally right, I actually noticed it but assumed `fiddle/import` was something specific in ruby-core setup and not an actual file in the `fiddle` gem. Let me look into it.

PR #12309, opened 2024-12-11
Output a Ruby bug report on ASAN error

peterzhu2118:

PR #12413, opened 2024-12-20
Don't print bug report in asan_death_callback when no VM

peterzhu2118:
If we don't have the VM (e.g. printing memory leaks in LSAN after shutdown) then we will crash when we try to print the bug report. This issue was reported in: https://github.com/ruby/ruby/pull/12309#issuecomment-2555766525


jhawthorn:
Thank you!

PR #12416, opened 2024-12-20
Remove and stop syncing IRB documentation

st0012:
Follow-up to #12325

Since we're not generating IRB documentation, we don't need to sync its `doc/` folder and keep the previous content.


st0012:
cc @tompng

tompng (approved):

PR #12428, opened 2024-12-22
Modular GC with relative load

nobu:


tool/m4/ruby_modular_gc.m4
```diff
@@ -17,8 +17,8 @@ AS_IF([test x"$modular_gc_dir" != x], [
     # Ensure that modular_gc_dir is always an absolute path so that Ruby
     # never loads a modular GC from a relative path
```
voxik, line 18: Is this comment still valid with this PR?

PR #12432, opened 2024-12-23
Simple doco fixes for ERB

zenspider:


jeremyevans:
Changes look good, but this should be sent upstream to ruby/erb.

k0kubun:
Yup, this should be filed on https://github.com/ruby/erb. Once we merge that, the change will also be synchronized to this repository. But it doesn't work the other way around.

zenspider:
Figured out how to pull the commits over using `git format-patch`... Closing this in favor of https://github.com/ruby/erb/pull/56

PR #12435, opened 2024-12-23
Add a test case for nested block `it`

yui-knk:


mame (approved):
https://bugs.ruby-lang.org/issues/20930

PR #12441, opened 2024-12-23
Fix [Bug #20779] Dedicated native thread creation failed bug

luke-gru:
When a dedicated native thread fails to create (OS can't create more threads), don't add the ruby-level thread to the thread scheduler. If it's added, then next time the thread is scheduled to run it will sleep forever, waiting for a native thread that doesn't exist to pick it up.


ko1:
Thank you!

PR #12446, opened 2024-12-24
not a incompatbile, but a feature

ko1:
The change of `Object#singleton_method` doesn't harm existing code so it should be new feature (bug fix).


jeremyevans:
This can change existing behavior if you have a method in the singleton class, and a method with the same name in a module prepended to the singleton class.  It now returns the method in the module prepended to the singleton class, instead of the method in the singleton class.  Other cases where the behavior differs would raise an exception previously.

PR #12449, opened 2024-12-24
Launchable: Exit 0 not to stop CI pipeline

ono-max:
For handling the error: https://github.com/ruby/ruby/actions/runs/12625056559/job/35176167637#step:6:713.


ono-max:
@XrXr 

Thank you for your suggestion.
[continue-on-error](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) is inappropriate in this case because we want to ignore only Launchable error. We want to let developers know if tests failed, so I added the pipeline `| true`.

PR #12457, opened 2024-12-24
Introduce a timeout to prevent `rb_thread_fd_select` from hanging with write(2) failure

shioimm:
Rarely, there are cases where a write(2) call from a child thread to notify the main thread of the completion of name resolution fails. If this happens while the main thread is waiting in `rb_thread_fd_select`, rb_thread_fd_select may not notice that the name resolution has completed and end up hanging.

This issue becomes a problem when there are no sockets currently being connected, no addresses ready for immediate connection attempts, and name resolution has already completed for one address family while the main thread is waiting for the name resolution of the other address family. (If name resolution is not completed for either address family, the chances of write(2) failing in both child threads are likely low.)

To avoid this issue, a timeout is introduced to rb_thread_fd_select under the above conditions. This way, even if the issue occurs,
the completion of name resolution should still be detected in the subsequent `if (!resolution_store.is_all_finished) ...` block.

### DIFF
--- a/ext/socket/ipsocket.c
+++ b/ext/socket/ipsocket.c
@@ -854,7 +854,18 @@ init_fast_fallback_inetsock_internal(VALUE v)
             delay = tv_to_timeout(ends_at, now);
             delay_p = &delay;
         } else {
-            delay_p = NULL;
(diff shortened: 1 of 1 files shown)


Pricer187:
What does that even mean


Christopher

PR #7599, opened 2023-03-28
vm_dump.c: Fix duplicate dump in LoongArch64.

zangruochen:
Fix some duplicate definitions in the LoongArch64 section of vm_dump.c

PR #12473, opened 2024-12-26
Revert LoongArch register changes

MingcongBai:
Commits 71b253cdf3 ("Added LARCH_REG_FP to dump results") and a07bf6d5ff ("Use constants for register numbers") introduced registers never defined in glibc or musl, the only two widely used C libraries available for LoongArch, causing build failures.

Reverting these two commits fixes build on glibc (musl build and tests pending) and passes all unit tests.

### DIFF
--- a/vm_dump.c
+++ b/vm_dump.c
@@ -1001,22 +1001,22 @@ rb_dump_machine_register(FILE *errout, const ucontext_t *ctx)
 #   elif defined __loongarch64
         dump_machine_register(mctx->__gregs[LARCH_REG_SP], "sp");
         dump_machine_register(mctx->__gregs[LARCH_REG_A0], "a0");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A1], "a1");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A2], "a2");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A3], "a3");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A4], "a4");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A5], "a5");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A6], "a6");
-        dump_machine_register(mctx->__gregs[LARCH_REG_A7], "a7");
+        dump_machine_register(mctx->__gregs[LARCH_REG_A0+1], "a1");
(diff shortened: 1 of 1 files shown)


xry111:
Glibc: https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/loongarch/sys/ucontext.h

Musl: https://git.musl-libc.org/cgit/musl/tree/arch/loongarch64/bits/signal.h

I can find `LA_REG_A5` nowhere.

ziyao233:
I've verified this does fix loongarch64 build on musl libc.

Without this patch, ruby master fails to build with:

```
compiling vm_dump.c
compiling vm_sync.c
vm_dump.c:1004:45: error: use of undeclared identifier 'LARCH_REG_A1'
 1004 |         dump_machine_register(mctx->__gregs[LARCH_REG_A1], "a1");
      |                                             ^
vm_dump.c:1005:45: error: use of undeclared identifier 'LARCH_REG_A2'
 1005 |         dump_machine_register(mctx->__gregs[LARCH_REG_A2], "a2");
```

hsbt:
Should we use https://github.com/ruby/ruby/pull/7599/files#diff-58f0e788b47677353796215b78b47f17abd66abb9c4ef223d7c726c7109e79e3R1008 directly for fp register?

xry111:
> Should we use https://github.com/ruby/ruby/pull/7599/files#diff-58f0e788b47677353796215b78b47f17abd66abb9c4ef223d7c726c7109e79e3R1008 directly for fp register?

I think so, as there seems no reasonable way to represent $fp otherwise.  `LARCH_REG_A0 + 8` or `LARCH_REG_S0 - 1` just looks so unreasonable...

(Also $fp is sometimes named $s9 if it's not really a frame pointer, thus writing `LARCH_REG_S0 + x` for any `x` can potentially promote people to incorrectly write `LARCH_REG_S0 + 9`.  So doing so seems not a good idea.)

nobu:
Does `LARCH_REG_S1` exist?

xry111:
> Does `LARCH_REG_S1` exist?

Yes, see the links to glibc and musl headers I provided above.

hsbt:
Does anyone try with https://github.com/ruby/ruby/pull/12515?

xry111:
> Does anyone try with #12515?

Works for me (applying on top of 3.4.1 release).

hsbt:
I merged https://github.com/ruby/ruby/pull/12515 instead of this.

PR #12476, opened 2024-12-26
Add dbg debugging helper method

pawurb:
Hi. I’m author of https://github.com/pawurb/dbg-rb gem. This PR implements a minimal `dbg` method without any external dependencies.

`dbg` method is inspired by Rust where it's built-in [into std-lib](https://doc.rust-lang.org/std/macro.dbg.html).  AFAIK in Ruby there's no simple mechanism to puts debug values together with caller info without using external dependencies. What’s more frustrating is that while `p nil` will output `“nil”` to the std, `puts nil` will print a blank line, sometimes making debugging sessions confusing. 

`dbg` produces verbose output together with informative file name and LOC info. I think that such built-in feature would be useful for many Ruby devs. 

Screenshot 2024-12-27 at 00 00 23

### DIFF
--- a/lib/dbg.rb
+++ b/lib/dbg.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# This is a simple debugging helper method that can be used to
+# print out the source file and line number of the caller
+# together with the debugged value.
+
+# Usage:
+# dbg("Hello world", [1, 2, 3])
+# => [file.rb:12] "Hello world"
+# => [file.rb:12] [1, 2, 3]
+
+def dbg(*msgs)
+  loc = caller_locations.first.to_s
+  matching_loc = loc.match(/.+(rb)\:\d+\:(in)\s/)
+  src = if matching_loc.nil?
+      loc
+    else
+      matching_loc[0][0..-5]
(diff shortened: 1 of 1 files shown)


hsbt:
Can you submit this proposal to https://bugs.ruby-lang.org/? We don't discuss feature request in GitHub.

lib/dbg.rb
```diff
@@ -0,0 +1,28 @@
...
+def dbg(*msgs)
+  loc = caller_locations.first.to_s
+  matching_loc = loc.match(/.+(rb)\:\d+\:(in)\s/)
+  src = if matching_loc.nil?
+      loc
+    else
+      matching_loc[0][0..-5]
+    end
+  file, line = src.split(":")
+  file = file.split("/").last(2).join("/")
```
nobu, line 21: `caller_locations` returns a list of `Thread::Backtrace::Location`, which has `absolute_path`, `path` and `lineno` methods.

pawurb: @nobu thank you! for now I've submitted proposal here https://bugs.ruby-lang.org/issues/20987

PR #12484, opened 2024-12-28
[Bug #20990] Fix tokenizing `"\M-\[mbchar]"` and `"\C-\[mbchar]"`

tompng:
https://bugs.ruby-lang.org/issues/20990


nobu:
Needs to fix prism as well.
See #12485

tompng:
Fixed by #12485

PR #12491, opened 2024-12-31
[Bug #20983] Stop auto-linking to `Ruby` and `Set`

st0012:
This excludes `Set` and `Ruby` from RDoc's auto-linking:

- Usages of the word `Set` don't need to be escaped anymore (~20 occurrences) but the references to the class need to be explicitly linked (1 occurrence).
- There's no escaped `Ruby` usage atm.

[[Bug #20983]](https://bugs.ruby-lang.org/issues/20983)

### DIFF
--- a/doc/standard_library.md
+++ b/doc/standard_library.md
@@ -59,7 +59,7 @@ of each.
 - Reline ([GitHub][reline]): GNU Readline and Editline in a pure Ruby implementation
 - Resolv ([GitHub][resolv]): Thread-aware DNS resolver library in Ruby
 - SecureRandom ([GitHub][securerandom]): Interface for a secure random number generator
-- Set ([GitHub][set]): Provides a class to deal with collections of unordered, unique values
+- [Set](rdoc-ref:Set) ([GitHub][set]): Provides a class to deal with collections of unordered, unique values
 - Shellwords ([GitHub][shellwords]): Manipulates strings with the word parsing rules of the UNIX Bourne shell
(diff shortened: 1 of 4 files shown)


st0012:
cc @nobu

doc/standard_library.md
```diff
@@ -59,7 +59,7 @@ of each.
 - Reline ([GitHub][reline]): GNU Readline and Editline in a pure Ruby implementation
 - Resolv ([GitHub][resolv]): Thread-aware DNS resolver library in Ruby
 - SecureRandom ([GitHub][securerandom]): Interface for a secure random number generator
-- Set ([GitHub][set]): Provides a class to deal with collections of unordered, unique values
+- [Set](rdoc-ref:Set) ([GitHub][set]): Provides a class to deal with collections of unordered, unique values
```
nobu, line 62: It should be able to use backquotes.

```suggestion
- `Set` ([GitHub][set]): Provides a class to deal with collections of unordered, unique values
```

nobu:
In `Init_version`, "The Ruby module" is the module and should be linked.

doc/standard_library.md
st0012: If it’s ok, I prefer using rdoc-ref for now as it has the missing reference check.

nobu: We will switch to explicit rdoc-ref from implicit autolinking?

st0012: No I don't plan to do it. But in the case where we want to link a not-auto-linked word, I think it makes sense to make it explicit and allow RDoc to check for missing references.

PR #12499, opened 2025-01-02
YJIT: Fix crash when yielding keyword args

nicholasdower:
Co-authored-by: @DaniTheLion

Code like the following is crashing for us on 3.4.1 ([ruby-2025-01-02-180907.ips.zip](https://github.com/user-attachments/files/18294151/ruby-2025-01-02-180907.ips.zip)):

```ruby
def a(&) = yield(x: 0)

1000.times { a { |x:| x } }
```

Output:
```
thread '<unnamed>' panicked at ./yjit/src/codegen.rs:8018:13:
assertion `left == right` failed
  left: 0
 right: 1
stack backtrace:
   0: rust_begin_unwind
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/90b35a6239c3d8bdabc530a6a0816f7ff89a0aaf/library/core/src/panicking.rs:74:14
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /Users/nickdower/.rustup/toolchains/stable-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/panicking.rs:367:5
   4: yjit::codegen::gen_send_iseq
```

The assertion that is failing seems to have been added in #9596. As far as we can tell, this code:

```rust
// Pop surplus positional arguments when yielding
if arg_setup_block {
    let extras = argc - required_num - opt_num;
    if extras > 0 {
        // Checked earlier. If there are keyword args, then
        // the positional arguments are not at the stack top.
        assert_eq!(0, kw_arg_num);
```

Is referring to this prior check:

```rust
// Bail if we can't drop extra arguments for a yield by just popping them
if supplying_kws && arg_setup_block && argc > (kw_arg_num + required_num + opt_num) {
```

But in one case `kw_arg_num` is being included in the sum, and in the other it isn't.

We attempted a fix by ignoring `kw_arg_num` in the initial check.

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -37,14 +37,18 @@ assert_equal "ok", %q{
 }
 
 # test discarding extra yield arguments
-assert_equal "2210150001501015", %q{
+assert_equal "22131300500015901015", %q{
   def splat_kw(ary) = yield *ary, a: 1
 
   def splat(ary) = yield *ary
 
-  def kw = yield 1, 2, a: 0
+  def kw = yield 1, 2, a: 3
+
+  def kw_only = yield a: 0
 
   def simple = yield 0, 1
+
+  def none = yield
 
   def calls
     [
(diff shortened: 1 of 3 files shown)


nicholasdower:
cc: @XrXr. Any advice you can provide would be greatly appreciated.

nobu:
`Co-authored-by:` in the middle isn't parsed as a trailer.
It needs to be placed at the end of the commit log.

nicholasdower:
> `Co-authored-by:` in the middle isn't parsed as a trailer. It needs to be placed at the end of the commit log.

Thank you very much @nobu. Fixed.

XrXr (commented):
I'm merging the test as-is, but going with a different fix. The suggested fix works by refusing to compile the previously problematic situation when it's fine to support it.

Thanks very much for taking taking the time to investigate!

maximecb:
Thanks for the report and the fix everyone 🙏

PR #12503, opened 2025-01-03
[DOC] Tweaks for Hash.new

BurdetteLamar:


hash.rb
```diff
@@ -1,39 +1,39 @@
+# :markup: markdown
```
peterzhu2118, line 1: Is there a reason why we're using markdown and not RDoc?

hash.rb
```diff
@@ -1,39 +1,39 @@
...
-  #   h = Hash.new {|hash, key| "Default value for #{key}" }
-  #   h.default # => nil
-  #   h.default_proc.class # => Proc
-  #   h[:nosuch] # => "Default value for nosuch"
-  #
-  # If both a block and a <tt>default_value</tt> are given, raises an +ArgumentError+
-  #
-  # If the optional keyword argument +capacity+ is given, the hash will be allocated
-  # with enough capacity to accommodate this many keys without having to be resized.
+  #   Hash.new(default_value = nil, capacity: 0) {|self, key| ... } -> new_hash
```
peterzhu2118, line 5: I think we still need two cases here, one without a block and with a `default_value`, and one with a block but without `default_value`.

hash.rb
BurdetteLamar: @st0012 believes (I think) that markdown is better going forward;  personally, I'm agnostic.  I'll be happy to switch if that's best.

@jeremyevans (and any other interested parties), should we say anything in the doc guide about choosing a markup format?

jeremyevans: I would prefer continuing to document in rdoc format, unless there are major benefits from using markdown for what is being documented.  If we do switch from rdoc to markdown, that will be done by a tool, with all changes being made at the same time.  After that point, we would write markdown.

st0012: Sorry that I think the confusion came from when @BurdetteLamar was rewriting IRB's documentation, I said my preference was Markdown. I didn't make it clear that it is only the case inside IRB based on the maintainers' preferences.

As for `ruby/ruby`, my personal opinion is:
-  For page documents, like contributing guides and maintainers list...etc., switching is beneficial as `.md` files are way easier to review on GH and/or preview locally through editor extensions. But there isn't any guideline on this yet AFAIK.
- For comment documentation, the benefit is less obvious as we review on the raw markup anyway. And in projects like `ruby/ruby`, keeping the markup consistent across API documentation is probably more beneficial for now.

BurdetteLamar: Restored to default markup (rdoc).

hash.rb
BurdetteLamar: Recast.

BurdetteLamar:
I think that `#[]`, `#default`, and `#default_proc` all need work to improve their cooperation, but will defer to next PRs.

hash.rb
```diff
@@ -1,39 +1,35 @@
...
-  # If both a block and a <tt>default_value</tt> are given, raises an +ArgumentError+
-  #
-  # If the optional keyword argument +capacity+ is given, the hash will be allocated
-  # with enough capacity to accommodate this many keys without having to be resized.
+  #   Hash.new(default_value = nil, capacity: 0)
+  #   Hash.new(capacity: 0) {|self, key| ... } -> new_hash
+  #
+  # Returns a new empty \Hash object;
+  # initializes the values of Hash#default and Hash#default_proc,
+  # which determine the value to be returned by method Hash::[] when the entry does not exist.
```
peterzhu2118, line 8: I think you mean the instance method `[]`, not class method.

```suggestion
  # which determine the value to be returned by method Hash#[] when the entry does not exist.
```

hash.rb
```diff
@@ -1,39 +1,35 @@
...
-  #   h = Hash.new {|hash, key| "Default value for #{key}" }
-  #   h.default # => nil
-  #   h.default_proc.class # => Proc
-  #   h[:nosuch] # => "Default value for nosuch"
-  #
-  # If both a block and a <tt>default_value</tt> are given, raises an +ArgumentError+
-  #
-  # If the optional keyword argument +capacity+ is given, the hash will be allocated
-  # with enough capacity to accommodate this many keys without having to be resized.
+  #   Hash.new(default_value = nil, capacity: 0)
```
jeremyevans, line 3:
```suggestion
  #   Hash.new(default_value = nil, capacity: 0) -> new_hash
```

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  #   Hash.new(capacity: 0) {|self, key| ... } -> new_hash
+  #
+  # Returns a new empty \Hash object;
+  # initializes the values of Hash#default and Hash#default_proc,
```
jeremyevans, line 7:
```suggestion
  # Returns a new empty \Hash object.
  # Initializes the values of Hash#default and Hash#default_proc,
```

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  # With a block given, initializes Hash#default to +nil+,
+  # and Hash#default_proc to a new Proc containing the block's code:
+  #
+  #   h = Hash.new {|hash, key| "Hash #{hash}: Default value for #{key}" }
+  #   h.default      # => nil
+  #   h.default_proc # => #<Proc:0x00000289b6fa7048 (irb):185>
+  #   h[:nosuch]     # => "Hash {}: Default value for nosuch"
+  #
+  # If optional keyword argument +capacity+ is given with a positive integer value +n+,
+  # initializes the hash with enough capacity to accommodate +n+ keys without resizing.
```
jeremyevans, line 28:
```suggestion
  # initializes the hash with enough capacity to accommodate +n+ entries without resizing.
```

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  #
+  # Raises ArgumentError if both +default_value+ and a block are given.
+  #
```
jeremyevans, line 31:
```suggestion
```

With the two given call-seqs, it's impossible to provide both a default value and a block, so I'm not sure we need to specify this.

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  # which determine the value to be returned by method Hash#[] when the entry does not exist.
+  #
+  # With no block given, initializes Hash#default to the given +default_value+,
+  # and Hash#default_proc to +nil+:
+  #
+  #   h = Hash.new        # => {}
+  #   h.default           # => nil
+  #   h.default_proc      # => nil
+  #   h = Hash.new(false) # => {}
+  #   h.default           # => false
```
jeremyevans, line 17:
```suggestion
  # By default, a hash has +nil+ values for both +default+ and +default_proc+:
  #
  #   h = Hash.new        # => {}
  #   h.default           # => nil
  #   h.default_proc      # => nil
  #
  # If the +default_value+ argument is provided, it sets the +default+ value for the hash:
  #
  #   h = Hash.new(false) # => {}
  #   h.default           # => false
  #   h.default_proc      # => nil
```

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  #
+  # With a block given, initializes Hash#default to +nil+,
+  # and Hash#default_proc to a new Proc containing the block's code:
```
jeremyevans, line 20:
```suggestion
  # If a block is given, it sets the +default_proc+ value:
```

jeremyevans (commented):
Looks good, but I have some minor wording change suggestions.

hash.rb
```diff
@@ -1,39 +1,35 @@
...
+  #   h.default      # => nil
+  #   h.default_proc # => #<Proc:0x00000289b6fa7048 (irb):185>
+  #   h[:nosuch]     # => "Hash {}: Default value for nosuch"
+  #
+  # If optional keyword argument +capacity+ is given with a positive integer value +n+,
+  # initializes the hash with enough capacity to accommodate +n+ keys without resizing.
+  #
+  # Raises ArgumentError if both +default_value+ and a block are given.
+  #
+  # See also [Methods for Creating a Hash](rdoc-ref:Hash@Methods+for+Creating+a+Hash).
```
st0012, line 32: We should use RDoc link syntax here:

```suggestion
  # See also {Methods for Creating a Hash}[rdoc-ref:Hash@Methods+for+Creating+a+Hash].
```

hash.rb
BurdetteLamar: @jeremyevans, my `Commit suggestion` button is grayed out (don't know why).  Can you force this last change?

hash.rb
```diff
@@ -1,39 +1,32 @@
...
   #
-  # Returns a new empty +Hash+ object.
+  # Returns a new empty \Hash object.
+  # Initializes the values of Hash#default and Hash#default_proc,
+  # which determine the value to be returned by method Hash#[] when the entry does not exist.
   #
-  # The initial default value and initial default proc for the new hash
-  # depend on which form above was used. See {Default Values}[rdoc-ref:Hash@Default+Values].
+  # With no block given, initializes Hash#default to the given +default_value+,
+  # and Hash#default_proc to +nil+:
```
jeremyevans, line 11:
```suggestion
  # By default, a hash has +nil+ values for both +default+ and +default_proc+:
  #
  #   h = Hash.new        # => {}
  #   h.default           # => nil
  #   h.default_proc      # => nil
```

hash.rb
```diff
@@ -1,39 +1,32 @@
...
-  # If neither an argument nor a block is given,
-  # initializes both the default value and the default proc to <tt>nil</tt>:
-  #   h = Hash.new
-  #   h.default # => nil
-  #   h.default_proc # => nil
+  #   h = Hash.new        # => {}
+  #   h.default           # => nil
+  #   h.default_proc      # => nil
+  #   h = Hash.new(false) # => {}
+  #   h.default           # => false
```
jeremyevans, line 17:
```suggestion
  # If the +default_value+ argument is provided, it sets the +default+ value for the hash:
  #
  #   h = Hash.new(false) # => {}
  #   h.default           # => false
  #   h.default_proc      # => nil
```

jeremyevans (commented):
GitHub doesn't allow applying suggestions if the code is outdated, and doesn't support suggestions covering deleted lines.  I split that suggestion into two.

st0012 (approved):
I think squashing commits for this PR will be great.

peterzhu2118 (approved):

PR #12505, opened 2025-01-03
Allow YJIT `mem-size` and `call-threshold` to be set at runtime via `YJIT.enable()`

annichai-stripe:
###  Summary

Allow the `mem-size` and `call-threshold` parameters for YJIT to be set at runtime via the `YJIT.enable` method.

### Motivation
Currently, YJIT's `mem-size` and `call-threshold` parameters can only be set through command-line options. There's no interface to configure these YJIT parameters when enabling YJIT at runtime.

At Stripe, our goal is to maximize the performance improvements brought by YJIT across our Ruby services. Each service has unique characteristics and resource constraints. We wish to 

- Fine-tune YJIT settings for each service without using command-line arguments, which can be difficult to manage in our codebase.
- Dynamically compute and adjust YJIT parameters based on service environment, traffic, and resources.
- More easily experiment with and optimize YJIT configurations across our service fleet.

### Changes

- Modified the `YJIT.enable` method in `yjit.rb` to accept two additional optional parameters: `mem_size` and `call_threshold`. 
- Updated the YJIT logic to set these runtime parameters when provided, prior to enabling.

### Usage

After this change, users can enable YJIT with custom `mem-size` and `call-threshold` values at runtime like this:

```ruby
YJIT.enable(mem_size: 64, call_threshold: 10)
```

Note: Because of https://github.com/ruby/ruby/blob/master/yjit/src/yjit.rs#L158, after custom config to call_threshold has been provided, it will not get increased to 120 when the number of ISEQs in the process reaches 40,000.
### Test

Tested with this script: https://gist.github.com/annichai-stripe/362c66429251e894ae5acc73932dc9db
Can confirm the options get set as expected.

### Would love review on
- The implementation approach, especially the interaction between Ruby, C, and Rust code
- Any potential edge cases or scenarios we may have overlooked
- Additional test cases that might be valuable


maximecb:
Hello Stripe 👋 

Not opposed to such a change. However, it might be better to use the new `yjit-mem-size` option (more intuitive) rather than `exec-mem-size` which we might deprecate.

yjit.rb
```diff
@@ -44,11 +44,18 @@ def self.reset_stats!
...
+
+    if exec_mem_size.nil? && call_threshold.nil?
+      # Proceed with original enable process
+      at_exit { print_and_dump_stats } if stats
+      call_yjit_hooks
+      Primitive.rb_yjit_enable(stats, stats != :quiet, log, log != :quiet)
+    else
+      # Call helper function for custom configuration
+      enable_with_custom_config(stats, log, exec_mem_size, call_threshold)
+    end
```
maximecb, line 58: This seems a bit complicated.

How about passing the Ruby values directly to `Primitive.rb_yjit_enable` ?

yjit.rb
```diff
@@ -254,6 +261,33 @@ class << self
...
+    def validate_config_params(exec_mem_size, call_threshold)
+      if exec_mem_size && !(exec_mem_size.is_a?(Integer) && exec_mem_size.between?(1, 2048))
+        raise ArgumentError, 'Invalid exec_mem_size: must be an integer between 1 and 2048'
+      end
+
+      return unless call_threshold
+      return if call_threshold.is_a?(Integer) && call_threshold >= 1
+
+      raise ArgumentError, 'Invalid call_threshold: must be an integer >= 1'
+    end
```
maximecb, line 289: Might be a bit simpler to directly validate the values directly in the enable function.

annichai-stripe:
@maximecb Sorry for not getting back to this sooner and thanks for your suggestions in the last review. I've updated the PR to move validation into `.rb`, and we now will set `mem-size` instead of `exec-mem-size`. Also added some test cases.

Also, do you know if those failing checks on CI must be resolved before merge? I'm unsure if they're relevant to the changes here. Thanks!

yjit.rb
```diff
@@ -259,6 +263,19 @@ def add_yjit_hook(hook)
...
+      if mem_size
+        raise ArgumentError, "mem_size must be a Integer" unless mem_size.is_a?(Integer)
+        raise ArgumentError, "mem_size must be between 1 and 2048 MB" unless (1..2048).include?(mem_size)
+      end
+
+      if call_threshold
+        raise ArgumentError, "call_threshold must be a Integer" unless call_threshold.is_a?(Integer)
+        raise ArgumentError, "call_threshold must be a positive integer" unless call_threshold.positive?
+      end
+    end
```
maximecb, line 277: nitpick: this doesn't really need to be in a separate method, it just forces people to go look elsewhere to see the validation code

maximecb (approved):
Small nitpick, otherwise looks good 👍

maximecb:
> @maximecb Sorry for not getting back to this sooner and thanks for your suggestions in the last review. I've updated the PR to move validation into `.rb`, and we now will set `mem-size` instead of `exec-mem-size`. Also added some test cases.
> 
> Also, do you know if those failing checks on CI must be resolved before merge? I'm unsure if they're relevant to the changes here. Thanks!

no problem there is no time crunch 👍 

I don't know what caused those jobs to fail. The error messages are not very useful. I'm trying to re-run the failed jobs.

maximecb:
We just have a lot of test failures on master at the moment. Seems safe to merge.

yjit/src/yjit.rs
```diff
@@ -185,8 +185,26 @@ pub extern "C" fn rb_yjit_code_gc(_ec: EcPtr, _ruby_self: VALUE) -> VALUE {
...
+            }
+            println!("mem_size set to: {} MB ({} bytes)", mem_size_mb, mem_size_bytes);
+        }
+
+        if !call_threshold.nil_p() {
+            let threshold = call_threshold.as_isize() >> 1;
+            unsafe {
+                rb_yjit_call_threshold = threshold as u64;
+            }
+            println!("rb_yjit_call_threshold set to: {}", threshold);
```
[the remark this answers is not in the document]
annichai-stripe: Thanks for the suggestions @maximecb @k0kubun

test/ruby/test_yjit.rb
```diff
@@ -142,6 +142,32 @@ def test_yjit_enable_with_monkey_patch
     RUBY
   end
 
+  def test_yjit_enable_with_valid_runtime_call_threshold_option
+    assert_in_out_err(['--yjit-disable', '-e',
+                       'RubyVM::YJIT.enable(call_threshold: 1); puts RubyVM::YJIT.enabled?']) do |stdout, stderr, _status|
+      assert_empty stderr
+      assert_include stdout.join, 'rb_yjit_call_threshold set to: 1'
```
k0kubun, line 149:
```suggestion
```

Now that you removed logging, we should remove this too.

maximecb:
Finally got all the tests passing. Sorry this took so long to merge!

annichai-stripe:
Thank you for all the help! @maximecb

PR #12518, opened 2025-01-06
[Bug #21006] Fix defined_expr compilation of method call with parenth…

tompng:
The argument `explicit_receiver` of `pm_compile_defined_expr0` is set to true for the optimized insn for chained method calls.

```c
if (PM_NODE_TYPE_P(cast->receiver, PM_CALL_NODE) && !BLOCK_P((const pm_call_node_t *) cast->receiver)) {
  Optimized insns created here needs insns with explicit_receiver=true
} else {
  Requires insns with explicit_receiver=false
}
```

Other changes: change explicit_receiver to false where it is always false.
Example: pushing explicit receiver to the stack while compiling implicit_node doesn't help compiling call_node's receiver.


kddnewton (approved):
Makes sense, thanks for the quick fix!

PR #12521, opened 2025-01-06
Ensure th->running_time_us is set back to 0 right before a thread switch

luke-gru:
Fixes [Bug #21007]


jhawthorn:
I think this is the same issue being solved by https://github.com/ruby/ruby/pull/12094

luke-gru:
Oh yeah, I wasn't aware of that issue. The other solution lgtm, should I close?

jhawthorn:
Closing in favour of 12094

PR #12523, opened 2025-01-07
Support more node types in compilation of PM_DEFINED_NODE

tompng:
Fix compiling `defined?(f(*))`, `defined?(f(**))`, `defined?([*])`, `defined?([**])`
```ruby
def f(*, **, &); defined?(p(*, **, &)); end
def f(*, **, &); defined?([*, **]); end
```

Add more node types into switch-case branch of pm_compile_defined_expr0 to support these codes.

```ruby
defined?((alias a b))
defined?((alias $a $b))
defined?(!(a..b))
def f(...); defined?(f(...)); end
defined?(!/foo/)
defined?(!/#{1}/)
defined?((END{}))
defined?((a rescue b))
defined?((undef a))
tap { defined?(it) }
```
and
```ruby
# shareable-constant-value: experimental_everything
defined?(A=1)
```

`!/foo/` `!/#{1}/` `!(a..b)` are `"method"`.
Standalone PM_MATCH_LAST_LINE_NODE, PM_INTERPOLATED_MATCH_LAST_LINE_NODE, PM_FLIP_FLOP_NODE is treated as `"expression"` in this pull request. It does not affect the result because they don't appear directly under PM_DEFINED_NODE.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -3845,9 +3845,10 @@ pm_compile_defined_expr0(rb_iseq_t *iseq, const pm_node_t *node, const pm_node_l
               }
               case PM_ASSOC_SPLAT_NODE: {
                 const pm_assoc_splat_node_t *assoc_splat = (const pm_assoc_splat_node_t *) element;
-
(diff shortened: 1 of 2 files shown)


tompng:
Fixed by #12584

PR #12526, opened 2025-01-08
Add the -g flag, an expanded version of -s

sampersand:


nobu:
We need the discussion for new features at the BTS.
Please file at https://bugs.ruby-lang.org/projects/ruby-master/issues/new.

nobu:
And this has conflicts probably because of wrong base commit.
You should fix it first.

sampersand:
@nobu Done! I was posted this first whilst I worked on the bug issue: https://bugs.ruby-lang.org/issues/21015?next_issue_id=21012

ruby.c
```diff
@@ -330,6 +330,7 @@ usage(const char *name, int help, int highlight, int columns)
         M("-e 'code'",     "",			   "Execute given Ruby code; multiple -e allowed."),
         M("-Eex[:in]",     ", --encoding=ex[:in]", "Set default external and internal encodings."),
         M("-Fpattern",	   "",			   "Set input field separator ($;); used with -a."),
+        M("-g",		   "",			   "Similar to -s, but supports posix shorthands and integers"),
```
nobu, line 333: Could this be merged to `-s`?

```C
        M("-s, -g",	   "",			   "Define global variables using switches following program path.\n"
            "-g supports posix shorthands and integers"),
```

ruby.c
```diff
@@ -355,7 +356,7 @@ usage(const char *name, int help, int highlight, int columns)
 #endif
         M("-h",		   "",			   "Print this help message; use --help for longer message."),
     };
-    STATIC_ASSERT(usage_msg_size, numberof(usage_msg) < 25);
+    STATIC_ASSERT(usage_msg_size, numberof(usage_msg) < 26);
```
nobu, line 359: And this can be reverted.

ruby.c
```diff
@@ -829,12 +841,44 @@ process_sflag(int sflag)
             if (s[1] == '-' && s[2] == '\0')
                 break;
 
+            // Handle `-abc` in extended switch mode. Also, removes the leading `-` for long-form
+            // `--foo`s in extended mode so they end up as `$foo`, not `$_foo`.
+            if (gflag && *++s != '-') {
+                while (*s != '\0') {
+                    char name = *s++;
+                    if (name != '_' && !ISALPHA(name)) {
```
nobu, line 849: `ruby -g -e 'p $α' -- -α` disallowed?

ruby.c
```diff
@@ -829,12 +841,44 @@ process_sflag(int sflag)
...
+                        VALUE name_error[2];
+                        name_error[0] = rb_str_new2("invalid name for global variable - ");
+                        char error_flag[3] = { '-', name, '\0' };
```
nobu, line 852: This can make incomplete multibyte string.

nobu:
Clear unexpected reverting changes.

ruby.c
sampersand: I like this idea

ruby.c
```diff
@@ -807,13 +808,24 @@ toplevel_context(rb_binding_t *bind)
     return &bind->block;
 }
 
+static VALUE
+parse_gflag_equals(const char *src)
+{
+        if (ISDIGIT(*src)) return rb_cstr_to_inum(src, 10, 0);
```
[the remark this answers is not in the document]
sampersand: Whoops! good point!

internal/cmdlineopt.h
```diff
@@ -30,7 +30,7 @@ typedef struct ruby_cmdline_options {
 
     const char *crash_report;
 
-    signed int sflag: 2;
+    signed int sflag: 3;
```
nobu, line 33: 2bits will be enough for -1..2.

sampersand: Aren't the bounds of a signed 2 bit integer `-2` thru `1`?
- `00` is `0`
- `01` is `1`
- `10` is `-2`
- `11` is `-1`

Granted, I could rework the code to use these bounds instead, or make it unsigned

nobu: Ah, yes, you're right.

ruby.c
sampersand: What's a good way to check for this? I could just do the remainder of the string, but that'd also bring in any other characters

PR #12528, opened 2025-01-08
Avoid opt_aset_with optimization inside multiple assignment

jeremyevans:
Previously, since the opt_aset_with optimization was introduced, use of the opt_aset_with optimization inside multiple assignment would result in a segfault or incorrect instructions.

Fixes [Bug #21012]

### DIFF
--- a/test/ruby/test_assignment.rb
+++ b/test/ruby/test_assignment.rb
@@ -248,6 +248,17 @@ class TestAssignment < Test::Unit::TestCase
     a,b,*c = *[*[1,2]]; assert_equal([1,2,[]], [a,b,c])
   end
 
+  def test_massign_optimized_literal_bug_21012
+    singleton_class.attr_reader :a
+    @a = []
+    def (@a).[]=(*args)
+      push args
+    end
+    a["a", "b"], = 1
+    a["a", 10], = 2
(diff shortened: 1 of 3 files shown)


test/ruby/test_assignment.rb
```diff
@@ -248,6 +248,17 @@ def test_massign_splat
     a,b,*c = *[*[1,2]]; assert_equal([1,2,[]], [a,b,c])
   end
 
+  def test_massign_optimized_literal_bug_21012
+    singleton_class.attr_reader :a
+    @a = []
+    def (@a).[]=(*args)
```
nobu, line 254: The kind of the receiver seems not essential.

```suggestion
    a = []
    def a.[]=(*args)
```

test/ruby/test_assignment.rb
```diff
@@ -248,6 +248,17 @@ def test_massign_splat
...
+      push args
+    end
+    a["a", "b"], = 1
+    a["a", 10], = 2
+    assert_equal [["a", "b", 1], ["a", 10, 2]], @a
```
nobu, line 259:
```suggestion
    assert_equal [["a", "b", 1], ["a", 10, 2]], a
```

nobu (approved):

PR #12535, opened 2025-01-08
coroutine/ppc64le: fix conditional registers got clobbered unexpectedly

liushuyu:
According to the [POWER ISA 64-bit ELF ABI Specification](https://openpowerfoundation.org/specifications/64bitelfabi/), section 2.2.2, page 27, special registers `cr2`, `cr3`, and `cr4` are non-volatile and MUST be preserved between function calls.

This pull request fixes this function called sequence violation by saving and restoring those registers at the Ruby fibre boundaries.


liushuyu:
CI failures are unrelated to this pull request: CI is already broken on the main branch.

PR #12539, opened 2025-01-09
[Feature #6012] Extend `source_location` for end position and columns

nobu:


eregon:
FYI I have found 2 issues with the current implementation:
https://bugs.ruby-lang.org/issues/21783
https://bugs.ruby-lang.org/issues/21784

PR #12551, opened 2025-01-11
[Bug #21024] <cstdbool> header is deprecated in C++17

nobu:
[[Bug #21024]](https://bugs.ruby-lang.org/issues/21024)

### DIFF
--- a/include/ruby/internal/stdbool.h
+++ b/include/ruby/internal/stdbool.h
@@ -27,7 +27,7 @@
 
 #elif defined(__cplusplus)
 # /* bool is a keyword in C++. */
-# if defined(HAVE_STDBOOL_H) && (__cplusplus >= 201103L)
+# if defined(HAVE_STDBOOL_H) && (__cplusplus >= 201103L) && (__cplusplus < 201703L)
 #  include <cstdbool>
 # endif
 #


include/ruby/internal/stdbool.h
```diff
@@ -27,7 +27,7 @@
 
 #elif defined(__cplusplus)
 # /* bool is a keyword in C++. */
-# if defined(HAVE_STDBOOL_H) && (__cplusplus >= 201103L)
+# if defined(HAVE_STDBOOL_H) && (__cplusplus >= 201103L) && (__cplusplus < 201703L)
```
jwakely, line 30: Why include it for C++11 and C++14? It's just as useless in C++11 as it is in C++17 and later.

junaruga: If you are taking about nobu's comment, "That patch will break C++-11 and C++-14." at https://bugs.ruby-lang.org/issues/21024#note-1, the "That patch" means the `https://bugs.ruby-lang.org/issues/21024#note-1` attached on the https://bugs.ruby-lang.org/issues/21024, it's not for this PR.

jwakely: I'm talking about t his commit. The committed change is to not include `<cstdbool>` for C++17 and later. But that means it's still included for C++11 and C++14. Why?

jwakely: IMHO the attachment on https://bugs.ruby-lang.org/issues/21024 is a better fix, and I don't see why (or how) it will break anything for C++11 and C++14.

nobu: I understand that `<cstdbool>` is deprecated since C++17, but not before it.
Is the header deprecated since C++11?

jwakely: It's deprecated in C++17 but it's useless in C++11 and serves no purpose.

Why does this code bother including it for C++11 and C++14?

Just because something is not deprecated doesn't mean it's useful. There are many other standard C++ headers which are not deprecated in C++11, should they also be included here?

jwakely: In https://bugs.ruby-lang.org/issues/21024 you claimed that removing `#include <cstdbool>` unconditionally would break C++11 and C++14. Could you explain why?

If a header is useless, it can be removed. It doesn't need to be deprecated for that to be true.

nobu: When someone tells you simply that a feature has been deprecated, don't you usually assume that it used to do something?

jwakely: If it's deprecated, I assume it still does exactly what it used to do, but is no longer recommended. That's what deprecated means.

In this case, it still does the same in C++17 as it did in C++11, which is nothing. If it's safe to remove in C++17 mode, then it's also safe to remove in C++11.

It was always a useless header and was never needed. This is your opportunity to stop including it! Conditionally including it for some versions of C++ and not others just makes the code unnecessarily complicated and will cause people to wonder why it was done that way. Is there actually any rationale for keeping it?

nobu: > In this case, it still does the same in C++17 as it did in C++11, which is nothing.

Thank you, that makes sense.

junaruga: For your information, I tried to find the context when the `<cstdbool>` header was added. It seems the line `#  include <cstdbool>` was added at the commit https://github.com/ruby/ruby/commit/9e6e39c3512f7a962c44dc3729c98a0f8be90341#diff-ed5cbb2924263dc20d70608d979ebf9404f4f4575c173ba2871e9c288c1f5224R29 (big change). But I couldn't find the context.

junaruga: FYI: nobu sent the following PR, and merged.
https://github.com/ruby/ruby/pull/12573

junaruga: jwakely, thank you for your comments!

PR #12552, opened 2025-01-11
[Doc] Encourage use of encoding constants

byroot:
Lots of documentation examples still use encoding APIs with encoding names rather than encoding constants. I think it would be preferable to direct users toward constants as it can help with auto-completion, static analysis and such.


BurdetteLamar (approved):

BurdetteLamar:
Thanks, @byroot.  I will propose an addition to the doc guide.

PR #12553, opened 2025-01-11
Make rb_tracearg_(parameters|eval_script|instruction_sequence) public C-API

richardboehme:
This allows C-Extension developers to call those methods to retrieve information about a TracePoint's parameters, eval script and instruction sequence.

Implements [Feature #20757]

### DIFF
--- a/include/ruby/debug.h
+++ b/include/ruby/debug.h
@@ -520,6 +520,18 @@ VALUE rb_tracearg_path(rb_trace_arg_t *trace_arg);
 
 RBIMPL_ATTR_NONNULL(())
 /**
+ *
+ * Queries the parameters passed on a call or return event.
+ *
+ * @param[in]  trace_arg         A trace instance
+ * @exception  rb_eRuntimeError  The tracing event does not support querying parameters.
+ * @return     Array of parameters in the format of `Method#parameters`.
+ *
+ */
+VALUE rb_tracearg_parameters(rb_trace_arg_t *trace_arg);
+
+RBIMPL_ATTR_NONNULL(())
+/**
  * Queries the method name of the point where the trace is at.
  *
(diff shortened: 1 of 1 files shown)


include/ruby/debug.h
```diff
@@ -595,6 +607,31 @@ RBIMPL_ATTR_NONNULL(())
...
+ * @retval     RUBY_Qnil         The script was loaded from a file.
+ * @retval     otherwise         The compiled source code.
+ *
+ */
+VALUE rb_tracearg_eval_script(rb_trace_arg_t *trace_arg);
+
+RBIMPL_ATTR_NONNULL(())
+/**
+ *
+ * Queries the compiled instruction sequence on a 'script_compiled' event.
```
ko1, line 626: Could you add it is MRI specific?
https://ruby-doc.org/core-2.6/TracePoint.html#method-i-instruction_sequence

nobu:
Shouldn't `RBIMPL_ATTR_NONNULL(())` be `RBIMPL_ATTR_NONNULL((1))`, including existing ones?

include/ruby/debug.h
richardboehme: I added the note from the docs you mentioned. Thanks for finding this!

richardboehme:
> Shouldn't `RBIMPL_ATTR_NONNULL(())` be `RBIMPL_ATTR_NONNULL((1))`, including existing ones?

As a non-C developer I'm not *entirely* sure but after reading about it makes sense. Each of these functions would crash when passed a null pointer so marking the argument as non-null seems to be correct. I added a separate commit to make this change.

richardboehme:
@ko1 Thanks for reviewing this! I'm sorry to ping, but is there anything I can do to move this PR forward and get it merged?

ko1:
thank you!

PR #12555, opened 2025-01-12
Stop passing `-t` and `-v` on ripper build

yui-knk:
Both of them are debug option.
Let's use `YFLAGS` for parse.y build if needed.

### DIFF
--- a/ext/ripper/depend
+++ b/ext/ripper/depend
@@ -12,7 +12,7 @@ ripper.o: ripper.c
 
 .y.c:
 	$(ECHO) compiling compiler $<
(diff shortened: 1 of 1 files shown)


nobu:
Isn't a chance to want to see the source for ripper?

yui-knk:
What source means? "ripper.y" is still created.
`-t` is same with `-Dparse.trace`. `-v` generates "ripper.output" file. Both of them are not related `.c` and `.h` files generated by Lrama, JFYI.

nobu (approved):
I see, ripper.output should not be needed as parse.output is available.

PR #12556, opened 2025-01-12
Use LRAMA instead of BISON, YACC

ydah:
Currently using lrama so we use LRAMA instead of BISON, YACC.


yui-knk:
I don't have strong opinion about the variable name, for example common.mk uses `YACC`.
@nobu I want to know your opinion.

nobu:
I don't have strong opinion too.
If you want to change `YACC` as well, note that it is used also in template/Makefile.in, win32/Makefile.sub and tool/make-snapshot, not only in common.mk.

simi:
Or maybe something generic like `PARSER` or `LALR_PARSER`?

ydah:
I changed it to `PARSER`. How does it look?

makenowjust:
`PARSER` looks very strange to me because `lrama` is not a parser, it is just a parser generator. `LRAMA` sounds best in this case.

simi:
> `PARSER` looks very strange to me because `lrama` is not a parser, it is just a parser generator. `LRAMA` sounds best in this case.

My bad, maybe GENERATOR would fit more generic needs. But LRAMA looks good also, since it seems there is no plan to change it soon.

ydah:
> `PARSER` looks very strange to me because `lrama` is not a parser, it is just a parser generator. `LRAMA` sounds best in this case.

Ah, Indeed it is. Thank you.

nobu:
> > `PARSER` looks very strange to me because `lrama` is not a parser, it is just a parser generator. `LRAMA` sounds best in this case.
> 
> My bad, maybe GENERATOR would fit more generic needs. But LRAMA looks good also, since it seems there is no plan to change it soon.

As `YACC` stands for "Yet Another CC (Compiler-Compiler)", I think `COMPILER_COMPILER` (or maybe `COMPILER_GENERATOR`) would be more appropriate than mere `GENERATOR`, but too long obviously.

Anyway, `LRAMA` looks fine to me.

PR #12560, opened 2025-01-12
[DOC] Hash defaults doc

BurdetteLamar:
Emphasis on what the user can do.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -2098,7 +2098,7 @@ rb_hash_stlike_lookup(VALUE hash, st_data_t key, st_data_t *pval)
  *    h[:foo] # => 0
  *
  *  If +key+ is not found, returns a default value
- *  (see {Default Values}[rdoc-ref:Hash@Default+Values]):
+ *  (see {Hash Default}[rdoc-ref:Hash@Hash+Default]):
  *    h = {foo: 0, bar: 1, baz: 2}
  *    h[:nosuch] # => nil
  */
@@ -2207,7 +2207,7 @@ rb_hash_fetch(VALUE hash, VALUE key)
  *
  *  Returns the default value for the given +key+.
  *  The returned value will be determined either by the default proc or by the default value.
- *  See {Default Values}[rdoc-ref:Hash@Default+Values].
+ *  See {Hash Default}[rdoc-ref:Hash@Hash+Default].
  *
  *  With no argument, returns the current default value:
  *    h = {}
@@ -2244,7 +2244,7 @@ rb_hash_default(int argc, VALUE *argv, VALUE hash)
  *    h.default = false # => false
  *    h.default # => false
  *
- *  See {Default Values}[rdoc-ref:Hash@Default+Values].
+ *  See {Hash Default}[rdoc-ref:Hash@Hash+Default].
  */
 
 static VALUE
@@ -2260,7 +2260,7 @@ rb_hash_set_default(VALUE hash, VALUE ifnone)
  *    hash.default_proc -> proc or nil
  *
  *  Returns the default proc for +self+
- *  (see {Default Values}[rdoc-ref:Hash@Default+Values]):
+ *  (see {Hash Default}[rdoc-ref:Hash@Hash+Default]):
  *    h = {}
  *    h.default_proc # => nil
  *    h.default_proc = proc {|hash, key| "Default value for #{key}" }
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
Marked as draft, b/c it may break links from elsewhere -- in particular, from the 10 related Hash methods (all mentioned here).  I'd like to update all those in this PR, but if that's too big I will make the links good and do the rest later.  @peterzhu2118, what do you prefer?

hash.c
```diff
@@ -6881,93 +6881,161 @@ static const rb_data_type_t env_data_type = {
...
+ *    h.fetch(:nosuch) {|key| "No key #{key}"}
+ *    # => "No key nosuch"
+ *    h.fetch(:nosuch)        # Raises KeyError.
  *
- *  Be aware that a default proc that modifies the hash is not thread-safe in the
- *  sense that multiple threads can call into the default proc concurrently for the
- *  same key.
+ *    h.fetch_values(:bar, :nosuch, :foo) {|key| "No key #{key}"}
+ *    # => [1, "No key nosuch", 0]
+ *    h.fetch_values(:bar, :nosuch, :foo) # Raises KeyError.
```
peterzhu2118, line 7038: IMO these sections have too much details about the methods themselves, which should belong in the method documentation rather than the class documentation. Class documentation should be about high level information, rather than the details about specific methods.

hash.c
```diff
@@ -6881,93 +6881,161 @@ static const rb_data_type_t env_data_type = {
  *
  *    reviews.length #=> 1
  *
- *  === Default Values
+ *  === Key Not Found?
```
peterzhu2118, line 6884: I find a question in a title a bit odd. How about:

```suggestion
 *  === Nonexistent keys
```

hash.c
```diff
@@ -6881,93 +6881,161 @@ static const rb_data_type_t env_data_type = {
...
  *
- *  The methods #[], #values_at and #dig need to return the value associated to a certain key.
- *  When that key is not found, that value will be determined by its default proc (if any)
- *  or else its default (initially `nil`).
+ *  When a method tries to retrieve and return the value for a key and that key <i>is found</i>,
+ *  the returned value is the value associated with the key.
  *
- *  You can retrieve the default value with method #default:
+ *  But what if the key <i>is not found</i>?
+ *  In that case, you have a wide choice of behaviors.
```
peterzhu2118, line 6890: Is there a wide choice of behaviors? There seems to only be two: default value or raise KeyError.

hash.c
BurdetteLamar: > I find a question in a title a bit odd. How about:

I like this title;  it captures the spirit of what I intend for the section.  So, stet?

hash.c
BurdetteLamar: New commit abbreviates the modified section, but does not (yet) make good the links in the 11 cited methods' doc; so still a draft.

hash.c
```diff
@@ -6881,93 +6881,92 @@ static const rb_data_type_t env_data_type = {
...
- *    synonyms[:world] << :universe
- *    synonyms[:world] # => [:hi, :universe], oops
- *    synonyms.keys # => [], oops
+ *  - #fetch(key).
+ *  - #fetch_values(*keys).
  *
- *  To use a mutable object as default, it is recommended to use a default proc
+ *  You can override these behaviors;
+ *  see {Hash Default}[rdoc-ref:Hash@Hash+Default]
+ *  and {Method Default}[rdoc-ref:Hash@Method+Default].
```
peterzhu2118, line 6913: I don't think this is true:

```ruby
h = {}
h.fetch(:blah) # => key not found: :blah (KeyError)
h.default = "hi"
h.fetch(:blah) # => key not found: :blah (KeyError)
```

BurdetteLamar:
@peterzhu2118, links in methods' doc now made good.

hash.c
peterzhu2118: @BurdetteLamar did you miss this comment?

BurdetteLamar: Or anything else I may define in a block?

peterzhu2118: But there's really only KeyError or default. AFAIK `nil` is just a special case of the default where the default is nil. Is there a method that returns nil even if `Hash#default` is not?

BurdetteLamar: What do you want this to say?

peterzhu2118: Maybe something like?

```suggestion
In that case, certain methods will return a default value while other will raise a \KeyError.
```

peterzhu2118 (approved):

PR #12562, opened 2025-01-12
rb_feature_p: skip `get_expanded_load_path` for absolute paths

byroot:
Ref: https://github.com/fxn/zeitwerk/pull/308

```ruby
require 'benchmark'

$LOAD_PATH << 'relative-path'

autoload :FOO, '/tmp/foo.rb'

puts Benchmark.realtime {
  500_000.times do
    Object.autoload?(:FOO)
  end
}
```

The above script takes 2.5 seconds on `master`, and only
50ms on this branch.

When we're looking for a feature with an absolute path, we don't
need to call the expensive `get_expanded_load_path`.

### DIFF
--- a/load.c
+++ b/load.c
@@ -597,7 +597,7 @@ rb_feature_p(rb_vm_t *vm, const char *feature, const char *ext, int rb, int expa
 
     loading_tbl = get_loading_table(vm);
     f = 0;
-    if (!expanded) {
+    if (!expanded && !rb_is_absolute_path(feature)) {
         struct loaded_feature_searching fs;
         fs.name = feature;
         fs.len = len;


fxn:
Since the thread in the linked issue is long, let me summarize the core issue.

In my laptop, this script takes 0.6s:

```ruby
require 'benchmark'

$LOAD_PATH << 'whatever' # (1)

autoload :FOO, '/tmp/foo'

puts Benchmark.realtime {
  100_000.times do
    Object.autoload?(:FOO)
  end
}
```

but if you uncomment the mutation of `$LOAD_PATH` pushing a relative path in (1), you get instead 0.01s. That is a 60x factor, and you'd argue that since the path passed to `autoload` is absolute, the contents of `$LOAD_PATH` should make no difference.

byroot:
Seems like all tests are passing, and it totally solves the perf issue reported on Zeitwerk.

So I think I can clean up the PR tomorrow and merge it, however if we want it backported, we'll have to report it as a bug on Redmine. Worth a try, but not sure if everyone will see it as a bug fix versus an optimization.

load.c
```diff
@@ -597,7 +597,7 @@ rb_feature_p(rb_vm_t *vm, const char *feature, const char *ext, int rb, int expa
 
     loading_tbl = get_loading_table(vm);
     f = 0;
-    if (!expanded) {
+    if (!expanded && !rb_is_absolute_path(feature)) {
```
nobu, line 600: What occurs when `feature` is an absolute path but not a real path, i.e., symlink?

fxn:
I just created this ticket https://bugs.ruby-lang.org/issues/21035, since I don't see this validation logic in `autoload?` to be necessary or intuitive.

load.c
byroot: I don't think it change anything? At least on profiles I can't see Ruby calling `realpath` even before that change, but maybe I missed it?

fxn: According to tests I did a while back:

* `Kernel#require` seems to base idempotence on real paths.
* `$LOADED_FEATURES` shows what the user entered.
  * If it is an absolute path, it remains as it is verbatim.
  * If it was `require "foo"`, the `foo` itself remains verbatim (mod extension), and the load path directory that acts as prefix appears as its real path.

I don't know if that logic applies here too, sharing in case it does.

nobu: My words were not enough.
When a library has been loaded by `$LOAD_PATH`, requiring a symlink to that path will do nothing now.
I think this PR will change it to load the same file again, no?

byroot: @nobu sorry but I'm really not sure what you have in mind, perhaps you could provide a test script?

I tried a few things:

```ruby
Dir.mkdir("/tmp/original") rescue nil
File.write("/tmp/original/a.rb", "p :loaded_original_a")

File.symlink("/tmp/original", "/tmp/symlink") rescue nil
$LOAD_PATH << "/tmp/original"

require "a"

$LOAD_PATH << "/tmp/original/symlink"

require "a"
```

```ruby
File.write("/tmp/a.rb", "p :loaded_a")
File.symlink("/tmp/a.rb", "/tmp/b.rb")
$LOAD_PATH << "/tmp"

require "a"
require "b"
```

None changed.

nobu: The case `require "a"` then `require "/tmp/original/symlink/a"`.
In the latter `require`, the path will be expanded to the original path, and not be loaded again, for now.
This PR skips that expansion, and the file will be loaded again, no?

byroot: I just tried:

```ruby
Dir.mkdir("/tmp/original") rescue nil
File.write("/tmp/original/a.rb", "p :loaded_original_a")

File.symlink("/tmp/original", "/tmp/symlink") rescue nil
$LOAD_PATH << "/tmp/original"

require "a"

$LOAD_PATH << "/tmp/symlink"

require "/tmp/symlink/a"
```

And the file is still only loaded once.

PR #12565, opened 2025-01-13
Array#fetch_values: expand ranges

byroot:
[Feature #20953]

This tries to implement the range expansion like desired by Matz.

However I don't know how to handle some edge cases such as endless ranges.

### DIFF
--- a/array.c
+++ b/array.c
@@ -1996,6 +1996,23 @@ rb_ary_last(int argc, const VALUE *argv, VALUE ary) // used by parse.y
     }
 }
 
+static VALUE
+ary_fetch(VALUE ary, long idx, VALUE ifnone, bool block_given)
+{
+    if (idx < 0) {
+        idx +=  RARRAY_LEN(ary);
+    }
+    if (idx < 0 || RARRAY_LEN(ary) <= idx) {
+        if (block_given) return rb_yield(LONG2NUM(idx));
+        if (UNDEF_P(ifnone)) {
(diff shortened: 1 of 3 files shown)


array.c
```diff
@@ -3855,6 +3873,82 @@ rb_ary_values_at(int argc, VALUE *argv, VALUE ary)
...
+rb_ary_fetch_values(int argc, VALUE *argv, VALUE ary)
+{
+    bool block_given = rb_block_given_p();
+
+    long i, olen = RARRAY_LEN(ary);
+    VALUE result = rb_ary_new_capa(argc);
+    for (i = 0; i < argc; ++i) {
+        append_fetch_values_single(result, ary, olen, argv[i], block_given);
+    }
+    return RB_GC_GUARD(result);
```
nobu, line 3949: `result` is the return value, which shouldn't be optimized away.

array.c
```diff
@@ -1996,6 +1997,23 @@ rb_ary_last(int argc, const VALUE *argv, VALUE ary) // used by parse.y
     }
 }
 
+static VALUE
+ary_fetch(VALUE ary, long idx, VALUE ifnone, bool block_given)
```
nobu, line 2001: Why is this function located away from its only caller, `append_fetch_values_single`?
Maybe you forgot to update `rb_ary_fetch` to use it too?

nobu:
Windows 2022 failure is the VC's bug.

array.c
byroot: Fixed.

PR #12566, opened 2025-01-13
Remove dead function rb_func_proc_new

peterzhu2118:


peterzhu2118:
The last reference I found was removed in 241244739f2b721ac7aa0961bd90d904c5e3fff6.

> but no chance to use it for other functions in the future?

I guess this can be reverted if it's needed in the future?

PR #12570, opened 2025-01-14
Restore lockfile for test-bundler

hsbt:
In the past, bundler updated platform without normalized variable like arm64-darwin-23. We ignored that update. But the current bundler uses arm64-darwin for that.


hsbt:
Thanks, I fixed up that at https://github.com/ruby/ruby/commit/c8d11edcf556088d10f0a245139741dcab1c7d56

PR #12572, opened 2025-01-14
Simplify gc/mmtk/extconf.rb

nobu:
- Split static recipes to depend file.
- Modify makefile configurations in the block to `create_makefile`.
- Expand rust sources in extconf.rb instead of GNU make extension.

TODO: pass `CARGO_TARGET_DIR` without shell syntax.


peterzhu2118 (approved):
Thank you!

PR #12589, opened 2025-01-16
Retire ubuntu-20.04

hsbt:
https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

GitHub have a plan to brownout `ubuntu-20.04` in Mar 2025. I don't want to see unintentional failure with that. So, we should remove `ubuntu-20.04` for Ruby 3.5 development.

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -37,8 +37,6 @@ jobs:
             timeout: 50
           - test_task: test-bundled-gems
           - test_task: check
-            os: ubuntu-20.04
-          - test_task: check
             os: ubuntu-24.04
           - test_task: check
             os: ubuntu-24.04-arm
--- a/.github/workflows/baseruby.yml
+++ b/.github/workflows/baseruby.yml
@@ -30,7 +30,7 @@ jobs:
   baseruby:
     name: BASERUBY
 
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
 
     if: >-
       ${{!(false
--- a/.github/workflows/rjit-bindgen.yml
+++ b/.github/workflows/rjit-bindgen.yml
@@ -33,7 +33,7 @@ jobs:
           - task: rjit-bindgen
       fail-fast: false
 
-    runs-on: ubuntu-20.04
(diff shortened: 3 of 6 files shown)


hsbt:
@ruby/yjit I changed `clang-12` to `clang-13` because `ubuntu-22.04` runner didn't have `clang-12`. But It's not working.

Does anyone look https://github.com/ruby/ruby/actions/runs/12804030074/job/35697714010?

maximecb:
> @ruby/yjit I changed `clang-12` to `clang-13` because `ubuntu-22.04` runner didn't have `clang-12`. But It's not working.
> 
> Does anyone look https://github.com/ruby/ruby/actions/runs/12804030074/job/35697714010?

Can you try a more recent version of clang?

hsbt:
I already tried with `clang-15`, but it has same issues. Should we use `ubuntu-24.04` and `clang-16` (or 17 or 18)?

XrXr:
[209a681](https://github.com/ruby/ruby/pull/12589/commits/209a681afef1471a09ed06faf0de3329e1d729e2) should (knock on wood) get it to work.

.github/workflows/ubuntu.yml
```diff
@@ -36,8 +36,6 @@ jobs:
           - test_task: test-bundler-parallel
             timeout: 50
           - test_task: test-bundled-gems
-          - test_task: check
-            os: ubuntu-20.04
```
nobu, line 40: Why not test on ubuntu-22.04?

maximecb:
> [b42f5ac](https://github.com/ruby/ruby/pull/12589/commits/b42f5acf68185af0688bbae648f13ba9ccd70eb7) should (knock on wood) get it to work.

Thank you Alan :)

maximecb (approved):

.github/workflows/ubuntu.yml
hsbt: We tested `make check` with default `runs-on` variable like `runs-on: ${{ matrix.os || 'ubuntu-22.04' }}`.

PR #12596, opened 2025-01-19
Avoid allocation for anonymous positional splat with no arguments

jeremyevans:
Anonymous positional splats cannot be directly accessed, they can only be passed as splats to other methods.  So if an anonymous positional splat would be empty, you can use a shared frozen empty array to save an allocation.

```ruby
def a(*) end
a()
```

This is similar to how anonymous empty keyword splats are optimized, except those use `nil` instead of a shared empty frozen hash.

This updates the allocation tests to check that the array allocations are avoided where possible.

It also makes a small change to test_iseq.rb to ensure an unfrozen hash is passed as the value of an anonymous splat parameter.

### DIFF
--- a/test/ruby/test_allocation.rb
+++ b/test/ruby/test_allocation.rb
@@ -524,6 +524,7 @@ class TestAllocation < Test::Unit::TestCase
     end
 
     def test_anonymous_splat_and_anonymous_keyword_splat_parameters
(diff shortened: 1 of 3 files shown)


peterzhu2118 (approved):

casperisfine:
This made some assertion fail: https://github.com/ruby/ruby/pull/13043

PR #12597, opened 2025-01-19
Avoid array allocation for *nil, by not calling nil.to_a

jeremyevans:
A method call such as `a(*nil)` previously allocated an array, because it calls `nil.to_a`, but I have determined this array allocation is unnecessary.  The instructions in this case are:

```
0000 putself                                                          (   1)[Li]
0001 putnil
0002 splatarray                             false
0004 opt_send_without_block                 <calldata!mid:a, argc:1, ARGS_SPLAT|FCALL>
0006 leave
```

The method call uses `ARGS_SPLAT` without `ARGS_SPLAT_MUT`, so the returned array doesn't need to be mutable.  I believe all cases where `splatarray false` are used allow the returned object to be frozen, since the `false` means to not duplicate the array.  The optimization in this case is to have `splatarray false` push a shared empty frozen array, instead of calling `nil.to_a` to return a newly allocated array.

There is a slightly backwards incompatibility with this optimization, in that `nil.to_a` is not called.  However, I believe the new behavior of `*nil` not calling `nil.to_a` is more consistent with how `**nil` does not call `nil.to_hash`.  Also, so much Ruby code would break if `nil.to_a` returned something different from the empty hash, that it's difficult to imagine anyone actually doing that in real code, though we have a few tests/specs for that.

I think it would be bad for consistency if `*nil` called `nil.to_a` in some cases and not others, so this changes other cases to not call `nil.to_a`:

For `[*nil]`, this uses `splatarray true`, which now allocates a new array for a `nil` argument without calling `nil.to_a`.

For `[1, *nil]`, this uses `concattoarray`, which now returns the first array if the second array is `nil`.

This updates the allocation tests to check that the array allocations are avoided where possible.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,9 @@ Note that each entry is kept to a minimum, see links for details.
 
 ## Language changes
 
+* `*nil` no longer calls `nil.to_a`, similar to how `**nil` does
+  not call `nil.to_hash`.  [[Feature #21047]]
+
 ## Core classes updates
 
 Note: We're only listing outstanding class updates.
@@ -84,4 +87,5 @@ The following bundled gems are updated.
 ## JIT
 
 [Feature #19908]: https://bugs.ruby-lang.org/issues/19908
+[Feature #21047]: https://bugs.ruby-lang.org/issues/21047
 [Bug #21049]:     https://bugs.ruby-lang.org/issues/21049
--- a/benchmark/nilclass.yml
+++ b/benchmark/nilclass.yml
@@ -1,6 +1,10 @@
+prelude: |
+  def a = nil
 benchmark:
   to_i: |
     nil.to_i
   to_f: |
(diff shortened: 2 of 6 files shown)


ko1:
Is this PR for performance? Please show us the benchmark results.
Anyway, I didn't know the details around this calling flag, so I believe @jeremyevans

jeremyevans:
This isn't just for performance, though it improves performance for the `*nil` case as it avoids a method call and avoids an array allocation. It's also for consistency with `**nil`.  The behavior change has been approved by Matz already (https://bugs.ruby-lang.org/issues/21047). 

I've pushed a benchmark, it shows a 4.45x speedup:

```
                            splat
      miniruby-after:  23202704.8 i/s 
     miniruby-before:   5213879.6 i/s - 4.45x  slower
```

PR #12598, opened 2025-01-20
Sync json

hsbt:


hsbt:
@byroot /cc @soutaro `make test-bundled-gems BUNDLED_GEMS=rbs` is stuck with this branch. Can you look this?

soutaro:
I found `RBS::SchemaTest` causes the issue. Adding the test to `rbs_skip_tests` would be a reasonable workaround.

I also confirmed `json-schema` gem doesn't work with the HEAD version of `json` gem. (However, I'm skeptical if anyone is using the JSON representation of RBS. Simply dropping the feature from `rbs` gem would be a solution.)

hsbt:
@soutaro Thank you for your investigation. I skipped `RBS::SchemaTest` in this PR.

byroot:
Fixed in https://github.com/ruby/json/commit/f8cfa2696a658f362f3c4700e7ef4af7450ebc0e.

You should be able to restore the functionality you disabled.

PR #12600, opened 2025-01-20
Restore rbs schema test

byroot:
Now that the root cause has been fixed, we can resume `BS::SchemaTest` testing.

Followup: https://github.com/ruby/ruby/pull/12598


hsbt (approved):
👍 Thanks for your quick fix.

PR #12604, opened 2025-01-20
.travis.yml: Drop arm64 case.

junaruga:
We started running a Linux arm64 case in GitHub Actions.[1] So, it's time to drop the Linux arm64 case in Travis CI.

[1] https://github.com/ruby/ruby/commit/a9d37ac3e5385c7aaec64e32de3d254903f4b5c0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,10 +55,6 @@ env:
         openssl
         zlib1g-dev
       - gcc-11 --version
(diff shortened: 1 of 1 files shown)


junaruga:
Travis CI passed for this PR.
https://app.travis-ci.com/github/ruby/ruby/builds/273914286

PR #12611, opened 2025-01-22
[DOC] Add a mention to git 2.32 in the building ruby documentation:

Edouard-chin:
Hello 👋 !  

Not sure if there is interest in tweaking the requirement recommendation for building Ruby but
the project now makes use of a feature added in git 2.32 (`GIT_CONFIG_SYSTEM` and `GIT_CONFIG_GLOBAL`). 

I had issues building Ruby as I was using git 2.30.
The error was:

> file2lastrev.rb: git rev-parse failed error: could not expand
> include path '~/.gitcinclude'

The default system git config includes paths making use of `$HOME`, but that env variable gets removed when building: https://github.com/ruby/ruby/blob/7070b1b1968f31f999636b42a1af872458cd5a81/tool/lib/vcs.rb#L546-L547

It works with git `>= 2.32` thanks to the `GIT_CONFIG_SYSTEM` feature which gets set to a Null IO in order to override the system gitconfig. https://github.com/ruby/ruby/blob/7070b1b1968f31f999636b42a1af872458cd5a81/tool/lib/vcs.rb#L549

I thought this could save other users from having to debug the problem.

Thanks!

### DIFF
--- a/doc/contributing/building_ruby.md
+++ b/doc/contributing/building_ruby.md
@@ -20,6 +20,9 @@
     * ruby - 3.0 or later
         * We can upgrade this version to system ruby version of the latest
           Ubuntu LTS.
+    * git - 2.32 or later
+        * Anterior versions may work; 2.32 or later will prevent build
+          errors in case your system `.gitconfig` uses `$HOME` paths.
 
 2. Install optional, recommended dependencies:
 


nobu:
Thank you for the report.

> I had issues building Ruby as I was using git 2.30. The error was:
> 
> > file2lastrev.rb: git rev-parse failed error: could not expand
> > include path '~/.gitcinclude'

Where does this path come from, your system's config file?

Edouard-chin:
Yes that's correct, the system's config (`/usr/local/git/etc/gitconfig`) file which gets generated by default. There are two paths that makes uses oh `$HOME`:

> [include]	# as of 1.7.10 https://github.com/git/git/commit/9b25a0b52e09400719366f0a33d0d0da98bbf7b0
	path = ~/.gitcinclude
	path = .githubconfig
	path = .gitcredential

> [core]
	legacyheaders = false # >git 1.5
	quotepath = false
	excludesfile = ~/.gitignore

Edouard-chin:
I was assuming this config was generated by git itself but it seems it's from the git-osx-tool installer (which was one of the [official way](https://git-scm.com/downloads/mac) to install git on macos) https://github.com/timcharper/git_osx_installer/blob/83b50f1923364eb491336dfce74375f65a599498/assets/etc/gitconfig.default

So I suppose this issue may depend on how users install git and whether their system config includes a HOME path (not sure what gets generated when install git from brew).

Anyway maybe this mention in the doc is not necessary and users encountering the problem will now find this issue indexed by Google 😄 . Feel free to close if it's not helpful! Thanks for the review.

nobu:
> I was assuming this config was generated by git itself but it seems it's from the git-osx-tool installer (which was one of the [official way](https://git-scm.com/downloads/mac) to install git on macos) https://github.com/timcharper/git_osx_installer/blob/83b50f1923364eb491336dfce74375f65a599498/assets/etc/gitconfig.default

It seems quite old and not recommended so much.
> Tim Harper provides an [installer](https://sourceforge.net/projects/git-osx-installer/) for Git. The latest version is [2.33.0](https://sourceforge.net/projects/git-osx-installer/files/git-2.33.0-intel-universal-mavericks.dmg/download?use_mirror=autoselect), which was released over 3 years ago, on 2021-08-30.

Edouard-chin:
> It seems quite old and not recommended so much.
> Tim Harper provides an [installer](https://sourceforge.net/projects/git-osx-installer/) for Git. The latest version is [2.33.0](https://sourceforge.net/projects/git-osx-installer/files/git-2.33.0-intel-universal-mavericks.dmg/download?use_mirror=autoselect), which was released over 3 years ago, on 2021-08-30.

Indeed, doesn't seem to be recommended so much anymore 😄. Thanks for checking !

PR #12619, opened 2025-01-23
Allow parsing URIs without throwing exceptions

TobiasBales:
Similar to what `Integer(value, exception: false)` does but for `URI.parse(value, exception: false)`. The change is implemented for the default parser (RFC3986) and also RFC2396.

The goal is to allow parsing invalid/user provided input without having to have control flow via exceptions or by having to wrap URI.parse in a method that catches the exception per project (or as a library).

This change can be seen via
`~/.rubies/ruby-master/bin/ruby -r 'uri' -e
'URI.parse("https://example.com/\[\]", exception: false)'` and `~/.rubies/ruby-master/bin/ruby -r 'uri' -e
'URI.parse("https://example.com/\[\]")'` respectively (or with `exception: true`) and is also covered by specs.

**Question/Failing CI**
The tests for checking the exception behavior fail in the Rubyspec for older versions
since the change is only available on `master` but all previous invocations will work
the same and as expected.

What is the correct way of handling this?
Hopefully not to not write tests for it

### DIFF
--- a/lib/uri/common.rb
+++ b/lib/uri/common.rb
@@ -204,8 +204,8 @@ module URI
   # It's recommended to first ::escape string +uri+
   # if it may contain invalid URI characters.
   #
-  def self.parse(uri)
-    DEFAULT_PARSER.parse(uri)
+  def self.parse(uri, exception: true)
+    DEFAULT_PARSER.parse(uri, exception:)
   end
 
   # Merges the given URI strings +str+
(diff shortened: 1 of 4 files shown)


nobu:
Please send to the upstream https://github.com/ruby/uri.
Note that URI still supports ruby 2.5 or later; keyword argument values cannot be omitted.

TobiasBales:
> Please send to the upstream https://github.com/ruby/uri. Note that URI still supports ruby 2.5 or later; keyword argument values cannot be omitted.

Thank you, I was not aware that it is a different repo and then pulled it - makes a lot of sense.
Will close this and reopen it in that repo

TobiasBales:
The [PR](https://github.com/ruby/uri/pull/149) is open

PR #12621, opened 2025-01-23
[Feature #21086] Declare rb_profile_frames() family as async-signal-safe

osyoyu:
Since 64fef3b870a8ed8147654531aef4c065d8a730c6, it should be safe to call rb_profile_frames() and rb_profile_thread_frames() in a signal handler.


tylerdezainde:
What is this is there any of my files for accepting my payments or more documents

PR #12622, opened 2025-01-23
Fix SystemCallError#message on windows

YO4:
SystemCallError#message was unpredictable if errno is out of range on windows.
Result of SystemCallError.new(-1).message changed from "The operation completed successfully." to "Unknown Error".

Also change specs that stands on the wrong assumption.
SystemCallError.new(-1).message returns
on Ubuntu "Unknown error -1" (glibc)
on Alpine "No error information" (musl libc)
on Windows "Unknown Error" (ruby/win32/win32.c(this patch), ucrt does same)
It depends on the environment.
I didn't want to make too much assumption, so I chose to predict only that it would contain readable characters.

[Bug#21083](https://bugs.ruby-lang.org/issues/21083)

### DIFF
--- a/spec/ruby/core/exception/system_call_error_spec.rb
+++ b/spec/ruby/core/exception/system_call_error_spec.rb
@@ -25,7 +25,7 @@ describe "SystemCallError.new" do
     @example_errno_class = Errno::EINVAL
     @last_known_errno = Errno.constants.size - 1
     @unknown_errno = Errno.constants.size
-    @some_human_readable = "[[:graph:]]+"
+    @some_human_readable = /[[:graph:]]+/
   end
 
   it "requires at least one argument" do
@@ -97,11 +97,11 @@ describe "SystemCallError.new" do
   end
 
   it "sets an 'unknown error' message when an unknown error number" do
-    SystemCallError.new(-1).message.should =~ Regexp.new(@some_human_readable)
+    SystemCallError.new(-1).message.should =~ @some_human_readable
   end
 
   it "adds a custom error message to an 'unknown error' message when an unknown error number and a custom message specified" do
-    SystemCallError.new("custom message", -1).message.should =~ Regexp.new("#{@some_human_readable}.* - custom message")
+    SystemCallError.new("custom message", -1).message.should =~ /#{@some_human_readable}.* - custom message/
   end
 
   it "converts to Integer if errno is a Complex convertible to Integer" do
@@ -142,7 +142,7 @@ describe "SystemCallError#message" do
       SystemCallError.new(2**28).message.should =~ /Error .*occurred/i
     end
     platform_is_not :aix do
-      SystemCallError.new(2**28).message.should =~ Regexp.new(@some_human_readable)
+      SystemCallError.new(2**28).message.should =~ @some_human_readable
     end
   end
 
--- a/spec/ruby/core/exception/system_call_error_spec.rb
+++ b/spec/ruby/core/exception/system_call_error_spec.rb
@@ -25,6 +25,7 @@ describe "SystemCallError.new" do
     @example_errno_class = Errno::EINVAL
     @last_known_errno = Errno.constants.size - 1
     @unknown_errno = Errno.constants.size
+    @some_human_readable = "[[:graph:]]+"
   end
 
   it "requires at least one argument" do
@@ -96,23 +97,11 @@ describe "SystemCallError.new" do
   end
 
   it "sets an 'unknown error' message when an unknown error number" do
-    platform_is_not :windows do
-      SystemCallError.new(-1).message.should =~ /Unknown error(:)? -1/
-    end
-
-    platform_is :windows do
-      SystemCallError.new(-1).message.should == "The operation completed successfully."
-    end
+    SystemCallError.new(-1).message.should =~ Regexp.new(@some_human_readable)
   end
 
   it "adds a custom error message to an 'unknown error' message when an unknown error number and a custom message specified" do
-    platform_is_not :windows do
-      SystemCallError.new("custom message", -1).message.should =~ /Unknown error(:)? -1 - custom message/
-    end
-
-    platform_is :windows do
-      SystemCallError.new("custom message", -1).message.should == "The operation completed successfully. - custom message"
-    end
+    SystemCallError.new("custom message", -1).message.should =~ Regexp.new("#{@some_human_readable}.* - custom message")
   end
 
   it "converts to Integer if errno is a Complex convertible to Integer" do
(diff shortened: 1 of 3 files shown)


spec/ruby/core/exception/system_call_error_spec.rb
```diff
@@ -25,6 +25,7 @@ def initialize
     @example_errno_class = Errno::EINVAL
     @last_known_errno = Errno.constants.size - 1
     @unknown_errno = Errno.constants.size
+    @some_human_readable = "[[:graph:]]+"
```
nobu, line 28: Shouldn't this be a regular expression, as the same as spec/ruby/optional/capi/kernel_spec.rb?
```suggestion
    @some_human_readable = /[[:graph:]]+/
```

spec/ruby/core/exception/system_call_error_spec.rb
```diff
@@ -96,23 +97,11 @@ def initialize
...
 
   it "sets an 'unknown error' message when an unknown error number" do
-    platform_is_not :windows do
-      SystemCallError.new(-1).message.should =~ /Unknown error(:)? -1/
-    end
-
-    platform_is :windows do
-      SystemCallError.new(-1).message.should == "The operation completed successfully."
-    end
+    SystemCallError.new(-1).message.should =~ Regexp.new(@some_human_readable)
```
nobu, line 100:
```suggestion
    SystemCallError.new(-1).message.should =~ @some_human_readable
```

spec/ruby/core/exception/system_call_error_spec.rb
```diff
@@ -153,7 +142,7 @@ def initialize
       SystemCallError.new(2**28).message.should =~ /Error .*occurred/i
     end
     platform_is_not :aix do
-      SystemCallError.new(2**28).message.should =~ /Unknown error/i
+      SystemCallError.new(2**28).message.should =~ Regexp.new(@some_human_readable)
```
nobu, line 145:
```suggestion
      SystemCallError.new(2**28).message.should =~ @some_human_readable
```

spec/ruby/core/exception/system_call_error_spec.rb
```diff
@@ -96,23 +97,11 @@ def initialize
...
 
   it "adds a custom error message to an 'unknown error' message when an unknown error number and a custom message specified" do
-    platform_is_not :windows do
-      SystemCallError.new("custom message", -1).message.should =~ /Unknown error(:)? -1 - custom message/
-    end
-
-    platform_is :windows do
-      SystemCallError.new("custom message", -1).message.should == "The operation completed successfully. - custom message"
-    end
+    SystemCallError.new("custom message", -1).message.should =~ Regexp.new("#{@some_human_readable}.* - custom message")
```
nobu, line 104:
```suggestion
    SystemCallError.new("custom message", -1).message.should =~ /#{@some_human_readable}.* - custom message/
```

spec/ruby/core/exception/system_call_error_spec.rb
```diff
@@ -153,7 +142,7 @@ def initialize
       SystemCallError.new(2**28).message.should =~ /Error .*occurred/i
     end
     platform_is_not :aix do
-      SystemCallError.new(2**28).message.should =~ /Unknown error/i
+      SystemCallError.new(2**28).message.should =~ @some_human_readable
     end
```
nobu, line 146: The AIX condition will not be needed.

win32/win32.c
```diff
@@ -2848,26 +2847,26 @@ rb_w32_strerror(int e)
...
+                    if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM |
+                                      FORMAT_MESSAGE_IGNORE_INSERTS, NULL, e,
+                                      MAKELANGID(LANG_ENGLISH,
+                                              SUBLANG_ENGLISH_US),
+                                      buffer, sizeof(buffer), NULL) != 0 ||
+                        FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM |
+                                      FORMAT_MESSAGE_IGNORE_INSERTS, NULL, e,
+                                      0,
+                                      buffer, sizeof(buffer), NULL) != 0) {
+                        if (p = strpbrk(buffer, "\r\n"))
```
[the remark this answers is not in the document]
YO4: The old code is from c43a23d49e407d8109d4540e6ae7205c261a1f97, says message contains newline from bccwin32 + winsock.
The output of FormatMessage() was terminated with ```\r\n```, so the current code is intended to cut that out.
Only ```\r``` is sufficient, but if using ```FORMAT_MESSAGE_MAX_WIDTH_MASK```, FormatMessage() will not output ```\r\n```, so strpbrk() itself does not seem necessary.

nobu:
Isn't it essentially just removing `if (e < 0) e = GetLastError();` and the next `else`?

YO4:
> Isn't it essentially just removing `if (e < 0) e = GetLastError();` and the next `else`?

The original code was passing the number not in the errmap directly to FormatMessage when ```e > sys_nerr```.
```ruby
irb(main):001> SystemCallError.new(140)
=> #<Errno::EWOULDBLOCK: A non-blocking socket operation could not be completed immediately.>
irb(main):002> SystemCallError.new(141)
=> #<SystemCallError: The system tried to SUBST a drive to a directory on a joined drive.>
```
Therefore, I have made a change to process only when ```e``` is found in the ```errmap```.
```ruby
> miniruby.exe -e "p SystemCallError.new(141)"
#<SystemCallError: Unknown Error>
```

YO4:
To simplify the git history, should I combine them into two commits, one for win32.c and one for specs?

YO4:
I found a mistake that ignoring the necessary errno range in my code and fixed it in 6fba1c8d4e926edc51213b01673a961867dc064f.

Current prev/after error messages are available on gist.
Among these, the following messages were incorrect without this PR.
EBADMSG
ECANCELED
EIDRM
ENODATA
ENOLINK
ENOMSG
ENOSR
ENOSTR
ENOTRECOVERABLE
ENOTSUP
EOVERFLOW
EOWNERDEAD
EPROTO
ETIME
ETXTBSY

nobu:
> Since it is treated as Win32 Error code, it appears to get mixed up with errno across API boundaries. Therefore, I have made a change to process only when `e` is found in the `errmap`. Then the error is correctly treated as an Unknown Error.
> 
> ```ruby
> > miniruby.exe -e "p SystemCallError.new(141)"
> #<SystemCallError: Unknown Error>
> ```

We consider this change somewhat questionable.
It is not included in the original topic, at least.

YO4:
Sorry to bother you. I would concentrate on the case of errno<0.
For win32/win32.c, I went back to the original code and fixed it.

PR #12633, opened 2025-01-24
Get ractor message passing working with > 1 thread sending and receiving values (WIP)

luke-gru:
A ractor can now have multiple threads calling r.take and r.send.

This is a WIP, comments welcome!
Windows builds that use thread_win32 are broken as expected. There's still some more work to get that working.


luke-gru:
Windows is working now, but I can't get RJIT to work. I could have messed up somehow with rjit-bindgen

PR #12634, opened 2025-01-25
[DOC] Adds section "Hash Inclusion"

BurdetteLamar:
New section will provide link target for sub/superset methods.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -3809,8 +3809,7 @@ hash_equal(VALUE hash1, VALUE hash2, int eql)
  *  Returns whether +self+ and +object+ are equal.
  *
  *  Returns +true+ if +object+ is a hash (or can be converted to a hash),
- *  and hashes +self+ and +object+ are equal
- *  (see {Hash Equality and Inclusion}[rdoc-ref:Hash@Hash+Equality+and+Inclusion]);
+ *  and hashes +self+ and +object+ are equal;
  *  otherwise, returns +false+:
  *
  *    h =  {'foo' => 'zero', 'bar' => 'one'}
@@ -3821,6 +3820,9 @@ hash_equal(VALUE hash1, VALUE hash2, int eql)
  *    h == {'FOO' => 'zero', 'bar' => 'one'} # => false  # Different key.
  *    h == {'foo' => 'ZERO', 'bar' => 'one'} # => false  # Different value.
  *
+ *  See {Hash Equality and Inclusion}[rdoc-ref:Hash@Hash+Equality+and+Inclusion].
+ *
+ *  Related: see {Methods for Comparing}[rdoc-ref:Hash@Methods+for+Comparing].
  */
 
 static VALUE
@@ -4639,8 +4641,7 @@ hash_le(VALUE hash1, VALUE hash2)
  *    self <= other_hash -> true or false
  *
  *  Returns +true+ if the entries of +self+ are a subset of the entries of +other_hash+,
- *  +false+ otherwise;
- *  (see {Hash Equality and Inclusion}[rdoc-ref:Hash@Hash+Equality+and+Inclusion];
+ *  +false+ otherwise:
  *
  *    h0 = {foo: 0, bar: 1}
  *    h1 = {foo: 0, bar: 1, baz: 2}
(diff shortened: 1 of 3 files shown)


hash.c
```diff
@@ -6964,6 +6965,67 @@ static const rb_data_type_t env_data_type = {
  *  - For either method #fetch or method #fetch_values,
  *    you can specify a per-key default via a block.
  *
+ *  === \Hash Comparison
```
peterzhu2118, line 6968: I'm not sure what the purpose of this section is. This should just be in the documentation of `#==`, `#>`, `#<`, etc.

BurdetteLamar: Closing.

BurdetteLamar:
I think we do need this.  the five comparison methods need to be able to link to this, instead of each having to say what makes an entry in a hash equal to another entry in another hash.

BurdetteLamar:
> Can you show how this will be used? I don't have enough context on how it will be used from this PR.

Let me go ahead and write up the five methods and add to this PR.

BurdetteLamar:
Marking as Draft (WIP, for now).

peterzhu2118 (commented):
I see why you want to have a "Hash Equality and Inclusion" section. But I don't think it should be in the class documentation. I would prefer if we had this in a separate document and did a `:include:` directive.

BurdetteLamar:
> I see why you want to have a "Hash Equality and Inclusion" section. But I don't think it should be in the class documentation. I would prefer if we had this in a separate document and did a `:include:` directive.

`:include:`?, not a link?

peterzhu2118:
I don't have a strong opinion on link vs `:include:`, but my preference is `:include:` because I find that following links in documentation makes it more cumbersome to navigate and take longer to read.

BurdetteLamar:
> I don't have a strong opinion on link vs `:include:`, but my preference is `:include:` because I find that following links in documentation makes it more cumbersome to navigate and take longer to read.

I think the text is way too large (and detailed) to include in the methods' discussion.  If not included, the text should remain as a section in `Hash.html`;  it has no use outside of Hash (I don't think).  So, stet?

I'll be filing a request over at `ruby/rdoc` for supporting `<details>` and `<summary>`, which I think would be perfect for this case and maybe many others -- would let us inline w/o forcing details onto the reader.

An alternative would be support for `<small>`, which I've seen elsewhere used to indicate that some info is detailed and may be skipped by some readers.

peterzhu2118:
> I think the text is way too large (and detailed) to include in the methods' discussion. If not included, the text should remain as a section in Hash.html; it has no use outside of Hash (I don't think). So, stet?

The class documentation should be a guide and a high level overview of the class. It should read continually and have good flow. It's not a dumping ground for random pieces of information that is shared across a few methods.

BurdetteLamar:
New file added in doc/.

hash.c
```diff
@@ -3804,21 +3804,25 @@ hash_equal(VALUE hash1, VALUE hash2, int eql)
 
 /*
  *  call-seq:
- *    hash == object -> true or false
+ *    self == object -> true or false
  *
- *  Returns +true+ if all of the following are true:
- *  * +object+ is a +Hash+ object.
- *  * +hash+ and +object+ have the same keys (regardless of order).
- *  * For each key +key+, <tt>hash[key] == object[key]</tt>.
```
peterzhu2118, line 3812: This new documentation hides this important information in a link to the new document. IMO this is important and so we should make this easy to access.

hash.c
```diff
@@ -4653,14 +4665,24 @@ rb_hash_le(VALUE hash, VALUE other)
...
- *  Returns +true+ if +hash+ is a proper subset of +other_hash+, +false+ otherwise:
- *    h1 = {foo: 0, bar: 1}
- *    h2 = {foo: 0, bar: 1, baz: 2}
- *    h1 < h2 # => true
- *    h2 < h1 # => false
- *    h1 < h1 # => false
+ *  Returns +true+ if the entries of +self+ are a proper subset of the entries of +other_hash+,
+ *  +false+ otherwise:
+ *
+ *    h = {'foo' => 'zero', 'bar' => 'one'}
```
peterzhu2118, line 4673: Is there a reason why you used strings here instead of symbols? The symbol syntax makes the code much shorter.

hash.c
BurdetteLamar: Fixed (I think).

hash.c

hash.c
```diff
@@ -4691,14 +4725,24 @@ rb_hash_ge(VALUE hash, VALUE other)
...
- *    h1 > h2 # => true
- *    h2 > h1 # => false
- *    h1 > h1 # => false
+ *  Returns +true+ if the entries of +self+ are a proper superset of the entries of +other_hash+,
+ *  +false+ otherwise:
+ *
+ *    h = {foo: 0, bar: 1, baz: 2}
+ *    h > {foo: 0, bar: 1}         # => true   # Proper superset.
+ *    h > {bar: 1, foo: 0}         # => true   # Order may differ.
+ *    h > h                                                 # => false  # Not a proper superset.
```
peterzhu2118, line 4736: The alignment of the comment of this line is off.

peterzhu2118 (approved):

PR #12645, opened 2025-01-27
win32/resolv: add headers to GetNetworkParams check.

jeremyd2019:
On 32-bit Cygwin at least, it was failing to find that function, presumably due to it being stdcall.

32-bit Cygwin is not supported anymore by the Cygwin folks, but this seems a correct change in general.

Note I just noticed the ruby/resolv repository has this code too.  I'll open a PR with this change there too, feel free to close the one that's in the wrong place.  I figure the GHA checks on this repository will be helpful to confirm I didn't break anything that used to work, at least.


hsbt:
We can sync https://github.com/ruby/resolv/pull/81 automatically after merging that. Don't need to create extra PR in this repo.

PR #12646, opened 2025-01-27
Fix Ractor#require with lots of ractors

luke-gru:
stack struct memory was receiving weird values with lots of ractors, even when struct declared volatile. I couldn't track down the exact issue, but I think it had to do with interrupts. A workaround is to allocate the cross_ractor_require struct to the heap and then free it after the require is done.

Fixes [Bug #21090]

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -4060,8 +4060,8 @@ rb_ractor_channel_close(rb_execution_context_t *ec, VALUE vch)
 
 struct cross_ractor_require {
     VALUE ch;
-    VALUE result;
-    VALUE exception;
+    volatile VALUE result;
+    volatile VALUE exception;
 
     // require
     VALUE feature;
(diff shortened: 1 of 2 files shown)


test/ruby/test_require.rb
```diff
@@ -1011,4 +1011,20 @@ class Object
       end
     RUBY
   end
+
+  def test_ractor_require_bug_21090
```
ono-max, line 1015: FYI: This is a flaky test.

https://app.launchableinc.com/organizations/ruby/workspaces/ruby/data/test-sessions/3791644?tab=flakes
https://app.launchableinc.com/organizations/ruby/workspaces/ruby/data/test-sessions/3796140?tab=flakes

luke-gru: Yeah I saw, thanks. I couldn't reproduce the flakiness on my machine, which is also running MacOS 15. I'll try running the test for a few hours and see what I get. If I can't reproduce it, I'll increase the timeout for the test. If that doesn't work I plan on taking a closer look at the versions of compilers in the macos-15 image and getting my environment close to the CI image.

PR #12647, opened 2025-01-27
Prefer `uname -n` over `hostname`.

ioquatix:
See <https://gitlab.archlinux.org/archlinux/packaging/packages/inetutils/-/issues/2#note_211062> for context.

### DIFF
--- a/spec/ruby/library/socket/socket/gethostname_spec.rb
+++ b/spec/ruby/library/socket/socket/gethostname_spec.rb
@@ -3,6 +3,6 @@ require_relative '../fixtures/classes'
 
 describe "Socket.gethostname" do
   it "returns the host name" do
-    Socket.gethostname.should == `hostname`.strip
+    Socket.gethostname.should == `uname -n`.strip
   end
 end


spec/ruby/library/socket/socket/gethostname_spec.rb
```diff
@@ -3,6 +3,6 @@
 
 describe "Socket.gethostname" do
   it "returns the host name" do
-    Socket.gethostname.should == `hostname`.strip
+    Socket.gethostname.should == `uname -n`.strip
```
nobu, line 6: This will break on **pure** Windows (GitHub Actions have msys).

ioquatix: Maybe we can have a conditional on Windows to use `hostname`?

eregon: Yes it would be good to avoid using `uname` on Windows as it might not be there/it's not the right thing to use.
You can use `platform_is :windows do` for that

eregon: I'll revert this, `hostname` is more portable, and Arch can add a dependency to `inetutils` to run specs

ioquatix: `hostname` is not more portable, it is a GNU-specific tool (`inettools`) and `uname -n` is standardised by POSIX/Open Group: https://pubs.opengroup.org/onlinepubs/009695399/utilities/uname.html and should be available everywhere. However, I accept on vanilla Windows, only `hostname.exe` exists.

ioquatix:
Rather than reverting this, I am okay to add a fallback to `hostname`: https://github.com/ruby/ruby/pull/12655

PR #12649, opened 2025-01-28
Fix to build bundled gems with cygwin

hsbt:


.github/workflows/cygwin.yml
```diff
@@ -55,6 +55,11 @@ jobs:
           ./configure --disable-install-doc
         shell: C:\cygwin\bin\bash.EXE --noprofile --norc -e -o igncr -o pipefail {0}
 
+      - name: Extract bundled gems
+        run: |
+          make ruby -j2
+          make extract-gems
```
jeremyd2019, line 61:
```suggestion
          make extract-gems
        shell: C:\cygwin\bin\bash.EXE --noprofile --norc -e -o igncr -o pipefail {0}
```

?

jeremyd2019:
somewhat off-topic, I ran into issues building with high -j values on Cygwin (https://cygwin.com/pipermail/cygwin/2025-January/257155.html).  Is that why you selected `-j2`?  That seemed to be reasonable compromise in https://github.com/msys2/MSYS2-packages/pull/5144#issuecomment-2614080664, but the MSYS2-default `-j5` on GHA seemed to fail pretty frequently.

PR #12654, opened 2025-01-28
Revert "Prefer `uname -n` over `hostname`."

eregon:
Reverts ruby/ruby#12647
See https://github.com/ruby/ruby/pull/12647#discussion_r1931980612


ioquatix:
`uname -n` is defined by POSIX but `hostname` has no such specification IIRC, so I’d prefer to add `hostname` as a fallback to `uname` rather than revert this change.

eregon:
-> https://github.com/ruby/ruby/pull/12655

PR #12655, opened 2025-01-28
Add fallback for `hostname` if `uname` isn't available.

ioquatix:
cc @segaja

### DIFF
--- a/spec/ruby/library/socket/socket/gethostname_spec.rb
+++ b/spec/ruby/library/socket/socket/gethostname_spec.rb
@@ -2,7 +2,15 @@ require_relative '../spec_helper'
 require_relative '../fixtures/classes'
 
 describe "Socket.gethostname" do
+  def system_hostname
+    # Most platforms implement this POSIX standard:
+    `uname -n`.strip
+  rescue
+    # Only really required for Windows without MSYS/MinGW/Cygwin etc:
+    `hostname`.strip
+  end
+
   it "returns the host name" do
-    Socket.gethostname.should == `uname -n`.strip
+    Socket.gethostname.should == system_hostname
   end
 end


spec/ruby/library/socket/socket/gethostname_spec.rb
```diff
@@ -2,7 +2,15 @@
 require_relative '../fixtures/classes'
 
 describe "Socket.gethostname" do
+  def system_hostname
+    # Most platforms implement this POSIX standard:
+    `uname -n`.strip
+  rescue
```
eregon, line 8: This is not good because it relies on deep implementation details of backticks, which might vary between platforms and Ruby implementations.
Specifically, it relies on `uname -n` to use no shell, because if it used a shell it wouldn't raise an exception and just emit some stderr output + exit code 1.
It's best to not rely  on deep implementation details like that in ruby/spec.

And anyway the real meaningful condition is POSIX vs not POSIX here, not try and rescue.

So it should use `platform_is/platform_is_not :windows` guards as I wrote you on Slack.

Also this is the second time you merge a PR and only ask for reviews after, this is uncomfortable and time-wasting so for the next PR wait for my review before merging.

ioquatix:
Here is the follow up PR as per your request: <https://github.com/ruby/ruby/pull/12682>

spec/ruby/library/socket/socket/gethostname_spec.rb
nobu: > And anyway the real meaningful condition is POSIX vs not POSIX here, not try and rescue.
>
> So it should use `platform_is/platform_is_not :windows` guards as I wrote you on Slack.

Why there isn't `platform_is :posix`?

eregon: > Why there isn't `platform_is :posix`?

`platform_is :posix` sounds nice but `platform_is_not :posix` seems not so clear, e.g. there might be platforms which are not POSIX and are not Windows (although I don't think we support any of those).

PR #12658, opened 2025-01-28
Update child module name when parent gets a temp name set

eileencodes:
When calling `set_temporary_name` on a parent module the child modules should get their name updated too.

Before:

```ruby
m = Module.new
m::N = Module.new
p m::N.name # => "#<Module:0x000000010d0a00b0>::N"

m.set_temporary_name("foo")
p m::N.name # => "#<Module:0x000000010d0a00b0>::N"
```

After:

```ruby
m = Module.new
m::N = Module.new
p m::N.name # => "#<Module:0x000000010d0a00b0>::N"

m.set_temporary_name("foo")
p m::N.name # => "foo::N"
```

[Bug #21094]

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -214,6 +214,22 @@ is_constant_path(VALUE name)
  *    c.new # => #<C:0x0....>
  */
 
+
+static VALUE
+build_const_path(VALUE head, ID tail);
+
+static enum rb_id_table_iterator_result
+set_temporary_name_i(ID id, VALUE v, void *payload)
+{
+    rb_const_entry_t *ce = (rb_const_entry_t *)v;
+    VALUE value = ce->value;
+    VALUE parental_path = *((VALUE *) payload);
+
+    rb_mod_set_temporary_name(value, build_const_path(parental_path, id));
+
+    return ID_TABLE_CONTINUE;
+}
+
 VALUE
 rb_mod_set_temporary_name(VALUE mod, VALUE name)
 {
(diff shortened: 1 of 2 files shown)


tenderlove (approved):

nobu:
Shouldn’t this change too?

```ruby
m = Module.new
m::N = Module.new
m::N::O = Module.new

m.set_temporary_name("foo")
p m::N::O
```

eileencodes:
Ah yea that should work too. I'll fix it tomorrow morning.

variable.c
```diff
@@ -214,6 +214,22 @@ is_constant_path(VALUE name)
...
+build_const_path(VALUE head, ID tail);
+
+static enum rb_id_table_iterator_result
+set_temporary_name_i(ID id, VALUE v, void *payload)
+{
+    rb_const_entry_t *ce = (rb_const_entry_t *)v;
+    VALUE value = ce->value;
+    VALUE parental_path = *((VALUE *) payload);
+
+    rb_mod_set_temporary_name(value, build_const_path(parental_path, id));
```
nobu, line 228: If the anonymous module has a constant that is not anonymous class or module, this will raise an exception.
And will crash on a non-class/module constant.

nobu:
[set_temporary_name.patch](https://github.com/user-attachments/files/18582075/set_temporary_name.patch)

nobu:
[set_temporary_name.patch](https://github.com/user-attachments/files/18582464/set_temporary_name.patch)

PR #12660, opened 2025-01-29
YJIT: Initialize locals in ISeqs defined with `...`

XrXr:
- **YJIT: Fix indentation [ci skip]**
- **YJIT: Initialize locals in ISeqs defined with `...`**

### DIFF
--- a/bootstraptest/test_yjit.rb
+++ b/bootstraptest/test_yjit.rb
@@ -5355,3 +5355,35 @@ assert_normal_exit %{
     new.foo
   end
 }
+
+# This used to trigger a "try to mark T_NONE"
+# due to an uninitialized local in foo.
+assert_normal_exit %{
+  def foo(...)
+    _local_that_should_nil_on_call = GC.start
+  end
+
+  def test_bug21021
+    puts [], [], [], [], [], []
+    foo []
+  end
+
+  GC.stress = true
+  test_bug21021
+}
+
+assert_equal 'nil', %{
+  def foo(...)
+    _a = _b = _c = binding.local_variable_get(:_c)
+
+    _c
+  end
+
+  # [Bug #21021]
+  def test_local_fill_in_forwardable
+    puts [], [], [], [], []
+    foo []
+  end
+
+  test_local_fill_in_forwardable.inspect
+}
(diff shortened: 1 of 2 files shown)


maximecb (approved):
Thanks Alan :)

benoittgt:
Thanks a lot

travisbell:
Can confirm this fixes the crashes we were also seeing. 👍🏼

jjb:
we just experienced this in 3.3, not sure if same root cause

```
<OBJ_INFO:gc_mark_ptr@gc.c:7072> 0x00007f8b781b8c20 [0 M    ] T_NONE 
/home/circleci/project/vendor/bundle/ruby/3.3.0/bundler/gems/rails-f042fc820675/activerecord/lib/active_record/result.rb:219: [BUG] try to mark T_NONE object
ruby 3.3.7 (2025-01-15 revision be31f993d7) +YJIT [x86_64-linux]

c:0243 p:---- s:1321 e:001320 CFUNC  :new
c:0242 p:0008 s:1315 e:001314 BLOCK  /home/circleci/project/vendor/bundle/ruby/3.3.0/bundler/gems/rails-f042fc820675/activerecord/lib/active_record/result.rb:219 [FINISH]
c:0241 p:---- s:1311 e:001310 CFUNC  :map
c:0240 p:0018 s:1307 e:001306 METHOD /home/circleci/project/vendor/bundle/ruby/3.3.0/bundler/gems/rails-f042fc820675/activerecord/lib/active_record/result.rb:219
```

k0kubun:
The forwarding optimization was added in Ruby 3.4 https://github.com/ruby/ruby/pull/10510, so I think Ruby 3.3 doesn't have the codegen that needs a backport of this.

Can you try that with Ruby 3.4.6? If it fails with 3.4.6 as well, it's new. Otherwise, we can look for a PR to backport, but it's probably not this PR.

jjb:
it was a fluke in CI, if it happens more frequently i'll do more investigation

fwiw we aren't able to use 3.4 in prod because of this https://github.com/puma/puma/issues/3620

k0kubun:
If you're one of the people who can still reproduce the issue (which is no longer possible for the people who tried to help us fix it), let's work together on fixing that. 

If you can reproduce it with `--yjit-dump-disasm=dir_to_save_logs` on Ruby 3.4.6 and share the log file and the full error backtrace, I think we can start looking into it.

PR #12666, opened 2025-01-29
Fix fiddle test dlopen for .so filename with more than one digit

lucaskanashiro:
In alpha architecture, we have libc.so.6.1 which causes a failure in this test:

 1) Failure:
TestFiddle#test_dlopen_linker_script_group_linux [/<<PKGBUILDDIR>>/test/fiddle/test_fiddle.rb:52]: <"libc.so"> expected but was
<"libc.so.6">.

Bug-Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1085238

### DIFF
--- a/test/fiddle/test_fiddle.rb
+++ b/test/fiddle/test_fiddle.rb
@@ -83,7 +83,7 @@ class TestFiddle < Fiddle::TestCase
     handle = Fiddle.dlopen("libc.so")
     begin
       assert_equal("libc.so",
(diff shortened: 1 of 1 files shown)


kou:
Thanks but could you open this in https://github.com/ruby/fiddle ?
ruby/fiddle is upstream.

lucaskanashiro:
Thanks, I'll do that.

PR #12668, opened 2025-01-29
YJIT: A64: Remove assert that trips when OOM at page boundary

XrXr:
With a well-timed OOM around a page switch in the backend, it can return RetryOnNextPage twice and crash due to the assert. (More places can signal OOM now since VirtualMem tracks Rust malloc heap size for --yjit-mem-size.)

Return error in these cases instead of crashing.

Fixes: https://github.com/Shopify/ruby/issues/566


maximecb (approved):
Ty Alan

kevin-sellers:
Wonderful turnaround time on this! I can confirm that this does indeed fix the issue we were seeing.

maximecb:
Alan saves our butt once again :)

PR #12669, opened 2025-01-29
Improve bundled gem warning messages

deivid-rodriguez:
The information about default gems no longer available at `require` time is currently a bit confusing once the default gem is no longer available.

This is because it suggests that the missing gem was actually loaded, and that the problem is a warning, not an actual error.

This PR improves the wording from something like

> irb was loaded from the standard library, but is not part of the default gems starting from Ruby 3.5.0

to something like

> irb used to be loaded from the standard library, but is not part of the default gems since Ruby 3.5.0

While looking into this improvement, I realized that the current implementation was much more complicated than it needed to, so I removed a bunch of code.

### DIFF
--- a/lib/bundled_gems.rb
+++ b/lib/bundled_gems.rb
@@ -33,8 +33,6 @@ module Gem::BUNDLED_GEMS # :nodoc:
     # "readline" => "3.5.0", # This is wrapper for reline. We don't warn for this.
   }.freeze
 
-  SINCE_FAST_PATH = SINCE.transform_keys { |g| g.sub(/\A.*\-/, "") }.freeze
-
   EXACT = {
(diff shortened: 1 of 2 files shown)


deivid-rodriguez:
@hsbt I have some more improvements that I'd like to build on top of this. Can you have a look when you get a moment?

hsbt:
@deivid-rodriguez Thanks! This change simplified the current complex logic. I'll merge this because the spec is passed now. If we have unexpected cases, I will fix that until next stable release.

deivid-rodriguez:
Great, thank you @hsbt. Now I'm planning to propose a slightly modified version of https://github.com/ruby/ruby/pull/11550 that does not have the issue that made you revert it.

PR #12670, opened 2025-01-29
Only count VM instructions in YJIT stats builds

tenderlove:
The instruction counter is slowing multi-Ractor applications.  I had changed it to use a thread local, but using a thread local is slowing single threaded applications.  This commit only enables the instruction counter in YJIT stats builds until we can figure out a way to gather the information with lower overhead.

### DIFF
--- a/internal/vm.h
+++ b/internal/vm.h
@@ -81,7 +81,7 @@ struct vm_ifunc *rb_current_ifunc(void);
 
 #if USE_YJIT
 /* vm_exec.c */
-extern RB_THREAD_LOCAL_SPECIFIER uint64_t rb_vm_insns_count;
+extern uint64_t rb_vm_insns_count;
 #endif
 
 extern bool rb_free_at_exit;
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -1637,7 +1637,7 @@ class TestYJIT < Test::Unit::TestCase
 
       [
         stats[:object_shape_count].is_a?(Integer),
-        stats[:ratio_in_yjit].is_a?(Float),
+        stats[:ratio_in_yjit].nil? || stats[:ratio_in_yjit].is_a?(Float),
       ].all?
     RUBY
   end
@@ -1648,7 +1648,7 @@ class TestYJIT < Test::Unit::TestCase
       3.times { test }
 
       # Collect single stat.
-      stat = RubyVM::YJIT.runtime_stats(:ratio_in_yjit)
+      stat = RubyVM::YJIT.runtime_stats(:yjit_alloc_size)
 
       # Ensure this invocation had stats.
       return true unless RubyVM::YJIT.runtime_stats[:all_stats]
--- a/vm_exec.c
+++ b/vm_exec.c
@@ -13,7 +13,7 @@
 
 #if USE_YJIT
 // The number of instructions executed on vm_exec_core. --yjit-stats uses this.
-RB_THREAD_LOCAL_SPECIFIER uint64_t rb_vm_insns_count = 0;
+uint64_t rb_vm_insns_count = 0;
 #endif
 
 #if VM_COLLECT_USAGE_DETAILS
--- a/vm_insnhelper.h
+++ b/vm_insnhelper.h
@@ -16,7 +16,7 @@ RUBY_EXTERN rb_serial_t ruby_vm_constant_cache_invalidations;
 RUBY_EXTERN rb_serial_t ruby_vm_constant_cache_misses;
 RUBY_EXTERN rb_serial_t ruby_vm_global_cvar_state;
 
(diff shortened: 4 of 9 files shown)


rjit.c
```diff
@@ -168,9 +168,17 @@ const struct ruby_opt_message rb_rjit_option_messages[] = {
...
 struct rb_rjit_runtime_counters rb_rjit_counters = { 0 };
 
+#if RJIT_STATS
+void
+rb_rjit_collect_vm_usage_insn(int insn)
+{
+    if (!rjit_stats_p) return;
+    rb_rjit_counters.vm_insns_count++;
+}
+#endif // YJIT_STATS
```
XrXr, line 178:
```suggestion
#endif // RJIT_STATS
```

configure.ac
```diff
@@ -3930,17 +3930,17 @@ AS_CASE(["${YJIT_SUPPORT}"],
...
     [dev_nodebug], [
 	rb_rust_target_subdir=dev_nodebug
-	CARGO_BUILD_ARGS='--profile dev_nodebug --features disasm'
+	CARGO_BUILD_ARGS='--profile dev_nodebug --features stats,disasm'
 	AC_DEFINE(YJIT_STATS, 1)
     ],
     [stats], [
 	rb_rust_target_subdir=stats
-	CARGO_BUILD_ARGS='--profile stats'
+	CARGO_BUILD_ARGS='--profile stats --features stats'
```
XrXr, line 3943: This whole hunk should be removed since we've removed the stats feature. It's not listed in Cargo.toml anymore. There's some comments leftover referencing it I'll cleanup separtely.

XrXr (commented):
Seems unnecessary to also go from a plain `++` per instruction to a call per instruction, but since it's in stats builds only it's fine.

k0kubun:
> Seems unnecessary to also go from a plain ++ per instruction to a call per instruction, but since it's in stats builds only it's fine.

I agree it's "fine", but I don't think we need to slow it down. Remember https://github.com/ruby/ruby/pull/12549 slowed down some benchmarks significantly, which is why we're doing this in the first place, and we need to run stats build in production.

Can we keep it `extern uint64_t rb_vm_insns_count;` (as before https://github.com/ruby/ruby/pull/12549) and just guard that with `#if YJIT_STATS`?

tenderlove:
> Yeah, let's not slow it down. Remember #12549 slowed down some benchmarks significantly, which is why we're doing this in the first place, and we need to run stats build in production. It'd be safer to keep it fast.
> 
> Can we keep it `uint64_t rb_vm_insns_count = 0;` (as before #12549) and just guard its increment with `#if YJIT_STATS`?

Ya, happy to change it to do that. I was debating whether or not to do what you say but just decided to just revert everything.  I'll change it to keep the global and guard with YJIT_STATS. Thank you!

configure.ac
```diff
@@ -4003,11 +4003,18 @@ AC_ARG_ENABLE(rjit,
...
     [dev], [
         # Link libcapstone for --rjit-dump-disasm
         AC_CHECK_LIB([capstone], [cs_disasm])
+
+        # Enable extra stats (vm_insns_count, ratio_in_rjit)
+        AC_DEFINE(RJIT_STATS, 1)
+    ],
+    [disasm], [
+        # Link libcapstone for --rjit-dump-disasm
+        AC_CHECK_LIB([capstone], [cs_disasm])
```
rwstauner, line 4017: I realize this was part of the revert but it looks a little funny...
To confirm, this is adding stats to the dev build,
And adding a disasm build in case anyone wants disasm without stats
(but there is no build for stats without disasm).

maximecb:
> Ya, happy to change it to do that. I was debating whether or not to do what you say but just decided to just revert everything. I'll change it to keep the global and guard with YJIT_STATS. Thank you!

Good move 👍

k0kubun:
Could we remove `RB_THREAD_LOCAL_SPECIFIER` from `rb_vm_insns_count`? That's what I wanted to mean in https://github.com/ruby/ruby/pull/12670#issuecomment-2625132232.

Currently, 99% (if not all) of the `ratio_in_yjit` users seem like our production environments (stats clusters). At the moment, having `RB_THREAD_LOCAL_SPECIFIER` would only hurt the performance of these environments.

tenderlove:
> Could we remove `RB_THREAD_LOCAL_SPECIFIER` from `rb_vm_insns_count`? That's what I wanted to mean in [#12670 (comment)](https://github.com/ruby/ruby/pull/12670#issuecomment-2625132232).

🤦 Yes definitely. Sorry, I reverted the revert but forgot that we only want to partially revert it.  I'll fix it up.

tenderlove:
@k0kubun do you mind taking another look? Sorry for so many reviews 😅

k0kubun (approved):
:+1:

PR #12677, opened 2025-01-30
Fix possible dereference of NULL

naya451:
get_nd_recv(node) is dereferenced in private_recv_p and may be NULL (that is checked in first if in function). Add a check to prevent NULL-dereference.


nobu:
> get_nd_recv(node) is dereferenced in private_recv_p and may be NULL (that is checked in first if in function).

Yes, it is checked before the dereference.
Why is it needed twice?

naya451:
You're right, sorry for the inconvenience.
This check was not added in previous versions where I found this bug first.

PR #12679, opened 2025-01-30
Update to ruby/spec@affef93

andrykonchin:

### DIFF
--- a/spec/ruby/core/marshal/dump_spec.rb
+++ b/spec/ruby/core/marshal/dump_spec.rb
@@ -47,6 +47,11 @@ describe "Marshal.dump" do
         Marshal.dump(-2**31 - 1).should == "\x04\bl-\a\x01\x00\x00\x80"
       end
     end
+
+    it "does not use object links for objects repeatedly dumped" do
+      Marshal.dump([0, 0]).should == "\x04\b[\ai\x00i\x00"
+      Marshal.dump([2**16, 2**16]).should == "\x04\b[\ai\x03\x00\x00\x01i\x03\x00\x00\x01"
+    end
   end
 
   describe "with a Symbol" do
@@ -93,6 +98,11 @@ describe "Marshal.dump" do
       value = [*value, value[0]]
       Marshal.dump(value).should == "\x04\b[\b#{symbol1}#{symbol2};\x00"
     end
+
+    it "uses symbol links for objects repeatedly dumped" do
+      symbol = :foo
+      Marshal.dump([symbol, symbol]).should == "\x04\b[\a:\bfoo;\x00" # ;\x00 is a link to the symbol object
+    end
   end
 
   describe "with an object responding to #marshal_dump" do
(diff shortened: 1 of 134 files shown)


andrykonchin:
cc @eregon

spec/ruby/core/marshal/dump_spec.rb
```diff
@@ -286,8 +286,10 @@ def _dump(level)
       ].should be_computed_by(:dump)
     end
 
-    it "uses object links for objects repeatedly dumped" do
-      Marshal.dump([0.0, 0.0]).should == "\x04\b[\af\x060@\x06" # @\x06 is a link to the float value
+    quarantine! do # fails on i686 with 'Expected "\x04\b[\af\x060f\x060" == "\x04\b[\af\x060@\x06" to be truthy but was false'
```
eregon, line 289: You can use a platform_is guard for this, i686 is 32-bit. Pack/unpack specs likely have guards like that.
But let's merge this PR now and refine that later.

andrykonchin: Yeah, it seems a proper way to do so.

TBH Now I see that I don't understand clearly the logic of dumping Float values.

The `Marshal.dump` may serialise a repeated Float value either with and without link .On macOS I always observe dumping repeated Float values with links for various Float values - small, 0, and big ones.

On macOS:
```ruby
f = 0.0; Marshal.dump([f, f])
# => "\x04\b[\af\x060@\x06"

f = 1234567891011121314.15; Marshal.dump([f, f])
# => "\x04\b[\af\x1A1.2345678910111214e18@\x06"

f = 1.23456789101112131415; Marshal.dump([f, f])
# => "\x04\b[\af\x171.2345678910111213@\x06"
```

In all the cases there is an object link `@\x06` at the end.

There are the following cases in CRuby's implementation of `Marshal.dump`:

1) Serialise a Float value and add it into an objects table (to refer it next time with an object link) if it's a "Flonum"

```c
if (FLONUM_P(obj)) {
    w_remember(obj, arg);
    w_byte(TYPE_FLOAT, arg);
    w_float(RFLOAT_VALUE(obj), arg);
    return;
}
```

2) Otherwise serialise a Float and **don't** add it into an object table

```c
case T_FLOAT:
  w_byte(TYPE_FLOAT, arg);
  w_float(RFLOAT_VALUE(obj), arg);
  break;
```

So I am wondering why on i686 Float values like `0.0` aren't "Flonum" and consequently aren't added into an objects table.

eregon: Oh, Marshal dumping with links or not depending on Flonum seems bad, I think that's worth discussion on the CRuby tracker, could you open an issue? I think it should be consistent for all floats (probably should always link).

In the meanwhile it would be fine to alter specs to allow either, since whether a Float is a Flonum or not is a fairly deep implementation detail.

> So I am wondering whether on i686 all Float values (e.g. 0.0) aren't "Flonum" and consequently aren't added into an objects table.

Correct, Flonum is only possible on 64-bit platforms, because it means using pointer tagging for a `double` (64-bit), and can't fit a double in 32-bit pointers (on 32-bit platforms).

andrykonchin: Created an issue https://bugs.ruby-lang.org/issues/21110.

PR #12682, opened 2025-01-31
Prefer `platform_is_not :windows`.

ioquatix:
See <https://github.com/ruby/ruby/pull/12655> for prior discussion.


eregon (approved):
Thanks!

PR #12689, opened 2025-02-02
[Bug #21108] Don't yield from Ractor.yield when terminating

osyoyu:
When the Ractor is in the process of termination, the basket could be basket_type_none (cleared), therefore invalid state for yielding. Yield nothing and return false in such cases.


osyoyu:
Closing in favor of #12646 (https://bugs.ruby-lang.org/issues/21108).

PR #12694, opened 2025-02-04
[DOC] Tweaks for Hash#[]

BurdetteLamar:


hash.c
```diff
@@ -2091,16 +2091,22 @@ rb_hash_stlike_lookup(VALUE hash, st_data_t key, st_data_t *pval)
...
- *    h[:foo] # => 0
+ *  Searches for a hash key equivalent to the given +key+;
+ *  see {Hash Key Equivalence}[rdoc-ref:Hash@Hash+Key+Equivalence].
+ *
+ *  If the key is found, returns its value:
+ *
+ *    {foo: 0, bar: 1, baz: 2}
+ *    h[:bar] # => 1
+ *
+ *  Otherwise, returns a default value (or raises an exception);
```
peterzhu2118, line 2104: In what cases does it raise an exception?

peterzhu2118: I still don't understand how it can raise an exception though? Can you provide an example?

BurdetteLamar: Can the default proc not raise an exception?  But I'm thinking that maybe we should just say 'returns +nil+' and let it go at that.

peterzhu2118: That's .... not very intuitive. That's like saying `Integer#times` can raise because the block can raise.

BurdetteLamar: @peterzhu2118, I have fouled my clone.  Would you fix this and merge?

peterzhu2118: Ok, I reverted it to the original documentation.

PR #12695, opened 2025-02-04
[DOC] Tweaks for Hash#[]=

BurdetteLamar:


hash.c
```diff
@@ -2901,26 +2901,31 @@ NOINSERT_UPDATE_CALLBACK(hash_aset_str)
...
  *  (see {Entry Order}[rdoc-ref:Hash@Entry+Order]):
+ *
  *    h = {foo: 0, bar: 1}
  *    h[:foo] = 2 # => 2
- *    h.store(:bar, 3) # => 3
- *    h # => {foo: 2, bar: 3}
+ *    h[:foo]     # => 2
  *
- *  If +key+ does not exist, adds the +key+ and +value+;
+ *  If the key is not found, creates a new entry for the given +key+ and +object+;
```
peterzhu2118, line 2919: Nit: I think we're referring to the variable `key` here.

```suggestion
 *  If +key+ is not found, creates a new entry for the given +key+ and +object+;
```

peterzhu2118 (approved):

PR #12707, opened 2025-02-06
Fix RJIT with bundled fiddle

nobu:


hsbt:
RJIT has been extracted at https://github.com/ruby/ruby/pull/12740

PR #12712, opened 2025-02-07
Add timeout to compilers workflow

peterzhu2118:
The default timeout on GitHub Actions is 360 minutes, the job usually takes around 20 to 30 minutes to complete. This commit sets the timeout to be 40 minutes so jobs that hang will timeout faster.


hsbt:
👍

PR #12713, opened 2025-02-07
PoC: Add option to use musttail in interpreter loop

jhawthorn:
This adds a compile-time option, `OPT_TAILCALL_THREADED_CODE`, to use musttail for threading in the VM loop.

This works best with the latest LLVM, though the latest GCC also has musttail.

This also attempts to use `__attribute__((preserve_none))` when available (LLVM 19 or so?). This changes the calling convention of the instruction methods to save all registers and use what would normally be callee-saved registers to pass arguments. This significantly reduces pushing and popping when we end up calling non-tailcall methods from our VM instructions (which is pretty common for us).

(I will make benchmarks before more seriously proposing merging this)

Aside from performance a thing I really like about this is that profilers (like linux perf) are able to show the VM instructions we're spending time inside. It also allows using objdump to print out the source for individual instructions, which is much easier to read.

```
❯ llvm-objdump -d --symbolize-operands --x86-asm-syntax=intel --no-show-raw-insn --no-leading-addr --disassemble-symbols=insn_func_pop -S ./miniruby

./miniruby:     file format elf64-x86-64

Disassembly of section .text:

<insn_func_pop>:
;     ADD_PC(INSN_ATTR(width));
                lea     rax, [r14 + 0x8]
                mov     qword ptr [r13], rax
;     COLLECT_USAGE_INSN(INSN_ATTR(bin));
                inc     qword ptr fs:[-0x18]
;     INC_SP(INSN_ATTR(sp_inc));
                add     qword ptr [r13 + 0x8], -0x8
;     END_INSN(pop);
                mov     rcx, qword ptr [r14 + 0x8]
                mov     r14, rax
                jmp     rcx
                nop
```

### DIFF
--- a/insns.def
+++ b/insns.def
@@ -1136,7 +1136,7 @@ leave
     }
 
     if (vm_pop_frame(ec, GET_CFP(), GET_EP())) {
-#if OPT_CALL_THREADED_CODE
+#if OPT_CALL_THREADED_CODE || OPT_TAILCALL_THREADED_CODE
         rb_ec_thread_ptr(ec)->retval = val;
         return 0;
 #else
@@ -1659,7 +1659,7 @@ opt_invokebuiltin_delegate_leave
     /* leave fastpath */
     /* TracePoint/return fallbacks this insn to opt_invokebuiltin_delegate */
     if (vm_pop_frame(ec, GET_CFP(), GET_EP())) {
-#if OPT_CALL_THREADED_CODE
+#if OPT_CALL_THREADED_CODE || OPT_TAILCALL_THREADED_CODE
         rb_ec_thread_ptr(ec)->retval = val;
         return 0;
 #else
--- a/vm_exec.h
+++ b/vm_exec.h
@@ -72,8 +72,8 @@ error !
 #define LABEL_PTR(x) &LABEL(x)
 
 #define INSN_FUNC_RET rb_control_frame_t *
-#define INSN_FUNC_PARAMS rb_execution_context_t *ec, rb_control_frame_t *reg_cfp
-#define INSN_FUNC_ARGS ec, reg_cfp
+#define INSN_FUNC_PARAMS rb_execution_context_t *ec, rb_control_frame_t *reg_cfp, const VALUE *reg_pc
+#define INSN_FUNC_ARGS ec, reg_cfp, reg_pc
 
 typedef INSN_FUNC_RET rb_insn_tailcall_func_t(INSN_FUNC_PARAMS);
 
@@ -91,7 +91,7 @@ typedef INSN_FUNC_RET rb_insn_tailcall_func_t(INSN_FUNC_PARAMS);
 #define NEXT_INSN() TC_DISPATCH(__NEXT_INSN__)
 
 #define START_OF_ORIGINAL_INSN(x) /* ignore */
-#define DISPATCH_ORIGINAL_INSN(x) return LABEL(x)(ec, reg_cfp);
+#define DISPATCH_ORIGINAL_INSN(x) MUSTTAIL return LABEL(x)(INSN_FUNC_ARGS);
 
 /************************************************/
 #elif OPT_TOKEN_THREADED_CODE || OPT_DIRECT_THREADED_CODE
--- a/vm_exec.h
+++ b/vm_exec.h
@@ -40,7 +40,7 @@ typedef rb_iseq_t *ISEQ;
 #if defined(DISPATCH_XXX)
 error !
 /************************************************/
-#elif OPT_CALL_THREADED_CODE
+#elif OPT_CALL_THREADED_CODE || OPT_TAILCALL_THREADED_CODE
 
 #define LABEL(x)  insn_func_##x
 #define ELABEL(x)
@@ -156,7 +156,7 @@ default:                        \
 
 #define VM_SP_CNT(ec, sp) ((sp) - (ec)->vm_stack)
 
-#if OPT_CALL_THREADED_CODE
+#if OPT_CALL_THREADED_CODE || OPT_TAILCALL_THREADED_CODE
 #define THROW_EXCEPTION(exc) do { \
     ec->errinfo = (VALUE)(exc); \
     return 0; \
--- a/vm_exec.h
+++ b/vm_exec.h
@@ -77,9 +77,12 @@ error !
 
 typedef INSN_FUNC_RET rb_insn_tailcall_func_t(INSN_FUNC_PARAMS);
 
+#define INSN_FUNC_ATTRIBUTES \
+    __attribute__((no_stack_protector))
+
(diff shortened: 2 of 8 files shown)


vm_exec.h
```diff
@@ -57,6 +57,60 @@ error !
...
 
+/************************************************/
+#elif OPT_TAILCALL_THREADED_CODE
+
+// https://blog.reverberate.org/2021/04/21/musttail-efficient-interpreters.html
+// https://sillycross.github.io/2022/11/22/2022-11-22/
+// https://github.com/wasm3/wasm3/blob/main/docs/Interpreter.md#m3-massey-meta-machine
+
+// TODO: move elsewhere
+#define MUSTTAIL __attribute__((musttail))
```
jhawthorn, line 68: TODO: Apparently `[[clang::musttail]]` has better support (works on GCC 15)

jhawthorn: Hmm. I didn't actually find evidence of this, 15.1 support __attribute__ https://godbolt.org/z/3dzzGc17e so maybe this was fixed by release

insns.def
```diff
@@ -869,16 +869,18 @@ sendforward
...
-    val = vm_sendish(ec, GET_CFP(), &adjusted_cd.cd, bh, mexp_search_method);
-    JIT_EXEC(ec, val);
+        val = vm_sendish(ec, GET_CFP(), &adjusted_cd.cd, bh, mexp_search_method);
+        JIT_EXEC(ec, val);
 
-    if (cd->cc != adjusted_cd.cd.cc && vm_cc_markable(adjusted_cd.cd.cc)) {
-        RB_OBJ_WRITE(GET_ISEQ(), &cd->cc, adjusted_cd.cd.cc);
+        if (cd->cc != adjusted_cd.cd.cc && vm_cc_markable(adjusted_cd.cd.cc)) {
+            RB_OBJ_WRITE(GET_ISEQ(), &cd->cc, adjusted_cd.cd.cc);
+        }
```
nobu, line 886: Looks like just enclosing in a brace pair.

insns.def
```diff
@@ -1110,16 +1112,18 @@ invokesuperforward
...
-    val = vm_sendish(ec, GET_CFP(), &adjusted_cd.cd, bh, mexp_search_super);
-    JIT_EXEC(ec, val);
+        val = vm_sendish(ec, GET_CFP(), &adjusted_cd.cd, bh, mexp_search_super);
+        JIT_EXEC(ec, val);
 
-    if (cd->cc != adjusted_cd.cd.cc && vm_cc_markable(adjusted_cd.cd.cc)) {
-        RB_OBJ_WRITE(GET_ISEQ(), &cd->cc, adjusted_cd.cd.cc);
+        if (cd->cc != adjusted_cd.cd.cc && vm_cc_markable(adjusted_cd.cd.cc)) {
+            RB_OBJ_WRITE(GET_ISEQ(), &cd->cc, adjusted_cd.cd.cc);
+        }
```
nobu, line 1130: Here too.

insns.def
jhawthorn: It is just enclosing a brace pair. We're declaring `struct rb_forwarding_call_data` and `struct rb_callinfo adjusted_ci`. Without the brace GCC is unsure whether or not the values used in the tailcall dispatch (`rb_execution_context_t *ec, rb_control_frame_t *reg_cfp, const VALUE *reg_pc`) might be pointing to something withing those structs.

Adding the brace makes it explicit to GCC that those go out of scope before dispatch

jhawthorn:
Here are the results from ruby-bench micro (`--category=micro`) benchmarks

```
ruby-clang: ruby 4.0.0dev (2025-12-07T16:51:02Z master 4f900c35bc) +PRISM [x86_64-linux]
ruby-tailcall: ruby 4.0.0dev (2025-12-08T00:48:48Z tailcall_2025 2e0aff55a0) +PRISM [x86_64-linux]
last_commit=Use tailcall interpreter if musttail is available

-------------------  ---------------  ----------  ------------------  ----------  ---------------------  ------------------------
bench                ruby-clang (ms)  stddev (%)  ruby-tailcall (ms)  stddev (%)  ruby-tailcall 1st itr  ruby-clang/ruby-tailcall
30k_ifelse           387.7            0.0         385.0               0.1         1.006                  1.007
30k_methods          307.1            0.2         310.3               0.1         0.993                  0.990
attr_accessor        91.0             0.6         83.9                0.7         1.084                  1.086
cfunc_itself         58.2             6.4         55.9                0.6         0.954                  1.042
fib                  125.6            3.3         116.5               0.2         1.102                  1.078
getivar              55.7             1.2         42.1                3.3         1.375                  1.325
getivar-module       117.0            2.2         93.8                1.7         1.489                  1.248
keyword_args         128.3            1.2         125.0               0.2         1.017                  1.026
loops-times          507.4            0.2         475.7               0.7         1.107                  1.067
object-new           40.1             0.2         39.6                0.2         1.013                  1.012
respond_to           130.0            0.3         114.4               0.5         1.123                  1.136
ruby-xor             66.6             0.4         64.4                0.2         1.017                  1.035
send_bmethod         129.8            0.2         128.8               0.6         0.986                  1.008
send_cfunc_block     160.6            0.9         147.5               0.7         1.064                  1.089
send_rubyfunc_block  66.4             0.7         64.5                0.2         1.039                  1.029
setivar              42.3             1.9         44.0                1.5         1.023                  0.961
setivar_object       72.0             0.6         70.1                0.3         1.013                  1.026
setivar_young        72.3             0.8         71.0                0.9         1.016                  1.019
str_concat           36.2             0.3         28.7                0.3         1.117                  1.260
structaref           87.2             3.5         98.6                1.9         0.920                  0.884
structaset           90.5             1.6         88.7                2.1         0.988                  1.020
throw                12.9             0.5         12.4                0.4         0.999                  1.043
-------------------  ---------------  ----------  ------------------  ----------  ---------------------  ------------------------

Legend:
- ruby-tailcall 1st itr: ratio of ruby-clang/ruby-tailcall time for the first benchmarking iteration.
- ruby-clang/ruby-tailcall: ratio of ruby-clang/ruby-tailcall time. Higher is better for ruby-tailcall. Above 1 represents a speedup.
```

YO4:
+1

I'm interested in this.
The Microsoft C Compiler (msvc) supports `[[msvc::musttail]]` and `__preserve_none` in C23 mode, and the use of tail calls has a significant impact in situations where support for OPT_DIRECT_THREADED_CODE is lacking. So I'd like to add support for msvc.

Is there anything I can do to help?

The CI failure in MinGW (CLANGARM64) appears to be caused by a conflict between `preserve_none` and `fastcall`.
I’m testing this at https://github.com/YO4/ruby/commits/jhawthorn/tailcall_2025
but I’m not sure how to proceed with Git to submit a patch in this situation.

PR #12718, opened 2025-02-09
gc.c: Remove no-op code

osyoyu:
In this context, `vm_locked` is a argument variable, and is not used later in the function.


osyoyu:
Unrelated CI failure?

peterzhu2118 (approved):

PR #12721, opened 2025-02-10
[Bug #21127] Thread deadlock does not display backtraces

pocke:
Previously, Ruby displayed backtraces for each thread on deadlock. However, it has not been shown since Ruby 3.0.
It should display the backtrace for debugging.

https://bugs.ruby-lang.org/issues/21127

### DIFF
--- a/eval_intern.h
+++ b/eval_intern.h
@@ -300,8 +300,8 @@ void rb_vm_set_progname(VALUE filename);
 VALUE rb_vm_cbase(void);
 
 /* vm_backtrace.c */
-#define BACKTRACE_START 0
-#define ALL_BACKTRACE_LINES -1
+#define RUBY_BACKTRACE_START 0
+#define RUBY_ALL_BACKTRACE_LINES -1
 VALUE rb_ec_backtrace_object(const rb_execution_context_t *ec);
 VALUE rb_ec_backtrace_str_ary(const rb_execution_context_t *ec, long lev, long n);
 VALUE rb_ec_backtrace_location_ary(const rb_execution_context_t *ec, long lev, long n, bool skip_internal);
--- a/eval_intern.h
+++ b/eval_intern.h
@@ -300,6 +300,8 @@ void rb_vm_set_progname(VALUE filename);
 VALUE rb_vm_cbase(void);
 
 /* vm_backtrace.c */
+#define BACKTRACE_START 0
+#define ALL_BACKTRACE_LINES -1
 VALUE rb_ec_backtrace_object(const rb_execution_context_t *ec);
 VALUE rb_ec_backtrace_str_ary(const rb_execution_context_t *ec, long lev, long n);
 VALUE rb_ec_backtrace_location_ary(const rb_execution_context_t *ec, long lev, long n, bool skip_internal);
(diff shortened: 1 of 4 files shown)


eval_intern.h
```diff
@@ -300,6 +300,8 @@ void rb_vm_set_progname(VALUE filename);
 VALUE rb_vm_cbase(void);
 
 /* vm_backtrace.c */
+#define BACKTRACE_START 0
+#define ALL_BACKTRACE_LINES -1
```
jeremyevans, line 304: I think if we are going to make these constants public API, they should start with `RUBY_` or `RB_.`

jeremyevans (commented):
Thank you for finding and fixing this issue.  As you expected, this was not intentional.  We must not have any tests that the thread backtraces were printed during deadlock.  Is it possible for you to add a test for that, to prevent regressions in the future?

pocke:
Thanks for your review! I added a test case and `RUBY_` prefix for the macro.

test/ruby/test_thread.rb
```diff
@@ -1557,4 +1557,37 @@ def test_pending_interrupt?
...
+    expected_stderr = [
+      /-:12:in 'Thread#join': No live threads left. Deadlock\? \(fatal\)\n/,
+      /2 threads, 2 sleeps current:\w+ main thread:\w+\n/,
+      /\* #<Thread:\w+ sleep_forever>\n/,
+      :*,
+      /^\s*-:6:in 'Object#frame_for_deadlock_test_2'/,
+      :*,
+      /\* #<Thread:\w+ -:10 sleep_forever>\n/,
+      :*,
+      /^\s*-:2:in 'Object#frame_for_deadlock_test_1'/,
```
pocke, line 1574: It depends on the order of `threads.set`.
https://github.com/pocke/ruby/blob/85fb7a31942884b5b0a0d95a2c25eb01dbfd6e15/thread.c#L5580

I'm concerned it causes a flaky test if the order is randomized. I guess it's not a problem because the main thread is always the first item on the list. Please tell me if it will cause a problem 🙏

test/ruby/test_thread.rb
```diff
@@ -1553,4 +1553,37 @@ def test_pending_interrupt?
     assert_equal(true, t.pending_interrupt?(Exception))
     assert_equal(false, t.pending_interrupt?(ArgumentError))
   end
+
+  def test_deadlock_backtrace
+    bug21127 = '[ruby-core:120930] [Bug #21127]'
+
```
jeremyevans, line 1559:
```suggestion
```

Probably only need one empty line here.

test/ruby/test_thread.rb
jeremyevans: I'm not sure if the order is guaranteed or deterministic.  But if it passes CI, it should be OK.  If it turns out to cause a test failure, we can always just the test so it works with either order.

jeremyevans (approved):
Change looks good, thank you for working on this!

test/ruby/test_thread.rb
pocke: Sorry for the mistake 🙈 I'll commit your suggestion.

PR #12726, opened 2025-02-10
[DOC] Tweaks for Hash#assoc

BurdetteLamar:


hash.c
```diff
@@ -4159,13 +4159,15 @@ assoc_i(VALUE key, VALUE val, VALUE arg)
 
 /*
  *  call-seq:
- *    hash.assoc(key) -> new_array or nil
+ *    assoc(key) -> entry or nil
```
peterzhu2118, line 4162: The other call-seq seems to standardize on having `hash.` prepended.

```suggestion
 *    hash.assoc(key) -> entry or nil
```

BurdetteLamar: > The other call-seq seems to standardize on having `hash.` prepended.

Doc guide says not: https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html#label-Calling+Sequence+-28for+methods+written+in+C-29.

Will change the others as we go.

peterzhu2118 (approved):

PR #12727, opened 2025-02-10
[DOC] Tweaks for Hash#clear

BurdetteLamar:


hash.c
```diff
@@ -2845,9 +2845,9 @@ clear_i(VALUE key, VALUE value, VALUE dummy)
 
 /*
  *  call-seq:
- *    hash.clear -> self
+ *    clear -> self
```
peterzhu2118, line 2848: The other call-seq seems to standardize on having `hash.` prepended.

```suggestion
 *    hash.clear -> self
```

BurdetteLamar: > The other call-seq seems to standardize on having `hash.` prepended.

Doc guide says not:  https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html#label-Calling+Sequence+-28for+methods+written+in+C-29.

Will change the others as we go.

peterzhu2118: I see, I fixed all of them in Hash and Array in https://github.com/ruby/ruby/pull/12728 to make it consistent.

peterzhu2118 (approved):

PR #12729, opened 2025-02-10
Always return jit_entry even if NULL

tenderlove:
We can just always return the jit_entry since it will be initialized to NULL.  There is no reason to specifically return NULL if yjit / rjit are disabled


k0kubun (approved):
:+1:

PR #12735, opened 2025-02-12
[Bug #21128] Include fcntl.h before checking for O_CLOEXEC

collinfunk:
On glibc O_CLOEXEC is defined in fcntl.h and not unistd.h so this change prevents the macro from being redefined.

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -22,10 +22,6 @@
 #include <unistd.h>
 #endif
 
-#ifndef O_CLOEXEC
-#  define O_CLOEXEC 0
-#endif
-
 #ifndef USE_OPENDIR_AT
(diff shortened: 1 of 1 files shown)


dir.c
```diff
@@ -35,8 +31,12 @@
 # endif
 #endif
 
-#if USE_OPENDIR_AT
-# include <fcntl.h>
+#if USE_OPENDIR_AT || HAVE_FCNTL_H
```
mame, line 34:
```suggestion
#ifdef HAVE_FCNTL_H
```

collinfunk: Amended with your suggestion and force pushed.

shyouhei (approved):
LGTM

PR #12737, opened 2025-02-12
Fixed infinite loop with GitHub Actions

hsbt:


hsbt:
This fixed the issued of `RJIT_FORCE_ENABLE` workflow.

PR #1234, opened 2016-02-02
Fix Ripper.lex error in dedenting squiggly heredoc

hanachin:
Ripper.lex raises error for squiggly heredoc.

reproduce code:

```console
% ruby -v -rripper -e 'Ripper.lex("<<~EOS\n  hi\nEOS")'
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-darwin15]
/usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:70:in `block in on_heredoc_dedent': undefined method `event' for #<Array:0x007fef29875238> (NoMethodError)
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:69:in `each'
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:69:in `on_heredoc_dedent'
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:61:in `parse'
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:61:in `parse'
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:55:in `lex'
	from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/2.3.0/ripper/lexer.rb:44:in `lex'
	from -e:1:in `<main>'
```

PR #12751, opened 2025-02-14
Backports

kddnewton:


k0kubun (approved):

k0kubun:
Btw, I'd appreciate it if you can write `[Bug #1234]` instead of `Bug #1234` in commit messages next time. The redmine integration and backport scripts parses it, so the bracket matters.

PR #12760, opened 2025-02-16
Remove undefined function from bindgen

tenderlove:
`rb_get_iseq_body_total_calls` was removed in cd8d20cd1fbcf9bf9d438b306beb65b2417fcc04, but it's still in the YJIT bindgen file.  This commit just removes it from bindgen


maximecb (commented):
Thanks Aaron!

PR #12775, opened 2025-02-18
[DOC] Tweaks for Hash#empty?

BurdetteLamar:


hash.c
```diff
@@ -3016,9 +3016,12 @@ rb_hash_size_num(VALUE hash)
  *  call-seq:
  *    empty? -> true or false
  *
- *  Returns +true+ if there are no hash entries, +false+ otherwise:
+ *  Returns whether <tt>self.size == 0</tt>:
```
peterzhu2118, line 3019: I prefer the original documentation (or maybe a mix of both?) because it gives an intuitive description of what this method does.

BurdetteLamar: Restored.

peterzhu2118 (approved):

PR #12781, opened 2025-02-18
Fix crash for special constants in too complex generic ivars

peterzhu2118:
We should skip reference updating for entries in too complex generic ivars that are special constants. This fixes the following crash:

```ruby
MAX_SHAPES = 0x80000

MAX_SHAPES.times do |i|
  o = []
  o.instance_variable_set(:"@foo#{i}", 1)
end

o = []

o.instance_variable_set(:"@a", 123)

GC.compact
```

### DIFF
--- a/test/ruby/test_gc_compact.rb
+++ b/test/ruby/test_gc_compact.rb
@@ -452,4 +452,21 @@ class TestGCCompact < Test::Unit::TestCase
       assert_raise(FrozenError) { a.set_a }
     end;
   end
+
+  def test_moving_too_compex_generic_ivar
+    omit "not compiled with SHAPE_DEBUG" unless defined?(RubyVM::Shape)
+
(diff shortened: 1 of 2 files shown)


XrXr (approved):

test/ruby/test_gc_compact.rb
```diff
@@ -452,4 +452,21 @@ def set_a
       assert_raise(FrozenError) { a.set_a }
     end;
   end
+
+  def test_moving_too_compex_generic_ivar
```
sanjioh, line 456: Hi, little typo here (`compex` -> `complex`)

peterzhu2118: Thanks for catching this. Fixed in 80a71bfb1cd58b4129e2b008fd0fccc7fc90a11e

PR #12785, opened 2025-02-20
Launchable: Send stdout and stderr

ono-max:
Currently, the Launchable team is developing a new feature to attach any logs. Attached log can be anything, such as system logs or stdout. Users can find these logs using any text search. Please note that this feature is a work in progress, so we can't use it yet. 

I'm going to attach stdout and stderr as attached logs because they will be useful for finding interpreter bugs. When running tests, we sometimes see interpreter itself crash, and the stack is output to stderr. When debugging the cause of the issue, this feature is useful.

### DIFF
--- a/.github/actions/launchable/setup/action.yml
+++ b/.github/actions/launchable/setup/action.yml
@@ -247,6 +247,10 @@ runs:
           mkdir "${test_spec_report_path}"
           echo test_spec_report_path="${test_spec_report_path}" >> $GITHUB_OUTPUT
         fi
+        stdout_report_path="${dir}${builddir:+${builddir}/}launchable_stdout.log"
+        stderr_report_path="${dir}${builddir:+${builddir}/}launchable_stderr.log"
+        echo stdout_report_path="${stdout_report_path}" >> $GITHUB_OUTPUT
+        echo stderr_report_path="${stderr_report_path}" >> $GITHUB_OUTPUT
       if: steps.enable-launchable.outputs.enable-launchable
       env:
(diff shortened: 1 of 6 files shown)


.github/actions/launchable/setup/action.yml
```diff
@@ -264,10 +264,11 @@ runs:
         shell: bash
         working-directory: ${{ inputs.srcdir }}
         post: |
-          [[ "${test_all_enabled}" = "true" ]] && \
+          if [[ "${test_all_enabled}" == "true" ]]; then \
             launchable record tests \
               --session "$(cat "${test_all_session_file}")" \
-              raw "${test_report_path}" || true
+              raw "${test_report_path}" \
+          fi
```
nobu, line 271: Do not escape the newline just before `fi`.
```suggestion
              raw "${test_report_path}"
          fi
```

.github/actions/launchable/setup/action.yml
```diff
@@ -264,10 +264,11 @@ runs:
         shell: bash
         working-directory: ${{ inputs.srcdir }}
         post: |
-          [[ "${test_all_enabled}" = "true" ]] && \
+          if [[ "${test_all_enabled}" == "true" ]]; then \
```
nobu, line 267: This escape is not needed too.
```suggestion
          if [[ "${test_all_enabled}" == "true" ]]; then
```

ono-max:
@nobu 

Can you please review changes under `.github/workflows`?

ono-max:
I got feedback via Slack from @nobu, so I believe that this PR is ready to merge.

PR #12786, opened 2025-02-20
Avoid some false positive bundled gem warnings

deivid-rodriguez:
When a bundled gem has already been removed, a `require` caller is rescuing `LoadError`, no warning/error messages should be displayed.

Instead, let the bundled gem message be part of `LoadError`, so that it's not displayed when rescued, but still gets to the user when not rescued.

This is an idea to reintroduce https://github.com/ruby/ruby/pull/11550, but without the issue that made @hsbt revert the original approach: If users upgrade Ruby from a version that did not show any warnings, to a version where the default gem has already been removed, then they'll miss any noticed about removal of the default gem. By making the notice part of the `LoadError`, they will not miss it.


deivid-rodriguez:
This still feels like a small usability improvement to me, even if it does not fix all issues. Thoughts?

deivid-rodriguez:
Alright, giving up.

PR #12798, opened 2025-02-24
[Feature #19908] Update to Unicode 15.1.0

ima1zumi:
https://bugs.ruby-lang.org/issues/19908

## Summary

Unicode 15.1.0 introduced a new grapheme cluster rule to handle Indic conjuncts. In UAX #29, rule GB9c was added to prevent grapheme breaks within certain Indic consonant+virama sequences. As a result, an orthographic syllable in scripts like Devanagari (e.g. “क्या”, consisting of KA + VIRAMA + YA) is now treated as a single extended grapheme cluster rather than split into two. This aligns the default segmentation with Indic writing system expectations.

[UAX #29: Unicode Text Segmentation](https://www.unicode.org/reports/tr29/tr29-43.html)

## New Enumerated Property

To support this, the Unicode Character Database (UCD) added a new enumerated property, Indic_Conjunct_Break (InCB), with values Consonant, Linker, Extend, and None, derived from existing properties. For example, virama characters in certain Brahmic scripts are classified as InCB=Linker, and base consonants as InCB=Consonant. These property values are listed in the UCD data (e.g. in DerivedCoreProperties.txt).

## Impact on Ruby

This change affected Ruby’s implementation of Unicode support in several ways:
- enc-unicode.rb
    - Ruby’s enc-unicode.rb script, which processes Unicode data, had to be updated to handle the new enumerated property. Previously, it expected only binary properties.
- Regex Engine
    - Support for InCB property values was added to Ruby’s regex engine (exposing them as InCB_Linker, InCB_Consonant, etc.).
- Grapheme Cluster Logic
    - Ruby’s grapheme cluster regex (\\X) logic was updated to incorporate GB9c, ensuring that Indic conjuncts are not split in string operations.


NEWS.md
```diff
@@ -31,7 +36,6 @@ The following bundled gems are promoted from default gems.
 * reline 0.6.0
 * readline 0.0.4
 * fiddle 1.1.6
-
```
kou, line 34: Could you revert a needless change?

tool/enc-unicode.rb
```diff
@@ -166,9 +166,9 @@ def parse_scripts(data, categories)
         categories[current] = file[:title]
         (names[file[:title]] ||= []) << current
         cps = []
-      elsif /^(\h+)(?:\.\.(\h+))?\s*;\s*(\w+)/ =~ line
-        current = $3
+      elsif /^(\h+)(?:\.\.(\h+))?\s*;\s*(\w(?>[\w\s;]*\w)?)/ =~ line
         $2 ? cps.concat(($1.to_i(16)..$2.to_i(16)).to_a) : cps.push($1.to_i(16))
+        current = $3.gsub(/\W+/, '_')
```
kou, line 171: Could you add a comment that shows an example matched line and `$1`/`$2`/`$3` for the case for easy to understand?

ima1zumi: I have added some comments. https://github.com/ruby/ruby/pull/12798/files#diff-66b721e4f3da90f52f789b006bac8743157852b8bea2643ce866d70296acaec9R171-R179

regparse.c
```diff
@@ -6086,10 +6089,55 @@ node_extended_grapheme_cluster(Node** np, ScanEnv* env)
...
+          R_ERR(create_node_from_array(LIST, core_alts+5, CC_list));
+        }
+
         /* [^Control CR LF] */
-        core_alts[5] = node_new_cclass();
-        if (IS_NULL(core_alts[5])) goto err;
-        cc = NCCLASS(core_alts[5]);
+        core_alts[6] = node_new_cclass();
+        if (IS_NULL(core_alts[6])) goto err;
+        cc = NCCLASS(core_alts[6]);
```
ima1zumi, line 6140: This part is covered by an automatically generated test in this file based on GraphemeBreakTest.txt.

https://github.com/ruby/ruby/blob/dfc25204235079e23eadf9e0ba860c1ebcb14325/test/ruby/enc/test_grapheme_breaks.rb
https://www.unicode.org/Public/15.1.0/ucd/auxiliary/GraphemeBreakTest.txt

regparse.c
```diff
@@ -6086,10 +6089,61 @@ node_extended_grapheme_cluster(Node** np, ScanEnv* env)
...
+              }
+
+              R_ERR(quantify_node(CC_list1+3, 0, REPEAT_INFINITE));
+            }
+
+            // \p{InCB=Consonant}
+            R_ERR(create_property_node(CC_list1+4, env, "InCB_Consonant"));
+
+            // ([\p{InCB=Extend} \p{InCB=Linker}]* \p{InCB=Linker} [\p{InCB=Extend} \p{InCB=Linker}]* \p{InCB=Consonant})
+            R_ERR(create_node_from_array(LIST, CC_list1, CC_list1+1));
```
tompng, line 6133: In the original source code, it looks like array passed to `create_node_from_array(kind, target, HERE)` always have a name so that it could be listed to "Memory layout for common node array" figure.

```
 *              node_array          0 1 2 3 4 5 6 7 8 9 A B C D E F G H I
 * target       array_var_name[4]           0 1 2 3*  // size=4(including termination null), position of the array head is 0
 * foo          foo[5]                    0 1 2 3 4*  // Wrong
 * foo          (foo+1)[4]                  0 1 2 3*  // Maybe OK?
 * foo          bar[4]                      0 1 2 3*  // Better
```

```c
create_something(foo+1)
create_something(foo+2)
create_something(foo+3)
// foo+4 is a termination null
create_node_from_array(foo, foo+1) // array is (foo+1) with size=4

// ↓

bar = foo + 1 // add a name to an array
create_something(bar+0)
create_something(bar+1)
create_something(bar+2)
// bar+3 is a termination null
create_node_from_array(foo, bar)
```

ima1zumi: Thanks for letting me know. I've fixed it.

regparse.c
```diff
@@ -5948,8 +5948,10 @@ create_node_from_array(int kind, Node **np, Node **node_array)
  * h_alt2+1     H_list2[3]                              0 1 2*
  * core_alts+4  XP_list[4]                              0 1 2 3*
  * XP_list+1    Ex_list[4]                                  0 1 2 3*
+ * core_alts+6  CC_list[2]                                  0 1*
+ * CC_list+1    CC_inner_list[4]                                0 1 2 3*
```
tompng, line 5952: I think this figure should be like this.
```
 *                  node_array          0 1 2 3 4 5 6 7 8 9 A B C D E F G H I
 * top_alts         alts[5]             0 1 2 3 4*
 * alts+2           list[4]                   0 1 2 3*
 * list+1           core_alts[8]                  0 1 2 3 4 5 6 7*
 * core_alts+0      H_list[4]                       0 1 2 3*
 * H_list+1         H_alt2[4]                           0 1 2 3*
 * h_alt2+1         H_list2[3]                              0 1 2*
 * core_alts+4      XP_list[3]                              0 1 2*
 * XP_list+1        Ex_list[4]                                  0 1 2 3*
 * core_alts+5      CC_list[3]                                0 1 2*
 * CC_list+1        CC_inner_list[5]                              0 1 2 3 4*
 * CC_inner_list    CC_alt1[3]                                      0 1 2*
 * CC_inner_list+2  CC_alt2[3]                                          0 1 2*
```

`[number]*` is where NULL is required.
Changes are:
- `alts+1 list[4]` is a mistake of `alts+2 list[4]`
- `core_alts+6` is a node, `core_alts+7` is NULL (need to add 7*), so the size of `core_alts` is changed to 8 in this pull request
- `XP_list[4]` is a mistake of `XP_list[3]` (`XP_list+1` exist, `XP_list+2` is NULL, size is 3)
- `core_alts+6 CC_list[2] ... 0 1` is a misitake of `core_alts+5 CC_list[3] ... 0 1 2*`
  It is assigned to `core_alts+5`, `CC_list+1` is a node, `CC_list+2` is NULL, size is 3
- `CC_inner_list+3` is a node, `CC_inner_list+4` is NULL, so the size of CC_inner_list is 5
- Add `CC_alt1` and `CC_alt2` because this is a list of all create_node_from_array.
- start offset of `CC_list` and `CC_inner_list` is wrong

Position of `CC_inner_list[2]` (column `I`) needs to be NULL, column `J` is not required, so NODE_COMMON_SIZE is 19.

ima1zumi: I completely understand! Thanks!
I fixed the comments and refactored making the alt node.
[f869ce4](https://github.com/ruby/ruby/pull/12798/commits/f869ce46744b5b714f46600fe276838a6ee927d3)
[40d6ec9](https://github.com/ruby/ruby/pull/12798/commits/40d6ec961104a2f1bc14973d89ac0279ea3e2847)

tool/enc-unicode.rb
```diff
@@ -161,14 +161,24 @@ def parse_scripts(data, categories)
...
     data_foreach(file[:fn]) do |line|
+      # Parse Unicode data files and store code points and properties.
       if /^# Total (?:code points|elements): / =~ line
         data[current] = cps
         categories[current] = file[:title]
         (names[file[:title]] ||= []) << current
         cps = []
-      elsif /^(\h+)(?:\.\.(\h+))?\s*;\s*(\w+)/ =~ line
-        current = $3
+      elsif /^(\h+)(?:\.\.(\h+))?\s*;\s*(\w(?>[\w\s;]*\w)?)/ =~ line
```
tompng, line 170: Is there a reason to use atomic group `(?>   )` here? I think `(?:   )` is enough.

ima1zumi: I've fixed it.

nurse (approved):

PR #12801, opened 2025-02-24
String#gsub! Elide MatchData allocation when we know it can't escape

byroot:
In gsub is used with a string replacement or a map that doesn't have a default proc, we know for sure no code can cause the MatchData to escape the `gsub` call.

In such case, we still have to allocate a new MatchData because we don't know what is the lifetime of the backref, but for any subsequent match we can re-use the MatchData we allocated ourselves, reducing allocations significantly.

This partially fixes [[Misc #20652]](https://bugs.ruby-lang.org/issues/20652), except when a block is used, and partially reduce the performance impact of
[abc0304cb28cb9dcc3476993bc487884c139fd11](https://github.com/ruby/ruby/commit/abc0304cb28cb9dcc3476993bc487884c139fd11) / [[Bug #17507]](https://bugs.ruby-lang.org/issues/17507)

```
compare-ruby: ruby 3.5.0dev (2025-02-24T09:44:57Z master 5cf146399f) +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-02-24T10:58:27Z gsub-elude-match da966636e9) +PRISM [arm64-darwin24]
warming up....

|                 |compare-ruby|built-ruby|
|:----------------|-----------:|---------:|
|escape           |      3.577k|    3.697k|
|                 |           -|     1.03x|
|escape_bin       |      5.869k|    6.743k|
|                 |           -|     1.15x|
|escape_utf8      |      3.448k|    3.738k|
|                 |           -|     1.08x|
|escape_utf8_bin  |      6.361k|    7.267k|
|                 |           -|     1.14x|
```

### DIFF
--- a/internal/re.h
+++ b/internal/re.h
@@ -14,12 +14,12 @@
 /* re.c */
 VALUE rb_reg_compile(VALUE str, int options, const char *sourcefile, int sourceline);
 VALUE rb_reg_check_preprocess(VALUE);
-long rb_reg_search0(VALUE, VALUE, long, int, int);
+long rb_reg_search0(VALUE, VALUE, long, int, int, VALUE *);
 VALUE rb_reg_match_p(VALUE re, VALUE str, long pos);
 bool rb_reg_start_with_p(VALUE re, VALUE str);
 VALUE rb_reg_hash(VALUE re);
 VALUE rb_reg_equal(VALUE re1, VALUE re2);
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
Great work!  I believe this optimization is safe for the reasons you mention, and the use of MATCH_BUSY should hopefully allow this to apply to most gsub/gsub! calls, as well as other methods like scan.

PR #12805, opened 2025-02-25
[Bug #21153] Add missing op-assign to top const in `command_asgn`

nobu:
It was defined in `arg` only; add that pattern to `command_asgn` as well.


k0kubun (approved):
Thanks!

PR #12825, opened 2025-02-27
[DOC] Tweaks for Hash#merge

BurdetteLamar:


hash.c
```diff
@@ -4111,53 +4111,49 @@ rb_hash_update_by(VALUE hash1, VALUE hash2, rb_hash_update_func *func)
 
 /*
  *  call-seq:
- *    merge -> copy_of_self
  *    merge(*other_hashes) -> new_hash
  *    merge(*other_hashes) { |key, old_value, new_value| ... } -> new_hash
+ *    merge -> copy_of_self
```
peterzhu2118, line 4116: Why was this placed at the end? I think we (implicitly) usually sort the call-seq in order of least compex to the most complex.

We could also choose to omit this call-seq, since calling with no arguments is equivalent to an empty array for `other_hashes`.

PR #12841, opened 2025-03-03
Backport: [Bug #21030] Fix step for non-numeric range

paracycle:
When the end points of an inclusive range equal, `Range#step` should yields the element once.


k0kubun (approved):
Thanks!

PR #12852, opened 2025-03-04
Replace tombstone when converting AR to ST hash [Bug #21170]

jhawthorn:

``` ruby
class MyObj
  attr_reader :hash
  def initialize(val) = @hash = val
end

values = 0.downto(-16).to_a
hash = {}
values.each do |val|
  hash[MyObj.new(val)] = val
end
p hash.values
# => [0, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13, -14, -15, false, -16]
```

---

st_table reserves -1 as a special hash value to indicate that an entry has been deleted. So that that's a valid value to be returned from the hash function, do_hash replaces -1 with 0 so that it is not mistaken for the sentinel.

Previously, when upgrading an AR table to an ST table, rb_st_add_direct_with_hash was used which did not perform the same conversion, this could lead to a hash in a broken state where one if its entries which was supposed to exist being marked as a tombstone.

The hash could then become further corrupted when the ST table required resizing as the falsely tombstoned entry would be skipped but it would be counted in num entries, leading to an uninitialized entry at index 15.

In most cases this will be really rare, unless using a very poorly implemented custom hash function.

This also adds two debug assertions, one that st_add_direct_with_hash does not receive the reserved hash value, and a second in rebuild_table_with, which ensures that after we rebuild/compact a table it contains the expected number of elements.

### DIFF
--- a/st.c
+++ b/st.c
@@ -103,12 +103,14 @@
 #ifdef NOT_RUBY
 #include "regint.h"
 #include "st.h"
+#include <assert.h>
 #elif defined RUBY_EXPORT
 #include "internal.h"
 #include "internal/bits.h"
 #include "internal/hash.h"
 #include "internal/sanitizers.h"
 #include "internal/st.h"
+#include "ruby_assert.h"
 #endif
 
 #include <stdio.h>
@@ -116,7 +118,6 @@
 #include <stdlib.h>
 #endif
 #include <string.h>
-#include <assert.h>
 
 #ifdef __GNUC__
 #define PREFETCH(addr, write_p) __builtin_prefetch(addr, write_p)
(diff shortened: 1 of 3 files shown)


st.c
```diff
@@ -784,6 +789,8 @@ rebuild_table_with(st_table *const new_tab, st_table *const tab)
         new_tab->num_entries++;
         ni++;
     }
+
+    RUBY_ASSERT(new_tab->num_entries == tab->num_entries);
```
XrXr, line 793: So people seem to copy paste this file out and compile outside of ruby, which is a problem for the asserts. 2 potential solutions:

1. Use assert(3) for `NOT_RUBY` and ruby_assert.h otherwise (might trip `make update-deps`):

```patch
diff --git a/st.c b/st.c
index 7979f8c453..f41120ecaf 100644
--- a/st.c
+++ b/st.c
@@ -103,12 +103,14 @@
 #ifdef NOT_RUBY
 #include "regint.h"
 #include "st.h"
+#include <assert.h>
 #elif defined RUBY_EXPORT
 #include "internal.h"
 #include "internal/bits.h"
 #include "internal/hash.h"
 #include "internal/sanitizers.h"
 #include "internal/st.h"
+#include "ruby_assert.h"
 #endif
 
 #include <stdio.h>
@@ -116,7 +118,6 @@
 #include <stdlib.h>
 #endif
 #include <string.h>
-#include <assert.h>
 
 #ifdef __GNUC__
 #define PREFETCH(addr, write_p) __builtin_prefetch(addr, write_p)
@@ -790,7 +791,7 @@ rebuild_table_with(st_table *const new_tab, st_table *const tab)
         ni++;
     }
 
-    RUBY_ASSERT(new_tab->num_entries == tab->num_entries);
+    assert(new_tab->num_entries == tab->num_entries);
 }
 
 static void
@@ -1180,7 +1181,7 @@ st_add_direct_with_hash(st_table *tab,
     st_index_t ind;
     st_index_t bin_ind;
 
-    RUBY_ASSERT(hash != RESERVED_HASH_VAL);
+    assert(hash != RESERVED_HASH_VAL);
 
     rebuild_table_if_necessary(tab);
     ind = tab->entries_bound++;
```

2. Surround RUBY_ASSERT() uses with `#ifdef RUBY` like other places in this file.

jhawthorn: 👍 Good call. I went with option 1

PR #12853, opened 2025-03-05
Restore vs2022 build with -O2 option.

hsbt:
from https://developercommunity.visualstudio.com/t/Use-of-the-result-of-bsr-before-that-i/10862351#T-N10863092-N10863860


hsbt:
We should wait to release VS 2022 17.14.0.

PR #12856, opened 2025-03-05
[DOC] Fix parameter name in YJIT.enable doc

a5-stable:
Fix a documentation-implementation mismatch in the parameter name; `compilation_log` -> `log`


XrXr (approved):
Thanks!

PR #12858, opened 2025-03-05
rdoc: Enable the "embed mixins" feature

flavorjones:
which will include methods, attributes, and constants from any included or extended Modules.

ref: https://ruby.github.io/rdoc/RDoc/Options.html#attribute-i-embed_mixins

As an example of the effect this change has, methods from `Enumerable` are documented in the `Array` docs:

image


nobu:
It would make those lists much longer by duplicate entries.
I'm not sure if it is preferable.

PR #12860, opened 2025-03-05
[Backport 3.2] [Bug #21170] Replace tombstone when converting AR to ST hash

jhawthorn:
https://github.com/ruby/ruby/pull/12852


hsbt:
@jhawthorn Thanks, I will fix the current CI failure today.

PR #12862, opened 2025-03-05
[Backport 3.4] [Bug #21170] Replace tombstone when converting AR to ST hash

jhawthorn:
https://github.com/ruby/ruby/pull/12852


k0kubun (approved):
Thanks for the backport PR. Great job on finding it by the way.

PR #12864, opened 2025-03-05
Update actions/cache for ruby_3_3

k0kubun:


k0kubun:
Let me quickly merge this to unblock https://github.com/ruby/ruby/pull/12863

PR #12879, opened 2025-03-07
Beginless max

nobu:


nobu:
https://github.com/ruby/ruby/pull/12879

PR #12880, opened 2025-03-07
Avoid pinning `storage_head` in `iseq_mark_and_move`

eileencodes:
This refactor changes the behavior of `iseq_mark_and_move` to avoid pinning the `storage_head`. Previously pinning was required because they could be gc'd during `iseq_set_sequence` it would be possible to end up with a half build array of instructions. However, in order to implement a moving immix algorithm we can't pin these objects so this refactoring changes the code to mark and move. To accomplish this, it was required to add `iseq_size`, `iseq_encoded`, and the `mark_bits` union to the `iseq_compile_data` struct. In addition `iseq_compile_data` sets a bool for whether there is a single or list of mark bits. While this change is needed for moving immix, it should be better for Ruby's GC as well.

cc/ @tenderlove @peterzhu2118

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -2609,12 +2609,18 @@ iseq_set_sequence(rb_iseq_t *iseq, LINK_ANCHOR *const anchor)
 
     bool needs_bitmap = false;
 
-    mark_offset_bits = ZALLOC_N(iseq_bits_t, ISEQ_MBITS_BUFLEN(code_index));
+    if (ISEQ_MBITS_BUFLEN(code_index) == 1) {
+        mark_offset_bits = &ISEQ_COMPILE_DATA(iseq)->mark_bits.single;
+        ISEQ_COMPILE_DATA(iseq)->is_single_mark_bit = true;
+    }
+    else {
+        mark_offset_bits = ZALLOC_N(iseq_bits_t, ISEQ_MBITS_BUFLEN(code_index));
+        ISEQ_COMPILE_DATA(iseq)->mark_bits.list = mark_offset_bits;
+        ISEQ_COMPILE_DATA(iseq)->is_single_mark_bit = false;
+    }
 
     ISEQ_COMPILE_DATA(iseq)->iseq_encoded = (void *)generated_iseq;
     ISEQ_COMPILE_DATA(iseq)->iseq_size = code_index;
-    ISEQ_COMPILE_DATA(iseq)->mark_bits.list = mark_offset_bits;
-    ISEQ_COMPILE_DATA(iseq)->is_single_mark_bit = false;
 
     list = FIRST_ELEMENT(anchor);
     insns_info_index = code_index = sp = 0;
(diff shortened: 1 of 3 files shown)


compile.c
```diff
@@ -2607,15 +2607,14 @@ iseq_set_sequence(rb_iseq_t *iseq, LINK_ANCHOR *const anchor)
...
-    }
-    else {
-        mark_offset_bits = ZALLOC_N(iseq_bits_t, ISEQ_MBITS_BUFLEN(code_index));
-    }
+    mark_offset_bits = ZALLOC_N(iseq_bits_t, ISEQ_MBITS_BUFLEN(code_index));
+
+    ISEQ_COMPILE_DATA(iseq)->iseq_encoded = (void *)generated_iseq;
+    ISEQ_COMPILE_DATA(iseq)->iseq_size = code_index;
+    ISEQ_COMPILE_DATA(iseq)->mark_bits.list = mark_offset_bits;
+    ISEQ_COMPILE_DATA(iseq)->is_single_mark_bit = false;
```
peterzhu2118, line 2617: Doesn't this render the single mark bit optimization useless? It's no longer conditionally allocating the `mark_offset_bits` but rather always allocating it and then freeing when it only needs a single word?

eileencodes: The reason we're doing this is because `tmp` is stack allocated, and the garbage collector needs to access that memory if a GC happens to occur during complication. We decided to always malloc so that if a GC happens, it can access those mark bits. Later we free it. So the optimization isn't lost because overall process memory should remain the same. In other words, in some cases we will need to do an extra malloc. Unless we can point at stack allocated memory, which I don't think we can.

peterzhu2118: We can take pointers to stack allocated memory, but we shouldn't need to do that as we can just point it to the `ISEQ_COMPILE_DATA(iseq)->mark_bits.single`.

peterzhu2118 (commented):
I think we also need to make the same changes to `ibf_load_code` for loading serialized iseqs.

compile.c
eileencodes: Looked at this with Aaron. We can't point `ISEQ_COMPILE_DATA(iseq)->mark_bits.single` to `tmp[0]` because it's a copy of the value (integer). Later in the function when the bits are marked, it'd be marking the value in `tmp[0]`, so the integer copy in `mark_bits.single` wouldn't get updated and the GC wouldn't see changes. So it's either this implementation or take a pointer to the stack memory. Are we misunderstanding what you're saying?

peterzhu2118: I pushed a commit that should do this.

eileencodes:
> I think we also need to make the same changes to ibf_load_code for loading serialized iseqs.

In `ibf_load_code` the `ISEQ_COMPILE_DATA(iseq)` is null, and there's no pinning so I think we're actually good here without changes to that function?

peterzhu2118:
Ah yeah it looks like for iseq loading there's a separate object `pinned_list` that guarantees that all the objects are marked and pinned.

peterzhu2118 (approved):
Looks good to me!

PR #12887, opened 2025-03-08
Organize the generation rules of `primary` using Parameterizing rules `?`

ydah:


nobu:
Is it better than the current code that selects by rule?

PR #12889, opened 2025-03-09
[Bug #21176] Fix case-insensitive matching for single-byte encodings

ima1zumi:
https://bugs.ruby-lang.org/issues/21176

### DIFF
--- a/regparse.c
+++ b/regparse.c
@@ -5669,31 +5669,44 @@ i_apply_case_fold(OnigCodePoint from, OnigCodePoint to[],
     if ((is_in != 0 && !IS_NCCLASS_NOT(cc)) ||
 	(is_in == 0 &&  IS_NCCLASS_NOT(cc))) {
       if (add_flag) {
-	if (ONIGENC_MBC_MINLEN(env->enc) > 1 || *to >= 0x80) {
-	  r = add_code_range0(&(cc->mbuf), env, *to, *to, 0);
-	  if (r < 0) return r;
-	}
-	else {
-	  BITSET_SET_BIT(bs, *to);
-	}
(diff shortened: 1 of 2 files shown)


regparse.c
```diff
@@ -5669,31 +5669,44 @@ i_apply_case_fold(OnigCodePoint from, OnigCodePoint to[],
...
+        if (ONIGENC_MBC_MAXLEN(env->enc) == 1) {
+          // single byte encoding
+          if (IS_NCCLASS_NOT(cc)) {
+            BITSET_CLEAR_BIT(bs, *to);
+          } else {
+            BITSET_SET_BIT(bs, *to);
+          }
+        } else {
+          // multi byte encoding && to >= 0x80
+          if (ONIGENC_MBC_MINLEN(env->enc) > 1 || *to >= 0x80) {
```
nobu, line 5698:
```suggestion
        if ((ONIGENC_MBC_MINLEN(env->enc) > 1) /* wide char encoding */ ||
            (ONIGENC_MBC_MAXLEN(env->enc) > 1 && *to >= 0x80) /* non-ASCII char */) {
```

nobu: How about a helper function:
```C
static inline bool
is_singlebyte_range(const OnigCodePoint *to, OnigEncoding enc)
{
    if (ONIGENC_MBC_MAXLEN(enc) == 1) return true; /* single byte encoding */
    if (ONIGENC_MBC_MINLEN(enc) > 1) return false; /* wide char encoding */
    return *to < 0x80;
}
```

and make the conditions `if (!is_singlebyte_range(to, env->enc))`?

nobu (approved):

PR #12893, opened 2025-03-10
retry on failure for weakref test

ko1:
On `test_latest_gc_info_weak_references_count`, it expects all of wmap keys are removed because of weakref. However, Ruby's conservertive GC can keep objects from conservertive logic. In other words, this test has the potential to fail.

(and my ASAN builds, it fails 26 times per 188 trials in a day)

This patch introduce "retry" on fail.


peterzhu2118:
I think it's better to just increase the tolerance, WDYT? https://github.com/ruby/ruby/pull/12894

PR #12900, opened 2025-03-11
Embed const caches that aren't cref dependent

byroot:
Opening this as a draft to get some CI coverage. I already know RJIT is broken.

Currently all `iseq_inline_constant_cache` have extra storage inside an IMEMO/constcache object.
That extension object stores one flag (so one bit), the reference to the cached value (8B) and the `cref` pointer (8B).

However the vast majority of const caches don't have a `cref`, which means we're allocating `40B`, and use `8B` to reference it, to store `8B` plus 1 bit.

Given that `iseq_inline_constant_cache.segments` is a pointer, we can use the lower 3 bits to store tags.

The idea here is to merge `iseq_inline_constant_cache_entry.flags` into `iseq_inline_constant_cache.segments` lower bits,
and also store an extra flag that records whether this cache is `cref` dependent.

When it's not (which is the vast majority of cases) we can store the cached value directly inside the inline cache, and not need the `_entry` object.

Expectations:

  - The vast majority of `IMEMO/constcache` instances are eliminated.
  - We're saving one pointer chasing, in exchange for a bitmask check and some pointer untagging.

cc @etiennebarrie

### DIFF
--- a/vm_core.h
+++ b/vm_core.h
@@ -309,6 +309,18 @@ vm_cc_flags(const struct iseq_inline_constant_cache *cc)
     return cc->entry->flags;
 }
 
+static inline VALUE
+vm_cc_value(const struct iseq_inline_constant_cache *cc)
+{
+    return cc->entry->value;
+}
+
+static inline void
+vm_cc_set_value(struct iseq_inline_constant_cache *cc, VALUE value)
+{
+    RB_OBJ_WRITE(cc->entry, &cc->entry->value, value);
+}
+
 struct iseq_inline_iv_cache_entry {
     uintptr_t value; // attr_index in lower bits, dest_shape_id in upper bits
     ID iv_set_name;
--- a/vm_core.h
+++ b/vm_core.h
@@ -312,6 +312,14 @@ vm_icc_is_set(const struct iseq_inline_constant_cache *cc)
 }
 
 static inline void
+vm_icc_reset(struct iseq_inline_constant_cache *cc)
+{
+    if (!UNDEF_P(cc->value)) {
+        cc->value = Qundef;
+    }
+}
+
+static inline void
 vm_icc_init(struct iseq_inline_constant_cache *cc, const ID *segments)
 {
     cc->value = Qundef;
--- a/vm_core.h
+++ b/vm_core.h
@@ -286,49 +286,49 @@ struct iseq_inline_constant_cache {
 };
 
 static inline void
-vm_cc_set_segments(struct iseq_inline_constant_cache *cc, const ID *segments)
+vm_icc_set_segments(struct iseq_inline_constant_cache *cc, const ID *segments)
 {
     cc->segments = segments;
 }
 
 static inline const ID *
-vm_cc_segments(const struct iseq_inline_constant_cache *cc)
+vm_icc_segments(const struct iseq_inline_constant_cache *cc)
 {
     return cc->segments;
 }
 
 static inline void
-vm_cc_set_flag(struct iseq_inline_constant_cache *cc, VALUE flag)
+vm_icc_set_flag(struct iseq_inline_constant_cache *cc, VALUE flag)
 {
     cc->entry->flags |= flag;
 }
 
 static inline VALUE
-vm_cc_flags(const struct iseq_inline_constant_cache *cc)
+vm_icc_flags(const struct iseq_inline_constant_cache *cc)
 {
     return cc->entry->flags;
 }
 
 static inline VALUE
-vm_cc_value(const struct iseq_inline_constant_cache *cc)
+vm_icc_value(const struct iseq_inline_constant_cache *cc)
 {
     return cc->entry->value;
 }
 
 static inline void
-vm_cc_set_value(struct iseq_inline_constant_cache *cc, VALUE value)
+vm_icc_set_value(struct iseq_inline_constant_cache *cc, VALUE value)
 {
     RB_OBJ_WRITE(cc->entry, &cc->entry->value, value);
 }
 
 static inline const rb_cref_t *
-vm_cc_cref(const struct iseq_inline_constant_cache *cc)
+vm_icc_cref(const struct iseq_inline_constant_cache *cc)
 {
     return cc->entry->ic_cref;
 }
 
 static inline void
-vm_cc_set_cref(struct iseq_inline_constant_cache *cc, const rb_cref_t *cref)
+vm_icc_set_cref(struct iseq_inline_constant_cache *cc, const rb_cref_t *cref)
 {
     cc->entry->ic_cref = cref;
 }
--- a/vm_core.h
+++ b/vm_core.h
@@ -297,6 +297,18 @@ vm_cc_segments(const struct iseq_inline_constant_cache *cc)
     return cc->segments;
 }
 
+static inline void
+vm_cc_set_flag(struct iseq_inline_constant_cache *cc, VALUE flag)
+{
+    cc->entry->flags |= flag;
+}
+
+static inline VALUE
+vm_cc_flags(const struct iseq_inline_constant_cache *cc)
+{
+    return cc->entry->flags;
+}
+
 struct iseq_inline_iv_cache_entry {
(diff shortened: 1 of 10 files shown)


byroot:
> I already know RJIT is broken.

Actually, it might not be, because it doesn't seem to use const caches that don't have a cref.

byroot:
@ruby/yjit would someone have a bit of time to help me update YJIT to handle the new constcache layout?

I suspect I need to check for the `HAS_EXT` flag in the generated YJIT code, but I'm not quite certain, it's the best solution. It might be better to just invalidate the code to not have to generate an extra branch?

vm_insnhelper.c
```diff
@@ -6386,33 +6386,41 @@ vm_inlined_ic_hit_p(VALUE flags, VALUE value, const rb_cref_t *ic_cref, const VA
...
-    if (rb_ractor_shareable_p(val)) ice->flags |= IMEMO_CONST_CACHE_SHAREABLE;
-    RB_OBJ_WRITE(iseq, &ic->entry, ice);
+    const rb_cref_t *cref = vm_get_const_key_cref(reg_ep);
+    if (cref) {
+        vm_icc_set_flag(ic, CONST_CACHE_HAS_EXT);
+        struct iseq_inline_constant_cache_entry *ice = IMEMO_NEW(struct iseq_inline_constant_cache_entry, imemo_constcache, 0);
+        RB_OBJ_WRITE(iseq, &ic->c.ext, ice);
+        RB_OBJ_WRITE(ic->c.ext, &ice->value, val);
+        ice->ic_cref = cref;
+    }
```
XrXr, line 6418: So @ko1 made it always allocate an imemo for ractor data race reasons, that way cache fills are always an atomic pointer swap: e7fc353f044f9280222ca41b029b1368d2bf2fe3

The multiple pointer sets here doesn't look thread safe.

byroot: Oh, that's good info thank you. I'll see if I can preserve the thread safety. I think with a bit of care it is possible.

byroot: Hum, actually, not so sure, the problem is really that both `RACTOR_SHAREABLE` flag and the `VALUE` have to be set atomically, so short of having access to 128bit atomics, I don't see how it could be achieved 😢

byroot:
Reopening, because this might actually be possible using `CAS`.

vm_core.h
```diff
@@ -269,22 +273,91 @@ STATIC_ASSERT(sizeof_iseq_inline_constant_cache_entry,
                sizeof(const rb_cref_t *)) <= RVALUE_SIZE);
 
 struct iseq_inline_constant_cache {
-    struct iseq_inline_constant_cache_entry *entry;
+    union {
+        struct iseq_inline_constant_cache_entry *ext;
```
k0kubun, line 277: What does "ext" stand for? "Ext"ended? I find this name confusing.

Could we just call this `extended` and rename the flag to `CONST_CACHE_HAS_EXTENDED` (if that's what it means)? Or maybe "heap" and `CONST_CACHE_IN_HEAP` (or `CONST_CACHE_EMBEDDED`).

byroot: Yes that what it stand for. If I managed to make this work, I'll rename it as you suggested.

byroot:
@XrXr I think I pushed a new version that work with ractors, but I'm not super accustomed to all the constraints so I'd really appreciate if you could review this. 

How it works now:

  - `IC->value` is initialized to `Qundef`, and the `segments` pointer is not tagged by default.
  - When a thread hits a miss, it does the constant resolution like before, and assign the cached value using `CAS`
  - `IC->value` is either the constant value, or an IMEMO, the caller has to check for that
  - The `CONST_CACHE_SHAREABLE` flag isn't set atomically, but we can call `is_shearable_p` when it is missing since we'd raise an exception anyway, so the overhead is negligible, and this should only happen when there is a race.

Here I somewhat assume `vm_icc_reset` is only called with the ractor lock, but if I'm wrong I think we can get away with removing the flag before setting `IC->value = Qundef`?

How does that sound?

vm_insnhelper.c
```diff
@@ -6372,47 +6372,73 @@ vm_ic_track_const_chain(rb_control_frame_t *cfp, IC ic, const ID *segments)
...
-    RB_OBJ_WRITE(iseq, &ic->entry, ice);
+    VALUE cache_val = val;
+
+    const rb_cref_t *cref = vm_get_const_key_cref(reg_ep);
+    if (RB_UNLIKELY(cref)) {
+        struct iseq_inline_constant_cache_entry *ice = IMEMO_NEW(struct iseq_inline_constant_cache_entry, imemo_constcache, 0);
+        RB_OBJ_WRITE(ice, &ice->value, val);
+        ice->ic_cref = cref;
+        cache_val = (VALUE)ice;
+    }
```
maximecb, line 6427: Should add a short comment to explain what this bit does. What does it mean for cref to be null or not null?

XrXr (commented):
I don't think letting `CONST_CACHE_SHAREABLE` race this way is correct. Since it's not set atomically with the cached value, it's possible to observe a stale set flag from another thread, for a value that's actually not sharable, allowing for smuggling of values between ractors without moving.

byroot:
😞, do you have any easy to understand literature you'd recommend on this topic? I'd like to better understand the constraints here.

While I get the write are not atomic, I don't understand if they're ordered, what's the granularity etc.

Alternatively I looked at 128bit atomics, but from what I could find it seems support is rather erratic.

XrXr:
Sorry, I misspoke. This is actually about atomicity, not ordering. The possible schedule is this:

```
Thread 1       Thread 2           Thread 3

set value
             set unsharable value
set flag                        
                                  read value
                                  read flag
             clear flag
```

Generally if two states are not set together atomically, all possible combinations of the two are observable. So even if you put clear before setting the value, a smuggling schedule is still possible.

Sorry, I know you two put effort into this and it's not fun to learn about the undocumented intentions, especially since the current code is not 100% correct either. Because it's not using atomics, the writes during a cache fill can be observed in any order, including one where the other thread sees the imemo pointer, but not the initialized contents of the pointed to imemo. (This one is actually about ordering.)

Memory ordering is a complex topic and takes a while to build up intuition. This is one of blog posts I read: https://preshing.com/20120710/memory-barriers-are-like-source-control-operations/ but I learned from a bunch of other sources. Another phrase to search around for is "Acquire-Release semantics".

byroot:
Ok, I think I understand what you mean, and thank you for the link.

If I understood the link correctly, given that the cache is only ever cleared from the main ractor (these are const cache), and AFAIK with the global lock held, we could clear the flag before clearing the value:

```c
vm_icc_reset(*ic)
{
        cc->tagged_segments &= ~CONST_CACHE_FLAGS_MASK;
        // Insert StoreStore fence (not sure how to do that)
        cc->value = Qundef;
}
```

This way, it's still possible to see a sharable value without the corresponding flag, but never an unsharable value with the flag. no?

If I'm still incorrect just say so, it's ok, I don't want to take your time any further.

byroot:
Ok, thank you. So short of having 128bit atomics, or of packing that bit in the `VALUE`, there isn't a proper way to re-inline these caches 😢

XrXr:
You could use a spin lock to synchronize fills and reads I guess, but that sacrifices single threaded speed.

byroot:
Or I could inline them only if no other ractor has been spawned yet?

XrXr:
That could work. Since there's no way to transition from embedded to not without 128B atomics, you can treat all embedded caches as busts in multi-ractor mode and fill and clear the flag. And make it so only share-ability claims in imemos be valid.

That sounds like it irons out all the inconsistent states. It's hard to reason about, though, and I might be missing something.

byroot:
> you can treat all embedded caches as busts in multi-ractor mode and fill and clear the flag.

I don't think you need to though. You can keep the embeded caches even after a second ractor has been spawned.

XrXr:
Ah right. No one writes embedded caches anymore after a ractor spawns.

PR #12901, opened 2025-03-11
[Bug #21177] Win32: Allow longer path name

nobu:
[[Bug #21177]](https://bugs.ruby-lang.org/issues/21177)


KostyaSha:
Hi, @nobu does it fix this issue https://bugs.ruby-lang.org/issues/18947 ?

PR #12905, opened 2025-03-11
Fix memory leak in rb_reg_search_set_match

peterzhu2118:
https://github.com/ruby/ruby/pull/12801 changed regexp matches to reuse the backref, which causes memory to leak if the original registers of the match is not freed.

For example, the following script leaks memory:

```ruby
10.times do
  1_000_000.times do
    "aaaaaaaaaaa".gsub(/a/, "")
  end

  puts `ps -o rss= -p #{$$}`
end
```

Before:

    774256
    1535152
    2297360
    3059280
    3821296
    4583552
    5160304
    5091456
    5114256
    4980192

After:

    12480
    11440
    11696
    11632
    11632
    11760
    11824
    11824
    11824
    11888

### DIFF
--- a/re.c
+++ b/re.c
@@ -1829,6 +1829,9 @@ rb_reg_search_set_match(VALUE re, VALUE str, long pos, int reverse, int set_back
     if (NIL_P(match)) {
         match = match_alloc(rb_cMatch);
     }
+    else {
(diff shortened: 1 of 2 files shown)


byroot (approved):
❤️ :MVP:

PR #12921, opened 2025-03-12
Lock free hash set for fstrings

jhawthorn:
This implements a hash set which is wait-free for lookup and lock-free for insert (unless resizing) to use for fstring de-duplication.

As highlighted in https://bugs.ruby-lang.org/issues/19288, heavy use of fstrings (frozen interned strings) can significantly reduce the parallelism of Ractors.

I was at first intimidated by writing the lock-free version. I previously tried a few other approaches: using an RWLock, striping a series of RWlocks (partitioning the hash N-ways to reduce lock contention), and putting a cache in front of it. All of these improved the situation, but were unsatisfying as all still required locks for writes (and granular locks are awkward, since we run the risk of needing to reach a vm barrier) and this table is somewhat write-heavy.

My main reference for this was Cliff Click's talk on a lock free hash-table for java https://www.youtube.com/watch?v=HJ-719EGIts. It turns out this lock-free hash set is made easier to implement by a few properties:
* We only need a hash set rather than a hash table (we only need keys, not values), and so the full entry can be written as a single `VALUE`
* As a set we only need lookup/insert/delete, no update
* Delete is only run inside GC so does not need to be atomic (It could be made concurrent)
* ~I use rb_vm_barrier for the (rare) table rebuilds (It could be made concurrent)~ We VM lock (but don't require other threads to stop) for table rebuilds, as those are rare
* The conservative garbage collector makes deferred replication easy, using a T_DATA object

Another benefits of having a table specific to fstrings is that we compare by value on lookup/insert, but by identity on delete, as we only want to remove the exact string which is being freed. This is faster and provides a second way to avoid the race condition in https://bugs.ruby-lang.org/issues/21172.

This is a pretty standard open-addressing hash table with quadratic probing. Similar to our existing st_table or id_table. Deletes (which happen on GC) replace existing keys with a tombstone, which is the only type of update which can occur. Tombstones are only cleared out on resize.

Unlike st_table, the VALUEs are stored in the hash table itself (st_table's bins) rather than as a compact index. This avoids an extra pointer dereference and is possible because we don't need to preserve insertion order. The table targets a load factor of 2 (it is enlarged once it is half full).

### DIFF
--- a/internal/string.h
+++ b/internal/string.h
@@ -83,6 +83,7 @@ VALUE rb_setup_fake_str(struct RString *fake_str, const char *name, long len, rb
 RUBY_SYMBOL_EXPORT_END
 
 VALUE rb_fstring_new(const char *ptr, long len);
+void rb_gc_free_fstring(VALUE obj);
 VALUE rb_obj_as_string_result(VALUE str, VALUE obj);
 VALUE rb_str_opt_plus(VALUE x, VALUE y);
 VALUE rb_str_concat_literals(size_t num, const VALUE *strary);
--- a/internal/string.h
+++ b/internal/string.h
@@ -84,6 +84,8 @@ RUBY_SYMBOL_EXPORT_END
 
 VALUE rb_fstring_new(const char *ptr, long len);
 void rb_gc_free_fstring(VALUE obj);
+bool rb_obj_is_fstring_table(VALUE obj);
+void Init_fstring_table();
 VALUE rb_obj_as_string_result(VALUE str, VALUE obj);
 VALUE rb_str_opt_plus(VALUE x, VALUE y);
 VALUE rb_str_concat_literals(size_t num, const VALUE *strary);
--- a/string.c
+++ b/string.c
@@ -383,11 +383,6 @@ fstring_hash(VALUE str)
 #define fstring_hash rb_str_hash
 #endif
 
-const struct st_hash_type rb_fstring_hash_type = {
-    fstring_cmp,
-    fstring_hash,
-};
-
 #define BARE_STRING_P(str) (!FL_ANY_RAW(str, FL_EXIVAR) && RBASIC_CLASS(str) == rb_cString)
 
 static inline st_index_t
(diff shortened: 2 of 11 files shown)


string.c
```diff
@@ -543,6 +518,150 @@ rb_fstring(VALUE str)
...
+    fstring_table.capacity = new_capacity;
+    fstring_table.entries = new_entries;
+    fstring_table.count = 0;
+    fstring_table.deleted_entries = 0;
+
+    for (int i = 0; i < old_capacity; i++) {
+        VALUE val = old_entries[i];
+        if (val == FSTRING_TABLE_EMPTY)     continue;
+        if (val == FSTRING_TABLE_TOMBSTONE) continue;
+        fstring_upsert(str_do_hash(val), val, true);
```
byroot, line 626: Is there a reason you use `str_do_hash`? I don't know if you've seen https://github.com/ruby/ruby/pull/12881 but `str_do_hash` does cause quite a lot of collisions in the `fstring_table`.

Also most if not all fstrings have a precomputed hash. So I think it would be better if you re-used `fstring_hash`.

jhawthorn: Thanks! I missed that, we can use `fstring_hash`.

One thing I also haven't settled on is whether the hash table should cache the hash value for each string. It's probably safer to do so but if we thought the precomputed hash was there very often maybe we could avoid it.

I saw https://bugs.ruby-lang.org/issues/21172 when working on this and I think have an additional fix for that in this PR: in delete we compare by identity, so it shouldn't be possible for equivalent strings to be evicted.

byroot: > if we thought the precomputed hash was there very often maybe we could avoid it.

All literal strings will have it (unless they can't be embedded), and all embeded strings that have enough space left.

Not too sure what the exact rate is in practice, but it would be easy to compute from a heap dump.

That said, in term of locality, it's an extra pointer chasing, but I don't know if that makes that much of a difference.

string.c
```diff
@@ -543,6 +518,150 @@ rb_fstring(VALUE str)
...
+    for (;;) {
+        VALUE candidate = table->entries[idx];
+
+        if (candidate == FSTRING_TABLE_EMPTY) {
+            // Not in table
+            if (insert) {
+                int prev_count = RUBY_ATOMIC_FETCH_ADD(table->count, 1);
+
+                if (UNLIKELY(prev_count > table->capacity / 2)) {
+                    fstring_try_resize();
```
luke-gru, line 557: On a resize, we've already added to `table->count`, next time around it will do so again, no?

string.c
```diff
@@ -543,6 +518,150 @@ rb_fstring(VALUE str)
...
+                // Skip it and mark it is a tombstone
+                RUBY_ATOMIC_VALUE_CAS(table->entries[idx], candidate, FSTRING_TABLE_TOMBSTONE);
+            } else {
+                return candidate;
+            }
+        }
+
+        // linear probing (TODO: something better)
+        idx++;
+        idx %= table->capacity;
```
luke-gru, line 591: Since `table->capacity` is a power of 2, I think you can do a trick to avoid this modulo by doing a simple `AND`. See https://stackoverflow.com/questions/6670715/mod-of-power-2-on-bitwise-operators . It might speed things up slightly in case of collisions.

string.c
jhawthorn: This isn't ready for review. I will be doing that after working on avoiding the calls to vm_barrier

internal/string.h
```diff
@@ -82,6 +82,8 @@ VALUE rb_setup_fake_str(struct RString *fake_str, const char *name, long len, rb
 RUBY_SYMBOL_EXPORT_END
 
 VALUE rb_fstring_new(const char *ptr, long len);
+void rb_gc_free_fstring(VALUE obj);
+void Init_fstring_table();
```
nobu, line 86: Use the prototype declaration.

```suggestion
void Init_fstring_table(void);
```

string.c
```diff
@@ -420,90 +420,65 @@ str_store_precomputed_hash(VALUE str, st_index_t hash)
...
-static int
-fstr_update_callback(st_data_t *key, st_data_t *value, st_data_t data, int existing)
-{
-    struct fstr_update_arg *arg = (struct fstr_update_arg *)data;
-    VALUE str = (VALUE)*key;
-
-    if (existing) {
-        /* because of lazy sweep, str may be unmarked already and swept
-         * at next time */
+static VALUE build_fstring(VALUE str, struct fstr_update_arg *arg) {
```
nobu, line 427: Would you align styles with the rest?

```suggestion
static VALUE
build_fstring(VALUE str, struct fstr_update_arg *arg)
{
```

PR #12927, opened 2025-03-13
Invoke `inherited` callbacks before `const_added`

byroot:
[[Misc #21143]](https://bugs.ruby-lang.org/issues/21143)

Conceptually this makes sense and is more consistent with using the `Name = Class.new(Superclass)` alternative method.

However the new class is still named before `inherited` is called.


spec/ruby/core/class/inherited_spec.rb
```diff
@@ -98,4 +99,16 @@ class << top; protected :inherited; end
     -> { Class.new(top) }.should_not raise_error
   end
 
+  it "is invoked after the class is named when using class definition syntax" do
+    CoreClassSpecs::Callbacks::Child::INHERITED_NAME.should == "CoreClassSpecs::Callbacks::Child"
+  end
+
+  ruby_version_is "3.5" do
```
eregon, line 106:
```suggestion
  ruby_version_is "3.5" do # https://bugs.ruby-lang.org/issues/21143
```
Seems nice to find out more about this change :)

eregon (approved):
LGTM

spec/ruby/core/class/fixtures/callback_order.rb
```diff
@@ -0,0 +1,21 @@
...
+        subclass.const_set(:INHERITED_NAME, subclass.name)
+        ORDER << [:inherited, subclass, eval("defined?(#{subclass.name})")]
+        super
+      end
+    end
+
+    ORDER = []
+
+    def self.const_added(const_name)
+      ORDER << [:const_added, const_name, eval("defined?(#{const_name})")]
```
byroot, line 14: I added this `defined?` check, to spec the other change, which is that the constant isn't yet defined when `inherited` is called.

If something causes backward compatibility issues, I think it will be this part. We could totally register the constant before, but the code would be sensibly more finicky, and I understood this was the behavior Matz wanted to try out

PR #12930, opened 2025-03-13
[BACKPORT] Push a real iseq in rb_vm_push_frame_fname()

tenderworks:
Previously, vm_make_env_each() (used during proc
creation and for the debug inspector C API) picked up the non-GC-allocated iseq that rb_vm_push_frame_fname() creates, which led to a SEGV when the GC tried to mark the non GC object.

Put a real iseq imemo instead. Speed should be about the same since the old code also did a imemo allocation and a malloc allocation.

Real iseq allows ironing out the special-casing of dummy frames in rb_execution_context_mark() and rb_execution_context_update(). A check is added to RubyVM::ISeq#eval, though, to stop attempts to run dummy iseqs.

[Bug #21180]

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -1749,6 +1749,7 @@ iseqw_s_compile_file(int argc, VALUE *argv, VALUE self)
                                          1, NULL, 0, ISEQ_TYPE_TOP, &option,
                                          Qnil));
     rb_ast_dispose(ast);
+    RB_GC_GUARD(ast_value);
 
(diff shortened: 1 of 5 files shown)


tenderworks:
@k0kubun yes, I think these failures are relevant. I need some time to understand the problem

tenderworks:
@k0kubun I think this is fixed now. We needed to backport 2 commits

PR #12936, opened 2025-03-15
Clear `cc->cme_` after invalidating `cc` during compaction

XrXr:
We observed T_NONE on `cc->cme_` on a --repeat-count=50 run a compaction
test:
http://ci.rvm.jp/results/trunk-repeat50@ruby-sp2-noble-docker/5654900

During reference updating for imemo_callcache in
rb_imemo_mark_and_move(), if `cc->class` is not live, but `cc->_cme` is
live and moved, we fall to the vm_cc_invalidate() path. Previously, that
left `cc->_cme` not updated and stale. In the next marking run after
compaction, it would've saw a T_NONE.


peterzhu2118 (approved):
This patch makes sense. Could we alternatively set `cme_ = NULL` in `vm_cc_invalidate` so the object doesn't have a reference to a dead object?

PR #12942, opened 2025-03-16
[DOC] Doc for Hash#transform_keys!

BurdetteLamar:


peterzhu2118:
Some of the documentation (e.g. Hash#merge!/Hash#update) refer the user to non-bang documentation, while others (e.g. this, Hash#select!/Hash#filter!, Hash#compact!) repeats it. Which do we want to standardize on?

BurdetteLamar:
> Some of the documentation (e.g. Hash#merge!/Hash#update) refer the user to non-bang documentation, while others (e.g. this, Hash#select!/Hash#filter!, Hash#compact!) repeats it. Which do we want to standardize on?

RDoc claims (on its own) that:
- Hash#update is primary, Hash#merge! is alias.
- Hash#select! is primary, Hash#filter! is alias.

So I've tried to cite the primary.

(Not sure whether this addresses your question.)

peterzhu2118:
No, what I mean is that for the bang version of the methods, some of them cite the non-bang version while others repeat the documentation.

For example, `Hash#update` says:

> Like Hash#merge, but modifies and returns self instead of a new hash

While `Hash#select!` is a repeat of `Hash#select` (but obviously saying that it modifies `self`).

This is inconsistent. How do we want to standardize. Referring to the documentation of the non-bang method reduces duplication and prevents them from going out-of-sync, but repeating the documentation prevents the user from needing to follow a link to read the documentation.

BurdetteLamar:
Got it.

BurdetteLamar:
@peterzhu2118, I'm thinking to document each method fully, instead of relying on a link.

BurdetteLamar:
Will update #update, but in a separate PR.  So, for here, stet?

hash.c
```diff
@@ -3325,13 +3325,98 @@ static int flatten_i(VALUE key, VALUE val, VALUE ary);
...
+ *
+ *      h = {foo: 0, bar: 1, baz: 2}
+ *      h.transform_keys!(baz: :BAZ, bar: :BAR, foo: :FOO)
+ *      # => {FOO: 0, BAR: 1, BAZ: 2}
+ *      h = {foo: 0, bar: 1, baz: 2}
+ *      h.transform_keys!(baz: :FOO, bar: :FOO, foo: :FOO)
+ *      # => {FOO: 2}
+ *
+ *  - If +old_key+ is not found, does nothing;
+ *    a duplicate key overwrites:
```
peterzhu2118, line 3374: If `old_key` is not found, it doesn't do anything, how can it overwrite?

PR #12944, opened 2025-03-16
[DOC] Doc for Hash#transform_values!

BurdetteLamar:


ioquatix:
Not sure if this breaks consistency, but I'd lead with a statement of fact about what this method does.

e.g

"Transform the values of a hash using a block."

BurdetteLamar:
> Not sure if this breaks consistency, but I'd lead with a statement of fact about what this method does.
> 
> e.g
> 
> "Transform the values of a hash using a block."

Thanks, @ioquatix.  I agree that this needs improvement, and have made a change in this PR.  I'll look at my other two active PRs to see whether they could also profit from this idea.  And I'll be on the lookout for other opportunities to apply it.

peterzhu2118 (approved):

PR #12945, opened 2025-03-17
Backport #7460 to Ruby 3.1

mdayaram:
This is a backport of PR #7460 to Ruby 3.1: [Correctly clean up `keeping_mutexes` before resuming any other
threads](https://github.com/ruby/ruby/pull/7460)

Fixes https://bugs.ruby-lang.org/issues/19480

We currently run Ruby 3.1.6 in our environment and have ran into this issue when using Fibers for request hedging.


hsbt:
Unfortunately, Ruby 3.1 is only security maintenance support. I don't backport this change.

PR #12950, opened 2025-03-19
Require 'delegate' for test_sum use of SimpleDelegate

headius:
test_sum in ruby/test_enum.rb uses SimpleDelegate, but the file does not require `delegate`.

I'm not sure how to handle test changes on maintenance branches, but we would like this change to go on the 3.4 branch for future JRuby 10 test updates.


k0kubun:
I'm fine with merging this diff to ruby_3_4 because missing a require is a bug in tests.

But it seems like you haven't made this change on master. In general, these stable branches are supposed to get only backports from master, which helps us avoid conflicts that could be made by changes coming from different branches.

Could you push the same change to master first? Then I'll merge this PR as a backport of a test fix.

headius:
@k0kubun Ok, that makes sense. I'll rebase on master and repush.

headius:
Is it expected that these test scripts are **only** runnable with `runner.rb`?

I think it would be best if the requires are local to the scripts where those libraries are used, rather than relying on a runner to load them.

nobu (approved):
Sorry to bother you.
I just described why these files work with `test/runner.rb`.

headius:
@nobu Oh thank you.

PR #12956, opened 2025-03-20
Trigger `inherited` and `const_set` callbacks after const has been defined

byroot:
[Misc #21143]
[Bug #21193]

The previous change caused a backward compatibility issue with code that called `Object.const_source_location` from the `inherited` callback.

To fix this, the order is now:

- Define the constant
- Invoke `inherited`
- Invoke `const_set`

This PR is trying to satisfy both Matz direction in [[Misc #21143]](https://bugs.ruby-lang.org/issues/21143) and also preserve backward compatibility. But if it's not working an alternative is to revert https://github.com/ruby/ruby/pull/12927

### DIFF
--- a/class.c
+++ b/class.c
@@ -1004,8 +1004,9 @@ rb_define_class(const char *name, VALUE super)
     }
     klass = rb_define_class_id(id, super);
     rb_vm_register_global_object(klass);
+    rb_const_set_raw(rb_cObject, id, klass);
     rb_class_inherited(super, klass);
-    rb_const_set(rb_cObject, id, klass);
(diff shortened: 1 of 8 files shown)


eregon (approved):

eileencodes (approved):
Can confirm this fixes the issue in our app too.

cc/ @matthewd since we were talking about the bug earlier

issyl0:
Also confirming that this fixes the issue GitHub was seeing with de48e47ddf78aba02fd9623bc7ce685540a10743. Thank you!

PR #12957, opened 2025-03-20
Remove leading `nop` from block when we don't need it

tenderlove:
Blocks insert a leading `nop` instruction in order to execute a "block call" tracepoint. Block compilation unconditionally inserts a leading `nop` plus a label after the instruction:

  https://github.com/ruby/ruby/blob/641f15b1c6bd8921407a1f045573d3b0605f00d3/prism_compile.c#L6867-L6869

This `nop` instruction is used entirely for firing the block entry tracepoint.  The label exists so that the block can contain a loop but the block entry tracepoint is executed only once.

For example, the following code is an infinite loop, but should only execute the b_call tracepoint once:

```ruby
-> { redo }.call
```

Previous to this commit, we would eliminate the `nop` instruction, but only if there were no other jump instructions inside the block.  This means that the following code would still contain a leading `nop` even though the label following the `nop` is unused:

```ruby
-> { nil if bar }
```

```
== disasm: #<ISeq:block in <main>@test.rb:1 (1,2)-(1,17)> (catch: FALSE)
0000 nop                                                              (   1)[Bc]
0001 putself                                [Li]
0002 opt_send_without_block                 <calldata!mid:bar, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0004 branchunless                           8
0006 putnil
0007 leave                                  [Br]
0008 putnil
0009 leave                                  [Br]
```

This commit checks to see if the label inserted after the `nop` is actually a jump target.  If it's not a jump target, then we should be safe to eliminate the leading `nop`:

```
> build-master/miniruby --dump=insns test.rb
== disasm: #<ISeq:<main>@test.rb:1 (1,0)-(1,17)>
0000 putspecialobject                       1                         (   1)[Li]
0002 send                                   <calldata!mid:lambda, argc:0, FCALL>, block in <main>
0005 leave

== disasm: #<ISeq:block in <main>@test.rb:1 (1,2)-(1,17)>
0000 putself                                                          (   1)[LiBc]
0001 opt_send_without_block                 <calldata!mid:bar, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0003 branchunless                           7
0005 putnil
0006 leave                                  [Br]
0007 putnil
0008 leave                                  [Br]
```

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -4367,9 +4367,18 @@ iseq_optimize(rb_iseq_t *iseq, LINK_ANCHOR *const anchor)
     list = FIRST_ELEMENT(anchor);
 
     int do_block_optimization = 0;
+    LABEL * block_loop_label = NULL;
 
-    if (ISEQ_BODY(iseq)->type == ISEQ_TYPE_BLOCK && !ISEQ_COMPILE_DATA(iseq)->catch_except_p) {
+    // If we're optimizing a block
+    if (ISEQ_BODY(iseq)->type == ISEQ_TYPE_BLOCK) {
         do_block_optimization = 1;
+
+        // If the block starts with a nop and a label,
+        // record the label so we can detect if it's a jump target
+        LINK_ELEMENT * le = FIRST_ELEMENT(anchor)->next;
+        if (IS_INSN(le) && IS_INSN_ID((INSN *)le, nop) && IS_LABEL(le->next)) {
+            block_loop_label = (LABEL *)le->next;
+        }
     }
 
     while (list) {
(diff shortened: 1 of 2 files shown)


compile.c
```diff
@@ -4386,9 +4395,27 @@ iseq_optimize(rb_iseq_t *iseq, LINK_ANCHOR *const anchor)
...
+                    // If the instruction has a jump target, check if the
+                    // jump target is the block loop label
+                    const char *types = insn_op_types(item->insn_id);
+                    for (int j = 0; types[j]; j++) {
+                        if (types[j] == TS_OFFSET) {
+                            // If the jump target is equal to the block loop
+                            // label, then we can't do the optimization because
+                            // the leading `nop` instruction fires the block
+                            // entry tracepoint
+                            LABEL * target = (LABEL *)OPERAND_AT(item, j);
```
XrXr, line 4412: Question since this is a bit confusing for me: TS_OFFSET during compilation isn't actually an offset yet, and is an instruction index?

XrXr (commented):
Could you add a b_call tracepoint test that use `redo` if there isn't one yet?

tenderlove:
> Could you add a b_call tracepoint test that use `redo` if there isn't one yet?

We have one [here](https://github.com/tenderlove/ruby/blob/aebf96f371c8d874398e0041b798892e545fa206/test/ruby/test_settracefunc.rb#L1728-L1736)

compile.c
tenderlove: Before assembly, it's a pointer to a label.  I'm not sure the best place to link to in order to explain, but if you check [the encoding routine](https://github.com/tenderlove/ruby/blob/aebf96f371c8d874398e0041b798892e545fa206/compile.c#L2651-L2657) when we encounter a TS_OFFSET, it pulls the label instance from the operand and encodes it.  I wanted this patch to be resilient in the case we have new instructions that take a jump target, so rather than checking for specific instructions like `jump`, I check the operands for `TS_OFFSET` and compare that to the block loop label.

XrXr (approved):

PR #12960, opened 2025-03-21
FCALL shouldn't be forwarded from caller

tenderlove:
When we forward an FCALL (a method call with an implicit self), we shouldn't forward the FCALL flag because it ignores method visibility checks.  This patch removes the FCALL flag from callers.

[Bug #21196]


tenderlove:
I don't think the build failures are related to my change.  I think [this PR fixes it](https://github.com/rubygems/rubygems/pull/8581), but since it causes the macOS build to fail I can't merge any PRs.

PR #12961, opened 2025-03-22
Fix handling of `error`/`errno` in `io_internal_wait`.

ioquatix:
With this test (provided by @mame), we can cause `rb_bug`:

```
$stdin.read_nonblock(1, exception: false)

trap(:HUP) { p :HUP }

Thread.new { sleep 1; Process.kill :HUP, $$ }

$stdin.timeout = 10

$stdin.gets
```

The reason is, `io_internal_wait` can assign `errno = 0` incorrectly. Instead, `errno` already has a value of 11 (EAGAIN) and we should not overwrite it.


ioquatix:
Had to fix https://github.com/ruby/ruby/pull/12963 first.

PR #12962, opened 2025-03-22
Increase timeout to fix flaky tests?

ioquatix:


ioquatix:
Applied in <https://github.com/ruby/ruby/pull/12963>.

PR #12975, opened 2025-03-24
Freeze $/ and make it ractor safe

etiennebarrie:
Implement [[Feature #21109]](https://bugs.ruby-lang.org/issues/21109) by freezing `$/`.

By always freezing when setting the global rb_rs variable, we can ensure it is not modified and can be accessed from a ractor.

We're also making sure it's an instance of String and does not have any instance variables.

Of course, if $/ is changed at runtime, it may cause surprising behavior but doing so is deprecated already anyway.

### DIFF
--- a/io.c
+++ b/io.c
@@ -8673,6 +8673,25 @@ deprecated_str_setter(VALUE val, ID id, VALUE *var)
     *var = val;
 }
 
+static void
+deprecated_rs_setter(VALUE val, ID id, VALUE *var)
+{
+    if (!NIL_P(val)) {
+        if (!RB_TYPE_P(val, T_STRING)) {
+            rb_raise(rb_eTypeError, "value of %"PRIsVALUE" must be String", rb_id2str(id));
+        }
+        if (rb_str_equal(val, rb_default_rs)) {
+            val = rb_default_rs;
+        }
+        else {
+            val = rb_str_frozen_bare_string(val);
+        }
+        rb_enc_str_coderange(val);
+        rb_warn_deprecated("'%s'", NULL, rb_id2name(id));
+    }
+    *var = val;
+}
+
 /*
  *  call-seq:
  *    print(*objects) -> nil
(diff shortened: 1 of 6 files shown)


spec/ruby/language/predefined_spec.rb
```diff
@@ -559,10 +559,20 @@ def foo
...
+      $/ = str
+      $/.should equal(str)
+    end
+  end
+
+  ruby_version_is "3.5" do
+    it "freezes the assigned String" do
+      str = +"abc"
+      $/ = str
+      $/.frozen?.should == true
```
nobu, line 574: I think `Expected "abc".frozen?` is better than `Expected false == true`.
And whether the string is assigned should be tested.

```suggestion
      $/.should.frozen?
      $/.should == str
```

io.c
```diff
@@ -15713,7 +15732,8 @@ Init_IO(void)
     rb_vm_register_global_object(rb_default_rs);
     rb_rs = rb_default_rs;
     rb_output_rs = Qnil;
-    rb_define_hooked_variable("$/", &rb_rs, 0, deprecated_str_setter);
+    rb_define_hooked_variable("$/", &rb_rs, 0, deprecated_rs_setter);
+    rb_gvar_ractor_local("$/"); // not local but ractor safe
     rb_define_hooked_variable("$-0", &rb_rs, 0, deprecated_str_setter);
```
nobu, line 15737: Also `$-0`?

spec/ruby/language/predefined_spec.rb
etiennebarrie: Thank you!

etiennebarrie:
We realized it was possible to smuggle non-ractor safe objects by setting them as instance variables on the String.

We want to set `$/` to a frozen "bare" string. But we didn't have a function available for that so I added `rb_str_frozen_bare_string`.

string.c
```diff
@@ -1480,6 +1480,14 @@ rb_str_new_frozen_String(VALUE orig)
     return str_new_frozen(rb_cString, orig);
 }
 
+
+VALUE
+rb_str_frozen_bare_string(VALUE orig)
+{
+    if (RB_LIKELY(BARE_STRING_P(orig))) return orig;
```
casperisfine, line 1487: Hum, that doesn't do what's marked on the tin. Here you are returning a string that is potentially mutable.

etiennebarrie: Oh yeah thanks for the catch, updated that function and added two tests, one checking that an non-frozen string turns into a (new) frozen one, and one checking that a frozen String with no instance variables just gets set.

PR #7182, opened 2023-01-25
Ractor.make_shareable(proc_obj) makes inner structure shareable

luke-gru:
Proc objects are now traversed like other objects when making them shareable.

Fixes [Bug #19372]

https://bugs.ruby-lang.org/issues/19372

PR #12977, opened 2025-03-24
(rebase) Ractor.make_shareable(proc_obj) makes inner structure shareable

jhawthorn:
This is a rebase of https://github.com/ruby/ruby/pull/7182 with one tiny fix (not setting RUBY_FL_SHAREABLE) and I've added some additional tests

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -1415,7 +1415,7 @@ rb_proc_ractor_make_shareable(VALUE self)
         proc->is_isolated = TRUE;
     }
 
-    FL_SET_RAW(self, RUBY_FL_SHAREABLE);
+    rb_obj_freeze(self);
     return self;
 }
 
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -1361,6 +1361,28 @@ assert_equal 'true', %q{
   Ractor.shareable?(pr)
 }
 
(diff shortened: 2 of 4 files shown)


vm.c
```diff
@@ -1415,7 +1415,7 @@ rb_proc_ractor_make_shareable(VALUE self)
         proc->is_isolated = TRUE;
     }
 
-    FL_SET_RAW(self, RUBY_FL_SHAREABLE);
+    rb_obj_freeze(self);
```
ko1, line 1418: does it marks shareable flag on a Proc?

jhawthorn: It does after traversing through the objects as part of `rb_ractor_make_shareable`. Setting it here before was allowing unfrozen objects to leak through.

ko1: Ah, ok, `mark_shareable` will set.
Strictly speaking the Proc is not needed to be frozen, but okay.
no problem to merge.

jhawthorn: I think we need to freeze to prevent instance variables from being set after it's made shareable.

ko1: Yes it is safer. But for example Ractor objects are shareable but not frozen. The instance variables can be accessed only from main ractor.

jhawthorn: Ah! I see, thank you. I didn't see that behaviour in `generic_ivtbl`

PR #12982, opened 2025-03-25
Ractor "after fork" fixes

tenderlove:
This PR fixes a couple bugs in Ractors after fork.  Ractors need to have their object cache cleared and also raise exceptions on `take` / `send`.

### DIFF
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -2210,3 +2210,10 @@ assert_equal 'ok', %q{
   end
   'ok'
 }
+
+# fork after creating Ractor
+assert_equal 'ok', %q{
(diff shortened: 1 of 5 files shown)


jhawthorn:
I think I've fixed the deadlock the new tests were creating in CI (they reproduced locally fairly consistently under llvm18, not sure why that compiler specifically). We also needed to acquire RB_VM_LOCK around the call to fork and now this seems to pass CI.

I'll rebase and merge tomorrow if there are no objections 🙇‍♂️

cc @ko1 @byroot @etiennebarrie @luke-gru @peterzhu2118

ko1:
I give up to understand completely so let's try!

PR #12984, opened 2025-03-26
Update to ruby/spec@5e579e2

andrykonchin:

cc @eregon


eregon:
Can you remove the [Rubyspec Version Guards Check / Rubyspec (ruby-3.0) (pull_request)](https://github.com/ruby/ruby/actions/runs/14090544107/job/39465861302?pr=12984) job since ruby/spec requires 3.1?

PR #12985, opened 2025-03-26
[DOC] Tweaks for Hash#update

BurdetteLamar:


hash.c
```diff
@@ -4160,13 +4160,37 @@ rb_hash_update_block_i(VALUE key, VALUE value, VALUE hash)
...
+ *
+ *  With no block given, for each successive entry +key+/+new_value+ in each successive +other_hash+:
+ *
+ *  - If +key+ is in +self+, sets <tt>self[key] = new_value</tt>, whose position is unchanged:
+ *
+ *      h.update({bat: 3, foo: -1}, {bat: 4, bar: -1}) # => {foo: -1, bar: -1, baz: 2, bat: 4}
+ *
+ *  - If +key+ is not in +self+, adds the entry at the end of +self+:
+ *
+ *      h = {foo: 0, bar: 1, baz: 2} # => {foo: 0, bar: 1, baz: 2}
```
peterzhu2118, line 4175: These examples are not right. `h` is not defined in the first example and this example doesn't demonstrate anything.

hash.c
```diff
@@ -4160,13 +4160,42 @@ rb_hash_update_block_i(VALUE key, VALUE value, VALUE hash)
...
+ *
+ *  With no block given, for each successive entry +key+/+new_value+ in each successive +other_hash+:
+ *
+ *  - If +key+ is in +self+, sets <tt>self[key] = new_value</tt>, whose position is unchanged:
+ *
+ *      h0 = {foo: 0, bar: 1, baz: 2}
+ *      h1 = {bat: 3, foo: -1}
+ *      h2 = {bat: 4, bar: -1}
+ *      h0.update(h1, h2) # => {foo: -1, bar: -1, baz: 2, bat: 4}
+ *      h0.update(h2, h1) # => {foo: -1, bar: -1, baz: 2, bat: 3}
```
peterzhu2118, line 4175: These examples are too complicated IMO. It's also showing what happens when a new entry is added. That's covered by the next section. I think we can simplify it to something like:

```suggestion
 *      h0 = {foo: 0, bar: 1, baz: 2}
 *      h1 = {bar: 3, foo: -1}
 *      h0.update(h1) # => {foo: -1, bar: 3, baz: 2}
```

peterzhu2118 (approved):

PR #12986, opened 2025-03-26
[DOC] Tweaks for Hash#has_value?

BurdetteLamar:
@peterzhu2118, there is something wrong here.  Even in https://docs.ruby-lang.org/en/master/Hash.html#method-i-has_value-3F, the `Source` button does not work (does nothing except re-orient the arrow).

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -3890,9 +3890,10 @@ rb_hash_search_value(VALUE key, VALUE value, VALUE arg)
 /*
  *  call-seq:
  *    has_value?(value) -> true or false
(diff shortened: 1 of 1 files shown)


peterzhu2118 (approved):

peterzhu2118:
Can you open an issue in https://github.com/ruby/rdoc for the issue?

BurdetteLamar:
> Can you open an issue in https://github.com/ruby/rdoc for the issue?

Done:  https://github.com/ruby/rdoc/issues/1335

PR #12992, opened 2025-03-27
Pend a test for ractor in etc

mame:
... because it is flaky


eregon:
I copied the commit back upstream in https://github.com/ruby/etc/commit/894a7e37fc699f0875da8e4ad3b9be76e6613603
and mentioned it in the PR adding the test.

PR #12995, opened 2025-03-27
Fix memory leaks in WASI setjmp handler

white-axe:
This pull request fixes two memory leaks in the implementation of setjmp in WASI (WebAssembly) builds of Ruby:

- The first commit fixes a problem where the stack pointer is not reset after a longjmp, leading to leaking of parts of the stack. This also causes leaking of Ruby objects since the garbage collector looks for GC roots on the stack.
- The second commit fixes a problem where the VM jump buffers are not freed after a longjmp. In WASI builds of Ruby, jump buffers are allocated on the heap instead of on the stack, so they need to be manually deallocated. I changed the implementations of `rb_vm_tag_jmpbuf_init()` and `rb_vm_tag_jmpbuf_deinit()` to clean up leaked jump buffers before initializing/deinitializing jump buffers.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -1369,7 +1369,6 @@ cont_init(rb_context_t *cont, rb_thread_t *th)
     /* save thread context */
     cont_save_thread(cont, th);
     cont->saved_ec.thread_ptr = th;
-    cont->saved_ec.tag = NULL;
     cont->saved_ec.local_storage = NULL;
     cont->saved_ec.local_storage_recursive_hash = Qnil;
     cont->saved_ec.local_storage_recursive_hash_for_trace = Qnil;
--- a/cont.c
+++ b/cont.c
@@ -1369,6 +1369,7 @@ cont_init(rb_context_t *cont, rb_thread_t *th)
     /* save thread context */
     cont_save_thread(cont, th);
     cont->saved_ec.thread_ptr = th;
+    cont->saved_ec.tag = NULL;
     cont->saved_ec.local_storage = NULL;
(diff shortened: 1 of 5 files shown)


white-axe:
@kateinoigakukun, since you maintain the WASI builds of Ruby, can you take a look at this?

kateinoigakukun:
Let me execute ruby.wasm test suite with this change: https://github.com/ruby/ruby.wasm/pull/586

white-axe:
Here's a test to see if a WASI build is affected by these two memory leaks. This script causes all of the current WASI builds from https://github.com/ruby/ruby.wasm to rapidly consume around 800 mebibytes of memory due to the jump buffer leak and then crash from what I believe is stack buffer overflow caused by the stack pointer leak:
```ruby
loop do
  array = [0]
  for item in array
    break
  end
end
```
To test it out, save it as test.rb in the current working directory, and then:
```
curl -Lo ruby.tar.gz https://github.com/ruby/ruby.wasm/releases/download/2.7.1/ruby-3.4-wasm32-unknown-wasip1-full.tar.gz
tar xzf ruby.tar.gz --strip-components=1
wasmtime --dir .::/ usr/local/bin/ruby test.rb
```

kateinoigakukun (approved):
Really good catch, the patch looks good to me.

kateinoigakukun:
Hi @white-axe, the change seems causing OOM on ruby.wasm test suite, so I'm reverting this for now https://github.com/ruby/ruby.wasm/pull/590

white-axe:
I'll find some other way to fix the second memory leak then

PR #12997, opened 2025-03-27
Ractor: Fix moving embedded objects

casperisfine:
`rb_obj_alloc(RBASIC_CLASS(obj))` will always allocate from the basic 40B pool, so if `obj` is larger than `40B`, we'll create a corrupted object when we later copy the shape_id.

I chose to use specialized `rb_*_dup` functions when possible to benefit from shared String/Arrays and trigger write barriers.

For some other types I resorted to the generic `rb_obj_dup`. It works but probably isn't ideal.

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -3585,8 +3585,27 @@ move_enter(VALUE obj, struct obj_traverse_replace_data *data)
         return traverse_skip;
     }
     else {
-        VALUE moved = rb_obj_alloc(RBASIC_CLASS(obj));
-        rb_shape_set_shape(moved, rb_shape_get_shape(obj));
+        VALUE moved;
+        switch (RB_BUILTIN_TYPE(obj)) {
+          case T_STRING:
+            moved = rb_str_dup(obj);
+            break;
+          case T_ARRAY:
+            moved = rb_ary_dup(obj);
+            break;
+          case T_HASH:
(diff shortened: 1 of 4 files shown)


ractor.c
```diff
@@ -3585,8 +3585,27 @@ move_enter(VALUE obj, struct obj_traverse_replace_data *data)
         return traverse_skip;
     }
     else {
-        VALUE moved = rb_obj_alloc(RBASIC_CLASS(obj));
```
casperisfine, line 3588: Also this was preserving the `metaclass` so perhaps using `dup` is incorrect.

casperisfine: That said, the previous behavior wasn't particularly useful:

```ruby
>> o = Object.new
=> #<Object:0x000000012b396a30>
>> def o.foo = 12
=> :foo
>> Ractor.new { Ractor.receive }.send(o, move: true).take
(irb):3: warning: Ractor is experimental, and the behavior may change in future versions of Ruby! Also there are many implementation issues.
<internal:ractor>:600:in 'Ractor#send': can't create instance of singleton class (TypeError)
```

casperisfine:
@jhawthorn pointed that `move_leave` also assume no WVA, so it needs to be updated too.

casperisfine:
Closing in favor of: https://github.com/ruby/ruby/pull/13008

PR #13008, opened 2025-03-28
Ractor: Fix moving embedded objects

casperisfine:
[[Bug #20271]](https://bugs.ruby-lang.org/issues/20271)
[[Bug #20267]](https://bugs.ruby-lang.org/issues/20267)
[[Bug #20255]](https://bugs.ruby-lang.org/issues/20255)

`rb_obj_alloc(RBASIC_CLASS(obj))` will always allocate from the basic 40B pool, so if `obj` is larger than `40B`, we'll create a corrupted object when we later copy the shape_id.

Instead we can use the same logic than ractor copy, which is to use `rb_obj_clone`, and later ask the GC to free the original object.

We then must turn it into a `T_OBJECT`, because otherwise just changing its class to `RactorMoved` leaves a lot of ways to keep using the object, e.g.:

```
a = [1, 2, 3]
Ractor.new{}.send(a, move: true)
[].concat(a) # Should raise, but wasn't.
```

If it turns out that `rb_obj_clone` isn't performant enough for some uses, we can always have carefully crafted specialized paths for the types that would benefit from it.

### DIFF
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -2101,13 +2101,3 @@ assert_equal 'ok', %q{
     :fail
   end
 }
-
-# moved objects keep their object_id
-assert_equal 'ok', %q{
-  ractor = Ractor.new { Ractor.receive }
-  obj = Object.new
-  id = obj.object_id
-  ractor.send(obj, move: true)
-  roundtripped_obj = ractor.take
-  roundtripped_obj.object_id == id ? :ok : :fail
-}
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -1987,3 +1987,127 @@ assert_equal 'ok', %q{
   GC.start
   :ok.itself
 }
+
+# moved objects being corrupted if embeded (String)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
+  obj = "foobarbazfoobarbazfoobarbazfoobarbaz"
+  ractor.send(obj.dup, move: true)
+  roundtripped_obj = ractor.take
+  roundtripped_obj == obj ? :ok : roundtripped_obj
+}
+
+# moved objects being corrupted if embeded (Array)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
+  obj = Array.new(10, 42)
+  ractor.send(obj.dup, move: true)
+  roundtripped_obj = ractor.take
+  roundtripped_obj == obj ? :ok : roundtripped_obj
+}
+
+# moved objects being corrupted if embeded (Hash)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
+  obj = { foo: 1, bar: 2 }
+  ractor.send(obj.dup, move: true)
+  roundtripped_obj = ractor.take
+  roundtripped_obj == obj ? :ok : roundtripped_obj
+}
+
+# moved objects being corrupted if embeded (MatchData)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
+  obj = "foo".match(/o/)
+  ractor.send(obj.dup, move: true)
+  roundtripped_obj = ractor.take
+  roundtripped_obj == obj ? :ok : roundtripped_obj
+}
+
+# moved objects being corrupted if embeded (Struct)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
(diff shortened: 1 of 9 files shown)


bootstraptest/test_ractor.rb
```diff
@@ -1987,3 +1987,127 @@ def require feature
   GC.start
   :ok.itself
 }
+
+# moved objects being corrupted if embeded (String)
+assert_equal 'ok', %q{
+  ractor = Ractor.new { Ractor.receive }
+  obj = "foobarbazfoobarbazfoobarbazfoobarbaz"
+  ractor.send(obj.dup, move: true)
+  rountriped_obj = ractor.take
```
peterzhu2118, line 1996: I think this is a typo? It should be `roundtripped_obj`. All of the tests have this typo.

```suggestion
  roundtripped_obj = ractor.take
```

gc.c
```diff
@@ -2659,6 +2661,23 @@ rb_gc_mark_roots(void *objspace, const char **categoryp)
 
 #define TYPED_DATA_REFS_OFFSET_LIST(d) (size_t *)(uintptr_t)RTYPEDDATA(d)->type->function.dmark
 
+void
+rb_gc_ractor_moved(VALUE dest, VALUE src)
+{
+    void *objspace = rb_gc_get_objspace();
+    st_data_t id = 0;
+    if (UNLIKELY(FL_TEST_RAW(src, FL_SEEN_OBJ_ID))) {
+        id = rb_gc_impl_object_id_move(objspace, dest, src);
```
peterzhu2118, line 2670: I think `rb_gc_impl_object_id_move` should set the `FL_SEEN_OBJ_ID` flag, so we don't need to do it below.

gc/default/default.c
```diff
@@ -1614,6 +1595,45 @@ rb_gc_impl_object_id(void *objspace_ptr, VALUE obj)
...
+
+   unsigned int lev = rb_gc_vm_lock();
+   st_data_t key = (st_data_t)src;
+   if (!st_delete(objspace->obj_to_id_tbl, &key, &id)) {
+       rb_bug("gc_move: object ID seen, but not in mapping table: %s", rb_obj_info(src));
+   }
+   st_insert(objspace->obj_to_id_tbl, (st_data_t)dest, id);
+   st_insert(objspace->id_to_obj_tbl, id, (st_data_t)dest);
+   FL_UNSET_RAW(src, FL_SEEN_OBJ_ID);
+   rb_gc_vm_unlock(lev);
```
peterzhu2118, line 1633: This is going to break `ObjectSpace._id2ref`. Is there a reason why we want the moved object to have the same object ID? It's a new object, so IMO it should have a different object ID.

byroot: > This is going to break `ObjectSpace._id2ref`.

How so?

> Is there a reason why we want the moved object to have the same object ID? It's a new object,

Semantically, it's not a new object. It's the same object but moved to another ractor.

I'm fine with not transferring the `object_id`, the previous code was doing it AFAICT, but it feels like the right thing to do.

peterzhu2118: Ah I misread the code. I didn't see that you removed the original object from the `obj_to_id_tbl` table. This makes sense now 👍

gc.c
byroot: The problem with that is that we're resetting the flags bellow.

byroot: nvm, I'm an idiot.

gc.c
```diff
@@ -2659,6 +2661,19 @@ rb_gc_mark_roots(void *objspace, const char **categoryp)
 
 #define TYPED_DATA_REFS_OFFSET_LIST(d) (size_t *)(uintptr_t)RTYPEDDATA(d)->type->function.dmark
 
+void
+rb_gc_ractor_moved(VALUE dest, VALUE src)
+{
+    void *objspace = rb_gc_get_objspace();
+    if (UNLIKELY(FL_TEST_RAW(src, FL_SEEN_OBJ_ID))) {
+        rb_gc_impl_object_id_move(objspace, dest, src);
+    }
```
casperisfine, line 2670: I suppose the big argument for not transferring the `object_id` is that we'd save on locking the VM.

peterzhu2118 (approved):
This looks good to me!

gc.c
ko1: I don't think ID should be inherited (moved object should be different object).

casperisfine: I did it because to me it seemed that semantically, `move: true` means we're just changing which ractor owns the object, and not allocating a new object.

But if you think it shouldn't be preserved, it's fine by me, I don't think it's very importan.

Removed.

PR #13009, opened 2025-03-28
Add Array#add to push and return item

aviflombaum:
Adds an item to an array and returns the added item and not the array itself.

```ruby
a = [1, 2, 3]
a.add(4) #=> 4
a #=> [1, 2, 3, 4]
```

Useful to avoid patterns like:

```ruby
# ❌
def add_message(message)
  messages << message
  message
end

# ✅
def add_message(message)
  messages.add(message)
end
```


array.rb
```diff
@@ -1,4 +1,19 @@
...
+  # Adds an item to +self+ and returns the item itself.
+  # Returns the added item:
+  #
+  #   a =          [1, 2, 3]
+  #   a.add(4) #=> 4
+  #   a        #=> [1, 2, 3, 4]
+  #
+  # Related: see {Methods for Assigning}[rdoc-ref:Array@Methods+for+Assigning].
+  def add item
+    item.tap { push(it) }
```
nobu, line 14: Why using unnecessary `tap`?

```suggestion
    self << item
    item
```

or, to avoid side effects by redefinition of other Array methods:

```suggestion
    Primitive.cexpr!(%q{rb_ary_push(self, item)})
    item
```

nobu (approved):
To add a new feature, we need discussion at https://bugs.ruby-lang.org.

nobu (changes requested):
Mistake.

PR #13013, opened 2025-03-30
ractor: don't inherit the default thread group

byroot:
[[Bug #17506]](https://bugs.ruby-lang.org/issues/17506)

`Thread.current.group` isn't shareable so it shouldn't be inherited by the main thread of a new Ractor.

This cause an extra allocation when spawning a ractor, which could be elided with a bit of extra work, but not sure if it's worth the effort.

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -106,6 +106,7 @@
 #endif
 
 static VALUE rb_cThreadShield;
+static VALUE cThGroup;
 
 static VALUE sym_immediate;
 static VALUE sym_on_blocking;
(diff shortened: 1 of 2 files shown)


thread.c
```diff
@@ -993,13 +1001,20 @@ rb_thread_create(VALUE (*fn)(void *), void *arg)
 VALUE
 rb_thread_create_ractor(rb_ractor_t *r, VALUE args, VALUE proc)
 {
+    VALUE thgroup = r->thgroup_default = rb_obj_alloc(cThGroup);
```
ko1, line 1004: init at `thread_do_start_proc()` seems better.

byroot: Ok, I'll change it.

byroot: https://github.com/ruby/ruby/pull/13016

PR #13023, opened 2025-03-31
Ractor: don't invoke `initialize_*` callbacks when moving an object

casperisfine:
[[Bug #20255]](https://bugs.ruby-lang.org/issues/20255)

Followup: https://github.com/ruby/ruby/pull/13008

Alternatively we could invoke it, but would need to do it in the context of the newly spawned Ractor so that accessing global non-shareable objects is not possible.

@ko1 @jhawthorn @tenderlove any opinion?


casperisfine:
This isn't possible because `Array` and other classes do rely on `initialize_clone` existing:

```ruby
class Array
  def initialize_clone(o)
    # noop
  end
end

p [1, 2, 3, 4, 5].clone # => []
```

PR #13024, opened 2025-04-01
[DOC] Doc for Thread::ConditionVariable

wks:
Documented the necessity of calling `wait` in a loop.

Mentioned spurious wake-up in the doc.

The example in the documentation of `Thread::ConditionVariable` is wrong.  Condition variables should use a separate boolean predicate, and should be waited in a loop due to nondeterministic scheduling and spurious wake-ups.  This pull request explains these issues and provides the right example.  FYI, the documentation of other programming languages or libraries (such as [POSIX thread](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_wait.html), [C++](https://en.cppreference.com/w/cpp/thread/condition_variable/wait), [Java](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java/util/concurrent/locks/Condition.html), [Rust](https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait) and [Python](https://docs.python.org/3.13/library/threading.html#condition-objects)) also describe such issues and provide relevant examples.

The current documentation does not mention whether *spurious wake-ups* are allowed, but from the function name `rb_thread_sleep_deadly_allow_spurious_wakeup` and the invocation of `sleep_hrtime(..., 0)`, it is apparent that spurious wake-ups are possible.  This pull request documented this, too.  The presence of spurious wake-ups does not affect the use of condition variables because they always need to be waited in a loop anyway.


thread_sync.c
```diff
@@ -1430,26 +1430,53 @@ struct rb_condvar {
  *
  *  ConditionVariable objects augment class Mutex. Using condition variables,
  *  it is possible to suspend while in the middle of a critical section until a
- *  resource becomes available.
+ *  condition is met, such as a resource becomes available.
+ *
+ *  Due to non-deterministic scheduling and spurious wake-ups, users of
```
peterzhu2118, line 1435: Does Ruby's condition variable have spurious wake-ups?

thread_sync.c
```diff
@@ -1430,26 +1430,53 @@ struct rb_condvar {
...
- *	 }
+ *      mutex.synchronize {
+ *        # Thread 'b' makes the resource available.
+ *        resource_available = true
+ *
+ *        # In case any thread is waiting on the condition variable, we notify
+ *        # one thread about the availability of the resource.
+ *        condvar.signal
+ *      }
  *    }
```
peterzhu2118, line 1480: I've noticed that this example is incorrect because we never call `a.join` or `b.join`, so the main thread terminates before the other two threads are able to do anything.

thread_sync.c
```diff
@@ -1430,26 +1430,53 @@ struct rb_condvar {
...
+ *        #     fast that it have already made the resource available before
+ *        #     thread 'a' start.  Therefore, we should check
+ *        #     'resource_available' before starting to wait.
+ *        # (2) The 'wait' method may spuriously wake up without signalling.
+ *        #     Therefore, we should recheck 'resource_available' after waking
+ *        #     up, and go back to wait if the condition is not actually met.
+ *        # (3) As soon as the resource is available, a third thread (not
+ *        #     necessarily waiting on 'condvar') may be scheduled first and
+ *        #     may have acquired the 'mutex' and consumed the resource
+ *        #     before thread 'a' wakes up.  This implies rechecking, too.
```
peterzhu2118, line 1464: I don't think the example demonstrates this. Is this note necessary here?

thread_sync.c
wks: We reach `mutex_sleep_begin` through the call chain `rb_condvar_wait` -> `do_sleep` -> `rb_mutex_sleep` -> `mutex_sleep_begin`.

```c
static VALUE
mutex_sleep_begin(VALUE _arguments)
{
    struct rb_mutex_sleep_arguments *arguments = (struct rb_mutex_sleep_arguments *)_arguments;
    VALUE timeout = arguments->timeout;
    VALUE woken = Qtrue;

    VALUE scheduler = rb_fiber_scheduler_current();
    if (scheduler != Qnil) {
        rb_fiber_scheduler_kernel_sleep(scheduler, timeout);
    }
    else {
        if (NIL_P(timeout)) {
            rb_thread_sleep_deadly_allow_spurious_wakeup(arguments->self, Qnil, 0);
        }
        else {
            struct timeval timeout_value = rb_time_interval(timeout);
            rb_hrtime_t relative_timeout = rb_timeval2hrtime(&timeout_value);
            /* permit spurious check */
            woken = RBOOL(sleep_hrtime(GET_THREAD(), relative_timeout, 0));
        }
    }

    return woken;
}
```

If the fiber scheduler is present, the user will implement it.  The documentation of `Fiber::Scheduler#kernel_sleep` does not specify whether spurious wake up is allowed.

The function name `rb_thread_sleep_deadly_allow_spurious_wakeup` clearly indicates it allows spurious wakeup.

The function `sleep_hrtime(th, rel, fl)` is defined like this:

```c
static int
sleep_hrtime(rb_thread_t *th, rb_hrtime_t rel, unsigned int fl)
{
    enum rb_thread_status prev_status = th->status;
    int woke;
    rb_hrtime_t end = rb_hrtime_add(rb_hrtime_now(), rel);

    th->status = THREAD_STOPPED;
    RUBY_VM_CHECK_INTS_BLOCKING(th->ec);
    while (th->status == THREAD_STOPPED) {
        native_sleep(th, &rel);
        woke = vm_check_ints_blocking(th->ec);
        if (woke && !(fl & SLEEP_SPURIOUS_CHECK))
            break;
        if (hrtime_update_expire(&rel, end))
            break;
        woke = 1;
    }
    th->status = prev_status;
    return woke;
}
```

When `fl` is 0 (the argument passed from `mutex_sleep_begin`) and `woke` is true (this thread is "spuriously interrupted"), the condition `woke && !(fl & SLEEP_SPURIOUS_CHECK)` will be true.  It will break and return.

So the conclusion is spurious wake-ups are possible.

thread_sync.c
wks: Well, the example is still correct if it is part of a larger program where the main thread runs almost forever.  But I can still add `a.join` and `b.join` to the end so that the user can copy and paste this example to a script and run it.

thread_sync.c
wks: It is an important issue to consider in real-world concurrent programming.  I can add another thread `a2` which behaves exactly like `a` and that will make the note (3) necessary.

wks:
@peterzhu2118  I have re-written part of the example and added a third thread `a2`.  I also moved the "Notes:" before the code example to keep the code terse.

peterzhu2118 (approved):

PR #13030, opened 2025-04-01
[DOC] Tweaks to String::try_convert

BurdetteLamar:


string.c
```diff
@@ -2874,14 +2874,16 @@ rb_check_string_type(VALUE str)
  *  call-seq:
  *    String.try_convert(object) -> object, new_string, or nil
  *
- *  If +object+ is a +String+ object, returns +object+.
+ *  Attempts to convert the given +object+ to a string.
+ *
+ *  If +object+ is already a string, returns +object+, unmodified:
```
peterzhu2118, line 2879: Should it be a period at the end?

```suggestion
 *  If +object+ is already a string, returns +object+, unmodified.
```

peterzhu2118 (approved):

PR #13033, opened 2025-04-01
io_buffer: Reimplement dcompact for IO::Buffer

hanazuki:
Proposed patch for [[Bug #21210]](https://bugs.ruby-lang.org/issues/21210)


byroot:
FYI @eightbitraptor I'll merge this given it's a bug.

PR #13041, opened 2025-04-02
Improve backtrace of errors raised by `Socket.tcp_with_fast_fallback`

casperisfine:
[[Bug #21211]](https://bugs.ruby-lang.org/issues/21211)

Socket errors raised from background threads are hard to track down because their backtrace starts from the spawned thread.

To solve this we can raise a new error with the old one as `cause`.


shioimm (approved):
Thank you for solving this!

PR #13043, opened 2025-04-02
Add a test for failing assertions in vm_args.c

casperisfine:
Git bisect points at https://github.com/ruby/ruby/pull/12596

@jeremyevans @peterzhu2118 I have very limited understanding of that code, so I can't tell if the assertions are incorrect or if something bad is happening.

I however added a test case that make the assertions fail.

### DIFF
--- a/bootstraptest/test_method.rb
+++ b/bootstraptest/test_method.rb
@@ -1419,3 +1419,11 @@ assert_equal 'ok', %q{
     "ok"
   end
 }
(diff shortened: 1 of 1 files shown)


jeremyevans:
Thanks for the report.  I'll look into this today, and if I can't fix it quickly, I'll revert the change.

casperisfine:
No rush!

PR #13045, opened 2025-04-02
Fix moving composite object parts

luke-gru:
Some bugs were introduced in commit 0350290262ea0fbc4e1807901797ee8a6970c2b9 that had to do with moving composite object parts. This fixes those bugs and adds tests for those cases.


byroot:
I merged the test at 72dc16aa65a9d46f032d7a8d0292f5a5df6aade5

luke-gru:
Yeah I saw your recent commit and figured the test could still be useful. Thanks @byroot!

PR #13046, opened 2025-04-02
Fix assertion failure with anonymous splats

jeremyevans:
When calling a method that accepts an anonymous splat and literal keywords without any arguments, an assertion failure was previously raised. Set rest_index to 0 when setting rest to the frozen array, so the args_argc calculation is accurate.

While here, add more tests for methods with anonymous splats with and without keywords and keyword splats to confirm behavior is correct.

Also add a basic bootstrap test that would hit the previous assertion failure.

Fixes #13043

### DIFF
--- a/bootstraptest/test_method.rb
+++ b/bootstraptest/test_method.rb
@@ -1419,3 +1419,11 @@ assert_equal 'ok', %q{
     "ok"
   end
 }
+
+assert_equal 'ok', <<~RUBY
+  def test(*, kw: false)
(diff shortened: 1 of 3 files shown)


byroot (approved):
❤️

jeremyevans:
Same Windows failures as in the master branch, so this seems safe to merge.

casperisfine:
Just to confirm, the assertion failure is gone and builds are green again.

PR #13048, opened 2025-04-02
Fix lldb debug scripts

eileencodes:
In ruby/ruby#13008 `RVALUE` was removed without replacement. This means the lldb scripts that relied on `RVALUE` stopped working.

I updated the ones that were using it just for the bytesize to use `slot_size` and then round to the nearest power of 40. We can't use `slot_size` directly because in debug mode it's `48` but `RVALUE` is `40` bytes.

For the `as_type` method, I updated it to check the type. It's only used for `bignum` and `array` so that's a simple change.

Lastly, for the `dump_page` method I replaced it with `struct free_slot` since that's looking at the freelist.

`struct RVALUE` has been removed from all the scripts and I verified that `rp` is fixed. I'm not confident the `dump_page` method is fixed, the freelist looks off, but for now this gets us closer.

### DIFF
--- a/misc/lldb_cruby.py
+++ b/misc/lldb_cruby.py
@@ -14,6 +14,7 @@ import sys
 import shlex
 import platform
 import glob
+import math
 
 from lldb_rb.constants import *
 
@@ -270,7 +271,6 @@ def lldb_inspect(debugger, target, result, val):
         print('immediate(%x)' % num, file=result)
     else:
         tRBasic = target.FindFirstType("struct RBasic").GetPointerType()
-        tRValue = target.FindFirstType("struct RVALUE")
 
         val = val.Cast(tRBasic)
         flags = val.GetValueForExpressionPath("->flags").GetValueAsUnsigned()
@@ -524,10 +524,11 @@ def rb_backtrace(debugger, command, result, internal_dict):
     bt.print_bt(val)
 
 def dump_bits(target, result, page, object_address, end = "\n"):
-    tRValue = target.FindFirstType("struct RVALUE")
+    slot_size = page.GetChildMemberWithName("heap").GetChildMemberWithName("slot_size").unsigned
+    byte_size = 40 ** math.floor(math.log(slot_size, 40))
     tUintPtr = target.FindFirstType("uintptr_t") # bits_t
 
-    num_in_page = (object_address & HEAP_PAGE_ALIGN_MASK) // tRValue.GetByteSize();
+    num_in_page = (object_address & HEAP_PAGE_ALIGN_MASK) // byte_size;
     bits_bitlength = tUintPtr.GetByteSize() * 8
     bitmap_index = num_in_page // bits_bitlength
     bitmap_offset = num_in_page & (bits_bitlength - 1)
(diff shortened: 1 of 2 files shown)


eileencodes:
That's weird I just tested it and `rp str` its not broken for me. 🤔

How is your Ruby configured? This is how I'm configuring

```
./configure --prefix=$HOME/.rubies/ruby-debug --disable-install-rdoc --with-openssl-dir=$(brew --prefix)/opt/openssl@3 optflags='-O0    ' debugflags='-g3' cppflags='-DRUBY_DEBUG'
```

And my lldd output:

```
lldb -o 'command script import -r ./misc/lldb_cruby.py' miniruby --  ./test.rb 
(lldb) target create "miniruby"
Current executable set to '/Users/eileencodes/src/github.com/open_source/ruby/miniruby' (arm64).
(lldb) settings set -- target.run-args  "./test.rb"
(lldb) command script import -r ./misc/lldb_cruby.py
lldb scripts for ruby has been installed.
(lldb) b rb_str_hash_m
Breakpoint 1: where = miniruby`rb_str_hash_m + 16 at string.c:4086:35, address = 0x0000000100311a18
(lldb) r
Process 39748 launched: '/Users/eileencodes/src/github.com/open_source/ruby/miniruby' (arm64)
Process 39748 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = breakpoint 1.1
    frame #0: 0x0000000100311a18 miniruby`rb_str_hash_m(str=0x00000001038f69f0) at string.c:4086:35
   4083	static VALUE
   4084	rb_str_hash_m(VALUE str)
   4085	{
-> 4086	    st_index_t hval = rb_str_hash(str);
   4087	    return ST2FIX(hval);
   4088	}
   4089	
Target 0: (miniruby) stopped.
(lldb) rp str
bits: [     ]
T_STRING: [UTF_8] [7BIT] (const char[6]) $1 = "string"
```

etiennebarrie:
I used something like this:
```console
../../configure --prefix=/Users/etienne/.ruby/debug --cache-file=../../config.cache --disable-shared --disable-install-doc --with-ext=openssl,psych,json,+ CFLAGS=-march=native --with-libyaml-dir=/opt/homebrew/opt/libyaml --with-openssl-dir=/opt/homebrew/opt/openssl@3 --with-gmp-dir=/opt/homebrew/opt/gmp --enable-debug-env 'optflags=-O0 -fno-omit-frame-pointer'
```
I removed the CFLAGS and get the same results. The debugging configure flags come from:
https://github.com/ruby/ruby/blob/36daa86022074fd2de07745d8ac2ebb0a390de90/doc/contributing/building_ruby.md?plain=1#L257

I tried with `debugflags='-g3'` too but it's not working still.

eileencodes:
Can you go back to the commit before 0350290262ea0fbc4e1807901797ee8a6970c2b9 and check that it works as expected?

etiennebarrie:
> Can you go back to the commit before [0350290](https://github.com/ruby/ruby/commit/0350290262ea0fbc4e1807901797ee8a6970c2b9) and check that it works as expected?

Arg sorry indeed I can't even get something to work in that case. Something must be wrong with my environment, I've used your configure options on a clean repo and I still get the same behavior. I tried on multiple computers (under macOS) too.

I did just manage to make it work under lima using clang, though 👍 

It does solve the issue in that case. cc @byroot

PR #13060, opened 2025-04-03
compile.c: deduplicate inline constant caches

casperisfine:
If a method reference the same constant twice or more, it can use a single constant cache for all the instructions.

```ruby
puts RubyVM::InstructionSequence.compile(<<~RUBY).disasm
  Foo::Bar + Foo::Bar + Bar + Bar
RUBY
```

Before

```
== disasm: #<ISeq:<compiled>@<compiled>:1 (1,0)-(1,31)>
0000 opt_getconstant_path                   <ic:0 Foo::Bar>           (   1)[Li]
0002 opt_getconstant_path                   <ic:1 Foo::Bar>
0004 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0006 opt_getconstant_path                   <ic:2 Bar>
0008 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0010 opt_getconstant_path                   <ic:3 Bar>
0012 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0014 leave
```

After:

```
== disasm: #<ISeq:<compiled>@<compiled>:1 (1,0)-(1,31)>
0000 opt_getconstant_path                   <ic:0 Foo::Bar>           (   1)[Li]
0002 opt_getconstant_path                   <ic:0 Foo::Bar>
0004 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0006 opt_getconstant_path                   <ic:1 Bar>
0008 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0010 opt_getconstant_path                   <ic:1 Bar>
0012 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0014 leave
```

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -12876,9 +12876,9 @@ ibf_dump_code(struct ibf_dump *dump, const rb_iseq_t *iseq)
                 {
                     IC ic = (IC)op;
                     VALUE arr = idlist_to_array(ic->segments);
-                    wv = ibf_dump_object(dump, arr);
+                    ibf_dump_write_small_value(dump, ibf_dump_object(dump, arr));
                 }
-                break;
+                /* fall through */
               case TS_ISE:
               case TS_IVC:
               case TS_ICVARC:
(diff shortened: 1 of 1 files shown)


casperisfine:
Tested on Shopify's monolith.

Before:

```
718.34 MB  <iseq> (IMEMO)
```

After:

```
 714.05 MB  <iseq> (IMEMO)
```

So the gain is relatively modest. But if we were to do the same for other types of caches, it might add up.

I also tried to graph these with a scatter plot.
Capture d’écran 2025-04-07 à 13 06 10

PR #13065, opened 2025-04-03
Fix regression caused by ruby/ruby#12995

white-axe:
#12995 was reverted by #13026 because it caused some of the tests from [ruby.wasm](https://github.com/ruby/ruby.wasm) to fail. This pull request reverts #13026 and then fixes the bug introduced by #12995 that caused the tests from ruby.wasm to fail.

Requesting a review from @kateinoigakukun.


white-axe:
Closing in favour of #13142.

PR #13067, opened 2025-04-04
Fix the if condition to skip test_ractor.rb correctly

ono-max:
Follow-up for https://github.com/ruby/ruby/commit/a2b03ba7cb721d698bebee74c535dea4583a9c28

### DIFF
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -1636,7 +1636,7 @@ assert_equal "ok", %q{
 
   1_000.times { idle_worker, tmp_reporter = Ractor.select(*workers) }
   "ok"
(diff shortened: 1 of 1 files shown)


hsbt (approved):

bootstraptest/test_ractor.rb
```diff
@@ -1636,7 +1636,7 @@ class C
 
   1_000.times { idle_worker, tmp_reporter = Ractor.select(*workers) }
   "ok"
-} if !yjit_enabled? || ENV['GITHUB_WORKFLOW'] == 'ModGC' # flaky
+} if !yjit_enabled? || ENV['GITHUB_WORKFLOW'] != 'ModGC' # flaky
```
nobu, line 1639: Maybe?
```suggestion
} if !yjit_enabled? && ENV['GITHUB_WORKFLOW'] != 'ModGC' # flaky
```

PR #13071, opened 2025-04-04
Simplify singleton class creation

callmesangio:
Hi all 👋🏻 

This PR aims at simplifying singleton classes' creation a bit, by dropping code that, to my understanding, seems redundant.

The following are my (hopefully correct) assumptions:

- WRT `make_singleton_class`, what I can see is that:
  - `orig_class` is prevented to be a singleton class by client code (see `singleton_class_of`), and
  - `orig_class` cannot be an iclass.

  Thus it should be safe, IIUC, to avoid the call `rb_class_real(orig_class)` entirely.

- WRT `make_metaclass`: 

  The creation of a class is always immediately followed by the creation of the associated singleton class - that is, before any module can be included/prepended to the class. Because of this, `super` cannot be an iclass, and we should be able to safely avoid the search of an actual superclass.

It goes without saying, but given the criticality of the code this PR touches, I'm totally open to make changes (or close the PR entirely) should the above assumptions be wrong, or if the current approach is deemed to be ok as it is, for any reason.
This code has been sitting there for a while, and I'm aware that changing it may be a long shot.

Thanks for your time.
Cheers!

### DIFF
--- a/class.c
+++ b/class.c
@@ -782,7 +782,6 @@ make_metaclass(VALUE klass)
     }
 
     super = RCLASS_SUPER(klass);
-    while (RB_TYPE_P(super, T_ICLASS)) super = RCLASS_SUPER(super);
     RCLASS_SET_SUPER(metaclass, super ? ENSURE_EIGENCLASS(super) : rb_cClass);
 
     // Full class ancestry may not have been filled until we reach here.
--- a/class.c
+++ b/class.c
@@ -808,7 +808,7 @@ make_singleton_class(VALUE obj)
     rb_singleton_class_attached(klass, obj);
     rb_yjit_invalidate_no_singleton_class(orig_class);
 
-    SET_METACLASS_OF(klass, METACLASS_OF(rb_class_real(orig_class)));
+    SET_METACLASS_OF(klass, METACLASS_OF(orig_class));
     return klass;
 }
 


mix127:
Failed tests:
TestProcess#test_uid_from_name: Test::Unit::AssertionFailedError: Exception(ArgumentError) with message matches to /\u{4e0d 5b58 5728}/.

Retrying...
[117/159] TestProcess#test_uid_from_name = 0.00 s                                                   
  1) Failure:
TestProcess#test_uid_from_name [/home/userii/ruby-3.4.2/test/ruby/test_process.rb:1694]:
Exception(ArgumentError) with message matches to /\u{4e0d 5b58 5728}/.
[ArgumentError] exception expected, not #<Errno::ENOENT: No such file or directory - getpwnam_r>.

Finished tests in 70.120731s, 458.7374 tests/s, 93539.3562 assertions/s.                           
32167 tests, 6559048 assertions, 1 failures, 0 errors, 360 skips

ruby -v: ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +PRISM [x86_64-linux]

callmesangio:
I worked a bit more on this and found one case of Ruby code in which my proposed change of `make_metaclass` causes a change in behavior:

```ruby
class C
  extend Module.new
end
C.clone.singleton_class
```

Specifically, the resulting inheritance chain of `C.clone.singleton_class` is different, so obviously the search of a superclass is required.

On the other hand, I wasn't able to write Ruby code that would require the call to `rb_class_real(orig_class)` in `make_singleton_class`. I'm stil curious though, so if somebody is willing to provide more context about that line I would really appreciate it.

All things considered, I think that this PR as it is makes little to no sense anymore, so I will proceed in closing it.
Thanks anyways and sorry for the noise.

PR #13073, opened 2025-04-05
Ensure builtin binaries are aligned to ibf_header

Gelbpunkt:
The `builtin_binary` buffers are cast to `ibf_header` later, but may not be aligned to the necessary boundaries. In particular, Ruby was crashing for me on ARMv7 + musl with LLVM 19.1.7 because it was not aligned to the 4-byte boundary, resulting in `SIGBUS` when dereferencing the header.

https://godbolt.org/z/ss8WeE88c

### DIFF
--- a/template/builtin_binary.inc.tmpl
+++ b/template/builtin_binary.inc.tmpl
@@ -6,7 +6,9 @@
 %   ary = RubyVM.enum_for(:each_builtin).to_a
 %   ary.each{|feature, iseq|
 
-static const unsigned char <%= feature %>_bin[] = {
+#include "compile.h"
+
+static const unsigned char <%= feature %>_bin[] __attribute__((aligned(alignof(struct ibf_header)))) = {
 %     iseq                   \
 %     . to_binary            \
 %     . each_byte            \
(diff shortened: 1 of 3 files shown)


template/builtin_binary.inc.tmpl
```diff
@@ -6,7 +6,9 @@
 %   ary = RubyVM.enum_for(:each_builtin).to_a
 %   ary.each{|feature, iseq|
 
-static const unsigned char <%= feature %>_bin[] = {
+#include "compile.h"
+
+static const unsigned char <%= feature %>_bin[] __attribute__((aligned(alignof(struct ibf_header)))) = {
```
nobu, line 11: This is non-portable clearly.

nobu (changes requested):

template/builtin_binary.inc.tmpl
Gelbpunkt: It works with GCC and Clang. It doesn't with MSVC, I guess I can come up with a way that will also work with MSVC. Is there any other compiler I need to be aware of?

nobu:
IBF dump string size should be multiple of 4 always.

PR #13074, opened 2025-04-06
Implement Set as a core class

jeremyevans:
Set has been an autoloaded standard library since Ruby 3.2.  The standard library Set is less efficient than it could be, as it uses Hash for storage, which stores unnecessary values for each key.

Implementation details:

* Core Set uses a modified version of `st_table`, named `set_table`.  Other than `s/st_/set_/`, the main difference is that the stored records do not have values, making them 1/3 smaller. `st_table_entry` stores `hash`, `key`, and `record` (value), while `set_table_entry` only stores `hash` and `key`.  This results in large sets using ~33% less memory compared to stdlib Set.  For small sets, core Set uses 20% less memory (160 bytes, while stdlib set uses 40 for Set and 160 for Hash).

* All methods are implemented as cfuncs, except the pretty_print methods, which were moved to `lib/pp.rb` (which is where the pretty_print methods for other core classes are defined).  As is typical for core classes, internal calls call C functions and not Ruby methods.  For example, to check if something is a Set, `rb_obj_is_kind_of` is used, instead of calling `is_a?(Set)` on the related object.

* Almost all methods use the same algorithm that the pure-Ruby implementation used.  The exception is when calling `Set#divide` with a block with 2-arity.  The pure-Ruby method used tsort to implement this.  I developed an algorithm that only allocates a single intermediate hash and does not need tsort.

* The `flatten_merge` protected method is no longer necessary, so it is not implemented (it could be).

* Similar to Hash/Array, subclasses of Set are no longer reflected in `inspect` output.

* RDoc from stdlib Set was moved to core Set, with minor updates.

This includes a comprehensive benchmark suite for all public Set methods.  As you would expect, the native version is faster in the vast majority of cases, and multiple times faster in many cases.  There are a few cases where it is significantly slower:

* Set.new with no arguments (~1.6x)
* Set#compare_by_identity for small sets (~1.3x)
* Set#clone for small sets (~1.5x)
* Set#dup for small sets (~1.7x)

These are slower as Set does not currently use the AR table optimization that Hash does, so a new set_table is initialized for each call.  I'm not sure it's worth the complexity to have an AR table-like optimization for small sets (for hashes it makes sense, as small hashes are used everywhere in Ruby).

The rbs and repl_type_completor bundled gems will need updates to support core Set.  The pull request marks them as allowed failures.

This passes all set tests with no changes.  The following specs needed modification:

* Modifying frozen set error message (changed for the better)
* `Set#divide` when passed a 2-arity block no longer yields the same object as both the first and second argument (this seems like an issue with the previous implementation).
* Set-like objects that override `is_a?` such that `is_a?(Set)` return `true` are no longer treated as Set instances.
* `Set.allocate.hash` is no longer the same as `nil.hash`
* `Set#join` no longer calls `Set#to_a` (it calls the underlying C function).
* `Set#flatten_merge` protected method is not implemented.

Previously, `set.rb` added a `SortedSet` autoload, which loads `set/sorted_set.rb`.  This replaces the `Set` autoload in `prelude.rb` with a `SortedSet` autoload, but I recommend removing it and `set/sorted_set.rb`.

This moves `test/set/test_set.rb` to `test/ruby/test_set.rb`, reflecting that switch to a core class.  This does not move the spec files, as I'm not sure how they should be handled.


jeremyevans:
`update_bundled_gems` CI failure is due to release of test-unit 3.6.8 (which was today), which moved the file that the CI workflow is trying to use.  Related test-unit commit: https://github.com/test-unit/test-unit/commit/b7d3c32f6e334e1823e30c053c2268893cf073ef

set.c
```diff
@@ -0,0 +1,3555 @@
...
+#endif
+#ifndef ANYARGS
+# ifdef __cplusplus
+#   define ANYARGS ...
+# else
+#   define ANYARGS
+# endif
+#endif
+
+typedef struct set_table set_table;
```
byroot, line 68: It would be good to split the `set_table` internal implementation and the `Set` type, because there are other places in the VM that could benefit from using a `set_table` instead of a `st_table`, one example being the `fstring_table`.

set.c
```diff
@@ -0,0 +1,3555 @@
...
+static CONSTFUNC(set_index_t set_hash_end(set_index_t h));
+static CONSTFUNC(set_index_t set_hash_start(set_index_t h));
+
+/* The type of hashes.  */
+typedef set_index_t set_hash_t;
+
+struct set_table_entry {
+    set_hash_t hash;
+    set_data_t key;
+};
```
byroot, line 147: Is it really worth having `set_hash_t / set_data_t / set_index_t`? I might be off, but I think re-using `st_*` types would be easier and would make it trivial to covert code using an `st_table` into using a `set_table`?

hsbt:
@jeremyevans 

>update_bundled_gems CI failure is due to release of test-unit 3.6.8 (which was today),

I fixed this at https://github.com/ruby/ruby/pull/13078

set.c
jeremyevans: That makes sense.  I'll work on that.

set.c
jeremyevans: I made the types deliberately different to avoid ambiguity. However, there is no technical reason we couldn't use the same types, and I agree that it could make conversion from `st_table` easier.

If we want to reuse `st_*` types, one option is to move the `set_table` code into `st.c`.  There are some functions that are the same between the two implementations, so it would eliminate some duplicate code.  What are your thoughts on that?

byroot: > What are your thoughts on that?

That would be fine by me.

For the record there's an in-flight PR that would also benefit from a `set_table`: https://github.com/ruby/ruby/pull/13077

jeremyevans: Giving some more thought to this, before making changes, I think we should discuss this at the developer meeting next week, as other developers may have thoughts about where the code should go.  I'll add this to the meeting agenda.

ParadoxV5:
🆙

---

Here’s 2¢ from my recent experience:
Instead of `HashSet[E]` implemented as `HashMap[E, truthy_dummy]`, it could be the converse, with `HashMap[K, V]` implemented as `HashSet[KVPair[K, V]]`.
```ruby
KVPair = Struct.new(:key, :value) do
  def hash = ~key.hash
  def eql?(other) = key.eql?(other.key) rescue nil
end
```

jeremyevans:
> 🆙
> 
> Here’s 2¢ from my recent experience: Instead of `HashSet[E]` implemented as `HashMap[E, truthy_dummy]`, it could be the converse, with `HashMap[K, V]` implemented as `HashSet[KVPair[K, V]]`.
> 
> ```ruby
> KVPair = Struct.new(:key, :value) do
>   def hash = ~key.hash
>   def eql?(other) = key.eql?(other.key) rescue nil
> end
> ```

Can you explain the advantage of doing that?  Seems to use more objects and is probably slower, plus it would require fairly invasive changes.  Additionally, `HashSet` does not support updating elements already in the set, while updating values for keys already in the map is a requirement of `HashMap`.  Supporting both `HashSet[E]` and `HashMap[K,V]` (as this pull request does) seems better to me, there is no requirement to implement one in terms of the other.

set.c
```diff
@@ -0,0 +1,3555 @@
...
+
+static const rb_data_type_t set_data_type = {
+    .wrap_struct_name = "set",
+    .function = {
+        .dmark = set_mark,
+        .dfree = set_free,
+        .dsize = set_size,
+        .dcompact = set_compact,
+    },
+    .flags = 0
```
byroot, line 1689: As discussed last night, you can set `TYPED_FREE_IMMEDIATELY`

byroot:
Shower thought: Probably doesn't matter, or at least largely offseted by the more efficient implementation, but we're losing the AR_TABLE optimization for small sets.

jeremyevans:
This PR now implements what was agreed to at the dev meeting:

* set_table now uses st_* types/functions where possible
* set_table implementation lives in st.c
* internal/set_table.h is added, allowing use of set_table
  by places other than set.c
* set.c only contains the Set-specific logic

Additionally, it implements the following improvements:

* Embeds set_table into object slot (currently results in
  a significant amount of wasted memory per Set, since
  struct set_table is 56 bytes, so the object is 88 bytes
  when embedded, just over the 80 byte limit, so it
  occupies a 160 byte slot)

* GC.compact support

* Write barrier support

* Optimizes some common Set methods when used with arrays,
  by using the array ptr to access elements instead of
  calling methods on the array.

I think this is now ready to be merged.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+                set_table *into = RSET_TABLE(set);
+                VALUE key;
+                int block_given = rb_block_given_p();
+                RARRAY_PTR_USE(other, ptr, {
+                    for(; len > 0; len--, ptr++) {
+                        key = *ptr;
+                        if (block_given) key = rb_yield(key);
+                        key = set_stringify(into, key);
+                        RB_OBJ_WRITTEN(set, Qundef, key);
+                        set_insert(into, (st_data_t)key);
```
byroot, line 508:
```suggestion
                        set_insert(into, (st_data_t)key);
                        RB_OBJ_WRITTEN(set, Qundef, key);
```

set.c
```diff
@@ -0,0 +1,2149 @@
...
+    if (set_iterating_p(set)) {
+        if (!set_lookup(RSET_TABLE(set), (st_data_t)item)) {
+            no_new_item();
+        }
+    }
+    else {
+        set_table *table = RSET_TABLE(set);
+        item = set_stringify(table, item);
+        RB_OBJ_WRITTEN(set, Qundef, item);
+        set_insert(table, (st_data_t)item);
```
byroot, line 694:
```suggestion
        set_insert(table, (st_data_t)item);
        RB_OBJ_WRITTEN(set, Qundef, item);
```

set.c
```diff
@@ -0,0 +1,2149 @@
...
+        if (!set_lookup(RSET_TABLE(set), (st_data_t)item)) {
+            no_new_item();
+        }
+        return Qnil;
+    }
+    else {
+        set_table *table = RSET_TABLE(set);
+        item = set_stringify(table, item);
+        RB_OBJ_WRITTEN(set, Qundef, item);
+        return set_insert(table, (st_data_t)item) ? Qnil : set;
```
byroot, line 724:
```suggestion
        if (set_insert(table, (st_data_t)item)) {
            return Qnil
        }
        RB_OBJ_WRITTEN(set, Qundef, item);
        return set;
```

We shouldn't trigger the write barrier if we didn't insert the item.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+};
+
+static int
+set_intersection_i(st_data_t key, st_data_t tmp)
+{
+    struct set_intersection_data *data = (struct set_intersection_data *)tmp;
+    if (set_lookup(data->other, key)) {
+        key = set_stringify(data->into, (VALUE)key);
+        RB_OBJ_WRITTEN(data->set, Qundef, (VALUE)key);
+        set_insert(data->into, key);
```
byroot, line 1010:
```suggestion
        set_insert(data->into, key);
        RB_OBJ_WRITTEN(data->set, Qundef, (VALUE)key);
```

set.c
```diff
@@ -0,0 +1,2149 @@
...
+  set_table *into;
+};
+
+static int
+set_merge_i(st_data_t key, st_data_t data)
+{
+    struct set_merge_args *args = (struct set_merge_args *)data;
+    key = set_stringify(args->into, (VALUE)key);
+    RB_OBJ_WRITTEN(args->set, Qundef, (VALUE)key);
+    set_insert(args->into, key);
```
byroot, line 1108:
```suggestion
    set_insert(args->into, key);
    RB_OBJ_WRITTEN(args->set, Qundef, (VALUE)key);
```

byroot: Might actually be worth having a small helper to insert and then trigger the write barrier if the value was actually inserted, given how frequent that pattern is.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+    return ST_CONTINUE;
+}
+
+static VALUE
+set_merge_block(RB_BLOCK_CALL_FUNC_ARGLIST(key, set))
+{
+    set_table *into = RSET_TABLE(set);
+    key = set_stringify(into, key);
+    RB_OBJ_WRITTEN(set, Qundef, key);
+    set_insert(into, (st_data_t)key);
```
byroot, line 1118: Here too.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+    else if (RB_TYPE_P(arg, T_ARRAY)) {
+        long len = RARRAY_LEN(arg);
+        if (RARRAY_LEN(arg) != 0) {
+            set_table *into = RSET_TABLE(set);
+            VALUE key;
+            RARRAY_PTR_USE(arg, ptr, {
+                for(; len > 0; len--, ptr++) {
+                    key = set_stringify(into, *ptr);
+                    RB_OBJ_WRITTEN(set, Qundef, key);
+                    set_insert(into, (st_data_t)key);
```
byroot, line 1141: Here too.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+    size_t size = set_table_size(old);
+    if (size > 0) {
+        set_table *new = set_init_table_with_size(NULL, type, size);
+        struct set_merge_args args = {
+            .set = set,
+            .into = new
+        };
+        set_iter(set, set_merge_i, (st_data_t)&args);
+        set_free_embedded(sobj);
+        memcpy(&sobj->table, new, sizeof(*new));
```
byroot, line 1193: We might have created new references here, no? If so we need to trigger write barriers.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+static int
+set_xor_i(st_data_t key, st_data_t data)
+{
+    VALUE set = (VALUE)data;
+    set_table *table = RSET_TABLE(set);
+    key = set_stringify(table, (VALUE)key);
+    RB_OBJ_WRITTEN(set, Qundef, (VALUE)key);
+    if (set_insert(table, key)) {
+        set_delete(table, &key);
+    }
```
byroot, line 1267: Here too the `RB_OBJ_WRITTEN` need to be moved.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+ *  given enumerable object.  <tt>(set ^ enum)</tt> is equivalent to
+ *  <tt>((set | enum) - (set & enum))</tt>.
+ *
+ *    Set[1, 2] ^ Set[2, 3]                   #=> #<Set: {3, 1}>
+ *    Set[1, 'b', 'c'] ^ ['b', 'd']           #=> #<Set: {"d", 1, "c"}>
+ */
+static VALUE
+set_i_xor(VALUE set, VALUE other)
+{
+    other = rb_funcall(rb_obj_class(set), id_new, 1, other);
```
byroot, line 1285: There is a is a few places like this where we call `Set.new` from C. Should we? Or should we just call `set_alloc` and sometime benefit from knowing the resulting size upfront?

I suppose it changes the behavior in case `Set.new` is replaced / decorated.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+}
+
+static int
+set_collect_i(st_data_t key, st_data_t data)
+{
+    VALUE set = (VALUE)data;
+    set_table *table = RSET_TABLE(set);
+    key = set_stringify(table, rb_yield((VALUE)key));
+    RB_OBJ_WRITTEN(set, Qundef, (VALUE)key);
+    set_insert(table, key);
```
byroot, line 1396: Here again.

st.c
```diff
@@ -2332,4 +2333,882 @@ rb_st_compact_table(st_table *tab)
...
+{
+    st_index_t bound = tab->entries_bound;
+
+    if (bound == set_get_allocated_entries(tab))
+        set_rebuild_table(tab);
+}
+
+/* Insert (KEY, VALUE) into table TAB and return zero.  If there is
+   already entry with KEY in the table, return nonzero and update
+   the value of the found entry.  */
```
byroot, line 2875: This comment need to be updated given there is no `VALUE`.

st.c
```diff
@@ -2332,4 +2333,882 @@ rb_st_compact_table(st_table *tab)
...
+        entry->key = key;
+        if (bin_ind != UNDEFINED_BIN_IND)
+            set_bin(tab->bins, set_get_size_ind(tab), bin_ind, ind + ENTRY_BASE);
+        return 0;
+    }
+    return 1;
+}
+
+/* Insert (KEY, VALUE, HASH) into table TAB.  The table should not have
+   entry with KEY before the insertion.  */
```
byroot, line 2919: No `VALUE` here either.

st.c
```diff
@@ -2332,4 +2333,882 @@ rb_st_compact_table(st_table *tab)
...
+   corresponding to deleted entries. */
+#define MARK_SET_BIN_DELETED(tab, i)				\
+    do {                                                        \
+        set_bin((tab)->bins, set_get_size_ind(tab), i, DELETED_BIN); \
+    } while (0)
+
+/* Delete entry with KEY from table TAB, set up *VALUE (unless
+   VALUE is zero) from deleted table entry, and return non-zero.  If
+   there is no entry with KEY in the table, clear *VALUE (unless VALUE
+   is zero), and return zero.  */
```
byroot, line 3003: No `VALUE`.

st.c
```diff
@@ -2332,4 +2333,882 @@ rb_st_compact_table(st_table *tab)
...
+}
+
+/* Traverse all entries in table TAB calling FUNC with current entry
+   key and value and zero.  If the call returns ST_STOP, stop
+   traversing.  If the call returns ST_DELETE, delete the current
+   entry from the table.  In case of ST_CHECK or ST_CONTINUE, continue
+   traversing.  The function returns zero unless an error is found.
+   CHECK_P is flag of set_foreach_check call.  The behavior is a bit
+   different for ST_CHECK and when the current element is removed
+   during traversing.  */
```
byroot, line 3047: No `value`

byroot (changes requested):

set.c
jeremyevans: Not needed in this function as the write barrier are added in `set_merge_i` (called a couple lines above).

set.c
jeremyevans: It does change behavior in that case, which is why I did it originally.  But I think it makes sense to switch to direct allocation, similar to how we call C functions instead of Ruby methods in general inside the implementation.

st.c
jeremyevans: Fixed this and the other cases.

jeremyevans (commented):
@byroot Thank you very much for your detailed review.  I just pushed commits to address all of the issues you pointed out.

set.c
```diff
@@ -0,0 +1,2149 @@
...
+static ID id_each_entry;
+static ID id_any_p;
+static ID id_new;
+static ID id_set_iter_lev;
+
+#define RSET_INITIALIZED FL_USER1
+#define RSET_LEV_MASK (FL_USER13 | FL_USER14 | FL_USER15 |                /* FL 13..19 */ \
+                        FL_USER16 | FL_USER17 | FL_USER18 | FL_USER19)
+#define RSET_LEV_SHIFT (FL_USHIFT + 13)
+#define RSET_LEV_MAX 127 /* 7 bits */
```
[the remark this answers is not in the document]
byroot: If your idea is what is mentioned in https://github.com/ruby/ruby/pull/13074/commits/9f9821ad1d4a2ea24cacc7a24eb2f36bc96c997c (getting rid of `typed_flag`), it's not as simple.

The GC need to store one bit of information to know if the object was actually embedded or not (since allocation size is dynamic, if you ask for more than the GC can provide, you won't be embeded). 

I have an idea how to reclaim that in the future, but it's a lot of work.

At the end of the day, I don't think `Set` being size `160B` is a big deal. It's a bit of a shame we're wasting 72B, but that's still smaller than what the previous implementation was.

jeremyevans:
> Looks good to me.
> 
> However I don't know if you cherry-picked the commits from [jeremyevans#2](https://github.com/jeremyevans/ruby/pull/2), but it seems like @mrnoname1000 authorship was lost in the process. Would be nice to add him back.

Certainly.  I squashed this into a single commit and added him as a co-author.  I'll merge after this passes CI.

jeremyevans:
CI is only failing on the same VM_DEBUG ractor btest that is also failing on master, so I am going to merge this.

MSP-Greg:
@jeremyevans

Haven't had time to look into this, but while running GHA tests today, the following patch worked:
```text
ruby 3.5.0dev (2025-04-24T18:33:11Z :detached: 71166f60d9) +PRISM [x86_64-linux]
```

A few hours later, the following:
```text
ruby 3.5.0dev (2025-04-26T11:56:42Z :detached: 687bd83724) +PRISM [x86_64-linux]
```
The logged error was:
```text
ActionView::Template::Error: no _dump_data is defined for class Set
    app/views/layouts/application.html.erb:7
```

This was running tests against turbo-rails in Puma...

byroot:
Thanks for pointing it out, that is indeed an oversight, won't be too hard to fix.

jeremyevans:
Yep, I'll work on a patch for Marshal support for Set.

jeremyevans:
PR submitted to support marshalling: https://github.com/ruby/ruby/pull/13185

PR #13077, opened 2025-04-06
Optimize callcache invalidation for refinements

alpaca-tc:
Fixes [Bug [#21201](https://bugs.ruby-lang.org/issues/21201)]

This change addresses a performance regression where defining methods inside `refine` blocks caused severe slowdowns. The issue was due to `rb_clear_all_refinement_method_cache()` triggering a full object space scan via `rb_objspace_each_objects` to find and invalidate affected callcaches, which is very inefficient.

To fix this, I introduce `vm->cc_refinement_table` to track callcaches related to refinements. This allows us to invalidate only the necessary callcaches without scanning the entire heap, resulting in significant performance improvement.

## benchmark

```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "benchmark-ips"
end

mod = Module.new
klass = Class.new

Benchmark.ips do |x|
  x.report(RUBY_VERSION) do
    mod.send(:refine, klass) do
      def call_1 = nil
      def call_2 = nil
      def call_3 = nil
    end
  end

  x.save! "/tmp/performance_regression_refine.bench"
  x.compare!
end
```

```
ruby 3.5.0dev (2025-06-05T14:33:43Z bugs/21201 e7833e82f4) +PRISM [arm64-darwin24]
         3.5.0_patch    50.158k i/100ms
         3.5.0_patch    520.371k (± 7.2%) i/s    (1.92 μs/i) -      2.608M in   5.038852s

Comparison:
               3.2.8:  1597482.0 i/s
         3.5.0_patch:   520371.1 i/s - 3.07x  slower (this PR)
               3.3.8:      213.0 i/s - 7499.45x  slower
               3.4.4:      138.2 i/s - 11561.20x  slower
3.5.0_master_0e0008da0f19d:      137.8 i/s - 11593.43x  slower
```


vm.c
```diff
@@ -4443,6 +4448,7 @@ Init_vm_objects(void)
     vm->loading_table = st_init_strtable();
     vm->ci_table = st_init_table(&vm_ci_hashtype);
     vm->frozen_strings = st_init_table_with_size(&rb_fstring_hash_type, 10000);
+    vm->cc_refinement_table = st_init_table(&vm_cc_refinement_hashtype);
```
byroot, line 4451:
```suggestion
    vm->cc_refinement_table = st_init_numtable();
```

I think you can just use a standard `numtable`, and get rid of your `hash` and `cmp` functions.

ko1:
I'm not sure about compaction, but is it safe for `GC.compact`?
cc/ @tenderlove

vm.c
alpaca-tc: Ah, I see! I force-pushed the fix.

byroot:
Nevermind, It's already handled because it's declared as a weak table.

vm.c
alpaca-tc: Does numtable compute hash values from address?
If so, it might be necessary to rebuild and recalculate the hash values after GC compaction.
Alternatively, using a unique object_id as in the original implementation could result in stable hash values.

Which is better: rebuilding or the original approach?
If rebuilding, where should it happen? I’ll investigate tomorrow.

However, this is my first time writing C code, so I might be saying something incorrect.

byroot: > Does numtable compute hash values from address?

Yes it does.

> If so, it might be necessary to rebuild and recalculate the hash values after GC compaction.

Indeed, I thought it being declared as a weak table would be enough, but looking closer at the code, it update the value in place, ignoring the hash.

So we indeed either need an address independent hash, or to use a more complex way to update that table after compaction.

> Alternatively, using a unique object_id as in the original implementation could result in stable hash values.

`object_id` insert that object in two other tables (`obj_to_id` and `id_to_obj`), so it's not ideal.

byroot: > or to use a more complex way to update that table after compaction.

One simple way could just be to rebuild it entirely after compaction. Iterate on the old one and insert in a new one, then swap them.

tenderlove:
> I'm not sure about compaction, but is it safe for `GC.compact`?

The reference updating code looks correct to me

vm.c
XrXr: You should probably reuse a `compare_by_identity` Hash. Search for hidden_identity_hash_new() usages. Some other refinement logic already uses it.

vm_method.c
```diff
@@ -303,27 +303,27 @@ rb_clear_method_cache(VALUE klass_or_module, ID mid)
...
-                }
-            }
-        }
+invalidate_cc_refinement(st_data_t key, st_data_t value, st_data_t data)
+{
+    VALUE v = (VALUE)value;
+    void *ptr = rb_asan_poisoned_object_p(v);
+    rb_asan_unpoison_object(v, false);
+
+    if (RBASIC(v)->flags) { // liveness check
```
XrXr, line 312: You need `rb_gc_pointer_to_heap_p(v) && !rb_objspace_garbage_object_p(v)` before you can dereference `v` now, since it's in a weak table. By the time this runs `v` could point to an unmapped memory. cc @peterzhu2118

alpaca-tc: Thank you for the review! I've applied the changes and force-pushed the update.

peterzhu2118: Can you rerun the benchmarks? `rb_gc_pointer_to_heap_p` is a pretty slow API, and will be slower the larger the heap (e.g. in a system with a lot of live objects such as a Rails app).

vm.c
alpaca-tc: I updated the code to perform rehashing (deletion and reinsertion) within the foreach loop, following the approach used in `vm_weak_table_gen_ivar_foreach`.

Also, `value` was added for compatibility with vm_weak_table_foreach_update_weak_key, but it's no longer needed after switching to foreach. So now I'm passing `1` instead of `cc`, as the value is unused.
Hope #13074 gets merged soon 👍

vm_method.c
alpaca-tc: Updated the benchmark in the description 👍 
スクリーンショット 2025-04-09 0 07 31

It's an unofficial benchmark, but the performance improvement is clear enough in our Rails application where the original performance regression was discovered and I'm happy with the result.

```
> $ bundle exec rails r 'puts Benchmark.measure { Rails.application.eager_load! }'
> before 12.689740   1.162582  13.852322 ( 14.480201)
> after   2.317027   0.605092   2.922119 (  3.348900)

> $ touch app/controllers/application_controller.rb && time curl http://localhost:3000/
> before 0.00s user 0.00s system 0% cpu 11.263 total
> after  0.00s user 0.01s system 0% cpu 2.503 total
```

alpaca-tc:
CI failed due to a network error, so I'll rebase and force-push to trigger it again.

vm.c
alpaca-tc: I considered the copy-then-swap approach, but if no objects move, it wastes memory by duplicating data unnecessarily.
I'm going with a naive implementation for now, though I'm not confident it's the right choice.
If GC compaction moves most cc, I'd prefer copy-then-swap.
I'd appreciate any insights on how GC compaction behaves.

alpaca-tc: copy-then-swap seems better after all. It's fast when there are few entries, and if there are many, ST_REPLACE will likely happen anyway.
@byroot  Updated the code, would appreciate a review 😄

gc.c
```diff
@@ -3532,6 +3541,64 @@ vm_weak_table_frozen_strings_foreach(st_data_t key, st_data_t value, st_data_t d
...
+    return ret;
+}
+
+static st_table *
+vm_weak_table_cc_refinement_rebuild(st_table *cc_refinement_table, struct global_vm_table_foreach_data *iter_data)
+{
+    st_table *new_tbl = NULL;
+    DURING_GC_COULD_MALLOC_REGION_START();
+    {
+        new_tbl = st_init_numtable();
```
byroot, line 3589:
```suggestion
        new_tbl = st_init_numtable_with_size(st_table_size(cc_refinement_table));
```

It's better to directly allocate it with a size large enough, so that it won't reallocate multiple times when inserting.

alpaca-tc: Agreed, updated accordingly.

vm.c
alpaca-tc: Thanks to everyone who's already reviewed this-no issues have been spotted so far. @peterzhu2118 , when you have a moment, could you take a look at the GC-related changes as well?

alpaca-tc:
I’m fixing the CI errors after rebasing 😕

PR #13081, opened 2025-04-08
Add an assertion that rb_sourceline() never return 0

byroot:
@mame what do you think? Or perhaps just a regular `RUBY_ASSERT` so it's optimized out, but I fear we're not that likely to run into it in the test suite.


byroot:
THis can happen legitimately with `eval`

PR #13084, opened 2025-04-08
Add type check for `rb_big_modulo` first argument

ianks:
This PR adds type checking to `rb_big_modulo` to ensure the first argument is a Bignum, preventing potential segmentation faults or undefined behavior when called with improper types.

I understand the API is meant to be called with a Bignum as first argument, but I think the check is worth it because it adds a safety layer that makes the codebase more robust against misuse.

Several other rb_big_* functions lack similar type checking and could benefit from the same fix, including:
- `rb_big_plus`
- `rb_big_minus`
- `rb_big_mul`
- `rb_big_div`
- `rb_big_idiv`
- `rb_big_divmod`
- `rb_big_and`
- `rb_big_or`
- `rb_big_xor`
- `rb_big_pack`

I wanted to see Ruby core's opinion on this approach before implementing similar fixes for these other functions.

### DIFF
--- a/bignum.c
+++ b/bignum.c
@@ -6136,6 +6136,9 @@ rb_big_modulo(VALUE x, VALUE y)
 {
     VALUE z;
 
+    if (!RB_BIGNUM_TYPE_P(x)) {
+        rb_raise(rb_eTypeError, "expected a bignum");
+    }
     if (FIXNUM_P(y)) {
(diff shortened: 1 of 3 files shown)


nobu:
I think it is the callers' responsibility to call with the correct receiver.
This is not only in bignum.c, in string.c, array.c, hash.c, etc. as well.

PR #13085, opened 2025-04-08
Restore the original order of const_added and inherited callbacks

fxn:
Originally, if a class was defined with the `class` keyword, the cref had a `const_added` callback, and the superclass an `inherited` callback, `const_added` was called first, and inherited second.

This was discussed in [#21143](https://bugs.ruby-lang.org/issues/21143) and an attempt at changing this order was made.

While both constant assignment and inheritance have happened before these callbacks are invoked, it was deemed nice to have the same order as in

    C = Class.new

This was mostly for alignment: In that last use case things happen at different times and therefore the order of execution is kind of obvious, whereas when the `class` keyword is involved, the order is opaque to the user and it is up to the interpreter.

However, soon in [#21193](https://bugs.ruby-lang.org/issues/21193), Matz decided to play safe and keep the existing order.

This reverts commits de097fbe5f3df105bd2a26e72db06b0f5139bc1a and de48e47ddf78aba02fd9623bc7ce685540a10743.


spec/ruby/optional/capi/class_spec.rb
```diff
@@ -385,17 +385,6 @@
...
-    ruby_version_is "3.5" do
-      it "calls .inherited before .const_added" do
-        ScratchPad.record([])
-        @s.rb_define_class_id_under(CApiClassSpecs::Callbacks, :Subclass, CApiClassSpecs::Callbacks)
-        ScratchPad.recorded.should == [
-          [:inherited, "CApiClassSpecs::Callbacks::Subclass", :location],
-          [:const_added, :Subclass, :location],
-        ]
-      end
-    end
```
byroot, line 398: I think we should keep these specs, just update them to match the old behavior.

fxn: I have that one in my [original/followup patch](https://github.com/ruby/ruby/pull/12759/files).

fxn: Let me add that my test verifies exactly callback execution order, which is what the test is about. I'll add @eregon suggestion too.

Then, I could add tests that say: In both callbacks, inheritance has happened, and constant assignment has happened (there is a location, `constants` has the constant, the class has a permanent name, etc.).

PR #13095, opened 2025-04-10
Remove unused stack space.

ioquatix:
I could not see the point of this code, so it looks like we can remove it.

It was introduced in https://github.com/ruby/ruby/commit/87e1616048edca270c86ab7e5d8b8599e1df79de


ioquatix:
I split this into two smaller PRs:

* #17435 removes the unused native thread stack size field and related dead code without changing stack-overflow detection behavior.
* #17436 is stacked on #17435 and contains the visible behavior change: using a page-sized stack-overflow detection window via `getpagesize()`.

This should make it easier to review the no-behavior cleanup separately from the stack-overflow detection heuristic change.

PR #13096, opened 2025-04-10
Fix LLDB scripts by using the toplevel RBasic

etiennebarrie:
With @eightbitraptor we figured the issue I had in #13048 by using `target.FindTypes` instead and looking at the results.

`target.FindFirstType("struct RBasic")` was returning the `yjit::cruby::autogened::RBasic` type instead of `struct RBasic` and then somehow getting the flags always returned 0.

Explicitly asking for the toplevel namespace `RBasic` by prefixing with `::` fixes that issue.

I also took the opportunity to fix the `heap_page` command by moving some parts of `RbInspector` into `LLDBInterface` and setting the `result` in the `call` method.

`rbbt` is still broken for me and `dump_page` just never returns or is very slow, but I still fixed the usage there for good measure.

**Before**
```console
(lldb) rp str
bits: [     ]
T_NONE: No value
(lldb) old_rp str
bits: [     ]
T_NONE: (yjit::cruby::autogened::RBasic) *$1 = {
  flags = (__0 = 5259269)
  klass = (__0 = 4304792640)
}
(lldb) heap_page str
Traceback (most recent call last):
  File "/Users/etienne/src/github.com/ruby/ruby/misc/lldb_rb/rb_base_command.py", line 48, in __call__
    self.call(debugger, command, exe_ctx, result)
  File "/Users/etienne/src/github.com/ruby/ruby/misc/lldb_rb/commands/heap_page_command.py", line 17, in call
    self._append_expression(debugger, "(struct heap_page *) %0#x" % page.GetValueAsUnsigned(), result)
AttributeError: 'HeapPageCommand' object has no attribute '_append_expression'
```

**After**
```console
(lldb) rp str
bits: [     ]
T_STRING: [UTF_8] [7BIT] (const char[6]) $1 = "string"
(lldb) old_rp str
bits: [     ]
T_STRING: [UTF_8] (const char[6]) $3 = "string"
(lldb) heap_page str
(struct heap_page *) $5 = 0x000000011e84b000
(struct heap_page) $6 = {
  slot_size = 40
  total_slots = 1638
# ✄
```

### DIFF
--- a/misc/lldb_cruby.py
+++ b/misc/lldb_cruby.py
@@ -68,7 +68,7 @@ class BackTrace:
         return self.VM_FRAME_MAGIC_NAME.get(frame_type, "(none)")
 
     def rb_iseq_path_str(self, iseq):
-        tRBasic = self.target.FindFirstType("struct RBasic").GetPointerType()
+        tRBasic = self.target.FindFirstType("::RBasic").GetPointerType()
 
         pathobj = iseq.GetValueForExpressionPath("->body->location.pathobj")
         pathobj = pathobj.Cast(tRBasic)
@@ -270,7 +270,7 @@ def lldb_inspect(debugger, target, result, val):
     elif num & RUBY_IMMEDIATE_MASK:
         print('immediate(%x)' % num, file=result)
     else:
-        tRBasic = target.FindFirstType("struct RBasic").GetPointerType()
+        tRBasic = target.FindFirstType("::RBasic").GetPointerType()
 
         val = val.Cast(tRBasic)
         flags = val.GetValueForExpressionPath("->flags").GetValueAsUnsigned()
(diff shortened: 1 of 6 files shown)


misc/lldb_cruby.py
```diff
@@ -68,7 +68,7 @@ def vm_frame_magic(self, cfp):
         return self.VM_FRAME_MAGIC_NAME.get(frame_type, "(none)")
 
     def rb_iseq_path_str(self, iseq):
-        tRBasic = self.target.FindFirstType("struct RBasic").GetPointerType()
+        tRBasic = self.target.FindFirstType("::RBasic").GetPointerType()
```
eightbitraptor, line 71: I had no idea you could reference constants like this! This is great.

eightbitraptor (approved):
Looks good to me

PR #13106, opened 2025-04-13
Use csv 3.3.4

kou:
csv 3.3.3 introduced an experimental executable `csv-filter` but 3.3.4 removed it. So we can use released version.

Fixes [Bug #21207]


hsbt (approved):
Thanks always.

PR #13108, opened 2025-04-13
[Bug #21256] Fix `it` parameter when splatting and `define_method` is used

Earlopain:
https://bugs.ruby-lang.org/issues/21256

It was failing to set the leads, like numblocks do, causing the result to be wrapped in an array


tenderlove (approved):

mame:
@Earlopain @tenderlove Sorry but this change caused an ASAN error. I would like to revert this for now. See https://github.com/ruby/ruby/pull/13947.

PR #13115, opened 2025-04-14
Lazily create `objspace->id_to_obj_tbl`

byroot:
This inverse table is only useful if `ObjectSpace._id2ref` is used, which is extremely rare. The only notable exception is the `drb` gem and even then it has an option not to rely on `_id2ref`.

So if we assume this table will never be looked up, we can just not maintain it, and if it turns out `_id2ref` is called, we can lock the VM and re-build it.

```
compare-ruby: ruby 3.5.0dev (2025-04-10T09:44:40Z master 684cfa42d7) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-04-10T10:13:43Z lazy-id-to-obj d3aa9626cc) +YJIT +PRISM [arm64-darwin24]
warming up..

|           |compare-ruby|built-ruby|
|:----------|-----------:|---------:|
|baseline   |     26.364M|   25.974M|
|           |       1.01x|         -|
|object_id  |     10.293M|   14.202M|
|           |           -|     1.38x|
```

Reopening of https://github.com/ruby/ruby/pull/13094 for logistical reasons.

### DIFF
--- a/benchmark/object_id.yml
+++ b/benchmark/object_id.yml
@@ -1,0 +1,4 @@
+benchmark:
+  baseline: "Object.new"
+  object_id: "Object.new.object_id"
+# loop_count: 100000
--- a/gc.c
+++ b/gc.c
@@ -1794,11 +1794,11 @@ id2ref(VALUE objid)
                     return ptr;
                 }
                 else {
-                    rb_raise(rb_eRangeError, "%p is not symbol id value", (void *)ptr);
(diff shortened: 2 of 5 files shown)


byroot:
The `macos-15` failure seem very consistent, but I don't get how it could be related to my change, and I haven't been able to reproduce it yet.

byroot:
Ah, looks like it's failing on other people's PRs.

PR #13127, opened 2025-04-18
Make `waiting_fd` behaviour per-IO.

ioquatix:
See <https://github.com/ruby/ruby/pull/5384> for full context.


ko1:
> io_reopen(VALUE io, VALUE nfile)

image

we should replace it.

internal/io.h
```diff
@@ -13,11 +13,23 @@
 #define HAVE_RB_IO_T
 struct rb_io;
 
-#include "ruby/io.h"            /* for rb_io_t */
+#include "ruby/io.h"
+
+#include "vm_core.h"
```
ko1, line 18: maybe it is for `rb_execution_context_t`, but we can use `struct rb_execution_context_struct *` and we don't need to include `vm_core.h`.

thread.c
```diff
@@ -2639,76 +2636,72 @@ rb_ec_reset_raised(rb_execution_context_t *ec)
...
+    rb_execution_context_t *ec = GET_EC();
+    io->closing_ec = ec;
+
+    // If there are no blocking operations, we are done:
+    if (ccan_list_empty(&io->blocking_operations)) return 0;
+
+    // This is used to ensure the correct execution context is woken up after the blocking operation is interrupted:
+    io->wakeup_mutex = rb_mutex_new();
+
+    size_t count = (size_t)rb_mutex_synchronize(io->wakeup_mutex, thread_io_close_notify_all, (VALUE)io);
```
ko1, line 2680: Mutex is needed?
At a glance, `thread_io_close_notify_all` seems to have no thread switching points.

thread.c
```diff
@@ -2639,76 +2636,72 @@ rb_ec_reset_raised(rb_execution_context_t *ec)
...
 
 void
 rb_thread_fd_close(int fd)
 {
-    struct rb_io_close_wait_list busy;
-
-    if (rb_notify_fd_close(fd, &busy)) {
-        rb_notify_fd_close_wait(&busy);
-    }
+    /* No-op */
```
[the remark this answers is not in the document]
ioquatix: Thanks for your investigation.

From your linked example, is it safe to call the current implementation of `rb_thread_fd_close` from within GC?

I think a better solution is to take a reference to an auto close IO instance. It is an easy fix and then the code can be updated to use `rb_io_wait` etc.

As an interim solution, what about issuing a warning and adding documentation about how to fix it?

thread.c
ioquatix: See https://github.com/ruby/ruby/pull/7884 for why it was introduced. It is also used for compatibility with the fiber scheduler, but maybe we can re-evaluate and consider removal.

internal/io.h
ioquatix: Agreed.

thread.c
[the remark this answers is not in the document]
ko1: Simplified code (as I understand):

```ruby
io1 = open(...)
io2 = IO.for_fd(io1.fileno) # https://github.com/vitoshalabs/ruby-pcap/blob/master/ext/pcap/Pcap.c#L455C94-L455C95

# maybe io2 is used
io2.read...

rb_notify_fd_close_wait(io1.fileno) # wait for closing interrupting io2
```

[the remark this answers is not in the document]
ioquatix: Why don’t they just use `rb_io_close`?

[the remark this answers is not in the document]
ioquatix: Okay, I considered the use case. If we want to support it, we can expose `rb_thread_io_close(VALUE io)` which does the same thing but per IO instance. However, I think the name should be considered, as the current name and `rb_thread_io_close` sound like they actually close the file descriptor, which is not true.

ioquatix:
@ko1 requested the following changes:

- [ ] Fix `common.mk` dependencies.
- [ ] Remove verbose warning details.
- [ ] Update NEWS.
- [ ] Remove mutex synchronisation around `thread_io_close_notify_all`.

thread.c
ioquatix: Okay, I've replaced the initial mutex usage with `RUBY_ASSERT_CRITICAL_SECTION_ENTER`/`RUBY_ASSERT_CRITICAL_SECTION_LEAVE` so we don't accidentally break this code.

ntkme:
@ioquatix I think this change is causing segmentation fault in my nightly CI run against ruby head:

https://github.com/sass-contrib/sass-embedded-host-ruby/actions/runs/15010437636/job/42178042427

To summarize the use case: sass-embedded spawn the dart-sass child process and communicate with child process’ stdio. When parent ruby process is forked, the stdio file descriptors of the child process is duplicated. However, because this gem does not support share the child process across multiple parent process, I need to close the stdio file descriptors in the forked ruby process. Right now it crashes when closing the duplicated file descriptors right after fork in the forked ruby process.

ioquatix:
Thanks for your report, I will take a look.

ioquatix:
I was able to reproduce the issue, I'll work on a fix.

ioquatix:
Okay, I found the root cause of the issue.

ioquatix:
Here is the reproduction:

```ruby
r, w = IO.pipe

thread = Thread.new do
  r.read
end

Thread.pass until thread.status == "sleep"

fork do
  r.close
end

w.close

Process.wait
thread.join
```

ioquatix:
Proposed fix: https://github.com/ruby/ruby/pull/13343

PR #13129, opened 2025-04-18
[DOC] Mention VS Code debugging setup in the debugging guide

st0012:
VS Code debugging is usually more friendly to new contributors so I think it's worth mentioning that Ruby already provides a template for setting it up (created by @k0kubun)

Screenshot 2025-04-18 at 13 37 30


k0kubun (approved):
:+1:

PR #13132, opened 2025-04-18
Don't emit stack trace for internal iseqs

zzak:


zzak:
Closing in favor of #13238

PR #13139, opened 2025-04-19
ZJIT: Add option to dump LIR

tekknolagi:
Now we can dump HIR, optimized HIR, LIR, and assembly.

```
plum% ../build-dev/miniruby --zjit-dump-hir --zjit-dump-lir /tmp/defint.rb
HIR:
fn foo:
bb0():
  v1:Fixnum[5] = Const Value(5)
  Return v1

LIR:
fn foo:
Assembler
    000 Label() -> None
    001 FrameSetup() -> None
    002 Add(A64Reg { num_bits: 64, reg_no: 19 }, 38_u64) -> Out64(0)
    003 Mov(A64Reg { num_bits: 64, reg_no: 19 }, Out64(0)) -> None
    004 Mov(Mem64[Reg(20) + 16], A64Reg { num_bits: 64, reg_no: 19 }) -> None
    005 FrameTeardown() -> None
    006 CRet(Value(VALUE(11))) -> None
plum%
```

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -10,6 +10,7 @@ use crate::backend::lir::{self, asm_comment, Assembler, Opnd, Target, CFP, C_ARG
 use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, CallInfo};
 use crate::hir::{Const, FrameState, Function, Insn, InsnId};
 use crate::hir_type::{types::Fixnum, Type};
+use crate::options::get_option;
 
 /// Ephemeral code generation state
 struct JITState {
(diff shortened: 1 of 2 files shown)


tekknolagi:
oh huh i just realized it didn't add @ruby/yjit to review it

k0kubun:
Can you use `ZJIT:` instead of `[zjit]` as the prefix for PR and commit messages? We use `[xxx]` for `[DOC]` and changes copied from other repositories, but we use `xxx:` for Ruby core and YJIT changes. So I expected we would use `ZJIT:`.

tekknolagi:
Sure -- I don't think I manually used `[zjit]` -- I think some tool added it

k0kubun (approved):

PR #13141, opened 2025-04-20
Adds a couple of specs

fxn:
In a class definition like this:

```ruby
class D < C
end
```

there might be a `const_added` hook, and also an `inherited` hook being triggered when `D` is just created.

Those are user-level hooks, and the following is observable:

1) When they are invoked, the class object stored in `D` has been created and it has the expected superclass.
2) When they are invoked, the constant has been created and constant-related state reflects that.

The specs proposed in this patch specify this contract.

(What I described above is the consensus achieved in discussions in https://bugs.ruby-lang.org/issues/21143, https://bugs.ruby-lang.org/issues/21193, and related PRs.)

/cc @byroot @eregon

### DIFF
--- a/spec/ruby/core/class/inherited_spec.rb
+++ b/spec/ruby/core/class/inherited_spec.rb
@@ -98,4 +98,21 @@ describe "Class.inherited" do
     -> { Class.new(top) }.should_not raise_error
   end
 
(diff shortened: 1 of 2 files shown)


byroot (approved):
LGTM. Will wait in case @eregon has any objections.

PR #13142, opened 2025-04-20
Fix jump buffer leak in WASI builds

white-axe:
This pull request fixes a memory leak in WASI (WebAssembly) builds of Ruby caused by the VM jump buffers not being freed in certain parts of the code.

I originally attempted to fix this memory leak in #12995 (it's the second memory leak mentioned in that pull request), but the fix was reverted by #13026 due to test failures. The fix has been rewritten to be much simpler and less invasive so that it has less of a chance of breaking things.

I've already tested building with this patch here: https://github.com/white-axe/ruby.wasm/actions/runs/14563460222/job/40849624514

For ease of reference, I'm mentioning again here that there's [a test](https://github.com/ruby/ruby/pull/12995#issuecomment-2764643641) to determine if a WASI build of Ruby is affected by the two memory leaks mentioned in #12995:

> Here's a test to see if a WASI build is affected by these two memory leaks. This script causes all of the current WASI builds from https://github.com/ruby/ruby.wasm to rapidly consume around 800 mebibytes of memory due to the jump buffer leak and then crash from what I believe is stack buffer overflow caused by the stack pointer leak:
> ```ruby
> loop do
>   array = [0]
>   for item in array
>     break
>   end
> end
> ```
> To test it out, save it as test.rb in the current working directory, and then:
> ```
> curl -Lo ruby.tar.gz https://github.com/ruby/ruby.wasm/releases/download/2.7.1/ruby-3.4-wasm32-unknown-wasip1-full.tar.gz
> tar xzf ruby.tar.gz --strip-components=1
> wasmtime --dir .::/ usr/local/bin/ruby test.rb
> ```

Note: The first memory leak has been fixed in the master branch so it won't immediately crash from stack buffer overflow anymore, but the test still crashes from running out of memory due to the second memory leak.

Requesting a review from @kateinoigakukun.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -1511,6 +1511,51 @@ cont_restore_thread(rb_context_t *cont)
             rb_raise(rb_eRuntimeError, "can't call across trace_func");
         }
 
+#if defined(__wasm__) && !defined(__EMSCRIPTEN__)
+        if (th->ec->tag != sec->tag) {
+            /* find the lowest common ancestor tag of the current EC and the saved EC */
+
+            struct rb_vm_tag *lowest_common_ancestor = NULL;
+            size_t num_tags = 0;
+            size_t num_saved_tags = 0;
+            for (struct rb_vm_tag *tag = th->ec->tag; tag != NULL; tag = tag->prev) {
+                ++num_tags;
+            }
+            for (struct rb_vm_tag *tag = sec->tag; tag != NULL; tag = tag->prev) {
+                ++num_saved_tags;
+            }
+
+            size_t min_tags = num_tags <= num_saved_tags ? num_tags : num_saved_tags;
+
+            struct rb_vm_tag *tag = th->ec->tag;
+            while (num_tags > min_tags) {
+                tag = tag->prev;
+                --num_tags;
+            }
+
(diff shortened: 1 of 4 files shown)


kateinoigakukun:
If I understand correctly, we still need to deinit jump buffers for intermediate frames as you did in the latest fix, shouldn't we?

white-axe:
Yeah, sorry, I forgot about one case where the longjmp caused by calling a continuation also leaks jump buffers, which I just fixed. Here's a test to see if continuations leak jump buffers:
```rb
require 'continuation'
callcc {|cc| $cc = cc}
array = [0]
for item in array
  $cc.call
end
```

Other than that, though, apparently all of the other longjmps in the Ruby code are made from inside of calls to `EC_JUMP_TAG()`, which just jumps to `GET_EC()->tag->buf`, i.e. it never skips over any tags, so no intermediate jump buffers are ever actually leaked by any of the longjmps other than the one in cont.c.

Jumping to a different fiber also doesn't leak jump buffers because the fiber's deinitialization will free the jump buffers that belong to the fiber, as far as I can tell.

kateinoigakukun (approved):
Sorry for taking a long time to review. It looks good to me, huge thanks!

PR #13151, opened 2025-04-22
Add missing lock in `rb_gc_impl_define_finalizer`

byroot:
`objspace->finalizer_table` must be synchronized,
otherwise concurrent insertion from multiple ractors will cause a crash.

Repro:

```ruby
ractors = 5.times.map do |i|
  Ractor.new do
    100_000.times.map do
      o = Object.new
      ObjectSpace.define_finalizer(o, ->(id) {})
      o
    end
  end
end

ractors.each(&:take)
```

FYI: @tenderlove @jhawthorn @etiennebarrie

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -2834,6 +2834,8 @@ rb_gc_impl_define_finalizer(void *objspace_ptr, VALUE obj, VALUE block)
 
     RBASIC(obj)->flags |= FL_FINALIZE;
 
+    int lev = rb_gc_vm_lock();
+
     if (st_lookup(finalizer_table, obj, &data)) {
         table = (VALUE)data;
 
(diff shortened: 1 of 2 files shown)


gc/default/default.c
```diff
@@ -2834,6 +2834,8 @@ rb_gc_impl_define_finalizer(void *objspace_ptr, VALUE obj, VALUE block)
 
     RBASIC(obj)->flags |= FL_FINALIZE;
 
+    int lev = rb_gc_vm_lock();
+
```
tekknolagi, line 2838: I can't add a line comment since you didn't modify that block, but do you have to unlock at 2849 before the early return?

gc/mmtk/mmtk.c
```diff
@@ -973,6 +973,8 @@ rb_gc_impl_define_finalizer(void *objspace_ptr, VALUE obj, VALUE block)
 
     RBASIC(obj)->flags |= FL_FINALIZE;
 
+    int lev = rb_gc_vm_lock();
```
tekknolagi, line 976: (as above with early return)

gc/default/default.c
byroot: Good catch 6d0dd7d86390458e02830684dd45830b5869a7aa

PR #13159, opened 2025-04-23
Move `object_id` in attributes

casperisfine:
And get rid of the `obj_to_id_tbl`

It's no longer needed, the `object_id` is now stored inline
in the object.

We still need the inverse table in case `_id2ref` is invoked, but
we lazily build it by walking the heap if that happens.


gc/default/default.c
```diff
@@ -1568,39 +1567,70 @@ rb_gc_impl_object_id(void *objspace_ptr, VALUE obj)
...
+            }
+            FL_SET_RAW(obj, FL_SEEN_OBJ_ID);
         }
+
+        rb_gc_vm_unlock(lock_lev);
+        return id;
+    }
+
+    if (rb_shape_has_object_id(shape)) {
+        // We could avoid locking if the object isn't shareable
```
casperisfine, line 1592: For `T_MODULE` and `T_CLASS` it might be worth finding `8B` in `rb_classext_t` so we can do some atomic CAS and find_or_create the object_id with no actual locking.

But if we need to also set the `FL_SEEN_OBJ_ID` flag, it might be more tricky than that?

gc.c
```diff
@@ -1213,9 +1207,15 @@ rb_data_free(void *objspace, VALUE obj)
     return true;
 }
 
+static void obj_free_object_id(VALUE obj);
+
 void
 rb_gc_obj_free_vm_weak_references(VALUE obj)
 {
+    if (FL_TEST_RAW(obj, FL_SEEN_OBJ_ID)) {
```
casperisfine, line 1215: `FL_SEEN_OBJ_ID` isn't very useful anymore, except for `SHAPE_TOO_COMPLEX`, but we could improve that too.

If we always check the shape flags, we could reclaim that flag.

Also more broadly, `shape_id_t` is `32bit` on 64bit systems, which is way more than we need. If needing to lookup the shape to know if it has an `object_id` (or is frozen) we could actually pack these tags in the most significant bits of the `shape_id`.

gc.c
```diff
@@ -1759,6 +1722,245 @@ rb_gc_pointer_to_heap_p(VALUE obj)
...
+
+static void gc_mark_tbl_no_pin(st_table *table);
+
+static void
+id_to_obj_tbl_mark(void *data)
+{
+    st_table *table = (st_table *)data;
+    if (UNLIKELY(!RB_POSFIXABLE(next_object_id))) {
+        // It's very unlikely, but if enough object ids were generated, keys may be T_BIGNUM
+        rb_mark_set(table);
```
byroot, line 1770: Even if you wrote a program to generate new object_ids in a loop, it would probably take decades to overflow. So perhaps we can just simplify all this.

Except perhaps on 32bit system?

gc.c
```diff
@@ -1759,6 +1722,245 @@ rb_gc_pointer_to_heap_p(VALUE obj)
...
+object_id(VALUE obj)
+{
+    VALUE id;
+    rb_shape_t *shape = rb_shape_get_shape(obj);
+    unsigned int lock_lev;
+
+    // We could avoid locking if the object isn't shareable
+    // but we'll lock anyway to lookup the next shape, and
+    // we'd at least need to generate the object_id using atomics.
+    lock_lev = rb_gc_vm_lock();
```
byroot, line 1834: Might be best to do in a followup, to it's easier to review, but we're not yet reducing contention at all here.

As for looking up the next shape, we can now do it without locking in the single-child happy path: https://github.com/ruby/ruby/pull/13191.

So combined with an atomic increment of `next_object_id` we could in many case not lock even on the initial write.

gc.c
```diff
@@ -1759,6 +1722,245 @@ rb_gc_pointer_to_heap_p(VALUE obj)
     return rb_gc_impl_pointer_to_heap_p(rb_gc_get_objspace(), (void *)obj);
 }
 
+#define OBJ_ID_INCREMENT (RUBY_IMMEDIATE_MASK + 1)
+#define OBJ_ID_INITIAL (OBJ_ID_INCREMENT)
+
+static unsigned long long next_object_id = OBJ_ID_INITIAL;
+static VALUE id_to_obj_value = 0;
+static st_table *id_to_obj_tbl = NULL;
```
byroot, line 1730: Perhaps we should move `object_id` related logic in its own file? Might be followup worthy.

gc.c
```diff
@@ -1759,6 +1722,245 @@ rb_gc_pointer_to_heap_p(VALUE obj)
...
+
+void
+rb_gc_obj_free_vm_weak_references(VALUE obj)
+{
+    obj_free_object_id(obj);
+
+    if (FL_TEST(obj, FL_EXIVAR)) {
+        rb_free_generic_ivar((VALUE)obj);
+        FL_UNSET(obj, FL_EXIVAR);
+    }
```
byroot, line 1936: I think it would be very easy to get rid of `FL_EXIVAR`, and replace it with something like:

```c
!RB_TYPE_ANY(obj, T_OBJECT, T_CLASS, T_MODULE) && get_shape(obj)->next_field_index > 0
```

It adds one extra pointer chase, but it could be optimized out because generic objects with no ivars will always have either the ROOT shape or the ROOT_FROZEN shape. So it could be a very quick operation.

variable.c
```diff
@@ -1478,62 +1507,102 @@ rb_attr_delete(VALUE obj, ID id)
...
+    rb_shape_t *shape = rb_shape_get_shape(obj);
+    if (rb_shape_has_object_id(shape)) {
+        rb_shape_t *object_id_shape = rb_shape_object_id_shape(obj);
+        VALUE id = rb_field_get(obj, object_id_shape);
+        st_insert(table, internal_object_id, id);
+
+        // We need to ensure the object ID is registered in id_to_obj_table
+        // before transitioning to too complex
+        rb_gc_update_id_to_obj_table(obj, id);
+    }
```
casperisfine, line 1602: This part I don't get.

It was added by @eightbitraptor in https://github.com/byroot/ruby/commit/1023700f019. To me it sounds like useless code because `rb_obj_copy_fields_to_hash_table` should already copy the `object_id` in the `st_table`.

But if I remove this code CI immediately fail, particularly with mmtk, so I assume there's some sort of race condition here.

casperisfine: We investigated this.

What seem to be going on:

  - A `T_DATA` or `T_FILE` enter `rb_gc_obj_free_vm_weak_references`.
  - Its shape indicate it has an `object_id`, but when accessing it, we get garbage.
  - That cause the `object_id_hash` function to crash, because it expect either `T_FIXNUM` or `T_BIGNUM` and what we get is neither (we observed a `T_FILE`).

We suspect `rb_gc_obj_free_vm_weak_references` is sometimes called twice on some objects. Before it wasn't a problem because `rb_gc_obj_free_vm_weak_references` was idempotent, but with this PR it no longer is.

Question:

  - Is it actually a contract that it must be idempotent? Or just an happy accident?
  - Is it expected that an object may be freed twice?

casperisfine: I figured that one out: 61fb42169693050a8a42bbff32d2ac8b89a1355f

What was happening is I made the mistake of declaring `internal_object_id` as a `static` in `shape.h`, so it was only initialized on `shape.c`, everywhere else it was either `0` or garbage. So this code was preventing the bug from trigerring by aliasing the `object_id` key with whatever the garbage ID was.

gc.c
peterzhu2118: A long long is guaranteed to be at least 64 bits in size, so probably won't overflow on 32 bit system either.

casperisfine: Sorry, overflow isn't what I meant. I mean not fit in a `T_FIXNUM`.

gc.c
```diff
@@ -3458,10 +3657,77 @@ vm_weak_table_gen_ivar_foreach_too_complex_replace_i(st_data_t *_key, st_data_t
...
+        GC_ASSERT(rb_shape_obj_has_id((VALUE)value));
+        return ST_DELETE;
+
+      case ST_REPLACE: {
+        VALUE new_value = (VALUE)value;
+        ret = iter_data->update_callback(&new_value, iter_data->data);
+        if (value != new_value) {
+            DURING_GC_COULD_MALLOC_REGION_START();
+            {
+                st_insert(id_to_obj_tbl, key, (st_data_t)new_value);
```
peterzhu2118, line 3683: Why do we need to do an `st_insert` here? If we only change the value, then we can just use `st_foreach_with_replace` and directly update the value.

gc.c
```diff
@@ -3587,12 +3853,32 @@ rb_gc_vm_weak_table_foreach(vm_table_foreach_callback_func callback,
...
+      case RB_GC_VM_ID_TO_OBJ_TABLE_KEYS: {
+        if (id_to_obj_tbl && !RB_POSFIXABLE(next_object_id)) {
+            st_foreach(
+                id_to_obj_tbl,
+                vm_weak_table_id_to_obj_keys_foreach,
+                (st_data_t)&foreach_data
+            );
+        }
+        break;
+      }
```
peterzhu2118, line 3875: I think we can merge these two cases together and only pass the key if `RB_POSFIXABLE(next_object_id)`.

gc.c
casperisfine: It's indeed a missed opportunity, I'll see about changing that.

gc.c
casperisfine: Good point.

shape.h
```diff
@@ -33,21 +33,24 @@ typedef uint32_t redblack_id_t;
...
     ID edge_name; // ID (ivar) for transition from parent to rb_shape
-    attr_index_t next_iv_index;
+    attr_index_t next_field_index; // Fields are either ivars or internal properties like `object_id`
     attr_index_t capacity; // Total capacity of the object with this shape
     uint8_t type;
     uint8_t heap_index;
+    uint8_t flags;
     shape_id_t parent_id;
     redblack_node_t *ancestor_index;
 };
```
tenderlove, line 56: What is the size of rb_shape now?  We might want to move `flags` to the end of the struct because I think this will leave wasted space between `flags` and `parent_id`.

On master it's 40 bytes:

```
(lldb) p sizeof(struct rb_shape)
(unsigned long) 40
```

I'm not sure how much we care about shape memory usage, but maybe we could split the type field. We might only need 4 bits for `type`.

tenderlove (approved):
This is fine with me. I think we might have an opportunity to decrease the size of `rb_shape`, but I'm fine if we merge this as-is.

shape.h
casperisfine: It's unchanged, I've put the flags here because there was some padding:

```diff
diff --git a/shape.c b/shape.c
index fb739d3614..8aec985496 100644
--- a/shape.c
+++ b/shape.c
@@ -1434,6 +1434,7 @@ Init_default_shapes(void)
 void
 Init_shape(void)
 {
+    fprintf(stderr, "sizeof(rb_shape_t) = %lu\n", sizeof(rb_shape_t));
 #if SHAPE_DEBUG
     /* Document-class: RubyVM::Shape
      * :nodoc: */
```

```
$ make -j run
sizeof(rb_shape_t) = 40
```

> I'm not sure how much we care about shape memory usage

Not that much, but I think we could save some spaces by not using a full pointer for `redblack_node_t *ancestor_index;`. We could just have an offset. I think It would be nice if shapes were 32B instead of 40B.

peterzhu2118 (approved):
Makes sense to me!

PR #13164, opened 2025-04-24
Use `cgi/util` instead of `cgi`

hsbt:
For https://bugs.ruby-lang.org/issues/21258

I have a plan to retire `cgi` library in the future. The libraries should use `cgi/util` instead of `cgi` if they used only like `CGI.escape`.

### DIFF
--- a/lib/rubygems/gemcutter_utilities/webauthn_listener.rb
+++ b/lib/rubygems/gemcutter_utilities/webauthn_listener.rb
@@ -85,10 +85,17 @@ module Gem::GemcutterUtilities
     end
 
     def parse_otp_from_uri(uri)
-      require "cgi"
+      query = uri.query
+      return unless query && !query.empty?
 
-      return if uri.query.nil?
-      CGI.parse(uri.query).dig("code", 0)
+      query.split('&') do |param|
+        key, value = param.split('=', 2)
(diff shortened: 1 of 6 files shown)


lib/rubygems/gemcutter_utilities/webauthn_listener.rb
```diff
@@ -85,10 +85,17 @@ def root_path?(uri)
     end
 
     def parse_otp_from_uri(uri)
-      require "cgi"
+      query = uri.query
+      return unless query && !query.empty?
 
-      return if uri.query.nil?
-      CGI.parse(uri.query).dig("code", 0)
+      query.split('&').each do |param|
```
nobu, line 91: `.each` is not needed.
```suggestion
      query.split('&') do |param|
```

lib/rubygems/gemcutter_utilities/webauthn_listener.rb
```diff
@@ -85,10 +85,17 @@ def root_path?(uri)
...
+        key, value = param.split('=', 2)
+        if key && value && Gem::URI.decode_www_form_component(key) == "code"
```
nobu, line 93: If `value` is non-nil, `key` should be too.

```suggestion
        if value && Gem::URI.decode_www_form_component(key) == "code"
```

PR #13165, opened 2025-04-24
Enabled to cleanup codeql-trap cache

hsbt:
codeql-trap cache is growing over 300MB. We should enable to cleanup needless cache from actions/cache.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -142,7 +142,7 @@ jobs:
 
       - name: Install libraries with vcpkg
         run: |
-          vcpkg install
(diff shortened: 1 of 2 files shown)


.github/workflows/windows.yml
```diff
@@ -142,7 +142,7 @@ jobs:
 
       - name: Install libraries with vcpkg
         run: |
-          vcpkg install
+          vcpkg install --vcpkg-root=C:\Users\runneradmin\scoop\apps\vcpkg\current
```
nobu, line 145: It can be `--vcpkg-root=%HOME%\scoop\apps\vcpkg\current` or `--vcpkg-root=%USERPROFILE%\scoop\apps\vcpkg\current`.

PR #13168, opened 2025-04-24
objspace_dump: Include `shareable` flag

byroot:
Given that the currently planned ractor local GC implementation performance will heavilly be influenced by the number of shareable objects it would be valuable to be able to know how many of them are in the heap.


byroot:
Merged as cb1ea54bbf6cdf49c53f42720fec1a151069810c to bypass the flaky CI.

PR #13175, opened 2025-04-25
Disabled built-in cache feature of vcpkg for Windows Arm build

hsbt:
It may conflict `x64-windows` build.


hsbt:
I'm working https://github.com/ruby/ruby/pull/13176 instead of this.

PR #13180, opened 2025-04-25
CODEQL_ACTION_CLEANUP_TRAP_CACHES needs `actions:write` permission

hsbt:
For reducing trap cache of CodeQL

```
Warning: Could not cleanup TRAP caches as the token did not have the required permissions. To clean up TRAP caches, ensure the token has the "actions:write" permission. See https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs for more information.
```


hsbt:
This configuration is not required now.

PR #13182, opened 2025-04-25
ZJIT: Bail out of HIR translation if we can't handle a send flag

tekknolagi:


zjit/src/hir.rs
```diff
@@ -2610,6 +2655,90 @@ mod tests {
...
+
+    #[test]
+    fn test_cant_compile_super_forward() {
+        eval("
+            def test(...) = super(...)
+        ");
+        assert_compile_fails("test", ParseError::UnknownOpcode("invokesuperforward".into()))
+    }
+
+    // TODO(max): Figure out how to generate a call with OPT_SEND flag
```
XrXr, line 2717: This should be impossible. The flag is only added for calls like `.send` during processing, and no call info should start with this flag. So since ZJIT doesn't add this flag we should never see it.

tekknolagi: Oh, it never makes it into the bytecode?

tekknolagi: Incredible, thanks

PR #13183, opened 2025-04-25
ZJIT: Replace GetConstantPath with Const if the IC is not empty

tekknolagi:
- **Add Class, Module HIR types**
- **Add rb_zjit_constcache_shareable**
- **Add rb_zjit_multi_ractor_p**
- **Replace GetConstantPath with Const if the IC is not empty**


tekknolagi:
This is based on / inspired by the YJIT impl:

```rust
fn gen_opt_getconstant_path(
    jit: &mut JITState,
    asm: &mut Assembler,
) -> Option<CodegenStatus> {
    let const_cache_as_value = jit.get_arg(0);
    let ic: *const iseq_inline_constant_cache = const_cache_as_value.as_ptr();
    let idlist: *const ID = unsafe { (*ic).segments };

    // Make sure there is an exit for this block as the interpreter might want
    // to invalidate this block from yjit_constant_ic_update().
    jit_ensure_block_entry_exit(jit, asm)?;

    // See vm_ic_hit_p(). The same conditions are checked in yjit_constant_ic_update().
    // If a cache is not filled, fallback to the general C call.
    let ice = unsafe { (*ic).entry };
    if ice.is_null() {
        // Prepare for const_missing

        // If this does not trigger const_missing, vm_ic_update will invalidate this block.
        extern "C" {
            fn rb_vm_opt_getconstant_path(ec: EcPtr, cfp: CfpPtr, ic: *const u8) -> VALUE;
        }
        let val = asm.ccall(
            rb_vm_opt_getconstant_path as *const u8,
            vec![EC, CFP, Opnd::const_ptr(ic as *const u8)],
        );

        let stack_top = asm.stack_push(Type::Unknown);
        asm.store(stack_top, val);

        return jump_to_next_insn(jit, asm);
    }

    let cref_sensitive = !unsafe { (*ice).ic_cref }.is_null();
    let is_shareable = unsafe { rb_yjit_constcache_shareable(ice) };
    let needs_checks = cref_sensitive || (!is_shareable && !assume_single_ractor_mode(jit, asm));

    if needs_checks {
        // Cache is keyed on a certain lexical scope. Use the interpreter's cache.
        let inline_cache = asm.load(Opnd::const_ptr(ic as *const u8));

        // Call function to verify the cache. It doesn't allocate or call methods.
        // This includes a check for Ractor safety
        let ret_val = asm.ccall(
            rb_vm_ic_hit_p as *const u8,
            vec![inline_cache, Opnd::mem(64, CFP, RUBY_OFFSET_CFP_EP)]
        );

        // Check the result. SysV only specifies one byte for _Bool return values,
        // so it's important we only check one bit to ignore the higher bits in the register.
        asm.test(ret_val, 1.into());
        asm.jz(Target::side_exit(Counter::opt_getconstant_path_ic_miss));

        let inline_cache = asm.load(Opnd::const_ptr(ic as *const u8));

        let ic_entry = asm.load(Opnd::mem(
            64,
            inline_cache,
            RUBY_OFFSET_IC_ENTRY
        ));

        let ic_entry_val = asm.load(Opnd::mem(
            64,
            ic_entry,
            RUBY_OFFSET_ICE_VALUE
        ));

        // Push ic->entry->value
        let stack_top = asm.stack_push(Type::Unknown);
        asm.store(stack_top, ic_entry_val);
    } else {
        // Invalidate output code on any constant writes associated with
        // constants referenced within the current block.
        jit.assume_stable_constant_names(asm, idlist);

        jit_putobject(asm, unsafe { (*ice).value });
    }

    jump_to_next_insn(jit, asm)
}
```

k0kubun (approved):

PR #13186, opened 2025-04-27
YJIT: Fix potential infinite loop when OOM

rianmcguire:
Fixes https://bugs.ruby-lang.org/issues/21257 [Bug #21257]

Avoid generating an infinite loop in the case where:
1. Block `first` is adjacent to block `second`, and the branch from `first` to `second` is a fallthrough, and
2. Block `second` immediately exits to the interpreter, and
3. Block `second` is invalidated and YJIT is OOM

While pondering how to fix this, I think I've stumbled on another related edge case:
1. Block `incoming_one` and `incoming_two` both branch to block `second`. Block `incoming_one` has a fallthrough
2. Block `second` immediately exits to the interpreter (so it starts with its exit)
3. When Block `second` is invalidated, the incoming fallthrough branch from `incoming_one` might be rewritten first, which overwrites the start of block `second` with a jump to a new branch stub.
4. YJIT runs of out memory
5. The incoming branch from `incoming_two` is then rewritten, but because we're OOM we can't generate a new stub, so we use `second`'s exit as the branch target. However `second`'s exit was already overwritten with a jump to the branch stub for `incoming_one`, so `incoming_two` will end up jumping to `incoming_one`'s branch stub.

I'm not sure what the consequences of calling the wrong branch stub are, but the comments in invalidate_block_version suggest the stubs are supposed to be unique.

I've attempted to address both of these cases with this change, but I can split it up if that would be preferable.

### DIFF
--- a/yjit/src/core.rs
+++ b/yjit/src/core.rs
@@ -4158,7 +4158,19 @@ pub fn invalidate_block_version(blockref: &BlockRef) {
     }
 
     // For each incoming branch
-    let incoming_branches = block.incoming.0.take();
+    let mut incoming_branches = block.incoming.0.take();
+
+    // An adjacent branch will write into the start of the block being invalidated, possibly
+    // overwriting the block's exit. If we run out of memory after doing this, any subsequent
+    // incoming branches we rewrite won't be able use the block's exit as a fallback when they
+    // are unable to generate a stub. To avoid this, if there's an incoming branch that's
+    // adjacent to the invalidated block, make sure we process it last.
+    incoming_branches.sort_by_key(|branchref| {
+        let branch = unsafe { branchref.as_ref() };
+        let target_next = block.start_addr == branch.end_addr.get();
+        target_next
+    });
+
     for (i, branchref) in incoming_branches.iter().enumerate() {
         let branch = unsafe { branchref.as_ref() };
         let target_idx = if branch.get_target_address(0) == Some(block_start) {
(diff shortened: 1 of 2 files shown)


yjit/src/core.rs
```diff
@@ -4158,7 +4158,20 @@ pub fn invalidate_block_version(blockref: &BlockRef) {
...
+    // An adjacent branch will write into the start of the block being invalidated, possibly
+    // overwriting the block's exit. If we run out of memory after doing this, any subsequent
+    // incoming branches we rewrite won't be able use the block's exit as a fallback when they
+    // are unable to generate a stub. To avoid this, if there's an incoming branch that's
+    // adjacent to the invalidated block, make sure we process it last.
+    incoming_branches.sort_by_key(|branchref| {
+        let branch = unsafe { branchref.as_ref() };
+        let target_next = block.start_addr == branch.end_addr.get();
+        target_next
+    });
```
rianmcguire, line 4172: This feels wrong.

Rather than sorting, I was originally hoping to do two passes over the list of branches, ignoring the adjacent branch on the first pass.

But I also didn't want to duplicate the code inside the loop, and everything I tried for reuse (inner fn, closure) felt uglier than sorting.

rianmcguire: Now that I've written that, finding the index of the adjacent branch (if any) and swapping it to the end of the slice would be more efficient than sorting. Let me try that...

XrXr:
Thank you so much for the investigation and the thorough fix! 🙏🏼 
(and sorry for not getting back to you quicker!)

PR #13190, opened 2025-04-28
Save one VALUE per embedded RTypedData

jeremyevans:
This halves the amount of memory used for embedded RTypedData if they
are one VALUE (8 bytes on 64-bit platforms) over the slot size limit.

For Set, on 64-bit it uses an embedded 56-byte struct.  With the
previous implementation, the embedded structs starts at offset 32,
resulting in a total size of 88.  Since that is over the 80 byte
limit, it goes to the next highest bucket, 160 bytes, wasting 72
bytes.  This allows it to fit in a 80 byte bucket, which reduces
the total size for small sets of from 224 bytes (160 bytes
embedded, 64 bytes malloc, 72 bytes wasted in embedding) to 144
bytes (80 bytes embedded, 64 bytes malloc, 0 bytes wasted in
embedding).

Any other embedded RTypedData will see similar advantages if they
are currently one VALUE over the limit.

To implement this, remove the typed_flag from struct RTypedData.
Embedded the typed_flag information in the type member, which is
now a tagged pointer using VALUE type, using the bottom low 2 bits
as flags (1 bit for typed flag, the other for the embedded flag).
To get the actual pointer, RTYPEDDATA_TYPE masks out
the low 2 bits and then casts.  That moves the RTypedData data
pointer from offset 32 to offset 24 (on 64-bit).

Vast amount of code in the internals (and probably external C
extensions) expects the following code to work for both RData and
non-embedded RTypedData:

```c
DATA_PTR(obj) = some_pointer;
```

Allow this to work by moving the data pointer in RData between
the dmark and dfree pointers, so it is at the same offset (24
on 64-bit).

Other than these changes to the include files, the only changes
needed were to gc.c, to account for the new struct layouts,
handle setting the low bits in the type member, and to use
RTYPEDDATA_TYPE(obj) instead of RTYPEDDATA(obj)->type.

The first commit fixes a non-deterministic failure in the gc tests
that was exposed by these changes.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1064,7 +1064,7 @@ rb_data_object_wrap(VALUE klass, void *datap, RUBY_DATA_FUNC dmark, RUBY_DATA_FU
 {
     RUBY_ASSERT_ALWAYS(dfree != (RUBY_DATA_FUNC)1);
     if (klass) rb_data_object_check(klass);
-    return newobj_of(GET_RACTOR(), klass, T_DATA, (VALUE)dmark, (VALUE)dfree, (VALUE)datap, !dmark, sizeof(struct RTypedData));
+    return newobj_of(GET_RACTOR(), klass, T_DATA, (VALUE)dmark, (VALUE)datap, (VALUE)dfree, !dmark, sizeof(struct RTypedData));
 }
 
 VALUE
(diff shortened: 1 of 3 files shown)


byroot (approved):
That's clever!

Looks good too to me, but I'd wait for Peter's expertise here.

test/ruby/test_gc.rb
```diff
@@ -411,6 +411,8 @@ def test_latest_gc_info_weak_references_count
       before_weak_references_count = GC.latest_gc_info(:weak_references_count)
       before_retained_weak_references_count = GC.latest_gc_info(:retained_weak_references_count)
 
+      # Clear ary, so if ary itself is somewhere on the stack, it won't hold all references
+      ary.clear
```
byroot, line 415: We can probably merge that fix now, Koichi reported it just this morning.

jeremyevans:
Rebased against master as the initial first commit was already merged.

peterzhu2118 (commented):
I think this should work. However, I'm a bit worried about users of `TypedData_Make_Struct` that wrap a non-allocated piece of memory or `char *` that is in the middle of a string that does not have the bottom address bit as zero.

peterzhu2118 (approved):
Ah yeah I misread, this makes sense. I think this is good 👍

nobu (approved):

PR #13191, opened 2025-04-28
get_next_shape_internal: Skip VM lock for single child case

casperisfine:
If the shape has only one child, we check it lock-free without compromising thread safety.

I haven't computed hard data as to how often that it the case, but we can assume that it's not too rare for shapes to have a single child that is often requested, typically when freezing and object.

### DIFF
--- a/shape.c
+++ b/shape.c
@@ -499,13 +499,26 @@ get_next_shape_internal(rb_shape_t *shape, ID id, enum shape_type shape_type, bo
 
     *variation_created = false;
 
+    // Fast path: if the shape has a single child, we can check it without a lock
+    struct rb_id_table *edges = RUBY_ATOMIC_PTR_LOAD(shape->edges);
+    if (edges && SINGLE_CHILD_P(edges)) {
+        rb_shape_t *child = SINGLE_CHILD(edges);
+        if (child->edge_name == id) {
+            return child;
+        }
+    }
+
     RB_VM_LOCK_ENTER();
     {
+        // The situation may have changed while we waited for the lock.
+        // So we load the edge again.
+        edges = RUBY_ATOMIC_PTR_LOAD(shape->edges);
+
         // If the current shape has children
-        if (shape->edges) {
+        if (edges) {
             // Check if it only has one child
-            if (SINGLE_CHILD_P(shape->edges)) {
-                rb_shape_t *child = SINGLE_CHILD(shape->edges);
+            if (SINGLE_CHILD_P(edges)) {
+                rb_shape_t *child = SINGLE_CHILD(edges);
                 // If the one child has a matching edge name, then great,
                 // we found what we want.
                 if (child->edge_name == id) {
@@ -515,7 +528,7 @@ get_next_shape_internal(rb_shape_t *shape, ID id, enum shape_type shape_type, bo
             else {
                 // If it has more than one child, do a hash lookup to find it.
                 VALUE lookup_result;
-                if (rb_id_table_lookup(shape->edges, id, &lookup_result)) {
+                if (rb_id_table_lookup(edges, id, &lookup_result)) {
                     res = (rb_shape_t *)lookup_result;
                 }
             }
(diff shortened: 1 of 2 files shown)


casperisfine:
> I haven't computed hard data as to how often that it the case

I did just that and it seems my assumption holds pretty well. From a big monolith:

```
1 children: 35949 shapes
2 children: 1661 shapes
3 children: 438 shapes
4 children: 159 shapes
5 children: 80 shapes
6 children: 60 shapes
7 children: 39 shapes
8 children: 28 shapes
9 children: 18 shapes
10 children: 15 shapes
11 children: 15 shapes
12 children: 10 shapes
```

Overall 72.8% of shapes have a single child.

shape.c
```diff
@@ -499,13 +499,23 @@ get_next_shape_internal(rb_shape_t * shape, ID id, enum shape_type shape_type, b
 
     *variation_created = false;
 
+    // Fast path: if the shape has a single child, we can check it without a lock
+    struct rb_id_table *edges = shape->edges;
```
jhawthorn, line 503: I think we should use atomics here and below where we write `shape->edges`. Otherwise the compiler could rearrange the initialization of the `shape` until after the pointer is written and on some platforms the hardware could do the same memory reordering.

There is definitely elsewhere in Ruby we're doing the same thing (ex. inline caches) and _apparently_ not seeing issues, but maybe that is due to insufficient testing on platforms with weaker memory models (arm), and luck that the compiler doesn't choose to rearrange too much/the hardware isn't running "concurrently enough" to hit the issue 🤷‍♂️.

It looks like we [have ATOMIC_PTR_LOAD](https://github.com/ruby/ruby/blob/58e3aa02240a9ec1b5fe6ce60d63828c2cf0c73a/include/ruby/atomic.h#L301-L302), but not ATOMIC_PTR_SET. I implemented [ATOMIC_VALUE_SET](https://github.com/ruby/ruby/commit/b28363a8381fd4e93d1254e71486e0185a84c9c5) here, which really should be the same thing. TBH I'm not sure why we have so much duplication in `ruby/atomic.h` where I think we know some of these types to be identical.

byroot: Thanks for the comment. I think I understand the concern. 

If I understand it correctly, this is mostly setting the pointer that need to use atomic, so that we guaranteed the shape content as fully been written.

I'm not sure I understand why we'd need an atomic load though. Presumably if we can see the pointer in `shape->edges`, the child shape content should be flushed, and we'd see it consistently too.

jhawthorn: The lazy answer is that the spec says we have to, otherwise it's UB and TSan will complain if we don't. This read of `shape->edges` could be combined with the one inside the lock (so we would see stale information after taking the lock). But I do think even aside from that it's still necessary to use an atomic.

I'm not 100% on this, but let me try for the less lazy answer 😅. My understanding of the memory model is that if we want writes from one thread to all be visible in another we need the writer to do an atomic operation with at least `memory_order_release` and then have it read with at least `memory_order_acquire` (all Ruby's atomics use seq_cst, which should just be stricter). The [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) link to [a wiki page with some examples](https://gcc.gnu.org/wiki/Atomic/GCCMM/AtomicSync).

On x86 I think that's true (all stores are equivalent to "release" and all loads are equivalent to "acquire"), but on ARM's weaker memory model [we end up with different instructions](https://godbolt.org/z/3qnMMhEov)

I tried making a program to show this happening but did not succeed 😅. So maybe there's something I don't understand about this or the cache lines didn't line up.

casperisfine:
@jhawthorn alright, I did what you suggested, hopefully this is correct now.

shape.c
```diff
@@ -499,13 +499,27 @@ get_next_shape_internal(rb_shape_t * shape, ID id, enum shape_type shape_type, b
 
     *variation_created = false;
 
+    // Fast path: if the shape has a single child, we can check it without a lock
+    struct rb_id_table *edges = RUBY_ATOMIC_PTR_LOAD(shape->edges);
+    if (edges && SINGLE_CHILD_P(edges)) {
```
casperisfine, line 504: I think in the case where `edges` is `NULL` we could also allocate the child atomically, and set it with CAS.

Not sure if it's really worth it though, because supposedly code running in ractors produce stable shapes, so write would become rare quickly.

jhawthorn: I agree we could do that, an issue I think that would create though is that if two Ractors are trying to insert the exact same shape, we would end up creating two shapes for it, for one Ractor the CAS would fail (as is normal with this pattern), but then there'd be no ability to free the shape it had created, since they are never freed (though I suppose we could create a freelist specifically for this case).

jhawthorn (approved):

shape.c
casperisfine: > (though I suppose we could create a freelist specifically for this case).

Yes, or just accept such case is rare enough that we can leak a few shapes.

PR #13192, opened 2025-04-28
Improve correctness for allocated object counts

etiennebarrie:
Currently the count of allocated object for a heap is incremented without regards to parallelism which leads to incorrect counts.

```ruby
5.times.map do
  Ractor.new do
    10_000_000.times { Object.new }
  end
end
  .each(&:take)

p GC.stat(:total_allocated_objects)
p GC.count
```

**Before**
```
17419220
2486
```

**After**
```
50008126
2074
```

By maintaining a local counter in the ractor newobj cache, and only syncing atomically with some granularity, we can improve the correctness without increasing contention.

The allocated object count is also synced when the ractor is freed.

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -161,6 +161,7 @@
 typedef struct ractor_newobj_heap_cache {
     struct free_slot *freelist;
     struct heap_page *using_page;
+    size_t allocated_objects_count;
 } rb_ractor_newobj_heap_cache_t;
 
 typedef struct ractor_newobj_cache {
@@ -2287,6 +2288,8 @@ rb_gc_impl_size_allocatable_p(size_t size)
     return size <= heap_slot_size(HEAP_COUNT - 1);
 }
 
+static const size_t ALLOCATED_COUNT_STEP = 1024;
+
 static inline VALUE
 ractor_cache_allocate_slot(rb_objspace_t *objspace, rb_ractor_newobj_cache_t *cache,
                            size_t heap_idx)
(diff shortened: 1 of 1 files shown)


ko1:
Thank you, but `rb_objspace_t` will be modified for Ractor-local GC soon so this proposal will not work soon.

byroot:
I'd like us to consider this anyway. Even if it no longer work once Ractor-local GC is merged, until then fixing Ractor bugs allow us to explore further and find more bugs and issues to fix.

gc/default/default.c
```diff
@@ -6261,6 +6274,15 @@ rb_gc_impl_ractor_cache_free(void *objspace_ptr, void *cache)
     rb_objspace_t *objspace = objspace_ptr;
 
     objspace->live_ractor_cache_count--;
+    rb_ractor_newobj_cache_t *newobj_cache = (rb_ractor_newobj_cache_t *)cache;
+
+    for (size_t heap_idx = 0; heap_idx < HEAP_COUNT; heap_idx++) {
+        rb_heap_t *heap = &heaps[heap_idx];
+        rb_ractor_newobj_heap_cache_t *heap_cache = &newobj_cache->heap_caches[heap_idx];
+        // TODO: Do we need atomics here? We're always inside VM_LOCK.
```
jhawthorn, line 6282: I think we probably do need the atomic. We're holding the VM lock, and we should assume also we have exclusive access to `newobj_cache` (because we're about to delete it 😅), but other threads not inside VM_LOCK could still be incrementing `heap->total_allocated_objects`

PR #13194, opened 2025-04-28
Update json gem

byroot:
Since there was a pretty big change with the introduction of SIMD code, I'm trying to check early if it cause any problem with Ruby rather than wait for Matzbot to discover it.


kateinoigakukun:
@byroot It looks like the problem is not specific for Wasm but `--with-static-linked-ext`.

tldr: Apply the following patch

```diff
diff --git a/ext/json/generator/extconf.rb b/ext/json/generator/extconf.rb
index 374f95a1af..e44890e2ed 100644
--- a/ext/json/generator/extconf.rb
+++ b/ext/json/generator/extconf.rb
@@ -35,7 +35,5 @@
     have_header('cpuid.h')
   end

-  create_header
-
   create_makefile 'json/ext/generator'
 end
```

----

With `static-linked-ext` enabled, `extmk.rb` adds a new `$defs` entry to avoid `Init_` symbol conflict for exts containing `/` in its feature name *during `create_makefile`*: https://github.com/ruby/ruby/blob/80a1a1bb8ae8435b916ae4f66a483e91ad31356a/ext/extmk.rb#L144

So generated `./ext/json/generator/extconf.h` is expected to look like:
```c
#ifndef EXTCONF_H
#define EXTCONF_H
#define JSON_GENERATOR 1
#define Init_generator Init_json_ext_generator
#define Init_generator Init_json_ext_generator
#define RUBY_EXPORT 1
#endif
```

And `ext/extinit.c` is generated to call such aliased Init symbol names.

However, the new `ext/json/generator/extconf.rb` calls `create_header` explicitly before `create_makefile`, so the aliasing hack was not applied, and it leads to inconsistent symbol name assumptions between extinit.c and the definition side.

byroot:
That worked, thank you very much!

samyron:
Removing that `create_header` will cause the `simd.h` to reference a file that does not exist.

```
../../../../../../ext/json/ext/generator/simd.h:1:10: fatal error: 'extconf.h' file not found
    1 | #include "extconf.h"
      |          ^~~~~~~~~~~
1 error generated.
gmake: *** [Makefile:251: generator.o] Error 1
```

If that include is removed, everything will compile, however SIMD will be excluded as it was relying on that header to define what is available.

samyron:
Yep, I believe that's correct. It seems like the `HAVE_*` definitions are passed via the command line and picked up appropriately.

After pulling master I see this:

```
scott@Scotts-MacBook-Air json % rake
mkdir -p tmp/arm64-darwin24/json/ext/generator/3.4.1/json/ext
cd tmp/arm64-darwin24/json/ext/generator/3.4.1
/Users/scott/.asdf/installs/ruby/3.4.1/bin/ruby -I. ../../../../../../ext/json/ext/generator/extconf.rb
checking for whether -std=c99 is accepted as CFLAGS... yes
checking for arm_neon.h... yes
checking for uint8x16_t in arm_neon.h... yes
checking for x86intrin.h... no
checking for cpuid.h... no
creating Makefile
...
```

kateinoigakukun:
I believe we don't need to include extconf.h as it's usually included  through ruby/c

PR #13196, opened 2025-04-28
ZJIT: Drop a duplicated call into optimize

k0kubun:
This PR removes a duplicated call into `optimize()`, which is a minor follow-up on https://github.com/Shopify/zjit/pull/109.

`compile_iseq()` does that, so no need to call it after that.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -170,11 +170,10 @@ fn gen_iseq(cb: &mut CodeBlock, iseq: IseqPtr) -> Option<(CodePtr, Vec<(Rc<Branc
     }
 
     // Convert ISEQ into High-level IR
-    let mut function = match compile_iseq(iseq) {
+    let function = match compile_iseq(iseq) {
         Some(function) => function,
         None => return None,
     };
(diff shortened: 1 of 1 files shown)


zjit/src/codegen.rs
```diff
@@ -170,11 +170,10 @@ fn gen_iseq(cb: &mut CodeBlock, iseq: IseqPtr) -> Option<(CodePtr, Vec<(Rc<Branc
     }
 
     // Convert ISEQ into High-level IR
```
tekknolagi, line 172:
```suggestion
    // Convert ISEQ into High-level IR and optimize HIR
```

tekknolagi (approved):

zjit/src/codegen.rs
tekknolagi: This comment is otherwise misleading

k0kubun: 👌 https://github.com/ruby/ruby/pull/13196/commits/8f56d47dd98815ff010c8c336e061345f46bfc6a

PR #13197, opened 2025-04-28
ZJIT: Handle ZJIT options properly

k0kubun:
This PR polishes the help and the interface to enable ZJIT.

* Hide `--zjit` from `--help` when `USE_ZJIT` is 0
* Show `--jit` for ZJIT-enabled build too
* Print user-facing ZJIT options in `--help`
* Enable ZJIT when `jit` feature is turned on and USE_ZJIT is 1
* Support `RUBY_ZJIT_ENABLE` environment variable like `RUBY_YJIT_ENABLE`

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -307,7 +307,9 @@ usage(const char *name, int help, int highlight, int columns)
 #define M(shortopt, longopt, desc) RUBY_OPT_MESSAGE(shortopt, longopt, desc)
 
 #if USE_YJIT
-# define PLATFORM_JIT_OPTION "--yjit"
+# define DEFAULT_JIT_OPTION "--yjit"
+#elif USE_ZJIT
+# define DEFAULT_JIT_OPTION "--zjit"
 #endif
 
     /* This message really ought to be max 23 lines.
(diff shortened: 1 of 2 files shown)


ruby.c
```diff
@@ -381,6 +385,9 @@ usage(const char *name, int help, int highlight, int columns)
         M("frozen-string-literal", "", "Freeze all string literals (default: disabled)."),
 #if USE_YJIT
         M("yjit",                  "", "In-process JIT compiler (default: disabled)."),
+#endif
```
tekknolagi, line 388: elif here too, as above?

ruby.c
```diff
@@ -338,13 +340,15 @@ usage(const char *name, int help, int highlight, int columns)
...
-#if USE_YJIT
-        M("--jit",         "",                     "Enable JIT for the platform; same as " PLATFORM_JIT_OPTION "."),
+#if USE_YJIT || USE_ZJIT
+        M("--jit",         "",                     "Enable the default JIT for the build; same as " DEFAULT_JIT_OPTION "."),
 #endif
 #if USE_YJIT
         M("--yjit",        "",                     "Enable in-process JIT compiler."),
 #endif
-        M("--zjit",        "",                     "Enable in-process JIT compiler."),
+#if USE_ZJIT
```
tekknolagi, line 349: elif here, as above?

ruby.c
```diff
@@ -419,6 +426,11 @@ usage(const char *name, int help, int highlight, int columns)
     printf("%s""YJIT options:%s\n", sb, se);
     rb_yjit_show_usage(help, highlight, w, columns);
 #endif
+#if USE_ZJIT
```
tekknolagi, line 429: elif

zjit/src/options.rs
```diff
@@ -155,6 +163,21 @@ fn update_profile_threshold(options: &Options) {
...
+/// `columns` is indent width of descriptions.
+#[unsafe(no_mangle)]
+pub extern "C" fn rb_zjit_show_usage(help: c_int, highlight: c_int, width: c_uint, columns: c_int) {
+    for &(name, description) in ZJIT_OPTIONS.iter() {
+        unsafe extern "C" {
+            fn ruby_show_usage_line(name: *const c_char, secondary: *const c_char, description: *const c_char,
+                                    help: c_int, highlight: c_int, width: c_uint, columns: c_int);
+        }
+        let name = CString::new(name).unwrap();
+        let description = CString::new(description).unwrap();
```
tekknolagi, line 176: assert 80 cols here?

ruby.c
XrXr: Actually, the elif above should be changed to if, since we'll have YJIT+ZJIT builds.

XrXr (approved):

ruby.c
tekknolagi: oh, ok

zjit/src/options.rs
k0kubun: The limitation is more tightly coupled to `show_usage_part` that formats each line with indentation (we can't even pass 80 cols of `description` from this function, for example). I would not add such an assertion on this side.

ruby.c
k0kubun: same discussion as https://github.com/ruby/ruby/pull/13197/files#r2066506891; this is deliberate.

ruby.c
k0kubun: It's an deliberate choice. There can be only one default option, hence `#elif` for it, but non-default options can co-exist.

PR #13198, opened 2025-04-28
Don't support blockarg in opt_new

tekknolagi:
Pairin' with Aaron.

cc @tenderlove

### DIFF
--- a/insns.def
+++ b/insns.def
@@ -916,11 +916,13 @@ opt_new
     VALUE argc = vm_ci_argc(cd->ci);
     VALUE val = TOPN(argc);
 
+    // The bookkeeping slot should be empty.
+    RUBY_ASSERT(TOPN(argc + 1) == Qnil);
+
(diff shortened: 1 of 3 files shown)


tekknolagi:
I would like to add a bunch of tests. Where should they go?

tekknolagi:
Linking to #13080

tenderlove:
> This testing code feels extremely ugly so if you have tips for how to improve, please let me know.

Were you able to make a short repro that causes a segv with RUBY_DEBUG enabled?  If so, I think that would be a good enough test.  I think we have a RUBY_DEBUG build so it should catch it.  The tests you added are fine, but they seem a little brittle because I think we could support `opt_new` in some of those cases, it will just take work.  That said, I'm fine merging this.

tekknolagi:
I think it's fine to have a test whose output we expect to change when we add an additional feature. We can migrate from the failing section to the passing section.

I got these to assert fail as expected (not segv), yes

PR #13199, opened 2025-04-28
ZJIT: Disable ZJIT instructions when USE_ZJIT is 0

k0kubun:


tekknolagi:
Seems reasonable. Does this implicitly rely on the zjit instructions a) being contiguous and b) being reliably ordered after all the other instructions?

XrXr (approved):

k0kubun:
> Does this implicitly rely on the zjit instructions a) being contiguous and b) being reliably ordered after all the other instructions?

Yes, it does. I added tests for the assumption https://github.com/ruby/ruby/pull/13199/commits/d67294e90f7d20829a3364f6bbac99a7e6f3651c.

I initially printed `#if 1` or `#if USE_ZJIT` for every instruction, but it made the source code in the distributed package harder to read. Also the offset calculation would be much harder to implement without relying on it.

> Ah, can you toss in a --disable-zjit build in .github/workflows/compilers.yml for good measure?

done https://github.com/ruby/ruby/pull/13199/commits/932f29cd1500b9d83d5d0ff4801a36942c7479f1

test/ruby/test_zjit.rb
```diff
@@ -479,6 +479,17 @@ def fib(n)
...
   end
 
+  # tool/ruby_vm/views/*.erb relies on the zjit instructions a) being contiguous and
+  # b) being reliably ordered after all the other instructions.
+  def test_instruction_order
+    zjit, others = RubyVM::INSTRUCTION_NAMES.map.with_index.partition { |name, _| name.start_with?('zjit_') }
+    zjit_indexes = zjit.map(&:last)
+    other_indexes = others.map(&:last)
+    zjit_indexes.product(other_indexes).each do |zjit_index, other_index|
+      assert zjit_index > other_index, "ZJIT instruction at #{zjit_index} must be defined after other instruction at #{other_index}"
```
tekknolagi, line 489: If you feel inclined, we can use the name in the error message instead of tossing it away

k0kubun: :+1: https://github.com/ruby/ruby/pull/13199/commits/26b9a45917c837a9c6f42840aaee7179f42d1ca8

PR #13200, opened 2025-04-29
Fix bug with multiline regexp and right anchor

composerinteralia:
https://bugs.ruby-lang.org/issues/21291

This seems to be an optimization bug. With `ONIG_DONT_OPTIMIZE` defined, `"aa\n" =~ /$.*/m` returns 2 as the author expected. But with optimizations turned on it does not find a match.

The optimization involves setting `ANCHOR_ANYCHAR_STAR_ML`, which ultimately limits the regexp search to the starting position. But in this case, because of the right anchor,  we won't find a match at the starting position. We need to keep searching, and that is what happens when the optimization is disable.

This commit adds the anchor information to the env, so we can skip the optimization when we've seen a right anchor prior to the quantifier.


composerinteralia:
Oops, I must have messed something up before pushing. Let me try that again

test/ruby/test_regexp.rb
```diff
@@ -2220,6 +2220,14 @@ def test_bug_20246 # [Bug #20246]
     assert_equal '1.2.3', '1.2.3'[/((?:\d|foo|bar)+)(\.\g<1>){2}/]
   end
 
+  def test_bug_21291 # [Bug #21291]
+    assert_equal 1, "a\n" =~ /$./m
+    assert_equal 1,  "a\n" =~ /$.*/m
+    assert_equal 2,  "aa\n" =~ /$./m
+    assert_equal 2,  "aa\n" =~ /$.*/m
+    assert_equal 2,  "aa\n" =~ /$[\S\s]*/m
```
byroot, line 2228: nitpick
```suggestion
    assert_equal 1, "a\n" =~ /$./m
    assert_equal 1, "a\n" =~ /$.*/m
    assert_equal 2, "aa\n" =~ /$./m
    assert_equal 2, "aa\n" =~ /$.*/m
    assert_equal 2, "aa\n" =~ /$[\S\s]*/m
```

composerinteralia: Oops! Thank you.

PR #13201, opened 2025-04-29
Improve syntax style consistency in shape.c and shape.h

casperisfine:
Most of this code use the `type * name` style, while the overwhemling majority of the rest of ruby use the `type *name` style.

This is a cosmetic change, but helps with readability.


byroot:
Merged as 18dac125cb48768b546a5913378e7349ee687492

PR #13203, opened 2025-04-29
ZJIT: Add `make zjit-test-lldb`

XrXr:
Handy for dropping into LLDB for a Rust test.


tekknolagi (approved):
Ooh, neat. Whatcha debugging?

k0kubun (approved):

PR #13204, opened 2025-04-29
ZJIT: Fix a couple of small bugs

tekknolagi:
- **Default instruction output type to Any**
- **Make sure to call find() on basic block arguments**

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -666,8 +666,12 @@ impl Function {
     // Add an instruction to the function without adding it to any block
     fn new_insn(&mut self, insn: Insn) -> InsnId {
         let id = InsnId(self.insns.len());
+        if insn.has_output() {
+            self.insn_types.push(types::Any);
+        } else {
+            self.insn_types.push(types::Empty);
+        }
         self.insns.push(insn);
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -680,6 +680,9 @@ impl Function {
         } else {
             self.blocks[block.0].insns.push(id);
         }
+        if self.insns[id.0].has_output() {
+            self.insn_types[id.0] = types::Any;
```
k0kubun, line 684: Is it possible to do this at `new_insn()` by checking `insn.has_output()`?

tekknolagi: Yeah. I'm trying to figure out what makes me want to do it here and what the differences are between an instruction that's been added to a block vs not

k0kubun: I'm worried about a possible situation that we initialize it with `types::Empty` at `new_insn()`, we update it to something before pushing the instruction, and then clear it with `types::Any` here. I'm thinking it's safer to set the default value first, not later like this.

tekknolagi: I agree with you and changed it. I think you are right

k0kubun (approved):

PR #13206, opened 2025-04-29
Don't call hash tombstone compaction from GC compaction

tenderlove:
Tombstone removal may possibly require allocation, and we're not allowed to allocate during GC.  This commit also renames `set_compact` to `set_gc_compact` to differentiate tombstone removal compaction with GC object compaction.

### DIFF
--- a/set.c
+++ b/set.c
@@ -179,10 +179,9 @@ set_replace_ref(st_data_t *key, st_data_t argp, int existing)
 }
 
 static void
-set_compact(void *ptr)
+set_update_references(void *ptr)
 {
     struct set_object *sobj = ptr;
-    set_compact_table(&sobj->table);
     set_foreach_with_replace(&sobj->table, set_foreach_replace, set_replace_ref, 0);
 }
 
@@ -192,7 +191,7 @@ static const rb_data_type_t set_data_type = {
         .dmark = set_mark,
         .dfree = set_free,
         .dsize = set_size,
-        .dcompact = set_compact,
+        .dcompact = set_update_references,
     },
     .flags = RUBY_TYPED_EMBEDDABLE | RUBY_TYPED_FREE_IMMEDIATELY | RUBY_TYPED_WB_PROTECTED | RUBY_TYPED_FROZEN_SHAREABLE
 };
--- a/test/ruby/test_set.rb
+++ b/test/ruby/test_set.rb
@@ -864,6 +864,19 @@ class TC_Set < Test::Unit::TestCase
       assert_equal(klass.new([a]), set, klass.name)
     }
   end
+
+  def test_set_gc_compact_does_not_allocate
+    assert_in_out_err([], <<-"end;", [], [])
+    def x
+      s = Set.new
+      s << Object.new
+      s
+    end
+
+    x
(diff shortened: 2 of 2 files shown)


test/ruby/test_set.rb
```diff
@@ -864,6 +864,19 @@ def test_reset
...
+  def test_set_gc_compact_does_not_allocate
+    assert_in_out_err([], <<-"end;", [], [])
+    def x
+      s = Set.new
+      s << Object.new
+      s
+    end
+
+    m = x
+    Process.warmup
```
byroot, line 877:
```suggestion
    GC.compact
```

nitpick: Could just be `GC.compact`.

set.c
```diff
@@ -192,7 +191,7 @@ static const rb_data_type_t set_data_type = {
         .dmark = set_mark,
         .dfree = set_free,
         .dsize = set_size,
-        .dcompact = set_compact,
+        .dcompact = set_gc_compact,
```
byroot, line 194:
```suggestion
        .dcompact = set_update_references,
```

Is a convention used for a few other types.

byroot (approved):
Nitpick, but LGTM

set.c
```diff
@@ -179,10 +179,9 @@ set_replace_ref(st_data_t *key, st_data_t argp, int existing)
 }
 
 static void
-set_compact(void *ptr)
+set_gc_compact(void *ptr)
```
byroot, line 182: need to update it here too. Sorry.

tenderlove: oops, I'll fix it

test/ruby/test_set.rb
```diff
@@ -864,6 +864,19 @@ def test_reset
...
+
+  def test_set_gc_compact_does_not_allocate
+    assert_in_out_err([], <<-"end;", [], [])
+    def x
+      s = Set.new
+      s << Object.new
+      s
+    end
+
+    m = x
```
byroot, line 876:
```suggestion
    x
```

Would fix:

```
  TC_Set#test_set_gc_compact_does_not_allocate [/Users/runner/work/ruby/ruby/src/test/ruby/test_set.rb:869]:
  pid 51454 exit 0
  | -:7: warning: assigned but unused variable - m
```

jeremyevans (approved):

test/ruby/test_set.rb
```diff
@@ -864,6 +864,19 @@ def test_reset
...
+  def test_set_gc_compact_does_not_allocate
+    assert_in_out_err([], <<-"end;", [], [])
+    def x
+      s = Set.new
+      s << Object.new
+      s
+    end
+
+    x
+    GC.compact
```
byroot, line 877: Ah mmtk, doesn't have `GC.compact`:

```
    1) Failure:
  TC_Set#test_set_gc_compact_does_not_allocate [/Users/runner/work/ruby/ruby/src/test/ruby/test_set.rb:869]:
  pid 52362 exit 1
  | -:8:in '<main>': compact() function is unimplemented on this machine (NotImplementedError)
  | 	from -:8:in '<main>'
  .
```

Sorry for suggesting that change :/

PR #13216, opened 2025-04-30
ZJIT: Compile opt_new to slow-path SendWithoutBlock

tekknolagi:
The optimizer will be able to handle this differently than the
interpreter; we will use type information to rewrite SendWithoutBlock
:new on any Class that doesn't override :new into the split
alloc+:initialize form.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1600,6 +1600,10 @@ fn compute_jump_targets(iseq: *const rb_iseq_t) -> Vec<u32> {
                 let offset = get_arg(pc, 0).as_i64();
                 jump_targets.insert(insn_idx_at_offset(insn_idx, offset));
             }
+            YARVINSN_opt_new => {
+                let offset = get_arg(pc, 1).as_i64();
+                jump_targets.insert(insn_idx_at_offset(insn_idx, offset));
+            }
             YARVINSN_leave | YARVINSN_opt_invokebuiltin_delegate_leave => {
                 if insn_idx < iseq_size {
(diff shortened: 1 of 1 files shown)


tekknolagi:
A question I have for the future: how hard is it to synthesize a CALL_DATA? If we're going to be turning this call to `:new` into a call to `:initialize`, the call to `:initialize` has to come from somewhere. I suppose we could try and keep track of the optimized "inlined" `:initialize` call from the `opt_new` compiler output, but that seems brittle. We'll probably also want to do more of this kind of stuff in the future unrelated to what's in the bytecode.

k0kubun (approved):

k0kubun:
> A question I have for the future: how hard is it to synthesize a CALL_DATA? If we're going to be turning this call to :new into a call to :initialize, the call to :initialize has to come from somewhere. I suppose we could try and keep track of the optimized "inlined" :initialize call from the opt_new compiler output, but that seems brittle. 

You'll need to discover an existing `cc` in the ISEQ to avoid extra allocation, and the same applies to `ci` when it has keyword arguments. And the `cd` itself should also be fetched from the bytecode to avoid consuming extra memory. Basically, we should not synthesize a `cd` but just discover it since the ISEQ already has one and we don't want to waste memory for JIT.

Since the bytecode has `send` instructions that call into `.new` and `#initialize`, we'll need to discover those instructions. I'm not sure what kind of algorithm we need to discover them, but if it's too hard, we could modify the interpreter to let the interpreter and ZJIT agree on where to find the instructions.

PR #13217, opened 2025-04-30
Use rb_current_ec_noinline in assertions

jhawthorn:
When doing a coroutine transfer from one thread to another, there's a risk that the compiler will reuse an address from TLS before the transfer to the new thread.

These VM assertions are all in places we would not otherwise be reading from TLS, but using the value of `ec` or `cr` passed in. Switching these to test against `rb_current_ec_noinline()` instead ensures there isn't an optimization applied to how we read `ruby_current_ec`.

Currently it seems we were hitting this on LLVM 18 specifically, but I don't know of any reason other versions wouldn't have the same issue.

cc @ko1

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -39,7 +39,8 @@ static void
 ASSERT_ractor_unlocking(rb_ractor_t *r)
 {
 #if RACTOR_CHECK_MODE > 0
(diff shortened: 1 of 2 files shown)


samuel-williams-shopify:
I think it would be a good idea to add documentation explaining the issue, as it may be non-obvious why the inline variant is used, and as there are no tests, nothing is preventing future regression.

PR #13219, opened 2025-04-30
YJIT: call alloc func directly in new

jhawthorn:
This removes an indirect branch by calling the allocator function from the class directly.

```
Benchmark 1: ./miniruby_master --yjit -e '20_000_000.times { Object.new }'
  Time (mean ± σ):      1.203 s ±  0.020 s    [User: 1.187 s, System: 0.011 s]
  Range (min … max):    1.168 s …  1.230 s    10 runs

Benchmark 2: ./miniruby_allocator --yjit -e '20_000_000.times { Object.new }'
  Time (mean ± σ):     967.5 ms ±  22.4 ms    [User: 951.6 ms, System: 11.5 ms]
  Range (min … max):   949.9 ms … 1026.0 ms    10 runs

Summary
  ./miniruby_allocator --yjit -e '20_000_000.times { Object.new }' ran
    1.24 ± 0.04 times faster than ./miniruby_master --yjit -e '20_000_000.times { Object.new }'
```

cc @tenderlove

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -4913,7 +4913,13 @@ fn gen_opt_new(
         // call rb_class_alloc to actually allocate
         jit_prepare_non_leaf_call(jit, asm);
 
-        let allocator = unsafe { rb_get_alloc_func(comptime_recv) };
+        let allocator =
+            if unsafe { FL_TEST(comptime_recv, VALUE(RUBY_FL_SINGLETON as usize)) } != VALUE(0) {
+                None
+            } else {
+                unsafe { rb_get_alloc_func(comptime_recv) }
+            };
+
         let allocator = allocator.unwrap_or(rb_obj_alloc);
         let obj = asm.ccall(allocator as _, vec![comptime_recv.into()]);
 


yjit/src/codegen.rs
```diff
@@ -4912,7 +4912,10 @@ fn gen_opt_new(
         // Fast path
         // call rb_class_alloc to actually allocate
         jit_prepare_non_leaf_call(jit, asm);
-        let obj = asm.ccall(rb_obj_alloc as _, vec![comptime_recv.into()]);
+
+        let allocator = unsafe { rb_get_alloc_func(comptime_recv) };
+        let allocator = allocator.unwrap_or(rb_obj_alloc);
+        let obj = asm.ccall(allocator as _, vec![comptime_recv.into()]);
```
k0kubun, line 4918: When an allocator function is found, it seems to skip the `if (rb_obj_class(obj) != rb_class_real(klass)) { rb_raise(rb_eTypeError, "wrong instance allocation"); }` check in `class_call_alloc_func`? Is the error unreachable?

tenderlove: Wouldn't this only be a problem for T_DATA objects?  I don't think they use `rb_class_new_instance_pass_kw` as their allocator.

jhawthorn: I think it's only reachable from a custom allocator which is implemented incorrectly (there's no reason to ever return an object with the wrong class), so I'm not sure if it's necessary to keep.

If it is I suppose we could allowlist just the most common builtin allocators (or even just the T_OBJECT allocator).

k0kubun: I honestly didn't know when it happens (it'd be nice to have a comment there) and thus asked a question. If the check is useless (on `opt_new`'s `rb_obj_alloc` call) on the interpreter too, it's fine.

tekknolagi:
I think we can revive this if you are still interested. We do something very similar in ZJIT after @byroot's shapes changes

PR #13221, opened 2025-05-01
Retire `windows-2019` and `ubuntu-20.04` for Ruby 3.4

hsbt:
For https://github.com/actions/runner-images/issues/12045

Note: I specified winsdk-10.0.22621.0 and Visual Studio 2019 toolchain(that version is 14.2). Because the default version of Visual Studio 2022 toolchain. see https://bugs.ruby-lang.org/issues/21167

### DIFF
--- a/.github/workflows/annocheck.yml
+++ b/.github/workflows/annocheck.yml
@@ -63,7 +63,7 @@ jobs:
       - run: id
         working-directory:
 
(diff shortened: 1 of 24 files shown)


hsbt:
I added extra commits for https://github.com/ruby/ruby/actions/runs/14769812299/job/41467921872?pr=13221

PR #13226, opened 2025-05-01
Namespace on read

tagomoris:
This change is to implement "Namespace on read" proposed by [Feature #21311](https://bugs.ruby-lang.org/issues/21311).

TODO:
* [x] Identify the cause and fix CI failures
* [x] Move prime classext readable/writable flags to RBasic flags (or abort this plan)
* [x] Rename `RCLASS_EXT` macro to prevent using it wrongly in the past way
* [x] Check review comments and fix code
* [x] Get a very basic benchmark to check performance regression
* [x] Show `[EXPERIMENTAL]` warning when namespace is enabled
* [x] Remove methods for debugging (`rb_class_debug_dump_*`)
* [x] Add a minimum doc, containing synopsis, known problems/limitations, list to be done, etc


namespace.c
```diff
@@ -0,0 +1,1139 @@
...
+
+const rb_data_type_t rb_namespace_data_type = {
+    "Namespace::Entry",
+    {
+        rb_namespace_entry_mark,
+        namespace_entry_free,
+        namespace_entry_memsize,
+        rb_namespace_gc_update_references,
+    },
+    0, 0, RUBY_TYPED_FREE_IMMEDIATELY
```
tenderlove, line 460: I'm not totally sure about the architecture, but I assume `NameSpace::Entry` objects live for a long time.  Should we make them `RUBY_TYPED_WB_PROTECTED`?

tagomoris: The lifecycle of `Namespace::Entry` is equal to `Namespace`, as long as the life of a namespace itself and will not be collected shortly. Is it better to specify `RUBY_TYPED_WB_PROTECTED` in such case?

fxn:
A curiosity. I have

```ruby
# test.rb
p Object.constants

ns = Namespace.new
ns.require_relative 'foo'
```

and

```ruby
# foo.rb
p Object.constants
```

If I run `make run` and compare, the main namespace has one extra constant: `:CROSS_COMPILING`.

My understanding is that said top-level constant must not belong to the root namespace and it is created in the user namespace. But, `test.rb` is not creating such constant. Which is the definition of root/user namespace?

fxn:
Oh, adding

```ruby
pp Object.const_source_location(:CROSS_COMPILING)
```

to `test.rb` segfaults.

tagomoris:
@fxn I ran the same script on `ruby` (and `miniruby`) and I got `:CROSS_COMPILING` in both output lines.
```
$ RUBY_NAMESPACE=1 build/exe/ruby -e 'p Object.constants.sort; ns = Namespace.new; ns.require_relative("foo")'
[:ARGF, :ARGV, :ArgumentError, :Array, :BasicObject, :Binding, :CROSS_COMPILING, :Class, # (snip)
[:ARGF, :ARGV, :ArgumentError, :Array, :BasicObject, :Binding, :CROSS_COMPILING, :Class, # (snip)
$ cat foo.rb 
p Object.constants.sort
```

On the other hand, when using `miniruby`, there were no `:CROSS_COMPILING` in both output lines.
```
$ RUBY_NAMESPACE=1 build/miniruby -e 'p Object.constants.sort; ns = Namespace.new; ns.require_relative("foo")'
[:ARGF, :ARGV, :ArgumentError, :Array, :BasicObject, :Binding, :Class, :ClosedQueueError, # (snip)
[:ARGF, :ARGV, :ArgumentError, :Array, :BasicObject, :Binding, :Class, :ClosedQueueError, # (snip)
```

`:CROSS_COMPILING` is set in `rbconfig.rb` for `ruby`, and in `#{archname}-fake.rb` for `miniruby`.
I just can imagine, does something (library?) require those files accidentally in your environment?

namespace.c
tagomoris: I fixed the type so.

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+ns = Namespace.new
+ns.require('foo')  # "This appears: 'foo'"
+
+p $foo      #=> nil
+p $VERBOSE  #=> false
+```
+
+### Top level constants
+
+Usually, top level constants are defined as constants of `Object`. In namespaces, top level constants are constants of `Object` in the namespace. And the namespace object `ns`'s constants are strictly equal to constants of `Object`.
```
eregon, line 218: > And the namespace object `ns`'s constants are strictly equal to constants of `Object`.

How is this implemented? Is the constants table shared between both the Namespace and that namespace's Object?
That seems quite dangerous implementation-wise as it might have far-reaching implications on constant lookup caching and invalidation.
I think it would be better to preserve the invariant that a constant table is only used for one Module, and implement `ns::Foo` another way, e.g. via `const_missing`.

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+The current namespace is the namespace that the executing code is in. `Namespace.current` returns the current namespace object.
+
+The loading namespace is an internally managed namespace to determine the namespace to load newly required/loaded files. For example, `ns` is the loading namespace when `ns.require("foo")` is called.
+
+## Discussions
+
+#### Namespace#inspect
+
+Currently, `Namespace#inspect` returns values like `"#<Namespace:0x00000001083a5660>"`. This results in the very redundant and poorly visible classpath outside the namespace.
```
eregon, line 304: Not only outside the namespace but also inside, so "always":
```
$ RUBY_NAMESPACE=1 ruby -ve 'ns = Namespace.new; File.write "ns.rb", "class C; end; p C"; ns.require "./ns"'    
ruby 3.5.0dev (2025-05-10T07:50:29Z namespace-on-read-.. bd4f57f96b) +PRISM [x86_64-linux]
ruby: warning: Namespace is experimental, and the behavior may change in the future!
See doc/namespace.md for know issues, etc.
#<Namespace:0x00007fd81ce67598>::C
```

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+* `NS1::C` (only when this namespace is created and assigned to NS1 in the current namespace)
+* `#<Namespace:user:1083a5660>::C` (with namespace type and without preceding 0)
+* or something else
+
+#### Namespace#eval
+
+Testing namespace features needs to create files to be loaded in namespaces. It's not easy nor casual.
+
+If `Namespace` class has an instance method `#eval` to evaluate code in the namespace, it can be helpful.
```
eregon, line 329: Yes, I was checking if this exists, I think it would be quite convenient.
That might also mean there is less need for `ns::C` and `ns.eval('C')` might be enough.

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### More builtin methods written in Ruby
+
+If namespace is enabled by default, builtin methods can be written in Ruby because it can't be overridden by users' monkey patches. Builtin Ruby methods can be JIT-ed, and it could bring performance reward.
```
eregon, line 333: > it can't be overridden by users' monkey patches

It would be good to expand this to explain what it would solve concretely with examples.
In https://bugs.ruby-lang.org/issues/21311#note-36 I wrote that method monkey patches are usually dealt with Primitives.
Maybe Namespace would make this nicer/less necessary?
Namespace I guess could make `Primitive.is_a?(obj, String)` safer in the case `String` would be re-assigned by some code (that's very unlikely though).

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### Monkey patching methods called by builtin methods
+
+Builtin methods sometimes call other builtin methods. For example, `Hash#map` calls `Hash#each` to retrieve entries to be mapped. Without namespace, Ruby users can overwrite `Hash#each` and expect the behavior change of `Hash#map` as a result.
+
+But with namespaces, `Hash#map` runs in the root namespace. Ruby users can define `Hash#each` only in user namespaces, so users cannot change `Hash#map`'s behavior in this case. To achieve it, users should override both`Hash#map` and `Hash#each` (or only `Hash#map`).
+
+It is a breaking change.
+
+It's an option to change the behavior of methods in the root namespace to refer to definitions in user namespaces. But if we do so, that means we can't proceed with "More builtin methods written in Ruby".
```
eregon, line 343: > It's an option to change the behavior of methods in the root namespace to refer to definitions in user namespaces.

How would that work though, e.g. if multiple namespaces have different Hash#each definitions?
You would always consider the methods of the Namespace.current also in calls inside builtin methods I guess.
That then becomes a problem for method lookup inline caching.

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### Context of \$LOAD\_PATH and \$LOADED\_FEATURES
+
+Global variables `$LOAD_PATH` and `$LOADED_FEATURES` control `require` method behaviors. So those namespaces are determined by the loading namespace instead of the current namespace.
+
+This could potentially conflict with the user's expectations. We should find the solution.
+
+#### Expose top level methods as a method of the namespace object
+
+Currently, top level methods in namespaces are not accessible from outside of the namespace. But there might be a use case to call other namespace's top level methods.
```
eregon, line 353: `Namespace#eval` would solve that

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### Split root and builtin namespace
+
+NOTE: "builtin" namespace is a different one from the "builtin" namespace in the current implementation
+
+Currently, the single "root" namespace is the source of classext CoW. And also, the "root" namespace can load additional files after starting main script evaluation by calling methods which contain lines like `require "openssl"`.
```
eregon, line 359: The root namespace, assuming it contains only the core library/builtin classes should have no method doing any `require`, isn't it?

doc/namespace.md
```diff
@@ -0,0 +1,417 @@
...
+
+The classext CoW is triggered when the contents of `rb_classext_t` are changed, including `cc_tbl`, `callable_m_tbl`, and `cvc_tbl`. But those three tables are changed by just calling methods or referring class variables. So, currently, classext CoW is triggered much more times than the original expectation.
+
+If we can move those three tables outside of `rb_classext_t`, the number of copied `rb_classext_t` will be much less than the current implementation.
+
+#### Object Shapes per namespace
+
+Now the classext CoW requires RClass and `rb_classext_t` to fallback its instance variable management from Object Shapes to the traditional `st_table`. It may have a performance penalty.
+
+If we can apply Object Shapes on `rb_classext_t` instead of `RClass`, per-namespace classext can have its own shapes, and it may be able to avoid the performance penalty.
```
eregon, line 404: Though that probably means an overhead for the non-Namespace case as the Shape is at least one more indirection away from the Class object.

doc/namespace.md
tagomoris: And, the "Discussions" section is to make discussions after merging this change, and I'm not going to make comments in this pull-requests on those points.
Thank you for making the comments.

doc/namespace.md
tagomoris: > How is this implemented? Is the constants table shared between both the Namespace and that namespace's Object?

Correct.
As far as I checked (and CI checked), it's working correctly. But I'll rethink if we find it'll not work well.

gc.c
```diff
@@ -2622,6 +2707,86 @@ mark_m_tbl(void *objspace, struct rb_id_table *tbl)
...
+
+struct mark_cc_entry_args {
+    rb_objspace_t *objspace;
+    VALUE klass;
+};
+
+static enum rb_id_table_iterator_result
+mark_cc_entry_i(ID id, VALUE ccs_ptr, void *data)
+{
+    struct mark_cc_entry_args *args = (struct mark_cc_entry_args *)data;
```
byroot, line 2735:
```
../gc.c:2735:32: warning: unused variable 'args' [-Wunused-variable]
 2735 |     struct mark_cc_entry_args *args = (struct mark_cc_entry_args *)data;
      | 
```

tagomoris: `args` is used only in the `VM_ASSERT`. I'll update it to directly cast and use it.

internal/class.h
```diff
@@ -79,103 +135,355 @@ STATIC_ASSERT(shape_max_variations, SHAPE_MAX_VARIATIONS < (1 << (sizeof(((rb_cl
...
+static inline rb_classext_t *
+RCLASS_EXT_TABLE_LOOKUP_INTERNAL(VALUE obj, const rb_namespace_t *ns)
 {
-    RUBY_ASSERT(RB_TYPE_P(obj, RUBY_T_CLASS) || RB_TYPE_P(obj, RUBY_T_MODULE));
-    if (rb_shape_obj_too_complex_p(obj)) {
-        uint32_t count;
+    st_data_t classext_ptr;
+    st_table *classext_tbl = RCLASS_CLASSEXT_TBL(obj);
+    if (classext_tbl) {
+        if (rb_st_lookup(classext_tbl, (st_data_t)ns->ns_object, &classext_ptr)) {
```
jhawthorn, line 352: It doesn't seem like this is safe to do without holding the VM lock. (The VM lock is held on write so it must also be held on read, otherwise the lock has no benefit)

PR #13227, opened 2025-05-01
Sync ruby/json

byroot:
Fix: https://github.com/ruby/json/issues/796

FYI: @hsbt


hsbt:
Thanks!

We can't test `crossbuild` until pushing master branch. Let's try this.

PR #13229, opened 2025-05-01
YJIT: ZJIT: Share identical glue functions

XrXr:
- **Add include guards for insns_info.inc**
- **YJIT: ZJIT: Share identical glue functions**

  Working towards having YJIT and ZJIT in the same build, we need to
  deduplicate some glue code that would otherwise cause name collision.
  Add jit.c for this and build it for YJIT and ZJIT builds. Update bindgen
  to look at jit.c; some shuffling of functions in the output, but the set
  of functions shouldn't have changed.


k0kubun:
I feel hesitant to include the name "glue" in the file name because people may put any future JIT glue code there, but we should probably use `yjit.c` or `zjit.c` if it's not shared.

Can we just call it `jit.c`? It should have functions that are shared by YJIT and ZJIT and would otherwise be defined in `yjit.c` or `zjit.c`. So just dropping `y` or `z` from the file name and not adding anything makes more sense to me.

k0kubun (approved):

tekknolagi:
Also: do we want to keep an `rb_jit` prefix for functions in `jit.c`?

XrXr:
> do we want to keep an rb_jit prefix for functions in jit.c?

Since these are for super simple operations, like reading off of a bit field, the extra prefex feel too heavy weight to me. (Similar reasoning for wrappers of `static` functions.)

PR #13234, opened 2025-05-01
dtoa.c: Use an on-stack buffer for small bigints

jhawthorn:
In https://github.com/ruby/ruby/pull/12991 I replaced a spinlock with malloc, which is about the same performance.

I think malloc implementations have gotten better since 1990, but this can still be a bit of a bottleneck. We seem to mostly be allocating a few small bigints (k <= 2) at a time, so this PR adds a small on-stack buffer we can allocate from without synchronization.

I need to test this a bit more before merging.

### DIFF
--- a/missing/dtoa.c
+++ b/missing/dtoa.c
@@ -501,8 +501,8 @@ extern double rnd_prod(double, double), rnd_quot(double, double);
 #define Kmax 15
 
 struct Bigint {
-    unsigned char k;
     unsigned short wds;
+    unsigned char k;
     unsigned char sign : 1;
     ULong x[1];
 };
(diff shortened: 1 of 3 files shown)


missing/dtoa.c
```diff
@@ -800,63 +849,41 @@ mult(Bigint *a, Bigint *b)
...
-            if (UNLIKELY(p5tmp)) {
-                Bfree(p5);
-                p5 = p5tmp;
-            }
-        }
-        FREE_DTOA_LOCK(1);
-    }
+    p5s = &p5s_static[0];
     for (;;) {
+        assert(p5s < &p5s_static[MAX_P5]);
```
jhawthorn, line 869: FIXME: Ran a fuzzer against this and we definitely sometimes exceed MAX_P5

PR #13236, opened 2025-05-02
Show the class variable name which caused a RactorIsolationError

osyoyu:
Class variable (@@var) access is prohibited in non-main Ractors and causes a RactorIsolationError.
Currently, RactorIsolationErrors do not carry any information about the actual class variable access which caused the violation. This makes it hard to debug non-Ractor-compatible code, especially when the culprit is in the middle of a long line.

This patch adds the class variable name which triggered the RactorIsolationError into its message. This follows the pattern of the "can not access non-shareable objects in constant FOO" error.

### DIFF
--- a/bootstraptest/test_ractor.rb
+++ b/bootstraptest/test_ractor.rb
@@ -1121,7 +1121,7 @@ values.join
 }
 
 # cvar in shareable-objects are not allowed to access from non-main Ractor
-assert_equal 'can not access class variables from non-main Ractors', %q{
+assert_equal 'can not access class variable @@cv from non-main Ractors', %q{
   class C
     @@cv = 'str'
   end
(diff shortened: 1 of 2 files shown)


osyoyu:
Example:

### Before

```
Ractor::IsolationError: can not access class variables from non-main Ractors
        from .../sinatra/lib/sinatra/show_exceptions.rb:26:in 'Sinatra::ShowExceptions#call'
        from .../sinatra/lib/sinatra/base.rb:227:in 'Sinatra::ExtendedRack#call'
        from .../sinatra/lib/sinatra/base.rb:2144:in 'Sinatra::Wrapper#call'
```

### After

```
Ractor::IsolationError: can not access class variable @@eats_errors from non-main Ractors
        from .../sinatra/lib/sinatra/show_exceptions.rb:26:in 'Sinatra::ShowExceptions#call'
        from .../sinatra/lib/sinatra/base.rb:227:in 'Sinatra::ExtendedRack#call'
        from .../sinatra/lib/sinatra/base.rb:2144:in 'Sinatra::Wrapper#call'
```

Error originating here.
https://github.com/sinatra/sinatra/blob/91cfb548c9e50a65324a9ce9e4ea5f10cd897027/lib/sinatra/show_exceptions.rb#L26

PR #13237, opened 2025-05-02
Test new pure C parser

soutaro:
This is to confirm the new pure C parser implemented in https://github.com/ruby/rbs/pull/2398 compiles in Ruby CI setting. This will break other bundled gems because RBS 4.0 has some breaking API changes.

### DIFF
--- a/gems/bundled_gems
+++ b/gems/bundled_gems
@@ -18,7 +18,7 @@ net-pop             0.1.2   https://github.com/ruby/net-pop
 net-smtp            0.5.1   https://github.com/ruby/net-smtp
 matrix              0.4.2   https://github.com/ruby/matrix
 prime               0.1.3   https://github.com/ruby/prime
-rbs                 3.9.2   https://github.com/ruby/rbs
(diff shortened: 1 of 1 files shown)


soutaro:
@Morriar @amomchilov @st0012 Can you help me to fix the compilation error?

https://github.com/ruby/ruby/actions/runs/14920483030/job/41914755919?pr=13237#step:5:1757

```
  /github/workspace/src/.bundle/gems/rbs-4.0.0.dev.2/ext/rbs_extension/../../src/util/rbs_allocator.c:53:39: error: use of undeclared identifier 'MAP_ANONYMOUS'
     53 |                         MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
        |                                       ^
  1 error generated.
```

Why `MAP_ANONYMOUS` is unavailable...??

Morriar:
It looks like `MAP_ANONYMOUS` is not available. I though it was POSIX 🤔 

We may need to provide a fallback. Working on a fix in https://github.com/ruby/rbs/pull/2470.

soutaro:
Hmm. It might be reasonable. Will review and test the PR, next week.

PR #13241, opened 2025-05-02
Make `rb_shape_t.type` an `enum shape_type`

byroot:
So it's easier to inspect shapes from gdb/lldb.

### DIFF
--- a/shape.h
+++ b/shape.h
@@ -41,12 +41,20 @@ typedef uint32_t redblack_id_t;
 
 typedef struct redblack_node redblack_node_t;
 
+enum shape_type {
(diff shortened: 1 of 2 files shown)


shape.h
```diff
@@ -41,12 +41,20 @@ typedef uint32_t redblack_id_t;
 
 typedef struct redblack_node redblack_node_t;
 
+enum shape_type {
```
ioquatix, line 44: I think it's better if it has an `rb_` prefix.

byroot:
This becomes messy with the yjit/zjit bindgen :/

PR #13242, opened 2025-05-02
ZJIT: Parse opt_newarray_send into HIR

tekknolagi:
This is a fast path that can avoid heap-allocating an intermediate Array
by using values on the stack.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -141,6 +141,7 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
                 write!(f, "BOPRedefined(")?;
                 match klass {
                     INTEGER_REDEFINED_OP_FLAG => write!(f, "INTEGER_REDEFINED_OP_FLAG")?,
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -1309,9 +1331,13 @@ impl Function {
             necessary[insn_id.0] = true;
             match self.find(insn_id) {
                 Insn::PutSelf | Insn::Const { .. } | Insn::Param { .. }
-                | Insn::NewArray { .. } | Insn::PatchPoint(..)
-                | Insn::GetConstantPath { .. } =>
+                | Insn::PatchPoint(..) | Insn::GetConstantPath { .. } =>
                     {}
+                Insn::ArrayMax { elements, state }
+                | Insn::NewArray { elements, state } => {
```
tekknolagi, line 1337: This is an extant bug that I noticed and fixed on the way

k0kubun (approved):

PR #13243, opened 2025-05-02
ZJIT: Parse topn, opt_length, opt_size into HIR

tekknolagi:


k0kubun (approved):

zjit/src/hir.rs
```diff
@@ -1573,6 +1573,12 @@ impl FrameState {
         self.stack[idx] = opnd;
     }
 
+    /// Get a stack operand at idx
+    fn stack_topn(&mut self, idx: usize) -> Result<InsnId, ParseError> {
+        let idx = self.stack.len() - idx - 1;
+        self.stack.get(idx).ok_or_else(|| ParseError::StackUnderflow(self.clone())).copied()
```
XrXr, line 1579: nit: Since there is an attempt to issue an error here: The math to compute idx can panics in case of underflow because of the rust build config we use. Use [`usize::checked_sub`](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.checked_sub) if you want.

XrXr (approved):

zjit/src/hir.rs
tekknolagi: ooh ok there are a couple of cases like this. i can go through and look

PR #10487, opened 2024-04-08
YJIT: Optimize local variables when EP == BP

k0kubun:
This PR optimizes local variable access in `getlocal` and `setlocal` from `REG_CFP->ep[offset]` (2 instructions) to `REG_SP[offset]` (1 instruction) when EP == BP. When EP escapes, the block gets invalidated.

Because local variable access is now based off of SP, this will allow us to handle local variables using `Opnd::Stack` and allocate registers for them in a future PR.

### Before
```asm
  # Insn: 0000 getlocal_WC_0 (stack_size: 0)
  0x59e90a49a058: mov rax, qword ptr [r13 + 0x20]
  # reg_temps: 00000000 -> 00000001
  0x59e90a49a05c: mov rsi, qword ptr [rax - 0x18]
```

### After
```asm
  # Insn: 0000 getlocal_WC_0 (stack_size: 0)
  # reg_temps: 00000000 -> 00000001
  0x6271819a1058: mov rsi, qword ptr [rbx - 0x20]
```

## Benchmark
This seems to improve the performance of the following benchmarks.

```
before: ruby 3.4.0dev (2024-04-17T02:45:52Z master 0b630d6441) +YJIT [x86_64-linux]
after: ruby 3.4.0dev (2024-04-17T05:46:52Z yjit-ep-bp 3b316da2e3) +YJIT [x86_64-linux]

--------------  -----------  ----------  ----------  ----------  -------------  ------------
bench           before (ms)  stddev (%)  after (ms)  stddev (%)  after 1st itr  before/after
erubi-rails     1219.7       0.1         1160.5      0.1         1.04           1.05
liquid-compile  74.1         0.5         73.6        0.8         1.00           1.01
liquid-render   94.8         0.3         93.0        0.6         1.01           1.02
psych-load      1785.5       0.1         1743.8      0.1         1.02           1.02
protoboeuf      59.0         0.6         58.2        0.8         1.01           1.01
--------------  -----------  ----------  ----------  ----------  -------------  ------------
```

PR #13245, opened 2025-05-02
YJIT: End the block after OPTIMIZE_METHOD_TYPE_CALL

k0kubun:
This PR lets `OPTIMIZE_METHOD_TYPE_CALL` end the block at the instruction, which currently keeps compiling. 

If we don't end the block and it's followed by a setlocal in the same block, we can't invalidate the setlocal instruction properly when `OPTIMIZE_METHOD_TYPE_CALL` escapes the EP.

See also: https://github.com/ruby/ruby/pull/10487, which relies on https://github.com/ruby/ruby/pull/10539

### DIFF
--- a/test/-ext-/debug/test_debug.rb
+++ b/test/-ext-/debug/test_debug.rb
@@ -75,4 +75,19 @@ class TestDebug < Test::Unit::TestCase
     end
     assert_equal true, x, '[Bug #15105]'
   end
+
+  # This is a YJIT test, but we can't test this without a C extension that calls
+  # rb_debug_inspector_open(), so we're testing it using "-test-/debug" here.
+  def test_yjit_invalidates_setlocal_after_inspector_call
+    val = setlocal_after_proc_call(proc { Bug::Debug.inspector; :ok })
+    assert_equal :ok, val
+  end if defined?(RubyVM::YJIT) && RubyVM::YJIT.enabled?
+
+  private
+
+  def setlocal_after_proc_call(block)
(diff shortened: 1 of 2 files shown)


udipl:
Are there plans to include this in the next release of v3.4 ?

k0kubun:
It was already backported https://github.com/ruby/ruby/pull/13261 and released as part of Ruby 3.4.4 https://github.com/ruby/ruby/compare/v3_4_3...v3_4_4.

udipl:
oh, interesting. I ran into [this Puma issue](https://github.com/puma/puma/issues/3620) running 3.4.9, so I guess this was not the cause.

k0kubun:
It was mentioned from it as "Potential YJIT fix" https://github.com/puma/puma/issues/3620#issuecomment-2849095895, but it was only potential, and later we learned that it was not the fix.

The actual fix was https://github.com/ruby/ruby/pull/16381. The [current branch maintainer](https://github.com/ruby/ruby/wiki/Release-Engineering#ruby_3_4) will probably release it when they get a chance. 

I'm the release manager for `ruby_4_0`, and I do regular releases with a published schedule as long as the branch is maintained/released by myself. The release managers for older versions like `ruby_3_4` don't publish their release schedule.

If you care about whether there are "plans to include this in the next release" or not, I highly recommend you to upgrade your Ruby version to Ruby 4.0 instead. It has a published release schedule, and I generally sweep all backport requests before each release.

PR #13248, opened 2025-05-03
Fix Set#to_h to be backwards compatible when called with a block

jeremyevans:
Set includes Enumerable, and Enumerable#to_h accepts a block, so there is likely code relying on the block being respected and not ignored.  If a block is given, call super (Enumerable#to_h) to restore backwards compatibility.

Also add tests and documentation for the method.


jeremyevans:
Closing, as `Set#to_h` was removed in #13247

PR #13250, opened 2025-05-03
Fix an ASAN error in Array#hash

mame:
The issue was introduced by c5fc1ce975ecdf1c6818714e47579c5d3531c4ca, and this change reverts that commit partially.

[Bug #21304]


jhawthorn (approved):
Thank you. This makes sense. Clearly `RARRAY_CONST_PTR` (or the length) can change because the `rb_hash` can call into user code and resize it 😩

PR #13253, opened 2025-05-03
Handle mutation of array being merged into set

jeremyevans:
Check length of array during every iteration, as a #hash method could truncate the array, resulting in heap-use-after-free.

Fixes [Bug #21305]


mame (approved):
Thank you! I confirmed that the ASAN error no longer occurs when running the reported test with this patch.

PR #13254, opened 2025-05-03
[DOC] Tweaks for String#-@"

BurdetteLamar:


string.c
```diff
@@ -3602,24 +3602,26 @@ str_uplus(VALUE str)
 
 /*
  * call-seq:
- *   -string -> frozen_string
- *   dedup -> frozen_string
+ *   -self -> frozen_string
  *
- * Returns a frozen, possibly pre-existing copy of the string.
```
byroot, line 3608: I think it would be good to keep the "possibly pre-existing" part.

I like your tweaks in general, but I think it makes it much less clear that it helps deduplicate strings.

BurdetteLamar: Note added.

string.c
```diff
@@ -3602,24 +3602,29 @@ str_uplus(VALUE str)
...
+ * Returns a frozen string containing the same content as +self+.
  *
- * The returned +String+ will be deduplicated as long as it does not have
- * any instance variables set on it and is not a String subclass.
+ * The returned +frozen_string+ is +self+, frozen, if +self+:
```
byroot, line 3609: `self` is only returned if it was already frozen (and then the two conditions you mentioned).

string.c
```diff
@@ -3602,24 +3602,29 @@ str_uplus(VALUE str)
 
 /*
  * call-seq:
- *   -string -> frozen_string
- *   dedup -> frozen_string
+ *   -self -> frozen_string
  *
- * Returns a frozen, possibly pre-existing copy of the string.
+ * Returns a frozen string containing the same content as +self+.
```
byroot, line 3607:
```suggestion
 * Returns a frozen string equal to +self+.
```

"containing content" is a bit redundant no?

string.c
BurdetteLamar: Recast.

byroot (approved):

string.c
```diff
@@ -3602,24 +3602,30 @@ str_uplus(VALUE str)
...
+ * - +self+ is an instance of \String (rather than of a subclass of \String)
+ * - +self+ has no instance variables set on it.
  *
- *    FILENAME = -'config/database.yml'
+ * Otherwise, the returned string is a frozen copy of +self+.
  *
- * while +dedup+ is better suitable for using the method in chains
- * of calculations:
+ * Returning +self+, when possible, saves duplicating data;
+ * see {Data deduplication}[https://en.wikipedia.org/wiki/Data_deduplication].
```
peterzhu2118, line 3618: While this is true, it's more powerful in that the returned string is such that for any other string `str` where `str == self`, then `-str` and `-self` will always be the same object. This allows for more deduplication that what this documentation implies. For example:

```ruby
str1 = "hello"
str2 = "hello"

puts "str1 and str2 are not the same object? #{str1.object_id != str2.object_id}"
puts "-str1 and -str2 are the same object? #{(-str1).object_id == (-str2).object_id}"
```

BurdetteLamar: Added.

peterzhu2118 (approved):

PR #13259, opened 2025-05-05
`constants_spec.rb`: Adding `Math::TAU`

Magayaga:
So, can `Math::TAU` be a mathematical constant? For example, `puts Math::TAU` is similar to `puts Math::PI * 2`. Thanks.


XrXr:
Please see https://github.com/ruby/ruby/wiki/How-To-Request-Features

hsbt:
and see https://bugs.ruby-lang.org/issues/17496

PR #13262, opened 2025-05-06
YJIT: ZJIT: Allow both JITs in the same build

XrXr:
This commit allows building YJIT and ZJIT simultaneously, a "combo
build". Previously, `./configure --enable-yjit --enable-zjit` failed. At
runtime, though, only one of the two can be enabled at a time.

Add a root Cargo workspace that contains both the yjit and zjit crate.
The common Rust build integration mechanisms are factored out into
defs/jit.mk.

Combo YJIT+ZJIT dev builds are supported; if either JIT uses
`--enable-*=dev`, both of them are built in dev mode.

The combo build requires Cargo, but building one JIT at a time with only
rustc in release build remains supported.


.github/workflows/zjit-macos.yml
```diff
@@ -35,6 +35,12 @@ jobs:
           - test_task: 'zjit-test'
             configure: '--enable-zjit=dev'
 
+          - test_task: 'ruby' # build test for combo build
```
k0kubun, line 38: We should probably do this in `.github/workflows/compilers.yml`. It used to be just a huge matrix but now it allows us to minimize the number of jobs (because we have limited parallelism) for this kind of test.

k0kubun:
Can we enable both by default when cargo is available? (out of scope in this PR?)

k0kubun (approved):
Great job :+1:

XrXr:
> Can we enable both by default when cargo is available? (out of scope in this PR?)

Out of scope for now, since we're not sure whether we want to auto enable in any situation.

.github/workflows/zjit-macos.yml
XrXr: I moved the Ubuntu ones, but there's no macOS runs in `compilers.yml` yet, so I'm going to leave these.

PR #13265, opened 2025-05-07
Update to ruby/spec@d8bacef

andrykonchin:

cc @eregon


andrykonchin:
New specs for Digest fail to build a native extension on CI:

```
    3300 .digest_spec.c:4:10: fatal error: ruby/digest.h: No such file or directory
      4 | #include "ruby/digest.h"
        |          ^~~~~~~~~~~~~~~
  compilation terminated.
  make[1]: *** [Makefile:252: digest_spec.o] Error 1
```

https://github.com/ruby/ruby/actions/runs/14882697018/job/41794332898?pr=13265
https://github.com/ruby/ruby/actions/runs/14882697027/job/41794334147?pr=13265

Wondering what's a proper way to fix this issue. Cannot reproduce this failure running specs manually with MSpec and Ruby 3.5.0 dev

eregon:
@nobu Could you help here? How should the `digest.h` header be found in `make test-spec`?

eregon:
The `extconf.rb` is created here: https://github.com/ruby/spec/blob/d8bacef459831bd410d6df375cd65e6e7b33c3b1/optional/capi/spec_helper.rb#L77-L84

If we need different behavior for `make test-spec` in CRuby you can use `MSpecScript.instance_variable_defined?(:@testing_ruby)` which returns true only in that case (example usages: https://github.com/search?q=repo%3Aruby%2Fspec%20testing_ruby&type=code).

andrykonchin:
Seems there is also a similar issue with `make` command, it fails with:

```
../src/spec/ruby/optional/capi/ext/digest_spec.c:4:10: fatal error: ruby/digest.h: No such file or directory
    4 | #include "ruby/digest.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```
https://github.com/ruby/ruby/actions/runs/14887591192/job/41811251631?pr=13265
https://github.com/ruby/ruby/actions/runs/14887591241/job/41811259477?pr=13265
https://github.com/ruby/ruby/actions/runs/14887591248/job/41811278856?pr=13265

eregon:
The problem with https://github.com/ruby/ruby/actions/runs/14935951886/job/41963330473?pr=13265 is it does:
```
/github/workspace/src/configure -C --with-gcc=clang-21 --enable-debug-env --disable-install-doc --with-ext=-test-/cxxanyargs,+ --enable-shared --without-valgrind --without-jemalloc --without-gmp CFLAGS= CXXFLAGS= optflags=-O1 cppflags= debugflags=-ggdb3
```
Notably that `--with-ext=-test-/cxxanyargs,+`.
That means there is no `.ext/include/ruby/digest`:
```
$ tree .ext                   
.ext
├── common
├── include
│   └── x86_64-linux
│       └── ruby
│           └── config.h
└── x86_64-linux
```

So there is no way to build those C API specs at that time.

I think building C API specs from Makefile is too hard to maintain, I try to disable it.

eregon:
IMO this is a perfect example of the complications of trying to run specs on not-yet-installed not-yet-full CRuby.
Other Ruby implementations run specs on a fully-built Ruby, and it just works and avoids all these headaches.
I think CRuby devs should really make it faster/easier to build a fully-built Ruby locally, make that the default, and stop hacking around makefiles in ways which are far too complicated.

PR #13268, opened 2025-05-07
ZJIT: Temporarily revert path compression

tekknolagi:
For reasons I don't understand yet, this causes an issue when trying to
boot yjit-bench. Temporarily revert it.


maximecb (approved):

tekknolagi:
One guess I have right now is that the second `borrow_mut` in this PR (undone in favor of `find!`) is in the same scope as the other `find!` calls but I cannot repro this error locally :/

k0kubun (approved):

XrXr:
Yeah... very strange. I don't think it's second `borrow_mut` in this PR either because adding a bare `self.union_find.borrow_mut()` after it doesn't crash, indicating that the borrow guard dropped on the same line.

PR #13270, opened 2025-05-07
ZJIT: Quietly abort codegen if we don't have enough registers

tekknolagi:
This temporarily paves over the register allocator issue that allocates
basic block parameters from a small fixed list of physical registers.


tekknolagi:
I'd like to clean this up before landing but anyone is welcome to steal it and land something similar

PR #13274, opened 2025-05-08
Register the `envtbl` global variable

byroot:
As far as I can tell, this global isn't registered anywhere. It won't be collected because it's referenced in a constant, but I don't see anything that would keep that global variable up to date in case of compaction.


byroot:
Nevermind, `rb_define_const` does pin objects.

PR #13276, opened 2025-05-08
ZJIT: 2 rest param fixes

XrXr:
See each commit:

- **ZJIT: Fix rest parameter not parsed into a BB parameter**
- **ZJIT: Avoid using SendWithoutBlockDirect for rest param callees for now**

Reported by @rwstauner.


zjit/src/hir.rs
```diff
@@ -1067,8 +1068,12 @@ impl Function {
                             // TODO(max): Allow non-iseq; cache cme
                             self.push_insn_id(block, insn_id); continue;
                         }
-                        self.push_insn(block, Insn::PatchPoint(Invariant::MethodRedefined { klass, method: mid }));
                         let iseq = unsafe { get_def_iseq_ptr((*cme).def) };
+                        // Reject callees with a rest parameter
+                        if unsafe { get_iseq_flags_has_rest(iseq) } {
```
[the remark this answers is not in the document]
XrXr: Sounds good, I'll drop the second commit.

XrXr:
Related to the second commit: In the future, do we want to have separate opcodes or split `SendWithoutBlockDirect` into smaller parts for complex parameter passing logic? 
Of course we could do everything at the LIR layer in a more or less one pass manner like YJIT, but I get a sense that we want to feed it through the optimizer in a multiple-pass way.

zjit/src/hir.rs
```diff
@@ -3112,6 +3107,18 @@ mod opt_tests {
...
     }
 
+    #[test]
+    fn test_rest_param_get_bb_param() {
+        eval("
+            def rest(*array) = array
+        ");
+        assert_optimized_method_hir("rest", expect![[r#"
+            fn rest:
+            bb0(v0:BasicObject):
```
tekknolagi, line 3117: I wonder if we can (easily) add type information here. Is it always an exact Array?

zjit/src/hir.rs
```diff
@@ -1719,7 +1714,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     let mut queue = std::collections::VecDeque::new();
     let mut entry_state = FrameState::new(iseq);
     for idx in 0..num_locals(iseq) {
-        if idx < num_lead_params(iseq) {
+        if idx < unsafe { get_iseq_body_param_size(iseq) }.as_usize() {
```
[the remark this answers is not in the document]
XrXr: Yes.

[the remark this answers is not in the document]
tekknolagi: Please add tests for those then

zjit/src/utils.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// all unsgined types.
+    fn as_usize(self) -> usize {
+        self.into()
+    }
+}
+
+/// The `Into<u64>` Rust does not provide.
+/// Convert to u64 with assurance that the value is preserved.
+/// Currently, `usize::BITS == 64` holds for all platforms we support.
+pub(crate) trait IntoU64 {
```
k0kubun, line 51: This is already defined in `cast.rs`. Could you move other traits to that file as well? We tried to stop creating a file that anything could go into.

zjit/src/hir.rs
XrXr: 👍🏼 done in [90cfe46](https://github.com/ruby/ruby/pull/13276/commits/90cfe46e8cd3c96c93baf75d91505b0a27413335) 

(weird ripple: keyword rest is not always a HashExact since it could be nil when anonymous)

XrXr:
I just realized that all the tests this adds will conflict with #13277, since they just won't parse anymore 😐. Not sure what to do there.

tekknolagi:
I mean if this PR now correctly parses all of these parameter types into HIR, then this very much supersedes my PR!

zjit/src/hir.rs
```diff
@@ -3112,6 +3121,52 @@ mod opt_tests {
...
+
+        assert_optimized_method_hir("rest", expect![[r#"
+            fn rest:
+            bb0(v0:ArrayExact):
+              Return v0
+        "#]]);
+        // extra hidden param for number of specified keywords
+        assert_optimized_method_hir("kw", expect![[r#"
+            fn kw:
+            bb0(v0:BasicObject, v1:BasicObject):
```
tekknolagi, line 3143: I don't know that this hidden param with the count should be reified into a HIR register if it's not being manipulated by Ruby code

zjit/src/hir.rs
```diff
@@ -1719,7 +1728,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     let mut queue = std::collections::VecDeque::new();
     let mut entry_state = FrameState::new(iseq);
     for idx in 0..num_locals(iseq) {
-        if idx < num_lead_params(iseq) {
+        if idx < unsafe { get_iseq_body_param_size(iseq) }.as_usize() {
```
tekknolagi, line 1731: Now that we apparently handle all cases (?) can you please add a comment explaining how different parameter types get turned into HIR? Positionals seem the most "obvious" but other stuff seems like it would take a logical leap

zjit/src/hir.rs
```diff
@@ -912,9 +913,23 @@ impl Function {
     fn infer_types(&mut self) {
         // Reset all types
         self.insn_types.fill(types::Empty);
-        for param in &self.blocks[self.entry_block.0].params {
+        let rest_param_idx = if !self.iseq.is_null() && unsafe { get_iseq_flags_has_rest(self.iseq) } {
```
tekknolagi, line 916: To avoid computing this every time we may want to store `param_types` on `Function` or something like that and copy that in instead

zjit/src/hir.rs
tekknolagi: oh boy

is this the `kw_bits` thing?

XrXr: Yep. At some point maybe DCE can also drop unused parameters?

XrXr:
> I mean if this PR now correctly parses all of these parameter types into HIR, then this very much supersedes my PR!

It panicks due to side exiting when calling into methods with a rest parameter, though, like the test case in https://github.com/ruby/ruby/commit/c72ccd91100b502b5560dfe0aaaafc8d47f6ed49. I'm also not sure why it's side exiting.

Problem for another PR.

zjit/src/hir.rs
tekknolagi: Perhaps!

zjit/src/hir.rs
XrXr: Sure, I tossed it in a `OnceCell`.

zjit/src/hir.rs
XrXr: I wrote one, but I'm not sure if it's the kind you're looking for. I think other than the hidden param for keyword params, the count is pretty much always the size of the comma separated parameter list in the source text.

Let me know if something is unclear.

zjit/src/hir.rs
```diff
@@ -912,9 +923,27 @@ impl Function {
...
+        });
+
+        // Fill parameter types
+        for (idx, param) in self.blocks[self.entry_block.0].params.iter().enumerate() {
             // We know that function parameters are BasicObject or some subclass
             self.insn_types[param.0] = types::BasicObject;
+
+            // Rest parameters are always ArrayExact
+            if let Ok(true) = c_int::try_from(idx).map(|idx| idx == rest_param_idx) {
+                self.insn_types[param.0] = types::ArrayExact;
```
tekknolagi, line 945: Can you store a `param_types` vec on `Function` instead of `rest_param_idx`? If you'd rather not do it in this PR I can stack a new commit.

tekknolagi: (& initialize in HIR builder, not OnceCell)

tekknolagi (approved):
Seems good to go +/- `param_types`

tekknolagi:
Yay! Thank you!

PR #13278, opened 2025-05-08
[DOC] ObjectSpace#each_object behavior in multi-Ractor mode

osyoyu:
This behavior of ObjectSpace#each_object has been around since Ruby 3.0 when Ractors were first introduced, but was never documented and has caused some amount of confusion:

https://bugs.ruby-lang.org/issues/17360
https://bugs.ruby-lang.org/issues/19387
https://bugs.ruby-lang.org/issues/21149

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1545,10 +1545,9 @@ os_obj_of(VALUE of)
  *     2.2250738585072e-308
  *     Total count: 7
  *
- *  Due to implementation limitations, this method will not yield
- *  Ractor-unshareable the process is in multi-Ractor mode (when
+ *  Due to a current known Ractor implementation issue, this method will not yield
+ *  Ractor-unshareable objects in multi-Ractor mode (when
  *  <code>Ractor.new</code> has been called within the process at least once).
- *  This behavior is not considered ideal.
  *  See https://bugs.ruby-lang.org/issues/19387 for more information.
  *
  *     a = 12345678987654321 # shareable
--- a/gc.c
+++ b/gc.c
@@ -1545,9 +1545,11 @@ os_obj_of(VALUE of)
  *     2.2250738585072e-308
  *     Total count: 7
  *
- *  When the process is in multi-Ractor mode (when <code>Ractor.new</code>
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -1582,6 +1582,17 @@ os_obj_of(VALUE of)
  *     2.2250738585072e-308
  *     Total count: 7
  *
+ *  When the process is in multi-Ractor mode (when <code>Ractor.new</code>
+ *  has been called within the process at least once), this method yields
+ *  Ractor-shareable objects only.
```
eregon, line 1587: I think we should clarify this is a bug (it's clearly incompatible), it just hasn't been solved yet.
And maybe link to one of these issues.

eregon: Also https://bugs.ruby-lang.org/issues/19387 is opened so it's not just "won't fix" and a known bug.

osyoyu: Understood. I made it clear that the current behavior is not ideal in 00bdbbaea8bdfa690a12cf78aad18d9c3f8cdb62. Does this look good to you?

gc.c
```diff
@@ -1545,6 +1545,19 @@ os_obj_of(VALUE of)
  *     2.2250738585072e-308
  *     Total count: 7
  *
+ *  Due to implementation limitations, this method will not yield
+ *  Ractor-unshareable the process is in multi-Ractor mode (when
+ *  <code>Ractor.new</code> has been called within the process at least once).
+ *  This behavior is not considered ideal.
+ *  See https://bugs.ruby-lang.org/issues/19387 for more information.
```
eregon, line 1552:
```suggestion
 *  Due to a current known Ractor implementation issue, this method will not yield
 *  Ractor-unshareable objects in multi-Ractor mode (when
 *  <code>Ractor.new</code> has been called within the process at least once).
 *  See https://bugs.ruby-lang.org/issues/19387 for more information.
```

eregon (approved):

osyoyu:
Thank you!

PR #13279, opened 2025-05-08
ZJIT: Fix a splitting condition for LHS

k0kubun:
This PR fixes a splitting logic in the x86_64 backend. I re-discovered this bug while writing a test case for side exits.

The `Opnd::Value` match arm added in https://github.com/Shopify/zjit/pull/38 was broken by the backend refactoring https://github.com/Shopify/zjit/pull/86 that removed the concept of `unmapped_opnd`, which had `Opnd::Value`. It now only deals with `Opnd::UImm` lowered from `Opnd::Value`, so the match arm should check for `Opnd::UImm`.

### DIFF
--- a/zjit/src/backend/x86_64/mod.rs
+++ b/zjit/src/backend/x86_64/mod.rs
@@ -183,7 +183,7 @@ impl Assembler
                                     *left = asm.load(*left);
                                 },
                                 // The first operand can't be an immediate value
-                                (Opnd::Value(_), _) => {
+                                (Opnd::UImm(_), _) => {
                                     *left = asm.load(*left);
                                 }
                                 _ => {}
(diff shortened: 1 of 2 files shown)


zjit/src/backend/x86_64/mod.rs
```diff
@@ -183,7 +183,7 @@ impl Assembler
                                     *left = asm.load(*left);
                                 },
                                 // The first operand can't be an immediate value
```
tekknolagi, line 185: What does this comment mean? It looks like now it *can* be an immediate? Or maybe I am totally misreading

k0kubun: There's no encoding where LHS is an immediate. This code turns such an immediate into a non-immediate by loading the immediate into a register.

XrXr (approved):

PR #13280, opened 2025-05-08
Fix `remove_instance_variable` on complex objects

byroot:
Introduced in: https://github.com/ruby/ruby/pull/13159

Now that there is no longer a unique TOO_COMPLEX shape with no children, checking `shape->type == TOO_COMPLEX` is incorrect.


jhawthorn (approved):
Looks like these are all the cases of the `== SHAPE_OBJ_TOO_COMPLEX` pattern 👍

PR #13282, opened 2025-05-08
YJIT: Split the block on optimized getlocal/setlocal

k0kubun:
This PR is a more generalized fix than https://github.com/ruby/ruby/pull/13245 to the problem the PR addressed.

It splits the block when getlocal or setlocal is optimized and thus needs to be invalidated when EP escapes. I'd like to still keep https://github.com/ruby/ruby/pull/13245 because send instructions are highly polymorphic, so it's generally nice to share the successor blocks.


yjit/src/codegen.rs
```diff
@@ -2536,6 +2541,11 @@ fn gen_setlocal_generic(
     }
 
     let (flags_opnd, local_opnd) = if level == 0 && jit.assume_no_ep_escape(asm) {
```
XrXr, line 2543: This seems like it would make 2 blocks with the ep escape assumption no? Before and after the defer_compilation(). I think only the newest block should assume the assumption.

k0kubun:
> No other message, and I haven't seen this symptom recently. Possibly uncovered by this change.

Addressing https://github.com/ruby/ruby/pull/13282#discussion_r2081660385 with https://github.com/ruby/ruby/pull/13282/commits/b4c207a0ed2e039793062952416cd1407b430e9c seems to have removed the symptom on this PR's CI.

Note that it has happened on master too:
https://github.com/ruby/ruby/actions/runs/14958270977/job/42016839015
http://ci.rvm.jp/results/trunk-yjit@ruby-sp2-noble-docker/5752983
maybe since namespace changes? Either way, the issue doesn't seem relevant to this PR.

PR #13287, opened 2025-05-09
bundle rbs-3.9.4 to fix test failure

soutaro:
<s>RBS 3.9.3 ships with updated test, which depends on `erb` instead of `set`.</s>

<s>rbs-3.9.3 is already in `bundled_gems`, but the test is failing because of `CGI`.
This PR skips the CGI tests in rbs test suite.</s>

Released rbs-3.9.4 to fix the failing tests.


hsbt (approved):

tool/rbs_skip_tests
```diff
@@ -72,3 +72,7 @@ test_iconv(JSONSingletonTest)
 # https://github.com/ruby/json/pull/774
 test_recurse_proc(JSONInstanceTest)
 test_recurse_proc(JSONSingletonTest)
+
+CGITest CGI is retired
+CGISingletonTest CGI is retired
+test_nph?(CGITest) skipping CGI doesn't work???
```
soutaro, line 78:
```suggestion
```

PR #13288, opened 2025-05-09
Remove dead code in `rb_gc_impl_ractor_cache_free`

byroot:
Followup: https://github.com/ruby/ruby/pull/13286

FYI: @jhawthorn


etiennebarrie (commented):
Just came to do the same 👍

PR #13289, opened 2025-05-09
Get rid of frozen shapes

casperisfine:
Instead `shape_id_t` higher bits contain flags, and the first one tells whether the shape is frozen.

This has multiple benefits:
  - Can check if a shape is frozen with a single bit check instead of
    dereferencing a pointer.
  - Guarantees it is always possible to transition to frozen.
  - This allow reclaiming `FL_FREEZE`  (TODO).
  - Allow to efficiently normalize the `shape_id` on `getivar` so that frozen status doesn't invalidate inline caches.

### DIFF
--- a/ext/objspace/objspace_dump.c
+++ b/ext/objspace/objspace_dump.c
@@ -818,9 +818,6 @@ shape_id_i(shape_id_t shape_id, void *data)
         dump_append_id(dc, shape->edge_name);
 
         break;
-      case SHAPE_FROZEN:
-        dump_append(dc, ", \"shape_type\":\"FROZEN\"");
-        break;
       case SHAPE_T_OBJECT:
         dump_append(dc, ", \"shape_type\":\"T_OBJECT\"");
         break;
--- a/shape.c
+++ b/shape.c
@@ -20,17 +20,7 @@
 #define SHAPE_DEBUG (VM_CHECK_MODE > 0)
 #endif
 
-#if SIZEOF_SHAPE_T == 4
-#if RUBY_DEBUG
-#define SHAPE_BUFFER_SIZE 0x8000
-#else
-#define SHAPE_BUFFER_SIZE 0x80000
-#endif
-#else
-#define SHAPE_BUFFER_SIZE 0x8000
-#endif
-
-#define ROOT_TOO_COMPLEX_SHAPE_ID 0x2
+#define ROOT_TOO_COMPLEX_SHAPE_ID 0x1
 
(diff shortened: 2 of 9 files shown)


ext/objspace/objspace_dump.c
```diff
@@ -416,7 +416,7 @@ dump_object(VALUE obj, struct dump_config *dc)
 
     size_t shape_id = rb_obj_shape_id(obj);
     dump_append(dc, ", \"shape_id\":");
-    dump_append_sizet(dc, shape_id);
+    dump_append_sizet(dc, shape_id >> SHAPE_ID_SHIFT);
```
casperisfine, line 419: One awkward thing I did envision is that now we could dump a `shape_id` that isn't listed in the shapes.

So I decided to "untag" the shape IDs for now.

shape.c
```diff
@@ -42,7 +42,7 @@
 #define SINGLE_CHILD_P(x) ((uintptr_t)(x) & SINGLE_CHILD_TAG)
 #define SINGLE_CHILD(x) (rb_shape_t *)((uintptr_t)(x) & SINGLE_CHILD_MASK)
 #define ANCESTOR_CACHE_THRESHOLD 10
-#define MAX_SHAPE_ID (SHAPE_BUFFER_SIZE - 1)
+#define MAX_SHAPE_ID ((SHAPE_BUFFER_SIZE - 1) >> SHAPE_ID_SHIFT)
```
byroot, line 45: TODO: bump `SHAPE_BUFFER_SIZE` accordingly on 64bit?

casperisfine:
Another note: This only make sense if we get rid of the `!SHAPE_IN_BASIC_FLAGS` codepaths. Otherwise to freeze a generic object (e.g. T_HASH) you need to create an entry in the generic ivar table which is crazy.

Or alternatively, we'd need to make the `!SHAPE_IN_BASIC_FLAGS` path a bit more complicated, and keep using `FL_FREEZE` in that case.

shape.h
```diff
@@ -8,34 +8,42 @@
...
+#define SHAPE_ID_OFFSET_NUM_BITS 19
 
 #else
 
 #define SIZEOF_SHAPE_T 2
 typedef uint16_t attr_index_t;
 typedef uint16_t shape_id_t;
-# define SHAPE_ID_NUM_BITS 16
+#define SHAPE_ID_NUM_BITS 16
+#define SHAPE_ID_OFFSET_NUM_BITS 15
```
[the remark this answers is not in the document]
casperisfine: https://github.com/ruby/ruby/pull/13500 would unlock all 32 bits.

PR #13290, opened 2025-05-09
 Refactor `rb_obj_hash` to rely on object address instead of object_id

casperisfine:
Whenever we observe the address of an object we record that as an object flag. Then if later on the object has to be moved, either by the GC compaction or through the Ractor API, then we store the old address inline inside the object fields.

cc @eightbitraptor

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -6095,7 +6095,7 @@ rb_gc_impl_writebarrier_remember(void *objspace_ptr, VALUE obj)
 struct rb_gc_object_metadata_names {
     // Must be ID only
     ID ID_wb_protected, ID_age, ID_old, ID_uncollectible, ID_marking,
-        ID_marked, ID_pinned, ID_object_id, ID_shareable;
+        ID_marked, ID_pinned, ID_object_id, ID_shareable, ID_address_seen;
 };
 
 #define RB_GC_OBJECT_METADATA_ENTRY_COUNT (sizeof(struct rb_gc_object_metadata_names) / sizeof(ID))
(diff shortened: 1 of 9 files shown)


ractor.c
```diff
@@ -3597,6 +3598,17 @@ move_leave(VALUE obj, struct obj_traverse_replace_data *data)
...
 
+    if (FL_TEST_RAW(obj, RUBY_FL_ADDRESS_SEEN) && !rb_obj_old_address_p(obj)) {
+        rb_shape_t *old_address_shape = rb_obj_old_address_shape(obj);
+        VALUE old_address;
+#if SIZEOF_LONG == SIZEOF_VOIDP
+        old_address = LONG2NUM(obj);
+#elif SIZEOF_LONG_LONG == SIZEOF_VOIDP
+        old_address = LL2NUM(obj);
+#endif
+        rb_obj_field_set(data->replacement, old_address_shape, old_address);
```
casperisfine, line 3609: TODO: we're not handling complex shape here, but I want to refactor `rb_obj_field_set` to handle it.

gc/default/default.c
```diff
@@ -5471,14 +5487,21 @@ gc_compact_move(rb_objspace_t *objspace, rb_heap_t *heap, VALUE src)
...
-            VALUE dest = rb_gc_impl_location(objspace, src);
             rb_gc_set_shape(dest, new_shape);
         }
         RMOVED(src)->original_shape_id = orig_shape;
     }
-
+    if (old_address) {
+        DURING_GC_COULD_MALLOC_REGION_START();
+        {
+            rb_obj_field_set(dest, old_address_shape, old_address);
```
casperisfine, line 5501: NB: With the introduction of namespaces, now classes potentially have one set of fields per namespaces, so storing the address in fields is no longer possible.

That being said, from my understanding this only apply to "core" classes, which I suspect are all pinned, so we could probably still get away with it, but we'd have to be careful.

PR #13292, opened 2025-05-09
ZJIT: Store CME on SendWithoutBlockDirect, not ISEQ

tekknolagi:
This lets us issue SendWithoutBlockDirect on non-ISEQs, such as C
functions. This means we can compile-time-lookup methods such as
`String#upcase`, which are (currently) written in C.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -255,7 +255,10 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::IfTrue { val, target } => return gen_if_true(jit, asm, opnd!(val), target),
         Insn::IfFalse { val, target } => return gen_if_false(jit, asm, opnd!(val), target),
         Insn::SendWithoutBlock { call_info, cd, state, .. } => gen_send_without_block(jit, asm, call_info, *cd, &function.frame_state(*state))?,
-        Insn::SendWithoutBlockDirect { iseq, self_val, args, .. } => gen_send_without_block_direct(cb, jit, asm, *iseq, opnd!(self_val), args)?,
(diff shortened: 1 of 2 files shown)


maximecb (approved):

zjit/src/codegen.rs
```diff
@@ -255,7 +255,10 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::IfTrue { val, target } => return gen_if_true(jit, asm, opnd!(val), target),
         Insn::IfFalse { val, target } => return gen_if_false(jit, asm, opnd!(val), target),
         Insn::SendWithoutBlock { call_info, cd, state, .. } => gen_send_without_block(jit, asm, call_info, *cd, &function.frame_state(*state))?,
-        Insn::SendWithoutBlockDirect { iseq, self_val, args, .. } => gen_send_without_block_direct(cb, jit, asm, *iseq, opnd!(self_val), args)?,
+        Insn::SendWithoutBlockDirect { cme, cd, call_info, state, .. } if unsafe { get_cme_def_type(*cme) } == VM_METHOD_TYPE_CFUNC =>
```
k0kubun, line 258: I feel like this implies that we want to use a different HIR instruction for C func calls. We need a different way to set up arguments from JIT-to-JIT calls, so there's not much to share between C func calls and ISEQ calls anyway.

maximecb: Seems like a good point 👍

tekknolagi: This is partially why I mentioned a potential split between Lookup and Call in the design doc. Maybe I should propose something more concrete

tekknolagi:
Closing in favor of an in-progress branch that @tenderlove and I are working on

PR #13293, opened 2025-05-09
Fix warnings in tests for ObjectSpace._id2ref

peterzhu2118:
There's a lot of warnings when running test_objectspace.rb because of ObjectSpace._id2ref. For example:

    test_objectspace.rb:19: warning: ObjectSpace._id2ref is deprecated


byroot (approved):
Thank you. I didn't think of that.

PR #13295, opened 2025-05-09
ZJIT: Stop padding side exits

k0kubun:
This PR lets the backend compile jumps to side exits using `Target::SideExitPtr` instead of `Target::CodePtr` to avoid filling unnecessary `nop` instructions after such jumps.

The backend assumes any `Target::CodePtr` could be invalidated due to patch points or chain guards. However, we never invalidate side exits. So, in YJIT, we introduced `Target::SideExitPtr` to distinguish side exits from other jumps.


tekknolagi (approved):

fivemhostingserver:
This is a well-structured codebase. I appreciate the clean separation of concerns.

maximecb (approved):

PR #13298, opened 2025-05-10
Reduce `Object#object_id` contention.

byroot:
Based on: https://github.com/ruby/ruby/pull/13297

If the object isn't shareable and already has a object_id we can access it without a lock.

If we need to generate an ID, we may need to lock to find the child shape.

We also currently still lock to increment the counter but we could use an atomic increment, but it means being limited to 32bits on some platforms.

Happy path benchmark:

```ruby
Warning[:experimental] = false

ractors = 10.times.map do
  Ractor.new do
    o = Object.new
    2_000_000.times do
      o.object_id
    end
    o.object_id
  end
end

ractors.each(&:take)
```

this branch vs 3.4.2:

```
Benchmark 1: ./ruby -v --disable-all ../test.rb
  Time (mean ± σ):     138.0 ms ±   3.6 ms    [User: 693.0 ms, System: 3.7 ms]
  Range (min … max):   132.4 ms … 145.8 ms    20 runs
 
Benchmark 2: ruby -v --disable-all ../test.rb
  Time (mean ± σ):      2.080 s ±  0.071 s    [User: 3.335 s, System: 5.702 s]
  Range (min … max):    1.971 s …  2.238 s    10 runs
 
Summary
  ./ruby -v --disable-all ../test.rb ran
   15.08 ± 0.65 times faster than ruby -v --disable-all ../test.rb
```

Roughly 15 times faster.


gc.c
```diff
@@ -1812,53 +1811,46 @@ static const rb_data_type_t id_to_obj_tbl_type = {
...
-    if (rb_shape_too_complex_p(shape)) {
-        st_table *table = ROBJECT_FIELDS_HASH(obj);
-        if (rb_shape_has_object_id(shape)) {
-            st_lookup(table, (st_data_t)ruby_internal_object_id, (st_data_t *)&id);
-            RUBY_ASSERT(id, "object_id missing");
+    // TODO: can we do 64bit atomics? Or accept that `object_id_counter`
+    // is sometimes only 32 bits?
+    unsigned int lock_lev = rb_gc_vm_lock();
+    id = ULL2NUM(++object_id_counter * OBJ_ID_INCREMENT);
+    rb_gc_vm_unlock(lock_lev);
```
byroot, line 1830: @jhawthorn no rush, but I wonder if you have opinions about what is best to do here.

`rb_atomic_t` is 32bits on i686 but also on Windows AFAICT.

byroot: Actually, I suppose I can just check for the size of `rb_atomic_t` and go the atomic route if it's large enough.

byroot:
object_id creation benchmark, here again on an happy path where we don't need to lock to find the shape transition:

```ruby
Warning[:experimental] = false

class Foo
  def initialize
    @bar = 1
  end
end

ractors = 10.times.map do
  Ractor.new do
    200_000.times do
      Foo.new.object_id
    end
  end
end

ractors.each(&:take)
```

```
$ hyperfine 'ruby -v --disable-all ../test.rb' './ruby -v --disable-all ../test.rb'
Benchmark 1: ruby -v --disable-all ../test.rb
  Time (mean ± σ):      1.688 s ±  0.176 s    [User: 1.207 s, System: 3.067 s]
  Range (min … max):    1.567 s …  2.137 s    10 runs
 
  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.
 
Benchmark 2: ./ruby -v --disable-all ../test.rb
  Time (mean ± σ):     121.4 ms ±   0.8 ms    [User: 472.0 ms, System: 45.1 ms]
  Range (min … max):   119.9 ms … 122.8 ms    24 runs
 
Summary
  ./ruby -v --disable-all ../test.rb ran
   13.90 ± 1.45 times faster than ruby -v --disable-all ../test.rb
```

Here again ~14 times faster.

PR #13299, opened 2025-05-10
Fix prism instruction generation differences for `def a(x, ...); b(...); end`

zenspider:
Diff instructions against parse.y using:

    DUMP="puts RubyVM::InstructionSequence.compile(ARGV.first).disasm.gsub(/^\d+ /, '')"
    RUBY="def a(x, ...); b(...); end"

    env -u RUBYOPT \
        diff -U99 \
        <(./build/ruby -v --parser=parse.y -e "$DUMP" "$RUBY" 2>/dev/null) \
        <(./build/ruby -v --parser=prism   -e "$DUMP" "$RUBY" 2>/dev/null)

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1763,7 +1763,7 @@ pm_setup_args_core(const pm_arguments_node_t *arguments_node, const pm_node_t *b
 
                 orig_argc += 2;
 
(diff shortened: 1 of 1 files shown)


zenspider:
@nobu I'm not sure but it looks like windows builds might have been broken around your recent clock changes.

zenspider:
I guess all these fixes eventually made its way into master over the past 3 months... would have been nice to be acknowledged. 🤷🏻

PR #13305, opened 2025-05-12
gc.c: Fix a race condition in `object_id`

casperisfine:
If an object is shareable and has no capacity left, it isn't safe to store the object ID in fields as it requires an object resize which can't be done unless all field reads are synchronized.
    
So in this case we have to store the ID externally like we used to.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1909,6 +1909,7 @@ obj_to_id_tbl_compact(void *data)
         st_foreach(obj_to_id_tbl, obj_to_id_tbl_any_moved_p_i, (st_data_t)&any_moved_args);
 
         if (any_moved_args.need_rebuild) {
+            fprintf(stderr, "-- rebyild---\n");
             st_table *new_table;
             DURING_GC_COULD_MALLOC_REGION_START();
             {
@@ -1969,7 +1970,7 @@ class_object_id(VALUE klass)
 static inline VALUE
 object_id_get(VALUE obj, shape_id_t shape_id)
 {
-    VALUE id;
+    VALUE id = Qundef;
 
     switch (rb_shape_has_object_id(shape_id)) {
       case SHAPE_NO_OBJ_ID:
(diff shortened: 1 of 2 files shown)


casperisfine:
So this fixes the obvious race, but there is a more fundamental issue that isn't and maybe can't be fixed.

Prior to moving the `object_id` inside the object itself, frozen (hence shareable) objects would never be mutated, so reading their instance variables was always safe.

Now that `object_id` may need to resize a frozen object, `ivar_get` would need to synchronize when called on a potentially shared object, and that's really not what we want.

It's not great, but I'm afraid we may need to go back to a synchronized table to store object ids, at least in the case where the object is frozen and would need to be enlarged.

gc.c
```diff
@@ -1901,23 +1956,46 @@ class_object_id(VALUE klass)
...
+    // resize it. So we have to store the object_id externally.
+    if (shareable && current_shape->next_field_index == current_shape->capacity) {
+        shape_id_t object_id_shape_id = rb_shape_transition_external_object_id(obj);
+
+        if (RB_UNLIKELY(!obj_to_id_tbl)) {
+            obj_to_id_tbl = st_init_numtable();
+            obj_to_id_value = TypedData_Wrap_Struct(0, &obj_to_id_tbl_type, obj_to_id_tbl);
+        }
+        st_insert(obj_to_id_tbl, obj, id);
+        rb_shape_set_shape_id(obj, object_id_shape_id);
```
casperisfine, line 1990: We're running into the same issue than https://github.com/ruby/ruby/pull/13289. If we on 32bits the shape_id is in the generic_ivar table...

gc/gc.h
```diff
@@ -135,6 +136,34 @@ gc_ref_update_table_values_only(st_table *tbl)
...
+
+    return ST_CONTINUE;
+}
+
+static void
+gc_ref_update_table_keys_only(st_table *tbl)
+{
+    if (!tbl || tbl->num_entries == 0) return;
+
+    // FIXME: this certainly isn't correct. If a key moved, we need to re-hash.
```
casperisfine, line 161: Note: don't merge without addressing that.

gc.c
```diff
@@ -1940,12 +2018,12 @@ object_id(VALUE obj)
 
     if (UNLIKELY(rb_gc_multi_ractor_p() && rb_ractor_shareable_p(obj))) {
```
byroot, line 2019:
```suggestion
    if (UNLIKELY(rb_gc_multi_ractor_p() && RB_OBJ_SHAREABLE_P(obj))) {
```

Faster and more restrictive check.

byroot:
Close in favor of https://github.com/ruby/ruby/pull/13609

PR #13306, opened 2025-05-12
[DOC] Tweaks for String#<<

BurdetteLamar:


string.c
```diff
@@ -4308,37 +4308,34 @@ rb_str_append_as_bytes(int argc, VALUE *argv, VALUE str)
...
  *    s << 33 # => "foo!"
  *
- *  If that codepoint is not representable in the encoding of
- *  _string_, RangeError is raised.
- *
- *    s = 'foo'
- *    s.encoding              # => <Encoding:UTF-8>
- *    s << 0x00110000         # 1114112 out of char range (RangeError)
- *    s = 'foo'.encode(Encoding::EUC_JP)
- *    s << 0x00800080         # invalid codepoint 0x800080 in EUC-JP (RangeError)
```
peterzhu2118, line 4332: We lost information about this exception case, is that deliberate?

string.c
```diff
@@ -4308,37 +4308,34 @@ rb_str_append_as_bytes(int argc, VALUE *argv, VALUE str)
...
+ *  changes the encoding to Encoding::ASCII_8BIT:
  *
  *    s = 'foo'.encode(Encoding::US_ASCII)
- *    s << 0xff
- *    s.encoding              # => #<Encoding:BINARY (ASCII-8BIT)>
+ *    s.encoding # => #<Encoding:US-ASCII>
+ *    s << 0xff  # => "foo\xFF"
+ *    s.encoding # => #<Encoding:BINARY (ASCII-8BIT)>
  *
- *  Related: String#concat, which takes multiple arguments.
```
peterzhu2118, line 4341: Is there a reason why you removed `String#concat` from the related section?

string.c
BurdetteLamar: Yes.  In general, the [doc guide](https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html#label-Corner+Cases+and+Exceptions) discourages mentioning exceptions.  In this case, I thought the error message for the raised exception explains itself.

peterzhu2118: The doc guide states:

> For corner cases of methods, such as atypical usage, briefly mention the behavior, but do not provide any examples.
>
> Only document exceptions raised if they are not obvious.

The error message might be obvious, but I don't think it's obvious what the behavior is for this exceptional case.

BurdetteLamar: Restored.

peterzhu2118 (approved):

PR #13313, opened 2025-05-13
Fix Namespace RubyGems isolation issue

jsxs0:
Ensure independent RubyGems environments per namespace

This fix resolves an issue where the Namespace feature fails to properly isolate RubyGems state. In the current implementation, RubyGems gets loaded into the root namespace and is shared across all user namespaces.  This causes gem versions loaded in one namespace to affect others,  leading to `Gem::LoadError` problems.

The fix explicitly removes any inherited `Gem` constant during main namespace and newly created namespace initialisation, then loads a fresh  RubyGems environment in each namespace. This allows each namespace to manage its own gem dependencies without conflicts between namespaces.

Fixes Bug #21324.

### DIFF
--- a/namespace.c
+++ b/namespace.c
@@ -567,6 +567,23 @@ namespace_initialize(VALUE namespace)
 
     setup_pushing_loading_namespace(ns);
 
+    // Initialize RubyGems for this new namespace, if gems are not disabled.
+    // Check if Gem constant exists in Object in this namespace context, and if so, remove it
+    // before initializing a fresh RubyGems environment
+    if (!GET_VM()->ruby_options->disable_gems)
+    {
+        // First check if we need to remove an inherited Gem constant in this namespace's Object
+        VALUE ns_object = rb_const_get(namespace, rb_intern("Object"));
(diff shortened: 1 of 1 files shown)


nobu:
Couldn't these be extracted as a function?

namespace.c
```diff
@@ -1007,6 +1023,18 @@ rb_initialize_main_namespace(void)
     rb_const_set(rb_cNamespace, rb_intern("MAIN"), main_ns);
 
     vm->main_namespace = th->ns = main_namespace = ns;
+
+    // If RubyGems was loaded in the root namespace during boot,
+    // remove it and load a fresh copy in the main namespace
```
tagomoris, line 1028: The line `th->ns = ns` specifies the main namespace as the current namespace. So the cleaning RubyGems status should be done above the line.

tagomoris:
I'm not familiar to RubyGems itself, but it seems better to clear RubyGems's internal status than to remove-and-require `rubygems`.
And I guess this code doesn't work well because the `.rb` files of rubygems are already included in `$LOADED_FEATURES`.

PR #13315, opened 2025-05-13
Fix `object_id` for classes and modules in namespace context

casperisfine:
Given classes and modules have a different set of fields in every namespace, we can't store the `object_id` in fields for them.

First because the `shape_id` is different if an object has an `object_id`, so we'd lookup the ID in a `st_table` that may not have it and crash.
But even if somehow each `class_ext` had a `shape_id`, the `object_id` would be different in every namespace, and as far as I understand that's not how namespaces are supposed to work.

Given that some space was freed in `RClass` we can store it there instead, which has the benefit of allowing it to be lockless (not fully in this PR but I have another one to generate object IDs atomically).

@tagomoris did I get namespaces correctly?

Also cc @jhawthorn

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1861,9 +1861,42 @@ static const rb_data_type_t id_to_obj_tbl_type = {
     .flags = RUBY_TYPED_WB_PROTECTED | RUBY_TYPED_FREE_IMMEDIATELY
 };
 
+#define RUBY_ATOMIC_VALUE_LOAD(x) (VALUE)(RUBY_ATOMIC_PTR_LOAD(x))
+
+static VALUE
+class_object_id(VALUE klass)
+{
+    VALUE id = RUBY_ATOMIC_VALUE_LOAD(RCLASS(klass)->object_id);
+    if (!id) {
+        unsigned int lock_lev = rb_gc_vm_lock();
+        id = ULL2NUM(next_object_id);
+        next_object_id += OBJ_ID_INCREMENT;
+        VALUE existing_id = RUBY_ATOMIC_VALUE_CAS(RCLASS(klass)->object_id, 0, id);
+        if (existing_id) {
+            id = existing_id;
+        }
+        else if (RB_UNLIKELY(id_to_obj_tbl)) {
+            st_insert(id_to_obj_tbl, id, klass);
+        }
+        rb_gc_vm_unlock(lock_lev);
+    }
+    return id;
+}
+
 static VALUE
 object_id(VALUE obj)
 {
+    switch (BUILTIN_TYPE(obj)) {
+      case T_CLASS:
+      case T_MODULE:
+        // With namespaces, classes and modules have different fields
+        // in different namespaces, so we cannot store the object id
+        // in fields.
+        return class_object_id(obj);
+      default:
+        break;
+    }
+
     VALUE id = Qfalse;
     unsigned int lock_lev;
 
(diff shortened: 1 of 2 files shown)


tagomoris:
@byroot When the `rb_classext_t` is copied at the first time (for the klass), it evict shapes: https://github.com/ruby/ruby/blob/master/internal/class.h#L406

As far as I understood, the root cause of this problem in the current implementation is `object_id` is not stored in the ivar tbl (not shape because evicted) in a namespace B after:
1. `MyClass.object_id` is not set yet
2. `MyClass`'s `rb_classext_t` is copied from namespace A to B
3. `MyClass.object_id` is set (stored into ivar table) and the flag about object_id is set in the namespace A
4. `MyClass.object_id` is referred in namespace B, the flag says the `object_id` is stored, but not exists in the ivar tbl

My original idea (popped up in talking w/ @ko1 ) to solve this problem is, when the `object_id` is set to ivar tbl, set the value into all ivar tables in each `rb_classext_t`. Once `object_id` value is set into the ivar tbl of the root namespace, it'll be copied to other namespaces.

tagomoris:
From my viewpoint:

* byroot's solution
  * Very simple, easy to understand
  * RClass (&classext) size: +1 VALUE
* my solution
  * Very complex, straightforward API doesn't exist (need to walk around the set of usual field functions)
  * No additional fields in RClass (especially from viewpoint of https://github.com/ruby/ruby/blob/master/doc/namespace.md#minimize-the-size-of-rclass-and-rb_classext_t )

byroot:
> RClass (&classext) size: +1 VALUE

Well, technically these bytes are currently wasted, so the size in unchanged. And you solution would mean one more `st_table_entry` per namepace, so `+3 VALUE` for each class that has an `object_id`.

tagomoris:
Ah, probably I misunderstand something.
@byroot I'm ok if adding the field in RClass is not a problem.

byroot:
> Ah, probably I misunderstand something.

No, you were right, I'm the one that was wrong, this indeed make classes one `VALUE` larger.

casperisfine:
> WDYT?

Yes, I'd rather get a fix out, because right now `object_id` crashes when namespaces are enabled, which prevent testing them any further. I'll wait for @jhawthorn's opinion though.

gc.c
```diff
@@ -1861,9 +1861,40 @@ static const rb_data_type_t id_to_obj_tbl_type = {
     .flags = RUBY_TYPED_WB_PROTECTED | RUBY_TYPED_FREE_IMMEDIATELY
 };
 
+static VALUE
+class_object_id(VALUE klass)
+{
+    VALUE id = RCLASS(klass)->object_id;
```
jhawthorn, line 1867: It might be nice to do an atomic load here, otherwise TSan will complain, but I think in practice it doesn't matter here (ie. an atomic load with memory order relaxed should work). So not a blocker.

jhawthorn (approved):
Looks good to me. One place I think TSan will complain, but I'm not sure it's worth caring about.

tagomoris:
How about adding `RCLASS_OBJECT_ID` macro to control the access to `RCLASS(klass)->object_id`?

gc.c
casperisfine: Actually, I couldn't make it work, so I used the same hack you used in `string.c`. Sorry.

PR #13319, opened 2025-05-13
Reclaim one `VALUE` from `rb_classext_t` by shrinking `super_classdepth`

casperisfine:
By making `super_classdepth` `uint16_t`, classes and modules can now fit in 160B slots again.

The downside of course is that before `super_classdepth` was large enough we never had to care about overflow, as you couldn't realistically create enough classes to ever go over it.

With this change, while it is stupid, you could realistically create an ancestor chain containing 65k classes and modules.

@jhawthorn thoughts?

Also cc @tagomoris


jhawthorn (approved):

jhawthorn:
Looks good to me!

> 65k classes and modules

It's actually only classes which are recorded in the superclasses Array, so this should be even less of an issue.

To me it also doesn't seem like namespaces should need to look up a separate superclass array per-namespace, since the superclass shouldn't be possible to change. I'll look at that in a separate PR.

byroot:
> To me it also doesn't seem like namespaces should need to look up a separate superclass array per-namespace

Yes, I think there are a bunch of things in `classext_t` that aren't namespace dependent and that we could avoid duplicating.

internal/class.h
```diff
@@ -145,14 +145,16 @@ struct RClass {
 };
 
 // Assert that classes can be embedded in heaps[2] (which has 160B slot size)
-// TODO: restore this assertion after removing several fields from rb_classext_t
-// STATIC_ASSERT(sizeof_rb_classext_t, sizeof(struct RClass) + sizeof(rb_classext_t) <= 4 * RVALUE_SIZE);
+// On 32bit platforms there is no variable width allocation so it doesn't matter.
```
casperisfine, line 148: I can't reenable that assertion if we merge: https://github.com/ruby/ruby/pull/13315

PR #13325, opened 2025-05-13
[DOC] Tweaks for String#=~

BurdetteLamar:
I could not find any non-Regexp +object+ that did not raise TypeError;  are there any?

- If so, some correction is needed.
- If not, should we say "Raises TypeError if +object+ is not a Regexp."?

### DIFF
--- a/string.c
+++ b/string.c
@@ -5319,10 +5319,10 @@ rb_str_byterindex_m(int argc, VALUE *argv, VALUE str)
 
 /*
  *  call-seq:
- *    self =~ regexp -> integer or nil
+ *    self =~ object -> integer or nil
  *
- *  Returns the index of the first matching substring in +self+
- *  matched by +regexp+,
+ *  When +object+ is a Regexp, returns the index of the first substring in +self+
+ *  matched by +object+,
  *  or +nil+ if no match is found;
  *  updates {Regexp-related global variables}[rdoc-ref:Regexp@Global+Variables]:
  *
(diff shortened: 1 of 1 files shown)


string.c
```diff
@@ -5319,30 +5319,30 @@ rb_str_byterindex_m(int argc, VALUE *argv, VALUE str)
...
  *    'foo' =~ /f/ # => 0
+ *    $~           # => #<MatchData "f">
  *    'foo' =~ /o/ # => 1
+ *    $~           # => #<MatchData "o">
  *    'foo' =~ /x/ # => nil
- *
- *  Note: also updates Regexp@Global+Variables.
- *
- *  If the given +object+ is not a Regexp, returns the value
- *  returned by <tt>object =~ self</tt>.
```
peterzhu2118, line 5335: We lost this documentation for the case when `regexp` is not a regexp.

BurdetteLamar:
Noted in the OP:

> I could not find any non-Regexp +object+ that did not raise TypeError; are there any?
> 
>     * If so, some correction is needed.
> 
>     * If not, should we say "Raises TypeError if +object+ is not a Regexp."?

string.c
BurdetteLamar: See my recent comment below.

peterzhu2118: You can just create your own object or use any non-string object that responds to `#=~` (e.g. nil):

```ruby
class Foo
  def =~(other)
    puts "calling =~ with #{other}"

    123
  end
end

p "hello" =~ Foo.new
# => calling =~ with hello
# => 123
p "hello" =~ nil
# => nil
```

BurdetteLamar: Added.

peterzhu2118 (approved):

PR #13328, opened 2025-05-14
[Bug #21336] Stop using `rb_ensure` during GC for hash iteration

kateinoigakukun:
bb180b87b43c45e17ff49735a26d7a188d5c8396 introduced some `rb_ensure` calls in `rb_hash_stlike_foreach{,with_replace}`. This caused "Cannot malloc during GC" on Wasm because:
1. `rb_ensure` calls `xmalloc` on Wasm to allocate jump buffers
2. `rb_hash_stlike_foreach{,with_replace}` is called during GC to mark hash values

All `rb_hash_stlike_foreach{,with_replace}` calls during GC are known not to raise exceptions, so we can safely use them without `rb_ensure`. This commit adds `_no_ensure` variants of these functions and uses them in the GC code to fix the Wasm issue.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -2675,10 +2675,10 @@ static void
 mark_hash(VALUE hash)
 {
     if (rb_hash_compare_by_id_p(hash)) {
-        rb_hash_stlike_foreach(hash, pin_key_mark_value, 0);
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
Closing in favor of https://github.com/ruby/ruby/pull/13329

PR #13339, opened 2025-05-14
ZJIT: Create more ergonomic type profiling API

tekknolagi:
- **Add ProfileOracle**
- **Use ProfileOracle in CCall code**
- **Use ProfileOracle in fixnum specialization code**
- **Use ProfileOracle in SendWithoutBlockDirect code**


zjit/src/hir.rs
```diff
@@ -2101,190 +2125,6 @@ mod union_find_tests {
     }
 }
 
-#[cfg(test)]
```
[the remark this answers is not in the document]
tekknolagi: Or to pass in a dummy iseq somehow

tekknolagi:
Eugh, going to fix some tests...

zjit/src/hir.rs
```diff
@@ -1717,8 +1723,35 @@ fn filter_translatable_calls(flag: u32) -> Result<(), ParseError> {
     Ok(())
 }
 
+#[derive(Debug)]
+struct ProfileOracle {
+    payload: &'static IseqPayload,
+    /// types is a map from insn_idx -> profiled type information at that idx
```
k0kubun, line 1729:
```suggestion
    /// types is a map from ISEQ insn_idx -> profiled type information at that HIR idx
```

Is this understanding ^ right?

Both "insn_idx" and "that idx" are a kind of an "instruction index" (both of which _could_ be called `InsnId`), so I found this comment confusing, even after looking at the type.

k0kubun: Well, I guess "that idx" may be pointing to "insn_idx", but then this comment doesn't explain what's `InsnId`. It'd be nice to clarify the difference between `usize` and `InsnId`.

tekknolagi: Ah, no, both refer to ISEQ -- thank you -- I should clarify

tekknolagi: At this point I want a type wrapper around iseq insn indices too but that is another PR

zjit/src/hir.rs
```diff
@@ -1717,8 +1723,35 @@ fn filter_translatable_calls(flag: u32) -> Result<(), ParseError> {
     Ok(())
 }
 
+#[derive(Debug)]
+struct ProfileOracle {
```
k0kubun, line 1727: It felt difficult to guess why this struct exits by looking at the name "profile oracle" and this definition. Could you add a comment above this struct to clarify the reason why this is necessary besides `IseqPayload::opnd_types`?

My understanding is that `IseqPayload::opnd_types` operates on ISEQ-level indexes whereas `ProfileOracle` translates them into HIR-level indexes, `InsnId`. But it's hard to get that idea from the name `ProfileOracle`. I'm also not familiar with how the word "oracle" is used in the compilers field in general.

k0kubun: Thanks for adding the comment :)

tekknolagi: Oracle largely means "mystical thing that holds the answers" (a reference to Greek mythology, I think), and is a fairly common term in CS theory. I am happy to swap out for a different name

zjit/src/hir.rs
k0kubun: sounds good. FYI it was called `IseqIdx` in YJIT.

tekknolagi:
Come to think of it, @k0kubun should I just have it profile at the top of every instruction? It'll only allocate if we actually recorded anything in the interpreter

k0kubun:
> should I just have it profile at the top of every instruction?

You mean adding a `zjit_` variant to every YARV instruction? Also, what made you think of it?

> It'll only allocate if we actually recorded anything in the interpreter

My assumption is that just calling a C (Rust) function and retrieving profiling metadata could be non-zero overhead for profiled runs. Remember we weren't even allowed to increment a global counter `vm_insns_count` on the interpreter on the default build, for example. It's somewhat mitigated by rewriting them back after num-profile cycles, but it's still overhead. At least I don't want to run a profiler on `nop` instruction (or any instruction that doesn't need to record anything).

tekknolagi:
Oh, no, sorry, I mean just in ISEQ->HIR translation. That will cause a read from `IseqPayload` in HIR creation but no other work (and, in particular, no other work in the interpreter).

tekknolagi:
Like lift the `profile_stack` so that it happens once per opcode.

k0kubun:
sounds good then :)

PR #13341, opened 2025-05-14
Add shape_id to RBasic under 32 bit

jhawthorn:
[[Feature #21353]](https://bugs.ruby-lang.org/issues/21353)

This adds a separate VALUE-sized shape_id field for 32-bit platforms. Allowing them to work much more similarly to how shapes always being available on 64-bit.

This does increase the size of objects by 4 bytes on 32 bit platforms, but shape access should be faster, more consistent, of the same size (32 bits) and much simpler.

### DIFF
--- a/include/ruby/internal/core/rbasic.h
+++ b/include/ruby/internal/core/rbasic.h
@@ -55,6 +55,12 @@ enum ruby_rvalue_flags {
     RVALUE_EMBED_LEN_MAX = RBIMPL_RVALUE_EMBED_LEN_MAX
 };
 
+#if (SIZEOF_VALUE < SIZEOF_UINT64_T)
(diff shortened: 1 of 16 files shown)


include/ruby/internal/core/rbasic.h
```diff
@@ -153,4 +166,13 @@ RBASIC_CLASS(VALUE obj)
     return RBASIC(obj)->klass;
 }
 
+static inline void
+RBASIC_RESET_FLAGS(VALUE obj)
```
eregon, line 170: Should this be in public headers (it is in the current state of this PR) or under `internal/**/*.h`?
IOW, do you think any C extension would need this, or just CRuby .c files?

byroot: This is meant to be internal only.

PR #13343, opened 2025-05-15
Ensure that forked process do not see invalid blocking operations.

ioquatix:


thread.c
```diff
@@ -2678,13 +2698,17 @@ thread_io_close_notify_all(struct rb_io *io)
 size_t
 rb_thread_io_close_interrupt(struct rb_io *io)
 {
+    if (io->fork_generation != GET_VM()->fork_gen) {
+
```
ntkme, line 2702: Is something missing here?

ioquatix: Ah thanks for calling that out, actually it's now handled later in `rb_io_blocking_operations`.

PR #13350, opened 2025-05-15
Fix `rb_gc_impl_copy_finalizer`

casperisfine:
- Add missing VM_LOCK
- Generate a new object_id when copying finalizers
- Dup the finalizer entry so adding a finalizer on a cloned object doesn't impact the source object.

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -2803,6 +2803,7 @@ rb_gc_impl_copy_finalizer(void *objspace_ptr, VALUE dest, VALUE obj)
 
     if (!FL_TEST(obj, FL_FINALIZE)) return;
 
+    int lev = rb_gc_vm_lock();
     if (RB_LIKELY(st_lookup(finalizer_table, obj, &data))) {
         table = (VALUE)data;
         st_insert(finalizer_table, dest, table);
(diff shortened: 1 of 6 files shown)


test/ruby/test_objectspace.rb
```diff
@@ -94,7 +94,7 @@ def test_count_objects
   end
 
   def test_finalizer
-    assert_in_out_err(["-e", <<-END], "", %w(:ok :ok :ok :ok), [])
+    assert_in_out_err(["-e", <<-END], "", %w(:ok :ok :ok), [])
```
casperisfine, line 97: My understanding is that this assertion was incorrect.

- We define 1 finalizer on the source object
- Clone the source object, so 1+1 => 2
- Add another finalizer on the copied object, so 2 + 1 => 3.

gc/default/default.c
```diff
@@ -2832,16 +2835,22 @@ run_final(rb_objspace_t *objspace, VALUE zombie)
     if (FL_TEST_RAW(zombie, FL_FINALIZE)) {
         FL_UNSET(zombie, FL_FINALIZE);
         st_data_t table;
+        unsigned int lev = rb_gc_vm_lock();
```
peterzhu2118, line 2838: I think we should do the locking in `finalize_list` instead. We lock the VM in `finalize_list` to add the slot back to the freelist, so we're just doing locking and unlocking twice for every zombie object instead of once total, which would be bad for performance.

casperisfine: Good catch, I just had to move the lock up by one line.

peterzhu2118 (approved):

PR #13351, opened 2025-05-15
Remove dependency on bits.h in default.c when BUILDING_MODULAR_GC

peterzhu2118:
We can assume that the compiler will have __builtin_clzll so we can implement nlz_int64 using that.

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -15,7 +15,11 @@
 # include <sys/user.h>
 #endif
 
-#include "internal/bits.h"
+#ifdef BUILDING_MODULAR_GC
+# define nlz_int64(x) (x == 0 ? 64 : (unsigned int)__builtin_clzll((unsigned long long)x))
+#else
+# include "internal/bits.h"
+#endif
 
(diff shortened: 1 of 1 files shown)


nobu:
> We can assume that the compiler will have __builtin_clzl

Why?

peterzhu2118:
Because modular GC is only supported on macOS and Linux and clang and GCC both have `__builtin_clzl`.

peterzhu2118:
> And other platforms may have dynamic loading feature, e.g., LoadLibrary on Windows.

That is true, however, we don't currently have any plans to support any platforms other than macOS and Linux for modular GC since we don't have the infrastructure to develop on it. This change only affects `default.c` when it is built as a modular GC. We may support more platforms when we make it not experimental.

> BTW, isn't the second commit to newobj_init applicable regardless the first one for bits.h?

This change was needed because `internal/bits.h` pulled in `internal/compilers.h`. Since we don't include `internal/bits.h` when building `default.c` as a modular GC, this fails to compile on GCC since it doesn't define `__has_feature`.

PR #13358, opened 2025-05-16
Prevent enabling yjit when zjit enabled

composerinteralia:
`ruby --yjit --zjit` already warns and exits, but it was still possible to enable both with 
`ruby --zjit -e 'RubyVM:YJIT.enable`.

This commit prevents that with a warning and an early return. (We could also exit, but that seems a bit unfriendly once we're already running the program.)

### DIFF
--- a/yjit.c
+++ b/yjit.c
@@ -29,6 +29,7 @@
 #include "iseq.h"
 #include "ruby/debug.h"
 #include "internal/cont.h"
+#include "zjit.h"
 
 // For mmapp(), sysconf()
 #ifndef _WIN32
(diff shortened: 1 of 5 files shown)


composerinteralia:
Oops, I think I know what I did wrong. I'll find a way to fix it and reopen this later.

yjit.c
```diff
@@ -29,6 +29,7 @@
 #include "iseq.h"
 #include "ruby/debug.h"
 #include "internal/cont.h"
+#include "zjit.h"
```
composerinteralia, line 32: For `rb_zjit_enabled_p` in `yjit.rb`

XrXr (approved):
Thanks!

nobu:
Isn't the variable that indicates which JIT is enabled better, than checking all combinations?

composerinteralia:
Ah, I guess it's possible to do `ruby --rjit -e 'RubyVM::YJIT.enable'` right now too. I didn't know a variable like that existed, but I'll look for it and update this.

composerinteralia:
I'm happy to update it, although I didn't notice a variable like that. Does that variable already exist, and do you remember what it's named?

zzak:
@composerinteralia I think nobu is proposing to have a function that returns _which JIT_ is enabled than _is JIT enabled_.

PR #13360, opened 2025-05-16
[Bug #21313] Handle `it` in rescue and ensure blocks.

nicholasdower:
https://bugs.ruby-lang.org/issues/21313

The following is crashing for me:

```shell
ruby --yjit --yjit-call-threshold=1 -e '1.tap { raise rescue p it }'

thread '<unnamed>' panicked at ./yjit/src/codegen.rs:2402:14:
...
```

It seems `it` sometimes points to the wrong value:

```shell
ruby -e '1.tap { raise rescue p it }'
false

ruby -e '1.tap { begin; raise; ensure; p it; end } rescue nil'
false
```

But only when `$!` is set:

```shell
ruby -e '1.tap { begin; nil; ensure; p it; end }'
1

ruby -e '1.tap { begin; nil; rescue; ensure; p it; end }'
1

ruby -e '1.tap { begin; raise; rescue; ensure; p it; end }'
1
```

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -9640,7 +9640,19 @@ pm_compile_node(rb_iseq_t *iseq, const pm_node_t *node, LINK_ANCHOR *const ret, 
         // -> { it }
         //      ^^
         if (!popped) {
-            PUSH_GETLOCAL(ret, location, scope_node->local_table_for_iseq_size, 0);
+            pm_scope_node_t *current_scope_node = scope_node;
+            int level = 0;
+
+            while (current_scope_node) {
+                if (current_scope_node->parameters && PM_NODE_TYPE_P(current_scope_node->parameters, PM_IT_PARAMETERS_NODE)) {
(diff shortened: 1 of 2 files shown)


XrXr (approved):
This looks good to me. Another test case related to the bug:

`"a".then{ /#{it}/o } #=> /false/` on prism without this patch.

This a third situation where the block housing `it` makes a child scope, so looping up to find the parameter is necessary.

nicholasdower:
> This looks good to me. Another test case related to the bug:
> 
> `"a".then{ /#{it}/o } #=> /false/` on prism without this patch.
> 
> This a third situation where the block housing `it` makes a child scope, so looping up to find the parameter is necessary.

Thanks for reviewing @XrXr. I've added a commit with the test case you suggested.

kddnewton (approved):
We could probably make scope node better to track this but this is a good fix

PR #13363, opened 2025-05-16
ZJIT: Allow DCE to remove some CCalls

tekknolagi:
Add is_elidable field that signals that there would be no discernible
effect if the call to the method were removed. The default is false.

### DIFF
--- a/zjit/src/cruby_methods.rs
+++ b/zjit/src/cruby_methods.rs
@@ -26,6 +26,8 @@ pub struct FnProperties {
     pub leaf: bool,
     /// What Type the C function returns
     pub return_type: Type,
+    /// Whether it's legal to remove the call if the result is unused
+    pub elidable: bool,
 }
 
 impl Annotations {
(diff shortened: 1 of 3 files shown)


zjit/src/cruby_methods.rs
```diff
@@ -26,6 +26,8 @@ pub struct FnProperties {
     pub leaf: bool,
     /// What Type the C function returns
     pub return_type: Type,
+    /// Whether it's legal to remove the call if the result is unused
+    pub is_elidable: bool,
```
k0kubun, line 30: "is elidable" doesn't seem to explain the actual property that makes it elidable, so things like "pure" or "side effect free" might clarify the nature better. But I'm sure sure if I like them more than "is elidable" 🤷‍♂️

k0kubun: It can be just `elidable` though. `leaf` is not `is_leaf`. `annotate!(..., elidable);` feels more fluent to me. But it could be just me.

tekknolagi: Ok, here's another side effect: a hypothetical HIR instruction that does a method lookup. This is actually a complicated operation, potentially, if it uses the global method lookup cache. This is a subtle side effect that we might not care about from a "should we keep it around" perspective but *does* require special care in code generation.

PR #13365, opened 2025-05-16
[DOC] Tweaks for String#byteindex

BurdetteLamar:


string.rb
```diff
@@ -342,6 +342,7 @@
 #
 # - #=~: Returns the index of the first substring that matches a given
 #   Regexp or other object; returns +nil+ if no match is found.
+# - #byteindex: Returns the byte indexof the first occurrence of a given substring.
```
peterzhu2118, line 345: I think there should be a space between these two words.

```suggestion
# - #byteindex: Returns the byte index of the first occurrence of a given substring.
```

peterzhu2118 (approved):

PR #13366, opened 2025-05-16
Add `String#ensure_suffix`

MatheusRich:
This method returns self if the string ends with the given suffix. It returns a new string with the suffix appended, otherwise.

    "Hell".ensure_suffix("o!")   # => "Hello!"
    "Hello!".ensure_suffix("o!") # => "Hello!"

---

In my microbenchmarks writing it in C seems to be faster than in Ruby, but happy to change the implementation if the team prefers that.

### DIFF
--- a/string.c
+++ b/string.c
@@ -11307,6 +11307,21 @@ rb_str_delete_suffix(VALUE str, VALUE suffix)
     return rb_str_subseq(str, 0, RSTRING_LEN(str) - suffixlen);
 }
 
+/*
+ *  call-seq:
+ *    ensure_suffix(suffix) -> self or new_string
+ *
+ *  :include: doc/string/ensure_suffix.rdoc
+ *
+ */
+static VALUE
+rb_str_ensure_suffix(VALUE str, VALUE suffix)
+{
+    if (rb_str_end_with(1, &suffix, str)) return str;
+
+    return rb_str_plus(str, suffix);
+}
+
 void
 rb_str_setter(VALUE val, ID id, VALUE *var)
 {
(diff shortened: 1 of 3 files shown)


string.c
```diff
@@ -11307,6 +11307,21 @@ rb_str_delete_suffix(VALUE str, VALUE suffix)
...
+ *
+ *  :include: doc/string/ensure_suffix.rdoc
+ *
+ */
+static VALUE
+rb_str_ensure_suffix(VALUE str, VALUE suffix)
+{
+    if (rb_str_end_with(1, &suffix, str)) return str;
+
+    return rb_str_plus(str, suffix);
```
MatheusRich, line 11322: I've used `+` here because I don't know if it is possible or desirable to use string interpolation in this case.

string.c
```diff
@@ -11307,6 +11307,21 @@ rb_str_delete_suffix(VALUE str, VALUE suffix)
     return rb_str_subseq(str, 0, RSTRING_LEN(str) - suffixlen);
 }
 
+/*
+ *  call-seq:
+ *    ensure_suffix(suffix) -> self or new_string
+ *
+ *  :include: doc/string/ensure_suffix.rdoc
```
nobu, line 11314: Why separate the file?

MatheusRich: No particular reason. I just follow what I saw in other methods. I can in-line the doc if that's better

nobu: It is for sharing the same description between *multiple* methods.

MatheusRich:
@nobu looks like [matz is cool with this](https://github.com/ruby/ruby/pull/13366). Anything I can do to move this forward?

PR #13375, opened 2025-05-19
Update fields count correctly when namespaces are disabled

tagomoris:
When namespace is disabled, Class's `rb_shape_obj_too_complex_p(klass)` should be false in most cases. In other words, when a class is under shape control, its prime-classext is always writable.
This change is to update max_iv_count appropriately when namespace is disabled.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -3261,8 +3261,23 @@ rb_gc_mark_children(void *objspace, VALUE obj)
       }
 
       case T_OBJECT: {
+        attr_index_t fields_count = ROBJECT_FIELDS_COUNT(obj);
+
         if (rb_shape_obj_too_complex_p(obj)) {
             gc_mark_tbl_no_pin(ROBJECT_FIELDS_HASH(obj));
+
+            if (fields_count) {
+                VALUE klass = RBASIC_CLASS(obj);
+
+                // Skip updating max_iv_count if the prime classext is not writable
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -3261,8 +3261,23 @@ rb_gc_mark_children(void *objspace, VALUE obj)
...
+
+                // Skip updating max_iv_count if the prime classext is not writable
+                // because GC context doesn't provide information about namespaces.
+                if (RCLASS_PRIME_CLASSEXT_WRITABLE_P(klass)) {
+                    // Increment max_iv_count if applicable, used to determine size pool allocation
+                    if (RCLASS_MAX_IV_COUNT(klass) < fields_count) {
+                        RCLASS_SET_MAX_IV_COUNT(klass, fields_count);
+                    }
+                }
+            }
```
jhawthorn, line 3280: I'm not sure this is useful to do for too_complex objects since they are using a hash anyways.

cc @tenderlove

tenderlove:
I think `RCLASS_SET_PRIME_CLASSEXT_WRITABLE` is always false when namespaces are disabled.  I made [a PR to demonstrate](https://github.com/ruby/ruby/pull/13385) but I'm not sure if my fix is correct.

tagomoris:
#13385 is the better fix

PR #13380, opened 2025-05-20
Fix Symbol#to_proc (rb_sym_to_proc) to be ractor safe

luke-gruber:
In non-main ractors, don't use `sym_proc_cache`. It is not thread-safe to add to this array without a lock and also it leaks procs from one ractor to another. Instead, we create a new proc each time. If this results in poor performance we can come up with a solution later.

Fixes [Bug #21354]

### DIFF
--- a/proc.c
+++ b/proc.c
@@ -22,6 +22,7 @@
 #include "method.h"
 #include "iseq.h"
 #include "vm_core.h"
+#include "ractor_core.h"
 #include "yjit.h"
 
 const rb_cref_t *rb_vm_cref_in_context(VALUE self, VALUE cbase);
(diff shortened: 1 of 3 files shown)


proc.c
```diff
@@ -1538,23 +1539,28 @@ rb_sym_to_proc(VALUE sym)
     long index;
     ID id;
 
-    if (!sym_proc_cache) {
-        sym_proc_cache = rb_ary_hidden_new(SYM_PROC_CACHE_SIZE * 2);
-        rb_vm_register_global_object(sym_proc_cache);
-        rb_ary_store(sym_proc_cache, SYM_PROC_CACHE_SIZE*2 - 1, Qnil);
-    }
+    bool is_main_ractor = ruby_single_main_ractor || rb_ractor_main_p_();
```
byroot, line 1542: Why not:

```suggestion
    bool is_main_ractor = rb_ractor_main_p();
```


?

proc.c
```diff
@@ -1538,23 +1539,28 @@ rb_sym_to_proc(VALUE sym)
...
+            return RARRAY_AREF(sym_proc_cache, index + 1);
+        }
+        else {
+            proc = sym_proc_new(rb_cProc, ID2SYM(id));
+            RARRAY_ASET(sym_proc_cache, index, sym);
+            RARRAY_ASET(sym_proc_cache, index + 1, proc);
+            return proc;
+        }
+    } else {
+        return sym_proc_new(rb_cProc, ID2SYM(id));
```
byroot, line 1563: Rather than to not use the cache at all, could we make these procs shareable?

proc.c
luke-gruber: For some reason, `rb_ractor_main_p_()` has a `VM_ASSERT` for multiple ractors running.

proc.c
luke-gruber: If we use the cache, we would need to lock around the cache, though. That would make even main `Symbol#to_proc` slower and I didn't want it to impact the non-ractor code path. I agree that we could make improvements later.

byroot: Not if the entry fits in a single `VALUE` as suggested in my other comment. I suspect it wouldn't be hard to implement.

proc.c
byroot: Note how my suggestion is `rb_ractor_main_p`, not `rb_ractor_main_p_`.

proc.c
```diff
@@ -1538,23 +1539,28 @@ rb_sym_to_proc(VALUE sym)
...
+            rb_vm_register_global_object(sym_proc_cache);
+            rb_ary_store(sym_proc_cache, SYM_PROC_CACHE_SIZE*2 - 1, Qnil);
+        }
+        if (RARRAY_AREF(sym_proc_cache, index) == sym) {
+            return RARRAY_AREF(sym_proc_cache, index + 1);
+        }
+        else {
+            proc = sym_proc_new(rb_cProc, ID2SYM(id));
+            RARRAY_ASET(sym_proc_cache, index, sym);
+            RARRAY_ASET(sym_proc_cache, index + 1, proc);
```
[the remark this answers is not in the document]
luke-gruber: Yeah, good idea. Maybe a separate PR?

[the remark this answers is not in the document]
byroot: Dunno, a separate PR would undo most of what you did in this one. It feels simpler to directly make the cache thread-safe IMO.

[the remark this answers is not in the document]
luke-gruber: It changes the language semantics, though.`:inspect.to_proc.instance_variable_set("@some", "value")` today makes it so that `:inspect.to_proc.instance_variable_get("@some") == "value"`. If we freeze the procs, it's a language change imo and would need to be discussed, no?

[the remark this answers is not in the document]
byroot: Hum, good point

byroot (approved):

PR #13385, opened 2025-05-20
Add assertion for RCLASS_SET_PRIME_CLASSEXT_WRITABLE

tenderlove:
When classes are booted, they should all be writeable unless namespaces are enabled.  This commit adds an assertion to ensure that classes are writable.

### DIFF
--- a/class.c
+++ b/class.c
@@ -702,7 +702,7 @@ class_alloc(VALUE flags, VALUE klass)
     RCLASS_PRIME_NS((VALUE)obj) = ns;
     // Classes/Modules defined in user namespaces are
     // writable directly because it exists only in a namespace.
-    RCLASS_SET_PRIME_CLASSEXT_WRITABLE((VALUE)obj, (NAMESPACE_USER_P(ns) || !rb_namespace_available()) ? true : false);
+    RCLASS_SET_PRIME_CLASSEXT_WRITABLE((VALUE)obj, !rb_namespace_available() || NAMESPACE_USER_P(ns) ? true : false);
 
     RCLASS_SET_ORIGIN((VALUE)obj, (VALUE)obj);
     RCLASS_SET_REFINED_CLASS((VALUE)obj, Qnil);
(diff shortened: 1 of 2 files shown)


class.c
```diff
@@ -702,7 +702,7 @@ class_alloc(VALUE flags, VALUE klass)
     RCLASS_PRIME_NS((VALUE)obj) = ns;
     // Classes/Modules defined in user namespaces are
     // writable directly because it exists only in a namespace.
-    RCLASS_SET_PRIME_CLASSEXT_WRITABLE((VALUE)obj, NAMESPACE_USER_P(ns) ? true : false);
+    RCLASS_SET_PRIME_CLASSEXT_WRITABLE((VALUE)obj, (NAMESPACE_USER_P(ns) || !rb_namespace_available()) ? true : false);
```
tagomoris, line 705: How about checking `rb_namespace_available()` at first? Checking `NAMESPACE_USER_P()` prior to `rb_namespace_available()` looks curious to me because types of namespace are nonsense when namespace is disabled.
```suggestion
    RCLASS_SET_PRIME_CLASSEXT_WRITABLE((VALUE)obj, !rb_namespace_available() || NAMESPACE_USER_P(ns) ? true : false);
```

gc.c
```diff
@@ -3280,12 +3280,14 @@ rb_gc_mark_children(void *objspace, VALUE obj)
...
             if (RCLASS_PRIME_CLASSEXT_WRITABLE_P(klass)) {
-                VM_ASSERT(rb_shape_obj_too_complex_p(klass));
                 // Increment max_iv_count if applicable, used to determine size pool allocation
                 if (RCLASS_MAX_IV_COUNT(klass) < fields_count) {
                     RCLASS_SET_MAX_IV_COUNT(klass, fields_count);
                 }
             }
+            else {
+                VM_ASSERT(rb_shape_obj_too_complex_p(klass));
+            }
```
tagomoris, line 3290: This else clause looks much smarter than #13375

PR #13389, opened 2025-05-21
Use rb_inspect for Ractor error

jhawthorn:
Previously the object was used directly, which calls `to_s` if defined. We should use rb_inspect to get a value suitable for display to the programmer. Looking through the codebase in other places it seems we've preferred just printing the class, but it's _really_ helpful to get more details in these shareable errors.

@tenderlove found this using 

```
require "socket"
Ractor.make_shareable Addrinfo.new "\x10\x02\x14\xE9\xE0\x00\x00\xFB\x00\x00\x00\x00\x00\x00\x00\x00".b
```

Which used to raise

```
<internal:ractor>:832:in 'Ractor.make_shareable': string contains null byte (ArgumentError)
	from test.rb:5:in '<main>'
```

But now appropriately raises

```
ruby test2.rb
<internal:ractor>:832:in 'Ractor.make_shareable': can not make shareable object for #<Addrinfo: unknown address family 528> (Ractor::Error)
        from test2.rb:4:in '<main>'
```

(or probably something better if you're on a mac and that addrinfo binary string makes sense to it)


tenderlove (approved):

ractor.c
```diff
@@ -3163,7 +3163,7 @@ make_shareable_check_shareable(VALUE obj)
             return traverse_cont;
         }
         else {
-            rb_raise(rb_eRactorError, "can not make shareable object for %"PRIsVALUE, obj);
+            rb_raise(rb_eRactorError, "can not make shareable object for %"PRIsVALUE, rb_inspect(obj));
```
cremno, line 3166:
```c
rb_raise(rb_eRactorError, "can not make shareable object for %+"PRIsVALUE, obj);
```

https://github.com/ruby/ruby/blob/97e774b95d5233f801666f873c27de64684b0cf0/doc/extension.rdoc?plain=1#L200-L203

jhawthorn: Thanks!

PR #13396, opened 2025-05-21
Fix an off-by-one error in the `rb_numparam_id_p`

mame:
[Bug #21356]


mame:
@nobu beats me. #13395

PR #13399, opened 2025-05-21
ZJIT: More type level docs in zjit::hir [DOC]

XrXr:
Given `InsnId` is at the top of the file and everywhere, hopefully this
will help first time readers.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1,4 +1,4 @@
-//! High level intermediary representation.
+//! High-level intermediary representation (IR) in static single-assignment (SSA) form.
 
 // We use the YARV bytecode constants which have a CRuby-style name
 #![allow(non_upper_case_globals)]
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -35,6 +36,7 @@ impl std::fmt::Display for InsnId {
     }
 }
 
+/// Reference to a [`Block`]. It's an index to help with circular reference.
```
tekknolagi, line 39: (as above)

zjit/src/hir.rs
```diff
@@ -530,6 +535,7 @@ impl std::fmt::Display for Insn {
     }
 }
 
+/// A basic block in a [`Function`].
```
tekknolagi, line 538: We should mention that it's an extended basic block, which might otherwise people if they see multiple control flow operations in it

zjit/src/hir.rs
```diff
@@ -20,6 +20,7 @@ use std::{
 };
 use crate::hir_type::{Type, types};
 
+/// An index of an [`Insn`] in a [`Function`]. See also: [`Function::find`].
```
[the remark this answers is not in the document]
XrXr: 👍🏼 . I've mentioned that it's basically a pointer, but didn't mention the "everything is index based" part since it feels a bit vague.

PR #13402, opened 2025-05-21
Avoid adding garbage objects to id2ref table

jhawthorn:
We didn't confirm that this is causing the CI failures, but this is our guess.

The id2ref table is built calling `disable_gc_no_rest`, which won't finish on ongoing GC, so it seems reasonable that that will end up with garbage objects inside the id2ref table.

cc @byroot


jhawthorn:
Hmm. Based on the CI failure I guess this wasn't it

PR #13411, opened 2025-05-22
Turn `rb_classext_t.fields` into a T_OBJECT

casperisfine:
This aims to solve two problems.

First, it solves the problem of namspaced classes having a single `shape_id`. Now each namespaced classext
has a `T_OBJECT` that can hold the namespace specific shape.

Second, it open the door to later make class instance variable writes atomics, hence be able to read class variables without locking the VM.
In the future, in multi-ractor mode, we can do the write on a copy of the `fields_obj` and then atomically swap it.

Considerations:

  - T_OBJECT isn't ideal, it probably should be a `T_IMEMO` but this require a very large refatoring.
  - Right now the `RClass` shape_id is always synchronized, but with namespace we should likely mark classes that have multiple namespace with `INVALID_SHAPE`.


tenderlove:
Just to follow up because @byroot asked me to look at the JIT side of things.  I don't think this impacts Z/YJIT because we only emit special machine code if the object is a T_OBJECT, [everything else just uses `rb_ivar_get`](https://github.com/ruby/ruby/blob/7b10660974dcdd7882a48a62cfc2a32c4a17aa85/yjit/src/codegen.rs#L2882-L2903).

byroot:
Ah thanks! I'll dig more into why the JITs are failing tomorrow then.

internal/imemo.h
```diff
@@ -257,4 +258,60 @@ MEMO_V2_SET(struct MEMO *m, VALUE v)
...
+            VALUE *ptr;
+        } external;
+        struct {
+            st_table *table; // TODO: embed
+        } complex;
+    } as;
+};
+
+#define OBJ_FIELD_EXTERNAL IMEMO_FL_USER0
+#define OBJ_FIELD_COMPLEX  IMEMO_FL_USER1
```
peterzhu2118: It seems like sometimes you check the shape for too complex, and sometimes you check `OBJ_FIELD_COMPLEX` flag. Isn't this duplicate information?

casperisfine: Indeed. I can probably get rid of that.

PR #13417, opened 2025-05-22
Update RDoc to 6.14.0 and adopt latest RDoc configs

st0012:
The first commit syncs RDoc to version 6.14.0 (latest release).
The second commit adopts some RDoc configs added to master after 3.4's release. 
- `autolink_excluded_words` reduces unintentional autolinks to make the text easier to read
- `canonical_root` will set `https://docs.ruby-lang.org/en/3.4/[page_path]` as the page's canonical root in html head


k0kubun (approved):

.rdoc_options
```diff
@@ -11,3 +11,13 @@ rdoc_include:
...
+
+autolink_excluded_words:
+- Class
+- Method
+- Module
+- Process
+- RDoc
+- Set
+
+canonical_root: https://docs.ruby-lang.org/en/3.4
```
p8, line 23: I think this should be set to https://docs.ruby-lang.org/en/master` as we want master to always be the version that shows up first in search results?

st0012: I don’t understand. This only applies to Ruby 3.4’s doc, not master’s. Should all versions set it to master doc’s url?

PR #13418, opened 2025-05-22
change Net::HTTPResponse.value to return self

max-power:
It's kind of a strange behaviour now. you call .value and get nil in a successful request, which contradicts the name. returning self would make it more useful especially with chaining.


hsbt:
Could you send this change to https://github.com/ruby/net-http/ instead of this?

PR #13421, opened 2025-05-22
Fix `make test-bundler-parallel` and `make test-bundled-gems` for Ruby 3.4

hsbt:
https://github.com/rubygems/rubygems/commit/d74fa0fdda481043112f9de179be572b74a711ed broke our test suites. 

https://github.com/ruby/actions/actions/runs/15194309897

/cc @deivid-rodriguez 

I have no idea to fix that issues with correct way yet.

### DIFF
--- a/spec/bundler/bin/parallel_rspec
+++ b/spec/bundler/bin/parallel_rspec
@@ -1,7 +1,0 @@
-#!/usr/bin/env ruby
-# frozen_string_literal: true
-
-require_relative "../bundler/support/setup"
-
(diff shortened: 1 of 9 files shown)


deivid-rodriguez:
I'll have a look. Just for my understanding, this is correctly passing in the master branch, but only causing issues in the stable branch, correct?

hsbt:
>this is correctly passing in the master branch, but only causing issues in the stable branch, correct?

Yes. It's only happened with `ruby_3_4` branch.

PR #13430, opened 2025-05-23
ZJIT: Add SideExit HIR opcode and use it instead of failing to compile

tekknolagi:
- **ZJIT: Side-exit into the interpreter on unknown opcodes**
- **ZJIT: Side-exit into the interpreter on unknown opt_newarray_send**
- **ZJIT: Side-exit into the interpreter on unknown call types**


k0kubun (approved):
:+1:

k0kubun:
I just merged the other PR. It seems like you need to rebase it to fix the conflict.

tekknolagi:
Can do!

tekknolagi:
launchable is timing out causing all omnibus to stall -.-

PR #13431, opened 2025-05-23
ractor_wakeup was broken when compiled with USE_RUBY_DEBUG_LOG

luke-gruber:
The `ractor_wakeup` function takes an optional `th` argument, so it can be NULL. There is a macro call to RUBY_DEBUG_LOG that dereferences `th` without checking if it's NULL first. To fix this, we never dereference `th` in this macro call.

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -576,9 +576,8 @@ ractor_wakeup(rb_ractor_t *r, rb_thread_t *th /* can be NULL */, enum rb_ractor_
 {
     ASSERT_ractor_locking(r);
 
-    RUBY_DEBUG_LOG("r:%u wait_by:%s -> wait:%s wakeup:%s",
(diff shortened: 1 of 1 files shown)


ractor.c
```diff
@@ -576,9 +576,8 @@ ractor_wakeup(rb_ractor_t *r, rb_thread_t *th /* can be NULL */, enum rb_ractor_
 {
     ASSERT_ractor_locking(r);
 
-    RUBY_DEBUG_LOG("r:%u wait_by:%s -> wait:%s wakeup:%s",
+    RUBY_DEBUG_LOG("r:%u wait:%s wakeup:%s",
                    rb_ractor_id(r),
-                   wait_status_str(th->ractor_waiting.wait_status),
```
jhawthorn, line 581: 👍 Per the comment above `/* can be NULL */`

PR #13433, opened 2025-05-23
Fix debug assertion in `thread_sched_enq` that results in infinite loop

luke-gruber:
Running a debug build, the following code can result in an infinite loop:

```ruby
rs = 100.times.map do
  Ractor.new do
    ths = 100.times.map do
      Thread.new do
        sleep 2
      end
    end
    ths.each(&:join)
  end
end
rs.each(&:take)
```

I was debugging another scheduler issue and ran into this infinite loop more than once. It's because the assertion is not running with the proper lock taken. Once we add the lock, it works well. NOTE: this lock is taken in the proper hierarchy order, which is:
1) thread_sched_lock
2) ractor_sched_lock

This ordering is enforced, for ex, in `thread_sched_setup_running_threads`


luke-gruber:
This debug assertion has been commented out. I'm closing this PR.

PR #13436, opened 2025-05-24
compile.c: Handle anonymous variables in `outer_variable_cmp`

byroot:
[[Bug #21370]](https://bugs.ruby-lang.org/issues/21370)

No idea if this is the correct way to fix it. I assume `it` is indeed meant to be anonymous, but perhaps it shouldn't?

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -13378,6 +13378,13 @@ outer_variable_cmp(const void *a, const void *b, void *arg)
 {
     const struct outer_variable_pair *ap = (const struct outer_variable_pair *)a;
     const struct outer_variable_pair *bp = (const struct outer_variable_pair *)b;
+
+    if (!ap->name) {
+        return -1;
+    } else if (!bp->name) {
+        return 1;
+    }
+
     return rb_str_cmp(ap->name, bp->name);
 }
 
--- a/test/ruby/test_iseq.rb
+++ b/test/ruby/test_iseq.rb
@@ -859,6 +859,25 @@ class TestISeq < Test::Unit::TestCase
     end
   end
 
+  def test_serialize_anonymous_outer_variables
+    iseq = RubyVM::InstructionSequence.compile(<<~'RUBY')
+      obj = Object.new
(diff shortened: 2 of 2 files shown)


tenderlove:
I think your patch seems alright.  I was going to suggest the following:

```diff
diff --git a/compile.c b/compile.c
index 7eb953203c..9bfff6961f 100644
--- a/compile.c
+++ b/compile.c
@@ -13366,6 +13366,9 @@ static enum rb_id_table_iterator_result
 store_outer_variable(ID id, VALUE val, void *dump)
 {
     struct outer_variable_list *ovlist = dump;
+    if (id & ID_INTERNAL) {
+        return ID_TABLE_CONTINUE;
+    }
     struct outer_variable_pair *pair = &ovlist->pairs[ovlist->num++];
     pair->id = id;
     pair->name = rb_id2str(id);
@@ -13399,7 +13402,7 @@ ibf_dump_outer_variables(struct ibf_dump *dump, const rb_iseq_t *iseq)
         struct outer_variable_list *ovlist = RB_ALLOCV(buff, buffsize);
         ovlist->num = 0;
         rb_id_table_foreach(ovs, store_outer_variable, ovlist);
-        ruby_qsort(ovlist->pairs, size, sizeof(struct outer_variable_pair), outer_variable_cmp, NULL);
+        ruby_qsort(ovlist->pairs, ovlist->num, sizeof(struct outer_variable_pair), outer_variable_cmp, NULL);
         for (size_t i = 0; i < size; ++i) {
             ID id = ovlist->pairs[i].id;
             ID val = ovlist->pairs[i].val;
```

But then maybe the iseqs wouldn't round-trip correctly.

I assume we're sorting these ids by name because when you load the iseq in a different process the id will change and the "name" is the only real stable key we have.  I'm not sure what impact that has on anonymous locals though.

byroot:
> I should assert the code is still functional once roundtriped.

I did this today with @etiennebarrie, seem to work fine.

PR #13437, opened 2025-05-24
Fix `IO#close` use-after-free in `blocking_operation_wait` and `fiber_interrupt` hooks.

ioquatix:
This PR fixes memory safety and interrupt handling issues discovered while debugging `IO#close` across fibers.

## Fix interrupt handling affecting `rb_io_blocking_operation_exit`

If `rb_io_blocking_operation_exit` executes with pending interrupts, user code that checks interrupts can exit unexpectedly. Instead, move the exception check into the `EC_PUSH_TAG` block so that by the time we reach `rb_io_blocking_operation_exit` the exception is already propagating and it behaves more like `ensure`.

Without this change, the following program can hang:

```ruby
#!/usr/bin/env ruby

require_relative 'lib/async'

Async do
  r, w = IO.pipe
  
  read_thread = Thread.new do
    Thread.current.report_on_exception = false
    r.read(5)
  end
  
  # Wait until read_thread blocks on I/O
  Thread.pass until read_thread.status == "sleep"
  
  close_task = Async do
    r.close
  end
  
  close_task.wait
  begin
    read_thread.join
  rescue => error
    puts "Caught exception: #{error.class} - #{error.message}"
  end
end
```

This is a follow up to https://github.com/ruby/ruby/pull/12839

## Fix use-after-free in `blocking_operation_wait` hook

`IO#close` could trigger stack-allocated memory access after the function returned. The new implementation uses a heap-allocated `BlockingOperation` instance which also exposes a C interface for (safe) native work pool execution.

https://bugs.ruby-lang.org/issues/21198

### Technical Implementation

The blocking operation system now provides:
- **Thread-safe cancellation** with atomic state transitions
- **Cross-platform atomic operations** supporting Windows, macOS, and Linux  
- **Anonymous BlockingOperation class** hidden from public API
- **Comprehensive state machine** for operation lifecycle management
- **GC-safe object management** with proper root registration

### API Improvements

- `rb_fiber_scheduler_blocking_operation_extract` - Safely extract operation data while holding GVL
- `rb_fiber_scheduler_blocking_operation_execute` - Execute in thread pools without GVL concerns  
- `rb_fiber_scheduler_blocking_operation_cancel` - Thread-safe cancellation support
- Consistent naming with `RB_FIBER_SCHEDULER_BLOCKING_OPERATION_STATUS_*` enum values

### Testing

Enhanced tests covering:
- IO#close across fibers 
- Memory safety verification
- Blocking operation cancellation
- Interrupt handling scenarios

### DIFF
--- a/include/ruby/fiber/scheduler.h
+++ b/include/ruby/fiber/scheduler.h
@@ -394,10 +394,53 @@ VALUE rb_fiber_scheduler_io_close(VALUE scheduler, VALUE io);
  */
 VALUE rb_fiber_scheduler_address_resolve(VALUE scheduler, VALUE hostname);
 
+// The state of the blocking operation execution.
 struct rb_fiber_scheduler_blocking_operation_state {
     void *result;
     int saved_errno;
 };
+
+// The opaque handle for the blocking operation.
+typedef struct rb_fiber_scheduler_blocking_operation rb_fiber_scheduler_blocking_operation_t;
+
+/**
+ * Extract the blocking operation handle from a BlockingOperationRuby object.
+ *
+ * This function safely extracts the opaque handle from a BlockingOperation VALUE
+ * while holding the GVL. The returned pointer can be passed to worker threads
+ * and used with rb_fiber_scheduler_blocking_operation_execute.
+ *
+ * @param[in] self The BlockingOperation VALUE to extract from
+ * @return The opaque struct pointer on success, NULL on error
+ * @note Experimental.
+ */
+rb_fiber_scheduler_blocking_operation_t *rb_fiber_scheduler_blocking_operation_extract(VALUE self);
+
+/**
+ * Execute blocking operation from handle (GVL not required).
+ *
+ * This function executes a blocking operation using the opaque handle
+ * obtained from rb_fiber_scheduler_blocking_operation_extract.
+ * It can be called from native threads without holding the GVL.
+ *
+ * @param[in] blocking_operation The opaque handle.
+ * @return 0 on success, -1 on error.
+ * @note Experimental. Can be called from any thread without holding the GVL
+ */
+int rb_fiber_scheduler_blocking_operation_execute(rb_fiber_scheduler_blocking_operation_t *blocking_operation);
+
+/**
+ * Cancel a blocking operation.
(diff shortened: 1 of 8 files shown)


samuel-williams-shopify:
Absolutely! The IO#close issue was the root cause that led us to discover all these problems. Here's a better title:

## New Title:
**Fix IO#close use-after-free in blocking_operation_wait and interrupt handling**

## New Description:

This PR fixes critical memory safety and interrupt handling issues discovered while debugging IO#close operations across fibers.

### Critical Fixes

**Use-After-Free Bug in `blocking_operation_wait`**: Fixed a critical memory safety issue where IO#close operations could trigger stack-allocated memory access after the calling function returned. The existing `blocking_operation_wait` scheduler hook was unsafe for native work pools. The new implementation uses heap-allocated Ruby objects with proper lifetime management.

**Interrupt Handling**: Fixed `rb_io_blocking_operation_exit` to not execute with pending interrupts, preventing unexpected exits during IO#close operations when user code checks for interrupts.

### Technical Implementation

The blocking operation system now provides:
- **Thread-safe cancellation** with atomic state transitions
- **Cross-platform atomic operations** supporting Windows, macOS, and Linux  
- **Anonymous BlockingOperation class** hidden from public API
- **Comprehensive state machine** for operation lifecycle management
- **GC-safe object management** with proper root registration

### API Improvements

- `rb_fiber_scheduler_blocking_operation_extract` - Safely extract operation data while holding GVL
- `rb_fiber_scheduler_blocking_operation_execute` - Execute in thread pools without GVL concerns  
- `rb_fiber_scheduler_blocking_operation_cancel` - Thread-safe cancellation support
- Consistent naming with `RB_FIBER_SCHEDULER_BLOCKING_OPERATION_STATUS_*` enum values

### Testing

Enhanced tests covering:
- IO#close across fibers 
- Memory safety verification
- Blocking operation cancellation
- Interrupt handling scenarios

This resolves the hanging behavior described in the original issue and makes IO#close operations safe in concurrent fiber applications.

---

Perfect! Now the title clearly indicates this started with IO#close issues, which is what will help people understand the context and importance of the fix.

PR #13441, opened 2025-05-26
shape.c: Implement a lock-free version of get_next_shape_internal

casperisfine:
Whenever we run into an inline cache miss when we try to set an ivar, we may need to take the global lock, just to be able to lookup inside `shape->edges`.

To solve that, when we're in multi-ractor mode, we can treat the `shape->edges` as immutable. When we need to add a new edge, we first copy the table, and then replace it with CAS.

This increases memory allocations, however we expect that creating new transitions becomes increasingly rare over time.

```ruby
class A
  def initialize(bool)
    @a = 1
    if bool
      @b = 2
    else
      @c = 3
    end
  end

  def test
    @d = 4
  end
end

def bench(iterations)
  i = iterations
  while i > 0
    A.new(true).test
    A.new(false).test
    i -= 1
  end
end

if ARGV.first == "ractor"
  ractors = 8.times.map do
    Ractor.new do
      bench(20_000_000 / 8)
    end
  end
  ractors.each(&:take)
else
  bench(20_000_000)
end
```

The above benchmark takes 27 seconds in Ractor mode on Ruby 3.4, and only 1.7s with this branch.

cc @etiennebarrie @jhawthorn @tenderlove @luke-gruber

### DIFF
--- a/id_table.c
+++ b/id_table.c
@@ -80,9 +80,10 @@ round_capa(int capa)
     return (capa + 1) << 2;
 }
 
-static struct rb_id_table *
-rb_id_table_init(struct rb_id_table *tbl, int capa)
+struct rb_id_table *
+rb_id_table_init(struct rb_id_table *tbl, size_t s_capa)
 {
+    int capa = (int)s_capa;
     MEMZERO(tbl, struct rb_id_table, 1);
     if (capa > 0) {
         capa = round_capa(capa);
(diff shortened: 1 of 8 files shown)


jhawthorn (approved):
Awesome! The atomic logic reads as correct to me

byroot:
> I suppose `shape->edges` might have been moved.

It seems I was right. For now I made it so these objects are pinned, which fixes the issue. That's not ideal, but I have no better solution short term.

My hope however is that if we move on with https://github.com/ruby/ruby/pull/13289, ultimately there would be no `SHAPE_T_OBJECT`, the `heap_id` data would just be a tag inside the shape_id, which means GC would no longer need to rebuild shapes, hence we'd no longer need to pin these.

But that's multiple big "ifs".

peterzhu2118 (approved):

PR #13451, opened 2025-05-27
[Bug #21377] Set#inspect should use the class name for path

zzak:


zzak:
Closing in favor of #13488

PR #13454, opened 2025-05-28
Namespace management using control frame

tagomoris:
This change is to update the namespace management. The main purpose is to fix bugs in corner cases to determine the current namespaces. I expect that this fix can solve bugs/problems to run RubyGems/Bundler with namespaces.

Core changes are:

* Determine the current namespace by the local context (`VM_CF_LEP` or `VM_EP_RUBY_LEP`) basically
* Stop using stacks for the namespace management
* Start using control frames (especially `cfp->ep[VM_ENV_DATA_INDEX_SPECVAL]`) to store the namespace of the frame
    * This is valid only when the frame type is `VM_FRAME_MAGIC_TOP` or `VM_FRAME_MAGIC_CLASS`
    * Those frame types don't have block handlers (block handlers are stored in specval for other frame types)
* Remove `vm->load_path` and other values, then start referring the root namespace (and its `load_path` and other fields) for simplicity
* Stop using refinements to determine the loading namespace by setting `VM_FRAME_FLAG_NS_REQUIRE`

TODO:
* [x] Delete the unused code about the current namespace management in namespace.c
* [x] Define singleton methods on `top_self` under namespaces (`#to_s` and `#inspect`)
* [x] Implement freeing `rb_namespace_t` (especially `loading_table_entry`)
* [x] Update `namespace_entry_memsize`
* [x] Implement global variable handling based on root/main namespaces
* [x] Check `vm_set_main_stack(ec, iseq); // TODO: not need to set the namespace?` → nothing to do

Add tests about:
* [ ] Basic namespace behavior
* [ ] Current namespace management under CFUNC calls
* [ ] Global variables in main/user namespaces

### DIFF
--- a/namespace.c
+++ b/namespace.c
@@ -826,6 +826,18 @@ Init_enable_namespace(void)
 
 #ifdef RUBY_DEBUG
 
+static VALUE
+rb_namespace_s_root(VALUE recv)
+{
+    return root_namespace->ns_object;
+}
+
+static VALUE
+rb_namespace_s_main(VALUE recv)
+{
+    return main_namespace->ns_object;
+}
+
 static const char *
 classname(VALUE klass)
 {
@@ -990,6 +1002,8 @@ Init_Namespace(void)
         rb_include_module(rb_cObject, rb_mNamespaceLoader);
 
 #ifdef RUBY_DEBUG
+        rb_define_singleton_method(rb_cNamespace, "root", rb_namespace_s_root, 0);
+        rb_define_singleton_method(rb_cNamespace, "main", rb_namespace_s_main, 0);
         rb_define_global_function("dump_classext", rb_f_dump_classext, 1);
 #endif
     }


namespace.c
```diff
@@ -70,75 +66,29 @@ rb_namespace_available(void)
...
 {
     if (!NAMESPACE_BUILTIN_P(ns)) {
         return false;
     }
-    /* Primitive methods (just to call C methods) covers/hides the effective
+    / Primitive methods (just to call C methods) covers/hides the effective
        namespaces, so ignore the methods' namespaces to expose user code's
        namespace to the implementation.
-     */
+     /
```
nobu, line 91: You can disable code contains comments by `#if 0`..`#endif`.

vm.c
```diff
@@ -118,7 +141,15 @@ PUREFUNC(static inline VALUE VM_CF_BLOCK_HANDLER(const rb_control_frame_t * cons
...
 VM_CF_BLOCK_HANDLER(const rb_control_frame_t * const cfp)
 {
-    const VALUE *ep = VM_CF_LEP(cfp);
+    const VALUE *ep;
+    if (VM_ENV_NAMESPACED_P(cfp->ep)) {
+        VM_ASSERT(VM_ENV_LOCAL_P(cfp->ep));
+        /* Never set black_handler for VM_FRAME_MAGIC_TOP or VM_FRAME_MAGIC_CLASS
+         * and the specval is used for namespace (rb_namespace_t) in these case
+         */
+        return VM_BLOCK_HANDLER_NONE;
```
tagomoris, line 169: For the anchor in the discussion: how to handle SPECVAL

vm.c
```diff
@@ -118,7 +141,15 @@ PUREFUNC(static inline VALUE VM_CF_BLOCK_HANDLER(const rb_control_frame_t * cons
 static inline VALUE
 VM_CF_BLOCK_HANDLER(const rb_control_frame_t * const cfp)
 {
-    const VALUE *ep = VM_CF_LEP(cfp);
+    const VALUE *ep;
+    if (VM_ENV_NAMESPACED_P(cfp->ep)) {
```
k0kubun, line 164: I'm confused by the fact that `VM_CF_BLOCK_HANDLER()`'s implementation is different from `GET_BLOCK_HANDLER()`'s. `GET_BLOCK_HANDLER()` checks `VM_ENV_NAMESPACED_P` on `GET_LEP()` while `VM_CF_BLOCK_HANDLER()` checks `cfp->ep` instead of using `VM_CF_LEP(cfp)`, which is the one actually queried later.

What allows them to be different? How is it safe for `VM_CF_BLOCK_HANDLER()` to not check the EP that has the specval to be queried?

tagomoris: This if-clause is a kind of early returrn because if `VM_ENV_NAMESPACED_P(ep)` is true, then `VM_ENV_LOCAL_P(ep)` is also true and `ep == VM_CF_LEP(ep)` (at least for now). `VM_ENV_BLOCK_HANDLER(ep)` also checks `VM_ENV_NAMESPACED_P(ep)`, so `GET_BLOCK_HANDLER()` returns the same result anyway.

I agree that it's confusing. So, I'll try to update `GET_BLOCK_HANDLER` definition to `VM_CF_BLOCK_HANDLER(GET_CFP())` and check if it will work correctly.

PR #13469, opened 2025-05-29
ZJIT: Implement side exits for entry frames

k0kubun:
This PR adds an initial implementation for side exits.

It exits to the interpreter correctly when the current frame is pushed by the interpreter. When it's pushed by JIT, however, it panics inside `rewind_caller_frames()`. It should materialize fields that JIT-to-JIT calls skip to write, which is to be done in a separate PR.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1818,7 +1818,7 @@ impl Assembler
     #[unsafe(no_mangle)]
     extern "C" fn rewind_caller_frames(addr: *const u8) {
         if ZJITState::is_iseq_return_addr(addr) {
-            unimplemented!("rewind_caller_frames is not implemented yet");
+            unimplemented!("Can't side-exit from JIT-JIT call: rewind_caller_frames is not implemented yet");
         }
     }
 }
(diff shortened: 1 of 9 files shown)


zjit/src/backend/lir.rs
```diff
@@ -1669,47 +1753,74 @@ impl Assembler
...
+        Some(())
     }
 
-    /// Assert that the next ccall() is going to be leaf.
-    pub fn expect_leaf_ccall(&mut self) {
-        self.leaf_ccall = true;
+    #[unsafe(no_mangle)]
+    extern "C" fn rewind_caller_frames(addr: *const u8) {
+        if ZJITState::is_iseq_return_addr(addr) {
+            unimplemented!("rewind_caller_frames is not implemented yet");
```
tekknolagi:
```suggestion
            unimplemented!("Can't side-exit from JIT-JIT call: rewind_caller_frames is not implemented yet");
```

tekknolagi (approved):
awesome, thank you for explaining. looks good. let's go for full frame push for now, as you suggested, to unblock the rest of the project

PR #13472, opened 2025-05-30
Sync lockfile from rubygems/rubygems

deivid-rodriguez:
This PR backports https://github.com/ruby/ruby/commit/c8d11edcf556088d10f0a245139741dcab1c7d56 to the stable branch. That way Bundler CI won't break every time a `rake` release happens.

I'll try to change Bundler specs to not break even if the rake version changes, but this is an easy fix already applied in master.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -233,7 +233,6 @@ lcov*.info
 
 # for `make test-bundler`
 /.rspec_status
-/tool/bundler/*.lock
 
 # /tool/
 /tool/config.guess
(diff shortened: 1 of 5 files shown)


deivid-rodriguez:
I addressed this in Bundler through https://github.com/rubygems/rubygems/pull/8735 but I think this PR is still good in case some other dependency change causes issues.

k0kubun:
I'm not sure what happened, but yeah it's green now.

PR #13475, opened 2025-05-30
Do not respect ruby2_keywords on method/proc with post arguments

jeremyevans:
Previously, ruby2_keywords could be used on a method or proc with
post arguments, but I don't think the behavior is desired:

```ruby
def a(*c, **kw) [c, kw] end
def b(*a, b) a(*a, b) end
ruby2_keywords(:b)

b({foo: 1}, bar: 1)
# Before: [[{foo: 1}], {bar: 1}]
#  After: [[{foo: 1}, {bar: 1}], {}]
```

This changes ruby2_keywords to emit a warning and not set the
flag on a method/proc with post arguments.

While here, fix the ruby2_keywords specs for warnings, since they
weren't testing what they should be testing.  They all warned
because the method didn't accept a rest argument, not because it
accepted a keyword or keyword rest argument

### DIFF
--- a/proc.c
+++ b/proc.c
@@ -4012,12 +4012,13 @@ proc_ruby2_keywords(VALUE procval)
     switch (proc->block.type) {
       case block_type_iseq:
         if (ISEQ_BODY(proc->block.as.captured.code.iseq)->param.flags.has_rest &&
+                !ISEQ_BODY(proc->block.as.captured.code.iseq)->param.flags.has_post &&
                 !ISEQ_BODY(proc->block.as.captured.code.iseq)->param.flags.has_kw &&
                 !ISEQ_BODY(proc->block.as.captured.code.iseq)->param.flags.has_kwrest) {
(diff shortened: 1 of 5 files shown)


jeremyevans:
> > It's not possible to pass keywords to such methods as the syntax doesn't allow it, since you cannot have keywords arguments before a positional argument, and a positional argument is needed for the post argument.
> 
> I'm not very clear on this, so let me try to paraphrase. Do you mean prior to this change, passing keyword args to a `ruby2_keywords` method that only has post parameters would always end up triggering the kwargs-to-hash conversion, and that the hash would always be absorbed by the last post parameter, so no keywords hash ends up in the rest parameter?

I originally thought it was not possible to trigger incorrect behavior, but apparently my original testing was insufficient.  It is possible to trigger incorrect behavior:

```ruby
def a(*c, **kw) [c, kw] end
def b(*a, b) a(*a, b) end
ruby2_keywords(:b)

b({foo: 1}, bar: 1)
# => [[{foo: 1}], {bar: 1}]
```

I'll open up a Redmine bug report for this, since I think we would want to backport this fix.

PR #13486, opened 2025-05-31
[DOC] Tweaks for String#bytes

BurdetteLamar:


doc/string/bytes.rdoc
```diff
@@ -1,6 +1,8 @@
-Returns an array of the bytes in +self+:
+Returns an array of the 8-bit bytes in +self+:
```
peterzhu2118: AFAIK all the platforms that Ruby runs on uses 8-bit bytes. In the case that a system somehow uses non 8-bit bytes, I think Ruby will also use the native byte size. So I think it's a bit redundant.

BurdetteLamar: Omitted.

peterzhu2118 (approved):

PR #13488, opened 2025-06-01
Simplify Set#inspect output

jeremyevans:
As Set is now a core collection class, it should have special inspect output.  Ideally, inspect output should be suitable to eval, similar to array and hash (assuming the elements are also suitable to eval):

```ruby
  set = Set[1, 2, 3]
  eval(set.inspect) == set # should be true
```

The simplest way to do this is to use the Set[] syntax.

This deliberately does not use any subclass name in the output, similar to array and hash. It is more important that users know they are dealing with a set than which subclass:

```ruby
  Class.new(Set)[]
  # this does: Set[]
  # not: #<Class:0x00000c21c78699e0>[]
```

This inspect change breaks the power_assert bundled gem tests, so add power_assert to TEST_BUNDLED_GEMS_ALLOW_FAILURES in the workflows.

### DIFF
--- a/.github/workflows/bundled_gems.yml
+++ b/.github/workflows/bundled_gems.yml
@@ -104,7 +104,7 @@ jobs:
         timeout-minutes: 30
         env:
           RUBY_TESTOPTS: '-q --tty=no'
-          TEST_BUNDLED_GEMS_ALLOW_FAILURES: 'typeprof,rbs,repl_type_completor'
+          TEST_BUNDLED_GEMS_ALLOW_FAILURES: 'typeprof,rbs,repl_type_completor,power_assert'
         if: ${{ steps.diff.outputs.gems }}
 
       - name: Commit
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -146,7 +146,7 @@ jobs:
         timeout-minutes: 60
         env:
           RUBY_TESTOPTS: '-q --tty=no'
-          TEST_BUNDLED_GEMS_ALLOW_FAILURES: 'typeprof,rbs,repl_type_completor'
+          TEST_BUNDLED_GEMS_ALLOW_FAILURES: 'typeprof,rbs,repl_type_completor,power_assert'
           PRECHECK_BUNDLED_GEMS: 'no'
           LAUNCHABLE_STDOUT: ${{ steps.launchable.outputs.stdout_report_path }}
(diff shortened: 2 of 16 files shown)


eregon (changes requested):
It should show the subclass name: https://bugs.ruby-lang.org/issues/21389?next_issue_id=21388&prev_issue_id=21390#note-2
I don't understand how you think it's a good idea to mislead the user and show in inspect code which wouldn't reproduce a corresponding instance.
Obviously Module and Object don't do that.
The subclass might have extra methods etc which means recreating a normal Set instead would be incompatible.

jeremyevans:
> It should show the subclass name: https://bugs.ruby-lang.org/issues/21389?next_issue_id=21388&prev_issue_id=21390#note-2 I don't understand how you think it's a good idea to mislead the user and show in inspect code which wouldn't reproduce a corresponding instance. 

This is not misleading the user any more than `{}` or `[]` are misleading the user for a Hash or Array subclass.  If we had syntax for set literals, `Set#inspect` would use it and would definitely not include subclass name.  `Set[]` is the closest we have to set literal syntax at present, so that is what is used.

Whether to use the subclass name is Matz's decision.  If  he decides to use the subclass name, then I'll update this accordingly.

> Obviously Module and Object don't do that.

They follow `Object#inspect` behavior.  This moves Set away from using `Object#inspect` behavior.

> The subclass might have extra methods etc which means recreating a normal Set instead would be incompatible.

Same is true for Hash and Array.

eregon:
> This is not misleading the user any more than `{}` or `[]` are misleading the user for a Hash or Array subclass.

It is, showing the wrong class name is more misleading than showing no class name.

> If we had syntax for set literals, `Set#inspect` would use it and would definitely not include subclass name.

Yes, agreed, but we don't have that and I think it's not likely to happen (`Set[]` is concise enough to build a Set, syntax for this is overkill IMO).

> `Set[]` is the closest we have to set literal syntax at present, so that is what is used.

Yep, agreed that's a good `inspect` for Set, except it should be `MySet[1, 2, 3]` for `MySet < Set` instances.
It's trivial to implement and the performance impact is very minimal, I see* no reason to needlessly confuse the user with a wrong class name here in the inspect output.

eregon:
Ah I missed (not really, it was written after) https://bugs.ruby-lang.org/issues/21389#note-3, that settles it then

notEthan:
`Set[:x]` is definitely nicer than `#<Set: {:x}>` (regardless of subclass name).
I think #pretty_print should be updated as well?

PR #13493, opened 2025-06-02
[3.4] compile.c: Handle anonymous variables in `outer_variable_cmp`

byroot:
[[Bug #21370]](https://bugs.ruby-lang.org/issues/21370)


byroot:
Interesting. Somehow the fix no longer works once backported :/ 

No idea to fix that.

rwstauner:
I backported a different commit to fix the bug that this test was highlighting.
https://github.com/ruby/ruby/pull/14573

`outer_variable_cmp` isn't being run in the context of the test, but I don't know if your fix might still be wanted for some other reason (maybe there is another way to induce the crash?).

k0kubun:
Thank you!

PR #13497, opened 2025-06-02
Allow pass special constants to the write barrier

peterzhu2118:
Some GC implementations want to always know when an object is written to, even if the written value is a special constant. Checking special constants in rb_obj_written was a micro-optimization that made assumptions about the GC implementation.

### DIFF
--- a/include/ruby/internal/abi.h
+++ b/include/ruby/internal/abi.h
@@ -24,7 +24,7 @@
  * In released versions of Ruby, this number is not defined since teeny
  * versions of Ruby should guarantee ABI compatibility.
  */
-#define RUBY_ABI_VERSION 1
+#define RUBY_ABI_VERSION 2
 
 /* Windows does not support weak symbols so ruby_abi_version will not exist
  * in the shared library. */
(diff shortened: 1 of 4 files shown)


composerinteralia:
Ah, I think this makes my comment over in https://github.com/ruby/strscan/pull/156#issuecomment-2907907492 incorrect. Do you think there are other places where folks might have skipped over special consts like I did because they didn't think it mattered? Will that cause problems with other GC implementations?

peterzhu2118:
> Do you think there are other places where folks might have skipped over special consts like I did because they didn't think it mattered?

For sure. There are plenty of micro-optimizations taken because of assumptions of how the garbage collector works. We're slowly removing these assumptions, but it won't be a small task. 

> Will that cause problems with other GC implementations?

Potentially., but not right now. We currently don't need this information to be accurate, just a best effort. We'll think about how to improve this when we need this information to be accurate.

include/ruby/internal/abi.h
```diff
@@ -24,7 +24,7 @@
  * In released versions of Ruby, this number is not defined since teeny
  * versions of Ruby should guarantee ABI compatibility.
  */
-#define RUBY_ABI_VERSION 1
+#define RUBY_ABI_VERSION 2
```
nobu, line 27: What needs this?

peterzhu2118: `include/ruby/internal/gc.h` changed in an incompatible way so we need to bump the ABI version.

PR #13501, opened 2025-06-03
Fix scheduler warning

casperisfine:
```
test/fiber/test_scheduler.rb:98: warning: Scheduler should implement #fiber_interrupt
```

@ioquatix


ioquatix:
Thanks, this looks okay to me.

PR #13502, opened 2025-06-03
Add missing lock in `rb_ivar_defined`

casperisfine:
If called on a class, we should acquire the lock.

Hopefully https://github.com/ruby/ruby/pull/13411 will soon allow to access class instance variables without a lock, but in the meantime it should be there.

cc @tenderlove @jhawthorn @etiennebarrie @luke-gruber

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -2131,12 +2131,11 @@ rb_obj_field_set(VALUE obj, shape_id_t target_shape_id, VALUE val)
     }
 }
 
-VALUE
-rb_ivar_defined(VALUE obj, ID id)
+static VALUE
+ivar_defined0(VALUE obj, ID id)
 {
     attr_index_t index;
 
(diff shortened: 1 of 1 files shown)


variable.c
```diff
@@ -2170,6 +2169,26 @@ rb_ivar_defined(VALUE obj, ID id)
...
+VALUE
+rb_ivar_defined(VALUE obj, ID id)
+{
+    if (SPECIAL_CONST_P(obj)) return Qfalse;
+
+    VALUE defined;
+    switch (BUILTIN_TYPE(obj)) {
+      case T_CLASS:
+      case T_MODULE:
+        RB_VM_LOCKING() {
```
casperisfine, line 2181: Actually the patch isn't ideal, we only need to lock in the `too_complex` case.

PR #13505, opened 2025-06-03
ZJIT: Add newrange support

st0012:
cc @tekknolagi

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -7,7 +7,7 @@ use crate::state::ZJITState;
 use crate::{asm::CodeBlock, cruby::*, options::debug, virtualmem::CodePtr};
 use crate::invariants::{iseq_escapes_ep, track_no_ep_escape_assumption};
 use crate::backend::lir::{self, asm_comment, Assembler, Opnd, Target, CFP, C_ARG_OPNDS, C_RET_OPND, EC, SP};
-use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, CallInfo};
+use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, CallInfo, RangeType};
 use crate::hir::{Const, FrameState, Function, Insn, InsnId};
 use crate::hir_type::{types::Fixnum, Type};
 use crate::options::get_option;
@@ -558,7 +558,7 @@ fn gen_new_range(
     asm: &mut Assembler,
     low: lir::Opnd,
     high: lir::Opnd,
-    flag: u32,
+    flag: RangeType,
     state: &FrameState,
 ) -> lir::Opnd {
     asm_comment!(asm, "call rb_range_new");
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -251,6 +251,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::PutSelf => gen_putself(),
         Insn::Const { val: Const::Value(val) } => gen_const(*val),
         Insn::NewArray { elements, state } => gen_new_array(jit, asm, elements, &function.frame_state(*state)),
+        Insn::NewRange { low, high, flag, state } => gen_new_range(asm, opnd!(low), opnd!(high), *flag, &function.frame_state(*state)),
         Insn::ArrayDup { val, state } => gen_array_dup(asm, opnd!(val), &function.frame_state(*state)),
         Insn::Param { idx } => unreachable!("block.insns should not have Insn::Param({idx})"),
         Insn::Snapshot { .. } => return Some(()), // we don't need to do anything for this instruction at the moment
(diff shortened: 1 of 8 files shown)


zjit/src/codegen.rs
```diff
@@ -550,6 +551,28 @@ fn gen_new_array(
...
 
+/// Compile a new range instruction
+fn gen_new_range(
+    asm: &mut Assembler,
+    low: lir::Opnd,
+    high: lir::Opnd,
+    flag: u32,
+    state: &FrameState,
+) -> lir::Opnd {
+    asm_comment!(asm, "call rb_range_new");
```
tekknolagi, line 564: Nit: this should go by the ccall

zjit/src/hir.rs
```diff
@@ -334,6 +334,7 @@ pub enum Insn {
     NewArray { elements: Vec<InsnId>, state: InsnId },
     /// NewHash contains a vec of (key, value) pairs
     NewHash { elements: Vec<(InsnId,InsnId)>, state: InsnId },
+    NewRange { low: InsnId, high: InsnId, flag: u32, state: InsnId },
```
tekknolagi: Please convert `flag` into a Rust-style enum if it's just `Inclusive|Exclusive`

tekknolagi: (What are the options?)

zjit/src/hir.rs
```diff
@@ -916,6 +926,7 @@ impl Function {
                 }
                 NewHash { elements: found_elements, state: find!(state) }
             }
+            &NewRange { low, high, flag, state } => NewRange { low: find!(low), high: find!(high), flag, state },
```
```suggestion
            &NewRange { low, high, flag, state } => NewRange { low: find!(low), high: find!(high), flag, find!(state) },
```

tekknolagi: We're not consistent about this and it doesn't/shouldn't matter but still

zjit/src/hir.rs
```diff
@@ -2709,6 +2734,52 @@ mod tests {
         "#]]);
     }
 
+    #[test]
+    fn test_new_range_with_one_element() {
```
tekknolagi: Please specify inclusive/exclusive with each test as you did in the Ruby-level tests

zjit/src/hir_type/mod.rs
```diff
@@ -137,6 +137,10 @@ fn is_hash_exact(val: VALUE) -> bool {
     val.class_of() == unsafe { rb_cHash } || (val.class_of() == VALUE(0) && val.builtin_type() == RUBY_T_HASH)
 }
 
+fn is_range_exact(val: VALUE) -> bool {
+    val.class_of() == unsafe { rb_cRange } || (val.class_of() == VALUE(0) && val.builtin_type() == RUBY_T_RANGE)
```
tekknolagi: We only do this because hash/array/... have this weird quasi-alive state if they are constant objects in the bytecode stream and need to be "resurrected". If this is not the case with range, we should just do a `class_of()` check like the others below

zjit/src/hir.rs
```diff
@@ -494,6 +496,14 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             }
+            Insn::NewRange { low, high, flag, .. } => {
+                let flag_str = match *flag {
+                    0 => "..",
+                    1 => "...",
```
tekknolagi: If you're not opposed, I think `NewRangeExclusive` and `NewRangeInclusive`, just for printing, would be maybe easier to read/skim than `..` vs `...`

tekknolagi (approved):
lgtm pending some comments!

tekknolagi:
Oh also I think PR/comment should be prefixed with "ZJIT:" like "ZJIT: Parse newrange into HIR"

zjit/src/hir.rs
tekknolagi: oh, it's a bool called `exclude_end`. i think `enum { Inclusive, Exclusive }` makes sense here

tekknolagi: Last, we don't have this yet, but in a stack @tenderlove and I are working on, we have an `ObjectAlloc` instruction, so we can probably eventually split this into `ObjectAlloc`+`FieldSet`+`FieldSet` or something similar. But again, these instructions don't exist now

zjit/src/codegen.rs
st0012: In `gen_array_dup` it also uses `call rb_ary_resurrect` while doing `asm.ccall`. Should we be consistent and update that and other occurrences like this to use `ccall {instruction name}`?

zjit/src/hir.rs
st0012: Do you mean we will be splitting `new_something` instructions into a combination of field set instructions + `ObjectAlloc`?

tekknolagi: Likely yes

zjit/src/codegen.rs
tekknolagi: Yeah we should probably move the comments. I think it is a quirk of the PC saving being added later

PR #13508, opened 2025-06-04
Fix a design issue in the debug inspector API

mame:
Previously, a user of the debug inspector API was supposed to use `rb_debug_inspector_backtrace_locations` to get an array of `Thread::Backtrace::Location`, and then used its index to get more information from `rb_debug_inspector _frame_binding_get(index)`, etc.

However, `rb_debug_inspector_backtrace_locations` returned an array of backtraces excluding rescue/ensure frames. On the other hand, `rb_debug_inspector_frame_binding_get(index)` interprets index with rescue/ensure frames. This led to inconsistency of the index, and it was very difficult to correctly use the debug inspector API.

This change deprecates `rb_debug_inspector_backtrace_locations` and introduces instead `rb_debug_inspector_frame_loc_get(index)`, which returns a `Thread::Backtrace::Location` object for the frame specified by `index`. Also `rb_debug_inspector_frame_count` is introduced to get the length of the backtrace.

### DIFF
--- a/ext/-test-/debug/inspector.c
+++ b/ext/-test-/debug/inspector.c
@@ -4,8 +4,7 @@
 static VALUE
 callback(const rb_debug_inspector_t *dbg_context, void *data)
 {
-    VALUE locs = rb_debug_inspector_backtrace_locations(dbg_context);
-    long i, len = RARRAY_LEN(locs);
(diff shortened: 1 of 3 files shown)


mame:
superseded by #13510

PR #13511, opened 2025-06-04
Get rid of TOO_COMPLEX shape type

casperisfine:
Instead it's now a `shape_id` flag.

This allows to check if an object is complex without having to chase the `rb_shape_t` pointer.

### DIFF
--- a/ext/objspace/objspace_dump.c
+++ b/ext/objspace/objspace_dump.c
@@ -821,9 +821,6 @@ shape_id_i(shape_id_t shape_id, void *data)
       case SHAPE_T_OBJECT:
         dump_append(dc, ", \"shape_type\":\"T_OBJECT\"");
         break;
-      case SHAPE_OBJ_TOO_COMPLEX:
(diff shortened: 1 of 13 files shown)


etiennebarrie:
Not sure what `SHAPE_FL_NON_CANONICAL_MASK` is, but should the new flag be added to that mask? or the mask removed since it's unused?

byroot:
> Not sure what `SHAPE_FL_NON_CANONICAL_MASK` is

It's the flags that need to be cleared on a `.dup`. So `frozen` and `object_id`, hence `too_complex` isn't part of it.

etiennebarrie:
But it's not used by `dup`? `rb_obj_copy_ivar` calls a similarly-named `rb_shape_canonical_p` but it doesn't use that specific mask.

byroot:
It's temporary: https://github.com/ruby/ruby/pull/13515/files#diff-15942e8d9ba3ea0d5b49d6c5fe5e10e6a363f21450ef76799a50d5fa1fa454ebR184

PR #13514, opened 2025-06-04
Get rid of TOO_COMPLEX shape type

casperisfine:
Complex shapes still exist, but it's now a `shape_id` bit flag rather than a materialized shape transition.

This allows to check for complex shape without dereferencing the shape.


casperisfine:
Wrong branch, sorry.

PR #13517, opened 2025-06-04
Integrate Ractor#join and Ractor#value with the fiber scheduler

luke-gruber:
Allow ractors to be used within a fiber scheduler context. Currently,
only Ractor.new { ... }.value or Ractor.new { ... }.join are supported
and tested.

When calling Ractor#join or Ractor#value within a non-blocking fiber
scheduler context, do not block the calling thread. Instead, call the
fiber scheduler's `block` method, which will transfer to another fiber
if coded correctly. When the ractor has terminated and is ready to give
its value to the calling thread, it will send an interrupt to that
thread telling it to switch back to its original fiber. If the thread is
blocked on IO (because it's in the fiber scheduler close hook), it will unblock
it by calling its unblock function. When the thread receives the interrupt,
it will call the fiber scheduler's `unblock` method, which will either
transfer to the original fiber or put it on a ready list, if coded correctly.

Ex:

```ruby
scheduler = Scheduler.new # your fiber scheduler
Fiber.set_scheduler scheduler
class << scheduler
  attr_reader :test_blockers
  def block(blocker, timeout=nil)
      (@test_blockers ||= []) << [blocker, timeout]
      super
  end
end
ordering = []
blocked_thread = nil
# in f1
Fiber.schedule do
  # in f2
  r = Ractor.new do
      # in f3
      sleep 0.5
  end
  ordering << "f2 before join"
  # Calling `r.join` should schedule us away from f2 back to f1. In f1, we end the script
  # and then Scheduler#close is called, which can block on IO.select or similar. When the ractor
  # is finished, it resumes fiber f2.
  blocked_thread = Thread.current
  r.join
  ordering << "f2 after join"
end
ordering << "f1 thread finish"
expected_ordering = ["f2 before join", "f1 thread finish", "f2 after join"]

assert_equal expected_ordering, ordering
assert_equal 1, scheduler.test_blockers.size
assert scheduler.test_blockers.first[0].is_a?(Thread) # the blocked thread that called join
assert_equal blocked_thread, scheduler.test_blockers.first[0]
```


samuel-williams-shopify:
This looks good to me.

- [ ] Let's split out the cosmetic changes so we can merge them separately.
- [ ] Then let's rebase this PR so we can focus on the review cycle with @ko1.

luke-gruber:
The PR is rebased and ready for review by @ko1. Thanks!

ractor_sync.c
```diff
@@ -1025,29 +1047,48 @@ ubf_ractor_wait(void *ptr)
...
-
     ccan_list_add_tail(&cr->sync.waiters, &waiter.node);
-
-    // resume another ready thread and wait for an event
-    rb_ractor_sched_wait(ec, cr, ubf_ractor_wait, &waiter);
+    VALUE scheduler = rb_fiber_scheduler_current();
+    rb_fiber_t *fiber = th->ec->fiber_ptr;
+    if (scheduler != Qnil && fiber && !rb_fiberptr_blocking(fiber)) {
+        waiter.fiber = fiber;
+        RACTOR_UNLOCK(cr);
```
ko1, line 1068: What happens when another Ractor wakes up this Ractor here?

common.mk
```diff
@@ -14113,6 +14114,26 @@ ractor.$(OBJEXT): $(top_srcdir)/internal/thread.h
 ractor.$(OBJEXT): $(top_srcdir)/internal/variable.h
 ractor.$(OBJEXT): $(top_srcdir)/internal/vm.h
 ractor.$(OBJEXT): $(top_srcdir)/internal/warnings.h
+ractor.$(OBJEXT): $(top_srcdir)/prism/defines.h
```
ko1, line 14117: why prism code are added here?

ko1:
Now that `Ractor::Port` was recently merged and is not yet mature, I don't want to introduce more complexity that would make it harder to debug.  
However, `Ractor::Port` is dramatically simpler than the previous `take`/`yield` pair, so the situation has improved.

samuel-williams-shopify:
> I don't want to introduce more complexity that would make it harder to debug.

Understood, but in this case, but given that the functionality is only present when the fiber scheduler is enabled, and it's a requirement not to block the scheduler, are you okay if we continue moving forward with this?

common.mk
luke-gruber: I don't know, I could look into it. I first opened the PR without these lines but the build failed saying it was missing these header dependencies, and tool/update-deps says that it's needed.

ractor_sync.c
luke-gruber: Good point. By looking at the code, I would say it might try to "unblock" the fiber before the fiber gets "blocked" by the ruby interrupt. I'll try to reproduce this behavior, and come back with a solution for this racy case. Thank you 😄

luke-gruber: So I checked out the situation by adding a `sleep` call after the `RACTOR_UNLOCK`, and for the use case of this PR (Only `Ractor#join` and `Ractor#take` can be called in fiber scheduler context), it is not racy. This is because the only possibility for wakeup would be the end of the ractor during `ractor_notify_exit`, and that uses an interrupt to signal this thread. Because interrupts are not checked in between the unlock and the time `rb_fiber_scheduler_block` is called, `unblock` cannot happen before `block` and the normal order of operations happens.
Edit: I also added a new commit that should fix some edge-case behavior.

ko1: `Ractor::Port#send` also wakes up the target ractor.

luke-gruber: Yes, but the scope of this PR is just to allow `Ractor#join` and `Ractor#value` inside the scheduler. This is why I only added tests for this behavior. Technically, you can call `Port#send` and it will also hook into the scheduler because I changed the generic `wait` and `wakeup` functions, so you are right. To address your concerns, we could do 1 of 3 things I can think of:

1) Only change the code path of `wakeup` when it's happening from `ractor_notify_exit`. This would add additional arguments to some functions. Also don't hook into `wait` when calling `Ractor.receive` or `Ractor.select` for these same reasons.

2) Get all ractor actions working properly with the fiber scheduler. This would be considerably more work and our team at Shopify don't want to do this right now, however we do want to do it in the future.

3) Leave this PR as is and commit to supporting the rest of the ractor actions with the fiber scheduler in the future.

For now I will add a commit that addresses point 1. 
Thank you!

ko1: How about to use Thread with Fiber scheduler app?

```ruby
class Ractor
  def join
    Thread.new{ self.orig_join }.join
  end
end
```

In this patch, `rb_threadptr_interrupt_exec` isn't used with thread creation, but interrupting at any timing is dangerous, so I think a similar technique is still necessary. After that, the cost of `Ractor#join` becomes equivalent.

luke-gruber: This is a good idea, and I did think about this. Ideally, we wanted to try to hook into the scheduler natively without thread creation. However, if you think it is too racy to use interrupts, I have no more ideas of how to accomplish this right now. In the worse case of this PR being rejected, we can use threads.

luke-gruber: So it was possible that during the call to `scheduler#block`, interrupts fired and the call to `scheduler#unblock` was made before `scheduler#block` finished transferring.

I have a few ideas on how to fix this:

1) in `ractor_wakeup_all`, just before calling `rb_threadptr_interrupt_exec`, we check that the fiber is not the current fiber, and if it is we put the waiter back on the list and try again, like this:

```c
                if (r->threads.running_ec->fiber_ptr == fiber) {
                    waiter->wakeup_status = wakeup_none;
                    ccan_list_add_tail(&r->sync.waiters, &waiter->node);
                    continue;
                }
```

I have tested this and it works well, but I'm just worried about the call to `RB_VM_CHECK_INTS` right after the call to `fiber_store` in `cont.c`, before `fiber_switch` returns. I would have to test this more for safety to ensure that if that interrupt switches the fiber back, then things don't go wrong. Or I could add a `fiber->switching` flag and make sure it's `false` as well before setting the interrupt on the thread.

2) We disable interrupts on the fiber's `ec` during the call to `scheduler#block` and re-enable them after it's scheduled back. A bit heavy-handed, but might work. Interrupt masking like this is currently not done when ruby code is being executed,  however. I am aware of `Thread.handle_interrupt` but it works differently, I believe.

3) We don't unlock the ractor in `ractor_wait`, and instead have the fiber unlock it after it finishes the transfer. I don't like this solution as much because I am wary of this type of locking/unlocking control flow.

Please let me know if any of these solutions would be acceptable to you. Sorry for the long post, and thank you for your time 😄

luke-gruber:
I'm closing this for now, as we're going to go forward using `Thread.new { }`. It would be nice to revisit this in the future.

PR #13518, opened 2025-06-04
Call `Set#add` from `initialize` on subclasses

tenderlove:
Set#new used to call add on items passed to initialize.  This adds backwards compatibility with subclasses of Set

[[Bug #21396]](https://bugs.ruby-lang.org/issues/21396)

### DIFF
--- a/set.c
+++ b/set.c
@@ -459,7 +459,12 @@ static VALUE
 set_initialize_with_block(RB_BLOCK_CALL_FUNC_ARGLIST(i, set))
 {
     VALUE element = rb_yield(i);
-    set_insert_wb(set, element, &element);
+    if (rb_obj_is_instance_of(set, rb_cSet)) {
+        set_insert_wb(set, element, &element);
+    }
+    else {
+        rb_funcall(set, rb_intern("add"), 1, element);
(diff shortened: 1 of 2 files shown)


jeremyevans:
I'm against this for reasons I described in the Redmine ticket.  If we want to be backwards compatible, we shouldn't do it piecemeal, we need to do a complete audit of all Set methods and in every case where stdlib Set is calling a Set method, we need to change core Set to call the method instead of directly calling the underlying C function.  That's likely to introduce race conditions that we'll have to deal with, and handling those safely will make the code even slower.

tenderlove:
@jeremyevans I understand. It's kind of a bummer though because it's breaking our existing code. It would be nice if there was some escape hatch, but I'm not sure what that would be.

ioquatix:
I think it's better to use composition rather than sub-classing if possible in scenarios like this. There are other code paths like `IO` where sub-classes of IO can override e.g. `read` and `write` in Ruby and it's a nightmare to maintain.

jeremyevans (commented):
With the addition of Set::SubclassCompatible, I think we can close this.

PR #13519, opened 2025-06-05
Get rid of SHAPE_T_OBJECT

casperisfine:
Now the `heap_index` is stored as flags inside `shape_id_t`.

This saves having to rebuild the entire tree when moving from one slot size to the other.

It also gives a cheap access to the slot size, whereas `rb_gc_obj_slot_size` often cause cache misses.


byroot:
Merged most of this as https://github.com/ruby/ruby/pull/13556

PR #13525, opened 2025-06-05
More comprehensive debugging configuration

etiennebarrie:
It took me a long time to figure out why I couldn't use `RubyVM::Shape` or the `RUBY_DEBUG_LOG`. I was relying on this documentation so I'm making it more comprehensive.

Let me know if you have other tips/flags that we could add.


byroot:
Merged as `22dfa250a89d11cc138ef73d1ed6c364c15b54c7` because CI is having weird issues.

PR #13528, opened 2025-06-05
ZJIT: Parse definedivar into HIR (+2 misc.)

XrXr:
- **ZJIT: Take a slice instead of Vec in test code**
- **ZJIT: Parse definedivar into HIR**
- **ZJIT: Fix insn arg index for `defined`, add tests**


zjit/src/hir.rs
```diff
@@ -398,6 +398,9 @@ pub enum Insn {
     GetIvar { self_val: InsnId, id: ID, state: InsnId },
     /// Set `self_val`'s instance variable `id` to `val`
     SetIvar { self_val: InsnId, id: ID, val: InsnId, state: InsnId },
+    /// Check whehter an insntace variable exists on `self_val`
```
```suggestion
    /// Check whether an insntace variable exists on `self_val`
```

zjit/src/hir.rs
```diff
@@ -2979,6 +3010,39 @@ mod tests {
...
+              Return v3
+        "#]]);
+    }
+
+    #[test]
+    fn defined() {
+        eval("
+            def test = return defined?(SeaChange), defined?(favourite), defined?($ruby)
+        ");
+        assert_method_hir_with_opcodes("test", &[YARVINSN_defined], expect![[r#"
```
```suggestion
        assert_method_hir_with_opcode("test", YARVINSN_defined, expect![[r#"
```

tekknolagi: (same above)

PR #13529, opened 2025-06-05
ZJIT: Pass self through basic block params

k0kubun:
This PR moves the storage of `self` from `cfp->self` to the first basic block argument to use a register.

Currently, ZJIT basic blocks are not allowed to refer to an argument of another basic block because the register allocator uses a fixed register for each basic block argument index, which would cause a conflict for such a reuse. I left a TODO comment in `as_args()` to fix it when we make the register allocator more flexible.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2606,9 +2606,10 @@ mod infer_tests {
         crate::cruby::with_rubyvm(|| {
             let mut function = Function::new(std::ptr::null());
             let param = function.push_insn(function.entry_block, Insn::Param { idx: SELF_PARAM_IDX });
+            function.param_types.push(types::BasicObject); // self
             let val = function.push_insn(function.entry_block, Insn::Test { val: param });
             function.infer_types();
-            assert_bit_equal(function.type_of(val), Type::from_cbool(false));
+            assert_bit_equal(function.type_of(val), types::CBool);
         });
     }
 


zjit/src/codegen.rs
```diff
@@ -324,13 +323,16 @@ fn gen_entry_prologue(asm: &mut Assembler, iseq: IseqPtr) {
...
+    asm.mov(self_param, Opnd::mem(VALUE_BITS, CFP, RUBY_OFFSET_CFP_SELF));
+
     let num_params = entry_block.params().len();
     if num_params > 0 {
         asm_comment!(asm, "set method params: {num_params}");
 
         // Allocate registers for basic block arguments
         let params: Vec<Opnd> = (0..num_params).map(|idx|
-            gen_param(asm, idx)
+            gen_param(asm, idx + 1)
```
```suggestion
            gen_param(asm, idx + 1)  // +1 for self
```

zjit/src/hir.rs
```diff
@@ -1981,8 +1989,11 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     // item between commas in the source increase the parameter count by one,
     // regardless of parameter kind.
     let mut entry_state = FrameState::new(iseq);
-    for idx in 0..num_locals(iseq) {
-        if idx < unsafe { get_iseq_body_param_size(iseq) }.as_usize() {
+    let mut self_param = fun.push_insn(fun.entry_block, Insn::Param { idx: SELF_PARAM_IDX });
+    fun.param_types.push(types::BasicObject);
```
```suggestion
    fun.param_types.push(types::BasicObject);  // self
```

zjit/src/hir.rs
```diff
@@ -1981,8 +1989,11 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     // item between commas in the source increase the parameter count by one,
     // regardless of parameter kind.
     let mut entry_state = FrameState::new(iseq);
-    for idx in 0..num_locals(iseq) {
-        if idx < unsafe { get_iseq_body_param_size(iseq) }.as_usize() {
+    let mut self_param = fun.push_insn(fun.entry_block, Insn::Param { idx: SELF_PARAM_IDX });
```
tekknolagi: Instead of making this mutable, maybe we should make it loop-local to shrink distance between def and use (might be lower cognitive overhead)

zjit/src/hir.rs
```diff
@@ -2386,15 +2398,15 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let id = ID(get_arg(pc, 0).as_u64());
                     // ic is in arg 1
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
-                    let self_val = fun.push_insn(block, Insn::PutSelf);
+                    let self_val = fun.push_insn(block, Insn::Param { idx: SELF_PARAM_IDX });
```
tekknolagi: This should be `self_param` instead of pushing a new `Param`, I think

zjit/src/hir.rs
```diff
@@ -2386,15 +2398,15 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
                     let result = fun.push_insn(block, Insn::GetIvar { self_val, id, state: exit_id });
                     state.stack_push(result);
                 }
                 YARVINSN_setinstancevariable => {
                     let id = ID(get_arg(pc, 0).as_u64());
                     // ic is in arg 1
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
-                    let self_val = fun.push_insn(block, Insn::PutSelf);
+                    let self_val = fun.push_insn(block, Insn::Param { idx: SELF_PARAM_IDX });
```
tekknolagi: This should be `self_param` instead of pushing a new `Param`, I think

zjit/src/hir.rs
```diff
@@ -2594,10 +2606,10 @@ mod infer_tests {
     fn test_unknown() {
         crate::cruby::with_rubyvm(|| {
             let mut function = Function::new(std::ptr::null());
-            let param = function.push_insn(function.entry_block, Insn::PutSelf);
+            let param = function.push_insn(function.entry_block, Insn::Param { idx: SELF_PARAM_IDX });
             let val = function.push_insn(function.entry_block, Insn::Test { val: param });
             function.infer_types();
-            assert_bit_equal(function.type_of(val), types::CBool);
+            assert_bit_equal(function.type_of(val), Type::from_cbool(false));
```
tekknolagi: Why did this change? We shouldn't know that it's `false`

zjit/src/hir.rs
```diff
@@ -3634,11 +3633,10 @@ mod tests {
         ");
         assert_method_hir_with_opcode("test", YARVINSN_setinstancevariable, expect![[r#"
             fn test:
-            bb0():
-              v1:Fixnum[1] = Const Value(1)
-              v3:BasicObject = PutSelf
-              SetIvar v3, :@foo, v1
-              Return v1
+            bb0(v0:BasicObject, v4):
```
tekknolagi: this bare `v4` should have a type

zjit/src/hir.rs
```diff
@@ -3619,10 +3619,9 @@ mod tests {
         ");
         assert_method_hir_with_opcode("test", YARVINSN_getinstancevariable, expect![[r#"
             fn test:
-            bb0():
-              v2:BasicObject = PutSelf
-              v3:BasicObject = GetIvar v2, :@foo
-              Return v3
+            bb0(v0:BasicObject, v3):
```
tekknolagi: this bare `v3` should have a type

zjit/src/hir.rs
```diff
@@ -4097,36 +4095,36 @@ mod opt_tests {
 
         assert_optimized_method_hir("rest", expect![[r#"
             fn rest:
-            bb0(v0:ArrayExact):
-              Return v0
+            bb0(v0:BasicObject, v1:BasicObject):
```
tekknolagi: we should still have ArrayExact

tekknolagi (changes requested):
I think there are some lingering indexing and typing issues but otherwise looks ok to me

zjit/src/codegen.rs
k0kubun: https://github.com/ruby/ruby/pull/13529/commits/3a0cf8b8c418b90ff70a387df8489ef7ac2604aa

zjit/src/hir.rs
k0kubun: https://github.com/ruby/ruby/pull/13529/commits/3a0cf8b8c418b90ff70a387df8489ef7ac2604aa

zjit/src/hir.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/13529/commits/59f878a1868d8259a71a65f3679e955ffb50defa

zjit/src/hir.rs
k0kubun: thanks for catching it :+1: https://github.com/ruby/ruby/pull/13529/commits/2e3f4b2863de9f5cefd09e978a8eb9d99fe38543

zjit/src/hir.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/13529/commits/2e3f4b2863de9f5cefd09e978a8eb9d99fe38543

zjit/src/hir.rs
k0kubun: fixed by https://github.com/ruby/ruby/pull/13529/commits/2e3f4b2863de9f5cefd09e978a8eb9d99fe38543

zjit/src/hir.rs
k0kubun: fixed by https://github.com/ruby/ruby/pull/13529/commits/2e3f4b2863de9f5cefd09e978a8eb9d99fe38543

zjit/src/hir.rs
k0kubun: fixed at https://github.com/ruby/ruby/pull/13529/commits/b5b4ad18847e561a9257c65996fc5445b20fdc70

zjit/src/hir.rs
```diff
@@ -2793,7 +2792,7 @@ mod tests {
         assert_method_hir_with_opcode("test", YARVINSN_newarray, expect![[r#"
```
tekknolagi, line 2795: I think this test is now wrong; instead of making an array of `a`, it makes an array of `self`

zjit/src/hir.rs
```diff
@@ -2804,7 +2803,7 @@ mod tests {
...
             fn test:
             bb0(v0:BasicObject, v1:BasicObject, v2:BasicObject):
-              v5:ArrayExact = NewArray v1, v2
+              v5:ArrayExact = NewArray v0, v1
```
tekknolagi: This is `[self, a]` instead of `[a, b]`

tekknolagi (changes requested):
I think something is still up with the indexing; see my comments about the modified tests

tekknolagi:
I can't push a commit but I think you want something like

```patch
commit d4d83a0d74b8ffe406ae9e5acd38b2a53f6829be (HEAD -> zjit-self-param)
Author: Max Bernstein <max@bernsteinbear.com>
Date:   Thu Jun 5 16:44:55 2025 -0400

    Only push self param for non-entry blocks

diff --git a/zjit/src/hir.rs b/zjit/src/hir.rs
index 1d26c99964..2711c7561b 100644
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1989,6 +1989,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     // item between commas in the source increase the parameter count by one,
     // regardless of parameter kind.
     let mut entry_state = FrameState::new(iseq);
+    fun.push_insn(fun.entry_block, Insn::Param { idx: SELF_PARAM_IDX });
     fun.param_types.push(types::BasicObject); // self
     for local_idx in 0..num_locals(iseq) {
         if local_idx < unsafe { get_iseq_body_param_size(iseq) }.as_usize() {
@@ -2012,11 +2013,14 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     while let Some((incoming_state, block, mut insn_idx)) = queue.pop_front() {
         if visited.contains(&block) { continue; }
         visited.insert(block);
-        let self_param = fun.push_insn(block, Insn::Param { idx: SELF_PARAM_IDX });
-        let mut state = if insn_idx == 0 { incoming_state.clone() } else {
+        let (self_param, mut state) = if insn_idx == 0 {
+            (fun.blocks[fun.entry_block.0].params[0], incoming_state.clone())
+        } else {
+            assert_eq!(fun.blocks[block.0].params.len(), 0);
+            let self_param = fun.push_insn(block, Insn::Param { idx: SELF_PARAM_IDX });
             let mut result = FrameState::new(iseq);
             let mut idx = 1;
-            for _ in 0..incoming_state.locals.len() {
+            for _ in incoming_state.locals {
                 result.locals.push(fun.push_insn(block, Insn::Param { idx }));
                 idx += 1;
             }
@@ -2024,7 +2028,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                 result.stack.push(fun.push_insn(block, Insn::Param { idx }));
                 idx += 1;
             }
-            result
+            (self_param, result)
         };
         // Start the block off with a Snapshot so that if we need to insert a new Guard later on
         // and we don't have a Snapshot handy, we can just iterate backward (at the earliest, to
```

zjit/src/hir.rs
k0kubun: Thanks, your patch fixed it https://github.com/ruby/ruby/pull/13529/commits/5e62c53d9685bdddd4239a31cbe7a4d67fa1f9db

zjit/src/hir.rs
k0kubun: Your patch fixed this, thanks https://github.com/ruby/ruby/pull/13529/commits/5e62c53d9685bdddd4239a31cbe7a4d67fa1f9db

zjit/src/hir.rs
```diff
@@ -2003,9 +2013,12 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     while let Some((incoming_state, block, mut insn_idx)) = queue.pop_front() {
         if visited.contains(&block) { continue; }
         visited.insert(block);
-        let mut state = if insn_idx == 0 { incoming_state.clone() } else {
+        let (self_param, mut state) = if insn_idx == 0 {
+            (fun.blocks[fun.entry_block.0].params[0], incoming_state.clone())
```
```suggestion
            (fun.blocks[fun.entry_block.0].params[SELF_PARAM_IDX], incoming_state.clone())
```

Per Kevin's recommendation, this might be a good idea. Sorry I forgot it in my last note

zjit/src/hir.rs
tekknolagi: This is still weird to me :/

tekknolagi: Aha, we need to do

```rust
            function.param_types.push(types::BasicObject); // self
```

tekknolagi: (in the test)

k0kubun: Thanks for pointing out the fix. Yeah I was looking into this and it had `Empty` and `is_known_falsy()` returned true (?). Pushed the fix at https://github.com/ruby/ruby/pull/13529/commits/5a543a65b08987fc5539e5399cab423f785de48b.

I wish we had less (ideally, no) synthetic tests like this. It seems to test something unreal and requires non-trivial effort for debugging it. The more integrated, the easier to maintain.

PR #13530, opened 2025-06-05
Use vm_sync API directly

ko1:
`RB_VM_*` APIs accept file:line information implicitly and `RUBY_DEBUG_LOG()` shows them. It is useful for debugging.

For modular GC, introduce some wrappers for them.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -130,32 +130,19 @@
 #include "builtin.h"
 #include "shape.h"
 
+#if USE_MODULAR_GC
 unsigned int
 rb_gc_vm_lock(void)
 {
-    unsigned int lev = 0;
-    RB_VM_LOCK_ENTER_LEV(&lev);
-    return lev;
+    RB_VM_LOCK();
+    return 1;
 }
 
 void
 rb_gc_vm_unlock(unsigned int lev)
 {
-    RB_VM_LOCK_LEAVE_LEV(&lev);
-}
-
-unsigned int
-rb_gc_cr_lock(void)
-{
-    unsigned int lev;
-    RB_VM_LOCK_ENTER_CR_LEV(GET_RACTOR(), &lev);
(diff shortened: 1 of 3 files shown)


peterzhu2118 (commented):
I don't like the amount of modular GC specific code.

Can we instead do something like this?

In `gc.c` we have functions like:

```c
unsigned void
rb_gc_vm_lock(const char *file, int line)
{
    rb_vm_lock(file, line);
}
```

In `gc/gc.h` we have:

```c
#define RB_GC_VM_LOCK() rb_gc_vm_lock(__FILE__, __LINE__)
```

Then in `default.c`, we replace `rb_gc_vm_lock()` with `RB_GC_VM_LOCK()`.

ko1:
or simply expose vm_sync.h functions when build for moduler gc and use `RB_VM` macros?

peterzhu2118:
> or simply expose vm_sync.h functions

Do you mean including `vm_sync.h` in modular GC? I'm against that because modular GC needs to have minimal dependancies to make it easy to build.

ko1:
okay, so I can agree with https://github.com/ruby/ruby/pull/13530#pullrequestreview-2902401042 idea.

can you make it or should I?

off-topic from this ticket, can we add `static` for `rb_gc_vm_...` functions unless `USE_MODULAR_GC`? They are not needed for non-modular gc versions.

peterzhu2118:
> can you make it or should I?

You can make it if you want. Otherwise, I can work on it tomorrow.

> off-topic from this ticket, can we add static for rb_gc_vm_... functions unless USE_MODULAR_GC?

I opened https://github.com/ruby/ruby/pull/13539.

peterzhu2118:
I opened https://github.com/ruby/ruby/pull/13550

PR #13531, opened 2025-06-05
Use a T_DATA for cross_ractor_require

jhawthorn:
Fixes [Bug #21090](https://bugs.ruby-lang.org/issues/21090)

The struct was previously allocated on the stack, which could be freed if the Thread is terminated. Moving this to a T_DATA on the heap should mean this is no longer an issue.

```
1000.times { Ractor.new { th = Thread.new { require "rbconfig" }; Thread.pass }.take }
```

Previously this would segfault, due to accessing the Thread's memory after it was terminated (when the Ractor exits). Now it prints 1000 times:

```
#<Thread:0x000077b21ddaa5e0 run> terminated with exception (report_on_exception is true):
The port was already closed (Ractor::ClosedError)
```

The error comes from trying to send the result of the require back through the port. Maybe we could have a better error message, but that seems pretty reasonable to me given how much of an edge case this is.

cc @ko1 @luke-gru

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -2339,8 +2339,8 @@ rb_ractor_require(VALUE feature)
     VALUE crr_obj = TypedData_Make_Struct(0, struct cross_ractor_require, &cross_ractor_require_data_type, crr);
     FL_SET_RAW(crr_obj, RUBY_FL_SHAREABLE);
 
-    // TODO: make feature shareable
-    crr->feature = feature;
+    // Convert feature to proper file path and make it shareable as fstring
+    crr->feature = rb_fstring(FilePathValue(feature));
     crr->port = ractor_port_new(GET_RACTOR());
     crr->result = Qundef;
     crr->exception = Qundef;
--- a/ractor.c
+++ b/ractor.c
@@ -2241,6 +2241,28 @@ struct cross_ractor_require {
     ID name;
 };
 
+static void
+cross_ractor_require_mark(void *ptr)
+{
+    struct cross_ractor_require *crr = (struct cross_ractor_require *)ptr;
+    rb_gc_mark(crr->port);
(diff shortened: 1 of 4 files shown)


ko1:
it should be problem with ractor local gc.

* pass simple malloc'ed memory
* make the object shareable

hmm, other ideas?

ractor.c
```diff
@@ -2237,6 +2237,28 @@ struct cross_ractor_require {
...
+static void
+cross_ractor_require_mark(void *ptr)
+{
+    struct cross_ractor_require *crr = (struct cross_ractor_require *)ptr;
+    rb_gc_mark(crr->port);
+    rb_gc_mark(crr->result);
+    rb_gc_mark(crr->exception);
+    rb_gc_mark(crr->feature);
+    rb_gc_mark(crr->module);
+}
```
paracycle, line 2253: Should we use declarative marking for this instead? https://github.com/Shopify/ruby/blob/master/doc/extension.rdoc/#declaratively-markingcompacting-struct-references-

jhawthorn:
I made the new CRR object shareable. I also fixed the todo comment and make `crr->feature` into a shareable fstring before sending.

ko1:
shareable objects increases a global GC pressure, so in this case, object shouldn't be used.

jhawthorn:
> shareable objects increases a global GC pressure, so in this case, object shouldn't be used.

Surely it can't be that significant for a cross-ractor-require, which are mostly going to be used at boot? Each one will also probably create an fstring, iseq, class, and more which are probably more significant GC pressure from shareable objects?

It's helpful to have an actual object only when both the main Ractor and the other Ractor have finished with it, and as the other Ractor can be terminated before, during, or after the main Ractor running the require, that is hard to do without relying on GC.

PR #13533, opened 2025-06-05
ZJIT: Panic unimplemented for OOB basic block args

havenwood:
- Go from 8 to 10 `ALLOC_REGS` registers for ARM64 and x86_64
- Panic about unimplemented register spilling when block args are out of bounds

Testing ZJIT I noticed an index out of bounds panic with:

```sh
ruby --zjit -e "pp 0"
```

Pending the implementation of register spilling, I propose:
1) Improve the panic message when there aren't enough registers for block arguments
2) Increase the number of basic block registers from 8 to 10 to cover many more cases in the interim.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -730,6 +730,12 @@ fn gen_push_frame(asm: &mut Assembler, recv: Opnd) {
 fn param_reg(idx: usize) -> Reg {
     // To simplify the implementation, allocate a fixed register for each basic block argument for now.
     // TODO: Allow allocating arbitrary registers for basic block arguments
+    if idx >= ALLOC_REGS.len() {
+        unimplemented!(
+            "register spilling not yet implemented, too many basic block arguments ({}/{})",
+            idx + 1, ALLOC_REGS.len()
+        );
+    }
     ALLOC_REGS[idx]
 }
 


zjit/src/backend/x86_64/mod.rs
```diff
@@ -90,6 +90,8 @@ pub const ALLOC_REGS: &'static [Reg] = &[
     R9_REG,
     R10_REG,
     RAX_REG,
+    R14_REG,
+    R15_REG,
```
k0kubun: r14 and r15 are callee-saved registers. The register allocator currently doesn't support callee-saved registers, so they're deliberately not added to the list at the moment. This branch could cause an undefined behavior depending on a C function called in JIT code. We'd need to preserve them when entering/leaving JIT code, but that's also overhead we need to carefully handle.

So our path forward is probably to support register spill first, before expanding the register pool to callee-saved registers.

zjit/src/codegen.rs
```diff
@@ -730,9 +730,26 @@ fn gen_push_frame(asm: &mut Assembler, recv: Opnd) {
...
+    }
     ALLOC_REGS[idx]
 }
 
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    #[should_panic(expected = "register spilling not yet implemented, too many basic block arguments (11/10)")]
```
k0kubun: The panic (by `unimplemented!`) itself works as an inline test, so I don't think you should test the test. It's also tedious to update the "11/10" part whenever you touch the `ALLOC_REGS` constant.

havenwood:
Thank you for explaining why I was in error, @k0kubun. I very much appreciate your thoughtful response. I've updated the PR accordingly, removing the new registers and unnecessary test.

I updated this PR to only include the unimplemented panic message.

k0kubun (approved):
Thanks :+1:

PR #13542, opened 2025-06-06
ignore confirming belonging while finrializer

ko1:
A finalizer registerred in Ractor A can be invoked in B.

```ruby
require "tempfile"
r = Ractor.new{
  10_000.times{|i|
    Tempfile.new(["file_to_require_from_ractor#{i}", ".rb"])
  }
}
sleep 0.1
```

For example, above script makes tempfiles which have finalizers on Ractor r, but at the end of the process, main Ractor will invoke finalizers and it violates belonging check. This patch just ignore the belonging check to avoid CI failure.

Of course it violates Ractor's isolation and wrong workaround. This issue will be solved with Ractor local GC.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -288,6 +288,7 @@ rb_gc_run_obj_finalizer(VALUE objid, long count, VALUE (*callback)(long i, void 
     saved.finished = 0;
     saved.final = Qundef;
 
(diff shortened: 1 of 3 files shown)


ko1:
See also https://bugs.ruby-lang.org/issues/21315

PR #13544, opened 2025-06-06
CI: Run Launchable in the build directory

nobu:
Use the build directory as the Launchable session directory as well as compilers/entrypoint.sh.

### DIFF
--- a/.github/actions/launchable/setup/action.yml
+++ b/.github/actions/launchable/setup/action.yml
@@ -57,12 +57,12 @@ inputs:
 
 outputs:
   stdout_report_path:
-    value: ${{ steps.variables.outputs.stdout_report_path }}
+    value: ${{ steps.global.outputs.stdout_report_path }}
     description: >-
       Report file path for standard output.
 
   stderr_report_path:
-    value: ${{ steps.variables.outputs.stderr_report_path }}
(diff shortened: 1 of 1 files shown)


.github/actions/launchable/setup/action.yml
```diff
@@ -236,124 +252,50 @@ runs:
...
-              "${stderr_report_path}"; \
+              "${stdout_report_file}" \
+              "${stderr_report_file}"; \
             launchable record tests \
               --session "$(cat "${test_spec_session_file}")" \
-              raw ${test_spec_report_path}/* || true; \
+              raw ${test_spec_report_dir}/* || true; \
           fi
-      if: ${{ always() && steps.enable-launchable.outputs.enable-launchable }}
+      if: always() && steps.setup-launchable.outcome == 'success'
```
ono-max: I think that we need to add the following command in `setup-launchable` step.

```
echo "outcome=true" >> $GITHUB_OUTPUT
```

nobu: It is a step context property, not an output.

https://docs.github.com/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#steps-context

ono-max (approved):
Thank you 🙏

PR #13545, opened 2025-06-06
Allow `Mutex#lock` in trap context.

ioquatix:
See <https://bugs.ruby-lang.org/issues/19473> for context.


ioquatix:
Another "fix" <https://github.com/ruby/ruby/pull/13684>.

eregon (approved):
I think this is a clear improvement, see https://bugs.ruby-lang.org/issues/19473#note-20 for the reasoning

PR #13546, opened 2025-06-06
Replicate heap_index in shape_id flags.

casperisfine:
This is preparation to getting rid of `T_OBJECT` transitions (https://github.com/ruby/ruby/pull/13519).

By first only replicating the information it's easier to ensure consistency.

### DIFF
--- a/internal/variable.h
+++ b/internal/variable.h
@@ -52,7 +52,7 @@ int rb_gen_fields_tbl_get(VALUE obj, ID id, struct gen_fields_tbl **fields_tbl);
 void rb_obj_copy_ivs_to_hash_table(VALUE obj, st_table *table);
 void rb_obj_init_too_complex(VALUE obj, st_table *table);
 void rb_evict_ivars_to_hash(VALUE obj);
-void rb_evict_fields_to_hash(VALUE obj);
+shape_id_t rb_evict_fields_to_hash(VALUE obj);
 VALUE rb_obj_field_get(VALUE obj, shape_id_t target_shape_id);
 void rb_ivar_set_internal(VALUE obj, ID id, VALUE val);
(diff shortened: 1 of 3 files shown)


byroot:
@XrXr do you have any idea what I'm doing wrong to cause YJIT/x86 to crash here? I can't make sense of that crash.

jhawthorn:
> @XrXr do you have any idea what I'm doing wrong to cause YJIT/x86 to crash here? I can't make sense of that crash.

I have a hunch. I added some print debugging to YJIT, and the crash happens immediately after a shape exceeds 2**31, so I think we're not handling that correctly

```
new_shape_id: 804002cf

thread '<unnamed>' panicked at yjit/src/asm/x86_64/mod.rs:553:30:
assertion `left == right` failed
  left: 64
 right: 32
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /usr/src/debug/rust/rustc-1.87.0-src/library/core/src/panicking.rs:380:5
   4: yjit::asm::x86_64::write_rm_multi
             at ./yjit/src/asm/x86_64/mod.rs:553:30
```

casperisfine:
THanks @jhawthorn, trying ae544b8674b96197bd87736fd13fa6d403327fa0 to workaround the issue for now.

casperisfine:
Seem to have worked, the YJIT crashes are gone. Thank you so much!

byroot:
I merged this as part of https://github.com/ruby/ruby/pull/13556.

Which means the workaround was merged, but it's super easy to remove as soon as YJIT can handle the higher bit being set.

PR #13550, opened 2025-06-06
Take file and line in GC VM locks

peterzhu2118:
This commit adds file and line to GC VM locking functions for debugging purposes and adds upper case macros to pass __FILE__ and __LINE__.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -131,45 +131,45 @@
 #include "shape.h"
 
 unsigned int
-rb_gc_vm_lock(void)
+rb_gc_vm_lock(const char *file, int line)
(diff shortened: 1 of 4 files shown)


ko1:
Compare with this patch, I think using `RB_VM_...` api directly and introducing replacement for mod gc is smarter, but it's only my preference.

peterzhu2118:
I don't have a strong opinion, but my goal is to reduce the number of  `USE_MODULAR_GC` /`BUILDING_MODULAR_GC` because it is divergent code path that need to be maintained separately.

PR #13555, opened 2025-06-07
Selective inspect of instance variables

byroot:
[Feature #21219]

Make Kernel#inspect ask which instance variables should be dumped by the result of `#instance_variables_to_inspect`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,26 @@ Note that each entry is kept to a minimum, see links for details.
 
 Note: We're only listing outstanding class updates.
 
+* Kernel
+
+    * `Kernel#inspect` now check for the existence of a `#instance_variables_to_inspect` method
+      allowing to control which instance variables are displayed in the `#inspect` string:
+
+      ```ruby
+      class DatabaseConfig
+        def initialize(host, user, password)
+          @host = host
+          @user = user
+          @password = password
+        end
+
+        private def instance_variables_to_inspect = [:@host, :@user]
+      end
+
+      conf = DatabaseConfig.new("localhost", "root", "hunter2")
+      conf.inspect #=> #<DatabaseConfig:0x0000000104def350 @host="localhost", @user="root">
+      ```
+
 * Binding
 
     * `Binding#local_variables` does no longer include numbered parameters.
--- a/spec/ruby/core/kernel/inspect_spec.rb
+++ b/spec/ruby/core/kernel/inspect_spec.rb
@@ -28,4 +28,34 @@ describe "Kernel#inspect" do
     end
     obj.inspect.should be_kind_of(String)
   end
+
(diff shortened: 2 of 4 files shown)


NEWS.md
```diff
@@ -14,6 +14,26 @@ Note that each entry is kept to a minimum, see links for details.
 
 Note: We're only listing outstanding class updates.
 
+* Kernel
+
+    * `Kernel#inspect` now check for the existence of a `#instance_variables_to_inspect` method
+      allowing to control which instance variables are displayed in the `#inspect` string:
```
joshuay03, line 20:
```suggestion
    * `Kernel#inspect` now checks for the existence of a `#instance_variables_to_inspect` method,
      allowing control over which instance variables are displayed in the `#inspect` string:
```

spec/ruby/core/kernel/inspect_spec.rb
```diff
@@ -28,4 +28,34 @@ class << obj
     end
     obj.inspect.should be_kind_of(String)
   end
+
+  ruby_version_is "3.5" do
+    it "calls #instance_variables_to_inspect private method to know which variables to display" do
```
joshuay03, line 33: Reading this I assumed it had to be a private method, but the implementation tells me it doesn't. What do you think about removing the mention of private here and making the method public in the empty override case (see next suggestion) so both are covered?

```suggestion
    it "calls #instance_variables_to_inspect method to know which variables to display" do
```

spec/ruby/core/kernel/inspect_spec.rb
```diff
@@ -28,4 +28,34 @@ class << obj
...
+      inspected.should == '#<Object:0x00 @host="localhost", @user="root">'
+
+      obj = Object.new
+      obj.instance_eval do
+        @host = "localhost"
+        @user = "root"
+        @password = "hunter2"
+      end
+      obj.singleton_class.class_eval do
+        private def instance_variables_to_inspect = []
```
joshuay03, line 54:
```suggestion
        def instance_variables_to_inspect = []
```

joshuay03 (commented):
This is going to be so useful! I had a few suggestions while reading this in passing—let me know if you agree with them, and I'd be happy to send a PR.

joshuay03:
> let me know if you agree with them, and I'd be happy to send a PR

Actually it's late here so I've just opened this if you decide to merge them: https://github.com/ruby/ruby/pull/13564

spec/ruby/core/kernel/inspect_spec.rb
byroot: > I assumed it had to be a private method,

No method ever have to be private. e.g. `method_missing` is supposed to be private, but nothing will break if it isn't.

PR #13559, opened 2025-06-09
[DOC] Update _regexp.rdoc examples Regexp.escape

hfl:
Regexp.escape example is wrong.


nobu:
`MatchData#inspect` escapes substrings as well.
The current description is correct.

PR #13560, opened 2025-06-09
rb_ary_size_as_embedded: only consider len, not capacity

byroot:
Repro:
```ruby
def test
  a = [1, 2, 3]
  a << 7 << 8
  a
end

ar = [test, test]
require 'objspace'
puts ObjectSpace.dump(ar[0])
puts ObjectSpace.dump(ar[1])

GC.verify_compaction_references(toward: :empty, expand_heap: true)
puts
puts
puts ObjectSpace.dump(ar[0])
```

In the above script, the array is being moved to size pool 320 instead of size pool 80. This is because the compaction logic consider the array's capacity to compute the ideal size instead of its length.

You could probably make an argument about preserving the capacity, but we're not doing that for strings.

### DIFF
--- a/array.c
+++ b/array.c
@@ -221,7 +221,7 @@ rb_ary_size_as_embedded(VALUE ary)
         real_size = ary_embed_size(ARY_EMBED_LEN(ary));
     }
     else if (rb_ary_embeddable_p(ary)) {
-        real_size = ary_embed_size(ARY_HEAP_CAPA(ary));
+        real_size = ary_embed_size(ARY_HEAP_LEN(ary));
     }
     else {
         real_size = sizeof(struct RArray);


byroot:
@peterzhu2118 I get that, but then String has the same API and does only consider the length.

peterzhu2118:
>  but then String has the same API and does only consider the length.

That's not true.

The function to get a string with a specific capacity only creates non-embedded strings:

https://github.com/ruby/ruby/blob/f9966b9b76706705202f83112e0e2dea0237aea1/string.c#L2035-L2049

If the string is not embedded, then we resize to the capacity:

https://github.com/ruby/ruby/blob/f9966b9b76706705202f83112e0e2dea0237aea1/string.c#L256

byroot:
My bad. I misread the code.

PR #13562, opened 2025-06-09
Fix build on alpine with statx change

Earlopain:
Since https://github.com/ruby/ruby/commit/18a036a6133bd141dfc25cd48ced9a2b78826af6 (https://github.com/ruby/ruby/pull/13474) building on alpine fails because usage of `__u32`, which is not portable.

> file.c: In function 'rb_stat_new':
> file.c:546:33: error: '__u32' undeclared (first use in this function)
>   \# define CP_32(m) .stx_ ## m = (__u32)st->st_ ## m

### DIFF
--- a/file.c
+++ b/file.c
@@ -543,7 +543,7 @@ rb_stat_new(const struct stat *st)
     if (st) {
 #if RUBY_USE_STATX
 # define CP(m) .stx_ ## m = st->st_ ## m
-# define CP_32(m) .stx_ ## m = (__u32)st->st_ ## m
+# define CP_32(m) .stx_ ## m = (uint32_t)st->st_ ## m
 # define CP_TS(m) .stx_ ## m = stat_ ## m ## spec(st)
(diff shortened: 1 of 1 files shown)


nobu:
How `struct statx` is define on Alpine?

Earlopain:
It looks identical, discarding the padding at the end https://git.musl-libc.org/cgit/musl/tree/src/stat/fstatat.c?h=v1.2.5#n10

nobu:
The man page `statx` on Alpine Linux 3.20.3 uses `__u16`, `__u32` and `__u64`.
Which should fix this mismatch, man or musl-libc?

Earlopain:
If I check https://pkgs.alpinelinux.org/package/v3.22/main/x86/man-pages, kernel.org is provided as the url so you're probably looking at the same docs like usual.

I guess it is fine the way it is, but I do not have much knowledge on it.

PR #13564, opened 2025-06-09
Follow-ups to #13555

joshuay03:
Addresses https://github.com/ruby/ruby/pull/13555#pullrequestreview-2909996726


byroot:
I'm fine with the NEWS.md change, but not for the spec change. The spec should test that the implementation is OK with calling a private method.

joshuay03:
I've dropped the spec change 👍🏽

PR #13565, opened 2025-06-09
ZJIT: Add support for putspecialobject

st0012:


vm_insnhelper.c
```diff
@@ -5550,6 +5550,13 @@ vm_get_special_object(const VALUE *const reg_ep,
     }
 }
 
+// ZJIT implementation is using the C function
+// and needs to call a non-static function
+VALUE rb_vm_get_special_object(const VALUE *reg_ep, enum vm_special_object_type type)
```
st0012: I'm not sure if exposing static VM functions for jit is allowed as I didn't see many examples. But we seems to do `rb_vm_concat_array` for YJIT?

st0012 (commented):
So the current tests cover 2 special object types: `VMCORE` and `CBASE`. But I'm not sure how to generate `CONST_BASE` 🤔

vm_insnhelper.c
tekknolagi: This is allowed

tekknolagi (approved):
lgtm! if you want to get fancy, you can add something to `infer_type` that in the case of `VM_SPECIAL_OBJECT_VMCORE` does `Type::from_value(rb_mRubyVMFrozenCore)`

tekknolagi:
> So the current tests cover 2 special object types: `VMCORE` and `CBASE`. But I'm not sure how to generate `CONST_BASE` 🤔

search for `CONST_BASE` in `prism_compile.c` -- looks like there are a couple of uses

tekknolagi:
`for I in []; end`

EDIT: Or even `Foo = 5`

vm_insnhelper.c
tekknolagi: Although maybe we just remove `static` sometimes? What do you think @XrXr / @k0kubun ?

zjit/src/hir.rs
```diff
@@ -138,6 +138,40 @@ impl Invariant {
...
+pub enum SpecialObjectType {
+    VMCore = 1,
+    CBase = 2,
+    ConstBase = 3,
+}
+
+impl From<u32> for SpecialObjectType {
+    fn from(value: u32) -> Self {
+        match value {
+            1 => SpecialObjectType::VMCore,
```
tekknolagi: Actually, since `rb_mRubyVMFrozenCore` is used expressly by YJIT already, we should be able to do the same. So please instead compile the `1` case into a `Const` instruction with `rb_mRubyVMFrozenCore`

test/ruby/test_zjit.rb
```diff
@@ -610,6 +610,18 @@ def test() = @foo = 1
     }
   end
 
+  def test_putspecialobject
```
tekknolagi: Can you please try to cover each subcase of `putspecialobject`, if possible, in the execution tests?

vm_insnhelper.c
k0kubun: I do the wrapper function one when I hope its performance impact on the interpreter will be smaller than just removing static (because static functions may be inlined), but they may actually result in the same compilation result. If you're unsure, just try both and pick whichever you think is better.

Also, like other functions in this file, CRuby uses this style:

```suggestion
VALUE
rb_vm_get_special_object(const VALUE *reg_ep, enum vm_special_object_type type)
```

zjit/src/hir.rs
```diff
@@ -138,6 +138,40 @@ impl Invariant {
...
+            2 => SpecialObjectType::CBase,
+            3 => SpecialObjectType::ConstBase,
```
k0kubun: I think you can use `VM_SPECIAL_OBJECT_*` constants in cruby_bindings.inc instead of magic numbers 1, 2, and 3.

st0012:
@k0kubun @tekknolagi I'm having issues adding test cases for `ConstBase`, which requires either
`Foo = 1` or `class Module::Bar; end` (class defined under a module). And ZJIT unit tests expects the target instruction to be inside a method, but neither of these can be performed inside methods.

For integration tests, I'm trying 

```rb
Foo = 1

def test
  Foo
end

test
```

which I'd expect to have hirs like:

```
$ build/miniruby --zjit-dump-hir --zjit-call-threshold=1 test.rb
HIR:
fn <main>:
bb0(v0:BasicObject):
  v2:Fixnum[1] = Const Value(1)
  v3:BasicObject = PutSpecialObject ConstBase
  SideExit

HIR:
fn test:
bb0(v0:BasicObject):
  v2:BasicObject = GetConstantPath 0x16fa15318
  Return v2
```

But the test fails with

```
TestZJIT#test_putspecialobject_const_base [/Users/hung-wulo/src/github.com/Shopify/ruby/test/ruby/test_zjit.rb:633]:
exited with status 6
stderr:
ZJIT: gen_function: unexpected insn Send { self_val: InsnId(5), call_info: CallInfo { method_name: "lambda" }, cd: 0x1536c4e00, blockiseq: 0x1031d21f0, args: [], state: InsnId(6) }
Failed to compile insn: v7 Send { self_val: InsnId(5), call_info: CallInfo { method_name: "lambda" }, cd: 0x1536c4e00, blockiseq: 0x1031d21f0, args: [], state: InsnId(6) }
ZJIT: gen_function: unexpected insn GetConstantPath { ic: 0x1536c4570 }
Failed to compile insn: v2 GetConstantPath { ic: 0x1536c4570 }
ZJIT: gen_function: unexpected insn SideExit { state: InsnId(2) }
Failed to compile insn: v3 SideExit { state: InsnId(2) }
ZJIT: gen_function: unexpected insn GetConstantPath { ic: 0x1536c45f0 }
Failed to compile insn: v2 GetConstantPath { ic: 0x1536c45f0 }
ruby: ZJIT has panicked. More info to follow...

thread '<unnamed>' panicked at zjit/src/codegen.rs:89:9:
Failed to compile: test@-e:7
fatal runtime error: failed to initiate panic, error 5
```

I'm not sure how those additional insn are generated 🤔

tekknolagi:
We don't have codegen for GetConstantPath right now (but we should add it later (tm)). We have to run it first once in the interpreter so the const IC is filled, then the optimizer will remove GetConstantPath and replace it with a Const instruction

tekknolagi:
I do not know where the lambdas are coming from

k0kubun:
> I'm not sure how those additional insn are generated

In ZJIT tests, we use `assert_compiles`. If you look at that method, it wraps a given script with a lambda, calls it, encode the result object, and writes it into a pipe. So if you call `assert_compiles` with `call_threshold: 1` (default), those things get compiled as well.

But those are compiled before `RubyVM::ZJIT.assert_compiles` forces ZJIT to never fail the compilation. You don't need to fix all these "Failed to compile insn"s for other ISEQs. You still have to make sure ISEQs you passed to `assert_compiles` (like one including `GetConstantPath`) can be compiled correctly though.

st0012:
So I found out why `GetConstantPath` was added: in `optimize_direct_send`, if the constant cache is empty, the `GetConstantPath` will be added back to the insn sets. So to avoid this, I warmed up the constant cache before ZJIT 
  compilation triggers.

zjit/src/hir.rs
```diff
@@ -3956,6 +4010,27 @@ mod tests {
         "#]]);
     }
 
+    #[test]
+    // Tests for ConstBase requires either constant or class definition, both
+    // of which can't be performed inside a method.
+    fn test_putspecialobject_vm_core_and_cbase() {
+        eval("
+            def test
+                alias aliased __callee__
```
```suggestion
              alias aliased __callee__
```

PR #13567, opened 2025-06-09
Get String#crypt working with multi-ractor in cases where !HAVE_CRYPT_R

luke-gruber:
In commit 12f7ba5ed4a, ractor safety was added to String#crypt, however in certain cases it can cause a deadlock. When we lock a native mutex, we cannot allocate ruby objects because they might trigger GC which starts a VM barrier. If the barrier is triggered and other native threads are waiting on this mutex, they will not be able to be woken up in order to join the barrier. To fix this, we don't allocate ruby objects when we hold the lock.

The following could reproduce the problem:

```ruby
strings = []
10_000.times do |i|
  strings << "my string #{i}"
end

STRINGS = Ractor.make_shareable(strings)

rs = []
100.times do
  rs << Ractor.new do
    STRINGS.each do |s|
      s.dup.crypt(s.dup)
    end
  end
end
while rs.any?
  r, obj = Ractor.select(*rs)
  rs.delete(r)
end
```

I will not be adding tests because I am almost finished a PR to enable running test-all test cases inside many ractors at once, which is how I found the issue.

### DIFF
--- a/string.c
+++ b/string.c
@@ -11149,11 +11149,6 @@ rb_str_oct(VALUE str)
 static struct {
     rb_nativethread_lock_t lock;
 } crypt_mutex = {PTHREAD_MUTEX_INITIALIZER};
-
-static void
-crypt_mutex_initialize(void)
-{
-}
 #endif
 
 /*
@@ -11224,6 +11219,7 @@ rb_str_crypt(VALUE str, VALUE salt)
     struct crypt_data *data;
 #   define CRYPT_END() ALLOCV_END(databuf)
 #else
+    char *tmp_buf;
     extern char *crypt(const char *, const char *);
 #   define CRYPT_END() rb_nativethread_lock_unlock(&crypt_mutex.lock)
 #endif
@@ -11258,7 +11254,6 @@ rb_str_crypt(VALUE str, VALUE salt)
 # endif
     res = crypt_r(s, saltp, data);
 #else
-    crypt_mutex_initialize();
     rb_nativethread_lock_lock(&crypt_mutex.lock);
     res = crypt(s, saltp);
 #endif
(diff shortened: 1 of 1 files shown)


string.c
```diff
@@ -11267,8 +11267,15 @@ rb_str_crypt(VALUE str, VALUE salt)
         CRYPT_END();
         rb_syserr_fail(err, "crypt");
     }
-    result = rb_str_new_cstr(res);
+
+    size_t res_size = strlen(res)+1;
+    char *dup = malloc(res_size); // need to hold onto lock while duplicating a potentially static buffer
+    memcpy(dup, res, res_size);
```
jhawthorn: Note for readers: We have to do this because of our `#define strdup ruby_strdup`. We want to malloc here, not xmalloc. And I don't think there's a portable way to determine the size of crypt's output buffer.

string.c
```diff
@@ -11267,8 +11267,15 @@ rb_str_crypt(VALUE str, VALUE salt)
...
     CRYPT_END();
+    // Don't allocate a ruby object while holding this lock, we could hit a VM barrier, which
+    // causes a deadlock if other ractors are waiting on this lock.
+    result = rb_str_new_cstr(dup);
+    free(dup);
```
jhawthorn: If `rb_str_new_cstr` erorrs, this pointer might leak.

I think that may not be worth dealing with as crypt is semi-deprecated https://bugs.ruby-lang.org/issues/14915

I prefer that (very unlikely) leak to the deadlock that will occur in the test suite under ractors.

jhawthorn (approved):

string.c
```diff
@@ -11267,8 +11267,15 @@ rb_str_crypt(VALUE str, VALUE salt)
         CRYPT_END();
         rb_syserr_fail(err, "crypt");
     }
-    result = rb_str_new_cstr(res);
+
+    size_t res_size = strlen(res)+1;
+    char *dup = malloc(res_size); // need to hold onto lock while duplicating a potentially static buffer
```
nobu: The encrypted result would be small enough in `alloca`.
And this copy should not be needed if `crypt_r` is available.

string.c
nobu: `res = (strdup)(res);` can avoid the macro.

string.c
luke-gruber: Thanks for the review. I've updated the PR to use `alloca`.
Edit: Some CI failures I'll look into it Friday.

luke-gruber:
@nobu, @jhawthorn: ready for re-review

PR #13569, opened 2025-06-10
ZJIT: Support get/set on global variables

tenderlove:
Adds support for code like:

```ruby
$foo
$foo = x
```


tekknolagi:
Please change PR title/commit to "ZJIT: ..."

tekknolagi:
Great stuff. Thank you

tekknolagi:
> Both C calls are not leaf and we'll need to do something about it later, but it's good as is for now.

Can you say more? What do you mean?

tenderlove:
> Can you say more? What do you mean?

I think @k0kubun means that they can raise an exception (they aren't leaf functions). @k0kubun isn't that what the snapshot instructions are for?

k0kubun:
During a non-leaf C call, it may dispatch a method that looks at the caller's Binding, for example. Because we don't spill locals onto the interpreter stack before the C call here, we need to either (1) spill them before each non-leaf call or (2) lazily move them from the C stack to the interpreter stack when you need to create a Binding.

We'll start from (1) and then optimize it into (2). I'm working on (1) right now, so it's helpful to leave it as is.

tenderlove:
I guess this is a case where we could hit a tracepoint or warning too. 😢

PR #13570, opened 2025-06-10
Fix handling of cancelled blocking operations.

samuel-williams-shopify:
Cancelled operations were returning `Qundef` causing a `rb_nogvl` to retry the execution of the given function.


samuel-williams-shopify:
cc @ioquatix

PR #13573, opened 2025-06-10
variable.c: Fix `generic_fields_lookup_ensure_size` to avoid use-after-free

casperisfine:
The following diff can cause a use-after-free:

```c
diff --git a/variable.c b/variable.c
index a54bebcec0..53e01c1d60 100644
--- a/variable.c
+++ b/variable.c
@@ -1818,6 +1818,7 @@ generic_fields_lookup_ensure_size(st_data_t *k, st_data_t *v, st_data_t u, int e
         }

         fields_tbl = gen_fields_tbl_resize(fields_tbl, RSHAPE_CAPACITY(fields_lookup->shape_id));
+        rb_gc_start();
         *v = (st_data_t)fields_tbl;
     }
```

```ruby
(2..).step(-1)
```

Since `gen_fields_tbl_resize` uses `xrealloc`, we might trigger GC after the old buffer has been freed, but the pointer in the generic ivar table hasn't been updated yet. So if the `xrealloc` trigger GC, we'll mark a freed buffer.


variable.c
```diff
@@ -1828,6 +1836,7 @@ generic_fields_lookup_ensure_size(st_data_t *k, st_data_t *v, st_data_t u, int e
         rb_obj_set_shape_id(fields_lookup->obj, fields_lookup->shape_id);
     }
 
+    xfree(old_fields_tbl);
```
casperisfine: The `use-after-free` is still there if GC trigger on `xfree` because the table hasn't been updated yet :/

casperisfine: Ok, it can be solved by delaying the `xfree` until after `st_update` returns.

casperisfine:
I think this bug is possible on 3.4.x

casperisfine:
Well, CI proves this wasn't the bug that cause the `test_array.rb` crash :/

casperisfine:
@peterzhu2118 tells me in `xrealloc` GC triggers before the `malloc` so GC shouldn't be able to trigger between `generic_fields_lookup_ensure_size` and the time we write the first attribute, except for when we grow from 0 to 3 capacity, but in that case it's not a problem.

PR #13576, opened 2025-06-10
{Y,Z}JIT: x86: Fix panic writing 32-bit number with top bit set

XrXr:
- **ZJIT: Revive most x64 backend tests**
- **ZJIT: Revive some A64 backend tests to fix unused warning**
- **YJIT: x86: Fix panic writing 32-bit number with top bit set**
- **ZJIT: x86: Fix panic writing 32-bit number with top bit set**

The test changes muddy up the diff a bit, but the actual fix isn't too complicated.

### DIFF
--- a/zjit/src/asm/x86_64/tests.rs
+++ b/zjit/src/asm/x86_64/tests.rs
@@ -1,11 +1,10 @@
 #![cfg(test)]
 
-//use crate::asm::x86_64::*;
+use crate::asm::x86_64::*;
 
-/*
 /// Check that the bytes for an instruction sequence match a hex string
 fn check_bytes<R>(bytes: &str, run: R) where R: FnOnce(&mut super::CodeBlock) {
-    let mut cb = super::CodeBlock::new_dummy(4096);
+    let mut cb = super::CodeBlock::new_dummy();
     run(&mut cb);
     assert_eq!(format!("{:x}", cb), bytes);
 }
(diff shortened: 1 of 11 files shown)


zjit/src/asm/x86_64/tests.rs
```diff
@@ -194,6 +194,7 @@ fn test_mov() {
     check_bytes("48c7470801000000", |cb| mov(cb, mem_opnd(64, RDI, 8), imm_opnd(1)));
     //check_bytes("67c7400411000000", |cb| mov(cb, mem_opnd(32, EAX, 4), imm_opnd(0x34))); // We don't distinguish between EAX and RAX here - that's probably fine?
     check_bytes("c7400411000000", |cb| mov(cb, mem_opnd(32, RAX, 4), imm_opnd(17)));
+    check_bytes("c7400401000080", |cb| mov(cb, mem_opnd(32, RAX, 4), uimm_opnd(0x80000001)));
```
k0kubun, line 196: This whole file is still commented out. I think this might just work today. Could you re-enable the tests?

zjit/src/assertions.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+        assert_eq!(format!("{:x}", $cb), $hex);
+    };
+}
+#[cfg(test)]
+pub(crate) use assert_disasm;
+
+/// Remove the minimum indent from every line, skipping the first line if `skip_first`.
+#[cfg(all(feature = "disasm", test))]
+pub fn unindent(string: &str, trim_lines: bool) -> String {
```
k0kubun: We have a refactored version of `unindent()` in `zjit/src/cruby.rs`. Could you drop either of these?

PR #13578, opened 2025-06-10
Fix RubyVM::Shape.transition_tree

etiennebarrie:
It's been broken since 3.3.0, a bisect points to 7c307e0379e3c6c07d821b863fefbdfdfc84c4f1 which seems to match.


byroot:
It's weird, CI fail on some syntax error but I can't see it:

```
 ../src/shape.c: In function ‘edges’:
../src/shape.c:1402:5: warning: no return statement in function returning non-void [-Wreturn-type]
 1402 |     }
      |     ^
../src/shape.c: At top level:
../src/shape.c:1403:5: error: expected identifier or ‘(’ before ‘return’
 1403 |     return hash;
      |     ^~~~~~
../src/shape.c:1404:1: error: expected identifier or ‘(’ before ‘}’ token
 1404 | }
      | ^
```

shape.c
```diff
@@ -1391,12 +1391,14 @@ static enum rb_id_table_iterator_result collect_keys_and_values(ID key, VALUE va
...
-        collect_keys_and_values(child->edge_name, (VALUE)child, &hash);
-    }
-    else {
-        rb_managed_id_table_foreach(edges, collect_keys_and_values, &hash);
+    if (edges) {
+        if (SINGLE_CHILD_P(edges)) {
+            rb_shape_t *child = SINGLE_CHILD(edges);
+            collect_keys_and_values(child->edge_name, (VALUE)child, &hash);
+        }
+        else
```
byroot:
```suggestion
        else {
```
Oh it's here.

etiennebarrie: erf sorry I switch from an `else if (edges != 0)` at the last minute to keep consistency with `rb_shape_edges` and completely failed 🤦

PR #13579, opened 2025-06-11
ZJIT: Write a callee frame on JIT-to-JIT calls

k0kubun:
This PR changes JIT-to-JIT calls to actually write a callee frame that is about to be pushed.

When JIT code makes a non-leaf call, it may dispatch an arbitrary method that looks at caller frames. When that happens, the method may read the locals, PC, ISEQ, and other things of the frame. Because that happens before the frame side-exits, we need to somehow make them available when needed.

As a first implementation to make that work, this PR writes them when a JIT frame is pushed. ZJIT should ideally optimize away the overhead, but I think it's nice to make things work first as we're onboarding more people to work on ZJIT.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -258,7 +258,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::IfTrue { val, target } => return gen_if_true(jit, asm, opnd!(val), target),
         Insn::IfFalse { val, target } => return gen_if_false(jit, asm, opnd!(val), target),
         Insn::SendWithoutBlock { call_info, cd, state, self_val, args, .. } => gen_send_without_block(jit, asm, call_info, *cd, &function.frame_state(*state), self_val, args)?,
-        Insn::SendWithoutBlockDirect { iseq, self_val, args, .. } => gen_send_without_block_direct(cb, jit, asm, *iseq, opnd!(self_val), args)?,
+        Insn::SendWithoutBlockDirect { cme, iseq, self_val, args, state, .. } => gen_send_without_block_direct(cb, jit, asm, *cme, *iseq, opnd!(self_val), args, &function.frame_state(*state))?,
         Insn::Return { val } => return Some(gen_return(asm, opnd!(val))?),
         Insn::FixnumAdd { left, right, state } => gen_fixnum_add(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state))?,
         Insn::FixnumSub { left, right, state } => gen_fixnum_sub(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state))?,
@@ -464,8 +464,15 @@ fn gen_send_without_block(
     self_val: &InsnId,
     args: &Vec<InsnId>,
 ) -> Option<lir::Opnd> {
-    // Spill the receiver and the arguments onto the stack. They need to be marked by GC and may be caller-saved registers.
+    // Spill locals onto the stack.
+    asm_comment!(asm, "spill locals");
+    for (idx, &insn_id) in state.locals().enumerate() {
+        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
+    }
+    // Spill the receiver and the arguments onto the stack.
+    // They need to be on the interpreter stack to let the interpreter access them.
     // TODO: Avoid spilling operands that have been spilled before.
+    asm_comment!(asm, "spill receiver and arguments");
     for (idx, &insn_id) in [*self_val].iter().chain(args.iter()).enumerate() {
         // Currently, we don't move the SP register. So it's equal to the base pointer.
         let stack_opnd = Opnd::mem(64, SP, idx as i32  * SIZEOF_VALUE_I32);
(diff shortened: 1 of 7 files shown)


zjit/src/codegen.rs
```diff
@@ -464,8 +464,15 @@ fn gen_send_without_block(
     self_val: &InsnId,
     args: &Vec<InsnId>,
 ) -> Option<lir::Opnd> {
-    // Spill the receiver and the arguments onto the stack. They need to be marked by GC and may be caller-saved registers.
+    // Spill locals onto the stack.
```
tekknolagi, line 467:
```suggestion
    // Spill locals onto the stack.
    // TODO: Don't spill locals eagerly; lazily reify frames
```

zjit/src/codegen.rs
```diff
@@ -495,10 +502,40 @@ fn gen_send_without_block_direct(
...
+    // Set up the new frame
+    // TODO: Lazily materialize caller frames on side exits or when needed
+    gen_push_frame(asm, args.len(), state, ControlFrame {
+        recv,
+        iseq,
+        cme,
+        frame_type: VM_FRAME_MAGIC_METHOD | VM_ENV_FLAG_LOCAL,
+    });
+
+    asm_comment!(asm, "switch to new SP register");
```
tekknolagi, line 535:
```suggestion
    asm_comment!(asm, "pop locals and stack off VM stack");
```

zjit/src/codegen.rs
```diff
@@ -517,7 +554,14 @@ fn gen_send_without_block_direct(
     jit.branch_iseqs.push((branch.clone(), iseq));
     // TODO(max): Add a PatchPoint here that can side-exit the function if the callee messed with
     // the frame's locals
-    Some(asm.ccall_with_branch(dummy_ptr, c_args, &branch))
+    let ret = asm.ccall_with_branch(dummy_ptr, c_args, &branch);
+
+    // If a callee side-exits, i.e. returns Qundef, propagate the return value to the caller.
```
tekknolagi, line 560:
```suggestion
    // If a callee side-exits, i.e. returns Qundef, propagate the return value to the caller.
    // The caller will side-exit the callee into the interpreter.
```

zjit/src/state.rs
```diff
@@ -88,9 +88,14 @@ impl ZJITState {
...
             method_annotations: cruby_methods::init(),
-            iseq_return_addrs: HashSet::new(),
+            exit_trampoline: None,
         };
         unsafe { ZJIT_STATE = Some(zjit_state); }
+
+        // Generate trampolines after initializing ZJITState, which Assembler will use
+        let cb = ZJITState::get_code_block();
+        let exit_trampoline = Self::gen_exit_trampoline(cb).unwrap();
+        ZJITState::get_instance().exit_trampoline = Some(exit_trampoline);
```
tekknolagi, line 98: Do you need to make a new option? You should be able to set `exit_trampoline` directly, it looks like

tekknolagi (approved):
After our conversation, overall it looks good.

I look forward to making this much lazier in the future.

I feel like we should have more tests, and more rigorous tests, for this new feature. Can we add some more to exercise this?

zjit/src/state.rs
k0kubun: The intention here was to assert that the trampoline is correctly set up early (you shouldn't hit OOM when you use the code region for the first time, but you can't be sure unless you have an assertion when you're investigating a bug) instead of failing later during codegen.

zjit/src/codegen.rs
k0kubun: They're not popped here. The caller's locals and stack are intact, and part of the stack is used as part of the callee's locals. We're rather pushing new locals (`local_size - args.len()`) and env data for the callee. 

I think the point here is that this piece of code is updating the register for JIT rather than the frame for the interpreter, which `gen_push_frame` takes care of. So the current comment makes more sense to me.

zjit/src/codegen.rs
k0kubun: applied at https://github.com/ruby/ruby/pull/13579/commits/bc214d2c61ad29f7d820a409481a0307d5ed4a80

zjit/src/codegen.rs
k0kubun: thanks, I forgot to leave a TODO for this one https://github.com/ruby/ruby/pull/13579/commits/829a7314f1033fc03c2e52990a8370295e1309ab

zjit/src/codegen.rs
tekknolagi: oh, oh, it grows the other direction. my bad - thank you

zjit/src/state.rs
tekknolagi: sg

k0kubun:
> I feel like we should have more tests, and more rigorous tests, for this new feature. Can we add some more to exercise this?

I added tests that dump non-argument locals on side exits https://github.com/ruby/ruby/pull/13579/commits/5b621b11052e772c3bd0caef23f9eda7753d881a and one that checks `cfp->pc` https://github.com/ruby/ruby/pull/13579/commits/7916f3dcc18002326c44d0494bdbfe2da6a21938.

Another test target would be the content of dumped locals. But to test the content of locals from a callee, we'd need to load a C extension in `test/-ext-/debug/test_debug.rb` like YJIT does. We're still not ready for running the current `make test-all` with ZJIT enabled.

Do you have other things you want to be tested in mind? We could just "have more tests" as you said, but I generally try to avoid bloating the number of tests by adding meaningless/duplicated test cases.

tekknolagi:
I don't mean to bloat the tests but did not (at the time of comment) have a good set of recommendations for things to test. 2 tests for a bunch of new functionality just seemed a bit light. I can imagine having a bunch of dark untested corner cases

k0kubun:
I think we should aim for enabling `make btest`, `make test-all`, and `make test-spec` on CI with ZJIT as soon as possible. They would test more corner cases than we can think of and write in `test_zjit.rb`. Even then, you'd encounter dark untested corner cases though. Anything we can come up with today is probably not a dark corner case.

This project should have got the test setup much earlier, but we were so focused on keeping generated code optimal that we never had the "rigorous tests" we could have got for free. This PR should bring us much closer to having it.

k0kubun:
You approved the PR and the CI passed, so I'll merge this PR to unblock your stacked PR. We can add more tests as we come up with untested scenarios that are worth testing.

PR #13584, opened 2025-06-11
do not need to keep interrupt_lock

ko1:
because no other threads are running just after fork/exec.


ko1:
@copilot resolve the merge conflicts in this pull request

PR #13587, opened 2025-06-11
fix: `Date#hash` for large years

Stranger6667:
Addresses https://bugs.ruby-lang.org/issues/21437

I am not 100% sure about the fix - this is the first time I am contributing to the C part of the codebase. Please let me know if the fix is sufficient


jeremyevans:
Please submit this upstream: https://github.com/ruby/date

Stranger6667:
Aha, thanks :) will do - I was about to create another PR (for a different issue), noticed your message right on time

PR #13588, opened 2025-06-11
ZJIT: Parse opt freeze insns to HIR

composerinteralia:
* `opt_hash_freeze`
* `opt_ary_freeze`
* `opt_str_freeze`
* `opt_str_uminus`

Largely borrowed from `opt_neq`, but there are no args for `freeze`

Fixes https://github.com/Shopify/ruby/issues/582
Fixes https://github.com/Shopify/ruby/issues/583
Fixes https://github.com/Shopify/ruby/issues/584
Fixes https://github.com/Shopify/ruby/issues/585

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -255,6 +255,34 @@ class TestZJIT < Test::Unit::TestCase
     }, insns: [:opt_ge], call_threshold: 2
   end
 
+  def test_opt_hash_freeze
+    assert_compiles '{}', <<~RUBY, insns: [:opt_hash_freeze]
+      def test = {}.freeze
+      test
+    RUBY
+  end
+
+  def test_opt_ary_freeze
+    assert_compiles '[]', <<~RUBY, insns: [:opt_ary_freeze]
+      def test = [].freeze
+      test
+    RUBY
+  end
+
+  def test_opt_str_freeze
+    assert_compiles '""', <<~RUBY, insns: [:opt_str_freeze]
+      def test = "".freeze
+      test
+    RUBY
+  end
+
+  def test_opt_str_uminus
+    assert_compiles '""', <<~RUBY, insns: [:opt_str_uminus]
+      def test = -""
+      test
+    RUBY
+  end
+
   def test_new_array_empty
     assert_compiles '[]', %q{
       def test = []
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -151,7 +151,9 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
                 write!(f, "BOPRedefined(")?;
                 match klass {
                     INTEGER_REDEFINED_OP_FLAG => write!(f, "INTEGER_REDEFINED_OP_FLAG")?,
+                    STRING_REDEFINED_OP_FLAG => write!(f, "STRING_REDEFINED_OP_FLAG")?,
                     ARRAY_REDEFINED_OP_FLAG => write!(f, "ARRAY_REDEFINED_OP_FLAG")?,
+                    HASH_REDEFINED_OP_FLAG => write!(f, "HASH_REDEFINED_OP_FLAG")?,
                     _ => write!(f, "{klass}")?,
                 }
                 write!(f, ", ")?;
@@ -167,6 +169,8 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
                     BOP_LE    => write!(f, "BOP_LE")?,
                     BOP_GT    => write!(f, "BOP_GT")?,
                     BOP_GE    => write!(f, "BOP_GE")?,
+                    BOP_FREEZE => write!(f, "BOP_FREEZE")?,
+                    BOP_UMINUS => write!(f, "BOP_UMINUS")?,
                     BOP_MAX    => write!(f, "BOP_MAX")?,
                     _ => write!(f, "{bop}")?,
                 }
@@ -1199,6 +1203,38 @@ impl Function {
         }
     }
 
+    fn rewrite_if_frozen(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId, klass: u32, bop: u32) {
+        let self_type = self.type_of(self_val);
+        if let Some(obj) = self_type.ruby_object() {
+            if obj.is_frozen() {
+                self.push_insn(block, Insn::PatchPoint(Invariant::BOPRedefined { klass, bop }));
+                self.make_equal_to(orig_insn_id, self_val);
+                return;
+            }
+        }
+        self.push_insn_id(block, orig_insn_id);
+    }
+
+    fn try_rewrite_freeze(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId) {
+        if self.is_a(self_val, types::StringExact) {
+            self.rewrite_if_frozen(block, orig_insn_id, self_val, STRING_REDEFINED_OP_FLAG, BOP_FREEZE);
+        } else if self.is_a(self_val, types::ArrayExact) {
+            self.rewrite_if_frozen(block, orig_insn_id, self_val, ARRAY_REDEFINED_OP_FLAG, BOP_FREEZE);
+        } else if self.is_a(self_val, types::HashExact) {
+            self.rewrite_if_frozen(block, orig_insn_id, self_val, HASH_REDEFINED_OP_FLAG, BOP_FREEZE);
+        } else {
+            self.push_insn_id(block, orig_insn_id);
+        }
+    }
+
+    fn try_rewrite_uminus(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId) {
+        if self.is_a(self_val, types::StringExact) {
+            self.rewrite_if_frozen(block, orig_insn_id, self_val, STRING_REDEFINED_OP_FLAG, BOP_UMINUS);
+        } else {
+            self.push_insn_id(block, orig_insn_id);
+        }
+    }
+
     /// Rewrite SendWithoutBlock opcodes into SendWithoutBlockDirect opcodes if we know the target
     /// ISEQ statically. This removes run-time method lookups and opens the door for inlining.
     fn optimize_direct_sends(&mut self) {
@@ -1229,6 +1265,10 @@ impl Function {
                         self.try_rewrite_fixnum_op(block, insn_id, &|left, right| Insn::FixnumGt { left, right }, BOP_GT, self_val, args[0], state),
                     Insn::SendWithoutBlock { self_val, call_info: CallInfo { method_name }, args, state, .. } if method_name == ">=" && args.len() == 1 =>
                         self.try_rewrite_fixnum_op(block, insn_id, &|left, right| Insn::FixnumGe { left, right }, BOP_GE, self_val, args[0], state),
+                    Insn::SendWithoutBlock { self_val, call_info: CallInfo { method_name }, args, .. } if method_name == "freeze" && args.len() == 0 =>
+                        self.try_rewrite_freeze(block, insn_id, self_val),
+                    Insn::SendWithoutBlock { self_val, call_info: CallInfo { method_name }, args, .. } if method_name == "-@" && args.len() == 0 =>
+                        self.try_rewrite_uminus(block, insn_id, self_val),
                     Insn::SendWithoutBlock { mut self_val, call_info, cd, args, state } => {
                         let frame_state = self.frame_state(state);
                         let (klass, guard_equal_to) = if let Some(klass) = self.type_of(self_val).runtime_exact_ruby_class() {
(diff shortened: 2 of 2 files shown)


composerinteralia:
(Max helped us get started here, but I'm still hoping for a thorough review since I'm new to this project and I want to know where I've done something wrong.)

test/ruby/test_zjit.rb
```diff
@@ -255,6 +255,34 @@ def test(a, b) = a >= b
     }, insns: [:opt_ge], call_threshold: 2
   end
 
+  def test_opt_hash_freeze
+    assert_compiles '{}', <<~RUBY, insns: [:opt_hash_freeze]
+      def test = {}.freeze
+      test
```
composerinteralia, line 261: In the future it might make sense to use `frozen?` in test tests, but I don't think we can compile builtins quite yet (perhaps I could look into that next).

tekknolagi: You can call that on the result of `test`; that is not compiled

tekknolagi: Like `a = test; [a, a.frozen?]`

tekknolagi: (here and below)

zjit/src/hir.rs
```diff
@@ -1199,6 +1203,34 @@ impl Function {
         }
     }
 
+    fn try_rewrite_freeze(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId) {
+        match self.find(self_val) {
+            Insn::Const { .. } if self.is_a(self_val, types::StringExact) => {
```
tekknolagi: You don't need to check what instruction produced `self_val`, only that `self.is_a(self_val, types::StringExact)` *and it is already frozen*.

Also, I'm not sure of the resurrection semantics here. Is it legal to rewrite `a.freeze` into `a` if `a` is frozen but in the bytecode stream and in that weird GC state? cc @k0kubun and @XrXr

zjit/src/hir.rs
```diff
@@ -2361,6 +2397,32 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let send = fun.push_insn(block, Insn::SendWithoutBlock { self_val: recv, call_info: CallInfo { method_name }, cd, args, state: exit_id });
                     state.stack_push(send);
                 }
+                YARVINSN_opt_hash_freeze |
+                YARVINSN_opt_ary_freeze |
+                YARVINSN_opt_str_freeze |
+                YARVINSN_opt_str_uminus => {
+                    let cd: *const rb_call_data = get_arg(pc, 1).as_ptr();
```
tekknolagi, line 2409: Can you add a NB about what the 0th arg is

zjit/src/hir.rs
```diff
@@ -2361,6 +2397,32 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                    let call_info = unsafe { rb_get_call_data_ci(cd) };
+                    if unknown_call_type(unsafe { rb_vm_ci_flag(call_info) }) {
+                        // Unknown call type; side-exit into the interpreter
+                        let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                        fun.push_insn(block, Insn::SideExit { state: exit_id });
+                        break;  // End the block
+                    }
+                    let argc = unsafe { vm_ci_argc((*cd).ci) };
+                    assert_eq!(0, argc, "freeze should not have args");
```
tekknolagi: This is not just freeze but also uminus -- either figure out a way to get the opcode name here or just say "freeze/uminus [...]"

test/ruby/test_zjit.rb
composerinteralia: I still get `Failed to compile: frozen?@<internal:kernel>:69` with:

```
  def test_opt_hash_freeze
    assert_compiles '[{}, true]', <<~RUBY, insns: [:opt_hash_freeze]
      def test = {}.freeze
      a = test
      [a, a.frozen?]
    RUBY
  end
```

It seems like it's trying to compile `Kernel#frozen?` but failing on one of the `invokebuiltin` type instructions.

zjit/src/hir.rs
composerinteralia: > and it is already frozen

Ah OK. This was the part I was least confident with so I'm not surprised I didn't get it quite right. I may need to learn a bit more first. I had thought that since these instructions all have a path where they [return the object directly without a frozen check](https://github.com/ruby/ruby/blob/95201299fd7bf0918dfbd8c127ce2b5b33ffa537/vm_insnhelper.c#L6121-L6152) that we'd be able to take a similar path here. Are you saying we need an explicit check along the lines of `rb_obj_frozen_p` or similar? Or is there some other way to know whether what we've got here is frozen?

test/ruby/test_zjit.rb
tekknolagi: Hm, that is weird. It should only be trying to compile `test` as far as I understand it. I can look into this later tonight or tomorrow

zjit/src/hir.rs
composerinteralia: I'm using `get_arg(pc, 0)` below, so maybe it'd be clearer if I pull that up.

zjit/src/hir.rs
k0kubun: > Also, I'm not sure of the resurrection semantics here. Is it legal to rewrite a.freeze into a if a is frozen but in the bytecode stream and in that weird GC state?

If you look at `vm_opt_str_freeze()` and `opt_str_freeze`, all it does is:

1. If the BOP is not redefined, push the string object in the bytecode as is
2. If the BOP is redefined, duplicate (resurrect) the string object in the bytecode and dispatch `#freeze` on it

Because it's frozen (and fstrings are de-duplicated, btw), you shouldn't need to resurrect it for (1).

zjit/src/hir.rs
```diff
@@ -1229,6 +1261,10 @@ impl Function {
                         self.try_rewrite_fixnum_op(block, insn_id, &|left, right| Insn::FixnumGt { left, right }, BOP_GT, self_val, args[0], state),
                     Insn::SendWithoutBlock { self_val, call_info: CallInfo { method_name }, args, state, .. } if method_name == ">=" && args.len() == 1 =>
                         self.try_rewrite_fixnum_op(block, insn_id, &|left, right| Insn::FixnumGe { left, right }, BOP_GE, self_val, args[0], state),
+                    Insn::SendWithoutBlock { self_val, call_info: CallInfo { method_name }, args, .. } if method_name == "freeze" && args.len() == 0 =>
+                        self.try_rewrite_freeze(block, insn_id, self_val),
```
k0kubun, line 1269: These frozen optimizations are only applied to object literals immediately followed by a `.freeze`, so I don't think we should go through `Insn::SendWithoutBlock` for this unlike Fixnum optimizations that also apply to non-literal receivers. The patchpoint + no-op should be directly converted from `opt_*_freeze`.

For example, `"a".freeze` and `("a").freeze` behave differently, but I guess this would not distinguish them. `"a".freeze` is a special literal and not just a `#freeze` call on a StringExact constant.

composerinteralia: Ah yeah, that makes sense to me. @ywenc I think this is also something you mentioned during our last pairing. Thank you!

(This currently does distinguish `"a".freeze` and `("a").freeze`, but only because I matched on `Insn::Const` and the one with parentheses is a `StringCopy`.)

tekknolagi: We can interrogate the object in the Type, though; we can add a primitive to ask if the object is frozen. So it could work on arbitrary objects not just literals

zjit/src/hir.rs
tekknolagi: No, I think in general variables should be defined close to their uses, but a note here saying that not a typo would be nice

zjit/src/hir.rs
tekknolagi: Amazing!

zjit/src/hir.rs
k0kubun: > So it could work on arbitrary objects not just literals

Can you provide an example Ruby script you want to cover with this?

k0kubun: > Can you provide an example Ruby script you want to cover with this?

I gave more thoughts on this. One valid example might be `"a".freeze.freeze`. On the interpreter, the second `.freeze` becomes a regular method call, but it seems fair to optimize away the second call as a compiler optimization.

I guess, if `Insn::Const` has the semantics that the `val` is either an immediate or a frozen object, it's alright. I don't know if every instruction that turns into `Insn::Const` satisfies that though.

tekknolagi: We can ask - since we know the pointer - if it's frozen when we're doing the optimization, no?

composerinteralia: > since we know the pointer

Is that only true when it's a `Insn::Const` though? (You mentioned above that we didn't need to match on `Insn::Const` but I'm having trouble reasoning about how we'd check at compile time for other instructions like `Insn::NewArray` or whatever)

composerinteralia: Maybe it makes sense to do both things here? For the opt freezes where we know we can return the value directly without checking, we can skip the `Insn::SendWithoutBlock` and compile directly to the optimized version with a bop check. But then we can still keep the direct send optimization (modified to check that the value is actually frozen) for cases like the second freeze in  `"".freeze.freeze`

tekknolagi: The key part of this is that we don't need to check the instruction, we need to check the type. The type can hold the pointer. So the API looks like:

```rust
let self_type = self.type_of(self_val);
if let Some(obj) = self_type.ruby_value() {
  if obj_is_frozen(obj) {
    // do a thing
  }
}
```

We don't need to know what instruction produced it. Maybe it's a `GuardBitEqual` or something, not a `Const` -- the question we are asking can and should be answered by looking at its `Type`

k0kubun: We're talking about two different scenarios:

1. `"string".freeze` ("frozen string literal"): `opt_str_freeze` in YARV. The interpreter doesn't make a method call of `#freeze`.
    * Because the interpreter doesn't, there's no reason for ZJIT to make a call either.
2. `string.freeze` (`string` is any expression that ZJIT knows is StringExact and frozen): `opt_send_without_block` in YARV. The interpreter makes a C method call of `#freeze`.
    * While the interpreter makes a call, because `rb_str_freeze` just returns the receiver when it's frozen, it's probably fair to optimize away such a C method call in ZJIT when the receiver is known to be frozen.

I'm fine with optimizing (2) as well as (1), but if we do (2), we definitely need `obj_is_frozen(obj)` API that Max mentioned. The frozen property isn't guaranteed for the general case (2) unlike the `opt_str_freeze` case (1).

composerinteralia: Thank you both very much for holding my hand a bit through this one. I think I'm understanding everything now (although let know if things look otherwise). I just pushed up some changes based on the feedback here.

test/ruby/test_zjit.rb
composerinteralia: I also have `invokebuiltin`/`invokebuiltin_delegate` semi-working (working for builtins without args so far—I'm still working on getting it working properly with args, but I might need some help with that), so I can go back and add the `frozen?` as part of that PR too.

zjit/src/hir.rs
```diff
@@ -1199,6 +1203,44 @@ impl Function {
...
+            Some(STRING_REDEFINED_OP_FLAG)
+        } else if self_type.is_subtype(types::ArrayExact) {
+            Some(ARRAY_REDEFINED_OP_FLAG)
+        } else if self_type.is_subtype(types::HashExact) {
+            Some(HASH_REDEFINED_OP_FLAG)
+        } else {
+            None
+        };
+
+        if let Some(obj) = self_type.ruby_object() {
```
composerinteralia, line 1208: Does this triple `if` look weird? Maybe it's better if I just write out the full `push_insn` 3 times? Let me know if you think I should keep playing around with it.

tekknolagi: It's a little awkward but I think it's fine for now. Means our APIs are a little awkward. We can think about it and come back to this later since it's a small change

zjit/src/hir.rs
```diff
@@ -2075,6 +2121,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
             let exit_state = state.clone();
             profiles.profile_stack(&exit_state);
 
+
```
```suggestion
```

zjit/src/hir.rs
```diff
@@ -5297,4 +5428,107 @@ mod opt_tests {
...
+              v3:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+              PatchPoint BOPRedefined(STRING_REDEFINED_OP_FLAG, BOP_FREEZE)
+              Return v3
+        "#]]);
+    }
+
+    #[test]
+    fn test_opt_str_freeze_refreeze() {
+        eval("
+            def test = -''.freeze.freeze
```
tekknolagi: Did you mean to also do uminus here?

tekknolagi (approved):

test/ruby/test_zjit.rb
tekknolagi: I am going to investigate why it's failing here

tekknolagi: But continue on your way, no need to wait up

tekknolagi: Oh. @k0kubun we're trying to call a built-in Ruby method that has an opcode we don't support. We currently have an assumption that if we're direct calling a specific ISEQ, we *must* be able to compile it, or we bail out of compiling the whole sub callgraph. Is there another way we can do this?

That is, if we know A calls ISEQ B, but B has some unknown instruction, can we do a more generic call to B without doing this require-it-to-be-jointly-JITed dance?

k0kubun: I guess https://github.com/ruby/ruby/pull/13579 + https://github.com/ruby/ruby/pull/13602 solves it?

zjit/src/hir.rs
composerinteralia: I did, but I'll split it off into another test to make things clearer.

test/ruby/test_zjit.rb
tekknolagi: It partially solves it, and maybe fully *if* we get codegen for all our HIR instructions in the near future

zjit/src/hir.rs
composerinteralia: After taking a break I found something I like a little better.

zjit/src/hir.rs
```diff
@@ -5297,4 +5421,121 @@ mod opt_tests {
               Return v2
         "#]]);
     }
+
+    #[test]
+    fn test_opt_hash_freeze() {
```
tekknolagi: A couple of test nits:

a) Can we please do tests that are descriptive of what they're testing? For example, `test_elide_freeze_with_frozen_hash` or something similar
b) Can you please also test that we don't do this optimization if there are args present

test/ruby/test_zjit.rb
k0kubun: Or codegen should generate a side exit without being instructed by HIR when it sees an unsupported HIR instruction. We'd need a `FrameState` but I guess we can use the last-seen `Snapshot`'s `FrameState`?

tekknolagi: Hmmmmmmmmm maybe

tekknolagi (approved):
ship it! :D

PR #13589, opened 2025-06-11
Fix generic_ivar_set_shape_field for table rebuild

jhawthorn:
Previously GC could trigger a table rebuild of the generic fields st_table in the middle of calling the st_update callback. This could cause entries to be reallocated or rearranged and the update to be for the wrong entry.

This commit adds an assertion to make that case easier to detect, and replaces the st_update with a separate st_lookup and st_insert.

This should be an additional fix to the issue we've been seeing in `test_array.rb` in CI (which has also had another workaround).

We used this as a (somewhat) reliable reproduction:

``` ruby
objs = 10_000.times.map do
  a = []
  a.instance_variable_set(:@a, 1)
  a
end

GC.stress = true
GC.auto_compact = true

steps = 1000.times.map do
  a = []
  a.instance_variable_set(:@a, 1)
  a.instance_variable_set(:@b, 1)
  a.instance_variable_set(:@c, 1)
  a.instance_variable_set(:@d, 1)
  a.instance_variable_set(:@e, 1)
  a
end

GC.stress = false
GC.auto_compact = false
```

### DIFF
--- a/st.c
+++ b/st.c
@@ -1495,7 +1495,16 @@ st_update(st_table *tab, st_data_t key,
         value = entry->record;
     }
     old_key = key;
+
+    unsigned int rebuilds_num = tab->rebuilds_num;
+
     retval = (*func)(&key, &value, arg, existing);
+
+    // We need to make sure that the callback didn't cause a table rebuild
+    // Ideally we would make sure no operations happened
+    assert(rebuilds_num == tab->rebuilds_num);
+    (void)rebuilds_num;
+
     switch (retval) {
       case ST_CONTINUE:
         if (! existing) {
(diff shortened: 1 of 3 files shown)


byroot (approved):

byroot:
We probably should check older versions, I fear this bug is present on at least 3.4, possibly older.

However to trigger on older versions you need 9 ivars (or 10 in debug mode).

casperisfine:
I was able to repro on 3.4, and the code had a similar shape for way over a decade, so I opened a report so we can backport: https://bugs.ruby-lang.org/issues/21438

byroot:
@jhawthorn shall we merge or is there some more changes you want to make?

PR #13597, opened 2025-06-12
[Bug #21439] Fix PM_SPLAT_NODE compilation error in for loops

paracycle:
This commit fixes a crash that occurred when using splat nodes (*) as the index variable in for loops. The error "Unexpected node type for index in for node: PM_SPLAT_NODE" was thrown because the compiler didn't know how to handle splat nodes in this context.

The fix allows code like `for *x in [[1,2], [3,4]]` to compile and execute correctly, where the splat collects each sub-array.


eileencodes (approved):

kddnewton (commented):
Thank you!

PR #13600, opened 2025-06-12
Add locks around accesses/modifications to global encodings table

luke-gruber:
Add locks around accesses/modifications to global encodings table

This fixes segfaults and errors of the type "Encoding not found" when
using encoding-related methods and internal encoding c functions across
ractors.

Example of a possible segfault in release mode or assertion error in debug mode:

```ruby
rs = []
100.times do
  rs << Ractor.new do
    "abc".force_encoding(Encoding.list.shuffle.first)
  end
end
while rs.any?
  r, obj = Ractor.select(*rs)
  rs.delete(r)
end
```

### DIFF
--- a/encoding.c
+++ b/encoding.c
@@ -93,12 +93,16 @@ static rb_encoding *global_enc_ascii,
                    *global_enc_utf_8,
                    *global_enc_us_ascii;
 
+// re-entrant lock
 #define GLOBAL_ENC_TABLE_LOCKING(tbl) \
     for (struct enc_table *tbl = &global_enc_table, **locking = &tbl; \
          locking; \
          locking = NULL) \
         RB_VM_LOCKING()
+#define GLOBAL_ENC_TABLE_LOCK_ENTER_LEV(tbl, lev) struct enc_table *tbl = &global_enc_table; RB_VM_LOCK_ENTER_LEV(lev)
+#define GLOBAL_ENC_TABLE_LOCK_LEAVE_LEV(lev) RB_VM_LOCK_LEAVE_LEV(lev)
 
+#define ASSERT_GLOBAL_ENC_TABLE_LOCKED() ASSERT_vm_locking()
 
 #define ENC_DUMMY_FLAG (1<<24)
 #define ENC_INDEX_MASK (~(~0U<<24))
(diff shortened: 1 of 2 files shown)


zzak:
Could you add a test and/or reproduction case?

Was there a ticket associated with this PR?

jhawthorn:
> Could you add a test and/or reproduction case?

It tends to be difficult to make reliable reproductions to concurrency issues, and at the moment all Ractor tests need an `assert_separately`, which is less than desirable. It's better to make things safe based on correct design and in this case it's pretty obvious that we'll need to hold a lock to read or write to a global st_table (and the ASSERT_GLOBAL_ENC_TABLE_LOCKED added ensure this won't regress)

> Was there a ticket associated with this PR?

As Ractors are experimental and there are many issues we aren't necessarily filing tickets as the fixes will not be backported.

zzak:
Cool just checking, I saw this and thought neat but there was no bug associated or other context, so wasn't sure if it was ready for review or not. :pray:

luke-gruber:
@zzak  Sorry yeah, I added a test that is fairly reproducible.

luke-gruber:
@jhawthorn This is ready for a review now 😅

encoding.c
```diff
@@ -1564,15 +1619,21 @@ enc_set_default_encoding(struct default_encoding *def, VALUE encoding, const cha
...
-    if (default_external.enc) return default_external.enc;
-
-    if (default_external.index >= 0) {
-        default_external.enc = rb_enc_from_index(default_external.index);
-        return default_external.enc;
-    }
-    else {
-        return rb_locale_encoding();
+    rb_encoding *enc = NULL;
+    // TODO: make lock-free
```
jhawthorn: What would it take to do this? I think we call these really a lot, and `rb_enc_from_index` only needs to lock in cases where the encoding is not one of the three fastpaths.

If it helps, I think it would be fine if we took a VM barrier to change internal or external encoding.

(ditto for internal encoding)

luke-gruber: Yeah, it's called quite a bit. I don't think it should be hard, I'll try to do it in this PR. I'll add a new commit so you can review it easily.

luke-gruber: @jhawthorn It's ready for review again. It turns out more than just the `tbl->names` stuff needs to be locked, because autoloading of encodings can happen inside ractors, which can call certain functions. However, the most-used encoding functions can still be without locks.

jhawthorn (approved):

PR #13601, opened 2025-06-12
Get transcoding to work across ractors by locking certain operations

luke-gruber:
Ex: `str.encode` and `str.encode!` work across ractors now.

The global table `transcoder_table` needs a lock around each st_lookup/st_insert, and it's a two-level
table so the second level also needs to be locked around insertion/deletion. In addition to this, the
transcoder entries (values in the second-level hash table) need to be locked around retrieving them and
loading them as they are loaded lazily.

The transcoding objects (`Encoding::Converter`) can't be made shareable, so their operations don't need to be locked.


luke-gruber:
@jhawthorn Ready for review 😅

PR #13602, opened 2025-06-12
ZJIT: Add codegen for SideExit

tekknolagi:


tekknolagi:
Stacked on top of https://github.com/ruby/ruby/pull/13579

PR #13606, opened 2025-06-13
Add SHAPE_ID_HAS_IVAR_MASK for quick ivar check

casperisfine:
This allow checking if an object has ivars with just a shape_id
mask.


string.c
```diff
@@ -2311,12 +2306,22 @@ rb_str_dup(VALUE str)
...
+VALUE
+rb_yjit_str_dup(VALUE str)
+{
+    if (LIKELY(BARE_STRING_P(str))) {
+        return str_duplicate(rb_cString, str);
+    }
+    return rb_obj_dup(str);
+}
+
+
```
k0kubun:
```suggestion

```

✂️

internal/string.h
```diff
@@ -30,6 +30,7 @@ enum ruby_rstring_private_flags {
 #endif
 
 /* string.c */
+VALUE rb_yjit_str_dup(VALUE str);
```
k0kubun:
```suggestion
```

I think YJIT can use the function without this declaration. bindgen alone should be enough.

k0kubun: Oh but bindgen needs to read this declaration 😅 nvm

casperisfine: TIL

k0kubun: Sorry that I confused you, but you need to do either of these:

1. Declare `rb_str_dup_m` here and let bindgen use it
2. Or just declare that yourself in Rust code

Please rerun `make yjit-bindgen` if you take the (1) route.

PR #13609, opened 2025-06-13
Fix a race condition in object_id for shareable objects

casperisfine:
If an object is shareable and has no capacity left, it isn't safe to store the object ID in fields as it requires an object resize which can't be done unless all field reads are synchronized.

In this very specific case we create the object_id in advance, before the object is made shareable.

Simpler alternative to: https://github.com/ruby/ruby/pull/13305

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -1357,9 +1357,26 @@ make_shareable_check_shareable(VALUE obj)
         }
     }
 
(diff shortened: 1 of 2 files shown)


tenderlove (approved):
This seems like a good and pragmatic solution to me.  Maybe we can do something different in the future, but I like this 👍

PR #13613, opened 2025-06-13
ZJIT: Partially enable btest on CI

k0kubun:
This PR adds a CI job to run bootstraptests that already run correctly with `--zjit-call-threshold=1`. We should enable more tests as we fix them. The setup is copied from what we used for YJIT before https://github.com/Shopify/ruby/pull/334.

Unlike YJIT, `--zjit-call-threshold=1` would miss profiling-based optimizations, but we should make `--zjit-call-threshold=1` work anyway. Since btests are not designed for `--zjit-call-threshold=2`, I didn't add such a job.

### DIFF
--- a/.github/workflows/zjit-macos.yml
+++ b/.github/workflows/zjit-macos.yml
@@ -42,6 +42,9 @@ jobs:
             configure: '--enable-zjit=dev'
             tests: '../src/test/ruby/test_zjit.rb'
 
+          - test_task: 'btest'
+            configure: '--enable-zjit=dev'
+
     env:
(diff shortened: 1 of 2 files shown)


tekknolagi (approved):
Whoa, amazing. Do we have an idea of the Big Things causing the other tests to fail?

k0kubun:
Majority of crashes comes from `not implemented: Invalidation on EP escape is not implemented yet`. The rest seems to come mostly from `not implemented: register spilling not yet implemented, too many basic block arguments`.

I'm thinking of working on them in that order.

PR #13625, opened 2025-06-16
ZJIT: Add codegen for StringCopy

composerinteralia:
Prior to this commit we compiled `putstring` and `putchilledstring` to `StringCopy`, but then failed to compile past HIR.

This commit adds codegen for `StringCopy` to call `rb_ec_str_ressurrect` as the VM does for these instructions.


k0kubun:
We merged another PR and it caused a conflict. Could you please fix it?

tekknolagi:
Thank you!

PR #13626, opened 2025-06-16
Refactor generic fields to use `T_IMEMO/fields` objects.

casperisfine:
Followup: https://github.com/ruby/ruby/pull/13589
    
This simplify a lot of things, as we no longer need to manually manage the memory, we can use the Read-Copy-Update pattern and avoid numerous race conditions.

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -4774,13 +4774,14 @@ rb_class_ivar_set(VALUE obj, ID id, VALUE val)
 
     if (new_fields_obj != original_fields_obj) {
         RCLASS_WRITABLE_SET_FIELDS_OBJ(obj, new_fields_obj);
-
-        // TODO: What should we set as the T_CLASS shape_id?
-        // In most case we can replicate the single `fields_obj` shape
-        // but in namespaced case?
-        // Perhaps INVALID_SHAPE_ID?
-        RBASIC_SET_SHAPE_ID(obj, RBASIC_SHAPE_ID(new_fields_obj));
     }
+
+    // TODO: What should we set as the T_CLASS shape_id?
+    // In most case we can replicate the single `fields_obj` shape
+    // but in namespaced case?
+    // Perhaps INVALID_SHAPE_ID?
+    RBASIC_SET_SHAPE_ID(obj, RBASIC_SHAPE_ID(new_fields_obj));
+
     return existing;
 }
 
--- a/imemo.c
+++ b/imemo.c
@@ -155,6 +155,13 @@ imemo_fields_trigger_wb_i(st_data_t key, st_data_t value, st_data_t arg)
     return ST_CONTINUE;
 }
 
+static int
+imemo_fields_complex_wb_i(st_data_t key, st_data_t value, st_data_t arg)
+{
+    RB_OBJ_WRITTEN((VALUE)arg, Qundef, (VALUE)value);
+    return ST_CONTINUE;
+}
+
 VALUE
 rb_imemo_fields_new_complex_tbl(VALUE klass, st_table *tbl)
 {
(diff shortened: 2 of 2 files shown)


variable.c
```diff
@@ -1815,119 +1782,174 @@ general_field_set(VALUE obj, shape_id_t target_shape_id, VALUE val, void *data,
...
+            RUBY_ASSERT(next_capacity > current_capacity);
+
+            fields_obj = rb_imemo_fields_new(rb_obj_class(obj), next_capacity);
+            if (original_fields_obj) {
+                attr_index_t fields_count = RSHAPE_LEN(current_shape_id);
+                VALUE *fields = rb_imemo_fields_ptr(fields_obj);
+                MEMCPY(fields, rb_imemo_fields_ptr(original_fields_obj), VALUE, fields_count);
+                for (attr_index_t i = 0; i < fields_count; i++) {
+                    RB_OBJ_WRITTEN(fields_obj, Qundef, fields[i]);
+                }
```
tenderlove, line 1850: I think we could do `rb_gc_writebarrier_remember(fields_obj)` here so we don't have to iterate the list.  Someone will eventually have to iterate this list, but at least we only need to put 1 item in the remembered set instead of N items.

variable.c
```diff
@@ -1815,119 +1782,174 @@ general_field_set(VALUE obj, shape_id_t target_shape_id, VALUE val, void *data,
...
-generic_ivar_set_shape_resize_fields(VALUE obj, attr_index_t _old_capa, attr_index_t new_capa, void *data)
-{
-    struct gen_fields_lookup_ensure_size *fields_lookup = data;
+    if (!existing) {
+        RBASIC_SET_SHAPE_ID(fields_obj, next_shape_id);
+    }
 
-    fields_lookup->resize = true;
-}
+    generic_update_fields_obj(obj, fields_obj, original_fields_obj);
```
tenderlove, line 1865: Maybe add a comment here that `generic_update_fields_obj` fires a WB? I was looking for missing WBs. This function clearly fires them from the fields object to the values, but I couldn't see where the obj -> fields_obj barrier is fired.

tenderlove (commented):
What is the difference between `generic_field_set` and `generic_ivar_set`? It seems like "fields" is just used for object ids?

I see some logic WRT shapes in `generic_update_fields_obj`, but I don't know if that's legit for both code paths.  Other than that, I can't see from this diff what the problem is.

casperisfine:
> What is the difference between `generic_field_set` and `generic_ivar_set`? It seems like "fields" is just used for object ids?

Yes, currently it's only used for `object_id`, eventually it would be used for "old_address" too https://github.com/ruby/ruby/pull/13290

Ideally I'd like `ivar_set` to just call into `field_set`, but I need to put more thoughts into it.

variable.c
casperisfine: I'm unfamiliar with `rb_gc_writebarrier_remember` but I'll read on it.

variable.c
```diff
@@ -1809,135 +1782,174 @@ general_field_set(VALUE obj, shape_id_t target_shape_id, VALUE val, void *data,
...
-    shape_id_t shape_id;
-    bool resize;
-};
+static inline void
+generic_update_fields_obj(VALUE obj, VALUE fields_obj, const VALUE original_fields_obj)
+{
+    if (fields_obj != original_fields_obj) {
+        if (original_fields_obj) {
+            // Clear root shape to avoid triggering cleanup such as free_object_id.
+            RBASIC_SET_SHAPE_ID(original_fields_obj, ROOT_SHAPE_ID);
```
casperisfine: Note to self: We actually can't do that, because if the shape was `TOO_COMPLEX` we lose that information and `imemo_fields_free` won't call `st_free_table`.

Possible solutions:
  - Only clear `SHAPE_ID_FL_HAS_OBJECT_ID`.
  - Store the owner object in `basic.klass`, and clear that to signify the fields have been replaced.

variable.c
casperisfine: Hum, I tried to use it as a replacement for firing write barriers on all fields, but instantly ran into GC crashes. So I suspect I don't understand how it's meant to be used.

PR #13627, opened 2025-06-16
ZJIT: objtostring to HIR

ywenc:
This PR adds the `objtostring` opcode for the HIR. Also adds a fast path for known strings at compile time. 

`objtostring` is an instruction that appears when doing string interpolation. It appears with `concatstrings` and `anytostring`. `concatstrings` does not do any string coercion and so relies on `objtostring` and `anytostring` to ensure the contents on the stack are strings. 

```
ywenc in ~/Documents/github/ruby  (zjit-objtostring) % ./miniruby --dump=i -e '"#{('foo')}"'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,10)>
0000 putobject                              ""                        (   1)[Li]
0002 putself
0003 opt_send_without_block                 <calldata!mid:foo, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0005 dup
0006 objtostring                            <calldata!mid:to_s, argc:0, FCALL|ARGS_SIMPLE>
0008 anytostring
0009 concatstrings                          2
0011 leave
```

For the interpreter, `objtostring` has fast paths for some classes like strings and symbols, and if the fast paths don't return anything, we call the method id `to_s`.  In the fast path for strings, [the original string is returned](https://github.com/ruby/ruby/blob/95201299fd7bf0918dfbd8c127ce2b5b33ffa537/vm_insnhelper.c#L6068-L6071). `Insn::ObjToString` is similarly implemented - this PR currently contains a fast path for strings that returns the value from the stack if we know it is a string, otherwise the instruction uses `SendWithoutBlock` with `to_s`. We added `Insn::ObjToString` as its own instruction so that it can be optimized for it specifically, instead of optimizing `SendWithoutBlock` with `to_s`. `to_s` can be overridden on the String class, which is what is called when calling `to_s` directly. 

--

Instead of replacing `Insn::ObjToString` with `Insn::SendWithoutBlock` as the fallback in `optimize_direct_sends`,  we also considered making `Insn::ObjToString` alias to `gen_send_without_block` in `gen_insn`. But have decided to keep it there for now so changes are in fewer files, and so that `Insn::ObjToString` makes some more sense inside `optimizing_direct_sends`.

We also considered doing the fast path and SendWithoutBlock as a fallback in the initial HIR construction, instead of in the optimizer. But have left things in the optimizer because [`infer_types` happens at the end of `iseq_to_hir`](https://github.com/ruby/ruby/blob/master/zjit/src/hir.rs#L2510).

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -444,6 +444,8 @@ pub enum Insn {
     FixnumGt   { left: InsnId, right: InsnId },
     FixnumGe   { left: InsnId, right: InsnId },
 
+    ObjToString { val: InsnId, call_info: CallInfo, cd: *const rb_call_data, state: InsnId },
+
     /// Side-exit if val doesn't have the expected type.
     GuardType { val: InsnId, guard_type: Type, state: InsnId },
     /// Side-exit if val is not the expected VALUE.
@@ -636,6 +638,7 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
             Insn::ToNewArray { val, .. } => write!(f, "ToNewArray {val}"),
             Insn::ArrayExtend { left, right, .. } => write!(f, "ArrayExtend {left}, {right}"),
             Insn::ArrayPush { array, val, .. } => write!(f, "ArrayPush {array}, {val}"),
+            Insn::ObjToString { val, .. } => { write!(f, "ObjToString {val}") },
             Insn::SideExit { .. } => write!(f, "SideExit"),
             insn => { write!(f, "{insn:?}") }
         }
@@ -950,6 +953,12 @@ impl Function {
             FixnumGe { left, right } => FixnumGe { left: find!(*left), right: find!(*right) },
             FixnumLt { left, right } => FixnumLt { left: find!(*left), right: find!(*right) },
             FixnumLe { left, right } => FixnumLe { left: find!(*left), right: find!(*right) },
+            ObjToString { val, call_info, cd, state } => ObjToString {
+                val: find!(*val),
+                call_info: call_info.clone(),
+                cd: *cd,
+                state: *state,
+            },
             SendWithoutBlock { self_val, call_info, cd, args, state } => SendWithoutBlock {
                 self_val: find!(*self_val),
                 call_info: call_info.clone(),
@@ -1076,6 +1085,7 @@ impl Function {
             Insn::GetIvar { .. } => types::BasicObject,
             Insn::ToNewArray { .. } => types::ArrayExact,
             Insn::ToArray { .. } => types::ArrayExact,
+            Insn::ObjToString { .. } => types::BasicObject,
         }
     }
 
@@ -1282,6 +1292,14 @@ impl Function {
                         self.push_insn(block, Insn::PatchPoint(Invariant::StableConstantNames { idlist }));
                         let replacement = self.push_insn(block, Insn::Const { val: Const::Value(unsafe { (*ice).value }) });
                         self.make_equal_to(insn_id, replacement);
+                    }
+                    Insn::ObjToString { val, call_info, cd, state, .. } => {
+                        if self.is_a(val, types::StringExact) {
+                            self.make_equal_to(insn_id, val);
+                        } else {
+                            let replacement = self.push_insn(block, Insn::SendWithoutBlock { self_val: val, call_info, cd, args: vec![], state });
+                            self.make_equal_to(insn_id, replacement)
+                        }
                     }
                     _ => { self.push_insn_id(block, insn_id); }
                 }
@@ -1643,6 +1661,10 @@ impl Function {
                 }
                 Insn::ArrayPush { array, val, state } => {
                     worklist.push_back(array);
+                    worklist.push_back(val);
+                    worklist.push_back(state);
+                }
+                Insn::ObjToString { val, state, .. } => {
                     worklist.push_back(val);
                     worklist.push_back(state);
                 }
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -5297,4 +5354,34 @@ mod opt_tests {
...
+        eval("
+            def test = \"#{1}\"
+        ");
+        assert_optimized_method_hir("test", expect![[r#"
+            fn test:
+            bb0(v0:BasicObject):
+              v2:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+              v3:Fixnum[1] = Const Value(1)
+              v8:BasicObject = SendWithoutBlock v3, :to_s
+              SideExit
```
ywenc, line 5991: This side exits because `objtostring` works together with `anytostring` and `concatstrings`, which are currently unimplemented. The iseq side exiting is also why this PR does not have a test in `test/ruby/test_zjit.rb`

k0kubun:
Please resolve the git conflict :pray:

zjit/src/hir.rs
```diff
@@ -1127,6 +1136,7 @@ impl Function {
             Insn::GetIvar { .. } => types::BasicObject,
             Insn::ToNewArray { .. } => types::ArrayExact,
             Insn::ToArray { .. } => types::ArrayExact,
+            Insn::ObjToString { .. } => types::BasicObject,
```
tekknolagi, line 1156: Do we have any guarantee that the result, if it did not raise, will be `String` or maybe even `StringExact` (not a subclass)? That would be neat if so

zjit/src/hir.rs
```diff
@@ -5381,4 +5438,34 @@ mod opt_tests {
               Return v2
         "#]]);
     }
+
+    #[test]
+    fn test_objtostring_string() {
+        eval("
+            def test = \"#{('foo')}\"
+        ");
```
tekknolagi: Nit: you can use raw strings to avoid escaping normal double quote

zjit/src/hir.rs
```diff
@@ -5381,4 +5438,34 @@ mod opt_tests {
...
+              v3:StringExact[VALUE(0x1008)] = Const Value(VALUE(0x1008))
+              v4:StringExact = StringCopy v3
+              SideExit
+        "#]]);
+    }
+
+    #[test]
+    fn test_objtostring_fallback() {
+        eval("
+            def test = \"#{1}\"
```
tekknolagi: As above

zjit/src/hir.rs
```diff
@@ -5381,4 +5438,34 @@ mod opt_tests {
...
+            bb0(v0:BasicObject):
+              v2:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+              v3:StringExact[VALUE(0x1008)] = Const Value(VALUE(0x1008))
+              v4:StringExact = StringCopy v3
+              SideExit
+        "#]]);
+    }
+
+    #[test]
+    fn test_objtostring_fallback() {
```
```suggestion
    fn test_objtostring_with_non_string() {
```

zjit/src/hir.rs
```diff
@@ -1334,6 +1344,14 @@ impl Function {
...
                         self.make_equal_to(insn_id, replacement);
                     }
+                    Insn::ObjToString { val, call_info, cd, state, .. } => {
+                        if self.is_a(val, types::StringExact) {
+                            self.make_equal_to(insn_id, val);
+                        } else {
+                            let replacement = self.push_insn(block, Insn::SendWithoutBlock { self_val: val, call_info, cd, args: vec![], state });
+                            self.make_equal_to(insn_id, replacement)
+                        }
+                    }
```
tekknolagi: Maybe better for a follow-up PR since this PR looks pretty done, but perhaps we should handle this instead by generating a send and then having a rewrite pass that rewrites all `a.to_s` when `a` is a `StringExact` into `a`. That is a little more generic

zjit/src/hir.rs
composerinteralia: I don't think so after `objtostring`, because somebody could write a `to_s` method that doesn't return a string. But `objtostring` is (always?) followed by `anytostring`, at which point we should know it's a `StringExact`.

zjit/src/hir.rs
composerinteralia: I believe @ywenc started with that approach, but it didn't work so well because in the general case of that optimization we'd want a method redefined patch point, but for `objtostring` specifically we don't want the patch point.

zjit/src/hir.rs
ywenc: Thanks!

zjit/src/hir.rs
tekknolagi: Can you leave a comment about not having a patchpoint in this rewrite? Seems like a potential future point of confusion

zjit/src/hir.rs
```diff
@@ -493,6 +493,8 @@ pub enum Insn {
     FixnumGt   { left: InsnId, right: InsnId },
     FixnumGe   { left: InsnId, right: InsnId },
 
+    ObjToString { val: InsnId, call_info: CallInfo, cd: *const rb_call_data, state: InsnId },
```
tekknolagi, line 500: This has slightly subtle/non-trivial semantics; please leave a one-liner comment explaining

tekknolagi (approved):

zjit/src/hir.rs
```diff
@@ -1386,6 +1397,15 @@ impl Function {
                         let replacement = self.push_insn(block, Insn::Const { val: Const::Value(unsafe { (*ice).value }) });
                         self.make_equal_to(insn_id, replacement);
                     }
+                    Insn::ObjToString { val, call_info, cd, state, .. } => {
+                        if self.is_a(val, types::StringExact) {
```
XrXr, line 1401: I think this is overly strict.

```suggestion
                        if self.is_a(val, types::String) {
```

Since the interpreter checks for T_STRING, it's noop on string subclasses as well.

cc @tekknolagi the `is_a` terminology is confusing to me, because `subclass_instance.is_a(String)`, but I think `self.is_a(val, types::StringExact)` is true only when `val.class == ::String`. I guess this just falls out of `StringExact` not having a Ruby level equivalent, and the hierarchy is just different.

I'm not going to hold up this PR over this, though and we can fix this separately.

XrXr (approved):

zjit/src/hir.rs
ywenc: I see `is_a` [here](https://github.com/ruby/ruby/blob/34eaa6418e5c5b8639add323dbfd531b32a7d4a3/zjit/src/hir.rs#L1081-L1083) where `self.is_a(val, types::StringExact)` if `self.type_of(val).is_subtype(type)`, would that cover string subcalsses? But I'm also down to go with `self.is_a(val, types::String)` here to be more clear about the intention.

XrXr: There is `types::StringSubclass` and `types::StringSubclass.is_subclass(types::StringExact) == false`. So no it's not covered.

ywenc: thank you!

PR #13628, opened 2025-06-16
ZJIT: Add codegen (and FrameState) for GetConstPath

XrXr:
Issue a call to rb_vm_opt_getconstant_path() like the
interpreter, but since that allocates the IC, we need to save the PC
before calling. To have the PC, GetConstPath now has a FrameState.

Fixes: https://github.com/Shopify/ruby/issues/586

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -650,6 +650,27 @@ class TestZJIT < Test::Unit::TestCase
     }
   end
 
+  def test_uncached_getconstant_path
+    assert_compiles RUBY_COPYRIGHT.dump, %q{
+      def test = RUBY_COPYRIGHT
+      test
+    }, call_threshold: 1, insns: [:opt_getconstant_path]
+  end
+
+  def test_getconstant_path_autoload
+    # A constant-referencing expression can run arbitrary code through Kernel#autoload.
+    Dir.mktmpdir('autoload') do |tmpdir|
+      autoload_path = File.join(tmpdir, 'test_getconstant_path_autoload.rb')
+      File.write(autoload_path, 'X = RUBY_COPYRIGHT')
+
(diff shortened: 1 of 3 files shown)


zjit/src/codegen.rs
```diff
@@ -293,6 +294,21 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
     Some(())
 }
 
+fn gen_get_constant_path(asm: &mut Assembler, ic: *const iseq_inline_constant_cache, state: &FrameState) -> Opnd {
+    unsafe extern "C" {
+        fn rb_vm_opt_getconstant_path(ec: EcPtr, cfp: CfpPtr, ic: *const iseq_inline_constant_cache) -> VALUE;
+    }
+
+    // Save PC since the call can allocate an IC
+    gen_save_pc(asm, state);
```
k0kubun, line 305: :+1:

k0kubun (approved):

test/ruby/test_zjit.rb
```diff
@@ -636,6 +636,26 @@ def test() = @foo = 1
...
+  def test_getconstant_path_autoload
+    Dir.mktmpdir('autoload') do |tmpdir|
+      autoload_path = File.join(tmpdir, "test_getconstant_path_autoload.rb")
+      File.write(autoload_path, 'X = RUBY_COPYRIGHT')
+
+      assert_compiles RUBY_COPYRIGHT.dump, %Q{
+        Object.autoload(:X, #{File.realpath(autoload_path).inspect})
+        def test = X
+        test
+      }, call_threshold: 1
```
tekknolagi: `insns: [:opt_getconstant_path]`?

tekknolagi (approved):

PR #13632, opened 2025-06-17
ZJIT: Support invokebuiltin opcodes

composerinteralia:
* `invokebuiltin`
* `invokebuiltin_delegate`
* `invokebuiltin_delegate_leave`

These instructions all call out to a C function, passing EC, self, and some number of arguments. `invokebuiltin` gets the arguments from the stack, whereas the `_delegate` instructions use a subset of the locals.

`opt_invokebuiltin_delegate_leave` has a fast path for `leave`, but I'm not sure we need to do anything special for that here (FWIW YJIT appears to treat the two delegate instructions the same).

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -76,6 +76,21 @@ class TestZJIT < Test::Unit::TestCase
     }
   end
 
+  def test_invokebuiltin
+    assert_compiles '["."]', %q{
+      def test = Dir.glob(".")
+      test
+    }
+  end
+
+  def test_invokebuiltin_delegate
+    assert_compiles '[[], true]', %q{
+      def test = [].clone(freeze: true)
+      r = test
+      [r, r.frozen?]
+    }
+  end
+
   def test_opt_plus_const
     assert_compiles '3', %q{
       def test = 1 + 2
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -260,6 +260,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::IfFalse { val, target } => return gen_if_false(jit, asm, opnd!(val), target),
         Insn::SendWithoutBlock { call_info, cd, state, self_val, args, .. } => gen_send_without_block(jit, asm, call_info, *cd, &function.frame_state(*state), self_val, args)?,
         Insn::SendWithoutBlockDirect { cme, iseq, self_val, args, state, .. } => gen_send_without_block_direct(cb, jit, asm, *cme, *iseq, opnd!(self_val), args, &function.frame_state(*state))?,
+        Insn::InvokeBuiltin { bf, args, state } => gen_invokebuiltin(jit, asm, &function.frame_state(*state), bf, args)?,
         Insn::Return { val } => return Some(gen_return(asm, opnd!(val))?),
         Insn::FixnumAdd { left, right, state } => gen_fixnum_add(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state))?,
         Insn::FixnumSub { left, right, state } => gen_fixnum_sub(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state))?,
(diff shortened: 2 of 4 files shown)


test/ruby/test_zjit.rb
```diff
@@ -76,6 +76,14 @@ def test3 = baz(4, 1)
     }
   end
 
+  def test_invokebuiltin_delegate
```
composerinteralia, line 86: I haven't really figured out how to test this, so let me know if you have ideas. This test kinda works, but it's indirect—there's no `invokebuiltin_delegate` directly in the code here, but `Kernel#clone` gets compiled along with `test` and that does have the instruction.

I couldn't get a good test for `invokebuiltin`. All of the common ones I could find (`Hash.new`, `Dir.glob`) use default args which we don't support yet. I was testing locally with `GC.start`, but there's not much to assert on for that one.

zjit/src/hir.rs
```diff
@@ -632,6 +635,13 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             }
+            Insn::InvokeBuiltin { bf, args, .. } => {
```
composerinteralia, line 642: An example of what this looks like for `GC.start`

```
fn start:
bb0(v0:BasicObject, v1:BasicObject, v2:BasicObject, v3:BasicObject, v4:BasicObject):
  v6:FalseClassExact = Const Value(false)
  v8:BasicObject = InvokeBuiltin gc_start_internal, v0, v1, v2, v3, v6
  Return v8
```

test/ruby/test_zjit.rb
XrXr: I suggest testing the no args code path of Integer#times for now:

```
== disasm: #<ISeq:times@<internal:numeric>:250 (250,2)-(261,5)> INLINE_BLOCK
[ 1] i@0
0000 putnil                                                           ( 252)[LiCa]
0001 defined                                yield, false, true
0005 branchif                               11
0007 opt_invokebuiltin_delegate             <builtin!_bi253/0>, 0     ( 253)[Li]
0010 leave                                  [Re]
0011 putobject_INT2FIX_0_                                             ( 255)[Li]
0012 setlocal_WC_0                          i@0
<snip>
```

You're running into a limitation of assert_compiles() here in that there is no way to test that an indirectly compiled method contains a certain YARV instruction. Probably still fine for now since the builtin instructions are the only ones we can't generate in normal user mode ruby.

XrXr: Related, you should be able to test the HIR directly for kernel methods in hir.rs, since it lets you target the method to assert HIR for more precisely:

```
    /// Get the ISeq of a specified method
    pub fn get_method_iseq(name: &str) -> *const rb_iseq_t {
        let wrapped_iseq = eval(&format!("RubyVM::InstructionSequence.of(method(:{}))", name));
        unsafe { rb_iseqw_to_iseq(wrapped_iseq) }
    }
```

Just call assert_method_hir_with_opcode() with say, `"loop"`, which should find Kernel#loop.

composerinteralia: Nice. Thank you for that. I'll add some tests there.

composerinteralia: `Integer#times` has a `defined` that we can't compile yet, but I found some others. I couldn't find anything in Kernel we could compile for `invokebuiltin`, so I went with `GC.start` for that and changed `get_method_iseq` to accept a receiver.

zjit/src/codegen.rs
```diff
@@ -311,6 +312,25 @@ fn gen_get_constant_path(asm: &mut Assembler, ic: *const iseq_inline_constant_ca
     val
 }
 
+fn gen_invokebuiltin(jit: &mut JITState, asm: &mut Assembler, state: &FrameState, bf: &rb_builtin_function, args: &Vec<InsnId>) -> Option<lir::Opnd> {
+    // ec, self, and arguments
+    if bf.argc + 2 > (C_ARG_OPNDS.len() as i32) {
+        return None;
```
tekknolagi, line 319: Can you add a TODO here to support this when we can do stack args? Or is this a fundamental limitation somehow?

zjit/src/hir.rs
```diff
@@ -1027,6 +1037,11 @@ impl Function {
                 args: args.iter().map(|arg| find!(*arg)).collect(),
                 state: *state,
             },
+            InvokeBuiltin { bf, args, state } => InvokeBuiltin {
+                bf: *bf,
+                args: args.iter().map(|arg| find!(*arg)).collect(),
```
tekknolagi: Nit: use `find_vec!`

test/ruby/test_zjit.rb
tekknolagi: Please add specific TODOs to add tests here when we support X specific feature

zjit/src/codegen.rs
composerinteralia: Maybe we don't need this at all, since it's already covered by asserts in the assembler? (e.g. https://github.com/ruby/ruby/blob/aaa956e8f19aa14ab9482474cc51eeb103623e7d/zjit/src/backend/x86_64/mod.rs#L343)

zjit/src/hir.rs
composerinteralia: Neat! Thanks

zjit/src/codegen.rs
composerinteralia: Although I guess that'll panic instead of failing to compile. I'll leave the TODO instead for now.

test/ruby/test_zjit.rb
composerinteralia: I added one more test that I think covers what we need (`Dir.glob` currently only compiles because it side exits, so the new test will become more meaningful once we can compile `checkkeyword` and it no longer side exits.)

k0kubun (approved):

PR #13633, opened 2025-06-17
ZJIT: SideExit on optional args

composerinteralia:
We haven't implemented optional args yet, so `SideExit` for now. Prior to this commit `def test(a=true) a` would compile to:

```
fn test:
  v3:TrueClassExact = Const Value(true)
  Return v3
```

But that is only correct for the case where the caller does not pass in the optional arg. We probably need a guard on the PC similar to what YJIT has.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2127,6 +2127,13 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
             let exit_state = state.clone();
             profiles.profile_stack(&exit_state);
 
+            // TODO: Remove this and implement optional arguments
+            if unsafe { get_iseq_flags_has_opt(iseq) } {
+                let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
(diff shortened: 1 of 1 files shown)


XrXr (commented):
Can you amend the test to cal the method with opt param in the eval() program? I wonder what happened before your change.

composerinteralia:
Sure, I can do that. Do you mean something like:

```
        eval("
            def test(a=nil) = a
            test
            test(true)
        ");
```

> I wonder what happened before your change.

It compiled, but it'd always return the default value regardless of what was passed in.

XrXr:
> It compiled, but it'd always return the default value regardless of what was passed in.

Thanks. In that case, never mind about changing what's inside eval(). I think we should instead add an execution test to test_zjit.rb.

composerinteralia:
Writing that test made me realize that what I have here doesn't actually work correctly (I think it's saving the wrong PC while exiting). It's probably easier here to bail out entirely with a `ParseError` (which might might make it difficult to write a test in `test_zjit`). Or maybe I can try to figure out how to actually implement this instead. I'll close this PR for now since it's not right and open something else later.

PR #13636, opened 2025-06-17
Fix minor typos in comments and docs

tas50:
Just a bit of minor cleanup

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -40,7 +40,7 @@ Usage: benchmark-driver [options] RUBY|YAML...
         --filter REGEXP              Filter out benchmarks with given regexp
         --run-duration SECONDS       Warmup estimates loop_count to run for this duration (default: 3)
         --timeout SECONDS            Timeout ruby command execution with timeout(1)
-    -v, --verbose                    Verbose mode. Multiple -v options increase visilibity (max: 2)
+    -v, --verbose                    Verbose mode. Multiple -v options increase visibility (max: 2)
 ```
 
 ## make benchmark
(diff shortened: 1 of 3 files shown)


spec/ruby/library/stringio/ungetbyte_spec.rb
```diff
@@ -9,7 +9,7 @@
     c = io.getc
     c.should == 'T'
     io.ungetbyte(83)
-    io.string.should == 'Shis is a simple string.'
+    io.string.should == 'This is a simple string.'
```
nobu: This is not a typo, `83.chr` is S.

spec/ruby/library/stringio/ungetbyte_spec.rb
```diff
@@ -27,7 +27,7 @@
     c = io.getc
     c.should == 'T'
     io.ungetbyte(83 | 0xff00)
-    io.string.should == 'Shis is a simple string.'
+    io.string.should == 'This is a simple string.'
```
nobu: Ditto.

nobu (changes requested):

spec/ruby/library/stringio/ungetbyte_spec.rb
nobu: May this avoid spell checkers?

```suggestion
    io.string.should == 'S' 'his is a simple string.'
```

spec/ruby/library/stringio/ungetbyte_spec.rb
nobu:
```suggestion
    io.string.should == 'S' 'his is a simple string.'
```

spec/ruby/library/stringio/ungetbyte_spec.rb
tas50: I just reverted it. Simple fix. I should have read the full spec

PR #13647, opened 2025-06-18
Use released version of debug-1.11.0

hsbt:


hsbt:
This is duplicated with https://github.com/ruby/ruby/pull/13238

PR #13648, opened 2025-06-18
Update bundled gems

hsbt:
We couldn't specify development version of RDoc. Because rdoc need to generate `markdown.rb` from that repository. So, development branch specified commit hash didn't contain that file.

Fixed https://rubyci.s3.amazonaws.com/osx1400arm-no-yjit/ruby-master/log/20250618T045007Z.fail.html.gz


hsbt:
I merged this manually.

PR #13649, opened 2025-06-18
Make `tsort` to bundled gems

hsbt:
https://bugs.ruby-lang.org/issues/21442


hsbt:
I confirmed to works with removing tsort from ruby/ruby. This pull-request should be closed because we need to migrate `tsort` after releasing Ruby 3.5.

hsbt:
I added that warning at https://github.com/ruby/ruby/pull/13841

PR #13650, opened 2025-06-18
Shink RClass when it is known they can't be namespaced

byroot:
Even when namespaces are enabled, only a few core classes created during init will eventually be namespaced.

For these it's OK to allocate a 320B slot to hold the extra namespace stuff.
    
But for any class created post init, we know we'll never need the namespace and we can fit in a 160B slot.

### DIFF
--- a/internal/class.h
+++ b/internal/class.h
@@ -293,6 +293,7 @@ static inline void RCLASS_WRITE_CLASSPATH(VALUE klass, VALUE classpath, bool per
 #define RCLASS_PRIME_CLASSEXT_WRITABLE FL_USER2
 #define RCLASS_IS_INITIALIZED FL_USER3
 // 3 is RMODULE_IS_REFINEMENT for RMODULE
+#define RCLASS_NAMESPACEABLE FL_USER4
 
 /* class.c */
 rb_classext_t * rb_class_duplicate_classext(rb_classext_t *orig, VALUE obj, const rb_namespace_t *ns);
--- a/internal/class.h
+++ b/internal/class.h
@@ -171,8 +171,6 @@ static inline rb_classext_t * RCLASS_EXT_WRITABLE_IN_NS(VALUE obj, const rb_name
 static inline rb_classext_t * RCLASS_EXT_WRITABLE(VALUE obj);
 
 // Raw accessor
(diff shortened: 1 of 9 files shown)


internal/class.h
```diff
@@ -146,14 +145,18 @@ struct RClass {
 
 // Assert that classes can be embedded in heaps[2] (which has 160B slot size)
 // On 32bit platforms there is no variable width allocation so it doesn't matter.
-// TODO: restore this assertion after shrinking rb_classext_t
-// STATIC_ASSERT(sizeof_rb_classext_t, sizeof(struct RClass) + sizeof(rb_classext_t) <= 4 * RVALUE_SIZE || SIZEOF_VALUE < SIZEOF_LONG_LONG);
+STATIC_ASSERT(sizeof_rb_classext_t, sizeof(struct RClass) + sizeof(rb_classext_t) <= 4 * RVALUE_SIZE || SIZEOF_VALUE < SIZEOF_LONG_LONG);
```
nobu: This is safer than adding `sizeof`s.
Of course, after `struct RClass_and_rb_classext_t` declaration.

```suggestion
#if SIZEOF_VALUE >= SIZEOF_LONG_LONG
STATIC_ASSERT(sizeof_RClass_and_rb_classext_t, sizeof(struct RClass_and_rb_classext_t) <= 4 * RVALUE_SIZE);
#endif
```

byroot: Done. Thank you @nobu!

jhawthorn:
> But for any class created post init, we know we'll never need the namespace and we can fit in a 160B slot.

I wonder if this will be true in the future because of singleton classes of pre-init namespaceable classes (they either need to also be made namespaceable, or we'd need to use namespacing to look up klass 😩) https://bugs.ruby-lang.org/issues/21343

byroot:
Good catch.

> they either need to also be made namespaceable,

I think that is the way, but I can probably handle that. When creating a singleton class I can check if the attached object class is namespaceable.

byroot:
I made it so that metaclasses created from a namespaceable class are also namespaceable. (I didn't fix https://bugs.ruby-lang.org/issues/21343 though.)

PR #13655, opened 2025-06-19
ZJIT: Add pass to clean CFG

tekknolagi:
We can fuse linked lists of blocks. This can be run any time, improves
future analyses, improves codegen, and also makes the HIR output look
nicer.

Inspired by my implementation of CleanCFG for Cinder, which was itself
inspired by Brett Simmers' implementation in HHVM.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1792,6 +1792,67 @@ impl Function {
         }
     }
 
+    fn absorb_dst_block(&mut self, num_in_edges: &Vec<u32>, block: BlockId) -> bool {
+        let Some(terminator_id) = self.blocks[block.0].insns.last()
+            else { return false };
+        let Insn::Jump(BranchEdge { target, args }) = self.find(*terminator_id)
+            else { return false };
+        if target == block {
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -1768,6 +1768,67 @@ impl Function {
...
+        // Link up params with block args
+        let params = std::mem::take(&mut self.blocks[target.0].params);
+        assert_eq!(args.len(), params.len());
+        for (arg, param) in args.iter().zip(params) {
+            self.make_equal_to(param, *arg);
+        }
+        // Remove branch instruction
+        self.blocks[block.0].insns.pop();
+        // Copy target instructions into block
+        let target_insns = std::mem::take(&mut self.blocks[target.0].insns);
```
XrXr: nit: since you're using `mem::take`
```suggestion
        self.blocks[block.0].insns.pop();
        // Move target instructions into block
        let target_insns = std::mem::take(&mut self.blocks[target.0].insns);
```

XrXr (approved):

PR #13658, opened 2025-06-19
ZJIT: `anytostring`

ywenc:
This PR adds support for `anytostring` with an optimization for known strings. 

`anytostring` is used with `objtostring` and `concatstrings` for string interpolation (see PR description in https://github.com/ruby/ruby/pull/13627 for more details). 

The implementation for `anytostring` pops two values from the stack. If the first value popped from the stack [is a `T_STRING`](https://github.com/ruby/ruby/blob/83fb07fb2c97b9922450979fa4a56f43324317a9/string.c#L2177), push it back on the stack, otherwise push the string coercion of the second.

This PR also piggybacks a fix for [string subclasses should also skip `:to_s` for `objtostring`](https://github.com/ruby/ruby/pull/13627#discussion_r2157269151) (cc @XrXr ).

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -283,6 +283,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::SetIvar { self_val, id, val, state: _ } => return gen_setivar(asm, opnd!(self_val), *id, opnd!(val)),
         Insn::SideExit { state } => return gen_side_exit(jit, asm, &function.frame_state(*state)),
         Insn::PutSpecialObject { value_type } => gen_putspecialobject(asm, *value_type),
+        Insn::AnyToString { val, str, state } => gen_anytostring(asm, opnd!(val), opnd!(str), &function.frame_state(*state))?,
         _ => {
             debug!("ZJIT: gen_function: unexpected insn {:?}", insn);
             return None;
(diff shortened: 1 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -1154,6 +1161,7 @@ impl Function {
             Insn::ToNewArray { .. } => types::ArrayExact,
             Insn::ToArray { .. } => types::ArrayExact,
             Insn::ObjToString { .. } => types::BasicObject,
+            Insn::AnyToString { .. } => types::StringExact,
```
tekknolagi: This is not guaranteed to be `StringExact`, unfortunately; if `str` is a subclass of `String`, `str` gets returned unmodified. We can add a conditional here based on the inferred type of `str` or figure out another way to split the IR, but otherwise the type must be `types::String`

zjit/src/codegen.rs
```diff
@@ -817,6 +818,18 @@ fn gen_fixnum_ge(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd) -> Opti
     Some(asm.csel_ge(Qtrue.into(), Qfalse.into()))
 }
 
+fn gen_anytostring(asm: &mut Assembler, val: lir::Opnd, str: lir::Opnd, state: &FrameState) -> Option<lir::Opnd> {
+    asm_comment!(asm, "call rb_obj_as_string_result");
```
tekknolagi: Please move the comment about calling a function down to the ccall so that it doesn't get confused with the PC-saving code

tekknolagi (approved):
lgtm pending comments

zjit/src/hir.rs
ywenc: you're right! good catch, thank you

k0kubun (approved):

PR #13664, opened 2025-06-20
Fix handling of PM_CONSTANT_PATH_NODE node in keyword arguments with ARGS_SPLAT

jeremyevans:
This was handled correctly in parse.y (NODE_COLON2), but not in prism. This wasn't caught earlier, because I only added tests for the optimized case and not the unoptimized case. Add tests for the unoptimized case.

In code terms:

```ruby
m(*a, kw: lvar::X)     # Does not require allocation for *a
m(*a, kw: method()::X) # Requires allocation for *a
```

This commit fixes the second case when prism is used.


kddnewton:
Thanks!

PR #13666, opened 2025-06-20
ZJIT: Optimize frozen array aref

tekknolagi:
If we have a frozen array `[..., a, ...]` and a compile-time fixnum index `i`, we can do the array load at compile-time.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -206,6 +206,7 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
                     BOP_FREEZE => write!(f, "BOP_FREEZE")?,
                     BOP_UMINUS => write!(f, "BOP_UMINUS")?,
                     BOP_MAX    => write!(f, "BOP_MAX")?,
+                    BOP_AREF   => write!(f, "BOP_AREF")?,
                     _ => write!(f, "{bop}")?,
                 }
                 write!(f, ")")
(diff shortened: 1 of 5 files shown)


zjit/src/hir.rs
```diff
@@ -1309,6 +1310,25 @@ impl Function {
         }
     }
 
+    fn try_rewrite_subscript(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId, idx_val: InsnId) {
```
```suggestion
    fn try_rewrite_aref(&mut self, block: BlockId, orig_insn_id: InsnId, self_val: InsnId, idx_val: InsnId) {
```

`#[]` is called `aref` very consistently in the CRuby codebase, so calling it differently could confuse contributors, at least including myself.

zjit/src/hir.rs
```diff
@@ -6038,4 +6060,49 @@ mod opt_tests {
...
+    fn test_eliminate_load_from_frozen_array_negative() {
+        eval(r##"
+            def test = [4,5,6].freeze[-3]
+        "##);
+        assert_optimized_method_hir("test", expect![[r#"
+            fn test:
+            bb0(v0:BasicObject):
+              PatchPoint BOPRedefined(ARRAY_REDEFINED_OP_FLAG, BOP_FREEZE)
+              PatchPoint BOPRedefined(ARRAY_REDEFINED_OP_FLAG, BOP_AREF)
+              v11:NilClassExact = Const Value(nil)
```
k0kubun, line 6104: Is it intentional that the result is nil? `-1` is 6, `-2` is 5, so `-3` should be 4.

```
$ ruby -e 'p [4,5,6].freeze[-3]'
4
```

You picked `rb_ary_elt` for your implementation, but as the comment above the function says, the behavior is different from `Array#[]`.

tekknolagi: Nice catch, thank you

PR #13671, opened 2025-06-20
call rb_bug() if we fail to create timer_thread

luke-gruber:
We should check this in general, but there are specific cases that worry me about timer thread creation.

In `rb_fork_ruby()`, if the call to `fork` fails, we handle the error (sleep or GC depending on errno), and try again. If fork fails because of resource issues, we can't assume the call in `after_fork_ruby()` to create the timer thread will succeed. If it fails, we get undefined behavior in the next iteration of the loop when we call pthread_join() on it in `before_fork_ruby()`.


ko1:
it is better to run without timer thread even if it has an issue.
How about to warn about that?

luke-gruber:
Okay, that makes sense. I'll make sure not call `pthread_join` on it in next iteration if fork fails and timer thread fails. I'll also warn the user.

Edit: I updated the PR. The failed CI run is due to a flake test.

PR #13673, opened 2025-06-20
Avoid allocation for positional splat for literal array keyword argument

jeremyevans:
If all nodes in the array are safe, then it is safe to avoid allocation for the positional splat:

```ruby
m(*a, kw: [:a])   # Safe
m(*a, kw: [meth]) # Unsafe
```

This avoids an unnecessary allocation in a Rails method call. Details: https://github.com/rails/rails/pull/54949/files#r2052645431


byroot (approved):
I'm not particularly knowledgeable around that area, but seems sound.

PR #13675, opened 2025-06-21
variable.c: avoid out of bound write in `generic_field_set`

byroot:
[[Bug #21445]](https://bugs.ruby-lang.org/issues/21445)

FYI: @yahonda, @Earlopain @composerinteralia


composerinteralia:
Thank you! Confirmed this fixes things for GitHub too.

PR #13678, opened 2025-06-23
Make the critical level an enum

nobu:


ioquatix:
@nobu are you working on fixing stack overflow issues that we talked about at RubyKaigi?

PR #13680, opened 2025-06-23
[Bug #21449] Fix Set#divide{|a,b|} using Union-find structure

tompng:
Implement `Set#divide` by union-find structure with path compression.
https://en.wikipedia.org/wiki/Disjoint-set_data_structure
Since `divide{|a,b|}` calls the given block `n**2` times in the worst case, there is no need to implement union-by-rank or union-by-size optimization.

### DIFF
--- a/set.c
+++ b/set.c
@@ -843,66 +843,68 @@ set_i_classify(VALUE set)
     return args[0];
 }
 
-struct set_divide_args {
-    VALUE self;
-    VALUE set_class;
-    VALUE final_set;
-    VALUE hash;
-    VALUE current_set;
-    VALUE current_item;
-    unsigned long ni;
-    unsigned long nj;
-};
-
-static VALUE
-set_divide_block0(RB_BLOCK_CALL_FUNC_ARGLIST(j, arg))
+// Union-find with path compression
+static long
+set_divide_union_find_root(VALUE uf_parents, long index, VALUE tmp_array)
 {
-    struct set_divide_args *args = (struct set_divide_args *)arg;
-    if (args->nj > args->ni) {
-        VALUE i = args->current_item;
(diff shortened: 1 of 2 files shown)


set.c
```diff
@@ -843,66 +843,68 @@ set_i_classify(VALUE set)
...
+    VALUE uf_parents = rb_ary_new_capa(size);
+    for (long i = 0; i < size; i++) {
+        rb_ary_store(uf_parents, i, LONG2FIX(i));
+    }
+    for (long i = 0; i < size - 1; i++) {
+        VALUE item1 = RARRAY_AREF(items, i);
+        for (long j = i + 1; j < size; j++) {
+            VALUE item2 = RARRAY_AREF(items, j);
+            if (RTEST(rb_yield_values(2, item1, item2)) &&
+                RTEST(rb_yield_values(2, item2, item1))) {
```
jeremyevans, line 891: This is probably vulnerable to the same type of bug as https://bugs.ruby-lang.org/issues/21306, since the block could get access to the items array and shrink it while you are iterating.

jeremyevans (commented):
Thank you for finding and fixing this.

set.c
tompng: `items` is a snapshot of `set_i_to_a(set)`
Since union-find uses index, and `items` is used in the latter part of this function, item needs to be unmodified.

To avoid `items` and other internal arrays to be modified through ObjectSpace, I changed these in 99c5bf4053b5dde8dec7dc97d2960802617a69da
- Freeze items `rb_ary_freeze(items)` before using it
- Use ALLOCV_N instead of ruby array that stores index.

jeremyevans (approved):
Awesome! Thank you again for fixing this.

PR #13681, opened 2025-06-23
ZJIT: Compile from offset PC for optional args

composerinteralia:
Prior to this commit we hard-coded the starting `insn_idx` to 0. But optional args work by looking up an offset in the iseq's `opt_table` and starting from that offset instead.

We still need a guard to prevent executing with a different offset than we compiled for, but I'll do that in a separate commit.

### DIFF
--- a/zjit/src/cruby.rs
+++ b/zjit/src/cruby.rs
@@ -262,7 +262,7 @@ pub struct ID(pub ::std::os::raw::c_ulong);
 pub type IseqPtr = *const rb_iseq_t;
 
 // Given an ISEQ pointer, convert PC to insn_idx
-pub fn iseq_pc_to_insn_idx(iseq: IseqPtr, pc: *mut VALUE) -> Option<u16> {
+pub fn iseq_pc_to_insn_idx(iseq: IseqPtr, pc: *mut VALUE) -> Option<u32> {
     let pc_zero = unsafe { rb_iseq_pc_at_idx(iseq, 0) };
     unsafe { pc.offset_from(pc_zero) }.try_into().ok()
 }
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -76,6 +76,20 @@ class TestZJIT < Test::Unit::TestCase
     }
   end
 
+  def test_providing_optional_arg
+    assert_compiles 'true', %q{
+      def test(opt=false) = opt
+      test(true)
+    }
+  end
+
+  def test_omitting_optional_arg
+    assert_compiles 'false', %q{
+      def test(opt=false) = opt
+      test
+    }
+  end
+
   def test_invokebuiltin
     assert_compiles '["."]', %q{
       def test = Dir.glob(".")
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -71,7 +71,7 @@ impl JITState {
 
 /// CRuby API to compile a given ISEQ
 #[unsafe(no_mangle)]
-pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, _ec: EcPtr) -> *const u8 {
+pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, ec: EcPtr) -> *const u8 {
     // Do not test the JIT code in HIR tests
     if cfg!(test) {
         return std::ptr::null();
@@ -80,7 +80,7 @@ pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, _ec: EcPtr) -> *co
     // Take a lock to avoid writing to ISEQ in parallel with Ractors.
     // with_vm_lock() does nothing if the program doesn't use Ractors.
     let code_ptr = with_vm_lock(src_loc!(), || {
-        gen_iseq_entry_point(iseq)
+        gen_iseq_entry_point(iseq, ec)
     });
 
     // Assert that the ISEQ compiles if RubyVM::ZJIT.assert_compiles is enabled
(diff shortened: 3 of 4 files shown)


zjit/src/cruby.rs
```diff
@@ -262,7 +262,7 @@ pub struct ID(pub ::std::os::raw::c_ulong);
 pub type IseqPtr = *const rb_iseq_t;
 
 // Given an ISEQ pointer, convert PC to insn_idx
-pub fn iseq_pc_to_insn_idx(iseq: IseqPtr, pc: *mut VALUE) -> Option<u16> {
+pub fn iseq_pc_to_insn_idx(iseq: IseqPtr, pc: *mut VALUE) -> Option<u32> {
```
composerinteralia, line 265: We're using u32 elsewhere for insn_idx, so I changed this to match. YJIT has https://github.com/ruby/ruby/blob/af6b98f7a25670cb569f6da59904b3c05482b16e/yjit/src/yjit.rs#L149-L151, so if we want to do something similar I think we could use u16 everywhere instead.

k0kubun:
Q: Do we intend to compile a whole function for each PC offset?

tekknolagi:
I don't love the idea of compiling multiple JIT functions per ISEQ and I also don't love the idea of multiple entry blocks.

If our goal is to eventually handle all calls in JIT code and trampolines, I think this check should happen inside the CFG somehow

composerinteralia:
Thanks both. I was planning to ask about the approach in the next PR (to exit if the PCs don't match), but I can see now I should have asked about it here. Maybe we don't need the start_idx at all if we're going to compile something upfront that handles all the possibilities. (I wasn't sure if we'd be able to do something similar to YJIT where we guard with a stub that compiles for another PC only if needed—it sounds like that's not what we want 👍🏻.)

Are you imagining something (roughly) like this:

`def test(a=true, b=:foo.nil?, c=42) = [a,b,c]`

```
bb0(v0:BasicObject, v1:BasicObject, v2:BasicObject, v3:BasicObject):
  PCGuard offset: 0, jump: bb1(v0, v1, v2, v3)
  PCGuard offset: 4, jump: bb2(v0, v1, v2, v3)
  PCGuard offset: 10, jump: bb3(v0, v1, v2, v3)
  Jump bb4(v0, v1, v2, v3)
bb1(v4:BasicObject, v5:BasicObject, v6:BasicObject, v7:BasicObject)
  v8:TrueClassExact = Const Value(true)
  Jump bb2(v4, v8, v6, v7)
bb2(v9:BasicObject, v10:BasicObject, v11:BasicObject, v12:BasicObject)
  v13:StaticSymbol[VALUE(0x94710c)] = Const Value(VALUE(0x94710c))
  v14:BasicObject = SendWithoutBlock v13, :nil?
  Jump bb3(v9, v10, v14, v12)
bb3(v15:BasicObject, v16:BasicObject, v17:BasicObject, v18:BasicObject)
  v19:Fixnum[1] = Const Value(1)
  Jump bb4(v15, v16, v17, v19)
bb3(v20:BasicObject, v21:BasicObject, v22:BasicObject, v23:BasicObject)
  v24:ArrayExact = NewArray v21, v22, v23
  Return v24
```

k0kubun:
It seems like you intend to compile all that HIR in a single `Assembler` instance. I have two questions on the proposed HIR:

* Given that the method parameter setup (`gen_method_params()`, which is part of the JIT entry trampoline `gen_entry()`) also seems to depend on the PC offset, would you move the parameter setup to the main JIT code, or check PC in the trampoline instead and keep setting up parameters there?
* How do we reuse this code on JIT-to-JIT calls? Does JIT-to-JIT calls need to set the callee PC now for this PC-based block dispatch?

composerinteralia:
> How do we reuse this code on JIT-to-JIT calls? Does JIT-to-JIT calls need to set the callee PC now for this PC-based block dispatch?

We could probably also dispatch on # of args provided, if that helps anything. We probably need some more effort with checking args on direct sends in general anyway—there's stuff like `./miniruby --zjit-dump-hir -e 'def test(valid) = valid ? :a.to_sym : :a.to_sym(1,2,3); test(true); test(false)'` that should raise an argument error but doesn't.

composerinteralia:
> gen_method_params() also seems to depend on the PC offset

Oh, I didn't realize it did. Is it just that it produces extra code to unnecessarily init omitted values to nil, or is there a bug there that I haven't noticed yet?

k0kubun:
> Is it just that it produces extra code to unnecessarily init omitted values to nil

It loads arguments from the interpreter stack into registers used by JIT code.

composerinteralia:
> It loads arguments from the interpreter stack into registers used by JIT code.

Right, but isn't it getting them from the local table, which would have `nil`s if you didn't pass in all the arguments? E.g. when we call `def test(a=true, b=:foo.nil?, c=42) = [a,b,c]` with only 1 arg doesn't the stack end up looking something like this at the time we are calling `gen_method_params()`?

```
test(:a)

 0x8 ┌────────────┐
     │            │
 0x7 ├────────────┤◄──SP
     │ < flags  > │   
 0x6 ├────────────┤◄──EP
     │ <no block> │
 0x5 ├────────────┤  
     │ <CME: foo> │  
 0x4 ├────────────┤  
     │    nil     │◄── gen_method_params loads this into a register
 0x3 ├────────────┤  
     │    nil     │◄── gen_method_params loads this into a register
 0x2 ├────────────┤  
     │    :a      │◄── gen_method_params loads this into a register 
 0x1 ├────────────┤  
     │    self    │  
 0x0 └────────────┘
```

Or am I misunderstanding something? My impression was that loading those `nil`s into registers wasn't hurting anything, but it also wasn't necessary.

composerinteralia:
I've got a branch that implements the above (although not for direct sends yet, which in this implementation would indeed require us to set up the initial callee PC). If that looks like the direction we want to go, I can clean that branch up and open a PR with it (and close this one).

https://github.com/ruby/ruby/compare/master...composerinteralia:ruby:zjit-optional-args?expand=1

tekknolagi:
I think we should do a little more intentional whole-system design around this first---I think with the right calling convention, we can handle a bunch of different issues at once.

I don't know the specifics, but it looks like the little prelude for optional handling is not limited to baked-in constants, meaning that this:

```ruby
def bar = puts "hi"

def foo(a=1, b=bar) = 7
```

Leads to a call to `bar` being emitted inside the body's bytecode:

```
== disasm: #<ISeq:foo@opt.rb:3 (3,0)-(3,25)>
local table (size: 2, argc: 0 [opts: 2, rest: -1, post: 0, block: -1, kw: -1@-1, kwrest: -1])
[ 2] a@0<Opt=0> [ 1] b@1<Opt=3>
0000 putobject_INT2FIX_1_                                             (   3)
0001 setlocal_WC_0                          a@0
0003 putself
0004 opt_send_without_block                 <calldata!mid:bar, argc:0, FCALL|VCALL|ARGS_SIMPLE>
0006 setlocal_WC_0                          b@1
0008 getlocal_WC_0                          a@0[Ca]
0010 getlocal_WC_0                          b@1
0012 opt_plus                               <calldata!mid:+, argc:1, ARGS_SIMPLE>[CcCr]
0014 leave                                  [Re]
```

This is a strong case for handling this argument processing control-flow inside the CFG: we may have to side-exit inside argument handling.

We may end up needing something like multiple entry blocks 😅 

I wonder how V8/SpiderMonkey/... handle this.

Pricer187:
What are u sayin

composerinteralia:
OK, I'm going to put this aside for now and let everything brew. Maybe I can come back to it later (if somebody doesn't beat me to it).

PR #13682, opened 2025-06-23
Fix lock ordering issue for rb_ractor_sched_wait() and rb_ractor_sched_wakeup()

luke-gruber:
In rb_ractor_sched_wait() (ex: Ractor.receive), we acquire RACTOR_LOCK(cr) and then thread_sched_lock(cur_th). However, on wakeup if we're a dnt, in thread_sched_wait_running_turn() we acquire thread_sched_lock(cur_th) after condvar wakeup and then RACTOR_LOCK(cr). This lock inversion can cause a deadlock with rb_ractor_wakeup_all() (ex: port.send(obj)), where we acquire RACTOR_LOCK(other_r) and then thread_sched_lock(other_th).

So, the error happens:

nt 1:   Ractor.receive
            rb_ractor_sched_wait() after condvar wakeup in thread_sched_wait_running_turn():
              - thread_sched_lock(cur_th) (condvar) # acquires lock
              - rb_ractor_lock_self(cr) # deadlock here: tries to acquire, HANGS

nt 2: port.send
            ractor_wakeup_all()
              - RACTOR_LOCK(port_r) # acquires lock
              - thread_sched_lock # tries to acquire, HANGS

One solution would be to rework `thread_sched_wait_running_turn()` with DNT's. I didn't do this because it would be a bigger architectural change. What I changed is to unlock RACTOR_LOCK before calling rb_ractor_sched_wakeup() in a pthread env. In a non-pthread env it's safe to hold this lock, and we should.

Fixes [Bug #21398]

### DIFF
--- a/test/ruby/test_ractor.rb
+++ b/test/ruby/test_ractor.rb
@@ -144,6 +144,45 @@ class TestRactor < Test::Unit::TestCase
     RUBY
   end
 
+  # [Bug #21398]
+  def test_port_receive_dnt_with_port_send
+    assert_ractor(<<~'RUBY', timeout: 30)
+      THREADS = 10
+      JOBS_PER_THREAD = 50
+      ARRAY_SIZE = 20_000
+      def ractor_job(job_count, array_size)
+        port = Ractor::Port.new
+        workers = (1..4).map do |i|
+          Ractor.new(port) do |job_port|
+            while job = Ractor.receive
+              result = job.map { |x| x * 2 }.sum
+              job_port.send result
+            end
+          end
+        end
+        jobs = Array.new(job_count) { Array.new(array_size) { rand(1000) } }
+        jobs.each_with_index do |job, i|
+          w_idx = i % 4
+          workers[w_idx].send(job)
+        end
+        results = []
+        jobs.size.times do
+          result = port.receive # dnt receive
+          results << result
+        end
+        results
+      end
+      threads = []
+      # creates 40 ractors (THREADSx4)
+      THREADS.times do
+        threads << Thread.new do
+          ractor_job(JOBS_PER_THREAD, ARRAY_SIZE)
+        end
+      end
+      threads.each(&:join)
+    RUBY
+  end
+
   def assert_make_shareable(obj)
     refute Ractor.shareable?(obj), "object was already shareable"
     Ractor.make_shareable(obj)
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -1351,6 +1351,7 @@ rb_ractor_sched_wait(rb_execution_context_t *ec, rb_ractor_t *cr, rb_unblock_fun
     }
 
     thread_sched_lock(sched, th);
+    rb_ractor_unlock_self(cr);
     {
         // setup sleep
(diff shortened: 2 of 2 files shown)


jhawthorn:
Here's another reproduction of the deadlock that I can confirm this fixes: https://bugs.ruby-lang.org/issues/20346

``` ruby
N = 1000
ractors = N.times.map do
  Ractor.new do
    Ractor.recv # Ractor doesn't start until explicitly told to
    # Do some calculations
    fib = ->(x) { x < 2 ? 1 : fib.call(x - 1) + fib.call(x - 2) }
    fib.call(20)
  end
end

threads = ractors.map { |r| Thread.new { r.value } }
ractors.each { |r| r.send(nil) }
threads.each_with_index do |th, i|
  p(i => th.value)
end
```

On master this often deadlocks with as little as N=10

ractor_sync.c
```diff
@@ -983,7 +983,16 @@ ractor_wakeup_all(rb_ractor_t *r, enum ractor_wakeup_status wakeup_status)
             VM_ASSERT(waiter->wakeup_status == wakeup_none);
 
             waiter->wakeup_status = wakeup_status;
-            rb_ractor_sched_wakeup(r, waiter->th);
+#ifdef RUBY_THREAD_PTHREAD_H
```
ko1: how about it?

```C
do {
  struct ractor_waiter *waiter;
  RACTOR_LOCK{
    waiter = ccan_list_pop(&r->sync.waiters, struct ractor_waiter, node);
    if (waiter) waiter->wakeup_status = wakeup_status;
  }
  if (waiter) rb_ractor_sched_wakeup(...);
} while (waiter);
```

This patch is same to your patch essentially, but it is simpler.
Now I'm not sure

ractor_sync.c
```diff
@@ -1018,10 +1028,18 @@ ubf_ractor_wait(void *ptr)
                 waiter->wakeup_status = wakeup_by_interrupt;
                 ccan_list_del(&waiter->node);
 
+#ifdef RUBY_THREAD_PTHREAD_H
+                should_wake = true;
+#else
```
ko1: we can use `should_wake` for non pthread and can remove ifdef macro.

luke-gruber:
So I made the changes, but there are windows failures in CI. I think the ractor lock is necessary to hold during broadcasting due to the windows implementation of `rb_native_cond_broadcast`. I changed it to `rb_native_cond_signal`, which should be thread-safe as long as 2 nts don't call it at the same time which is currently not possible in this scenario.

Edit: I added another commit on top to try to fix windows failures but I'm still getting some. I'm going to move on to other things for now, but I would appreciate it if you could take another look or if you have any ideas please share 🙇

luke-gruber:
We were getting Windows timeout failures in CI with the approach of unlocking the ractor lock before wakeup. It's very hard to debug Windows issues especially because I don't have a computer with Windows installed (I tried for a couple of days), so I changed the waiter side instead. Everything seems to be working well. @ko1 Let me know if this is an acceptable solution for the time being. We plan on adding more assertions for lock ordering to prevent this kind of issue in the future, but that will be done in a separate PR.

PR #13683, opened 2025-06-23
ZJIT: Parse putspecialobject(VMCore) into Const

tekknolagi:
This way we get more information in HIR for the optimizer.

Fix https://github.com/Shopify/ruby/issues/587


zjit/src/hir_type/mod.rs
```diff
@@ -68,6 +69,7 @@ fn write_spec(f: &mut std::fmt::Formatter, printer: &TypePrinter) -> std::fmt::R
     let ty = printer.inner;
     match ty.spec {
         Specialization::Any | Specialization::Empty => { Ok(()) },
+        Specialization::Object(val) if val == unsafe { rb_mRubyVMFrozenCore } => write!(f, "[VMFrozenCore]"),
```
tekknolagi, line 72: This is optional and just for easier reading but we can omit if it feels icky

k0kubun (approved):

PR #13689, opened 2025-06-24
ZJIT: Implement getlocal and setlocal (for nested scopes)

XrXr:
- **ZJIT: `getlocal` and `setlocal` to HIR**
- **ZJIT: Add codegen for GetLocal and SetLocal**

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -81,6 +81,28 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 3, insns: [:getlocal, :setlocal, :getlocal_WC_0, :setlocal_WC_1]
   end
 
+  def test_read_local_written_by_children_iseqs
+    omit "This test fails right now because Send doesn't compile."
+
+    assert_compiles '[1, 2]', %q{
+      def test
+        l1 = nil
+        l2 = nil
+        tap do |_|
+          l1 = 1
+          tap do |_|
+            l2 = 2
+          end
+        end
+
+        [l1, l2]
+      end
+
+      test
+      test
+    }, call_threshold: 2
+  end
+
   def test_send_without_block
     assert_compiles '[1, 2, 3]', %q{
       def foo = 1
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2543,11 +2543,17 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     break;  // Don't enqueue the next block as a successor
                 }
                 YARVINSN_getlocal_WC_0 => {
+                    // TODO(alan): This implementation doesn't read from EP, so will miss writes
+                    // from nested ISeqs. This will need to be amended when we add codegen for
+                    // Send.
                     let ep_offset = get_arg(pc, 0).as_u32();
                     let val = state.getlocal(ep_offset);
                     state.stack_push(val);
                 }
                 YARVINSN_setlocal_WC_0 => {
+                    // TODO(alan): This implementation doesn't write to EP, where nested scopes
+                    // read, so they'll miss these writes. This will need to be amended when we
+                    // add codegen for Send.
                     let ep_offset = get_arg(pc, 0).as_u32();
                     let val = state.stack_pop()?;
                     state.setlocal(ep_offset, val);
@@ -3527,16 +3533,16 @@ mod tests {
             expect![[r#"
                 fn block (3 levels) in <compiled>:
                 bb0(v0:BasicObject):
-                  v2:BasicObject = GetLocal l2, 4
-                  SetLocal l1, 3, v2
-                  v4:BasicObject = GetLocal l1, 3
-                  v5:BasicObject = GetLocal l2, 4
+                  v2:BasicObject = GetLocal l2, EP@4
+                  SetLocal l1, EP@3, v2
+                  v4:BasicObject = GetLocal l1, EP@3
+                  v5:BasicObject = GetLocal l2, EP@4
                   v7:BasicObject = SendWithoutBlock v4, :+, v5
-                  SetLocal l2, 4, v7
-                  v9:BasicObject = GetLocal l2, 4
-                  v10:BasicObject = GetLocal l3, 5
+                  SetLocal l2, EP@4, v7
+                  v9:BasicObject = GetLocal l2, EP@4
+                  v10:BasicObject = GetLocal l3, EP@5
                   v12:BasicObject = SendWithoutBlock v9, :+, v10
-                  SetLocal l3, 5, v12
+                  SetLocal l3, EP@5, v12
                   Return v12
             "#]]
         );


XrXr:
~I've removed the codegen part of this PR for now since there is a x86-only bug with it.~

EDIT: fixed

k0kubun (approved):

XrXr:
Also added 2 minor cleanups.

test/ruby/test_zjit.rb
```diff
@@ -62,6 +62,25 @@ def test(n)
...
+      1.times do |l2|
+        1.times do |l1|
+          define_method(:test) do
+            l1 = 1
+            l2 = 2
+            l3 = 3
+            [l1, l2, l3]
+          end
+        end
+      end
```
tekknolagi, line 76: This PR adds support for the inner functions to write to outer locals but it doesn't look like it modifies those outer functions to handle their locals being written to

What I mean is if you read from `l2` in the outer `1.times` loop, you will see a `getlocal_WC_0` which we have folded away in our bytecode->IR pass. In other words, we'll miss the fact that the inner function writes to the frame.

I think this happens to "work" accidentally right now because we abort compiling if we have send with block (`Send`) and therefore never have a JIT compiled function have its locals modified like this.

Since block ISEQs are known statically, we can/should detect this kind of nested scope local writing in the block ISEQs and penalize the locals in the upper frame(s) that might get written to like this.

I don't know that there is necessarily a full fix needed in this current PR, but:

1) We should write this down somewhere, probably a TODO in the code and a Ruby issue, and
2) We should add a test for catching that local being written to from a nested ISEQ; it will pass right now

test/ruby/test_zjit.rb
```diff
@@ -791,7 +810,7 @@ def assert_compiles(expected, test_script, insns: [], **opts)
       result = {
         ret_val:,
         #{ unless insns.empty?
-          'insns: RubyVM::InstructionSequence.of(_test_proc).enum_for(:each_child).map(&:to_a)'
+          'insns: RubyVM::InstructionSequence.of(method(:test)).to_a'
```
tekknolagi, line 835: The comment above is no longer correct.

@k0kubun do we rely, implicitly or explicitly, on this checking that *every* function is compiled? I would guess that this was introduced near your direct-call-JIT patch

zjit/src/codegen.rs
```diff
@@ -298,6 +301,37 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
...
+    let ep_opnd = Opnd::mem(64, CFP, RUBY_OFFSET_CFP_EP);
+    let mut ep_opnd = asm.load(ep_opnd);
+
+    for _ in 0..level {
+        // Get the previous EP from the current EP
+        // See GET_PREV_EP(ep) macro
+        // VALUE *prev_ep = ((VALUE *)((ep)[VM_ENV_DATA_INDEX_SPECVAL] & ~0x03))
+        let offs = SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL;
+        ep_opnd = asm.load(Opnd::mem(64, ep_opnd, offs));
+        ep_opnd = asm.and(ep_opnd, Opnd::Imm(!0x03));
```
tekknolagi: Can we make the `0x03` a special named const here, even though it's not in the original C? Helps document things

zjit/src/codegen.rs
```diff
@@ -298,6 +301,37 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
...
+fn gen_get_ep(asm: &mut Assembler, level: u32) -> Opnd {
+    // Load environment pointer EP from CFP into a register
+    let ep_opnd = Opnd::mem(64, CFP, RUBY_OFFSET_CFP_EP);
+    let mut ep_opnd = asm.load(ep_opnd);
+
+    for _ in 0..level {
+        // Get the previous EP from the current EP
+        // See GET_PREV_EP(ep) macro
+        // VALUE *prev_ep = ((VALUE *)((ep)[VM_ENV_DATA_INDEX_SPECVAL] & ~0x03))
+        let offs = SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL;
```
```suggestion
        let offset = SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL;
```

Let's not arbitrarily shorten please

zjit/src/codegen.rs
```diff
@@ -298,6 +301,37 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
...
+        ep_opnd = asm.load(Opnd::mem(64, ep_opnd, offs));
+        ep_opnd = asm.and(ep_opnd, Opnd::Imm(!0x03));
+    }
+
+    ep_opnd
+}
+
+fn gen_nested_getlocal(asm: &mut Assembler, ep_offset: u32, level: NonZeroU32) -> Option<lir::Opnd> {
+    let ep = gen_get_ep(asm, level.get());
+    let offs = -(SIZEOF_VALUE_I32 * i32::try_from(ep_offset).ok()?);
```
```suggestion
    let offset = -(SIZEOF_VALUE_I32 * i32::try_from(ep_offset).ok()?);
```

zjit/src/codegen.rs
```diff
@@ -298,6 +301,37 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
...
+    Some(asm.load(Opnd::mem(64, ep, offs)))
+}
+
+fn gen_nested_setlocal(asm: &mut Assembler, val: Opnd, ep_offset: u32, level: NonZeroU32) -> Option<()> {
+    let ep = gen_get_ep(asm, level.get());
+    let offs = -(SIZEOF_VALUE_I32 * i32::try_from(ep_offset).ok()?);
```
```suggestion
    let offset = -(SIZEOF_VALUE_I32 * i32::try_from(ep_offset).ok()?);
```

tekknolagi: It's a little confusing that this is also an EP offset but it's different somehow. Maybe there are better names?

zjit/src/codegen.rs
```diff
@@ -446,19 +480,6 @@ fn gen_method_params(asm: &mut Assembler, iseq: IseqPtr, entry_block: &Block) {
     }
 }
 
-/// Set branch params to basic block arguments
```
tekknolagi: Can you put this back where it was to minimize changes?

zjit/src/hir.rs
```diff
@@ -447,6 +448,10 @@ pub enum Insn {
     /// Check whether an instance variable exists on `self_val`
     DefinedIvar { self_val: InsnId, id: ID, pushval: VALUE, state: InsnId },
 
+    /// Get local varaible from non-immediate scope
```
```suggestion
    /// Get local variable from non-immediate scope
```

zjit/src/hir.rs
```diff
@@ -447,6 +448,10 @@ pub enum Insn {
...
+    GetLocal { level: NonZeroU32, ep_offset: u32 },
+    /// Set a local varaible in a non-immediate scope
```
```suggestion
    /// Set a local variable in a non-immediate scope
```

tekknolagi: Is "non-immediate" a Ruby term? Can we say "higher" scope or something?

zjit/src/hir.rs
```diff
@@ -1741,6 +1755,7 @@ impl Function {
                 | Insn::Return { val }
                 | Insn::Defined { v: val, .. }
                 | Insn::Test { val }
+                | Insn::SetLocal{ val, .. }
```
```suggestion
                | Insn::SetLocal { val, .. }
```

zjit/src/hir.rs
```diff
@@ -2174,6 +2189,7 @@ pub enum CallType {
 #[derive(Debug, PartialEq)]
 pub enum ParseError {
     StackUnderflow(FrameState),
+    MalformIseq(u32), // insn_idx into iseq_encoded
```
```suggestion
    MalformedIseq(u32), // insn_idx into iseq_encoded
```

zjit/src/hir.rs
```diff
@@ -696,6 +702,8 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
             Insn::SetIvar { self_val, id, val, .. } => write!(f, "SetIvar {self_val}, :{}, {val}", id.contents_lossy().into_owned()),
             Insn::GetGlobal { id, .. } => write!(f, "GetGlobal :{}", id.contents_lossy().into_owned()),
             Insn::SetGlobal { id, val, .. } => write!(f, "SetGlobal :{}, {val}", id.contents_lossy().into_owned()),
+            Insn::GetLocal { level, ep_offset } => write!(f, "GetLocal l{level}, {ep_offset}"),
+            Insn::SetLocal { val, level, ep_offset } => write!(f, "SetLocal l{level}, {ep_offset}, {val}"),
```
```suggestion
            Insn::GetLocal { level, ep_offset } => write!(f, "GetLocal l{level}, EP@{ep_offset}"),
            Insn::SetLocal { val, level, ep_offset } => write!(f, "SetLocal l{level}, EP@{ep_offset}, {val}"),
```

tekknolagi (approved):
Looks good -- some comments

test/ruby/test_zjit.rb
tekknolagi: Maybe it's a bad guess but I kind of expect upper-local writes like this to be common but other get-my-caller-and-write-its-locals shenanigans to be more rare.

XrXr: 👍🏼 6e85878c357ccf9906bb480ab8a520087ed8b8d0. Actually, the test fails right now because assert_compile() catches that Send doesn't compile. So I'm checking it in as skipped for now. https://github.com/Shopify/ruby/issues/596

test/ruby/test_zjit.rb
XrXr: It's more like "assert that all compilation attempts are successful at a certain call threshold". I think it's a possible reading of the current comment, but maybe it's not very clear?

PR #13695, opened 2025-06-25
Fix Windows CI for Ruby 3.4

hsbt:
https://github.com/ruby/ruby/actions/runs/15866161213/job/44733343866 is failing now. It may be fixed with the following commit.

e1adb6cb15129a54df0c55a337e98b92b2a55e3f

```
Win: Suppress false warnings from Visual C 17.14.1

https://developercommunity.visualstudio.com/t/warning-C5287:-operands-are-different-e/10877942?

It is not able to silence "operands are different enum types" warnings, even using an explicit cast, as the message says.
```

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -96,13 +96,6 @@ jobs:
           windows-toolchain: none
         if: ${{ matrix.os != '2025' }}
 
-      - name: Export GitHub Actions cache environment variables
-        uses: actions/github-script@v7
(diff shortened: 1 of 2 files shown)


hsbt:
I will replace binary cache strategy for vcpkg after resolving the current issue.

hsbt:
https://github.com/ruby/ruby/actions/runs/15868941715/job/44757107844 vcpkg cache is working now.

PR #13696, opened 2025-06-25
Fix missing write barrier when on too_complex

jhawthorn:
We were creating a new fields object, and then inserting into it without a write barrier.

I was looking into this because of this CI failure: https://rubyci.s3.amazonaws.com/debian11/ruby-master/log/20250624T213003Z.fail.html.gz

```
/home/chkbuild/chkbuild/tmp/build/20250624T213003Z/ruby/test/openssl/utils.rb:234: [BUG] try to mark T_NONE object
ruby 3.5.0dev (2025-06-24T20:41:45Z master 3b602c952d) +PRISM [x86_64-linux]
```

When running OpenSSL::TestSSL, wbcheck reports:

    OpenSSL::TestSSL#test_tlsext_hostname
    WBCHECK ERROR: Missed write barrier detected!
      Parent object: 0x7ce8b3a390c0 (wb_protected: true)
        rb_obj_info_dump: 0x00007ce8b3a390c0 T_IMEMO/<fields>
      Reference counts - snapshot: 1, writebarrier: 1, current: 4, missed: 2
      Missing reference to: 0x7c08b4110750
        rb_obj_info_dump: 0x00007c08b4110750 OpenSSL/X509/OpenSSL::X509::Certificate OpenSSL/X509
      Missing reference to: 0x7c08b4128dd0
        rb_obj_info_dump: 0x00007c08b4128dd0 OpenSSL/EVP_PKEY/OpenSSL::PKey::RSA OpenSSL/EVP_PKEY

Another thing I noticed is that in some other cases (for T_OBJECT?), we're allocating an st_table first and only after assigning it to the object. That _might_ be fine, but I wonder if that could fail under autocompact. I kept this building the table first and then inserting with a writebarrier after for that reason.


byroot:
> we're allocating an st_table first and only after assigning it to the object. That might be fine, but I wonder if that could fail under autocompact.

It's supposed to be fine because the table is created with the right size, so it's not supposed to allocate.

variable.c
```diff
@@ -1645,6 +1645,32 @@ rb_obj_init_too_complex(VALUE obj, st_table *table)
...
 
+static int
+imemo_fields_complex_from_obj_i(ID key, VALUE val, st_data_t arg)
+{
+    VALUE fields = (VALUE)arg;
+    st_table *table = rb_imemo_fields_complex_tbl(fields);
+
+    RUBY_ASSERT(!st_lookup(table, (st_data_t)key, NULL));
+    st_add_direct(table, (st_data_t)key, (st_data_t)val);
+    RB_OBJ_WRITTEN(fields, Qundef, key);
```
byroot: The key is an `ID` so it doesn't need a write barrier AFAIK.

PR #13701, opened 2025-06-25
Introduce Namespace#eval

tenderlove:
This commit adds an `eval` method to `Namespace` that takes a string and evaluates the string as Ruby code within the context of that namespace.

For example:

```ruby
n = Namespace.new

n.eval("class TestClass; def hello; 'from namespace'; end; end")

instance = n::TestClass.new
instance.hello # => "from namespace"
```

[Feature #21365]


tagomoris:
LGTM 👍

PR #13703, opened 2025-06-25
Fix the exception when trying to copy procs for Ractors

tenderlove:
`rb_obj_clone` can possibly raise an exception, and the exception isn't always helpful.  This commit catches the exception and raises an IsolationError telling the user what went wrong.

Before this commit:

```
> ruby -e'Ractor.make_shareable(->{}, copy:true)'
<internal:ractor>:828:in 'Ractor.make_shareable': allocator undefined for Proc (TypeError)
	from -e:1:in '<main>'
```

After this commit:

```
> ./miniruby -e'Ractor.make_shareable(->{}, copy:true)'
-e:1:in 'Ractor.make_shareable': cannot copy #<Proc:0x000000011f311a80 -e:1 (lambda)> (Ractor::IsolationError)
	from -e:1:in '<main>'
-e:1:in 'Ractor.make_shareable': allocator undefined for Proc (TypeError)
	from -e:1:in '<main>'
```


ractor.c
```diff
@@ -1917,6 +1917,12 @@ ractor_move(VALUE obj)
     }
 }
 
+static VALUE
+copy_clone_wrapper(VALUE obj)
+{
+    return rb_obj_clone(obj);
+}
+
```
nobu: Is this wrapper needed?

tenderlove: No. I removed it. Thank you!

PR #13705, opened 2025-06-25
Assert AR table entry stability on replace

sferik:
Improve hash internal iteration by verifying that AR table entries remain stable after `st_update_callback_func` replaces a key-value pair. Fixes a [TODO](https://github.com/ruby/ruby/commit/72825c35b0d8b9d566663de961fddbf4f010fff7#diff-2894a90fd17590e4f7afd08dfd8505526d5283e09fea645cc7a7d4873936bdffR864) by @ko1 from 2019.

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -882,8 +882,9 @@ ar_general_foreach(VALUE hash, st_foreach_check_callback_func *func, st_update_c
                 if (replace) {
                     (*replace)(&key, &val, arg, TRUE);
 
-                    // TODO: pair should be same as pair before.
-                    pair = RHASH_AR_TABLE_REF(hash, i);
+                    // Pair should not have moved
+                    HASH_ASSERT(pair == RHASH_AR_TABLE_REF(hash, i));
+
                     pair->key = (VALUE)key;
                     pair->val = (VALUE)val;
                 }


hash.c
```diff
@@ -880,10 +880,11 @@ ar_general_foreach(VALUE hash, st_foreach_check_callback_func *func, st_update_c
                 return 0;
               case ST_REPLACE:
                 if (replace) {
+                    ar_table_pair *orig_pair = pair;
                     (*replace)(&key, &val, arg, TRUE);
 
-                    // TODO: pair should be same as pair before.
                     pair = RHASH_AR_TABLE_REF(hash, i);
+                    HASH_ASSERT(orig_pair == pair);
```
jhawthorn: Reading this, I think what the original TODO may have intended is that in non-debug builds, maybe we don't need to do the `RHASH_AR_TABLE_REF` lookup a second time? What do you think?

```suggestion
                    (*replace)(&key, &val, arg, TRUE);

                    // Pair should not have moved
                    HASH_ASSERT(pair == RHASH_AR_TABLE_REF(hash, i));
```

sferik: This makes sense to me. I tried committing your patch but got the error:

> Applying suggestions on deleted lines is currently not supported.

I'm going to manually apply your patch and push it for review.

I'd love for @ko1 to comment on his original intent for this TODO.a

sferik:
@jhawthorn Anything else you need me to change before this is merged?

jhawthorn:
Nope! I meant to merge this last week.

jhawthorn:
It's nice to see you opening PRs to Ruby Erik ❤️

sferik:
Thanks! Now that they're being merged, I'm encouraged to make some more!

PR #13707, opened 2025-06-25
Freeze UNUSABLE_CHARS constant

sferik:
The constant listing invalid characters for temporary file names is now frozen to prevent accidental modification, aligning it with the other constants in the module.


nobu:
This file has `frozen_string_literal: true` already.

PR #13712, opened 2025-06-26
Fix race condition in signal handler query

sferik:
Introduced a static mutex when checking non-POSIX signal handlers, making the operation thread-safe and preventing races around the signal retrieval and restoration. Fixes TODO added by @luke-gruber in https://github.com/ruby/ruby/commit/1d4822a175a0dfccca8f252b0e757a1991bd54f9#diff-47e173231fbc1ddb6c555f7116c157cb32ab63d04e85047b55ff378e71fb965eR679.

### DIFF
--- a/signal.c
+++ b/signal.c
@@ -678,9 +678,12 @@ signal_ignored(int sig)
     // SIG_GET: Returns the current value of the signal.
     func = signal(sig, SIG_GET);
 #else
-    // TODO: this is not a thread-safe way to do it. Needs lock.
-    sighandler_t old = signal(sig, SIG_DFL);
+    static rb_nativethread_lock_t sig_check_lock = RB_NATIVETHREAD_LOCK_INIT;
+    sighandler_t old;
+    rb_native_mutex_lock(&sig_check_lock);
+    old = signal(sig, SIG_DFL);
     signal(sig, old);
+    rb_native_mutex_unlock(&sig_check_lock);
     func = old;
 #endif
     if (func == SIG_IGN) return 1;
--- a/signal.c
+++ b/signal.c
@@ -664,7 +664,9 @@ ruby_nativethread_signal(int signum, sighandler_t handler)
 #endif
 #endif
 
+#if !defined(POSIX_SIGNAL) && !defined(SIG_GET)
 static rb_nativethread_lock_t sig_check_lock;
+#endif
 
 static int
 signal_ignored(int sig)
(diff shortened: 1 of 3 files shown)


luke-gruber:
Thanks for taking a look at this! It looks like Windows stuff is failing because there's no static initializer for Window's version of `rb_nativethread_lock_t` (see `thread_native.h` for internals). You can make it file local and init in during `Init_signal`.

This mutex also needs to be reinitialized after a call to `fork` or else the child might hit a deadlock. In `rb_thread_atfork_internal()` you can add a `rb_signal_atfork()` and do it there.

sferik:
@luke-gruber Thanks for the feedback! I will work on making these changes and update the PR.

sferik:
@luke-gruber I've made the changes you've recommend:
* Initialized this mutex during signal setup so it’s ready before any handler queries happen.
* Provided a fork hook to reinitialize the lock after forking and invoked it from the thread fork path to prevent deadlocks in the child.
* Tests now pass on all platforms, including Windows.

Please let me know if you have any additional feedback.

signal.c
```diff
@@ -664,6 +664,8 @@ ruby_nativethread_signal(int signum, sighandler_t handler)
 #endif
 #endif
 
+static rb_nativethread_lock_t sig_check_lock;
```
nobu, line 668: This variable seems to be required only under certain conditions.

signal.c
```diff
@@ -1561,3 +1566,13 @@ Init_signal(void)
 
     rb_enable_interrupt();
 }
+
+ #if defined(HAVE_WORKING_FORK)
```
nobu: This condition doesn’t look matching the above.
And the function is always needed, so the condition can be in the function body.

sferik:
I've made the allocation and initialization of `sig_check_lock` conditional on `POSIX_SIGNAL` and `SIG_GET` being defined. I've also changed the code to always define `rb_signal_atfork`, moving the conditions inside the function body.

Please let me know if I've gotten anything wrong and if you'd like me to make any other changes (e.g. squashing these changes into a single commit). Thanks so much for reviewing this code!

nobu (approved):
By the way, do you know any platform that is still alive, neither POSIX nor Windows?

sferik:
I guess some bespoke real-time OSes, stuff within the government (e.g. systems that control rocket ships), and very old mainframes that are still technically "alive", for some definition of that word, but just about everything else is POSIX or Windows.

nobu:
Ah, sorry, I meant any of the platforms supported by ruby.

PR #13715, opened 2025-06-26
Extract Ractor safe table used for frozen strings

peterzhu2118:
This commit extracts the Ractor safe table used for frozen strings into ractor_safe_table.c, which will allow it to be used elsewhere, including for the global symbol table.


ractor_safe_table.c
```diff
@@ -0,0 +1,323 @@
...
+          case RACTOR_SAFE_TABLE_EMPTY:
+            // We didn't find our entry to delete.
+            return 0;
+          case RACTOR_SAFE_TABLE_DELETED:
+            break;
+          case RACTOR_SAFE_TABLE_MOVED:
+            rb_bug("rb_ractor_safe_table_delete: moved entry");
+            break;
+          default:
+            if (table->funcs->cmp(key, curr_key)) {
```
jhawthorn: In the existing code we compared by identity here, right? I'd really like to keep that as otherwise it's easy to end up with [Bug #21172](https://bugs.ruby-lang.org/issues/21172)

jhawthorn: Maybe we could call this `rb_ractor_safe_table_delete_by_identity`? I feel like we're going to want to do this for the symbol table as well (since we're likely to build duplicate dsymbols when concurrently inserting).

ractor_safe_table.c
```diff
@@ -0,0 +1,323 @@
...
+        ractor_safe_table_try_resize_without_locking(old_table_obj, table_obj_ptr);
+    }
+}
+
+VALUE
+rb_ractor_safe_table_find_or_insert(VALUE *table_obj_ptr, VALUE key, void *data)
+{
+    bool inserting = false;
+    VALUE table_obj;
+
```
jhawthorn: Just for future proofing new uses of this table, let's add

```suggestion
RUBY_ASSERT(key >= RACTOR_SAFE_TABLE_SPECIAL_VALUE_COUNT);

```

jhawthorn:
One comment, otherwise looks good! How do you feel about the naming? I think it might be better to clarify that this is only a hash set not a table, but not totally necessary (and this is just internal so we can change our mind any time)

peterzhu2118:
I don't have strong opinions on the name. I changed it to `ractor_safe_set`.

PR #13716, opened 2025-06-26
Fix flaky TestGc#test_heaps_grow_independently

peterzhu2118:
The test sometimes fails with "Expected 2062788 to be < 2000000" because heap 0 has not been cleared yet by GC. This commit fixes it to run GC before the assertion to ensure that it does not flake.

### DIFF
--- a/test/ruby/test_gc.rb
+++ b/test/ruby/test_gc.rb
@@ -701,6 +701,11 @@ class TestGc < Test::Unit::TestCase
         allocate_large_object
       end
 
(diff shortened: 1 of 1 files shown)


test/ruby/test_gc.rb
```diff
@@ -701,6 +701,8 @@ def allocate_large_object = Array.new(10)
         allocate_large_object
       end
 
+      GC.start
```
tekknolagi, line 707: Nit: add a comment explaining why you GC?

PR #13717, opened 2025-06-26
add VM Lock around `rb_const_remove` operations (Module#remove_const)

luke-gruber:
Without a VM Lock, there's an unlocked `rb_id_table_delete` for the class's const_tbl which can cause problems. Example:

```ruby
class C
  CONSTANT = 3
end
$VERBOSE = nil
rs = []
100.times do
  rs << Ractor.new do
    10_000.times do
      if defined?(C::CONSTANT)
        C.send(:remove_const, :CONSTANT) rescue NameError
      else
        C.send(:const_set, :CONSTANT, 3)
      end
    end
  end
end
while rs.any?
  r, obj = Ractor.select(*rs)
  rs.delete(r)
end
```

Without lock:
../ruby-release/test.rb:14: [BUG] Segmentation fault at 0x0000000000000001 -- Control frame information ----------------------------------------------- miniruby(82790,0x16f49f000) malloc: *** error for object 0x600000f880a0: pointer being freed was not allocated miniruby(82790,0x16f49f000) malloc: *** set a breakpoint in malloc_error_break to debug

### DIFF
--- a/constant.h
+++ b/constant.h
@@ -44,7 +44,8 @@ void rb_free_const_table(struct rb_id_table *tbl);
 VALUE rb_const_source_location(VALUE, ID);
 
 int rb_autoloading_value(VALUE mod, ID id, VALUE *value, rb_const_flag_t *flag);
-rb_const_entry_t *rb_const_lookup(VALUE klass, ID id, rb_const_entry_t *entry_out);
+rb_const_entry_t *rb_const_lookup_locked(VALUE klass, ID id);
+bool rb_const_lookup_unlocked(VALUE klass, ID id, rb_const_entry_t *entry_out);
 VALUE rb_public_const_get_at(VALUE klass, ID id);
 VALUE rb_public_const_get_from(VALUE klass, ID id);
 int rb_public_const_defined_from(VALUE klass, ID id);
--- a/variable.c
+++ b/variable.c
@@ -3025,8 +3025,8 @@ autoload_synchronized(VALUE _arguments)
     struct autoload_arguments *arguments = (struct autoload_arguments *)_arguments;
     rb_const_entry_t constant_entry = {0};
 
-    rb_const_entry_t *ce = rb_const_lookup(arguments->module, arguments->name, &constant_entry);
-    if (ce && !UNDEF_P(constant_entry.value)) {
+    bool found = rb_const_lookup_unlocked(arguments->module, arguments->name, &constant_entry);
+    if (found && !UNDEF_P(constant_entry.value)) {
         return Qfalse;
     }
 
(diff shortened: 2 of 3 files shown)


jhawthorn:
Seems like we might be missing locking for const lookup as well. 

``` ruby
class C
  CONSTANT = 3
end
$VERBOSE = nil
rs = []
100.times do
  rs << Ractor.new do
    10_000.times do
      if defined?(C::CONSTANT)
        C.send(:remove_const, :CONSTANT) rescue NameError
      else
        C.send(:const_set, :CONSTANT, 3)
      end
      C.const_get(:CONSTANT) rescue NameError
      C.const_get(:CONSTANT) rescue NameError
    end
  end
end
while rs.any?
  r, obj = Ractor.select(*rs)
  rs.delete(r)
end
```

```
make run
./miniruby -I./lib -I. -I.ext/common  -r./x86_64-linux-fake  ./test.rb
<OBJ_INFO:obj_traverse_i@ractor.c:1276> out-of-heap:0x000055700ced3ea0
./test.rb:14: [BUG] unreachable
ruby 3.5.0dev (2025-06-26T16:21:30Z ractor_safe_remove.. c3420bfd76) +PRISM [x86_64-linux]

c:0005 p:---- s:0019 e:000018 CFUNC  :const_get
c:0004 p:0047 s:0014 e:000013 BLOCK  ./test.rb:14
c:0003 p:0024 s:0011 e:000010 METHOD <internal:numeric>:257
c:0002 p:0005 s:0006 e:000005 BLOCK  ./test.rb:8 [FINISH]
c:0001 p:---- s:0003 e:000002 DUMMY  [FINISH]

-- Ruby level backtrace information ----------------------------------------
./test.rb:8:in 'block (2 levels) in <main>'
<internal:numeric>:257:in 'times'
./test.rb:14:in 'block (3 levels) in <main>'
./test.rb:14:in 'const_get'
```

byroot:
Should `const_set` and `remove_const` even be allowed from non-main ractor? It seems to me that if you can't set a class ivar, you shouldn't be able to set a const, no?

luke-gruber:
Yeah it's a good point, but for whatever reason this has always been allowed in ractors, and it was by design not by accident.  As for why, I don't know but maybe it's to do with supporting autoloading in ractors. If you support autoloading constants, it's a small step away from supporting setting constants directly.

constant.h
```diff
@@ -44,7 +44,7 @@ void rb_free_const_table(struct rb_id_table *tbl);
 VALUE rb_const_source_location(VALUE, ID);
 
 int rb_autoloading_value(VALUE mod, ID id, VALUE *value, rb_const_flag_t *flag);
-rb_const_entry_t *rb_const_lookup(VALUE klass, ID id);
+rb_const_entry_t *rb_const_lookup(VALUE klass, ID id, rb_const_entry_t *entry_out);
```
jhawthorn: Should we make this return an int bool like `st_lookup`?

jhawthorn:
`const_set` and `remove_const` from a non-main ractor surprised me as well, but I don't think changing that would make a big difference from an implementation standpoint. We'd still have to lock for both reads and writes.

constant.h
luke-gruber: It's a little trickier than that because sometimes we want to do a lookup and get back the raw `ce` to modify it. I changed `rb_const_lookup` to 2 functions: `rb_const_lookup_locked` and `*_unlocked`. Locked variant returns raw `ce` and asserts VM lock is taken.

luke-gruber:
Jean is working on lock-free const get/set/remove with RCU pattern, so this will probably not be merged but I'm going to keep it open until then.

PR #13727, opened 2025-06-27
Reduce allocations in `Gem::BUNDLED_GEMS.warning?`

joshuay03:
When profiling our (@buildkite's) Rails application's boot, this `#split` consistently showed up as taking 3-4% of allocation samples (~125k+ allocations) with a sample rate of 100. I suspect this may be related to our use of bootsnap given the comments in this method, though I haven't tested without it to confirm.

<details>

<summary>Unfortunately I can't share that specific a profile, but I tried to compare profiles using a micro test script</summary>

<br>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "vernier"

  gem "rails", require: false
  gem "puma", require: false
  gem "sidekiq", require: false
  gem "rspec", require: false

  gem 'bootsnap', require: false
end

require "bootsnap"
Bootsnap.setup(development_mode: true, cache_dir: "tmp/cache")

# Vernier.profile(out: "../time_build.json") do
Vernier.profile(out: "../alloc_build.json", allocation_interval: 10) do
  require "rails"
  require "puma"
  require "sidekiq"
  require "rspec"
end

# rm -rf ~/.rubies/ruby-master/lib/ruby/gems/3.5.0+2/gems && rm -rf tmp/cache && make && make install && ~/.rubies/ruby-master/bin/ruby ../test.rb
```

Viewer: https://vernier.prof/ - filter data source to allocations and invert call tree stack
- [alloc_master.json](https://github.com/user-attachments/files/20949591/alloc_master.json)
- [alloc_build.json](https://github.com/user-attachments/files/20949597/alloc_build.json)

</details>

I didn't open an issue for this, but I did come across https://bugs.ruby-lang.org/issues/20641, so it seems this method is known to be somewhat of a hotspot.


lib/bundled_gems.rb
```diff
@@ -105,11 +105,11 @@ def self.warning?(name, specs: nil)
       # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
       feature.delete_prefix!(ARCHDIR)
       feature.delete_prefix!(LIBDIR)
-      segments = feature.split("/")
+      segments = feature.split("/", 3)
```
joshuay03: This branch needs to have three segments at most.

lib/bundled_gems.rb
```diff
@@ -105,11 +105,11 @@ def self.warning?(name, specs: nil)
...
       name = segments.shift
       name = EXACT[name] || name
       if !SINCE[name]
-        name = [name, segments.shift].join("-")
+        name = "#{name}-#{segments.shift}"
```
joshuay03, line 109: While I was here I thought I'd eliminate this array allocation. This was only showing up at around 0.5%, but why not 🤷🏽‍♂️

joshuay03:
> I suspect this may be related to our use of bootsnap given the comments in this method, though I haven't tested without it to confirm.

Actually I did just confirm this. Without bootsnap, I see many features like `psych/y`, but with bootsnap they're mostly like `/Users/me/Projects/buildkite/buildkite/.bundle/ruby/3.5.0+2/gems/psych-5.2.6/lib/psych/y`. So there's a lot of unnecessary splitting that'll just end up being discarded—no warning.

lib/bundled_gems.rb
tenderlove: Do you know why it only needs 3 segments at most? I don't doubt it, but it's not obvious (to me) from the code why that is.  I think a comment explaining the number would be helpful.

tenderlove: Ah, because only two shifts. 🤦

joshuay03: Yes two shifts and then just one needed for the `segments.any?` return value. I'll add a comment 👍🏽

lib/bundled_gems.rb
```diff
@@ -105,11 +105,14 @@ def self.warning?(name, specs: nil)
       # and `require "syslog"` to `require "#{ARCHDIR}/syslog.so"`.
       feature.delete_prefix!(ARCHDIR)
       feature.delete_prefix!(LIBDIR)
-      segments = feature.split("/")
+      # 1. A segment for the basic EXACT and SINCE check e.g. `logger`
+      # 2. A segment for the multi-part SINCE check e.g. `resolv-replace`
```
nobu: I don't think that  resolv-replace, which provides `resolv-replace.rb` as the name says, is the targets of this code.
The targets are net/* family, net-http gem for `require "net/http"`.

I suspect that this code can be removed since such libraries are no longer in `SINCE` table.

nobu: "This code" is joining the first two segments with `-`.

joshuay03: Thanks, I've removed that and updated the comments. I also had to remove a failing test for `net/smtp`. I think I can also remove this one? https://github.com/ruby/ruby/blob/eab4a0bc8d883be2071a090c87914efc9c12d10c/spec/bundled_gems_spec.rb#L347-L364

It passes, but it's probably a false positive now.

As well as the `SINCE` entry here? https://github.com/ruby/ruby/blob/eab4a0bc8d883be2071a090c87914efc9c12d10c/spec/bundled_gems_spec.rb#L66

joshuay03: I've gone ahead and removed them.

nobu: @hsbt What do you think? Or we should keep the feature?

hsbt: I don't want to remove these tests.

joshuay03: Okay I've put all of that back and kept the comments example-free, given the spec has good examples.

hsbt (approved):

deivid-rodriguez:
Thank you for restoring performance after my change @joshuay03 ❤️

lib/bundled_gems.rb
deivid-rodriguez: In my opinion, it's better to not keep code not necessary in real life, but just meant to make some specs pass, so I would remove the code and tests as suggested.

If we want to make sure we don't regress if gems with the special naming are ever added back to the `SINCE` list, we could make sure the specs cover the real `SINCE` list more thoroughly, and that they start failing if a gem with the special naming is added in the future.

PR #13729, opened 2025-06-27
ZJIT: `throw` to HIR

XrXr:
- **ZJIT: Use initialization shorthand**
- **ZJIT: `throw` to HIR**

No codegen for this for now since the calling convention is
somewhat in flux, and I'm not sure how best to do a unwind
with a value other than `Qundef`.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1072,7 +1072,7 @@ impl Function {
             ArraySet { array, idx, val } => ArraySet { array: find!(*array), idx: *idx, val: find!(*val) },
             ArrayDup { val , state } => ArrayDup { val: find!(*val), state: *state },
             &HashDup { val , state } => HashDup { val: find!(val), state },
-            &CCall { cfun, ref args, name, return_type, elidable } => CCall { cfun: cfun, args: find_vec!(args), name: name, return_type: return_type, elidable },
+            &CCall { cfun, ref args, name, return_type, elidable } => CCall { cfun, args: find_vec!(args), name, return_type, elidable },
             &Defined { op_type, obj, pushval, v } => Defined { op_type, obj, pushval, v: find!(v) },
             &DefinedIvar { self_val, pushval, id, state } => DefinedIvar { self_val: find!(self_val), pushval, id, state },
             NewArray { elements, state } => NewArray { elements: find_vec!(*elements), state: find!(*state) },
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -488,6 +488,7 @@ pub enum Insn {
 
     /// Control flow instructions
     Return { val: InsnId },
+    Throw { state: u32, val: InsnId },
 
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -488,6 +488,7 @@ pub enum Insn {
 
     /// Control flow instructions
     Return { val: InsnId },
+    Throw { state: u32, val: InsnId },
```
tekknolagi: Let's not use `state`; we're probably going to need that for side exits and that's our conventional use case for the Snapshot InsnId. Maybe `throw_state`?

(Do we need to have a FrameState for throw? It feels like yes but I don't actually know how this works)

zjit/src/hir.rs
```diff
@@ -1013,6 +1031,7 @@ impl Function {
                     }
                 },
             Return { val } => Return { val: find!(*val) },
+            &Throw { state, val } => Throw { state, val: find!(val) },
```
tekknolagi: note to self: we can probably make a smarter macro `find!(x)` that expands to `x: find!(x)`, etc

zjit/src/hir.rs
```diff
@@ -1117,7 +1136,7 @@ impl Function {
         match &self.insns[insn.0] {
             Insn::Param { .. } => unimplemented!("params should not be present in block.insns"),
             Insn::SetGlobal { .. } | Insn::ArraySet { .. } | Insn::Snapshot { .. } | Insn::Jump(_)
-            | Insn::IfTrue { .. } | Insn::IfFalse { .. } | Insn::Return { .. }
```
tekknolagi, line 1122: note to self: we should probably use `has_output` here

zjit/src/hir.rs
```diff
@@ -2696,6 +2716,10 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     fun.push_insn(block, Insn::Return { val: state.stack_pop()? });
                     break;  // Don't enqueue the next block as a successor
                 }
+                YARVINSN_throw => {
+                    fun.push_insn(block, Insn::Throw { state: get_arg(pc, 0).as_u32(), val: state.stack_pop()? });
+                    break;  // Don't enqueue the next block as a successor
```
tekknolagi, line 2736: note to self: we can probably extract this into `is_terminator` below? maybe?

tekknolagi (approved):
awesome

tekknolagi:
Since codegen for this (and others) is challenging, maybe we should be compiling hard-to-codegen instructions into explicit side-exits for now instead of aborting compilation. Though that does require that they have a FrameState

zjit/src/hir.rs
XrXr: We do need to have a FrameState to check interrupts. Return should technically do that as well, and it doesn't as of yet. Maybe there should be a dedicated CheckInterrupts, though.

tekknolagi: I think there should be a dedicated CheckInterrupts

PR #13731, opened 2025-06-27
Implements [Feature #21455] Add a block argument to Array#join

leoarnold:
I sometimes come across code like this where
the `Array#join` at the end can easily
be overlooked or stands out like a sore thumb:

```ruby
hex_string = string.bytes.map do |byte|
  format('%02X', byte)
end.join(' ')
```

It seems idiomatic and more succinct
to pass the block to `Array#join` directly:

```ruby
hex_string = string.bytes.join(' ') do |byte|
  format('%02X', byte)
end
```

### DIFF
--- a/array.c
+++ b/array.c
@@ -2925,6 +2925,8 @@ rb_ary_join(VALUE ary, VALUE sep)
     return result;
 }
 
+static VALUE rb_ary_collect(VALUE ary);
+
 /*
  *  call-seq:
  *    join(separator = $,) -> new_string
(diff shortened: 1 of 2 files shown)


jeremyevans:
If you would like this to be considered, please submit a feature request on https://bugs.ruby-lang.org/, explaining what you see as the costs and benefits of the feature, and linking to this pull request.

leoarnold:
https://bugs.ruby-lang.org/issues/21455

PR #13735, opened 2025-06-29
Add Set C-API

jeremyevans:
This should be a minimal C-API needed to deal with Set objects. It
supports creating the sets, checking whether an element is the set,
adding and removing elements, iterating over the elements, clearing
a set, and returning the size of the set.

As some of the function names I wanted to use in the C-API were
already used internally, I had to rename some existing internal functions.

### DIFF
--- a/include/ruby/intern.h
+++ b/include/ruby/intern.h
@@ -51,6 +51,7 @@
 #include "ruby/internal/intern/re.h"
 #include "ruby/internal/intern/ruby.h"
 #include "ruby/internal/intern/select.h"
+#include "ruby/internal/intern/set.h"
 #include "ruby/internal/intern/signal.h"
 #include "ruby/internal/intern/sprintf.h"
 #include "ruby/internal/intern/string.h"
--- a/include/ruby/internal/intern/set.h
+++ b/include/ruby/internal/intern/set.h
@@ -1,0 +1,111 @@
+#ifndef RBIMPL_INTERN_SET_H                         /*-*-C++-*-vi:se ft=cpp:*/
+#define RBIMPL_INTERN_SET_H
+/**
+ * @file
+ * @author     Ruby developers <ruby-core@ruby-lang.org>
+ * @copyright  This  file  is   a  part  of  the   programming  language  Ruby.
+ *             Permission  is hereby  granted,  to  either redistribute  and/or
+ *             modify this file, provided that  the conditions mentioned in the
+ *             file COPYING are met.  Consult the file for details.
+ * @warning    Symbols   prefixed  with   either  `RBIMPL`   or  `rbimpl`   are
+ *             implementation details.   Don't take  them as canon.  They could
+ *             rapidly appear then vanish.  The name (path) of this header file
+ *             is also an  implementation detail.  Do not expect  it to persist
+ *             at the place it is now.  Developers are free to move it anywhere
+ *             anytime at will.
+ * @note       To  ruby-core:  remember  that   this  header  can  be  possibly
+ *             recursively included  from extension  libraries written  in C++.
+ *             Do not  expect for  instance `__VA_ARGS__` is  always available.
+ *             We assume C99  for ruby itself but we don't  assume languages of
+ *             extension libraries.  They could be written in C++98.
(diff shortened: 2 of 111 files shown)


include/ruby/internal/intern/set.h
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Creates a new, empty set object.
+ *
+ * @return  An allocated new instance of ::rb_cSet.
+ */
+VALUE rb_set_new(void);
+
+/**
+ * Identical to rb_set_new(), except it additionally specifies how many elements
```
nobu:
```suggestion
 * Identical to rb_set_new(), except it additionally specifies how many elements
```

st0012:
Should `rb_cSet` also be exposed in `include/ruby/internal/globals.h`?
Having it exposed will also allow ZJIT's to optimize for Set objects more easily ([PR](https://github.com/ruby/ruby/pull/13743)).

jeremyevans:
> Should `rb_cSet` also be exposed in `include/ruby/internal/globals.h`? Having it exposed will also allow ZJIT's to optimize for Set objects more easily ([PR](https://github.com/ruby/ruby/pull/13743)).

Yes, it should.  I updated this to include that change.

set.c
```diff
@@ -1913,6 +1913,56 @@ compat_loader(VALUE self, VALUE a)
...
+}
+
+VALUE
+rb_set_new(void)
+{
+    return set_alloc_with_size(rb_cSet, 0);
+}
+
+VALUE
+rb_set_new_capa(unsigned long capa)
```
rhenium: Should this take a `size_t`, to match the type used in `rb_set_size()`?

st0012:
Headsup: I've exposed `rb_cSet` [in `zjit.c`](https://github.com/ruby/ruby/blob/master/zjit.c#L35) for now. So you'll need to remove it when rebasing this PR.

set.c
jeremyevans: The underlying function takes the size as `st_index_t`. I didn't want to have the public header depend on `st.h` types. I think either `size_t` or `unsigned long` would work. For the common platforms, `st_index_t` is `unsigned long` (I think it's `unsigned long long` on Windows).

rhenium: I think `rb_set_new_capa()` and `rb_set_size()` should use the same type. Avoiding `st_index_t` here seems like a good idea to me as well.

I think `size_t` would be simpler, since we seem to assume it has the same size as a pointer type (and happens to be compatible with `st_index_t`), whereas `unsigned long` can sometimes be smaller.

spec/ruby/optional/capi/set_spec.rb
```diff
@@ -0,0 +1,96 @@
+require_relative 'spec_helper'
+
+load_extension("set")
+
+describe "C-API Set function" do
```
eregon, line 5: Thank you for adding these C API specs, that will definitely help implementing those functions on other Rubies.

jeremyevans:
> Headsup: I've exposed `rb_cSet` [in `zjit.c`](https://github.com/ruby/ruby/blob/master/zjit.c#L35) for now. So you'll need to remove it when rebasing this PR.

Rebased and updated to remove it from `zjit.c`.

set.c
jeremyevans: Makes sense to me. When I rebased today, I included that change.

include/ruby/intern.h
```diff
@@ -51,6 +51,7 @@
 #include "ruby/internal/intern/re.h"
 #include "ruby/internal/intern/ruby.h"
 #include "ruby/internal/intern/select.h"
+#include "ruby/internal/intern/set.h"
```
nobu, line 54: May this be unnecessary?
Maybe it's enough to include "ruby/set.h" only when Set-API is used, as well as "ruby/io.h"?

jeremyevans: I agree that it isn't strictly necessary.  However, I'd like the Set C-API to be usable in a similar manner to the Array/Hash C-API, and I don't believe those require separate includes.  If you do not feel that is appropriate, please let me know and I'll submit a PR to switch to `ruby/set.h`

PR #13736, opened 2025-06-29
fix building issue when using gcc15, because C23 is default

clan:
otherwise build is failed with:

> error: static declaration of ‘rb_io_closed_p’ follows non-static declaration

### DIFF
--- a/ext/socket/rubysocket.h
+++ b/ext/socket/rubysocket.h
@@ -292,8 +292,8 @@ extern VALUE rb_eResolution;
 #ifdef SOCKS
 extern VALUE rb_cSOCKSSocket;
 #  ifndef SOCKS5
-void SOCKSinit();
-int Rconnect();
+void SOCKSinit(char *);
+int Rconnect(int, const struct sockaddr *, socklen_t);
 #  endif
 #endif
 


nobu:
Please send the patches for [io-console] and [strscan] to each upstreams.

[io-console]: https://github.com/ruby/io-console
[strscan]: https://github.com/ruby/strscan

nobu:
And a typo in the last commit log,  socketc should be socket, I guess.

clan:
> Please send the patches for [io-console](https://github.com/ruby/io-console) and [strscan](https://github.com/ruby/strscan) to each upstreams.

ah, i think this is the reason why these two functions (maybe more) are implemented again?

nobu:
I don't think this fix is the best.
Probably it can be fixed by passing the header name to `have_func` calls, such ash `have_func("rb_io_path", "ruby/io.h")`.

nobu:
Anyway, this may need backports to stable branches.
Could you file an issue at the [BTS](https://bugs.ruby-lang.org)?

clan:
> Anyway, this may need backports to stable branches. Could you file an issue at the [BTS](https://bugs.ruby-lang.org)?

ok, problem on ext/strscan had been fixed in commit 4585ccd90f4251f4d42bfc338a5e14100236fa15, backport is needed.

will file an issue when free.

clan:
https://bugs.ruby-lang.org/issues/21497

clan:
io-console had been fixed in https://github.com/ruby/io-console/commit/dd013030dd276a7372df34cf43ada1c14d0cbc21

clan:
does ext/socket has upstream?

nobu:
No, `socket` is bound to ruby core itself deeply and a few libraries including it are needed to work rubygems.

PR #13743, opened 2025-06-30
ZJIT: Add new ZJIT types for Set

st0012:
This allows ZJIT to identify Set objects and provide special optimization for them later on.

One issue is that `rb_cSet` is not currently exposed yet, which I also [proposed](https://github.com/ruby/ruby/pull/13735#issuecomment-3018773453) in [Jeremy's PR for `Set`'s C-API](https://github.com/ruby/ruby/pull/13735).

### DIFF
--- a/zjit/src/hir_type/mod.rs
+++ b/zjit/src/hir_type/mod.rs
@@ -1,6 +1,6 @@
 #![allow(non_upper_case_globals)]
 use crate::cruby::{Qfalse, Qnil, Qtrue, VALUE, RUBY_T_ARRAY, RUBY_T_STRING, RUBY_T_HASH};
-use crate::cruby::{rb_cInteger, rb_cFloat, rb_cArray, rb_cHash, rb_cString, rb_cSymbol, rb_cObject, rb_cTrueClass, rb_cFalseClass, rb_cNilClass, rb_cRange};
+use crate::cruby::{rb_cInteger, rb_cFloat, rb_cArray, rb_cHash, rb_cString, rb_cSymbol, rb_cObject, rb_cTrueClass, rb_cFalseClass, rb_cNilClass, rb_cRange, rb_cSet};
 use crate::cruby::ClassRelationship;
 use crate::cruby::get_class_name;
 use crate::cruby::rb_mRubyVMFrozenCore;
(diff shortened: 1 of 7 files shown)


include/ruby/internal/globals.h
```diff
@@ -93,6 +93,7 @@ RUBY_EXTERN VALUE rb_cRandom;                 /**< `Random` class. */
 RUBY_EXTERN VALUE rb_cRange;                  /**< `Range` class. */
 RUBY_EXTERN VALUE rb_cRational;               /**< `Rational` class. */
 RUBY_EXTERN VALUE rb_cRegexp;                 /**< `Regexp` class. */
+RUBY_EXTERN VALUE rb_cSet;                    /**< `Set` class. */
```
tekknolagi: To avoid poking too much at Jeremy's PR too much, you can do the usual ZJIT approach and do this in `zjit.c`. I think that should work?

zjit/src/hir_type/mod.rs
```diff
@@ -143,6 +143,10 @@ fn is_range_exact(val: VALUE) -> bool {
     val.class_of() == unsafe { rb_cRange }
 }
 
+fn is_set_exact(val: VALUE) -> bool {
```
tekknolagi: Please inline this; the other ones are only outlined because of the prism GC-hiding weirdness

zjit/src/hir_type/mod.rs
```diff
@@ -585,6 +593,19 @@ mod tests {
...
     }
 
+    #[test]
+    fn set() {
+        assert_subtype(types::SetExact, types::Set);
+        assert_subtype(types::SetSubclass, types::Set);
+    }
+
+    #[test]
+    fn set_has_ruby_class() {
```
tekknolagi, line 599: This should probably use the with_vm thing because otherwise we're messing with uninitialized pointers. And even though we're not dereferencing them, it makes me uneasy

test/ruby/test_zjit.rb
```diff
@@ -843,6 +843,16 @@ def test_instruction_order
     end
   end
 
+  def test_set
+    assert_compiles '3', %q{
+      def test_set
+        s = Set.new([1, 2, 3])
```
tekknolagi: This isn't testing anything in your PR, really, at least until you add cruby_methods stuff. And even then you will probably want to do the constants trick because the JIT can't (yet) see through Class#new

tekknolagi (approved):
lgtm pending comments

you can do cruby_methods in the next PR if you like, or stack a commit on top

st0012:
I think none of the test failures are related and this is ready for merge 🙏  @ruby/jit

PR #13744, opened 2025-06-30
ZJIT: setglobal should not return output

k0kubun:
This PR fixes a crash in btest `test_class.rb`. 

`gen_setglobal()` returns an output, so `Insn::SetGlobal` should be considered to have output to avoid assertion failures.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -278,7 +278,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::PatchPoint(_) => return Some(()), // For now, rb_zjit_bop_redefined() panics. TODO: leave a patch point and fix rb_zjit_bop_redefined()
         Insn::CCall { cfun, args, name: _, return_type: _, elidable: _ } => gen_ccall(jit, asm, *cfun, args)?,
         Insn::GetIvar { self_val, id, state: _ } => gen_getivar(asm, opnd!(self_val), *id),
-        Insn::SetGlobal { id, val, state: _ } => return gen_setglobal(asm, *id, opnd!(val)),
+        Insn::SetGlobal { id, val, state: _ } => return Some(gen_setglobal(asm, *id, opnd!(val))),
         Insn::GetGlobal { id, state: _ } => gen_getglobal(asm, *id),
         &Insn::GetLocal { ep_offset, level } => gen_nested_getlocal(asm, ep_offset, level)?,
         Insn::SetLocal { val, ep_offset, level } => return gen_nested_setlocal(asm, opnd!(val), *ep_offset, *level),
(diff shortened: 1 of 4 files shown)


tekknolagi:
I don't think setglobal has an output -- our SetGlobal pretty neatly maps to setglobal in insns.def, which has no stack output. This may be related to some neighboring instruction though

zjit/src/codegen.rs
```diff
@@ -451,12 +451,10 @@ fn gen_getglobal(asm: &mut Assembler, id: ID) -> Opnd {
...
 /// Set global variables
-fn gen_setglobal(asm: &mut Assembler, id: ID, val: Opnd) -> Opnd {
+fn gen_setglobal(asm: &mut Assembler, id: ID, val: Opnd) -> Option<()> {
     asm_comment!(asm, "call rb_gvar_set");
-    asm.ccall(
-        rb_gvar_set as *const u8,
-        vec![Opnd::UImm(id.0), val],
-    )
+    asm.ccall(rb_gvar_set as *const u8, vec![Opnd::UImm(id.0), val]);
+    Some(())
```
tekknolagi: Given that this does not have any failure paths, we can either return nothing or `()` here and handle (all of?) the type in the caller

tekknolagi (approved):
lgtm with PR title change

zjit/src/codegen.rs
k0kubun: Ah right, I forgot that's what I did for `gen_return`. Fixed in https://github.com/ruby/ruby/pull/13744/commits/deafdea0cc5f88ecb77822998d70d5a99c5acade

zjit/src/codegen.rs
```diff
@@ -278,7 +278,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::PatchPoint(_) => return Some(()), // For now, rb_zjit_bop_redefined() panics. TODO: leave a patch point and fix rb_zjit_bop_redefined()
         Insn::CCall { cfun, args, name: _, return_type: _, elidable: _ } => gen_ccall(jit, asm, *cfun, args)?,
         Insn::GetIvar { self_val, id, state: _ } => gen_getivar(asm, opnd!(self_val), *id),
-        Insn::SetGlobal { id, val, state: _ } => return gen_setglobal(asm, *id, opnd!(val)),
+        Insn::SetGlobal { id, val, state: _ } => return Some(gen_setglobal(asm, *id, opnd!(val))),
```
tekknolagi, line 281: This might have to be `{ gen_setglobal(); return Some(()) }`

k0kubun: May I ask why? We already do this for `gen_return()`.

k0kubun: Well `gen_return()` was actually doing something different, but this still compiles correctly in my understanding.

tekknolagi: In this case because you totally nuked the return type of gen_setglobal

tekknolagi: Oh, wow, I guess I did not understand that that was (). Never mind

PR #13745, opened 2025-06-30
ZJIT: IsNil optimization and codegen support

ywenc:
This PR adds optimizations and codegen support for `IsNil`. 

Currently:
```
./miniruby --zjit-debug -e "def test = 1&.itself; test; test"
ZJIT: gen_function: unexpected insn IsNil v15
Failed to compile insn: v5 IsNil v15
```

---

Closes: https://github.com/Shopify/ruby/issues/600

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -831,6 +831,15 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 2
   end
 
+  def test_branchnil
+    assert_compiles '[2, nil]', %q{
+      def test(x)
+        x&.succ
+      end
+      [test(1), test(nil)]
+    }, call_threshold: 1, insns: [:branchnil]
+  end
+
   # tool/ruby_vm/views/*.erb relies on the zjit instructions a) being contiguous and
   # b) being reliably ordered after all the other instructions.
   def test_instruction_order
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -272,6 +272,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::FixnumLe { left, right } => gen_fixnum_le(asm, opnd!(left), opnd!(right))?,
         Insn::FixnumGt { left, right } => gen_fixnum_gt(asm, opnd!(left), opnd!(right))?,
         Insn::FixnumGe { left, right } => gen_fixnum_ge(asm, opnd!(left), opnd!(right))?,
+        Insn::IsNil { val } => gen_isnil(asm, opnd!(val))?,
         Insn::Test { val } => gen_test(asm, opnd!(val))?,
         Insn::GuardType { val, guard_type, state } => gen_guard_type(jit, asm, opnd!(val), *guard_type, &function.frame_state(*state))?,
         Insn::GuardBitEquals { val, expected, state } => gen_guard_bit_equals(jit, asm, opnd!(val), *expected, &function.frame_state(*state))?,
@@ -888,6 +889,13 @@ fn gen_fixnum_gt(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd) -> Opti
 fn gen_fixnum_ge(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd) -> Option<lir::Opnd> {
     asm.cmp(left, right);
     Some(asm.csel_ge(Qtrue.into(), Qfalse.into()))
+}
+
+// Compile val == nil
+fn gen_isnil(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
+    asm.cmp(val, Qnil.into());
+    // TODO: Implement and use setcc
+    Some(asm.csel_e(Opnd::Imm(1), Opnd::Imm(0)))
 }
 
 fn gen_anytostring(asm: &mut Assembler, val: lir::Opnd, str: lir::Opnd, state: &FrameState) -> Option<lir::Opnd> {
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -571,6 +571,7 @@ impl Insn {
             Insn::FixnumGt   { .. } => false,
             Insn::FixnumGe   { .. } => false,
             Insn::GetLocal   { .. } => false,
+            Insn::IsNil      { .. } => false,
             Insn::CCall { elidable, .. } => !elidable,
             _ => true,
         }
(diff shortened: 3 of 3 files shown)


tekknolagi:
Optimizations should not be required to run for correctness, so this kind of papers over the core problem of not having codegen for IsNil (which is a fine thing to do for now but is not the general solution). Would you like to also add codegen for IsNil? https://github.com/Shopify/ruby/issues/600

zjit/src/hir.rs
```diff
@@ -1440,6 +1440,17 @@ impl Function {
                         let replacement = self.push_insn(block, Insn::Const { val: Const::Value(unsafe { (*ice).value }) });
                         self.make_equal_to(insn_id, replacement);
                     }
+                    Insn::IsNil { val } => {
```
tekknolagi: Nice

zjit/src/hir.rs
```diff
@@ -6145,6 +6193,27 @@ mod opt_tests {
...
+    #[test]
+    fn test_opt_isnil_pattern_matching() {
+        eval("def test = ([] in [])");
+
+        assert_optimized_method_hir("test",  expect![[r#"
+            fn test:
+            bb0(v0:BasicObject):
+              v2:NilClassExact = Const Value(nil)
+              v4:ArrayExact = NewArray
+              v15:StaticSymbol[VALUE(0x1000)] = Const Value(VALUE(0x1000))
```
tekknolagi: Not for this PR, but (note to self) we should print the symbol value here for debugging in the future

tekknolagi:
Please title PRs "ZJIT: ..."

zjit/src/hir.rs
tekknolagi: #13746

ywenc:
> Would you like to also add codegen for IsNil?

~I'll do that! I will put it on a follow up PR if that sounds good, since not sure if IsNil can get complicated because it's generated as a part of `branchnil` and is not an actual yarv instruction~

The codegen is part of this same PR

tekknolagi:
I think it's cmp and cmov, which should not be too tricky

EDIT: or setcc, if we have that

test/ruby/test_zjit.rb
```diff
@@ -830,6 +830,17 @@ def test = Foo
...
 
+  def test_branchnil
+    assert_compiles '1', %q{
+      def test(x)
+        x.to_i&.itself
+      end
+
+      test(1)
+      test(1)
+    }, call_threshold: 2
```
tekknolagi:
```suggestion
    }, call_threshold: 2, insns: [:branchnil]
```

test/ruby/test_zjit.rb
```diff
@@ -830,6 +830,17 @@ def test = Foo
     }, call_threshold: 2
   end
 
+  def test_branchnil
```
tekknolagi, line 834: This should be split into two tests: one with `call_threshold:1` and one with `:2`. Then you should test true and false for each. This will test both IsNil (not optimized) on truthy/falsey and a guarded type that will always return true/false on truthy/falsey

zjit/src/codegen.rs
```diff
@@ -890,6 +891,12 @@ fn gen_fixnum_ge(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd) -> Opti
     Some(asm.csel_ge(Qtrue.into(), Qfalse.into()))
 }
 
+// Compile val == nil
+fn gen_isnil(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
+    asm.cmp(val, Qnil.into());
+    Some(asm.csel_e(Opnd::Imm(1),Opnd::Imm(0)))
+}
```
```suggestion
// Compile val == nil
fn gen_isnil(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
    asm.cmp(val, Qnil.into());
    // TODO: Implement and uset setcc
    Some(asm.csel_e(Opnd::Imm(1), Opnd::Imm(0)))
}
```

zjit/src/hir.rs
tekknolagi: Actually, I think the bigger problem here is not that it's not getting expressly optimized out but that `IsNil.has_effects()` is returning `true`. Can you revert this and add an entry to `has_effects` instead please? We already have this logic in `infer_type`

zjit/src/hir.rs
```diff
@@ -6145,6 +6183,27 @@ mod opt_tests {
         "#]]);
     }
 
+    #[test]
+    fn test_opt_isnil_pattern_matching() {
+        eval("def test = ([] in [])");
+
+        assert_optimized_method_hir("test",  expect![[r#"
```
```suggestion
        assert_optimized_method_hir("test", expect![[r#"
```

nitpick: You seem to like placing two spaces here, but it seems inconsistent with the style of other test code.

zjit/src/hir.rs
```diff
@@ -4526,6 +4527,43 @@ mod tests {
         "#]]);
     }
 
+    #[test]
+    fn test_branchnil_pattern_matching() {
+        eval("def test = ([] in [])");
+
+        assert_method_hir_with_opcode("test",  YARVINSN_branchnil, expect![[r#"
```
```suggestion
        assert_method_hir_with_opcode("test", YARVINSN_branchnil, expect![[r#"
```

ditto

zjit/src/codegen.rs
k0kubun: > // TODO: Implement and uset setcc

You mean `s/uset/use/`?

I think it's the backend's job to choose `setcc` instruction for this LIR though. We shouldn't need to change this codegen. Maybe clarify that in the comment too?

ywenc:
After some feedback conversation with @tekknolagi , I've changed the tests/direction of the PR to be more focused on IsNil, and use safe navigation for a simpler example, and will leave pattern matching for a future PR.

zjit/src/hir.rs
ywenc: ah so sorry! I must have kept copying that over from my other tests, I'll be more mindful in the future, should no longer be on this branch now.

test/ruby/test_zjit.rb
```diff
@@ -830,6 +830,16 @@ def test = Foo
...
+    assert_compiles '1', %q{
+      def test(x)
+        x.itself&.itself
+      end
+
+      test(1)
+    }, call_threshold: 1, insns: [:branchnil]
+  end
+
```
tekknolagi:
```suggestion
  def test_branchnil
    assert_compiles '[2, nil]', %q{
      def test(x)
        x&.succ
      end

      [test(1), test(nil)]
    }, call_threshold: 1, insns: [:branchnil]
  end

```

Out of curiosity why two calls vs `x&.itself`?

zjit/src/codegen.rs
```diff
@@ -890,6 +891,13 @@ fn gen_fixnum_ge(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd) -> Opti
     Some(asm.csel_ge(Qtrue.into(), Qfalse.into()))
 }
 
+// Compile val == nil
+fn gen_isnil(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
+    asm.cmp(val, Qnil.into());
+    // TODO: Implement and use setcc
+    Some(asm.csel_e(Opnd::Imm(1),Opnd::Imm(0)))
```
```suggestion
    Some(asm.csel_e(Opnd::Imm(1), Opnd::Imm(0)))
```

tekknolagi (approved):
nice

zjit/src/hir.rs
```diff
@@ -6176,6 +6177,42 @@ mod opt_tests {
         "#]]);
     }
 
+    #[test]
+    fn test_branchnil_nil() {
+        eval("
+            def test
+                x = nil
+                x&.itself
```
```suggestion
              x = nil
              x&.itself
```

I know it's complicated, but we don't indent Ruby code with 4 spaces while we indent the Rust code (like this string itself) with 4 spaces.

zjit/src/hir.rs
```diff
@@ -6176,6 +6177,42 @@ mod opt_tests {
...
+              Return v3
+        "#]]);
+    }
+
+    #[test]
+    fn test_branchnil_truthy() {
+        eval("
+            def test
+                x = 1
+                x&.itself
```
```suggestion
              x = 1
              x&.itself
```

ditto

test/ruby/test_zjit.rb
ywenc: ah I was setting up something that will be a `BasicObject` to go before the `&.`, but I think we already get this by passing x into test. I took your suggestion here, thank you!

k0kubun (approved):

PR #13746, opened 2025-06-30
ZJIT: Pretty-print symbols in HIR dump

tekknolagi:
This lets us better see what is going on, for example in pattern
matching code, which has a bunch of dynamic method lookups and
`respond_to?` sends.

### DIFF
--- a/zjit/src/cruby.rs
+++ b/zjit/src/cruby.rs
@@ -738,6 +738,11 @@ fn ruby_str_to_rust(v: VALUE) -> String {
     }
 }
 
+pub fn ruby_sym_to_rust(v: VALUE) -> String {
+    let ruby_str = unsafe { rb_sym2str(v) };
+    ruby_str_to_rust(ruby_str)
+}
+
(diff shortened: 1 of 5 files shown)


zjit/src/cruby.rs
```diff
@@ -738,6 +738,11 @@ fn ruby_str_to_rust(v: VALUE) -> String {
     }
 }
 
+pub fn ruby_sym_to_rust(v: VALUE) -> String {
```
k0kubun:
```suggestion
pub fn ruby_sym_to_rust_string(v: VALUE) -> String {
```

It's not the Rust itself, so the name feels a bit weird. Given that we have `cstr_to_rust_string`, maybe we should call it `rust_string`? Same applies to `ruby_str_to_rust` above this.

If that's too long, I'd just call it `string` instead of `rust_string`.

tekknolagi: I went for matching the function immediately above but I agree that both names need improvement

PR #13747, opened 2025-06-30
[DOC] Tweaks for Case Mapping doc

BurdetteLamar:
I don't think we need to mention the unimplemented argument value `:lithuanian`.  Omitting it means that instead of arguments `*options`, we can have one argument, `mapping`.

Marking this as Draft.  If agreed upon, I'll add commits changing the argument name `*options` to `mapping` in the dozen or so affected methods' calling sequences.

### DIFF
--- a/doc/case_mapping.rdoc
+++ b/doc/case_mapping.rdoc
@@ -37,7 +37,7 @@ Context-dependent case mapping as described in
 {Table 3-17 (Context Specification for Casing) of the Unicode standard}[https://www.unicode.org/versions/latest/ch03.pdf]
 is currently not supported.
 
-In most cases, case conversions of a string have the same number of characters.
+In most cases, the case conversion of a string has the same number of characters as before.
 There are exceptions (see also +:fold+ below):
 
   s = "\u00DF" # => "ß"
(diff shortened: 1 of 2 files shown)


BurdetteLamar:
> I don't think we need to mention the unimplemented argument value `:lithuanian`. Omitting it means that instead of arguments `*options`, we can have one argument, `mapping`.
> 
> Marking this as Draft. If agreed upon, I'll add commits changing the argument name `*options` to `mapping` in the dozen or so affected methods' calling sequences.

@peterzhu2118, if you like this, I'll make the additional changes.

peterzhu2118:
Yeah I like this 👍

BurdetteLamar:
@peterzhu2118, in `string.c` I have modified _only_ the argument name `*options`, changing it to `mapping`;  any further tweaks will come later.

BurdetteLamar:
@peterzhu2118, it looks like mapping `:lithuanian` may be implemented after all (see `check_case_options` in string .c).  Should we revise this documentation?

peterzhu2118:
It looks like `:lithuanian` is "supported" as in it does not raise an error, but doesn't actually do anything. I'm not sure what is missing from the `:lithuanian` implementation. Maybe @duerst could give an example?

PR #13749, opened 2025-06-30
ZJIT: Don't compile functions with unhandled parameter types

tekknolagi:
Follow-up on #13633.


tekknolagi:
Aside from optionals, what other kinds should we filter here?

tekknolagi:
This might help https://github.com/Shopify/ruby/issues/608

tekknolagi:
We really need to fix the thing where we fail an entire connected component of JIT calls if a leaf can't be compiled

EDIT: https://github.com/Shopify/ruby/issues/613

PR #13750, opened 2025-06-30
ZJIT: Mark GetLocal as having no effects

tekknolagi:
This removes the GetLocal of l3 from:

    def test
      l3 = 3
      1.times do |l2|
        _ = l3
        1
      end
    end

PR #13752, opened 2025-06-30
ZJIT: Stop tracking EP == BP assumption on JIT entry

k0kubun:


tekknolagi:
I'm unfamiliar with this particular assumption. When might EP not be BP? Is this a global effect or scoped to a frame? An ISEQ? Something else? Is this recognizable statically in HIR or do we have to track it happening somewhere in the runtime with a hook?

.github/workflows/zjit-macos.yml
```diff
@@ -128,17 +129,16 @@ jobs:
...
           ../src/bootstraptest/test_syntax.rb \
+          ../src/bootstraptest/test_thread.rb \
           ../src/bootstraptest/test_yjit_30k_ifelse.rb \
           ../src/bootstraptest/test_yjit_30k_methods.rb \
           ../src/bootstraptest/test_yjit_rust_port.rb
         # ../src/bootstraptest/test_eval.rb \
-        # ../src/bootstraptest/test_gc.rb \
         # ../src/bootstraptest/test_insns.rb \
         # ../src/bootstraptest/test_massign.rb \
         # ../src/bootstraptest/test_method.rb \
```
st0012: I think this is fixed by the PR too.

k0kubun:
> When might EP not be BP?

When an environment is escaped to the heap. As written in the comment, `<main>` and `eval` starts with EP pointing to the heap. After entry, methods like `proc {}` or `binding` can escape stack-local EP to the heap afterward.

> Is this a global effect or scoped to a frame? An ISEQ? Something else? 

Scoped to a frame. An ISEQ type (`<main>`/`eval` or not) tells you whether EP should point to the heap from the beginning or not. 

> Is this recognizable statically in HIR or do we have to track it happening somewhere in the runtime with a hook?

Remember https://github.com/ruby/ruby/pull/13750#issuecomment-3020589350, this is where I suggested any `method_call` could escape EP of the caller frame (the block given to `Numeric#times`). Because Ruby allows escaping the EP of any frame in the call stack at any time, yes, we'd have to track it happening somewhere in the runtime with a hook.

Unless we recursively inline every method call, including ones called inside C functions, it's impossible to statically track every arbitrary method call that may happen under the current frame at compile time.

.github/workflows/zjit-macos.yml
k0kubun: :+1: https://github.com/ruby/ruby/pull/13752/commits/e54a4accf06bc35c81f51536ec9925458866c5ec

st0012 (approved):

tekknolagi (approved):
nice, thanks, i understand now

PR #13760, opened 2025-07-01
ZJIT: Support `Regexp` type

st0012:


zjit/src/hir_type/mod.rs
```diff
@@ -197,6 +197,9 @@ impl Type {
         else if is_string_exact(val) {
             Type { bits: bits::StringExact, spec: Specialization::Object(val) }
         }
+        else if val.class_of() == unsafe { rb_cRegexp } {
```
st0012, line 200: Prism doesn't hide Regexp literals with `rb_obj_hide` like it does for Hash. So unless there are other ways to hide an object's class, Regexp objs don't need the special class check.

zjit/src/hir.rs
```diff
@@ -6310,4 +6323,21 @@ mod opt_tests {
               Return v7
         "#]]);
     }
+
```
tekknolagi, line 6313: can you also add a test like

```ruby
x = /hello/i.casefold?
1
```

so we can see that the call gets elided?

zjit/src/hir.rs
```diff
@@ -6310,4 +6323,21 @@ mod opt_tests {
...
+    #[test]
+    fn test_regexp_casefold() {
+        eval("
+            def test = /hello/i.casefold?
+            test
+            test
```
tekknolagi: you shouldn't have to call it at all here; you are working with a regexp literal

```suggestion
            def test = /hello/i.casefold?
```

zjit/src/hir.rs
```diff
@@ -6310,4 +6323,21 @@ mod opt_tests {
               Return v7
         "#]]);
     }
+
+    #[test]
+    fn test_regexp_casefold() {
```
```suggestion
    fn test_regexp_casefold_specialized_to_ccall() {
```

tekknolagi (approved):
more or less looks good *but* i am not sure it is always a leaf: there is a check there, `rb_reg_check`, that may raise if the regexp is uninitialized. in what circumstances might a regexp be uninitialized?

zjit/src/hir.rs
tekknolagi: (unless we determine it's not leaf)

st0012:
Hmm TIL you can `Regexp.allocate` an uninitialized `Regexp` object. And calling any methods, including `casefold?` on it, would fail `rb_reg_check`.

```
Regexp.allocate.casefold?
(irb):2:in 'Regexp#casefold?': uninitialized Regexp (TypeError)
```

I'll just drop the second commit for now.

zjit/src/hir.rs
st0012: Yeah it's not a leaf. Removed the `casefold?` commit.

zjit/src/hir.rs
```diff
@@ -6310,4 +6310,18 @@ mod opt_tests {
               Return v7
         "#]]);
     }
+
+    #[test]
+    fn test_regexp_type() {
+        eval("
+            def test = /a/
+            test
```
```suggestion
```

tekknolagi (approved):
noice

PR #13761, opened 2025-07-01
ZJIT: Support spilling basic block arguments

k0kubun:
This PR fixes the `register spilling not yet implemented, too many basic block arguments` panic.

It bumps the C stack pointer on the first basic block to allocate C stack slots for spilled basic block arguments, lets `param_opnd` pick a slot out of the allocated region, and restores the C stack pointer on `Return` instructions or side exits.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -576,7 +576,10 @@ fn gen_const(val: VALUE) -> lir::Opnd {
 
 /// Compile a basic block argument
 fn gen_param(asm: &mut Assembler, idx: usize) -> lir::Opnd {
+    // Allocate a register or a stack slot
     match param_opnd(idx) {
+        // If it's a register, insert LiveReg instruction to reserve the register
+        // in the register pool for register allocation.
         param @ Opnd::Reg(_) => asm.live_reg_opnd(param),
         param => param,
     }


k0kubun:
`test_too_many_basic_block_args` is passing on my x86_64-linux machine, but it's failing on both CIs. Maybe not happy with the alignment? I'll convert this to a draft PR until I figure it out.

k0kubun:
Fixed it at https://github.com/ruby/ruby/pull/13761/commits/acb2d2becbf91264781e01cc3b0a221950862906.

test/ruby/test_zjit.rb
```diff
@@ -693,6 +693,40 @@ def fib(n)
     }, call_threshold: 5, num_profiles: 3
   end
 
+  def test_too_many_basic_block_args
```
tekknolagi: Nit: "too many" is not really what's happening here---we can handle it---so a more descriptive name might say something about "more than fit in the physical register pool"

zjit/src/codegen.rs
```diff
@@ -992,17 +1022,15 @@ fn gen_push_frame(asm: &mut Assembler, argc: usize, state: &FrameState, frame: C
     asm.mov(cfp_opnd(RUBY_OFFSET_CFP_BLOCK_CODE), 0.into());
 }
 
-/// Return a register we use for the basic block argument at a given index
-fn param_reg(idx: usize) -> Reg {
+/// Return an operand we use for the basic block argument at a given index
+fn param_opnd(idx: usize) -> Opnd {
     // To simplify the implementation, allocate a fixed register for each basic block argument for now.
```
tekknolagi: "...or stack slot"

zjit/src/codegen.rs
```diff
@@ -992,17 +1022,15 @@ fn gen_push_frame(asm: &mut Assembler, argc: usize, state: &FrameState, frame: C
...
     // TODO: Allow allocating arbitrary registers for basic block arguments
-    if idx >= ALLOC_REGS.len() {
-        unimplemented!(
-            "register spilling not yet implemented, too many basic block arguments ({}/{})",
-            idx + 1, ALLOC_REGS.len()
-        );
+    if idx < ALLOC_REGS.len() {
+        Opnd::Reg(ALLOC_REGS[idx])
+    } else {
+        Opnd::mem(64, NATIVE_STACK_PTR, -((idx - ALLOC_REGS.len() + 1) as i32) * SIZEOF_VALUE_I32)
```
tekknolagi, line 1036: Does this need to deal with the stack alignment stuff too?

zjit/src/codegen.rs
```diff
@@ -555,7 +575,10 @@ fn gen_const(val: VALUE) -> lir::Opnd {
 
 /// Compile a basic block argument
 fn gen_param(asm: &mut Assembler, idx: usize) -> lir::Opnd {
-    asm.live_reg_opnd(Opnd::Reg(param_reg(idx)))
+    match param_opnd(idx) {
+        param @ Opnd::Reg(_) => asm.live_reg_opnd(param),
```
tekknolagi, line 583: Can you please add a comment about what's happening here?

zjit/src/codegen.rs
```diff
@@ -1063,6 +1092,28 @@ fn iseq_entry_escapes_ep(iseq: IseqPtr) -> bool {
...
+    let reverse_post_order = function.rpo();
+    reverse_post_order.iter().map(|&block_id| {
+        let block = function.block(block_id);
+        block.params().len()
+    }).max().unwrap_or(0)
+}
+
+/// Given the number of spill slots needed for a function, return the number of bytes
+/// the function needs to allocate on the stack for the stack frame.
+pub fn aligned_stack_bytes(size: usize) -> usize {
```
tekknolagi: I think my comment disappeared but I think `size` here is more accurately `count` or `num_slots` and only becomes `size` once it gets multiplied by `SIZEOF_VALUE`. Same with `c_stack_size`. It's more like `max_num_block_args`

k0kubun: > I think my comment disappeared

because you commented on a commit https://github.com/ruby/ruby/commit/acb2d2becbf91264781e01cc3b0a221950862906#r161284414 , not on the PR 🙂 

> size here is more accurately count or num_slots and only becomes size once it gets multiplied by SIZEOF_VALUE

Renamed it to `num_slots` https://github.com/ruby/ruby/pull/13761/commits/2bcee40e030786d5f41374d097ea8d5611dbd772

> Same with c_stack_size. It's more like max_num_block_args

I get what you mean, but it's unfortunately not equal to the max number of block args (because we subtract it by the number of registers). I did https://github.com/ruby/ruby/pull/13761/commits/34691a0852da2689a90c0f4fab005883b9aedce8 instead.

test/ruby/test_zjit.rb
k0kubun: ok, "spilled" from the register pool https://github.com/ruby/ruby/pull/13761/commits/b8e69f428d7157733dc4c809d3133d9536604294

zjit/src/codegen.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/13761/commits/d2a9baaeef28eb891ecb89e1d439625551fba042

zjit/src/codegen.rs
k0kubun: done https://github.com/ruby/ruby/pull/13761/commits/5c40b83e48777e6059813138926248285c36cd04

zjit/src/codegen.rs
k0kubun: I don't think so. This only refers to a stack slot that is already allocated by the stack pointer bump, so what we do here should not cause the stack pointer to be unaligned. Reading bytes from an 8-byte-aligned address seems fine, and which slot we leave unused shouldn't matter either.

PR #13762, opened 2025-07-01
ZJIT: Avoid panicing with "Option::unwrap() on None"

k0kubun:
This PR attempts to provide a better error message than `called Option::unwrap() on a None value` for register allocator-related crashes.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1523,7 +1523,12 @@ impl Assembler
             // Convert live_ranges to live_regs: the number of live registers at each index
             let mut live_regs: Vec<usize> = vec![];
             for insn_idx in 0..insns.len() {
-                let live_count = live_ranges.iter().filter(|range| range.start() <= insn_idx && insn_idx <= range.end()).count();
+                let live_count = live_ranges.iter().filter(|range|
+                    match (range.start, range.end) {
+                        (Some(start), Some(end)) => start <= insn_idx && insn_idx <= end,
+                        _ => false,
+                    }
+                ).count();
                 live_regs.push(live_count);
             }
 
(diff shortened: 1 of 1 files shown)


zjit/src/backend/lir.rs
```diff
@@ -1523,7 +1523,12 @@ impl Assembler
             // Convert live_ranges to live_regs: the number of live registers at each index
             let mut live_regs: Vec<usize> = vec![];
             for insn_idx in 0..insns.len() {
-                let live_count = live_ranges.iter().filter(|range| range.start() <= insn_idx && insn_idx <= range.end()).count();
+                let live_count = live_ranges.iter().filter(|range|
```
st0012, line 1526: Dumb Q: what makes a register live/non-live (dead?)

k0kubun: The range becomes `(None, None)` when a VReg was previously allocated/used but deleted in a later pass.

For example, when codegen generates:

```
    Add(X86Reg { num_bits: 64, reg_type: GP, reg_no: 13 }, 38_u64) -> VReg64(2),
    Mov(X86Reg { num_bits: 64, reg_type: GP, reg_no: 13 }, VReg64(2)) -> None,
```

`x86_split` merges these two instructions into:

```
Add(X86Reg { num_bits: 64, reg_type: GP, reg_no: 13 }, 38_u64) -> X86Reg { num_bits: 64, reg_type: GP, reg_no: 13 },
```

Here, we no longer use `VReg64(2)`, so nobody writes to the `start` and `end` of `live_ranges[2]` in the Assembler created by `x86_split`. 

We didn't have this problem when we were using the instruction index as the index to virtual registers (the live range for the removed instruction is automatically gone). But shifting indexes for all virtual registers and live ranges whenever we insert/remove instructions was so hard that it caused a lot of other troubles and we stopped doing that at ZJIT https://github.com/Shopify/zjit/pull/86. So, unlike YJIT, ZJIT may have a live range for a deleted VReg.

XrXr: huh this index maintenance problem sounds like what Union-Find solves. I've no idea if it actually makes things easier here, though.

PR #13764, opened 2025-07-01
ZJIT: Annotate `NilClass#nil?` and `Kernel#nil?`

st0012:
These methods return fixed `true` or `false` so we can be certain about their return types.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -840,6 +840,25 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 1, insns: [:branchnil]
   end
 
+  def test_nil_nil
+    assert_compiles 'true', %q{
+      def test = nil.nil?
+      test
+      test
+    }
+  end
+
+  def test_obsolete_nil_nil
+    assert_compiles '1', %q{
+      def test
+        nil.nil?
+        1
+      end
+      test
+      test
+    }
+  end
+
   # tool/ruby_vm/views/*.erb relies on the zjit instructions a) being contiguous and
   # b) being reliably ordered after all the other instructions.
   def test_instruction_order
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -844,19 +844,14 @@ class TestZJIT < Test::Unit::TestCase
     assert_compiles 'true', %q{
       def test = nil.nil?
       test
-      test
-    }
+    }, insns: [:opt_nil_p]
   end
 
-  def test_obsolete_nil_nil
-    assert_compiles '1', %q{
-      def test
-        nil.nil?
-        1
-      end
+  def test_non_nil_nil
+    assert_compiles 'false', %q{
+      def test = 1.nil?
       test
-      test
-    }
+    }, insns: [:opt_nil_p]
   end
 
   # tool/ruby_vm/views/*.erb relies on the zjit instructions a) being contiguous and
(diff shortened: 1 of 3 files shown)


test/ruby/test_zjit.rb
```diff
@@ -840,6 +840,44 @@ def test(x)
...
+
+  def test_obsolete_non_nil_nil
+    assert_compiles '2', %q{
+      def test
+        1.nil?
+        2
+      end
+      test
+      test
+    }
```
k0kubun: I think you meant to add `call_threshold: 2` to all these tests.

k0kubun (approved):

test/ruby/test_zjit.rb
st0012: Ah I tested the code with `./build/miniruby --zjit --zjit-dump-hir` and I assumed the threshold there is the same as the tests. But looking at `eval_with_jit`, the default is actually `1`. So I only need to call `test` once if `./build/miniruby --zjit --zjit-dump-hir --zjit-call-threshold=1` can already generate the expected hir?

k0kubun: Actually, yeah, it'd work in this case. Because the receiver is a literal in the method, type inference probably gives you an optimized method call without a profiling cycle. You may remove the second `test` call from them.

st0012: Done 👍

test/ruby/test_zjit.rb
```diff
@@ -855,6 +854,22 @@ def test
...
+  end
+
+  def test_non_nil_nil
+    assert_compiles 'false', %q{
+      def test = 1.nil?
+      test
+    }
+  end
+
+  def test_obsolete_non_nil_nil
```
tekknolagi: Nit: I don't think we need this one as a Ruby test. We're checking the behavior of calling the method in the other test and we're not checking that it's removed on the Ruby side, only on the HIR side

test/ruby/test_zjit.rb
```diff
@@ -855,6 +854,22 @@ def test
...
       end
       test
+    }
+  end
+
+  def test_non_nil_nil
+    assert_compiles 'false', %q{
+      def test = 1.nil?
+      test
+    }
```
tekknolagi: Let's do `insns: [:opt_nil]` or whatever it is so we don't accidentally get constant folded by the bytecode compiler. It doesn't normally happen with method calls I think but there are some surprising cases

test/ruby/test_zjit.rb
```diff
@@ -840,6 +840,25 @@ def test(x)
...
 
+  def test_nil_nil
+    assert_compiles 'true', %q{
+      def test = nil.nil?
+      test
+      test
+    }
+  end
+
+  def test_obsolete_nil_nil
```
tekknolagi: (not useful, ..., as with the other test)

test/ruby/test_zjit.rb
```diff
@@ -840,6 +840,25 @@ def test(x)
     }, call_threshold: 1, insns: [:branchnil]
   end
 
+  def test_nil_nil
```
tekknolagi, line 843: ( `insns: ...` as with the other test)

tekknolagi (approved):
nice, lgtm

tekknolagi:
Looks like it has a conflict :(

st0012:
@tekknolagi Resolved 👍

PR #13765, opened 2025-07-02
Object#hash: use the object address for T_OBJECT

casperisfine:
Updated attempt at: https://github.com/ruby/ruby/pull/13290

Rather than to generate an object_id, we can use the object address. However we have to record that we did, so that if the object is later moved, we know to record its original address inline, so that it keeps a stable hash.

We can't use the same strategy for other objects because they delegate their fields to another IMEMO/fields object. So if they don't yet have fields or if the fields object is full, during compaction we'd need to allocate another one but that not allowed during GC.

### DIFF
--- a/include/ruby/internal/fl_type.h
+++ b/include/ruby/internal/fl_type.h
@@ -216,12 +216,15 @@ ruby_fl_type {
      */
     RUBY_FL_PROMOTED    = (1<<5),
 
-    /**
-     * This flag is no longer in use
-     *
-     * @internal
-     */
-    RUBY_FL_UNUSED6    = (1<<6),
+
+   /**
+    * This flag is used to mark when an object address has been observed by
+    * `rb_obj_hash` or similar. This indicates to the GC that the old address
+    * must be recorded inside the moved object to keep `rb_obj_hash` stable.
+    *
+    * @internal
+    */
+    RUBY_FL_ADDRESS_SEEN = (1<<6),
 
     /**
      * This flag has  something to do with finalisers.  A  ruby object can have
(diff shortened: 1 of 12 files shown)


include/ruby/internal/fl_type.h
```diff
@@ -216,12 +216,15 @@ ruby_fl_type {
...
-    RUBY_FL_UNUSED6    = (1<<6),
+
+   /**
+    * This flag is used to mark when an object address has been observed by
+    * `rb_obj_hash` or similar. This indicates to the GC that the old address
+    * must be recorded inside the moved object to keep `rb_obj_hash` stable.
+    *
+    * @internal
+    */
+    RUBY_FL_ADDRESS_SEEN = (1<<6),
```
casperisfine, line 227: If only T_OBJECT can use this strategy, we can use a FL_USER flag instead of a global flag.

variable.c
```diff
@@ -4782,3 +4788,53 @@ rb_const_lookup(VALUE klass, ID id)
...
+
+    if (!rb_shape_has_old_address(shape_id)) {
+#if SIZEOF_LONG == SIZEOF_VOIDP
+        old_address = LONG2NUM(old_address);
+#elif SIZEOF_LONG_LONG == SIZEOF_VOIDP
+        old_address = LL2NUM(old_address);
+#else
+#error "Unexpected VALUE size"
+#endif
+        rb_obj_field_set(obj, rb_shape_transition_old_address(shape_id), 0, old_address);
```
casperisfine, line 4838: Actually this doesn't work well. `rb_shape_transition_old_address` will lookup or insert into a `managed_id_table` which may be a `T_MOVED`.

So it's currently not possible to make any shape transition (other than shape_id flags) as part of compaction.

casperisfine: Another reason it doesn't work, is that if the shape currently has a single child, creating a second child need to allocate a `managed_id_table`.

So for this whole PR to work, the compaction `rb_gc_update_moved_object` would need to be invoked once it's now OK to allocate objects (or the transition being created before allocations are disallowed).

PR #13770, opened 2025-07-02
ZJIT: Reject ISEQs with too-large stack_max

k0kubun:


st0012 (approved):

tekknolagi (approved):
I wonder if there are any cases where the optimizer could fold a pathological case like this away. Should we check it by another signal later in the pipeline somehow?

k0kubun:
Once we stop eagerly spilling stack temps and locals on side exits, maybe. For now, I think we're most likely going to hit it anyway when spilling them on side exits.

PR #13771, opened 2025-07-02
Adapt to some changes in rubygems/rubygems

deivid-rodriguez:
This is to catch up with https://github.com/rubygems/rubygems/pull/8793.


deivid-rodriguez:
Thank you @hsbt!

PR #13772, opened 2025-07-02
ZJIT: Support inference of ClassExact type

st0012:


zjit/src/hir_type/mod.rs
```diff
@@ -396,13 +422,15 @@ impl Type {
...
         if self.is_subtype(types::ArrayExact) { return Some(unsafe { rb_cArray }); }
+        if self.is_subtype(types::ClassExact) { return Some(unsafe { rb_cClass }); }
         if self.is_subtype(types::FalseClassExact) { return Some(unsafe { rb_cFalseClass }); }
         if self.is_subtype(types::FloatExact) { return Some(unsafe { rb_cFloat }); }
         if self.is_subtype(types::HashExact) { return Some(unsafe { rb_cHash }); }
         if self.is_subtype(types::IntegerExact) { return Some(unsafe { rb_cInteger }); }
         if self.is_subtype(types::NilClassExact) { return Some(unsafe { rb_cNilClass }); }
         if self.is_subtype(types::ObjectExact) { return Some(unsafe { rb_cObject }); }
         if self.is_subtype(types::RangeExact) { return Some(unsafe { rb_cRange }); }
+        if self.is_subtype(types::RegexpExact) { return Some(unsafe { rb_cRegexp }); }
```
st0012, line 416: I missed this in #13760

tekknolagi: Can you please split that out into a separate commit?

zjit/src/hir_type/mod.rs
```diff
@@ -145,6 +145,28 @@ fn is_range_exact(val: VALUE) -> bool {
...
 }
 
+fn is_class_exact(val: VALUE) -> bool {
+    // Only consider objects with RUBY_T_CLASS type
+    if val.builtin_type() != RUBY_T_CLASS {
+        return false;
+    }
+
+    // rb_mRubyVMFrozenCore is created with rb_class_new, so it has type RUBY_T_CLASS
+    // but zjit shouldn't treat it as a class
```
tekknolagi: Why shouldn't ZJIT treat it as a class?

zjit/src/hir_type/mod.rs
```diff
@@ -145,6 +145,28 @@ fn is_range_exact(val: VALUE) -> bool {
...
+    if val == unsafe { rb_mRubyVMFrozenCore } {
+        return false;
+    }
+
+    // rb_cModule is not a class
```
tekknolagi: Why is Module not a class?

zjit/src/hir.rs
```diff
@@ -5731,6 +5731,71 @@ mod opt_tests {
...
+            bb0(v0:BasicObject):
+              PatchPoint SingleRactorMode
+              PatchPoint StableConstantNames(0x1000, C)
+              v7:ClassExact[VALUE(0x1008)] = Const Value(VALUE(0x1008))
+              Return v7
+        "#]]);
+    }
+
+    #[test]
+    fn core_class_type_inference() {
```
tekknolagi: Can you also check `Class`, `BasicObject`?

tekknolagi (commented):
I am a little confused on some of the restrictions

zjit/src/hir_type/mod.rs
st0012: My bad. `Module` is a class but not its instances.

zjit/src/hir_type/mod.rs
st0012: Because it's an internal class hidden from the user ([source](https://github.com/Shopify/ruby/blob/master/vm.c#L4014-L4036)). It's used in things like merging hash in keyword splat arguments.

So if we don't exclude it, `def test(a) = foo **a, b: 1` will generate

```
    fn test:
      v3:ClassExact[VMFrozenCore] = Const Value(VALUE(0x1000)) # see this
      v5:HashExact = NewHash
      v7:BasicObject = SendWithoutBlock v3, :core#hash_merge_kwd, v5, v1
      v8:ClassExact[VMFrozenCore] = Const Value(VALUE(0x1000)) # and this
      v9:StaticSymbol[:b] = Const Value(VALUE(0x1008))
      v10:Fixnum[1] = Const Value(1)
      v12:BasicObject = SendWithoutBlock v8, :core#hash_merge_ptr, v7, v9, v10
      SideExit
```

which I think it's wrong.

Also, because it's only used internally, users won't be able to call `new` (or other `Class`'s instance methods) on it.
I think for `VMFrozenCore` values we can create a type specifically for it?

zjit/src/hir.rs
st0012: Added

zjit/src/hir_type/mod.rs
st0012: Done

zjit/src/hir_type/mod.rs
tekknolagi: A lot of things aren't user facing but that doesn't mean the optimizer should treat them specially. If they behave the same way internally---if they use the same class lookup & manipulation infrastructure---we should treat them the same way. We're going to see the `:core#hash_merge_kwd` in IR dumps anyway

tekknolagi: We have a special type for Qundef, for example, which should not be user facing---we still need to handle and consider it in the optimizer

zjit/src/hir.rs
```diff
@@ -5731,6 +5731,82 @@ mod opt_tests {
...
+              PatchPoint StableConstantNames(0x1000, C)
+              v7:ClassExact[VALUE(0x1008)] = Const Value(VALUE(0x1008))
+              Return v7
+        "#]]);
+    }
+
+    #[test]
+    fn core_classes_type_inference() {
+        eval("
+            def test = [String, Class, BasicObject]
```
tekknolagi: add Module and remove separate test (it is now class exact)

zjit/src/hir_type/mod.rs
st0012: Ok that's fair. I'll update it 👍

tekknolagi (approved):

PR #13774, opened 2025-07-03
ZJIT: Skip a hanging ractor test

k0kubun:
After https://github.com/ruby/ruby/pull/13773, this is the only test in `bootstraptest/test_ractor.rb` that doesn't work. It hangs and never finishes.

I haven't looked into how it ends up hanging, but given that it doesn't even work for YJIT, I'm tempted to skip it for now so that we can start running other test cases in `test_ractor.rb` on CI, leaving an issue to revisit it. WDYT?

We need both PRs to be merged to enable `bootstraptest/test_ractor.rb`, so I haven't modified the workflow yet. I'll make sure it's enabled before or after a second PR gets merged.

### DIFF
--- a/bootstraptest/runner.rb
+++ b/bootstraptest/runner.rb
@@ -891,4 +891,8 @@ def yjit_enabled?
   ENV.key?('RUBY_YJIT_ENABLE') || ENV.fetch('RUN_OPTS', '').include?('yjit') || BT.ruby.include?('yjit')
 end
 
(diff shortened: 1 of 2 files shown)


tekknolagi:
As long as we file an issue to track it, sure

tekknolagi (approved):

k0kubun:
filed https://github.com/Shopify/ruby/issues/618

PR #13780, opened 2025-07-03
ZJIT: Add a simple HIR validator

Fidget-Spinner:
This PR adds a simple validator for ZJIT's HIR.

See issue https://github.com/Shopify/ruby/issues/591


Fidget-Spinner:
cc @tekknolagi

zjit/src/hir.rs
```diff
@@ -2005,6 +2005,49 @@ impl Function {
...
+    fn validate_block_terminators_and_jumps(&self) {
+        for block_id in self.rpo() {
+            let mut block_has_terminator = false;
+            let insns = &self.blocks[block_id.0].insns;
+            for (idx, insn_id) in insns.iter().enumerate() {
+                let insn = self.find(*insn_id);
+                if !insn.is_terminator() {
+                    continue;
+                }
+                if let Insn::Jump(BranchEdge{target, args}) = &insn {
```
tekknolagi: This should also check parameter matches for conditional jumps--IfFalse and IfTrue--which are not terminators

zjit/src/hir.rs
```diff
@@ -2005,6 +2005,49 @@ impl Function {
             None => {},
         }
     }
+
+
+    /// Validates the following:
+    /// 1. Basic block jump args match parameter arity.
+    /// 2. Every terminator must be in the last position.
+    /// 3. Every block must have a terminator.
+    fn validate_block_terminators_and_jumps(&self) {
```
tekknolagi: Let's have this return `Result<(), ValidationError>` so we can stop compilation gracefully

zjit/src/hir.rs
```diff
@@ -2005,6 +2005,49 @@ impl Function {
...
+                }
+            }
+            if !block_has_terminator {
+                panic!("Block ID has no terminator: {}", block_id)
+            }
+        }
+    }
+
+    fn validate_block_uses(&self) {
+        // TODO(kenjin): Need to validate all uses and kills are properly gen-ed.
```
tekknolagi: Will you be doing this in this PR? If not, let's remove this function for now

zjit/src/hir.rs
```diff
@@ -4701,6 +4744,7 @@ mod opt_tests {
         unsafe { crate::cruby::rb_zjit_profile_disable(iseq) };
         let mut function = iseq_to_hir(iseq).unwrap();
         function.optimize();
+        function.validate();
```
tekknolagi: Let's also run this after construction in `iseq_to_hir`

tekknolagi (changes requested):
Nice start--looking good

zjit/src/hir.rs
Fidget-Spinner: Removed this function, will be done in another follow-up, as it might be a little slow to run.

Fidget-Spinner:
Sorry: new to ruby contributing. Should I force push or will you squash merge at the end?

tekknolagi:
I can squash at the end. If you have multiple logical commits that you would like preserved, you should force push and I can rebase

Fidget-Spinner:
> I can squash at the end. If you have multiple logical commits that you would like preserved, you should force push and I can rebase

Nope I don't. Please squash and overwrite the commit message with your own at the end. Thanks!

tekknolagi:
If you have time, please add some tests for this, but it's not dire. I will otherwise merge tonight or Monday

Fidget-Spinner:
I would add tests, but I'm not too sure how to at the moment. We'd basically need to pass a malformed iseq to the translation function and have it fail on validation right? Not sure how to construct an iseq in Rust in the first place considering `rb_iseq_t` is an opaque struct in Rust. There's probably some way to do it in C land, but I'm not familiar enough with the Ruby internals to achieve that yet.

tekknolagi:
No, we'd construct the HIR manually---not from an ISEQ

Fidget-Spinner:
Ah ok, I was under the impression you wanted an integration test for `iseq_to_hir` with the validator. That makes more sense now :). I'll add tests for this in another PR if you're okay with it. The fact it passes everything is a pretty strong indicator it's at least not detecting any false positives.

Fidget-Spinner:
Oh nevermind, I just found out `rpo_tests` exists. I can just copy some tests from there. Will have it done by tomorrow.

Fidget-Spinner:
The failures look to be flakiness of Ractor tests. This PR doesnt touch Ractors at all so I'm not concerned.

PR #13782, opened 2025-07-03
ZJIT: Panic on BOP redefinition only when needed

k0kubun:
This PR changes `rb_zjit_bop_redefined` to panic only when JIT code relies on the BOP assumption while it used to panic whenever any BOP is redefined. It allows us to run `bootstraptest/test_insns.rb` without crashing.

It also removes LIR `Insn::PadInvalPatch` because the instruction's semantics don't seem useful when we're not doing block-level invalidation. I think we want to invalidate JIT code on HIR instruction boundaries, so padding nops based on the number of instructions from the beginning of the block seems useless.


tekknolagi (approved):
nice

PR #13783, opened 2025-07-03
Fix missing write barriers of imemo_fields in Ractor copy/move traversal

jhawthorn:
Found by wbcheck. This fixes two cases where write barriers could be missed when Ractor copying/moving objects with generic fields.

* When we move a geniv object, the new object needs a write barrier because it gains a reference to the fields. (incremental GC could happen as during inserting the new key into the table)
* When we're copying a geniv object, the write barrier needs to come from the fields object rather than the object itself.

cc @byroot

### DIFF
--- a/ractor.c
+++ b/ractor.c
@@ -1653,10 +1653,10 @@ obj_traverse_replace_i(VALUE obj, struct obj_traverse_replace_data *data)
         obj = replacement;
     }
 
-#define CHECK_AND_REPLACE(v) do { \
+#define CHECK_AND_REPLACE(parent_obj, v) do { \
     VALUE _val = (v); \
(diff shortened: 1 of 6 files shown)


ractor.c
```diff
@@ -1681,7 +1681,7 @@ obj_traverse_replace_i(VALUE obj, struct obj_traverse_replace_data *data)
             uint32_t fields_count = RSHAPE_LEN(RBASIC_SHAPE_ID(obj));
```
jhawthorn, line 1681: I feel like up here I need to also make a change so that the complex fields also have the correct write barrier, but I want to make a test first that surfaces that to wbcheck

jhawthorn: Added tests and found this and one more issue (when a large list of IVs is resized)

PR #13784, opened 2025-07-03
ZJIT: Handle GCed objects in profiling data

k0kubun:
This PR fixes a crash in `VALUE::class_of()`. It allows us to run some micro-benchmarks on yjit-bench's default harness.

The assertions were copied from YJIT, and that was valid for YJIT because it only looks at objects that are on stack during compilation. However, it's not necessarily the case for ZJIT because it looks at objects collected in past profiling cycles, which could be GC-ed. So they could be freed or moved in ZJIT.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -915,6 +915,12 @@ class TestZJIT < Test::Unit::TestCase
     end
   end
 
+  def test_require_rubygems
+    assert_runs 'true', %q{
+      require 'rubygems'
+    }, call_threshold: 2
+  end
+
   private
 
   # Assert that every method call in `test_script` can be compiled by ZJIT
--- a/zjit/src/cruby.rs
+++ b/zjit/src/cruby.rs
@@ -423,7 +423,7 @@ impl VALUE {
     }
 
     pub fn string_p(self) -> bool {
-        self.class_of() == unsafe { rb_cString }
+        self.class_of() == Some(unsafe { rb_cString })
     }
 
     /// Read the flags bits from the RBasic object, then return a Ruby type enum (e.g. RUBY_T_ARRAY)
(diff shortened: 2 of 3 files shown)


st0012 (approved):
I think you'll need to rebase for https://github.com/ruby/ruby/pull/13772 but otherwise LGTM 👍

k0kubun:
It's possible but you have to either slowly scan every profiled object or consume extra memory to maintain a lookup table.
We actually need to do that for moved objects at some point, so that'll be implemented in the future. But I don't think we should prioritize compaction support (or just cleaning up profiling data) over making it work now.

tekknolagi:
I am not sure I understand what's going on. Are we doing "undefined behavior" right now in implicitly storing weakrefs in the profiling table and this kind of patches over that in the happy case where the object hasn't been reused yet post-GC?

If so, that makes me kind of nervous. And this change makes me more nervous since it normalizes (kinda) inspecting objects of questionable provenance. I would be happier if we instead filtered for valid objects (checking T_MOVED or whatever) on profile ingestion as a patch over that UB, since that at least localizes the sketchiness to one place.

(And I would be happiest, of course, if we registered profile data as actual weak references or strong references, so that the GC may manage them normally.)

If any of those solutions are too challenging for bringup right now, then we should pick a suitable future path and add it as a TODO & an issue so that we may revert this change to `class_of` in the future.

k0kubun:
Actually, unlike moving references for `T_MOVED`, I don't think we want to remove freed objects from profiling data because it would mean to hook every single free of any objects in GC and look up a table with it.

What we should probably do instead is to mark objects in profiled data to keep them alive until they're used, and drop the profiled objects as soon as we're done with it.

I'm not sure if triggering that from ISEQ mark (like YJIT does) fixes this `T_NONE` crash, but I'll give it a shot.

k0kubun:
I implemented the GC-mark callback https://github.com/ruby/ruby/pull/13784/commits/69b83728f9ec5297a7d57a0ce20084c27e6f8618. It seems to successfully fix the problem on yjit-bench.

tekknolagi (approved):
There are a lot of changes I'm not super familiar with but looks broad strokes great

PR #13790, opened 2025-07-04
ZJIT: Add a ccall macro that also adds an LIR comment

XrXr:
This DRYs up the `asm_comment!` + `asm.ccall` combo, and makes ccalls
have a comment by default.
The macro itself removes visual noise and puts the callee's name and
arguments first. `[]` on the outside to make it visually distinctive
from the `()` inside.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -117,7 +117,7 @@ impl Opnd
     }
 
     /// Constructor for constant pointer operand
-    pub fn const_ptr(ptr: *const u8) -> Self {
+    pub fn const_ptr<T>(ptr: *const T) -> Self {
         Opnd::UImm(ptr as u64)
     }
 
@@ -1836,7 +1836,7 @@ impl Assembler
                 }
 
                 asm_comment!(self, "save cfp->pc");
-                self.load_into(Opnd::Reg(Assembler::SCRATCH_REG), Opnd::const_ptr(pc as *const u8));
+                self.load_into(Opnd::Reg(Assembler::SCRATCH_REG), Opnd::const_ptr(pc));
                 self.store(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_PC), Opnd::Reg(Assembler::SCRATCH_REG));
 
                 asm_comment!(self, "save cfp->sp");
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -5,7 +5,7 @@ use crate::codegen::local_size_and_idx_to_ep_offset;
 use crate::cruby::{Qundef, RUBY_OFFSET_CFP_PC, RUBY_OFFSET_CFP_SP, SIZEOF_VALUE_I32};
 use crate::hir::SideExitReason;
 use crate::options::{debug, get_option};
-use crate::{cruby::VALUE};
+use crate::cruby::VALUE;
 use crate::backend::current::*;
 use crate::virtualmem::CodePtr;
 use crate::asm::{CodeBlock, Label};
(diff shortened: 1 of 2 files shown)


zjit/src/codegen.rs
```diff
@@ -810,20 +781,12 @@ fn gen_new_array(
 
     let length: ::std::os::raw::c_long = elements.len().try_into().expect("Unable to fit length of elements into c_long");
 
-    asm_comment!(asm, "call rb_ary_new");
```
XrXr, line 833: Mismatch like this will large be gone.

st0012 (approved):

zjit/src/backend/lir.rs
```diff
@@ -2243,6 +2243,15 @@ macro_rules! asm_comment {
 }
 pub(crate) use asm_comment;
 
+/// Convenience macro over [`Assembler::ccall`] that also adds a comment with the function name.
+macro_rules! ccall {
+    [$fn_name:ident ( $($args:expr),* ) , $asm: ident] => {{
```
tekknolagi: Can this instead use the `asm` from the existing scope?

tekknolagi (approved):
very cool

zjit/src/backend/lir.rs
```diff
@@ -2243,6 +2243,15 @@ macro_rules! asm_comment {
 }
 pub(crate) use asm_comment;
 
+/// Convenience macro over [`Assembler::ccall`] that also adds a comment with the function name.
+macro_rules! ccall {
```
k0kubun: I called `asm_comment!` as `asm_comment!` so that it looks closer to `asm.comment`. It clarifies that generated code makes the C call, not the Rust code. So I would use a syntax like `asm_ccall!(asm, rb_gvar_set(id.0.into(), val))` for consistency, but you may not like it.

Or maybe `asm_ccall!(asm, rb_gbar_set, vec![id.0.into(), val])` if the purpose of the `[` `]` syntax was to avoid too many `))`s, which looks even closer to `asm.ccall!(rb_gbar_set, vec![id.0.into(), val])`. But I guess you prefer the vector-less syntax 🙂

zjit/src/backend/lir.rs
k0kubun: I'm usually the guy who suggests that kind of implicit stuff to make things shorter, but we will not be able to use such a macro in `compile_side_exits()` where `asm` is called `self` (unless you add `let asm = self;` for the C call).

zjit/src/backend/lir.rs
tekknolagi: I have a slight preference for `asm_ccall!(asm, rb_gbar_set, vec![id.0.into(), val])` as well

tekknolagi: Or at least making it look slightly less like a function call that is happening at compile-time

zjit/src/backend/lir.rs
XrXr: Unfortunately this is not possible due to macro hygiene [rules](https://doc.rust-lang.org/reference/macros-by-example.html#r-macro.decl.hygiene.intro). 

> Macros by example have mixed-site hygiene. This means that [loop labels](https://doc.rust-lang.org/reference/expressions/loop-expr.html#loop-labels), [block labels](https://doc.rust-lang.org/reference/expressions/loop-expr.html#labelled-block-expressions), and local variables are looked up at the macro definition site while other symbols are looked up at the macro invocation site.

If you move `opnd!` out of `gen_insn` for example, it won't build anymore.

zjit/src/backend/lir.rs
XrXr: How about `asm_ccall!(asm, rb_gvar_get, id.0.into(), val)`? This looks similar to the HIR CCall printout.

k0kubun (approved):
P.S. Pushing a fix without making any comment doesn't send an email notification, so I've failed to notice this PR is updated. It's partly why I make "Fixed at xxxx" kinds of comments.

PR #13791, opened 2025-07-04
ZJIT: Start testing against `/test/ruby/` and update all ZJIT related testing targets/docs

st0012:
- `make zjit-check` has replaced the **old** `make zjit-test-all`, which runs ZJIT unit tests (Rust) and `make test-all TESTS=test/ruby/test_zjit.rb`
- `make zjit-test-all` now runs ZJIT with tests under `/test/ruby/` with some test cases excluded under `test/.excludes-zjit/` due to known crash/panic/failure

There are 3 ways tests fail to execute with ZJIT:

1. The behaviour is incorrect so the test failed
2. Ruby crashes (due to assertion failure...etc.)
3. ZJIT panicked

We can avoid 1 and 2 by excluding those specific cases with `exclude(:test_method_name, "reason")`. But this doesn't avoid ZJIT panics.
"I think" this is because the exclusion is done [through removing methods](https://github.com/ruby/ruby/blob/3cf32e936416e6c4aa272e63a9ac626c55087891/tool/lib/test/unit.rb#L1327-L1336), so if ZJIT panics due to compiling certain iseqs, it'd panic before those methods even get removed?


k0kubun:
> zjit-test | zjit-test-rust

I'm reluctant to make the name `make zjit-test` longer. Unlike YJIT, we can't use the bare `cargo test` for ZJIT, so we have to type and run the command a lot. It seems pretty fair that `cargo test` for ZJIT is called `zjit-test` because every other test is just part of regular CRuby tests.

> zjit-test-ruby

Running a single test file in `test-all` is something you have to do a lot regardless of ZJIT, so I'm not sure if it's worth a new make target. But if you want to give a name to `test-all TESTS='test/ruby/test_zjit.rb'`, maybe repurpose `make zjit-test-all` to it because it's a test case in `test-all` for ZJIT?

> zjit-test-all | zjit-test-suite

I admit that `zjit-test-all` was not a good name, but if we're renaming it, how about `make zjit-check`? `make check` runs all make targets for testing CRuby, so `make zjit-check` can be a ZJIT version of it.

> zjit-test-ruby-all

It's confusing to insert "ruby" in between "test-all". `test-all` is a very well-known make target name (probably one of the most frequently used names), so calling it "test-ruby-all" makes it hard to recognize what it is.

I don't come up with a good name for this other than `make zjit-test-all`, but I think you're just making too many make targets. I'd suggest dropping either "zjit-test-ruby" or "zjit-test-ruby-all". At some point we'll just be able to use the actual `make test-all` for this, so maybe drop `zjit-test-ruby-all` (or both)?

If you need a place to maintain a set of skipped test cases for ZJIT, you might want to try "excludes" directory. Look at `test/.excludes*` for example. We could have `test/.excludes-zjit` and use that.

st0012:
@k0kubun Thanks for the suggestions! I didn't know about `.excludes*` folders but it's indeed better.
I'll get back to this after https://github.com/Shopify/ruby/issues/620 is resolved as it fails a lot tests right now.

.github/workflows/zjit-macos.yml
```diff
@@ -38,9 +38,8 @@ jobs:
           - test_task: 'ruby' # build test for combo build
             configure: '--enable-yjit --enable-zjit'
 
-          - test_task: 'test-all'
```
st0012, line 41: I'm not sure if it's worth testing `test_zjit` by itself on CI when we'll be running it in `zjit-test-all` anyway. Also the `test-all` name will be confusing along with `zjit-test-all`.

test/.excludes-zjit/TestProc.rb
```diff
@@ -0,0 +1,6 @@
+exclude(/test_/, 'Tests make ZJIT panic')
+
+# exclude(:test_proc_args_pos_rest_block, 'Test crashes with ZJIT')
+# exclude(:test_proc_args_rest_post_block, 'Test crashes with ZJIT')
+
+# exclude(:test_binding_receiver, 'Test fails with ZJIT')
```
st0012, line 6: As explained in description, I need to exclude all test methods to avoid the ZJIT panic. But I also identified these tests that'd also fail/crash even without the panic.

zjit/zjit.mk
```diff
@@ -49,11 +49,15 @@ update-zjit-bench:
...
-.PHONY: zjit-test-all
-zjit-test-all:
+.PHONY: zjit-check
+zjit-check:
 	$(MAKE) zjit-test
 	$(MAKE) test-all TESTS='$(top_srcdir)/test/ruby/test_zjit.rb'
 
+.PHONY: zjit-test-all
+zjit-test-all:
+	$(MAKE) test-all RUST_BACKTRACE=1 TEST_EXCLUDES='--excludes-dir=$(top_srcdir)/test/.excludes-zjit --name=!/memory_leak/' RUN_OPTS='--zjit --zjit-call-threshold=1' TESTS='$(top_srcdir)/test/ruby/'
```
k0kubun:
```suggestion
	$(MAKE) test-all RUST_BACKTRACE=1 TEST_EXCLUDES='--excludes-dir=$(top_srcdir)/test/.excludes-zjit --name=!/memory_leak/' RUN_OPTS='--zjit-call-threshold=1' TESTS='$(top_srcdir)/test/ruby/'
```

Every `--zjit*` also does `--zjit`, so this is unnecessary. We don't write options that way in the CI configurations, etc.

k0kubun (approved):
Great job :+1: Let's merge it once the CI passes.

zjit/zjit.mk
st0012: Updated!

.github/workflows/zjit-macos.yml
tekknolagi: As long as it is running in CI somehow. 

Although it may give more informative and early warnings than other test suites because it's more targeted

PR #13794, opened 2025-07-05
ZJIT: Support inference of ModuleExact type

st0012:


zjit/src/hir_type/mod.rs
```diff
@@ -202,7 +197,10 @@ impl Type {
         else if is_string_exact(val) {
             Type { bits: bits::StringExact, spec: Specialization::Object(val) }
         }
-        else if is_class_exact(val) {
+        else if val.builtin_type() == RUBY_T_MODULE {
```
tekknolagi: Modules can be subclassed and I think this check would accidentally include module subclasses. Please add a test

zjit/src/hir_type/gen_hir_type.rb
```diff
@@ -75,6 +75,7 @@ def base_type name
 base_type "Set"
 base_type "Regexp"
 base_type "Class"
+base_type "Module"
```
tekknolagi: I wonder if we should explicitly mark Class as subclassing Module in the type hierarchy

tekknolagi (approved):

zjit/src/hir_type/gen_hir_type.rb
st0012: Updated 👍

zjit/src/hir_type/mod.rs
st0012: Updated

zjit/src/hir_type/gen_hir_type.rb
```diff
@@ -74,19 +74,21 @@ def base_type name
 base_type "Range"
 base_type "Set"
 base_type "Regexp"
-base_type "Class"
+_, module_exact = base_type "Module"
+module_exact.subtype "Class"
```
tekknolagi: Since `Class` is final we can remove `ClassExact` and just use that (consistent with `Fixnum`, etc)

tekknolagi: Also we should make `Class` subclass `Module`, not `ModuleExact`

zjit/src/hir_type/mod.rs
```diff
@@ -145,9 +146,13 @@ fn is_range_exact(val: VALUE) -> bool {
...
 
-fn is_class_exact(val: VALUE) -> bool {
-    // Objects with RUBY_T_CLASS type and not instances of Module
-    val.builtin_type() == RUBY_T_CLASS && val.class_of() != unsafe { rb_cModule }
+fn is_module_exact(val: VALUE) -> bool {
+    if val.builtin_type() != RUBY_T_MODULE {
+        return false;
+    }
+
+    let klass = unsafe { rb_obj_class(val) };
```
tekknolagi, line 157: We should add a comment about how this is different from `class_of` because it looks so similar but does a different thing

tekknolagi (approved):
lgtm pending comments

zjit/src/hir_type/gen_hir_type.rb
st0012: Updated

tekknolagi (approved):

tekknolagi:
lmk when to merge

st0012:
@tekknolagi I think it's ready

PR #13795, opened 2025-07-05
[DOC] Tweaks for String#capitalize

BurdetteLamar:


string.c
```diff
@@ -8282,20 +8282,29 @@ rb_str_capitalize_bang(int argc, VALUE *argv, VALUE str)
...
- *     s.capitalize       # => "Hello world!"
+ *  - The first character is upcased.
+ *  - Other characters are downcased.
  *
- *  The casing may be affected by the given +mapping+;
- *  see {Case Mapping}[rdoc-ref:case_mapping.rdoc].
+ *  Examples:
+ *
+ *    'hello'.capitalize # => "Hello"
+ *    'HELLO'.capitalize # => "Hello"
```
nobu: It should be clarified that the first character of the whole string, not each words, will be upcased, as the previous examples, I think.

BurdetteLamar: > It should be clarified that the first character of the whole string, not each words, will be upcased, as the previous examples, I think.

Thanks, @nobu.  I should have realized that.  Will fix.

peterzhu2118 (approved):

PR #13797, opened 2025-07-05
ZJIT: Support guarding *Exact types

st0012:
ZJIT already can generate guard type instructions for *Exact types.

For example:

```
def test(strings)
  strings.map do |string|
    string.bytesize
  end
end

test(["foo", "bar"])
```

```
HIR:
fn block in test:
  PatchPoint MethodRedefined(String@0x1014be890, bytesize@0x19f1)
  v7:StringExact = GuardType v1, StringExact
  v8:Fixnum = CCall bytesize@0x16fa4cc18, v7
  Return v8

```

But zjit only supported guarding fixnums so this script would panic.

This commit adds support for guarding *Exact types.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -915,6 +915,201 @@ class TestZJIT < Test::Unit::TestCase
     end
   end
 
+  def test_module_name_with_guard_passes
+    assert_compiles '"Integer"', %q{
+      def test(mod)
+        mod.name
+      end
+
+      test(String)
+      test(Integer)
+    }, call_threshold: 2
+  end
+
+  def test_module_name_with_guard_fallthrough
+    # This test demonstrates that the guard side exit works correctly
+    # In this case, when we call with a non-Class object, it should fall back to interpreter
+    assert_compiles '["String", "Integer", "Bar"]', %q{
+      class MyClass
+        def name = "Bar"
+      end
+
+      def test(mod)
+        mod.name
+      end
+
+      results = []
+      results << test(String)
+      results << test(Integer)
+      results << test(MyClass.new)
+
+      results
+    }, call_threshold: 2
+  end
+
+  def test_string_bytesize_with_guard
+    assert_compiles '5', %q{
+      def test(str)
+        str.bytesize
+      end
+
+      test('hello')
+      test('world')
+    }, call_threshold: 2
+  end
+
+  def test_nil_value_nil_opt_with_guard
+    assert_compiles 'true', %q{
+      def test(val) = val.nil?
+
+      test(nil)
+      test(nil)
+    }, call_threshold: 2, insns: [:opt_nil_p]
+  end
+
+  def test_nil_value_nil_opt_with_guard_fallthrough
+    assert_compiles 'false', %q{
+      def test(val) = val.nil?
+
+      test(nil)
+      test(nil)
+      test(1)
+    }, call_threshold: 2, insns: [:opt_nil_p]
+  end
+
+  def test_true_nil_opt_with_guard
+    assert_compiles 'false', %q{
+      def test(val) = val.nil?
+
+      test(true)
+      test(true)
+    }, call_threshold: 2, insns: [:opt_nil_p]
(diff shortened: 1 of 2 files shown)


zjit/src/codegen.rs
```diff
@@ -979,6 +979,16 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         // Check if opnd is Fixnum
         asm.test(val, Opnd::UImm(RUBY_FIXNUM_FLAG as u64));
         asm.jz(side_exit(jit, state)?);
+    } else if let Some(expected_class) = guard_type.runtime_exact_ruby_class() {
+        // Guard for exact class types (e.g., ClassExact, ArrayExact, StringExact, etc.)
+        asm_comment!(asm, "guard exact class");
+
+        // Get the class of the value
+        let klass = asm.ccall(rb_yarv_class_of as *const u8, vec![val]);
```
tekknolagi, line 981: If we can inline this somehow, that would be great (needs a check if it's a special const, then load `RBASIC(obj)->klass`)

zjit/src/codegen.rs
```diff
@@ -979,6 +979,16 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
...
+
+        // Compare with expected class
+        asm.cmp(klass, Opnd::UImm(expected_class.into()));
```
tekknolagi: `Value`, not `UImm`

test/ruby/test_zjit.rb
```diff
@@ -915,6 +915,51 @@ def test_instruction_order
...
+    }, call_threshold: 2
+  end
+
+  def test_guard_type_side_exit
+    # This test demonstrates that the guard side exit works correctly
+    # In this case, when we call with a non-Class object, it should fall back to interpreter
+    assert_compiles '["String", "Integer", "Foo"]', %q{
+      class Foo
+        def name
+          "Foo"
```
tekknolagi: Let's use a different value than the class name to disambiguate

zjit/src/codegen.rs
tekknolagi: (this is a very handwavy sketch)

```
if expected_class == FalseClassExact {
  cmp val, 0
  jnz interpreter
} else if expected_class == NilClassExact {
  // fast check for nil
}
...
else {
  assert!(!expected_class.special_const_p());
  test val, RUBY_IMMEDIATE_MASK
  jz interpreter
  load cls, val->klass
  cmp cls, expected_class
  jnz interpreter
}
```

tekknolagi: Could be a TODO

zjit/src/hir.rs
```diff
@@ -6517,4 +6517,44 @@ mod opt_tests {
...
+            bb0(v0:BasicObject, v1:BasicObject):
+              PatchPoint MethodRedefined(Class@0x1000, name@0x1008)
+              v7:ClassExact = GuardType v1, ClassExact
+              v8:StringExact|NilClassExact = CCall name@0x1010, v7
+              Return v8
+        "#]]);
+    }
+
+    #[test]
+    fn test_guard_string_for_bytesize() {
```
tekknolagi: can you add tests for immediates like false, nil, true, symbol, float

tekknolagi (approved):
lgtm

zjit/src/hir.rs
st0012: I added them in https://github.com/ruby/ruby/pull/13825

tekknolagi: for posterity, per my comment there, you will also need to update this PR to actually codegen and execute the optimized code

zjit/src/codegen.rs
st0012: I've added Ruby execution tests for different classes (integer, float, true, false...etc.) but for inlining I plan to add them in a separate PR. I created https://github.com/Shopify/ruby/issues/624 for it.

PR #13799, opened 2025-07-06
`atomic.h`: Add C11 implmentation of atomic operations

fuhsnn:
Introduce an implementation of atomic operations with C11 _Atomic qualifier and <stdatomic.h> functions. This allows C compilers relying on C11 model as the sole atomic interface to compile Ruby. The existing configure definition `HAVE_STDATOMIC_H` is used as the feature switch.

To be as non-intrusive as possible, new `#elif` blocks are consistently positioned right before `#error`, all currently supported systems are expected to be unaffected by the change, that is, no additional `#include` or library dependency or raising of `-std` version.

Several `#else` branches that were only used for 32-bit Windows and Oracle platforms are changed to explicit `#elif Win || Oracle` to maintain a consistent order of implementation branches.


nobu:
> the currently supported systems: GNU/Windows/Oracle

Do you mean Solaris by Oracle here?
I was confused because Oracle provides also Oracle Linux.

fuhsnn:
> Do you mean Solaris by Oracle here?

I used Oracle to describe whatever compiler that evals  `(defined(__sun) && defined(HAVE_ATOMIC_H))` to `1`, honestly I don't know which exact systems.

Edit:
[Oracle Developer Studio 12.6 manual](https://docs.oracle.com/cd/E77782_01/html/E77803/uc-cc-1.html):
> Note that __sun is only defined on Oracle Solaris platforms

[cpredef/predef](https://github.com/cpredef/predef/blob/master/OperatingSystems.md) claims Solaris, SunOS and illumos

Should I change it to Solaris?

nobu:
> > Do you mean Solaris by Oracle here?
> 
> I used Oracle to describe whatever compiler that evals `(defined(__sun) && defined(HAVE_ATOMIC_H))` to `1`, honestly I don't know which exact systems.
> 
> Edit: [Oracle Developer Studio 12.6 manual](https://docs.oracle.com/cd/E77782_01/html/E77803/uc-cc-1.html):
> 
> > Note that __sun is only defined on Oracle Solaris platforms
> 
> [cpredef/predef](https://github.com/cpredef/predef/blob/master/OperatingSystems.md) claims Solaris, SunOS and illumos
> 
> Should I change it to Solaris?

At least illumos is based on OpenSolaris and not owned by Oracle.
And, as I wrote before, it will be confusing with "Oracle Linux" to call Solaris as "Oracle" simply.

fuhsnn:
If `Solaris` is not preferable, How about:
```
These branches are only active for `defined(_WIN32)` and
`(defined(__sun) && defined(HAVE_ATOMIC_H))` platforms,
codify the fact by changing `#else` to `#elif`'s that explicitly
include those targets, and `#error`-out otherwise.
```
```
The implementation is only active if `HAVE_STDATOMIC_H` is defined,
and only after the compiler fails to match all currently supported
systems.
```

nobu:
> If `Solaris` is not preferable, How about:

I haven't said that Solaris is not preferable, at all, and the word Solaris is not in your commit log.
I'm saying that it is ambiguous whether it means Oracle Linux or Oracle Solaris by just Oracle.
Rather I think you should use Solaris instead.

fuhsnn:
> I haven't said that Solaris is not preferable

Sorry for the confusion, reworded and rebased.

fuhsnn:
Is it acceptable for the `macOS result` CI jobs to fail? Several recent commits on master also had this one failing:
1582bd9382f202b45c11c874c8c26ca8727276de
12d44dbc496f131fcd2af8864d19471c0720800d
0b23a8db6069c726af0c5b090e65790fda347dff
1317377fa74811cad088e3ee17926007a7d06732

PR #13802, opened 2025-07-07
ZJIT: Fix crashes and a miscomp for param spilling

XrXr:
Easiest to review each commit.

- **ZJIT: Add failing test to test_spilled_method_args()**
- **ZJIT: Ban `asm.load_into(Mem, ..)` and avoid it in gen_entry_params()**
- **ZJIT: Fix crashes and a miscomp for param spilling**

Still not completely debugged since there is still some issue with
it clobbering the self parameter, which is x0 on ARM.

```ruby
# this crashes on ARM due to `self` clobber
def a(n1,n2,n3,n4,n5,n6,n7,n8,n9) = [self, n1, n9]
p a(1,0,0,0,0,0,0,0,9)
```

It gets clobbered in and outside of gen_entry_params(), so it's not
easily fixable by just rejigging the order in there.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -32,13 +32,13 @@ pub enum MemBase
 pub struct Mem
 {
     // Base register number or instruction index
-    pub(super) base: MemBase,
+    pub base: MemBase,
 
     // Offset relative to the base pointer
-    pub(super) disp: i32,
+    pub disp: i32,
 
     // Size in bits
-    pub(super) num_bits: u8,
+    pub num_bits: u8,
 }
 
 impl fmt::Debug for Mem {
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -2150,6 +2150,7 @@ impl Assembler {
     }
 
     pub fn load_into(&mut self, dest: Opnd, opnd: Opnd) {
+        assert!(matches!(dest, Opnd::Reg(_) | Opnd::VReg{..}), "Destination of load_into must be a register");
         match (dest, opnd) {
             (Opnd::Reg(dest), Opnd::Reg(opnd)) if dest == opnd => {}, // skip if noop
             _ => self.push_insn(Insn::LoadInto { dest, opnd }),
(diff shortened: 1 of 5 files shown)


zjit/src/codegen.rs
```diff
@@ -216,6 +215,8 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Optio
                 let new_sp = asm.sub(NATIVE_STACK_PTR, jit.c_stack_bytes.into());
                 asm.mov(NATIVE_STACK_PTR, new_sp);
             }
+
+            gen_entry_params(&mut asm, iseq, function.block(BlockId(0)));
```
k0kubun: We deliberately put this in `gen_entry()` instead of here so that JIT-to-JIT calls don't have to read arguments off of EP. When we optimize frame push further by removing eager spilling, this would stop working.

Could we adjust operands in `jit.opnds` for params if they point to memory based off of `NATIVE_STACK_PTR` instead? If it doesn't work, let's leave a TODO comment to optimize argument access on JIT-to-JIT calls.

k0kubun: > Could we adjust operands in jit.opnds for params if they point to memory based off of NATIVE_STACK_PTR instead?

Or let the backend do it? https://github.com/ruby/ruby/pull/13789/files#r2190770455

XrXr: Wrote to the slots at entry point in [604c715](https://github.com/ruby/ruby/pull/13802/commits/604c71571c6ea1c70de061d36a6993c542e16a6e).

k0kubun (approved):
Thanks!

zjit/src/backend/lir.rs
```diff
@@ -2150,6 +2150,7 @@ impl Assembler {
     }
 
     pub fn load_into(&mut self, dest: Opnd, opnd: Opnd) {
+        assert!(matches!(dest, Opnd::Reg(_) | Opnd::VReg{..}), "Destination of load_into must be a register");
```
tekknolagi, line 2153: This feels a little too lax. Isn't there a chance the vreg is allocated to memory?

XrXr: There is a lot of code around register allocation that assume VReg means reg and I think there's no way a VReg becomes memory at the moment (because of no auto spills). I figure when they could become memory LIR would make some rewrites to completely erase these.

Also, maybe this is beside the point and we should remove these constrained `mov`s (like `Store`) and use the general `mov` for everything. I guess these are still useful in spots where we don't run the split (legalization) backend pass...

PR #13805, opened 2025-07-07
ZJIT: Add opnds macro for Vec<InsnId> to Vec<Opnd>

composerinteralia:
Along the same lines as the `opnd` macro we already have, but for a `Vec<InsnId>` instead of a single `InsnId`.

This gets a few for loops and `jit.get_opnd` calls out of the `gen_` functions.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -255,13 +255,21 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         };
     }
 
+    macro_rules! opnds {
+        ($insn_ids:ident) => {
+            {
+                Option::from_iter($insn_ids.iter().map(|insn_id| jit.get_opnd(*insn_id)))?
+            }
+        };
+    }
+
     if !matches!(*insn, Insn::Snapshot { .. }) {
         asm_comment!(asm, "Insn: {insn_id} {insn}");
     }
 
     let out_opnd = match insn {
         Insn::Const { val: Const::Value(val) } => gen_const(*val),
-        Insn::NewArray { elements, state } => gen_new_array(jit, asm, elements, &function.frame_state(*state)),
+        Insn::NewArray { elements, state } => gen_new_array(asm, opnds!(elements), &function.frame_state(*state)),
         Insn::NewRange { low, high, flag, state } => gen_new_range(asm, opnd!(low), opnd!(high), *flag, &function.frame_state(*state)),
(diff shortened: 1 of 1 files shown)


zjit/src/codegen.rs
```diff
@@ -255,13 +255,21 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         };
     }
 
+    macro_rules! opnds {
+        ($insn_ids:ident) => {
+            {
+                $insn_ids.iter().map(|insn_id| jit.get_opnd(*insn_id)).collect::<Option<Vec<Opnd>>>()?
```
XrXr: TIL about [`Option::from_iter`](https://doc.rust-lang.org/std/option/enum.Option.html#method.from_iter) cc @tekknolagi cause heard you asking about an idiom that does just this IRL.

> **Takes each element in the [Iterator](https://doc.rust-lang.org/std/iter/trait.Iterator.html): if it is [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None), no further elements are taken, and the [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None) is returned. Should no [None](https://doc.rust-lang.org/std/option/enum.Option.html#variant.None) occur, a container of type V containing the values of each [Option](https://doc.rust-lang.org/std/option/enum.Option.html) is returned.**

Also:

> If you want to create a collection from the contents of an iterator, the [Iterator::collect()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect) method is preferred. However, when you need to specify the container type, [FromIterator::from_iter()](https://doc.rust-lang.org/std/iter/trait.FromIterator.html#tymethod.from_iter) can be more readable than using a turbofish (e.g. ::<Vec<_>>()). See the [Iterator::collect()](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect) documentation for more examples of its use.

nit: I agree with the docs in this case that `Option::from_iter` is more readable than collect with a turbofish. But they're really close, so 🤷🏼 .

XrXr (approved):

zjit/src/codegen.rs
composerinteralia: Neat, thanks! Updated to use `Option::from_iter` 👍🏻

tekknolagi: oh wow I really like `Option::from_iter`. that's so dope

tekknolagi (approved):
very nice

k0kubun (approved):
nice!

PR #13806, opened 2025-07-07
Allow concurrent use of system(), IO.popen and Process.spawn in ractors

luke-gruber:
These ruby methods call `fork` or `vfork` internally, so we need to grab the fork lock in the parent and release it after the call. We don't reset the lock in the child because we're going to exec right after anyway. Without this lock, strange things were happening having to do with the timer thread. Sometimes multiple timers were running.

### DIFF
--- a/process.c
+++ b/process.c
@@ -4010,6 +4010,7 @@ retry_fork_async_signal_safe(struct rb_process_status *status, int *ep,
 
     while (1) {
         prefork();
+        rb_thread_acquire_fork_lock();
         disable_child_handler_before_fork(&old);
 #ifdef HAVE_WORKING_VFORK
         if (!has_privilege())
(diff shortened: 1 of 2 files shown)


luke-gruber:
Some CI failures right now. It looks like the added test may be triggering a ractor bug that we already know about that has a [PR](https://github.com/ruby/ruby/pull/13682) but has not been merged yet. I'll continue to investigate.

jhawthorn:
Hmm. The test doesn't reproduce for me (at least on linux, testing macos...). I feel like we shouldn't need the fork lock for these operations because we're going to exec immediately after and so whatever inconsistency that was trying to avoid shouldn't be an issue (but maybe there is some problem before the `exec` occurs).

luke-gruber:
You're right, sorry! Closing.

PR #13807, opened 2025-07-07
Make `rclass_ext->const_tbl` a managed object

casperisfine:
The table itself is a `rb_id_marked_table` and the values are `IMEMO/constentry`.
    
This open the door to do RCU updates of constant tables.

### DIFF
--- a/internal/imemo.h
+++ b/internal/imemo.h
@@ -15,7 +15,7 @@
 #include "ruby/internal/stdbool.h"     /* for bool */
 #include "ruby/ruby.h"          /* for rb_block_call_func_t */
 
-#define IMEMO_MASK   0x0f
+#define IMEMO_MASK   0x0f // NOTE: The debug gem mirror this constant.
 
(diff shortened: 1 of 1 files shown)


luke-gru:
@byroot Could the CI failures be related to [this](https://github.com/ruby/debug/blob/master/ext/debug/iseq_collector.c#L19)? The mask appears to be different.

casperisfine:
Hum, that's a good catch. Thank you.

casperisfine:
Ref: https://github.com/ruby/debug/pull/1151

casperisfine:
> Could the CI failures be related to [this](https://github.com/ruby/debug/blob/master/ext/debug/iseq_collector.c#L19)? The mask appears to be different.

It did indeed fix one issue. Unfortunately the `debug` gem suite is still failing when running in `parse.y` mode, unclear why.

This also is unrelated to the write barrier issue :/. It's progress though.

PR #13814, opened 2025-07-07
ZJIT: Add def-use validator via dataflow analysis

Fidget-Spinner:
This PR adds a validator based on dataflow analysis to ZJIT. It checks that all uses are dominated by a GEN-DEF prior.

See issue https://github.com/Shopify/ruby/issues/591

This is especially useful in validating optimizations don't zap away instructions that are actually needed, e.g. DCE.

Also included: a slight refactor of the DCE code to its own function, so I can reuse it.

Note: the algorithm uses the worklist algorithm rather than the iterative version for faster convergence.

Co-Authored-By: Max Bernstein <donotemailthisaddress@bernsteinbear.com>


zjit/src/hir.rs
```diff
@@ -2065,9 +2072,77 @@ impl Function {
...
+                let mut operands = VecDeque::new();
+                let insn = self.find(insn_id);
+                self.worklist_traverse_single_insn(insn.clone(), &mut operands);
+                for operand in operands {
+                    if !assigned.get(operand) {
+                        let fun_string = format!("{:?}", self);
+                        return Err(ValidationError::OperandNotDefined(fun_string, block, insn_id));
+                    }
+                }
+                assigned.insert(insn_id);
```
tekknolagi: Still only if `insn.has_output()`

zjit/src/hir.rs
```diff
@@ -2065,9 +2072,77 @@ impl Function {
...
+                        let Some(block_in) = assigned_in[target.target.0].as_mut() else {
+                            let fun_string = format!("{:?}", self);
+                            return Err(ValidationError::JumpTargetNotInRPO(fun_string, target.target));
+                        };
+                        // jump target's block_in was modified, we need to queue the block for processing.
+                        if block_in.intersect_with(&assigned) {
+                            worklist.push_back(target.target);
+                        }
+                    }
+                    _ => {
```
```suggestion
                    _ if insn.has_output() => {
```

zjit/src/hir.rs
tekknolagi: Good to add a test for using an instruction that has no output as an operand

zjit/src/hir.rs
```diff
@@ -3199,6 +3274,39 @@ mod validation_tests {
...
+        let entry = function.entry_block;
+        // Create an instruction without making it belong to anything.
+        let dangling = function.new_insn(Insn::Const{val: Const::CBool(true)});
+        let val = function.push_insn(function.entry_block, Insn::ArrayDup { val: dangling, state: InsnId(0usize) });
+        let fun_string = format!("{:?}", function);
+        assert_matches_err(function.validate_definite_assignment(), ValidationError::OperandNotDefined(fun_string, entry, val));
+    }
+
+    #[test]
+    fn not_dominated_by_diamond() {
```
tekknolagi, line 3321: can you add some success cases? for example, having a definition on both sides of a diamond

zjit/src/hir.rs
```diff
@@ -3199,6 +3274,39 @@ mod validation_tests {
...
+        let val = function.push_insn(entry, Insn::Const { val: Const::CBool(false) });
+        function.push_insn(entry, Insn::IfFalse { val, target: BranchEdge { target: side, args: vec![] } });
+        let v1 = function.push_insn(entry, Insn::Const { val: Const::Value(VALUE::fixnum_from_usize(4)) });
+        function.push_insn(entry, Insn::Jump(BranchEdge { target: exit, args: vec![v1] }));
+        let param = function.push_insn(exit, Insn::Param { idx: 0 });
+        let val = function.push_insn(exit, Insn::ArrayDup { val: v0, state: InsnId(0usize) });
+        crate::cruby::with_rubyvm(|| {
+            function.infer_types();
+        });
+        let fun_string = format!("{:?}", function);
```
```suggestion
            function.infer_types();
            let fun_string = format!("{:?}", function);
        });
```

tekknolagi (approved):
looking really excellent, thanks for being open to changes

zjit/src/hir.rs
Fidget-Spinner: This caused the tests to fail and I found out why: things can point to `Insn::Snapshot { .. }`, but it's marked as no output in the has_output function.

tekknolagi: Ah, then it should be marked as having an output!

tekknolagi: That can/should be stacked as a separate commit in the PR

zjit/src/hir.rs
Fidget-Spinner: Added in latest commit.

zjit/src/hir.rs
Fidget-Spinner: I'm not sure if this change is right?

zjit/src/hir.rs
Fidget-Spinner: Added a TODO and a temporary workaround.

Fidget-Spinner: Added test in latest commit.

zjit/src/hir.rs
tekknolagi: eh, sure, the variable defn doesn't reach, but i am trying to make the point that i think formatting expects the ruby vm to be around. it will still be around now that we've called the function but potentially we should move more into the closure

zjit/src/bitset.rs
```diff
@@ -27,12 +28,33 @@ impl<T: Into<usize> + Copy> BitSet<T> {
...
         debug_assert!(idx.into() < self.num_bits);
         let entry_idx = idx.into() / ENTRY_NUM_BITS;
         let bit_idx = idx.into() % ENTRY_NUM_BITS;
         (self.entries[entry_idx] & (1 << bit_idx)) != 0
     }
+
+    /// Modify `self` to only have bits set if they are also set in `other`. Returns true if `self`
+    /// was modified, and false otherwise.
+    /// `self` and `other` must have the same number of bits.
+    pub fn intersect_with(&mut self, other: &Self) -> bool {
```
tekknolagi, line 48: Please add tests

zjit/src/bitset.rs
```diff
@@ -27,12 +28,33 @@ impl<T: Into<usize> + Copy> BitSet<T> {
         newly_inserted
     }
 
+    /// Set all bits to 1.
+    pub fn insert_all(&mut self) {
```
tekknolagi, line 32: Please add tests

zjit/src/hir.rs
```diff
@@ -2065,9 +2072,81 @@ impl Function {
...
+        worklist.push_back(self.entry_block);
+        while let Some(block) = worklist.pop_front() {
+            let mut assigned = assigned_in[block.0].clone().unwrap();
+            for &param in &self.blocks[block.0].params {
+                assigned.insert(param);
+            }
+            for &insn_id in &self.blocks[block.0].insns {
+                let insn_id = self.union_find.borrow().find_const(insn_id);
+                let insn = self.find(insn_id);
+                match insn {
```
```suggestion
                let insn_id = self.union_find.borrow().find_const(insn_id);
                match self.find(insn_id) {
```

zjit/src/hir.rs
```diff
@@ -2065,9 +2072,81 @@ impl Function {
...
+                            let fun_string = format!("{:?}", self);
+                            return Err(ValidationError::JumpTargetNotInRPO(fun_string, target.target));
+                        };
+                        // jump target's block_in was modified, we need to queue the block for processing.
+                        if block_in.intersect_with(&assigned) {
+                            worklist.push_back(target.target);
+                        }
+                    }
+                    // TODO fix has_output to include snapshots.
+                    _ if insn.has_output() || matches!(insn, Insn::Snapshot {..}) => {
```
```suggestion
                    insn if insn.has_output() || matches!(insn, Insn::Snapshot {..}) => {
```

zjit/src/hir.rs
```diff
@@ -2065,9 +2072,81 @@ impl Function {
...
+                let insn = self.find(unioned_insn_id);
+                self.worklist_traverse_single_insn(insn.clone(), &mut operands);
+                for operand in operands {
+                    if !assigned.get(operand) {
+                        return Err(ValidationError::OperandNotDefined(format!("{:?}", self), block, unioned_insn_id, operand));
+                    }
+                }
+                if insn.has_output() || matches!(insn, Insn::Snapshot {..}) {
+                    assigned.insert(insn_id);
+                }
```
```suggestion
                let insn_id = self.union_find.borrow().find_const(insn_id);
                let mut operands = VecDeque::new();
                let insn = self.find(insn_id);
                self.worklist_traverse_single_insn(&insn, &mut operands);
                for operand in operands {
                    if !assigned.get(operand) {
                        return Err(ValidationError::OperandNotDefined(format!("{:?}", self), block, insn_id, operand));
                    }
                }
                if insn.has_output() || matches!(insn, Insn::Snapshot {..}) {
                    assigned.insert(insn_id);
                }
```

tekknolagi: We want to always report about the union-find representative

Also, we don't want to make unnecessary clones

zjit/src/hir.rs
```diff
@@ -3199,6 +3278,75 @@ mod validation_tests {
...
+    fn dominated_by_diamond() {
+        // This tests that both branches with a definition succeeds.
+        let mut function = Function::new(std::ptr::null());
+        let entry = function.entry_block;
+        let side = function.new_block();
+        let exit = function.new_block();
+        let param1 = function.push_insn(side, Insn::Param { idx: 1 });
+        let v0 = function.push_insn(side, Insn::Const { val: Const::Value(VALUE::fixnum_from_usize(3)) });
+        function.push_insn_id(entry, v0);
+        function.push_insn(side, Insn::Jump(BranchEdge { target: exit, args: vec![param1] }));
```
tekknolagi: This is testing that things get passed by parameter but not testing that in A -> B, A -> C, B -> D, C -> D situation that D can use an InsnId from A. If that's intentional due to backend constraints right now, great --- please document it and add a TODO to allow "reaching backward" to dominating blocks. But for that we wouldn't need dataflow analysis at all; it would all be block-local. So instead I think you should test something that doesn't use block params

tekknolagi (commented):
nearly there

zjit/src/hir.rs
Fidget-Spinner: I can't pass the Insn here without cloning as it is consumed by the `worklist_traverse_single_insn`.

However, I realised it's probably cheaper to just find another `self.find` lookup which will only cost one union find and (IIUC) not allocate anything.

tekknolagi: We can pass a reference to it though

zjit/src/bitset.rs
```diff
@@ -68,4 +90,27 @@ mod tests {
...
+        let mut set = BitSet::with_capacity(4);
+        let zero_mask = BitSet::with_capacity(4);
+        let mut one_mask: BitSet<usize> = BitSet::with_capacity(4);
+        one_mask.insert_all();
+        assert_eq!(set.intersect_with(&zero_mask), false);
+        set.insert(0usize);
+        assert_eq!(set.intersect_with(&one_mask), false);
+        assert_eq!(set.intersect_with(&zero_mask), true);
+        assert_eq!(set.get(0usize), false);
+    }
```
tekknolagi: So this test, as written, could be testing that `intersect_with` clears the `BitSet`. Let's do this instead
```suggestion
    #[test]
    #[should_panic]
    fn intersect_with_panics_with_different_num_bits() {
        let mut left: BitSet<usize> = BitSet::with_capacity(3);
        let right = BitSet::with_capacity(4);
        left.intersect_with(&right);
    }

    #[test]
    fn intersect_with_keeps_only_common_bits() {
        let mut left = BitSet::with_capacity(3);
        let mut right = BitSet::with_capacity(3);
        left.insert(0usize);
        left.insert(1usize);
        right.insert(1usize);
        right.insert(2usize);
        left.intersect_with(&right);
        assert_eq!(left.get(0usize), false);
        assert_eq!(left.get(1usize), true);
        assert_eq!(left.get(2usize), false);
    }
```

zjit/src/hir.rs
```diff
@@ -3215,6 +3293,73 @@ mod validation_tests {
...
+        function.push_insn_id(entry, v0);
+        function.push_insn(side, Insn::Jump(BranchEdge { target: exit, args: vec![] }));
+        let val = function.push_insn(entry, Insn::Const { val: Const::CBool(false) });
+        function.push_insn(entry, Insn::IfFalse { val, target: BranchEdge { target: side, args: vec![] } });
+        function.push_insn(entry, Insn::Jump(BranchEdge { target: exit, args: vec![] }));
+        let _val = function.push_insn(exit, Insn::ArrayDup { val: v0, state: v0 });
+        crate::cruby::with_rubyvm(|| {
+            function.infer_types();
+            // Just checking that we don't panic.
+            function.validate_definite_assignment().unwrap();
```
tekknolagi: assert!(...is_ok()), generally, I think

zjit/src/hir.rs
```diff
@@ -3215,6 +3293,73 @@ mod validation_tests {
...
+
+    #[test]
+    fn dominated_by_diamond() {
+        // This tests that both branches with a definition succeeds.
+        let mut function = Function::new(std::ptr::null());
+        let entry = function.entry_block;
+        let side = function.new_block();
+        let exit = function.new_block();
+        let v0 = function.push_insn(side, Insn::Const { val: Const::Value(VALUE::fixnum_from_usize(3)) });
+        function.push_insn_id(entry, v0);
```
tekknolagi: This v0 should not appear in two blocks

(also, separately, we should have a validator that each insn only appears once per CFG)

tekknolagi: v0 should probably remain in entry

tekknolagi: https://github.com/ruby/ruby/pull/13846

zjit/src/hir.rs
Fidget-Spinner: worklist_traverse_single_insn takes stuff out of Insn and adds it to a worklist, so we can't.

tekknolagi: We totally can; `InsnId` are copyable

tekknolagi: I'll make a patch

tekknolagi: 705c84034e

Fidget-Spinner: Interesting, I was wrong then! Thanks!

tekknolagi:
Amazing, thank you!! This will be very helpful

Fidget-Spinner:
PR merged in 214983bd9be88903833558043a20ba1c2469c333. Not sure why it doesn't show merged, but closing this PR now.

PR #13815, opened 2025-07-08
Fix a few small race conditions inserting into the shape tree

jhawthorn:


shape.c
```diff
@@ -841,6 +844,9 @@ rb_shape_get_next_iv_shape(shape_id_t shape_id, ID id)
 {
     rb_shape_t *shape = RSHAPE(shape_id);
     rb_shape_t *next_shape = shape_get_next_iv_shape(shape, id);
+    if (!next_shape) {
+        return INVALID_SHAPE_ID;
```
jhawthorn, line 848: I'm not totally sure about this. I guess it would have been an rb_bug before?

shape.h
```diff
@@ -122,6 +122,12 @@ RUBY_SYMBOL_EXPORT_BEGIN
 RUBY_EXTERN rb_shape_tree_t rb_shape_tree;
 RUBY_SYMBOL_EXPORT_END
 
+static inline shape_id_t
+rb_shape_get_shape_count(void)
```
byroot: nitpick:
```suggestion
rb_shapes_count(void)
```

Much shorter, avoid the repetition etc.

jhawthorn: Good call

PR #13818, opened 2025-07-08
Bump up resolv-0.6.2 for Ruby 3.4

hsbt:
@k0kubun resolv-0.6.2 fixed https://www.ruby-lang.org/en/news/2025/07/08/dos-resolv-cve-2025-24294/

But that contained a bit for feature changes for Windows. I rewrite the feature about the DNS server with Powershell if failing to load `win32-registry`.

If you want to only vulnerability changes for Ruby 3.4, I will create and release resolv-0.6.0.1 for that.

### DIFF
--- a/ext/win32/lib/win32/resolv.rb
+++ b/ext/win32/lib/win32/resolv.rb
@@ -4,13 +4,8 @@
 
 =end
 
-require 'win32/registry'
-
 module Win32
   module Resolv
-    API = Registry::API
-    Error = Registry::Error
-
     def self.get_hosts_path
       path = get_hosts_dir
       path = File.expand_path('hosts', path)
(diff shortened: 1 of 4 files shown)


k0kubun (approved):

k0kubun:
> If you want to only vulnerability changes for Ruby 3.4, I will create and release resolv-0.6.0.1 for that.

We could publish resolv-0.6.0.1 whenever somebody requests that.

Because we can do that afterwards for gems, the next 3.4 release will have resolv-0.6.2 (unless we end up releasing resolv-0.6.0.1 first) and other regular changes. FYI, we've planned on releasing 3.4.5 in July https://www.ruby-lang.org/en/news/2025/05/14/ruby-3-4-4-released/, and it's scheduled next Monday in my calendar.

PR #13820, opened 2025-07-08
Fix unused variable warnings in default.c in modular GC

peterzhu2118:
The asan and valgrind macros when BUILDING_MODULAR_GC don't use the variables which could the compiler to emit unused variable warnings.


nobu:
Casted macro arguments should be parenthesized.

PR #13821, opened 2025-07-08
Make `defined? (x;)` return `expression` when using parse.y parser

S-H-GAMELINKS:
Follow up [Bug #21029].

Currently,  `defined? (x;)` returns `expression` when using Prism parser.

See:
- https://github.com/ruby/ruby/pull/12949
- https://bugs.ruby-lang.org/issues/21029

However, `defined? (x;)` returns nil when using parse.y, as reported in bug ticket comment and test-all (when using parse.y parser) test result.

This change adds a context flag to track trailing semicolons in defined? scope. When a trailing semicolon is detected within a defined? scope, the generated AST node is wrapped with NODE_BLOCK. This change ensures consistent behavior with `defined? (;x)` .

However, I have a little skeptical about whether this change is a good idea. If any other good suggestions, I would appreciate it if you could comment.

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -314,6 +314,7 @@ struct lex_context {
     unsigned int in_argdef: 1;
     unsigned int in_def: 1;
     unsigned int in_class: 1;
+    unsigned int has_trailing_semicolon: 1;
     BITFIELD(enum rb_parser_shareability, shareable_constant_value, 2);
     BITFIELD(enum rescue_context, in_rescue, 2);
     unsigned int cant_return: 1;
@@ -4041,6 +4042,7 @@ arg		: asgn(arg_rhs)
                     {
                         p->ctxt.in_defined = $3.in_defined;
                         $$ = new_defined(p, $4, &@$);
+                        p->ctxt.has_trailing_semicolon = $3.has_trailing_semicolon;
                     /*% ripper: defined!($:4) %*/
                     }
                 | def_endless_method(endless_arg)
@@ -4428,6 +4430,7 @@ primary		: inline_primary
                 {
                     p->ctxt.in_defined = $4.in_defined;
                     $$ = new_defined(p, $5, &@$);
+                    p->ctxt.has_trailing_semicolon = $4.has_trailing_semicolon;
                 /*% ripper: defined!($:5) %*/
                 }
             | keyword_not '(' expr rparen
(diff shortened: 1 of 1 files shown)


parse.y
```diff
@@ -2970,7 +2971,7 @@ rb_parser_ary_free(rb_parser_t *p, rb_parser_ary_t *ary)
 
 %rule compstmt(stmts) <node>
                     : stmts terms?
-                        {
+		    {
```
ydah: Is this change intentional?

parse.y
```diff
@@ -13013,6 +13021,9 @@ kwd_append(rb_node_kw_arg_t *kwlist, rb_node_kw_arg_t *kw)
 static NODE *
 new_defined(struct parser_params *p, NODE *expr, const YYLTYPE *loc)
 {
+    int had_trailing_semicolon = p->ctxt.has_trailing_semicolon;
+    p->ctxt.has_trailing_semicolon = 0;
```
nobu, line 13027: I suspect that the `has_trailing_semicolon` saved at `begin_defined` should be restored after `new_defined()` calls.

parse.y
S-H-GAMELINKS: Not intentional.
I'll fix it.

parse.y
S-H-GAMELINKS: Are you suggesting that I should either remove this line where I assign 0 to `has_trailing_semicolon`, or restore the original value by assigning `had_trailing_semicolon` back to `has_trailing_semicolon`?
If so, I'll remove this initialization as it was added by mistake.

nobu: I suggest that `has_trailing_semicolon` should be restored from the context saved at `begin_defined`, after `new_defined()` is called.

```diff
diff --git a/parse.y b/parse.y
index 3138061b988..a92df0ae81f 100644
--- a/parse.y
+++ b/parse.y
@@ -4041,6 +4041,7 @@ arg		: asgn(arg_rhs)
                     {
                         p->ctxt.in_defined = $3.in_defined;
                         $$ = new_defined(p, $4, &@$);
+                        p->ctxt.has_trailing_semicolon = $3.has_trailing_semicolon;
                     /*% ripper: defined!($:4) %*/
                     }
                 | def_endless_method(endless_arg)
@@ -4428,6 +4429,7 @@ primary		: inline_primary
                 {
                     p->ctxt.in_defined = $4.in_defined;
                     $$ = new_defined(p, $5, &@$);
+                    p->ctxt.has_trailing_semicolon = $4.has_trailing_semicolon;
                 /*% ripper: defined!($:5) %*/
                 }
             | keyword_not '(' expr rparen
```

Or, pass the `struct lex_context` to `new_defined` and restore it and `in_defined` together there?

S-H-GAMELINKS: Thank you for the reply and I understand.
I've restore `has_trailing_semicolon` after `new_defined()`.

> Or, pass the struct lex_context to new_defined and restore it and in_defined together there?

I think it would be better not to do that.
Instead, I thought it's better to restore `has_trailing_semicolon` after calling `new_defined()`.
If we pass the `struct lex_context` to `new_defined()`, we might need to create a temporary copy or something similar.
It would be simpler and better to just reassign the value in the action part.

PR #13822, opened 2025-07-08
ZJIT: Add BitSet and use in HIR

tekknolagi:
This reduces memory usage by ~8x for some very short-lived data structures.

### DIFF
--- a/zjit/src/bitset.rs
+++ b/zjit/src/bitset.rs
@@ -1,0 +1,71 @@
+type Entry = u128;
+
+const ENTRY_NUM_BITS: usize = Entry::BITS as usize;
+
+// TODO(max): Make a `SmallBitSet` and `LargeBitSet` and switch between them if `num_bits` fits in
+// `Entry`.
+pub struct BitSet<T: Into<usize> + Copy> {
+    entries: Vec<Entry>,
+    num_bits: usize,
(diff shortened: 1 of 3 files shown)


k0kubun (approved):

zjit/src/bitset.rs
```diff
@@ -0,0 +1,70 @@
...
+pub struct BitSet<T: Into<usize> + Copy> {
+    storage: Vec<Entry>,
+    num_bits: usize,
+    phantom: std::marker::PhantomData<T>,
+}
+
+impl<T: Into<usize> + Copy> BitSet<T> {
+    pub fn with_capacity(num_bits: usize) -> Self {
+        // +1 because we are rounding down
+        let num_entries = num_bits / (Entry::BITS as usize) + 1;
```
st0012: Does it make sense to store `Entry::BITS as usize` in a constant?

st0012: Also, maybe this is simpler?
```suggestion
        let num_entries = num_bits.div_ceil(Entry::BITS as usize);
```

tekknolagi: Re: constant: eh, maybe

Re: div_ceil: ooh, nice, thank you

tekknolagi:
Merged in c691095f2e.

PR #13825, opened 2025-07-08
ZJIT: Profile `opt_nil_p` instruction

st0012:
1. Enable profiling for `nil?` calls
    - Added `zjit_profile` attribute to opt_nil_p instruction
    - Updated profiler to track `nil?` call receivers
    - Changed to use `runtime_exact_ruby_class` for more accurate type profiling
  2. Add comprehensive test coverage
    - Tests for type guard generation with all major Ruby types:
        - nil → returns true
      - false, true, symbol, fixnum, float, string → returns false
    - Verifies correct guard types and CCall optimizations


st0012:
Yeah that's the plan. Can't add them before #13797 or they'd panic and fail.

tekknolagi:
Merged in 342ada1546.

PR #13828, opened 2025-07-09
Skip test_rm_r_no_permissions test under root

jiajia123-wind:
Skip the test_rm_r_no_permissions test under the root user, as deletion always succeeds.


nobu:
Please send to the upstream, https://github.com/ruby/fileutils

jiajia123-wind:
Thank you for the reminder. I've resubmitted the patch here:
https://github.com/ruby/fileutils/pull/133

PR #13830, opened 2025-07-09
ZJIT: Profile `opt_and` and annotate `&` method for `true` and `false`

st0012:


k0kubun (approved):

k0kubun:
I'm not sure why the test cases are focused on testing true and false because we never use `&` or `|` for booleans. It's a bit-wise operator, which is normally used for Integers (or Arrays, which YARV and YJIT don't optimize). We usually don't see the instruction in Rails-like benchmarks, but the most typical workload for `opt_and` and `opt_or` is, Optcarrot.

k0kubun:
> I'm not sure why the test cases are focused on testing true and false

Oh, because that's what's being annotated here, nvm.

I don't think we want to pay the maintenance cost for the methods which we aren't sure if are used. Did you see the methods in any benchmark of yjit-bench?

Profiling `opt_and` and `opt_or` is fine (it seems fine to do those two in a single PR btw), but the goal should be something that is actually used in benchmarks or production, or the ones that the interpreter optimizes (so that enabling ZJIT will not slow it down). So maybe consider optimizing `Integer` instead? YARV optimizes only `Integer` on `opt_and`/`opt_or`, and we know it's used by a benchmark we kinda care about.

test/ruby/test_zjit.rb
```diff
@@ -330,6 +330,42 @@ def test(x, y) = x | y
...
+  def test_fixnum_and
+    assert_compiles '1', %q{
+      def test(a, b) = a & b
+      test(2, 2)
+      test(2, 2)
+      test(5, 3)
+    }, call_threshold: 2, insns: [:opt_and]
+  end
+
+  def test_fixnum_and_fallthrough
```
tekknolagi: Nit: "fallthrough" isn't a thing; it's side-exit

zjit/src/hir.rs
```diff
@@ -5033,6 +5049,42 @@ mod opt_tests {
         "#]]);
     }
 
+    #[test]
+    fn test_fold_fixnum_and() {
```
tekknolagi: It doesn't fold, though. You would need to add an optimizer case for that in the constant folding pass

tekknolagi (approved):
very nice

zjit/src/hir.rs
st0012: Ah right I confused myself. I don't think these tests are super useful then given we already have guard tests. I'll just remove them for now.

test/ruby/test_zjit.rb
st0012: I added a commit to update these and other `_fallthrough` names to `_side_exit`.

PR #13839, opened 2025-07-09
[DOC] Tweaks for String#count

BurdetteLamar:
I'd like to get rid of file `doc/character_selectors.rdoc`.  I think it's better for the methods' documentation to handle things on their own (as I've shown here).

Right now, it's also referred to at:

- `#delete`/`delete!`.
- `#squeeze/#squeeze!`.

When these are all 'tweaked', those references will be gone and we can delete the file.

### DIFF
--- a/doc/string/count.rdoc
+++ b/doc/string/count.rdoc
@@ -1,0 +1,50 @@
+Returns the total number of characters in +self+ that are specified by the given selectors.
+
+For one 1-character selector,
+returns the count of instances of that character:
+
+  s = 'abracadabra'
+  s.count('a') # => 5
+  s.count('b') # => 2
+  s.count('x') # => 0
+  s.count('')  # => 0
+
+  s = 'тест'
+  s.count('т')  # => 2
+  s.count('е')  # => 1
+
+  s = 'よろしくお願いします'
+  s.count('よ')  # => 1
+  s.count('し')  # => 2
+
+For one multi-character selector,
+returns the count of instances for all specified characters:
+
+  s = 'abracadabra'
+  s.count('ab')     # => 7
+  s.count('abc')    # => 8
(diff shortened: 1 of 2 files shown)


doc/string/count.rdoc
```diff
@@ -0,0 +1,50 @@
...
+
+  s = 'abcdefg'
+
+  s.count('abcde', 'dcbfg') # => 3
+  s.count('bcd')            # => 3
+
+  s.count('abc', 'def')     # => 0
+  s.count('')               # => 0
+
+Selectors may include negations, ranges, and escapes:
```
peterzhu2118: What do you mean by "escapes"? Like `"\n"`? That's a feature of the string rather than `String#count`.

BurdetteLamar: Not sure what "escapes" is supposed to mean.  It's mentioned in the previous documentation, though.  What do you thinki?

peterzhu2118: Looking at the original documentation, it means escaping the special characters like `^` and `-`

```ruby
"1^2^3".count("\\^1") # => 3
"^-^".count("\\^\\-") # => 3
```

BurdetteLamar:
@peterzhu2118, on second thought, maybe it's better to retain file `doc/character_selectors.rdoc`.  I was mistaken in thinking how few methods use it (there are four more than I realized).

Maybe review that file for possible improvements, and add/revise the examples in its nine referring methods?

If agreed, I'll close this PR.

peterzhu2118:
It's your call, but I think there's only 3 other places that use it (`String#tr`, `String#delete`, `String#squeeze`).

BurdetteLamar:
I'm rethinking this.

BurdetteLamar:
Re-rethinking this.  And marking as draft.

PR #13840, opened 2025-07-09
ZJIT: Avoid optimizing locals on eval

k0kubun:


zjit/src/hir.rs
```diff
@@ -2682,37 +2682,47 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     break;  // Don't enqueue the next block as a successor
                 }
                 YARVINSN_getlocal_WC_0 => {
-                    // TODO(alan): This implementation doesn't read from EP, so will miss writes
```
tekknolagi, line 2685: Please add some HIR parsing tests

tekknolagi (approved):
lgtm with tests

zjit/src/hir.rs
k0kubun: Could you give me some example code to do so? I'm not sure how to extract an ISEQ pointer (or `RubyVM::InstructionSequence`) out of an eval, which has `ISEQ_TYPE_EVAL`.

tekknolagi: I was thinking we'd capture stdout with `--zjit-dump-hir` or something but that's pretty different from what we're doing now... maybe later we can have a hook that does something for each compiled function, like adding the printed HIR to a hashmap


I suppose we will leave it for now

zjit/src/hir.rs
```diff
@@ -2682,37 +2682,47 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
-                    // from nested ISeqs. This will need to be amended when we add codegen for
-                    // Send.
-                    let ep_offset = get_arg(pc, 0).as_u32();
-                    let val = state.getlocal(ep_offset);
-                    state.stack_push(val);
+                    if iseq_type == ISEQ_TYPE_EVAL {
+                        let ep_offset = get_arg(pc, 0).as_u32();
+                        state.stack_push(fun.push_insn(block, Insn::GetLocal { ep_offset, level: 0 }));
```
tekknolagi, line 2688: If something side-exits after this, will we write a nil to the frame for each local? I think the answer is yes (so we should still write to our FrameState)

k0kubun: Fixed in https://github.com/ruby/ruby/pull/13840/commits/75dc66d645d9fe3c6f16e32c61b50e00ee97ab61

PR #13844, opened 2025-07-10
Make `rb_enc_autoload_p` atomic

casperisfine:
Using `encoding->max_enc_len` as a way to check if the encoding has been loaded isn't atomic, because it's not atomically set last.
    
Instead we can use a dedicated atomic value inside the encoding table.

### DIFF
--- a/encoding.c
+++ b/encoding.c
@@ -24,6 +24,7 @@
 #include "internal/string.h"
 #include "internal/vm.h"
 #include "regenc.h"
+#include "ruby/atomic.h"
 #include "ruby/encoding.h"
 #include "ruby/util.h"
 #include "ruby_assert.h"
@@ -60,6 +61,7 @@ VALUE rb_cEncoding;
 static VALUE rb_encoding_list;
 
 struct rb_encoding_entry {
+    rb_atomic_t loaded;
     const char *name;
     rb_encoding *enc;
     rb_encoding *base;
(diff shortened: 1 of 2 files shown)


encoding.c
```diff
@@ -810,11 +824,21 @@ rb_enc_autoload(rb_encoding *enc)
     return i;
 }
 
+bool
+rb_enc_autoload_p(rb_encoding *enc)
+{
+    int idx = ENC_TO_ENCINDEX(enc);
+    return !RUBY_ATOMIC_LOAD(global_enc_table.list[idx].loaded);
```
nobu: Is it possible that `enc` and `&global_enc_table.list[idx]` are different?

casperisfine: I don't think so, but I can add an assertion for that.

encoding.c
```diff
@@ -810,11 +824,22 @@ rb_enc_autoload(rb_encoding *enc)
     return i;
 }
 
+bool
+rb_enc_autoload_p(rb_encoding *enc)
+{
+    int idx = ENC_TO_ENCINDEX(enc);
+    RUBY_ASSERT(rb_enc_from_index(idx) == enc);
```
casperisfine, line 831: FYI: @nobu

nobu: If these are equal, can't the next line use `enc->loaded` simply?

byroot: `enc->loaded` doesn't exist. `loaded` is stored in `struct rb_encoding_entry` not in `rb_encoding`.

Putting it in `rb_encoding` would be much simpler, but I assumed patching `onigmo` wasn't acceptable.

PR #13845, opened 2025-07-10
[DOC] Remove documentation for RDoc

peterzhu2118:
RDoc has been moved from a default gem to a bundled gem, which means that the source code is no longer in the ruby/ruby repository. We should remove all documentation for RDoc.


BurdetteLamar (approved):

nobu:
How about moving duplicate links to footnotes?

PR #13850, opened 2025-07-10
ZJIT: Implement patch points on BOP redefinition

k0kubun:
This PR compiles a side exit for patch points and rewrites them with a jump to the side exit on BOP redefinition.

To make sure these jump instructions overwrite each other, nop instructions are generated in between them. When multiple adjacent patch points have the same FrameState, we could merge them together to reduce the number of nop instructions, but it's out of scope in this PR.


tekknolagi (approved):
Let's go!!

PR #13852, opened 2025-07-10
ZJIT: Print a message about ZJIT_RB_BUG when unused

k0kubun:
This PR improves a message of the panic hook https://github.com/Shopify/zjit/pull/43.

When Rust panics on non-release builds, we do not show Rust and Ruby backtraces by default. You need to give `RUST_BACKTRACE=1` and/or `ZJIT_RB_BUG=1` to turn on each of them. The `RUST_BACKTRACE=1` is already instructed in the panic message, but `ZJIT_RB_BUG=1` isn't. So this PR changes it to print a message like:

```
thread '<unnamed>' panicked at zjit/src/codegen.rs:141:5:
...
note: run with `ZJIT_RB_BUG=1` environment variable to display a Ruby backtrace
ZJIT panicked while holding VM lock acquired at zjit/src/codegen.rs:86. Aborting...
```


tekknolagi (approved):
Oh, nice

PR #13862, opened 2025-07-11
ZJIT: Mark objects baked in JIT code

k0kubun:
This PR updates `rb_zjit_iseq_mark` to mark objects in JIT code. It also implements `rb_zjit_iseq_update_references` to avoid pinning objects in `rb_zjit_iseq_mark`.

It was a TODO in YJIT that we want to use `CodePtr` instead of `u32` for GC offsets. So I updated `compile*()` functions to use `CodePtr`. While doing so, I noticed there's some leftover code for retrying YJIT's code layout switches. Because we no longer have the concept of code layout splitting, I cleaned up the code as well.

### DIFF
--- a/zjit/src/gc.rs
+++ b/zjit/src/gc.rs
@@ -1,7 +1,7 @@
 // This module is responsible for marking/moving objects on GC.
 
 use std::ffi::c_void;
-use crate::{cruby::*, profile::IseqProfile, virtualmem::CodePtr};
+use crate::{cruby::*, profile::IseqProfile, state::ZJITState, virtualmem::CodePtr};
 
 /// This is all the data ZJIT stores on an ISEQ. We mark objects in this struct on GC.
 #[derive(Debug)]
@@ -12,12 +12,17 @@ pub struct IseqPayload {
     /// JIT code address of the first block
     pub start_ptr: Option<CodePtr>,
 
-    // TODO: Add references to GC offsets in JIT code
+    /// GC offsets of the JIT code. These are the addresses of objects that need to be marked.
+    pub gc_offsets: Vec<CodePtr>,
 }
 
 impl IseqPayload {
     fn new(iseq_size: u32) -> Self {
-        Self { profile: IseqProfile::new(iseq_size), start_ptr: None }
+        Self {
+            profile: IseqProfile::new(iseq_size),
+            start_ptr: None,
+            gc_offsets: vec![],
+        }
     }
 }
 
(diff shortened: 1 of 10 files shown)


zjit/src/gc.rs
```diff
@@ -66,17 +71,65 @@ pub extern "C" fn rb_zjit_iseq_mark(payload: *mut c_void) {
...
+    // Mark objects baked in JIT code
+    let cb = ZJITState::get_code_block();
+    for &offset in payload.gc_offsets.iter() {
+        let value_ptr: *const u8 = offset.raw_ptr(cb);
+        // Creating an unaligned pointer is well defined unlike in C.
+        let value_ptr = value_ptr as *const VALUE;
+
+        unsafe {
+            let object = value_ptr.read_unaligned();
+            rb_gc_mark_movable(object);
```
tekknolagi, line 88: Does this mean that we need to mark all code pages W on GC and then X again after?

zjit/src/hir_type/mod.rs
```diff
@@ -288,6 +288,14 @@ impl Type {
         }
     }
 
+    /// Return a mutable reference to the object specialization, if any.
+    pub fn ruby_object_mut(&mut self) -> Option<&mut VALUE> {
```
tekknolagi: This should also get Type and TypeExact cases and probably be documented as only being used for GC

zjit/src/gc.rs
k0kubun: I'm not sure if I follow what you're asking. `read_unaligned()` doesn't write anything to the code region, and `rb_gc_mark_movable()` only marks an object on the Ruby heap, which doesn't impact the code region.

This doesn't mark code pages writable, so it doesn't need to call `mark_all_executable()` afterwards.

zjit/src/hir_type/mod.rs
k0kubun: good catch, thanks https://github.com/ruby/ruby/pull/13862/commits/e63bac9a9194a77479808b249b0ff802fdb155db

zjit/src/gc.rs
tekknolagi: Is the Ruby GC moving? If so, it'll want to update embedded pointers in the code page(s). If so, we'll have to mark them writable on GC.

k0kubun: Not moving by default.

If you look at the caller of these hooks, you can see we call `rb_zjit_iseq_update_references()` instead of `rb_zjit_iseq_mark()` when we indeed move objects.

PR #13875, opened 2025-07-14
ZJIT: Support invalidating on method redefinition

st0012:
This commit adds support for the MethodRedefined invariant to be invalidated when a method is redefined.

Changes:
- Added CME pointer to the `MethodRedefined` invariant in HIR
- Updated all places where `MethodRedefined` invariants are created to include the CME pointer
- Added handling for `MethodRedefined` invariants in `gen_patch_point` to call `track_cme_assumption`, which registers the patch point for invalidation when `rb_zjit_cme_invalidate` is called

This ensures that when a method is redefined, all JIT code that depends on that method will be properly invalidated.

### DIFF
--- a/vm_method.c
+++ b/vm_method.c
@@ -122,6 +122,7 @@ vm_cme_invalidate(rb_callable_method_entry_t *cme)
     RB_DEBUG_COUNTER_INC(cc_cme_invalidate);
 
     rb_yjit_cme_invalidate(cme);
+    rb_zjit_cme_invalidate(cme);
 }
 
 static int
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -127,6 +127,8 @@ pub enum Invariant {
         klass: VALUE,
         /// The method ID of the method we want to assume unchanged
         method: ID,
+        /// The callable method entry that we want to track
+        cme: *const rb_callable_method_entry_t,
     },
     /// A list of constant expression path segments that must have not been written to for the
     /// following code to be valid.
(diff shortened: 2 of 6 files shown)


st0012:
I'm investigating the segfault on ubuntu atm.

zjit/src/hir.rs
```diff
@@ -222,12 +224,13 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
...
-            Invariant::MethodRedefined { klass, method } => {
+            Invariant::MethodRedefined { klass, method, cme } => {
                 let class_name = get_class_name(klass);
-                write!(f, "MethodRedefined({class_name}@{:p}, {}@{:p})",
+                write!(f, "MethodRedefined({class_name}@{:p}, {}@{:p}, cme:{:p})",
                     self.ptr_map.map_ptr(klass.as_ptr::<VALUE>()),
                     method.contents_lossy(),
-                    self.ptr_map.map_id(method.0)
+                    self.ptr_map.map_id(method.0),
+                    self.ptr_map.map_ptr(cme.cast::<u8>())
```
tekknolagi: does map_ptr not work on arbitrary pointers?

zjit/src/invariants.rs
```diff
@@ -99,3 +103,53 @@ pub fn track_bop_assumption(
...
+                    let mut asm = Assembler::new();
+                    asm_comment!(asm, "CME is invalidated: {:?}", cme);
+                    asm.jmp(jump.to.into());
+                    asm.compile(cb).expect("can write existing code");
+                });
+            }
+            cb.mark_all_executable();
+        }
+
+        // Also invalidate the compiled code of the method itself
```
tekknolagi: Why invalidate the compiled iseq?

st0012: My implementation is [crashing on Ubuntu](https://github.com/ruby/ruby/actions/runs/16323762653/job/46108291739?pr=13875) and my hypothesis is that it's caused by not cleaning up associated jit pointers in iseq objects. So the change (and the 2nd commit) is to try verifying that.

But now seeing there are [similar build failures on master](https://github.com/ruby/ruby/actions/runs/16325452354/job/46114254086) this may not be the case and I'll drop this change if that's confirmed true.

tekknolagi: If we can have multiple CMEs for a given ISEQ, then we should not invalidate the code. Also, we may have live frames trying to execute this code

tekknolagi:
Try another rebase?

tekknolagi (approved):
Nice, lgtm

vm_method.c
```diff
@@ -122,6 +122,7 @@ vm_cme_invalidate(rb_callable_method_entry_t *cme)
     RB_DEBUG_COUNTER_INC(cc_cme_invalidate);
 
     rb_yjit_cme_invalidate(cme);
+    rb_zjit_cme_invalidate(cme);
```
k0kubun, line 125: There were other existing `rb_yjit_cme_invalidate` calls, and they should have had `rb_zjit_cme_invalidate` too. I'll add that in https://github.com/ruby/ruby/pull/14376 that happens to trigger the problem.

st0012: Ah good catch and sorry 🥲

PR #13883, opened 2025-07-14
[DOC] Tweaks for String#dump

BurdetteLamar:


doc/string/dump.rdoc
```diff
@@ -0,0 +1,26 @@
...
+  s << 8   # Back space.
+  s << 9   # Horizontal tab.
+  s << 10  # Line feed.
+  s << 11  # Vertical tab.
+  s << 12  # Form feed.
+  s << 13  # Carriage return.
+  s        # => "\a\b\t\n\v\f\r"
+  s.dump   # => "\"\\a\\b\\t\\n\\v\\f\\r\""
+
+Multi-byte characters are rendered in unicode notation:
```
nobu: It is only for Unicode encodings.

BurdetteLamar: Does this mean that there are multi-byte characters that are not in Unicode encodings?  If so, I'll need examples.

BurdetteLamar:
@peterzhu2118, I'll need help with this:
- I think I need examples of multi-byte characters that will not dump in Unicode notation.
- Old doc notwithstanding, I found to character than dumps in hexadecimal notation.  Are there any?

peterzhu2118:
> I think I need examples of multi-byte characters that will not dump in Unicode notation.

For example:

```ruby
'тест'.dump # => "\"\\u0442\\u0435\\u0441\\u0442\""
'тест'.encode('utf-16le').dump # => "\"B\\x045\\x04A\\x04B\\x04\".dup.force_encoding(\"UTF-16LE\")"
```

> Old doc notwithstanding, I found to character than dumps in hexadecimal notation. Are there any?

Sorry, I don't understand what you mean by this.

BurdetteLamar:
Thanks, @peterzhu2118.  What you've written above answers both questions (however poorly they're posed).

doc/string/dump.rdoc
BurdetteLamar: Fixed.

doc/string/dump.rdoc
```diff
@@ -0,0 +1,45 @@
...
+  s.encoding                # => #<Encoding:UTF-8>
+  s.dump                    # => "\"\\u0442\\u0435\\u0441\\u0442\""
+  s.encode('utf-16').dump   # => "\"\\xFE\\xFF\\x04B\\x045\\x04A\\x04B\".dup.force_encoding(\"UTF-16\")"
+  s.encode('utf-16le').dump # => "\"B\\x045\\x04A\\x04B\\x04\".dup.force_encoding(\"UTF-16LE\")"
+
+  s = 'こんにちは'
+  s.encoding                # => #<Encoding:UTF-8>
+  s.dump                    # => "\"\\u3053\\u3093\\u306B\\u3061\\u306F\""
+  s.encode('utf-16').dump   # => "\"\\xFE\\xFF0S0\\x930k0a0o\".dup.force_encoding(\"UTF-16\")"
+  s.encode('utf-16le').dump # => "\"S0\\x930k0a0o0\".dup.force_encoding(\"UTF-16LE\")"
```
peterzhu2118, line 43: I think it would be better to move the examples of non-UTF8 encodings to a separate section with some text describing it (e.g. using hexadecimal format and adding `dup.force_encoding(<encoding name>)`. This is because non-UTF8 is more of an edge case rather than a commonly used case.

BurdetteLamar: I've moved the cited lines to the end.  I think you want other changes, but I'm not sure what exactly is needed.  Can you fix up one, as a guide for me?

BurdetteLamar: @peterzhu2118, I'll take another shot at this;  marking as Draft in the interim.

BurdetteLamar: @peterzhu2118, I take it back.  I don't know what to do with this.

peterzhu2118: I opened https://github.com/ruby/ruby/pull/13965

PR #13892, opened 2025-07-15
Update default gems with `-flto` support

hsbt:
from https://bugs.ruby-lang.org/issues/21496

I released and sync the following gems with `-flto` support.

* io-console
* io-wait
* io-nonblock
* etc and `Ractor#take` workaround

They are contained other changes. But I believe these changes are minor and will not cause new issues.

Unfotunately, the following gems couldn't release with `-flto` support because they have a lot of changes from bundled version of Ruby 3.4. So, I only backport `-flto` support for them.

* strscan
* openssl
* json

BTW, we confirm to work all of default gems with GCC 15.1 at Fedora 42 CI at https://rubyci.s3.amazonaws.com/fedora42-arm/ruby-3.4/summary.html.

This issue happend with Gentoo linux https://bugs.ruby-lang.org/issues/21497

### DIFF
--- a/ext/etc/etc.c
+++ b/ext/etc/etc.c
@@ -56,7 +56,7 @@ static VALUE sGroup;
 #endif
 RUBY_EXTERN char *getlogin(void);
 
-#define RUBY_ETC_VERSION "1.4.5"
+#define RUBY_ETC_VERSION "1.4.6"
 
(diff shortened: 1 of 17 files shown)


k0kubun (approved):
Thanks!

PR #13893, opened 2025-07-15
Make Pathname#mkpath builtin

casperisfine:
[[Feature #17473]](https://bugs.ruby-lang.org/issues/17473)

FYI: @hsbt


hsbt:
👍  Thanks!

PR #13899, opened 2025-07-15
ZJIT: Only specialize direct positional-positional calls

tekknolagi:
This is temporary until we have a unified calling convention.


k0kubun (approved):
:+1:

PR #13900, opened 2025-07-15
ZJIT: Fix SP alignment on JIT entry for x86_64

k0kubun:


zjit/src/codegen.rs
```diff
@@ -511,10 +507,6 @@ fn gen_entry_prologue(asm: &mut Assembler, iseq: IseqPtr) {
     asm.cpush(CFP);
     asm.cpush(EC);
     asm.cpush(SP);
-    // On x86_64, maintain 16-byte stack alignment
-    if cfg!(target_arch = "x86_64") {
-        asm.cpush(SP);
-    }
```
XrXr, line 517: Nice, this looked suspect to me, so I'm glad we're fixing it.

zjit/src/codegen.rs
```diff
@@ -1157,12 +1152,19 @@ fn max_num_params(function: &Function) -> usize {
...
+        num_slots + 1
+    } else if cfg!(target_arch = "aarch64") && num_slots % 2 == 1 {
+        // On arm64, the stack pointer is always aligned to 16 bytes, so we need to round up
+        // `num_slots`` to an even number.
         num_slots + 1
+    } else {
+        num_slots
     };
+
+    const _: () = assert!(SIZEOF_VALUE == 8, "aligned_stack_bytes() assumes SIZEOF_VALUE == 8");
```
```suggestion
    const { assert!(SIZEOF_VALUE == 8, "aligned_stack_bytes() assumes SIZEOF_VALUE == 8"); }
```

https://doc.rust-lang.org/reference/expressions/block-expr.html#const-blocks New rust feature we can use now.

tekknolagi: oh wow

k0kubun: cool, thanks! https://github.com/ruby/ruby/pull/13900/commits/c32bdc297e3e400a8dc4fd0d121e65e250d685bc

zjit/src/codegen.rs
```diff
@@ -1157,12 +1152,19 @@ fn max_num_params(function: &Function) -> usize {
...
-        num_slots
-    } else {
+    let num_slots = if cfg!(target_arch = "x86_64") && num_slots % 2 == 0 {
+        // On x86_64, since the call instruction bumps the stack pointer by 8 bytes on entry,
+        // we need to round up `num_slots` to an odd number.
+        num_slots + 1
+    } else if cfg!(target_arch = "aarch64") && num_slots % 2 == 1 {
+        // On arm64, the stack pointer is always aligned to 16 bytes, so we need to round up
+        // `num_slots`` to an even number.
         num_slots + 1
```
amomchilov, line 1162: If we ever need to resurrect this code, it's worth checking if there's a benefit to making it branchless. I'm guessing LLVM could optimize it the same way (and replace that `% 2` with a bitcheck, of course), but worth confirming.

```diff
-    let num_slots = if cfg!(target_arch = "x86_64") {
+    let num_slots = if cfg!(target_arch = "x86_64") && num_slots % 2 == 0 {
-        num_slots + 1
+        num_slots + (1 - num_slots % 2)
-   } else if cfg!(target_arch = "aarch64") {
+   } else if cfg!(target_arch = "aarch64") && num_slots % 2 == 1 {
-        num_slots + 1
+        num_slots + (num_slots % 2)
```

PR #13901, opened 2025-07-15
ZJIT: Profile type+shape distributions

tekknolagi:
ZJIT uses the interpreter to take type profiles of what objects pass through
the code. It stores a compressed record of the history per opcode for the
opcodes we select.

Before this change, we re-used the HIR Type data-structure, a shallow type
lattice, to store historical type information. This was quick for bringup but
is quite lossy as profiles go: we get one bit per built-in type seen, and if we
see a non-built-in type in addition, we end up with BasicObject. Not very
helpful. Additionally, it does not give us any notion of cardinality: how many
of each type did we see?

This change brings with it a much more interesting slice of type history: a
histogram. A Distribution holds a record of the top-N (where N is fixed at Ruby
compile-time) `(Class, ShapeId)` pairs and their counts. It also holds an
*other* count in case we see more than N pairs.

Using this distribution, we can make more informed decisions about when we
should use type information. We can determine if we are strictly monomorphic,
very nearly monomorphic, or something else. Maybe the call-site is polymorphic,
so we should have a polymorphic inline cache. Exciting stuff.

I also plumb this new distribution into the HIR part of the compilation
pipeline.

### DIFF
--- a/jit.c
+++ b/jit.c
@@ -454,15 +454,3 @@ rb_set_cfp_sp(struct rb_control_frame_struct *cfp, VALUE *sp)
 {
     cfp->sp = sp;
 }
-
-const uint32_t
-RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID;
-
-const uint32_t
-RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID;
-
-bool
-rb_zjit_singleton_class_p(VALUE klass)
-{
-    return RCLASS_SINGLETON_P(klass);
-}
--- a/jit.c
+++ b/jit.c
@@ -454,3 +454,15 @@ rb_set_cfp_sp(struct rb_control_frame_struct *cfp, VALUE *sp)
 {
     cfp->sp = sp;
 }
+
+const uint32_t
+RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID;
+
+const uint32_t
+RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID;
+
+bool
+rb_zjit_singleton_class_p(VALUE klass)
+{
+    return RCLASS_SINGLETON_P(klass);
+}
--- a/zjit.c
+++ b/zjit.c
@@ -351,11 +351,10 @@ VALUE rb_zjit_assert_compiles(rb_execution_context_t *ec, VALUE self);
 VALUE rb_zjit_stats(rb_execution_context_t *ec, VALUE self);
 VALUE rb_zjit_stats_enabled_p(rb_execution_context_t *ec, VALUE self);
 
-const uint32_t
-RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID;
-
-const uint32_t
-RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID;
+enum {
+    RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID,
+    RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID,
+};
 
 bool
 rb_zjit_singleton_class_p(VALUE klass)
--- a/zjit.c
+++ b/zjit.c
@@ -346,11 +346,6 @@ rb_zjit_shape_obj_too_complex_p(VALUE obj)
     return rb_shape_obj_too_complex_p(obj);
 }
 
-// Primitives used by zjit.rb. Don't put other functions below, which wouldn't use them.
-VALUE rb_zjit_assert_compiles(rb_execution_context_t *ec, VALUE self);
-VALUE rb_zjit_stats(rb_execution_context_t *ec, VALUE self);
-VALUE rb_zjit_stats_enabled_p(rb_execution_context_t *ec, VALUE self);
-
 enum {
     RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID,
     RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID,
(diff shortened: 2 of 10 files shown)


tekknolagi:
How can I fix the build so we let these special shape symbols leak? Or should I do something else?

k0kubun:
The failure is saying that you published symbols that do not follow our naming convention. It seems like you actually didn't use those symbols from other object files (even `cruby_bindings.inc.rs` doesn't use them), so you should be able to hide/remove those symbols.

If you want to just use the value of `SPECIAL_CONST_SHAPE_ID` and `INVALID_SHAPE_ID` from Rust, you should be able to define them in a way that it doesn't expose those symbols. One way we've done is to define an anonymous enum and refer to them in its variants.

jit.c
```diff
@@ -454,3 +454,15 @@ rb_set_cfp_sp(struct rb_control_frame_struct *cfp, VALUE *sp)
...
 }
+
+const uint32_t
+RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID;
+
+const uint32_t
+RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID;
+
+bool
+rb_zjit_singleton_class_p(VALUE klass)
```
k0kubun: Until we share them with YJIT, they should be defined in zjit.c. It's confusing to put non-shared things in the shared place, and ZJIT-only implementations should not bloat the YJIT-only binaries.

zjit/src/profile.rs
```diff
@@ -79,25 +80,100 @@ fn profile_insn(profiler: &mut Profiler, bare_opcode: ruby_vminsn_type) {
...
+/// * Integer + SPECIAL_CONST_SHAPE_ID == Fixnum
+/// * Float + SPECIAL_CONST_SHAPE_ID == Flonum
+/// * Symbol + SPECIAL_CONST_SHAPE_ID == StaticSymbol
+/// * NilClass == Nil
+/// * TrueClass == True
+/// * FalseClass == False
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ProfiledType {
+    class: VALUE,
+    shape: ShapeId,
```
k0kubun, line 118: For non-ivar instructions, is there any plan to use any information other than `SPECIAL_CONST_SHAPE_ID`?

I feel like `immediate: bool` (or a set of object flags including that) is what you actually wanted to use. Object shapes in Ruby are basically a cache key to access ivars today, so using all those 32 bits to store the single flag `SPECIAL_CONST_SHAPE_ID` seems space-inefficient.

For shapes, some instructions would need to query non-`class`/`immediate` information, so it might make sense to prepare a storage for instruction-specific profiling data, and store shapes only for ivar instructions.

tekknolagi: The tuple of (class, shape) is useful for ivar instructions and also sends --- we may see cases where a send `a.b` is actually a send to an `attr_accessor`, inline that, and then can use the shape information to optimize that into a hypothetical `LoadField`.

Re: other methods: we could absolutely do a discriminated union of shape|bitfield. If we come up with more bits, I would be delighted to do that!

k0kubun: Sure, attr accessors are a good point, but the point is that it's still applicable to only a limited number of instructions. Whether you come up with other bits or not, it'd be more space-efficient to make the storage size for profiling data instruction-specific, instead of using the largest number of bits for every instruction.

It's probably too early to optimize for space efficiency now, but from the maintenance point of view, it seems misleading to store shapes (instead of `immediate: bool`) for every instruction when (non-`SPECIAL_CONST_SHAPE_ID`) shapes are used only for 3 out of 110 instructions (getivar, setivar, opt_send_without_block).

tekknolagi: Yes, though if we do that we either need to keep them aligned (in which case they are all the same size anyway) or do variable-width profiles (which imo requires additional design).

tekknolagi: I think starting a doc of what profile information we might want per instruction is a great idea. So far I have:

* ivar stuff: (class, shape)
* sends: (class, shape)
* math ops: (class, immediate bit (which we could even fold into the class pointer?))
* opt_nil_p: (bool nilness)
* conditional branch instructions: (class)

tekknolagi:
Now I am getting this weird diff :(

```diff
-pub const RB_SPECIAL_CONST_SHAPE_ID: _bindgen_ty_12 = 33554432;
-pub const RB_INVALID_SHAPE_ID: _bindgen_ty_12 = 4294967295;
-pub type _bindgen_ty_12 = u32;
+pub const RB_SPECIAL_CONST_SHAPE_ID: _bindgen_ty_38 = 33554432;
+pub const RB_INVALID_SHAPE_ID: _bindgen_ty_38 = 4294967295;
+pub type _bindgen_ty_38 = u32;
```

Where did you recommend putting the enum? in shape.h?

k0kubun:
zjit.c. You don't need to share it with shape.h users. What's wrong with the diff? You should be able to apply the diff on the CI to your branch.

tekknolagi:
The diff is not appearing locally, only on CI. I wonder why they differ

zjit.c
```diff
@@ -351,5 +351,16 @@ VALUE rb_zjit_assert_compiles(rb_execution_context_t *ec, VALUE self);
...
+enum {
+    RB_SPECIAL_CONST_SHAPE_ID = SPECIAL_CONST_SHAPE_ID,
+    RB_INVALID_SHAPE_ID = INVALID_SHAPE_ID,
+};
+
+bool
+rb_zjit_singleton_class_p(VALUE klass)
+{
+    return RCLASS_SINGLETON_P(klass);
+}
```
k0kubun, line 358: In the comment on L349, I was proposing to not put things like this after L350-352 (similarly to YJIT) because leaving them last makes it clear that these functions declarations are used only by `zjit.rbinc` in the C code.

If you don't like the idea, it's ok, but please update L349 in that case.

k0kubun:
> I wonder why they differ

Probably using a different libclang version?

zjit.c
tekknolagi: Sorry, totally missed that! That makes a lot of sense

k0kubun (approved):

PR #13902, opened 2025-07-15
[Feature #21513] Raise on converting endless range to set

viralpraxis:
Before this patch, trying to convert endless range (e.g. `(1..)`) to set (using `to_set`) would hang


range.c
```diff
@@ -1017,6 +1017,30 @@ range_to_a(VALUE range)
...
+ *    ('a'..).to_set # => raises RangeError
+ *
+ */
+
+
+static VALUE
+range_to_set(VALUE range)
+{
+    if (NIL_P(RANGE_END(range))) {
+        rb_raise(rb_eRangeError, "cannot convert endless range to a set");
```
jeremyevans: I believe that endless range error should be the same error type as beginless range error.  Calling `to_set` on a beginless range currently raises `TypeError`.  This method should probably have beginless range raise `RangeError` instead, and include a test and an example for that.

range.c
```diff
@@ -1017,6 +1017,30 @@ range_to_a(VALUE range)
...
+ *  call-seq:
+ *    to_a -> set
+ *
+ *  Returns a set containing the elements in +self+, if a finite collection;
+ *  raises an exception otherwise.
+ *
+ *    (1..4).to_set     # => Set[1, 2, 3, 4]
+ *    (1...4).to_set    # => Set[1, 2, 3]
+ *    ('a'..'d').to_set # => Set["a", "b", "c", "d"]
+ *    ('a'..).to_set # => raises RangeError
```
jeremyevans:
```suggestion
 *    ('a'..).to_set    # RangeError
```

The use of `# =>` in example code implies a returned result. You also want `#` to line up.

jeremyevans (changes requested):
Thank you for working on this.

range.c
viralpraxis: Makes sense. Should we also tweak `Range#to_a` to raise `RangeError` in both cases?

```shell
./ruby -e '(..1).to_a'
-e:1:in 'Range#each': can't iterate from NilClass (TypeError)
```

jeremyevans: > Makes sense. Should we also tweak `Range#to_a` to raise `RangeError` in both cases?

@nobu what do you think?

viralpraxis:
There's a SEGV in the ZJIT test suite: https://github.com/ruby/ruby/actions/runs/16305765996/job/46051383888?pr=13902

Not sure if it's related to my changes

tekknolagi:
> There's a SEGV in the ZJIT test suite: https://github.com/ruby/ruby/actions/runs/16305765996/job/46051383888?pr=13902
> 
> Not sure if it's related to my changes

ZJIT error seems entirely unrelated to your change from a quick glance

tekknolagi:
Looks like a weird GC issue related to marking though so paging in @k0kubun for good measure

range.c
```diff
@@ -1017,6 +1017,31 @@ range_to_a(VALUE range)
...
+ *    (1...4).to_set    # => Set[1, 2, 3]
+ *    ('a'..'d').to_set # => Set["a", "b", "c", "d"]
+ *    (..'d').to_set    # RangeError
+ *    ('a'..).to_set    # RangeError
+ *
+ */
+
+
+static VALUE
+range_to_set(VALUE range)
```
nobu: Although passing arguments to `Enumerable#to_set` is deprecated, it is allowed yet.
Shouldn't this consistent with it too?

range.c
nobu: It feels clearer for me.
@knu Any thoughts?

If we will change it too, maybe ruby/spec also needs to be updated?

range.c
viralpraxis: Thanks, I missed that. It should accept arguments now.

k0kubun:
I agree that it's probably not related to the changes. We're looking into GC issues with ZJIT.

range.c
knu: We should perhaps check the size of a given Enumerable in Set.new and raise an error when it is infinity.

viralpraxis: > We should perhaps check the size of a given Enumerable in Set.new and raise an error when it is infinity.

Seems good to me. I'll try to follow this path

viralpraxis: I tried out this idea and it seems to work fine (we now raise `ArgumentError` error for any endless enumerable, e.g. `1.upto(Float::INFINITY)`.

```shell
$ ./ruby -e 'Set.new((1..nil))'
-e:1:in 'Set#initialize': cannot initialize Set from a enumerable with infinite size (ArgumentError)
	from -e:1:in '<main>'
$ ./ruby -e '(1..nil).to_set'  
-e:1:in 'Enumerable#to_set': cannot initialize Set from a enumerable with infinite size (ArgumentError)
	from -e:1:in '<main>'

```

viralpraxis:
Could someone please take a look?

set.c
```diff
@@ -494,6 +494,14 @@ set_i_initialize(int argc, VALUE *argv, VALUE set)
     rb_check_arity(argc, 0, 1);
 
     if (argc > 0 && (other = argv[0]) != Qnil) {
+        ID id_size = rb_intern("size");
```
jeremyevans: Should move this to the following `else` block, so it won't be called for arrays, as arrays never have infinite size.

set.c
```diff
@@ -494,6 +494,14 @@ set_i_initialize(int argc, VALUE *argv, VALUE set)
...
+        if (rb_respond_to(other, id_size)) {
+            VALUE size = rb_funcall(other, id_size, 0);
+            if (RB_TYPE_P(size, T_FLOAT) && RFLOAT_VALUE(size) == INFINITY) {
+                rb_raise(rb_eArgError, "cannot initialize Set from a enumerable with infinite size");
```
jeremyevans: You aren't checking that the argument is enumerable, only that the size is infinite.  It would be more correct to say "an object" instead of "a enumerable".  Even if you keep the current wording, it should be "an enumerable" instead of "a enumerable".

jeremyevans (changes requested):

viralpraxis:
@jeremyevans thank you very much for helping me with that! ❤️

viralpraxis:
CI failures are unrelated:

```bash
   TestGemRequest#test_verify_certificate_extra_message:
  Test::Unit::ProxyError: X509_dup
      /home/runner/work/ruby/ruby/src/lib/rubygems/request.rb:92:in 'OpenSSL::X509::StoreContext#current_cert'
      /home/runner/work/ruby/ruby/src/lib/rubygems/request.rb:92:in 'Gem::Request.verify_certificate'
      /home/runner/work/ruby/ruby/src/test/rubygems/test_gem_request.rb:373:in 'block in TestGemRequest#test_verify_certificate_extra_message'
      /home/runner/work/ruby/ruby/src/lib/rubygems/user_interaction.rb:46:in 'Gem::DefaultUserInteraction.use_ui'
      /home/runner/work/ruby/ruby/src/lib/rubygems/user_interaction.rb:69:in 'Gem::DefaultUserInteraction#use_ui'
      /home/runner/work/ruby/ruby/src/test/rubygems/test_gem_request.rb:372:in 'TestGemRequest#test_verify_certificate_extra_message'
```

knu (approved):

PR #13903, opened 2025-07-15
ZJIT: Profile each instruction at most num_profiles times

k0kubun:
This PR changes ZJIT instructions to only profile at most `--zjit-num-profiles` times.

ZJIT currently profiles all instructions for `num_profiles` calls, but when the ISEQ contains a while loop, each instruction can be profiled more than `num_profiles` times, which I didn't expect. So this PR optimizes such cases. We don't really use while loops in reality, but loop methods themselves use a while loop, so this will be helpful for real-world code too.

There's a trade-off that we need to consume more memory for the per-insn counters, but the impact seems significant enough to justify it. In addition, we could make it more compact and/or restrict some instructions to profile only once to save memory, but it's out of scope in this PR.

```
$ ./run_benchmarks.rb getivar --chruby 'before --zjit;after --zjit' --once
Running benchmark "getivar" (1/1)
+ setarch x86_64 -R taskset -c 10 /opt/rubies/before/bin/ruby --zjit -I harness /home/k0kubun/src/github.com/Shopify/yjit-bench/benchmarks/getivar.rb
ruby 3.5.0dev (2025-07-15T20:31:58Z master 3cf32e9364) +ZJIT +PRISM [x86_64-linux]
itr:   time
 #1:  684ms
RSS: 14.9MiB
MAXRSS: 16.7MiB
Running benchmark "getivar" (1/1)
+ setarch x86_64 -R taskset -c 10 /opt/rubies/after/bin/ruby --zjit -I harness /home/k0kubun/src/github.com/Shopify/yjit-bench/benchmarks/getivar.rb
ruby 3.5.0dev (2025-07-15T22:07:56Z zjit-profile-count b0e903fad8) +ZJIT +PRISM [x86_64-linux]
itr:   time
 #1:  128ms
RSS: 15.0MiB
MAXRSS: 16.7MiB
Total time spent benchmarking: 1s

before: ruby 3.5.0dev (2025-07-15T20:31:58Z master 3cf32e9364) +ZJIT +PRISM [x86_64-linux]
after: ruby 3.5.0dev (2025-07-15T22:07:56Z zjit-profile-count b0e903fad8) +ZJIT +PRISM [x86_64-linux]

getivar  684.6        0.0         128.6       0.0         5.324          5.324
Legend:

```

### DIFF
--- a/zjit/src/profile.rs
+++ b/zjit/src/profile.rs
@@ -73,7 +73,7 @@ fn profile_insn(profiler: &mut Profiler, bare_opcode: ruby_vminsn_type) {
     }
 
     // Once we profile the instruction num_profiles times, we stop profiling it.
-    profile.num_profiles[profiler.insn_idx] += 1;
+    profile.num_profiles[profiler.insn_idx] = profile.num_profiles[profiler.insn_idx].saturating_add(1);
     if profile.num_profiles[profiler.insn_idx] == get_option!(num_profiles) {
         unsafe { rb_zjit_iseq_insn_set(profiler.iseq, profiler.insn_idx as u32, bare_opcode); }
     }


tekknolagi:
What do you think about having num profiles correspond to the sum of:

* function call
* backward jump

This way we can still keep one count per iseq but we don't have a profile-time explosion with loops.

zjit/src/profile.rs
```diff
@@ -38,43 +38,49 @@ impl Profiler {
...
             let argc = unsafe { vm_ci_argc((*cd).ci) };
             // Profile all the arguments and self (+1).
-            profile_operands(profiler, (argc + 1) as usize);
+            profile_operands(profiler, profile, (argc + 1) as usize);
         }
         _ => {}
     }
+
+    // Once we profile the instruction num_profiles times, we stop profiling it.
+    profile.num_profiles[profiler.insn_idx] += 1;
```
tekknolagi: saturating add?

tekknolagi (approved):
seems a reasonable solution for this approach but curious what you think of an alternative

k0kubun:
> What do you think about having num profiles correspond to the sum of:
> function call
> backward jump
> This way we can still keep one count per iseq but we don't have a profile-time explosion with loops.

With that implementation, when there's a loop, non-looped instructions or instructions in a different loop will have a significantly smaller number of profiles. While it satisfies "we don't have a profile-time explosion with loops", I don't think it's the intended behavior.

zjit/src/profile.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/13903/commits/094692d09adec24cf3e2384bee1a10149d89ee87

tekknolagi:
> With that implementation, when there's a loop, non-looped instructions or instructions in a different loop will have a significantly smaller number of profiles.

I think that's fine, actually, and mirrors what we have today as well. It's kind of sampling-profiler-esque.

k0kubun:
I think many instructions will need only one profile (as suggested in the PR description), so I'm fine with that for such instructions. I feel uncomfortable having only one profile for other instructions.

However, if the intention is that we assume it never happens in reality because people don't use while loops and it doesn't matter for loop methods themselves, I'm open to merging such a change.

PR #13904, opened 2025-07-16
ZJIT: A64 backend miscomp fixes (and a CI fix)

XrXr:
I noticed that SP movement still weren't single instruction after my other
PR and that led me to find a few bugs.

Please check commit message for each.

- **ZJIT: Have `make zjit-test` use the same Cargo features as miniruby**
    We should now be running the disassembly tests in the backend on CI.
- **ZJIT: A64: Fix `asm.add_into(NATIVE_STACK_POINTER, ...)`**
    Probably the most important one because we use this for side exits.
- **ZJIT: A64: Fix the optimization merging `asm.add(reg, imm)` with Mov**
- **ZJIT: A64: Fix bad operand swapping in `asm.sub(imm, reg)`**


zjit/Cargo.toml
```diff
@@ -18,3 +18,4 @@ expect-test = "1.5.1"
 [features]
 # Support --yjit-dump-disasm and RubyVM::YJIT.disasm using libcapstone.
 disasm = ["capstone"]
+runtime_checks = []
```
tekknolagi, line 21: Please add a TODO(#issuenumber) to turn this back on to track this

tekknolagi (commented):
I don't know much about Arm but looks reasonable at first glance. Delegating to Kokubun

Cargo.toml
```diff
@@ -22,7 +22,8 @@ path = "jit.rs"
 
 [features]
 disasm = ["yjit?/disasm", "zjit?/disasm"]
-runtime_checks = []
+# TODO(GH-642) Turning this on trips a btest failure.
```
k0kubun, line 25: It's hard for other contributors to figure out it refers to Shopify/ruby when you're looking at ruby/ruby, so it might be nice to mention the repository name. It confused me at first too.

PR #13910, opened 2025-07-16
[DOC] Bump autoconf requirement

st0012:
Autoconf 2.70 or later is required to properly handle multi-line AC_CHECK_FUNCS lists. The recent changes in #13907 that split function lists across multiple lines require autoconf 2.70's improved whitespace handling.

Without autoconf 2.70+, the configure script generation fails with syntax errors like:
../configure: line 29306: syntax error near unexpected token `pthread_attr_get_np'

This is because autoconf 2.69 and earlier don't properly handle multi-line function lists in AC_CHECK_FUNCS without backslash continuations, while 2.70+ requires square brackets and handles newlines correctly.

cc @nobu

### DIFF
--- a/doc/contributing/building_ruby.md
+++ b/doc/contributing/building_ruby.md
@@ -14,7 +14,7 @@
 
     If you want to build from the git repository, you will also need:
 
-    * [autoconf] - 2.67 or later
+    * [autoconf] - 2.70 or later
     * [gperf] - 3.1 or later
         * Usually unneeded; only if you edit some source files using gperf
     * ruby - 3.0 or later


tekknolagi:
Is it possible to make the build fail with a better message in addition to bumping the documentation? Like detect an old autoconf

nobu:
Thank you.
Should we revert 32bfb61d34?

nobu:
@tekknolagi It would be possible by bumping up `AC_PREREQ`.

st0012:
I'm not familiar with the criteria required to bump Ruby's build dependencies. But given that autoconf `2.67` is released in 2010 and `2.70` is 2020, this is likely a big jump to some devs/build systems.
So if we don't get much out of https://github.com/ruby/ruby/commit/32bfb61d349b49ddedb7d34d9e434063324aafcc other than fixing warnings, I think it's better to revert it.

Earlopain:
For reference, debian bullseye only contains autoconf 2.69: https://packages.debian.org/bullseye/autoconf

PR #13912, opened 2025-07-16
ZJIT: run `zjit-test-all` on CI with seed

st0012:
ZJIT is still unstable and can crash intermittently running with Ruby tests. Adding a seed to avoid flakiness while we work on stability.


k0kubun:
> ZJIT: Request team review for workflow changes too

For the record, I deliberately skipped these files because the team don't necessarily need to review dependabot updates for GitHub Actions. It's noisy in notifications, and they're auto-merged anyway.

st0012:
Oh I see. Ok I can drop it.

k0kubun (approved):
Thanks!

PR #13918, opened 2025-07-16
ZJIT: Remove unneeded test exclusions

st0012:
After several ZJIT fix PRs, the number of failing/crashing tests for ZJIT has dropped significantly. This removes the unneeded test exclusions.

### DIFF
--- a/.github/auto_request_review.yml
+++ b/.github/auto_request_review.yml
@@ -10,6 +10,7 @@ files:
   'zjit/src/cruby_bindings.inc.rs': []
   'doc/zjit*': [team:jit]
   'test/ruby/test_zjit*': [team:jit]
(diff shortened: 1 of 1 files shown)


.github/workflows/zjit-ubuntu.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
           - test_task: 'zjit-test-all'
             configure: '--enable-zjit=dev'
-            testopts: '--seed=11831'
+            testopts: '--seed=18140'
```
st0012, line 45: With different tests running, ubuntu now needs a different seed to pass.
The old seed was causing [this crash](https://github.com/ruby/ruby/actions/runs/16329885791/job/46129574671).

k0kubun (approved):
:+1:

PR #13919, opened 2025-07-16
Detect rehash in array-backed hashes

sferik:
Added safeguards for array-backed hashes by storing the table pointer and bound in `hash_foreach_arg` and checking them during iteration to raise an exception if a rehash occurs.

Fixed [TODO added in 2018](https://github.com/ruby/ruby/commit/8f675cdd00e#diff-2894a90fd17590e4f7afd08dfd8505526d5283e09fea645cc7a7d4873936bdffR1113) by @ko1.


hash.c
```diff
@@ -1300,8 +1300,11 @@ hash_ar_foreach_iter(st_data_t key, st_data_t value, st_data_t argp, int error)
 
     if (error) return ST_STOP;
 
+    ar_table *tbl = RHASH_AR_TABLE(arg->hash);
     int status = (*arg->func)((VALUE)key, (VALUE)value, arg->arg);
-    /* TODO: rehash check? rb_raise(rb_eRuntimeError, "rehash occurred during iteration"); */
+    if (!RHASH_AR_TABLE_P(arg->hash) || RHASH_AR_TABLE(arg->hash) != tbl) {
```
ko1: I forget what happens on rehash, but `RHASH_AR_TABLE(arg->hash) != tbl` is correct checking?

XrXr: I know this is copying from similar code using `RHASH_ST_TABLE`, but that check might actually never trigger. RHASH_ST_TABLE() just adds an offset to the argument, so for the same hash, it's always going to be the same. `arg->hash` is never going to be changed by the callback, so that check will always be false.

sferik: I changed this condition to check if the bound increases. That should work, right?

XrXr:
@ko1 is this check and the on for ST table meant to be a debug assert? Seems like it's not possible to accurately check for rehash here efficiently. Maybe we should just remove the code and TODO, or turn them into `RUBY_ASSERT`. 

```c
st_table *tbl = RHASH_ST_TABLE(arg->hash);
int status = (*arg->func)((VALUE)key, (VALUE)value, arg->arg);

if (RHASH_ST_TABLE(arg->hash) != tbl) {
    rb_raise(rb_eRuntimeError, "rehash occurred during iteration");
}
```
With recent changes it never raises:
```c
static inline st_table *
RHASH_ST_TABLE(VALUE h)
{
    return (st_table *)((uintptr_t)h + sizeof(struct RHash));
}
```

Also, Hash#rehash disallows running while iterating, so it's unclear how these checks could fail in practice:
```c
if (hash_iterating_p(hash)) {
    rb_raise(rb_eRuntimeError, "rehash during iteration");
}
```

ko1:
> @ko1 is this check and the on for ST table meant to be a debug assert? Seems like it's not possible to accurately check for rehash here efficiently. Maybe we should just remove the code and TODO, or turn them into `RUBY_ASSERT`.

+1 for removing the code and TODO.

sferik:
Okay, I've removed the code I added and the TODO. Would you like me to squash this into a single commit or will you squash when you merge?

PR #13920, opened 2025-07-16
FileUtils.remove_dir checks directory

sferik:
Ensure `FileUtils.remove_dir` is passed a directory. Add a unit test verifying that `Errno::ENOTDIR` is raised when calling `remove_dir` on a non-directory/file.

This may cause some existing "working" programs to error, but in a way that prevents users from accidentally deleting files, which seems like a safer approach than allowing the program to delete the file and continue running.

Fixes [FIXME added in 2005](https://github.com/ruby/ruby/commit/55ec2ad1637d1aa4bbc7c433da8d4d4b7b1f3ecc#diff-f76dd327efaf28e3cf96422b26cd2ce702f7996c28963bc26d3f46cb29fb69ceR37-R719) by @aamine.

### DIFF
--- a/lib/fileutils.rb
+++ b/lib/fileutils.rb
@@ -1491,7 +1491,8 @@ module FileUtils
   # Related: {methods for deleting}[rdoc-ref:FileUtils@Deleting].
   #
   def remove_dir(path, force = false)
-    remove_entry path, force   # FIXME?? check if it is a directory
+    raise Errno::ENOTDIR, path unless File.directory?(path)
+    remove_entry path, force
   end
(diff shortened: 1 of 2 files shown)


sferik:
> ❌ [Tests Failed](https://app.launchableinc.com/organizations/ruby/workspaces/ruby/data/builds/4240112)
> ✖️no tests failed ✔️62095 tests passed(2 flakes)

I'm not sure what this means or how to fix this.

hsbt:
Could you submit this to https://github.com/ruby/fileutils instead of this?

PR #13921, opened 2025-07-16
YJIT: Side-exit on String#dup when it's not leaf

k0kubun:
This fixes https://github.com/ruby/ruby/commit/ccbbe06a0277db74cfbf7dc1f26722cbf4904b72.

`rb_obj_dup` is not leaf, so it gives a wrong backtrace if you skip pushing a frame when the string has ivars and falls back to `rb_obj_dup`. This PR changes it to side-exit, reversing the simplification we made in https://github.com/ruby/ruby/pull/13606.

### DIFF
--- a/shape.h
+++ b/shape.h
@@ -48,7 +48,8 @@ enum shape_id_fl_type {
 
 // This masks allows to check if a shape_id contains any ivar.
 // It rely on ROOT_SHAPE_WITH_OBJ_ID==1.
-#define SHAPE_ID_HAS_IVAR_MASK (SHAPE_ID_FL_TOO_COMPLEX | (SHAPE_ID_OFFSET_MASK - 1))
+#define SHAPE_ID_HAS_IVAR_MASK 0x807fffe
+STATIC_ASSERT(shape_id_has_ivar_mask, SHAPE_ID_HAS_IVAR_MASK == (SHAPE_ID_FL_TOO_COMPLEX | (SHAPE_ID_OFFSET_MASK - 1)));
 
 // The interpreter doesn't care about frozen status or slot size when reading ivars.
 // So we normalize shape_id by clearing these bits to improve cache hits.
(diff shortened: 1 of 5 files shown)


shape.h
```diff
@@ -48,7 +48,8 @@ enum shape_id_fl_type {
 
 // This masks allows to check if a shape_id contains any ivar.
 // It rely on ROOT_SHAPE_WITH_OBJ_ID==1.
-#define SHAPE_ID_HAS_IVAR_MASK (SHAPE_ID_FL_TOO_COMPLEX | (SHAPE_ID_OFFSET_MASK - 1))
+#define SHAPE_ID_HAS_IVAR_MASK 0x807fffe
```
tekknolagi: Why do you need this constant inline?

tekknolagi (approved):
lgtm with one q

shape.h
k0kubun: For bindgen.

XrXr: This is a bindgen limitation where it can't tell whether a macro is a constant. An alternative is to put the expression in an anonymous enum, which bindgen can figure out since it's a language level thing to force a value to be compile time known and not allocate storage for it.
Could you try that? If it doesn't work for some C code or whatever, as is isn't too horrible.

k0kubun: sure thing

tekknolagi: A secret third thing is to make a `uint32_t abc = SHAPE_ID_HAS_IVAR_MASK` in `jit.c` and then bindgen that

tekknolagi: (If we care about signedness, which it looks like we do, we can't really rely on enum signedness. and I think enum width, too)

k0kubun: like this? https://github.com/ruby/ruby/pull/13921/commits/fae20391650d03ac075141702759b02caae278cc

XrXr: No need to worry about signedness because the operands in the expression come from enum themselves 😁 Go integer promotion.

byroot (approved):

PR #13926, opened 2025-07-17
Ractor shareable proc

ko1:
Ractor.shareable_proc

call-seq:
  Ractor.sharable_proc(self: nil){} -> sharable proc

It returns shareable Proc object. The Proc object is
shareable and the self in a block will be replaced with
the value passed via `self:` keyword.

In a shareable Proc, the outer variables are read-only
and the outer variables should point only shareable objects.

```
    a = 42
    b = []
    Ractor.shareable_proc{ p a }.call # ok
    Ractor.shareable_proc{ p b }
    #=> can not make shareable Proc because it can refer unshareable object [] from variable 'b' (Ractor::IsolationError)
```

Ractor.sharaeble_lambda is also introduced.

Ractor.make_sharable(a_proc) is no longer supported.

[Feature #21039]

### DIFF
--- a/gems/bundled_gems
+++ b/gems/bundled_gems
@@ -13,7 +13,7 @@ test-unit           3.7.0   https://github.com/test-unit/test-unit
 rexml               3.4.2   https://github.com/ruby/rexml
 rss                 0.3.1   https://github.com/ruby/rss
 net-ftp             0.3.8   https://github.com/ruby/net-ftp
-net-imap            0.5.10  https://github.com/ruby/net-imap
+net-imap            0.5.10  https://github.com/ruby/net-imap 71c0288b9a8f78a7125a4ce2980ab421acdf5836
 net-pop             0.1.2   https://github.com/ruby/net-pop
 net-smtp            0.5.1   https://github.com/ruby/net-smtp
 matrix              0.4.3   https://github.com/ruby/matrix


bootstraptest/test_ractor.rb
```diff
@@ -1137,41 +1156,17 @@ def /(other)
   [a.frozen?, a[0].frozen?] == [true, false]
 }
 
-# Ractor.make_shareable(a_proc) makes a proc shareable.
+# Ractor.make_shareable(a_proc) is not supported now.
```
eregon, line 1159: Actually this only error because self is not shareable:
```
$ ruby -ve 'pr = Proc.new{}; Ractor.make_shareable(pr)'
ruby 3.5.0dev (2025-09-23T19:05:52Z master 0fe1099d9a) +PRISM [x86_64-linux]
-e:1:in 'Ractor.make_shareable': Proc's self is not shareable: #<Proc:0x00007f9341328148 -e:1> (Ractor::IsolationError)
	from -e:1:in '<main>'
```

It seems to work the same as `shareable_proc` if `self` is shareable (good for compatibility):
```
$ ruby -e 'nil.instance_exec { pr = Proc.new{}; p Ractor.make_shareable(pr) }'     
#<Proc:0x00007fd5eb078050 -e:1>
```

(so this comment should probably be updated)

ko1: Ah, my comment was wrong. `make_shareable(proc)` is not prohibited now.

eregon: https://github.com/ruby/ruby/pull/15026

PR #13928, opened 2025-07-17
Flag rsock_raise_user_specified_timeout() as NORETURN

osyoyu:
This suppresses this warning:

```
../../../ext/socket/ipsocket.c: In function ‘rsock_raise_user_specified_timeout’: ../../../ext/socket/ipsocket.c:30:1: warning: function might be candidate for attribute ‘noreturn’ [-Wsuggest-attribute=noreturn]
   30 | rsock_raise_user_specified_timeout()
      | ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

@shioimm Can you check this, as this code was introduced in https://github.com/ruby/ruby/pull/13909?


ext/socket/ipsocket.c
```diff
@@ -27,7 +27,7 @@ struct inetsock_arg
 };
 
 void
-rsock_raise_user_specified_timeout()
+rsock_raise_user_specified_timeout(void)
```
nobu, line 30: I missed this.

PR #13932, opened 2025-07-17
Test IBM platform

junaruga:
Test IBM ppc64le/s390x platform on GitHub Actions because the Ruby project was approved at https://github.com/IBM/actionspz/issues/4#issuecomment-3075984322.


junaruga:
OK. I confirmed that the following images worked.

* ubuntu-24.04-ppc64le
* ubuntu-22.04-ppc64le
* ubuntu-24.04-s390x
* ubuntu-22.04-s390x

PR #13935, opened 2025-07-17
ZJIT: Precise GC writebarriers

jhawthorn:
This issues writebarriers for objects added via gc_offsets or by profiling. This may be slower than writebarrier_remember, but we would like it to be more debuggable.

We found the lack of gc_offset writebarriers via [wbcheck](https://github.com/ruby/ruby/pull/13557)

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -4,7 +4,7 @@ use std::rc::Rc;
 use crate::asm::Label;
 use crate::backend::current::{Reg, ALLOC_REGS};
 use crate::invariants::track_bop_assumption;
-use crate::gc::get_or_create_iseq_payload;
+use crate::gc::{get_or_create_iseq_payload, append_gc_offsets};
 use crate::state::ZJITState;
 use crate::{asm::CodeBlock, cruby::*, options::debug, virtualmem::CodePtr};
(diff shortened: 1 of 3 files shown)


XrXr (approved):
Sounds good to me.

Side note: we should really have a proper object pool instead of baking VALUEs directly into code. It's one of those things that we never got around to doing for YJIT. Should be faster to mark an array of VALUEs than to scavenge VALUEs everywhere in generated code.

k0kubun:
looks good to me too. 

> This issues writebarriers for objects added via gc_offsets or by profiling. This may be slower than writebarrier_remember, but we would like it to be more debuggable.

seems fair.

---

> Side note: we should really have a proper object pool instead of baking VALUEs directly into code.

About the object pool idea, I'm still worried about a potential overhead added by introducing one indirection that doesn't exist today. If production gets better CoW efficiency, it'd be a fair trade-off in terms of overall performance, but I'd like us to carefully evaluate the speed when we get to work on it.

k0kubun:
> [tekknolagi](https://github.com/tekknolagi) disabled auto-merge

Sorry, the latter half of my comment might have confused you, but it has nothing to do with this PR. You don't need to disable auto-merge 😅 I'll issue one again.

tekknolagi:
Haha, no, unrelated to your comment -- I got excited and forgot about how we have been normally letting committers merge on their own post-review and non-committers expressly request a merge

k0kubun:
ok, I'll disable mine too then

st0012 (approved):

PR #13937, opened 2025-07-17
ZJIT: Mark the code region executable on partial failures

k0kubun:
This PR fixes a random bus error on yjit-bench for arm64.

ZJIT compilation currently has three phases that run `asm.compile()`:

1. `gen_function()`: Compiles the first ISEQ body
2. `gen_entry()`: Compiles the trampoline to enter the ISEQ's JIT code 
3. `gen_iseq()`: Compiles callee ISEQs for JIT-to-JIT calls

Once it reaches the `emit` pass of (1), it modifies the code region state. So, even if (2) or (3) fails, you need to run `cb.mark_all_executable()` after (1). Currently, it does not call `cb.mark_all_executable()` when (3) fails, so that's the problem this PR fixes. To reduce the possibility of making this mistake again, I made a wrapper function that does it.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -96,8 +96,22 @@ pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, _ec: EcPtr) -> *co
     code_ptr
 }
 
-/// Compile an entry point for a given ISEQ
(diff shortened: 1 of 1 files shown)


tekknolagi (approved):
Awesome!

XrXr (approved):

PR #13941, opened 2025-07-17
ZJIT: Support jitdump

tekknolagi:
This lets us ZJIT compiled functions show up in the profiles of, say,
perf, or samply.

Fix https://github.com/Shopify/ruby/issues/634

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -154,6 +154,20 @@ fn gen_iseq_entry_point_body(cb: &mut CodeBlock, iseq: IseqPtr) -> *const u8 {
     start_ptr.map(|start_ptr| start_ptr.raw_ptr(cb)).unwrap_or(std::ptr::null())
 }
 
+fn register_with_perf(iseq_name: String, start_ptr: usize, code_size: usize) {
+    use std::io::Write;
+    let perf_map = format!("/tmp/perf-{}.map", std::process::id());
+    debug!("Registering JIT code in perf map: {perf_map}");
+    let Ok(mut file) = std::fs::OpenOptions::new().create(true).append(true).open(&perf_map) else {
+        debug!("Failed to open perf map file: {perf_map}");
+        return;
+    };
+    let Ok(_) = write!(file, "{:x} {:x} zjit::{}\n", start_ptr, code_size, iseq_name) else {
+        debug!("Failed to write {iseq_name} to perf map file: {perf_map}");
+        return;
+    };
+}
+
 /// Compile a JIT entry
 fn gen_entry(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function, function_ptr: CodePtr, c_stack_bytes: usize) -> Option<CodePtr> {
     // Set up registers for CFP, EC, SP, and basic block arguments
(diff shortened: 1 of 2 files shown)


tekknolagi:
Before:
Screenshot 2025-07-17 at 4 15 25 PM

After:
Screenshot 2025-07-17 at 4 15 33 PM

tekknolagi:
There are still some unnamed pointers in there. Perhaps I am not getting the full code range?

k0kubun:
> There are still some unnamed pointers in there. Perhaps I am not getting the full code range?

You recorded only the JIT entry trampoline. The frame for the main function would be unnamed, and the name of callees in JIT-to-JIT calls would never appear. I think both of them should have symbols.

tekknolagi:
Oh, silly

zjit/src/codegen.rs
```diff
@@ -154,6 +154,19 @@ fn gen_iseq_entry_point_body(cb: &mut CodeBlock, iseq: IseqPtr) -> *const u8 {
     start_ptr.map(|start_ptr| start_ptr.raw_ptr(cb)).unwrap_or(std::ptr::null())
 }
 
+fn register_with_perf(iseq_name: String, start_ptr: usize, code_size: usize) {
+    use std::io::Write;
+    let perf_map = format!("/tmp/perf-{}.map", std::process::id());
+    let Ok(mut file) = std::fs::OpenOptions::new().create(true).append(true).open(&perf_map) else {
```
k0kubun, line 161: I think we had found doing something like this was slow for YJIT. Just like YJIT, could you create the file when the option is parsed, keep the fd, and just write to the fd here?

tekknolagi:
@k0kubun where should I be adding this? I think I am not finding the right place to get the (start, end) pair

k0kubun:
`gen_function()` maybe? That seems to cover both cases.

zjit/src/codegen.rs
tekknolagi: YJIT looks like it writes addresses at the end of a function, so it'll do so roughly as much as this PR does now

tekknolagi: (and does not look like it keeps a fd)

tekknolagi:
That looks better but still has some magic numbers
Screenshot 2025-07-17 at 5 21 51 PM

zjit/src/codegen.rs
k0kubun: Oh sorry, my bad. I mixed up disasm and perf. Never mind!

k0kubun:
You seem to have removed the symbol from the JIT entry trampoline. By both cases I meant the main function and callee ISEQs, not including the JIT entry. So you need to record symbols in two places now.

tekknolagi:
Awesome, thank you. Very cool
Screenshot 2025-07-17 at 5 30 17 PM

zjit/src/codegen.rs
```diff
@@ -172,7 +186,16 @@ fn gen_entry(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function, function_pt
...
     asm.cret(C_RET_OPND);
 
-    asm.compile(cb).map(|(start_ptr, _)| start_ptr)
+    let result = asm.compile(cb).map(|(start_ptr, _)| start_ptr);
+    if let Some(start_addr) = result {
+        if get_option!(perf) {
+            let start_ptr = start_addr.raw_ptr(cb) as usize;
+            let end_ptr = cb.get_write_ptr().raw_ptr(cb) as usize;
+            let code_size = end_ptr - start_ptr;
+            register_with_perf("entry".into(), start_ptr, code_size);
```
k0kubun: I think it's helpful to include the ISEQ name in the symbol for the entry, also keeping "entry" as part of the symbol. We'd be interested in which ISEQs are called from C, which you could see by looking at the callee frame but it's easier to understand if it's annotated in the frame itself. Using the same "entry" symbol for all entries could be problematic depending on how you aggregate perf profiles too.

tekknolagi: Will do before landing

st0012 (approved):
Should we add profiling related entries to `doc/zjit.md` too?

PR #13948, opened 2025-07-18
Use concurrent set for global symbol table

peterzhu2118:
This PR switches the global symbol table and allows for symbol lookup and creation of dynamic symbols without needing the global VM lock. Creating static symbols still require the global VM lock but I will be investigating how to make that lock-free in the future.

Here's a microbenchmark demonstrating the speedup:

```ruby
ARGV[0].to_i.times.map do
  Ractor.new do
    1_000_000.times do |i|
      "hello#{i}".to_sym
    end
  end
end.map(&:value)
```

Results:

chart

Raw results:

| Ractor count | Branch (s) | Master (s) |
|--------------|------------|------------|
| 1            | 0.364      | 0.401      |
| 2            | 0.555      | 1.149      |
| 3            | 0.583      | 3.890      |
| 4            | 0.680      | 3.288      |
| 5            | 0.789      | 5.107      |

### DIFF
--- a/concurrent_set.c
+++ b/concurrent_set.c
@@ -119,7 +119,7 @@ concurrent_set_try_resize_without_locking(VALUE old_set_obj, VALUE *set_obj_ptr)
         RUBY_ASSERT(key != CONCURRENT_SET_MOVED);
 
         if (key < CONCURRENT_SET_SPECIAL_VALUE_COUNT) continue;
-        if (rb_objspace_garbage_object_p(key)) continue;
+        if (!RB_SPECIAL_CONST_P(key) && rb_objspace_garbage_object_p(key)) continue;
 
         VALUE hash = RUBY_ATOMIC_VALUE_LOAD(entry->hash);
         if (hash == 0) {
@@ -231,7 +231,7 @@ rb_concurrent_set_find_or_insert(VALUE *set_obj_ptr, VALUE key, void *data)
             VALUE curr_hash = RUBY_ATOMIC_VALUE_LOAD(entry->hash);
             if ((curr_hash == hash || curr_hash == 0) && set->funcs->cmp(key, curr_key)) {
                 // We've found a match.
-                if (UNLIKELY(rb_objspace_garbage_object_p(curr_key))) {
+                if (UNLIKELY(!RB_SPECIAL_CONST_P(curr_key) && rb_objspace_garbage_object_p(curr_key))) {
                     // This is a weakref set, so after marking but before sweeping is complete we may find a matching garbage object.
                     // Skip it and mark it as deleted.
                     RUBY_ATOMIC_VALUE_CAS(entry->key, curr_key, CONCURRENT_SET_DELETED);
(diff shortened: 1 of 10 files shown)


concurrent_set.c
```diff
@@ -167,6 +175,61 @@ concurrent_set_try_resize(VALUE old_set_obj, VALUE *set_obj_ptr)
...
+
+    VALUE set_obj;
+
+  retry:
+    set_obj = RUBY_ATOMIC_VALUE_LOAD(*set_obj_ptr);
+    RUBY_ASSERT(set_obj);
+    struct concurrent_set *set = RTYPEDDATA_GET_DATA(set_obj);
+
+    struct concurrent_set_probe probe;
+    VALUE hash = set->funcs->hash(key);
```
jhawthorn: (non-blocking) I doubt it makes a significant difference, but I think `hash` won't change and could go above `retry:`

peterzhu2118: Good catch 👍 I copied the code from `rb_concurrent_set_find_or_insert` which also has this issue so I fixed it in both places.

jhawthorn: Ah, I see. That didn't work because we only load `set` later (even though we know a new set will still have the same funcs).

I don't think this change is worth making 😅

concurrent_set.c
```diff
@@ -61,6 +61,14 @@ rb_concurrent_set_new(const struct rb_concurrent_set_funcs *funcs, int capacity)
     return obj;
 }
 
+rb_atomic_t
+rb_concurrent_size(VALUE set_obj)
```
casperisfine:
```suggestion
rb_concurrent_set_size(VALUE set_obj)
```

We might have other `rb_concurrent_<data-structure>` in the future.

gc.c
```diff
@@ -3166,7 +3168,6 @@ rb_gc_mark_children(void *objspace, VALUE obj)
     switch (BUILTIN_TYPE(obj)) {
       case T_FLOAT:
       case T_BIGNUM:
-      case T_SYMBOL:
```
casperisfine, line 3169:
```
* RSYMBOL(obj)->fstr intentionally not marked. See log for 96815f1e
         * ("symbol.c: remove rb_gc_mark_symbols()") */
```

That comment is outdated now, right?

symbol.c
```diff
@@ -82,23 +93,301 @@ enum id_entry_type {
...
+
+    str = rb_enc_str_new(RSTRING_PTR(str), RSTRING_LEN(str), enc);
+
+    rb_encoding *ascii = rb_usascii_encoding();
+    if (enc != ascii && sym_check_asciionly(str, false)) {
+        rb_enc_associate(str, ascii);
+    }
+    OBJ_FREEZE(str);
+
+    str = rb_fstring(str);
```
casperisfine, line 201: The code was already like that, but after you merge I might followup to use encindex instead of `rb_encoding *` here, because `rb_enc_get` and other functions that deal with `rb_encoding *` are way slower.

concurrent_set.c
peterzhu2118: Good catch!

concurrent_set.c
peterzhu2118: I changed it to skip rehashing if the hash is not 0.

peterzhu2118:
Closing because I manually merged it in 6b0e5de4e68bc60c10ce7974cb69f246678a9512 due to the rbs CI failing and preventing this PR from merging.

PR #13950, opened 2025-07-18
ZJIT: Run against all tests

st0012:
- `make zjit-test-all` now runs all test under `/test`, which includes several default gems' tests
- Some additional exclusions have been added to `/test/.excludes-zjit/`
- Since now `zjit-test-all` takes quite a while, I think we can run `zjit-check` instead of `zjit-test` on CI to get a faster feedback, even tho `test_zjit.rb` will be run twice (another time in `zjit-test-all`)

### DIFF
--- a/.github/workflows/zjit-macos.yml
+++ b/.github/workflows/zjit-macos.yml
@@ -32,7 +32,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - test_task: 'zjit-test'
+          - test_task: 'zjit-check'
             configure: '--enable-yjit=dev --enable-zjit'
 
           - test_task: 'ruby' # build test for combo build
@@ -84,7 +84,7 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: nextest@0.9
-        if: ${{ matrix.test_task == 'zjit-test' }}
+        if: ${{ matrix.test_task == 'zjit-check' }}
 
       - name: Install Rust # TODO(alan): remove when GitHub images catch up past 1.85.0
         run: rustup default 1.85.0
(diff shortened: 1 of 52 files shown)


tekknolagi:
Can you explain the YAML changes please? I'm not sure the current description clears it up for me. There are too many suites and commands for me to keep track of.

st0012:
`zjit-test` runs ZJIT's Rust tests. `zjit-check` runs both the Rust tests and `test_zjit.rb`.
So currently we have one job running ZJIT Rust tests, and let `zjit-test-all` run all Ruby tests, including `test_zjit.rb`.
After the change, one job will run both ZJIT Rust tests and `test_zjit.rb`, and `zjit-test-all` would still run `test_zjit.rb` but I think that's fine.

zjit/src/codegen.rs
```diff
@@ -102,6 +102,10 @@ fn gen_iseq_entry_point(iseq: IseqPtr) -> *const u8 {
     let cb = ZJITState::get_code_block();
     let code_ptr = gen_iseq_entry_point_body(cb, iseq);
 
+    // REMOVE ME
+    // Just to test if the crash is related to write barriers
+    unsafe { rb_gc_writebarrier_remember(iseq.into()) };
```
tekknolagi: There was another one, too, somewhere else in this file

st0012: I think in that commit the other change is unrelated:

```
-    asm.cmp(val, Opnd::UImm(expected.into()));
+    asm.cmp(val, Opnd::Value(expected));
```

which has also been shipped in https://github.com/ruby/ruby/pull/13935

tekknolagi: No, I mean there was another call to remember

st0012: I checked out back to the old commits, and the other 2 `rb_gc_writebarrier_remember` we added in pairing later became part of `append_gc_offsets`.
The other one in `profile.rs` was changed into `rb_gc_writebarrier` in #13935

PR #13953, opened 2025-07-18
ZJIT: Load stack relative slot before frame teardown

XrXr:
Or else the test returns garbage since it loads after
moving SP. Prior bad disassembly:

    # Block: bb0(v0, v1, v2, v3, v4, v5, v6, v7, v8)
    stp x29, x30, [sp, #-0x10]!
    mov x29, sp
    # bump C stack pointer
    sub sp, sp, #0x10
    # Insn: v10 Return v8
    # pop stack frame
    adds x19, x19, #0x38
    stur x19, [x20, #0x10]
    # restore C stack pointer
    add sp, sp, #0x10
    mov sp, x29
    ldp x29, x30, [sp], #0x10
    ldur x0, [sp]
    ret

### DIFF
--- a/doc/string/dump.rdoc
+++ b/doc/string/dump.rdoc
@@ -1,0 +1,52 @@
+Returns a printable version of +self+, enclosed in double-quotes:
+
+  'hello'.dump     # => "\"hello\""
+
+Certain special characters are rendered with escapes:
+
+  '"'.dump  # => "\"\\\"\""
+  '\\'.dump # => "\"\\\\\""
+
+Non-printing characters are rendered with escapes:
+
+  s = ''
+  s << 7   # Alarm (bell).
+  s << 8   # Back space.
+  s << 9   # Horizontal tab.
+  s << 10  # Line feed.
+  s << 11  # Vertical tab.
+  s << 12  # Form feed.
+  s << 13  # Carriage return.
+  s        # => "\a\b\t\n\v\f\r"
+  s.dump   # => "\"\\a\\b\\t\\n\\v\\f\\r\""
+
+If +self+ is encoded in UTF-8 and contains Unicode characters, renders Unicode
+characters in Unicode escape sequence:
+
+  'тест'.dump     # => "\"\\u0442\\u0435\\u0441\\u0442\""
+  'こんにちは'.dump # => "\"\\u3053\\u3093\\u306B\\u3061\\u306F\""
+
+If the encoding of +self+ is not ASCII-compatible (i.e., +self.encoding.ascii_compatible?+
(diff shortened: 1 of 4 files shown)


zjit/src/codegen.rs
```diff
@@ -890,14 +890,23 @@ fn gen_new_range(
...
     let incr_cfp = asm.add(CFP, RUBY_SIZEOF_CONTROL_FRAME.into());
     asm.mov(CFP, incr_cfp);
     asm.mov(Opnd::mem(64, EC, RUBY_OFFSET_EC_CFP), CFP);
 
+    // Load any stack relative memory since we're about to destroy the current frame
+    const NATIVE_SP_REG_NO: u8 = {
+        let Opnd::Reg(reg) = NATIVE_STACK_PTR else { unreachable!(); };
+        reg.reg_no
+    };
+    if let Opnd::Mem(lir::Mem { base: lir::MemBase::Reg(NATIVE_SP_REG_NO), .. }) = val {
```
tekknolagi: Not sure I am following. Don't we also need to check this post-regalloc then?

zjit/src/codegen.rs
```diff
@@ -890,14 +890,23 @@ fn gen_new_range(
...
+        val = asm.load(val);
```
k0kubun: Could we load it into C_RET_REG directly?

zjit/src/codegen.rs
XrXr: We'd need to check after regalloc if it supported spilling, but it doesn't right now. The SP-relative operands here are coming from param_opnd() in the same file.

https://github.com/ruby/ruby/blob/6b0e5de4e68bc60c10ce7974cb69f246678a9512/zjit/src/codegen.rs#L1110-L1119

We refer to the same `lir::Opnd` for each param.

The does feel like something LIR should provide an API for so we can't get it wrong. I'm adding a `MemBase::FrameBase` so we can refer to the right slot even after operations like cpush, but that won't get us all the way there.

tekknolagi: We support at least block param spilling now, though, after @k0kubun 's change

XrXr: I believe those are implemented here, not in the backend, so they're visible like this before regalloc.

zjit/src/codegen.rs
XrXr: Ah yes. That gets rid of the confusing check for SP base as well. [edc4572](https://github.com/ruby/ruby/pull/13953/commits/edc4572cd766e1a7c1b02d307678c0ef09803c47)

zjit/src/codegen.rs
k0kubun: Yeah. It should be good for now.

PR #13954, opened 2025-07-18
ZJIT: Remove no-op movs after register allocation

XrXr:
Previously `no_dead_mov_from_vreg` generated:

        0x0: ldur x0, [x0]
        0x4: mov x0, x0
        0x8: ret

    Because of phase ordering. Split couldn't recognize that the no-op mov
    because at that point it sees a `VReg`.

~Re the capstone change, it seems fairly easy to fix upstream. I'll try and toss
them a PR.~ https://github.com/capstone-rust/capstone-rs/pull/184

### DIFF
--- a/zjit/src/backend/arm64/mod.rs
+++ b/zjit/src/backend/arm64/mod.rs
@@ -1425,6 +1425,20 @@ mod tests {
     }
 
     #[test]
+    fn no_dead_mov_from_vreg() {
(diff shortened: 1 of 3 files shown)


st0012 (approved):

tekknolagi (approved):

k0kubun (approved):
makes sense

PR #13961, opened 2025-07-19
Add comments to autogen script

altxtech:
### Context

Was trying to understand how the build process works and found the autogen script hard to interpret. 

### Changes

- Added some explanatory comments
- Some formatting adjustments

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -5,22 +5,18 @@ PWD=
 
 # Figure out the source directory for this script
 # configure.ac should be in the same place
-srcdir() {
-  case "$0" in
-    */* ) dirname "$0" ;; # Called with path
-    *   ) echo ""    ;; # Otherwise
-  esac
-}
+case "$0" in
+    */* )  srcdir=`dirname "$0"` ;; # Called with path
+    *   )  srcdir="";; # Otherwise
+esac
 
 # If install-only is explicitly requested, disbale symlink flags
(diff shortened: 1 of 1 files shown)


nobu:
Could you tell what was hard to you?

> necessary evil for bash scripts

This file is a plain shell script, not a bash script.

altxtech:
For example, in the first part

```
case "$0" in
*/*) srcdir=`dirname $0`;;
*) srcdir="";;
esac
```

$0 --> Name of the script being run. Something I knew, but needed a few seconds to remember
\*/\* --> Pattern matching looking for slash. Didn't know what that was. Had to google it.

From there I understood here we are checking whether the script was called from a path or not. But it was not obvious.

nobu:
> For example, in the first part
> 
> ```
> case "$0" in
> */*) srcdir=`dirname $0`;;
> *) srcdir="";;
> esac
> ```
> 
> $0 --> Name of the script being run. Something I knew, but needed a few seconds to remember */* --> Pattern matching looking for slash. Didn't know what that was. Had to google it.
> 
> From there I understood here we are checking whether the script was called from a path or not. But it was not obvious.

I thought it was a very common idiom as sh scripts.
OK, it'd be good to add comments.

altxtech:
@nobu 

Some comments already help a lot.
Removed the wrapper functions so the original script is not changed too radically.

PR #13968, opened 2025-07-22
experiment with speeding up Time.local

catlee:
On macOS, methods like `Time.local` can be up to hundreds of times slower when run in a forked child process than when run in the parent process. This simple benchmark demonstrates the issue:

```ruby
require "benchmark/ips"

def time_local
  Time.local(2025,7,21,14,23)
end

Benchmark.ips do |x|
  x.report("non-forked") { time_local }
  x.report("forked") do |times|
    pid = fork do
      times.times { time_local }
    end
    Process.wait(pid)
  end
  x.compare!
end
```

```shell
ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM [arm64-darwin20]
          non-forked     7.873k i/100ms
              forked   131.000 i/100ms
          non-forked     77.203k (± 1.0%) i/s   (12.95 μs/i) -    393.650k in   5.099458s
              forked      1.248k (± 2.1%) i/s  (801.54 μs/i) -      6.288k in   5.042401s

Comparison:
          non-forked:    77202.9 i/s
              forked:     1247.6 i/s - 61.88x  slower
```

This impacts all kinds of things from logging, parsing times, working with zipfiles with `rubyzip`, etc.

This PR implements a small cache for `rb_localtime_r` to avoid multiple calls to libc's `localtime_r` for the same time value.

The cache is cleared if the timezone is changed.

I looked into caching the timezone offset by first calling gmtime_r and using the year/month/day/hour/minute as the cache key, but ran into problems handling leap seconds.


nobu:
That benchmark looks including `fork` and `waitpid` times and doesn't feel fair enough.

Rather, I feel that this problem should be reported to Apple.
Might be intentional since `fork` is declared as deprecated, though.

```C
#include <stdio.h>
#include <sys/types.h>
#include <sys/wait.h>
#include <time.h>
#include <unistd.h>

void
localtime_benchmark(const char *cond)
{
    struct timespec ts, te;
    clock_gettime(CLOCK_MONOTONIC , &ts);
    for (long n = 0; n < 10000; ++n) {
        time_t lt = 1753075380;
        struct tm t;
        localtime_r(&lt, &t);
    }
    clock_gettime(CLOCK_MONOTONIC , &te);
    time_t diff = te.tv_sec - ts.tv_sec;
    unsigned long ndiff = te.tv_nsec;
    if (te.tv_nsec < ts.tv_nsec) {
        --diff;
        ndiff += 1000000000UL;
    }
    ndiff -= ts.tv_nsec;
    printf("%s: %ld.%.9lu\n", cond, diff, ndiff);
}

int
main(void)
{
    localtime_benchmark("non-forked");
    pid_t forked = fork();
    if (forked) {
        int stat;
        waitpid(forked, &stat, 0);
    }
    else {
        localtime_benchmark("    forked");
    }
    return 0;
}
```

```
non-forked: 0.001993000
    forked: 2.986874000
```

time.c
```diff
@@ -764,12 +772,265 @@ update_tz(void)
...
+/* Calculate day of year from year/month/day */
+static int
+calculate_yday(int year, int month, int day)
+{
+    int yday = 0;
+    for (int m = 1; m < month; m++) {
+        yday += get_days_in_month(year, m);
+    }
+    return yday + day - 1; /* tm_yday is 0-based */
+}
```
nobu: These look like duplicates to the existing functions.

catlee:
> Rather, I feel that this problem should be reported to Apple.

Indeed. I did some searching around and found a few references which make me think this won't be fixed:

https://developer.apple.com/forums/thread/747499
https://developer.apple.com/forums/thread/737464?answerId=764686022#764686022
https://stackoverflow.com/questions/27932330/why-is-tzset-a-lot-slower-after-forking-on-mac-os-x

And from fork's [manpage](https://www.unix.com/man_page/osx/2/fork/):
> CAVEATS
     There are limits to what you can do in the child process.	To be totally safe you should restrict yourself to only executing async-signal
     safe operations until such time as one of the exec functions is called.  All APIs, including global data symbols, in any framework or library
     should be assumed to be unsafe after a fork() unless explicitly documented to be safe or async-signal safe.  If you need to use these frame-
     works in the child process, you must exec.  In this situation it is reasonable to exec yourself.

AIUI there is IPC signaling in place to notify processes of timezone changes, and these don't work after fork() before exec().

time.c
```diff
@@ -764,12 +777,158 @@ update_tz(void)
...
+    if (month < 3) {
+        month += 12;
+        year--;
+    }
+    int k = year % 100;
+    int j = year / 100;
+    int h = (day + (13 * (month + 1)) / 5 + k + k / 4 + j / 4 - 2 * j) % 7;
+    /* Convert to tm_wday format (Sunday = 0) */
+    return (h + 6) % 7;
+}
```
nobu: You can use `calc_wday`.

time.c
```diff
@@ -703,6 +704,11 @@ static struct vtm *localtimew(wideval_t timew, struct vtm *result);
 
 static int leap_year_p(long y);
 #define leap_year_v_p(y) leap_year_p(NUM2LONG(modv((y), INT2FIX(400))))
+static int calc_tm_yday(long tm_year, int tm_mon, int tm_mday);
+
```
nobu, line 709:
```suggestion
static int calc_wday(int year_mod400, int month, int day);

```

time.c
```diff
@@ -860,6 +1020,88 @@ static const int8_t leap_year_days_in_month[] = {
...
+            }
+        }
+
+        /* Update tm structure */
+        tm->tm_year = year - 1900;
+        tm->tm_mon = month - 1;
+        tm->tm_mday = day;
+
+        /* Recalculate wday and yday */
+        tm->tm_wday = calculate_wday(year, month, day);
```
nobu:
```suggestion
        tm->tm_wday = calc_wday(year % 400, month, day);
```

time.c
```diff
@@ -764,12 +777,158 @@ update_tz(void)
...
+#define OFFSET_CACHE_ZONE_LENGTH 64
+
+/* Cache key: 15-minute precision */
+typedef struct {
+    int year;
+    int month;
+    int day;
+    int hour;
+    int quarter;  /* 0-3 representing 00, 15, 30, 45 minute intervals */
+} offset_cache_key;
```
nobu: I guess it would be possible to pack the cache keys, so that the validness can be checked atomically.

catlee: Great idea. I think now we can probably just use our time_t value/15 for our cache key?

I haven't quite convinced myself that it's safe to use 15 minute quarters to cache by..... There are some historical timezones with odd offsets. I don't know of any yet where a dst boundary doesn't happen on a 15 minute boundary in UTC time though.

nobu: Perhaps modern time zones do not have such odd UTC offsets, but "zdump(1)" refers to "Europe/Astrahan".

```console
$ LC_TIME=C /bin/date -z Europe/Astrakhan -j {-f,+}%Y-%m-%dT%H:%M:%S%Z 1924-04-30T00:00:00GMT
1924-04-30T03:12:12LMT
```

nobu: I think it'd be OK to abandon such cases, i.e., do not cache if the `gmtoff % (15 * 60) != 0`.

configure.ac
```diff
@@ -2253,6 +2253,20 @@ AC_CHECK_FUNCS(waitpid)
...
+	[macos_localtime_cache=$enableval], [macos_localtime_cache=no])
+AS_IF([test "$macos_localtime_cache" = yes], [
+    AS_CASE(["$target_os"],
+        [darwin*], [
+            AC_DEFINE(ENABLE_MACOS_LOCALTIME_CACHE, 1, [Enable macOS localtime cache optimization])
+            AC_MSG_NOTICE([macOS localtime cache optimization enabled])
+        ],
+        [AC_MSG_WARN([--enable-macos-localtime-cache is only supported on macOS, ignoring])]
+    )
+])
```
nobu, line 2267: This may not be limited to macOS in the future.
And do you have any reason not to enable it on macOS by the default?

```suggestion
AC_ARG_ENABLE(localtime-cache,
	AS_HELP_STRING([--enable-localtime-cache],
		       [enable localtime cache optimization to improve forked process performance]),
	[localtime_cache=$enableval],
	[AS_CASE(["$target_os"], [darwin*], [localtime_cache=yes], [localtime_cache=no])])
AS_IF([test "$localtime_cache" = yes], [
    AC_DEFINE(ENABLE_LOCALTIME_CACHE, 1, [Enable localtime cache optimization])
    AC_MSG_NOTICE([localtime cache optimization enabled])
])
```

catlee: I checked on Linux and it doesn't seem to have the same performance problem running localtime in a forked process.

time.c
catlee: I don't think we can use value/15 in general.

E.g. in [Newfoundland](https://www.timeanddate.com/time/change/canada/newfoundland-and-labrador?year=2011) up until 2011, the daylight savings time change happened at 12:01:00, which is within the same 15-minute chunk of time as 12:00:00 or 12:02:00.

I do think a minute-based cache is generally safe.

catlee:
@akr how does this look?

samuel-williams-shopify:
I have tested this PR locally and can confirm the results. However, the best results come when hitting the same and/or similar dates, but the worst case scenario is still present - i.e. when we benchmark a wide variety of different dates, the cache is almost completely ineffective (unfortunately).

With only 64 cache entries and a hash function that causes collisions, the cache can only handle very tightly clustered timestamps. The moment you have timestamps spanning:

- More than ~64 different minutes, or
- Hash collisions (which happen quickly)

You're back to hitting the slow macOS `localtime_r` path on almost every call.

Given the extra complexity this introduces, I'm slightly negative on this fix as proposed... and I don't have a good idea about how to improve this either :( For reference, I did file a bug report with Apple... but never heard back from them.

## Benchmark

Here are the test cases I explored with Claude:

```ruby
def run_test(name, &block)
  # Non-forked
  start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  10000.times { block.call }
  end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  non_forked = end_time - start_time

  # Forked
  start_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  pid = fork do
    10000.times { block.call }
  end
  Process.wait(pid)
  end_time = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  forked = end_time - start_time

  slowdown = forked / non_forked
  printf("%-50s: forked %6.2fx slower (forked: %.3fs, non-forked: %.3fs)\n",
         name, slowdown, forked, non_forked)

  slowdown
end

puts "=" * 90
puts "Real-World Scenario Testing"
puts "=" * 90
puts

# Scenario 1: Processing a log file (entries clustered in time)
# Logs typically have many entries per second, so we'd hit cache
puts "Scenario 1: Processing log file (entries within same second)"
base_time = Time.local(2025, 10, 6, 14, 30, 0)
log_times = 100.times.map { |i| base_time + (i / 10.0) }  # 10 entries per second for 10 seconds
slowdown1 = run_test("  Log processing (clustered)") do
  time = log_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min, time.sec)
end

# Scenario 2: Processing timestamps from different sources throughout a day
puts "\nScenario 2: Processing events throughout a day"
day_times = 1440.times.map { |minute| Time.local(2025, 10, 6, minute / 60, minute % 60) }
slowdown2 = run_test("  Events spread across 1 day (1440 minutes)") do
  time = day_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min)
end

# Scenario 3: Processing file timestamps from a zipfile (could be from different days)
puts "\nScenario 3: Zipfile with files from various dates"
file_times = 200.times.map { |i| Time.local(2025, rand(1..10), rand(1..28), rand(0..23), rand(0..59)) }
slowdown3 = run_test("  Zipfile timestamps (mixed dates)") do
  time = file_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min)
end

# Scenario 4: Web server handling requests (timestamps very close together)
puts "\nScenario 4: Web server processing requests"
now = Time.now
request_times = 50.times.map { |i| now + (i * 0.1) }  # Request every 100ms for 5 seconds
slowdown4 = run_test("  Web requests (within 5 seconds)") do
  time = request_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min, time.sec)
end

# Scenario 5: Data processing with timestamps from the past month
puts "\nScenario 5: Analytics processing last month's data"
month_times = 30.times.map { |day|
  Time.local(2025, 9, day + 1, rand(0..23), rand(0..59))
}
slowdown5 = run_test("  Monthly data processing") do
  time = month_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min)
end

# Scenario 6: Parsing diverse historical data
puts "\nScenario 6: Parsing historical records (wide time range)"
historical_times = 1000.times.map { |i|
  Time.local(rand(2000..2025), rand(1..12), rand(1..28), rand(0..23), rand(0..59))
}
slowdown6 = run_test("  Historical data (25 years)") do
  time = historical_times.sample
  Time.local(time.year, time.month, time.day, time.hour, time.min)
end

puts
puts "=" * 90
puts "Analysis"
puts "=" * 90

practical_scenarios = [
  ["Log processing", slowdown1],
  ["Web requests", slowdown4],
]

limited_scenarios = [
  ["Daily events", slowdown2],
  ["Zipfile timestamps", slowdown3],
  ["Monthly data", slowdown5],
  ["Historical data", slowdown6],
]

puts "\n✅ Cache is EFFECTIVE for these scenarios:"
practical_scenarios.each do |name, slowdown|
  puts "  #{name}: #{slowdown < 2 ? "NO slowdown" : "#{slowdown.round(1)}x slower"}"
end

puts "\n❌ Cache has LIMITED benefit for these scenarios:"
limited_scenarios.each do |name, slowdown|
  puts "  #{name}: #{slowdown.round(1)}x slower (still hitting macOS issue)"
end

avg_limited = limited_scenarios.map { |_, s| s }.sum / limited_scenarios.size
puts "\n📊 Conclusion:"
puts "  - Cache helps ONLY when timestamps are very tightly clustered (same second/minute)"
puts "  - For moderately diverse timestamps: still #{avg_limited.round(0)}x slower"
puts "  - 64-entry cache with hash collisions = limited real-world applicability"
```

### Results

```
(pr-13968)> make runruby
RUBY_ON_BUG='gdb -x ./.gdbinit -p' ./miniruby -I./lib -I. -I.ext/common  ./tool/runruby.rb --extout=.ext  -- --disable-gems  ./test.rb 
==========================================================================================
Real-World Scenario Testing
==========================================================================================

Scenario 1: Processing log file (entries within same second)
  Log processing (clustered)                      : forked   1.14x slower (forked: 0.006s, non-forked: 0.005s)

Scenario 2: Processing events throughout a day
  Events spread across 1 day (1440 minutes)       : forked 263.66x slower (forked: 2.136s, non-forked: 0.008s)

Scenario 3: Zipfile with files from various dates
  Zipfile timestamps (mixed dates)                : forked 232.91x slower (forked: 1.998s, non-forked: 0.009s)

Scenario 4: Web server processing requests
  Web requests (within 5 seconds)                 : forked   1.42x slower (forked: 0.006s, non-forked: 0.004s)

Scenario 5: Analytics processing last month's data
  Monthly data processing                         : forked 178.69x slower (forked: 1.106s, non-forked: 0.006s)

Scenario 6: Parsing historical records (wide time range)
  Historical data (25 years)                      : forked 241.79x slower (forked: 2.128s, non-forked: 0.009s)
```

✅ Cache is EFFECTIVE for these scenarios:
  - Log processing: NO slowdown
  - Web requests: NO slowdown

❌ Cache has LIMITED benefit for these scenarios:
  - Daily events: 263.7x slower (still hitting macOS issue)
  - Zipfile timestamps: 232.9x slower (still hitting macOS issue)
  - Monthly data: 178.7x slower (still hitting macOS issue)
  - Historical data: 241.8x slower (still hitting macOS issue)

📊 Conclusion:
  - Cache helps ONLY when timestamps are very tightly clustered (same second/minute)
  - For moderately diverse timestamps: still 229x slower
  - 64-entry cache with hash collisions = limited real-world applicability

catlee:
> With only 64 cache entries and a hash function that causes collisions, the cache can only handle very tightly clustered timestamps. The moment you have timestamps spanning:
> 
> * More than ~64 different minutes, or
> * Hash collisions (which happen quickly)

We can certainly play around with the cache size.

I had two primary assumptions when working on this:
* You are likely to be calling localtime frequently for increasing timestamps, e.g. via calls to a logging framework from a server process
* Things like zip files are likely to contain clusters of the same timestamp rather than having completely random timestamps

One approach to mitigate cache eviction for these two different cases could be to use a different cache for times close to now vs times in the past; or adjust the cache indexing function to skew times close to the present closer together.

configure.ac
```diff
@@ -2252,6 +2252,20 @@ AC_CHECK_FUNCS(waitpid)
...
+		       [enable localtime cache optimization for macOS to improve forked process performance]),
+	[macos_localtime_cache=$enableval], [macos_localtime_cache=yes])
+AS_IF([test "$macos_localtime_cache" = yes], [
+    AS_CASE(["$target_os"],
+        [darwin*], [
+            AC_DEFINE(ENABLE_MACOS_LOCALTIME_CACHE, 1, [Enable macOS localtime cache optimization])
+            AC_MSG_NOTICE([macOS localtime cache optimization enabled])
+        ],
+        [AC_MSG_WARN([--enable-macos-localtime-cache is only supported on macOS, ignoring])]
+    )
```
nobu, line 2266: It is not nice this warning is shown unless this option is explicitly disabled on other platforms.
https://github.com/ruby/ruby/actions/runs/18602388839/job/53043740969#step:6:531

```suggestion
	[AS_CASE(["$target_os"],
	    [darwin*], [macos_localtime_cache=$enableval],
	    [*], [AC_MSG_WARN([--enable-macos-localtime-cache is only supported on macOS, ignoring])]
	)],
	[AS_CASE(["$target_os"], [darwin*], [macos_localtime_cache=yes], [macos_localtime_cache=no])])
AS_IF([test "$macos_localtime_cache" = yes], [
    AC_DEFINE(ENABLE_MACOS_LOCALTIME_CACHE, 1, [Enable macOS localtime cache optimization])
    AC_MSG_NOTICE([macOS localtime cache optimization enabled])
```

time.c
```diff
@@ -703,6 +704,12 @@ static struct vtm *localtimew(wideval_t timew, struct vtm *result);
 
 static int leap_year_p(long y);
 #define leap_year_v_p(y) leap_year_p(NUM2LONG(modv((y), INT2FIX(400))))
+static int calc_tm_yday(long tm_year, int tm_mon, int tm_mday);
+static int calc_wday(int year_mod400, int month, int day);
+
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 710: It doesn't look nice to repeat multiple conditions many times.

```suggestion
#if !defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
# undef ENABLE_MACOS_LOCALTIME_CACHE
#endif

#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

time.c
```diff
@@ -746,6 +753,10 @@ get_tzname(int dst)
 }
 #endif
 
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 756:
```suggestion
#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

time.c
```diff
@@ -754,6 +765,9 @@ ruby_reset_timezone(const char *val)
     w32_tz.use_tzkey = !val || !*val;
 #endif
     ruby_reset_leap_second_info();
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 768:
```suggestion
#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

time.c
```diff
@@ -764,12 +778,107 @@ update_tz(void)
     tzset();
 }
 
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 781:
```suggestion
#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

time.c
```diff
@@ -764,12 +778,107 @@ update_tz(void)
...
+#endif
+
 static struct tm *
 rb_localtime_r(const time_t *t, struct tm *result)
 {
 #if defined __APPLE__ && defined __LP64__
     if (*t != (time_t)(int)*t) return NULL;
 #endif
+
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 821:
```suggestion
#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

time.c
```diff
@@ -779,6 +888,7 @@ rb_localtime_r(const time_t *t, struct tm *result)
         if (tmp) *result = *tmp;
     }
 #endif
+#endif /* __APPLE__ */
```
nobu, line 891:
```suggestion
#endif /* ENABLE_MACOS_LOCALTIME_CACHE */
```

time.c
```diff
@@ -860,6 +970,88 @@ static const int8_t leap_year_days_in_month[] = {
 #define days_in_month_in(y) days_in_month_of(leap_year_p(y))
 #define days_in_month_in_v(y) days_in_month_of(leap_year_v_p(y))
 
+#if defined(__APPLE__) && defined(ENABLE_MACOS_LOCALTIME_CACHE)
```
nobu, line 973:
```suggestion
#if defined(ENABLE_MACOS_LOCALTIME_CACHE)
```

PR #13974, opened 2025-07-22
POC: Explicit memory ordering

jhawthorn:
This adds atomic helpers which use explicit memory ordering. For example:

```
rbimpl_atomic_sub_explicit(&set->size, 1, RUBY_ATOMIC_RELAXED);
```

I defined these using macros in hopes of making these easier and less error prone to declare for a variety of operations and types. Explicit ordering is only available for platforms with `stdatomic.h` (C11) or GCC's `__atomic`. When not available they fall back to the  non-explicit rbimpl_* definitions which should implement the strictest SEQ_CST ordering.

Note that these are being added to `ruby_atomic.h`, so these are only available within the Ruby VM but not native extensions.

### DIFF
--- a/ruby_atomic.h
+++ b/ruby_atomic.h
@@ -73,6 +73,8 @@ rbimpl_atomic_##name##load_explicit(type *ptr, int memory_order) \
 
 /* Generate atomic load function with explicit memory ordering */
 DEFINE_ATOMIC_LOAD_EXPLICIT(, rb_atomic_t)
+DEFINE_ATOMIC_LOAD_EXPLICIT(value_, VALUE)
+DEFINE_ATOMIC_LOAD_EXPLICIT(ptr_, void *)
 
 #undef DEFINE_ATOMIC_LOAD_EXPLICIT
 #undef DEFINE_ATOMIC_LOAD_EXPLICIT_BODY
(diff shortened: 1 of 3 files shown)


ruby_atomic.h
```diff
@@ -27,16 +49,107 @@
...
+#define DEFINE_ATOMIC_STORE_EXPLICIT(gcc_op, stdatomic_op, type_prefix, type) \
+static inline void \
+rbimpl_atomic_##type_prefix##stdatomic_op##_explicit(volatile type *ptr, type val, int memory_order) \
+{ \
+    ATOMIC_OP_EXPLICIT_BODY(gcc_op, stdatomic_op, ptr, val, memory_order, type, type_prefix); \
+}
+
+/* Generate atomic operations with explicit memory ordering */
+DEFINE_ATOMIC_STORE_EXPLICIT(store_n, store, , rb_atomic_t)
+DEFINE_ATOMIC_OP_EXPLICIT(add_fetch, add, , rb_atomic_t)
```
tavianator, line 112: I don't think there's an `atomic_add_explicit()` in `<stdatomic.h>` (https://en.cppreference.com/w/c/header/stdatomic.html)

PR #13977, opened 2025-07-22
ZJIT: Use rb_vm_env_write() when writing locals in higher scopes

XrXr:
We weren't firing write barriers before when writing to imemo/env
objects. Wbcheck caught this with test/ruby/test_refinement.rb:

    ruby -v: ruby 3.5.0dev (2025-07-22T17:05:58Z wbcheck 2569a80954) +ZJIT dev +PRISM +GC[wbcheck] [x86_64-linux]
    WBCHECK ERROR: Missed write barrier detected!
      Parent object: 0x558de9f4e6e0 (wb_protected: true)
        rb_obj_info_dump: 0x0000558de9f4e6e0 T_IMEMO/<env>
      Reference counts - snapshot: 3, writebarrier: 0, current: 4, missed: 1
      Missing reference to: 0x558decf37c30
        rb_obj_info_dump: 0x0000558decf37c30 method/UnboundMethod method

    WBCHECK SUMMARY: Found 1 objects with missed write barriers (1 total violations)

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1,5 +1,6 @@
 use std::cell::Cell;
 use std::rc::Rc;
+use std::ffi::{c_int};
 
 use crate::asm::Label;
 use crate::backend::current::{Reg, ALLOC_REGS};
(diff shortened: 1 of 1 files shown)


k0kubun:
Similarly to YJIT, could we skip the call when the `val` is known to be an immediate? I guess we could let the HIR generator check the type and set a flag to `Insn::SetLocal` like `write_barrier: true`.

Feel free to merge this first and leave it for another PR though.

k0kubun (approved):

tekknolagi:
tbh we don't need a flag in the instruction -- we can make the decision at HIR->LIR codegen time with the type of `val`

tekknolagi:
@st0012 does this (with the class.h fix) fix the problems you see and enable more tests?

k0kubun:
"with the type of `val`", yes. What I'm saying is that I thought the HIR generator has better/more access to the type of `val`, but I might be wrong. If we want to optimize only the case that it points to `Opnd::Value` (likely created by `Insn::Const`) and the `VALUE` of it is an immediate, the HIR can stay as is. I was worried about cases like `val` pointing to `Insn::GuardType`, but maybe `find()` resolves it into `Insn::Const` when possible?

XrXr:
I figure things like GuardType on a constant should be handled by optimizer, so HIR->LIR translation should have the same visibility on whether something is constant.

tekknolagi:
I mean that in HIR->LIR you can do `if val.is_subtype(types::Immediate)` and not care at all about what instruction it comes from. That will use all available information from the optimizer

tekknolagi:
That gives (much) more information than looking at the exact value, which we might not know. It would handle if it's any kind of Fixnum, for example

k0kubun:
> I mean that in HIR->LIR you can do `if val.is_subtype(types::Immediate)`

I'm not sure if I follow. `is_subtype()` is a function on `Type`, but `val` is `hir::InsnId` in HIR and `lir::Opnd` in codegen. How do you call `is_subtype()` on `val`?

k0kubun:
Ah, I guess `Function::insn_types` keeps them for each `InsnId`. Nvm then. I guess we don't need to bake type-related information in the instruction itself because the function has it separately.

PR #13978, opened 2025-07-22
ZJIT: Guide WB skipping for Insn::SetLocal using HIR type info

XrXr:
Also tweak the opnd!() macro so that it works with both an `&InsnId` and
an `InsnId`.

Following discussions in https://github.com/ruby/ruby/pull/13977


zjit/src/codegen.rs
```diff
@@ -353,7 +353,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::SetGlobal { id, val, state: _ } => return Some(gen_setglobal(asm, *id, opnd!(val))),
         Insn::GetGlobal { id, state: _ } => gen_getglobal(asm, *id),
         &Insn::GetLocal { ep_offset, level } => gen_getlocal_with_ep(asm, ep_offset, level)?,
-        Insn::SetLocal { val, ep_offset, level } => return gen_setlocal_with_ep(asm, opnd!(val), *ep_offset, *level),
+        &Insn::SetLocal { val, ep_offset, level } => return gen_setlocal_with_ep(asm, opnd!(val), function.is_a(val, types::Immediate), ep_offset, level),
```
k0kubun: > function.is_a(val, types::Immediate)

I feel like this kind of type checking (not just fetching the type, but checking that it's "immediate") should be part of the codegen logic (i.e. live inside `gen_setlocal_with_ep()`). You generally wouldn't want to go back and forth between `gen_insn()` and `gen_setlocal_with_ep()` too often when you read/write the logic of codegen.

One way to do that would be to give `val_type: Type` as an argument. Another way would be to just pass around a reference to `function` (which seems fine to me) if we want to use the `Function::is_a` API.

k0kubun (approved):
nice improvement on `opnd!` :+1:

XrXr:
Now this crashes when marking an Iseq:

```
  /home/runner/work/ruby/ruby/src/test/ruby/test_weakkeymap.rb:142: [BUG] Segmentation fault at 0x0000000000000048
  ruby 3.5.0dev (2025-07-22T23:12:12Z pull/13978/merge 844ed0ff83) +ZJIT dev +PRISM [x86_64-linux]
  
  /home/runner/work/ruby/ruby/build/ruby(rb_print_backtrace+0x8) [0x55e25b22c1e6] ../src/vm_dump.c:843
  /home/runner/work/ruby/ruby/build/ruby(rb_vm_bugreport) ../src/vm_dump.c:1175
  /home/runner/work/ruby/ruby/build/ruby(rb_bug_for_fatal_signal+0xf8) [0x55e25b1e2d58] ../src/error.c:1130
  /home/runner/work/ruby/ruby/build/ruby(sigsegv+0x46) [0x55e25acc0786] ../src/signal.c:946
  /lib/x86_64-linux-gnu/libc.so.6(0x7f2b99a42520) [0x7f2b99a42520]
  /home/runner/work/ruby/ruby/build/ruby(RVALUE_WB_UNPROTECTED+0x17) [0x55e25aba5c10] ../src/gc/default/default.c:1195
  /home/runner/work/ruby/ruby/build/ruby(rgengc_check_relation) ../src/gc/default/default.c:4300
  /home/runner/work/ruby/ruby/build/ruby(RVALUE_MARKED+0x0) [0x55e25abb1eb2] ../src/gc/default/default.c:4369
  /home/runner/work/ruby/ruby/build/ruby(gc_mark_set) ../src/gc/default/default.c:4311
  /home/runner/work/ruby/ruby/build/ruby(gc_mark) ../src/gc/default/default.c:4370
  /home/runner/work/ruby/ruby/build/ruby(rb_zjit_iseq_mark+0x193) [0x55e25ade7a13] zjit/src/gc.rs:88
```

This could be some bad HIR optimization or some other GC bug...

zjit/src/codegen.rs
```diff
@@ -445,21 +445,19 @@ fn gen_getlocal_with_ep(asm: &mut Assembler, local_ep_offset: u32, level: u32) -
...
-        // If we're writing a constant, non-heap VALUE, do a raw memory write without
-        // running write barrier.
-        lir::Opnd::Value(const_val) if const_val.special_const_p() => {
-            let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).ok()?);
-            asm.mov(Opnd::mem(64, ep, offset), val);
-        }
+
+    // When we've proved that we're writing an immediate,
+    // we can skip the write barrier.
+    if function.is_a(val, types::Immediate) {
```
tekknolagi: nit: `.is_immediate()`

tekknolagi:
```patch
diff --git a/zjit/src/codegen.rs b/zjit/src/codegen.rs
index 11a83bed5d..41fd06ce5b 100644
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -442,6 +442,13 @@ fn gen_getlocal_with_ep(asm: &mut Assembler, local_ep_offset: u32, level: u32) -
     Some(asm.load(Opnd::mem(64, ep, offset)))
 }

+#[unsafe(no_mangle)]
+pub extern "C" fn assert_immediate(val: VALUE) {
+    if !val.special_const_p() {
+        panic!("Expected an immediate value, got: {val}");
+    }
+}
+
 /// Set a local variable from a higher scope or the heap. `local_ep_offset` is in number of VALUEs.
@@ -452,6 +459,7 @@ fn gen_setlocal_with_ep(asm: &mut Assembler, jit: &JITState, function: &Function
     // we can skip the write barrier.
     if function.is_a(val, types::Immediate) {
         let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).ok()?);
+        asm_ccall!(asm, assert_immediate, jit.get_opnd(val)?);
         asm.mov(Opnd::mem(64, ep, offset), jit.get_opnd(val)?);
     } else {
         // We're potentially writing a reference to an IMEMO/env object,
```

tekknolagi:
I wonder: if you rebase this, will you still see crashes? John has landed some WB fixes elsewhere in Ruby since then.

tekknolagi:
Looks like it passes 👀

st0012 (approved):

PR #13981, opened 2025-07-23
[DOC] Tweaks for String#each_grapheme_cluster

BurdetteLamar:


doc/string/each_grapheme_cluster.rdoc
```diff
@@ -1,12 +1,35 @@
...
-  s.each_grapheme_cluster {|gc| print gc, ' ' }
+  a = []
+  'hello'.each_grapheme_cluster do |grapheme_cluster|
+    a.push(grapheme_cluster)
+  end
+  a  # => ["h", "e", "l", "l", "o"]
 
-Output:
+  a = []
+  # => []
```
peterzhu2118: Should this line be removed? It's showing the output of an empty array.

doc/string/each_grapheme_cluster.rdoc
```diff
@@ -1,12 +1,35 @@
...
+  a = []
+  'こんにちは'.each_grapheme_cluster do |grapheme_cluster|
+    a.push(grapheme_cluster)
+  end
+  a # => ["こ", "ん", "に", "ち", "は"]
 
-Returns an enumerator if no block is given.
+  a = []
+  s = "\u0061\u0308-\u0062\u0308-\u0063\u0308"
+  # => "ä-b̈-c̈"
```
peterzhu2118: Why are we using unicode here instead of just directly using the string `"ä-b̈-c̈"`.

BurdetteLamar: Search me.  These were in the previous example.  Should we remove this last section (as redundant)?

peterzhu2118: Sure, I'm fine with either way.

doc/string/each_grapheme_cluster.rdoc
BurdetteLamar: Removed.

doc/string/each_grapheme_cluster.rdoc
BurdetteLamar: Removed.

PR #13983, opened 2025-07-23
ZJIT: Fix clobbering register for `self` in gen_entry_params()

XrXr:
Previously, for 8+ params we wound up clobbering the self param when
putting the last param in memory in the JIT entry point:

    # ZJIT entry point: a@../test.rb:5
    <snip>
    ldur x0, [x19, #0x18]
    # set method params: 8
    ldur x1, [x21, #-0x58]
    ldur x2, [x21, #-0x50]
    ldur x3, [x21, #-0x48]
    ldur x4, [x21, #-0x40]
    ldur x5, [x21, #-0x38]
    ldur x11, [x21, #-0x30]
    ldur x12, [x21, #-0x28]
    ldur x0, [x21, #-0x20]
    stur x0, [sp, #-0x20]
    bl #0x11e17018c

Doing the memcpys for parameters in memory first avoids this clobbering.

    # set method params: 8
    ldur x0, [x21, #-0x20]
    stur x0, [sp, #-0x20]
    ldur x12, [x21, #-0x28]
    ldur x11, [x21, #-0x30]
    ldur x5, [x21, #-0x38]
    ldur x4, [x21, #-0x40]
    ldur x3, [x21, #-0x48]
    ldur x2, [x21, #-0x50]
    ldur x1, [x21, #-0x58]
    ldur x0, [x19, #0x18]

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -811,6 +811,12 @@ class TestZJIT < Test::Unit::TestCase
       def a(n1,n2,n3,n4,n5,n6,n7,n8) = n8
       a(1,1,1,1,1,1,1,0)
     }
+
+    # self param with spilled param
+    assert_compiles '"main"', %q{
+      def a(n1,n2,n3,n4,n5,n6,n7,n8) = self
+      a(1,0,0,0,0,0,0,0).to_s
+    }
   end
 
   def test_opt_aref_with
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -187,6 +187,10 @@ fn gen_entry(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function, function_pt
     asm.frame_teardown();
     asm.cret(C_RET_OPND);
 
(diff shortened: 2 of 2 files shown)


k0kubun (approved):
Looks good to me. Thanks for fixing this.

I thought the `LiveReg` instruction issued by `gen_param()` would reserve the register during `gen_entry_params()` and thus writing memory would not use those registers.

What happened in reality seems to be that nobody used the result of `LiveReg` in the JIT entry code, so the live range was empty. We would need to insert a new LIR instruction right before `asm.ccall` if we want to make it happen.

But then exhausting all registers first would make it impossible to assign spilled arguments (unless you explicitly use `SCRATCH_REG`), so we'd need this change anyway. And it obviates the need of such a new instruction in JIT entry for now.

XrXr:
I had another solution that used parallel_mov with memory operands and the registers, which also worked and feels cleaner live range wise, but that basically gives the same code. It's stretching what parallel_mov is for though, and it's not free to build up the Vec to pass to it. This is like a pre-solved version of the parallel move.

PR #13986, opened 2025-07-23
[DOC] Update documentation of `ObjectSpace.memsize_of`

st0012:
`RVALUE` no longer exists and objects can now have various sizes due to VWA. So I think this API's doc should be updated.

### DIFF
--- a/ext/objspace/objspace.c
+++ b/ext/objspace/objspace.c
@@ -38,10 +38,11 @@
  *  information as only a *HINT*. Especially, the size of +T_DATA+ may not be
  *  correct.
  *
- *  This method is only expected to work with C Ruby.
+ *  This method is only expected to work with CRuby.
  *
- *  From Ruby 2.2, memsize_of(obj) returns a memory size includes
- *  sizeof(RVALUE).
+ *  From Ruby 3.2 with Variable Width Allocation, it returns the actual slot
+ *  size used plus any additional memory allocated outside the slot (such
+ *  as external strings, arrays, or hash tables).
  */
 
 static VALUE


ext/objspace/objspace.c
```diff
@@ -38,10 +38,14 @@
...
- *  This method is only expected to work with C Ruby.
- *
- *  From Ruby 2.2, memsize_of(obj) returns a memory size includes
- *  sizeof(RVALUE).
+ *  This method is only expected to work with CRuby.
+ *
+ *  From Ruby 2.2, memsize_of(obj) returns a memory size that includes
+ *  the slot size occupied by the object. Prior to Ruby 3.2, this was a
+ *  fixed size (sizeof(RVALUE)). From Ruby 3.2 with Variable Width
+ *  Allocation, this is the actual slot size used (1×, 2×, 4×, 8×, or
```
st0012: I got this information from `rb_gc_impl_objspace_init`: https://github.com/Shopify/ruby/blob/9eac1c3a888088912fcc1cbb6fc677fd1ad5683c/gc/default/default.c#L9331-L9337

Let me know if it's correct?

ext/objspace/objspace.c
```diff
@@ -38,10 +38,14 @@
...
- *  From Ruby 2.2, memsize_of(obj) returns a memory size includes
- *  sizeof(RVALUE).
+ *  This method is only expected to work with CRuby.
+ *
+ *  From Ruby 2.2, memsize_of(obj) returns a memory size that includes
+ *  the slot size occupied by the object. Prior to Ruby 3.2, this was a
+ *  fixed size (sizeof(RVALUE)). From Ruby 3.2 with Variable Width
+ *  Allocation, this is the actual slot size used (1x, 2x, 4x, 8x, or
+ *  16x BASE_SLOT_SIZE), plus any additional memory allocated outside
+ *  the slot (such as external strings, arrays, or hash tables).
```
peterzhu2118: We're only documenting for the latest version of Ruby, so I don't think we should include info about prior versions.

ext/objspace/objspace.c
```diff
@@ -38,10 +38,14 @@
...
- *
- *  From Ruby 2.2, memsize_of(obj) returns a memory size includes
- *  sizeof(RVALUE).
+ *  This method is only expected to work with CRuby.
+ *
+ *  From Ruby 2.2, memsize_of(obj) returns a memory size that includes
+ *  the slot size occupied by the object. Prior to Ruby 3.2, this was a
+ *  fixed size (sizeof(RVALUE)). From Ruby 3.2 with Variable Width
+ *  Allocation, this is the actual slot size used (1x, 2x, 4x, 8x, or
+ *  16x BASE_SLOT_SIZE), plus any additional memory allocated outside
```
peterzhu2118: I don't think we should include this kind of implementation detail, especially since slot sizes may not exist when we use modular GC with an alternative implementation.

st0012:
@peterzhu2118 So does it make sense to just drop everything after `This method is only expected to work with CRuby.`?

peterzhu2118:
No, I think there's still value in making it clear that it returns the size of the slot + externally allocated memory.

PR #13988, opened 2025-07-23
Fix `RCLASS_WRITE_M_TBL_WORKAROUND` to avoid Ruby crash

st0012:
On the current master, running `test_refinement.rb` with GC stress would cause Ruby to crash:

```
make test-all TESTS='../test/ruby/test_refinement.rb --gc-stress'
```

```
TestRefinement#test_refine_alias_in_subclass../internal/class.h:613:
Assertion Failed: RCLASS_WRITE_M_TBL_WORKAROUND:!check_promoted || !RB_OBJ_PROMOTED(klass)
```

This commit fixes the crash by applying write barrier remember to the klass, which is still a workaround, but avoids crashing Ruby with an assertion.

### DIFF
--- a/internal/class.h
+++ b/internal/class.h
@@ -8,6 +8,7 @@
  *             file COPYING are met.  Consult the file for details.
  * @brief      Internal header for Class.
  */
+#include "gc.h"
 #include "id.h"
 #include "id_table.h"           /* for struct rb_id_table */
 #include "internal/namespace.h" /* for rb_current_namespace */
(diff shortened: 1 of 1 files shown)


internal/class.h
```diff
@@ -604,16 +605,16 @@ RCLASS_SET_M_TBL_WORKAROUND(VALUE klass, struct rb_id_table *table, bool check_p
...
 }
 
-#define RCLASS_WRITE_M_TBL_EVEN_WHEN_PROMOTED(klass, table) RCLASS_WRITE_M_TBL_WORKAROUND(klass, table, false)
-#define RCLASS_WRITE_M_TBL(klass, table) RCLASS_WRITE_M_TBL_WORKAROUND(klass, table, true)
+#define RCLASS_WRITE_M_TBL_EVEN_WHEN_PROMOTED(klass, table) RCLASS_WRITE_M_TBL_WORKAROUND(klass, table)
+#define RCLASS_WRITE_M_TBL(klass, table) RCLASS_WRITE_M_TBL_WORKAROUND(klass, table)
 
 static inline void
-RCLASS_WRITE_M_TBL_WORKAROUND(VALUE klass, struct rb_id_table *table, bool check_promoted)
+RCLASS_WRITE_M_TBL_WORKAROUND(VALUE klass, struct rb_id_table *table)
```
st0012, line 612: I remove the argument as it's not used anymore, but I'm not sure if it's against any convention or we actually would want to keep it for documentation purpose...etc. Let me know if I should undo this change.

st0012:
This is wrong so I'm closing it now.

PR #13998, opened 2025-07-23
ZJIT: Support invalidating constant patch points

st0012:


vm_method.c
```diff
@@ -148,6 +148,7 @@ rb_clear_constant_cache_for_id(ID id)
     }
 
     rb_yjit_constant_state_changed(id);
+    rb_zjit_constant_state_changed(id);
```
tekknolagi, line 151: If this is only keyed on one ID, does that mean we will invalidate `A::B::C` if `D::E::C` changes (because they both share `C`)?

k0kubun:
YJIT also invalidates constants on `rb_yjit_constant_ic_update`. Does ZJIT need the same thing?

vm_method.c
st0012: Yes when `C` is redefined the current implementation invalidates all jumps associated with it (confirmed with debugger and print messages). But since the callback only receives one `ID`, I'm not sure if there's a way to reconstruct the full constant path based on it.

It looks like YJIT behaves the same way? 

https://github.com/Shopify/ruby/blob/8f634e19599c97f9ecae31afd651a65ca06693a1/yjit/src/invariants.rs#L337-L343

@k0kubun is this correct?

tekknolagi:
lgtm after addressing ^

vm_method.c
k0kubun: Yeah, it does seem like it. `rb_clear_constant_cache_for_id(ID id)` is called when the `id` is used in the scope of a module. So if you invalidate constants with an `id`, that would invalidate constants with the same name under different modules.

I do think we need to do that anyway though. If the ISEQ is on a module, any random class could include it and use the JIT code. We can't really scope the invalidation into specific classes/modules unless you scan every class/module that could use it, which seems unrealistic.

k0kubun:
So, for correctness, `rb_zjit_constant_state_changed` seems to catch everything. The reason why YJIT still needs the other hook `rb_yjit_constant_ic_update` today is that YJIT (and ZJIT) relies on the interpreter's inline cache to know the content, so it wants the interpreter to notify YJIT when the interpreter updates the inline cache. 

Since ZJIT currently only discards invalid code and never recompiles them, we probably don't need the hook yet. Let's merge this as is for now. Thanks!

vm_method.c
st0012: Thanks for the clarification!

PR #14015, opened 2025-07-25
Split autogenerated dependency to depend file from common.mk

hsbt:
`common.mk` is over the 20000+ lines. GitHub couldn't render that file in Web interface. I tried to split that from `common.mk`.


hsbt:
@nobu How about this?

PR #14016, opened 2025-07-25
Avoid GC while operands in inconsistent state

jhawthorn:
compile_data_calloc2 may run GC (though because it allocates from an arena this is rare in practice). When this happened when resizing operands there was a risk of seeing the insn in an inconsistent state.

To solve this we need to make any allocations before we start modifying the instrucitons. This refactors the code to use a new insn_replace_with_operands() function that allocates the new operands array before modifying the instruction object.


tenderlove (approved):
Looks great! Nice refactor

PR #14017, opened 2025-07-25
ZJIT: Inline guard type checks for some built-in types

st0012:
This implements similar fast-path guard type checks as YJIT.


zjit/src/codegen.rs
```diff
@@ -1047,9 +1047,27 @@ fn gen_test(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
 /// Compile a type check with a side exit
 fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard_type: Type, state: &FrameState) -> Option<lir::Opnd> {
     if guard_type.is_subtype(Fixnum) {
-        // Check if opnd is Fixnum
```
tekknolagi, line 1050: Is there a particular reason you deleted this comment? I think we can have a similar one that explains the mask

zjit/src/codegen.rs
```diff
@@ -1047,9 +1047,27 @@ fn gen_test(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
...
+        // Flonum: (val & RUBY_FLONUM_MASK) == RUBY_FLONUM_FLAG
+        let masked = asm.and(val, Opnd::UImm(RUBY_FLONUM_MASK as u64));
+        asm.cmp(masked, Opnd::UImm(RUBY_FLONUM_FLAG as u64));
+        asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_subtype(StaticSymbol) {
+        // Static symbols have (val & 0xff) == RUBY_SYMBOL_FLAG
+        // Use 8-bit comparison like YJIT does
+        asm.cmp(val.with_num_bits(8).unwrap(), Opnd::UImm(RUBY_SYMBOL_FLAG as u64));
+        asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_subtype(NilClass) {
```
tekknolagi: Until https://github.com/ruby/ruby/pull/14029 lands, these should be Exact variants

zjit/src/codegen.rs
```diff
@@ -1047,9 +1047,27 @@ fn gen_test(asm: &mut Assembler, val: lir::Opnd) -> Option<lir::Opnd> {
...
+        asm.cmp(val, Qnil.into());
+        asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_subtype(TrueClass) {
+        asm.cmp(val, Qtrue.into());
+        asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_subtype(FalseClass) {
```
tekknolagi: Can we add an assert that `Qfalse == 0` here?

zjit/src/codegen.rs
```diff
@@ -11,7 +11,7 @@ use crate::{asm::CodeBlock, cruby::*, options::debug, virtualmem::CodePtr};
 use crate::backend::lir::{self, asm_comment, asm_ccall, Assembler, Opnd, SideExitContext, Target, CFP, C_ARG_OPNDS, C_RET_OPND, EC, NATIVE_STACK_PTR, SP};
 use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, CallInfo, Invariant, RangeType, SideExitReason, SideExitReason::*, SpecialObjectType, SELF_PARAM_IDX};
 use crate::hir::{Const, FrameState, Function, Insn, InsnId};
-use crate::hir_type::{types::Fixnum, Type};
+use crate::hir_type::{types::{Fixnum, Flonum, StaticSymbol, NilClass, TrueClass, FalseClass}, Type};
```
tekknolagi: let's just import `types` and use `types::Fixnum`, etc

zjit/src/codegen.rs
st0012: I felt it's explaining the intention of the branch which is pretty obvious IMO. I think here we only need to comment the implementation nuances, which to me are the masking and the 8-bit comparison for symbols.

k0kubun (approved):

PR #14027, opened 2025-07-28
Port Oniguruma patches to prevent integer overflow

makenowjust:
I ported Oniguruma patches referred to kkos/oniguruma#164, which are to prevent integer overflow related to `reg->dmax`, `reg->dmin`, etc.


regexec.c
```diff
@@ -4960,7 +4960,7 @@ forward_search_range(regex_t* reg, const UChar* str, const UChar* end, UChar* s,
 
   p = s;
   if (reg->dmin != 0) {
-    if (end - p <= reg->dmin) return 0; /* fail */
+    if ((ptrdiff_t)(end - p) <= (ptrdiff_t)reg->dmin) return 0; /* fail */
```
nobu: Isn’t `end - p` `ptrdiff_t`?

PR #14028, opened 2025-07-28
Make static symbol ID atomic

peterzhu2118:
We don't need the VM lock if we make static symbol IDs atomic.

### DIFF
--- a/symbol.c
+++ b/symbol.c
@@ -93,7 +93,7 @@ enum id_entry_type {
 };
 
 typedef struct {
-    rb_id_serial_t last_id;
+    rb_atomic_t next_id;
     VALUE sym_set;
 
     VALUE ids;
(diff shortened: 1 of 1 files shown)


symbol.c
```diff
@@ -1061,7 +1036,7 @@ rb_make_temporary_id(size_t n)
 {
     const ID max_id = RB_ID_SERIAL_MAX & ~0xffff;
     const ID id = max_id - (ID)n;
-    if (id <= ruby_global_symbols.last_id) {
+    if (id < RUBY_ATOMIC_LOAD(ruby_global_symbols.next_id)) {
```
jhawthorn, line 1039: It feels like this could race with `next_id` being incremented (ie. this is loaded, then a new symbol is allocated). But with the existing code, even without threads, I think we'd have the same problem 🤷‍♂️.

peterzhu2118: Yes, this could race, but this is more of an edge case where we've allocated so many static symbols that we start colliding.

symbol.c
```diff
@@ -983,7 +958,7 @@ rb_sym2id(VALUE sym)
 
             if (UNLIKELY(!(id & ~ID_SCOPE_MASK))) {
                 VALUE fstr = RSYMBOL(sym)->fstr;
-                ID num = next_id_base_with_lock(symbols);
+                ID num = next_id_base();
```
nobu, line 961: Also `next_id_base` shouldn’t have the `symbols` argument?

peterzhu2118: We don't need it anymore because we can just use the `ruby_global_symbols` global variable since `next_id_base` does not require the VM lock anymore.

PR #14029, opened 2025-07-28
ZJIT: Remove some builtin subclasses from lattice

tekknolagi:
While Integer can technically be subclassed, instances of subclasses
cannot be created. Remove it from the type lattice.


tekknolagi:
This only makes sense if there's not some scary corner of the C API that can create instances of integer subclasses. In that case, though, I would just call rename Integer to IntegerExact and if we encounter an instance of a subclass, oh well, it's BasicObject

st0012 (approved):
Should we adopt the `final_type` helper in other similar types (e.g. TrueClass), either in a separate PR or commit?

tekknolagi:
Oh! I didn't realize it was true for them as well. Same commit & PR makes sense

st0012 (approved):

k0kubun (approved):

PR #14030, opened 2025-07-28
ZJIT: Use type profile if class is known

tekknolagi:


tekknolagi:
With #14042 this passes

PR #6187, opened 2022-07-27
New constant caching instruction: opt_getconstant_path 

jhawthorn:
_I previously shared an early experimental version of this in #6049, the main change from that PR is finishing the implementation, {Y,M}JIT support, and moving the `ID` segment list into the `IC` structure._

Previously YARV bytecode implemented constant caching by having a pair of instructions, `opt_getinlinecache` and `opt_setinlinecache`, wrapping a series of `getconstant` calls (with `putobject` providing supporting arguments).

```
# old
$ ruby --dump=insns -e 'Foo::Bar::Baz'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,13)> (catch: FALSE)
0000 opt_getinlinecache                     17, <is:0>                (   1)[Li]
0003 putobject                              true
0005 getconstant                            :Foo
0007 putobject                              false
0009 getconstant                            :Bar
0011 putobject                              false
0013 getconstant                            :Baz
0015 opt_setinlinecache                     <is:0>
0017 leave
```

This commit replaces that pattern with a new instruction, `opt_getconstant_path`, handling both getting/setting the inline cache and fetching the constant on a cache miss.

This is implemented by storing the full constant path as a null-terminated array of IDs inside of the IC structure. idNULL is used to signal an absolute constant reference.

```
# new
$ ./miniruby --dump=insns -e '::Foo::Bar::Baz'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,13)> (catch: FALSE)
0000 opt_getconstant_path                   <ic:0 ::Foo::Bar::Baz>      (   1)[Li]
0002 leave
```

The motivation for this is that we had increasingly found the need to disassemble the instructions between the `opt_getinlinecache` and `opt_setinlinecache` in order to determine the constant we are fetching, or otherwise store metadata.

This disassembly was previously done:
* In `opt_setinlinecache`, to register the `IC` against the constant names it is using for granular invalidation.
* In `rb_iseq_free`, to unregister the IC from the invalidation table.
* In YJIT to find the position of a `opt_getinlinecache` instruction to invalidate it when the cache is populated
* In YJIT to register the constant names being used for invalidation.

With this change we no longer need disassemly for these (in fact `rb_iseq_each` is now unused), as the list of constant names being referenced is held in the `IC`. This should also make it possible to make more optimizations in the future.

This may also reduce the size of iseqs, as previously each segment required 32 bytes (on 64-bit platforms) for each constant segment. This implementation only stores one `ID` per-segment.

[...]

PR #14033, opened 2025-07-28
YJIT: Remove a dead test for getinlinecaches

st0012:
The test was added in #5221 4 years ago but:

1. The insn it targets was removed in 2022 in #6187
2. The YJIT API `blocks_for` seems to be dropped in 2022 when it switched to use Rust in #5826

So this test has not been run in more than 3 years and can't be run anymore. I think we can remove it.


st0012:
@XrXr I think this is ready for merge now

PR #14034, opened 2025-07-28
ZJIT: Add --zjit-stats

k0kubun:
This PR introduces `--zjit-stats` and ports `ratio_in_yjit` to ZJIT.

When benchmarks are slow because you can't compile some code, compatibility changes that increase `ratio_in_yjit` the most have given the most significant impact on YJIT's performance. So let's make it available it for ZJIT. We should also count exit reasons to make it actionable, but it's out of scope in this PR, which focuses on establishing the infrastructure.

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -35,6 +35,8 @@
 #include "iseq.h"
 #include "symbol.h" // This includes a macro for a more performant rb_id2sym.
 #include "yjit.h"
+#include "insns.inc"
+#include "zjit.h"
 #include "ruby/st.h"
 #include "ruby/vm.h"
 #include "vm_core.h"
@@ -45,8 +47,6 @@
 #include "ractor_core.h"
 #include "vm_sync.h"
 #include "shape.h"
-#include "insns.inc"
-#include "zjit.h"
 
 #include "builtin.h"
 
--- a/yjit.c
+++ b/yjit.c
@@ -23,13 +23,13 @@
 #include "insns_info.inc"
 #include "vm_sync.h"
 #include "yjit.h"
+#include "zjit.h"
 #include "vm_insnhelper.h"
 #include "probes.h"
 #include "probes_helper.h"
 #include "iseq.h"
 #include "ruby/debug.h"
 #include "internal/cont.h"
-#include "zjit.h"
 
 // For mmapp(), sysconf()
 #ifndef _WIN32
--- a/zjit.c
+++ b/zjit.c
@@ -15,6 +15,7 @@
 #include "builtin.h"
 #include "insns.inc"
 #include "insns_info.inc"
+#include "zjit.h"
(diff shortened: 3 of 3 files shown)


zjit.rb
```diff
@@ -1,6 +1,54 @@
...
+    end
+
+    stats
+  end
+
+  # Get the summary of ZJIT statistics as a String
+  def stats_string
+    # Lazily require StringIO to avoid breaking miniruby
+    require 'stringio'
+    strio = StringIO.new
```
tekknolagi: This seems like the kind of thing that could just append to a string because it's not super perf sensitive. Then we wouldn't have to worry about circularity and requires and stuff

zjit/src/hir.rs
```diff
@@ -566,6 +566,9 @@ pub enum Insn {
 
     /// Side-exit into the interpreter.
     SideExit { state: InsnId, reason: SideExitReason },
+
+    /// Increment a counter in ZJIT stats
+    IncrCounter(Counter),
```
tekknolagi, line 571: Can you make this Display nicely?

tekknolagi:
I get

```
In file included from ../vm.c:44:
../vm_insnhelper.h:19:19: warning: 'ZJIT_STATS' is not defined, evaluates to 0 [-Wundef]
   19 | #if YJIT_STATS || ZJIT_STATS // We want vm_insns_count only on stats builds.
      |                   ^
1 warning generated.
```

is there something I have to do to fix this?

zjit.rb
k0kubun: seems fair. I like the fact that the terminal gets streaming updates because the YJIT one can be so big that it doesn't finish instantly, which also gives partial results on failure. But I guess I don't really read it until it's done, so it might not matter.

Let's try printing all at once https://github.com/ruby/ruby/pull/14034/commits/3d58c8a5ce71d7a5307850ac2283ccb1d9147b3e, and maybe revisit this when it gets super slow for some reason.

zjit/src/hir.rs
k0kubun: I thought `IncrCounter(zjit_insns_count)` was good enough but I guess you don't like parens. Fixed in https://github.com/ruby/ruby/pull/14034/commits/5015f69b5e9b977d03baa2c95d8d636754b8c932

k0kubun:
> is there something I have to do to fix this?

We shouldn't randomly introduce a new compile error to any environment, so it's not your fault.

Try rebuilding with https://github.com/ruby/ruby/pull/14034/commits/963eeb3e06e01a2811b6d0c9b345c52720bdfacd? One of the "Compilations" job (`RUBY_DEVEL` in `omnibus compilations, #7`) failed in the same way, so I'll use it as a test for it.

PR #14039, opened 2025-07-29
Allow for lock-free lookups in `RClass.cc_tbl`

casperisfine:
This PR is composed of three commits for easier review.

One of the biggest remaining contention point is `RClass.cc_table`. The logical solution is to turn it into a managed object, so we can use an RCU strategy, given it is expected to be read heavy.

However, that's not currently possible because the table can't be freed before the owning class, given the class free function MUST go over all the CC entries to invalidate them. By relying on `rb_gc_mark_weak`, we can break this dependency. However this means that `cc.klass` now defaults to `Qundef` instead of `Qfalse`.

The second commit turns `RClass.cc_tbl` into a managed object. It's a sub-type of `rb_managed_id_table`, and is in change of freeing the `rb_class_cc_entries`.

Then the third commit refactor `vm_lookup_cc` to be lock-free on the happy path. We still lock the VM when we need to evict an invalidated CC, and when we miss. When in single Ractor mode, we still directly mutate the `cc_tbl`, however in multi-ractor mode, we first deep-dup it, modify it, and then atomically swap it.

There are a number of small optimizations that could be done on the slowpath, but I chose to try to keep the PR as simple as possible to ease the review.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1208,7 +1208,6 @@ classext_free(rb_classext_t *ext, bool is_prime, VALUE namespace, void *arg)
     struct classext_foreach_args *args = (struct classext_foreach_args *)arg;
 
     rb_id_table_free(RCLASSEXT_M_TBL(ext));
-    rb_cc_tbl_free(RCLASSEXT_CC_TBL(ext), args->klass);
 
     if (!RCLASSEXT_SHARED_CONST_TBL(ext) && (tbl = RCLASSEXT_CONST_TBL(ext)) != NULL) {
         rb_free_const_table(tbl);
@@ -1239,7 +1238,6 @@ classext_iclass_free(rb_classext_t *ext, bool is_prime, VALUE namespace, void *a
     if (RCLASSEXT_CALLABLE_M_TBL(ext) != NULL) {
         rb_id_table_free(RCLASSEXT_CALLABLE_M_TBL(ext));
     }
-    rb_cc_tbl_free(RCLASSEXT_CC_TBL(ext), args->klass);
 
     rb_class_classext_free_subclasses(ext, args->klass);
 
@@ -2263,24 +2261,6 @@ rb_gc_after_updating_jit_code(void)
 #endif
 }
 
-static enum rb_id_table_iterator_result
-cc_table_memsize_i(VALUE ccs_ptr, void *data_ptr)
-{
-    size_t *total_size = data_ptr;
-    struct rb_class_cc_entries *ccs = (struct rb_class_cc_entries *)ccs_ptr;
-    *total_size += sizeof(*ccs);
-    *total_size += sizeof(ccs->entries[0]) * ccs->capa;
-    return ID_TABLE_CONTINUE;
-}
-
-static size_t
-cc_table_memsize(struct rb_id_table *cc_table)
-{
-    size_t total = rb_id_table_memsize(cc_table);
-    rb_id_table_foreach_values(cc_table, cc_table_memsize_i, &total);
-    return total;
-}
-
 static void
 classext_memsize(rb_classext_t *ext, bool prime, VALUE namespace, void *arg)
 {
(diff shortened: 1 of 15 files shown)


vm_method.c
```diff
@@ -22,6 +22,162 @@ static inline rb_method_entry_t *lookup_method_table(VALUE klass, ID id);
...
+    rb_managed_id_table_foreach(old_table, vm_cc_table_dup_i, (void *)new_table);
+    return new_table;
+}
+
+void
+rb_vm_cc_table_delete(VALUE table, ID mid)
+{
+    struct rb_class_cc_entries *ccs;
+    if (rb_managed_id_table_lookup(table, mid, (VALUE *)&ccs)) {
+        // FIXME: What are the invalidation considerations? e.g. invalidate_method_cache_in_cc_table?
```
casperisfine: This is likely the reason for the crash in YJIT. I need to figure out when exactly to trigger the invalidation.

vm_method.c
```diff
@@ -22,6 +22,184 @@ static inline rb_method_entry_t *lookup_method_table(VALUE klass, ID id);
...
+vm_cc_table_dup_i(ID key, VALUE old_ccs_ptr, void *data)
+{
+    struct rb_class_cc_entries *old_ccs = (struct rb_class_cc_entries *)old_ccs_ptr;
+    struct rb_class_cc_entries *new_ccs = ALLOC(struct rb_class_cc_entries);
+    MEMCPY(new_ccs, old_ccs, struct rb_class_cc_entries, 1);
+#if VM_CHECK_MODE > 0
+    new_ccs->debug_sig = ~(VALUE)new_ccs;
+#endif
+    new_ccs->entries = ALLOC_N(struct rb_class_cc_entries_entry, new_ccs->capa);
+    MEMCPY(new_ccs->entries, old_ccs->entries, struct rb_class_cc_entries_entry, new_ccs->capa);
```
casperisfine, line 155: In a followup, I'd like to inline `ccs->entries` inside `ccs`, should reduce the malloc churn and remove one pointer chase.

byroot:
Also cc @etiennebarrie (can't add you as reviewer, sorry)

gc.c
```diff
@@ -1743,12 +1742,15 @@ rb_objspace_free_objects(void *objspace)
...
-    return rb_gc_impl_garbage_object_p(rb_gc_get_objspace(), obj);
+    return !SPECIAL_CONST_P(obj) && rb_gc_impl_garbage_object_p(rb_gc_get_objspace(), obj);
 }
 
 bool
 rb_gc_pointer_to_heap_p(VALUE obj)
 {
+    if (SPECIAL_CONST_P(obj)) {
+        return false;
+    }
```
peterzhu2118: Hmmm, can we not make this change? I don't think it's technically correct to return true or false here for special consts. Since most places use this to check whether a pointer is alive or not, we should not return false for special consts. But since special consts are not GC managed, we cannot return true here either.

I think it's best to assert that `obj` is not a special const.

imemo.c
```diff
@@ -353,25 +353,20 @@ rb_imemo_mark_and_move(VALUE obj, bool reference_updating)
          */
         struct rb_callcache *cc = (struct rb_callcache *)obj;
         if (reference_updating) {
-            if (!cc->klass) {
-                // already invalidated
+            if (moved_or_living_object_strictly_p(cc->klass) &&
```
peterzhu2118: Since `klass` is marked weak, we should never run into the case where `moved_or_living_object_strictly_p(cc->klass) == false` right?

peterzhu2118 (commented):
Some minor comments, otherwise looks good!

gc.c
byroot: > Hmmm, can we not make this change? 

Yes, absolutely. It's a left over of me trying to workaround the mmtk bug.

I'll leave it to you to decide whether it should be made independently. It seemed fine to be because if something is immediate then it's not a pointer to the heap, that is certain. But I don't have that much context.

imemo.c
byroot: I think you are right.

PR #14041, opened 2025-07-29
ZJIT: Create delta debugging script to narrow JIT failures

tekknolagi:
Add support for `--zjit-allowed-iseqs=SomeFile` and
`--zjit-save-compiled-iseqs=SomeFile` so we can restrict and inspect
which ISEQs get compiled.

Then add `jit_bisect.rb` which we can run to try and narrow a failing
script. For example:

    plum% ./jit_bisect.rb ../build-dev/miniruby "test.rb"
    I, [2025-07-29T12:41:18.657177 #96899]  INFO -- : Starting with JIT list of 4 items.
    I, [2025-07-29T12:41:18.657229 #96899]  INFO -- : Verifying items
    I, [2025-07-29T12:41:18.726213 #96899]  INFO -- : step fixed[0] and items[4]
    I, [2025-07-29T12:41:18.726246 #96899]  INFO -- : 4 candidates
    I, [2025-07-29T12:41:18.797212 #96899]  INFO -- : 2 candidates
    Reduced JIT list:
    bar@test.rb:8
    plum%

We start with 4 compiled functions and shrink to just one.

### DIFF
--- a/tool/zjit_bisect.rb
+++ b/tool/zjit_bisect.rb
@@ -1,0 +1,97 @@
+#!/usr/bin/env ruby
+require 'logger'
+require 'open3'
+require 'tempfile'
+require 'timeout'
+
+RUBY = ARGV[0] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
+OPTIONS = ARGV[1] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
+TIMEOUT_SEC = 5
+LOGGER = Logger.new($stdout)
+
+# From https://github.com/tekknolagi/omegastar
+# MIT License
+# Copyright (c) 2024 Maxwell Bernstein and Meta Platforms
+# Attempt to reduce the `items` argument as much as possible, returning the
+# shorter version. `fixed` will always be used as part of the items when
+# running `command`.
+# `command` should return True if the command succeeded (the failure did not
+# reproduce) and False if the command failed (the failure reproduced).
+def bisect_impl(command, fixed, items, indent="")
+  LOGGER.info("#{indent}step fixed[#{fixed.length}] and items[#{items.length}]")
+  while items.length > 1
+    LOGGER.info("#{indent}#{fixed.length + items.length} candidates")
+    # Return two halves of the given list. For odd-length lists, the second
+    # half will be larger.
+    half = items.length / 2
+    left = items[0...half]
+    right = items[half..]
+    if !command.call(fixed + left)
+      items = left
+      next
+    end
+    if !command.call(fixed + right)
+      items = right
+      next
+    end
+    # We need something from both halves to trigger the failure. Try
+    # holding each half fixed and bisecting the other half to reduce the
+    # candidates.
+    new_right = bisect_impl(command, fixed + left, right, indent + "< ")
+    new_left = bisect_impl(command, fixed + new_right, left, indent + "> ")
+    return new_left + new_right
+  end
+  items
+end
+
+# From https://github.com/tekknolagi/omegastar
+# MIT License
+# Copyright (c) 2024 Maxwell Bernstein and Meta Platforms
+def run_bisect(command, items)
+  LOGGER.info("Verifying items")
+  if command.call(items)
+    raise StandardError.new("Command succeeded with full items")
+  end
+  if !command.call([])
+    raise StandardError.new("Command failed with empty items")
+  end
+  bisect_impl(command, [], items)
+end
+
+def run_with_jit_list(ruby, options, jit_list)
+  # Make a new temporary file containing the JIT list
+  Tempfile.create("jit_list") do |temp_file|
+    temp_file.write(jit_list.join("\n"))
+    temp_file.flush
+    temp_file.close
+    # Run the JIT with the temporary file
+    Open3.capture3("#{ruby} --zjit-allowed-iseqs=#{temp_file.path} #{options}")
+  end
+end
+
+# Try running with no JIT list to get a stable baseline
+_, stderr, status = run_with_jit_list(RUBY, OPTIONS, [])
+if !status.success?
+  raise "Command failed with empty JIT list: #{stderr}"
+end
+# Collect the JIT list from the failing Ruby process
+jit_list = nil
+Tempfile.create "jit_list" do |temp_file|
+  Open3.capture3("#{RUBY} --zjit-log-compiled-iseqs=#{temp_file.path} #{OPTIONS}")
+  jit_list = File.readlines(temp_file.path).map(&:strip).reject(&:empty?)
+end
+LOGGER.info("Starting with JIT list of #{jit_list.length} items.")
+# Now narrow it down
+command = lambda do |items|
+  status = Timeout.timeout(TIMEOUT_SEC) do
+    _, _, status = run_with_jit_list(RUBY, OPTIONS, items)
+    status
+  end
+  status.success?
+end
+result = run_bisect(command, jit_list)
+File.open("jitlist.txt", "w") do |file|
+  file.puts(result)
+end
+puts "Reduced JIT list (available in jitlist.txt):"
+puts result
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -284,6 +284,10 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Optio
             let iseq_name = iseq_get_location(iseq, 0);
             register_with_perf(iseq_name, start_usize, code_size);
         }
+        if ZJITState::should_log_compiled_iseqs() {
+            let iseq_name = iseq_get_location(iseq, 0);
+            ZJITState::log_compile(iseq_name);
+        }
     }
     result
 }
--- a/zjit/src/options.rs
+++ b/zjit/src/options.rs
@@ -1,6 +1,7 @@
 use std::{ffi::{CStr, CString}, ptr::null};
 use std::os::raw::{c_char, c_int, c_uint};
 use crate::cruby::*;
+use std::collections::HashSet;
 
 /// Number of calls to start profiling YARV instructions.
 /// They are profiled `rb_zjit_call_threshold - rb_zjit_profile_threshold` times,
@@ -19,7 +20,7 @@ pub static mut rb_zjit_call_threshold: u64 = 2;
 #[allow(non_upper_case_globals)]
 static mut zjit_stats_enabled_p: bool = false;
 
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub struct Options {
     /// Number of times YARV instructions should be profiled.
     pub num_profiles: u8,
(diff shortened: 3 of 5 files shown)


zjit/src/options.rs
```diff
@@ -36,6 +37,12 @@ pub struct Options {
 
     /// Dump code map to /tmp for performance profilers.
     pub perf: bool,
+
+    /// List of ISEQs that can be compiled, identified by their iseq_get_location()
+    pub allowed_iseqs: Option<HashSet<String>>,
+
+    /// Path to a file where compiled ISEQs will be saved on shutdown.
```
tekknolagi:
```suggestion
    /// Path to a file where compiled ISEQs will be saved.
```

zjit/jit_bisect.rb
k0kubun, line 1: This script should probably be placed under `misc/` (or `tool/` otherwise) because it's not part of the ZJIT implementation.

zjit/src/state.rs
```diff
@@ -148,7 +176,6 @@ pub extern "C" fn rb_zjit_init(options: *const u8) {
 
         let options = unsafe { Box::from_raw(options as *mut Options) };
         ZJITState::init(*options);
-        std::mem::drop(options);
```
k0kubun, line 151: This should be kept. We heap-allocated `options` and then assigned it to the permanent storage, so this shouldn't be floating around on the heap forever.

zjit/src/codegen.rs
```diff
@@ -101,6 +101,9 @@ pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, _ec: EcPtr) -> *co
 /// See [gen_iseq_entry_point_body]. This wrapper is to make sure cb.mark_all_executable()
 /// is called even if gen_iseq_entry_point_body() partially fails and returns a null pointer.
 fn gen_iseq_entry_point(iseq: IseqPtr) -> *const u8 {
+    if !ZJITState::can_compile_iseq(iseq_get_location(iseq, 0)) {
```
k0kubun: Can you avoid allocating a String here when `--zjit-allowed-iseqs` is not in use? 

In case you missed it, `asm_comment!(asm, "ZJIT entry point: {}", iseq_get_location(iseq, 0));` doesn't allocate any string when `--zjit-dump-disasm` is not enabled.

zjit/src/state.rs
```diff
@@ -136,6 +136,34 @@ impl ZJITState {
     pub fn get_counters() -> &'static mut Counters {
         &mut ZJITState::get_instance().counters
     }
+
+    pub fn should_save_compiled_iseqs() -> bool {
+        ZJITState::get_instance().options.save_compiled_iseqs.is_some()
+    }
+
+    pub fn log_compile(iseq_name: String) {
```
k0kubun, line 146: Could you add a `///` comment for this function? This is a fairly complicated logic, so it's not immediately clear what it's doing from the function signature.

zjit/src/state.rs
tekknolagi: I think it gets moved since I turned off Copy. If I keep this, it complains about use-after-move or something like that. I can get the exact error

zjit/jit_bisect.rb
```diff
@@ -0,0 +1,95 @@
...
+  if !command.call([])
+    raise StandardError.new("Command failed with empty items")
+  end
+  bisect_impl(command, [], items)
+end
+
+def run_with_jit_list(ruby, options, jit_list)
+  # Make a new temporary file containing the JIT list
+  temp_file = Tempfile.new("jit_list")
+  temp_file.write(jit_list.join("\n"))
```
XrXr: Some untested nit picks about this since we had some discussions about FS tempfile race:

```suggestion
  # Make a new temporary file containing the JIT list
  Tempfile.create("jit_list") do |tempfile|
      temp_file.puts(jit_list)
      temp_file.flush
      temp_file.close
      # Run the JIT with the temporary file
  Open3.capture3("#{ruby} --zjit-allowed-iseqs=#{temp_file.path} #{options}")
  end
```

Or maximally covert and race-free:

```suggestion
  # Make a new temporary file containing the JIT list
  Tempfile.create("jit_list", anonymous: true) do |tempfile|
      File.unlink(tempfile)
      temp_file.puts(jit_list)
      temp_file.flush
      fd = temp_file.fileno
      # Run the JIT with the temporary file
      Open3.capture3("#{ruby} --zjit-allowed-iseqs=/dev/fd/#{fd} #{options}", { fd => fd })
  end
```

But that leaves no file with a pathname you can easily inspect for debugging, a slight downside.


- IO#puts takes an array and puts each on a line
- `Tempfile.create` is recommended over `.new` for reasons listed in the [docs](https://docs.ruby-lang.org/en/master/Tempfile.html#method-c-new).

zjit/src/state.rs
tekknolagi:
```
error[E0382]: use of moved value: `options`
   --> zjit/src/state.rs:183:24
    |
182 |         ZJITState::init(*options);
    |                         -------- value moved here
183 |         std::mem::drop(options);
    |                        ^^^^^^^ value used here after move
    |
    = note: move occurs because `*options` has type `Options`, which does not implement the `Copy` trait
```

zjit/src/options.rs
```diff
@@ -165,6 +194,19 @@ fn parse_option(options: &mut Options, str_ptr: *const std::os::raw::c_char) ->
 
         ("perf", "") => options.perf = true,
 
+        ("allowed-iseqs", _) if opt_val != "" => options.allowed_iseqs = Some(parse_jit_list(opt_val)),
+        ("save-compiled-iseqs", _) if opt_val != "" => {
```
k0kubun: WDYT about calling it `--zjit-log-compiled-iseqs`? This is a compilation log. "save-compiled-iseqs" could sound like a compilation cache, so that'd be confusing when we actually introduce it.

Note that YJIT calls it `--yjit-log`, so it might make sense to call it `--zjit-log` too, which should probably support printing to stderr as well in that case.

zjit/src/state.rs
XrXr: No it's fine to remove. Box is owning, so it gets naturally dropped (or in this case, moved into `ZJITState::init()` and immortalized into a global). With `Copy` it was duping the whole thing then dropping it...

zjit/src/options.rs
tekknolagi: Sure, can can support printing to stderr with `/dev/stderr`

zjit/jit_bisect.rb
tekknolagi: I will switch to create

zjit/src/state.rs
k0kubun: ~I mean, sure, but not liking `Copy` trait is not a reason for introducing a memory leak. It just seems like you need to `clone()`. It's on the heap, so it's natural that you need to "copy" it to the global variable.~

k0kubun: OK I guess Alan is right. Sorry.

XrXr: There is no leaks here or without the `mem::drop` `mem::drop` is an empty function that doesn't do anything extra than if options here went out of scope naturally.

zjit/src/options.rs
```diff
@@ -57,6 +64,8 @@ pub fn init_options() -> Options {
         dump_lir: false,
         dump_disasm: false,
         perf: false,
+        allowed_iseqs: None,
+        log_compiled_iseqs: None,
     }
 }
```
k0kubun, line 71: We might want to add `--zjit-log-compiled-iseqs` to `ZJIT_OPTIONS` (`ruby --help`). The allowed-iseqs option should probably be secret, but compilation logs seem useful for end-users.

tool/zjit_bisect.rb
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env ruby
+require 'logger'
+require 'open3'
+require 'tempfile'
+require 'timeout'
+
+RUBY = ARGV[0] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
+OPTIONS = ARGV[1] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
```
k0kubun, line 8: Could you make it a local variable? It seems like you refer to this constant only in the top-level binding. It's generally more maintainable to restrict the scope of variables.

tool/zjit_bisect.rb
```diff
@@ -0,0 +1,97 @@
...
+# Collect the JIT list from the failing Ruby process
+jit_list = nil
+Tempfile.create "jit_list" do |temp_file|
+  Open3.capture3("#{RUBY} --zjit-log-compiled-iseqs=#{temp_file.path} #{OPTIONS}")
+  jit_list = File.readlines(temp_file.path).map(&:strip).reject(&:empty?)
+end
+LOGGER.info("Starting with JIT list of #{jit_list.length} items.")
+# Now narrow it down
+command = lambda do |items|
+  status = Timeout.timeout(5) do
```
k0kubun: You probably meant:

```suggestion
  status = Timeout.timeout(TIMEOUT_SEC) do
```

tool/zjit_bisect.rb
k0kubun: Well, I guess you might have wanted to make it seem like read-only (because it's supposedly a "constant"). It might make sense as is in that case.

k0kubun (approved):

zjit/src/options.rs
```diff
@@ -67,6 +76,8 @@ pub const ZJIT_OPTIONS: &'static [(&str, &str)] = &[
     ("--zjit-num-profiles=num",   "Number of profiled calls before JIT (default: 1, max: 255)."),
     ("--zjit-stats",              "Enable collecting ZJIT statistics."),
     ("--zjit-perf",               "Dump ISEQ symbols into /tmp/perf-{}.map for Linux perf."),
+    ("--zjit-log-compiled-iseqs=path",
+     "Log compiled ISEQs to the specified file. The file will be truncated on each run."),
```
k0kubun: This description is too long that it doesn't pass `TestRubyOptions#test_usage_long` in `test/ruby/test_rubyoptions.rb`.

You might want to align the indentation of this line to make the `80-char limit --> |` comment actually useful.

tekknolagi: oh it's not clear that it should be aligned with the name too. i will give it a go

zjit/src/options.rs
```diff
@@ -67,6 +76,8 @@ pub const ZJIT_OPTIONS: &'static [(&str, &str)] = &[
     ("--zjit-num-profiles=num",   "Number of profiled calls before JIT (default: 1, max: 255)."),
     ("--zjit-stats",              "Enable collecting ZJIT statistics."),
     ("--zjit-perf",               "Dump ISEQ symbols into /tmp/perf-{}.map for Linux perf."),
+    ("--zjit-log-compiled-iseqs=path", "Log compiled ISEQs to the specified file. The file will
+                                        be truncated on each run."),
```
k0kubun: If you try `./miniruby --help`, this renders into:

```
  --zjit-log-compiled-iseqs=path
                  Log compiled ISEQs to the specified file. The file will
                                                          be truncated on each run.
```

The indentation should not be part of the string content.

Because `--zjit-log-compiled-iseqs=path` is already long by itself, the help automatically moves the line to the next line, so let's start the description in the next line in the source code as well.

And it's still too long by the way (try `make test-all TESTS="test/ruby/test_rubyoptions.rb"`), so you need to make it as short as (Note that the position of the `80-char limit` comment was actually wrong. I can fix it separately):

```suggestion
    ("--zjit-log-compiled-iseqs=path",
                     "Log compiled ISEQs to the file. The file will be truncated."),
```

PR #14042, opened 2025-07-29
ZJIT: Catch more failed recursive compilations

tekknolagi:
Untangle the logic a bit and specifically:

* catch `gen_entry` failures
* don't set `start_ptr` until all recursive calls succeed

Co-authored-by: Alan Wu <alanwu@ruby-lang.org>

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -120,21 +120,18 @@ fn gen_iseq_entry_point_body(cb: &mut CodeBlock, iseq: IseqPtr) -> *const u8 {
     };
 
     // Compile the High-level IR
-    let (start_ptr, gc_offsets, mut branch_iseqs) = match gen_function(cb, iseq, &function) {
-        Some((start_ptr, gc_offsets, jit)) => {
-            // Compile an entry point to the JIT code
-            if let Some(ptr) = gen_entry(cb, iseq, &function, start_ptr) {
-                (ptr, gc_offsets, jit.branch_iseqs)
-            } else {
-                debug!("Failed to compile iseq: gen_entry failed: {}", iseq_get_location(iseq, 0));
-                return std::ptr::null();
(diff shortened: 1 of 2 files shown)


zjit/src/codegen.rs
```diff
@@ -120,39 +120,46 @@ fn gen_iseq_entry_point_body(cb: &mut CodeBlock, iseq: IseqPtr) -> *const u8 {
...
-        if let Some((callee_ptr, callee_branch_iseqs)) = gen_iseq(cb, iseq) {
-            let callee_addr = callee_ptr.raw_ptr(cb);
-            branch.regenerate(cb, |asm| {
-                asm.ccall(callee_addr, vec![]);
-            });
-            branch_iseqs.extend(callee_branch_iseqs);
-        } else {
+        let Some((callee_ptr, callee_branch_iseqs)) = gen_iseq(cb, iseq) else {
             // Failed to compile the callee. Bail out of compiling this graph of ISEQs.
+            debug!("Failed to compile iseq: could not compile callee: {} -> {}",
```
XrXr, line 145: a little bit of fun

```suggestion
            debug!("Failed to compile iseq: could not compile callee: {} ⇝ {}",
```

[Rightwards Squiggle Arrow](https://www.compart.com/en/unicode/U+21DD)

XrXr (approved):

zjit/src/codegen.rs
tekknolagi: If we can do some kind of terminal vs non-terminal split, I think special chars are fine for terminal but irritating for non-TTY output. I did something like this for YJIT backend but it somehow got lost in the port to ZJIT

tekknolagi:
ah. failures because the early start_ptr setting "ties the knot" for recursive functions. we need to keep that but have a rollback/undo button or something

PR #14044, opened 2025-07-29
ZJIT: Prepare for sharing JIT hooks with ZJIT

k0kubun:
This PR renames `with_yjit` to `with_jit` and prepares for sharing JIT hooks with ZJIT. It should help us compare the performance of YJIT and ZJIT in a fair setup, and speed up ZJIT on those methods in the future.

It also moves `Options` out of `ZJITState` because we refer to it during builtins, which is before `rb_zjit_init()` initializes `ZJITState`. Note that `zjit_stats_enabled_p` is removed, so the number of global variables stayed the same.

Unfortunately, ZJIT currently seems to have multiple bugs when it actually compiles `Array#each` in Ruby. So the feature is turned off inside `with_jit` for now. However, I still want to land this fix to make sure we have the change of `Options` and make it easier to debug `Array#each`.

### DIFF
--- a/jit_hook.rb
+++ b/jit_hook.rb
@@ -3,7 +3,7 @@ class Module
   # This method is removed in jit_undef.rb.
   private def with_jit(&block) # :nodoc:
     # ZJIT currently doesn't compile Array#each properly, so it's disabled for now.
-    if defined?(RubyVM::ZJIT) && Primitive.rb_zjit_option_enabled_p && false # TODO: remove `&& false`
+    if defined?(RubyVM::ZJIT) && Primitive.rb_zjit_option_enabled_p && false # TODO: remove `&& false` (Shopify/ruby#667)
       # We don't support lazily enabling ZJIT yet, so we can call the block right away.
       block.call
     elsif defined?(RubyVM::YJIT)


jit_hook.rb
```diff
@@ -0,0 +1,13 @@
+class Module
+  # Internal helper for built-in initializations to define methods only when JIT is enabled.
+  # This method is removed in jit_undef.rb.
+  private def with_jit(&block) # :nodoc:
+    # ZJIT currently doesn't compile Array#each properly, so it's disabled for now.
+    if defined?(RubyVM::ZJIT) && Primitive.rb_zjit_option_enabled_p && false # TODO: remove `&& false`
```
tekknolagi: Can you please attach this to an issue?

tekknolagi (approved):

jit_hook.rb
k0kubun: filed https://github.com/Shopify/ruby/issues/667 and added a comment linking to it https://github.com/ruby/ruby/pull/14044/commits/e980143bc744d68efff2f945bd9ada3d0de00255

PR #14045, opened 2025-07-29
ZJIT: Fix more recursive compile issues

tekknolagi:
- **Abort stub**
- **Don't set start_ptr or GC offsets until entire graph compiles**


tekknolagi:
Superseded by #14052

PR #14052, opened 2025-07-30
ZJIT: Stub JIT-to-JIT calls

k0kubun:
This PR introduces a stub for JIT-to-JIT calls that allows you to lazily compile the callees in a compiled ISEQ.

Previously, ZJIT was forced to compile every callee ISEQ recursively and give up compiling the first-seen ISEQ if any of it fails. In this PR, since we no longer compile the callee ISEQs at the same time when compiling an ISEQ, the caller ISEQ can be successfully compiled even when it fails to compile callee ISEQs.

We could make the stub code smaller by having a trampoline shared by all stubs, but it's out of scope in this PR. I want to make it work first without complicating the first step too much.

### DIFF
--- a/jit.c
+++ b/jit.c
@@ -442,3 +442,15 @@ rb_yarv_ary_entry_internal(VALUE ary, long offset)
 {
     return rb_ary_entry_internal(ary, offset);
 }
+
+void
+rb_set_cfp_pc(struct rb_control_frame_struct *cfp, const VALUE *pc)
+{
+    cfp->pc = pc;
+}
+
+void
+rb_set_cfp_sp(struct rb_control_frame_struct *cfp, VALUE *sp)
+{
+    cfp->sp = sp;
+}
--- a/yjit.c
+++ b/yjit.c
@@ -499,18 +499,6 @@ rb_yjit_str_simple_append(VALUE str1, VALUE str2)
     return rb_str_cat(str1, RSTRING_PTR(str2), RSTRING_LEN(str2));
 }
 
-void
-rb_set_cfp_pc(struct rb_control_frame_struct *cfp, const VALUE *pc)
-{
-    cfp->pc = pc;
-}
-
-void
-rb_set_cfp_sp(struct rb_control_frame_struct *cfp, VALUE *sp)
-{
-    cfp->sp = sp;
-}
-
 extern VALUE *rb_vm_base_ptr(struct rb_control_frame_struct *cfp);
 
 // YJIT needs this function to never allocate and never raise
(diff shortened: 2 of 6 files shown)


XrXr (approved):
Am I reading this right that hitting a stubbed function patches all of its callers? If so it's maybe more complex than what we should start with -- with YJIT this kind of setup brought a lot of complexity with invalidation and patching and related memory management situations. That kind of setup is pretty much necessary for YJIT, but for ZJIT we might be able to sacrifice a tiny bit of efficiency for simplicity.

I think a simpler setup where each iseq owns its stub(s) where we patch the stub, not the callers might be a better starting point. That way, patching is localized to each iseq. From chatting with @tekknolagi that seems to play better with the multi entry point setup he wants to go towards.

k0kubun:
> Am I reading this right that hitting a stubbed function patches all of its callers?

No? If you look at `function_stub_hit`, each stub receives a single callsite (`Rc<Branch>`), so only that caller is patched as it's being called. When other callsites of the same callee are called, it discovers the pointer from the payload, and update that caller, one at a time.

> I think a simpler setup where each iseq owns its stub(s) where we patch the stub, not the callers might be a better starting point. That way, patching is localized to each iseq.

By "each iseq", which ISEQ do you mean, the caller ISEQ or the callee ISEQ? If you want a caller ISEQ to own a stub, that's what this PR does. If you want a callee ISEQ to own a stub and share it among all its callers without patching them, the caller would need to first jump to some shared trampoline, and then the trampoline (which will be the single patching location when the ISEQ is compiled) would jump to the compiled code, even when it has already been compiled. The double jumps seem like an extra overhead compared to what this PR generates, so I'm not sure if that's what we want.

XrXr:
> No? If you look at function_stub_hit, each stub receives a single callsite (Rc<Branch>), so only that caller is patched as it's being called. When other callsites of the same callee are called, it discovers the pointer from the payload, and update that caller, one at a time.

Ah. It looks like one iseq can get multiple stubs if called from multiple places, so if you have `A -> B`, `C -> B`, there would be two stubs for the callsite in A and C. Wouldn't unhit stubs take up memory until they're hit? Also, tricky object graph issue where the stub might mark the callee's iseq and force the callee to stay alive longer than necessary are the type of  issues I was alluding to.

> If you want a callee ISEQ to own a stub and share it among all its callers without patching them, the caller would need to first jump to some shared trampoline... The double jumps seem like an extra overhead compared to what this PR generates, so I'm not sure if that's what we want.

Yes, that's the kind of setup I'm suggesting. It's one more jump than this, but simpler. Anyways, we're going with what's here right now and it's fine. What I'm proposing here is for future discussions.

k0kubun:
We don't reclaim code, so it takes up space forever. So this is a speed/space trade-off. 

I'm open to discussing codegen slower than YJIT, but I'd be more comfortable landing that after we achieve the YJIT parity first and have a nice benchmark environment to evaluate the speed we want to compromise. Reproducing the YJIT's performance with a different compiler architecture is hard by itself. This first problem would be "simpler" to solve if we didn't complicate it by adding extra overhead YJIT didn't have, such as extra jumps or extra memory loads (for object pool).

PR #14054, opened 2025-07-30
Use write barrier version for rb_proc_call_kw

jhawthorn:
RARRAY_CONST_PTR requires that the elements in the Array are never modified. However inside [invoke_block_from_c_proc the last hash arg (which I think was assumed to be on the stack) could be replaced by a new object](https://github.com/ruby/ruby/blob/7cece235ab9d3834b314f4395d45b8c97399470c/vm.c#L1756).

Because of this that either needs a write barrier (which is challenging because there is no reference to the args array at that point), or we need to mark the array as WB unprotected (which will slow GC).

This was found by wbcheck

@jeremyevans is there another way we could do that in [invoke_block_from_c_proc](https://github.com/ruby/ruby/blob/7cece235ab9d3834b314f4395d45b8c97399470c/vm.c#L1756) ?

### DIFF
--- a/proc.c
+++ b/proc.c
@@ -992,7 +992,12 @@ rb_proc_call_kw(VALUE self, VALUE args, int kw_splat)
     VALUE vret;
     rb_proc_t *proc;
     int argc = check_argc(RARRAY_LEN(args));
-    const VALUE *argv = RARRAY_CONST_PTR(args);
+
+    // rb_vm_invoke_proc may end up modifying argv as part of calling and so we
+    // must use RARRAY_PTR, which marks the array as WB_UNPROTECTED instead of
+    // RARRAY_CONST_PTR. Unfortunately this is worse for GC.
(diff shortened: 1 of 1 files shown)


nobu:
It feels surprising to me that the argument content may be modified silently.
May we need to copy and splat the last argument in earlier stages?

jeremyevans:
I assumed the pointer wouldn't change, that's why I used `RARRAY_CONST_PTR`. That turns out to be a bad assumption to the current implementation (not sure whether it was a bad assumption when the code was written), so using `RARRAY_PTR` seems to be an appropriate fix.  I don't have an idea of a more optimized approach, unfortunately.

jhawthorn:
Thank you both. I also found it very surprising.

Before merging I'm going to look at modifying `invoke_block_from_c_proc` to see how much that extra allocation+memcpy hurts performance (and it feels like we should be able to `alloca` there for small values).

PR #14055, opened 2025-07-30
Make cross_ractor_require write barrier protected

peterzhu2118:


jhawthorn (approved):
Thank you! Looks correct to me

PR #14059, opened 2025-07-31
ZJIT: Codegen for NewHash

corona10:
This is my first contribution to ZJIT. I started by handling `NewHash`, and also noticed that more complex cases like `HashDup` and `Send` should be handled as well. These might be optimizable into other opcodes.

I'm sure my mentor @tekknolagi will take a look :)


zjit/src/codegen.rs
```diff
@@ -913,6 +914,36 @@ fn gen_new_array(
...
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    elements: &Vec<(InsnId, InsnId)>,
+    state: &FrameState,
+) -> lir::Opnd {
+    // Save PC
+    gen_save_pc(asm, state);
+
+    asm_comment!(asm, "call rb_hash_new");
+    let cap: ::std::os::raw::c_long = (elements.len() / 2).try_into().expect("Unable to fit length of elements into c_long");
```
tekknolagi: Why div by 2? We have `elements.len()` elements

zjit/src/codegen.rs
```diff
@@ -913,6 +914,36 @@ fn gen_new_array(
...
+        rb_hash_new_with_size as *const u8,
+        vec![lir::Opnd::Imm(cap)],
+    );
+
+    for (key_id, val_id) in elements.iter() {
+        let key = jit.get_opnd(*key_id).unwrap();
+        let val = jit.get_opnd(*val_id).unwrap();
+        asm_comment!(asm, "call rb_hash_aset");
+        asm.ccall(
+            rb_hash_aset as *const u8,
```
tekknolagi: I think this can call arbitrary user code so we have to save more state than just the PC. I'm not 100% sure what, in this case. I'll get back to you

zjit/src/codegen.rs
```diff
@@ -913,6 +914,36 @@ fn gen_new_array(
...
+
+    asm_comment!(asm, "call rb_hash_new");
+    let cap: ::std::os::raw::c_long = (elements.len() / 2).try_into().expect("Unable to fit length of elements into c_long");
+    let new_hash = asm.ccall(
+        rb_hash_new_with_size as *const u8,
+        vec![lir::Opnd::Imm(cap)],
+    );
+
+    for (key_id, val_id) in elements.iter() {
+        let key = jit.get_opnd(*key_id).unwrap();
```
tekknolagi: Let's not use `unwrap` but instead `?` operator and return `Option<lir::Opnd>`

zjit/src/codegen.rs
```diff
@@ -913,6 +914,36 @@ fn gen_new_array(
...
+    asm: &mut Assembler,
+    elements: &Vec<(InsnId, InsnId)>,
+    state: &FrameState,
+) -> lir::Opnd {
+    // Save PC
+    gen_save_pc(asm, state);
+
+    asm_comment!(asm, "call rb_hash_new");
+    let cap: ::std::os::raw::c_long = (elements.len() / 2).try_into().expect("Unable to fit length of elements into c_long");
+    let new_hash = asm.ccall(
```
tekknolagi: use `asm_ccall!` macro -- it's new

test/ruby/test_zjit.rb
```diff
@@ -429,6 +429,43 @@ def test(a, b) = a >= b
...
+    }, insns: [:newhash]
+  end
+
+  def test_new_hash_with_computation
+    assert_compiles '{"sum" => 5, "product" => 6}', %q{
+      def test(a, b)
+        {"sum" => a + b, "product" => a * b}
+      end
+      test(2, 3)
+    }, insns: [:newhash]
```
tekknolagi, line 537: Can we add some tests that it calls the right user eq and hash functions?

tekknolagi (changes requested):
Thank you for the very fast PR! On the right track -- needs some more work.

tekknolagi:
Also I wonder if this failure is related:

```
  make[1]: *** [yes-test-all] Error 1
    1) Failure:
  make: *** [zjit-test-all] Error 2
  TestRubyOptimization#test_side_effect_in_popped_splat [/Users/runner/work/ruby/ruby/src/test/ruby/test_optimization.rb:984]:
  Expected "[ruby-core:84340] [Bug #14201]" to be nil.
```

corona10:
> Also I wonder if this failure is related:

Not sure.. I can not reproduce at my mahcine.

corona10:
yeah it's related.. I need to understand intended behavior.

zjit/src/codegen.rs
tekknolagi: I think we need to write all the locals and stack and SP to the frame, unfortunately. See for example the beginning of gen_send_without_block. I would probably copy that code

tekknolagi: Oops, no, sorry, it's going to be a kind of combination of that and gen_send_without_block_direct, I think. Like

```rust
    // Save cfp->pc and cfp->sp for the caller frame
    gen_save_pc(asm, state);
    gen_save_sp(asm, state.stack().len());

    // Spill the virtual stack and the locals of the caller onto the stack
    // TODO: Lazily materialize caller frames on side exits or when needed
    asm_comment!(asm, "spill locals and stack");
    for (idx, &insn_id) in state.locals().enumerate() {
        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
    }
    for (idx, &insn_id) in state.stack().enumerate() {
        asm.mov(Opnd::mem(64, SP, idx as i32 * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
    }
```

tekknolagi: Medium term we should probably have a function like `prepare_for_reentrant_code()`

corona10:
@tekknolagi Thank for the review explain many things :) I've updated the PR!

zjit/src/codegen.rs
```diff
@@ -917,6 +917,41 @@ fn gen_new_array(
...
+
+    asm_comment!(asm, "call rb_hash_new");
+    let cap: ::std::os::raw::c_long = elements.len().try_into().expect("Unable to fit length of elements into c_long");
+    let new_hash = asm_ccall!(asm, rb_hash_new_with_size, lir::Opnd::Imm(cap));
+
+    for (key_id, val_id) in elements.iter() {
+        let key = jit.get_opnd(*key_id)?;
+        let val = jit.get_opnd(*val_id)?;
+        asm_comment!(asm, "call rb_hash_aset");
+        asm_ccall!(asm, rb_hash_aset, new_hash, key, val);
```
k0kubun: Not using the bulk-insert API could make ZJIT slower than the interpreter, e.g. rehash multiple times. Could you use the same function that the interpreter uses, `rb_hash_bulk_insert`? YJIT uses that too.

tekknolagi:
Except for @k0kubun 's feedback I think this looks great! Bulk insert is a good idea.

(Kokubun, do you think the frame setup bit is ok?)

k0kubun:
> Kokubun, do you think the frame setup bit is ok?

Yeah, it's a non-leaf instruction, so it makes sense that we set PC/SP and spill locals/stack slots (for now). Filed https://github.com/ruby/ruby/pull/14100 to define a helper for sharing the logic.

When there's no element to insert, it's probably leaf (while it does trigger GC). I think we can skip the SP update and spills in that case. We still have to set PC for allocation tracing.

zjit/src/codegen.rs
k0kubun: > Medium term we should probably have a function like prepare_for_reentrant_code()

Filed https://github.com/ruby/ruby/pull/14100

zjit/src/codegen.rs
```diff
@@ -917,6 +917,41 @@ fn gen_new_array(
...
+
+    // Spill the virtual stack and the locals of the caller onto the stack
+    // TODO: Lazily materialize caller frames on side exits or when needed
+    asm_comment!(asm, "spill locals and stack");
+    for (idx, &insn_id) in state.locals().enumerate() {
+        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
+    }
+    for (idx, &insn_id) in state.stack().enumerate() {
+        asm.mov(Opnd::mem(64, SP, idx as i32 * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
+    }
```
k0kubun: I've merged https://github.com/ruby/ruby/pull/14100. Could you rebase your branch and use `gen_prepare_non_leaf_call()` here?

zjit/src/codegen.rs
corona10: @k0kubun Out of curiosity, how would we handle it if a user-defined key throws an exception while inserting multiple keys at once?

cc @tekknolagi

k0kubun: When something raises an exception in the middle of a C call, it would longjmp to a frame that did setjmp. At that point, the frames of this JIT code and the C function inserting keys will be expired, so there's nothing to handle in the JIT code. Thanks to `gen_prepare_non_leaf_call`, the exception handler could use the stack spilled by the JIT code to take over the execution on the interpreter when there's a catch table for the ISEQ.

So whether you bulk-insert keys or not doesn't really change how the exception is handled. When it raises, you don't come back from `rb_hash_aset` to the JIT code either.

tekknolagi:
There are some test failures that are not immediately clear what's going on. I can try and take a look tomorrow or the day after. Frustrating :/ Probably something small

zjit/src/codegen.rs
```diff
@@ -907,6 +907,43 @@ fn gen_new_array(
...
+        }
+
+        for (i, pair) in pairs.iter().enumerate() {
+            let stack_opnd = Opnd::mem(64, SP, i as i32 * SIZEOF_VALUE_I32);
+            asm.mov(stack_opnd, *pair);
+        }
+
+        let argv = asm.lea(Opnd::mem(64, SP, 0));
+
+        asm_comment!(asm, "call rb_hash_bulk_insert");
```
```suggestion
```

`asm_ccall!` does this

test/ruby/test_zjit.rb
```diff
@@ -429,6 +429,117 @@ def test(a, b) = a >= b
     }, insns: [:opt_ge], call_threshold: 2
   end
 
+
```
k0kubun:
```suggestion
```

zjit/src/codegen.rs
```diff
@@ -907,6 +907,43 @@ fn gen_new_array(
...
+        let mut pairs = Vec::new();
+        for (key_id, val_id) in elements.iter() {
+            let key = jit.get_opnd(*key_id)?;
+            let val = jit.get_opnd(*val_id)?;
+            pairs.push(key);
+            pairs.push(val);
+        }
+
+        for (i, pair) in pairs.iter().enumerate() {
+            let stack_opnd = Opnd::mem(64, SP, i as i32 * SIZEOF_VALUE_I32);
```
k0kubun: This seems unsafe. `rb_hash_bulk_insert` calls hash functions, which may not be leaf (hence you're calling `gen_prepare_non_leaf_call`).

Your code here is writing above the VM stack stop, so if anything is pushed to the VM stack, the content of the array in the argument gets clobbered. Because it's not leaf, it could call another method, and at that point things will be pushed to the VM stack.

But you can't bump the `cfp->sp` to protect them because it would miss the point of `gen_prepare_non_leaf_call` (part of the dumped stack should be a valid array usable for `rb_hash_bulk_insert` currently, but we'd want to optimize it away later anyway), which makes it safe for the interpreter to take over the stack as is.

So I think we need to use the C stack here. Bump NATIVE_STACK_PTR, use memory operands based off of `NATIVE_BASE_PTR` (because `NATIVE_STACK_PTR` can be moved by the backend), and restore NATIVE_STACK_PTR.

k0kubun: Well, I remembered we never update the `SP` register, so I guess I misread it a bit. But then it writes to random locations in the VM stack that already have spilled stack slots, so it can still corrupt the state for the interpreter in the wrong way. As suggested above, we should write to the C stack.

tekknolagi: Happy to pair on this this week or next @corona10 ! Sorry it's been a bit of a rollercoaster

tekknolagi:
Another thing we can do to chunk the work is to add a new PR that just handles `NewHash` with no operands. That should be a) much easier to compile and b) still handle a huge chunk of the "failed to compile" cases which are just allocating empty hashes.

corona10:
> Another thing we can do to chunk the work is to add a new PR that just handles NewHash with no operands. That should be a) much easier to compile and b) still handle a huge chunk of the "failed to compile" cases which are just allocating empty hashes.

Yeah, let's start with simple stuff first. (I was also pretty busy with other stuff, but happy to start pair with you too :))

tekknolagi:
I wonder if you are running into a manifestation of https://bugs.ruby-lang.org/issues/14201 that https://github.com/ruby/ruby/pull/14205 fixes

corona10:
> I wonder if you are running into a manifestation of https://bugs.ruby-lang.org/issues/14201 that https://github.com/ruby/ruby/pull/14205 fixes

Maybe it is worth rebasing PR once you merge it.

tekknolagi:
Feel free to also cherry-pick

zjit/src/codegen.rs
```diff
@@ -930,6 +930,62 @@ fn gen_new_array(
...
+/// Compile a new hash instruction
+fn gen_new_hash(
+    jit: &mut JITState,
+    asm: &mut Assembler,
+    elements: &Vec<(InsnId, InsnId)>,
+    state: &FrameState,
+) -> Option<lir::Opnd> {
+    gen_prepare_non_leaf_call(jit, asm, state)?;
+
+    asm_comment!(asm, "call rb_hash_new");
```
k0kubun: `asm_ccall!` is `asm_comment!` + `asm.ccall`, so you shouldn't need this.

```suggestion
```

zjit/src/codegen.rs
```diff
@@ -930,6 +930,62 @@ fn gen_new_array(
...
+    let cap: ::std::os::raw::c_long = elements.len().try_into().expect("Unable to fit length of elements into c_long");
```
k0kubun: `::std::os::raw::c_long` seems a bit too long in this line. Since we import `c_int` and `c_void` in this file, we can import `c_long` as well.

k0kubun:
Now that the PR is merged, let's try rebasing from the latest master then? We can't merge a PR with a failing ZJIT test, which is relatively stable these days.

tekknolagi:
This might be related to our failure to handle locals in the JIT with `eval`:

```ruby
  def test_side_effect_in_popped_splat
    bug = '[ruby-core:84340] [Bug #14201]'
    eval("{**(bug = nil; {})};42")
    assert_nil(bug)

    bug = '[ruby-core:85486] [Bug #14459]'
    h = {}
    assert_equal(bug, eval('{ok: 42, **h}; bug'))
    assert_equal(:ok, eval('{ok: bug = :ok, **h}; bug'))
    assert_empty(h)
  end
```

Notice how it tries to modify `bug`...

tekknolagi:
Going to skip for now but we can do EP==BP invalidation soon, I think?

tekknolagi:
```ruby
def test
  bug = 'BUG!!!!'
  eval("bug = 123; 456")
  bug
end

p test
p test
p test
p test
p test
```

PR #14063, opened 2025-07-31
Get rid of RSHAPE_PARENT in favor of RSHAPE_DIRECT_CHILD_P

casperisfine:
Fix: https://github.com/ruby/ruby/pull/14058

`RSHAPE_PARENT` is error prone because it returns a raw untagged shape_id.

To check if a shape is a direct parent of another, tags should be discarded. So providing a comparison function is better than exposing untagged ids.

### DIFF
--- a/shape.h
+++ b/shape.h
@@ -271,9 +271,16 @@ rb_shape_root(size_t heap_id)
 }
 
 static inline shape_id_t
-RSHAPE_PARENT(shape_id_t shape_id)
+RSHAPE_PARENT_RAW_ID(shape_id_t shape_id)
 {
     return RSHAPE(shape_id)->parent_id;
+}
+
+static inline bool
+RSHAPE_DIRECT_CHILD_P(shape_id_t parent_id, shape_id_t child_id)
+{
+    return (parent_id & SHAPE_ID_FLAGS_MASK) == (child_id & SHAPE_ID_FLAGS_MASK) &&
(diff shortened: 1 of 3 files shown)


shape.h
```diff
@@ -271,11 +271,17 @@ rb_shape_root(size_t heap_id)
...
 
 static inline shape_id_t
-RSHAPE_PARENT(shape_id_t shape_id)
+RSHAPE_PARENT_RAW_ID(shape_id_t shape_id)
 {
     return RSHAPE(shape_id)->parent_id;
 }
 
+static inline bool
+RSHAPE_PARENT_OF(shape_id_t parent_id, shape_id_t child_id)
```
nobu: `PARENT_OF` doesn't feel like a predicate.

casperisfine: Good point.

casperisfine: I renamed it `RSHAPE_DIRECT_CHILD_P`

casperisfine:
Hum, actually the required logic is more complex than that. We need to keep some of the tags, such as the slot size one:

```
[BUG] shape_id heap_index flags mismatch: shape_id_slot_size=88, gc_slot_size=760
```

casperisfine:
Should be good to go now.

jhawthorn (approved):
Thank you! I think it makes the most sense to have `RSHAPE_DIRECT_CHILD_P` as you have done rather than picking an arbitrary "possible" parent

PR #14064, opened 2025-07-31
ZJIT: Implement `ToArray` and `ArrayExtend`

st0012:
- ToArray is used in both `concattoarray` and `splatarray` instructions to convert a value to an array OR put the value in an array
- ArrayExtend is used in `concattoarray` to extend `ary1` with `ary2`

With this change, ZJIT can actually execute `concattoarray` and `splatarray` instructions.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -489,6 +489,42 @@ class TestZJIT < Test::Unit::TestCase
     }
   end
 
+  def test_array_splat_custom_to_a
+    assert_compiles '[1, 2, 3]', %q{
+      class Foo
+        def to_a
+          [1, 2, 3]
+        end
+      end
+      def bar(*a) = a
+      def test = bar(*Foo.new)
+      test
+    }, insns: [:splatarray]
+  end
+
+  def test_array_splat_nil
+    assert_compiles '1', %q{
+      def foo = 1
+      def test = foo(*nil)
+      test
+    }, insns: [:splatarray]
+  end
+
+  def test_array_splat
+    assert_compiles '"foo"', %q{
+      def foo(b) = b
+      def test(a) = foo(*a)
+      test("foo")
+    }, insns: [:splatarray]
+  end
+
+  def test_array_concat_implicit
+    assert_compiles '[1, 2, 3]', %q{
+      def test(a) = [1, *a]
+      test([2, 3])
+    }, insns: [:concattoarray]
+  end
+
   def test_new_range_inclusive
     assert_compiles '1..5', %q{
       def test(a, b) = a..b
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -367,7 +367,8 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::AnyToString { val, str, state } => gen_anytostring(asm, opnd!(val), opnd!(str), &function.frame_state(*state))?,
         Insn::Defined { op_type, obj, pushval, v } => gen_defined(jit, asm, *op_type, *obj, *pushval, opnd!(v))?,
         &Insn::IncrCounter(counter) => return Some(gen_incr_counter(asm, counter)),
-        Insn::ArrayExtend { .. }
+        Insn::ToArray { val, state } => gen_to_array(jit, asm, opnd!(val), &function.frame_state(*state))?,
+        Insn::ArrayExtend { left, right, state } => return gen_array_extend(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state)),
         | Insn::ArrayMax { .. }
         | Insn::ArrayPush { .. }
         | Insn::DefinedIvar { .. }
@@ -379,7 +380,6 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         | Insn::Send { .. }
         | Insn::StringIntern { .. }
         | Insn::Throw { .. }
-        | Insn::ToArray { .. }
         | Insn::ToNewArray { .. }
         | Insn::Const { .. }
         => {
@@ -768,12 +768,8 @@ fn gen_send_without_block(
     self_val: Opnd,
     args: Vec<Opnd>,
 ) -> Option<lir::Opnd> {
-    // Spill locals onto the stack.
-    // TODO: Don't spill locals eagerly; lazily reify frames
-    asm_comment!(asm, "spill locals");
-    for (idx, &insn_id) in state.locals().enumerate() {
-        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
-    }
+    spill_locals(jit, asm, state);
+
     // Spill the receiver and the arguments onto the stack.
     // They need to be on the interpreter stack to let the interpreter access them.
     // TODO: Avoid spilling operands that have been spilled before.
@@ -819,13 +815,8 @@ fn gen_send_without_block_direct(
 
     // Spill the virtual stack and the locals of the caller onto the stack
     // TODO: Lazily materialize caller frames on side exits or when needed
-    asm_comment!(asm, "spill locals and stack");
-    for (idx, &insn_id) in state.locals().enumerate() {
-        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
-    }
-    for (idx, &insn_id) in state.stack().enumerate() {
-        asm.mov(Opnd::mem(64, SP, idx as i32 * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
-    }
+    spill_locals(jit, asm, state);
+    spill_stack(jit, asm, state);
 
     // Set up the new frame
     // TODO: Lazily materialize caller frames on side exits or when needed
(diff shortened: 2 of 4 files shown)


zjit/src/codegen.rs
```diff
@@ -1042,6 +1042,28 @@ fn gen_anytostring(asm: &mut Assembler, val: lir::Opnd, str: lir::Opnd, state: &
     Some(asm_ccall!(asm, rb_obj_as_string_result, str, val))
 }
 
+fn gen_to_array(asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> Option<lir::Opnd> {
+    // Save PC
+    gen_save_pc(asm, state);
```
tekknolagi, line 1038: This does a lot more than just allocate -- it can run arbitrary user code. This means we have to save more state to the EC than just the PC, I think

zjit/src/codegen.rs
```diff
@@ -1042,6 +1042,28 @@ fn gen_anytostring(asm: &mut Assembler, val: lir::Opnd, str: lir::Opnd, state: &
...
+    //
+    // rb_Array satisfies the above conditions, so we can use it here
+    Some(asm_ccall!(asm, rb_Array, val))
+}
+
+fn gen_array_extend(asm: &mut Assembler, left: lir::Opnd, right: lir::Opnd, state: &FrameState) -> Option<()> {
+    // Save PC
+    gen_save_pc(asm, state);
+
+    asm_ccall!(asm, rb_ary_concat, left, right);
```
tekknolagi, line 1064: As above

test/ruby/test_zjit.rb
```diff
@@ -489,6 +489,21 @@ def test = [1,2,3]
     }
   end
 
+  def test_array_splat
+    assert_compiles '"foo"', %q{
+      def foo(b) = b
+      def test(a) = foo(*a)
```
tekknolagi, line 516: can you also test with splatting nil, calling custom `to_a`, ...?

tekknolagi (changes requested):
Needs a bit more work

zjit/src/codegen.rs
st0012: My understanding is that we also need to save SP as the user code can push args to the stack and then raise exception and that can mess up the stack pointer. But I can't find a script that produces such theoretical crash 🤔

tekknolagi: We may need to write out a full frame :/ cc @k0kubun and @XrXr ?

k0kubun: When you are about to call arbitrary methods, which can look at the Binding of caller frames, you'd need to dump locals into the interpreter stack (for now).

When you are about to call arbitrary methods that are not wrapped by `vm_exec` (which does setjmp and protects the caller JIT frame from expiring on longjmp), which can replace the JIT frame with an interpreter frame without going through a side-exit code by raising in the method and catching it on the frame, you'd need to dump stack slots into the interpreter stack (for now).

Both of them seem to apply, so I suspect you need to dump locals and stack slots. Other things should have been already written by `gen_push_frame`.

st0012: Thanks for the explanation!

test/ruby/test_zjit.rb
st0012: Added 👍

zjit/src/codegen.rs
```diff
@@ -768,12 +768,9 @@ fn gen_send_without_block(
     self_val: Opnd,
     args: Vec<Opnd>,
 ) -> Option<lir::Opnd> {
-    // Spill locals onto the stack.
-    // TODO: Don't spill locals eagerly; lazily reify frames
-    asm_comment!(asm, "spill locals");
-    for (idx, &insn_id) in state.locals().enumerate() {
-        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id)?);
-    }
+
```
```suggestion
```

zjit/src/codegen.rs
```diff
@@ -1042,6 +1033,38 @@ fn gen_anytostring(asm: &mut Assembler, val: lir::Opnd, str: lir::Opnd, state: &
...
+    spill_stack(jit, asm, state);
+
+    // ToArray is used in both concattoarray and splatarray instructions
+    // It needs to satisfy the following conditions:
+    // - Check if the value is already an array. If it is, return it
+    // - Check if the value can be converted into an array. If it is, return it
+    // - Return an array with just the value in it
+    //
+    // rb_Array satisfies the above conditions, so we can use it here
+    Some(asm_ccall!(asm, rb_Array, val))
```
k0kubun, line 1052: > ToArray is used in both concattoarray and splatarray instructions

Both `concattoarray` and `splatarray` seem to use `rb_check_to_array`. Can't we use that instead? Generally, when we're just delegating the implementation to the C functions of the interpreter, you shouldn't discover and pick a different C function; you should use the same C function used by the YARV instruction for the best compatibility.

Looking at what `vm_concat_to_array` and `vm_splat_array` do, what `rb_Array` does seems similar to `vm_splat_array`, but not so much for `vm_concat_to_array`. So I don't know if it's a legit translation.

Could you explain the reason why HIR for `concattoarray` needs to call `rb_Array` instead of `vm_concat_to_array` (maybe in the code comment as well)? `vm_concat_to_array` doesn't seem to call `rb_ary_new3` when `rb_check_to_array` returns `nil`, unlike `rb_Array`.

k0kubun: So the HIR for concattoarray is currently `ToArray` + `ArrayExtend`.

What concattoarray does is, when a given `ary2` is not T_ARRAY and doesn't respond to `#to_a`, it pushes `ary2` to `ary1`. So in that case, what HIR needs to do after `rb_check_to_array` is more like `ArrayPush`, not `ArrayExtend`. `rb_Array` allocates a new array in that case, but it shouldn't.

`rb_Array` also checks for `#to_ary` but we shouldn't do that because `concattoarray` doesn't. More reason to avoid functions that the interpreter doesn't use.

So, I feel like the HIR doesn't capture what we need to actually implement. I would suggest to start from introducing an HIR instruction that does `concattoarray` if we aren't sure how to break it up in HIR. Right now, it just seems to be translated to something different from the interpreter's behavior.

st0012: Yeah I didn't notice the `idTo_a` and `idTo_ary` difference on the underlying functions. 
I also struggled to find a good way to extract the supposedly shared ToArray part here and `rb_Array`'s behaviour is the closest that I can find.

I agree that for now it's better to have a dedicated `concattoarray` HIR instruction that just calls into `vm_concat_to_array` (and perhaps do the same with `splatarray`?). And later we can perhaps split the HIR implementations with some new C functions that do just that.

st0012: To make things easier to review I'll close this PR and create a new one.

st0012: Actually, we may be able to make it work by adding a new attribute to the `ToArray` HIR and creating a new C function?

For `splatarray`, `ToArray` needs to do this:

```c
    VALUE tmp = rb_check_to_array(ary);
    if (NIL_P(tmp)) {
        return rb_ary_new3(1, ary);
    }
    else {
        return tmp;
    }
```

For `concattoarray`, `ToArray` needs to do:

```c
    VALUE tmp2 = rb_check_to_array(ary2);
```

So if we have this C function:

```c
VALUE rb_zjit_to_array(VALUE ary, bool wrap_value)
{
    VALUE tmp = rb_check_to_array(ary);
    if (wrap_value && NIL_P(tmp)) {
        return rb_ary_new3(1, ary);
    }
    else {
        return tmp;
    }
}
```

And we add the `wrap_value` to `ToArray`, "I think" we can reproduce the same behaviour?

But for `concattoarray` insn, we still needs to have `ArrayExtend` to handle:

```c
    if (NIL_P(tmp2)) {
        return rb_ary_push(ary1, ary2);
    }
    else {
        return rb_ary_concat(ary1, tmp2);
    }
```

which will need access to `ary1`, `ary2`, result of `rb_check_to_array` on `ary2`.

tekknolagi: This is why we have ToArray and ToNewArray, which I *thought* fully captured the semantics here

tekknolagi:
```rust
    /// Call `to_a` on `val` if the method is defined, or make a new array `[val]` otherwise.
    ToArray { val: InsnId, state: InsnId },
    /// Call `to_a` on `val` if the method is defined, or make a new array `[val]` otherwise. If we
    /// called `to_a`, duplicate the returned array.
    ToNewArray { val: InsnId, state: InsnId },
```

tekknolagi:
```rust
                YARVINSN_concattoarray => {
                    let right = state.stack_pop()?;
                    let left = state.stack_pop()?;
                    let right_array = fun.push_insn(block, Insn::ToArray { val: right, state: exit_id });
                    fun.push_insn(block, Insn::ArrayExtend { left, right: right_array, state: exit_id });
                    state.stack_push(left);
                }
                YARVINSN_splatarray => {
                    let flag = get_arg(pc, 0);
                    let result_must_be_mutable = flag.test();
                    let val = state.stack_pop()?;
                    let obj = if result_must_be_mutable {
                        fun.push_insn(block, Insn::ToNewArray { val, state: exit_id })
                    } else {
                        fun.push_insn(block, Insn::ToArray { val, state: exit_id })
                    };
                    state.stack_push(obj);
                }
```

k0kubun: > we may be able to make it work by adding a new attribute to the ToArray HIR and creating a new C function?

It makes sense to me to let `ToArray` have the mode that doesn't allocate a new array, but ideally the overhead of the `rb_zjit_to_array` -> `rb_check_to_array` C call should not exist. When `wrap_value` is false, JIT code should call `rb_check_to_array` directly. So you should be able to remove the `wrap_value` from the function for the other case. Defining that function in array.c might allow us to reduce the call overhead too.

> But for concattoarray insn, we still needs to have ArrayExtend to handle:
> which will need access to ary1, ary2, result of rb_check_to_array on ary2.

Yeah, and we might need to skip that behavior for other use of `ArrayExtend`.

> This is why we have ToArray and ToNewArray, which I thought fully captured the semantics here

The difference between `ToArray` and `ToNewArray` seems to apply only when `#to_a` is defined on `val`. But we're talking about the case that `#to_a` is not defined; `splatarray` allocates a new array, but `concattoarray` doesn't.

Since `concattoarray` does not allocate a new array in the case that `val` doesn't have `#to_a`, ZJIT should not use the `ToArray` that "makes a new array `[val]`" for `concattoarray`. Such a `val` should be pushed straight to the other array without creating an extra object.

PR #14066, opened 2025-07-31
Make ARGF not pin references during marking

peterzhu2118:
ARGF supports compaction, but it pins all of its references, which means that none of it can move. This commit changes it to actually support compaction.


michaeltyra19-collab:
Okay

michaeltyra19-collab:
peterzhu2118:argf-mark-and-move

PR #14072, opened 2025-08-01
Fix transition of object_id to too complex shape

peterzhu2118:
object_id0 tried to generate a new object ID, but this object ID may betoo complex. rb_obj_field_set didn't handle this transition and thus caused crashes.

This commit changes object_id to use the regular rb_ivar_set_internal API to handle the too complex transition.

For example, the following script crashes:

```ruby
o = Object.new
o.instance_variable_set(:https://github.com/hello, "hello!")

RubyVM::Shape.exhaust_shapes

puts o.object_id
```

With:

```
gc.c:1908: Assertion Failed: object_id0:RBASIC_SHAPE_ID(obj) == object_id_shape_id
```

### DIFF
--- a/ext/objspace/objspace_dump.c
+++ b/ext/objspace/objspace_dump.c
@@ -818,9 +818,6 @@ shape_id_i(shape_id_t shape_id, void *data)
         dump_append_id(dc, RSHAPE_EDGE_NAME(shape_id));
 
         break;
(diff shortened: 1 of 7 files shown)


casperisfine:
I have a much simpler fix: https://github.com/ruby/ruby/pull/14073

PR #14076, opened 2025-08-01
Add `void rb_gc_update_reference(VALUE *ptr)` API

casperisfine:
The existing `VALUE rb_gc_location(VALUE obj)` API isn't ideal in my opinion, because the idomatic way to use it is:

```c
struct->field = rb_gc_location(struct->field);
```

Which means that when running `GC.compact`, any reference that is movable is rewritten, even if the referenced object didn't move.

I suspect this is a lot of unnecessary writes that are trashing shared pages.

Inside the GC itself, there's the much more convenient `UPDATE_IF_MOVED` macro, that only update the reference if it's actually needed.

I only rewrote a single `dcompact` function to show how it works, and haven't made the API public. But @peterzhu2118 and @eightbitraptor: if you agree this is a good idea, I'd like to propose this to be made a public API.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -2878,6 +2878,15 @@ rb_gc_location(VALUE value)
     return gc_location_internal(rb_gc_get_objspace(), value);
 }
 
+void
+rb_gc_update_reference(VALUE *value_ptr)
+{
+    void *objspace = rb_gc_get_objspace();
+    if (rb_gc_impl_object_moved_p(objspace, *value_ptr)) {
+        *value_ptr = gc_location_internal(objspace, *value_ptr);
+    }
+}
+
 #if defined(__wasm__)
 
 
--- a/internal/gc.h
+++ b/internal/gc.h
@@ -210,6 +210,7 @@ bool rb_gc_size_allocatable_p(size_t size);
 size_t *rb_gc_heap_sizes(void);
 size_t rb_gc_heap_id_for_size(size_t size);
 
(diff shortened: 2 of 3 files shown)


casperisfine:
> IMO for most cases it's just better to use `rb_gc_mark_and_move`.

That's a private function. So yes for internal uses. But I'm thinking for `TypedData` here. Can you use `rb_gc_mark_and_move` with `TypedData`?

peterzhu2118:
> That's a private function.

Yeah I think we can make it public.

> Can you use rb_gc_mark_and_move with TypedData?

Just did that in https://github.com/ruby/ruby/pull/14066.

casperisfine:
Ok. `mark_and_move` currently have the same issue, but the interface does allow to fix it:

```c
void
rb_gc_impl_mark_and_move(void *objspace_ptr, VALUE *ptr)
{
    rb_objspace_t *objspace = objspace_ptr;

    if (RB_UNLIKELY(objspace->flags.during_reference_updating)) {
        GC_ASSERT(objspace->flags.during_compacting);
        GC_ASSERT(during_gc);

        *ptr = rb_gc_impl_location(objspace, *ptr);
    }
    else {
        gc_mark(objspace, *ptr);
    }
}
```

Wouldn't be hard to fix it. I'll probably PR that in a bit.

casperisfine:
Superseeded by https://github.com/ruby/ruby/pull/14081

PR #7511, opened 2023-03-13
YJIT: Introduce no_gc attribute

k0kubun:
* Introduce `no_gc` attribute to skip that `jit_prepare_routine_call` when possible
  * If a method doesn't allocate, we should be able to skip `jit_prepare_routine_call`.

`Kernel#class` is one of such methods that are called often on railsbench https://github.com/Shopify/yjit-bench/pull/168.

PR #14078, opened 2025-08-01
ZJIT: Annotate builtin functions

st0012:
This is an attempt to support annotating builtin functions, like `Float`, `GC.start`...etc.

The current implementation only uses annotation to provide the return type info. So instead of 

```
Optimized HIR:
fn test@test.rb:2:
bb0(v0:BasicObject):
  v2:Fixnum[1] = Const Value(1)
  v4:BasicObject = SendWithoutBlock v0, :Float, v2
  Return v4

Optimized HIR:
fn Float@<internal:kernel>:195:
  v4:BasicObject = InvokeBuiltin rb_f_float1, v0, v1
  Return v4
```

we can have return type information with annotations like `annotate_builtin!("rb_f_float", types::Flonum);`

```
Optimized HIR:
fn Float@<internal:kernel>:195:
  v4:Flonum = InvokeBuiltin rb_f_float1, v0, v1
  Return v4
```

### DIFF
--- a/zjit/src/cruby_methods.rs
+++ b/zjit/src/cruby_methods.rs
@@ -99,6 +99,8 @@ fn annotate_builtin_method(props_map: &mut HashMap<*mut c_void, FnProperties>, c
 
         // Scan through the ISEQ to find invokebuiltin instructions
         let mut insn_idx: u32 = 0;
+        let mut func_ptr = std::ptr::null_mut::<c_void>();
+
         while insn_idx < encoded_size {
             // Get the PC for this instruction index
             let pc = rb_iseq_pc_at_idx(iseq, insn_idx);
@@ -112,13 +114,22 @@ fn annotate_builtin_method(props_map: &mut HashMap<*mut c_void, FnProperties>, c
                 // The first operand is the builtin function pointer
                 let bf_value = *pc.add(1);
                 let bf_ptr = bf_value.as_ptr() as *const rb_builtin_function;
-                let func_ptr = (*bf_ptr).func_ptr as *mut c_void;
-                props_map.insert(func_ptr, props);
+
+                if func_ptr.is_null() {
+                    func_ptr = (*bf_ptr).func_ptr as *mut c_void;
+                } else {
+                    panic!("Multiple invokebuiltin instructions found in ISEQ for {}#{}",
+                        std::ffi::CStr::from_ptr(rb_class2name(class)).to_str().unwrap_or("?"),
+                        method_name);
+                }
             }
 
             // Move to the next instruction using the proper length
             insn_idx = insn_idx.saturating_add(rb_insn_len(VALUE(opcode as usize)).try_into().unwrap());
         }
+
+        // Only insert the properties if its iseq has exactly one invokebuiltin instruction
+        props_map.insert(func_ptr, props);
     }
 }
 


zjit/src/codegen.rs
```diff
@@ -337,6 +337,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
             gen_send_without_block(jit, asm, *cd, &function.frame_state(*state), opnd!(self_val), opnds!(args))?,
         Insn::SendWithoutBlockDirect { cme, iseq, self_val, args, state, .. } => gen_send_without_block_direct(cb, jit, asm, *cme, *iseq, opnd!(self_val), opnds!(args), &function.frame_state(*state))?,
         Insn::InvokeBuiltin { bf, args, state } => gen_invokebuiltin(asm, &function.frame_state(*state), bf, opnds!(args))?,
+        Insn::BuiltinCall { bf, args, state, return_type: _ } => gen_invokebuiltin(asm, &function.frame_state(*state), bf, opnds!(args))?,
```
tekknolagi: I think we can keep the `InvokeBuiltin` instruction. It's already quite specialized and has a function pointer built-in. You can probably use that in `Annotations`

zjit/src/cruby_methods.rs
```diff
@@ -74,6 +81,22 @@ pub fn init() -> Annotations {
         }
     }
 
+    macro_rules! annotate_builtin {
+        ($name:literal, $return_type:expr) => {
+            annotate_builtin!($name, $return_type, no_gc, leaf, elidable)
```
tekknolagi: Is there a way to look up a builtin by string name? If so we should do that here so we can correlate by fptr already in InvokeBuiltin

st0012: I don't see such structure exists but maybe I overlooked. @k0kubun @XrXr do you know if there's a table of builtin functions (like `Float`) and their names available during compile time?

zjit/src/codegen.rs
st0012: Updated

zjit/src/cruby_methods.rs
k0kubun: If you look at the source code (`kernel.rbinc` for `#Float`), there's a table of `rb_builtin_function[]`, but it's not kept after `rb_load_with_builtin_functions`.

The only thing you can do is probably to look at the ISEQ (you'd need to specify `rb_mKernel` and `"Float"` and look up a method with it) and look for the invokebuiltin instruction (scan every invokebuiltin and check if each of `bf->name` matches, I guess).

zjit/src/cruby_methods.rs
```diff
@@ -83,7 +124,13 @@ pub fn init() -> Annotations {
     annotate!(rb_cNilClass, "nil?", types::TrueClass, no_gc, leaf, elidable);
     annotate!(rb_mKernel, "nil?", types::FalseClass, no_gc, leaf, elidable);
 
+    // Annotate builtin functions
+    annotate_builtin!("rb_f_float", types::Flonum);
+    annotate_builtin!("rb_f_float1", types::Flonum);
```
st0012: For methods like `Kernel#class` that use `cexpr!`, we'd need to annotate them with there generated function name, which will be `_bi#{lineno}`, like `_bi20` for `#class`.

zjit/src/cruby_methods.rs
```diff
@@ -74,6 +99,22 @@ pub fn init() -> Annotations {
...
+            annotate_builtin!($name, $return_type, no_gc, leaf, elidable)
+        };
+        ($name:literal, $return_type:expr, $($properties:ident),+) => {
+            let mut props = FnProperties {
+                no_gc: false,
+                leaf: false,
+                elidable: false,
+                return_type: $return_type
+            };
+            $(props.$properties = true;)+
```
k0kubun, line 163: So in CRuby, we already have a way to annotate builtin, which is `Primitive.attr! :leaf`, which YJIT uses for optimization. We also used to have `:no_gc`, but we abandoned it since the patch failed the CI for reasons.

While ZJIT currently doesn't verify the annotated leafness, the CI verifies the validity of `Primitive.attr! :leaf`. So for `leaf`, it's probably better to support loading those annotations and annotate leaf with that when possible, at least until we remove YJIT.

Just to be clear, I'm not proposing to use `Primitive.attr!` for non-leaf stuff here. We'd still need ZJIT-level annotations for describing types in the ZJIT way.

zjit/src/cruby_methods.rs
tekknolagi: I wonder if we can have a hook like `zjit_observe_builtin_function()` on load

zjit/src/cruby_methods.rs
k0kubun: So do you intend to use `annotate_builtin!("_bi20", ...);` for `Kernel#class` here?

First, this function name is not unique. Any other class on line 20 can also define `_bi20` with a different definition (because they're static functions). The annotation should be scoped to the class (`Kernel`) at least if we do this.

Second, matching a function by the line number seems fragile even if it's scoped to the class. I don't think we want to fail the CI and update this Rust code whenever somebody contributes to builtin source code. In the worst case, every line number would end up matching a different function, and silently break the behavior without noticing.

---

My proposal is to do one of these:

1. Support only ISEQs that have only one invokebuiltin, and annotate them based on the class and the method.
    * This should cover the vast majority of methods that are converted from C methods.
2. Update `mk_builtin_loader.rb` to give a unique name to builtin functions for `cexpr!`
    1. Automatically name it like `_bi_kernel_class_0` (`kernel`: file name, `class`: ISEQ name, `0`: the index of the invokebuiltin in the ISEQ)
    2. Let `Primitive.cexpr!` optionally take an identifier and use it, e.g. `Primitive.cexpr! %{ ... }, :kernel_class`

(1) would be easier to implement than (2), so it might make sense to do (1) as a first step.

zjit/src/cruby_methods.rs
k0kubun: It'd be called before the ZJIT state is initialized, so there's no initialized storage for it and that might complicate the implementation, but not impossible.

But whether we need it or not kind of depends on what we do for https://github.com/ruby/ruby/pull/14078#discussion_r2248671070.

st0012: I switched to use `annotate_builtin!(rb_mKernel, "Float", types::Flonum);` for annotation, which allows ZJIT to find the iseq, get the `invokebuiltin*` insns, and get the function pointer as @k0kubun suggested.

zjit/src/cruby_methods.rs
```diff
@@ -59,10 +66,66 @@ fn annotate_c_method(props_map: &mut HashMap<*mut c_void, FnProperties>, class:
     props_map.insert(fn_ptr, props);
 }
 
+/// Look up a method and find its builtin function pointer by parsing its ISEQ
+fn annotate_builtin_method(props_map: &mut HashMap<*mut c_void, FnProperties>, class: VALUE, method_name: &'static str, props: FnProperties) {
+    unsafe {
```
st0012, line 72: @k0kubun I used `rb_yjit_constant_ic_update` as a reference for this and I see it has `with_vm_lock` before going through the iseq. Is it safe to not have it here given that it's executed at compile time?

zjit/src/cruby_methods.rs
```diff
@@ -59,10 +66,66 @@ fn annotate_c_method(props_map: &mut HashMap<*mut c_void, FnProperties>, class:
...
+        if iseq.is_null() {
+            panic!("Failed to get ISEQ for {}#{}",
+                std::ffi::CStr::from_ptr(rb_class2name(class)).to_str().unwrap_or("?"),
+                method_name);
+        }
+
+        // Get the size of the ISEQ in instruction units
+        let encoded_size = rb_iseq_encoded_size(iseq);
+
+        // Scan through the ISEQ to find invokebuiltin instructions
```
k0kubun, line 100: The annotation may not be valid for all the invokebuiltin functions for the ISEQ at the same time. As I mentioned at https://github.com/ruby/ruby/pull/14078#discussion_r2248671070, it should support only ISEQs that have only one invokeblock instruction, rejecting ISEQs with multiple invokebuiltin instructions. If methods that currently have only one invokeblock instruction suddenly gets another invokeblock instruction, that will be a problem. So they should be rejected.

If we want to support other kinds of ISEQs, we need to change `mk_builtin_loader.rb` to have better identifiers for invokeblock functions.

zjit/src/cruby_methods.rs
k0kubun: > given that it's executed at compile time?

Is it? 

`cruby_methods::init()` is called in `ZJITState::init()` during boot. Until we support lazily initializing ZJIT, there should be no non-main Ractor when the interpreter is not even initialized. Even when we do, I don't think it's going to be this function's responsibility to take the VM lock; we will need it for the whole `rb_zjit_init()`.

st0012: > cruby_methods::init() is called in ZJITState::init() during boot.

🤦‍♂️ I confused myself about the macro/function's execution time. 

> Even when we do, I don't think it's going to be this function's responsibility to take the VM lock; we will need it for the whole rb_zjit_init().

Thanks for clarifying it

zjit/src/cruby_methods.rs
st0012: Thanks. Updated 👍

zjit/src/cruby_methods.rs
```diff
@@ -59,10 +66,86 @@ fn annotate_c_method(props_map: &mut HashMap<*mut c_void, FnProperties>, class:
...
+                method_name);
+        }
+
+        // Check if the ISEQ has the BUILTIN_ATTR_LEAF attribute
+        let builtin_attrs = rb_zjit_get_iseq_body_builtin_attrs(iseq);
+        if (builtin_attrs & BUILTIN_ATTR_LEAF) != 0 {
+            // If the method is marked as leaf via Primitive.attr!, set leaf to true
+            props.leaf = true;
+            // Leaf methods don't trigger GC
+            props.no_gc = true;
```
k0kubun: Could we avoid deriving `no_gc` from `leaf`, and just annotate `no_gc` separately in ZJIT for now? 

We have the history of adding `:no_gc` attribute while `:leaf` attribute already exists https://github.com/ruby/ruby/pull/7511, so they're considered separate things. `leaf` just means that it doesn't call other methods, so semantically it's still allowed to allocate.

k0kubun (approved):
looks good once we fix https://github.com/ruby/ruby/pull/14078#discussion_r2248953737

zjit/src/cruby_methods.rs
st0012: I see. Updated 👍

zjit/src/cruby_methods.rs
```diff
@@ -59,10 +66,84 @@ fn annotate_c_method(props_map: &mut HashMap<*mut c_void, FnProperties>, class:
...
+            panic!("Failed to get ISEQ for {}#{}",
+                std::ffi::CStr::from_ptr(rb_class2name(class)).to_str().unwrap_or("?"),
+                method_name);
+        }
+
+        // Check if the ISEQ has the BUILTIN_ATTR_LEAF attribute
+        let builtin_attrs = rb_zjit_get_iseq_body_builtin_attrs(iseq);
+        if (builtin_attrs & BUILTIN_ATTR_LEAF) != 0 {
+            // If the method is marked as leaf via Primitive.attr!, set leaf to true
+            props.leaf = true;
```
k0kubun: I think the idea of https://github.com/ruby/ruby/pull/14078#discussion_r2248624814 is the most useful when we don't annotate the method at all but it's automatically populated by `Primitive.attr!`.

If we use it this way (use the attribute only when the same method is annotated inside ZJIT), it kinda defeats the purpose. It's rather confusing that those ZJIT annotations don't have `leaf` on ZJIT's end. 

Could we just drop this builtin attribute sync for now, leaving https://github.com/ruby/ruby/pull/14078#discussion_r2248624814 for another PR to address, and just write `leaf` (and `no_gc`) on `annotate_builtin!` calls inside ZJIT? Not writing `leaf` on `annotate_builtin!` can be confusing when they're indeed leaf. Sorry for the back and forth.

st0012: Ok I created https://github.com/Shopify/ruby/issues/670 to track it as well

zjit/src/cruby_methods.rs
```diff
@@ -83,7 +173,13 @@ pub fn init() -> Annotations {
     annotate!(rb_cNilClass, "nil?", types::TrueClass, no_gc, leaf, elidable);
     annotate!(rb_mKernel, "nil?", types::FalseClass, no_gc, leaf, elidable);
 
+    // Annotate builtin functions using the new method-based approach
+    annotate_builtin!(rb_mKernel, "Float", types::Flonum);
+    annotate_builtin!(rb_mKernel, "Integer", types::Integer);
+    annotate_builtin!(rb_mKernel, "class", types::Class);
```
k0kubun: In https://github.com/ruby/ruby/pull/14078#discussion_r2248964104, I was suggesting to add

```suggestion
    annotate_builtin!(rb_mKernel, "class", types::Class, leaf);
```

for the methods we choose to annotate in ZJIT. Even if it's duplicated with `Primitive.attr!`, it seems confusing that this method misses `leaf` while it's possible to annotate that with this `annotate_builtin!` macro.

zjit/src/cruby_methods.rs
```diff
@@ -83,7 +173,13 @@ pub fn init() -> Annotations {
     annotate!(rb_cNilClass, "nil?", types::TrueClass, no_gc, leaf, elidable);
     annotate!(rb_mKernel, "nil?", types::FalseClass, no_gc, leaf, elidable);
 
+    // Annotate builtin functions using the new method-based approach
```
k0kubun: I'm not sure if I follow what's "new" about it? If you mean it's new because it's just implemented, it may not hold for future readers.

To clarify that it's an annotation on _the only_ builtin function, I would write:

```suggestion
    // Annotate builtin functions for ISEQs that have only one invokeblock instruction
```

st0012: Yeah I agree it's not too helpful in the future so I've deleted it.

st0012:
@k0kubun thanks for all the pointers 🙌

zjit/src/cruby_methods.rs
```diff
@@ -83,7 +174,12 @@ pub fn init() -> Annotations {
     annotate!(rb_cNilClass, "nil?", types::TrueClass, no_gc, leaf, elidable);
     annotate!(rb_mKernel, "nil?", types::FalseClass, no_gc, leaf, elidable);
 
+    annotate_builtin!(rb_mKernel, "Float", types::Flonum);
```
eregon, line 177: Does `types::Flonum` mean the same as just `Float`?
I would think `Kernel#Float` can return either a flonum (an immediate/tagged pointer) or a heap-allocated Float, depending on the argument, and so it seems wrong to say the return type is "flonum" when it can be either.
(same issue in the PR description)

k0kubun: Good catch. I agree this should be `types::Float`.

st0012: Thanks. I'll open a PR for it

st0012: Opened https://github.com/ruby/ruby/pull/14123

PR #14083, opened 2025-08-01
ZJIT: Fix side-exit panicking when there's too many locals

XrXr:
Previously, ARM64 panicked due to compiled_side_exits() when the memory
displacement got large enough to exceed the 9 bits limit. Usually, we split
these kind of memory operands, but compiled_side_exits() runs after
split.

Using scratch registers, implement `Insn::Store` on ARM such that it can
handle large displacements without split(). Do this for x86 as well, and
remove arch specific code from compiled_side_exits(). We can now pass
`TestKeywordArguments`.

Since `Insn::Store` doesn't need splitting now, users enjoy lower
register pressure.

Downside is, using `Assembler::SCRATCH_REG` as a base register is now
sometimes an error, depending on whether `Insn::Store` also needs to
use the register. It seems a fair trade off since `SCRATCH_REG` is
not often used, and we don't put it as a base register anywhere at the
moment.

Fixes: https://github.com/Shopify/ruby/issues/650

### DIFF
--- a/zjit/src/backend/arm64/mod.rs
+++ b/zjit/src/backend/arm64/mod.rs
@@ -196,10 +196,10 @@ impl Assembler
     /// Avoid using if you can since this is used to lower [Insn] internally and
     /// so conflicts are possible.
     pub const SCRATCH_REG: Reg = X16_REG;
-    const SCRATCH0: A64Opnd = A64Opnd::Reg(Assembler::SCRATCH_REG);
-    const SCRATCH0_REG: Reg = Assembler::SCRATCH_REG;
-    const SCRATCH1: A64Opnd = A64Opnd::Reg(X17_REG);
+    const SCRATCH0_REG: Reg = Self::SCRATCH_REG;
     const SCRATCH1_REG: Reg = X17_REG;
+    const SCRATCH0: A64Opnd = A64Opnd::Reg(Self::SCRATCH0_REG);
+    const SCRATCH1: A64Opnd = A64Opnd::Reg(Self::SCRATCH1_REG);
 
     /// Get the list of registers from which we will allocate on this platform
     pub fn get_alloc_regs() -> Vec<Reg> {
--- a/zjit/src/backend/arm64/mod.rs
+++ b/zjit/src/backend/arm64/mod.rs
@@ -191,11 +191,15 @@ pub const ALLOC_REGS: &'static [Reg] = &[
 
 impl Assembler
 {
-    // Special scratch registers for intermediate processing.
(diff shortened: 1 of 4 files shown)


michaeltyra19-collab:
> Previously, ARM64 panicked due to compiled_side_exits() when the memory displacement got large enough to exceed the 9 bits limit. Usually, we split these kind of memory operands, but compiled_side_exits() runs after split.
> 
> Using scratch registers, implement `Insn::Store` on ARM such that it can handle large displacements without split(). Do this for x86 as well, and remove arch specific code from compiled_side_exits(). We can now run `TestFixnum` and `TestKeywordArguments`.
> 
> Since `Insn::Store` doesn't need splitting now, users enjoy lower register pressure.
> 
> Downside is, using `Assembler::SCRATCH_REG` as a base register is now sometimes an error, depending on whether `Insn::Store` also needs to use the register. It seems a fair trade off since `SCRATCH_REG` is not often used, and we don't put it as a base register anywhere at the moment.

B

zjit/src/backend/arm64/mod.rs
```diff
@@ -191,11 +191,15 @@ pub const ALLOC_REGS: &'static [Reg] = &[
...
-    // Special scratch registers for intermediate processing.
-    // This register is caller-saved (so we don't have to save it before using it)
+    /// Special scratch registers for intermediate processing.
+    /// This register is call-clobbered (so we don't have to save it before using it).
+    /// Avoid using if you can since this is used to lower [Insn] internally and
+    /// so conflicts are possible.
     pub const SCRATCH_REG: Reg = X16_REG;
     const SCRATCH0: A64Opnd = A64Opnd::Reg(Assembler::SCRATCH_REG);
+    const SCRATCH0_REG: Reg = Assembler::SCRATCH_REG;
     const SCRATCH1: A64Opnd = A64Opnd::Reg(X17_REG);
```
k0kubun: This would make it easier to understand their relationship:

```suggestion
    const SCRATCH0: A64Opnd = A64Opnd::Reg(Assembler::SCRATCH0_REG);
    const SCRATCH0_REG: Reg = Assembler::SCRATCH_REG;
    const SCRATCH1: A64Opnd = A64Opnd::Reg(Assembler::SCRATCH1_REG);
```

k0kubun (approved):
I hope we won't keep hitting cases like this where we need to move split the pass to emit. As long as we don't have a conflict on those scratch registers, it might be fine though.

If this becomes too complicated, we might want to consider compiling side exits in separate Assembler instances (which comes with its own complexity).

PR #14084, opened 2025-08-01
Ensure CC entries always marked, add missing WB

jhawthorn:
Previously we were issuing writebarriers for each cc, but were missing the cme.

We also need to avoid it being possible to run GC after we've copied the values into the allocated array, but before they're visible in the object.

### DIFF
--- a/vm_method.c
+++ b/vm_method.c
@@ -136,17 +136,19 @@ rb_vm_cc_table_create(size_t capa)
 static enum rb_id_table_iterator_result
 vm_cc_table_dup_i(ID key, VALUE old_ccs_ptr, void *data)
 {
+    VALUE new_table = (VALUE)data;
     struct rb_class_cc_entries *old_ccs = (struct rb_class_cc_entries *)old_ccs_ptr;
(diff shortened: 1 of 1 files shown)


vm_method.c
```diff
@@ -136,17 +136,19 @@ rb_vm_cc_table_create(size_t capa)
 static enum rb_id_table_iterator_result
 vm_cc_table_dup_i(ID key, VALUE old_ccs_ptr, void *data)
 {
+    VALUE new_table = (VALUE)data;
     struct rb_class_cc_entries *old_ccs = (struct rb_class_cc_entries *)old_ccs_ptr;
     size_t memsize = vm_ccs_alloc_size(old_ccs->capa);
-    struct rb_class_cc_entries *new_ccs = ruby_xmalloc(memsize);
+    struct rb_class_cc_entries *new_ccs = ruby_xcalloc(1, memsize);
```
byroot, line 142: Is `calloc` really required? `new_table` is right sized so inserting in it should never result in an allocation.

Not a huge deal though

byroot:
Thanks !

PR #14086, opened 2025-08-01
ZJIT: Enable IncrCounter for arm64

k0kubun:
This PR fixes `--zjit-stats`'s panic on arm64.

The backend did not have label support when it was initially ported from YJIT, but we added that later. So there's no reason to comment out the `IncrCounter` implementation today.


XrXr (approved):

tekknolagi:
Thank you!

PR #14087, opened 2025-08-02
Add `Math.log1p` and `Math.expm1`

mame:
This commit adds two new methods to the `Math` module:

* `Math.log1p(x)`: Computes `Math.log(x + 1)`
* `Math.expm1(x)`: Computes `Math.exp(x) - 1`

These methods are often more accurate than the straightforward computation, especially when `x` is close to zero. The corresponding functions, `log1p` and `expm1`, are defined in the C99 standard math library.

### DIFF
--- a/math.c
+++ b/math.c
@@ -457,6 +457,34 @@ math_exp(VALUE unused_obj, VALUE x)
     return DBL2NUM(exp(Get_Double(x)));
 }
 
+/*
+ *  call-seq:
+ *    Math.expm1(x) -> float
+ *
+ *  Returns "exp(x) - 1", +e+ raised to the +x+ power, minus 1.
+ *
+ *
+ *  - Domain: <tt>[-INFINITY, INFINITY]</tt>.
+ *  - Range: <tt>[-1.0, INFINITY]</tt>.
+ *
+ *  Examples:
+ *
+ *    expm1(-INFINITY) # => 0.0
+ *    expm1(-1.0)      # => -0.6321205588285577 # 1.0/E - 1
+ *    expm1(0.0)       # => 0.0
+ *    expm1(0.5)       # => 0.6487212707001282  # sqrt(E) - 1
+ *    expm1(1.0)       # => 1.718281828459045   # E - 1
+ *    expm1(2.0)       # => 6.38905609893065    # E**2 - 1
+ *    expm1(INFINITY)  # => Infinity
+ *
+ */
+
+static VALUE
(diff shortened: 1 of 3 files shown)


math.c
```diff
@@ -457,6 +457,33 @@ math_exp(VALUE unused_obj, VALUE x)
     return DBL2NUM(exp(Get_Double(x)));
 }
 
+/*
+ *  call-seq:
+ *    Math.expm1(x) -> float
+ *
+ *  Returns +e+ raised to the +x+ power minus 1.
```
notEthan: The precedence is not indicated very well with the texts, this could be `Math.exp(x) - 1` or `Math.exp(x - 1)`

The other,  "Returns the base E [logarithm](https://en.wikipedia.org/wiki/Logarithm) of `x` + 1." is more confusing since operator `+` is generally lower precedence.

math.c
```diff
@@ -457,6 +457,33 @@ math_exp(VALUE unused_obj, VALUE x)
...
+ *
+ *  Examples:
+ *
+ *    expm1(-INFINITY) # => 0.0
+ *    expm1(-1.0)      # => -0.6321205588285577 # 1.0/E - 1
+ *    expm1(0.0)       # => 0.0
+ *    expm1(0.5)       # => 0.6487212707001282  # sqrt(E) - 1
+ *    expm1(1.0)       # => 1.718281828459045   # E - 1
+ *    expm1(2.0)       # => 6.38905609893065    # E**2 - 1
+ *    expm1(INFINITY)  # => Infinity
```
notEthan, line 478: Minor, but the casing is inconsistent.

`e` is always lowercase as I've seen it, apart from `Math::E` where it has to be uppercase, but these aren't referring to that constant so I think should be `e`.

INFINITY uppercase some places and Infinity others (Float::INFINITY exists but top-level doesn't)

nobu: Because `Float::INFINITY.to_s` returns `"Infinity"`.

math.c
mame: Thanks, I will change them to:

* `Returns "exp(x) - 1", +e+ raised to the +x+ power, minus 1.`
* `Returns "log(x + 1)", the base E logarithm of (+x+ + 1).`

math.c
mame: The examples are copyed from the doc of `Math.exp`. I guess the examples are assuming `include Math`. In any way, please create another issue if needed.

PR #14088, opened 2025-08-03
Add link from help text to options doc

BurdetteLamar:


ruby.c
```diff
@@ -411,6 +411,8 @@ usage(const char *name, int help, int highlight, int columns)
 
     if (highlight) sb = esc_standout;
 
+    printf("%s""Options:%s\n", sb, se);
+    printf("%s""(details and examples at https://docs.ruby-lang.org/en/master/ruby/options_md.html)%s\n", sb, se);
```
peterzhu2118, line 415: Ideally, I think we should link to the released documentation for released versions, rather than always linking to the master branch docs since it may not apply to that version. So for example, when Ruby 3.5 comes out later this year, it should link to `https://docs.ruby-lang.org/en/3.5/ruby/options_md.html`

BurdetteLamar: So for now, would be 3.4?  How does it get updated as we go forward?

peterzhu2118: No, Ruby 3.4 is already released, so it would not get this change. The current changes will be reflected in Ruby 3.5. It can be automatically changed by using the `RUBY_VERSION_MAJOR` and `RUBY_VERSION_MINOR` macros.

BurdetteLamar: Ah, yes.  As I said in the OP, I haven't tested this (and in fact can't without building Ruby (which I avoid)).  Can you either fix now and merge, or merge and fix in later PR?

peterzhu2118: I implemented this in https://github.com/ruby/ruby/pull/14142. However, I think `options.md` needs some updating before we should merge it. Specifically, the [JIT options are no longer experimental](https://docs.ruby-lang.org/en/master/ruby/options_md.html#label-Experimental+Options). It's also missing a bunch of options described [here](https://docs.ruby-lang.org/en/master/yjit/yjit_md.html#command-line-options).

BurdetteLamar: This PR is superseded, and should be closed?  I think updates to `options.md` is out of scope here?  If agreed, I'll open a new PR for that.

peterzhu2118: Yes, I think it's out of scope. But we should update it.

BurdetteLamar: > Yes, I think it's out of scope. But we should update it.

So, we have the no-longer-experimental `--jit*` family of options, and the `--yjit*` family as well?  I see the write-up for YJIT, but none for JIT;  should there be?

peterzhu2118: It actually looks like all of the `--jit*` options are dead except for `--jit`, which is just an alias for `--yjit`.

PR #14093, opened 2025-08-03
[DOC] Tweaks for GC.start

BurdetteLamar:


gc.rb
```diff
@@ -7,31 +7,39 @@
...
-  # The +immediate_sweep+ keyword argument determines whether or not to defer
-  # sweeping (using lazy sweep). When set to +false+, sweeping is performed in
-  # steps that are interleaved with future Ruby code execution, so sweeping might
-  # not be completed during this method call. When set to +true+, sweeping is
-  # completed during the call to this method.
-  #
-  # Note: These keyword arguments are implementation and version-dependent. They
-  # are not guaranteed to be future-compatible and may be ignored if the
-  # underlying implementation does not support them.
+  # Initiates garbage collection, even if explicitly disabled by GC.disable .
```
peterzhu2118: Nit: whitespace

```suggestion
  # Initiates garbage collection, even if explicitly disabled by GC.disable.
```

gc.rb
```diff
@@ -7,31 +7,39 @@
...
+  #
+  # Keyword arguments:
+  #
+  # - +full_mark+:
+  #   its boolean value determines whether to perform a major garbage collection cycle:
+  #
+  #   - +true+: initiates a major garbage collection cycle is run,
```
peterzhu2118: I think the "is run" part is redundant here:

```suggestion
  #   - +true+: initiates a major garbage collection cycle,
```

PR #14094, opened 2025-08-04
[Bug #21528] Always produce strings with valid encoding when truncating syntax errors

Earlopain:
https://bugs.ruby-lang.org/issues/21528

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -10706,6 +10706,21 @@ pm_parse_errors_format_line(const pm_parser_t *parser, const pm_newline_list_t *
         start += column_start;
     }
 
+    // Take care to not produce a string that is invalid in its encoding.
+    if (truncate_end) {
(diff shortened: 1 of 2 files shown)


prism_compile.c
```diff
@@ -10706,6 +10706,21 @@ pm_parse_errors_format_line(const pm_parser_t *parser, const pm_newline_list_t *
...
+        while (end_adjusted < end) {
+            const size_t next_width = pm_encoding_utf_8_char_width(end_adjusted, 4);
+            // We should only get here with valid utf-8
+            RUBY_ASSERT(next_width > 0);
+
+            if (end_adjusted + next_width > end) {
+                end = end_adjusted;
+            }
+            end_adjusted += next_width;
+        }
```
nobu, line 10721: Doesn't prism/encoding.c provide something similar to `rb_enc_left_char_head`?

PR #14095, opened 2025-08-04
Struct: keep direct reference to IMEMO/fields when space allows

etiennebarrie:


byroot:
Superseeded by: https://github.com/ruby/ruby/pull/14129

PR #14097, opened 2025-08-04
ZJIT: Remove passing tests from exclusions

st0012:


test/.excludes-zjit/TestTimeout.rb
```diff
@@ -1,3 +1 @@
-exclude(:test_timeout, 'Test hangs with ZJIT')
```
k0kubun: It doesn't seem to be passing on the CI.

st0012: Hmm interesting it was passing in [the previous run](https://github.com/ruby/ruby/actions/runs/16730475848/job/47357055918)

test/.excludes-zjit/TestTimeout.rb
```diff
@@ -1,3 +1,2 @@
 exclude(:test_timeout, 'Test hangs with ZJIT')
-exclude(:test_nested_timeout, 'Test hangs with ZJIT')
-exclude(:test_nested_timeout_error_identity, 'Test hangs with ZJIT')
```
k0kubun: This is also failing the CI.

test/.excludes-zjit/TestThread.rb
```diff
@@ -1,2 +0,0 @@
-exclude(:test_switch_while_busy_loop, 'Test hangs with ZJIT')
```
k0kubun, line 1: This failed too https://github.com/ruby/ruby/actions/runs/16733873773/job/47368171297?pr=14097. 

When you file a PR that enables more test-all test cases next time, could you consider marking the PR ready for review after both test-all jobs (x86_64 and arm64) pass on the CI? It seems like the CI produces different results from your local environment.

st0012: Sorry, It's been removed and both are passing now.
In my last push before changing the PR status the CI [failed with 1 failure](https://github.com/ruby/ruby/actions/runs/16731180401/job/47359364529), so I naively assumed if I added it back it'd be passing 😕

k0kubun (approved):

test/.excludes-zjit/TestThread.rb
k0kubun: I think this should have been addressed. The last job on this PR passed it, but it didn't once, so it means it still randomly fails.

Since the CI right after the merge failed it https://github.com/ruby/ruby/actions/runs/16735509694/job/47373381258, I don't think we should enable this test yet. I'll revert this part of the PR.

k0kubun: disabled that test again https://github.com/ruby/ruby/commit/591849a1a565e2177bee82bee8d0118a7c48eaa9

k0kubun: The other test failed too after disabling it. Disabled that one too https://github.com/ruby/ruby/commit/196155812be09fdc4c36f8d7d6d1fbc69948765c.

st0012: Apology and thanks for adding them back 🙏

PR #14100, opened 2025-08-04
ZJIT: Add helpers to prepare for C calls

k0kubun:
This PR adds helpers similar to `jit_prepare_call_with_gc`/`jit_prepare_non_leaf_call` in YJIT. Note that we're renaming `jit_*` functions to `gen_*` in ZJIT because we thought they were used for the same purpose.

* Define `gen_prepare_call_with_gc`: It's just an alias to `gen_save_pc`, but it clarifies the purpose by the function name.
* Define `gen_prepare_non_leaf_call`: It saves the whole frame state.
* Use `gen_prepare_non_leaf_call` for `invokebuiltin` and `opt_getconstant_path`: They are not leaf.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -340,7 +340,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::SendWithoutBlockDirect { cd, state, self_val, args, .. } if args.len() + 1 > C_ARG_OPNDS.len() => // +1 for self
             gen_send_without_block(jit, asm, *cd, &function.frame_state(*state), opnd!(self_val), opnds!(args))?,
         Insn::SendWithoutBlockDirect { cme, iseq, self_val, args, state, .. } => gen_send_without_block_direct(cb, jit, asm, *cme, *iseq, opnd!(self_val), opnds!(args), &function.frame_state(*state))?,
-        Insn::InvokeBuiltin { bf, args, state, .. } => gen_invokebuiltin(asm, &function.frame_state(*state), bf, opnds!(args))?,
+        Insn::InvokeBuiltin { bf, args, state, .. } => gen_invokebuiltin(jit, asm, &function.frame_state(*state), bf, opnds!(args))?,
         Insn::Return { val } => return Some(gen_return(asm, opnd!(val))?),
         Insn::FixnumAdd { left, right, state } => gen_fixnum_add(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state))?,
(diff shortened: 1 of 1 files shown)


zjit/src/codegen.rs
```diff
@@ -490,25 +490,26 @@ fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, local_ep_offset: u32, le
...
+fn gen_invokebuiltin(jit: &JITState, asm: &mut Assembler, state: &FrameState, bf: &rb_builtin_function, args: Vec<Opnd>) -> Option<lir::Opnd> {
     // Ensure we have enough room fit ec, self, and arguments
     // TODO remove this check when we have stack args (we can use Time.new to test it)
     if bf.argc + 2 > (C_ARG_OPNDS.len() as i32) {
         return None;
     }
 
-    gen_save_pc(asm, state);
+    // Anything can happen inside builtin functions
+    gen_prepare_non_leaf_call(jit, asm, state)?;
```
st0012, line 512: I know the information is not passed in here, but do we want to use the annotated leaf attribute here in the future?
Or we'd mostly use it on the HIR level only?

zjit/src/codegen.rs
```diff
@@ -490,25 +490,26 @@ fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, local_ep_offset: u32, le
...
 
-fn gen_get_constant_path(asm: &mut Assembler, ic: *const iseq_inline_constant_cache, state: &FrameState) -> Opnd {
+fn gen_get_constant_path(jit: &JITState, asm: &mut Assembler, ic: *const iseq_inline_constant_cache, state: &FrameState) -> Option<Opnd> {
     unsafe extern "C" {
         fn rb_vm_opt_getconstant_path(ec: EcPtr, cfp: CfpPtr, ic: *const iseq_inline_constant_cache) -> VALUE;
     }
 
-    // Save PC since the call can allocate an IC
-    gen_save_pc(asm, state);
+    // Anything could be called on const_missing
```
st0012, line 498: Thanks for this comment. Does autoloading happen during this operation too?

st0012 (approved):

zjit/src/codegen.rs
k0kubun: Yeah, we'd need to use that information here. I don't have a strong preference on how the HIR should present it to codegen, but at least it's not worth adding a different `Insn` for it depending on whether it's leaf or not.

zjit/src/codegen.rs
k0kubun: It does. I kinda intended to imply that as one of "const" is "missing" situations, but I guess it's not technically accurate to associate const"_"missing to autoload. Anyway, mentioning const_missing should be enough to prove its non-leafness.

PR #14101, opened 2025-08-04
ZJIT: Implement `defined?` codegen for non-yield calls

st0012:
I found this when studying call preparations in YJIT and use [its implementation](https://github.com/ruby/ruby/blob/master/yjit/src/codegen.rs#L3320-L3346) as a reference.


zjit/src/hir.rs
```diff
@@ -1172,7 +1172,7 @@ impl Function {
             &ArrayDup { val, state } => ArrayDup { val: find!(val), state },
             &HashDup { val, state } => HashDup { val: find!(val), state },
             &CCall { cfun, ref args, name, return_type, elidable } => CCall { cfun, args: find_vec!(args), name, return_type, elidable },
-            &Defined { op_type, obj, pushval, v } => Defined { op_type, obj, pushval, v: find!(v) },
+            &Defined { op_type, obj, pushval, v, state } => Defined { op_type, obj, pushval, v: find!(v), state: find!(state) },
```
st0012, line 1176: I actually am not sure if I need to use `find!` for state here. How should I decide what value requires `find!` and what doesn't 🤔

st0012:
Changing this to draft because it panics on Ubuntu.

zjit/src/hir.rs
tekknolagi: Right now state is the odd one out. It "should" be find-ed ("found"?) but we will never actually rewrite it so it doesn't matter in practice

zjit/src/codegen.rs
```diff
@@ -455,7 +455,22 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, _obj: VALUE,
                 Some(Qnil.into())
             }
         }
-        _ => None
+        _ => {
+            // Skip on x86_64 due to index out of bounds on live ranges
```
tekknolagi: Please make this actionable; add a TODO

zjit/src/codegen.rs
```diff
@@ -455,7 +455,22 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, _obj: VALUE,
...
+            // May be related to https://github.com/Shopify/ruby/issues/646
+            if cfg!(target_arch = "x86_64") {
+                None
+            } else {
+                // Save the PC and SP because the callee may allocate or call #respond_to?
+                gen_prepare_non_leaf_call(jit, asm, state)?;
+
+                // Call vm_defined(ec, reg_cfp, op_type, obj, v)
+                let def_result = asm.ccall(rb_vm_defined as *const u8, vec![EC, CFP, op_type.into(), obj.into(), tested_value]);
```
tekknolagi: Is there a reason you are not using the ccall! macro?

zjit/src/codegen.rs
```diff
@@ -455,7 +455,22 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, _obj: VALUE,
...
+
+                asm.test(def_result, Opnd::UImm(255));
```
tekknolagi: Can you add a todo to split this up into the sub-cases? They are (I think) known at compile-time

tekknolagi: Why are you testing against 255? rb_vm_defined returns a bool

k0kubun:
> Skipping this codegen for x86_64 & omitting the newly added test would make the tests pass. But I'm not sure if we'd want to ship an implementation for just one architecture first instead of both. Thoughts?

This seems to add more problems (skipped tests and arch-specific codegen, which could require zjit-test-all to disable a different set of tests depending on the arch) than solving ones. We've deliberately avoided this kind of arch-specific situations by using the same number of registers (and the same size of code pages in YJIT) for both arches. We should be able to just fix https://github.com/Shopify/ruby/issues/646 first if it's blocking something else.

st0012:
Thanks for both of your input. I’ll convert this back to draft now.

zjit/src/codegen.rs
st0012: It can return `Qfalse` too from `rb_vm_defined`. But yeah `Qfalse` and `false` are both `0` so we can just do `asm.test(def_result, def_result)`

st0012: Added a comment and created https://github.com/Shopify/ruby/issues/679

XrXr (changes requested):
The `test reg, 0xff` is in fact load bearing. Since `bool` is 1 byte, ABI has it that it does not have to fill the entire register, so there might be junk in the upper bits for `false`.

https://github.com/ruby/ruby/blob/24d0b458cd6ccde1e052c1db15af120def349693/yjit/src/codegen.rs#L10399-L10402

tekknolagi:
Can we just do `testb` equivalent?

tekknolagi:
I feel like `test(a, a)` is not meaningfully visually different from `test(a, 1)` unless you assume that the assembler always generates the shortest instruction (which may not be the case)

zjit/src/codegen.rs
st0012: Let me correct myself: because `rb_vm_defined` returns `bool`, the potential `VALUE` object returned from `vm_defined` would be implicitly converted into `bool` too. So `rb_vm_defined` can only return `bool`.

zjit/src/codegen.rs
```diff
@@ -455,7 +455,20 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, _obj: VALUE,
...
+
+            // TODO: Inline the cases for each op_type
+            // Call vm_defined(ec, reg_cfp, op_type, obj, v)
+            let def_result = asm_ccall!(asm, rb_vm_defined, EC, CFP, op_type.into(), obj.into(), tested_value);
+
+            // Mask the bool result with 0xFF to check only the low byte
+            // C ABI only guarantees the low byte is valid for bool returns
+            // upper bits are undefined and may contain garbage
+            asm.test(def_result, Opnd::UImm(255));
+            Some(asm.csel_nz(pushval.into(), Qnil.into()))
```
k0kubun: > I feel like test(a, a) is not meaningfully visually different from test(a, 1) unless you assume that the assembler always generates the shortest instruction (which may not be the case)

`cmp(a, 0)` + `csel_ne` might be what Max wants. The x86 backend converts `cmp(a, 0)` followed by `csel_ne` into `test(a, a)` because it in fact generates a shorter instruction.

Combined with the `_Bool` problem Alan pointed out, we probably want to use:

```suggestion
            asm.cmp(def_result.with_num_bits(8).unwrap(), 0.into);
            Some(asm.csel_ne(pushval.into(), Qnil.into()))
```

k0kubun: sorry, I meant `with_num_bits`. let me update the comment

zjit/src/codegen.rs
k0kubun: This seems to have come from YJIT's `gen_defined()`, and from what Alan said on Slack, it might have been a carry over from the assembler when it was written in C.

I don't know why we didn't use `test(a, 1)` at the time, but `test(a, 255)` might have been a way to say we want to mask higher bits without having a proper API like `with_num_bits(8)`. `test(reg_64, 1_u64)` would mask higher bits too, so for `test`, I guess we don't have a reason to use `255` today.

zjit/src/codegen.rs
st0012: Ah thanks for the `with_num_bits` update. I was quite confused but now it makes sense.

PR #14102, opened 2025-08-04
ZJIT: Remove GC offsets overwritten by invalidation

k0kubun:
This PR removes GC offsets then they are overwritten on invalidation. This allows `require "bundler/setup"` to run successfully on yjit-bench. ~It also adds `rb_zjit_iseq_free` hook that frees the ZJIT payload and the invariants when an ISEQ is freed.~

P.S. I moved the location of `append_gc_offsets` so that `rb_zjit_iseq_mark` and `rb_zjit_iseq_update_references` are placed next to each other. I also wrote a test of `require "bundler/setup"` in test_zjit.rb, but I didn't add it since I couldn't make it fail on the test-all harness with the master branch.

### DIFF
--- a/zjit/src/gc.rs
+++ b/zjit/src/gc.rs
@@ -1,6 +1,6 @@
 // This module is responsible for marking/moving objects on GC.
 
-use std::ffi::c_void;
+use std::{ffi::c_void, ops::Range};
 use crate::{cruby::*, profile::IseqProfile, state::ZJITState, virtualmem::CodePtr};
 
 /// This is all the data ZJIT stores on an ISEQ. We mark objects in this struct on GC.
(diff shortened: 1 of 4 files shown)


zjit/src/gc.rs
```diff
@@ -150,3 +133,47 @@ pub extern "C" fn rb_zjit_iseq_update_references(payload: *mut c_void) {
...
+        let value_ptr = value_ptr as *const VALUE;
+        unsafe {
+            let object = value_ptr.read_unaligned();
+            rb_gc_writebarrier(iseq.into(), object);
+        }
+    }
+}
+
+/// Remove GC offsets that overlap with a given removed_range
+pub fn remove_gc_offsets(iseq: IseqPtr, removed_range: &Range<CodePtr>) {
```
tekknolagi: Can we add a doc comment either here or at call-sites or both to explain why? It's very subtle

k0kubun: added a comment https://github.com/ruby/ruby/pull/14102/commits/61266a56f21e79152a2dc8d323de346af20324a4

k0kubun:
not merging this yet since I'm debugging a random assertion failure on this branch, e.g. https://github.com/ruby/ruby/actions/runs/16737690453/job/47380192499.

k0kubun:
I resolved a git conflict and updated the patch to use an `IseqPayload` pointer instead of `IseqPtr` as a fix to the random crash by compaction. It seems more efficient and maintainable to work on the payload than maintaining `IseqPtr`, which moves, and getting the same thing out of it.

I also reduced the scope of this PR to only fix the problem, and leave it for another PR to implement a missing `rb_zjit_iseq_free` https://github.com/Shopify/ruby/issues/682, which is an existing problem that exists separately from the scope of this issue.

PR #14107, opened 2025-08-05
variable.c: refactor accesses to the `generic_fields_tbl`

casperisfine:
It's now mostly all encapsulated inside:

  - `rb_obj_fields`
  - `rb_obj_set_fields`
  - `rb_obj_replace_fields`

### DIFF
--- a/vm_insnhelper.c
+++ b/vm_insnhelper.c
@@ -1267,8 +1267,8 @@ vm_getivar(VALUE obj, ID id, const rb_iseq_t *iseq, IVC ic, const struct rb_call
         }
       default:
         if (rb_obj_exivar_p(obj)) {
-            VALUE fields_obj = 0;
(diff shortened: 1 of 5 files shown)


vm_insnhelper.c
```diff
@@ -1484,14 +1480,15 @@ vm_setivar_default(VALUE obj, ID id, VALUE val, shape_id_t dest_shape_id, attr_i
         return Qundef;
     }
 
-    rb_gen_fields_tbl_get(obj, 0, &fields_obj);
+    VALUE fields_obj = rb_obj_fields(obj, id);
+    RUBY_ASSERT(fields_obj);
+    RB_OBJ_WRITE(obj, &rb_imemo_fields_ptr(fields_obj)[index], val);
```
jhawthorn: Is this write to `obj`, or is it a write to `fields_obj`?

casperisfine: Good catch. That's a preexisting bug :/

PR #14116, opened 2025-08-05
Array#fill doc fix to indicate return is self

notEthan:
doc currently indicates the return value as `new_array` but then in the first sentence explains "always returns +self+ (never a new array)".


notEthan:
I decided to check other `new_array` return values and found `map!`/`collect!` also has this; added that one as well.

nobu:
Could you add `[DOC]` label at each commit title?
Abbreviated `doc` alone cannot trigger the documentation preview.

notEthan:
okay, pushed

PR #14117, opened 2025-08-05
ZJIT: Add a graphviz dumper for HIR

tekknolagi:
This is moderately useful just in stdout (copy and paste into a renderer) but potentially more useful alongside a tool that parses stdout looking for `digraph G { ... }` and renders those automatically.


tekknolagi:
Right now this is optimized for web: it does HTML encoding automatically. Should we avoid that? Make it an option like HIR=raw/all/...?

tekknolagi:
@tenderlove does this meet the needs you had when you started prototyping graphviz output?

k0kubun:
> Should we avoid that? Make it an option like HIR=raw/all/...?

If HTML is what you use, I don't think we need to support anything else.

PR #14120, opened 2025-08-05
ZJIT: Fix line numbers in HIR tests

tekknolagi:
This is caused by #14093 I think.


tekknolagi:
(merge at will)

st0012 (approved):

k0kubun:
This seems too fragile to maintain. Similarly to addresses, could we avoid matching the actual line numbers? HIR tests shouldn't fail just because documentation or new methods are added to gc.rb or dir.rb.

st0012:
I opened #14124 to remove line numbers from test expectations

k0kubun:
We've merged https://github.com/ruby/ruby/pull/14124 instead, so closing this one.

PR #14121, opened 2025-08-05
ZJIT: Implement SingleRactorMode invalidation

st0012:
Additional changes:
- To make `ractor.c` (or any files that don't include `insns.inc`) able to include `zjit.h`, `zjit.h` can't use the unexported `enum ruby_vminsn_type` in it. So I replaced it with `int` in `zjit.h`
- I added `compile_jumps!` macro to reduce the jump compilation boilerplate
- Fixed 2 typos

### DIFF
--- a/zjit.h
+++ b/zjit.h
@@ -33,6 +33,6 @@ static inline void rb_zjit_cme_invalidate(const rb_callable_method_entry_t *cme)
 static inline void rb_zjit_invalidate_ep_is_bp(const rb_iseq_t *iseq) {}
 static inline void rb_zjit_constant_state_changed(ID id) {}
 static inline void rb_zjit_before_ractor_spawn(void) {}
-#endif // #if USE_YJIT
+#endif // #if USE_ZJIT
 
 #endif // #ifndef ZJIT_H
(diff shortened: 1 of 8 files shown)


zjit.h
```diff
@@ -14,23 +14,25 @@ extern bool rb_zjit_enabled_p;
 extern uint64_t rb_zjit_call_threshold;
 extern uint64_t rb_zjit_profile_threshold;
 void rb_zjit_compile_iseq(const rb_iseq_t *iseq, rb_execution_context_t *ec, bool jit_exception);
-void rb_zjit_profile_insn(enum ruby_vminsn_type insn, rb_execution_context_t *ec);
+void rb_zjit_profile_insn(int insn, rb_execution_context_t *ec);
```
k0kubun: This seems to effectively hold for our supported platforms, but it still seems implementation-dependent in C in general. If you want to do this, can you update the signature of `rb_zjit_profile_insn` accordingly and explicitly cast it inside the Rust function?

> Each enumerated type shall be compatible with char, a signed integer type, or an unsigned integer type. The choice of type is implementation-defined, but shall be capable of representing the values of all the members of the enumeration.

st0012: Updated 👍

tekknolagi:
Can you please add some tests?

st0012:
@tekknolagi Added a test that'd crash without the invalidation

k0kubun (approved):

PR #14124, opened 2025-08-05
ZJIT: Avoid matching HIR line numbers in tests

st0012:
This should avoid issues mentioned in #14120


zjit/src/hir.rs
```diff
@@ -2241,7 +2241,18 @@ impl Function {
...
+            let full_location = iseq_get_location(fun.iseq, 0);
+
+            if let Some(colon_pos) = full_location.rfind(':') {
+                full_location[..colon_pos].to_string()
+            } else {
+                full_location
+            }
+        } else {
+            iseq_get_location(fun.iseq, 0)
+        };
```
k0kubun: How about limiting the workaround only to builtin ISEQs? We should test the actual behavior when possible. There's no need to hide a stable line number at the beginning of an eval owned by ZJIT tests.

```suggestion
        let iseq_name = iseq_get_location(fun.iseq, 0);
        // In tests, strip the line number for builtin ISEQs to make tests stable across line changes
        let iseq_name = if cfg!(test) && iseq_name.contains("@<internal:") {
            iseq_name[..iseq_name.rfind(':').unwrap()].to_string()
        } else {
            iseq_name
        };
```

st0012: 👍 Updated

PR #14125, opened 2025-08-06
ZJIT: Fix "immediate value too large" on cmp for x86_64

k0kubun:
This PR fixes an "immediate value too large" error when encoding `cmp` for x86_64.

When the rhs is `-1` (`UImm(0xffffffffffffffff)`), it doesn't fit in 32 bits. So it needs to be loaded into a register first.

### DIFF
--- a/zjit/src/backend/x86_64/mod.rs
+++ b/zjit/src/backend/x86_64/mod.rs
@@ -377,7 +377,7 @@ impl Assembler
                 },
                 Opnd::UImm(value) => {
                     // 32-bit values will be sign-extended
-                    if imm_num_bits(*value as i64) > 32 {
+                    if uimm_num_bits(*value) > 32 {
                         mov(cb, Assembler::SCRATCH0, opnd.into());
                         Assembler::SCRATCH0
                     } else {
@@ -963,7 +963,9 @@ mod tests {
         asm.cmp(Opnd::Reg(RAX_REG), Opnd::UImm(0xFF));
         asm.compile_with_num_regs(&mut cb, 0);
 
-        assert_eq!(format!("{:x}", cb), "4881f8ff000000");
+        assert_disasm!(cb, "4881f8ff000000", "
+            0x0: cmp rax, 0xff
+        ");
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


zjit/src/backend/x86_64/mod.rs
```diff
@@ -377,7 +377,7 @@ impl Assembler
                 },
                 Opnd::UImm(value) => {
                     // 32-bit values will be sign-extended
```
tekknolagi: Is this sign-extension thing still true?

tekknolagi (approved):

zjit/src/backend/x86_64/mod.rs
XrXr: This is still valid. The immediate to CMP is always sign extended when the destination is bigger.

zjit/src/backend/x86_64/mod.rs
```diff
@@ -377,7 +377,7 @@ impl Assembler
...
-                    if imm_num_bits(*value as i64) > 32 {
+                    if uimm_num_bits(*value) > 32 {
                         mov(cb, Assembler::SCRATCH0, opnd.into());
                         Assembler::SCRATCH0
                     } else {
```
XrXr, line 383: It's possible to directly encode `cmp rax, -1` (Shortest form seems to be `48 83 f8 ff` with imm8, but the form that takes imm32 also works). It might just work if we keep the old check and put `imm_opnd(*value as i64)` in the else branch.

If that doesn't work for some reason this is still a good bugfix and we can revisit this later. 

EDIT: See also https://github.com/Shopify/ruby/commit/78d60df322fde38b20baa063f3c000f202213f7e

k0kubun: Thanks, I took your patch https://github.com/ruby/ruby/pull/14125/commits/dfa8635f57c2e3e79360ac52c81d8d1fa05d4ec6.

It appears that `test` does not support short encoding for `Imm` unlike `UImm`, so it ended up making a test case for `test` generate a bigger code. But it does work, so I'll have a look at it in a separate PR.

PR #14126, opened 2025-08-06
ZJIT: Inline attr_reader/attr_accessor

tekknolagi:
We can rewrite SendWithoutBlock to GetIvar.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -879,6 +879,38 @@ class TestZJIT < Test::Unit::TestCase
     }
   end
 
+  def test_attr_reader
+    assert_compiles '[4, 4]', %q{
+      class C
+        attr_reader :foo
+
+        def initialize
+          @foo = 4
+        end
+      end
+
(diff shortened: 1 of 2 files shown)


k0kubun:
It's not something added by this PR, but I think we should call `gen_prepare_non_leaf_call` before `rb_ivar_get` on `gen_getivar`, similarly to YJIT. YJIT added that because it could raise RactorIsolationError.

https://github.com/ruby/ruby/blob/e378a21a32bb0ca26f206f975073760a4e1a8bef/variable.c#L1380-L1386

k0kubun (approved):
The diff looks good though. The interpreter doesn't push a frame for `VM_METHOD_TYPE_IVAR` either, so the unleafness shouldn't prevent us from inlining the call. (We do need `gen_prepare_non_leaf_call` though)

tekknolagi:
Can we patchpoint single ractor mode instead? That way we don't need to push anything to the stack

tekknolagi:
https://github.com/Shopify/ruby/issues/677

PR #14127, opened 2025-08-06
ZJIT: x86: split: Fix live ranges index-out-of-range panic

XrXr:
Combined with the fix from GH-14125, we now pass TestFixnum on x86.
Just need to be careful when landing.

- **ZJIT: x86: split: Query new assembler for live ranges**
- **ZJIT: Run TestFixnum**
- **[drop later] fix for "immediate value too large"**

### DIFF
--- a/zjit/src/backend/x86_64/mod.rs
+++ b/zjit/src/backend/x86_64/mod.rs
@@ -148,6 +148,15 @@ impl Assembler
                 };
             }
 
+            // When we split an operand, we can create a new VReg not in `live_ranges`.
+            // So when we see a VReg with out-of-range index, it's created from splitting
+            // from the loop above and we know it doesn't outlive the current instruction.
+            let vreg_outlives_insn = |vreg_idx| {
+                live_ranges
+                    .get(vreg_idx)
+                    .map_or(false, |live_range: &LiveRange| live_range.end() > index)
+            };
+
             // We are replacing instructions here so we know they are already
             // being used. It is okay not to use their output here.
             #[allow(unused_must_use)]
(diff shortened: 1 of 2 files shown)


zjit/src/backend/x86_64/mod.rs
```diff
@@ -128,6 +128,17 @@ impl Assembler
         let mut iterator = self.insns.into_iter().enumerate().peekable();
         let mut asm = Assembler::new_with_label_names(take(&mut self.label_names), live_ranges.len());
 
+        // When we add news instructions to `asm`, we also create new VRegs not in the old
+        // assembler. All VRegs with index larger than the old `live_ranges` live in the
+        // new assembler since all old VRegs are inherited and keep their index.
+        fn vreg_outlives_insn(vreg_idx: usize, older_live_ranges: &Vec<LiveRange>, new_asm: &Assembler, insn_idx: usize) -> bool {
+            older_live_ranges
+                .get(vreg_idx)
+                .map_or_else(|| new_asm.live_ranges.get(vreg_idx), Some)
```
k0kubun: Hmm. I feel like `vreg_outlives_insn()` should return `false` without checking its live range against `insn_idx` when `vreg_idx` is not found in `older_live_ranges`. `insn_idx` is an index to instructions in the old Assembler, so querying live ranges for the new Assembler with shifted indexes with `insn_idx` doesn't seem correct.

Looking at what `x86_split` is doing, such new VRegs seem to come only from the lowering of `Opnd::Value`. That reference to `Opnd::Value` is not shared across multiple instructions, so IIUC it should never outlive the insn. However, because the index for the old Assembler is queried against the indexes in the new Assembler, when there are too many past instructions merged into fewer instructions, it could accidentally return true.

XrXr: Hmm yeah, I was trying to avoid specific property of split because that feels too arcane, but comparing across index system is definitely incorrect.

k0kubun (approved):
:+1:

st0012 (approved):

PR #14129, opened 2025-08-06
Struct: keep direct reference to IMEMO/fields when space allows

casperisfine:
Improved version of https://github.com/ruby/ruby/pull/14095, enabled by the refactoring in https://github.com/ruby/ruby/pull/14107

It's not rare for structs to have additional ivars, hence are one of the most common, if not the most common type in the `gen_fields_tbl`.

This can cause Ractor contention, but even in single ractor mode means having to do a hash lookup to access the ivars, and increase GC work.

Instead, unless the struct is perfectly right sized, we can store a reference to the associated IMEMO/fields object right after the last struct member.

```
compare-ruby: ruby 3.5.0dev (2025-08-06T12:50:36Z struct-ivar-fields-2 9a30d141a1) +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-06T12:57:59Z struct-ivar-fields-2 2ff3ec237f) +PRISM [arm64-darwin24]
warming up.....

|                      |compare-ruby|built-ruby|
|:---------------------|-----------:|---------:|
|member_reader         |    590.317k|  579.246k|
|                      |       1.02x|         -|
|member_writer         |    543.963k|  527.104k|
|                      |       1.03x|         -|
|member_reader_method  |    213.540k|  213.004k|
|                      |       1.00x|         -|
|member_writer_method  |    192.657k|  191.491k|
|                      |       1.01x|         -|
|ivar_reader           |    403.993k|  569.915k|
|                      |           -|     1.41x|
```

cc @etiennebarrie

### DIFF
--- a/internal/struct.h
+++ b/internal/struct.h
@@ -11,10 +11,23 @@
 #include "ruby/internal/stdbool.h"     /* for bool */
 #include "ruby/ruby.h"          /* for struct RBasic */
 
+/* Flags of RStruct
+ *
+ * 1-7: RSTRUCT_EMBED_LEN
+ *          If non-zero, the struct is embedded (its contents follow the
+ *          header, rather than being on a separately allocated buffer) and
+ *          these bits are the length of the Struct.
+ * 8:   RSTRUCT_GEN_FIELDS
+ *          The struct is embedded and has no space left to store the
+ *          IMEMO/fields reference. Any ivar this struct may have will be in
+ *          the generic_fields_tbl. This flag doesn't imply the struct has
+ *          ivars.
+ */
 enum {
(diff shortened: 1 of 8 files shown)


internal/struct.h
```diff
@@ -11,10 +11,23 @@
 #include "ruby/internal/stdbool.h"     /* for bool */
 #include "ruby/ruby.h"          /* for struct RBasic */
 
+/* Flags of RStruct
+ *
+ * 1-7: RSTRUCT_EMBED_LEN
+ *          If non-zero, the struct is embedded (its contents follow the
+ *          header, rather than being on a separately allocated buffer) and
+ *          these bits are the length of the Struct.
+ * 8:   RSTRUCT_GEN_FIELDS
```
casperisfine, line 20: Another option would be to evacuate the struct when the first ivar is assigned. It's a bit more complex code, but would be more localized.

peterzhu2118 (approved):
Makes sense to me

PR #14133, opened 2025-08-06
Introduce Time#am? and Time#pm?

MatheusRich:


nobu:
It would be able to implement in timev.rb.

MatheusRich:
@nobu I can do that. Are there any differences/advantages in doing so? Just curious, so I can apply this on future pull requests.

nobu:
> @nobu I can do that. Are there any differences/advantages in doing so? Just curious, so I can apply this on future pull requests.

No big difference in this case.
It is just **possible**.

It is said simpler and that JIT will be able to optimize such methods.
The current implementations look simple enough though.

One more difference is documentation.
C source code must be ASCII only, and comments can’t contain `*/`.
Neither of them applies here.

nobu:
[[Feature #21533]](https://bugs.ruby-lang.org/issues/21533) was rejected.

PR #14134, opened 2025-08-06
RTypedData: keep direct reference to IMEMO/fields

byroot:
Similar to f3206cc79bec2fd852e81ec56de59f0a67ab32b7 but for TypedData.

It's quite common for TypedData objects to have a mix of reference in their struct and some ivars.
Since we do happen to have 8B free in the RtypedData struct, we could use it to keep a direct reference to the IMEMO/fields saving having to synchronize the VM and lookup the `gen_fields_tbl` on every ivar access.

For old school Data classes however, we don't have free space, but this API is soft-deprecated and no longer very common.

cc @tenderlove as mentioned.

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1231,13 +1231,20 @@ rb_obj_fields(VALUE obj, ID field_name)
     VALUE fields_obj = 0;
     if (rb_shape_obj_has_fields(obj)) {
         switch (BUILTIN_TYPE(obj)) {
+          case T_DATA:
+            if (LIKELY(RTYPEDDATA_P(obj))) {
+                fields_obj = RTYPEDDATA(obj)->fields_obj;
+                break;
+            }
+            goto generic_fields;
           case T_STRUCT:
             if (LIKELY(!FL_TEST_RAW(obj, RSTRUCT_GEN_FIELDS))) {
                 fields_obj = RSTRUCT_FIELDS_OBJ(obj);
                 break;
             }
-            // fall through
+            goto generic_fields;
           default:
+          generic_fields:
             RB_VM_LOCKING() {
                 if (!st_lookup(generic_fields_tbl_, (st_data_t)obj, (st_data_t *)&fields_obj)) {
                     rb_bug("Object is missing entry in generic_fields_tbl");
@@ -1254,13 +1261,20 @@ rb_free_generic_ivar(VALUE obj)
     if (rb_obj_exivar_p(obj)) {
         st_data_t key = (st_data_t)obj, value;
         switch (BUILTIN_TYPE(obj)) {
+          case T_DATA:
+            if (LIKELY(RTYPEDDATA_P(obj))) {
+                RB_OBJ_WRITE(obj, &RTYPEDDATA(obj)->fields_obj, 0);
+                break;
+            }
+            goto generic_fields;
           case T_STRUCT:
             if (LIKELY(!FL_TEST_RAW(obj, RSTRUCT_GEN_FIELDS))) {
                 RSTRUCT_SET_FIELDS_OBJ(obj, 0);
                 break;
             }
-            // fall through
+            goto generic_fields;
           default:
+          generic_fields:
             RB_VM_LOCKING() {
                 st_delete(generic_fields_tbl_no_ractor_check(), &key, &value);
             }
@@ -1279,13 +1293,20 @@ rb_obj_set_fields(VALUE obj, VALUE fields_obj, ID field_name, VALUE original_fie
 
     if (fields_obj != original_fields_obj) {
         switch (BUILTIN_TYPE(obj)) {
+          case T_DATA:
+            if (LIKELY(RTYPEDDATA_P(obj))) {
+                RB_OBJ_WRITE(obj, &RTYPEDDATA(obj)->fields_obj, fields_obj);
+                break;
+            }
+            goto generic_fields;
           case T_STRUCT:
             if (LIKELY(!FL_TEST_RAW(obj, RSTRUCT_GEN_FIELDS))) {
                 RSTRUCT_SET_FIELDS_OBJ(obj, fields_obj);
                 break;
             }
-            // fall through
+            goto generic_fields;
           default:
+          generic_fields:
             RB_VM_LOCKING() {
                 st_insert(generic_fields_tbl_, (st_data_t)obj, (st_data_t)fields_obj);
             }
--- a/gc.c
+++ b/gc.c
@@ -1047,7 +1047,7 @@ rb_data_object_wrap(VALUE klass, void *datap, RUBY_DATA_FUNC dmark, RUBY_DATA_FU
 {
     RUBY_ASSERT_ALWAYS(dfree != (RUBY_DATA_FUNC)1);
     if (klass) rb_data_object_check(klass);
-    return newobj_of(GET_RACTOR(), klass, T_DATA, (VALUE)dmark, (VALUE)datap, (VALUE)dfree, !dmark, sizeof(struct RTypedData));
+    return newobj_of(GET_RACTOR(), klass, T_DATA, (VALUE)dmark, (VALUE)dfree, (VALUE)datap, !dmark, sizeof(struct RTypedData));
 }
 
 VALUE
@@ -1064,7 +1064,7 @@ typed_data_alloc(VALUE klass, VALUE typed_flag, void *datap, const rb_data_type_
     RBIMPL_NONNULL_ARG(type);
     if (klass) rb_data_object_check(klass);
     bool wb_protected = (type->flags & RUBY_FL_WB_PROTECTED) || !type->function.dmark;
-    return newobj_of(GET_RACTOR(), klass, T_DATA, ((VALUE)type) | IS_TYPED_DATA | typed_flag, (VALUE)datap, 0, wb_protected, size);
+    return newobj_of(GET_RACTOR(), klass, T_DATA, 0, ((VALUE)type) | IS_TYPED_DATA | typed_flag, (VALUE)datap, wb_protected, size);
 }
 
 VALUE
@@ -3173,10 +3173,15 @@ rb_gc_mark_children(void *objspace, VALUE obj)
         break;
 
       case T_DATA: {
-        void *const ptr = RTYPEDDATA_P(obj) ? RTYPEDDATA_GET_DATA(obj) : DATA_PTR(obj);
+        bool typed_data = RTYPEDDATA_P(obj);
+        void *const ptr = typed_data ? RTYPEDDATA_GET_DATA(obj) : DATA_PTR(obj);
+
+        if (typed_data) {
+            gc_mark_internal(RTYPEDDATA(obj)->fields_obj);
+        }
 
         if (ptr) {
-            if (RTYPEDDATA_P(obj) && gc_declarative_marking_p(RTYPEDDATA_TYPE(obj))) {
+            if (typed_data && gc_declarative_marking_p(RTYPEDDATA_TYPE(obj))) {
                 size_t *offset_list = TYPED_DATA_REFS_OFFSET_LIST(obj);
 
                 for (size_t offset = *offset_list; offset != RUBY_REF_END; offset = *offset_list++) {
@@ -3184,7 +3189,7 @@ rb_gc_mark_children(void *objspace, VALUE obj)
                 }
             }
             else {
-                RUBY_DATA_FUNC mark_func = RTYPEDDATA_P(obj) ?
+                RUBY_DATA_FUNC mark_func = typed_data ?
                     RTYPEDDATA_TYPE(obj)->function.dmark :
                     RDATA(obj)->dmark;
                 if (mark_func) (*mark_func)(ptr);
@@ -4121,9 +4126,15 @@ rb_gc_update_object_references(void *objspace, VALUE obj)
       case T_DATA:
         /* Call the compaction callback, if it exists */
         {
-            void *const ptr = RTYPEDDATA_P(obj) ? RTYPEDDATA_GET_DATA(obj) : DATA_PTR(obj);
+            bool typed_data = RTYPEDDATA_P(obj);
+            void *const ptr = typed_data ? RTYPEDDATA_GET_DATA(obj) : DATA_PTR(obj);
+
+            if (typed_data) {
+                UPDATE_IF_MOVED(objspace, RTYPEDDATA(obj)->fields_obj);
+            }
+
             if (ptr) {
-                if (RTYPEDDATA_P(obj) && gc_declarative_marking_p(RTYPEDDATA_TYPE(obj))) {
+                if (typed_data && gc_declarative_marking_p(RTYPEDDATA_TYPE(obj))) {
                     size_t *offset_list = TYPED_DATA_REFS_OFFSET_LIST(obj);
 
                     for (size_t offset = *offset_list; offset != RUBY_REF_END; offset = *offset_list++) {
@@ -4131,7 +4142,7 @@ rb_gc_update_object_references(void *objspace, VALUE obj)
                         *ref = gc_location_internal(objspace, *ref);
                     }
                 }
-                else if (RTYPEDDATA_P(obj)) {
+                else if (typed_data) {
                     RUBY_DATA_FUNC compact_func = RTYPEDDATA_TYPE(obj)->function.dcompact;
                     if (compact_func) (*compact_func)(ptr);
                 }
--- a/include/ruby/internal/abi.h
+++ b/include/ruby/internal/abi.h
@@ -24,7 +24,7 @@
  * In released versions of Ruby, this number is not defined since teeny
  * versions of Ruby should guarantee ABI compatibility.
  */
-#define RUBY_ABI_VERSION 2
+#define RUBY_ABI_VERSION 3
 
 /* Windows does not support weak symbols so ruby_abi_version will not exist
  * in the shared library. */
--- a/include/ruby/internal/core/rdata.h
+++ b/include/ruby/internal/core/rdata.h
@@ -133,12 +133,6 @@ struct RData {
      */
     RUBY_DATA_FUNC dmark;
 
-    /** Pointer to the actual C level struct that you want to wrap.
-      * This is in between dmark and dfree to allow DATA_PTR to continue
-      * to work for both RData and non-embedded RTypedData.
-      */
-    void *data;
-
     /**
      * This function is called when the object  is no longer used.  You need to
      * do whatever necessary to avoid memory leaks.
@@ -147,6 +141,12 @@ struct RData {
      *           impossible at that moment (that is why GC runs).
      */
     RUBY_DATA_FUNC dfree;
+
+    /** Pointer to the actual C level struct that you want to wrap.
+      * This is in between dmark and dfree to allow DATA_PTR to continue
+      * to work for both RData and non-embedded RTypedData.
+      */
+    void *data;
 };
 
 RBIMPL_SYMBOL_EXPORT_BEGIN()
--- a/include/ruby/internal/core/rtypeddata.h
+++ b/include/ruby/internal/core/rtypeddata.h
@@ -355,6 +355,9 @@ struct RTypedData {
     /** The part that all ruby objects have in common. */
     struct RBasic basic;
 
+    /** Direct reference to the slots that holds instance variables, if any **/
+    VALUE fields_obj;
+
     /**
      * This is a `const rb_data_type_t *const` value, with the low bits set:
      *


jeremyevans:
I think this may cause issues for embedded RTypedData, as the extra space is used for the embedded data. Also, a lot of code depends on `DATA_PTR(obj)` working both when `obj` is RData and non-embedded RTypedData, so moving fields around is going to be problematic. I found this out when working on ce51ef30df5bf07ec3881a377f0011b8f20ec507.

byroot:
> I think this may cause issues for embedded RTypedData, as the extra space is used for the embedded data.

It's a possibility, and I admit I didn't go over them all to verify, but the overwhelming majority of them likely have the extra space and won't be bumped to the next slot. e.g.  `Set` has `8B` free.

> so moving fields around is going to be problematic.

Yes, my intent is for them to keep being at the same offset, seems like I got it wrong, but it's the plan that they stay that way.

jeremyevans:
> It's a possibility, and I admit I didn't go over them all to verify, but the overwhelming majority of them likely have the extra space and won't be bumped to the next slot. e.g. `Set` has `8B` free.

I wonder how Set ended up with 8B free. It was right at the limit when ce51ef30df5bf07ec3881a377f0011b8f20ec507 was committed. If it already had 8B free, I wouldn't have needed to implement ce51ef30df5bf07ec3881a377f0011b8f20ec507 to free up 8B so it would fit in an 80B slot.

byroot:
> I wonder how Set ended up with 8B free. I

My bad. Seems like I miscounted.

byroot:
> my intent is for them to keep being at the same offset, seems like I got it wrong

Note to self. The offset still match AFAICT. The crash is with T_ZOMBIE, there might be a mismatch there.

jeremyevans:
For the non-embedded RTypedData and for embedded RTypedData with 8B free, this seems a clear win as the space is unused. For the embedded RTypedData with less than 8B free, this doubles the amount of space the objects take. At that point, it becomes a memory vs CPU tradeoff.  You could handle that case by using one of the flag bits in the pointer, so it can be an per-object/type decision:

```c
// in include/ruby/internal/core/rtypeddata.h
#define TYPED_DATA_EMBEDDED_FIELDS ((VALUE)4)
#define TYPED_DATA_PTR_FLAGS ((VALUE)7)
```

I think that's the last free flag bit on 32-bit arch. Alternatively, you could make a decision based on size of the embedded object (assuming you can get that easily), and only embed if the object has more than 8B free in the object slot. This would also increase complexity, because you would need to handle the slower fallback case. I assume the fallback code is the current code, so this is would involve moving the new code to an if branch, and keeping the current code in an else branch.  

Assuming you don't want to go that route due to the complexity it would entail, how many RTypedData actually use ivars? You mention that it is "quite common for TypedData", but I don't know if that is 50%, 5%, or something else. Is it possible to include a benchmark that shows the CPU speedup from this approach, so we can compare it to the potential doubling of space for certain objects when making a decision regarding the tradeoff?

byroot:
> how many RTypedData actually use ivars? 

It's all gonna depend on what gems you run. But even in Ruby-core, there are a bunch of typed data that use ivars, e.g. `Enumerator`. Then in gems, you have quite a bunch in OpenSSL, etc. 

Beyond how much it actually is used, I like the idea of not penalizing the pattern of doing a few core functionality in C, and then extending the class from Ruby like we do in Trilogy: https://github.com/trilogy-libraries/trilogy/blob/029db7d874b9d09d3403e4c2418802f5d54e2c4c/contrib/ruby/lib/trilogy.rb

> Is it possible to include a benchmark that shows the CPU speedup from this approach,

I included a benchmark in my previous PR where I did something similar for `T_STRUCT`, it's quite significant: https://github.com/ruby/ruby/commit/f3206cc79bec2fd852e81ec56de59f0a67ab32b7, and that's a single threaded benchmark. If you start a Ractor, Ruby then need to acquire the VM lock to access the object's ivars.

casperisfine:
Ok, so I figured what the bug was, I forgot to update `rb_data_object_wrap(` to account for the fields reordering.

Now for the size. The types that would be bumped:

  - `Set` as discussed (I didn't notice because I was checking in debug mode, and in debug mode size 2 is 88B...)
  - `Process::Status`. Used to fit in 40B, because the struct is 16B.
  - `frame_info` AKA `Thread::Backtrace::Location`.  Also used to be `40B`, would now be `80B`.

There's also `pinned_list`, but this one is of dynamic size, so it's not really the same.

Out of all of those, `Process::Status` and `frame_info` tend to be very temporary objects, and it's rare to allocate a ton of them, so to me it's a rounding error.

`Set` is a bit more annoying, and I wish it would still fit in 80B, but in 3.4 it was `200B` (`40B T_OBJECT` + `160B T_HASH`), so it's not like it's really a regression.

Of course as you mentioned there's always the possibility of having yet another flag to make this a conditional, but it's very tricky as it means different offsets based on the flags, so would be extremely finicky to get it working.

casperisfine:
Since `Set` seem to be the contentious point, another option is to shrink `set_table`, e.g.: https://github.com/ruby/ruby/compare/master...Shopify:ruby:set-entries-bounds

Would that satisfy you?

jeremyevans:
> Since `Set` seem to be the contentious point, another option is to shrink `set_table`, e.g.: [master...Shopify:ruby:set-entries-bounds](https://github.com/ruby/ruby/compare/master...Shopify:ruby:set-entries-bounds)
> 
> Would that satisfy you?

I should be clear that I don't object to the change in this PR, even if it increases the size of Set. My comments were just discussing the potential impact on memory and whether issue could be mitigated. In general, you don't need Set for small sets (Array generally works for those), so the size of small Set is not critical.

That being said, assuming that the Set change doesn't significantly impact the performance of Set, it would be great to keep Set in an 80B slot. However, if that change works for Set, couldn't it also work for Hash and allow Hash to fit in 80B (if you decrease the ar table size from 8 to 3)?

byroot:
> allow Hash to fit in 80B (if you decrease the ar table size from 8 to 3)?

Interesting. I think you might be right that it would be possible, but then there's even more tradeoffs to consider.

byroot:
It's a bit of an arbitrary application, but I decided to run the [YJIT bench shipit benchmark](https://github.com/Shopify/yjit-bench/tree/250b5a00bfc0cbaf80ef25329ccd3ed8546a0d3e/benchmarks/shipit) with a modified Ruby to record how often generic ivars are accessed:

```ruby
[["VM/thread", 4886971],
 ["T_STRING", 248114],
 ["T_HASH", 235517],
 ["SQLite3::Backup", 127320],
 ["xmlDoc", 24300],
 ["T_ARRAY", 9212],
 ["OpenSSL/Cipher", 4200],
 ["xmlNodeSet", 2700],
 ["xmlNode", 2700],
 ["T_SYMBOL", 533],
 ["time", 398],
 ["encoding", 363],
 ["T_REGEXP", 213],
 ["proc", 136],
 ["T_STRUCT", 74],
 ["set", 32],
 ["FFI::Type::Builtin", 21],
 ["FFI::StructField", 15],
 ["zstream", 8],
 ["OpenSSL/X509/STORE", 4],
 ["T_FLOAT", 4],
 ["Psych/parser", 3],
 ["msgpack:buffer_view", 2],
 ["FFI::Type::Mapped", 1],
 ["T_RATIONAL", 1],
 ["FFI::DynamicLibrary", 1],
 ["T_BIGNUM", 1],
]
```

Some do surprise me a bit (e.g. `T_FLOAT`, `T_SYMBOL`), but I suspect it's just some very generic codepath checking for ivars, and we should just bail out early for these types.

I'm quite surprised about `T_HASH` as well, I wasn't expecting it to be this common, I need to check where these comes from.

`VM/Thread` is most likely [`ActiveSupport::ExecutionState`](https://github.com/rails/rails/blob/bb3ddbf032c3a24c2c94f911c8c5ca9f6939c6d9/activesupport/lib/active_support/isolated_execution_state.rb)

Then there's a bunch of common C extensions (sqlite3, nokogiri, openssl).

byroot:
Actually, revised number with only object that do actually have ivars. Since trying to access ivars on an object that doesn't have any bails out fast:

```ruby
[["VM/thread", 4886969],
 ["T_HASH", 229501],
 ["SQLite3::Backup", 122531],
 ["T_STRING", 70597],
 ["xmlDoc", 23625],
 ["T_ARRAY", 9039],
 ["OpenSSL/Cipher", 2800],
 ["xmlNode", 2025],
 ["encoding", 358],
 ["time", 199],
 ["proc", 68],
 ["T_STRUCT", 38],
 ["OpenSSL/X509/STORE", 3],
 ["Psych/parser", 2],
 ["set", 1],
```

It makes much more sense.

byroot:
So for `T_HASH` it seems to mostly come from Rack's `HeaderHash`, and a bit from `ActiveSupport::OrderedOptions`.

For `T_STRING` without even checking I'm sure it's `String#html_safe`, but I took care of that on Rails main a few weeks ago: https://github.com/rails/rails/pull/55352

As for `T_ARRAY`, it seems to mostly be `ActiveSupport::Inflector::Inflections::Uncountables`.

byroot:
Out of curiosity I also looked at the sequel benchmark:

```
[
  ["SQLite3::Backup", 4073197],
  ["T_ARRAY", 171011],
  ["T_STRING", 56]
}
```

Here T_ARRAY seem to come from the older version of sqlite3-ruby that used Array and Hash subclasses for its result sets, but it has been deprecated and removed recently, so that's good new: https://github.com/sparklemotion/sqlite3-ruby/commit/002443e1e84c109c32cff7b1bd89d46add5eb4c0

byroot:
@jeremyevans I opened https://github.com/ruby/ruby/pull/14179 to shrink sets back, but figured an independent PR was best.

variable.c
```diff
@@ -1279,13 +1293,20 @@ rb_obj_set_fields(VALUE obj, VALUE fields_obj, ID field_name, VALUE original_fie
 
     if (fields_obj != original_fields_obj) {
         switch (BUILTIN_TYPE(obj)) {
+          case T_DATA:
+            if (LIKELY(RTYPEDDATA_P(obj))) {
+                RB_OBJ_WRITE(obj, &RTYPEDDATA(obj)->fields_obj, fields_obj);
+                break;
+            }
+            goto generic_fields;
```
peterzhu2118, line 1301: Why do we need to insert into the `generic_fields_tbl` if we already store the `fields_obj`?

peterzhu2118: Ah never mind, this is for the non-TypedData DATA types.

composerinteralia:
GitHub is seeing some http://github.com/github/memcached errors after this commit. Is that expected because it's using the old data classes (and some instance variables via SWIG that I haven't quite wrapped my head around yet). The error from memcached looks like:

```
  TypeError: Expected argument 0 of type memcached_st *, but got Rlibmemcached::MemcachedSt #<Rlibmemcached::MemcachedSt:0... (TypeError)
  	in SWIG method 'memcached_create'
  
    Lib.memcached_create(empty_struct)
                         ^^^^^^^^^^^^
.../vendor/gems/3.5.0/ruby/3.5.0+3/bundler/gems/memcached-13d52378016c/lib/memcached/exceptions.rb:63:in 'Rlibmemcached.memcached_create'
```

quickly followed by a segfault. That error is triggered by [this code](https://github.com/github/memcached/blob/13d52378016cb3a2f3d96197eee8454739876812/ext/rlibmemcached/rlibmemcached_wrap.c#L6950-L6953).

I'll keep looking into it, but figured I'd ask here in case I'm missing something obvious.

byroot:
That is weird. Make sure the extension is recompiled, because the `RData` struct was re-ordered. I bumped the ABI version but perhaps your build system persisted the old compiled code?

Otherwise a repro script would be welcome (`memcached` is a bit of a shitshow though...)

composerinteralia:
> memcached is a bit of a shitshow though...

Yeah, it seems that way.


Pretty sure it was recompiled (we reinstalled all the gems from scratch into `vendor/gems/3.5.0/ruby/3.5.0+3/...`). I'll try to narrow it down. Thank you.

byroot:
There's a non-zero chance it's the same bug than https://github.com/ruby/ruby/pull/14291

composerinteralia:
I don't think it's the same, unfortunately. I didn't have much luck debugging the memcached error (adding print statements near the crash made it go away half the time, so I struggled to get too far with that).

But I think I found a similar crash with the gpgme gem (hopefully reproducible 🤞🏻 — I couldn't reproduce on my Mac). On the previous commit this works fine:

```
$ gem install gpgme
$ ruby -r gpgme -e "p RUBY_DESCRIPTION; GPGME::Key.import('whatever')"
"ruby 3.5.0dev (2025-08-12T19:56:57Z master 85c52079aa) +PRISM [x86_64-linux]"
```

But on this commit and on master it crashes, although the crash varies depending on how exactly I run it (e.g. it crashes differently with prism vs. parse.y — showing the parse.y version only because it looks more similar to the memcached one).

```
$ ruby --parser=parse.y -e "p RUBY_DESCRIPTION; require 'gpgme'; GPGME::Key.import('whatever')" 
"ruby 3.5.0dev (2025-08-21T22:52:15Z master 6fbe2dd36e) [x86_64-linux]"
/workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/ctx.rb:82:in 'GPGME.gpgme_release': wrong argument type GPGME::Ctx (expected Data) (TypeError)

          GPGME::gpgme_release(ctx)
                               ^^^
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/ctx.rb:82:in 'GPGME::Ctx.new'
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/key.rb:126:in 'GPGME::Key.import'
	from -e:1:in '<main>'
/workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/ctx.rb:437:in 'GPGME.gpgme_op_import': wrong argument type GPGME::Ctx (expected Data) (TypeError)

      err = GPGME::gpgme_op_import(self, keydata)
                                   ^^^^^^^^^^^^^
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/ctx.rb:437:in 'GPGME::Ctx#import_keys'
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/key.rb:127:in 'block in GPGME::Key.import'
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/ctx.rb:80:in 'GPGME::Ctx.new'
	from /workspaces/.rubies/master/lib/ruby/gems/3.5.0+3/gems/gpgme-2.0.25/lib/gpgme/key.rb:126:in 'GPGME::Key.import'
	from -e:1:in '<main>'
-e: [BUG] Segmentation fault at 0x0000000000000000
ruby 3.5.0dev (2025-08-21T22:52:15Z master 6fbe2dd36e) [x86_64-linux]

c:0001 p:0000 s:0003 E:001940 DUMMY  [FINISH]


Total ractor count: 1
Ruby thread count for this ractor: 1

-- Machine register context ------------------------------------------------
 RIP: 0x00006283b11679d2 RBP: 0x0000768c668bf510 RSP: 0x00007ffee61eb408
 RAX: 0x03d9410d8d480003 RBX: 0x0000768c80872000 RCX: 0x0000768c668bed18
 RDX: 0x0000768c668bed18 RDI: 0x0000768c668bf510 RSI: 0x00006283b13c4e20
  R8: 0x0000000000000000  R9: 0x00000000000000c0 R10: 0x0000768c80a008c0
 R11: 0x0000000000000010 R12: 0x0000768c668c0000 R13: 0x0000000000000002
 R14: 0x0000000000000028 R15: 0x0000000000000000 EFL: 0x0000000000010297

/workspaces/.rubies/master/bin/ruby(rb_print_backtrace+0x11) [0x6283b11b09af] ../vm_dump.c:843
/workspaces/.rubies/master/bin/ruby(rb_vm_bugreport) ../vm_dump.c:1175
/workspaces/.rubies/master/bin/ruby(rb_bug_for_fatal_signal+0x102) [0x6283b1167612] ../error.c:1130
/workspaces/.rubies/master/bin/ruby(sigsegv+0x46) [0x6283b0eb4da6] ../signal.c:948
/lib/x86_64-linux-gnu/libpthread.so.0(__restore_rt+0x0) [0x768c81151420]
/workspaces/.rubies/master/bin/ruby(rb_typeddata_inherited_p+0xa) [0x6283b11679d2] ../error.c:1373
/workspaces/.rubies/master/bin/ruby(rb_typeddata_is_kind_of) ../error.c:1383
/workspaces/.rubies/master/bin/ruby(rb_obj_is_thread+0x14) [0x6283b0f6fab4] ../vm.c:3640
/workspaces/.rubies/master/bin/ruby(rb_gc_shutdown_call_finalizer_p+0x27) [0x6283b0da4e97] ../gc.c:341
/workspaces/.rubies/master/bin/ruby(rb_gc_impl_shutdown_call_finalizer) ../gc/default/default.c:3045
/workspaces/.rubies/master/bin/ruby(rb_objspace_call_finalizer) ../gc.c:1731
/workspaces/.rubies/master/bin/ruby(rb_ec_finalize+0x16) [0x6283b0d8050e] ../eval.c:165
/workspaces/.rubies/master/bin/ruby(rb_ec_cleanup) ../eval.c:256
/workspaces/.rubies/master/bin/ruby(rb_main+0x21) [0x6283b0d7b757] ../main.c:42
/workspaces/.rubies/master/bin/ruby(main) ../main.c:62
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xf3) [0x768c80f54083]
[0x6283b0d7b7ae]
```

Sorry I didn't manage to get further with this. I'll be out for the next 10 days, so I won't have a chance to look again until I get back.

Is it possible an RData is somehow getting marked with `IS_TYPED_DATA` and being treated like RTypedData when it shouldn't?

byroot:
> Is it possible an RData is somehow getting marked with `IS_TYPED_DATA` and being treated like RTypedData when it shouldn't?

That's what the crash look like yes.

byroot:
I tried to reproduce using docker containers, but without success.

I also tried to setup ruby-head CI for that repo and it also doesn't fail. https://github.com/byroot/ruby-gpgme/pull/1

😞

byroot:
I also tried to repro in the `memcached` repo but it's in such terrible state that I can't seem to make it work https://github.com/byroot/memcached/actions/runs/17152779869/job/48662664145

composerinteralia:
Thank you for trying! I'll plan to look at this again when I'm back in early September.

composerinteralia:
I'm back and had a chance to look more at this. It looks like the crash happens for both gpgme and memcached when the `dfree` function is at an odd numbered address. For the gpgme example above, [`dmark` is 0 and `dfree` is `gpgme_release`](https://github.com/ueno/ruby-gpgme/blob/e15d3fc436b91a3c51d1dcd489ebff3261e1cf35/ext/gpgme/gpgme_n.c#L98C29-L98C42). If I'm understanding things correctly `dmark` used to line up with the `RTypedData` `type` field, and obviously 0 will never have the `IS_TYPED_DATA` bit set. But now `dfree` lines up with `type`, and when `gpgme_release` is at an odd address it does have the `IS_TYPED_DATA` bit set and the whole thing gets treated incorrectly as an `RTypedData`.

I think this shows the odd offset for `gpgme_release`:

```
objdump -t $(gem which gpgme_n) | grep ' gpgme_release'
00000000000553eb g     F .text	00000000000001e2              gpgme_release
```

If I muck around with the library and recompile it such that it ends up at an even address, everything works fine. On my Mac it's always at an even address regardless. I'm not sure how to reproduce this reliably outside a GitHub codespace I'm working with, since I assume it's platform specific.

Same deal for memcached, except the sometimes-odd-addressed dfree in question there is `memcached_free`.

Would swapping the order of the `RData` `dmark` and `dfree` fields be a possibility to make things work as before? I suppose the longer-term solution is to stop using the deprecated API in these gems, but I'd rather not touch github/memcached too much if I can help it (hopefully we can stop using it at some point... it causes us trouble on every Ruby upgrade).

byroot:
> Would swapping the order of the `RData` `dmark` and `dfree` fields be a possibility to make things work as before? I

I'm at a conference all weeks so I don't have time to think hard about this. We can try your suggestion but I I suspect `dmark` and `dfree` both being function pointers would have the same constraints?

also cc @jeremyevans since you have knowledge about this, so perhaps you have opinions.

composerinteralia:
> I suspect dmark and dfree both being function pointers would have the same constraints?

I would think so, but I wonder if dmark is also more likely to be 0 for the gems still using this older API? I assume that's why we didn't have problems with these particular gems before.

Enjoy the conference!

jeremyevans:
I think in order for the tagged pointer to work, the assumption is that function pointers will not have the low bit set. `0x553eb` has the low bit set, and that violates the assumption. Swapping `dmark` and `dfree` may fix the gpgme issue, but it could break libraries that set `dmark` and not `dfree`.

If we want to work on platforms where function pointers don't enforce alignment of at least 2, then we will need to change the implementation. We cannot revert to the previous implementation, as that would have the same issue with `dmark` that it now has with `dfree`. Our only option I think would be to use object flags. Maybe one of `RUBY_FL_UNUSED{6,9,10}` could be renamed and used for that bit?

composerinteralia:
That all makes sense. Thank you. I'll see if I can get things working with an object flag.

byroot:
> Maybe one of `RUBY_FL_UNUSED{6,9,10}` could be renamed and used for that bit?

We absolutely can. I freed these recently. One would argue that generic flags are hard to come by so using them for a single type is a bit of a waste. But until the flag is actually needed it might be fine.

byroot:
Another possibility would be to try using the high bits of the function pointer for tagging?

byroot:
Using RBasic.flags: https://github.com/ruby/ruby/pull/14470

PR #14136, opened 2025-08-06
ZJIT: Optimize class guards by directly reading klass field

st0012:
Replace `rb_yarv_class_of` call with:
- a constant check for special constants (nil, fixnums, symbols, etc)
- a check for false
- direct memory read at offset 8 for regular heap objects for the class check

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -1432,6 +1432,30 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 2, insns: [:opt_nil_p]
   end
 
+  def test_basic_object_guard_works_with_immediate
+    assert_compiles 'NilClass', %q{
+      class Foo; end
+
+      def test(val) = val.class
+
+      test(Foo.new)
+      test(Foo.new)
+      test(nil)
+    }, call_threshold: 2
+  end
+
+  def test_basic_object_guard_works_with_false
+    assert_compiles 'FalseClass', %q{
+      class Foo; end
+
+      def test(val) = val.class
+
+      test(Foo.new)
+      test(Foo.new)
+      test(false)
+    }, call_threshold: 2
+  end
+
   private
 
   # Assert that every method call in `test_script` can be compiled by ZJIT
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1068,17 +1068,29 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         asm.cmp(val, Qtrue.into());
         asm.jne(side_exit(jit, state, GuardType(guard_type))?);
     } else if guard_type.is_subtype(types::FalseClass) {
-        assert!(Qfalse.as_i64() == 0);
-        asm.test(val, val);
+        asm.cmp(val, Qfalse.into());
         asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_immediate() {
+        // All immediate types' guard should have been handled above
+        panic!("unexpected immediate guard type: {guard_type}");
     } else if let Some(expected_class) = guard_type.runtime_exact_ruby_class() {
-        asm_comment!(asm, "guard exact class");
+        asm_comment!(asm, "guard exact class for non-immediate types");
 
-        // Get the class of the value
-        let klass = asm.ccall(rb_yarv_class_of as *const u8, vec![val]);
+        let side_exit = side_exit(jit, state, GuardType(guard_type))?;
+
+        // Check if it's a special constant
+        asm.test(val, (RUBY_IMMEDIATE_MASK as u64).into());
+        asm.jnz(side_exit.clone());
+
+        // Check if it's false
+        asm.cmp(val, Qfalse.into());
+        asm.je(side_exit.clone());
+
+        // Load the class from the object's klass field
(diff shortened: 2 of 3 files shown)


XrXr (commented):
I think you also need to test for 0. Try making a test case that goes down this code path and then pass that code `false`. That should make it segfault.

st0012:
@XrXr Good catch! Thanks

tekknolagi:
I don't think we want to side exit when we're trying to load the class of an immediate unless for some reason it happens not very often

tekknolagi:
This change goes from slower single operation to potentially slower rest of function

XrXr:
Since we've ruled out the cases where we expect an immediate above, seeing an immediate here guarantees a class mismatch, which side exits just likes when we called `CLASS_OF`.  So I think this impl is as accurate as the old one.

Though we should probably toss in a `!gurad_type.is_immediate()` to make this explicit.

zjit/src/codegen.rs
```diff
@@ -1073,8 +1073,17 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
...
+        // Check if it's false
+        assert!(Qfalse.as_i64() == 0);
+        asm.test(val, val);
+        asm.jz(side_exit(jit, state, GuardType(guard_type))?);
+
+        // Load the class from the object's klass field
+        let klass = asm.load(Opnd::mem(64, val, RUBY_OFFSET_RBASIC_KLASS));
 
         asm.cmp(klass, Opnd::Value(expected_class));
         asm.jne(side_exit(jit, state, GuardType(guard_type))?);
```
XrXr: I'd use a local for the side exit target to show that all these jump to same side exit.

st0012:
> I don't think we want to side exit when we're trying to load the class of an immediate unless for some reason it happens not very often

In that condition the guard type should definitely NOT an immediate type. So I think side exiting is what we want to when the `val` is an immediate type value?

> Though we should probably toss in a !gurad_type.is_immediate() to make this explicit.

Yeah that makes sense. I added a condition to panic when we see any immediate guard type that's not already handled in the previous branches.

tekknolagi:
> Since we've ruled out the cases where we expect an immediate above, seeing an immediate here guarantees a class mismatch, which side exits just likes when we called CLASS_OF. So I think this impl is as accurate as the old one.

That's fair. I was thinking of a (hypothetical) future case where we're looking up methods on `Integer` for a `Fixnum` but in that case I suppose we should still guard that it's a `Fixnum`. So never mind! Great stuff.

test/ruby/test_zjit.rb
```diff
@@ -1432,6 +1432,18 @@ def test(val) = val.nil?
     }, call_threshold: 2, insns: [:opt_nil_p]
   end
 
+  def test_non_special_constant_guard_works_with_false
+    assert_compiles 'FalseClass', %q{
```
tekknolagi:
```suggestion
  def test_basic_object_guard_works_with_immediate
    assert_compiles 'FalseClass', %q{
```

k0kubun:
Have you seen this failure? https://github.com/ruby/ruby/actions/runs/16807814799/job/47604843959?pr=14136

```
    1) Failure:
  TestStruct::TopStruct#test_named_structs_are_not_rooted [/Users/runner/work/ruby/ruby/src/test/ruby/test_struct.rb:541]:
  pid 46863 killed by SIGABRT (signal 6)
  | ../src/vm_callinfo.h:421: Assertion Failed: vm_cc_cme:cc->klass != Qundef || !vm_cc_markable(cc)
```

zjit/src/codegen.rs
```diff
@@ -1071,14 +1071,28 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
...
+        let side_exit = side_exit(jit, state, GuardType(guard_type))?;
+
+        // Check if it's a special constant
+        asm.test(val, (RUBY_IMMEDIATE_MASK as u64).into());
+        asm.jnz(side_exit.clone());
+
+        // Check if it's false
+        assert!(Qfalse.as_i64() == 0);
+        asm.test(val, val);
+        asm.jz(side_exit.clone());
```
k0kubun: I think it's a result of a micro-optimization in YJIT, but the backend is capable of doing that today. Without adding the assertion, you can write:

```suggestion
        asm.cmp(val, Qfalse.into());
        asm.je(side_exit.clone());
```

and the x86 backend will convert `cmp(val, 0)` to `test(val, val)`.

st0012: Thanks. I also update the condition for false type guard.

zjit/src/codegen.rs
```diff
@@ -1068,17 +1068,29 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         asm.cmp(val, Qtrue.into());
         asm.jne(side_exit(jit, state, GuardType(guard_type))?);
     } else if guard_type.is_subtype(types::FalseClass) {
-        assert!(Qfalse.as_i64() == 0);
-        asm.test(val, val);
+        asm.cmp(val, Qfalse.into());
         asm.jne(side_exit(jit, state, GuardType(guard_type))?);
+    } else if guard_type.is_immediate() {
+        // All immediate types' guard should be inlined
```
k0kubun: Every case is inlined at this point, so this comment doesn't make sense to me.

```suggestion
        // All immediate types' guard should have been handled above
```

k0kubun (approved):

PR #14137, opened 2025-08-06
[DOC] Tweaks for GC.config

BurdetteLamar:


gc.rb
```diff
@@ -250,71 +250,58 @@ def self.stat_heap heap_name = nil, hash_or_key = nil
 
   # call-seq:
   #     GC.config -> hash
-  #     GC.config(hash) -> hash
+  #     GC.config(hash_to_merge) -> merged_hash
   #
-  # Sets or gets information about the current \GC config.
+  # This method is expected to be defined, useful, and well-behaved
+  # only in the CRuby implementation.
```
peterzhu2118: I think this is too wordy. Maybe something like "This method is implementation-specific to CRuby" is better?

gc.rb
```diff
@@ -250,71 +250,58 @@ def self.stat_heap heap_name = nil, hash_or_key = nil
...
   #
-  # Configuration parameters are \GC implementation-specific and may change
-  # without notice.
```
peterzhu2118, line 258: We should keep this. Ruby now allows alternate GC implementation, and the parameters specified here may not apply to a different implementation.

gc.rb
```diff
@@ -250,71 +250,58 @@ def self.stat_heap heap_name = nil, hash_or_key = nil
...
-  #   A flag will be set to notify that a full mark has been
-  #   requested. This flag is accessible using
-  #   <code>GC.latest_gc_info(:need_major_by)</code>
-  #
-  #   The user can trigger a major collection at any time using
-  #   <code>GC.start(full_mark: true)</code>
-  #
-  #   When +false+, Young to Old object promotion is disabled. For performance
-  #   reasons, it is recommended to warm up an application using +Process.warmup+
+  #   Setting this parameter to +false+ disables young-to-old promotion .
```
peterzhu2118: Nit: space before period

```suggestion
  #   Setting this parameter to +false+ disables young-to-old promotion.
```

gc.rb
```diff
@@ -250,71 +250,58 @@ def self.stat_heap heap_name = nil, hash_or_key = nil
...
-  #   only young objects will be marked. When the heap space is exhausted, new
-  #   pages will be allocated immediately instead of running a full mark.
+  #   - +true+ (default): \GC interleaves major and minor collections.
+  #   - +false+: \GC does not initiate a full marking cycle unless
+  #     explicitly directed by user code;
+  #     see GC.start.
   #
-  #   A flag will be set to notify that a full mark has been
-  #   requested. This flag is accessible using
-  #   <code>GC.latest_gc_info(:need_major_by)</code>
```
peterzhu2118, line 310: This is useful information we should keep.

gc.rb
BurdetteLamar: Done.

gc.rb
BurdetteLamar: Restored.

gc.rb
BurdetteLamar: Fixed.

gc.rb
BurdetteLamar: Restored (with slight reorganization).

PR #14139, opened 2025-08-07
Favor `rb_gc_mark_and_move` over `rb_gc_location`

casperisfine:
Previous discussion in https://github.com/ruby/ruby/pull/14076

The `p->field = rb_gc_location(p->field)` isn't ideal because it means all
references are rewritten on compaction, regardless of whether the referenced
object has moved. This isn't good for caches nor for Copy-on-Write.

`rb_gc_mark_and_move` avoid needless writes, and most of the time allow to
have a single function for both marking and updating references.

### DIFF
--- a/time.c
+++ b/time.c
@@ -1888,39 +1888,25 @@ force_make_tm(VALUE time, struct time_object *tobj)
 }
 
 static void
-time_mark(void *ptr)
+time_mark_and_move(void *ptr)
 {
     struct time_object *tobj = ptr;
     if (!FIXWV_P(tobj->timew)) {
-        rb_gc_mark_movable(w2v(tobj->timew));
+        rb_gc_mark_and_move(&WIDEVAL_GET(tobj->timew));
     }
-    rb_gc_mark_movable(tobj->vtm.year);
-    rb_gc_mark_movable(tobj->vtm.subsecx);
-    rb_gc_mark_movable(tobj->vtm.utc_offset);
-    rb_gc_mark_movable(tobj->vtm.zone);
-}
-
-static void
-time_compact(void *ptr)
-{
-    struct time_object *tobj = ptr;
-    if (!FIXWV_P(tobj->timew)) {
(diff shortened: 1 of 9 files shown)


time.c
```diff
@@ -1888,39 +1888,25 @@ force_make_tm(VALUE time, struct time_object *tobj)
 }
 
 static void
-time_mark(void *ptr)
+time_mark_and_move(void *ptr)
 {
     struct time_object *tobj = ptr;
     if (!FIXWV_P(tobj->timew)) {
-        rb_gc_mark_movable(w2v(tobj->timew));
+        rb_gc_mark_and_move(&WIDEVAL_GET(tobj->timew));
```
Earlopain, line 1895: I feel like this still needs to use `w2v` in some way, previously there was https://github.com/ruby/ruby/commit/b8de3cfb04e4d510d3a3af73029a9ca5ab39f79e. On some architecures with musl, I am getting the following build error:

```
time.c: In function 'time_mark_and_move':
time.c:1895:29: error: passing argument 1 of 'rb_gc_mark_and_move' from incompatible pointer type [-Wincompatible-pointer-types]
1895 |         rb_gc_mark_and_move(&WIDEVAL_GET(tobj->timew));
|                             ^
|                             |
|                             wideval_t * {aka long long unsigned int *}
In file included from ./shape.h:4,
from internal/variable.h:16,
from time.c:43:
./internal/gc.h:213:33: note: expected 'VALUE *' {aka 'long unsigned int *'} but argument is of type 'wideval_t *' {aka 'long long unsigned int *'}
213 | void rb_gc_mark_and_move(VALUE *ptr);
|                          ~~~~~~~^~~
```

Maybe so?

```suggestion
        VALUE timew = w2v(tobj->timew);
        rb_gc_mark_and_move(&timew);
```

byroot: 👀

byroot: https://github.com/ruby/ruby/pull/14161

PR #14142, opened 2025-08-07
Add link to Ruby options doc in help text

peterzhu2118:
Adds link to https://docs.ruby-lang.org/en/master/ruby/options_md.html in Ruby help text (-h and --help).

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -61,6 +61,7 @@
 #include "ruby/util.h"
 #include "ruby/version.h"
 #include "ruby/internal/error.h"
+#include "version.h"
 
 #define singlebit_only_p(x) !((x) & ((x)-1))
 STATIC_ASSERT(Qnil_1bit_from_Qfalse, singlebit_only_p(Qnil^Qfalse));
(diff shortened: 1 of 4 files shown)


st0012:
Just confirmed that html comments work with RDoc:

```
<!---
  Important! This page's link on docs.ruby-lang.org is displayed in Ruby's help message.
  Please make sure you update the link when renaming or moving this file.
--->
```

ruby.c
```diff
@@ -403,7 +404,10 @@ usage(const char *name, int help, int highlight, int columns)
     unsigned int w = (columns > 80 ? (columns - 79) / 2 : 0) + 16;
 #define SHOW(m) show_usage_line(&(m), help, highlight, w, columns)
 
-    printf("%sUsage:%s %s [options] [--] [filepath] [arguments]\n", sb, se, name);
+    printf("%sUsage:%s %s [options] [--] [filepath] [arguments]\n\n", sb, se, name);
+    printf("Details and examples at https://docs.ruby-lang.org/en/%s/ruby/options_md.html\n",
+           RUBY_PATCHLEVEL == -1 ? "master" : STRINGIZE(RUBY_VERSION_MAJOR) "." STRINGIZE(RUBY_VERSION_MINOR));
+
```
nobu, line 410: Detailed info should be only when `--help`.
And keep revision info only in version.c.

peterzhu2118: I'm not sure what you mean here.

> Detailed info should be only when --help.

We're not showing any detailed info? Just linking to a web page that could explain and give examples about the options in `-h` and `--help`. I think it should be shown for both of these cases. Do you have other opinions?

> And keep revision info only in version.c.

Do you mean the use of `RUBY_PATCHLEVEL`? How would we get the information about whether to use `master` or a version otherwise?

mame:
@peterzhu2118 The output of `ruby -h` is intended to fit the traditional 80x25 size. Adding the message line breaks the requirement. So this change requires Matz's approval.

In addition, the test references the undefined constant `RUBY_VERSION_MAJOR`, which is found a blocker for the Ruby 4.0.0 release.

I am sorry that this is realized at the last minute, but @nurse will revert this PR for Ruby 4.0.0.

peterzhu2118:
Ah sorry about that. Thanks @nurse and @mame for reverting it!

> The output of ruby -h is intended to fit the traditional 80x25 size. 

Ah I wasn't aware of it. I could put the URL on a new line to fit within 80 characters.

PR #14143, opened 2025-08-07
ZJIT: Improve bisect script

tekknolagi:


st0012 (approved):

tekknolagi:
cc @XrXr who pointed out some issues

tool/zjit_bisect.rb
```diff
@@ -1,12 +1,26 @@
...
 require 'open3'
+require 'optparse'
+require 'shellwords'
 require 'tempfile'
 require 'timeout'
 
-RUBY = ARGV[0] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
-OPTIONS = ARGV[1] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> <options>")
-TIMEOUT_SEC = 5
+ARGS = {:timeout => 5}
```
k0kubun:
```suggestion
ARGS = {timeout: 5}
```

nitpick: Nobody prefers the Ruby 1.8 syntax in 2025

tool/zjit_bisect.rb
```diff
@@ -58,40 +72,63 @@ def run_bisect(command, items)
   bisect_impl(command, [], items)
 end
 
+def run_ruby *cmd
+  stdout_data = nil
+  stderr_data = nil
+  status = nil
+  Open3.popen3(Shellwords.join cmd) do |stdin, stdout, stderr, wait_thr|
```
k0kubun: IIRC, if you do it, it would shell out. It seems better to pass the array as is.

```suggestion
  Open3.popen3(*cmd) do |stdin, stdout, stderr, wait_thr|
```

k0kubun: And why don't you use `Open3.capture3`? I guess you wanted to wrap a `Timeout.timeout` but doing that around `capture3` seems enough?

tool/zjit_bisect.rb
```diff
@@ -58,40 +72,63 @@ def run_bisect(command, items)
...
-  status.success?
+  _, _, exitcode = run_with_jit_list(RUBY, OPTIONS, items)
+  exitcode == 0
 end
 result = run_bisect(command, jit_list)
 File.open("jitlist.txt", "w") do |file|
   file.puts(result)
 end
+puts "Run:"
+command = Shellwords.join [RUBY, "--zjit-allowed-iseqs=jitlist.txt", *OPTIONS]
```
k0kubun: This is just a personal preference, but in case you missed it, there's also:

```suggestion
command = [RUBY, "--zjit-allowed-iseqs=jitlist.txt", *OPTIONS].shelljoin
```

tool/zjit_bisect.rb
tekknolagi: Haha my Ruby knowledge is so outdated 😅

tool/zjit_bisect.rb
tekknolagi: Doesn't work with capture3 :/ the process doesn't always die. I originally tried that but it will just wait until child exits

k0kubun: ok, makes sense then

XrXr: Yeah, you shouldn't need any shell parsing here and can forward everything after "--" to the child ruby process as it. Only need to do shell parsing if you take all the options as one element of ARV.

XrXr: You also don't need to *cmd, I think just passing `cmd` will just work.

tool/zjit_bisect.rb
```diff
@@ -1,12 +1,26 @@
...
+  opts.on("--timeout=TIMEOUT_SEC", "Seconds until child process is killed") do |timeout|
+    ARGS[:timeout] = Integer(timeout)
+  end
+  opts.on("-h", "--help", "Prints this help") do
+    puts opts
+    exit
+  end
+end.parse!
+
+RUBY = ARGV[0] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> -- <options>")
```
XrXr, line 21: WDYT about using `RbConfig.ruby` to get the current ruby path? Feels easier to use, and we can have an option to pass  a path too.

tekknolagi: Eh, prefer explicit here

PR #14144, opened 2025-08-07
ZJIT: Remove the need for unwrap() on with_num_bits()

k0kubun:
This PR lets `with_num_bits()` return `Opnd` instead of `Option<Opnd>`. It also adds `try_num_bits()` that returns `Option<Opnd>` to make the `StaticSymbol` check safer.

I realized YJIT and ZJIT always call `unwrap()` after `Opnd::with_num_bits()`. If we always panic anyway, it seems more useful to panic inside `with_num_bits()` so that you can see what `Opnd` it was.

However, one of the `with_num_bits()` usages in ZJIT seemed not 100% safe to me, so I added `try_num_bits()` for that place and added a debug assertion to expect `Some` in the debug mode.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -146,14 +146,25 @@ impl Opnd
         }
     }
 
-    pub fn with_num_bits(&self, num_bits: u8) -> Option<Opnd> {
+    /// Return Some(Opnd) with a given num_bits if self has num_bits.
+    /// None if self doesn't have a num_bits field.
+    pub fn try_num_bits(&self, num_bits: u8) -> Option<Opnd> {
         assert!(num_bits == 8 || num_bits == 16 || num_bits == 32 || num_bits == 64);
         match *self {
             Opnd::Reg(reg) => Some(Opnd::Reg(reg.with_num_bits(num_bits))),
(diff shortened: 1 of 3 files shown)


zjit/src/backend/lir.rs
```diff
@@ -146,17 +146,28 @@ impl Opnd
...
             Opnd::Mem(Mem { base, disp, .. }) => Some(Opnd::Mem(Mem { base, disp, num_bits })),
             Opnd::VReg { idx, .. } => Some(Opnd::VReg { idx, num_bits }),
-            //Opnd::Stack { idx, stack_size, num_locals, sp_offset, reg_mapping, .. } => Some(Opnd::Stack { idx, num_bits, stack_size, num_locals, sp_offset, reg_mapping }),
             _ => None,
         }
     }
 
+    /// Return Opnd with a given num_bits if self has num_bits.
+    /// Panic otherwise. This should be used only when you know which Opnd self is.
+    pub fn with_num_bits(&self, num_bits: u8) -> Opnd {
```
XrXr:
```suggestion
    /// Panic otherwise. This should be used only when you know which Opnd self is.
    #[track_caller]
    pub fn with_num_bits(&self, num_bits: u8) -> Opnd {
```

Just a bit of stack trace editing in case it does panic.

XrXr (approved):

PR #14146, opened 2025-08-08
gc.c: Filter garbage in build_id2ref_i

byroot:
Given `ObjectSpace._id2ref` is a deprecated API, it is assumed it won't ever be called, hence we optimistically don't keep an `id -> ref` mapping.

So when `object_id_to_ref` is called for the first time, we go over the entire heap to build that missing table.

At that point we must be very careful about garbage objects. If they have an object id, the GC will attempt to remove them from the `id2ref` table when sweeping them.

Until now, the strategy was to insert them anyway, but this actually cause issue with objects that rely on `IMEMO/fields` because the object fields may be already sweeped (hence ASAN poisoned).

Unfortunately I'm not able to craft a reproduction script to validate this theory, as the GC need to be in a precise state for the bug to happen.

Hoping this would fix the following ASAN issue:

```
=================================================================
==252918==ERROR: AddressSanitizer: use-after-poison on address 0x7f63c776f4a8 at pc 0x6340aff886b3 bp 0x7fffcfbe8730 sp 0x7fffcfbe8728
READ of size 8 at 0x7f63c776f4a8 thread T0
    #0 0x6340aff886b2 in RB_FL_TEST_RAW /tmp/ruby/src/trunk_asan/include/ruby/internal/fl_type.h:466:25
    #1 0x6340aff886b2 in rb_imemo_fields_ptr /tmp/ruby/src/trunk_asan/internal/imemo.h:291:9
    #2 0x6340aff886b2 in rb_obj_field_get /tmp/ruby/src/trunk_asan/variable.c:1356:18
    #3 0x6340afc18fe3 in object_id /tmp/ruby/src/trunk_asan/gc.c:1938:12
    #4 0x6340afc18fe3 in rb_find_object_id /tmp/ruby/src/trunk_asan/gc.c:2171:12
    #5 0x6340afc18fe3 in rb_obj_id /tmp/ruby/src/trunk_asan/gc.c:2226:12
    #6 0x6340afc2952e in build_id2ref_i /tmp/ruby/src/trunk_asan/gc.c:1958:35
    #7 0x6340afc2952e in rb_gc_impl_each_object /tmp/ruby/src/trunk_asan/gc/default/default.c:3079:17
    #8 0x6340afc2952e in object_id_to_ref /tmp/ruby/src/trunk_asan/gc.c:1984:13
    #9 0x6340afc2952e in id2ref /tmp/ruby/src/trunk_asan/gc.c:2143:17
    #10 0x6340afc2952e in os_id2ref /tmp/ruby/src/trunk_asan/gc.c:2157:12
    #11 0x6340b0016c57 in vm_call_cfunc_with_frame_ /tmp/ruby/src/trunk_asan/vm_insnhelper.c:3848:11
```

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1755,7 +1755,6 @@ rb_gc_pointer_to_heap_p(VALUE obj)
 #define LAST_OBJECT_ID() (object_id_counter * OBJ_ID_INCREMENT)
 static VALUE id2ref_value = 0;
 static st_table *id2ref_tbl = NULL;
-static bool id2ref_tbl_built = false;
 
 #if SIZEOF_SIZE_T == SIZEOF_LONG_LONG
 static size_t object_id_counter = 1;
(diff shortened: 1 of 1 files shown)


gc.c
```diff
@@ -1938,24 +1937,50 @@ object_id(VALUE obj)
...
+            }
+            else {
+                st_insert(id2ref_tbl, RCLASS(obj)->object_id, obj);
+            }
         }
         break;
       case T_IMEMO:
+        if (IMEMO_TYPE_P(obj, imemo_fields) && rb_shape_obj_has_id(obj) && rb_gc_impl_garbage_object_p(objspace, obj)) {
+            rb_imemo_fields_clear(obj);
+        }
```
byroot, line 1970: A change I've meant to do for a while, and that may simplify all this.

I think it'd be helpful if `imemo_fields.klass` pointed to the owner object rather than its class. In situations like this, it would allow to deal with `object_id` entirely from the `T_IMEMO`, and we could ignore the associated object, breaking the interdependency.

gc.c
```diff
@@ -1938,24 +1937,50 @@ object_id(VALUE obj)
     return object_id0(obj);
 }
 
+struct build_id2ref_args {
+    st_table *id2ref_tbl;
+    rb_objspace_t *objspace;
```
peterzhu2118, line 1942: Nit: `rb_objspace_t` no longer exists outside of the default GC, should be a `void *` here.

byroot:
> I think the simpler solution is to (1) complete incremental GC if we're in one, and (2) disable GC while building the id2ref table.

I'm fine with that. We already do (2). How do I do (1) ?

peterzhu2118:
`rb_objspace_gc_disable` or `rb_gc_impl_gc_disable(objspace, true)`.

byroot:
Thanks for the suggestions, closing in favor of: https://github.com/ruby/ruby/pull/14150

PR #14147, opened 2025-08-08
imemo_fields: store owner object in RBasic.klass

casperisfine:
Followup on https://github.com/ruby/ruby/pull/14146

It is much more convenient than storing the klass, especially when dealing with `object_id` as it allows to update the id2ref table without having to dereference the owner, which may be garbage at that point.


byroot:
Superseeded by https://github.com/ruby/ruby/pull/14202

PR #14152, opened 2025-08-08
ZJIT: Avoid compiling and direct sends to forwardable ISEQs

XrXr:
These `...` ISEQs have a special calling convention in the interpreter
and our stubs and JIT calling convention don't handle these correctly.

The attached test passes if we just avoid direct sends, but also
avoid compiling ISEQs with a `...` parameter to limit exposure for now.

`SendWithoutBlock`, which does dynamic dispatch using interpreter code,
seems to handle calling into forwardable ISEQs correctly, so they are
fine -- we can't predict where these dynamic sends land anyways.


zjit/src/hir.rs
```diff
@@ -2549,6 +2553,7 @@ pub const SELF_PARAM_IDX: usize = 0;
 
 fn filter_unknown_parameter_type(iseq: *const rb_iseq_t) -> Result<(), ParseError> {
     if unsafe { rb_get_iseq_body_param_opt_num(iseq) } != 0 { return Err(ParseError::UnknownParameterType(ParameterType::Optional)); }
+    if unsafe { rb_get_iseq_flags_forwardable(iseq) } { return Err(ParseError::UnknownParameterType(ParameterType::Forwardable)); }
```
tekknolagi, line 2556: can you also add a HIR test for failing to compile methods with forwarding params? we have some prior art

XrXr:
Fixes the crash on yjit-bench `activerecord` for me on ARM. Might fix all other crashes on https://github.com/Shopify/ruby/issues/671 as well.

tekknolagi (approved):
Nice!

tekknolagi:
Does this also fix the railtie issue?

tekknolagi:
Also check on `test_param_forms_get_bb_param` :P

k0kubun (approved):

XrXr:
> Does this also fix the railtie issue?

I don't know. Probably? I'm kinda scared to run https://github.com/tekknolagi/store because `Gemfile.lock` is so huge. Could you test it? `gh pr checkout 14152`.

tekknolagi:
Haha sure

tekknolagi:
Passes!

PR #14153, opened 2025-08-08
ZJIT: Fix "memory operand with non-register base"

k0kubun:
Follow-up on: https://github.com/ruby/ruby/pull/14136

This fixes a crash on benchmarks with "internal error: entered unreachable code: memory operand with non-register base".

`val` can be of any operand type. When it's a memory operand, e.g. a spilled basic block argument, you cannot use it as the base of `Opnd::mem(64, val, RUBY_OFFSET_RBASIC_KLASS)`. So this PR loads such an operand into a register. It does it a bit early to hopefully speed up checks on `val`.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1088,9 +1088,14 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
     } else if let Some(expected_class) = guard_type.runtime_exact_ruby_class() {
         asm_comment!(asm, "guard exact class for non-immediate types");
 
-        let side_exit = side_exit(jit, state, GuardType(guard_type))?;
+        // If val isn't in a register, load it to use it as the base of Opnd::mem later.
+        let val = match val {
+            Opnd::Reg(_) | Opnd::VReg { .. } => val,
+            _ => asm.load(val),
+        };
 
         // Check if it's a special constant
+        let side_exit = side_exit(jit, state, GuardType(guard_type))?;
         asm.test(val, (RUBY_IMMEDIATE_MASK as u64).into());
         asm.jnz(side_exit.clone());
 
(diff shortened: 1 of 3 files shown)


zjit/src/codegen.rs
```diff
@@ -1088,9 +1088,14 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
     } else if let Some(expected_class) = guard_type.runtime_exact_ruby_class() {
         asm_comment!(asm, "guard exact class for non-immediate types");
 
-        let side_exit = side_exit(jit, state, GuardType(guard_type))?;
+        // If val isn't in a register, load it to use it as the base of Opnd::mem later.
+        let val = match val {
```
tekknolagi, line 1093: This seems like it belongs at a lower level than the HIR->LIR transformation. HIR->LIR imo should not have to care about the shapes of the operands except to create them.

That being said, I don't have a good way to do this cleanly across test + cmp + Opnd::mem that avoids loading it multiple times. Let's leave a TODO and task.

tekknolagi:
This also fixes `rake test` in a Rails app. Nice!

zjit/src/codegen.rs
k0kubun: Filed an issue https://github.com/Shopify/ruby/issues/685 and added a TODO comment linking to it https://github.com/ruby/ruby/pull/14153/commits/f69426f579aee7ac4d4c2dad0d08517c0183fc1a.

> That being said, I don't have a good way to do this cleanly across test + cmp + Opnd::mem that avoids loading it multiple times.

To do that in the backend, you probably need to introduce a scan of the whole LIR sequence to discover memory operands that need to be loaded and are used multiple times. While it stays `O(N)`, we know from the experience of YJIT that it slows down the compilation speed whenever you add a LIR pass.

Another way to do this would be to let one pass (`split`) do the scan and let the next pass (`alloc_regs`) do the actual split so that it doesn't increase the number of scans. But overloading compiler passes for different use complicates the implementation, and I'm not sure if it's possible to alloc regs without finishing the split.

So, yeah, I don't have a good way to do this either. I'm probably biased in a bad way because it's been kind of a common pattern in YJIT, but this looks like the most reasonable implementation as is for sharing it across three instructions.

st0012:
Good catch! But why isn't this a problem for immediate guard checks 🤔

k0kubun:
Are there any "memory operand" created by them?

st0012:
Oh sorry I misunderstood. I thought the error is raised in `asm.test` but it's actually from `Opnd::mem`.

PR #14154, opened 2025-08-08
ZJIT: Implement `concatstrings` insn

st0012:
@amomchilov and I added both a new HIR (`StringConcat`) + codegen for it in this PR.

We also added `stack_pop_n` to `FrameState` as it's been a repeated pattern for HIRs.

PR #14155, opened 2025-08-08
ZJIT: Extract `stack_popn` helper

amomchilov:
There's a number of places where we extract a variable number of values off the stack to build up a new Vec. Now that the logic is centralized, we can sprinkle in some `unsafe` wizardry to fill the Vec back-to-front, skipping the need for a `reverse()` step.

There was one other similar case that I didn't touch:

https://github.com/ruby/ruby/blob/8a63c17158ec0e3c73c8b36c7c4dedcf7fdd0215/zjit/src/hir.rs#L2840-L2846

We could extract out some generic thing that takes a lambda to call `n` times, but ... meh.

Update: I didn't realize the `stack` is also just a Vec. If we have random access into it and aren't limited to `pop_last()`, then we don't need this `unsafe` trick to write back-to-front.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2468,6 +2468,22 @@ impl FrameState {
         self.stack.pop().ok_or_else(|| ParseError::StackUnderflow(self.clone()))
     }
 
+    /// Return a vec of the top n stack operands, in the same order as they were on the stack.
+    fn stack_popn(&mut self, n: usize) -> Result<Vec<InsnId>, ParseError> {
+        let mut elements = Vec::with_capacity(n);
+        self.stack_popn_into(n, &mut elements)?;
+        Ok(elements)
+    }
+
+    fn stack_popn_into(&mut self, n: usize, destination: &mut Vec<InsnId>) -> Result<(), ParseError> {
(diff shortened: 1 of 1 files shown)


zjit/src/hir.rs
```diff
@@ -2468,6 +2468,22 @@ impl FrameState {
...
 
+    /// Return a vec of the top n stack operands, in the same order as they were on the stack.
+    fn stack_popn(&mut self, n: usize) -> Result<Vec<InsnId>, ParseError> {
+        let mut elements = Vec::with_capacity(n);
+        self.stack_popn_into(n, &mut elements)?;
+        Ok(elements)
+    }
+
+    fn stack_popn_into(&mut self, n: usize, destination: &mut Vec<InsnId>) -> Result<(), ParseError> {
+        unsafe { // Unsafely set the length, so we don't have to initialize the slots we're about to overwrite.
```
tekknolagi, line 2479: Let's not do unsafe. We don't need perf for this. We can assume random access.

tekknolagi:
oops sorry did not realize it was a draft

zjit/src/hir.rs
amomchilov: Read the description :)

amomchilov:
Stan rolled this right into #14154

PR #14158, opened 2025-08-08
ZJIT: Add compile/profile/GC/invalidation time stats

k0kubun:
This PR introduces the following counters to `RubyVM::ZJIT.stats`:

* `compile_time_ns`
* `profile_time_ns`
* `gc_time_ns`
* `invalidation_time_ns`

They should allow us to investigate which part of ZJIT is adding overhead to yjit-bench.

They are added as default counters, which are available without `--zjit-stats`. We often need these metrics in production on production clusters that don't enable stats. It also refactors and improves the stats system a little.

### DIFF
--- a/zjit.rb
+++ b/zjit.rb
@@ -42,7 +42,7 @@ class << RubyVM::ZJIT
       :zjit_insns_count,
       :ratio_in_zjit,
     ].each do |key|
-      # Some stats like vm_insns_count and ratio_in_yjit are not supported on the release build
+      # Some stats like vm_insns_count and ratio_in_zjit are not supported on the release build
       next unless stats.key?(key)
       value = stats[key]
 
--- a/zjit.rb
+++ b/zjit.rb
@@ -17,6 +17,7 @@ class << RubyVM::ZJIT
   # Return ZJIT statistics as a Hash
   def stats
     stats = Primitive.rb_zjit_stats
+    return nil if stats.nil?
 
     if stats.key?(:vm_insns_count) && stats.key?(:zjit_insns_count)
       stats[:total_insns_count] = stats[:vm_insns_count] + stats[:zjit_insns_count]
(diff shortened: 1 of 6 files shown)


zjit.rb
```diff
@@ -32,15 +33,29 @@ def stats_string
...
+      :compile_time_ns,
+      :profile_time_ns,
+      :gc_time_ns,
+      :invalidation_time_ns,
       :total_insns_count,
       :vm_insns_count,
       :zjit_insns_count,
       :ratio_in_zjit,
     ].each do |key|
+      # Some stats like vm_insns_count and ratio_in_yjit are not supported on the release build
```
st0012:
```suggestion
      # Some stats like vm_insns_count and ratio_in_zjit are not supported on the release build
```

zjit/src/stats.rs
```diff
@@ -49,14 +86,30 @@ pub extern "C" fn rb_zjit_stats(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+
+        if unsafe { rb_vm_insns_count } > 0 {
+            set_stat(hash, "vm_insns_count", unsafe { rb_vm_insns_count });
+        }
     }
 
     hash
 }
+
+/// Measure the time taken by func() and add that to yjit_compile_time.
```
st0012:
```suggestion
/// Measure the time taken by func() and add that to zjit_compile_time.
```

zjit/src/gc.rs
```diff
@@ -80,7 +82,29 @@ pub extern "C" fn rb_zjit_iseq_mark(payload: *mut c_void) {
...
+        // else has an overlapping &mut IseqPayload.
+        unsafe {
+            rb_assert_holding_vm_lock();
+            &mut *(payload as *mut IseqPayload)
+        }
+    };
+    with_time_stat(gc_time_ns, || iseq_update_references(payload));
+}
 
+fn iseq_mark(payload: &IseqPayload) {
```
tekknolagi, line 107: What do you think about also including code marking time in the stats?

tekknolagi: oops, sorry, it's in the caller

PR #14160, opened 2025-08-08
Fix id2ref table build when GC in progress

jhawthorn:
Previously, if GC was in progress when we're initially building the id2ref table, it could see the empty table and then crash when trying to remove ids from it. This commit fixes the bug by only publishing the table after GC is done.


byroot:
Thank you!

PR #6597, opened 2022-10-20
YJIT: Lazily enable YJIT after prelude

k0kubun:
Currently, YJIT compiles code for prelude as well. But some code used by prelude may not be used ever again after prelude, which could bring less localized code and unnecessary pressure on code GC.

JITing prelude seems just useless if it doesn't contribute to speeding up prelude. And the cost of JITing prelude seems to outweigh the JITed code's speedup. So I'd like to propose not JITing prelude.

### Before
```
$ time ruby -v --yjit -e ''
ruby 3.2.0dev (2022-10-20T00:45:36Z master 2e3b764ad1) +YJIT [arm64-darwin21]
ruby -v --yjit -e ''  0.05s user 0.02s system 91% cpu 0.071 total
```

### After
```
$ time ruby -v --yjit -e ''
ruby 3.2.0dev (2022-10-20T00:45:54Z yjit-lazy-init f60a1b4ce3) +YJIT [arm64-darwin21]
ruby -v --yjit -e ''  0.04s user 0.01s system 90% cpu 0.058 total
```

I've tried both multiple times, and this PR was reliably faster.

PR #14171, opened 2025-08-11
YJIT, ZJIT: Fix JITs compiling prelude

k0kubun:
This PR resurrects the behavior of https://github.com/ruby/ruby/pull/6597 that JITs are enabled after prelude.

Prelude often runs code that is never used after boot. To avoid wasting memory and compilation time for such code (and tidy up the output of `--*jit-dump-disasm`), we enable JITs after prelude. For some reason, https://github.com/ruby/ruby/pull/13650 put `ruby_init_prelude` after JIT inits. The PR or the commit message https://github.com/ruby/ruby/commit/96a0c2065a95d076978de41e8bfacbd19858d0bb didn't explain why, so I thought it was oversight and thus fixed it.

### DIFF
--- a/ruby.c
+++ b/ruby.c
@@ -1819,8 +1819,10 @@ ruby_opt_init(ruby_cmdline_options_t *opt)
 
     if (rb_namespace_available())
         rb_initialize_main_namespace();
+    rb_namespace_init_done();
+    ruby_init_prelude();
 
-    // Initialize JITs after prelude because JITing prelude is typically not optimal.
+    // Initialize JITs after ruby_init_prelude() because JITing prelude is typically not optimal.
 #if USE_YJIT
     rb_yjit_init(opt->yjit);
 #endif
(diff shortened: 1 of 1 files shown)


ruby.c
```diff
@@ -1819,8 +1819,10 @@ ruby_opt_init(ruby_cmdline_options_t *opt)
 
     if (rb_namespace_available())
         rb_initialize_main_namespace();
+    rb_namespace_init_done();
```
st0012, line 1822: For anyone curious, this function only sets a flag:

```c
void
rb_namespace_init_done(void)
{
    ruby_namespace_init_done = true;
}
```

So moving it before JIT initialization is probably fine.

byroot: What it does is that any class defined before this point use a larger object slot, so that it can have multiple `classext_t`.

And yes, I think moving it here is fine.

PR #14172, opened 2025-08-11
[DOC] ZJIT: Add ZJIT to autolink_excluded_words

st0012:
This tells RDoc to not automatically link to the `ZJIT` module so we don't need to keep escaping the word ZJIT in the documentation/comments.

### DIFF
--- a/.rdoc_options
+++ b/.rdoc_options
@@ -20,5 +20,6 @@ autolink_excluded_words:
 - Ruby
 - Set
 - ZJIT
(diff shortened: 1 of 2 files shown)


XrXr (approved):
TIL

k0kubun (approved):
Oh wow, nice. Could you do that for YJIT too?

st0012:
I've added a commit for YJIT

PR #14173, opened 2025-08-11
CI: Surface Rust warnings on PRs that touch any Rust code

XrXr:
Idea is, if you create PRs touching Rust code often, you probably don't
mind being shown new warnings introduced in new Rust releases. They've
been trivial to fix and this will help us stay on top of them.

Closes https://github.com/Shopify/ruby/pull/693

### DIFF
--- a/.github/workflows/rust-warnings.yml
+++ b/.github/workflows/rust-warnings.yml
@@ -1,0 +1,53 @@
+# Surface Rust warnings on PRs that touch any Rust code.
+# Not a required check so we never block people over new warnings
+# that might come from a new Rust version being released.
+name: Rust warnings
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    paths:
+      - '**.rs'
+      - '!**.inc.rs'
+  merge_group:
+
+concurrency:
+  group: ${{ github.workflow }} / ${{ startsWith(github.event_name, 'pull') && github.ref_name || github.sha }}
+  cancel-in-progress: ${{ startsWith(github.event_name, 'pull') }}
+
+permissions:
+  contents: read
+
+jobs:
+  make:
+    env:
+      GITPULLOPTIONS: --no-tags origin ${{ github.ref }}
+
+    runs-on: ubuntu-24.04
+
+    if: >-
+      ${{!(false
+      || contains(github.event.head_commit.message, '[DOC]')
+      || contains(github.event.head_commit.message, 'Document')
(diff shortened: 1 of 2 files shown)


XrXr:
> Rust PRs will have a failed CI step if they trigger any warnings.
This helps us stay on top of warnings from new Rust releases and
also ones we accidentally write.
>
> Fix a typo for demo, since this only runs when Rust files are changed.

Since we're clean on warnings on `master` right now, the check passes. See https://github.com/Shopify/ruby/pull/693 for how it looks like when the check fails.

st0012 (approved):

.github/workflows/rust-lints-on-prs.yml
k0kubun, line 1: Could we call it `.github/workflows/rust-warnings.yml` (or `rust-warnings-on-prs.yml`)? I would recall something like `cargo clippy` (which is more of a stylistic preference and less important than fixing Rust warnings) from the name `rust-lints-on-prs.yml`, but it's actually done in a different workflow.

XrXr: Hmm I have a preference for "lints" since that's what rustc and cargo calls them -- the compiler lints in addition to compiling. https://doc.rust-lang.org/rustc/lints/index.html

~~It looks like we lost the clippy check so maybe I can add it back in the same file. Would that make the name sound better?~~

nvm, keeping the file name and job name consistent is important.

k0kubun: It's not lost, still in `yjit-ubuntu.yml`. But yeah, it would.

k0kubun: Oh, you changed it 👌

.github/workflows/rust-warnings.yml
```diff
@@ -0,0 +1,53 @@
...
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install Rust
+        run: rustup default beta
+
+      - name: Rust warnings
+        run: |
+          set -euo pipefail
+          cargo check --quiet --all-features --message-format=json \
```
k0kubun, line 49: I wonder if there's a way to extract not only warnings but also errors with jq. It just silently fails right now.

Screenshot 2025-08-15 at 09 14 35

https://github.com/ruby/ruby/actions/runs/16994039992/job/48180380659

PR #14175, opened 2025-08-11
ZJIT: Add --zjit-exec-mem-size

k0kubun:
This PR adds `--zjit-exec-mem-size` option and stops hard-coding the max size of the code region.

I wanted to add `--zjit-mem-size` at first, but `#[global_allocator]` to implement that conflicts with YJIT on the combo build. So I just added `--zjit-exec-mem-size` for now.

### DIFF
--- a/zjit/src/options.rs
+++ b/zjit/src/options.rs
@@ -176,9 +176,12 @@ fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
 
         ("mem-size", _) => match opt_val.parse::<usize>() {
             Ok(n) => {
+                // Reject 0 or too large values that could overflow.
+                // The upper bound is 1 TiB but we could make it smaller.
                 if n == 0 || n > 1024 * 1024 {
                     return None
                 }
+
                 // Convert from MiB to bytes internally for convenience
                 options.exec_mem_bytes = n * 1024 * 1024;
             }


zjit/src/stats.rs
```diff
@@ -113,3 +109,8 @@ pub fn with_time_stat<F, R>(counter: Counter, func: F) -> R where F: FnOnce() ->
     incr_counter(counter, nanos as u64);
     ret
 }
+
```
tekknolagi, line 112: nit: slightly neater diff if the function doesn't move

zjit/src/options.rs
```diff
@@ -163,6 +174,17 @@ fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
     match (opt_name, opt_val) {
         ("", "") => {}, // Simply --zjit
 
+        ("mem-size", _) => match opt_val.parse::<usize>() {
+            Ok(n) => {
+                if n == 0 || n > 1024 * 1024 {
```
tekknolagi, line 181: Can you add a comment about what's going on here because it's not clear to me

tekknolagi (approved):

zjit/src/stats.rs
k0kubun: It is a deliberate move. I decided `zjit_alloc_size` (the implementation for one of the counters) should not sit in between `make_counters!`/`incr_counter` (the core of the counter) and `rb_zjit_stats` (ditto). `zjit_alloc_size` is closer to the category of `with_time_stat`, which is an implementation for specific stats.

Part of the reason why I moved it here was because I had to add `SystemAllocator` and its `impl` next to `zjit_alloc_size` for `--zjit-mem-size`, and that no longer applies if we take the `jit` crate route. But I still prefer this ordering for the reason explained above.

zjit/src/options.rs
k0kubun: done https://github.com/ruby/ruby/pull/14175/commits/d9b4d979b7dc1d1ea95d1757d49179ecda61fb34

YJIT had `if n == 0 || n > 2 * 1024 * 1024 {` (and no comment) but I thought `2 *` isn't really needed, so I simplified that part at least. 1 GiB _could_ have a customer in weird edge cases, so I kept `1024 *` to make it larger than that. But I wouldn't object to lowering it to something smaller for ZJIT, even the 1 GiB limit. Maybe I'll change it so when adding `--zjit-mem-size` before 3.5 is released.

k0kubun: It's not only a check for overflow, but also a nice check (given my version is `n > 1MiB`) to say "you're writing the size in bytes, not in MiB".

PR #14176, opened 2025-08-11
ZJIT: Use asm_ccall macro in `gen_send_without_block`

st0012:


k0kubun (approved):

XrXr:
I don't know if we want the double comment? "call #{} with dynamic dispatch" seems descriptive enough.

k0kubun:
Oh true. I missed the diff folded in the PR view. We would like to see the method name, so the current code looks good as is.

st0012:
@XrXr Oh good call. I didn't notice that there's a `asm_comment!` a few lines above.
My intention for this change is more for consistency as this is the only place that we call a specific C function without using `asm_ccall!`

st0012:
Yeah if `asm_ccall`'s main purpose is to enforce comments on the C function, it's not helping too much:

```
  # call #frozen? with dynamic dispatch
  # call rb_vm_opt_send_without_block
  0x11e520170: mov x2, #0xfa0
  0x11e520174: movk x2, #0x4a91, lsl #16
  0x11e520178: movk x2, #1, lsl #32
  0x11e52017c: mov x1, x19
  0x11e520180: mov x0, x20
  0x11e520184: mov x16, #0x8ec0
  0x11e520188: movk x16, #0x290, lsl #16
  0x11e52018c: movk x16, #1, lsl #32
  0x11e520190: blr x16
```

Closing then

PR #14177, opened 2025-08-12
ZJIT: Avoid splitting add_into/sub_into

k0kubun:
This PR addresses https://github.com/ruby/ruby/pull/14154#discussion_r2268083538.

We were splitting `asm.add_into` to two instructions. By fixing it, this PR also updates `asm.sub_into` to use an implementation that looks similar to `asm.add_into`.

## Before

`add_into` in concatstrings looked like this:

```asm
  # restore C stack pointer
  0x652fd19ee1b8: mov rdi, rsp
  0x652fd19ee1bb: add rdi, 0x20
```

## After

It's now:

```asm
  # restore C stack pointer
  0x5629f326e1b8: add rsp, 0x20
```

The instructions generated by `sub_into` in concatstrings stayed the same because merging sub + mov works fine.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1889,9 +1889,8 @@ impl Assembler {
         out
     }
 
-    pub fn add_into(&mut self, left: Opnd, right: Opnd) -> Opnd {
+    pub fn add_into(&mut self, left: Opnd, right: Opnd) {
         self.push_insn(Insn::Add { left, right, out: left });
-        left
     }
 
     #[must_use]
@@ -2233,10 +2232,8 @@ impl Assembler {
         out
     }
 
-    pub fn sub_into(&mut self, left: Opnd, right: Opnd) -> Opnd {
-        let out = self.sub(left, right);
-        self.mov(left, out);
-        out
+    pub fn sub_into(&mut self, left: Opnd, right: Opnd) {
+        self.push_insn(Insn::Sub { left, right, out: left });
     }
 
     #[must_use]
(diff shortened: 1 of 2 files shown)


zjit/src/backend/lir.rs
```diff
@@ -1889,9 +1889,8 @@ impl Assembler {
         out
     }
 
-    pub fn add_into(&mut self, left: Opnd, right: Opnd) -> Opnd {
+    pub fn add_into(&mut self, left: Opnd, right: Opnd) {
```
tekknolagi, line 1892: I don't love that this breaks SSA if we use a vreg operand as an argument. Can we require it only take memory/phyreg? (here and below)

k0kubun: I changed them to only take `Opnd::Reg` for now https://github.com/ruby/ruby/pull/14177/commits/a43e62e9eaa1304abc61f67cc6ac7826e8bad2ed since it's the only expected usage of them.

Btw if you want LIR to be SSA, you need to fix `cpop_into`, `lea_into`, `load_into`, `mov`, and `store` as well.

tekknolagi: Thank you for the change. Didn't know that. I thought LIR was SSA...

k0kubun: Almost. You can make it happen with just ^. We just haven't come up with anything that needs the guarantee yet.

tekknolagi: Sketching in https://github.com/ruby/ruby/pull/14188

PR #14179, opened 2025-08-12
set.c: Store `set_table->bins` at the end of `set_table->entries`

byroot:
Followup: https://github.com/ruby/ruby/pull/14134

This saves one pointer in `struct set_table`, which would allow `Set` objects to still fit in 80B TypedData slots even if RTypedData goes from 32B to 40B large.

The existing set benchmark seem to show this doesn't have a very significant impact. Smaller sets are a bit faster, larger sets a bit slower.

It seem consistent over multiple runs, but it's unclear how much of that is just error margin.

```
compare-ruby: ruby 3.5.0dev (2025-08-12T02:14:57Z master 428937a536) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-12T07:22:26Z set-entries-bounds da30024fdc) +YJIT +PRISM [arm64-darwin24]
warming up........

|new_0                    |     15.459M|   15.823M|
|new_10                   |      3.484M|    3.574M|
|                         |           -|     1.03x|
|new_100                  |    546.992k|  564.679k|
|                         |           -|     1.03x|
|new_1000                 |     49.391k|   48.169k|
|                         |       1.03x|         -|
|aref_0                   |     18.643M|   19.350M|
|                         |           -|     1.04x|
|aref_10                  |      5.941M|    6.006M|
|                         |           -|     1.01x|
|aref_100                 |    822.197k|  814.219k|
|                         |       1.01x|         -|
|aref_1000                |     83.230k|   79.411k|
|                         |       1.05x|         -|
```

### DIFF
--- a/internal/set_table.h
+++ b/internal/set_table.h
@@ -15,13 +15,16 @@ struct set_table {
     const struct st_hash_type *type;
     /* Number of entries currently in the table.  */
     st_index_t num_entries;
-    /* Array of bins used for access by keys.  */
-    st_index_t *bins;
+
     /* Start and bound index of entries in array entries.
        entries_starts and entries_bound are in interval
        [0,allocated_entries].  */
     st_index_t entries_start, entries_bound;
-    /* Array of size 2^entry_power.  */
+
+    /**
+     * Array of size 2^entry_power.
+     * Followed by st_index_t *bins, Array of bins used for access by keys.
+     */
     set_table_entry *entries;
 };
 
--- a/set.c
+++ b/set.c
@@ -139,7 +139,6 @@ set_mark(void *ptr)
 static void
 set_free_embedded(struct set_object *sobj)
 {
-    free((&sobj->table)->bins);
     free((&sobj->table)->entries);
 }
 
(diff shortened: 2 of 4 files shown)


jeremyevans (approved):
Great work! I think it would be worth considering this approach for st_table, and potentially shrinking hash from 160 bytes to 80 bytes (reducing AR table size from 8 to 3). There are far more hashes than sets, and I would guess that the majority of hashes are likely to have 0-3 entries. Not sure if the potential performance hit for 4-8 entry hashes is worth the memory savings. Hopefully the benchmarks could help determine that.

byroot:
> I think it would be worth considering this approach for st_table,

AFAIK, `RHash` is only `24B`, so `st_table` can already fit in a `80B` slot AFAIK.

@peterzhu2118 do you remember if there is a reason hashes default to 160B slots?

peterzhu2118:
> do you remember if there is a reason hashes default to 160B slots?

Yes, because Hashes using AR tables are 160 bytes in size.

byroot:
> because Hashes using AR tables are 160 bytes in size.

Yes we know that, but they could be whatever size we want. I suppose they were this size before WVA?

PR #14098, opened 2025-08-04
[DOC] New .md file to replace doc/globals.rdoc

BurdetteLamar:
Besides changing format from .rdoc to .md, adds summary tables.

I've avoided changing actual text in the reformatting -- changing just markup to markdown.  (More can be done with the text, but I don't want to muddy the waters in this reformatting PR.)

PR #14187, opened 2025-08-12
[DOC] Warn about missing rdoc-ref target

st0012:
RDoc resolves `rdoc-ref` when generating links and with this option it can emit warnings on missing targets, like:

```
exceptions_md.html: `rdoc-ref:Exceptione` can't be resolved for `Built-In Exception Class Hierarchy`
```

Currently there's no missing `rdoc-ref` but it'd be helpful capturing new ones.

### DIFF
--- a/.rdoc_options
+++ b/.rdoc_options
@@ -23,3 +23,4 @@ autolink_excluded_words:
 - YJIT
 
 canonical_root: https://docs.ruby-lang.org/en/master
+warn_missing_rdoc_ref: true


nobu:
Isn’t it enabled now?
https://github.com/ruby/ruby/pull/14098#issuecomment-3175248506

st0012:
Ah yes it is. It's default to `true` in RDoc now but I forgot to update its documentation 😅

st0012:
Also, I'm happy to see it catching issues 😄

PR #14188, opened 2025-08-12
ZJIT: Make sure output operands are not VRegs

tekknolagi:
Make LIR SSA.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1684,6 +1684,7 @@ impl Assembler {
     }
 
     pub fn cpop_into(&mut self, opnd: Opnd) {
+        assert!(matches!(opnd, Opnd::Reg(_)), "Destination of cpop_into must be a register, got: {opnd:?}");
         self.push_insn(Insn::CPopInto(opnd));
     }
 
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
> Make LIR SSA.

I mean this PR doesn't because there's still `load_into` that allows writes to VReg, ~but we do write to VReg using `load_into` for , so I agree we should merge this first and then work on `load_into` separately if we want to.~

Huh. We return `Opnd::Reg` from `param_opnd` and we `load_into` that register, and `VReg` is only used by its wrapper `gen_param`. So I guess we can ban `VReg` from `load_into` now?

k0kubun:
I just updated the comment ^. I think we can do:

```diff
     pub fn load_into(&mut self, dest: Opnd, opnd: Opnd) {
-        assert!(matches!(dest, Opnd::Reg(_) | Opnd::VReg{..}), "Destination of load_into must be a register");
+        assert!(matches!(dest, Opnd::Reg(_)), "Destination of load_into must be a register, got: {dest:?}"); 
```

as well.

zjit/src/backend/lir.rs
```diff
@@ -2160,7 +2162,7 @@ impl Assembler {
     }
 
     pub fn load_into(&mut self, dest: Opnd, opnd: Opnd) {
-        assert!(matches!(dest, Opnd::Reg(_) | Opnd::VReg{..}), "Destination of load_into must be a register");
+        assert!(matches!(dest, Opnd::Reg(_), "Destination of load_into must be a register");
```
k0kubun:
```suggestion
        assert!(matches!(dest, Opnd::Reg(_)), "Destination of load_into must be a register, got: {dest:?}");
```

tekknolagi: ty

PR #14189, opened 2025-08-12
[DOC] Tweaks for GC.count

BurdetteLamar:


gc.rb
```diff
@@ -102,9 +102,14 @@ def self.stress=(flag)
   end
 
   # call-seq:
-  #    GC.count -> Integer
+  #   self.count -> integer
+  #
+  # Returns the number of times garbage collection has occurred in the current process:
```
peterzhu2118: I think this (and the original) description is kind of misleading since it seems to suggest that forking a new process resets this counter (it does not). Maybe something like this is less ambiguous:

```suggestion
  # Returns the total number of times garbage collection has occurred:
```

PR #14192, opened 2025-08-12
Add missing write barriers to Random

jhawthorn:
We recently converted this to be WB_PROTECTED, but missed some locations the seed was set.

This is a little bit confusing because as far as I can tell some of the seeds are set in Ractor local storage, and others on actual objects.

### DIFF
--- a/random.c
+++ b/random.c
@@ -142,18 +142,21 @@ static const rb_random_interface_t random_mt_if = {
 };
 
 static rb_random_mt_t *
-rand_mt_start(rb_random_mt_t *r)
+rand_mt_start(rb_random_mt_t *r, VALUE obj)
 {
     if (!genrand_initialized(&r->mt)) {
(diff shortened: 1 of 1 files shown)


random.c
```diff
@@ -142,18 +142,21 @@ static const rb_random_interface_t random_mt_if = {
 };
 
 static rb_random_mt_t *
-rand_mt_start(rb_random_mt_t *r)
+rand_mt_start(rb_random_mt_t *r, VALUE obj)
 {
     if (!genrand_initialized(&r->mt)) {
         r->base.seed = rand_init(&random_mt_if, &r->base, random_seed(Qundef));
+        if (obj) {
+            RB_OBJ_WRITTEN(obj, Qundef, r->base.seed);
```
peterzhu2118, line 150: Do you know why this is needed? AFAICT `rand_init` doesn't seem to be touching `seed` at all.

PR #14195, opened 2025-08-12
ZJIT: Avoid compiling failed ISEQs repeatedly

k0kubun:
This PR lets `function_stub_hit` exit quickly when the ISEQ has failed to compile before, instead of attempting it repeatedly.

I updated `start_ptr: Option<CodePtr>` to `status: IseqStatus` to avoid consuming more memory. We should also fix the compiler to emit a side-exit when there's an unsupported instruction so that we exit more efficiently without calling the Rust function, but it's for another PR.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1377,6 +1377,8 @@ c_callable! {
             }
 
             // If we already know we can't compile the ISEQ, fail early without cb.mark_all_executable().
+            // TODO: Alan thinks the payload status part of this check can happen without the VM lock, since the whole
+            // code path can be made read-only. But you still need the check as is while holding the VM lock in any case.
             let cb = ZJITState::get_code_block();
             let payload = get_or_create_iseq_payload(iseq);
             if cb.has_dropped_bytes() || payload.status == IseqStatus::CantCompile {


zjit/src/codegen.rs
```diff
@@ -1356,27 +1362,38 @@ macro_rules! c_callable {
...
                 let cfp = unsafe { get_ec_cfp(ec) };
                 let pc = unsafe { rb_iseq_pc_at_idx(iseq, 0) }; // TODO: handle opt_pc once supported
                 unsafe { rb_set_cfp_pc(cfp, pc) };
                 unsafe { rb_set_cfp_sp(cfp, sp) };
+            }
 
+            // If we already know we can't compile the ISEQ, fail early without cb.mark_all_executable().
+            let cb = ZJITState::get_code_block();
+            let payload = get_or_create_iseq_payload(iseq);
+            if cb.has_dropped_bytes() || payload.status == IseqStatus::CantCompile {
```
XrXr, line 1384: I think the payload status part of this check can happen without the VM lock, since the whole code path can be made read-only. But you still need the check as is while holding the VM lock in any case. Ractor concerns are far out, though. Could you leave a TODO?

XrXr (approved):

tekknolagi:
Does this speed up some of the issues on yjit-bench?

st0012 (approved):

zjit/src/codegen.rs
k0kubun: Left a TODO, quoting your comment as is https://github.com/ruby/ruby/pull/14195/commits/28f9f3c027a7de586f8c138a0fd92fc7ba4c0601.

TBH I didn't understand what you said though. I interpreted your comment like "you can check `payload.status == IseqStatus::CantCompile` without the VM lock, but you still need the check with the VM lock". Can you elaborate more on when you can check it without the VM lock and when it needs to be checked with the VM lock?

Also, as explained above `fn function_stub_hit`, the primary use of this path is for `cb.has_dropped_bytes()` like YJIT (once we implement side exits for unsupported instructions). To get `cb`, we'd still need the VM lock (we could have an atomic global variable separately from `cb` if needed though).

k0kubun:
> Does this speed up some of the issues on yjit-bench?

Oh yeah they're waaay faster. I'll update the issue.

zjit/src/codegen.rs
XrXr: Both checks are necessary. If you find `payload.status == IseqStatus::CantCompile` outside the lock, you can exit without taking the lock. But if you do contend for the lock, by the time you get it you might've lost the race so you repeat the same check once you have the lock.

k0kubun: Ah, that makes sense. Thanks.

PR #14196, opened 2025-08-12
ZJIT: Fix `ObjToString` rewrite

st0012:
Currently, the rewrite for `ObjToString` always replaces it with a `SendWithoutBlock(to_s)` instruction when the receiver is not a string literal. This is incorrect because it calls `to_s` on the receiver even if it's already a string.

This change fixes it by:
- Avoiding the `SendWithoutBlock(to_s)` rewrite
- Implement codegen for `ObjToString`

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -1343,6 +1343,71 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 2
   end
 
+  def test_objtostring_calls_to_s_on_non_strings
+    assert_compiles '["foo", "foo"]', %q{
+      results = []
+
+      class Foo
+        def to_s
+          "foo"
+        end
+      end
+
+      def test(str)
+        "#{str}"
+      end
+
+      results << test(Foo.new)
+      results << test(Foo.new)
+
+      results
+    }
+  end
+
+  def test_objtostring_rewrite_does_not_call_to_s_on_strings
+    assert_compiles '["foo", "foo"]', %q{
+      results = []
+
+      class String
+        def to_s
+          "bad"
+        end
+      end
+
+      def test(foo)
+        "#{foo}"
+      end
+
+      results << test("foo")
+      results << test("foo")
+
+      results
+    }
+  end
+
+  def test_objtostring_rewrite_does_not_call_to_s_on_string_subclasses
+    assert_compiles '["foo", "foo"]', %q{
+      results = []
+
+      class StringSubclass < String
+        def to_s
+          "bad"
+        end
+      end
+
+      foo = StringSubclass.new("foo")
+
+      def test(str)
+        "#{str}"
+      end
+
+      results << test(foo)
+      results << test(foo)
+
+      results
+    }
+  end
+
   def test_string_bytesize_with_guard
     assert_compiles '5', %q{
       def test(str)
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -378,6 +378,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::AnyToString { val, str, state } => gen_anytostring(asm, opnd!(val), opnd!(str), &function.frame_state(*state))?,
         Insn::Defined { op_type, obj, pushval, v, state } => gen_defined(jit, asm, *op_type, *obj, *pushval, opnd!(v), &function.frame_state(*state))?,
         &Insn::IncrCounter(counter) => return Some(gen_incr_counter(asm, counter)),
+        Insn::ObjToString { val, cd, state, .. } => gen_objtostring(jit, asm, opnd!(val), *cd, &function.frame_state(*state))?,
         Insn::ArrayExtend { .. }
         | Insn::ArrayMax { .. }
         | Insn::ArrayPush { .. }
@@ -386,7 +387,6 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         | Insn::FixnumMod { .. }
         | Insn::HashDup { .. }
         | Insn::NewHash { .. }
-        | Insn::ObjToString { .. }
         | Insn::Send { .. }
         | Insn::StringIntern { .. }
         | Insn::Throw { .. }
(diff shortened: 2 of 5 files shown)


zjit/src/codegen.rs
```diff
@@ -439,6 +439,24 @@ fn gen_get_ep(asm: &mut Assembler, level: u32) -> Opnd {
...
+
+    // TODO: Specialize for immediate types
+    // Call rb_vm_objtostring(iseq, recv, cd)
+    let ret = asm_ccall!(asm, rb_vm_objtostring, iseq_opnd, val, (cd as usize).into());
+
+    // TODO: Call `to_s` on the receiver if rb_vm_objtostring returns Qundef
+    // Need to replicate what CALL_SIMPLE_METHOD does
+    asm_comment!(asm, "side-exit if rb_vm_objtostring returns Qundef");
+    asm.cmp(ret, Qundef.into());
+    asm.je(side_exit(jit, state, ObjToStringFallback)?);
```
st0012, line 461: I'm not sure how to implement the conditional "call to_s on val when we got Qundef" without generating a lot of assembly (e.g. calling `gen_send_without_block` and `je` to it). So I decided to side exit in that case for now. This will at least fix the bug we're seeing
Thoughts?

test/ruby/test_zjit.rb
```diff
@@ -1319,6 +1319,48 @@ def test(mod)
...
+      end
+
+      results << test(Foo.new)
+      results << test(Foo.new)
+
+      results
+    }
+  end
+
+  def test_objtostring_rewrite_does_not_call_to_s_on_strings
```
tekknolagi, line 1367: Can you please add a test for string subclass?

zjit/src/codegen.rs
k0kubun: You can profile the receiver object, guard on the type (and side-exit), and generate only what's supposed to be executed. This is what YJIT does.

---
> So I decided to side exit in that case for now.

I also agree it's fine for now, but if you're interested in fixing it in this PR, I would suggest:

Profile the receiver type. Generate one of the following HIRs:

1. For T_STRING: `GuardType(String)`
2. (Optional: For T_SYMBOL/T_MODULE/T_CLASS/T_NIL/T_TRUE/T_FALSE/T_FIXNUM: `GuardType(type)` → `ObjToString` that side-exits on Qundef like the current PR)
3. For others: `GuardTypeNot(String)` (new HIR) → `SendWithoutBlock`

Behavior-wise, (1) + (3) should fix any issues. (2) is more of an optional optimization that you can leverage your current patch for. `GuardTypeNot(String)` shouldn't be too hard to implement if you handle only `T_STRING` (and bail out otherwise) for now.

k0kubun:
~Could we just guard that the receiver is not T_STRING and leave a method call handling in the JIT code? If the fallback case is a method call, it's easy to understand that you can optimize it like a regular method call, but once you delegate the method call to a C function, you cannot easily touch it. When we add a specialized codegen for `to_s` C functions, we'd like to optimize it using them.~

~That's the bugfix, but ideally you should also profile the receiver object and generate different HIR (guard that it's T_STRING and do nothing else) to avoid exiting because "the receiver is not T_STRING".~

I updated the comment at https://github.com/ruby/ruby/pull/14196#discussion_r2271452248 and merged my suggestions there.

k0kubun:
If it's too complicated to figure out the right HIR for it now, I guess it's fine to start with this and work on it in a separate PR.

Could you file another issue to optimize `ObjToString` further as a reminder? We could probably optimize away `gen_prepare_non_leaf_call` if we guard on types, and avoid side-exiting on types that are not handled in `vm_objtostring`.

zjit/src/codegen.rs
XrXr: I'm not a fan of hiding a send in a non-send HIR instruction either due to optimizer visibility etc. One way around this is to have `insn::ObjToString` only call rb_vm_objtostring() and build the if-else logic into HIR during parsing. This way we have a regular send in the else BB, and the optimizer should still be able to handle folding away string input situations.

To go faster, see Kokubun's suggestion on how to speculate.

It still sounds quite involved to me though. I think side exiting for now is fine and we can make an issue and optimize this later.

zjit/src/codegen.rs
```diff
@@ -439,6 +439,24 @@ fn gen_get_ep(asm: &mut Assembler, level: u32) -> Opnd {
     ep_opnd
 }
 
+fn gen_objtostring(jit: &mut JITState, asm: &mut Assembler, val: Opnd, cd: *const rb_call_data, state: &FrameState) -> Option<Opnd> {
+    gen_prepare_non_leaf_call(jit, asm, state)?;
```
k0kubun, line 449: I couldn't figure out which part of `rb_vm_objtostring` calls a method. Could you point out which part of it is not leaf?

XrXr: When `recv.singleton_class?` `rb_mod_to_s()` calls can call inspect on the attached object. Diabolical.

k0kubun: Oh I realized there's `rb_inspect` in the non-`rb_any_to_s` branch, which does `rb_funcall`. So yeah, `rb_vm_objtostring` is not leaf.

It doesn't matter at this point, but I didn't get what's not leaf about `rb_any_to_s`. Does `rb_sprintf("#<%"PRIsVALUE":%p>", cname, (void*)obj)` dispatch a method through `PRIsVALUE`? (I don't know which code formats `"%"PRIsVALUE`)

XrXr: Yes, in general `PRIsVALUE` can call `to_s` by going through `PRI_EXTRA_MARK` and ruby__sfvextra(). Only for non-T_STRINGs does ruby__sfvextra() make a call, though, and it looks like `cname` is always a string here. Though there is also an encoding check in there so the print might rise.

zjit/src/codegen.rs
st0012: Thank you for your suggestions. I created https://github.com/Shopify/ruby/issues/701 as a follow up.

For now I'll keep side exit so we can avoid the incorrect behaviour.

test/ruby/test_zjit.rb
st0012: Added

PR #14198, opened 2025-08-12
ZJIT: Two combo build fixes

XrXr:
- **ZJIT: Look for RUBY_ZJIT_ENABLE in combo build with YJIT**
- **ZJIT: Fix assert failure on JIT combo build**

### DIFF
--- a/version.c
+++ b/version.c
@@ -66,11 +66,6 @@ const int ruby_api_version[] = {
 #else
 #define ZJIT_DESCRIPTION " +ZJIT"
 #endif
-#if USE_ZJIT
-#define JIT_DESCRIPTION ZJIT_DESCRIPTION
-#else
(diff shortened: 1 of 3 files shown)


version.c
```diff
@@ -200,6 +195,8 @@ rb_ruby_default_parser_set(ruby_default_parser_enum parser)
 static void
 define_ruby_description(const char *const jit_opt)
 {
+#define JIT_DESCRIPTION YJIT_DESCRIPTION ZJIT_DESCRIPTION
```
k0kubun, line 205: I was wondering if we could reduce the number of wasted bytes in the buffer, but even if we somehow pick the longer description here at preprocessor time, when a JIT with a shorter description is enabled, we'd have to waste some bytes anyway. So I guess this is fine.

PR #14200, opened 2025-08-13
ZJIT: Compile toregexp

composerinteralia:
`toregexp` is fairly similar to `concatstrings`, so this commit extracts a helper for pushing and popping operands on the native stack.

There's probably opportunity to move some of this into lir (e.g. Alan suggested a push_many that could use STP on ARM to push 2 at a time), but I might save that for another day.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -346,6 +346,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::StringConcat { strings, .. } if strings.is_empty() => return None,
         Insn::StringConcat { strings, state } => gen_string_concat(jit, asm, opnds!(strings), &function.frame_state(*state)),
         Insn::StringIntern { val, state } => gen_intern(asm, opnd!(val), &function.frame_state(*state)),
+        Insn::ToRegexp { opt, values, state } => gen_toregexp(jit, asm, *opt, opnds!(values), &function.frame_state(*state)),
         Insn::Param { idx } => unreachable!("block.insns should not have Insn::Param({idx})"),
         Insn::Snapshot { .. } => return Some(()), // we don't need to do anything for this instruction at the moment
         Insn::Jump(branch) => no_output!(gen_jump(jit, asm, branch)),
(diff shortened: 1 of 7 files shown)


zjit/src/hir.rs
```diff
@@ -2823,6 +2842,14 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let insn_id = fun.push_insn(block, Insn::StringConcat { strings, state: exit_id });
                     state.stack_push(insn_id);
                 }
+                YARVINSN_toregexp => {
+                    let opt = get_arg(pc, 0).as_usize();
```
tekknolagi, line 2907: Can we call this something more descriptive and also store it in a way that's more conducive to examination than a raw int?

zjit/src/hir.rs
```diff
@@ -2823,6 +2842,14 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                    let count = get_arg(pc, 1).as_usize();
+                    let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    let values = state.stack_pop_n(count as usize)?;
```
tekknolagi: you shouldn't need to as-usize twice

zjit/src/hir.rs
tekknolagi: Or at least, if we store it as a raw int---which we'll need to pass into codegen later---print it out more nicely as OR-ed flags?

zjit/src/codegen.rs
```diff
@@ -1457,43 +1458,66 @@ pub fn gen_stub_exit(cb: &mut CodeBlock) -> Option<CodePtr> {
...
-    let n = strings.len();
-
-    // concatstrings shouldn't have 0 strings
-    // If it happens we abort the compilation for now
-    if n == 0 {
-        return None;
-    }
-
-    gen_prepare_non_leaf_call(jit, asm, state)?;
+fn push_opnds(jit: &mut JITState, asm: &mut Assembler, opnds: &[Opnd]) -> lir::Opnd {
```
k0kubun: I would like the `gen_` prefix to be kept. The naming consistency allows us to identify which functions are side effect-free (doesn't generate instructions) easily.

```suggestion
fn gen_push_opnds(jit: &mut JITState, asm: &mut Assembler, opnds: &[Opnd]) -> lir::Opnd {
```

zjit/src/codegen.rs
```diff
@@ -1457,43 +1458,66 @@ pub fn gen_stub_exit(cb: &mut CodeBlock) -> Option<CodePtr> {
...
         );
     }
 
-    let first_string_ptr = asm.lea(Opnd::mem(64, NATIVE_BASE_PTR, -total_offset_from_base));
-
-    let result = asm_ccall!(asm, rb_str_concat_literals, n.into(), first_string_ptr);
+    asm.lea(Opnd::mem(64, NATIVE_BASE_PTR, -total_offset_from_base))
+}
 
+fn pop_opnds(asm: &mut Assembler, opnds: &[Opnd]) {
```
k0kubun: ditto

```suggestion
fn gen_pop_opnds(asm: &mut Assembler, opnds: &[Opnd]) {
```

zjit/src/hir.rs
k0kubun: > Can we call this something more descriptive

I want this to be consistent with whatever name insns.def uses (currently `opt`). Since we have to compare the interpreter implementation and the ZJIT implementation for checking compatibility, it'd be harder to maintain if the names have inconsistency.

If you need a description, we should add a comment. If you do need a rename of the variable, please rename that in insns.def first.

tekknolagi: Sure, fair.

zjit/src/hir.rs
composerinteralia: Oops! That happened when I merged the concatstring stuff into my existing branch.

zjit/src/hir.rs
composerinteralia: > I want this to be consistent with whatever name insns.def uses

That's the rule I've been following for naming things too. 👍🏻 

> print it out more nicely as OR-ed flags?

I added some fancier printing (I wasn't printing the options at all before), plus a comment about what opt is. Does that seem sufficient?

zjit/src/hir.rs
```diff
@@ -689,6 +692,24 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
...
+            Insn::ToRegexp { values, opt, .. } => {
+                write!(f, "ToRegexp")?;
+                let mut prefix = " ";
+                for value in values {
+                    write!(f, "{prefix}{value}")?;
+                    prefix = ", ";
+                }
+
+                let opt = *opt as u32;
+                if opt & (ONIG_OPTION_MULTILINE | ONIG_OPTION_IGNORECASE | ONIG_OPTION_EXTEND) > 0 {
```
tekknolagi: I think there are a bunch more of these like `ONIG_OPTION_DONT_CAPTURE_GROUP`. Maybe something like this?

```rust
/// make a pair of stringified and int
macro_rules! strenum {
    ($name:ident) => {
        (stringify!($name), $name)
    };
}

static ALL_FLAGS: &[(&str, u32)] = &[
    strenum!(ONIG_OPTION_NONE),
    strenum!(ONIG_OPTION_IGNORECASE),
    strenum!(ONIG_OPTION_EXTEND),
    strenum!(ONIG_OPTION_MULTILINE),
    strenum!(ONIG_OPTION_SINGLELINE),
    strenum!(ONIG_OPTION_FIND_LONGEST),
];

/// iterate over ALL_FLAGS and call bitflag!
macro_rules! bitflags {
    ($writer:expr, $flag:expr) => {
        let mut sep = "";
        for (name, flag) in ALL_FLAGS {
            if $flag & flag != 0 {
                write!($writer, "{sep}{name}")?;
                sep = "|";
            }
        }
    };
}

impl std::fmt::Display for F {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        bitflags!(f, self.options);
        Ok(())
    }
}
```

Please pardon the Rust---I'm very much not an expert

tekknolagi: Actually this is cleaner I think

```rust
macro_rules! bitflags {
    ($writer:expr, $bitfield:expr, $($flag:ident),*) => {
        let mut sep = "";
        $(
            if $bitfield & $flag != 0 {
                write!($writer, "{sep}{}", stringify!($flag))?;
                sep = "|";
            }
        )*
    };
}

impl std::fmt::Display for F {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        bitflags!(f, self.options, ONIG_OPTION_NONE, ONIG_OPTION_IGNORECASE, ONIG_OPTION_EXTEND, ONIG_OPTION_MULTILINE, ONIG_OPTION_SINGLELINE, ONIG_OPTION_FIND_LONGEST);
        Ok(())
    }
}
```

tekknolagi: (but please get all the relevant flags from `include/ruby/onigmo.h`)

composerinteralia: Hm, I don't think all of those onigmo flags match exactly what is in `opt` here. I modeled this printing after https://github.com/ruby/ruby/blob/a04555c8ab109355d00275d28d867fce06805c55/re.c#L322-L331 but I guess there actually are some other options for encodings that don't typically get printed. This might be the more complete list: https://github.com/ruby/ruby/blob/a04555c8ab109355d00275d28d867fce06805c55/re.c#L4863-L4872

So 16 is `ARG_ENCODING_FIXED` and 32 is `ARG_NO_ENCODING`. Those don't match up with the ognimo options with the same values. NOENCODING with the `n` flag, for example:

```
% ruby --dump=insn -e '/#{}/n'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,6)>
0000 putobject                              ""                        (   1)[Li]
0002 putnil
0003 dup
0004 objtostring                            <calldata!mid:to_s, argc:0, FCALL|ARGS_SIMPLE>
0006 anytostring
0007 toregexp                               32, 2
0010 leave

% ruby -e 'p Regexp::NOENCODING; p //n.options'
32
32
```

composerinteralia: ARG_NO_ENCODING is easy enough to display as `n`. I'm not sure how best to print ARG_ENCODING_FIXED though since it seems like the option itself doesn't tell us whether it's an e, s, or u.

tekknolagi: .......huh. That's a weird subset. I don't think we need to print as one-char; printing as OR-ed long-name is probably fine (like the function call flags). Thank you for investigating. I trust your judgement here

composerinteralia: I'll try printing it as the Ruby constant name like `IGNORECASE|EXTENDED|MULTILINE|FIXEDENCODING|NOENCODING`

k0kubun (approved):

zjit/src/hir.rs
composerinteralia: OK, I've got something that works and I think covers all the possible options. I didn't end up using the macros you offered above since using the Ruby constant name made them somewhat less useful, but I'm open to suggestions if there's a more elegant way to do this.

PR #14203, opened 2025-08-13
[WIP] Add safe dig

CuddlyBunion341:
Digging hashes has always been a pain in Ruby.

When handling API responses, you may only care about the result if you can precisely dig there and don't want to catch intermediary errors such as dig not being defined. Sometimes APIs respond inconsistently, sometimes with strings and sometimes with hashes.

Currently dig does not work well with such mixed types:

```ruby

my_api_response = { status: "none" }
my_api_response.dig(:status, :code) # => TypeError: String does not have #dig method (TypeError)
```

It would be very nice if we could have a safe dig that returns early when it cannot dig further:

```ruby
my_api_response = { status: "none" }
my_api_response.safe_dig(:status, :code) # => nil
```

### DIFF
--- a/hash.c
+++ b/hash.c
@@ -4902,6 +4902,12 @@ rb_hash_dig(int argc, VALUE *argv, VALUE self)
     return rb_obj_dig(argc, argv, self, Qnil);
 }
 
+static VALUE
+rb_safe_hash_dig(int argc, VALUE *argv, VALUE self)
+{
+    return 42;
+}
+
 static int
 hash_le_i(VALUE key, VALUE value, VALUE arg)
 {
@@ -7468,6 +7474,7 @@ Init_Hash(void)
 
     rb_define_method(rb_cHash, "any?", rb_hash_any_p, -1);
     rb_define_method(rb_cHash, "dig", rb_hash_dig, -1);
+    rb_define_method(rb_cHash, "safe_dig", rb_safe_hash_dig, -1);
 
     rb_define_method(rb_cHash, "<=", rb_hash_le, 1);
     rb_define_method(rb_cHash, "<", rb_hash_lt, 1);
--- a/spec/ruby/core/hash/safe_dig_spec.rb
+++ b/spec/ruby/core/hash/safe_dig_spec.rb
@@ -1,0 +1,13 @@
+require_relative '../../spec_helper'
+
+describe "Hash#safe_dig" do
+  it "behaves like dig when not hitting rock bottom" do
+    h = { foo: { bar: "baz" }}
+    h.safe_dig(:foo, :bar).should == "baz"
+  end
+
+  it "returns nil instead of raising dig error when hitting rock bottom" do
+    h = { foo: { bar: '' } }
+    h.safe_dig(:foo, :bar, :baz).should == nil
+  end
+end


hash.c
```diff
@@ -4902,6 +4902,69 @@ rb_hash_dig(int argc, VALUE *argv, VALUE self)
...
+ *    h.safe_dig(key, ...) -> object or nil
+ *
+ *  Traverses the given keys and returns the value found or +nil+ if any step
+ *  is not a Hash or Array, or the key/index is missing.
+ *
+ *  Accepts both String and Symbol keys for Hash lookup.
+ *
+ *    { foo: { bar: 1 } }.safe_dig(:foo, :bar)         #=> 1
+ *    { foo: "none" }.safe_dig(:foo, :bar)             #=> nil
+ *    { items: [{id: 1}, {id: 2}] }.safe_dig(:items, 1, :id)  #=> 2
```
cb341:
```suggestion
*    { foo: { bar: 1 } }.safe_dig(:foo, :bar)                #=> 1
*    { foo: "none" }.safe_dig(:foo, :bar)                    #=> nil
*    { items: [{id: 1}, {id: 2}] }.safe_dig(:items, 1, :id)  #=> 2
```

nobu:
1. New features need discussions at the [ITS](https://bugs.ruby-lang.org/issues).
2. The word “safe” feels improper as the behavior not to raise an exception.
3. Other than `Hash` and `Array` are not supported.
4. Mixing `Symbol` and `String` as `Hash` keys is bad.

> Many APIs return inconsistent structures, sometimes nested hashes and sometimes plain strings or other types, which makes this process error prone.

Is it so common?

cb341:
1. Thanks for the notice @nobu.
I'll start a conversation on the official issue tracker.

2. The naming was inspired by safe_navigate. Which does not raise either.

3. **Regarding the other types:**
The idea would be to dig only on such objects?
Why would I want to `safe_dig` on a string or an integer?

Maybe it makes sense to implement dig on the object level..

4. Also thanks for the remark regarding the mixing of key types.
What would be your proposed solution to retrieving hash data?

rnestler:
A better name could be `try_dig`? Similar to the `rails` helper `try` compared to the safe navigator:
```ruby
[1] pry(main)> "foo".try(:length)
=> 3
[2] pry(main)> "foo".try(:bar)
=> nil
[3] pry(main)> nil.try(:length)
=> nil
```

```ruby
[1] pry(main)> "foo"&.length
=> 3
[2] pry(main)> "foo"&.bar
NoMethodError: undefined method `bar' for an instance of String (NoMethodError)

"foo"&.bar
     ^^^^^
from (pry):2:in `__pry__'
[3] pry(main)> nil&.length
=> nil
```

cb341:
Thanks for the suggestion @rnestler!
try_dig does make more sense than safe_dig regarding the error handling behavior.

I have opened a feature request regarding this functionality:
https://bugs.ruby-lang.org/issues/21545

hash.c
nobu: Why unaligned in the same comment block?

nobu:
> 3. **Regarding the other types:**
>    The idea would be to dig only on such objects?
>    Why would I want to `safe_dig` on a string or an integer?

`dig` also does not support `String` or `Integer`, but it supports `Struct`, `Data`, and objects with `#dig` methods.

> 4. Also thanks for the remark regarding the mixing of key types.
>    What would be your proposed solution to retrieving hash data?

I’d just distinguish them as-is.

cb341:
Closing this PR as the feature makes more sense in active support opposed to ruby core and was rejected.

PR #14206, opened 2025-08-13
Drop Travis CI.

junaruga:
Travis CI has not been running since 2 months ago. https://app.travis-ci.com/github/ruby/ruby/builds

As we are migrating the ppc64le/s390x CI cases from Travis to GitHub Actions, I will drop the `.travis.yml`, while the GitHub Actions ppc64le case is not available yet.


junaruga:
It seems the the fix for the failures on the CI Window case is coming at the https://github.com/ruby/ruby/pull/14220.

PR #14207, opened 2025-08-13
ZJIT: Implement `StringIntern` codegen

st0012:


zjit/src/hir.rs
```diff
@@ -4494,6 +4495,26 @@ mod tests {
...
+        "#);
+        assert_method_hir_with_opcode("test", YARVINSN_intern, expect![[r#"
+            fn test@<compiled>:3:
+            bb0(v0:BasicObject):
+              v2:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+              v3:Fixnum[123] = Const Value(123)
+              v5:BasicObject = ObjToString v3
+              v7:String = AnyToString v3, str: v5
+              v9:StringExact = StringConcat v2, v7
+              v11:StringExact = StringIntern v9
```
tekknolagi: lol oops can you fix this to return Symbol?

tekknolagi (approved):

zjit/src/hir.rs
st0012: ohh good catch

zjit/src/codegen.rs
```diff
@@ -591,6 +591,15 @@ fn gen_getglobal(asm: &mut Assembler, id: ID) -> Opnd {
     asm_ccall!(asm, rb_gvar_get, id.0.into())
 }
 
+/// Intern a string
+fn gen_intern(asm: &mut Assembler, val: Opnd, state: &FrameState) -> Option<Opnd> {
+    // Save the PC and SP because we might allocate
```
k0kubun: Saying "we might allocate" doesn't seem to add any information to what the name of `gen_prepare_call_with_gc` explains, and saving PC is an implementation detail abstracted away by `gen_prepare_call_with_gc`, so I'd say this comment is not necessary. You might want to sometimes explain why a C function allocates when it's not clear from the function name, but it's not what this comment is doing now.

Also note that it doesn't save SP. This is part of the reason why you shouldn't duplicate the comment about the implementation of `gen_prepare_call_with_gc`. Please see the comments in the function.

```suggestion
```

st0012: Ah my bad, sorry

zjit/src/codegen.rs
```diff
@@ -591,6 +591,14 @@ fn gen_getglobal(asm: &mut Assembler, id: ID) -> Opnd {
     asm_ccall!(asm, rb_gvar_get, id.0.into())
 }
 
+/// Intern a string
+fn gen_intern(asm: &mut Assembler, val: Opnd, state: &FrameState) -> Option<Opnd> {
+    gen_prepare_call_with_gc(asm, state);
+
+    // Call rb_str_intern(str)
```
k0kubun: same applies to this comment. `Some(asm_ccall!(asm, rb_str_intern, val))` calling `rb_str_intern` seems very obvious. If you want to explain val == str, you should probably rename the field of `Insn::StringIntern` and the argument of this function instead.

```suggestion
```

PR #14208, opened 2025-08-13
YJIT: Fix `defined?(yield)` and `block_given?` at top level

XrXr:
- [ ] File backport: 

      $ ruby --yjit-call-threshold=1 -ve 'p block_given?'
      ruby 3.4.5 (2025-07-16 revision 20cda200d3) +YJIT +PRISM [arm64-darwin24]
      true

---

Previously, YJIT returned truthy for the block given query at the top
level. That's incorrect because the top level script never receives a
block, and `yield` is a syntax error there.

Inside methods, the number of hops to get from `iseq` to
`iseq->body->local_iseq` is the same as the number of
`VM_ENV_PREV_EP(ep)` hops to get to an environment with
`VM_ENV_FLAG_LOCAL`. YJIT and the interpreter both rely on this as can
be seen in get_lvar_level(). However, this identity does not hold for
the top level frame because of vm_set_eval_stack(), which sets up
`TOPLEVEL_BINDING`.

Since only methods can take a block that `yield` goes to, have ISEQs
that are the child of a non-method ISEQ return falsy for the block given
query. This fixes the issue for the top level script and is an
optimization for non-method contexts such as inside `ISEQ_TYPE_CLASS`.

---

I think this also fixes the issue in <https://github.com/ruby/ruby/pull/13454>. There's no need for a runtime check.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -6656,16 +6656,24 @@ fn gen_block_given(
 ) {
     asm_comment!(asm, "block_given?");
 
-    // Same as rb_vm_frame_block_handler
-    let ep_opnd = gen_get_lep(jit, asm);
-    let block_handler = asm.load(
-        Opnd::mem(64, ep_opnd, SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL)
-    );
+    // `yield` goes to the block handler stowed in the "local" iseq which is
+    // the current iseq or a parent. Only the "method" iseq type can be passed a
+    // block handler. (e.g. `yield` in the top level script is a syntax error.)
+    let local_iseq = unsafe { rb_get_iseq_body_local_iseq(jit.iseq) };
+    if unsafe { rb_get_iseq_body_type(local_iseq) } == ISEQ_TYPE_METHOD {
+        // Same as rb_vm_frame_block_handler
+        let ep_opnd = gen_get_lep(jit, asm);
+        let block_handler = asm.load(
+            Opnd::mem(64, ep_opnd, SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL)
+        );
 
-    // Return `block_handler != VM_BLOCK_HANDLER_NONE`
-    asm.cmp(block_handler, VM_BLOCK_HANDLER_NONE.into());
-    let block_given = asm.csel_ne(true_opnd, false_opnd);
-    asm.mov(out_opnd, block_given);
+        // Return `block_handler != VM_BLOCK_HANDLER_NONE`
+        asm.cmp(block_handler, VM_BLOCK_HANDLER_NONE.into());
+        let block_given = asm.csel_ne(true_opnd, false_opnd);
+        asm.mov(out_opnd, block_given);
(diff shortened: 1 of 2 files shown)


yjit/src/codegen.rs
```diff
@@ -6656,16 +6656,24 @@ fn gen_block_given(
...
     asm_comment!(asm, "block_given?");
 
-    // Same as rb_vm_frame_block_handler
-    let ep_opnd = gen_get_lep(jit, asm);
-    let block_handler = asm.load(
-        Opnd::mem(64, ep_opnd, SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL)
-    );
+    // `yield` goes to the block handler stowed in the "local" iseq which is
+    // the current iseq or a parent. Only the "method" iseq type can be passed a
+    // block handler. (e.g. `yield` in the top level script is a syntax error.)
```
k0kubun, line 6661: I was kind of confused about `ISEQ_TYPE_BLOCK` since you can pass a block to a block, but I guess we should still return `false` because the `block_given?` method refers to the local environment (`LEP`) that a block would inherit from a frame of non-`ISEQ_TYPE_BLOCK` ISEQ.

I'm also unsure what `ISEQ_TYPE_MAIN` (when there's already `ISEQ_TYPE_TOP`) and `ISEQ_TYPE_PLAIN` are, but the assumption probably holds for the rest of the types.

k0kubun: It might be useful to have an assertion for debug builds in either the original C function or generated code to improve our confidence, but compatibility-wise it seems optional since it's passing tests anyway.

k0kubun (approved):
Thanks for the patch. If the same codegen makes https://github.com/ruby/ruby/pull/14208 work, that'd be amazing.

yjit/src/codegen.rs
k0kubun: > I guess we should still return false because the block_given? method refers to the local environment (LEP) that a block would inherit from a frame of non-ISEQ_TYPE_BLOCK ISEQ.

Well, but then the LEP frame could indeed be `ISEQ_TYPE_METHOD`.

Does it support this kind of situation? (Is the `{ block_given? }` ISEQ not a `ISEQ_TYPE_BLOCK`?)

```rb
irb(main)[01:0]> def foo = self.then { block_given? }
=> :foo
irb(main)[02:0]> foo
=> false
irb(main)[03:0]> foo {}
=> true
```

XrXr: > Does it support this kind of situation? (Is the { block_given? } ISEQ not a ISEQ_TYPE_BLOCK?)

Yes it works; lots of test code call `block_given?` from within a block. The block ISEQ in this example is indeed `ISEQ_TYPE_BLOCK`, but the ISEQ type check happens on `block_iseq->body->local_iseq`, not on `block_iseq`, and `block_iseq->body->local_iseq->type == ISEQ_TYPE_METHOD`.

k0kubun: Sorry, I missed the `rb_get_iseq_body_local_iseq` portion. Thanks for the explanation!

PR #14210, opened 2025-08-13
Don't change Encoding.default_internal in assert_raise_with_message

peterzhu2118:
For most tests (except two), we don't need to change Encoding.default_internal in assert_raise_with_message. We're trying to run the test suite across multiple Ractors and modifying Encoding.default_internal can cause other concurrently running tests to fail.

### DIFF
--- a/test/ruby/test_numeric.rb
+++ b/test/ruby/test_numeric.rb
@@ -22,10 +22,12 @@ class TestNumeric < Test::Unit::TestCase
     assert_raise_with_message(TypeError, /:\u{3042}/) {1&:"\u{3042}"}
     assert_raise_with_message(TypeError, /:\u{3042}/) {1|:"\u{3042}"}
     assert_raise_with_message(TypeError, /:\u{3042}/) {1^:"\u{3042}"}
-    assert_raise_with_message(TypeError, /:"\\u3042"/) {1+:"\u{3042}"}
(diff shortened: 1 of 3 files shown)


eregon:
Note: basically a redo of https://github.com/ruby/ruby/commit/c32ce5fc27075d64e0509fee5e6703098aa617c4 + https://github.com/ruby/ruby/commit/f356bd817cc636c6c91d106505a3b09d785b2f32. That got reverted in https://github.com/ruby/ruby/commit/cf4d4c32f341ef342e91c367dbd414147ed08542.
BTW r54522 is https://github.com/ruby/ruby/commit/404bf57aaffd64cb251574e2916e393d272f77a3.
 I don't recall why the revert of both my commits but probably some CI was failing, maybe on some exotic platform or Windows or so which has/had a non-default UTF-8 external encoding (though this is about internal encoding). Encodings have changed a bit on Windows so it might be fine now.

In any case, :+1: from me.

eregon (approved):

eregon:
I think back then I also thought it's really not great that any core `#inspect` depends on internal encoding, which is maybe what I meant by `More general fix coming.`. Not sure what happened to that, maybe I filed a redmine issue.

I wonder what's the use case for `Encoding.default_internal` actually, it's `nil` by default and I don't recall any real code setting it.

PR #14211, opened 2025-08-13
[DOC] Tweaks for GC.stat

BurdetteLamar:
I have re-formatted, but not otherwise modified, the trailing list.

### DIFF
--- a/gc.rb
+++ b/gc.rb
@@ -110,85 +110,127 @@ module GC
   end
 
   # call-seq:
-  #    GC.stat -> Hash
-  #    GC.stat(hash) -> Hash
-  #    GC.stat(:key) -> Numeric
+  #   GC.stat -> new_hash
+  #   GC.stat(key) -> value
+  #   GC.stat(hash) -> hash
   #
-  # Returns a Hash containing information about the \GC.
+  # This method is implementation-specific to CRuby.
   #
-  # The contents of the hash are implementation-specific and may change in
-  # the future without notice.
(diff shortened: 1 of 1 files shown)


gc.rb
```diff
@@ -110,85 +110,127 @@ def self.count
...
+  # returns that hash with GC statistics merged into its content;
+  # this form may be useful in minimizing {probe effects}[https://en.wikipedia.org/wiki/Probe_effect]:
+  #
+  #   h = {foo: 0, bar: 1}
+  #   GC.stat(h)
+  #   h.keys.take(5) # => [:foo, :bar, :count, :time, :marking_time]
+  #
+  # The hash includes entries such as:
+  #
+  # - +count:+ -
```
peterzhu2118: Why is there a dash after each of the keys? It generates as a dash in the rendered HTML:

Screenshot 2025-08-13 at 3 30 01 PM

BurdetteLamar: I had `+foo:+:`, but it looked weird to have the two colons; so I used `+foo:+ -` instead.  I think we have done this a couple of other places, for similar reasons.

peterzhu2118: Since it's a symbol, we can (and probably should) put the colon at the beginning, like this:

```
  # - +:count+:
```

BurdetteLamar: Duh.  What was I thinking?  Will fix.

peterzhu2118 (approved):

PR #14212, opened 2025-08-13
ZJIT: Side-exit on unknown instructions

tekknolagi:
Don't abort the entire compilation.

Fix https://github.com/Shopify/ruby/issues/700

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -281,11 +281,15 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Optio
         // Compile all instructions
         for &insn_id in block.insns() {
             let insn = function.find(insn_id);
-            if gen_insn(cb, &mut jit, &mut asm, function, insn_id, &insn).is_none() {
-                debug!("Failed to compile insn: {insn_id} {insn}");
+            if let Err(last_snapshot) = gen_insn(cb, &mut jit, &mut asm, function, insn_id, &insn) {
+                debug!("ZJIT: gen_function: Failed to compile insn: {insn_id} {insn}. Generating side-exit.");
                 incr_counter!(failed_gen_insn);
-                return None;
-            }
+                gen_side_exit(&mut jit, &mut asm, &SideExitReason::UnhandledInstruction(insn_id), &function.frame_state(last_snapshot));
+                // Don't bother generating code after a side-exit. We won't run it.
+                // TODO(max): Generate ud2 or equivalent.
+                break;
+            };
+            // It's fine; we generated the instruction
         }
         // Make sure the last patch point has enough space to insert a jump
         asm.pad_patch_point();
@@ -316,7 +320,7 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Optio
 }
 
 /// Compile an instruction
-fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, function: &Function, insn_id: InsnId, insn: &Insn) -> Option<()> {
+fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, function: &Function, insn_id: InsnId, insn: &Insn) -> Result<(), InsnId> {
     // Convert InsnId to lir::Opnd
     macro_rules! opnd {
         ($insn_id:ident) => {
@@ -334,7 +338,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
 
     macro_rules! no_output {
         ($call:expr) => {
-            { let () = $call; return Some(()); }
+            { let () = $call; return Ok(()); }
         };
     }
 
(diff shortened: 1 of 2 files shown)


k0kubun:
It seems like this PR fails some tests.

Assuming `last_snapshot` created by `Snapshot` is accurate, one possible issue in the implementation could be that we return `None` from side exit-unsafe points. We addressed this problem in YJIT by making sure `return None` doesn't happen after a point that's unsafe to side-exit.

A more elegant solution might be to get a snapshot of `Assembler` before calling `gen_insn` and then rewind `Assembler` to that state before generating a side exit. Save the length of `insns`, `live_ranges`, `label_names`, and drop any new entries in these vectors. Remaining `live_ranges` with `end` pointing to deleted insns would need a scan of `insns` to be fixed.

tekknolagi:
What is a side-exit unsafe point?

tekknolagi:
This is a minimal repro of a crash with this diff, btw:

```ruby
def test
  require_labels = ["replace_require", "require"]
  uplevel = 0
  require_found = false
  Thread.each_caller_location do |cl|
    if require_found
      unless require_labels.include?(cl.base_label)
        return uplevel
      end
    end
    cl.path.end_with?("bundle")
  end
end

test
test
test
```

tekknolagi:
Oh, you are saying they may have emitted some code before aborting. Hm.

tekknolagi:
A lot of these can only abort because of `jit.get_opnd` returning `None`. I personally don't mind that being a noisy failure instead; we *should* (whatever that means) be precluding this with the HIR verifier now.

k0kubun:
> What is a side-exit unsafe point?
> Oh, you are saying they may have emitted some code before aborting.

For YJIT, it was generally when it writes to the stack in a way that cannot be restored by spills on side-exit. The method call implementation had so many stack manipulations that it had "do not side exit beyond this" kind of point (which I can't find right now, maybe we fixed it by `ctx` rewinding?).

I thought ZJIT's stack writes by send may not be unsafe (particularly `gen_send_without_block` not replicating the layout of the interpreter stack seems bad), but I guess side-exit will overwrite everything, so that might be safe for ZJIT. I can't point out an unsafe point at the moment.

> A lot of these can only abort because of jit.get_opnd returning None. I personally don't mind that being a noisy failure instead; we should (whatever that means) be precluding this with the HIR verifier now.

Sure. Let's `unwrap()` them then. We'll see less fail-able points and it'll be easier to debug.

Separately from that, if we can rewind Assembler to a before-`gen_insn` state efficiently, we probably should. There's no need to leave unnecessary instructions if they're going to be wasted after the side exit anyway.

tekknolagi:
Moving back to draft while I solve some underlying issues.

tekknolagi:
Still panicking about VReg. Smaller repro:

```ruby
def test
  uplevel = 0
  Thread.each_caller_location do |cl|
    return uplevel + 5
  end
end

p test
p test
p test
```

tekknolagi:
~~Ah, I am getting too late a FrameState somehow.~~

I think it's related to https://github.com/Shopify/ruby/issues/614

tekknolagi:
Well, no crashes, now it just fails btest :)

tekknolagi:
I think it's because we don't handle a JITed block Throw-ing to a JITed caller:

```ruby
def test
  result = []
  [1,2].each {
    result << 2
    break
    $a << 3
  }
  result
end
result = test
expected = [2]
if result != expected
  raise "Interpreter Test failed: Expected #{expected} but got #{result.inspect}"
else
  puts "Interpreter Test passed"
end
result = test
if result != expected
  raise "ZJIT Test failed: Expected #{expected} but got #{result.inspect}"
else
  puts "ZJIT Test passed"
end
```

tekknolagi:
Although........

```
ZJIT: gen_function: Failed to compile insn: v9 Send v7, 0x16d9da950, :each. Generating side-exit.
...............
ZJIT: gen_function: Failed to compile insn: v7 Throw TAG_BREAK, v6. Generating side-exit.
```

tekknolagi:
If we bisect, we don't even need to compile the Send -- compiling the block is enough:

```
plum% ../build-dev/miniruby --zjit-allowed-iseqs\=jitlist.txt testrescue.rb
JIT list:
  block in test@testrescue.rb:3
Interpreter Test passed
testrescue.rb:19:in '<main>': ZJIT Test failed: Expected [2] but got [2, 2] (RuntimeError)
plum%
```

tekknolagi:
.........actually I think this is because we have the wrong FrameState. Going to require all unimplemented functions to truck around a Snapshot InsnId for their implicit SideExit

tekknolagi:
Ah, and Throw is the only unimplemented one that currently doesn't do that

tekknolagi:
Putting this back up for review @ruby/jit -- maybe we can compile more functions now!

zjit/src/codegen.rs
```diff
@@ -281,11 +281,16 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Optio
...
+            let Err(last_snapshot) = gen_insn(cb, &mut jit, &mut asm, function, insn_id, &insn) else {
+                // It's fine; we generated the instruction
+                continue;
+            };
+            debug!("ZJIT: gen_function: Failed to compile insn: {insn_id} {insn}. Generating side-exit.");
+            incr_counter!(failed_gen_insn);
+            gen_side_exit(&mut jit, &mut asm, &SideExitReason::UnhandledInstruction(insn_id), &function.frame_state(last_snapshot));
+            // Don't bother generating code after a side-exit. We won't run it.
+            // TODO(max): Generate ud2 or equivalent.
+            break;
```
k0kubun: I'm a fan of small indentation and doing early returns (it's actually `continue` here) for errors, but with this code, it feels like `continue;` was the error and the other was the main body of the loop, especially when you had to explain "it's fine".

Maybe that's just me, but it feels less confusing to keep the previous structure:

```suggestion
            if let Err(last_snapshot) = gen_insn(cb, &mut jit, &mut asm, function, insn_id, &insn) {
                debug!("ZJIT: gen_function: Failed to compile insn: {insn_id} {insn}. Generating side-exit.");
                incr_counter!(failed_gen_insn);
                gen_side_exit(&mut jit, &mut asm, &SideExitReason::UnhandledInstruction(insn_id), &function.frame_state(last_snapshot));
                // Don't bother generating code after a side-exit. We won't run it.
                // TODO(max): Generate ud2 or equivalent.
                break;
            }
```

k0kubun (approved):
Very nice!

PR #14213, opened 2025-08-13
Require rbconfig in resolv.rb

zenspider:
Uses ::RbConfig::CONFIG['host_os']

Found with ruby --disable-gems -e '...'

### DIFF
--- a/lib/resolv.rb
+++ b/lib/resolv.rb
@@ -4,6 +4,7 @@ require 'socket'
 require 'timeout'
 require 'io/wait'
 require 'securerandom'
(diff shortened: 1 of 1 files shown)


nobu:
Is this the fix for ruby/resolv#97?

zenspider:
> Is this the fix for [ruby/resolv#97](https://github.com/ruby/resolv/issues/97)?

hadn't seen that, but yeah... I can pull the PR over to there if you'd like

XrXr:
Merged in https://github.com/ruby/resolv/pull/99

PR #14221, opened 2025-08-14
Fix documentation about struct RData's data field

etiennebarrie:
Also adds a static assertion to ensure the documented behavior stays true, namely that the data field is at the same position in the RData and RTypedData structs.

This was changed in 360be94d0492f766b08cc39e33f5e248f49a89b7. cc @byroot

### DIFF
--- a/include/ruby/internal/core/rdata.h
+++ b/include/ruby/internal/core/rdata.h
@@ -143,8 +143,8 @@ struct RData {
     RUBY_DATA_FUNC dfree;
 
     /** Pointer to the actual C level struct that you want to wrap.
-      * This is in between dmark and dfree to allow DATA_PTR to continue
-      * to work for both RData and non-embedded RTypedData.
+      * This is after dmark and dfree to allow DATA_PTR to continue to work for
+      * both RData and non-embedded RTypedData.
       */
     void *data;
 };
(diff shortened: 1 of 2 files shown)


nobu:
Do not mix code change into documentation change.

junaruga:
@etiennebarrie Let me explain more context that this PR's commit which caused the CI failures (https://github.com/ruby/ruby/pull/14229).

Previously, the CI skipped checks when a PR commit's title or message includes "document" (case insensitive), assuming the commit is only modifying the document. However, this PR's commit was using the title including "document", but actually modifying the code, and the PR skipped the CI checks.

So, we stopped skipping the CI when the commit title or message include "document" by the https://github.com/ruby/ruby/commit/70b4b6fea0eeb66647539bcb3b9a50d027d92e51. 

There is still a logic to skip CI when a commit's title or message includes "[DOC]". But please keep in mind that you do not modify a code, when using the "[DOC]".

etiennebarrie:
> Do not mix code change into documentation change.

Sorry! 

Thanks for the details @junaruga, I hadn't realized some checks were skipped. Thank everyone and @k0kubun in particular for the quick fix.
And again sorry everyone for the trouble.

PR #14222, opened 2025-08-14
CI: ubuntu-ibm.yml: Add GitHub Actions ppc64le case

junaruga:
Note that the default configure option `./configure cppflags=-DRUBY_DEBUG`
causes the following Ractor test and other tests failing. So, we don't set the
option in ppc64le cases.

```
$ make btest BTESTS=bootstraptest/test_ractor.rb
```

You can check the following ticket for details.
https://bugs.ruby-lang.org/issues/21534

### DIFF
--- a/.github/workflows/ubuntu-ibm.yml
+++ b/.github/workflows/ubuntu-ibm.yml
@@ -27,10 +27,7 @@ jobs:
         test_task: [check]
         configure: ['']
         os:
-          # FIXME Comment out ppc64le due to failing tests on GitHub Actions
-          # ppc64le
-          # https://bugs.ruby-lang.org/issues/21534
-          # - ubuntu-24.04-ppc64le
+          - ubuntu-24.04-ppc64le
           - ubuntu-24.04-s390x
           # Add a x86_64 case to make this CI pass on fork repositories.
           - ubuntu-24.04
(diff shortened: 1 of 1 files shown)


junaruga:
The GitHub Actions ppc64le sometimes delays to start the job. In my latest case, I need to wait for the job to start more than 25 minutes. This is a blocker to merge this PR in my opinion. I reported the issue https://github.com/IBM/actionspz/issues/34.

junaruga:
Some tests in the `make test-all` failed with timeout on the ppc64le case below.
https://github.com/ruby/ruby/actions/runs/16993265630/job/48177676417?pr=14222#step:16:1446

junaruga:
I am seeing many timeout related failures, while the the tests passe on RubyCI Ubuntu (ppc64le) server. I suspect the reason of the failures is a CI machine's slow performance. So, I added the logic to measure the performance of the `make check` by `mpstat`, `vmstat`, and `iostat` commands on the latest commit in this PR.

junaruga:
I reported the GitHub Actions ppc64le (and s390x) performance issue https://github.com/IBM/actionspz/issues/38 coming from the wrong CPU settings in the containers.

junaruga:
I found the root cause https://github.com/IBM/actionspz/issues/38#issuecomment-3221870826.

junaruga:
As the issue https://github.com/IBM/actionspz/issues/34 still blocks to add GitHub Actions ppc64le to ruby/ruby, I sent the https://github.com/ruby/ruby/pull/14345 to merge only the 1st commit related to the `nproc`.

junaruga:
People at IBM said they they fixed the issue https://github.com/IBM/actionspz/issues/34 that jobs sometimes not starting, and I confirmed the GitHub Actions ppc64le/390x jobs started without delay on this PR's latest CI. So, let's merge this PR adding the s390x and ppc64le, and we will see how these work.

junaruga:
I see my fork repository's CI: https://github.com/junaruga/ruby/actions/runs/17273836937 is failing because there is not cases in the `.github/workflows/ubuntu-ibm.yml`. I want to make the CI green in the fork repository case too.

PR #14226, opened 2025-08-14
[DOC] Tweaks for GC.stat_heap

BurdetteLamar:


gc.rb
```diff
@@ -194,61 +194,140 @@ def self.stat hash_or_key = nil
...
+  #      force_incremental_marking_finish_count: 0,
+  #      total_allocated_objects: 1230,
+  #      total_freed_objects: 309}}
+  #
+  # In the example above, the keys in the outer hash are the heap identifiers:
+  #
+  #   GC.stat_heap.keys # => [0, 1, 2, 3, 4]
+  #
+  # On CRuby, each heap identifier is an integer;
+  # no other implementations, a heap identifier may be a string.
```
peterzhu2118: Typo:

```suggestion
  # on other implementations, a heap identifier may be a string.
```

gc.rb
```diff
@@ -194,61 +194,140 @@ def self.stat hash_or_key = nil
...
+  #    force_incremental_marking_finish_count: 0,
+  #    total_allocated_objects: 225018,
+  #    total_freed_objects: 206647}
+  #
+  # With arguments +heap_id+ and +key+ given,
+  # returns the value for the given key in the given heap:
+  #
+  #   GC.stat_heap(2, :slot_size) # => 160
+  #
+  # With arguments +nil+ and# +hash+ given,
```
peterzhu2118:
```suggestion
  # With arguments +nil+ and +hash+ given,
```

gc.rb
```diff
@@ -194,61 +194,140 @@ def self.stat hash_or_key = nil
...
+  # merges the statistics for all heaps into the given hash:
+  #
+  #   GC.stat_heap(nil, h).keys # => [:foo, :bar, 0, 1, 2, 3, 4]
```
peterzhu2118, line 289: I don't think the variable `h` was defined anywhere.

BurdetteLamar: Defn added (two places).

PR #14229, opened 2025-08-14
Fix a static assertion incompatible with C++98

k0kubun:
The CI was skipped for the PR, but https://github.com/ruby/ruby/pull/14221 has failed the C++98 CI. https://github.com/ruby/ruby/actions/runs/16973372638/job/48115906097

cc: @etiennebarrie @byroot


k0kubun:
https://github.com/ruby/ruby/pull/14232

PR #14235, opened 2025-08-15
zjit_bisect.rb: Fix pipe deadlock; log when timed out

XrXr:
I spent a long time bisecting some largish program before realising
that failure was in fact coming from the child timing out due to
zjit_bisect.rb not clearing the stdout and stderr pipe.

Fix the pipe dead lock by redirecting to /dev/null. This sacrifices
the debug output during boot, but for that we also get to remove
a lot of code and the dependency on Open3.

---

> Warning Use [communicate()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate) rather than [.stdin.write](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stdin), [.stdout.read](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stdout) or [.stderr.read](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.stderr) to avoid deadlocks due to any of the other OS pipe buffers filling up and blocking the child process.

https://docs.ruby-lang.org/en/master/Open3.html#method-i-popen3

> Take care to avoid deadlocks. Output streams stdout and stderr have fixed-size buffers, so reading extensively from one but not the other can cause a deadlock when the unread buffer fills. To avoid that, stdout and stderr should be read simultaneously (using threads or [IO.select](https://docs.ruby-lang.org/en/master/IO.html#method-c-select)).

### DIFF
--- a/tool/zjit_bisect.rb
+++ b/tool/zjit_bisect.rb
@@ -1,6 +1,5 @@
 #!/usr/bin/env ruby
 require 'logger'
-require 'open3'
 require 'optparse'
 require 'shellwords'
 require 'tempfile'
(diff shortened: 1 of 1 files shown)


tekknolagi (approved):
Thank you -- this is nice

PR #14238, opened 2025-08-15
ZJIT: Remove a bunch of Option from codegen

tekknolagi:
I want to reduce the chance of accidentally having a not-implemented or
internal-compiler-error related side-exit after some code for an instruction
has already been emitted. So in this PR I make some of these ICEs cause a noisy
error and remove the spurious use of `Option`s.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -439,6 +439,7 @@ pub enum SideExitReason {
     CalleeSideExit,
     ObjToStringFallback,
     UnknownSpecialVariable(u64),
+    UnhandledDefinedType(usize),
 }
 
 impl std::fmt::Display for SideExitReason {
@@ -2957,6 +2958,11 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let pushval = get_arg(pc, 2);
                     let v = state.stack_pop()?;
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    if op_type == DEFINED_METHOD.try_into().unwrap() {
+                        // TODO(Shopify/ruby#703): Fix codegen for defined?(method call expr)
+                        fun.push_insn(block, Insn::SideExit { state: exit_id, reason: SideExitReason::UnhandledDefinedType(op_type)});
+                        break; // End the block
+                    }
                     state.stack_push(fun.push_insn(block, Insn::Defined { op_type, obj, pushval, v, state: exit_id }));
                 }
                 YARVINSN_definedivar => {
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -1245,6 +1245,14 @@ class TestZJIT < Test::Unit::TestCase
     }, insns: [:defined]
   end
 
+  def test_defined_with_method_call
+    assert_compiles '["method", nil]', %q{
+      def test = return defined?("x".reverse(1)), defined?("x".reverse(1).reverse)
+
+      test
+    }, insns: [:defined]
+  end
+
   def test_defined_yield
     assert_compiles "nil", "defined?(yield)"
     assert_compiles '[nil, nil, "yield"]', %q{


zjit/src/codegen.rs
```diff
@@ -475,65 +480,64 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, obj: VALUE,
...
     }
 }
 
 /// Get a local variable from a higher scope or the heap. `local_ep_offset` is in number of VALUEs.
 /// We generate this instruction with level=0 only when the local variable is on the heap, so we
 /// can't optimize the level=0 case using the SP register.
-fn gen_getlocal_with_ep(asm: &mut Assembler, local_ep_offset: u32, level: u32) -> Option<lir::Opnd> {
+fn gen_getlocal_with_ep(asm: &mut Assembler, local_ep_offset: u32, level: u32) -> lir::Opnd {
     let ep = gen_get_ep(asm, level);
-    let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).ok()?);
```
tekknolagi, line 503: This is one of the few cases I am uncertain about. Maybe we should catch these cases in the caller for a clearer signal

zjit/src/codegen.rs
```diff
@@ -475,65 +480,64 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, obj: VALUE,
...
 /// can't optimize the level=0 case using the SP register.
-fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, local_ep_offset: u32, level: u32) -> Option<()> {
+fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, local_ep_offset: u32, level: u32) {
     let ep = gen_get_ep(asm, level);
     match val {
         // If we're writing a constant, non-heap VALUE, do a raw memory write without
         // running write barrier.
         lir::Opnd::Value(const_val) if const_val.special_const_p() => {
-            let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).ok()?);
+            let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).unwrap());
```
tekknolagi: This is one of the few cases I am uncertain about. Maybe we should catch these cases in the caller for a clearer signal

zjit/src/codegen.rs
```diff
@@ -475,65 +480,64 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, obj: VALUE,
...
             asm.mov(Opnd::mem(64, ep, offset), val);
         }
         // We're potentially writing a reference to an IMEMO/env object,
         // so take care of the write barrier with a function.
         _ => {
-            let local_index = c_int::try_from(local_ep_offset).ok().and_then(|idx| idx.checked_mul(-1))?;
+            let local_index = c_int::try_from(local_ep_offset).ok().and_then(|idx| idx.checked_mul(-1)).unwrap();
```
tekknolagi: This is one of the few cases I am uncertain about. Maybe we should catch these cases in the caller for a clearer signal

tekknolagi:
Minimal repro, x86 only

```ruby
class C
  def assert_equal expected, actual
    if expected != actual
      raise "NO"
    end
  end

  def test_defined_method
    assert_equal(nil, defined?("x".reverse(1).reverse))
  end
end

c = C.new
c.test_defined_method
c.test_defined_method
c.test_defined_method
```

tekknolagi:
Although I don't know what this is supposed to be testing because `"x".reverse(1)` raises `ArgumentError` in isolation....

tekknolagi:
oh `defined?` should be able to find that it's a method though. hm

tekknolagi:
Can repro on ARM after all

tekknolagi:
Opening this up for review

tekknolagi:
`Defined` was failing to compile before. Fixed a bug in DCE related to `Defined`. Now it is compiling but incorrect in one case, so made a temporary `SideExit` in HIR building.

zjit/src/codegen.rs
k0kubun: Could we include the value of `local_ep_offset` in the panic message then?

zjit/src/hir.rs
```diff
@@ -439,6 +439,7 @@ pub enum SideExitReason {
     CalleeSideExit,
     ObjToStringFallback,
     UnknownSpecialVariable(u64),
+    TemporaryBug,
```
k0kubun: I get that you want to leave the bug only temporarily, but it's always nice to describe what it is. Could we call it `UnknownDefinedType(op)` (also I feel like they are not "Unknown" to us but just "Unhandled" btw) or `DefinedMethod`?

k0kubun (approved):
nice

PR #14242, opened 2025-08-15
ZJIT: Make --zjit-dump-hir work with HIR opt disabled

k0kubun:
This PR allows `--zjit-dump-hir` to work when `--zjit-disable-hir-opt`. ref: https://github.com/ruby/ruby/pull/14181

I know `--zjit-dump-hir-init` still works, but I want to keep using `--zjit-dump-hir` when switching `--zjit-disable-hir-opt` because I use it as "dump HIR used by codegen", which should work even when HIR optimizer is disabled.


XrXr:
Nice! I ran into this too.

PR #14249, opened 2025-08-15
CI: Also show errors in rust-warnings.yml

XrXr:
Demos:
- errors: https://github.com/ruby/ruby/actions/runs/16998802932
- warnings: https://github.com/ruby/ruby/actions/runs/16998864751/job/48196086544?pr=14249
- success: this PR

At under a minute, this check runs faster than a lot of the other CI checks,
so we might as well show errors from `cargo check` to serve
as a smoke check in addition to surfacing warnings.

https://github.com/ruby/ruby/pull/14173#discussion_r2279409225

### DIFF
--- a/.github/workflows/rust-warnings.yml
+++ b/.github/workflows/rust-warnings.yml
@@ -42,10 +42,10 @@ jobs:
         run: rustup default beta
 
       - name: Rust warnings
+        shell: bash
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
Thank you!

PR #14253, opened 2025-08-16
[DOC] Tweaks for String#hash

BurdetteLamar:


string.c
```diff
@@ -4133,10 +4133,13 @@ rb_str_hash_cmp(VALUE str1, VALUE str2)
  * call-seq:
  *   hash -> integer
  *
- * Returns the integer hash value for +self+.
- * The value is based on the length, content and encoding of +self+.
+ * Returns the integer hash value for self.
```
peterzhu2118: I think the documentation uses `+self+` for consistency:

```suggestion
 * Returns the integer hash value for +self+.
```

string.c
```diff
@@ -4134,9 +4134,12 @@ rb_str_hash_cmp(VALUE str1, VALUE str2)
  *   hash -> integer
  *
  * Returns the integer hash value for +self+.
- * The value is based on the length, content and encoding of +self+.
  *
- * Related: Object#hash.
+ * Two hashes with the same content and encoding have the same hash value
+ * (and will compare using #eql?);
```
nobu: What will compare?

BurdetteLamar: Actually, I thought the same thing.  I borrowed this language from elsewhere (Array, Complex, and Struct), but will revise.  I've also made a note to myself to rationalize the language in all #hash methods.

BurdetteLamar:
Marking as Draft pending further commits.

string.c
BurdetteLamar: Revised and expanded.

peterzhu2118 (approved):

PR #14260, opened 2025-08-17
[DOC] Tweaks for Objec#hash

BurdetteLamar:


hash.c
```diff
@@ -321,40 +321,37 @@ objid_hash(VALUE obj)
...
- * do not call the #hash method when used as a hash key.
+ * \Class Hash uses both #hash and #eql? to determine whether two objects
+ * used as hash keys are to be treated as the same key.
+ * A hash value that exceeds the capacity of an Integer is truncated before being used.
  *
- * When implementing your own #hash based on multiple values, the best
- * practice is to combine the class and any values using the hash code of an
- * array:
+ * Many core classes override method Object#hash;
+ * visit {core methods}[https://docs.ruby-lang.org/en/master/table_of_contents.html#methods]
```
peterzhu2118: I don't think we should include this link, and if we do, it should not link to the master branch because it will be incorrect for released versions of Ruby.

peterzhu2118 (approved):

PR #14262, opened 2025-08-18
Properly unlock locked mutexes on thread cleanup.

luke-gruber:
Mutexes were being improperly unlocked on thread cleanup. This bug was introduced in 
050a8954395.

We must keep a reference from the mutex to the thread, because if the fiber
is collected before the mutex is, then we cannot unlink it from the thread
in `mutex_free`. If it's not unlinked from the the thread when it's freed, it 
causes bugs in `rb_thread_unlock_all_locking_mutexes`.

Fixes [Bug #21342]

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -442,7 +442,7 @@ rb_threadptr_unlock_all_locking_mutexes(rb_thread_t *th)
         th->keeping_mutexes = mutex->next_mutex;
 
         // rb_warn("mutex #<%p> was not unlocked by thread #<%p>", (void *)mutex, (void*)th);
-
+        VM_ASSERT(mutex->fiber);
         const char *error_message = rb_mutex_unlock_th(mutex, th, mutex->fiber);
         if (error_message) rb_bug("invalid keeping_mutexes: %s", error_message);
     }
--- a/thread_sync.c
+++ b/thread_sync.c
@@ -8,6 +8,7 @@ static VALUE rb_eClosedQueueError;
 /* Mutex */
 typedef struct rb_mutex_struct {
     rb_fiber_t *fiber;
+    VALUE thread; // even if the fiber is collected, we might need access to the thread in mutex_free
     struct rb_mutex_struct *next_mutex;
     struct ccan_list_head waitq; /* protected by GVL */
 } rb_mutex_t;
@@ -106,8 +107,6 @@ static const char* rb_mutex_unlock_th(rb_mutex_t *mutex, rb_thread_t *th, rb_fib
  *
  */
 
-#define mutex_mark ((void(*)(void*))0)
-
 static size_t
 rb_mutex_num_waiting(rb_mutex_t *mutex)
 {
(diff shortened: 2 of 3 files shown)


thread_sync.c
```diff
@@ -106,7 +107,14 @@ static const char* rb_mutex_unlock_th(rb_mutex_t *mutex, rb_thread_t *th, rb_fib
  *
  */
 
-#define mutex_mark ((void(*)(void*))0)
+static void
+mutex_mark(void * ptr)
+{
+    rb_mutex_t *mutex = ptr;
+    if (mutex->fiber) {
+        rb_gc_mark_movable(rb_fiberptr_self(mutex->fiber));
```
byroot: You use `rb_gc_mark_movable` but `mutex_data_type` doesn't register a `dcompact` method.

luke-gruber: 👋 Doesn't this just mark the fiber without pinning it? Oh right, then we would need our own way to update its references.
Edit: John just mentioned to me that we would also probably need write barriers for `fiber`. John and I will discuss later this week.

byroot: Yes precisely. And if you don't have a `dcompact` callback, by not pinning it, you allow it to move, if that does happen, the ref might be corrupted.

Now in this case it turns out it's alright because the ref is inside a struct you don't own, and the Fiber will update its own reference. But I'd either add a very clear comment mentioning this assumption, or just implement `dcompact` here too.

thread.c
```diff
@@ -437,14 +437,15 @@ rb_threadptr_join_list_wakeup(rb_thread_t *thread)
 void
 rb_threadptr_unlock_all_locking_mutexes(rb_thread_t *th)
 {
-    while (th->keeping_mutexes) {
-        rb_mutex_t *mutex = th->keeping_mutexes;
-        th->keeping_mutexes = mutex->next_mutex;
+    rb_mutex_t *mutexes = th->keeping_mutexes;
+    while (mutexes) {
+        rb_mutex_t *mutex = mutexes;
```
samuel-williams-shopify: Is this change strictly needed?

luke-gruber: No I don't think it was, I've switched it back.

thread_sync.c
```diff
@@ -238,6 +263,8 @@ mutex_locked(rb_thread_t *th, VALUE self)
 {
     rb_mutex_t *mutex = mutex_ptr(self);
 
+    mutex->thread = th->self;
+    RB_OBJ_WRITTEN(self, Qundef, th->self);
```
jhawthorn, line 269:
```suggestion
    RB_OBJ_WRITE(self, &mutex->thread, th->self);
```

thread_sync.c
```diff
@@ -369,6 +397,7 @@ do_mutex_lock(VALUE self, int interruptible_p)
                 // unlocked by another thread while sleeping
                 if (!mutex->fiber) {
                     mutex->fiber = fiber;
+                    mutex->thread = th->self;
```
jhawthorn: These probably both need write barriers now that we mark them (and all the similar assignments), but I can test more thoroughly with wbcheck to be sure

PR #14265, opened 2025-08-18
ZJIT: Remove more Option from codegen

tekknolagi:


tekknolagi:
segfault in not-ZJIT :(

zjit/src/codegen.rs
```diff
@@ -1184,7 +1180,7 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         // Static symbols have (val & 0xff) == RUBY_SYMBOL_FLAG
         // Use 8-bit comparison like YJIT does
         debug_assert!(val.try_num_bits(8).is_some(), "GuardType should not be used for a known constant, but val was: {val:?}");
-        asm.cmp(val.try_num_bits(8)?, Opnd::UImm(RUBY_SYMBOL_FLAG as u64));
+        asm.cmp(val.try_num_bits(8).unwrap(), Opnd::UImm(RUBY_SYMBOL_FLAG as u64));
```
k0kubun: If we have an HIR validator for `GuardType` not using known operands and we're going to unwrap this, let's remove the `try_num_bits` function from `Opnd` and use `with_num_bits(8)` instead.

zjit/src/codegen.rs
```diff
@@ -541,22 +544,15 @@ fn gen_get_constant_path(jit: &JITState, asm: &mut Assembler, ic: *const iseq_in
...
 
-fn gen_invokebuiltin(jit: &JITState, asm: &mut Assembler, state: &FrameState, bf: &rb_builtin_function, args: Vec<Opnd>) -> Option<lir::Opnd> {
-    // Ensure we have enough room fit ec, self, and arguments
-    // TODO remove this check when we have stack args (we can use Time.new to test it)
-    if bf.argc + 2 > (C_ARG_OPNDS.len() as i32) {
-        return None;
-    }
-
+fn gen_invokebuiltin(jit: &JITState, asm: &mut Assembler, state: &FrameState, bf: &rb_builtin_function, args: Vec<Opnd>) -> lir::Opnd {
+    assert!(bf.argc + 2 <= C_ARG_OPNDS.len() as i32, "Builtin function {} has too many arguments: {}", unsafe { std::ffi::CStr::from_ptr(bf.name).to_str().unwrap() }, bf.argc);
```
k0kubun: ~~There's no such guarantee. You shouldn't crash just because a Primitive takes 5+ arguments.~~

k0kubun: Oh, you're checking it on `gen_insn`. I hope the panic message explains that it should not happen because it's checked by the caller `gen_insn`. There's nothing wrong in "Builtin function {} has too many arguments" itself.

k0kubun: e.g.

```suggestion
    assert!(
        bf.argc + 2 <= C_ARG_OPNDS.len() as i32,
        "gen_invokebuiltin should not be called for builtin function {} with too many arguments: {}",
        unsafe { std::ffi::CStr::from_ptr(bf.name).to_str().unwrap() }, bf.argc,
    );
```

zjit/src/codegen.rs
k0kubun: > let's remove the try_num_bits function from Opnd

https://github.com/ruby/ruby/pull/14272

PR #14270, opened 2025-08-19
ZJIT: Allocate register for VRegs that begin and end at the same index

tekknolagi:
If the LiveRange looks like (idx, idx), we will currently not allocate a
register. This change allocates a register and then immediately
deallocates it.

Fix https://github.com/Shopify/ruby/issues/614

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1398,6 +1398,9 @@ impl Assembler
                 _ => None,
             };
             if vreg_idx.is_some() {
+                if live_ranges[vreg_idx.unwrap()].end() == index {
+                    debug!("Allocating a register for VReg({}) at instruction index {} even though it does not live past this index", vreg_idx.unwrap(), index);
+                }
                 // This is going to be the output operand that we will set on the
                 // instruction. CCall and LiveReg need to use a specific register.
                 let mut out_reg = match insn {
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -1391,8 +1391,7 @@ impl Assembler
                 }
             }
 
-            // If the output VReg of this instruction is used by another instruction,
-            // we need to allocate a register to it
+            // Allocate a register for the output operand if it exists
             let vreg_idx = match insn.out_opnd() {
                 Some(Opnd::VReg { idx, .. }) => Some(*idx),
                 _ => None,
@@ -1480,6 +1479,8 @@ impl Assembler
                 }
             }
 
+            // If we have an output that dies at its definition (it is unused), free up the
+            // register
             if let Some(idx) = vreg_idx {
                 if live_ranges[idx].end() == index {
                     if let Some(reg) = reg_mapping[idx] {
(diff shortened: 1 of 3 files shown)


XrXr:
I guess this would turn what's currently an arcane crash message due to unused output to a silent transient register pressure. It's an improvement, but I'd want some feedback mechanism, maybe a `debug!()`? 

But also, test failures.

k0kubun:
We used to control every LIR instruction in YJIT's design, so it was useful to have a notification to remove unused LIR instructions because it was easy to identify why it happens and remove it, which also helps us simplify the implementation.

Now that we have an extra abstraction layer of HIR with multiple possible behaviors (HIR opt or not), I guess it's harder to remove unused instructions in ZJIT? I'm open to taking this direction, but I'd like to understand what's happening and why we need this, which I don't think are explained in https://github.com/Shopify/ruby/issues/614#issue-3189659922 or https://github.com/Shopify/ruby/issues/614#issuecomment-3201015150.

k0kubun:
Alternatively, if this happens only when you disable the HIR optimizer, we could take this path when it's disabled, and make it panic (probably with a better message than what we see today) when it's enabled. At least for the production mode, unused instructions are something to be fixed, somewhere in HIR, LIR, or the LIR backend.

tekknolagi:
In https://github.com/Shopify/ruby/issues/614 it happens because the HIR ends up looking like this in the block:

```
Optimized HIR:
fn block in test@testvreg.rb:4:
  v2:NilClass = Const Value(nil)
  v3:Any = Snapshot FrameState { pc: 0x0, stack: [], locals: [v1, v2] }
  v4:BasicObject = GetLocal l1, EP@3
  v5:Fixnum[1] = Const Value(1)
  Return v5
```

The `GetLocal` generates a `load` but it's never used. It's not cleaned up because we haven't run the optimizer.

k0kubun:
Thanks for the explanation. It makes sense to allow such LIR to be compiled then. I still want the `debug!` log (at least when the HIR optimizer is enabled) in case the HIR optimizer misses it though.

zjit/src/backend/lir.rs
```diff
@@ -1477,6 +1477,16 @@ impl Assembler
                 }
             }
 
+            if let Some(idx) = vreg_idx {
```
k0kubun, line 1484: It's also nice to add a comment above this to explain this is for VReg never used by other instructions.

zjit/src/backend/lir.rs
```diff
@@ -1397,7 +1397,7 @@ impl Assembler
                 Some(Opnd::VReg { idx, .. }) => Some(*idx),
```
k0kubun, line 1396: > // If the output VReg of this instruction is used by another instruction,

This comment needs some modification too. We now allocate a register for VReg not used by another instruction.

k0kubun (approved):

k0kubun:
I think you can call `rb_zjit_prepare_options()` in the `called Option::unwrap() on a None value`-ing tests to make `debug!()` work.

PR #14271, opened 2025-08-19
Fix special consts unit tests for i686

tekknolagi:
32-bit platforms do not have flonum and something about the static symbol test was flaky.


tekknolagi:
VM_CHECK_MODE messing with this for some reason...

tekknolagi:
Oh, `SHAPE_DEBUG` is defined as `(VM_CHECK_MODE > 0)` and some stuff in shape.c is gated on `SHAPE_DEBUG`---such as defining `RubyVM::Shape` at all---but I would expect that to raise a totally different error

k0kubun (approved):

PR #14273, opened 2025-08-19
ZJIT: 2 newrange fixes

XrXr:
- **ZJIT: Prepare for rb_range_new() calling <=>**
- **ZJIT: Mark Insn::NewRange as having side effects**

Fixes: https://github.com/Shopify/ruby/issues/708


tekknolagi (approved):
Look good -- we can rewrite to a hypothetical NewRangeFixnum in a pass later

k0kubun (approved):

PR #14274, opened 2025-08-19
ZJIT: Inline cfuncs into HIR

tekknolagi:


tekknolagi:
Going for https://github.com/ruby/ruby/pull/14679 instead

PR #14280, opened 2025-08-20
Adjust snt < max_cpu calculation

jhawthorn:
[[Bug #20146]](https://bugs.ruby-lang.org/issues/20146)

Previously we dealt with the main Ractor not being enabled for M:N by incrementing `snt_cnt++`. This worked for comparing against ractor count, but meant that we always had one less SNT than was specified by `RUBY_MAX_CPU`.

This was notably a problem for `RUBY_MAX_CPU=1`, which would cause Ractors to hang.

This commit instead of adjusting snt, adjusts a "`schedulable_ractor_cnt`". This way `snt_cnt` will actually reach `RUBY_MAX_CPU`.

### DIFF
--- a/thread_pthread_mn.c
+++ b/thread_pthread_mn.c
@@ -397,11 +397,15 @@ native_thread_check_and_create_shared(rb_vm_t *vm)
 
     rb_native_mutex_lock(&vm->ractor.sched.lock);
     {
-        unsigned int snt_cnt = vm->ractor.sched.snt_cnt;
-        if (!vm->ractor.main_ractor->threads.sched.enable_mn_threads) snt_cnt++; // do not need snt for main ractor
+        unsigned int schedulable_ractor_cnt = vm->ractor.cnt;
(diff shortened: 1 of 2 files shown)


thread_pthread_mn.c
```diff
@@ -397,11 +397,15 @@ native_thread_check_and_create_shared(rb_vm_t *vm)
 
     rb_native_mutex_lock(&vm->ractor.sched.lock);
     {
-        unsigned int snt_cnt = vm->ractor.sched.snt_cnt;
-        if (!vm->ractor.main_ractor->threads.sched.enable_mn_threads) snt_cnt++; // do not need snt for main ractor
```
luke-gruber, line 401: Is it possible to not increase snt_cnt here and have it work?

jhawthorn: That and no other changes? It would work, but we'd spawn one more SNT than we need until we hit `vm->ractor.sched.max_cpu` (because we'd be spawning one "for" the main Ractor).

PR #14284, opened 2025-08-20
ZJIT: Minor ergonomic things

tekknolagi:
- **ZJIT: Make a function to iterate over Insn operands**
- **ZJIT: Use Function::block and Function::block_mut**
- **ZJIT: Add BlockEditor**
- **Use BlockEditor more**


tekknolagi:
Fixed in #16305

PR #14288, opened 2025-08-20
Pre-mark classes as shareable

jhawthorn:
Classes are supposed to be always shareable. It's better for us to eagerly allocate them this way, otherwise we end up unsafely concurrently mutating their flags.

cc @ko1


ko1:
Thank you!!

I already has completely same patch with newly introducing macro, so could you wait for it?
I'll merge it in a week.

jhawthorn:
Sure! It's no problem to wait. This solves a TSAN error, but to my knowledge isn't yet causing wider issues

PR #14295, opened 2025-08-21
Fixes to encoding/transcoding for ractors.

jhawthorn:
Not all ractor-related encoding issues were fixed by 1afc07e815051e2f73493f055f2130cb642ba12a. I found more by running my test-all branch with 3 ractors for each test.

### DIFF
--- a/depend
+++ b/depend
@@ -1401,7 +1401,6 @@ compile.$(OBJEXT): $(top_srcdir)/prism/pack.h
 compile.$(OBJEXT): $(top_srcdir)/prism/parser.h
 compile.$(OBJEXT): $(top_srcdir)/prism/prettyprint.h
 compile.$(OBJEXT): $(top_srcdir)/prism/prism.h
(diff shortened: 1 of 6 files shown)


luke-gruber:
The CI is fixed. The issue was actually with `enc_arg` in `transcode.c`:
```
Encoding::Converter.asciicompat_encoding
    econv_s_asciicompat_encoding
        enc_arg
            rb_to_encoding_index
                rb_enc_find_index
                    load_encoding // creates thread when multi ractor     
```
So, that test was creating `7x500=3500` threads, and was waiting for them sequentially. I lowered it to `7x50=350` threads, but even that might be high.

I tried removing the call to `load_encoding` in `rb_enc_find_index`, because I don't like it that we try to load encodings that aren't registered. It's been there since 2008 (0052259d5e9934de20411943ce2077909e82d1d3), and when I removed it `test-all` still passed. It's possible that all the other test suites also pass without it. We should at least have a test case that covers this behavior.

We could look into removing it, but I'm not sure that's possible as it would be a breaking change.

PR #14301, opened 2025-08-21
[DOC] Write up NEWS entries for JITs

k0kubun:
This PR adds JIT-related entries to NEWS. This should be included in the release note of the upcoming preview2 release.

Since preview1 was released right before ZJIT was upstreamed, this will be the first release with ZJIT. However, I don't think we want people to test out ZJIT just yet, so I buried ZJIT entries to the very end of NEWS. In future releases, I'm hoping to move ZJIT entries to the headline section once it starts delivering meaningful speedups on benchmarks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -243,6 +243,22 @@ The following bundled gems are updated.
 
 ## JIT
 
+* YJIT
+    * YJIT stats
+        * `ratio_in_yjit` no longer works in the default build.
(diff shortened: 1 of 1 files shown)


k0kubun:
Since we don't know the exact release timing, I'll merge this now to make sure we have something. Let's make changes in follow-up PRs as neeed.

PR #14303, opened 2025-08-22
Migrate `pathname.c` to `pathname_builtin.rb`

hsbt:
Import from https://github.com/ruby/pathname/pull/57


hsbt:
I could reproduce the failures of `test_allocation.rb` and confirmed failure number is changed with code size of `pathname_builtin.rb`. So that is similar with https://github.com/ruby/ruby/pull/13906.

eregon:
Thank you for this, regarding the extra commits I think it would be best to have all/some of them in ruby/pathname too.
I can help with that.

* realdirpath should be trivial to define in Ruby.
* the TypeError is good to add in ruby/pathname too
* Does anything rely on Kernel.Pathname do be public?

test/ruby/test_allocation.rb
```diff
@@ -54,7 +54,11 @@ def self.num_allocations
         RB
 
         if num_arrays != $allocations[0]
-          failures << "Expected \#{num_arrays} array allocations for \#{check_code.inspect}, but \#{$allocations[0]} arrays allocated"
+          # TODO: Must fix this condition
+          # Windows platforms may allocate an extra array after https://github.com/ruby/ruby/pull/14303
+          unless ($allocations[0] - num_arrays == 1 && RUBY_PLATFORM =~ /mswin|mingw/)
+            failures << "Expected \#{num_arrays} array allocations for \#{check_code.inspect}, but \#{$allocations[0]} arrays allocated"
+          end
```
eregon, line 62: This is really surprising, how is this possible given this file doesn't seem to use Pathname at all?

eregon: Ah probably because of https://github.com/ruby/ruby/pull/14303#issuecomment-3213229471. I don't quite understand how having a bigger `pathname_builtin.rb` can cause that though.

eregon:
I would like to help with "upstreaming" these commits to ruby/pathname this weekend, if that's OK.
I think the changes needed here are mostly revealing missing tests in `test_pathname.rb`.

eregon:
Looking at https://github.com/ruby/ruby/actions/runs/17144600559/job/48638223862 (failures on the first commit) that's what showed the differences, so ruby/spec has extra tests that test_pathname.rb is missing. We should either run that part of ruby/spec in ruby/pathname, or port those specs to test_pathname.rb.

eregon:
All commits except [Skip rbs tests for pathname_bundle.rb](https://github.com/ruby/ruby/pull/14303/commits/9f86ecafeaefac52188bf3f0d96b1c025536b268) and [Omit extra allocation of Array object in Windows platform temporary](https://github.com/ruby/ruby/pull/14303/commits/69a231069af57480fcc98a891043430fbd7506f4) upstreamed in https://github.com/ruby/pathname/pull/60

test/ruby/test_allocation.rb
hsbt: Because I requested to separate the original PR of https://github.com/ruby/pathname/pull/57#issuecomment-3088583643.

But you ignored my suggestions and stubbornly refuse to accept them. As a result, Windows platform is broken and I waste a many of investigation time.

eregon: > Because I requested to separate the original PR of [ruby/pathname#57 (comment)](https://github.com/ruby/pathname/pull/57#issuecomment-3088583643).

As far as I can tell smaller PRs would not have helped or changed anything about that.
This PR revealed that ruby/pathname did not test against ruby/spec, which I fixed in https://github.com/ruby/pathname/pull/60.
I also found significant bugs in test_pathname.rb which I fixed in that PR too.

> But you ignored my suggestions and stubbornly refuse to accept them.

I asked you to make concrete suggestions on the PR in https://github.com/ruby/pathname/pull/57#issuecomment-3088583643, you did not in one month, so I couldn't understand what you wanted.
I also explained why I couldn't see how to meaningfully make smaller PRs for that change.

I did not understand precisely what you meant by https://github.com/ruby/ruby/pull/13906 since that seemed unrelated to Pathname as far as I could tell.
It turns out a related issue happened on this PR, but that is clearly a bug of `ObjectSpace.count_objects`, not of Pathname. BTW I tried to fix that bug https://github.com/ruby/ruby/pull/14329 to help move things forward.
Excluding the test like you did here seems fine to me, it was a bug unrelated to Pathname, basically a transient test failure waiting to happen (it even already failed before I guess as the reason for #13906) due to a bug in `ObjectSpace.count_objects`.

> As a result, Windows platform is broken

No CI was broken. Only failures on this PR when updating to latest ruby/pathname.
As far as I understand there was no urgency to merge this PR or make it green.

> and I waste a many of investigation time.

I am sorry about that, I sympathize with that feeling, it's always annoying when unexpected failures happen.
I tried to react here and on https://bugs.ruby-lang.org/issues/21532 as fast as possible and offer help.
If such a situation would happen again, please ask help and let others help, rather than spending too much of your time on it and getting frustrated.

I would also like to help with the sync of ruby/pathname <-> ruby/ruby in future, if you would like so.

PR #14304, opened 2025-08-22
Remove warnings from test suite

etiennebarrie:
This removes warnings when the test suite runs.

These appear when running the test suite with `RUBYOPT=-w` like [chkbuild does](https://github.com/ruby/chkbuild/blob/72c06b8d9d891df248d53192670181652c468510/chkbuild/ruby.rb#L687).


etiennebarrie:
cc @byroot I don't think the timeout for ppc64le is related.

PR #14308, opened 2025-08-22
ZJIT: Spill whole FrameState in `Insn::SendWithoutBlock`

XrXr:
Previously, we only spilled the arguments necessary for the particular
send. In case the callee raises and a rescue resumes the ISEQ, that
did not present a complete stack state. E.g. in `[1, (raise rescue 2)]`
the raise send only spills `self`, when `1` also needs to be spilled.

Spill the whole stack. Adjust parsing for `opt_aref_with` since the
key argument for the send now needs to be described by the frame state
of the send.

This changes the contract for `Insn::SendWithoutBlock` to use arguments
from the interpreter stack as described by its frame state.

---

`self_val` and `args` are now redundant, but I'd like to do the removal
in a separate diff to keep this one focused on the bugfix.


k0kubun (approved):
Nice! 

It seems like you need to fix HIR tests to pass zjit-check.

zjit/src/codegen.rs
```diff
@@ -859,26 +859,19 @@ fn gen_send_without_block(
...
+    asm_comment!(asm, "spill frame state");
 
     // Save PC and SP
     gen_save_pc(asm, state);
-    gen_save_sp(asm, 1 + args.len()); // +1 for receiver
+    gen_save_sp(asm, state.stack().len());
+
+    // Spill locals and stack
+    gen_spill_locals(jit, asm, state);
+    gen_spill_stack(jit, asm, state);
```
st0012, line 874: After the change, this section seems to do what `gen_prepare_non_leaf_call` does now. Do we want to invoke it here instead?

https://github.com/ruby/ruby/blob/d2995a5d11928dc75cbc784e7e3013114cedda37/zjit/src/codegen.rs#L1277-L1289

zjit/src/hir.rs
```diff
@@ -3195,9 +3204,11 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let aref_arg = fun.push_insn(block, Insn::Const { val: Const::Value(get_arg(pc, 0)) });
                     let args = vec![aref_arg];
 
+                    let mut send_state = state.clone();
+                    send_state.stack_push(aref_arg);
+                    let send_state = fun.push_insn(block, Insn::Snapshot { state: send_state });
                     let recv = state.stack_pop()?;
-                    let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
-                    let send = fun.push_insn(block, Insn::SendWithoutBlock { self_val: recv, cd, args, state: exit_id });
+                    let send = fun.push_insn(block, Insn::SendWithoutBlock { self_val: recv, cd, args, state: send_state });
```
tekknolagi, line 3211: We have the (loosely documented, if at all) invariant that the FrameState/Snapshot operand is the state *before* the instruction executes. For `opt_aref_with`, that includes leaving `key` off the stack. This is how we handle side-exits throughout HIR. For that reason, we should undo this change.

If we for some reason need to have a special method-fallback FrameState on a new instruction `OptArefWith` *in addition* to its side-exit FrameState, that's fine. But uniformity is really important here.

zjit/src/codegen.rs
```diff
@@ -859,26 +859,19 @@ fn gen_send_without_block(
...
-    // for opt_aref_with, which pushes to the stack in the middle of the instruction.
-    asm_comment!(asm, "spill receiver and arguments");
-    for (idx, &val) in [self_val].iter().chain(args.iter()).enumerate() {
-        // Currently, we don't move the SP register. So it's equal to the base pointer.
-        let stack_opnd = Opnd::mem(64, SP, idx as i32 * SIZEOF_VALUE_I32);
-        asm.mov(stack_opnd, val);
-    }
+    // Note that it's incorrect to use this frame state to side exit because
+    // the state might not be on the boundary of an interpreter instruction.
+    // For example, `opt_aref_with` pushes to the stack and then sends.
```
tekknolagi, line 865: This kind of comment is a symptom of having the wrong API in place internally. We can fix things instead of papering over them.

This is why we have `self_val` and `args`: they encode the call information needed for a `SendWithoutBlock`.

tekknolagi (commented):
I think we should do a partial revert here.

zjit/src/codegen.rs
tekknolagi: This is why I originally wanted to delegate this to some C helper function that took a vector of args from the native stack: we could set up the VM frame with the FrameState stack and then let the C helper set up the VM stack appropriately.

tekknolagi: `rb_call0` type thing

XrXr: Since this is ultimately an interpreter operation, it's more natural and efficient to follow the interpreter's API. 

> This is why we have self_val and args: they encode the call information needed for a SendWithoutBlock.

We also have CD, which contains the call info that the interpreter uses, that also describes where `self_val` and `args` are. I understand it's nicer and more natural to have the operands as fields, but these they're not used and ultimately redundant.

We could choose to do redundant work with an API that more directly use `self_val`, `args` but we'd be sacrificing efficiency in the process, and also introduce a moving part in mapping between the two APIs.

zjit/src/codegen.rs
XrXr: I opted against it because

> // TODO: Lazily materialize caller frames when needed
 
Presumably gen_prepare_non_leaf_call() would do light weight frame stuff when that's implemented, but what this calls for is a direct materialization of the frame state.

This is future coding and I can go either way on it, though.

zjit/src/codegen.rs
k0kubun: I think we just need two different `FrameState` arguments in `gen_send_without_block`, one being `Option<FrameState>` which `opt_aref_with` sets `Some` to. It should be all the API change we need here.

This is how `opt_aref_with` works:

1. Call `rb_hash_aref` if `recv` is Hash
2. If not, _push `key` to the stack and_ dispatch `opt_send_without_block`

Since it's `gen_send_without_block` here, we definitely need a `FrameState` that is used by `gen_spill_stack` right before the call, which has `key` on stack. It's unfortunately incorrect to side-exit to `opt_aref_with` with `key` on stack, so we need a different `FrameState` for side-exit.

tekknolagi: > I understand it's nicer and more natural to have the operands as fields, but these they're not used and ultimately redundant.

They are not redundant! They are used to directly model the inputs to an operation. We use them in, for example, the optimizer.

The FrameState is related but it answers different questions.

k0kubun: Sorry, I made the above comment while thinking about `gen_send_without_block_direct`, but this was `opt_send_without_block` doing dynamic dispatch.

One of the reasons why we use dynamic dispatch on JIT is to avoid side exits. We shouldn't need to side-exit in `gen_send_without_block` in the first place, so I guess we won't need an extra `FrameState` argument in this function.

The generated code is exactly what we need before dynamic dispatch today, so I don't see any reason to introduce a call to a C helper function.

k0kubun: Anyway, if we want to give `state: FrameState` that doesn't have `key` to HIR for `opt_aref_with` for whatever reason, we need another `FrameState` that has `key` for the same HIR and use that in `gen_send_without_block()`.

zjit/src/codegen.rs
st0012: Thanks the clarifying 👍

PR #14311, opened 2025-08-23
Fix thread_profile_frames crashing due to uninitialized PC

XrXr:
ZJIT leaves cfp->pc uninitialized in many instances, and the conditions
in thread_profile_frames() was outdated, leading sampling profilers to
crash.

Check `cfp->pc` for validity in a way that works for both YJIT and ZJIT.

### DIFF
--- a/vm_backtrace.c
+++ b/vm_backtrace.c
@@ -1761,15 +1761,22 @@ thread_profile_frames(rb_execution_context_t *ec, int start, int limit, VALUE *b
             }
 
             if (lines) {
-                // The topmost frame may not have an updated PC because the JIT
-                // may not have set one.  The JIT compiler will update the PC
-                // before entering a new function (so that `caller` will work),
-                // so only the topmost frame could possibly have an out of date PC
-                if (cfp == top && cfp->jit_return) {
+                const VALUE *pc = cfp->pc;
+                VALUE *iseq_encoded = ISEQ_BODY(cfp->iseq)->iseq_encoded;
+                VALUE *pc_end = iseq_encoded + ISEQ_BODY(cfp->iseq)->iseq_size;
+
+                // The topmost frame may have an invalid PC because the JIT
+                // may leave it uninitialized for speed. JIT code must update the PC
+                // before entering a non-leaf method (so that `caller` will work),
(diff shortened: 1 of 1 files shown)


XrXr:
Unfortunately, it's hard to write a stable test for this type of thing. I did verify that the change fixes ZJIT+Stackprof/Vernier usage in an internal app though.

vm_backtrace.c
```diff
@@ -1761,15 +1761,22 @@ thread_profile_frames(rb_execution_context_t *ec, int start, int limit, VALUE *b
...
+                VALUE *iseq_encoded = ISEQ_BODY(cfp->iseq)->iseq_encoded;
+                VALUE *pc_end = iseq_encoded + ISEQ_BODY(cfp->iseq)->iseq_size;
+
+                // The topmost frame may have an invalid PC because the JIT
+                // may leave it uninitialized for speed. JIT code must update the PC
+                // before entering a non-leaf method (so that `caller` will work),
+                // so only the topmost frame could possibly have an out-of-date PC.
+                // ZJIT doesn't set `cfp->jit_return`, so it's not a reliable signal.
+                //
+                // Avoid passing invalid PC to calc_lineno() to avoid crashing.
```
ko1, line 1774: Just curious:
if unitialized PC (random value) in correct range (in condition), does it work well?
(of course it is too rare)

XrXr: If the uninitialized value happens to be in range, the line number we report would be wrong. But at least we won't SEGV.

ko1: I see. `calc_lineno()` doesn't require alignments.

vm_backtrace.c
```diff
@@ -1761,15 +1761,22 @@ thread_profile_frames(rb_execution_context_t *ec, int start, int limit, VALUE *b
...
+                if (cfp == top && (pc < iseq_encoded || pc > pc_end)) {
```
jhawthorn, line 1775: Should this be `pc >= pc_end`?

XrXr: pc == pc_end is valid, because PC is bumped by instruction width before the body of the instruction, so the last instruction in an iseq can have PC at one-past-end.

PR #14313, opened 2025-08-24
[DOC] Tweaks for String#inspect

BurdetteLamar:


doc/string/inspect.rdoc
```diff
@@ -0,0 +1,57 @@
+
```
peterzhu2118: Nit: why is there a blank line on the first line?

doc/string/inspect.rdoc
```diff
@@ -0,0 +1,57 @@
...
+  [23, "\"\\x17\""]
+  [24, "\"\\x18\""]
+  [25, "\"\\x19\""]
+  [26, "\"\\x1A\""]
+  [27, "\"\\e\""]     # ESC
+  [28, "\"\\x1C\""]
+  [29, "\"\\x1D\""]
+  [30, "\"\\x1E\""]
+  [31, "\"\\x1F\""]
+  [127, "\"\\x7F\""]  # DEL
```
peterzhu2118: I think we should have a section for specially handled ASCII control characters (e.g. BEL, BS, etc.) and another section for the non specially handled ASCII control characters that just output as hexadecimal. That way, we also don't need to show examples of every ASCII control character.

doc/string/inspect.rdoc
BurdetteLamar: Removed.

doc/string/inspect.rdoc
BurdetteLamar: Fixed.

peterzhu2118 (approved):

PR #14314, opened 2025-08-24
[DOC] Tweaks for String#intern

BurdetteLamar:


doc/string/intern.rdoc
```diff
@@ -0,0 +1,16 @@
...
+  'foo'.intern       # => :foo
+  'тест'.intern      # => :тест
+  'こんにちは'.intern # => :こんにちは
+
+May be useful for creating a symbol that cannot be created as a symbol literal:
+
+  'foo bar'.intern              # => :"foo bar"
+  ':foo'.intern                 # => :":foo"
+  '\\foo'.intern                # => :"\\foo"
+  "foo\tbar\nbaz\tbat\n".intern # => :"foo\tbar\nbaz\tbat\n"
```
peterzhu2118: I realize that you didn't write this and this was how the documentation was, but I'm not sure what this actually means. You can create these symbols literals by simply encasing them in quotes like how the return value looks, for example:

```ruby
'foo bar'.intern == :'foo bar' => true
```

BurdetteLamar: Removed.

peterzhu2118 (approved):

PR #14315, opened 2025-08-24
[DOC] Tweaks for String#length

BurdetteLamar:


doc/string/length.rdoc
```diff
@@ -2,11 +2,12 @@ Returns the count of characters (not bytes) in +self+:
 
   'foo'.length        # => 3
   'тест'.length       # => 4
-  'こんにちは'.length   # => 5
+  'こんにちは'.length  # => 5
```
peterzhu2118, line 5: This breaks the alignment rendered:

Before:

Screenshot 2025-08-25 at 9 39 01 AM

After:

Screenshot 2025-08-25 at 9 39 15 AM

BurdetteLamar: I think the alignment is broken either way.  I think I've usually done what I've done here.  Does it matter?

peterzhu2118: I think it's better in the original way since it's aligned in the source code, but RDoc rendering is making it not aligned.

BurdetteLamar: It's also not aligned in the source:
image

The hash in the first Japanese line is at column 19 (per RubyMine), but the hashes above it are at column 23.

I'm a total agnostic on this point;  just gimme a rule I can work with.

peterzhu2118 (approved):

PR #14326, opened 2025-08-25
ZJIT: Migrate to insta

tekknolagi:
Migrate from expect-test to insta snapshot testing library. While expect-test
is very small and has a limited surface, it does not handle nextest well; it
while doing inline snapshot updates, it races other test processes to update
the file, leading to frequent "Failed to process macro invocation" errors.

insta handles this by doing batches; it writes to temporary files which can
then be committed in a batch using `cargo insta review` or
`INSTA_UPDATE=always`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,19 +32,32 @@ dependencies = [
 ]
 
 [[package]]
-name = "dissimilar"
-version = "1.0.10"
+name = "console"
+version = "0.15.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tekknolagi:
Putting this up for commentary: what do you think of the change from expect-test to insta?

XrXr:
Seems nice. Any way to get it to indent the expectation the same way?

tekknolagi:
Huh. I had not noticed the indentation difference. That would be nice....

tekknolagi:
If I am reading the source right, then no, there is no way to configure it a different way than indenting at the same indentation level as the macro invocation.

k0kubun (approved):
I can live with the same-level indentation if insta needs it. It's still better than no indentation for those heredocs.

PR #14333, opened 2025-08-25
ZJIT: Resurrect string key of `opt_aref_with`

XrXr:
The immediate problem this fixes is the frozen status of the key object
in the `recv[key]` call.

This also repairs an invariant that `FrameState` should always be at a
YARV instruction boundary. It's important to have this because if a
`SendWithoutBlock` is reduced to `SendWithoutBlockDirect`, the direct
variant may want to side exit, which needs a `FrameState` as described.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -370,6 +370,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::InvokeBuiltin { bf, .. } if bf.argc + 2 > (C_ARG_OPNDS.len() as i32) => return None,
         Insn::InvokeBuiltin { bf, args, state, .. } => gen_invokebuiltin(jit, asm, &function.frame_state(*state), bf, opnds!(args)),
         Insn::Return { val } => no_output!(gen_return(asm, opnd!(val))),
+        &Insn::ArefWith { receiver: _, key, cd, send_state } => gen_opt_aref_with(jit, asm, cd, opnd!(key), &function.frame_state(send_state)),
         Insn::FixnumAdd { left, right, state } => gen_fixnum_add(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state)),
         Insn::FixnumSub { left, right, state } => gen_fixnum_sub(jit, asm, opnd!(left), opnd!(right), &function.frame_state(*state)),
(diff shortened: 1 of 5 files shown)


zjit/src/hir.rs
```diff
@@ -564,6 +564,9 @@ pub enum Insn {
         return_type: Option<Type>,  // None for unannotated builtins
     },
 
+    /// Parallel of `opt_aref_with` in the interpreter.
+    ArefWith { receiver: InsnId, key: InsnId, send_state: InsnId, cd: *const rb_call_data },
```
tekknolagi, line 568: Let's do self_val to remain consistent with everything else and then rename all at once

tekknolagi, line 568: Let's call it `Aref` and then split out the resurrection behavior into a separate opcode `StringResurrect`. Modular IR

tekknolagi, line 568: Let's call it `state` to remain consistent, not `send_state`

XrXr:
ha 

>  TestHashOnly#test_AREF_fstring_key [/Users/runner/work/ruby/ruby/src/test/ruby/test_hash.rb:2011]:
  <31620> expected but was
  <31625>.

This is asserting allocation count, so the test essentially forces ZJIT to do the same optimization as the interpreter in `opt_aref_with`. Similar stuation as https://github.com/Shopify/ruby/issues/678

zjit/src/codegen.rs
```diff
@@ -960,6 +959,31 @@ fn gen_send_without_block_direct(
     ret
 }
 
+/// Compile the equivalent of the general path of `opt_aref_with`
+fn gen_opt_aref_with(jit: &mut JITState, asm: &mut Assembler, cd: *const rb_call_data, key: Opnd, state: &FrameState) -> Opnd {
+    // First, resurrect the string key. Save PC before allocating.
+    gen_save_pc(asm, state);
```
k0kubun, line 965:
```suggestion
    gen_prepare_call_with_gc(asm, state);
```

Doing this would be useful for things like https://github.com/Shopify/ruby/issues/709.

At this point, maybe we should rename `gen_save_pc` to `gen_prepare_call_with_gc` to make sure we use the correct one. The need of "prepare call with GC" applies to all current callsites of `gen_save_pc`.

st0012:
JFYI https://github.com/ruby/ruby/pull/14336 will remove `opt_aref_with` completely

XrXr:
Since I don't have time to do this this week and Aaron might just remove the instruction I'll just let this sit.

PR #14300, opened 2025-08-21
ZJIT: Check interrupts

tekknolagi:
- **ZJIT: Add CheckInterrupts HIR instruction**
- **ZJIT: Eagerly place CheckInterrupts**
- **ZJIT: Update tests**

Fix https://github.com/Shopify/ruby/issues/589

PR #14336, opened 2025-08-25
Remove `opt_aref_with` and `opt_aset_with`

tenderlove:
When these instructions were introduced it was common to read from a hash with mutable string literals.  However, these days, I think these instructions are fairly rare.

I tested this with the lobsters benchmark, and saw no difference in speed.  In order to be sure, I tracked down every use of this instruction in the lobsters benchmark, and there were only 4 places where it was used.

Additionally, this patch fixes a case where "chilled strings" should emit a warning but they don't.

```ruby
class Foo
  def self.[](x)= x.gsub!(/hello/, "hi")
end

Foo["hello world"]
```

Removing these instructions shows this warning:

```
> ./miniruby -vw test.rb
ruby 3.5.0dev (2025-08-25T21:36:50Z rm-opt_aref_with dca08e286c) +PRISM [arm64-darwin24]
test.rb:2: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```


k0kubun:
Since https://github.com/ruby/ruby/pull/8118 benchmarked YJIT's `opt_aref_with` with `erbui-rails`, I also benchmarked `erubi-rails`, but it didn't show a difference either.

```
before: ruby 3.5.0dev (2025-08-25T20:16:34Z master e49c29ed75) +PRISM [x86_64-linux]
after: ruby 3.5.0dev (2025-08-25T23:53:54Z rm-opt_aref_with 2651e8b4a3) +PRISM [x86_64-linux]

erubi-rails  1957.2       0.4         1941.1      0.3         1.122          1.008
```

```
before: ruby 3.5.0dev (2025-08-25T20:16:34Z master e49c29ed75) +YJIT +PRISM [x86_64-linux]
after: ruby 3.5.0dev (2025-08-25T23:53:54Z rm-opt_aref_with 2651e8b4a3) +YJIT +PRISM [x86_64-linux]

erubi-rails  904.4        0.3         907.3       0.9         0.813          0.997
```

k0kubun (approved):
seems fine to me.

I hope ZJIT will eventually get rid of the unnecessary `rb_str_resurrect` call through HIR.

yjit/src/codegen.rs
```diff
@@ -4017,38 +3983,6 @@ fn gen_opt_aset(
...
-        rb_vm_opt_aref_with as *const u8,
-        vec![
-            recv_opnd,
-            key_opnd
-        ],
-    );
-    asm.stack_pop(1); // Keep it on stack during GC
-
-    asm.cmp(val_opnd, Qundef.into());
-    asm.je(Target::side_exit(Counter::opt_aref_with_qundef));
```
st0012, line 4044: I think we can now delete `opt_aref_with_qundef` and `opt_aset_with_qundef` from YJIT's counters too?

st0012 (approved):

yjit/src/codegen.rs
tenderlove: Good catch! Thank you!

tekknolagi:
Hm. You may want to add a test exclude for ZJIT

k0kubun:
> TestGc#test_interrupt_in_finalizer

Was https://github.com/ruby/ruby/pull/14300 not enough? I agree with adding it to test excludes for now.

tekknolagi:
It worked before CheckInterrupts and still works after---on macOS arm64. I wonder what the deal is with Ubuntu or x86.

tenderlove:
> Was #14300 not enough? I agree with adding it to test excludes for now.

I think I added one. I've never added test exclusion before so I'm not sure if I did it correctly.

PR #14339, opened 2025-08-26
Revert prism-deps branch changes

nobu:
Failing on BSD make, revert for now.


hsbt:
I added GNU make to our OpenBSD instance and `MAKE=gmake` for chkbuild. It resolved the build issue now.

jeremyevans:
> I added GNU make to our OpenBSD instance and `MAKE=gmake` for chkbuild. It resolved the build issue now.

I think we should still revert. Ruby is designed to build correctly with both BSD and GNU make. Prism needs to build with both.

PR #14341, opened 2025-08-26
Add RubyVM::AST::Node#parent (only for NODE_SCOPE)

mame:
`RubyVM::AST.of(Method object)` returns a NODE_SCOPE node of the method. This internal API allows to get a reference to the outer NODE_DEFN (or NODE_DEFS) node.

This is preparation for [Feature #21543].


mame:
Superseded by #14359

PR #14345, opened 2025-08-26
CI: Use `nproc` to count only on-line CPUs for GNUMAKEFLAGS

junaruga:
Use `nproc` rather than `nproc --all`. Because the number by the `nproc` is different from the number by the `nproc --all` on GitHub Actions ppc64le/s390x. This caused the `make` command runs much more jobs than the number of on-line CPUs. The make command ran 193 jobs in parallel for 4 on-line CPUs in GitHub Actions ppc64le, and ran 9 jobs in parallel for 4 on-line CPUs in GitHub Actions s390x.

And this caused the high load average 34.58 on ppc64le, and 6.69 on s390x. These values should be less than 4.0. I believe we should use the `nproc` rather than `nproc --all`.

```
+ nproc
4
+ nproc --all
192
```

```
+ nproc
4
+ nproc --all
8
```

See https://github.com/IBM/actionspz/issues/38 for details.

### DIFF
--- a/.github/actions/compilers/entrypoint.sh
+++ b/.github/actions/compilers/entrypoint.sh
@@ -26,7 +26,7 @@ export CONFIGURE_TTY='never'
 export RUBY_DEBUG='ci rgengc'
 export RUBY_TESTOPTS='-q --color=always --tty=no'
 export RUBY_DEBUG_COUNTER_DISABLE='1'
-export GNUMAKEFLAGS="-j$((1 + $(nproc --all)))"
+export GNUMAKEFLAGS="-j$((1 + $(nproc)))"
 
 case "x${INPUT_ENABLE_SHARED}" in
 x | xno | xfalse )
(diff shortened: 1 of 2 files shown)


junaruga:
@shyouhei I see you added the `.github/actions/setup/directories/action.yml` at the commit https://github.com/ruby/ruby/commit/420d069cc2b9abcc1c7cc6a5d3cd6731bac4f54c. Perhaps you also might added the logic for the `nproc --all`. I would like you to review this PR's change. Thanks!

shyouhei (approved):
At that time I added this code, hyper threading was still a thing.  Under such circumstance `nproc --all` boosted overall CI speed.  But that was an old story.  Now that datacenter CPUs employ completely different setups. Removing `--all` just makes sense today.  LGTM.

junaruga:
Thanks for your review and info! I updated the commit message adding the context about the hyper threading, and rebasing this PR on the latest master branch.

PR #14347, opened 2025-08-26
ZJIT: Give --zjit-dump-hir-graphviz optional filename

tekknolagi:
- **ZJIT: Graphviz point to top of params**
- **ZJIT: Dump graphviz output to given filename**

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2244,8 +2244,11 @@ impl Function {
             None => {},
         }
 
-        if get_option!(dump_hir_graphviz) {
-            println!("{}", FunctionGraphvizPrinter::new(&self));
+        if let Some(filename) = &get_option!(dump_hir_graphviz) {
+            use std::fs::OpenOptions;
+            use std::io::Write;
+            let mut file = OpenOptions::new().append(true).open(filename).unwrap();
+            writeln!(file, "{}", FunctionGraphvizPrinter::new(&self)).unwrap();
         }
     }
 
--- a/zjit/src/options.rs
+++ b/zjit/src/options.rs
@@ -44,7 +44,8 @@ pub struct Options {
     /// Dump High-level IR after optimization, right before codegen.
     pub dump_hir_opt: Option<DumpHIR>,
 
-    pub dump_hir_graphviz: bool,
+    /// Dump High-level IR to the given file in Graphviz format after optimization
+    pub dump_hir_graphviz: Option<std::path::PathBuf>,
 
     /// Dump low-level IR
     pub dump_lir: bool,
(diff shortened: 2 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -5701,26 +5704,26 @@ mod graphviz_tests {
...
-        </TABLE>>];
-          bb0:v6 -> bb1:params;
-          bb1 [label=<<TABLE BORDER="0" CELLBORDER="1" CELLSPACING="0">
-        <TR><TD ALIGN="LEFT" PORT="params" BGCOLOR="gray">bb1(v11:BasicObject, v12:BasicObject)&nbsp;</TD></TR>
-        <TR><TD ALIGN="left" PORT="v14">v14:Fixnum[4] = Const Value(4)&nbsp;</TD></TR>
-        <TR><TD ALIGN="left" PORT="v16">CheckInterrupts&nbsp;</TD></TR>
-        <TR><TD ALIGN="left" PORT="v17">Return v14&nbsp;</TD></TR>
-        </TABLE>>];
-        }
+            digraph G { # test@&lt;compiled&gt;:3
```
k0kubun: So the same-level indentation https://github.com/ruby/ruby/pull/14326#issuecomment-3221347282 was not required?

tekknolagi: ........huh. I actually don't know. For the last PR I had just let insta rewrite everything

tekknolagi: This spacing was the result of a weird rebase I think

zjit/src/hir.rs
```diff
@@ -2244,8 +2244,11 @@ impl Function {
             None => {},
         }
 
-        if get_option!(dump_hir_graphviz) {
-            println!("{}", FunctionGraphvizPrinter::new(&self));
+        if let Some(filename) = &get_option!(dump_hir_graphviz) {
+            use std::fs::OpenOptions;
+            use std::io::Write;
+            let mut file = OpenOptions::new().append(true).open(filename).unwrap();
```
XrXr, line 2250: Reopening the file multiple times doesn't work well when the process changes CWD. Happens in some benchmarks in `yjit-bench` https://github.com/ruby/ruby/pull/11004

tekknolagi: augh

tekknolagi: might require this to be a fully qualified path then

XrXr: I guess this can be fixed by just calling https://doc.rust-lang.org/beta/std/fs/fn.canonicalize.html to normalize to an absolute path.

tekknolagi: done. thanks

PR #13972, opened 2025-07-22
CI: ubuntu.yml: Add GitHub Actions s390x case

junaruga:
/cc @hsbt

Add the ubuntu-22.04-ppc64le and ubuntu-22.04-s390x to align the following
hosts running Ubuntu 22.04 on Ruby CI.

https://rubyci.org/
* ppc64le (Ubuntu)
* s390x (Ubuntu)

These GitHub Actions ppc64le/s390x pipelines are managed by the following project.
https://github.com/IBM/actionspz

The onboarding document is below.
https://github.com/IBM/actionspz/blob/main/docs/onboarding.md

This PR is related to https://github.com/IBM/actionspz/issues/4.

PR #14353, opened 2025-08-26
Build ppc64le and s390x jobs in a separate workflow

k0kubun:
## Description
IBM-sponsored environments (s390x in particular) seem to have a significantly lower capacity compared to publicly-available ones (there's a job not starting after 1h). Because it's currently part of `ubuntu.yml`, a required status check, it's preventing pull requests from being auto-merged quickly.

To address this issue, this PR removes these jobs from the required status check.

Screenshot 2025-08-26 at 11 40 17

## Note
We deliberately avoided using matrix builds and enumerated every job for `ubuntu.yml` to save CI time https://github.com/ruby/ruby/pull/10327. https://github.com/ruby/ruby/pull/13972 introduced matrix builds again for using `excludes`, but the requirement of `excludes` is very specific to IBM jobs, so separating them into a separate workflow allows us to stay away from matrix builds in `ubuntu.yml`.

https://github.com/ruby/ruby/pull/13972 also added a non-trivial amount of s390x/ppc64le-specific steps, so it deserves a separate workflow from that point of view as well.

### DIFF
--- a/.github/workflows/ubuntu-ibm.yml
+++ b/.github/workflows/ubuntu-ibm.yml
@@ -1,4 +1,4 @@
-name: Ubuntu
+name: Ubuntu IBM
 on:
   push:
     paths-ignore:
@@ -28,8 +28,6 @@ jobs:
         configure: ['']
         arch: ['']
         os:
-          - ubuntu-24.04
-          - ubuntu-24.04-arm
           # FIXME Comment out ppc64le due to failing tests on GitHub Actions
           # ppc64le
           # https://bugs.ruby-lang.org/issues/21534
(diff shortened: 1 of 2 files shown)


k0kubun:
@junaruga Let me know if you have a problem with this change. I'll merge this to address the stuck auto-merge issue, but I'd appreciate your thoughts after that too.

k0kubun:
> there's a job not starting after 1h

For the record, it could be 2h+.

Screenshot 2025-08-26 at 12 57 49

junaruga:
> @junaruga Let me know if you have a problem with this change. I'll merge this to address the stuck auto-merge issue, but I'd appreciate your thoughts after that too.

Thank you for the PR! No problem! I definitely agree on managing the Ubuntu ppc64le/s390x cases as a separate workflow YML file from the `.github/workflows/ubuntu.yml`, considering passing all the cases in the `.github/workflows/ubuntu.yml` is required to merge PRs, and the current situation that ppc64le/s390x jobs are sometimes not starting. I already reported a similar issue https://github.com/IBM/actionspz/issues/34 for the ppc64le. However, today I have also seen the not-starting s390x jobs too. Perhaps people at IBM may be doing something related to the issue https://github.com/IBM/actionspz/issues/38. But I still wish they wouldn't stop the CI services.

Duplicated contents between the `.github/workflows/ubuntu.yml` and `.github/workflows/ubuntu-ibm.yml` are a challenge. However, we still have to split the `ubuntu.yml` for now. After observing the GitHub Actions ppc64le/s390x for a certain time, and if we see the CI services are stable and mature enough, we may be able to move ppc64le/s390x cases to the `ubuntu.yml` again in the future. But now is definitely not a good time to do it.

PR #14357, opened 2025-08-26
ZJIT: Implement side exit stats

k0kubun:
This PR adds stats that count every side exit. They are counted for each YARV instruction like YJIT does. To compare them with how often we hit failed ISEQs, such ISEQs get a JIT entry that just increments `exit_compilation_failure`.

It also removes `failed_gen_insn` and `failed_gen_insn_unexpected` because they don't fail compilation since https://github.com/ruby/ruby/pull/14212. `failed_iseq_stack_too_large` is now counted as a compilation failure reason.

### Example
On lobsters,

```
***ZJIT: Printing ZJIT statistics on exit***
compilation failure reasons:
  hir_compile: 646 (70.1%)
  asm_compile: 276 (29.9%)
compiled_iseq_count: 6,431
compilation_failure: 922
compile_time:        4,642ms
profile_time:        56ms
gc_time:             63ms
invalidation_time:   11ms
side_exit_count:     5,099,598
total_insns_count:   168,991,101
vm_insns_count:      74,679,899
zjit_insns_count:    94,311,202
ratio_in_zjit:       55.8%
Top-20 side exit reasons (99.9% of all 41):
          compilation_failure: 1,830,456 (26.4%)
                         send: 1,605,136 (23.2%)
       opt_send_without_block:   886,705 (12.8%)
                  invokesuper:   609,514 ( 8.8%)
           getblockparamproxy:   522,395 ( 7.5%)
                  invokeblock:   318,324 ( 4.6%)
                  definedivar:   286,818 ( 4.1%)
         opt_getconstant_path:   221,503 ( 3.2%)
                    opt_nil_p:   197,064 ( 2.8%)
  zjit_opt_send_without_block:   120,425 ( 1.7%)
                 checkkeyword:    88,575 ( 1.3%)
                  expandarray:    79,829 ( 1.2%)
            opt_case_dispatch:    48,437 ( 0.7%)
            opt_newarray_send:    41,922 ( 0.6%)
                   splatarray:    29,403 ( 0.4%)
                        throw:    16,003 ( 0.2%)
             getclassvariable:     9,904 ( 0.1%)
                  objtostring:     7,264 ( 0.1%)
                getblockparam:     3,393 ( 0.0%)
                      opt_neq:     1,563 ( 0.0%)
```

### DIFF
--- a/zjit.rb
+++ b/zjit.rb
@@ -52,11 +52,13 @@ class << RubyVM::ZJIT
       :gc_time_ns,
       :invalidation_time_ns,
 
+      :side_exit_count,
       :total_insns_count,
       :vm_insns_count,
       :zjit_insns_count,
       :ratio_in_zjit,
     ], buf:, stats:)
+    print_counters_with_prefix(prefix: 'exit_', prompt: 'side exit reasons', buf:, stats:, limit: 20)
 
     buf
   end
@@ -70,7 +72,7 @@ class << RubyVM::ZJIT
   private
 
   def print_counters(keys, buf:, stats:)
-    left_pad = keys.map(&:size).max + 1
+    left_pad = keys.map { |key| key.to_s.sub(/_time_ns\z/, '_time').size }.max + 1
     keys.each do |key|
       # Some stats like vm_insns_count and ratio_in_zjit are not supported on the release build
       next unless stats.key?(key)
@@ -90,11 +92,28 @@ class << RubyVM::ZJIT
     end
   end
 
-  def print_counters_with_prefix(buf:, stats:, prefix:, prompt:)
-    keys = stats.keys.select { |key| key.start_with?(prefix) && stats[key] > 0 }
-    unless keys.empty?
-      buf << "#{prompt}:\n"
-      print_counters(keys, buf:, stats:)
+  def print_counters_with_prefix(buf:, stats:, prefix:, prompt:, limit: nil)
+    counters = stats.select { |key, value| key.start_with?(prefix) && value > 0 }
+    return if stats.empty?
+
+    counters.transform_keys! { |key| key.to_s.delete_prefix!(prefix) }
+    left_pad = counters.keys.map(&:size).max
+    right_pad = counters.values.map { |value| number_with_delimiter(value).size }.max
+    total = counters.values.sum
+    count = counters.size
+
+    counters = counters.to_a
+    counters.sort_by! { |_, value| -value }
+    counters = counters.first(limit) if limit
+
+    buf << "Top-#{counters.size} " if limit
+    buf << "#{prompt}"
+    buf << " (%.1f%% of all #{count})" % (100.0 * counters.map(&:last).sum / total) if limit
+    buf << ":\n"
+    counters.each do |key, value|
+      padded_key = key.rjust(left_pad, ' ')
+      padded_value = number_with_delimiter(value).rjust(right_pad, ' ')
+      buf << "  #{padded_key}: #{padded_value} (%4.1f%%)\n" % (100.0 * value / total)
     end
   end
 
--- a/zjit.rb
+++ b/zjit.rb
@@ -29,9 +29,9 @@ class << RubyVM::ZJIT
     stats = Primitive.rb_zjit_stats(key)
     return stats if stats.nil? || !key.nil?
 
-    if stats.key?(:vm_insns_count) && stats.key?(:zjit_insns_count)
-      stats[:total_insns_count] = stats[:vm_insns_count] + stats[:zjit_insns_count]
-      stats[:ratio_in_zjit] = 100.0 * stats[:zjit_insns_count] / stats[:total_insns_count]
+    if stats.key?(:vm_insn_count) && stats.key?(:zjit_insn_count)
+      stats[:total_insn_count] = stats[:vm_insn_count] + stats[:zjit_insn_count]
+      stats[:ratio_in_zjit] = 100.0 * stats[:zjit_insn_count] / stats[:total_insn_count]
     end
 
     stats
@@ -53,9 +53,9 @@ class << RubyVM::ZJIT
       :invalidation_time_ns,
 
       :side_exit_count,
-      :total_insns_count,
-      :vm_insns_count,
-      :zjit_insns_count,
+      :total_insn_count,
+      :vm_insn_count,
+      :zjit_insn_count,
       :ratio_in_zjit,
     ], buf:, stats:)
     print_counters_with_prefix(prefix: 'exit_', prompt: 'side exit reasons', buf:, stats:, limit: 20)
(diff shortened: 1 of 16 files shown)


zjit/src/backend/lir.rs
```diff
@@ -1584,14 +1585,27 @@ impl Assembler
...
                 asm_comment!(self, "save cfp->pc");
-                self.load_into(Opnd::Reg(Assembler::SCRATCH_REG), Opnd::const_ptr(pc));
+                self.load_into(SCRATCH_OPND, Opnd::const_ptr(pc));
                 self.store(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_PC), Opnd::Reg(Assembler::SCRATCH_REG));
 
                 asm_comment!(self, "save cfp->sp");
-                self.lea_into(Opnd::Reg(Assembler::SCRATCH_REG), Opnd::mem(64, SP, stack.len() as i32 * SIZEOF_VALUE_I32));
+                self.lea_into(SCRATCH_OPND, Opnd::mem(64, SP, stack.len() as i32 * SIZEOF_VALUE_I32));
                 let cfp_sp = Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SP);
                 self.store(cfp_sp, Opnd::Reg(Assembler::SCRATCH_REG));
```
tekknolagi:
```suggestion
                self.store(cfp_sp, SCRATCH_OPND);
```

zjit/src/backend/lir.rs
```diff
@@ -1584,14 +1585,27 @@ impl Assembler
                 }
 
                 asm_comment!(self, "save cfp->pc");
-                self.load_into(Opnd::Reg(Assembler::SCRATCH_REG), Opnd::const_ptr(pc));
+                self.load_into(SCRATCH_OPND, Opnd::const_ptr(pc));
                 self.store(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_PC), Opnd::Reg(Assembler::SCRATCH_REG));
```
tekknolagi:
```suggestion
                self.store(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_PC), SCRATCH_OPND);
```

zjit/src/backend/lir.rs
```diff
@@ -1584,14 +1585,27 @@ impl Assembler
...
+                    asm_comment!(self, "increment an exit counter");
+                    self.load_into(SCRATCH_OPND, Opnd::const_ptr(exit_counter_ptr(pc)));
+                    let counter_opnd = if cfg!(target_arch = "aarch64") { // See arm64_split()
+                        // Using C_CRET_OPND since arm64_emit uses both SCRATCH0 and SCRATCH1 for IncrCounter.
+                        self.lea_into(C_RET_OPND, Opnd::mem(64, SCRATCH_OPND, 0));
+                        C_RET_OPND
+                    } else { // x86_emit expects Opnd::Mem
+                        Opnd::mem(64, SCRATCH_OPND, 0)
+                    };
+                    self.incr_counter(counter_opnd, 1.into());
```
tekknolagi, line 1606: What if we just pass in the `Opnd::const_ptr(exit_counter_ptr(pc))` and let `incr_counter`/`IncrCounter` figure out what to do with it per architecture? This feels like it unnecessarily leaks architecture details upward

zjit/src/codegen.rs
```diff
@@ -1596,6 +1592,20 @@ pub fn gen_function_stub_hit_trampoline(cb: &mut CodeBlock) -> Option<CodePtr> {
     })
 }
 
+/// Generate a trampoline that increments exit_compilation_failure and jumps to exit_trampoline.
+pub fn gen_function_stub_exit_trapmoline(cb: &mut CodeBlock, exit_trampoline: CodePtr) -> Option<CodePtr> {
```
tekknolagi: s/trapmoline/trampoline

zjit.rb
```diff
@@ -52,11 +52,13 @@ def stats_string
       :gc_time_ns,
       :invalidation_time_ns,
 
+      :side_exit_count,
```
st0012, line 55: nit: should this be plural `side_exits_count` to match other `*_insns_count`?

zjit/src/codegen.rs
```diff
@@ -1518,7 +1514,7 @@ c_callable! {
             } else {
                 // Exit to the interpreter
                 spill_stack(iseq, cfp, sp);
-                ZJITState::get_exit_trampoline()
+                ZJITState::get_function_stub_exit_trampoline()
```
tekknolagi: Can we give this a different name that better communicates what it does? It's very similar to the other one and it's all confusing to me

zjit/src/backend/lir.rs
k0kubun: It's a deliberate choice to avoid introducing `SCRATCH2` for arm64.

Nothing's leaking for regular `IncrCounter` when it can use the split pass. In `compile_side_exits`, we can't use the split pass or the register allocator, so it's expected that some stuff leaks, especially when you need registers.

So let's talk about the no-split case. `arm64_emit` uses three registers for `IncrCounter`. We have two scratch registers: `SCRATCH0` and `SCRATCH1`, missing one more. If you "let `IncrCounter` figure out" the register, there's no register available because it's past `alloc_regs` and thus all `ALLOC_REGS` could be in use. So it seems natural to me to let the split pass allocate such a register for the general case. For the no-split case, since none of the `ALLOC_REGS` is in use here, we shouldn't need `SCRATCH2`. So I decided to take this design to avoid `SCRATCH2`.

However, it does mean that arm64 suffers from an extra register pressure that x86_64 doesn't have on `IncrCounter`. So I'm fine with introducing `SCRATCH2` to fix it since we do need three scratch registers for this instruction. I'll separately file a PR for this to discuss it.

zjit/src/backend/lir.rs
k0kubun: Thanks, fixed in https://github.com/ruby/ruby/pull/14357/commits/7190aea605156bc6e8e768b215cd82c7e3de6a58

zjit/src/backend/lir.rs
k0kubun: https://github.com/ruby/ruby/pull/14357/commits/7190aea605156bc6e8e768b215cd82c7e3de6a58

zjit/src/codegen.rs
k0kubun: https://github.com/ruby/ruby/pull/14357/commits/23035cf662d07662d977530f7a474581f3c1b130 and then https://github.com/ruby/ruby/pull/14357/commits/5d5115a856cf894af53bf6794bfd5b5741c8b480

zjit/src/codegen.rs
k0kubun: How about this https://github.com/ruby/ruby/pull/14357/commits/5d5115a856cf894af53bf6794bfd5b5741c8b480?

zjit.rb
k0kubun: I made them consistent with YJIT, but I didn't notice the inconsistency. Now that I think about it, "side-exit count" feels more natural to me than "side-exits count" or "side-exit's count". `{singular}_count` seems more common than `{plural}_count` in YJIT stats and other variable names in CRuby too.

So I'd pick `side_exit_count` over `*_insns_count`. I just renamed the other ones https://github.com/ruby/ruby/pull/14357/commits/863d766e6140a88c1e07f8705991d6464683e38e, while leaving the YJIT stats names as is for compatibility.

zjit/src/backend/lir.rs
k0kubun: Filed https://github.com/ruby/ruby/pull/14367

PR #14358, opened 2025-08-27
[3.4 backport] Prism ^D fix

XrXr:
- **When reading from stdin, put a wrapper around the IO object**
- **Bump PRISM version number to avoid collision with RubyGems version**

Because this bugfix touches the Prism gem, there is
a problem of having two versions of the gem sharing
the same version number. With that users would get
different copies depending on the version they
install the gem from, or get a checksum error from
RubyGems or Bundler, which is undesirable.

I don't think we can bump to the latest Prism gem
because that would also bring in syntax changes
that are only in Ruby 3.5. 

I've picked a version number that ruby/prism hasn't
used yet. Maybe as an optional step, we should also
release this version of Prism to RubyGems when
releasing the next 3.4 version. Otherwise, Prism
needs to remember to avoid this version number
in the future to make sure versions are uniquely
identifiable.

CC @kddnewton

### DIFF
--- a/lib/prism/ffi.rb
+++ b/lib/prism/ffi.rb
@@ -80,6 +80,7 @@ module Prism
     end
 
     callback :pm_parse_stream_fgets_t, [:pointer, :int, :pointer], :pointer
+    callback :pm_parse_stream_feof_t, [:pointer], :int
     enum :pm_string_init_result_t, %i[PM_STRING_INIT_SUCCESS PM_STRING_INIT_ERROR_GENERIC PM_STRING_INIT_ERROR_DIRECTORY]
     enum :pm_string_query_t, [:PM_STRING_QUERY_ERROR, -1, :PM_STRING_QUERY_FALSE, :PM_STRING_QUERY_TRUE]
 
@@ -95,7 +96,7 @@ module Prism
       "pm_string_query_local",
       "pm_string_query_constant",
       "pm_string_query_method_name",
-      [:pm_parse_stream_fgets_t]
+      [:pm_parse_stream_fgets_t, :pm_parse_stream_feof_t]
     )
 
     load_exported_functions_from(
@@ -273,13 +274,15 @@ module Prism
           end
         }
 
+        eof_callback = -> (_) { stream.eof?  }
+
         # In the pm_serialize_parse_stream function it accepts a pointer to the
         # IO object as a void* and then passes it through to the callback as the
         # third argument, but it never touches it itself. As such, since we have
         # access to the IO object already through the closure of the lambda, we
         # can pass a null pointer here and not worry.
-        LibRubyParser.pm_serialize_parse_stream(buffer.pointer, nil, callback, dump_options(options))
-        Prism.load(source, buffer.read)
+        LibRubyParser.pm_serialize_parse_stream(buffer.pointer, nil, callback, eof_callback, dump_options(options))
+        Prism.load(source, buffer.read, options.fetch(:freeze, false))
       end
     end
 
--- a/lib/prism/prism.gemspec
+++ b/lib/prism/prism.gemspec
@@ -2,7 +2,7 @@
 
 Gem::Specification.new do |spec|
   spec.name = "prism"
-  spec.version = "1.2.0"
+  spec.version = "1.3.1"
   spec.authors = ["Shopify"]
   spec.email = ["ruby@shopify.com"]
 
--- a/prism/extension.c
+++ b/prism/extension.c
@@ -879,6 +879,14 @@ profile_file(int argc, VALUE *argv, VALUE self) {
     return Qnil;
 }
 
+static int
+parse_stream_eof(void *stream) {
+    if (rb_funcall((VALUE) stream, rb_intern("eof?"), 0)) {
+        return 1;
+    }
+    return 0;
+}
+
 /**
  * An implementation of fgets that is suitable for use with Ruby IO objects.
  */
@@ -919,7 +927,7 @@ parse_stream(int argc, VALUE *argv, VALUE self) {
     pm_parser_t parser;
     pm_buffer_t buffer;
 
-    pm_node_t *node = pm_parse_stream(&parser, &buffer, (void *) stream, parse_stream_fgets, &options);
+    pm_node_t *node = pm_parse_stream(&parser, &buffer, (void *) stream, parse_stream_fgets, parse_stream_eof, &options);
     rb_encoding *encoding = rb_enc_find(parser.encoding->name);
 
     VALUE source = pm_source_new(&parser, encoding);
--- a/prism/prism.c
+++ b/prism/prism.c
@@ -22624,11 +22624,11 @@ pm_parse(pm_parser_t *parser) {
  * otherwise return true.
  */
 static bool
-pm_parse_stream_read(pm_buffer_t *buffer, void *stream, pm_parse_stream_fgets_t *fgets) {
+pm_parse_stream_read(pm_buffer_t *buffer, void *stream, pm_parse_stream_fgets_t *stream_fgets, pm_parse_stream_feof_t *stream_feof) {
 #define LINE_SIZE 4096
     char line[LINE_SIZE];
 
-    while (memset(line, '\n', LINE_SIZE), fgets(line, LINE_SIZE, stream) != NULL) {
+    while (memset(line, '\n', LINE_SIZE), stream_fgets(line, LINE_SIZE, stream) != NULL) {
         size_t length = LINE_SIZE;
         while (length > 0 && line[length - 1] == '\n') length--;
 
@@ -22659,6 +22659,12 @@ pm_parse_stream_read(pm_buffer_t *buffer, void *stream, pm_parse_stream_fgets_t 
             case 9:
                 if (strncmp(line, "__END__\r\n", 9) == 0) return false;
                 break;
+        }
+
+        // All data should be read via gets.  If the string returned by gets
+        // _doesn't_ end with a newline, then we assume we hit EOF condition.
+        if (stream_feof(stream)) {
+            break;
         }
     }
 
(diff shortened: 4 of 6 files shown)


kddnewton:
A couple of things:

* It's actually okay to use the latest version, since we gate all syntax changes on a flag for the version of Ruby that you want to parse.
* If I understand the problem you're describing, I think maybe we should rethink the change a make it so that it can support both versions (with and without the callback). Is the issue basically that libprism that is shipped with Ruby is different?

k0kubun:
> Maybe as an optional step, we should also release this version of Prism to RubyGems when releasing the next 3.4 version.

This is the general expectations for default gems when they cannot be upgraded to the latest version of the gem in Ruby's stable branches.

If we introduced a feature/syntax to Prism that Ruby 3.4 shouldn't have, ruby/prism probably should cut a ruby_3_4 branch and make releases separately from the `main` branch.

I'll merge this now, but can you either release such a version and backport the version bump to this branch _now_, or make me an owner of the gem so that I can do it myself now or when I release Ruby 3.4.6?

tenderlove:
> I'll merge this now, but can you either release such a version and backport the version bump to this branch _now_, or make me an owner of the gem so that I can do it myself now or when I release Ruby 3.4.6?

I don't understand what we're trying to accomplish here.

Are we backporting the Prism fix just because it's a bugfix?  As such, we need to ship a new version of Prism?  Bumping the prism version in `ruby/ruby` without doing it upstream first doesn't seem like what we should be doing (unless I'm misunderstanding).  I guess I don't understand why we needed to bump the Prism version.

XrXr:
> It's actually okay to use the latest version, since we gate all syntax changes on a flag for the version of Ruby that you want to parse.

Cool, that's a valid option, to bump to the latest version. We'd still need to coordinate a release to keep the version number unique and meaningful though. Looks like no current release contains the API change in this diff yet. Though a bump like that might be risky for the branch maintainer, it's useful to know that it's an option.

> If I understand the problem you're describing, I think maybe we should rethink the change a make it so that it can support both versions (with and without the callback). Is the issue basically that libprism that is shipped with Ruby is different?

The problem is that making any changes at all will make the default gem version 1.2.0 of Prism different from the 1.2.0 one can get on rubygems.org. Making changes to support both versions still runs into this problem.

PS. we may want to pick a better version name than I did. I picked 1.3.1 because this is a breaking change, but this version is more closely related to 1.2.0 than 1.3.0... Maybe we need a separate versioning scheme for these backport Prism versions.

k0kubun:
> > It's actually okay to use the latest version, since we gate all syntax changes on a flag for the version of Ruby that you want to parse.
>
> it's useful to know that it's an option.

Sorry I guess I overlooked this discussion. 

> Though a bump like that might be risky for the branch maintainer

As a branch maintainer, I'm generally open to bumping gem versions to the latest, which may include new features, as long as they are backward-compatible. We backport rubygems and bundler which tend to have way riskier changes :p

So which one do you all want in the `ruby_3_4` branch, the latest version or a ruby_3_4-specific release? With the flag maintaining the integrity of the syntax for the Ruby version, I'm fine with both ways. If a version with the fix is already published and you don't see any issue, let's just synchronize the whole Prism code in the branch to it.

XrXr:
I vote for cutting a new Prism gem release and pointing to it. It's been a while since the last Prism release anyways: https://github.com/ruby/prism/issues/3553 and also, doing that will also do the backport for https://bugs.ruby-lang.org/issues/21337

Will take a little bit of integration work to set the syntax version correctly so we don't accidentally pull in https://bugs.ruby-lang.org/issues/20925 and https://github.com/ruby/prism/pull/3337 which Matz just approved but this seems like the best long term play.

k0kubun:
@kddnewton WDYT about making @tenderlove and @eileencodes gem owners of prism.gem? I, as a release manager of Ruby 3.4, think we need at least one person who has full-time attention to Ruby and can cut a release of Prism as needed.

kddnewton:
v1.5.0 is released. Please let me know if there's something I should do with this specific ticket or any backport.

k0kubun:
Thanks! I'll work on sync-ing Prism to that version on `ruby_3_4` and let you know if I need help.

Have you also invited them to be gem owners?

k0kubun:
It doesn't compile 😞 

```
/Users/runner/work/ruby/ruby/build/rbconfig.rb:315:in 'String#replace': can't modify frozen String: "$(SDKROOT)$(prefix)/include" (FrozenError)
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:315:in 'RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:308:in 'block in RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'String#gsub'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:308:in 'block in RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'String#gsub'
compiling ../src/enc/cesu_8.c
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:308:in 'block in RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'String#gsub'
compiling ../src/enc/trans/cesu_8.c
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:301:in 'RbConfig.expand'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:319:in 'block in <module:RbConfig>'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:318:in 'Hash#each_value'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:318:in '<module:RbConfig>'
	from /Users/runner/work/ruby/ruby/build/rbconfig.rb:11:in '<top (required)>'
	from ../src/ext/extmk.rb:42:in 'Kernel#require'
	from ../src/ext/extmk.rb:42:in '<main>'
```

https://github.com/ruby/ruby/actions/runs/17684937914/job/50267562914?pr=14532#step:9:176
https://github.com/ruby/ruby/pull/14532

k0kubun:
It builds correctly on my local Linux environment, but every macOS-based CI seems to be failing with it.

k0kubun:
It works fine on my local macOS environment too 🤔

k0kubun:
It seems like we also can't use Prism 1.5.0 for Ruby 3.4 anyway? Do we need to create a Prism branch and version for `ruby_3_4`?

<pre>
    1) Failure:
  TestSyntax#test_methoddef_endless_command [/home/runner/work/ruby/ruby/src/test/ruby/test_syntax.rb:1800]:
  /home/runner/work/ruby/ruby/src/test/ruby/test_syntax.rb:1800:in 'TestSyntax#test_methoddef_endless_command'.
  ```
  private def foo = puts "Hello"
  ```
  no-newline.
  SyntaxError expected but nothing was raised.
  
    2) Failure:
  Prism::TestCompilePrism#test_concatenated_StringNode [/home/runner/work/ruby/ruby/src/test/ruby/test_compile_prism.rb:624]:
  @/home/runner/work/ruby/ruby/src/test/ruby/test_compile_prism.rb:624.
  <false> expected but was
  <true>.
  
  Finished tests in 221.929572s, 147.4747 tests/s, 29655.7955 assertions/s.
  32729 tests, 6581498 assertions, 2 failures, 0 errors, 364 skips
</pre>

https://github.com/ruby/ruby/actions/runs/17684937953/job/50267562966?pr=14532

Earlopain:
It contains 3.5 specific syntax now and currently ruby always uses the latest version that prism knows about. You have to explicitly request 3.4 to get the correct behaviour like https://github.com/ruby/ruby/pull/14523

k0kubun:
I see. Thanks for working on that. I hope backporting it will fix the rest of the problems too.

kddnewton:
@k0kubun would you be able to make it so that the ruby 3.4 branch requests ruby 3.4 syntax?

Earlopain:
There is this commit: https://github.com/ruby/ruby/commit/7dbd9c26361719a45fa39838f46a76d67dc3c2e9

I lack the context but it looks like it was never reverse synced (or mentioned in https://bugs.ruby-lang.org/issues/21187 for that matter)

k0kubun:
Hm. So I guess this needs to be fixed (or just reverted) on Prism's end first?

k0kubun:
So on July 21st at ruby-lang Slack (#general-ja), hsbt reported (probably non-GitHub Actions) CI failures because of the reverted commit and shared that he was going to revert it. Aaron acknowledged it and said he would fix Prism and Ruby together, but I guess we've never got to fix it.

For now, I'd like to suggest reverse sync-ing the revert to Prism, make another Prism release, and re-open the Redmine ticket if necessary. I'll just forward-fix the problem by merging the "almost Prism 1.5.0" (with the revert) to ruby_3_4, but it'd be nice to have a release with the revert soon. Otherwise I'll (revert backports and) postpone Prism backports to Ruby 3.4.7.

Earlopain:
The problem that this change intended to fix still seems to be fixed even with the revert. There is more context in https://github.com/ruby/ruby/pull/13966 for those that understand but I tested the repro from https://bugs.ruby-lang.org/issues/21187 and it is as expected on current master. 3.4.5 still fails since it never got backported.

So reverting seems ok? Although I cannot say why @tenderlove made that change to begin with. I assume there were reasons.

k0kubun:
Huh, ok. Let's start from reverting it then. Unfortunately Aaron seems to be out next week (as well as this week), so I guess I'll need Kevin to make another release. I'll prepare the reverse sync PR.

k0kubun:
So `ruby_3_4` currently has Prism 1.5.0 with the revert https://github.com/ruby/ruby/pull/14532, which does fix the build issue on `ruby_3_4` as well. I need a Prism release with the revert to make it consistent with the version it points to on rubygems.org.

@kddnewton Could you merge https://github.com/ruby/prism/pull/3643 and make another release of Prism? @tenderlove is out until Sep 21st, so I don't think he has time to attempt to fix his patch before the Ruby 3.4.6 release (next Monday).

kddnewton:
v1.5.1 is out with the revert.

PR #14362, opened 2025-08-27
Fix bad NameError raised using sendforward instruction through vcall

luke-gruber:
If you called a VCALL method and the method takes forwarding arguments and then you forward those arguments along using the sendforward instruction, the method_missing class was wrongly chosen as NameError instead of NoMethodError. This is because the VM looked at the CallInfo of the vcall and determined it needed to raise NameError. Now we detect that case and raise NoMethodError.

Fixes [Bug #21535]


luke-gruber:
cc @tenderlove

PR #14365, opened 2025-08-27
vm_setivar: Don't populate cache on size transition

byroot:
Any setivar that enlarge the object won't ever hit, so it's wasteful to populate the cache as it will only result in more work on the next match.


byroot:
Nevermind, this case can still happen, so this optimization isn't valid.

PR #14366, opened 2025-08-27
Populate ivar caches for types other than T_OBJECT

byroot:
`vm_setinstancevariable` had a codepath to try to match the inline cache for types other than T_OBJECT, but the cache population path in `vm_setivar_slowpath` was exclusive to T_OBJECT, so `vm_setivar_default` would never match anything.

This commit improves `vm_setivar_slowpath` so that it is capable of filling the cache for all types, and adds a `vm_setivar_class` codepath for `T_CLASS` and `T_MODULE`.

`vm_setivar`, `vm_setivar_default` and `vm_setivar_class` could be unified, but based on the very explicit `NOINLINE` I assume they were split to minimize codesize.

```
compare-ruby: ruby 3.5.0dev (2025-08-27T14:58:58Z merge-vm-setivar-d.. 5b749d8e53) +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-27T16:30:31Z setivar-cache-gene.. 4fe78ff296) +PRISM [arm64-darwin24]
# Iteration per second (i/s)

|vm_ivar_set_on_instance  |     161.809|   164.688|
|vm_ivar_set_on_generic   |      58.769|   115.638|
|                         |           -|     1.97x|
|vm_ivar_set_on_class     |      70.034|   141.042|
|                         |           -|     2.01x|
```

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1300,7 +1300,7 @@ rb_free_generic_ivar(VALUE obj)
     }
 }
 
-void
+static void
 rb_obj_set_fields(VALUE obj, VALUE fields_obj, ID field_name, VALUE original_fields_obj)
 {
     ivar_ractor_check(obj, field_name);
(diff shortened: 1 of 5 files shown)


variable.c
```diff
@@ -1554,6 +1554,7 @@ rb_ivar_delete(VALUE obj, ID id, VALUE undef)
             }
             else {
                 val = rb_ivar_delete(fields_obj, id, undef);
+                RBASIC_SET_SHAPE_ID(obj, RBASIC_SHAPE_ID(fields_obj));
```
byroot: This was a bug, in some cases the IMEMO/fields_obj shape_id wasn't forwarded back to the T_CLASS.

But here too it's unclear what that means for namespaces.

PR #14367, opened 2025-08-27
ZJIT: Introduce SCRATCH2 for IncrCounter on arm64

k0kubun:
This PR addresses Max's suggestion at https://github.com/ruby/ruby/pull/14357#discussion_r2302699921.

Adding `SCRATCH2` allows `arm64_emit` to emit `IncrCounter` without relying on `arm64_split`, which lets `compile_side_exits` remove the `cfg!(target_arch = "aarch64")` branch. I don't think taking `Opnd::const_ptr` in `IncrCounter` as Max suggested isn't worth it, which will duplicate more code in `arm64_emit`, so it still takes `Opnd::Mem`.

It generally doesn't seem like a good idea to waste a register when you can save it with no run-time overhead, especially when it's for `--zjit-stats`. But at the same time, it seems useful to remove an extra register pressure that the other architecture doesn't have for the same LIR instruction.

`x18` is a platform register. Since we control the use of every register in JIT code, it seems safe to use it as an extra caller-saved register. We don't seem to have any more caller-saved registers.

### DIFF
--- a/zjit/src/asm/arm64/opnd.rs
+++ b/zjit/src/asm/arm64/opnd.rs
@@ -112,6 +112,7 @@ pub const X14_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 14 };
 pub const X15_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 15 };
 pub const X16_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 16 };
 pub const X17_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 17 };
+pub const X18_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 18 };
 
 // callee-save registers
 pub const X19_REG: A64Reg = A64Reg { num_bits: 64, reg_no: 19 };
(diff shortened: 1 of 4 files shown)


XrXr (commented):
> x18 is a platform register. Since we control the use of every register in JIT code, it seems safe to use it as an extra caller-saved register. We don't seem to have any more caller-saved registers.

It's not a great idea to run afoul macOS's platform rules for this https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers

It's hard to predict what can go wrong. Maybe they use it in the kernel for something in which case we actually don't have full control over that register because we could be preempted in the region of code that uses the register.

k0kubun:
> https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms#Respect-the-purpose-of-specific-CPU-registers
> The platforms reserve register x18. Don’t use this register.

Heh, interesting. Thanks for pointing that out.

If we don't have an additional caller-saved register, I'd rather leave the `cfg!(target_arch = "aarch64")` branch in `compile_side_exits` as is. `Assembler` is not designed to hide abstraction leaks after `split`/`alloc_regs` in the first place. We should treat `compile_side_exits` like part of the `emit` pass, which happens to have slightly richer API.

PR #14368, opened 2025-08-27
ZJIT: Support updating insta snapshots

st0012:


doc/zjit.md
```diff
@@ -50,12 +52,24 @@ You can also run a single test case by specifying the function name:
...
+```
+make zjit-test-update
+```
+
+This requires `cargo-insta` to be installed. If not already installed:
 
 ```
-UPDATE_EXPECT=1 make zjit-test ZJIT_TESTS=test_putobject
+cargo install cargo-insta
 ```
```
tekknolagi: This is suggested above. Do we need to mention again?

zjit/zjit.mk
```diff
@@ -80,7 +80,18 @@ zjit-test: libminiruby.a
...
+		$(ZJIT_TESTS) || true
+	@echo ""
+	@if [ -f '$(top_srcdir)/zjit/src/.hir.rs.pending-snap' ]; then \
+		echo "Pending snapshots found. Accept with: make zjit-test-update"; \
+	fi
+
+# Accept all pending snapshots (requires cargo-insta)
+# Install with: cargo install cargo-insta
+zjit-test-update:
+	@PATH="$(HOME)/.cargo/bin:$$PATH"; \
```
tekknolagi: I think this does not belong in the Makefile; people should have it in their rc files

tekknolagi (approved):
Nice, thank you

zjit/zjit.mk
```diff
@@ -80,7 +80,17 @@ zjit-test: libminiruby.a
 	$(ZJIT_NEXTEST_ENV) $(CARGO) nextest run \
 		--manifest-path '$(top_srcdir)/zjit/Cargo.toml' \
 		'--features=$(ZJIT_TEST_FEATURES)' \
-		$(ZJIT_TESTS)
+		$(ZJIT_TESTS) || true
```
k0kubun: Doesn't this prevent `make zjit-test` from failing?

tekknolagi: Oh I missed this (or it wasn't there originally). I recommend `INSTA_FORCE_PASS=1` and `--no-fail-fast` in the `update` command

st0012: @tekknolagi The update command only applies the `pending-snap` file and doesn't run tests. Those options need to be added to `zjit-test` command if we want them (which I do)

tekknolagi: I don't want those as part of `zjit-test`. `zjit-test` should *definitely* fail if need be

st0012: Do you mean fail or fail fast?
I've fixed the command so `zjit-test` should run the correct exit code when tests fail. I do think we can have `--no-fail-fast` tho as to me it's always nicer to see all failing tests at once. It'd make updating snapshots easier too.

tekknolagi: I am fine with `--no-fail-fast` but not `|| true`

doc/zjit.md
```diff
@@ -50,12 +52,18 @@ You can also run a single test case by specifying the function name:
...
 
-If you expect that your changes cause tests to fail and they do, you can have
-`expect-test` fix the expected value for you by putting `UPDATE_EXPECT=1`
-before your test command, like so:
+#### Snapshot Testing
+
+ZJIT uses [insta](https://insta.rs/) for snapshot testing. When tests fail due to snapshot mismatches, pending snapshots are created. The test command will notify you if there are pending snapshots:
+
+```
+Pending snapshots found. Accept with: make zjit-test-update
```
tekknolagi, line 60: I might also suggest reviewing one by one with `cargo insta review`

zjit/zjit.mk
st0012: `|| true` is gone 🙂

doc/zjit.md
st0012: Added

PR #14375, opened 2025-08-27
ZJIT: Add Ractor mode PatchPoint for ivar get/set

st0012:
This will tick off another failing test in `bootstraptest/test_yjit.rb`

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1671,7 +1671,14 @@ impl Function {
                                 self_val = self.push_insn(block, Insn::GuardType { val: self_val, guard_type: Type::from_profiled_type(profiled_type), state });
                             }
                             let id = unsafe { get_cme_def_body_attr_id(cme) };
-                            self.push_insn(block, Insn::PatchPoint { invariant: Invariant::SingleRactorMode, state });
+
+                            // Check if we're accessing ivars of a Class or Module object as they require single-ractor mode
+                            if unsafe { rb_zjit_singleton_class_p(klass) } {
+                                let attached = unsafe { rb_class_attached_object(klass) };
+                                if unsafe { RB_TYPE_P(attached, RUBY_T_CLASS) || RB_TYPE_P(attached, RUBY_T_MODULE) } {
+                                    self.push_insn(block, Insn::PatchPoint { invariant: Invariant::SingleRactorMode, state });
+                                }
+                            }
                             let getivar = self.push_insn(block, Insn::GetIvar { self_val, id, state });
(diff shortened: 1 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -3325,13 +3333,17 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let id = ID(get_arg(pc, 0).as_u64());
                     // ic is in arg 1
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    // TODO: We only really need this if self_val is a class/module
```
k0kubun, line 3338: From just looking at the HIR code, it wasn't obvious to me why we need the patch point.

```suggestion
                    // Assume single-Ractor mode to omit gen_prepare_non_leaf_call on gen_getivar
                    // TODO: We only really need this if self_val is a class/module
```

zjit/src/hir.rs
```diff
@@ -3325,13 +3333,17 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                    fun.push_insn(block, Insn::PatchPoint { invariant: Invariant::SingleRactorMode, state: exit_id });
                     let result = fun.push_insn(block, Insn::GetIvar { self_val: self_param, id, state: exit_id });
                     state.stack_push(result);
                 }
                 YARVINSN_setinstancevariable => {
                     let id = ID(get_arg(pc, 0).as_u64());
                     // ic is in arg 1
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    // TODO: We only really need this if self_val is a class/module
```
k0kubun, line 3348: ditto

```suggestion
                    // Assume single-Ractor mode to omit gen_prepare_non_leaf_call on gen_setivar
                    // TODO: We only really need this if self_val is a class/module
```

zjit/src/hir.rs
```diff
@@ -1671,6 +1671,14 @@ impl Function {
                                 self_val = self.push_insn(block, Insn::GuardType { val: self_val, guard_type: Type::from_profiled_type(profiled_type), state });
                             }
                             let id = unsafe { get_cme_def_body_attr_id(cme) };
+
+                            // Check if we're accessing ivars of a Class or Module object as they require single-ractor mode
```
```suggestion
                            // Check if we're accessing ivars of a Class or Module object as they require single-ractor mode.
                            // We omit gen_prepare_non_leaf_call on gen_getivar, so it's unsafe to raise for multi-ractor mode.
```

PR #14376, opened 2025-08-27
ZJIT: Compile Insn::Send

k0kubun:
This PR adds codegen for `Insn::Send` by dynamic dispatch. HIR is also changed to reload every local after `Insn::Send`.

It increases `ratio_in_zjit` on lobsters from 56.7% to 61.8%.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -1665,6 +1665,23 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 2
   end
 
+  def test_method_redefinition_with_module
+    assert_runs '["original", "redefined"]', %q{
+      module Foo
+        def self.foo = "original"
+      end
+
+      def test = Foo.foo
+      test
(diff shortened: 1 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -4698,6 +4709,7 @@ mod tests {
         bb0(v0:BasicObject, v1:BasicObject):
           v3:BasicObject = GetLocal l0, EP@3
           v5:BasicObject = Send v3, 0x1000, :each
+          v6:BasicObject = GetLocal l0, EP@3
```
st0012, line 4712: In this case, if there's anything/what would access `v6`? Or is the `GetLocal` just for `state.setlocal`?

st0012 (approved):

zjit/src/hir.rs
k0kubun: CheckInterrupts. The re-loaded local is spilled on the side exit when there's an interrupt.

PR #14377, opened 2025-08-28
Backport vcpkg related improvement to `ruby_3_4`

hsbt:
We gave up to tweak cache management by vcpkg.

These commits make to skip vcpkg build when `actions/cache` detect the past artifacts by vcpkg and only save the new artifacts to `actions/cache`.

And I suggest to backport to detect TEMP dir by `RUNNER_TEMP` instead by hard coded path.


k0kubun (approved):

hsbt:
Thanks!

PR #14378, opened 2025-08-28
ZJIT: Refactor stats implementations

k0kubun:
This PR tweaks some stats related code.

### DIFF
--- a/zjit.rb
+++ b/zjit.rb
@@ -25,16 +25,8 @@ class << RubyVM::ZJIT
   end
 
   # Return ZJIT statistics as a Hash
-  def stats(key = nil)
-    stats = Primitive.rb_zjit_stats(key)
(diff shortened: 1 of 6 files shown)


zjit.rb
```diff
@@ -44,6 +36,8 @@ def stats_string
 
     print_counters_with_prefix(prefix: 'failed_', prompt: 'compilation failure reasons', buf:, stats:)
     print_counters([
+      :num_send_dynamic,
```
tekknolagi: if we're going to rename it, let's do a suffix _count instead of prefix num_

k0kubun: ok https://github.com/ruby/ruby/pull/14378/commits/47635e8ee35c6b11f6a0bdbdca536fdaf1e32ea1

PR #14380, opened 2025-08-28
rb_ivar_delete: don't reembed objects

byroot:
With the recent refactoring of shapes, it is now guaranteed that embedded and external objects shape_id are different.

Hence we no longer need that extra complexity, and this can be turned into a unit test instead.


byroot:
Nevermind, I overlooked something.

The issue remain in some cases.

PR #14383, opened 2025-08-28
YJIT simplify gen_get_ivar and gen_set_ivar

byroot:
The `shape_id` now includes 3 bits for the `heap_id`. It is always non-zero for `T_OBJECT` and always zero for all other types.

Hence all these allocator checks are no longer necessary.

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2848,22 +2848,10 @@ fn gen_get_ivar(
     recv: Opnd,
     recv_opnd: YARVOpnd,
 ) -> Option<CodegenStatus> {
-    let comptime_val_klass = comptime_receiver.class_of();
-
     // If recv isn't already a register, load it.
     let recv = match recv {
         Opnd::InsnOut { .. } => recv,
         _ => asm.load(recv),
-    };
-
-    // Check if the comptime class uses a custom allocator
-    let custom_allocator = unsafe { rb_get_alloc_func(comptime_val_klass) };
-    let uses_custom_allocator = match custom_allocator {
-        Some(alloc_fun) => {
-            let allocate_instance = rb_class_allocate_instance as *const u8;
-            alloc_fun as *const u8 != allocate_instance
-        }
-        None => false,
     };
 
     // Check if the comptime receiver is a T_OBJECT
@@ -2874,12 +2862,9 @@ fn gen_get_ivar(
         gen_counter_incr(jit, asm, Counter::num_getivar_megamorphic);
     }
 
-    // If the class uses the default allocator, instances should all be T_OBJECT
-    // NOTE: This assumes nobody changes the allocator of the class after allocation.
-    //       Eventually, we can encode whether an object is T_OBJECT or not
-    //       inside object shapes.
+    // NOTE: This assumes T_OBJECT can't ever have the same shape_id as any other type.
     // too-complex shapes can't use index access, so we use rb_ivar_get for them too.
-    if !receiver_t_object || uses_custom_allocator || comptime_receiver.shape_too_complex() || megamorphic {
+    if !receiver_t_object || comptime_receiver.shape_too_complex() || megamorphic {
         // General case. Call rb_ivar_get().
         // VALUE rb_ivar_get(VALUE obj, ID id)
         asm_comment!(asm, "call rb_ivar_get()");
@@ -3073,8 +3058,6 @@ fn gen_set_ivar(
     recv_opnd: YARVOpnd,
     ic: Option<*const iseq_inline_iv_cache_entry>,
 ) -> Option<CodegenStatus> {
-    let comptime_val_klass = comptime_receiver.class_of();
-
     // If the comptime receiver is frozen, writing an IV will raise an exception
     // and we don't want to JIT code to deal with that situation.
     if comptime_receiver.is_frozen() {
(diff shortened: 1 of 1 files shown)


tekknolagi:
This is awesome. Do you have plans to also encode the embed status in the shape ID? Leaving review to those more familiar with YJIT.

byroot:
> Do you have plans to also encode the embed status in the shape ID?

It's already somewhat in it, because `rb_ivar_delete` ensure any embeddable object is always embeded: https://github.com/ruby/ruby/blob/4fc0db7389d2c93eebf232e802769a38702eea8e/variable.c#L1617-L1623

So you can't have two `T_OBJECT` with the same ivars and same `shape_id` but one embedded and the other not, as it would mean they're in different heaps, hence different `heap_id`, hence different `shape_id`.

We may have to more explicitly encode the embed status later on if we start compiling code for types other than T_OBJECT though (I'm pairing on that with Aaron tonight).

byroot:
> We may have to more explicitly encode the embed status later on if we start compiling code for types other than T_OBJECT though

Actually https://github.com/ruby/ruby/pull/14384 should be enough.

tekknolagi:
If I can tell from a bit on the shape ID whether the object is embedded or not, that makes YJIT's and ZJIT's lives easier but it's also fine to look it up on the class somehow

byroot:
> but it's also fine to look it up on the class somehow

I assume you mean on the object?

In the end it's just:

```rust
    pub fn embedded_p(self) -> bool {
        unsafe {
            FL_TEST_RAW(self, VALUE(ROBJECT_HEAP as usize)) == VALUE(0)
        }
    }
```

So I don't think having it in the shape_id would simply much.

On paper it wouldn't be very hard to add it, but it's potentially a bit finicky, so not 100% sure it's worth the effort.

tekknolagi:
No, I mean the class. Right now ZJIT's profiling only has access to a historical statistical distribution of (class, shapeID) pairs. We want to use that (and maybe additional information if need be) to specialize ivar access. We could get the embed-bit from each profiled object and store it as a flag on our profiling info alongside (class, shape) if it's not somehow redundantly computable.

byroot:
> Right now ZJIT's profiling only has access to a historical statistical distribution of (class, shapeID)

I see.

I might look into getting the embedded bit in there then.

tekknolagi:
I mean, to be clear, we absolutely can sneak a separate embed bit into our profiling info right now, but if it's trivially known from shapeID in the long term, great! If it's not an obvious win to you, I don't want to otherwise push for it

tekknolagi:
For example: https://github.com/ruby/ruby/pull/14386

k0kubun:
I don't know what custom allocators are, so I didn't follow this logic:

> The shape_id now includes 3 bits for the heap_id. It is always non-zero for T_OBJECT and always zero for all other types.
> 
> Hence all these allocator checks are no longer necessary.

So this removes a compile-time check. I wonder which run-time code was guarding against a custom allocator.

My current guess is:

1. `T_OBJECT` never has a custom allocator. Other `T_*` types can.
2. Whether it's `T_OBJECT` or not, the kind of a custom allocator (if any) is encoded in shape_id.
3. While we checked `T_OBJECT` at compile time, the previous code was shared among non-`T_OBJECT` types as well.

Is my understanding correct?

k0kubun:
Well, (1) + (2) doesn't explain why we had to check `uses_custom_allocator` besides `receiver_t_object`. My theory is gone.

byroot:
> So this removes a compile-time check. I wonder which run-time code was guarding against a custom allocator.

Well, I suppose it was never needed. Because previously `T_OBJECT` and genivar had different shape roots. So I guess they could already never match even before my refactor.

XrXr (commented):
> Well, (1) + (2) doesn't explain why we had to check uses_custom_allocator besides receiver_t_object. My theory is gone.

The custom allocator was probably a left-over from https://bugs.ruby-lang.org/issues/18007
in that you could have the same class with some instances T_DATA and some T_OBJECT. The compile time sample is T_OBJECT, but you know there is a good chance of seeing T_DATA in the future, so it uses the general function call.

> Well, I suppose it was never needed. Because previously T_OBJECT and genivar had different shape roots. So I guess they could already never match even before my refactor.

This check existed pre-shapes in C YJIT, and was necessary for correctness because of the mixed T_OBJECT, T_DATA type.

https://github.com/ruby/ruby/blob/0a3704f218f0aec7f92f3a46a2293175b0a7d2b3/yjit_codegen.c#L1651-L1696

Anyways it sounds fine to remove now.

XrXr:
BTW, I believe there is no need for an embed bit. For T_OBJECTS, anyways. You can tell the slot size using the `heap_id`, and `embed_p = slot_size(heap_id) >= capacity(shape_id)`.

yjit/src/codegen.rs
```diff
@@ -2875,11 +2863,9 @@ fn gen_get_ivar(
     }
 
     // If the class uses the default allocator, instances should all be T_OBJECT
```
XrXr:
```suggestion
```
No more allocator check

byroot:
> You can tell the slot size using the `heap_id`, and `embed_p = slot_size(heap_id) >= capacity(shape_id)`.

Indeed. And I already use that in shape code. The embed bit is only needed if you want to avoid both these statements de-referencing a pointer each, which I suppose doesn't matter in the compiler.

byroot:
> This check existed pre-shapes in C YJIT

That makes sense.

PR #14385, opened 2025-08-28
ZJIT: Disable profiling in compile_iseq

tekknolagi:
This catches both the interpreter-JIT and JIT-JIT cases.

Fixes https://github.com/Shopify/ruby/issues/719


zjit/src/codegen.rs
```diff
@@ -1378,6 +1378,9 @@ fn compile_iseq(iseq: IseqPtr) -> Option<Function> {
         return None;
     }
 
+    // Convert ZJIT instructions back to bare instructions
+    unsafe { crate::cruby::rb_zjit_profile_disable(iseq) };
```
k0kubun: I don't think we have a motivation to put this later than the stack_max bail-out. We shouldn't pay the profiling overhead if we're never going to compile the ISEQ. Let's put this at the top of `compile_iseq`.

PR #14388, opened 2025-08-28
ZJIT: Specialize monomorphic GetIvar

tekknolagi:
Specialize monomorphic `GetIvar` into:

* `GuardType(HeapObject)`
* `GuardShape`
* `LoadIvarEmbedded` or `LoadIvarExtended`

This requires profiling self for `getinstancevariable` (it's not on the operand
stack).

This also optimizes `GetIvar`s that happen as a result of inlining
`attr_reader` and `attr_accessor`.

Also move some (newly) shared JIT helpers into jit.c.


zjit.c
```diff
@@ -363,6 +360,12 @@ rb_zjit_defined_ivar(VALUE obj, ID id, VALUE pushval)
     return result ? pushval : Qnil;
 }
 
+bool
+rb_zjit_shape_too_complex_p(shape_id_t shape_id)
```
k0kubun: Let's put this in jit.c as `rb_jit_shape_too_complex_p` and share it among YJIT and ZJIT. They're doing the same thing and we can avoid bloating the binary size for it.

tekknolagi:
This PR causes a bunch of (new) GuardShape exits on optcarrot in particular. Many millions. However, even with that, we go from 6000ms per iteration to 4000 ms per iteration. I would like to figure out why there are so many exits.

XrXr:
> I would like to figure out why there are so many exits.

Maybe we need something like [`--yjit-exit-locations`](https://docs.ruby-lang.org/en/master/RubyVM/YJIT.html#method-c-dump_exit_locations) 😁

k0kubun (approved):

k0kubun:
Note for self:

> https://github.com/ruby/ruby/pull/14383
> The shape_id now includes 3 bits for the heap_id. It is always non-zero for T_OBJECT and always zero for all other types.
rb_ivar_delete ensure any embeddable object is always embeded
>
> So you can't have two T_OBJECT with the same ivars and same shape_id but one embedded and the other not, as it would mean they're in different heaps, hence different heap_id, hence different shape_id.

> https://github.com/ruby/ruby/pull/14384
> we must ensure two IMEMO/fields can't have the same shape_id but diffent embed/heap status.

PR #14390, opened 2025-08-28
YJIT: getinstancevariable cache indexes for types other than T_OBJECT

byroot:
Based on https://github.com/ruby/ruby/pull/14383

While accessing the ivars of other types is too complicated to realistically generate the ASM for it, we can at least provide the ivar index as to not have to lookup the shape tree every time.

```
compare-ruby: ruby 3.5.0dev (2025-08-27T14:58:58Z merge-vm-setivar-d.. 5b749d8e53) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-28T21:23:38Z yjit-get-exivar 3cc21b76d4) +YJIT +PRISM [arm64-darwin24]

|                           |compare-ruby|built-ruby|
|:--------------------------|-----------:|---------:|
|vm_ivar_get_on_obj         |     975.981|   975.772|
|                           |       1.00x|         -|
|vm_ivar_get_on_class       |     136.214|   470.912|
|                           |           -|     3.46x|
|vm_ivar_get_on_generic     |     148.315|   299.122|
|                           |           -|     2.02x|
```

~~NB: I was expecting a larger improvement, but I suspect preparing for a non-leaf call adds a massive overhead? Still I think it's simple enough to be worth implementing.~~

### DIFF
--- a/variable.c
+++ b/variable.c
@@ -1494,6 +1494,26 @@ rb_ivar_get_at(VALUE obj, attr_index_t index, ID id)
 }
 
 VALUE
+rb_ivar_get_at_no_ractor_check(VALUE obj, attr_index_t index)
+{
+    // Used by JITs, but never for T_OBJECT.
+
+    VALUE fields_obj;
+    switch (BUILTIN_TYPE(obj)) {
+      case T_OBJECT:
+        UNREACHABLE_RETURN(Qundef);
+      case T_CLASS:
+      case T_MODULE:
+        fields_obj = RCLASS_WRITABLE_FIELDS_OBJ(obj);
+        break;
+      default:
+        fields_obj = rb_obj_fields_no_ractor_check(obj);
+        break;
+    }
+    return rb_imemo_fields_ptr(fields_obj)[index];
+}
+
+VALUE
 rb_attr_get(VALUE obj, ID id)
 {
     return rb_ivar_lookup(obj, id, Qnil);
--- a/variable.c
+++ b/variable.c
@@ -1464,6 +1464,36 @@ rb_ivar_get(VALUE obj, ID id)
 }
 
 VALUE
+rb_ivar_get_at(VALUE obj, attr_index_t index, ID id)
+{
+    RUBY_ASSERT(rb_is_instance_id(id));
+    // Used by JITs, but never for T_OBJECT.
+
+    switch (BUILTIN_TYPE(obj)) {
+      case T_OBJECT:
+        UNREACHABLE_RETURN(Qundef);
+      case T_CLASS:
+      case T_MODULE:
+        {
+            VALUE fields_obj = RCLASS_WRITABLE_FIELDS_OBJ(obj);
+            VALUE val = rb_imemo_fields_ptr(fields_obj)[index];
+
+            if (UNLIKELY(!rb_ractor_main_p()) && !rb_ractor_shareable_p(val)) {
+                rb_raise(rb_eRactorIsolationError,
+                        "can not get unshareable values from instance variables of classes/modules from non-main Ractors");
+            }
+
+            return val;
+        }
+      default:
+        {
+            VALUE fields_obj = rb_obj_fields(obj, id);
+            return rb_imemo_fields_ptr(fields_obj)[index];
+        }
+    }
+}
+
+VALUE
 rb_attr_get(VALUE obj, ID id)
 {
     return rb_ivar_lookup(obj, id, Qnil);
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -2864,7 +2864,7 @@ fn gen_get_ivar(
 
     // NOTE: This assumes T_OBJECT can't ever have the same shape_id as any other type.
     // too-complex shapes can't use index access, so we use rb_ivar_get for them too.
-    if !receiver_t_object || comptime_receiver.shape_too_complex() || megamorphic {
+    if !comptime_receiver.heap_object_p() || comptime_receiver.shape_too_complex() || megamorphic {
         // General case. Call rb_ivar_get().
         // VALUE rb_ivar_get(VALUE obj, ID id)
         asm_comment!(asm, "call rb_ivar_get()");
@@ -2900,9 +2900,6 @@ fn gen_get_ivar(
     // Guard heap object (recv_opnd must be used before stack_pop)
     guard_object_is_heap(asm, recv, recv_opnd, Counter::getivar_not_heap);
 
-    // Compile time self is embedded and the ivar index lands within the object
-    let embed_test_result = comptime_receiver.embedded_p();
-
     let expected_shape = unsafe { rb_obj_shape_id(comptime_receiver) };
     let shape_id_offset = unsafe { rb_shape_id_offset() };
     let shape_opnd = Opnd::mem(SHAPE_ID_NUM_BITS as u8, recv, shape_id_offset);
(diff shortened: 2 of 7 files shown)


byroot:
> I suspect preparing for a non-leaf call adds a massive overhead?

If that's the case, I wonder if it would be worth returning `Qundef` instead of raising, then we can more cheaply test for `== Qundef` and then do a non-leaf call?

byroot:
So I was benchmarking in debug mode 🤦, it's a bit better in release mode:

```
compare-ruby: ruby 3.5.0dev (2025-08-27T14:58:58Z merge-vm-setivar-d.. 5b749d8e53) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-28T17:58:32Z yjit-get-exivar efaa8c9b09) +YJIT +PRISM [arm64-darwin24]
# Iteration per second (i/s)

|                           |compare-ruby|built-ruby|
|:--------------------------|-----------:|---------:|
|vm_ivar_get_uninitialized  |    136.136M|  137.560M|
|                           |           -|     1.01x|
|vm_ivar_get_on_obj         |     978.502|   963.354|
|                           |       1.02x|         -|
|vm_ivar_get_on_class       |     134.569|   428.425|
|                           |           -|     3.18x|
|vm_ivar_get_on_generic     |     147.503|   273.851|
|                           |           -|     1.86x|
```

If I remove the `jit_prepare_non_leaf_call` (which is incorrect but help decide if it's worth pursuing), it doesn't seem to make that big of a difference:

```
compare-ruby: ruby 3.5.0dev (2025-08-27T14:58:58Z merge-vm-setivar-d.. 5b749d8e53) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-28T17:58:32Z yjit-get-exivar efaa8c9b09) +YJIT +PRISM [arm64-darwin24]
# Iteration per second (i/s)

|                           |compare-ruby|built-ruby|
|:--------------------------|-----------:|---------:|
|vm_ivar_get_uninitialized  |    138.801M|  138.452M|
|                           |       1.00x|         -|
|vm_ivar_get_on_obj         |     930.458|   936.865|
|                           |           -|     1.01x|
|vm_ivar_get_on_class       |     134.471|   431.622|
|                           |           -|     3.21x|
|vm_ivar_get_on_generic     |     146.679|   284.408|
|                           |           -|     1.94x|
```

yjit/src/codegen.rs
```diff
@@ -2946,28 +2929,33 @@ fn gen_get_ivar(
...
             } else {
-                // Compile time value is *not* embedded.
+                asm_comment!(asm, "call rb_ivar_get_at()");
 
-                // Get a pointer to the extended table
-                let tbl_opnd = asm.load(Opnd::mem(64, recv, ROBJECT_OFFSET_AS_HEAP_FIELDS as i32));
-
-                // Read the ivar from the extended table
-                let ivar_opnd = Opnd::mem(64, tbl_opnd, (SIZEOF_VALUE * ivar_index) as i32);
+                // The function could raise RactorIsolationError.
```
k0kubun: How about using `assume_single_ractor_mode()` and then `jit_prepare_non_leaf_call()` if it returns false?

variable.c
```diff
@@ -1463,6 +1463,36 @@ rb_ivar_get(VALUE obj, ID id)
...
+            VALUE fields_obj = RCLASS_WRITABLE_FIELDS_OBJ(obj);
+            VALUE val = rb_imemo_fields_ptr(fields_obj)[index];
+
+            if (UNLIKELY(!rb_ractor_main_p()) && !rb_ractor_shareable_p(val)) {
+                rb_raise(rb_eRactorIsolationError,
+                        "can not get unshareable values from instance variables of classes/modules from non-main Ractors");
+            }
+
+            return val;
+        }
```
k0kubun, line 1487: It's probably nice to put the unreachable case later so that `T_CLASS` and `T_MODULE` cases can run without checking if it's `T_OBJECT`.

```suggestion
      case T_CLASS:
      case T_MODULE:
        {
            VALUE fields_obj = RCLASS_WRITABLE_FIELDS_OBJ(obj);
            VALUE val = rb_imemo_fields_ptr(fields_obj)[index];

            if (UNLIKELY(!rb_ractor_main_p()) && !rb_ractor_shareable_p(val)) {
                rb_raise(rb_eRactorIsolationError,
                        "can not get unshareable values from instance variables of classes/modules from non-main Ractors");
            }

            return val;
        }
      case T_OBJECT:
        UNREACHABLE_RETURN(Qundef);
```

k0kubun (approved):

variable.c
byroot: That makes a difference? I had assumed the UNREACHEABLE would have allowed the compiler to ignore T_OBJECT.

I guess I'll play a bit with godbolt.

k0kubun: From the name, I assumed there are cases where it's compiled to a bare `return`, which does seem true. I see that it's probably `__builtin_unreachable()` for environments we primarily care about, but we also seem to support environments that don't have one.

byroot: Yeah, the unreacheable basically eliminate the branch.

yjit/src/codegen.rs
byroot: I'll look into it, thank for the pointers.

variable.c
XrXr: The `case` labels are not mandated to be checked in source code order (e.g. jump tables) so I'd say just do whatever looks nicest.

yjit/src/codegen.rs
byroot: Is feb35677cc1c47a11e86178842337243be6a54c1 how it's supposed to be used?

Do I understand correctly that the code is invalidated/recompiled if we enter multi-ractor mode? If so it would be possible to have a version of `rb_ivar_get_at` that doesn't do ractor checks.

byroot: > it would be possible to have a version of `rb_ivar_get_at` that doesn't do ractor checks.

That seems to raise the `vm_ivar_get_on_class` case at 3.5x baseline.

k0kubun: > Is https://github.com/ruby/ruby/commit/feb35677cc1c47a11e86178842337243be6a54c1 how it's supposed to be used?

Yes

> Do I understand correctly that the code is invalidated/recompiled if we enter multi-ractor mode?

Also yes. `rb_yjit_before_ractor_spawn` does `invalidate_block_version` on such blocks, and it re-inserts stubs to the incoming blocks of the invalidated block.

byroot: eabe69c4fece52a763ee26e694b05f179f3c4793

```
compare-ruby: ruby 3.5.0dev (2025-08-27T14:58:58Z merge-vm-setivar-d.. 5b749d8e53) +YJIT +PRISM [arm64-darwin24]
built-ruby: ruby 3.5.0dev (2025-08-28T21:23:38Z yjit-get-exivar 3cc21b76d4) +YJIT +PRISM [arm64-darwin24]

|                           |compare-ruby|built-ruby|
|:--------------------------|-----------:|---------:|
|vm_ivar_get_on_obj         |     975.981|   975.772|
|                           |       1.00x|         -|
|vm_ivar_get_on_class       |     136.214|   470.912|
|                           |           -|     3.46x|
|vm_ivar_get_on_generic     |     148.315|   299.122|
|                           |           -|     2.02x|
```

PR #14391, opened 2025-08-28
Add predicate attribute readers

havenwood:
This patch lets `attr_reader` take predicate method names that correspond to instance variables of the base name without a trailing question mark.

If the base name method doesn't already exist, a predicate name defines a base name method attribute reader to alias then undefines the base name.

For example, this creates an `enabled?` method that reads `@enabled`:
```ruby
attr_reader :enabled?
```

This feature is only supported for `attr_reader` and `attr`, not `attr_writer` or `attr_accessor`, since setter methods cannot have question marks.

Example:
```ruby
class Example
  attr_reader :valid?, :meaning

  def initialize
    @valid = [true, false].sample
    @meaning = 42
  end
end

Example.new.valid? #=> true
```

I created a corresponding [bugs.ruby-lang.org issue](https://bugs.ruby-lang.org/issues/21555).

### DIFF
--- a/object.c
+++ b/object.c
@@ -2361,7 +2361,14 @@ rb_mod_attr_reader(int argc, VALUE *argv, VALUE klass)
 
     for (i=0; i<argc; i++) {
         VALUE name_sym = argv[i];
-        VALUE name_str = rb_sym2str(name_sym);
+        VALUE name_str;
+        if (SYMBOL_P(name_sym)) {
+            name_str = rb_sym2str(name_sym);
+        }
+        else {
+            name_str = StringValue(name_sym);
+            name_sym = rb_str_intern(name_str);
+        }
 
         if (rb_is_predicate_name(name_str)) {
             const char *name_cstr = StringValueCStr(name_str);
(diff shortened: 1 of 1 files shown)


object.c
```diff
@@ -2349,9 +2360,33 @@ rb_mod_attr_reader(int argc, VALUE *argv, VALUE klass)
     VALUE names = rb_ary_new2(argc);
 
     for (i=0; i<argc; i++) {
-        ID id = id_for_attr(klass, argv[i]);
-        rb_attr(klass, id, TRUE, FALSE, TRUE);
-        rb_ary_push(names, ID2SYM(id));
+        VALUE name_sym = argv[i];
+        VALUE name_str = rb_sym2str(name_sym);
```
nobu: `Module#attr_reader` (and so on) accepts attribute name strings as well as symbols, while `rb_sym2str` argument must be `T_SYMBOL`.

rubyFeedback:
I think this should be suggested to an upcoming ruby developer meeting, considering how many ruby folks suggested this over the years.

The underlying use case is good, in my opinion. The trailing '?' at attr_reader is a bit ugly though. What about another attr-method name? Do we need to include the trailing '?'?

Anyway I think it would be useful to ask matz about the underlying idea first; and the syntax in a separate step.

The primary argument for the functionality would be that ruby users often use methods that end with a '?'. It makes the code easier to read:

    if game_over?
      do_cleanup_and_then_notify_the_user_that_the_game_is_over
    end

object.c
havenwood: Ah, right, that makes perfect sense. Thanks for explaining!

havenwood:
Closing since features rejected. Thank you for the feedback!

PR #14395, opened 2025-08-29
Skip `test_exivar_resize_with_compaction_stress` on s390x

hsbt:
https://rubyci.s3.amazonaws.com/s390x/ruby-3.3/log/20250828T220136Z.fail.html.gz


hsbt:
I missed https://github.com/ruby/ruby/pull/14269. This PR is duplicated with it.

PR #14397, opened 2025-08-29
Document Makefile rules in tool/enc-unicode.rb [ci skip]

XrXr:


nobu:
The PR with `[ci skip]` never start.

XrXr:
Thank you. It's all good, I'm posting this just in case anyone has comments.

PR #14400, opened 2025-08-29
ZJIT: Add InvokeSuper and InvokeSuperWithoutBlock

tekknolagi:
Support the same limited set of calls as SendWithoutBlock. Don't attempt
to cache yet.


tekknolagi:
I don't like this PR as-is. The reason I wanted to split up lookup and invoke instructions is because we can simplify our code: we can have Lookup, LookupSuper, ... and then just Send, SendWithoutBlock. We don't need to fully explode the powerset of combinations (and their optimized variants) into the IR. Ideally we only have ~1 kind of Ruby send that we can analyze, simplify, inline, etc. I don't want to treat invokesuper differently from e.g. invokeblock.

k0kubun:
That sounds right. YJIT lets those instructions share the same `gen_send_iseq`/`gen_send_cfunc` optimizations and we shouldn't give it up in ZJIT.

At the same time, however, all these instructions need separate fallback implementations. We might not need two `InvokeSuper*` instructions, but we need one for the `invokesuper` fallback (and another for `invokeblock`), just like `Insn::Send` and (non-direct) `Insn::SendWithoutBlock`.

tekknolagi:
Closing since it looks like you got to it.

PR #14407, opened 2025-08-29
ZJIT: Clear jit entry from iseqs after TracePoint activation

st0012:
This will allow TracePoint events to be triggered when calling previously JITed methods from the interpreter.
I also opened #14420 to find ways to add a new patch point for TracePoint but it's not working atm.

### DIFF
--- a/zjit/src/invariants.rs
+++ b/zjit/src/invariants.rs
@@ -269,3 +269,25 @@ pub extern "C" fn rb_zjit_before_ractor_spawn() {
         cb.mark_all_executable();
     });
 }
+
+#[unsafe(no_mangle)]
+pub extern "C" fn rb_zjit_tracing_invalidate_all() {
+    use crate::gc::{get_or_create_iseq_payload, IseqStatus};
+    use crate::cruby::{for_each_iseq, rb_iseq_reset_jit_func};
+
+    if !zjit_enabled_p() {
+        return;
+    }
+
+    // Stop other ractors since we are going to patch machine code.
+    with_vm_lock(src_loc!(), || {
+        debug!("Invalidating all ZJIT compiled code due to TracePoint");
+
+        for_each_iseq(|iseq| {
+            let payload = get_or_create_iseq_payload(iseq);
+
+            payload.status = IseqStatus::NotCompiled;
+            unsafe { rb_iseq_reset_jit_func(iseq) };
+        });
+    });
+}
--- a/depend
+++ b/depend
@@ -19732,6 +19732,7 @@ vm_trace.$(OBJEXT): {$(VPATH)}vm_opts.h
 vm_trace.$(OBJEXT): {$(VPATH)}vm_sync.h
 vm_trace.$(OBJEXT): {$(VPATH)}vm_trace.c
 vm_trace.$(OBJEXT): {$(VPATH)}yjit.h
+vm_trace.$(OBJEXT): {$(VPATH)}zjit.h
 weakmap.$(OBJEXT): $(CCAN_DIR)/check_type/check_type.h
 weakmap.$(OBJEXT): $(CCAN_DIR)/container_of/container_of.h
 weakmap.$(OBJEXT): $(CCAN_DIR)/list/list.h
--- a/jit.c
+++ b/jit.c
@@ -492,3 +492,51 @@ rb_jit_vm_unlock(unsigned int *recursive_lock_level, const char *file, int line)
 {
     rb_vm_lock_leave(recursive_lock_level, file, line);
 }
+
+void
+rb_iseq_reset_jit_func(const rb_iseq_t *iseq)
+{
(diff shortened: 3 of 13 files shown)


zjit/src/invariants.rs
```diff
@@ -269,3 +269,53 @@ pub extern "C" fn rb_zjit_before_ractor_spawn() {
...
+                // TODO: Do we free the IseqPayload?
+                payload.gc_offsets.clear();
+                payload.iseq_calls.clear();
+            }
+
+            // Reset output code entry point
+            unsafe { rb_iseq_reset_jit_func(iseq) };
+        });
+
+        // TODO: Do we need to clear variants?
```
st0012: @XrXr @k0kubun I try to mimic YJIT's structure of handling TracePoint in this PR. And I think walking through iseqs and call `rb_iseq_reset_jit_func` on them after TracePoint activation already allows TracePoint to run in some capacity. (I decide to add support for C return and line events separately).

But I'm not sure how to properly clean up/free the related payload and structs (like mentioned in the todo comments). Any pointer would be appreciated 🙏

k0kubun: > clean up/free the related payload and structs

Can you leave it to https://github.com/Shopify/ruby/issues/682 for now? Doing that only for TracePoint might rather complicate the situation.

zjit/src/codegen.rs
```diff
@@ -67,12 +67,17 @@ impl JITState {
...
 #[unsafe(no_mangle)]
-pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, _ec: EcPtr) -> *const u8 {
+pub extern "C" fn rb_zjit_iseq_gen_entry_point(iseq: IseqPtr, ec: EcPtr) -> *const u8 {
     // Do not test the JIT code in HIR tests
     if cfg!(test) {
         return std::ptr::null();
     }
 
+    // Check if c_call or c_return tracing is enabled - if so, don't compile
+    if unsafe { rb_c_method_tracing_currently_enabled(ec) } {
```
k0kubun: Why check that here? If you're trying to mimic what YJIT does, it should be a side exit on HIR instructions for C method calls (or dynamic dispatch that can be C method calls). We generally don't want to fail the compilation (`return null()`).

zjit/src/invariants.rs
k0kubun: If you really need that now for whatever reason, you just need to call `Box::from_raw` on the payload's raw pointer, clear `iseq->body->zjit_payload`, and that's it.

k0kubun:
> I think walking through iseqs and call rb_iseq_reset_jit_func on them after TracePoint activation already allows TracePoint to run in some capacity. (I decide to add support for C return and line events separately).

It only impacts JIT entries after TracePoint is enabled, which would miss returning from a non-leaf call to JIT code since we have no such patch points yet, but I guess it's fine to start from supporting just that.

zjit/src/invariants.rs
```diff
@@ -269,3 +269,53 @@ pub extern "C" fn rb_zjit_before_ractor_spawn() {
...
+        let invariants = ZJITState::get_invariants();
+        invariants.ep_escape_iseqs.clear();
+        invariants.no_ep_escape_iseqs.clear();
+        invariants.bop_patch_points.clear();
+        invariants.cme_patch_points.clear();
+        invariants.constant_state_patch_points.clear();
+        invariants.single_ractor_patch_points.clear();
```
k0kubun: Also, I don't think you should remove(disable) patch points until you introduce patch points for TracePoint and actually invalidate them https://github.com/ruby/ruby/pull/14407#issuecomment-3238256415. Clearing them only makes code less safe when TracePoint is not invalidated properly.

zjit/src/invariants.rs
st0012: I'll drop this bit for now 👍

st0012:
> It only impacts JIT entries after TracePoint is enabled, which would miss returning from a non-leaf call to JIT code since we have no such patch points yet, but I guess it's fine to start from supporting just that.

Thanks for the pointer. I tried to add a new patch point for that in #14420 but it's currently not working 😞  I'll log more info about it there

zjit/src/gc.rs
```diff
@@ -222,3 +222,14 @@ pub fn remove_gc_offsets(payload_ptr: *mut IseqPayload, removed_range: &Range<Co
 fn ranges_overlap<T>(left: &Range<T>, right: &Range<T>) -> bool where T: PartialOrd {
     left.start < right.end && right.start < left.end
 }
+
+/// Iterate over all existing ISEQs
+pub fn for_each_iseq<F: FnMut(IseqPtr)>(mut callback: F) {
```
k0kubun: How about putting this in `cruby.rs`? Unless `gc.rs` uses it by itself, it has nothing to do with GC.

k0kubun (approved):

zjit/src/gc.rs
st0012: Updated

PR #14409, opened 2025-08-29
ZJIT: NewRangeFixnum instruction

altxtech:
### Context
Closes Shopify/ruby#710

### Changes

**HIR**  
- Added a new instruction `NewRangeFixnum` that mirrors `NewRange` but is specialized for Fixnum endpoints.  
- Marked as leaf and effect-free (unlike `NewRange`, which may call `<=>`).

**Codegen**  
- Added `gen_new_range_fixnum`, wired to the new instruction.  
- Uses `gen_prepare_call_with_gc(asm, state)` since range allocation may trigger GC, but does not call Ruby methods.  
- Emits a direct call to `rb_range_new`.

**Optimization**  
- Added `optimize_ranges` pass: conservatively rewrites `NewRange` into `NewRangeFixnum` when one or both ends are statically Fixnum literals (with guards added for the dynamic side if needed).  
- This covers the simple literal/guardable cases that can be proven cheaply.

At this point I haven’t added rewrites for parameter–parameter ranges.  
From looking at how `opt_plus` works, it seems to rely on profiling info, but I don’t fully understand whether that same info is available for `newrange`.  
Since I’m still new to this codebase (and to low-level Rust in general), I decided to keep the optimization conservative for now and get maintainer feedback before broadening it.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -501,6 +501,7 @@ pub enum Insn {
     /// NewHash contains a vec of (key, value) pairs
     NewHash { elements: Vec<(InsnId,InsnId)>, state: InsnId },
     NewRange { low: InsnId, high: InsnId, flag: RangeType, state: InsnId },
+    NewRangeFixnum { low: InsnId, high: InsnId, flag: RangeType },
     ArrayDup { val: InsnId, state: InsnId },
     ArrayMax { elements: Vec<InsnId>, state: InsnId },
     /// Extend `left` with the elements from `right`. `left` and `right` must both be `Array`.
@@ -689,6 +690,7 @@ impl Insn {
             // TODO: NewRange is effects free if we can prove the two ends to be Fixnum,
             // but we don't have type information here in `impl Insn`. See rb_range_new().
             Insn::NewRange { .. } => true,
+            Insn::NewRangeFixnum { .. } => false,
             _ => true,
         }
     }
@@ -733,6 +735,9 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
             }
             Insn::NewRange { low, high, flag, .. } => {
                 write!(f, "NewRange {low} {flag} {high}")
+            }
+            Insn::NewRangeFixnum { low, high, flag, .. } => {
+                write!(f, "NewRangeFixnum {low} {flag} {high}")
             }
             Insn::ArrayMax { elements, .. } => {
                 write!(f, "ArrayMax")?;
@@ -1313,6 +1318,7 @@ impl Function {
                 NewHash { elements: found_elements, state: find!(state) }
             }
             &NewRange { low, high, flag, state } => NewRange { low: find!(low), high: find!(high), flag, state: find!(state) },
+            &NewRangeFixnum { low, high, flag } => NewRangeFixnum { low: find!(low), high: find!(high), flag },
             &ArrayMax { ref elements, state } => ArrayMax { elements: find_vec!(elements), state: find!(state) },
             &SetGlobal { id, val, state } => SetGlobal { id, val: find!(val), state },
             &GetIvar { self_val, id, state } => GetIvar { self_val: find!(self_val), id, state },
@@ -1383,6 +1389,7 @@ impl Function {
             Insn::NewHash { .. } => types::HashExact,
             Insn::HashDup { .. } => types::HashExact,
             Insn::NewRange { .. } => types::RangeExact,
+            Insn::NewRangeFixnum { .. } => types::RangeExact,
             Insn::CCall { return_type, .. } => *return_type,
             Insn::GuardType { val, guard_type, .. } => self.type_of(*val).intersection(*guard_type),
             Insn::GuardBitEquals { val, expected, .. } => self.type_of(*val).intersection(Type::from_value(*expected)),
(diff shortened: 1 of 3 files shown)


altxtech:
@tekknolagi 

Could you take a look at this draft PR? Want to make sure this is the right direction.

tekknolagi:
I'm on vacation for the next two weeks but someone else on the team can give it a look!

tekknolagi:
cc @ruby/jit

zjit/src/hir.rs
```diff
@@ -2082,6 +2135,11 @@ impl Function {
                 worklist.push_back(high);
                 worklist.push_back(state);
             }
+            &Insn::NewRangeFixnum { low, high, state, .. } => {
```
st0012: This can be stacked with `NewRange` like

```rs
            &Insn::NewRangeFixnum { low, high, state, .. } |
            &Insn::NewRange { low, high, state, .. } => {
                worklist.push_back(low);
                worklist.push_back(high);
                worklist.push_back(state);
            }
```

zjit/src/hir.rs
```diff
@@ -6548,6 +6608,116 @@ mod opt_tests {
         ");
     }
 
+    #[test]
+    fn test_optimize_range_fixnum_inclusive_literals() {
+        eval(r#"
```
st0012: I think none of the new tests new `r#` as there's no character needs escaping.

zjit/src/hir.rs
```diff
@@ -6548,6 +6608,116 @@ mod opt_tests {
...
+        ");
+    }
+
+    #[test]
+    fn test_optimize_range_fixnum_exclusive_low_guarded() {
+        eval(r#"
+            def test(a)
+              (a...10)
+            end
+            test(2); test(3)
```
st0012, line 6704: Can you add these cases to `test/ruby/test_zjit.rb` too as integration tests? Remember to add `call_threshold: 2` option so we actually run the optimzied HIR.

zjit/src/hir.rs
altxtech: Will check

zjit/src/hir.rs
altxtech: Fixed

zjit/src/hir.rs
altxtech: Fixed

zjit/src/hir.rs
altxtech: Tests added

zjit/src/hir.rs
```diff
@@ -1940,6 +1947,52 @@ impl Function {
...
+                    Insn::NewRange { low, high, flag, state } => {
+
+                        // The NewRange rewrite triggers mostly on literals because that is the
+                        // case we can easily prove Fixnum statically and cheaply guard the other
+                        // side.
+                        let low_is_fix  = self.is_a(low,  types::Fixnum);
+                        let high_is_fix = self.is_a(high, types::Fixnum);
+
+                        if low_is_fix && high_is_fix {
+                            // Both statically fixnum → specialize directly
```
k0kubun: > →

Could we use `=>` or `->` instead? IIRC somebody said we don't want non-ASCII characters in source code (including comments) in a past ZJIT PR, and I have the same preference.

The ruby/ruby repository has a few scripts that process source code and it tends to have problems with files with such characters. That particular issue is probably fixed today, but it still worries me (partly because we still have the problem to date in other places like commit messages). I want to stay away from it if possible.

Also, my input method configuration doesn't allow me to input "→" on my editor, so it would be tedious for me to write a similar comment when I want to follow the pattern and write similar code.

altxtech: Fixed

zjit/src/hir.rs
```diff
@@ -6548,6 +6604,116 @@ mod opt_tests {
...
+          CheckInterrupts
+          Return v2
+        ");
+    }
+
+    #[test]
+    fn test_optimize_range_fixnum_exclusive_literals() {
+        eval("
+            def test()
+              (1...2)
```
tekknolagi, line 6628: You may have noticed that this doesn't produce a New range instruction. For that you probably need to use a local variable (makes it harder for the bytecode compiler to reason about, but the JIT sees through it). If you have time, please rewrite these tests.

zjit/src/hir.rs
```diff
@@ -1940,6 +1947,52 @@ impl Function {
...
+                        let high_is_fix = self.is_a(high, types::Fixnum);
+
+                        if low_is_fix && high_is_fix {
+                            // Both statically fixnum => specialize directly
+                            let repl = self.push_insn(block, Insn::NewRangeFixnum { low, high, flag, state });
+                            self.make_equal_to(insn_id, repl);
+                            self.insn_types[repl.0] = self.infer_type(repl);
+                        } else if low_is_fix {
+                            // Only left is fixnum => guard right
+                            let high_fix = self.coerce_to_fixnum(block, high, state);
```
tekknolagi, line 1972: There's some code above to do this for you: see arguments_likely_fixnums and coerce something or other

zjit/src/hir.rs
altxtech: Will handle this later this week, along with you other comment.

zjit/src/hir.rs
altxtech: Didn't found `arguments_likely_fixnum` to be applicable here, unless we change how this optimization works.  

Profiling info is not enabled for `NewRange`, so this optimization fires if *either* end of the range is a fixnum, without looking  at profiling info. For example:

```ruby
def test(a)
  (1..a)
end
```

Is a case where the optimization fires even if we don't have profiling info for `a`, because the start of the range is a Fixnum literal.  

We *could* enable profiling for `NewRange` and just use `arguments_likely_fixnum` as you said. In effect, this would make this more similar to other Fixnum operations.  

~~I submitted a draft PR ruby#14607. Would like you to take a look before marking it ready for review.~~. Decided to just mark it as ready.

PR #14412, opened 2025-08-30
Initialize class dup/clone before calling initialize_dup/initialize_clone

jeremyevans:
Previously, you could override the class initialize_dup/initialize_clone method and the class hierarchy would not be set correctly inside the method before calling super.

This removes Class#initialize_copy, and instead makes Object#dup/clone call the underlying C function (rb_mod_init_copy) before calling the appropriate initialize_dup/initialize_clone method.

This results in the following fixes:

* The appropriate initialize_dup method is called (dup on a class will respect superclass initialize_dup).

* Inside class initialize_dup/initialize_clone/initialize_copy, class ancestor hierarchy is correct.

* Calling singleton_class inside initialize_dup no longer raises a TypeError later in dup.

* Calling singleton_class.ancestors inside initialize_dup no longer results in missing ancestors.

Fixes [Bug #21538]

### DIFF
--- a/test/ruby/test_class.rb
+++ b/test/ruby/test_class.rb
@@ -259,6 +259,48 @@ class TestClass < Test::Unit::TestCase
     assert_raise(TypeError) { BasicObject.dup }
   end
 
+  def test_class_hierarchy_inside_initialize_dup_bug_21538
+    initialize_dup_called = false
+    ancestors = sc_ancestors = nil
+    b = Class.new
+    b.define_singleton_method(:initialize_dup) do |x|
+      ancestors = self.ancestors
+      sc_ancestors = singleton_class.ancestors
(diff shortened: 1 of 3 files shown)


test/ruby/test_class.rb
```diff
@@ -259,6 +259,48 @@ def test_initialize_copy
     assert_raise(TypeError) { BasicObject.dup }
   end
 
+  def test_class_hierarchy_inside_initialize_dup_bug_21538
+    initialize_dup_called = false
```
nobu: Is this variable used?

jeremyevans: No. I was originally going to use it, and then decided on a different approach.  I'll remove it.

PR #14416, opened 2025-08-31
Fix console cursor on BSDs in async context

havenwood:
`IO.console.cursor` errors on kqueue-based systems when used with the fiber scheduler.

```ruby
require 'async'

Async { IO.console(:cursor) }.wait
#!> Errno::EINVAL: Invalid argument - IO_Event_Selector_KQueue_io_wait:IO_Event_Selector_KQueue_Waiting_register
```

It works on Linux but errors like above on macOS.

I don't know if it's an acceptable fix, but this patch just dups file descriptors for stdin/stdout/stderr if detected alongside kqueue.

cc @ioquatix

### DIFF
--- a/ext/io/console/console.c
+++ b/ext/io/console/console.c
@@ -96,6 +96,8 @@ extern VALUE rb_scheduler_timeout(struct timeval *timeout);
 # define rb_fiber_scheduler_make_timeout rb_scheduler_timeout
 #endif
 
+extern VALUE rb_fiber_scheduler_current(void);
+
 #ifndef HAVE_RB_IO_DESCRIPTOR
 static int
 io_descriptor_fallback(VALUE io)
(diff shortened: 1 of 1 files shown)


ext/io/console/console.c
```diff
@@ -1701,6 +1703,22 @@ console_dev(int argc, VALUE *argv, VALUE klass)
         out = rb_io_open_descriptor(klass, fd, FMODE_WRITABLE | FMODE_SYNC, path, Qnil, NULL);
 #endif
         fd = rb_cloexec_open(CONSOLE_DEVICE_FOR_READING, O_RDWR, 0);
```
nobu, line 1705: Thia `fd` will leak when the `fd` is duped below.
Probably, initializing `fd` to -1, and open if it is -1  **after** the `HAVE_SYS_EVENT_H` block.

ext/io/console/console.c
```diff
@@ -1701,6 +1703,22 @@ console_dev(int argc, VALUE *argv, VALUE klass)
...
+#ifdef HAVE_SYS_EVENT_H
+        VALUE scheduler = rb_fiber_scheduler_current();
+        if (!NIL_P(scheduler) && (isatty(0) || isatty(1) || isatty(2))) {
+            if (isatty(0)) {
+                fd = dup(0);
+                path = rb_obj_freeze(rb_str_new2("<STDIN>"));
+            } else if (isatty(1)) {
+                fd = dup(1);
+                path = rb_obj_freeze(rb_str_new2("<STDOUT>"));
+            } else if (isatty(2)) {
```
nobu, line 1715: Just a little style alignment.

```suggestion
            }
            else if (isatty(1)) {
                fd = dup(1);
                path = rb_obj_freeze(rb_str_new2("<STDOUT>"));
            }
            else if (isatty(2)) {
```

havenwood:
Opened a PR in the correct place. Thank you! https://github.com/ruby/io-console/pull/100

PR #14428, opened 2025-09-02
ZJIT: Count exits coming from jit_exception

k0kubun:
This PR adds a compile-error counter `compile_error_exception_handler`.

We currently don't allow entering JIT code for handling exceptions, which requires different frame handling like popping frames not pushed by JIT frames. To evaluate its importance to support it, I want to count exits coming from it.

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -445,7 +445,7 @@ jit_compile(rb_execution_context_t *ec)
 
         // At call-threshold, compile the ISEQ with ZJIT.
         if (body->jit_entry_calls == rb_zjit_call_threshold) {
-            rb_zjit_compile_iseq(iseq, ec, false);
+            rb_zjit_compile_iseq(iseq, false);
         }
     }
 #endif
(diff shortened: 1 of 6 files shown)


vm.c
```diff
@@ -493,8 +493,25 @@ jit_compile_exception(rb_execution_context_t *ec)
...
+
+        // At call-threshold, compile the ISEQ with ZJIT.
+        if (body->jit_exception_calls == rb_zjit_call_threshold) {
+            rb_zjit_compile_iseq(iseq, true);
+        }
+    }
+#endif
+
 #if USE_YJIT
+    // Increment the ISEQ's call counter and trigger JIT compilation if not compiled
```
aidenfoxivey, line 514: Is this comment also applicable to the USE_ZJIT block?

zjit.c
```diff
@@ -159,18 +159,22 @@ rb_zjit_reserve_addr_space(uint32_t mem_size)
...
+        uint8_t *rb_zjit_iseq_gen_entry_point(const rb_iseq_t *iseq, bool jit_exception); // defined in Rust
+        uintptr_t code_ptr = (uintptr_t)rb_zjit_iseq_gen_entry_point(iseq, jit_exception);
 
-        // TODO: support jit_exception
-        iseq->body->jit_entry = (rb_jit_func_t)code_ptr;
-}
+        if (jit_exception) {
+            iseq->body->jit_exception = (rb_jit_func_t)code_ptr;
+        }
+        else {
```
aidenfoxivey, line 174: nit: `else` could be on same line as closing brace for if stmt

aidenfoxivey (commented):
This PR makes sense to me. I added a few questions.

zjit.c
k0kubun: and that's not what ruby/ruby uses. 

There's even a committer who changes every commit of `} else {`  to `}\nelse {` too.

vm.c
k0kubun: The logic is different, so it's broken up into multiple comments for ZJIT, similarly to `jit_compile`.

zjit.c
aidenfoxivey: Ah interesting - good to know. Thank you!

PR #14429, opened 2025-09-02
ZJIT: Add miscellaneous fixes from `cargo clippy`

aidenfoxivey:


aidenfoxivey:
> For VS Code users, you can use the setting below to make clippy warnings show up in the editor:
> 
> ```json
>   {
>     "rust-analyzer.check.command": "clippy"
>   }
> ```

Would it be useful/relevant for me to write a Github Actions workflow to check for `clippy` conformance on each PR?

zjit/src/asm/arm64/arg/bitmask_imm.rs
```diff
@@ -93,8 +93,7 @@ impl BitmaskImmediate {
 impl BitmaskImmediate {
     /// Encode a bitmask immediate into a 32-bit value.
     pub fn encode(self) -> u32 {
-        0
-        | ((self.n as u32) << 12)
+        (((self.n as u32) << 12))
```
k0kubun: What's the point in the added parens? I don't get why it needs to be `(((self.n as u32) << 12))` instead of `((self.n as u32) << 12)`.

aidenfoxivey: Right - good point. Let me see if that's just a `clippy`-ism.

aidenfoxivey: Turns out yes, `clippy` just doesn't batch these together.

XrXr: Probably because otherwise it wouldn't parse right. IMO Clippy is just off base on this one and the current look is better.

k0kubun: > because otherwise it wouldn't parse right.

I think you're talking about `(self.n as u32) << 12` vs `((self.n as u32) << 12)`, not `(((self.n as u32) << 12))` vs `((self.n as u32) << 12)` that I brought up.

k0kubun (approved):
I'm fine with the change, but I'll leave it for Alan to make another comment since he doesn't seem to like part of the changes.

zjit/src/asm/arm64/arg/bitmask_imm.rs
XrXr: I guess the extra parentheses maintains vertical alignment with lines below, which still looks worse than using `0 |` and looks less intentional, as we both found out.

aidenfoxivey: Would it be preferred to remove that aspect of this PR?

XrXr: Yes please remove it. This is a clear intentional aesthetic choice that clippy doesn't understand. Suggestions from it makes the code looks worse in the end.

aidenfoxivey: Fixed!

aidenfoxivey:
Apologies for the force pushes - I just had noticed that I setup my git email as `--global`, which is rather uninformative.

PR #14435, opened 2025-09-03
vm_core.h: Add ppc64le to set the EC (Execution Context)

junaruga:


junaruga:
@nobu and @luke-gruber, could you review this PR? This PR is like a ppc64le version of nobu's past PR https://github.com/ruby/ruby/commit/f7059af50a31a4d27a04ace0beadb60616f3f971 for arm64. Thanks.

junaruga:
The ppc64le case passed: https://github.com/ruby/ruby/actions/runs/17440974950/job/49523628997?pr=14435 when when configuring with cppflags=-DRUBY_DEBUG and -O3.

nobu (approved):
A new macro for the preprocessor condition may be nice.

junaruga:
> A new macro for the preprocessor condition may be nice.

Thank you for your review. That's good idea. I will send another PR about that.

junaruga:
I updated the code comment with the `// TLS can not be accessed across .so on arm64 and perhaps ppc64le too.`, as I am not sure if the TLS (Thread-Local Storage) canot be accessed across .so is the case of ppc64le or not.

PR #14436, opened 2025-09-03
Improve format specification docs

deivid-rodriguez:
One example to describe how "*" works actually prints a warning:

```
$ ruby -we "sprintf('%d', 20, 14)"
=> -e:1: warning: too many arguments for format string
```

I think it's better to not use examples that print warnings, so I
propose to merge * docs with "width" specifier docs, and only include the
"correct" example.

After all I believe "*" is not an actual flag, but a special value that the
width specifier can take.


nobu:
Thank you, the change looks reasonable to me.
In addition to `Width Specifier`, `*` can be used as a `Precision Specifier`.
Could you add it too?

deivid-rodriguez:
Thank you @nobu, nice catch! I included your suggestion, plus some minor extra tweaks!

PR #14438, opened 2025-09-03
ZJIT: Further `clippy` fixes

aidenfoxivey:


aidenfoxivey:
Apologies for the mountain of commits - it should be easier to view this as individual line changes.

zjit/src/asm/arm64/inst/halfword_imm.rs
```diff
@@ -12,10 +12,10 @@ enum Index {
     None = 0b00,
 
     /// Mutate the register after the read.
-    PostIndex = 0b01,
+    Post = 0b01,
```
k0kubun, line 15: > https://github.com/ruby/ruby/pull/14438/commits/171f7d2a3f6f671ca79bdd10f15305200eb62bac

Out of curiosity, what kind of warning does clippy show for this?

aidenfoxivey: clippy::enum_variant_names

https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names

aidenfoxivey: Seems to be mostly stylistic, but of course ZJIT does have its own style that we may want to avoid writing over.

XrXr: This is bad. These are architectural names in A64. Revert.

PR #14439, opened 2025-09-03
ZJIT: Ensure `clippy` passes and silence unnecessary warnings

aidenfoxivey:


aidenfoxivey:
Noting the error - will figure out what the issue here is.

aidenfoxivey:
Screenshot 2025-09-03 at 2 24 52 PM
Looks like clippy passes.

zjit/src/codegen.rs
```diff
@@ -229,8 +232,10 @@ fn gen_iseq_body(cb: &mut CodeBlock, iseq: IseqPtr, function: Option<&Function>,
     Ok(start_ptr)
 }
 
+type IseqCallRef = Rc<RefCell<IseqCall>>;
```
k0kubun: Let's put that around `pub struct IseqCall { ... }`. This place is just one of its usages. It makes more sense to me to place that as part of the definition of the type.

aidenfoxivey:
test_reorder_c_args_no_cycle seems to be failing - sorting that out.

zjit/src/hir.rs
```diff
@@ -1083,8 +1080,7 @@ fn can_direct_send(iseq: *const rb_iseq_t) -> bool {
     else if unsafe { rb_get_iseq_flags_has_kw(iseq) } { false }
     else if unsafe { rb_get_iseq_flags_has_kwrest(iseq) } { false }
     else if unsafe { rb_get_iseq_flags_has_block(iseq) } { false }
-    else if unsafe { rb_get_iseq_flags_forwardable(iseq) } { false }
-    else { true }
+    else { !unsafe { rb_get_iseq_flags_forwardable(iseq) } }
```
k0kubun: The same discussion as https://github.com/ruby/ruby/pull/14429#discussion_r2317197402 seems to apply. Let's just turn it off.

aidenfoxivey: got it!

k0kubun: like I don't want to modify this line when I want to put another condition after this. It just makes it diff-noisy and less consistent.

aidenfoxivey: Okay, I've added a flag to clippy to just let the original version stay as is.

zjit/src/codegen.rs
aidenfoxivey: Sorted this out.

zjit/src/hir.rs
```diff
@@ -630,23 +633,17 @@ pub enum Insn {
 impl Insn {
     /// Not every instruction returns a value. Return true if the instruction does and false otherwise.
     pub fn has_output(&self) -> bool {
-        match self {
-            Insn::Jump(_)
+        !matches!(self, Insn::Jump(_)
```
k0kubun: I'm fine-ish with `matches!` for `is_terminator`, but with `!`, what's happening here seems less obvious than the previous code. Can we just turn it off for this file and leave `has_output` as is?

aidenfoxivey: Absolutely - will do!

zjit/src/hir.rs
```diff
@@ -1807,12 +1805,10 @@ impl Function {
                             // entered the compiler.  That means we can just return nil for this
                             // shape + iv name
                             Insn::Const { val: Const::Value(Qnil) }
+                        } else if recv_type.flags().is_embedded() {
+                            Insn::LoadIvarEmbedded { self_val, id, index: ivar_index }
                         } else {
-                            if recv_type.flags().is_embedded() {
```
k0kubun, line 1811: This also seems like an intentional stylistic choice to emphasize that these two are similar to each other but less relevant to the other `Insn::Const` case. Maybe nice to turn that off too?

aidenfoxivey: Works for me!

zjit/src/backend/lir.rs
```diff
@@ -1243,17 +1244,16 @@ impl Assembler
...
+        fn find_safe_move(moves: &[(Reg, Opnd)]) -> Option<usize> {
             moves.iter().enumerate().find(|&(_, &(dest_reg, _))| {
                 moves.iter().all(|&(_, src_opnd)| src_opnd != Opnd::Reg(dest_reg))
             }).map(|(index, _)| index)
         }
 
         // Remove moves whose source and destination are the same
-        let mut old_moves: Vec<(Reg, Opnd)> = old_moves.clone().into_iter()
-            .filter(|&(reg, opnd)| Opnd::Reg(reg) != opnd).collect();
+        let mut old_moves: Vec<(Reg, Opnd)> = old_moves.to_owned();
```
k0kubun: I don't think this does "Remove moves whose source and destination are the same" anymore. Whatever it is that forces you to remove the filter, please just turn it off.

aidenfoxivey: Ah right - this is what I was just thinking - let me see if that was my own doing or Clippy's.

aidenfoxivey: I think I have a better solution since Reg and Opnd implement Copy,

```rust
        let mut old_moves: Vec<(Reg, Opnd)> = old_moves.iter()
            .copied()
            .filter(|(reg, opnd)| Opnd::Reg(*reg) != *opnd)
            .collect();
```

aidenfoxivey: It passes the tests on my machine - I will see whether CI treats it similarly.

zjit/src/hir.rs
```diff
@@ -3,6 +3,10 @@
 // We use the YARV bytecode constants which have a CRuby-style name
 #![allow(non_upper_case_globals)]
 
+#![allow(clippy::if_same_then_else)]
+#![allow(clippy::too_many_arguments)]
```
XrXr: Let's lift this `clippy::too_many_arguments` to `lib.rs` and just make it crate wide. (remove other instances)

zjit/src/virtualmem.rs
```diff
@@ -3,6 +3,8 @@
 // usize->pointer casts is viable. It seems like a lot of work for us to participate for not much
 // benefit.
 
+#![allow(clippy::enum_variant_names)]
```
XrXr: ditto, make it ZJIT crate-wide

zjit/src/hir.rs
XrXr: Revert this hunk.

zjit/src/backend/lir.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::double_must_use)]
```
XrXr: I think the `#[must_use]` was just left-over from when the return type was an `Option`. We can fix it by removing the `#[must_use]`.

zjit/src/hir.rs
k0kubun: You marked the thread resolved but the diff has not been reverted yet.

XrXr (commented):
There is some Clippy warnings for x86 we should address too https://github.com/ruby/ruby/actions/runs/17443589349/job/49532376056?pr=14439#step:3:31

k0kubun:
Could you rebase the branch to fix the conflicts?

k0kubun (approved):

zjit/src/profile.rs
```diff
@@ -148,7 +149,7 @@ impl Flags {
 /// opt_send_without_block/opt_plus/... should store:
 /// * the class of the receiver, so we can do method lookup
 /// * the shape of the receiver, so we can optimize ivar lookup
-/// with those two, pieces of information, we can also determine when an object is an immediate:
+///   with those two, pieces of information, we can also determine when an object is an immediate:
```
tekknolagi, line 152: this is now more confusing than it was before. now it looks like it continues the previous list item :(

aidenfoxivey: Fixing!

aidenfoxivey: https://github.com/ruby/ruby/pull/14580

PR #14444, opened 2025-09-04
ZJIT: Compile invokesuper with dynamic dispatch

k0kubun:
This PR starts compiling `invokesuper` with dynamic dispatch, partly taking over https://github.com/ruby/ruby/pull/14400. Non-dynamic dispatch calls are out of scope in this PR.

It also relaxes the `has_send` (renamed to `has_blockiseq`) condition in `compute_bytecode_info` to instructions with `blockiseq`, which are required to modify parent locals in those instructions without acquiring Binding.

It increases `ratio_in_zjit` on lobsters from 64.2% to 66.8%.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -565,10 +565,13 @@ pub enum Insn {
     /// `name` is for printing purposes only
     CCall { cfun: *const u8, args: Vec<InsnId>, name: ID, return_type: Type, elidable: bool },
 
-    /// Send without block with dynamic dispatch
+    /// Un-optimized fallback implementation (dynamic dispatch) for send-ish instructions
     /// Ignoring keyword arguments etc for now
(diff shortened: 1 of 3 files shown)


zjit/src/hir.rs
```diff
@@ -4971,7 +5023,6 @@ mod tests {
         assert_snapshot!(hir_string("test"), @r"
         fn test@<compiled>:2:
         bb0(v0:BasicObject, v1:BasicObject):
-          v5:BasicObject = GetLocal l0, EP@3
```
XrXr, line 4974: The check on blockiseq is more accurate now. Nice. Which reminds me, could you add a `super &nil` HIR test?

st0012 (approved):

zjit/src/hir.rs
k0kubun: like this? https://github.com/ruby/ruby/pull/14444/commits/11d8a10a52cf9914df72e30ab2aaccddb99bf736

PR #14446, opened 2025-09-04
ZJIT: Fix 30k if stmt test

aidenfoxivey:
Closes [#726](https://github.com/Shopify/ruby/issues/726)

This should be treated as a draft now.

I have sorted this out for ARM64, but am in process of investigating whether it is an issue for x86_64.


zjit/src/backend/arm64/mod.rs
```diff
@@ -765,11 +765,36 @@ impl Assembler
...
+                            // NOTE: I'm skeptical of the `.try_into().unwrap()`
+                            let load_insns: i32 = emit_load_size(dst_addr.try_into().unwrap()).into();
+                            bcond(cb, Condition::inverse(CONDITION), (load_insns + 2).into());
+                            emit_load_value(cb, Assembler::SCRATCH0, dst_addr.try_into().unwrap());
+                            br(cb, Assembler::SCRATCH0);
+                            load_insns + 2
+                        };
+
+                        assert!(num_insns <= cb.conditional_jump_insns());
+                        for _ in num_insns..cb.conditional_jump_insns() { nop(cb); }
```
k0kubun: We don't rewrite label jumps, so we shouldn't need `nop`s here.

XrXr: It's a conditional branch though, so the fallthrough side of the branch need the `nop`s

k0kubun:
Can you add a test for it? It probably requires a relatively large ISEQ, so maybe use an `eval` to define a method with repeated instructions and call it in `test/ruby/test_zjit.rb`. 

> am in process of investigating whether it is an issue for x86_64.

IIRC it's probably fine, but you can find this out by adding that test (which x86_64 CI will run when pushed) too.

aidenfoxivey:
Certainly - I’ll have a test right up!

My assumption is that the nops were there to increment the instruction pointer should the branch not be taken.

zjit/src/backend/arm64/mod.rs
k0kubun: I'm not sure if I follow. Could you provide an example script that would trigger the problem?

k0kubun: Oh, sorry, I get it now. `label_ref` takes `num_bytes` and it has to fill out the bytes it promised when linked. Never mind.

zjit/src/backend/arm64/mod.rs
```diff
@@ -765,11 +765,36 @@ impl Assembler
...
+                        } else {
+                            // Since we cannot encode the offset into an instruction, load it into
+                            // some number of registers and then call br.
+                            // NOTE: I'm skeptical of the `.try_into().unwrap()`
+                            let load_insns: i32 = emit_load_size(dst_addr.try_into().unwrap()).into();
+                            bcond(cb, Condition::inverse(CONDITION), (load_insns + 2).into());
+                            emit_load_value(cb, Assembler::SCRATCH0, dst_addr.try_into().unwrap());
+                            br(cb, Assembler::SCRATCH0);
+                            load_insns + 2
+                        };
```
k0kubun: Now that it basically does the same thing as the `Target::CodePtr` one, how about carving them out as a separate function (maybe defined inside `emit_conditional_jump`) and using it?

aidenfoxivey: Right - good idea.

aidenfoxivey: In the prior block,
```
                    if let Target::CodePtr(_) = target {
                        // We need to make sure we have at least 6 instructions for
                        // every kind of jump for invalidation purposes, so we're
                        // going to write out padding nop instructions here.
                        assert!(num_insns <= cb.conditional_jump_insns());
                        for _ in num_insns..cb.conditional_jump_insns() { nop(cb); }
                    }
```

This seems redundant given that to end up calling this, we would have already hit this

```
match target {
                Target::CodePtr(dst_ptr) => {
```

Does that seem right?

k0kubun: Yeah, I think it's just because we used to have another `Target` that is compiled in the same way as `Target` and then removed it for ZJIT. Feel free to remove the `if let Target::CodePtr(_) = target {` check and share the nop logic with labels too.

zjit/src/backend/arm64/mod.rs
```diff
@@ -706,70 +706,50 @@ impl Assembler
...
+                    // Since we cannot encode the offset in a bcond directly, generate
+                    // an unconditional branch and then jump past it only if the inverse
+                    // of the original condition is met.
+                    bcond(cb, Condition::inverse(CONDITION), 2.into());
+                    b(cb, offset);
+                    2
+                } else {
+                    // Since we cannot encode the offset into an instruction, load it into
+                    // some number of registers and then call br.
+                    // NOTE: I'm skeptical of the `.try_into().unwrap()`
```
XrXr: It's because it's wrong. For cases where it'd panic (negative numbers) we actually always want to always preserve the bit string of the address. So this should always use a `i64 as u64` cast. Userspace addresses only have the sign bit set in illumos/Solaris descendants, though.

Can we take old `CodePtr` version instead of this rewrite?
You also have some legit CI crashes.

aidenfoxivey: > Userspace addresses only have the sign bit set in illumos/Solaris descendants, though.
Interesting - I hadn't heard of that.

> Can we take old CodePtr version instead of this rewrite?
Sure!

zjit/src/backend/arm64/mod.rs
```diff
@@ -706,67 +706,67 @@ impl Assembler
...
+                    // took to write out the destination address + 1 for the
+                    // b.cond and 1 for the br.
+                    load_insns + 2
+                };
+
+                // We need to make sure we have at least 6 instructions for
+                // every kind of jump for invalidation purposes, so we're
+                // going to write out padding nop instructions here.
+                assert!(num_insns <= cb.conditional_jump_insns());
+                (num_insns..cb.conditional_jump_insns()).for_each(|_| nop(cb));
```
aidenfoxivey, line 756: I opted for this since it's a bit more idiomatic than a `for _ in`. Happy to revert though.

zjit/src/backend/arm64/mod.rs
```diff
@@ -706,67 +706,67 @@ impl Assembler
...
+                    // instruction used for branching.
+                    bcond(cb, Condition::inverse(CONDITION), 2.into());
+                    b(cb, InstructionOffset::from_bytes((dst_addr - (src_addr + 4)) as i32)); // + 4 for bcond
+
+                    // We've only written out 2 instructions.
+                    2
+                } else {
+                    // Otherwise, we need to load the address into a
+                    // register and use the branch register instruction.
+                    let load_insns: i32 = emit_load_size(dst_addr as u64).into();
```
aidenfoxivey, line 737: Previously, the code calculated `dst_addr` directly in this scope. I've removed it - but I suspect I am wrong.

aidenfoxivey: If I am wrong to do this though, it appears that `make zjit-check` doesn't catch this.

k0kubun:
Let's make it a draft to signal the status of the PR properly, until you add a test and the fix that passes it. The current version doesn't change the `Target::Label` case at all, so IIUC the diff is not a bugfix at the moment.

aidenfoxivey:
> Let's make it a draft to signal the status of the PR properly, until you add a test and the fix that passes it. The current version doesn't change the `Target::Label` case at all, so IIUC the diff is not a bugfix at the moment.

I'm not actually sure I can mark things as a draft myself yet - how should I do that going forward?

k0kubun:
## Opening a draft PR

When you open a PR, you can choose "Create draft pull request". 

When you're not sure if every CI job is gonna pass (or you want to discuss a WIP patch), I would recommend opening a draft PR first, wait for the CI results, and then mark it ready for reviews.

Screenshot 2025-09-05 at 09 17 49

## Converting a PR to a draft PR

On the right side, under "Reviewers" section, there's "Convert to draft".

Screenshot 2025-09-05 at 09 18 24

aidenfoxivey:
> ## Opening a draft PR
> When you open a PR, you can choose "Create draft pull request".
> 
> When you're not sure if every CI job is gonna pass (or you want to discuss a WIP patch), I would recommend opening a draft PR first, wait for the CI results, and then mark it ready for reviews.
> 
> Screenshot 2025-09-05 at 09 17 49
> ## Converting a PR to a draft PR
> On the right side, under "Reviewers" section, there's "Convert to draft".
> 
> Screenshot 2025-09-05 at 09 18 24

strange - it didn't seem like the convert to draft was showing up for me earlier. Thank you for the note - I will make sure to follow this.

k0kubun:
Sorry, I did it myself this time. It's probably why.

Screenshot 2025-09-05 at 10 09 26

aidenfoxivey:
```
 
    1) Error:
  TestOpenURI#test_read_timeout:
  Test::Unit::ProxyError: Broken pipe
      D:/a/ruby/ruby/src/test/open-uri/test_open-uri.rb:80:in 'IO#write'
      D:/a/ruby/ruby/src/test/open-uri/test_open-uri.rb:80:in 'IO#print'
      D:/a/ruby/ruby/src/test/open-uri/test_open-uri.rb:80:in 'block (2 levels) in TestOpenURI#test_read_timeout'
  
  Finished tests in 572.174219s, 50.3518 tests/s, 10697.4568 assertions/s.
  28810 tests, 6120809 assertions, 0 failures, 1 errors, 270 skips
  
  ruby -v: ruby 3.5.0dev (2025-09-08T16:36:33Z pull/14446/merge ebd0420598) +PRISM [x64-mswin64_140]
  NMAKE : fatal error U1077: '@ .\ruby.exe -I../src/lib -I".ext/x64-mswin64_140" -I. "../src/tool/runruby.rb" --extout=".ext"  -- --disable-gems -r../src/tool/lib/_tmpdir  "../src/test/runner.rb" --ruby=".\ruby.exe -I../src/lib -I".ext/x64-mswin64_140" -I. "../src/tool/runruby.rb" --extout=".ext"  -- --disable-gems"  --excludes-dir=../src/test/.excludes --name=!/memory_leak/ -j6 --launchable-test-reports=launchable_test_all_report.json' : return code '0x1'
```
this is from the windows case - this strikes me like flakey due to the timeout.

zjit/src/asm/mod.rs
```diff
@@ -310,6 +310,18 @@ impl CodeBlock {
 
         Self::new(Rc::new(RefCell::new(virt_mem)), false)
     }
+
+    pub fn new_dummy_sized(mem_size: usize) -> Self {
```
k0kubun, line 306: Can you update `new_dummy()` to use `new_dummy_sized()` to share the code?

k0kubun:
Yeah don't worry about the flaky failure. If your PR is ready, please press "Ready for review".

zjit/src/asm/mod.rs
aidenfoxivey: Absolutely - will do.

zjit/src/backend/arm64/mod.rs
```diff
@@ -706,70 +706,68 @@ impl Assembler
...
+                    load_insns + 2
+                };
+
+                // We need to make sure we have at least 6 instructions for
+                // every kind of jump for invalidation purposes, so we're
+                // going to write out padding nop instructions here.
+                assert!(num_insns <= cb.conditional_jump_insns());
+                (num_insns..cb.conditional_jump_insns()).for_each(|_| nop(cb));
+            }
+
```
k0kubun, line 758: I think we generally don't put empty lines consecutively

```suggestion
```

PR #14448, opened 2025-09-04
ZJIT: Invalidate local variables on EP escape

k0kubun:
This PR implements invalidation for local variables in case EP is escaped. It fixes 9 failures on `test-spec`.

There have been TODO comments in `gen_send*` to add a patch point after these `send`s, but it's not appropriate because its `FrameState` is a before-send state and there are many other non-leaf instructions. So this PR adds a patch point before getlocal/setlocal that follows any non-leaf instruction.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -136,6 +136,8 @@ pub enum Invariant {
     },
     /// TracePoint is not enabled. If TracePoint is enabled, this is invalidated.
     NoTracePoint,
+    /// cfp->ep is not escaped to the heap on the ISEQ
+    NoEPEscape(IseqPtr),
     /// There is one ractor running. If a non-root ractor gets spawned, this is invalidated.
     SingleRactorMode,
 }
@@ -250,6 +252,7 @@ impl<'a> std::fmt::Display for InvariantPrinter<'a> {
                 write!(f, ")")
             }
             Invariant::NoTracePoint => write!(f, "NoTracePoint"),
+            Invariant::NoEPEscape(iseq) => write!(f, "NoEPEscape({})", &iseq_name(iseq)),
             Invariant::SingleRactorMode => write!(f, "SingleRactorMode"),
         }
     }
@@ -2705,6 +2708,15 @@ pub struct FrameState {
     locals: Vec<InsnId>,
 }
 
+impl FrameState {
+    /// Return itself without locals. Useful for side-exiting without spilling locals.
+    fn without_locals(&self) -> Self {
+        let mut state = self.clone();
+        state.locals.clear();
+        state
+    }
+}
+
 /// Print adaptor for [`FrameState`]. See [`PtrPrintMap`].
 pub struct FrameStatePrinter<'a> {
     inner: &'a FrameState,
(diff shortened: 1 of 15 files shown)


zjit/src/hir.rs
```diff
@@ -3096,12 +3108,16 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                 profiles.profile_stack(&exit_state);
             }
 
-            unsafe extern "C" {
-                fn rb_iseq_event_flags(iseq: IseqPtr, pos: usize) -> rb_event_flag_t;
+            // Flag a future getlocal/setlocal to add a patch point if this instruction may not be leaf.
+            if unsafe { !rb_zjit_insn_is_always_leaf(opcode as i32) } {
```
XrXr: I think leafness of an instruction is solely defined by the opcode and the operands of the instruction, and since we have access to both here, I'm not sure I understand why we need to add a new `rb_zjit_insn_is_always_leaf()`. If there's a function to check the leafness of a whole instruction, we should use that. (I'm can't finding it, though.)

tool/ruby_vm/views/_leaf_helpers.erb
XrXr, line 1: Off topic `leafness_of_check_ints` is unused now and could be cleaned up. That's a blast from the past: https://github.com/ruby/ruby/commit/5d74894f2bc4a3a18aec952d946ead3d784cb4b4

PC movement is consistent now.

zjit/src/hir.rs
XrXr: Practically, this only impacts `leafness_of_checkmatch` and `defined` after #14449.

k0kubun: > If there's a function to check the leafness of a whole instruction, we should use that.

Because vm.inc can render different function names for each instruction in the insn_entry directly, there's no such function. https://github.com/ruby/ruby/commit/83bc6ced920c51ffc3783b03f2e1c7f89f5c2fd5 was the only function we've used at JIT for checking leafness, and I was just adjusting the behavior and the location for ZJIT's use, assuming the behavior is the best we can get at compile time.

However,

> I think leafness of an instruction is solely defined by the opcode and the operands of the instruction

this sounds right. I guess I just didn't care about `defined` and `checkmatch` too much, so I probably chose to simplify that when I added `insn_may_depend_on_sp_or_pc`. But we shouldn't compromise it for ZJIT. I'll just add a function that does that then. We do need a new function.

tool/ruby_vm/views/_leaf_helpers.erb
k0kubun: It took a little for me to process what's going on in the commit since that person didn't use a constant or a macro for the thing that doesn't need to be a variable, but yeah we can clean that up now.

zjit/src/hir.rs
k0kubun: Added `insn_leaf` function https://github.com/ruby/ruby/pull/14448/commits/ff43c57d2e7440ad63b5618ee9e4b89ba8aee69c

tool/ruby_vm/views/_leaf_helpers.erb
k0kubun: fixed in https://github.com/ruby/ruby/commit/799d57bd0178168c977cc0c2b45ff3a5f01d1297

XrXr (approved):

zjit/src/hir.rs
```diff
@@ -7324,10 +7360,11 @@ mod opt_tests {
           v1:NilClass = Const Value(nil)
           v6:ArrayExact = NewArray
           PatchPoint MethodRedefined(Array@0x1000, itself@0x1008, cme:0x1010)
-          v18:BasicObject = CCall itself@0x1038, v6
-          v11:Fixnum[1] = Const Value(1)
+          v20:BasicObject = CCall itself@0x1038, v6
+          PatchPoint NoEPEscape(test)
```
tekknolagi, line 7364: It might be worth having a pass to add these no-EP-escape PatchPoints "in post"---after optimization---because we plan on lowering a bunch of bytecode-effectful operations into HIR-non-effectful operations (FixnumAdd, CCall itself, etc) in the optimizer. We can still use the FrameState following the effectful operation.

k0kubun: makes sense

PR #14455, opened 2025-09-05
Allow IO#each_codepoint to work with unetc even when encoding conversion is active

YO4:
Using IO#each_codepoint together with IO#ungetc causes an unwanted exception when encoding conversion is active.

```
C:\>ruby -e "open('NUL', 'rt') { |f| f.ungetc('aa'); f.each_codepoint { |c| p c }}"
97
-e:1:in 'IO#each_codepoint': byte oriented read for character buffered IO (IOError)
        from -e:1:in 'block in <main>'
        from -e:1:in 'Kernel#open'
        from -e:1:in '<main>'
```

This PR allows calling IO#each_codepoint after IO#ungetc even when encoding conversion is enabled.
Fixes [Bug #21562]

### DIFF
--- a/io.c
+++ b/io.c
@@ -4947,7 +4947,7 @@ rb_io_each_codepoint(VALUE io)
             fptr->cbuf.off += n;
             fptr->cbuf.len -= n;
             rb_yield(UINT2NUM(c));
-            rb_io_check_byte_readable(fptr);
+            rb_io_check_char_readable(fptr);
         }
     }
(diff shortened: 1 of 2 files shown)


nobu:
Do you mean [[Bug #21562]](https://bugs.ruby-lang.org/issues/21562)?

YO4:
Yes, that is correct.
I have updated the PR text.

PR #14456, opened 2025-09-05
Add a macro to manage the condition of no-inline version rb_current_ec

junaruga:
This PR comes from the https://github.com/ruby/ruby/pull/14435#pullrequestreview-3184697090.

Add the macro RB_THREAD_CURRENT_EC_NOINLINE to manage the condition to use no-inline version rb_current_ec for a better maintainability.

### DIFF
--- a/thread_pthread.h
+++ b/thread_pthread.h
@@ -17,6 +17,11 @@
 #define RB_NATIVETHREAD_LOCK_INIT PTHREAD_MUTEX_INITIALIZER
 #define RB_NATIVETHREAD_COND_INIT PTHREAD_COND_INITIALIZER
 
+// TLS can not be accessed across .so on arm64 and perhaps ppc64le too.
+#if defined(__arm64__) || defined(__aarch64__) || defined(__powerpc64__)
+# define RB_THREAD_CURRENT_EC_NOINLINE
+#endif
+
 // this data should be protected by timer_th.waiting_lock
 struct rb_thread_sched_waiting {
     enum thread_sched_waiting_flag {
@@ -133,8 +138,7 @@ struct rb_thread_sched {
 #ifdef RB_THREAD_LOCAL_SPECIFIER
   NOINLINE(void rb_current_ec_set(struct rb_execution_context_struct *));
 
-  # if defined(__arm64__) || defined(__aarch64__) || defined(__powerpc64__)
-    // TLS can not be accessed across .so on arm64 and perhaps ppc64le too.
+  # ifdef RB_THREAD_CURRENT_EC_NOINLINE
     NOINLINE(struct rb_execution_context_struct *rb_current_ec(void));
   # else
     RUBY_EXTERN RB_THREAD_LOCAL_SPECIFIER struct rb_execution_context_struct *ruby_current_ec;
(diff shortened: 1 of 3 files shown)


junaruga:
The `vm_core.h` has the `#include THREAD_IMPL_H`.

https://github.com/ruby/ruby/blob/0d91997cba15f7b9150b2e1d61f09d0761697cdb/vm_core.h#L179

According to the following logic, it seems that the value of the `THREAD_IMPL_H` can be `thread_none.h`, `thread_pthread.h` or `thread_win32.h`. So, I am not sure if we need to add the same definition of `RB_THREAD_CURRENT_EC_NOINLINE` macro in `thread_none.h` and `thread_win32.h` too.

https://github.com/ruby/ruby/blob/0d91997cba15f7b9150b2e1d61f09d0761697cdb/tool/m4/ruby_thread.m4#L63-L74

nobu:
It seems OK by using `#ifdef RB_THREAD_CURRENT_EC_NOINLINE`, instead of `#if`.

junaruga:
> It seems OK by using `#ifdef RB_THREAD_CURRENT_EC_NOINLINE`, instead of `#if`.

@nobu I rebased this PR using the `#ifdef RB_THREAD_CURRENT_EC_NOINLINE`, instead of `#if`.

According to the following logic, the `thread_win32.h` is included, when the target OS is MinGW, and `thread_none.h` is included when the target OS is WASI (WebAssembly).

https://github.com/ruby/ruby/blob/ac24f70fb0c0cc80d30ec6e96b3369079775d0dc/tool/m4/ruby_thread.m4#L3-L33

In the MInGW case below, the `make` passed without warnings related to the `RB_THREAD_CURRENT_EC_NOINLINE`.

https://github.com/ruby/ruby/actions/runs/17551203407/job/49843860669?pr=14456#step:9:2

In the WebAssembly case below, the `make mini ruby` passed without warnings related to the `RB_THREAD_CURRENT_EC_NOINLINE`.

https://github.com/ruby/ruby/actions/runs/17551203381/job/49843860613?pr=14456#step:10:2

There are is a change by this PR.

If `./configure --with-thread=none` is used in the arm64 case, or if there is an arm64 version of the MInGW or WebAssembly, the no-inline version rb_current_ec was previously used for the arm64 version, but now the cases use the inline version rb_current_ec.

So, what do you think now? The PR looks good to you?

junaruga:
How about creating a new header file such as a `thread_common.h` included commonly from `thread_none.h`, `thread_pthread.h` and `thread_win32.h`, and defining the `RB_THREAD_CURRENT_EC_NOINLINE` in the `thread_common.h`?

nobu (approved):
In the thread=none case, thread feature is not used at all, including TLS.
In the thread=win32 case, `RB_THREAD_LOCAL_SPECIFIER` is undefined and use of TLS is disabled.
That means, we need to consider TLS only in the thread=pthread case.

junaruga:
> In the thread=none case, thread feature is not used at all, including TLS. In the thread=win32 case, `RB_THREAD_LOCAL_SPECIFIER` is undefined and use of TLS is disabled. That means, we need to consider TLS only in the thread=pthread case.

Thank you for your review! I updated the commit message including what you said, and rebased ths PR on the latest master branch. I am going to merge the PR after I confirm that CI passes.

PR #14461, opened 2025-09-05
ZJIT: Fix backtraces on opt_new

k0kubun:


zjit/src/hir.rs
```diff
@@ -3340,16 +3353,30 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                        target: BranchEdge { target, args: state.as_args(self_param) }
+                    });
                     queue.push_back((state.clone(), target, target_idx, local_inval));
-                    break;  // Don't enqueue the next block as a successor
+
+                    // Move on to the fast path
+                    let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    let insn_id = fun.push_insn(block, Insn::ObjectAlloc { val, state: exit_id });
+                    state.stack_setn(argc, insn_id);
+                    state.stack_setn(argc + 1, insn_id);
```
st0012, line 3379: For my own learning: why do we need to set these?

st0012 (approved):

zjit/src/hir.rs
k0kubun: It's how `opt_new` is compiled in the ISEQ and implemented in `insns.def`; it does `TOPN(argc) = val;` and `TOPN(argc + 1) = val;` on this path.

The `TOPN(argc)` one is used as a receiver for `#initialize`, and the `TOPN(argc + 1)` one is used as a return value of the whole optimized sequence for `opt_new`, ignoring the return value from the `#initialize` call.

zjit/src/hir.rs
```diff
@@ -1417,6 +1427,7 @@ impl Function {
             Insn::HashDup { .. } => types::HashExact,
             Insn::NewRange { .. } => types::RangeExact,
             Insn::NewRangeFixnum { .. } => types::RangeExact,
+            Insn::ObjectAlloc { .. } => types::HeapObject,
```
tekknolagi, line 1430: Let's specialize this with the class of passed into `ObjectAlloc`! Then we could get rid of a `GuardType` here

k0kubun: sounds good :+1:

k0kubun: Filed https://github.com/Shopify/ruby/issues/752

PR #14462, opened 2025-09-05
ZJIT, YJIT: Make the workflow names consistent with file names

k0kubun:
This PR makes the workflow names of `yjit-macos.yml` and `zjit-macos.yml` consistent with their file names like other workflows.

It's true that part of the point of having `*jit-macos.yml` is to test the arm64 backend, but it feels weird that only one of the architectures is named in the workflows, and the workflow names are so long that it makes it hard to read the details of the job in some UI.

### DIFF
--- a/.github/workflows/yjit-macos.yml
+++ b/.github/workflows/yjit-macos.yml
@@ -1,4 +1,4 @@
-name: YJIT macOS Arm64
+name: YJIT macOS
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


k0kubun:
For this rename, I will:

1. Temporarily remove "YJIT macOS Arm64" from required status checks
2. Merge this
3. Wait for 1+ days for people to catch up with the rename on active PRs they're working on
5. Add "YJIT macOS" to required status checks

Now that ZJIT CI is more stabilized than before, I'll add ZJIT jobs to required status checks after that as well.

k0kubun:
Resurrected "YJIT macOS" as a required status check now. 

I think ZJIT jobs has become more unstable than before, so I don't think it's the time to add them as required status checks yet. I'll revisit it after we fix https://github.com/Shopify/ruby/issues/733.

PR #14463, opened 2025-09-06
Make invalid & operator type error message consistent with */**

jeremyevans:
If #to_proc is defined, this uses the following error message format, matching the error message format used for * when to_a returns non-Array and for ** when to_hash returns non-Hash:

```
can't convert ClassName to Proc (ClassName#to_proc gives OtherClassName)
```

If #to_proc is not defined, this uses the following error message format, matching the error message format used when ** is called on a non-Hash not implementing to_hash.

```
no implicit conversion of ClassName into Proc
```

There isn't a similar error for * when called on a non-Array not implementing to_a, as Ruby does not raise for that case.

Fixes [Bug #21563]

### DIFF
--- a/test/ruby/test_exception.rb
+++ b/test/ruby/test_exception.rb
@@ -1525,4 +1525,31 @@ $stderr = $stdout; raise "\x82\xa0"') do |outs, errs, status|
       assert_in_out_err(%W[-r#{lib} #{main}], "", [], [:*, "\n""path=#{main}\n", :*])
     end
   end
+
+  class Ex; end
+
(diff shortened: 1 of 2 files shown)


byroot (approved):
Looks good, but it would be nice to add a spec for it. I can do it if you wish.

jeremyevans:
> Looks good, but it would be nice to add a spec for it. I can do it if you wish.

I added a test for it that shows the consistency for */**/&.

soulcutter:
Thanks so much for your work on this!

PR #14469, opened 2025-09-08
ZJIT: Optimize `ObjToString` with type guards

altxtech:
### Context

Closes #Shopify/ruby/issues/701

Work in progress

### Changes

- Profile the receiver for `ObjToString`
- For profiled strings, rewrite to `GuardType(String)` (AnyToString, also folds)
- Added new HIR `GuardTypeNot`
- For profiled nonstrings, Guard against strings and do a direct send with `SendWithoutBlock`


zjit/src/codegen.rs
```diff
@@ -1325,6 +1326,10 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         // for a known VALUE, which with_num_bits() does not support.
         asm.cmp(val.with_num_bits(8), Opnd::UImm(RUBY_SYMBOL_FLAG as u64));
         asm.jne(side_exit(jit, state, GuardType(guard_type)));
+    } else if guard_type.is_subtype(types::String) {
+        let tag = asm.and(val, Opnd::UImm(RUBY_T_MASK as u64));
```
altxtech: I do not know if this is the correct way to check if `val` is a `String`. Repeated the same pattern here on ` gen_guard_type_not`.

Experimented with other ways of doing it, which often blew up tests. This makes the pipeline green, but I'm not confident it is right.

altxtech:
We got one failed check in the pipeline, but I think it was a networking hiccup.

zjit/src/codegen.rs
k0kubun: You're masking the object pointer, but `RUBY_T_MASK` is supposed to be used on object flags.

You'd need to do something like:

```rust
let flags = asm.load(Opnd::mem(VALUE_BITS, val, RUBY_OFFSET_RBASIC_FLAGS));
let tag = asm.and(flags, Opnd::UImm(RUBY_T_MASK as u64));
```

zjit/src/hir.rs
```diff
@@ -1780,12 +1787,26 @@ impl Function {
...
+                            self.make_equal_to(insn_id, val); continue;
+                        }
+
+                        let frame_state = self.frame_state(state);
+                        let Some(recv_type) = self.profiled_type_of_at(val, frame_state.insn_idx) else {
+                            self.push_insn_id(block, insn_id); continue
+                        };
+
+                        if recv_type.is_string() {
+                            let guard = self.push_insn(block, Insn::GuardType { val: val, guard_type: types::String, state: state });
```
k0kubun, line 1813: Here, we use `Insn::GuardType { val, guard_type: String }` for checking if it's a `T_STRING` object (`String` or its subclass), but I think we currently use `Insn::GuardType` for checking if `val` is "exactly" the class of `guard_type` (not a subclass of it) because it cannot be used for caching a method lookup otherwise.

So I think HIR needs to have a way to communicate whether it wants exactly a `String` or a `T_STRING`. How about adding `GuardSubtype` and `GuardSubtypeNot` (or `GuardNotSubtype`), and letting them own `T_*` kinds of checks? Since we want the `T_STRING` check, we probably don't need `GuardTypeNot` for now.

test/ruby/test_zjit.rb
```diff
@@ -2011,6 +2011,48 @@ def test(str)
...
+  end
+
+  def test_objtostring_profiled_nonstring_guard_exits_when_string
+    assert_compiles '"foo"', %q{
+      def test(str)
+        "#{str}"
+      end
+
+      test([1,2,3]); # profiles as nonstring
+      test('foo');
```
k0kubun, line 2142:
```suggestion
      test([1,2,3]) # profile as nonstring
      test([1,2,3])
    }, call_threshold: 2
  end

  def test_objtostring_profiled_nonstring_guard_exits_when_string
    assert_compiles '"foo"', %q{
      def test(str)
        "#{str}"
      end

      test([1,2,3]) # profiles as nonstring
      test('foo')
```

k0kubun:
The HIR looks good :+1:

zjit/src/hir.rs
altxtech: In  `hir_type`, we have:

```rust
//...
  pub const String: u64 = StringExact | StringSubclass;
  pub const StringExact: u64 = 1u64 << 38;
  pub const StringSubclass: u64 = 1u64 << 39;
//...
```

So I thought  `GuardType` could be used in both cases, and whether you accept subclasses or only the exact class is determined by type you guard:

`GuardType(String)` --> String or subclasses  
`GuardType(StringExact)` --> Exactly a string
`GuardType(StringSubclass)` --> Subclasses of string, but not string

altxtech: I guess, we could add an HIR test for that? If the optimization triggers with a subclass of String, then the behavior is correct, at least.

zjit/src/codegen.rs
```diff
@@ -1325,6 +1326,10 @@ fn gen_guard_type(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd, guard
         // for a known VALUE, which with_num_bits() does not support.
         asm.cmp(val.with_num_bits(8), Opnd::UImm(RUBY_SYMBOL_FLAG as u64));
         asm.jne(side_exit(jit, state, GuardType(guard_type)));
+    } else if guard_type.is_subtype(types::String) {
```
XrXr: This puts an overly loose runtime check for `types::StringExact`, which is a subtype of `types::String`. This if-else is ordered from most specific to least, and so this should go before the heap object branch.

zjit/src/hir.rs
XrXr, line 1813: `recv_type.is_string()` currently checks for `types::StringExact` but we should be checking for `types::String` here, to match up with the type-bound on `GuardType`.

k0kubun: > thought GuardType could be used in both cases
> GuardType(StringExact) --> Exactly a string

Hmm right, that sounds true. And `Type::from_profiled_type` gives `types::StringExact` for `rb_cString`, so I guess it's already thought out. HIR probably doesn't need to change.

k0kubun: I guess https://github.com/ruby/ruby/pull/14469/files#r2330915176 is the point. The branch should not handle `types::StringExact`. As long as it's fixed, it should be fine.

zjit/src/codegen.rs
altxtech: Thanks for the help.

Fixed it. The solution is pretty much that, but we also need to check for some edge cases first, like false, special constants, etc.

zjit/src/codegen.rs
altxtech: Fixed it.

zjit/src/hir.rs
altxtech: Fixed it. 

Also added the following test to verify the optimization also triggers for subclasses of string.

```rust
    #[test]
    fn test_optimize_objtostring_anytostring_recv_profiled_string_subclass() {
        eval("
            class MyString < String; end

            def test(a)
              \"#{a}\"
            end
            foo = MyString.new('foo')
            test(MyString.new(foo)); test(MyString.new(foo))
        ");

        fn test@<compiled>:5:
          v5:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
          v17:String = GuardType v1, String
          v11:StringExact = StringConcat v5, v17
          CheckInterrupts
          Return v11
        ");
    }
```

k0kubun (approved):

PR #14470, opened 2025-09-08
Move `IS_TYPED_DATA` in RBasic.flags

byroot:
Ref: https://github.com/ruby/ruby/pull/14134#issuecomment-3207733725

We can't safely use low-bit pointer tagging anymore because `RTypedData.type` lines up with `RData.dfree` and there is no aligment guarantee on function pointers, as evidenced by `memcached` and `gpgme` gems.

We also can't use FL_USER* for this, because extensions may use these for other purposes.

Using a general flag for this is a bit unfortunate, as general flags are hard to come by, however I recently freed several of them, and we still have two or three free ones left.


composerinteralia (commented):
You beat me to it. Thanks!

I had a branch just like this one and confirmed it fixed everything for GitHub. 👍🏻

nobu:
This broke CIs build incrementally instead of full build each time.
Probably the ABI version needs to be bumped up?

nobu:
Conversely, we may be able to check binary compatibility by using the cached CAPI extension binaries.

byroot:
> Probably the ABI version needs to be bumped up?

I didn't expect this to change the ABI, but it's possible. I'll bump it on master.

PR #14474, opened 2025-09-08
[DOC] Tweaks for String#lstrip

BurdetteLamar:


string.c
```diff
@@ -10393,16 +10393,18 @@ rb_str_lstrip_bang(VALUE str)
 /*
  *  call-seq:
  *    lstrip -> new_string
+
```
peterzhu2118: Was this blank line committed by accident?

BurdetteLamar: Removed.  (Note to self:  I always look at the HTML;  need also to look at the diff.)

PR #14476, opened 2025-09-08
[DOC] Tweaks for String#match

BurdetteLamar:


string.c
```diff
@@ -5067,34 +5067,37 @@ static VALUE get_pat(VALUE);
...
- *  If Integer argument +offset+ is given, the search begins at index +offset+:
+ *  If integer argument +offset+ is given, the search begins at index +offset+:
  *
  *    'foo'.match('f', 1) # => nil
  *    'foo'.match('o', 1) # => #<MatchData "o">
  *
- *  With a block given, calls the block with the computed +matchdata+
- *  and returns the block's return value:
+ *  With a block given, calls the block with the computed +matchdata+ or +nil+;
+ *  returns the block's return value:
```
peterzhu2118: The block is not called if `matchdata` is `nil`, for example:

```ruby
'foo'.match(/x/) {|matchdata| raise } # does not raise
```

string.c
```diff
@@ -5067,34 +5067,38 @@ static VALUE get_pat(VALUE);
...
  *    'foo'.match('o', 1) # => #<MatchData "o">
  *
- *  With a block given, calls the block with the computed +matchdata+
- *  and returns the block's return value:
+ *  With a block given and computed +matchdata+ non-nil, calls the block with +matchdata+;
+ *  returns the block's return value:
  *
  *    'foo'.match(/o/) {|matchdata| matchdata } # => #<MatchData "o">
  *    'foo'.match(/x/) {|matchdata| matchdata } # => nil
  *    'foo'.match(/f/, 1) {|matchdata| matchdata } # => nil
```
peterzhu2118: These examples still show cases where the block isn't called.

BurdetteLamar: Removed.

PR #14477, opened 2025-09-08
[DOC] Tweaks for String#match?

BurdetteLamar:


string.c
```diff
@@ -5116,24 +5116,26 @@ rb_str_match_m(int argc, VALUE *argv, VALUE str)
...
+ *  does not update {Regexp Global Variables}[rdoc-ref:Regexp@Global+Variables].
  *
- *  Note: does not update Regexp@Global+Variables.
+ *  Computes +regexp+ by converting +pattern+ (if not already a Regexp):
  *
- *  Computes +regexp+ by converting +pattern+ (if not already a Regexp).
  *    regexp = Regexp.new(pattern)
  *
- *  Returns +true+ if <tt>self+.match(regexp)</tt> returns a MatchData object,
+ *  Returns +true+ if <tt>self[offset].match(regexp)</tt> returns a MatchData object,
```
peterzhu2118: `self[offset]` just returns the character at `offset`. I think what you mean is maybe something like `self[offset..]`?

If that's the case, then we don't need to consider the case where `offset` is given vs. not given, since `offset` is 0 by default.

PR #14479, opened 2025-09-08
ZJIT: Add RubyVM::ZJIT.reset_stats! method

rwstauner:
This allows for more precise tracking of stats programmatically
which is particularly useful for our nightly benchmarking suite.

- Define rust function
- Expose to C
- Wrap with Ruby API
- Add a test

### DIFF
--- a/zjit/src/stats.rs
+++ b/zjit/src/stats.rs
@@ -247,9 +247,7 @@ pub extern "C" fn rb_zjit_reset_stats_bang(_ec: EcPtr, _self: VALUE) -> VALUE {
     *counters = Counters::default();
 
     // Reset exit counters for YARV instructions
-    for i in 0..(VM_INSTRUCTION_SIZE as usize) {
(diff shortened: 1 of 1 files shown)


zjit/src/stats.rs
```diff
@@ -237,6 +237,23 @@ pub fn exit_counter_ptr(reason: crate::hir::SideExitReason) -> *mut u64 {
...
+    let counters = ZJITState::get_counters();
+    let exit_counters = ZJITState::get_exit_counters();
+
+    // Reset all counters to zero
+    *counters = Counters::default();
+
+    // Reset exit counters for YARV instructions
+    for i in 0..(VM_INSTRUCTION_SIZE as usize) {
+        exit_counters[i] = 0;
+    }
```
XrXr:
```suggestion
    exit_counters.as_mut_slice().fill(0);
```

rwstauner: Thank you

PR #14483, opened 2025-09-08
ZJIT: Implement getblockparamproxy

k0kubun:
This PR implements `getblockparamproxy` instruction.

Many of previous exits on `getblockparamproxy` just turn into `unhandled_call_block_arg` now, but it still improves `ratio_in_zjit` on lobsters from 65.9% to 66.3% by not failing to compile the entire ISEQ.


zjit/src/codegen.rs
```diff
@@ -548,6 +549,29 @@ fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, val_type: Type, local_ep
...
+    // So to check for either of those cases we can use: val & 0x1 == 0x1
+    const _: () = assert!(RUBY_SYMBOL_FLAG & 1 == 0, "guard below rejects symbol block handlers");
+
+    // Bail ouf if the block handler is neither ISEQ nor ifunc
+    let block_handler = asm.load(Opnd::mem(64, ep, SIZEOF_VALUE_I32 * VM_ENV_DATA_INDEX_SPECVAL));
+    asm.test(block_handler, 0x1.into());
+    asm.jz(side_exit(jit, state, SideExitReason::BlockParamProxyNotIseqOrIfunc));
+
+    // Return the rb_block_param_proxy instance
+    Opnd::const_ptr(unsafe { rb_block_param_proxy.0 } as *const u8)
```
```suggestion
    // Return the rb_block_param_proxy instance (GC root, so put as a number to avoid unnecessary GC tracing)
    unsafe { rb_block_param_proxy }.as_u64().into()
```

XrXr (approved):

zjit/src/codegen.rs
```diff
@@ -398,6 +398,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::GetGlobal { id, state } => gen_getglobal(jit, asm, *id, &function.frame_state(*state)),
         &Insn::GetLocal { ep_offset, level } => gen_getlocal_with_ep(asm, ep_offset, level),
         &Insn::SetLocal { val, ep_offset, level } => no_output!(gen_setlocal_with_ep(asm, opnd!(val), function.type_of(val), ep_offset, level)),
+        &Insn::GetBlockParamProxy { level, state } => gen_get_block_param_proxy(jit, asm, level, &function.frame_state(state)),
```
tekknolagi, line 401: Can we split this into two guard instructions and a Const? Just like the VMFrozenCore or whatever it's called

k0kubun: We'd still need a new instruction to check the frame flag, but yeah we could carve out the Const part. Is that what you wanted?

tekknolagi: Yes please. I may tackle on plane else can file an issue tomorrow

k0kubun: I filed https://github.com/ruby/ruby/pull/14583, but it makes it hard to maintain https://github.com/ruby/ruby/pull/14483#discussion_r2334134019.

If we just make it `Const::CUint64`, the HIR tests would constantly generate a diff like:

```
        6       │-  v17:CUInt64[135329854491400] = Const CUInt64(135329854491400)
              6 │+  v17:CUInt64[132207273872120] = Const CUInt64(132207273872120)
```

Do we want an u64 printer to print a stable output for `rb_block_param_proxy`, or a flag in `Insn::Const` or `Const::Value` to lower it to `Opnd::UImm` instead of `Opnd::Value`?

k0kubun: Also, if we make it "two guard instructions", you'd also need another instruction to get EP if you want to share the EP reference among those two guards to avoid fetching it twice. Do you really want to add three new instructions just for `getblockparamproxy`?

I feel like we don't really need to split the guard until we have other instructions that reuse part of it or add a pass that optimizes the split guards, although it does make it easier to understand what's being guarded from reading the HIR.

tekknolagi: seems great as-is

tekknolagi: We can have a flag on `Const` to mark something as non-rooted later, if we need to

PR #14486, opened 2025-09-09
ZJIT: Add --zjit-mem-size

aidenfoxivey:
Closes https://github.com/Shopify/ruby/issues/686

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -266,6 +266,7 @@ MAKE_LINK = $(MINIRUBY) -rfileutils -e "include FileUtils::Verbose" \
 YJIT_RUSTC_ARGS = --crate-name=yjit \
 	--crate-type=staticlib \
 	--edition=2021 \
+	--cfg 'feature="stats_allocator"' \
 	-g \
 	-C lto=thin \
 	-C opt-level=3 \
@@ -276,6 +277,7 @@ YJIT_RUSTC_ARGS = --crate-name=yjit \
 ZJIT_RUSTC_ARGS = --crate-name=zjit \
 	--crate-type=staticlib \
 	--edition=2024 \
+	--cfg 'feature="stats_allocator"' \
 	-g \
 	-C lto=thin \
 	-C opt-level=3 \
--- a/zjit/src/options.rs
+++ b/zjit/src/options.rs
@@ -35,6 +35,10 @@ pub struct Options {
     /// Note that the command line argument is expressed in MiB and not bytes.
     pub exec_mem_bytes: usize,
 
+    /// Hard limit of ZJIT's total memory usage.
+    /// Note that the command line argument is expressed in MiB and not bytes.
+    pub mem_bytes: usize,
+
     /// Number of times YARV instructions should be profiled.
     pub num_profiles: u8,
 
@@ -79,6 +83,7 @@ impl Default for Options {
     fn default() -> Self {
         Options {
             exec_mem_bytes: 64 * 1024 * 1024,
+            mem_bytes: 128 * 1024 * 1024,
             num_profiles: DEFAULT_NUM_PROFILES,
             stats: false,
             print_stats: false,
@@ -100,9 +105,8 @@ impl Default for Options {
 /// Note that --help allows only 80 chars per line, including indentation, and it also puts the
 /// description in a separate line if the option name is too long.  80-char limit --> | (any character beyond this `|` column fails the test)
 pub const ZJIT_OPTIONS: &[(&str, &str)] = &[
-    // TODO: Hide --zjit-exec-mem-size from ZJIT_OPTIONS once we add --zjit-mem-size (Shopify/ruby#686)
-    ("--zjit-exec-mem-size=num",
-                     "Size of executable memory block in MiB (default: 64)."),
+    ("--zjit-mem-size=num",
+                     "Max amount of memory that ZJIT can use (in MiB)."),
     ("--zjit-call-threshold=num",
                      "Number of calls to trigger JIT (default: 2)."),
     ("--zjit-num-profiles=num",
--- a/yjit/Cargo.toml
+++ b/yjit/Cargo.toml
@@ -25,3 +25,4 @@ disasm = ["capstone"]
 # from cfg!(debug_assertions) so that we can see disasm of the code
 # that would run in the release mode.
 runtime_checks = []
+stats_allocator = []
--- a/yjit/src/stats.rs
+++ b/yjit/src/stats.rs
@@ -1,9 +1,8 @@
 //! Everything related to the collection of runtime stats in YJIT
 //! See the --yjit-stats command-line option
 
-use std::alloc::{GlobalAlloc, Layout, System};
 use std::ptr::addr_of_mut;
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::Ordering;
 use std::time::Instant;
 use std::collections::HashMap;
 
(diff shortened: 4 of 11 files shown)


k0kubun:
Could you fix https://github.com/ruby/ruby/blob/0dbaf7e981081eee5665a4144b500efd6da4208b/zjit/src/state.rs#L88 as well?

common.mk
```diff
@@ -262,10 +262,34 @@ MAKE_LINK = $(MINIRUBY) -rfileutils -e "include FileUtils::Verbose" \
 	  -e "noraise {ln(src, dest)} or" \
 	  -e "cp(src, dest)"
 
+
```
k0kubun: Your PRs tend to include repeated empty lines, but we don't want that 🙂 

```suggestion
```

k0kubun (approved):
I like this diff (because it's what I asked), but I'd appreciate @XrXr's eyes before moving forward.

aidenfoxivey:
> Could you fix
> 
> https://github.com/ruby/ruby/blob/0dbaf7e981081eee5665a4144b500efd6da4208b/zjit/src/state.rs#L88
> 
> as well?

Fixed!

zjit/src/options.rs
```diff
@@ -100,9 +105,6 @@ impl Default for Options {
 /// Note that --help allows only 80 chars per line, including indentation, and it also puts the
 /// description in a separate line if the option name is too long.  80-char limit --> | (any character beyond this `|` column fails the test)
 pub const ZJIT_OPTIONS: &[(&str, &str)] = &[
-    // TODO: Hide --zjit-exec-mem-size from ZJIT_OPTIONS once we add --zjit-mem-size (Shopify/ruby#686)
-    ("--zjit-exec-mem-size=num",
-                     "Size of executable memory block in MiB (default: 64)."),
```
k0kubun, line 105: :+1: but let's add an entry for "--zjit-mem-size=num" too. It's probably the most useful option for end-users that should stay at the top of `ZJIT_OPTIONS`.

common.mk
aidenfoxivey: Ah sorry about that! I've noticed my git tooling doesn't seem to flag those. I'll more carefully examine them in the future.

zjit/src/options.rs
aidenfoxivey: Done.

aidenfoxivey:
Does the `parse.y` issue seem like a flake? I can't think of anything I changed that would have caused it.

k0kubun:
It is. We'll deal with it separately.

common.mk
```diff
@@ -262,10 +262,33 @@ MAKE_LINK = $(MINIRUBY) -rfileutils -e "include FileUtils::Verbose" \
...
+	$(ECHO) 'building Rust JIT library'
+	+$(Q) $(RUSTC) $(JIT_RUSTC_ARGS)
+	$(JIT_LIB_TOUCH)
+
 # For release builds
 YJIT_RUSTC_ARGS = --crate-name=yjit \
 	--crate-type=staticlib \
 	--edition=2021 \
+	--extern jit=$(CARGO_TARGET_DIR)/release/libjit.rlib \
+	-L dependency=$(CARGO_TARGET_DIR)/release \
```
XrXr: As mentioned over chat, I'd like a solution that doesn't involve a separate crate for release rustc-only builds.

aidenfoxivey: Hmm, okay, admittedly I'm a _little_ less sure on how to go about that, but I will look into it.

aidenfoxivey: So the preference is to not have a separate crate at all right? Presumably then I'll just write a file called `jit.rs` and then write something in the makefile to compile it to an rlib and link it.

XrXr: It's fine to have the crate, and cargo builds will use it as you currently have it. Release build just won't consume it as an rlib. You should be able to add 

```
#[cfg(feature = "stats_allocator")]
#[path = "...."]
mod jit;
```

to stats.rs and with a rustc flag enabling `stats_allocator` and it should build fine. Cargo builds won't enable this feature since they consume the code as a crate.

aidenfoxivey: Hmm okay, I've given that a try, I'm having a bit of trouble getting it to not fail CI. I will take a look as to where it's failing.

aidenfoxivey: It's mostly confusing because I see the cfg declaration exists, the feature is defined in the `Cargo.toml`, but yet passing the flag doesn't seem to do anything.

aidenfoxivey: Okay, I think I may have just been writing the flag incorrectly.

aidenfoxivey: Looks like it's working!

k0kubun:
I wonder if we can let CI test builds in environments that don't have `cargo`, assuming every job has `cargo` at the moment.

For the `RUSTC='rustc +1.58.0'` job in yjit-ubuntu and `rust_version: '1.85.0'` job in zjit-ubuntu (removing `--enable-yjit=dev` or `=dev` in `--enable-zjit=dev`), is it possible to remove `cargo` from the environment? If such jobs pass, I think we can confidently say we're not changing build dependencies.

aidenfoxivey:
> I wonder if we can let CI test builds in environments that don't have `cargo`, assuming every job has `cargo` at the moment.
> 
> For the `RUSTC='rustc +1.58.0'` job in yjit-ubuntu and `rust_version: '1.85.0'` job in zjit-ubuntu (removing `--enable-yjit=dev` or `=dev` in `--enable-zjit=dev`), is it possible to remove `cargo` from the environment? If such jobs pass, I think we can confidently say we're not changing build dependencies.

Let me see if this is a thing. I would gather we can specify just `rustc`, but I wonder if `rustc` implicitly depends on having Cargo.

aidenfoxivey:
> > I wonder if we can let CI test builds in environments that don't have `cargo`, assuming every job has `cargo` at the moment.
> > For the `RUSTC='rustc +1.58.0'` job in yjit-ubuntu and `rust_version: '1.85.0'` job in zjit-ubuntu (removing `--enable-yjit=dev` or `=dev` in `--enable-zjit=dev`), is it possible to remove `cargo` from the environment? If such jobs pass, I think we can confidently say we're not changing build dependencies.
> 
> Let me see if this is a thing. I would gather we can specify just `rustc`, but I wonder if `rustc` implicitly depends on having Cargo.

My guess is that we'd need to use Docker, since the default Github actions runner image contains `cargo`.

k0kubun:
What if you just `sudo rm` it?

aidenfoxivey:
> What if you just `sudo rm` it?

Let me give that a try!

.github/workflows/yjit-ubuntu.yml
```diff
@@ -145,6 +145,10 @@ jobs:
         if: ${{ matrix.rust_version }}
         run: rustup install ${{ matrix.rust_version }} --profile minimal
 
+      - name: Remove cargo
+        if: ${{ matrix.rust_version }}
```
XrXr: Could you change this condition to search for `rustc` in configure instead? It's not obvious why specifying a rust version entails deleting cargo.

common.mk
```diff
@@ -276,6 +298,8 @@ YJIT_RUSTC_ARGS = --crate-name=yjit \
 ZJIT_RUSTC_ARGS = --crate-name=zjit \
 	--crate-type=staticlib \
 	--edition=2024 \
+	--extern jit=$(CARGO_TARGET_DIR)/release/libjit.rlib \
+	-L dependency=$(CARGO_TARGET_DIR)/release \
```
XrXr: Let's apply the same trick to ZJIT and remove all the stuff related to BUILT_JIT_LIB.

aidenfoxivey: Certainly! Will do.

.github/workflows/yjit-ubuntu.yml
aidenfoxivey: Sure, that works for me.

aidenfoxivey:
I think this is a flake test. One quick q - would it be preferred for me to clean up my commit history before merge?

XrXr:
Yes, a nice commit log would be nice. Too many happening for squashing to one commit.

aidenfoxivey:
> Yes, a nice commit log would be nice. Too many happening for squashing to one commit.

Should be fixed!

PR #14490, opened 2025-09-09
ZJIT: Remove unused `self_val` from HIR `Send` and `InvokeSuper` HIRs

st0012:
It's not causing any issues but IMO if we're not using a HIR field (in codegen, optimization...etc.) then it's better to avoid having it for now to avoid confusion.


k0kubun (approved):
Could you resolve the conflict please?

PR #14494, opened 2025-09-10
Improve performance of UnicodeNormalize.canonical_ordering_one

tompng:
This is what canonical_ordering_one does:
```
These numbers are ccc(combining character class) of each char
0 3 1 2 0 2 0 0 1 2 2 1 0 3 2 0 3 1 3
↓ Group ccc != 0 chars
0, [3, 1, 2], 0, [2], 0, 0, [1, 2, 2, 1], 0, [3, 2], 0, [3, 1, 3]
↓ Stable sort
0, [1, 2, 3], 0, [2], 0, 0, [1, 1, 2, 2], 0, [2, 3], 0, [1, 3, 3]
↓ Concat
0 1 2 3 0 2 0 0 1 1 2 2 0 2 3 0 1 3 3
```
Change canonical ordering from modified-bubble-sort to simply do the above with `Array#sort!`

Faster than before in every cases.
```ruby
canonical_ordering_one("\u0300") # 20% faster
canonical_ordering_one("s\u0323") # 40% faster
canonical_ordering_one("s\u0307\u0323") # 60% faster
canonical_ordering_one("\u03b1\u0345\u0313\u0300") # 85% faster
```

Detail of stable sort by combining character classes

- Sort an array of `ccc * chars.size + index` gets the same result as `sort_by { [ccc, index] }` with less time and object allocation.
- `sorted_numbers.map { index = it % chars.size;  chars[index] }` is the sorted result

### DIFF
--- a/lib/unicode_normalize/normalize.rb
+++ b/lib/unicode_normalize/normalize.rb
@@ -82,16 +82,22 @@ module UnicodeNormalize  # :nodoc:
 
   ## Canonical Ordering
   def self.canonical_ordering_one(string)
-    sorting = string.each_char.collect { |c| [c, CLASS_TABLE[c]] }
-    (sorting.length-2).downto(0) do |i| # almost, but not exactly bubble sort
-      (0..i).each do |j|
-        later_class = sorting[j+1].last
-        if 0<later_class and later_class<sorting[j].last
-          sorting[j], sorting[j+1] = sorting[j+1], sorting[j]
-        end
+    result = ''
+    unordered = []
+    chars = string.chars
+    n = chars.size
+    chars.each_with_index do |char, i|
+      ccc = CLASS_TABLE[char]
+      if ccc == 0
+        unordered.sort!.each { result << chars[it % n] }
+        unordered.clear
+        result << char
+      else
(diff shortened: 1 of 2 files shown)


duerst:
I will look at this after looking at https://bugs.ruby-lang.org/issues/21559.

duerst:
@tompng: I wonder how you measured the speedup (20%~85%). My guess is that it was achieved on a busy loop just calling `canonical_ordering_one`. Getting such speedups isn't a bad thing, but this method, as well as the other methods ending in _one, are not significant contributors to overall normalization speed in practice.

It would be interesting to know how much speedup one gets e.g. when running `test/test_unicode_normalize.rb`. This is still quite far away from some realistic text, but at least it only calls `canonical_ordering_one` only once for each run it normalizes.

Even more realistic would be to check speedup on actual texts, such as https://github.com/duerst/eprun/tree/master/benchmark.

Also, if we really want to concentrate on speeding up `canonical_ordering_one`, special-casing two-character strings where the first character has a CCC of 0 might be worth looking at, because in practice, these are most frequent and don't need any work. But as said above, this method isn't on the hot path in practice.

PR #14496, opened 2025-09-10
ZJIT: Support `invokeblock` insn

st0012:
Basically like other send-ish insns, except that it doesn't pop the receiver.


k0kubun (approved):
:+1:

PR #14501, opened 2025-09-10
ZJIT: Compile sendforward with dynamic dispatch

k0kubun:


zjit/src/hir.rs
```diff
@@ -3783,6 +3800,33 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                    }
+                    let argc = unsafe { vm_ci_argc((*cd).ci) };
+
+                    let args = state.stack_pop_n(argc as usize + usize::from(forwarding))?;
+                    let recv = state.stack_pop()?;
+                    let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
+                    let send_forward = fun.push_insn(block, Insn::SendForward { recv, cd, blockiseq, args, state: exit_id });
+                    state.stack_push(send_forward);
+
+                    // Reload locals that may have been modified by the blockiseq.
```
tekknolagi: Can we gate this on having a blockiseq? Or can that come in through the forwarding arguments somehow?

zjit/src/codegen.rs
```diff
@@ -1041,6 +1042,29 @@ fn gen_send(
...
+
+    gen_prepare_non_leaf_call(jit, asm, state);
+
+    asm_comment!(asm, "call #{} with dynamic dispatch", ruby_call_method_name(cd));
+    unsafe extern "C" {
+        fn rb_vm_sendforward(ec: EcPtr, cfp: CfpPtr, cd: VALUE, blockiseq: IseqPtr) -> VALUE;
+    }
+    asm.ccall(
+        rb_vm_sendforward as *const u8,
+        vec![EC, CFP, (cd as usize).into(), VALUE(blockiseq as usize).into()],
```
tekknolagi, line 1064: What are the `as usize` casts for?

zjit/src/hir.rs
```diff
@@ -5457,14 +5501,17 @@ mod tests {
     }
 
     #[test]
-    fn test_cant_compile_forwarding() {
+    fn test_compile_forwarding() {
         eval("
             def test(...) = foo(...)
```
tekknolagi, line 5509: Can you test with a prefix positional too?

zjit/src/codegen.rs
k0kubun: You get this without it.

```
error[E0277]: the trait bound `lir::Opnd: From<*const autogened::rb_call_data>` is not satisfied
    --> zjit/src/codegen.rs:1064:28
     |
1064 |         vec![EC, CFP, (cd).into(), VALUE(blockiseq).into()],
     |                            ^^^^ the trait `From<*const autogened::rb_call_data>` is not implemented for `lir::Opnd`
     |
     = help: the following other types implement trait `From<T>`:
               `lir::Opnd` implements `From<cruby::VALUE>`
               `lir::Opnd` implements `From<i32>`
               `lir::Opnd` implements `From<i64>`
               `lir::Opnd` implements `From<u32>`
               `lir::Opnd` implements `From<u64>`
               `lir::Opnd` implements `From<usize>`
     = note: required for `*const autogened::rb_call_data` to implement `Into<lir::Opnd>`
```

 `something_usize.into()` is possible but we don't necessarily have `pointer_of_random_type.into()`.

zjit/src/hir.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/14501/commits/961bfcece20b8fc91a308eb238fe2c9aa0dc9517

zjit/src/hir.rs
k0kubun: added the test case https://github.com/ruby/ruby/pull/14501/commits/b11c87c76ec705f03dc262f243c7d45d2986b5fa. It's no longer a forwarding call though.

PR #14508, opened 2025-09-10
YJIT: Print more disassembly in release builds

XrXr:
- **YJIT: Print more disassembly in release builds**
- **YJIT: Remove dead code: `asm_comment!` checks `--yjit-dump-disasm`**


yjit/src/codegen.rs
```diff
@@ -1096,8 +1096,6 @@ pub fn gen_entry_prologue(
     let mut asm = Assembler::new(unsafe { get_iseq_body_local_table_size(iseq) });
     if get_option_ref!(dump_disasm).is_some() {
```
k0kubun: I guess we don't need this `if` guard either. Just `asm_comment!(asm, "YJIT entry point: {}", iseq_get_location(iseq, 0));` here seems enough.

k0kubun (approved):

k0kubun:
Let's backport this to `ruby_3_4` as well 🙏

PR #14520, opened 2025-09-12
ZJIT: Share more code with YJIT in jit.c

k0kubun:
This PR moves some functions from `yjit.c` and `zjit.c` to `jit.c` to share them.

They are mostly just copied from YJIT, so there's no need to separately define them, which would unnecessarily increase the binary size on combo builds. Also, leaving only ZJIT-specific things in `zjit.c` makes it easier to understand whether it's really about ZJIT or not.

### DIFF
--- a/jit.c
+++ b/jit.c
@@ -538,16 +538,16 @@ rb_jit_get_page_size(void)
 {
 #if defined(_SC_PAGESIZE)
     long page_size = sysconf(_SC_PAGESIZE);
-    if (page_size <= 0) rb_bug("zjit: failed to get page size");
+    if (page_size <= 0) rb_bug("jit: failed to get page size");
 
     // 1 GiB limit. x86 CPUs with PDPE1GB can do this and anything larger is unexpected.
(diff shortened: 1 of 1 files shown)


jit.c
```diff
@@ -533,10 +533,184 @@ for_each_iseq_i(void *vstart, void *vend, size_t stride, void *data)
     return 0;
 }
 
+uint32_t
+rb_jit_get_page_size(void)
+{
+#if defined(_SC_PAGESIZE)
+    long page_size = sysconf(_SC_PAGESIZE);
+    if (page_size <= 0) rb_bug("zjit: failed to get page size");
```
XrXr:
```suggestion
    if (page_size <= 0) rb_bug("jit: failed to get page size");
```

Also need to remove references to ZJIT everywhere in the file.

k0kubun: thanks :+1: https://github.com/ruby/ruby/pull/14520/commits/9cda52e81a2759681f5ff92de0f0fb0fbfd45415

PR #14522, opened 2025-09-12
[3.4 backport] Reject some cases with `return` and command calls

Earlopain:
Backports https://github.com/ruby/prism/pull/3630


k0kubun:
I think now we want to stop doing this way of backports for Prism (like https://github.com/ruby/ruby/pull/14358) actually. We'd want to cut a prism.gem release, sync a whole version to this branch, and merge it to `ruby_3_4`.

Even if we backport things this way, we still need to have a branch in the Prism repository to release what `ruby_3_4` has separately, so it doesn't change the fact that we need a Prism release. If so, it's probably the best to sync Prism to the `main` branch-based version, as long as it has a flag to maintain its syntax compatibility for the Ruby version.

Earlopain:
Yes, I'am aware. If there is a prism release this is not needed.

Earlopain:
BTW, here is a PR to handle multi-version when backporting by version: https://github.com/ruby/ruby/pull/14523

There's still nothing in prism that will affect this at the moment, but at the latest for 3.5 something like that will be needed.

Earlopain:
Very quickly became not needed anymore

PR #14523, opened 2025-09-12
Explicitly use a ruby version for prism to parse the code as

Earlopain:
Prism can parse multiple versions of ruby. Because of that branch release managers are ok with simply bumping prism to its latest version.

However, if no version is specified, it will parse as the latest known version, which can be ahead of the maintenance branch. So we need to explicitly pass a version to not accidentally introduce new syntax to maintenance branches.

Since prism and ruby are developed independently from each other, the version can't be derived from `RUBY_VERSION`, since by the time that the next ruby version is bumped, prism does not yet know about that new ruby version. So I check for development versions and simply use LATEST there.

I believe this is what you were talking about in https://github.com/ruby/ruby/pull/14358#issuecomment-3229656013 @xrxr. For now, there is still no 3.5-specific syntax.

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -1,4 +1,5 @@
 #include "prism.h"
+#include "version.h"
 
 /**
  * This compiler defines its own concept of the location of a node. We do this
@@ -11352,6 +11353,8 @@ pm_parse_file(pm_parse_result_t *result, VALUE filepath, VALUE *script_lines)
     pm_options_filepath_set(&result->options, RSTRING_PTR(filepath));
     RB_GC_GUARD(filepath);
 
+    pm_options_version_for_current_ruby_set(&result->options);
+
     pm_parser_init(&result->parser, pm_string_source(&result->input), pm_string_length(&result->input), &result->options);
     pm_node_t *node = pm_parse(&result->parser);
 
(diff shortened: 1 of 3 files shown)


prism_compile.c
```diff
@@ -11492,6 +11497,19 @@ pm_parse_stdin(pm_parse_result_t *result)
     return pm_parse_process(result, node, NULL);
 }
 
+#define PM_VERSION_FOR_RELEASE(major, minor) PM_OPTIONS_VERSION_CRUBY_##major##_##minor
+
+void pm_options_version_for_current_ruby_set(pm_options_t *options) {
+# if RUBY_PATCHLEVEL == -1
```
XrXr: Checking the patchlevel can create release-only problems that development CI does not catch. Seems just the else branch is enough? Prism knows about development versions of ruby.

eregon: Agreed. Also I guess `RUBY_PATCHLEVEL == -1` might hold for commits on the release branches as well, and there we would not want to use `PM_OPTIONS_VERSION_LATEST`.

Earlopain: I thought the version was only added to prism _after_ ruby itself got bumped but turns out it was the other way around, so that was fine last time: https://github.com/ruby/prism/pull/3336

In fact, this code did not work as given, sorry... I only checked with literals while testing and did not realize it behaves different when the argument is itself a macro. So I definitly see your point lol.

That said, when the major version gets bumped it needs a bit of coordination since 4.0 is at the moment not known to prism. I hear that 3.5 might become 4.0 but I guess no one knows until it actually happens, at which point this code is kind of in the way. Maybe @kddnewton has some idea?

k0kubun: The coordination just needs to happen when it's decided. When we decide to do that on CRuby, we'd just change both Prism and CRuby to change the version checks from 3.5 to 4.0. At the moment we should handle it like another 3.x release, and past (<= 3.4) version flags shouldn't break when the latest version decides to use a 4.0 check instead of a 3.5 check.

PR #14525, opened 2025-09-12
Restore test example for argument forwarding

etiennebarrie:
Since cb419e3912f0514b8151469b0a4a4b83cbbcce78 we're no longer testing this case because foo is redefined on obj1.

This also removes a warning (which was showing the issue).

cc @byroot


byroot:
Merged as 691a5545e6161cf8c2f574ae9fc69df1232cda97

PR #14529, opened 2025-09-12
ZJIT: Test against bundled gems on CI

st0012:
Closes https://github.com/Shopify/ruby/issues/743

### DIFF
--- a/.github/workflows/zjit-macos.yml
+++ b/.github/workflows/zjit-macos.yml
@@ -101,9 +101,6 @@ jobs:
       - name: Run configure
         run: ../src/configure -C --disable-install-doc ${{ matrix.configure }}
 
-      - run: make prepare-gems
-        if: ${{ matrix.test_task == 'test-bundled-gems' }}
-
       - run: make
 
       - name: Verify that --zjit-dump-disasm works
--- a/.github/workflows/zjit-ubuntu.yml
+++ b/.github/workflows/zjit-ubuntu.yml
@@ -72,6 +72,10 @@ jobs:
             configure: '--enable-zjit=dev --with-gcc=clang-14'
             libclang_path: '/usr/lib/llvm-14/lib/libclang.so.1'
 
+          - test_task: 'test-bundled-gems'
+            configure: '--enable-zjit=dev'
+            run_opts: '--zjit-call-threshold=1'
+
     env:
       GITPULLOPTIONS: --no-tags origin ${{ github.ref }}
       RUN_OPTS: ${{ matrix.run_opts }}
--- a/gems/bundled_gems
+++ b/gems/bundled_gems
@@ -41,7 +41,7 @@ benchmark           0.4.1   https://github.com/ruby/benchmark
 logger              1.7.0   https://github.com/ruby/logger
 rdoc                6.14.2  https://github.com/ruby/rdoc
 win32ole            1.9.2   https://github.com/ruby/win32ole
-irb                 1.15.2  https://github.com/ruby/irb 331c4e851296b115db766c291e8cf54a2492fb36
+irb                 1.15.2  https://github.com/ruby/irb d43c3d764ae439706aa1b26a3ec299cc45eaed5b
 reline              0.6.2   https://github.com/ruby/reline
 readline            0.0.4   https://github.com/ruby/readline
 fiddle              1.1.8   https://github.com/ruby/fiddle
--- a/tool/test-bundled-gems.rb
+++ b/tool/test-bundled-gems.rb
@@ -1,6 +1,7 @@
 require 'rbconfig'
 require 'timeout'
 require 'fileutils'
+require 'shellwords'
 require_relative 'lib/colorize'
 require_relative 'lib/gem_env'
 
@@ -25,6 +26,7 @@ colorize = Colorize.new
 rake = File.realpath("../../.bundle/bin/rake", __FILE__)
 gem_dir = File.realpath('../../gems', __FILE__)
 rubylib = [gem_dir+'/lib', ENV["RUBYLIB"]].compact.join(File::PATH_SEPARATOR)
+run_opts = ENV["RUN_OPTS"]&.shellsplit
 exit_code = 0
 ruby = ENV['RUBY'] || RbConfig.ruby
 failed = []
@@ -33,7 +35,7 @@ File.foreach("#{gem_dir}/bundled_gems") do |line|
   next if bundled_gems&.none? {|pat| File.fnmatch?(pat, gem)}
   next unless File.directory?("#{gem_dir}/src/#{gem}/test")
 
-  test_command = [ruby, "-C", "#{gem_dir}/src/#{gem}", rake, "test"]
+  test_command = [ruby, *run_opts, "-C", "#{gem_dir}/src/#{gem}", rake, "test"]
   first_timeout = 600 # 10min
 
   toplib = gem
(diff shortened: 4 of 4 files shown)


k0kubun:
I think you meant to enable ZJIT but they don't. Could you pass either `--zjit` (this is basically what YJIT CI does) or `--zjit-call-threshold=1` (thorough testing) or `--zjit-call-threshold=2` (most testing with profiles)? I don't know how to propagate that to `test-bundled-gems` though (does `RUN_OPTS` work?).

k0kubun:
The way it works for YJIT is these jobs using `RUBY_YJIT_ENABLE=1`, but I don't like applying it for every job in the workflow (or running tests with a relatively high call threshold), so ZJIT jobs don't have the environment variable.

st0012:
Ah thanks for the context. Converting to draft to see what option works.

st0012:
I passed `--zjit-debug` to `RUN_OPTS` and it did print the expected output. So I think `RUN_OPTS` works for this too.

k0kubun:
> I passed --zjit-debug to RUN_OPTS and it did print the expected output.

What was the output you checked?

I'm looking at the CI output but it doesn't seem to have an indicator that ZJIT is enabled.

```
Testing the minitest gem
  /Users/runner/work/ruby/ruby/build/ruby -C /Users/runner/work/ruby/ruby/src/gems/src/minitest /Users/runner/work/ruby/ruby/src/.bundle/bin/rake test
  TESTOPTS is deprecated in Minitest::TestTask. Use A instead
  TESTOPTS is deprecated in Minitest::TestTask. Use A instead
  Run options: --seed 30956
  # Running:
  ......SS..........................................................................................SS.S..SSS....S.....S.....S.....S....................................................................................................................................................................................................................................................................................................................................
  Finished in 0.287374s, 1579.8228 runs/s, 4568.9589 assertions/s.
  454 runs, 1313 assertions, 0 failures, 0 errors, 12 skips
  You have skipped tests. Run with --verbose for details.
```

k0kubun:
How is `tool/test-bundled-gems.rb` propagating `RUN_OPTS` to the `test_command` it spawns?

st0012:
Ah the output I saw was from the script itself 🤦‍♂️

.github/workflows/zjit-macos.yml
```diff
@@ -101,9 +101,6 @@ jobs:
       - name: Run configure
         run: ../src/configure -C --disable-install-doc ${{ matrix.configure }}
 
-      - run: make prepare-gems
-        if: ${{ matrix.test_task == 'test-bundled-gems' }}
```
st0012, line 105: I think YJIT only test against bundled gems on Ubuntu. If we follow it, we won't need this here.

gems/bundled_gems
```diff
@@ -41,7 +41,7 @@ benchmark           0.4.1   https://github.com/ruby/benchmark
 logger              1.7.0   https://github.com/ruby/logger
 rdoc                6.14.2  https://github.com/ruby/rdoc
 win32ole            1.9.2   https://github.com/ruby/win32ole
-irb                 1.15.2  https://github.com/ruby/irb 331c4e851296b115db766c291e8cf54a2492fb36
+irb                 1.15.2  https://github.com/ruby/irb d43c3d764ae439706aa1b26a3ec299cc45eaed5b
```
st0012, line 44: This version of IRB has higher test timeout on CI. Without it, ZJIT would fail on some integration tests.

tool/test-bundled-gems.rb
```diff
@@ -100,7 +100,15 @@
     group = :pgroup
     pg = "-"
   end
-  pid = Process.spawn(test_command, group => true)
+  if rubyopt = ENV["TEST_BUNDLED_GEMS_RUBYOPT"]
+    env = ENV.to_h
+    env["RUBYOPT"] = rubyopt
+    puts "TEST_BUNDLED_GEMS_RUBYOPT is specified. Adding `#{rubyopt}` to RUBYOPT"
+    pid = Process.spawn(env, test_command, group => true)
```
st0012: Since the `test_command` is running a rake task, using `RUBYOPT` is the only way I can think of to pass the options in.

k0kubun: But the `test_command` appears to be `"#{ruby} -C #{gem_dir}/src/#{gem} #{rake} test"` (+ `" stdlib_test validate"`) or `"#{ruby} -C #{gem_dir}/src/#{gem} test/run.rb"`. 

It doesn't seem too hard to modify the script such that `rubyopt` is inserted before or after the `-C #{gem}` option, does it?

.github/workflows/zjit-ubuntu.yml
```diff
@@ -173,8 +177,18 @@ jobs:
...
         continue-on-error: ${{ matrix.continue-on-test_task || false }}
 
+      - name: make test-bundled-gems
+        run: |
+          set -x
+          make -s test-bundled-gems
+        env:
+          # This env is specifically added to tool/test-bundled-gems.rb as a way to pass options to Rake tasks
+          TEST_BUNDLED_GEMS_RUBYOPT: ${{ matrix.run_opts }}
+        if: ${{ matrix.test_task == 'test-bundled-gems' }}
```
k0kubun: Why split the step? Do any of these existing environment variables have an unwanted impact on `make test-bundled-gems`?

st0012: Using `RUN_OPTS` used to cause issues on the testing script itself when I first opened the PR a while ago, which is why I created a new env var for this. But now it seems to work this is not needed anymore.

st0012:
The failed bundled gems test is unrelated to the changes and should be fixed by https://github.com/ruby/ruby/pull/14747

tool/test-bundled-gems.rb
```diff
@@ -33,7 +34,7 @@
   next if bundled_gems&.none? {|pat| File.fnmatch?(pat, gem)}
   next unless File.directory?("#{gem_dir}/src/#{gem}/test")
 
-  test_command = [ruby, "-C", "#{gem_dir}/src/#{gem}", rake, "test"]
+  test_command = [ruby, run_opts, "-C", "#{gem_dir}/src/#{gem}", rake, "test"].compact
```
k0kubun: `RUN_OPTS` generally accepts multiple options (but this doesn't).

How about (adding `require 'shellwords'` and) using `run_opts = ENV["RUN_OPTS"]&.shellsplit` and changing it to `*run_opts` here?

st0012: Sounds good. Updated 👍

tool/test-bundled-gems.rb
```diff
@@ -33,7 +35,7 @@
   next if bundled_gems&.none? {|pat| File.fnmatch?(pat, gem)}
   next unless File.directory?("#{gem_dir}/src/#{gem}/test")
 
-  test_command = [ruby, "-C", "#{gem_dir}/src/#{gem}", rake, "test"]
+  test_command = [ruby, *run_opts, "-C", "#{gem_dir}/src/#{gem}", rake, "test"].compact
```
k0kubun: You shouldn't need `.compact` for `*nil`.

```suggestion
  test_command = [ruby, *run_opts, "-C", "#{gem_dir}/src/#{gem}", rake, "test"]
```

tool/test-bundled-gems.rb
```diff
@@ -71,7 +73,7 @@
     load_path = true
 
   when "test-unit"
-    test_command = [ruby, "-C", "#{gem_dir}/src/#{gem}", "test/run.rb"]
+    test_command = [ruby, *run_opts, "-C", "#{gem_dir}/src/#{gem}", "test/run.rb"].compact
```
k0kubun:
```suggestion
    test_command = [ruby, *run_opts, "-C", "#{gem_dir}/src/#{gem}", "test/run.rb"]
```

k0kubun:
> The failed bundled gems test is unrelated to the changes and should be fixed by https://github.com/ruby/ruby/pull/14747

Is it intentional that https://github.com/ruby/ruby/pull/14747 also fails the CI?

st0012:
It's not but it's failing on git pull `fatal: Not possible to fast-forward, aborting.` even though it should be on the latest base already.

PR #14532, opened 2025-09-12
Bump Prism version to 1.5.0

k0kubun:
```
tool/sync_default_gems.rb prism
```

on `ruby_3_4` with the latest Prism https://github.com/ruby/prism/commit/b4865f91dbdc3e30d20fd3219740cc742f71dbeb.

### DIFF
--- a/prism/prism.c
+++ b/prism/prism.c
@@ -5279,10 +5279,6 @@ pm_interpolated_string_node_append(pm_interpolated_string_node_t *node, pm_node_
 
     switch (PM_NODE_TYPE(part)) {
         case PM_STRING_NODE:
-            // If inner string is not frozen, clear flags for this string
-            if (!PM_NODE_FLAG_P(part, PM_STRING_FLAGS_FROZEN)) {
-                CLEAR_FLAGS(node);
-            }
             part->flags = (pm_node_flags_t) ((part->flags | PM_NODE_FLAG_STATIC_LITERAL | PM_STRING_FLAGS_FROZEN) & ~PM_STRING_FLAGS_MUTABLE);
             break;
         case PM_INTERPOLATED_STRING_NODE:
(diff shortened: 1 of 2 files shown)


satmandu:
Is it possible that this is breaking ruby on Linux/i686?

I'm getting segfaults when I build ruby against the 3.4 branch with prism 1.5.1 and run `ruby -e "require 'concurrent-ruby' ; puts 'success'"`

k0kubun:
It's passing the Ubuntu i686 CI job, so I'm not sure if it's a Linux/i686 problem; it could be an issue in concurrent-ruby or your environment. Please file a PR to add a test case that fails on that CI if you believe that's a CRuby issue and need something in CRuby to be fixed.

satmandu:
> It's passing the Ubuntu i686 CI job, so I'm not sure if it's a Linux/i686 problem; it could be an issue in concurrent-ruby or your environment. Please file a PR to add a test case that fails on that CI if you believe that's a CRuby issue and need something in CRuby to be fixed.

Thanks for that. I'm trying a rebuild off of the ruby_3_4 branch with `#(install_prefix}/bin/ruby -e \"require 'concurrent-ruby' ; puts 'success'\"", exception: true` in my install to check to see if a build directly causes this error.

Is there a good way to leverage the build in the Ubuntu i686 CI? Are there artifacts available?

I'm happy to provide docker based reproducers that we are generating in Chromebrew.

k0kubun:
The definition is at https://github.com/ruby/ruby/blob/8cb7224af6aa1edba5e642ce23c31fa05f1ff861/.github/workflows/ubuntu.yml#L30-L31 You can see how it's built in the workflow and compare that with how you build your Ruby. You can also run the CI by forking the repository and pushing your build process to the workflow.

satmandu:
Thanks for that! I opened an issue at https://bugs.ruby-lang.org/issues/21607 as I didn't realize 3.4.6 was going to be cut so soon, so perhaps that is helpful. It includes our build run log with the output from `make check`.

PR #14533, opened 2025-09-12
ZJIT: Add stat for `def_type` of send fallbacks

st0012:
I thought about creating a new HIR like `SendWithoutBlockFailedToOptimize` that can carry very specific reasons later. But it'll mean adding it to every branch matching `SendWithoutBlock` and may make code unnecessarily complicated.

So I take the easier path for now:

```
Top-4 send fallback def_types (100.0% of total 21,375,357):
      cfunc: 20,164,487 (94.3%)
  optimized:  1,197,897 ( 5.6%)
    attrset:     12,953 ( 0.1%)
      alias:         20 ( 0.0%)
```

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -13,10 +13,10 @@ use crate::invariants::{track_bop_assumption, track_cme_assumption, track_no_ep_
 use crate::gc::{append_gc_offsets, get_or_create_iseq_payload, get_or_create_iseq_payload_ptr, IseqPayload, IseqStatus};
 use crate::state::ZJITState;
 use crate::stats::{exit_counter_for_compile_error, incr_counter, incr_counter_by, CompileError};
-use crate::stats::{counter_ptr, with_time_stat, Counter, Counter::{compile_time_ns, exit_compile_error}};
+use crate::stats::{counter_ptr, with_time_stat, Counter, send_fallback_counter, Counter::{compile_time_ns, exit_compile_error}};
 use crate::{asm::CodeBlock, cruby::*, options::debug, virtualmem::CodePtr};
 use crate::backend::lir::{self, asm_comment, asm_ccall, Assembler, Opnd, Target, CFP, C_ARG_OPNDS, C_RET_OPND, EC, NATIVE_STACK_PTR, NATIVE_BASE_PTR, SCRATCH_OPND, SP};
-use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, Invariant, RangeType, SideExitReason, SideExitReason::*, SpecialObjectType, SpecialBackrefSymbol, SELF_PARAM_IDX};
+use crate::hir::{iseq_to_hir, Block, BlockId, BranchEdge, Invariant, RangeType, SideExitReason, SideExitReason::*, MethodType, SpecialObjectType, SpecialBackrefSymbol, SELF_PARAM_IDX};
 use crate::hir::{Const, FrameState, Function, Insn, InsnId};
 use crate::hir_type::{types, Type};
 use crate::options::get_option;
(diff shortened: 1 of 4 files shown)


zjit/src/codegen.rs
```diff
@@ -364,7 +364,10 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::IfTrue { val, target } => no_output!(gen_if_true(jit, asm, opnd!(val), target)),
         Insn::IfFalse { val, target } => no_output!(gen_if_false(jit, asm, opnd!(val), target)),
         &Insn::Send { cd, blockiseq, state, .. } => gen_send(jit, asm, cd, blockiseq, &function.frame_state(state)),
-        Insn::SendWithoutBlock { cd, state, .. } => gen_send_without_block(jit, asm, *cd, &function.frame_state(*state)),
+        Insn::SendWithoutBlock { cd, state, def_type, .. } => {
+            if let Some(def_type) = def_type { update_send_fallback_counter(asm, *def_type); }
```
k0kubun: Why here and not in `gen_send_without_block`? `gen_insn` is supposed to be a dispatch function and not a place to add implementation details for each instruction.

zjit/src/hir.rs
```diff
@@ -467,6 +467,42 @@ pub enum SideExitReason {
     StackOverflow,
 }
 
+#[derive(Debug, Clone, Copy)]
+pub enum SendWithoutBlockDefType {
```
k0kubun: It's not really about `send_without_block` (this enum itself should be used for `send` too, for example), so maybe just call it `MethodType`, given it's called `rb_method_type_t` in C?

zjit/src/stats.rs
```diff
@@ -225,6 +239,26 @@ pub fn exit_counter_ptr(reason: crate::hir::SideExitReason) -> *mut u64 {
     counter_ptr(counter)
 }
 
+pub fn send_without_block_def_type_counter(def_type: crate::hir::SendWithoutBlockDefType) -> Counter {
```
k0kubun: The counter prefix `send_fallback_` doesn't imply the `_without_block` part. Maybe just call it `send_fallback_counter` (or `send_fallback_method_type_counter` for now)?

k0kubun:
Sorry, we've just made a conflict. Could you fix that please?

k0kubun (approved):

zjit/src/hir.rs
```diff
@@ -1798,7 +1841,12 @@ impl Function {
                             let getivar = self.push_insn(block, Insn::GetIvar { self_val: recv, id, state });
                             self.make_equal_to(insn_id, getivar);
                         } else {
-                            self.push_insn_id(block, insn_id); continue;
+                            let send_without_block = self.push_insn(
+                                block,
+                                Insn::SendWithoutBlock { recv, cd, args, state, def_type: Some(MethodType::from(def_type)) }
```
tekknolagi: nit: the instructions *are* mutable. we could mutate in-place if you like. the reason we use union-find is to rewrite references for downstream users, but mutating `def_type` in-place has no semantic or naming difference

zjit/src/hir.rs
```diff
@@ -1798,7 +1841,13 @@ impl Function {
                             let getivar = self.push_insn(block, Insn::GetIvar { self_val: recv, id, state });
                             self.make_equal_to(insn_id, getivar);
                         } else {
-                            self.push_insn_id(block, insn_id); continue;
+                            if let Insn::SendWithoutBlock { def_type: insn_def_type, .. } = &mut self.insns[insn_id.0] {
```
st0012, line 1844: I think we need to do this to get a mutable `Insn`, unless we want to make `find` return a mutable `Insn`?

tekknolagi: `find` might return a copy. this seems fiiiiiiine-ish (it's slightly icky because it presumes the instruction has not been rewritten, but this is the first optimization pass, but also we should try to avoid correctness-related ordering problems). 

perhaps consider instead bailing out or just not mutating if it's not what you expect

k0kubun: I'm not sure if I understand the situation, so I have a question.

What's wrong with just doing `self.push_insn(block, Insn::SendWithoutBlock { recv, cd, args, state, def_type: Some(MethodType::from(def_type)) })` (and calling `make_equal_to` with it) here? Do we generally want to mutate existing instructions for optimizations when the opcode doesn't change?

tekknolagi: nothing was wrong with it -- it just made a massive diff. so i was saying "fyi, if you want, you *can* mutate in-place."

in general, we are not otherwise replacing A with an extraordinarily similar A'---we are changing the opcode, etc.

tekknolagi: I would probably have preferred a side-table instead of an (otherwise unused) field on the instruction, but that was a foregone conclusion

k0kubun: > it just made a massive diff

Oh, because of the HIR tests?

Either of them seems fine to me then. Are we okay to merge it as is?

st0012: Just pushed another update to make it even shorter.

PR #14534, opened 2025-09-12
ZJIT: Register spill for Opnd::VReg (WIP)

k0kubun:
very much WIP, but sharing it here for those who are curious about our current patch.


k0kubun:
I decided to implement this differently. Closing.

PR #14536, opened 2025-09-13
Fill in lead num for blocks with `it`

kddnewton:
Fixes [Bug #21256]


Earlopain:
Huh, I was pretty close. Thanks

kddnewton:
Very close! It just shouldn't have a name, that's all.

PR #14538, opened 2025-09-13
Fix prism error messages with multibyte truncation

kddnewton:
When a line is going to be displayed in an error message that contains multibyte characters, we need to respect the encoding of the source and truncate only at a character boundary, as opposed to a raw byte boundary.

Fixes [Bug #21528]

### DIFF
--- a/prism_compile.c
+++ b/prism_compile.c
@@ -10621,7 +10621,26 @@ pm_parse_errors_format_line(const pm_parser_t *parser, const pm_newline_list_t *
     // Here we determine if we should truncate the end of the line.
     bool truncate_end = false;
     if ((column_end != 0) && ((end - (start + column_end)) >= PM_ERROR_TRUNCATE)) {
-        end = start + column_end + PM_ERROR_TRUNCATE;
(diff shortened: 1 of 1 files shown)


prism_compile.c
```diff
@@ -10621,7 +10621,26 @@ pm_parse_errors_format_line(const pm_parser_t *parser, const pm_newline_list_t *
...
+
+            // If this next character would go past the end candidate,
+            // then we need to truncate before it.
+            if (ptr + char_width > end_candidate) {
+                end_candidate = ptr;
+                break;
+            }
+
+            ptr += char_width;
+        }
```
nobu, line 10641: Doesn't prism/encoding.c provide something similar to `rb_enc_left_char_head`?
Backward scan should be better than (or equal in worst cases) forward scan.

nobu (approved):

prism_compile.c
kddnewton: I don't have that implemented for every encoding since I haven't needed it yet. Since this is the error case and it's always going to be < 100 bytes, I'm not worried about performance.

PR #14547, opened 2025-09-13
Make Regexp immutable (all Regexp are frozen, not just literal regexps)

eregon:
https://bugs.ruby-lang.org/issues/8948

### DIFF
--- a/spec/ruby/core/regexp/shared/new.rb
+++ b/spec/ruby/core/regexp/shared/new.rb
@@ -5,22 +5,30 @@ describe :regexp_new, shared: true do
     Regexp.send(@method, '').is_a?(Regexp).should == true
   end
 
-  it "works by default for subclasses with overridden #initialize" do
-    class RegexpSpecsSubclass < Regexp
-      def initialize(*args)
-        super
-        @args = args
+  ruby_version_is "4.0" do
+    it "is frozen" do
+      Regexp.send(@method, '').should.frozen?
+    end
+  end
+
+  ruby_version_is ""..."4.0" do
+    it "works by default for subclasses with overridden #initialize" do
+      class RegexpSpecsSubclass < Regexp
+        def initialize(*args)
+          super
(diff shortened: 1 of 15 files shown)


spec/ruby/core/regexp/shared/new.rb
```diff
@@ -376,7 +384,7 @@ def obj.to_int() ScratchPad.record(:called) end
     end
 
     it "accepts a multiple byte character which need not be escaped" do
-      Regexp.send(@method, "\�").should == /#{"�"}/
+      Regexp.send(@method, "\�").should == /#{"�"}/
```
nobu: This backslash will be dropped in the parsers, so this wouldn't make sense as a test for the Regexp methods.

spec/ruby/optional/capi/encoding_spec.rb
```diff
@@ -475,12 +475,20 @@
       @s.rb_enc_copy("string", @obj).encoding.should == Encoding::US_ASCII
     end
 
-    it "raises a RuntimeError if the second argument is a Symbol" do
+    it "raises a RuntimeError if the first argument is a Symbol" do
```
nobu, line 492: Just a typo fix unrelated to this feature?

spec/ruby/core/regexp/shared/new.rb
eregon: Right, I see you fixed that in 67cc574dfb1943a9cb052478cc84d0c8823d0527, thanks

spec/ruby/optional/capi/encoding_spec.rb
eregon: Yeah just noticed the typo while looking at this file.

eregon:
@nobu Could you review maybe?

eregon:
Replaced by https://github.com/ruby/ruby/pull/16530

PR #14553, opened 2025-09-14
ZJIT: Debug mechanism to verify leafness for `gen_prepare_call_with_gc`

altxtech:
### Context

Closes Shopify#709

### Main changes

**Backend**  
- Mechanism to verify *leafness* of a `CCall` by arming/disarming a stack canary operand
- The canary always  disarms on `ccall`, ensuring the check applies only to the immediately following call.
- In short, arming the canary means *the next* `ccall` is expected to be leaf.

**Codegen**

- When runtime checks are enabled, `gen_prepare_call_with_gc` arms the canary
- Refactor functions for non-leaf instructions that were using `gen_prepare_call_with_gc`. Mostly duplicated code that could be replaced by `gen_prepare_non_leaf_call`
- Marked `ObjectAlloc` as non-leaf 

The `ObjectAlloc` change was necessary because I was hitting a leafness verification error while running tests, e.g
```bash
make test-all TESTS="test/digest/test_digest.rb --name=TestDigest::TestBase#test_base" RUN_OPTS="--zjit-call-threshold=1";
``` 

Since `rb_obj_alloc` can raise, maybe this instruction is not guaranteed to always be leaf.

### Testing

This mechanism is for debugging only, so I haven’t added unit tests. Instead, manual testing can be done by intentionally misclassifying a non-leaf call as leaf and observing the failure.

For example, force `gen_to_new_array` to use `prepare_call_with_gc` instead of `prepare_non_leaf_call`:
```rust
fn gen_to_new_array(jit: &mut JITState, asm: &mut Assembler, val: Opnd, state: &FrameState) -> lir::Opnd {
    //gen_prepare_non_leaf_call(jit, asm, state);
    gen_prepare_call_with_gc(asm, state);
    asm_ccall!(asm, rb_vm_splat_array, Opnd::Value(Qtrue), val)
}
```

And run this script:  
```ruby
# to_new_array.rb
class X
  def to_a
    [1,2]
  end
end

def test(a)
  [*a]
end

100.times{ test(X.new) }
```

```bash
make miniruby && ./miniruby --zjit --zjit-dump-hir to_new_array.rb
```

Expected output should include:
```
We are killing the stack canary set by splatarray, at <RubyVM::InstructionSequence:test@to_new_array.rb:7>@pc=2                                                                                                                                                
watch out the C stack trace.                                      
```

<details>
<summary>Example of full crash output</summary>
We are killing the stack canary set by splatarray, at <RubyVM::InstructionSequence:test@to_new_array.rb:7>@pc=2                                                                                                                                                
watch out the C stack trace.                                                                                                                                                                                                                                   
== disasm: #<ISeq:test@to_new_array.rb:7 (7,0)-(9,3)>                                                                                                                                                                                                          
local table (size: 1, argc: 1 [opts: 0, rest: -1, post: 0, block: -1, kw: -1@-1, kwrest: -1])                                                                                                                                                                  
[ 1] a@0<Arg>                                                                                                                                                                                                                                                  
0000 getlocal_WC_0                          a@0                       (   8)[LiCa]                                                                                                                                                                             
0002 splatarray                             true                                                                                                                                                                                                               
0004 leave                                                            (   9)[Re]                                                                                                                                                                               
to_new_array.rb:8: [BUG] see above.                                                                                                                                                                                                                            
ruby 3.5.0dev (2025-09-16T01:52:28Z altxtech/prepare-c.. e682145c6e) +ZJIT dev +PRISM [x86_64-linux]                                                                                                                                                           
                                                                                                                                                                                                                                                               
c:0005 p:0004 s:0020 e:000018 METHOD to_new_array.rb:8                                                                                                                                                                                                         
c:0004 p:0018 s:0014 e:000013 BLOCK  to_new_array.rb:11                                                                                                                                                                                                        
c:0003 p:0024 s:0011 e:000010 METHOD <internal:numeric>:257                                                                                                                                                                                                    
c:0002 p:0016 s:0006 e:000005 EVAL   to_new_array.rb:11 [FINISH]                                                                                                                                                                                               
c:0001 p:0000 s:0003 E:0026d0 DUMMY  [FINISH]                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                               
-- Ruby level backtrace information ----------------------------------------                                                                                                                                                                                   
to_new_array.rb:11:in '<main>'                                                                                                                                                                                                                                 
<internal:numeric>:257:in 'times'                                                                                                                                                                                                                              
to_new_array.rb:11:in 'block in <main>'                                                                                                                                                                                                                        
to_new_array.rb:8:in 'test'                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                               
Total ractor count: 1                                                                                                                                                                                                                                          
Ruby thread count for this ractor: 1                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                               
/home/andre/repos/ruby/miniruby(rb_print_backtrace+0x5) [0x55bd3d577ef6] /home/andre/repos/ruby/vm_dump.c:843                                                                                                                                                  
/home/andre/repos/ruby/miniruby(rb_vm_bugreport) /home/andre/repos/ruby/vm_dump.c:1175                                                                                                                                                                         
/home/andre/repos/ruby/miniruby(rb_bug_without_die_internal+0x9f) [0x55bd3d22894b] /home/andre/repos/ruby/error.c:1097                                                                                                                                         
/home/andre/repos/ruby/miniruby(rb_bug) /home/andre/repos/ruby/error.c:1115                                                                                                                                                                                    
[0x55bd3d53cc69]                                                                                                                                                                                                                                               
/home/andre/repos/ruby/miniruby(vm_call0_body+0x4d8) [0x55bd3d55b1c8] /home/andre/repos/ruby/vm_eval.c:211                                                                                                                                                     
/home/andre/repos/ruby/miniruby(vm_call0_cc+0x125) [0x55bd3d55ba85] /home/andre/repos/ruby/vm_eval.c:101                                                                                                                                                       
/home/andre/repos/ruby/miniruby(rb_vm_call0+0x51) [0x55bd3d562aa3] /home/andre/repos/ruby/vm_eval.c:61                                                                                                                                                         
/home/andre/repos/ruby/miniruby(rb_vm_call_kw) /home/andre/repos/ruby/vm_eval.c:326                                                                                                                                                                            
/home/andre/repos/ruby/miniruby(rb_check_funcall_default_kw) /home/andre/repos/ruby/vm_eval.c:712                                                                                                                                                              
/home/andre/repos/ruby/miniruby(convert_type_with_id+0x5) [0x55bd3d3e3765] /home/andre/repos/ruby/object.c:3195                                                                                                                                                
/home/andre/repos/ruby/miniruby(rb_check_convert_type_with_id) /home/andre/repos/ruby/object.c:3287                                                                                                                                                            
/home/andre/repos/ruby/miniruby(vm_splat_array+0x13) [0x55bd3d53e9d9] /home/andre/repos/ruby/vm_insnhelper.c:5733                                                                                                                                              
/home/andre/repos/ruby/miniruby(rb_vm_splat_array) /home/andre/repos/ruby/vm_insnhelper.c:5750                                                                                                                                                                 
[0x55bd3920f471]                                        
-- Other runtime information -----------------------------------------------                                                                                                                                                                                   
                                                                                                                                                                                                                                                               
* Loaded script: to_new_array.rb                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                               
* Loaded features:                                                                                                                                                                                                                                             

    0 enumerator.so
    1 thread.rb
    2 fiber.so
    3 rational.so
    4 complex.so
    5 pathname.so
    6 ruby2_keywords.rb
    7 set.rb

* Process memory map:

55bd3920f000-55bd39210000 r-xp 00000000 00:00 0 
55bd39210000-55bd3d20f000 ---p 00000000 00:00 0 
55bd3d21b000-55bd3d9f0000 r-xp 00000000 00:21 23291106                   /home/andre/repos/ruby/miniruby
55bd3d9f0000-55bd3e251000 r--p 007d5000 00:21 23291106                   /home/andre/repos/ruby/miniruby
55bd3e251000-55bd3e3b2000 r--p 01035000 00:21 23291106                   /home/andre/repos/ruby/miniruby
55bd3e3b2000-55bd3e3b5000 rw-p 01196000 00:21 23291106                   /home/andre/repos/ruby/miniruby
55bd3e3b5000-55bd3e3c9000 rw-p 00000000 00:00 0 
55bd71681000-55bd717f4000 rw-p 00000000 00:00 0                          [heap]
7f32a9130000-7f32a9150000 rw-p 00000000 00:00 0 
7f32a915f000-7f32a91a0000 rw-p 00000000 00:00 0 
7f32a91af000-7f32a91b0000 ---p 00000000 00:00 0 
7f32a91b0000-7f32a9251000 rw-p 00000000 00:00 0 
7f32a9251000-7f32a9252000 ---p 00000000 00:00 0 
7f32a9252000-7f32a92f3000 rw-p 00000000 00:00 0 
7f32a92f3000-7f32a92f4000 ---p 00000000 00:00 0 
7f32a92f4000-7f32a9395000 rw-p 00000000 00:00 0 
7f32a9395000-7f32a9396000 ---p 00000000 00:00 0 
7f32a9396000-7f32a9437000 rw-p 00000000 00:00 0 
7f32a9437000-7f32a9438000 ---p 00000000 00:00 0 
7f32a9438000-7f32a94d9000 rw-p 00000000 00:00 0 
7f32a94d9000-7f32a94da000 ---p 00000000 00:00 0 
7f32a94da000-7f32a957b000 rw-p 00000000 00:00 0 
7f32a957b000-7f32a957c000 ---p 00000000 00:00 0 
7f32a957c000-7f32a961d000 rw-p 00000000 00:00 0 
7f32a961d000-7f32a961e000 ---p 00000000 00:00 0 
7f32a961e000-7f32a96bf000 rw-p 00000000 00:00 0 
7f32a96bf000-7f32a96c0000 ---p 00000000 00:00 0 
7f32a96c0000-7f32a9761000 rw-p 00000000 00:00 0 
7f32a9761000-7f32a9762000 ---p 00000000 00:00 0 
7f32a9762000-7f32a9803000 rw-p 00000000 00:00 0 
7f32a9803000-7f32a9804000 ---p 00000000 00:00 0 
7f32a9804000-7f32a98a5000 rw-p 00000000 00:00 0 
7f32a98a5000-7f32a98a6000 ---p 00000000 00:00 0 
7f32a98a6000-7f32a9947000 rw-p 00000000 00:00 0 
7f32a9947000-7f32a9948000 ---p 00000000 00:00 0 
7f32a9948000-7f32a99e9000 rw-p 00000000 00:00 0 
7f32a99e9000-7f32a99ea000 ---p 00000000 00:00 0 
7f32a99ea000-7f32a9a8b000 rw-p 00000000 00:00 0 
7f32a9a8b000-7f32a9a8c000 ---p 00000000 00:00 0 
7f32a9a8c000-7f32a9b2d000 rw-p 00000000 00:00 0 
7f32a9b2d000-7f32a9b2e000 ---p 00000000 00:00 0 
7f32a9b2e000-7f32a9bcf000 rw-p 00000000 00:00 0 
7f32a9bcf000-7f32a9bd0000 ---p 00000000 00:00 0 
7f32a9bd0000-7f32a9c71000 rw-p 00000000 00:00 0 
7f32a9c71000-7f32a9c72000 ---p 00000000 00:00 0 
7f32a9c72000-7f32a9d13000 rw-p 00000000 00:00 0 
7f32a9d13000-7f32a9d14000 ---p 00000000 00:00 0 
7f32a9d14000-7f32a9db5000 rw-p 00000000 00:00 0 
7f32a9db6000-7f32a9e57000 rw-p 00000000 00:00 0                                                                                                                                                                                              23:03:16 [32/1920]
7f32a9e57000-7f32a9e58000 ---p 00000000 00:00 0 
7f32a9e58000-7f32a9ef9000 rw-p 00000000 00:00 0 
7f32a9ef9000-7f32a9efa000 ---p 00000000 00:00 0 
7f32a9efa000-7f32a9f9b000 rw-p 00000000 00:00 0 
7f32a9f9b000-7f32a9f9c000 ---p 00000000 00:00 0 
7f32a9f9c000-7f32aa03d000 rw-p 00000000 00:00 0 
7f32aa03d000-7f32aa03e000 ---p 00000000 00:00 0 
7f32aa03e000-7f32aa0df000 rw-p 00000000 00:00 0 
7f32aa0df000-7f32aa0e0000 ---p 00000000 00:00 0 
7f32aa0e0000-7f32aa181000 rw-p 00000000 00:00 0 
7f32aa181000-7f32aa182000 ---p 00000000 00:00 0 
7f32aa182000-7f32aa223000 rw-p 00000000 00:00 0 
7f32aa223000-7f32aa224000 ---p 00000000 00:00 0 
7f32aa224000-7f32aa2c5000 rw-p 00000000 00:00 0 
7f32aa2c5000-7f32aa2c6000 ---p 00000000 00:00 0 
7f32aa2c6000-7f32aa367000 rw-p 00000000 00:00 0 
7f32aa367000-7f32aa368000 ---p 00000000 00:00 0 
7f32aa368000-7f32aa409000 rw-p 00000000 00:00 0 
7f32aa409000-7f32aa40a000 ---p 00000000 00:00 0 
7f32aa40a000-7f32aa4ab000 rw-p 00000000 00:00 0 
7f32aa4ab000-7f32aa4ac000 ---p 00000000 00:00 0 
7f32aa4ac000-7f32aa54d000 rw-p 00000000 00:00 0 
7f32aa54d000-7f32aa54e000 ---p 00000000 00:00 0 
7f32aa54e000-7f32aa5ef000 rw-p 00000000 00:00 0 
7f32aa5ef000-7f32aa5f0000 ---p 00000000 00:00 0 
7f32aa5f0000-7f32aadf0000 rw-p 00000000 00:00 0 
7f32aadf0000-7f32c4200000 rw-p 00000000 00:00 0 
7f32c4200000-7f32d2070000 r--p 00000000 00:21 13944013                   /usr/lib/locale/locale-archive
7f32d2080000-7f32d20d0000 rw-p 00000000 00:00 0 
7f32d20df000-7f32d2110000 rw-p 00000000 00:00 0 
7f32d2115000-7f32d2219000 rw-p 00000000 00:00 0 
7f32d2219000-7f32d2388000 r-xp 00000000 00:21 13944590                   /usr/lib64/libc.so.6
7f32d2388000-7f32d23fd000 r--p 0016f000 00:21 13944590                   /usr/lib64/libc.so.6
7f32d23fd000-7f32d2401000 r--p 001e3000 00:21 13944590                   /usr/lib64/libc.so.6
7f32d2401000-7f32d2403000 rw-p 001e7000 00:21 13944590                   /usr/lib64/libc.so.6
7f32d2403000-7f32d240d000 rw-p 00000000 00:00 0 
7f32d240d000-7f32d2430000 r-xp 00000000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f32d2430000-7f32d2437000 r--p 00023000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f32d2437000-7f32d2438000 r--p 0002a000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f32d2438000-7f32d2439000 rw-p 00000000 00:00 0 
7f32d2439000-7f32d24b5000 r-xp 00000000 00:21 13944592                   /usr/lib64/libm.so.6
7f32d24b5000-7f32d2525000 r--p 0007c000 00:21 13944592                   /usr/lib64/libm.so.6
7f32d2525000-7f32d2526000 r--p 000eb000 00:21 13944592                   /usr/lib64/libm.so.6
7f32d2526000-7f32d2527000 rw-p 000ec000 00:21 13944592                   /usr/lib64/libm.so.6
7f32d2527000-7f32d2540000 r-xp 00000000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f32d2540000-7f32d255a000 r--p 00019000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f32d255a000-7f32d255b000 r--p 00032000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f32d255b000-7f32d255c000 rw-p 00000000 00:00 0 
7f32d255c000-7f32d2573000 r-xp 00000000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f32d2573000-7f32d257d000 r--p 00017000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f32d257d000-7f32d257e000 r--p 00020000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f32d257e000-7f32d257f000 rw-p 00021000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f32d2599000-7f32d259a000 rw-p 00000000 00:00 0 
7f32d259a000-7f32d259b000 rw-p 00000000 00:00 0 
7f32d259b000-7f32d259d000 rw-p 00000000 00:00 0 
7f32d259d000-7f32d25a1000 r--p 00000000 00:00 0                          [vvar]
7f32d25a1000-7f32d25a3000 r--p 00000000 00:00 0                          [vvar_vclock]
7f32d25a3000-7f32d25a5000 r-xp 00000000 00:00 0                          [vdso]
7f32d25a5000-7f32d25ce000 r-xp 00000000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f32d25ce000-7f32d25d9000 r--p 00029000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f32d25d9000-7f32d25db000 r--p 00034000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f32d25db000-7f32d25dc000 rw-p 00036000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f32d25dc000-7f32d25dd000 rw-p 00000000 00:00 0 
7fff35de3000-7fff35e05000 rw-p 00000000 00:00 0                          [stack]
ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]


zjit/src/codegen.rs
```diff
@@ -1260,8 +1230,8 @@ fn gen_new_range_fixnum(
     asm_ccall!(asm, rb_range_new, low, high, (flag as i64).into())
 }
 
-fn gen_object_alloc(asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
-    gen_prepare_call_with_gc(asm, state);
+fn gen_object_alloc(jit: &JITState, asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
+    gen_prepare_non_leaf_call(jit, asm, state);
```
altxtech, line 1236: Dug a little deeper on this.  

Checked out to master and created a probed version of `rb_obj_alloc`.

```diff
diff --git a/eval.c b/eval.c
index 019a2d19a2..0ab95615ea 100644
--- a/eval.c
+++ b/eval.c
@@ -671,6 +671,11 @@ rb_exc_exception(VALUE mesg, enum ruby_tag_type tag, VALUE cause)
     rb_longjmp(GET_EC(), tag, mesg, cause);
 }
 
+extern int rb_obj_alloc_probe_depth;
+static inline int rb_obj_alloc_in_scope(void) {
+    return rb_obj_alloc_probe_depth > 0;
+}
+
 /**
  * Raises an exception in the current thread.
  * @param[in] mesg an Exception class or an `Exception` object.
@@ -681,6 +686,9 @@ rb_exc_exception(VALUE mesg, enum ruby_tag_type tag, VALUE cause)
 void
 rb_exc_raise(VALUE mesg)
 {
+    if (rb_obj_alloc_in_scope()) {
+        fprintf(stderr, "[probe] raise while in rb_obj_alloc\n");
+    }
     rb_exc_exception(mesg, TAG_RAISE, Qundef);
 }
 
diff --git a/gc.c b/gc.c
index 0433da2e67..9eb0071b15 100644
--- a/gc.c
+++ b/gc.c
@@ -131,6 +131,7 @@
 #include "builtin.h"
 #include "shape.h"
 
+
 unsigned int
 rb_gc_vm_lock(const char *file, int line)
 {
@@ -4251,9 +4252,17 @@ rb_gc_update_object_references(void *objspace, VALUE obj)
     UPDATE_IF_MOVED(objspace, RBASIC(obj)->klass);
 }
 
+extern int rb_obj_alloc_probe_depth;
+static inline int rb_obj_alloc_in_scope(void) {
+    return rb_obj_alloc_probe_depth > 0;
+}
+
 VALUE
 rb_gc_start(void)
 {
+    if (rb_obj_alloc_in_scope()) {
+        fprintf(stderr, "[probe] GC while in rb_obj_alloc\n");
+    }
     rb_gc();
     return Qnil;
 }
@@ -5516,3 +5525,4 @@ ruby_annotate_mmap(const void *addr, unsigned long size, const char *name)
     errno = 0;
 #endif
 }
+
diff --git a/include/ruby/internal/intern/object.h b/include/ruby/internal/intern/object.h
index 9daad7d046..b754826b68 100644
--- a/include/ruby/internal/intern/object.h
+++ b/include/ruby/internal/intern/object.h
@@ -160,6 +160,7 @@ VALUE rb_obj_is_kind_of(VALUE obj, VALUE klass);
  * @see        #TypedData_Make_Struct
  */
 VALUE rb_obj_alloc(VALUE klass);
+VALUE rb_obj_alloc_with_probe(VALUE klass);
 
 /**
  * Produces a shallow copy of the given object.  Its list of instance variables
diff --git a/object.c b/object.c
index 56afc7e99b..995aa7915b 100644
--- a/object.c
+++ b/object.c
@@ -2180,9 +2180,11 @@ class_get_alloc_func(VALUE klass)
     rb_alloc_func_t allocator;
 
     if (!RCLASS_INITIALIZED_P(klass)) {
+        printf("KABOOM - Class not initialized\n");
         rb_raise(rb_eTypeError, "can't instantiate uninitialized class");
     }
     if (RCLASS_SINGLETON_P(klass)) {
+        printf("KABOOM - Singleton class\n");
         rb_raise(rb_eTypeError, "can't create instance of singleton class");
     }
     allocator = rb_get_alloc_func(klass);
@@ -2207,6 +2209,7 @@ class_call_alloc_func(rb_alloc_func_t allocator, VALUE klass)
     return obj;
 }
 
+
 VALUE
 rb_obj_alloc(VALUE klass)
 {
@@ -2214,6 +2217,19 @@ rb_obj_alloc(VALUE klass)
     return rb_class_alloc(klass);
 }
 
+int rb_obj_alloc_probe_depth = 0;
+
+VALUE
+rb_obj_alloc_with_probe(VALUE klass)
+{
+    Check_Type(klass, T_CLASS);
+
+    rb_obj_alloc_probe_depth++;
+    VALUE ret = rb_class_alloc(klass);
+    rb_obj_alloc_probe_depth--;
+
+    return ret;
+}
 /*
  *  call-seq:
  *     class.new(args, ...)    ->  obj
diff --git a/zjit/bindgen/src/main.rs b/zjit/bindgen/src/main.rs
index 7d66bf0ecf..82c2554aa4 100644
--- a/zjit/bindgen/src/main.rs
+++ b/zjit/bindgen/src/main.rs
@@ -324,6 +324,7 @@ fn main() {
         .allowlist_function("rb_obj_equal")
         .allowlist_function("rb_class_new_instance_pass_kw")
         .allowlist_function("rb_obj_alloc")
+        .allowlist_function("rb_obj_alloc_with_probe")
         .allowlist_function("rb_obj_info")
         .allowlist_function("ruby_xfree")
         .allowlist_function("rb_profile_frames")
diff --git a/zjit/src/codegen.rs b/zjit/src/codegen.rs
index 1f71ad9db3..8076309ca9 100644
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1262,7 +1262,7 @@ fn gen_new_range_fixnum(
 
 fn gen_object_alloc(asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
     gen_prepare_call_with_gc(asm, state);
-    asm_ccall!(asm, rb_obj_alloc, val)
+    asm_ccall!(asm, rb_obj_alloc_with_probe, val)
 }
 
 /// Compile code that exits from JIT code with a return value
diff --git a/zjit/src/cruby_bindings.inc.rs b/zjit/src/cruby_bindings.inc.rs
index 9ee4b1bb74..b20250be33 100644
--- a/zjit/src/cruby_bindings.inc.rs
+++ b/zjit/src/cruby_bindings.inc.rs
@@ -812,6 +812,7 @@ unsafe extern "C" {
     ) -> VALUE;
     pub fn rb_obj_is_kind_of(obj: VALUE, klass: VALUE) -> VALUE;
     pub fn rb_obj_alloc(klass: VALUE) -> VALUE;
+    pub fn rb_obj_alloc_with_probe(klass: VALUE) -> VALUE;
     pub fn rb_obj_frozen_p(obj: VALUE) -> VALUE;
     pub fn rb_class_inherited_p(scion: VALUE, ascendant: VALUE) -> VALUE;
     pub fn rb_backref_get() -> VALUE;
```

Then I ran the test that was giving me trouble:
`make test-all TESTS="-v test/digest/test_digest.rb --name=TestDigest::TestBase#test_base" RUN_OPTS="--zjit-call-threshold=1";`  

```
Run options: 
  --seed=8140
  "--ruby=./miniruby -I./lib -I. -I.ext/common  ./tool/runruby.rb --extout=.ext  -- --zjit-call-threshold=1"
  --excludes-dir=./test/.excludes
  --name=!/memory_leak/
  -v
  --name=TestDigest::TestBase#test_base

# Running tests:

[1/1] TestDigest::TestBase#test_base[probe] raise while in rb_obj_alloc
 = 0.07 s
[probe] raise while in rb_obj_alloc
[probe] raise while in rb_obj_alloc
Finished tests in 0.246294s, 4.0602 tests/s, 12.1806 assertions/s.
1 tests, 3 assertions, 0 failures, 0 errors, 0 skips

ruby -v: ruby 3.5.0dev (2025-09-15T13:25:20Z master b0ce1fd549) +ZJIT dev +PRISM [x86_64-linux]
[probe] raise while in rb_obj_alloc
```  

So it really seems like it raises sometimes.

zjit/src/backend/lir.rs
```diff
@@ -1190,9 +1192,20 @@ impl Assembler
             insns: Vec::with_capacity(ASSEMBLER_INSNS_CAPACITY),
             live_ranges,
             label_names,
+            leaf_canary_addr: None
         }
     }
 
+    pub fn arm_leaf_canary(&mut self, sp_addr: Opnd) {
```
k0kubun: I felt so confused by "arm" because we support Arm in ZJIT. Can we maybe just say "set" and "unset"?

zjit/src/codegen.rs
```diff
@@ -1533,6 +1503,29 @@ fn gen_incr_counter(asm: &mut Assembler, counter: Counter) {
...
+fn gen_set_vm_stack_canary(asm: &mut Assembler, state: &FrameState) {
+    // compute address of *(cfp->sp) after current virtual stack
+    let stack_size = state.stack_size();
+    let sp_addr = asm.lea(Opnd::mem(64, SP, (stack_size as i32) * SIZEOF_VALUE_I32));
+
+    // save cfp->sp for the interpreter
+    asm.mov(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_SP), sp_addr);
+
+    // write canary
+    let canary_val = asm_ccall!(asm, rb_vm_stack_canary, );
```
k0kubun: If the last `, ` is forced by how the macro is defined, could we modify it to make this `, ` optional?

zjit/src/codegen.rs
```diff
@@ -1533,6 +1503,29 @@ fn gen_incr_counter(asm: &mut Assembler, counter: Counter) {
 /// Unlike YJIT, we don't need to save the stack slots to protect them from GC
 /// because the backend spills all live registers onto the C stack on CCall.
 fn gen_prepare_call_with_gc(asm: &mut Assembler, state: &FrameState) {
+    gen_save_pc(asm, state);
+    if cfg!(feature = "runtime_checks") {
```
k0kubun: How did you test this locally?

With `--enable-yjit=dev --enable-zjit=dev`, it'd probably work, but I think this doesn't work with `--enable-zjit=dev` alone, so the CI wouldn't use it. You probably need to change `configure.ac` to add `runtime_checks` to `rb_cargo_features` on the `${ZJIT_SUPPORT}` check.

zjit/src/codegen.rs
```diff
@@ -1533,6 +1503,29 @@ fn gen_incr_counter(asm: &mut Assembler, counter: Counter) {
...
+        gen_set_vm_stack_canary(asm, state);
+    }
+}
+
+fn gen_set_vm_stack_canary(asm: &mut Assembler, state: &FrameState) {
```
k0kubun: YJIT was setting it right before a `call` instruction, but in your version the canary could be set way before a `call` instruction. There could be random code that could erase the canary in between `gen_prepare_call_with_gc` and `asm.ccall`, so I think it's better to write the canary right before the `call` instruction (just write and clear the canary inside `asm.ccall`) to make sure it's checked, just like YJIT does.

zjit/src/codegen.rs
tekknolagi: I don't love this. Is there a clean way to isolate when it might raise? ObjectAlloc is supposed to be effect-free

zjit/src/backend/lir.rs
altxtech: Sure. Can do.

zjit/src/codegen.rs
altxtech: I would need to investigate more.

zjit/src/codegen.rs
altxtech: Maybe the comma is not required. Just of an artifact of other ways I experimented writing this. 

But yes, I can try to remove it, even if it needs to edit the macro.

zjit/src/codegen.rs
altxtech: Yes I enabled both YJIT and ZJIT for this, because I did some experiments on  how YJIT behaved too.  Used some adhoc log statements to make sure these checks were firing.

zjit/src/codegen.rs
k0kubun: If the canary is set up correctly, it should crash when it raises. The `[BUG]` output would give you backtraces, so you should be able to see where it's coming from.

You shouldn't need a custom diff to investigate that if it's supposed to provide debug information on its own 🙂

altxtech: Oh, yeah, true. I could have just done that :laughing: 

To clarify, the above diff was after I checked out to the master branch, so the canary mechanism wasn't present. I do get the expect crash when I run it with the canary enabled, and the backtrace points to the 
`rb_digest_base_alloc`.


```
/home/andre/repos/ruby/ruby(rb_print_backtrace+0x5) [0x557b73bc4256] /home/andre/repos/ruby/
...
/home/andre/repos/ruby/ruby(rb_raise+0xa6) [0x557b73b78de6] /home/andre/repos/ruby/error.c:3799
/home/andre/repos/ruby/.ext/x86_64-linux/digest.so(rb_digest_base_alloc+0x18) [0x7f85c8910c6e] /home/andre/repos/ruby/ext/digest/digest.c:641
/home/andre/repos/ruby/.ext/x86_64-linux/digest.so(algo_init) /home/andre/repos/ruby/ext/digest/digest.c:634
/home/andre/repos/ruby/.ext/x86_64-linux/digest.so(rb_digest_base_alloc) /home/andre/repos/ruby/ext/digest/digest.c:648
/home/andre/repos/ruby/ruby(class_call_alloc_func+0x5) [0x557b733de355] /home/andre/repos/ruby/object.c:2202
/home/andre/repos/ruby/ruby(rb_class_alloc) /home/andre/repos/ruby/object.c:2174
/home/andre/repos/ruby/ruby(rb_obj_alloc) /home/andre/repos/ruby/object.c:2214
```

Looking at that function:  

```c
static VALUE
rb_digest_base_alloc(VALUE klass)
{
    rb_digest_metadata_t *algo;
    VALUE obj;
    void *pctx;

    if (klass == rb_cDigest_Base) {
        rb_raise(rb_eNotImpError, "Digest::Base is an abstract class");
    }

    algo = get_digest_base_metadata(klass);

    obj = rb_data_typed_object_zalloc(klass, algo->ctx_size, &digest_type);
    pctx = RTYPEDDATA_DATA(obj);
    algo_init(algo, pctx);

    return obj;
}
```

So, in the case of `Base::Digest` specifically an error can happen during allocation.

tekknolagi: @k0kubun @XrXr is this part of the contract? Can people legally raise in an allocator function? If so, maybe we can further specialize based on the class's allocator or something. Ideally a "normal" or "fast path" alloc should have *zero* side effects

XrXr: Yes allocator functions can raise/do whatever else they need. They are not different from C methods because once you set a custom alloc function, it's reachable by calling `.allocate` in rubyland.

k0kubun: So yeah it seems rather right to use `gen_prepare_non_leaf_call` here.

You'd need to specialize `Insn::ObjectAlloc` further for each class to get rid of the `gen_prepare_non_leaf_call` prep. YJIT does `jit_prepare_non_leaf_call` before `rb_obj_alloc` too.

zjit/src/codegen.rs
altxtech: How much logic should I move into the backend for this?

The address for canary is calculated from `FrameState`  stack size. I could go as far as passing the reference to the frame state which `Assembler` will keep until the time of setting the canary. 

Or I could just move just this snippet to be executed before the ccall.
```
    // write canary
    let canary_val = asm_ccall!(asm, rb_vm_stack_canary, );
    asm.store(Opnd::mem(64, sp_addr, 0), canary_val);
```

k0kubun: YJIT does this in `Assembler::ccall`, and I think ZJIT can do the same, which is not really "move into the backend". 

You'd need to keep something like `Option<FrameState>` on `Assembler` and let `Assembler::ccall` react to it.

k0kubun: It's only used for the stack size, so you could just keep the stack size instead of the whole `FrameState`.

zjit/src/codegen.rs
k0kubun: By the way, you shouldn't need `asm_ccall!` for this C function in the first place. In https://github.com/Shopify/ruby/issues/709#issuecomment-3271456444, I kind of expected you to look at YJIT as a reference for implementing this, but I'm sorry if I wasn't clear 🙂 It can be called at the compile time like YJIT does.

zjit/src/codegen.rs
k0kubun: ok. Anyway, let's fix `configure.ac`. ZJIT shouldn't need `--enable-yjit=dev` to make it work.

zjit/src/codegen.rs
altxtech: Can the stack size change between `gen_prepare_call_with_gc` and `ccall`? If that's the case, it would be better to save a reference to FrameState, no?

k0kubun: Each HIR instruction has only one `FrameState` (or stack size) for side exits, so as long as `gen_prepare_call_with_gc` is called for `asm.ccall` for the same HIR, it shouldn't change. It should be used only when you're about to make a C call with GC, so they should never span across different HIR instructions.

Also, FrameState referenced by an instruction is not mutated at any point, so it doesn't help.

zjit/src/codegen.rs
tekknolagi: ok. in that case, for now, let's at least have a check for *uses custom allocator?* in codegen to avoid prepare non-leaf. the default allocator shouldn't, right?

zjit/src/backend/lir.rs
altxtech: Done. Renamed to `set_stack_canary` and `clear_stack_canary`.  This should be clearer, and is also consistent with YJIT

zjit/src/codegen.rs
altxtech: Done.

zjit/src/codegen.rs
altxtech: Done.

zjit/src/codegen.rs
altxtech: Done. Also made the implementation more similar to YJIT in general. I think we can start from that point and then change anything that should be done differently in ZJIT.

zjit/src/codegen.rs
altxtech: This is the only item I haven't taken action on yet. 

One option is leave `obj_alloc` as leaf for now and disable the test that breaks it, and the deal with it in a separate issue / PR

tekknolagi: Feel free to change to `gen_prepare_non_leaf_call` as long as you make a Shopify/ruby issue and tag it in a TODO that explains the need to split the opcode for better codegen

zjit/src/codegen.rs
```diff
@@ -1189,6 +1158,7 @@ fn gen_new_array(
     elements: Vec<Opnd>,
     state: &FrameState,
 ) -> lir::Opnd {
+
```
```suggestion
```

zjit/src/backend/lir.rs
```diff
@@ -1190,9 +1193,32 @@ impl Assembler
...
+                let canary_addr = self.lea(Opnd::mem(64, SP, (stack_size as i32) * SIZEOF_VALUE_I32));
+                let canary_opnd = Opnd::mem(64, canary_addr, 0);
+                self.mov(canary_opnd, vm_stack_canary().into());
+                return Some(canary_opnd)
+            }
+        }
+        None
+    }
+
+    fn clear_stack_canary(&mut self, canary_opnd: Option<Opnd>){
```
k0kubun:
```suggestion
    pub fn expect_leaf_ccall(&mut self, stack_size: usize) {
        self.leaf_ccall_stack_size = Some(stack_size);
    }

    fn set_stack_canary(&mut self) -> Option<Opnd> {
        if cfg!(feature = "runtime_checks") {
            if let Some(stack_size) = self.leaf_ccall_stack_size.take(){
                let canary_addr = self.lea(Opnd::mem(64, SP, (stack_size as i32) * SIZEOF_VALUE_I32));
                let canary_opnd = Opnd::mem(64, canary_addr, 0);
                self.mov(canary_opnd, vm_stack_canary().into());
                return Some(canary_opnd)
            }
        }
        None
    }

    fn clear_stack_canary(&mut self, canary_opnd: Option<Opnd>) {
```

This spacing seems inconsistent with other places

zjit/src/codegen.rs
altxtech: TODO included

zjit/src/codegen.rs
```diff
@@ -1059,9 +1039,9 @@ fn gen_send_without_block_direct(
     asm.jbe(side_exit(jit, state, StackOverflow));
 
     // Save cfp->pc and cfp->sp for the caller frame
-    gen_prepare_call_with_gc(asm, state);
+    gen_save_pc(asm, state);
```
tekknolagi: please use `gen_prepare_call_with_gc`

zjit/src/codegen.rs
```diff
@@ -1059,9 +1039,9 @@ fn gen_send_without_block_direct(
...
+    // Special SP math. Can't use gen_prepare_non_leaf_call
     gen_save_sp(asm, state.stack().len() - args.len() - 1); // -1 for receiver
-
```
tekknolagi, line 1064: Please don't change whitespace

zjit/src/codegen.rs
tekknolagi: Please don't change whitespace

zjit/src/codegen.rs
tekknolagi: unless there is some specific reason i am missing here

altxtech: We can't use that here because this is a non-leaf call. 

If you look at the sequence of calls here, we got almost 1-to-1 what `gen_prepare_non_leaf_call` does. We just can't replace directly this for `gen_prepare_non_leaf_call` because we got some special SP math.

So I just replaced `gen_prepare_call_with_gc` to avoid calling `asm.expect_leaf_ccall()`

k0kubun: How about calling `gen_prepare_non_leaf_call` with a little modified `FrameState` here by introducing a function like `state.with_stack_size(state.stack().len() - args.len() - 1)`?

I think we don't need to spill things above the SP, so it should work.

altxtech: Actually, I don't think that works either.  The stack size, whether you getting it through `state.stack_size()` or `state.stack().len()`, is derived from the underlying `state.stack`.

What about if we add a field/flag to `FrameState` that modifies how `stack_size` is calculated? Then we just ensure stack size is always calculated through `state.stack_len()` and not `state.stack().len()`?

tekknolagi: I think Kokubun is proposing having a builder-type method on `FrameState` that makes a new one, truncates the underlying stack, and then we can use that for `gen_prepare_non_leaf_call`

altxtech: Ok, ok. It's worth a shot.

k0kubun: Yes. As I said, it should be used like `state.with_stack_size(state.stack().len() - args.len() - 1)` in codegen.rs, and I don't see any reason why you can't derive it from `state` and `args` there.

altxtech: Tried doing that refactor, but I running into few failures.

It's a valid refactor, but will need some debugging. Maybe we can do that in separate PR?

<details><summary>Error 1: test_send_exit_unitialed_locals</summary>
<p>

[1/1] TestZJIT#test_send_exit_with_uninitialized_localsOptimized HIR:
fn block in script_shell_encode@/home/andre/repos/ruby/test/ruby/test_zjit.rb:2785:
  v6:BasicObject = SendWithoutBlock v1, :ascii_only?
  CheckInterrupts
  v9:CBool = Test v6
  IfFalse v9, bb1(v0, v1)
  PatchPoint NoEPEscape(block in script_shell_encode)
  CheckInterrupts
  Return v1
bb1(v18:BasicObject, v19:BasicObject):
  v21:StringExact[VALUE(0x7fb68307fba0)] = Const Value(VALUE(0x7fb68307fba0))
  PatchPoint NoEPEscape(block in script_shell_encode)
  v25:BasicObject = SendWithoutBlock v19, :codepoints
  v26:Fixnum[0] = Const Value(0)
  v30:BasicObject = SendWithoutBlock v25, :[], v26
  v34:BasicObject = SendWithoutBlock v21, :%, v30
  CheckInterrupts
  Return v34

 = 0.02 s
  1) Failure:
TestZJIT#test_send_exit_with_uninitialized_locals [/home/andre/repos/ruby/test/ruby/test_zjit.rb:418]:
<"nil"> expected but was
<"1">.


</p>
</details> 

<details><summary>Error 2/3: test_require_rubygems/with_autocompact</summary>
<p>


[1/1] TestZJIT#test_require_rubygems_with_auto_compactOptimized HIR:
fn block in script_shell_encode@/home/andre/repos/ruby/test/ruby/test_zjit.rb:2785:
  v6:BasicObject = SendWithoutBlock v1, :ascii_only?
  CheckInterrupts
  v9:CBool = Test v6
  IfFalse v9, bb1(v0, v1)
  PatchPoint NoEPEscape(block in script_shell_encode)
  CheckInterrupts
  Return v1
bb1(v18:BasicObject, v19:BasicObject):
  v21:StringExact[VALUE(0x7f5dd9d4fbc0)] = Const Value(VALUE(0x7f5dd9d4fbc0))
  PatchPoint NoEPEscape(block in script_shell_encode)
  v25:BasicObject = SendWithoutBlock v19, :codepoints
  v26:Fixnum[0] = Const Value(0)
  v30:BasicObject = SendWithoutBlock v25, :[], v26
  v34:BasicObject = SendWithoutBlock v21, :%, v30
  CheckInterrupts
  Return v34

 = 1.31 s
  1) Failure:
TestZJIT#test_require_rubygems_with_autocompact [/home/andre/repos/ruby/test/ruby/test_zjit.rb:1903]:
exited with status 134
stderr:
```
Allocating a register for VReg(0) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 2 even though it does not live past this index
Allocating a register for VReg(6) at instruction index 40 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 7 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 2 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(4) at instruction index 12 even though it does not live past this index
Allocating a register for VReg(14) at instruction index 37 even though it does not live past this index
Allocating a register for VReg(28) at instruction index 81 even though it does not live past this index
Allocating a register for VReg(29) at instruction index 86 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 12 even though it does not live past this index
Allocating a register for VReg(4) at instruction index 10 even though it does not live past this index
Allocating a register for VReg(15) at instruction index 45 even though it does not live past this index
Allocating a register for VReg(21) at instruction index 52 even though it does not live past this index
spilling VReg is not implemented yet, can't evacuate C_RET_REG on CCall
RegisterSpillOnCCall: gen_iseq failed: block in expand@/home/andre/repos/ruby/rbconfig.rb:307
Allocating a register for VReg(2) at instruction index 4 even though it does not live past this index
Allocating a register for VReg(3) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 13 even though it does not live past this index
Allocating a register for VReg(15) at instruction index 52 even though it does not live past this index
Allocating a register for VReg(23) at instruction index 69 even though it does not live past this index
Allocating a register for VReg(29) at instruction index 84 even though it does not live past this index
Allocating a register for VReg(44) at instruction index 134 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(22) at instruction index 125 even though it does not live past this index
Allocating a register for VReg(24) at instruction index 127 even though it does not live past this index
Allocating a register for VReg(31) at instruction index 155 even though it does not live past this index
Allocating a register for VReg(33) at instruction index 157 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(29) at instruction index 182 even though it does not live past this index
Allocating a register for VReg(34) at instruction index 187 even though it does not live past this index
Allocating a register for VReg(35) at instruction index 195 even though it does not live past this index
Allocating a register for VReg(40) at instruction index 200 even though it does not live past this index
Allocating a register for VReg(50) at instruction index 240 even though it does not live past this index
Allocating a register for VReg(53) at instruction index 258 even though it does not live past this index
Allocating a register for VReg(56) at instruction index 276 even though it does not live past this index
Allocating a register for VReg(59) at instruction index 294 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 2 even though it does not live past this index
Allocating a register for VReg(4) at instruction index 19 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 7 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(12) at instruction index 45 even though it does not live past this index
Allocating a register for VReg(13) at instruction index 47 even though it does not live past this index
Allocating a register for VReg(14) at instruction index 49 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(10) at instruction index 64 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(30) at instruction index 151 even though it does not live past this index
Allocating a register for VReg(45) at instruction index 233 even though it does not live past this index
Allocating a register for VReg(53) at instruction index 279 even though it does not live past this index
Allocating a register for VReg(56) at instruction index 296 even though it does not live past this index
Allocating a register for VReg(64) at instruction index 326 even though it does not live past this index
Allocating a register for VReg(65) at instruction index 332 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 2 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(13) at instruction index 55 even though it does not live past this index
Allocating a register for VReg(15) at instruction index 57 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 5 even though it does not live past this index
Allocating a register for VReg(7) at instruction index 17 even though it does not live past this index
Allocating a register for VReg(11) at instruction index 28 even though it does not live past this index
Allocating a register for VReg(15) at instruction index 39 even though it does not live past this index
Allocating a register for VReg(26) at instruction index 85 even though it does not live past this index
Allocating a register for VReg(27) at instruction index 89 even though it does not live past this index
Allocating a register for VReg(38) at instruction index 121 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 6 even though it does not live past this index
Allocating a register for VReg(0) at instruction index 0 even though it does not live past this index
Allocating a register for VReg(7) at instruction index 42 even though it does not live past this index
Allocating a register for VReg(8) at instruction index 44 even though it does not live past this index
spilling VReg is not implemented yet, can't evacuate C_RET_REG on CCall
RegisterSpillOnCCall: gen_iseq failed: parse@/home/andre/repos/ruby/lib/rubygems/requirement.rb:104
object.c:970: Assertion Failed: rb_obj_is_kind_of:RB_TYPE_P(cl, T_CLASS)
ruby 3.5.0dev (2025-09-16T18:40:59Z altxtech/prepare-c.. 60880bed59) +ZJIT dev +PRISM [x86_64-linux]

c:0014 p:---- s:0109 e:000108 CFUNC  :===
c:0013 p:0006 s:0104 e:000103 METHOD /home/andre/repos/ruby/lib/rubygems/requirement.rb:104
c:0012 p:0007 s:0097 e:000096 BLOCK  /home/andre/repos/ruby/lib/rubygems/requirement.rb:141 [FINISH]
c:0011 p:---- s:0093 e:000092 CFUNC  :map!
c:0010 p:0036 s:0089 e:000088 METHOD /home/andre/repos/ruby/lib/rubygems/requirement.rb:141 [FINISH]
c:0009 p:0010 s:0084 e:000082 METHOD /home/andre/repos/ruby/lib/rubygems/requirement.rb:77
c:0008 p:0199 s:0079 e:000035 CLASS  /home/andre/repos/ruby/lib/rubygems/specification.rb:152
c:0007 p:0050 s:0032 e:000031 TOP    /home/andre/repos/ruby/lib/rubygems/specification.rb:40 [FINISH]
c:0006 p:---- s:0029 e:000028 CFUNC  :require_relative
c:0005 p:0057 s:0024 e:000023 TOP    /home/andre/repos/ruby/lib/rubygems.rb:1434 [FINISH]
c:0004 p:---- s:0017 e:000016 CFUNC  :require
c:0003 p:0012 s:0012 e:000011 BLOCK  -e:2
c:0002 p:0010 s:0009 E:000df8 EVAL   -e:3 [FINISH]
c:0001 p:0000 s:0003 E:0020c0 DUMMY  [FINISH]

-- Ruby level backtrace information ----------------------------------------
-e:3:in '<main>'
-e:2:in 'block in <main>'
-e:2:in 'require'
/home/andre/repos/ruby/lib/rubygems.rb:1434:in '<top (required)>'
/home/andre/repos/ruby/lib/rubygems.rb:1434:in 'require_relative'
/home/andre/repos/ruby/lib/rubygems/specification.rb:40:in '<top (required)>'
/home/andre/repos/ruby/lib/rubygems/specification.rb:152:in '<class:Specification>'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:77:in 'default'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:141:in 'initialize'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:141:in 'map!'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:141:in 'block in initialize'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:104:in 'parse'
/home/andre/repos/ruby/lib/rubygems/requirement.rb:104:in '==='

Total ractor count: 1
Ruby thread count for this ractor: 1

/home/andre/repos/ruby/ruby(rb_print_backtrace+0x5) [0x558bfa56bb56] /home/andre/repos/ruby/vm_dump.c:843
/home/andre/repos/ruby/ruby(rb_vm_bugreport) /home/andre/repos/ruby/vm_dump.c:1175
/home/andre/repos/ruby/ruby(rb_assert_failure_detail+0x150) [0x558bf9e4a873] /home/andre/repos/ruby/error.c:1215
/home/andre/repos/ruby/ruby(RB_FL_TEST_RAW+0x0) [0x558bf9e4a88f] /home/andre/repos/ruby/error.c:1191
/home/andre/repos/ruby/ruby(rb_assert_failure) (null):0
/home/andre/repos/ruby/ruby(RB_FL_ABLE+0x0) [0x558bf9e1b95c] /home/andre/repos/ruby/object.c:970
/home/andre/repos/ruby/ruby(RB_FL_TEST_RAW) ./include/ruby/internal/fl_type.h:465
/home/andre/repos/ruby/ruby(RCLASS_CLASSEXT_TBL) internal/class.h:301
/home/andre/repos/ruby/ruby(RCLASS_EXT_TABLE_LOOKUP_INTERNAL) internal/class.h:371
/home/andre/repos/ruby/ruby(RCLASS_EXT_READABLE_LOOKUP) internal/class.h:383
/home/andre/repos/ruby/ruby(RCLASS_EXT_READABLE) internal/class.h:413
/home/andre/repos/ruby/ruby(RCLASS_EXT_READABLE) internal/class.h:402
/home/andre/repos/ruby/ruby(class_real) /home/andre/repos/ruby/object.c:297
/home/andre/repos/ruby/ruby(rb_obj_class) /home/andre/repos/ruby/object.c:316
/home/andre/repos/ruby/ruby(rb_obj_is_instance_of) /home/andre/repos/ruby/object.c:913
/home/andre/repos/ruby/ruby(vm_call_cfunc_with_frame_+0xe2) [0x558bfa01d492] /home/andre/repos/ruby/vm_insnhelper.c:3873
/home/andre/repos/ruby/ruby(vm_sendish+0x157) [0x558bfa024f87] /home/andre/repos/ruby/vm_insnhelper.c:6087
/home/andre/repos/ruby/ruby(vm_exec_core+0x81) [0x558bfa03e101] /home/andre/repos/ruby/insns.def:900
/home/andre/repos/ruby/ruby(vm_exec_loop+0x4e) [0x558bfa02f41a] /home/andre/repos/ruby/vm.c:2666
/home/andre/repos/ruby/ruby(rb_vm_exec) /home/andre/repos/ruby/vm.c:2642
/home/andre/repos/ruby/ruby(rb_yield+0x16b) [0x558bfa0341db] /home/andre/repos/ruby/vm.c:1720
/home/andre/repos/ruby/ruby(rb_ary_collect_bang+0x4d) [0x558bfa45b86d] /home/andre/repos/ruby/array.c:3690
/home/andre/repos/ruby/ruby(vm_call_cfunc_with_frame_+0xe2) [0x558bfa01d492] /home/andre/repos/ruby/vm_insnhelper.c:3873
/home/andre/repos/ruby/ruby(vm_sendish+0x157) [0x558bfa024f87] /home/andre/repos/ruby/vm_insnhelper.c:6087
/home/andre/repos/ruby/ruby(rb_vm_send+0x3c) [0x558bfa044e9c] /home/andre/repos/ruby/vm_insnhelper.c:6105
[0x558bf5dc7362]

-- Other runtime information -----------------------------------------------

* Loaded script: -e

* Loaded features:

    0 enumerator.so
    1 thread.rb
    2 fiber.so
    3 rational.so
    4 complex.so
    5 pathname.so
    6 ruby2_keywords.rb
    7 set.rb
    8 /home/andre/repos/ruby/.ext/x86_64-linux/enc/encdb.so
    9 /home/andre/repos/ruby/.ext/x86_64-linux/enc/trans/transdb.so
   10 /home/andre/repos/ruby/rbconfig.rb
   11 /home/andre/repos/ruby/lib/rubygems/defaults.rb
   12 /home/andre/repos/ruby/lib/rubygems/deprecate.rb
   13 /home/andre/repos/ruby/lib/rubygems/errors.rb
   14 /home/andre/repos/ruby/lib/rubygems/target_rbconfig.rb
   15 /home/andre/repos/ruby/lib/rubygems/unknown_command_spell_checker.rb
   16 /home/andre/repos/ruby/lib/rubygems/exceptions.rb
   17 /home/andre/repos/ruby/lib/rubygems/basic_specification.rb
   18 /home/andre/repos/ruby/lib/rubygems/stub_specification.rb
   19 /home/andre/repos/ruby/lib/rubygems/platform.rb
   20 /home/andre/repos/ruby/lib/rubygems/specification_record.rb
   21 /home/andre/repos/ruby/lib/rubygems/util/list.rb
   22 /home/andre/repos/ruby/lib/rubygems/version.rb
   23 /home/andre/repos/ruby/lib/rubygems/requirement.rb

* Process memory map:

558bf5dc3000-558bf5dc8000 r-xp 00000000 00:00 0 
558bf5dc8000-558bf9dc3000 ---p 00000000 00:00 0 
558bf9e0f000-558bfa5e3000 r-xp 00000000 00:21 23366582                   /home/andre/repos/ruby/ruby
558bfa5e3000-558bfae4c000 r--p 007d4000 00:21 23366582                   /home/andre/repos/ruby/ruby
558bfae4c000-558bfafad000 r--p 0103d000 00:21 23366582                   /home/andre/repos/ruby/ruby
558bfafad000-558bfafb0000 rw-p 0119e000 00:21 23366582                   /home/andre/repos/ruby/ruby
558bfafb0000-558bfafc4000 rw-p 00000000 00:00 0 
558c15137000-558c15450000 rw-p 00000000 00:00 0                          [heap]
7f9e20680000-7f9e206b0000 rw-p 00000000 00:00 0 
7f9e206bf000-7f9e207b0000 rw-p 00000000 00:00 0 
7f9e207bf000-7f9e207c0000 ---p 00000000 00:00 0 
7f9e207c0000-7f9e20801000 rw-p 00000000 00:00 0 
7f9e20801000-7f9e20802000 ---p 00000000 00:00 0 
7f9e20802000-7f9e20843000 rw-p 00000000 00:00 0 
7f9e20843000-7f9e20844000 ---p 00000000 00:00 0 
7f9e20844000-7f9e20885000 rw-p 00000000 00:00 0 
7f9e20885000-7f9e20886000 ---p 00000000 00:00 0 
7f9e20886000-7f9e208c7000 rw-p 00000000 00:00 0 
7f9e208c7000-7f9e208c8000 ---p 00000000 00:00 0 
7f9e208c8000-7f9e20909000 rw-p 00000000 00:00 0 
7f9e20909000-7f9e2090a000 ---p 00000000 00:00 0 
7f9e2090a000-7f9e2094b000 rw-p 00000000 00:00 0 
7f9e2094b000-7f9e2094c000 ---p 00000000 00:00 0 
7f9e2094c000-7f9e2098d000 rw-p 00000000 00:00 0 
7f9e2098d000-7f9e2098e000 ---p 00000000 00:00 0 
7f9e2098e000-7f9e209cf000 rw-p 00000000 00:00 0 
7f9e209cf000-7f9e209d0000 ---p 00000000 00:00 0 
7f9e209d0000-7f9e20a11000 rw-p 00000000 00:00 0 
7f9e20a11000-7f9e20a12000 ---p 00000000 00:00 0 
7f9e20a12000-7f9e20a53000 rw-p 00000000 00:00 0 
7f9e20a53000-7f9e20a54000 ---p 00000000 00:00 0 
7f9e20a54000-7f9e20a95000 rw-p 00000000 00:00 0 
7f9e20a95000-7f9e20a96000 ---p 00000000 00:00 0 
7f9e20a96000-7f9e20ad7000 rw-p 00000000 00:00 0 
7f9e20ad7000-7f9e20ad8000 ---p 00000000 00:00 0 
7f9e20ad8000-7f9e20b19000 rw-p 00000000 00:00 0 
7f9e20b19000-7f9e20b1a000 ---p 00000000 00:00 0 
7f9e20b1a000-7f9e20b5b000 rw-p 00000000 00:00 0 
7f9e20b5b000-7f9e20b5c000 ---p 00000000 00:00 0 
7f9e20b5c000-7f9e20b9d000 rw-p 00000000 00:00 0 
7f9e20b9d000-7f9e20b9e000 ---p 00000000 00:00 0 
7f9e20b9e000-7f9e20bdf000 rw-p 00000000 00:00 0 
7f9e20bdf000-7f9e20be0000 ---p 00000000 00:00 0 
7f9e20be0000-7f9e20c21000 rw-p 00000000 00:00 0 
7f9e20c21000-7f9e20c22000 ---p 00000000 00:00 0 
7f9e20c22000-7f9e20c63000 rw-p 00000000 00:00 0 
7f9e20c63000-7f9e20c64000 ---p 00000000 00:00 0 
7f9e20c64000-7f9e20ca5000 rw-p 00000000 00:00 0 
7f9e20ca5000-7f9e20ca6000 ---p 00000000 00:00 0 
7f9e20ca6000-7f9e20ce7000 rw-p 00000000 00:00 0 
7f9e20ce7000-7f9e20ce8000 ---p 00000000 00:00 0 
7f9e20ce8000-7f9e20d29000 rw-p 00000000 00:00 0 
7f9e20d29000-7f9e20d2a000 ---p 00000000 00:00 0 
7f9e20d2a000-7f9e20d6b000 rw-p 00000000 00:00 0 
7f9e20d6b000-7f9e20d6c000 ---p 00000000 00:00 0 
7f9e20d6c000-7f9e20dad000 rw-p 00000000 00:00 0 
7f9e20dad000-7f9e20dae000 ---p 00000000 00:00 0 
7f9e20dae000-7f9e20def000 rw-p 00000000 00:00 0 
7f9e20def000-7f9e20df0000 ---p 00000000 00:00 0 
7f9e20df0000-7f9e20e31000 rw-p 00000000 00:00 0 
7f9e20e31000-7f9e20e32000 ---p 00000000 00:00 0 
7f9e20e32000-7f9e20e73000 rw-p 00000000 00:00 0 
7f9e20e73000-7f9e20e74000 ---p 00000000 00:00 0 
7f9e20e74000-7f9e20eb5000 rw-p 00000000 00:00 0 
7f9e20eb5000-7f9e20eb6000 ---p 00000000 00:00 0 
7f9e20eb6000-7f9e20ef7000 rw-p 00000000 00:00 0 
7f9e20ef7000-7f9e20ef8000 ---p 00000000 00:00 0 
7f9e20ef8000-7f9e20f39000 rw-p 00000000 00:00 0 
7f9e20f39000-7f9e20f3a000 ---p 00000000 00:00 0 
7f9e20f3a000-7f9e20f7b000 rw-p 00000000 00:00 0 
7f9e20f7b000-7f9e20f7c000 ---p 00000000 00:00 0 
7f9e20f7c000-7f9e20fbd000 rw-p 00000000 00:00 0 
7f9e20fbd000-7f9e20fbe000 ---p 00000000 00:00 0 
7f9e20fbe000-7f9e20fff000 rw-p 00000000 00:00 0 
7f9e20fff000-7f9e21000000 ---p 00000000 00:00 0 
7f9e21000000-7f9e21800000 rw-p 00000000 00:00 0 
7f9e21800000-7f9e3ac00000 rw-p 00000000 00:00 0 
7f9e3ac1f000-7f9e3acc0000 rw-p 00000000 00:00 0 
7f9e3accf000-7f9e3ad00000 rw-p 00000000 00:00 0 
7f9e3ad0a000-7f9e3ae0e000 rw-p 00000000 00:00 0 
7f9e3ae0e000-7f9e3af7d000 r-xp 00000000 00:21 13944590                   /usr/lib64/libc.so.6
7f9e3af7d000-7f9e3aff2000 r--p 0016f000 00:21 13944590                   /usr/lib64/libc.so.6
7f9e3aff2000-7f9e3aff6000 r--p 001e3000 00:21 13944590                   /usr/lib64/libc.so.6
7f9e3aff6000-7f9e3aff8000 rw-p 001e7000 00:21 13944590                   /usr/lib64/libc.so.6
7f9e3aff8000-7f9e3b002000 rw-p 00000000 00:00 0 
7f9e3b002000-7f9e3b025000 r-xp 00000000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f9e3b025000-7f9e3b02c000 r--p 00023000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f9e3b02c000-7f9e3b02d000 r--p 0002a000 00:21 13673671                   /usr/lib64/libgcc_s-15-20250808.so.1
7f9e3b02d000-7f9e3b02e000 rw-p 00000000 00:00 0 
7f9e3b02e000-7f9e3b0aa000 r-xp 00000000 00:21 13944592                   /usr/lib64/libm.so.6
7f9e3b0aa000-7f9e3b11a000 r--p 0007c000 00:21 13944592                   /usr/lib64/libm.so.6
7f9e3b11a000-7f9e3b11b000 r--p 000eb000 00:21 13944592                   /usr/lib64/libm.so.6
7f9e3b11b000-7f9e3b11c000 rw-p 000ec000 00:21 13944592                   /usr/lib64/libm.so.6
7f9e3b11c000-7f9e3b135000 r-xp 00000000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f9e3b135000-7f9e3b14f000 r--p 00019000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f9e3b14f000-7f9e3b150000 r--p 00032000 00:21 8839375                    /usr/lib64/libcrypt.so.2.0.0
7f9e3b150000-7f9e3b151000 rw-p 00000000 00:00 0 
7f9e3b151000-7f9e3b168000 r-xp 00000000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f9e3b168000-7f9e3b172000 r--p 00017000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f9e3b172000-7f9e3b173000 r--p 00020000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f9e3b173000-7f9e3b174000 rw-p 00021000 00:21 13673695                   /usr/lib64/libz.so.1.3.1.zlib-ng
7f9e3b185000-7f9e3b187000 r-xp 00000000 00:21 23365589                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/trans/transdb.so
7f9e3b187000-7f9e3b188000 r--p 00002000 00:21 23365589                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/trans/transdb.so
7f9e3b188000-7f9e3b189000 r--p 00002000 00:21 23365589                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/trans/transdb.so
7f9e3b189000-7f9e3b18a000 rw-p 00003000 00:21 23365589                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/trans/transdb.so
7f9e3b18a000-7f9e3b18b000 r-xp 00000000 00:21 23365549                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/encdb.so
7f9e3b18b000-7f9e3b18c000 r--p 00001000 00:21 23365549                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/encdb.so
7f9e3b18c000-7f9e3b18d000 r--p 00001000 00:21 23365549                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/encdb.so
7f9e3b18d000-7f9e3b18e000 rw-p 00002000 00:21 23365549                   /home/andre/repos/ruby/.ext/x86_64-linux/enc/encdb.so
7f9e3b18e000-7f9e3b18f000 rw-p 00000000 00:00 0 
7f9e3b18f000-7f9e3b190000 rw-p 00000000 00:00 0 
7f9e3b190000-7f9e3b192000 rw-p 00000000 00:00 0 
7f9e3b192000-7f9e3b196000 r--p 00000000 00:00 0                          [vvar]
7f9e3b196000-7f9e3b198000 r--p 00000000 00:00 0                          [vvar_vclock]
7f9e3b198000-7f9e3b19a000 r-xp 00000000 00:00 0                          [vdso]
7f9e3b19a000-7f9e3b1c3000 r-xp 00000000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f9e3b1c3000-7f9e3b1ce000 r--p 00029000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f9e3b1ce000-7f9e3b1d0000 r--p 00034000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f9e3b1d0000-7f9e3b1d1000 rw-p 00036000 00:21 13944587                   /usr/lib64/ld-linux-x86-64.so.2
7f9e3b1d1000-7f9e3b1d2000 rw-p 00000000 00:00 0 
7ffc29118000-7ffc2913a000 rw-p 00000000 00:00 0                          [stack]
ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]

</p>
</details>

Comment (k0kubun) [reply zjit/src/codegen.rs]: okay but can you show us the diff you tried?

Comment (k0kubun) [reply zjit/src/codegen.rs]: uhh okay I get it. we do need to spill AND have SP lower than that. The callee loads them as arguments.

> Maybe we can do that in separate PR?

You're effectively reverting https://github.com/ruby/ruby/pull/14427, and it should be the scope of this PR to minimize the impact of it. It's kinda inevitable with your design (unless you let `gen_prepare_non_leaf_call` clear the `leaf_ccall_stack_size` flag, which I'm actually fine with if that allows us to get rid of `gen_save_pc`), but I don't want the use of `gen_save_pc` to be spread outside of `gen_prepare*` helpers.

Can you create a variant of `gen_prepare_non_leaf_call` that takes both `state` and `stack_size` (like `gen_prepare_non_leaf_call_with_stack_size`?), let `gen_prepare_non_leaf_call` call `stack_size` with `state.stack.len()`, and use the one with `state.stack().len() - args.len() - 1` as `stack_size`?

Comment (altxtech) [reply zjit/src/codegen.rs]: I works if I add an extra `gen_spill_stack` after `gen_prepare_non_leaf_call`.  

```rust
let prep_state = state.with_stack_size(state.stack().len() - args.len() - 1);
gen_prepare_non_leaf_call(jit, asm, &prep_state);
gen_spill_stack(jit, asm, state); // still needed
```  

That solves the correctness issue, but it feels too hacky for a clean refactor. Probably better to keep the current special-case SP math in `gen_send_without_block_direct` and do a bigger clean up in a follow up PR.

altxtech: Diff I tried
```diff
diff --git a/zjit/src/codegen.rs b/zjit/src/codegen.rs
index 7a481eb496..e4c0fce2e6 100644
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1038,12 +1038,8 @@ fn gen_send_without_block_direct(
     asm.cmp(CFP, stack_limit);
     asm.jbe(side_exit(jit, state, StackOverflow));
 
-    // Save cfp->pc and cfp->sp for the caller frame
-    gen_save_pc(asm, state);
-    // Special SP math. Can't use gen_prepare_non_leaf_call
-    gen_save_sp(asm, state.stack().len() - args.len() - 1); // -1 for receiver
-
-    gen_spill_locals(jit, asm, state);
+    let prep_state = state.with_stack_size(state.stack().len() - args.len() -1);
+    gen_prepare_non_leaf_call(jit, asm, &prep_state);
     gen_spill_stack(jit, asm, state);
 
     // Set up the new frame
diff --git a/zjit/src/hir.rs b/zjit/src/hir.rs
index 1a268e62da..6e953b0d87 100644
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2775,6 +2775,12 @@ impl FrameState {
         state.locals.clear();
         state
     }
+
+    pub fn with_stack_size(&self, size: usize) -> Self {
+        let mut state = self.clone();
+        state.stack.truncate(size);
+        state
+    }
 }
```

k0kubun: OK, so I'm not asking you to do either of these now.

Please forget all past suggestions. Instead of having the current `gen_save_pc` and `gen_prepare_call_with_gc`, can we just have the following `gen_prepare_call_with_gc` and `gen_prepare_leaf_call_with_gc`?

```rs
fn gen_prepare_leaf_call_with_gc(asm: &mut Assembler, state: &FrameState) {
    gen_prepare_call_with_gc(asm, state, true);
}

fn gen_prepare_call_with_gc(asm: &mut Assembler, state: &FrameState, leaf: bool) {
    let opcode: usize = state.get_opcode().try_into().unwrap();
    let next_pc: *const VALUE = unsafe { state.pc.offset(insn_len(opcode) as isize) };

    asm_comment!(asm, "save PC to CFP");
    asm.mov(Opnd::mem(64, CFP, RUBY_OFFSET_CFP_PC), Opnd::const_ptr(next_pc));

    if leaf {
        asm.expect_leaf_ccall(state.stack_size());
    }
}
```

Nothing's really hacky about it, and it's very relevant to this PR introducing `gen_save_pc` (reverting https://github.com/ruby/ruby/pull/14427).

k0kubun: Then you should be able to replace this `gen_save_pc(asm, state)` with `gen_prepare_call_with_gc(asm, state, false)`, leaving other calls like `gen_save_sp` and `gen_spill_locals`/`gen_spill_stack`.

Nothing about "call with GC" really implies leafness, so if we start verifying its leafness in the helper, it's more reasonable to call it `leaf` or pass a flag to indicate that, which is what I suggested above.

altxtech: Done

k0kubun (approved):
Thanks :pray:

PR #14556, opened 2025-09-15
Remove "assigned but unused variable" warning

etiennebarrie:
```console
$ make test/ruby/test_module.rb RUBYOPT=-w >/dev/null
test/ruby/test_module.rb:420: warning: assigned but unused variable - new_mod
```

cc @byroot


test/ruby/test_module.rb
```diff
@@ -417,7 +417,7 @@ def test_initialize_copy
     klass = Class.new { include mod }
     instance = klass.new
     assert_equal(:first, instance.foo)
-    new_mod = Module.new { define_method(:foo) { :second } }
```
byroot, line 420: I suspect the intent is to keep a reference on that module. And the `4.times { GC.start }` is to make it reach old gen.

So the proper fix is probably to prefix the variable with `_` to avoid the warning.

nobu: Another fix would be adding `new_mod = new_mod`.

etiennebarrie: I did this but looking at 5c7dfe85a1dc49334e2828791f0ade42eee662db I think we want to remove the test entirely instead, because it was testing the `TypeError` that was raised on `initialize_copy`, it wasn't just related to making the module oldgen.
The repro script from https://bugs.ruby-lang.org/issues/17048 doesn't crash anymore, but some of the "broken semantics" have reappeared regarding constants.

However other semantics have been fixed from https://bugs.ruby-lang.org/issues/21538, namely the hierarchy of `initialize_{copy,clone}` being called.

So I think this test should be removed now that we don't expect a TypeError anymore, WDYT?

byroot: Let's ask @jeremyevans what the intent was here.

jeremyevans: I agree with @etiennebarrie we could safely remove `test_initialize_copy`. It made sense when the logic was in `Module#initialize_copy`, but now that the logic is part of `dup`/`clone`, before calling `initialize_dup`/`initialize_clone`, it serves no purpose.

etiennebarrie:
Updated the commit, we're still removing the warning by removing the whole test now.

PR #14563, opened 2025-09-15
[DOC] More tweaks for String#hex

BurdetteLamar:


string.c
```diff
@@ -10713,22 +10713,55 @@ rb_str_scan(VALUE str, VALUE pat)
 
 
 /*
- *  call-seq:
- *    hex -> integer
+ *  call-seq: hex -> integer
```
peterzhu2118: Shouldn't the call-seq be two lines?

string.c
```diff
@@ -10713,22 +10713,55 @@ rb_str_scan(VALUE str, VALUE pat)
...
+ *    '0d999'.hex   # => 55705
  *
- *  Returns zero if there is no such leading substring:
+ *  Also unlike #oct, this method does not interpret
+ *  a substring starting with <tt>'0o'</tt> or <tt>'0O'</tt> as octal;
+ *  instead it interprets <tt>'o'</tt> or <tt>'O'</tt> as simply "not hexadecimal":
  *
- *    'zzz'.hex # => 0
+ *    '0o777'.hex   # => 0
+ *    '0O777'.hex   # => 0
```
peterzhu2118: IMO this is too wordy just to explain that "unlike #oct, it only interprets hexadecimal". I think it's surprising that #oct supports all the other formats and which is why we have all that description, but I think that level of description is unnecessary here.

BurdetteLamar: Simplified.

string.c

BurdetteLamar: @peterzhu2118, missed this comment before.  Sorry.

BurdetteLamar:
@peterzhu2118, I think this ready to go?

PR #14564, opened 2025-09-15
GitHub CI: JITs: Dump RUBY_CRASH_REPORT on failure

XrXr:
- **Unset RUBY_CRASH_REPORT in tests that crash on purpose**
- **GitHub CI: JITs: Dump RUBY_CRASH_REPORT on failure**

Example dumps from a failure:
https://github.com/Shopify/ruby/actions/runs/17746006938/job/50431200179


k0kubun (approved):
Thanks!

.github/workflows/yjit-macos.yml
```diff
@@ -166,6 +168,10 @@ jobs:
         if: ${{ matrix.test_task == 'check' && matrix.skipped_tests }}
         continue-on-error: ${{ matrix.continue-on-skipped_tests || false }}
 
+      - if: ${{ failure() }}
+        continue-on-error: true
+        run: tail --verbose --lines=+1 rb_crash_*.txt
```
aidenfoxivey, line 173: Out of curiosity, if using `tail` to view the whole file (I think this is what the `--lines=+1` flag means), then why not just use `cat`.

aidenfoxivey (approved):
Had a minor question. Looks good to me!

.github/workflows/yjit-macos.yml
nobu: Probably, I guess because `tail` inserts the headers like `==> <filename> <==`.

PR #14565, opened 2025-09-16
ZJIT: Revert VM_CALL_ARGS_SPLAT and VM_CALL_KWARG support

k0kubun:
This PR partially reverts https://github.com/ruby/ruby/pull/14518. 

`VM_CALL_ARGS_SPLAT` breaks binarytrees, graphql, graphql-native, hexapdf, mail, and ruby-lsp, and `VM_CALL_KWARG` breaks rubocop, shipit, lobsters. So it filters out those call types for now.

They are currently a relatively small source of exits, so the counters are renamed from the `unhandled_call_` to the global `exit_` counters.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -450,7 +450,7 @@ pub enum SideExitReason {
     UnknownSpecialVariable(u64),
     UnhandledHIRInsn(InsnId),
     UnhandledYARVInsn(u32),
-    UnhandledTailCall,
+    UnhandledCallType(CallType),
     FixnumAddOverflow,
     FixnumSubOverflow,
     FixnumMultOverflow,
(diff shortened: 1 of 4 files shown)


XrXr (approved):
Thanks for extracting the tests out of the failures!

zjit/src/hir.rs
```diff
@@ -3600,10 +3604,10 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let blockiseq: IseqPtr = get_arg(pc, 1).as_iseq();
                     let call_info = unsafe { rb_get_call_data_ci(cd) };
                     let flags = unsafe { rb_vm_ci_flag(call_info) };
-                    if is_tailcall(flags) {
+                    if let Err(call_type) = unhandled_call_type(flags) {
                         // Can't handle tailcall; side-exit into the interpreter
                         let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
-                        fun.push_insn(block, Insn::SideExit { state: exit_id, reason: SideExitReason::UnhandledTailCall });
+                        fun.push_insn(block, Insn::SideExit { state: exit_id, reason: SideExitReason::UnhandledCallType(call_type) });
```
tekknolagi, line 3610: When we eventually re-enable for send/invokesuper/invokeblock we should add tests for all of these kinds of calls

PR #14575, opened 2025-09-16
ZJIT: Support variadic C calls

st0012:
This reduces the `dynamic_send_count` in `liquid-render` by ~20%

**Before**

```
Top-4 dynamic send types (100.0% of total 37,691,693):
  send_without_block: 31,176,518 (82.7%)
                send:  6,207,698 (16.5%)
         invokeblock:    274,932 ( 0.7%)
         invokesuper:     32,545 ( 0.1%)
Top-4 send fallback def_types (100.0% of total 15,891,032):
      cfunc: 15,733,625 (99.0%)
  optimized:    144,322 ( 0.9%)
    attrset:     13,065 ( 0.1%)
      alias:         20 ( 0.0%)
dynamic_send_count:  37,691,693
```

**After**

```
Top-4 dynamic send types (100.0% of total 30,161,408):
  send_without_block: 23,358,749 (77.4%)
                send:  6,482,940 (21.5%)
         invokeblock:    286,934 ( 1.0%)
         invokesuper:     32,785 ( 0.1%)
Top-4 send fallback def_types (100.0% of total 7,411,863):
      cfunc: 7,248,096 (97.8%)
  optimized:   150,682 ( 2.0%)
    attrset:    13,065 ( 0.2%)
      alias:        20 ( 0.0%)
dynamic_send_count:  30,161,408
```

It adds a new `CCallVariadic` HIR for C methods with signature `func(int argc, VALUE *argv, VALUE recv)`.
If my implementation is not too far off, I think we may be able to split it into few HIRs:
- One for setting up frame
- One for `gen_push_opnds`
- `CCall`
- One for `gen_pop_opnds`
- One for clean up

Is this the direction we're expecting? 🤔 
If that's the case, I prefer to merge the implementation with one HIR, and open an issue for the split as a todo.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2100,10 +2100,14 @@ impl Function {
                 }
                 // Variadic method
                 -1 => {
+                    if unsafe { rb_zjit_method_tracing_currently_enabled() } {
+                        return Err(());
+                    }
                     // The method gets a pointer to the first argument
                     // func(int argc, VALUE *argv, VALUE recv)
                     let ci_flags = unsafe { vm_ci_flag(call_info) };
                     if ci_flags & VM_CALL_ARGS_SIMPLE != 0 {
+                        fun.push_insn(block, Insn::PatchPoint { invariant: Invariant::NoTracePoint, state });
                         fun.push_insn(block, Insn::PatchPoint {
                             invariant: Invariant::MethodRedefined {
                                 klass: recv_class,
@@ -6958,9 +6962,9 @@ mod opt_tests {
           v4:StringExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
           v6:StringExact = StringCopy v4
           PatchPoint MethodRedefined(Object@0x1008, puts@0x1010, cme:0x1018)
-          v15:BasicObject = CCallVariadic puts@0x1040, v0, v6
+          v16:BasicObject = CCallVariadic puts@0x1040, v0, v6
           CheckInterrupts
-          Return v15
+          Return v16
         ");
     }
 
--- a/zjit.c
+++ b/zjit.c
@@ -159,6 +159,23 @@ rb_zjit_defined_ivar(VALUE obj, ID id, VALUE pushval)
 }
 
 bool
+rb_zjit_method_tracing_currently_enabled(void)
+{
+    rb_event_flag_t tracing_events;
+    if (rb_multi_ractor_p()) {
+        tracing_events = ruby_vm_event_enabled_global_flags;
+    }
+    else {
+        // At the time of writing, events are never removed from
+        // ruby_vm_event_enabled_global_flags so always checking using it would
+        // mean we don't compile even after tracing is disabled.
+        tracing_events = rb_ec_ractor_hooks(GET_EC())->events;
+    }
+
+    return tracing_events & (RUBY_EVENT_C_CALL | RUBY_EVENT_C_RETURN);
+}
+
+bool
 rb_zjit_insn_leaf(int insn, const VALUE *opes)
 {
     return insn_leaf(insn, opes);
--- a/zjit/bindgen/src/main.rs
+++ b/zjit/bindgen/src/main.rs
@@ -274,6 +274,7 @@ fn main() {
         .allowlist_function("rb_zjit_local_id")
         .allowlist_function("rb_set_cfp_(pc|sp)")
         .allowlist_function("rb_c_method_tracing_currently_enabled")
+        .allowlist_function("rb_zjit_method_tracing_currently_enabled")
         .allowlist_function("rb_full_cfunc_return")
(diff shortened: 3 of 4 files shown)


zjit/src/codegen.rs
```diff
@@ -640,6 +643,56 @@ fn gen_ccall(asm: &mut Assembler, cfun: *const u8, args: Vec<Opnd>) -> lir::Opnd
...
+    asm_comment!(asm, "call variadic cfunc: argc={}", argc);
+
+    let argv_ptr = gen_push_opnds(jit, asm, &args);
+    let result = asm.ccall(cfun, vec![argc.into(), argv_ptr, recv]);
+    gen_pop_opnds(asm, &args);
+
+    asm_comment!(asm, "pop C frame");
+    let new_cfp = asm.add(CFP, RUBY_SIZEOF_CONTROL_FRAME.into());
+    asm.mov(CFP, new_cfp);
+    asm.store(Opnd::mem(64, EC, RUBY_OFFSET_EC_CFP), CFP);
```
st0012, line 681: Why does C frame needs to be popped but `gen_send_without_block_direct` doesn't seem to do it 🤔

zjit/src/hir.rs
```diff
@@ -2330,6 +2378,7 @@ impl Function {
             }
             &Insn::Send { recv, ref args, state, .. }
             | &Insn::SendWithoutBlock { recv, ref args, state, .. }
+            | &Insn::CCallVariadic { recv,ref args, state, .. }
```
```suggestion
            | &Insn::CCallVariadic { recv, ref args, state, .. }
```

k0kubun (approved):

zjit/src/codegen.rs
k0kubun: because the callee ISEQ runs `Insn::Return` whereas the implementation of a C method doesn't pop the frame by itself.

st0012: I see, thanks!

k0kubun:
I guess we're missing some deoptimization for C calls? 

https://github.com/ruby/ruby/actions/runs/17810760232/job/50633769374?pr=14575

```
    1) Failure:
  TestRubyOptimization#test_trace_optimized_methods [/home/runner/work/ruby/ruby/src/test/ruby/test_optimization.rb:242]:
  [ruby-core:87638].
  <[:-@,
   :max,
   :min,
   :+,
   :-,
   :*,
   :/,
   :%,
   :==,
   :<,
   :<=,
   :>,
   :>=,
   :<<,
   :&,
   :|,
   :[],
   :[]=,
   :length,
   :empty?,
   :nil?,
   :succ,
   :!,
   :=~]> expected but was
  <[:-@,
   :+,
   :-,
   :*,
   :/,
   :%,
   :==,
   :<,
   :<=,
   :>,
   :>=,
   :<<,
   :&,
   :|,
   :[],
   :[]=,
   :length,
   :empty?,
   :nil?,
   :succ,
   :!,
   :=~]>.
```

Are we doing something similar to YJIT's `if c_method_tracing_currently_enabled(jit)` check and `record_global_inval_patch(asm, CodegenGlobals::get_outline_full_cfunc_return_pos())`?

st0012:
Looks like I need to bail the codegen when TracePoint events `c_call` or `c_return` is enabled. I'll see how I can pass `ec` in to call `rb_c_method_tracing_currently_enabled`

zjit/src/codegen.rs
```diff
@@ -397,6 +405,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         &Insn::GuardBlockParamProxy { level, state } => no_output!(gen_guard_block_param_proxy(jit, asm, level, &function.frame_state(state))),
         Insn::PatchPoint { invariant, state } => no_output!(gen_patch_point(jit, asm, invariant, &function.frame_state(*state))),
         Insn::CCall { cfun, args, name: _, return_type: _, elidable: _ } => gen_ccall(asm, *cfun, opnds!(args)),
+        Insn::CCallVariadic { state, .. } if unsafe { rb_c_method_tracing_currently_enabled(jit.ec) } => return Err(*state),
```
tekknolagi: This seems like a *ton* of plumbing and code changes for what seems to me like it should be a quick change in the optimizer:

* Check if method tracing is enabled. If so, don't optimize to `CCallVariadic`. Add a TODO to add method tracing opcodes later.
* Add a PatchPoint that method tracing is not enabled (if this does not already exist).

Instead of plumbing `ec` through the entire JIT, let's fetch it using `GET_EC` in a ZJIT helper function or something.

st0012: > to add method tracing opcodes later

Is it for generate optimized code with tracing support? I thought we don't want to compile anything after TracePoint has been enabled?

tekknolagi: Then let's not do a TODO

st0012:
@tekknolagi I think this is ready to merge

PR #14577, opened 2025-09-17
Sync default gems

hsbt:


etiennebarrie:
Hello! We can also revert caa5d8cdd7483647013af5e3d1701b0ed3ec8785 now. Thank you and sorry for the issues with ruby/json.

PR #14579, opened 2025-09-17
ZJIT: Specialize ObjectAlloc with known class pointer

tekknolagi:
This has fewer effects (can be elided!) and will eventually get better
codegen, too.


aidenfoxivey (approved):
I read through it all and it looks good to me!

zjit/src/hir.rs
```diff
@@ -1850,6 +1856,17 @@ impl Function {
                             self.push_insn_id(block, insn_id);
                         }
                     }
+                    Insn::ObjectAlloc { val, state } => {
```
k0kubun, line 1920: It feels weird that this rewrite lives in `optimize_direct_sends` when it's also explained as "Rewrite SendWithoutBlock opcodes into SendWithoutBlockDirect opcodes". I think the function name should be something like `specialize_insns` now?

zjit/src/codegen.rs
```diff
@@ -1231,12 +1232,16 @@ fn gen_new_range_fixnum(
...
 fn gen_object_alloc(jit: &JITState, asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
-    // TODO: this is leaf in the vast majority of cases,
-    // Should specialize to avoid `gen_prepare_non_leaf_call` (Shopify#747)
     gen_prepare_non_leaf_call(jit, asm, state);
     asm_ccall!(asm, rb_obj_alloc, val)
 }
 
+fn gen_object_alloc_class(asm: &mut Assembler, class: VALUE, state: &FrameState) -> lir::Opnd {
+    // TODO(max): directly call `class_call_alloc_func`
+    gen_prepare_call_with_gc(asm, state, /*leaf=*/true);
```
k0kubun: Why not just `gen_prepare_leaf_call_with_gc` here, especially when you had to add the `/*leaf=*/` comment? I didn't expect the use of `gen_prepare_call_with_gc` to be spread outside `gen_prepare_leaf_call_with_gc` or ones that need `leaf=false`.

zjit/src/codegen.rs
```diff
@@ -1231,12 +1232,16 @@ fn gen_new_range_fixnum(
 }
 
 fn gen_object_alloc(jit: &JITState, asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
-    // TODO: this is leaf in the vast majority of cases,
-    // Should specialize to avoid `gen_prepare_non_leaf_call` (Shopify#747)
     gen_prepare_non_leaf_call(jit, asm, state);
```
k0kubun, line 1239: The difference between `gen_object_alloc` and `gen_object_alloc_class` seems super unobvious now. `gen_prepare_non_leaf_call` should generally have a comment that explains why it's not leaf.

In this case, it's also nice to explain why `rb_obj_alloc` on `class: VALUE` never (raises an exception on the `T_CLASS` check or) triggers a custom allocator (I still don't remember what condition lets the class use a custom allocator).

zjit/src/codegen.rs
tekknolagi: oh, yes, i massively prefer that. thank you

vm_insnhelper.c
```diff
@@ -2353,6 +2353,11 @@ vm_search_method(VALUE cd_owner, struct rb_call_data *cd, VALUE recv)
     return vm_cc_cme(cc);
 }
 
+const struct rb_callable_method_entry_struct *
+rb_zjit_vm_search_method(VALUE cd_owner, struct rb_call_data *cd, VALUE recv) {
```
k0kubun: It's just not how CRuby code is styled

```suggestion
rb_zjit_vm_search_method(VALUE cd_owner, struct rb_call_data *cd, VALUE recv)
{
```

zjit/src/codegen.rs
```diff
@@ -1231,12 +1232,19 @@ fn gen_new_range_fixnum(
...
 fn gen_object_alloc(jit: &JITState, asm: &mut Assembler, val: lir::Opnd, state: &FrameState) -> lir::Opnd {
-    // TODO: this is leaf in the vast majority of cases,
-    // Should specialize to avoid `gen_prepare_non_leaf_call` (Shopify#747)
+    // Allocating an object from an unknown class is non-leaf; see doc for `ObjectAlloc`.
     gen_prepare_non_leaf_call(jit, asm, state);
     asm_ccall!(asm, rb_obj_alloc, val)
 }
 
+fn gen_object_alloc_class(asm: &mut Assembler, class: VALUE, state: &FrameState) -> lir::Opnd {
+    // Allocating an object for a known class with default allocator is leaf; see doc for
```
k0kubun, line 1244: I missed the code that does the default allocator check. Which part of the HIR generator is checking if this class uses the default allocator?

k0kubun:
Could you add test cases for method redefinition and custom allocators? Since we haven't added the side-by-side thing in the HIR tests yet, I think we need test cases in `test_zjit.rb` for testing the behavior.

vm_insnhelper.c
k0kubun: Well, I guess I didn't need to make a comment because CI points it out https://github.com/ruby/ruby/actions/runs/17805392000/job/50615809556?pr=14579

st0012 (approved):

tekknolagi:
So we already have

```ruby
  def test_opt_new_does_not_push_frame
    assert_compiles 'nil', %q{
      class Foo
        attr_reader :backtrace

        def initialize
          @backtrace = caller
        end
      end
      def test = Foo.new

      foo = test
      foo.backtrace.find do |frame|
        frame.include?('Class#new')
      end
    }, insns: [:opt_new]
  end

  def test_opt_new_with_redefinition
    assert_compiles '"foo"', %q{
      class Foo
        def self.new = "foo"

        def initialize = raise("unreachable")
      end
      def test = Foo.new

      test
    }, insns: [:opt_new]
  end
```

I will add for method redefinition

k0kubun:
What about custom allocators? Could we reproduce this particular raise https://github.com/ruby/ruby/pull/14553#discussion_r2349864922 (and rescue it) in `test_zjit.rb`? The leafness check of https://github.com/ruby/ruby/pull/14553 would then crash the test if we call `rb_class_alloc` with `gen_prepare_leaf_call_with_gc` in that case.

I still don't get how this HIR generator avoids specializing `ObjectAlloc` into `ObjectAllocClass` for classes that have the custom allocator like https://github.com/ruby/ruby/pull/14553#discussion_r2349864922, so I would appreciate a test for it. `make check` might test it but it may not when it runs out of the code region.

vm_insnhelper.c
k0kubun: You all really like to mark a conversation as resolved when the diff is still not marked "Outdated", but you haven't fixed it. My previous reply didn't mean you don't need to fix it. It's failing the CI https://github.com/ruby/ruby/actions/runs/17805392000/job/50615809556?pr=14579.

tekknolagi:
Yeah I am trying to figure that one out. I can't require `"digest"` in test_zjit.rb, I think because we run with miniruby or something

EDIT: Figured it out

tekknolagi:
```
  1) Failure:
TestZJIT#test_opt_new_with_custom_allocator [/Users/emacs/Documents/code/ruby/test/ruby/test_zjit.rb:867]:
failed to get iseq disassembly.
<"YARVInstructionSequence/SimpleDataFormat"> expected but was
<nil>.
```

zjit/src/codegen.rs
tekknolagi: I added it in the optimizer

tekknolagi:
I thought having a custom allocator would require having a custom new but it seems like it does not. Added a special check for custom allocator

vm_insnhelper.c
tekknolagi: Sorry, thought I fixed this locally but accidentally rebased it out.

tekknolagi: Feel free to resolve when you see fit.

zjit/src/codegen.rs
```diff
@@ -388,7 +389,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::FixnumAnd { left, right } => gen_fixnum_and(asm, opnd!(left), opnd!(right)),
         Insn::FixnumOr { left, right } => gen_fixnum_or(asm, opnd!(left), opnd!(right)),
         Insn::IsNil { val } => gen_isnil(asm, opnd!(val)),
-        &Insn::IsMethodCfunc { val, cd, cfunc } => gen_is_method_cfunc(jit, asm, opnd!(val), cd, cfunc),
+        &Insn::IsMethodCfunc { val, cd, cfunc, state: _ } => gen_is_method_cfunc(jit, asm, opnd!(val), cd, cfunc),
```
k0kubun, line 392: What is `state` added for if it's not being used?

k0kubun (approved):

zjit/src/codegen.rs
tekknolagi: It's used if rewriting to PatchPoint

tekknolagi:
Thanks for shepherding this along.

PR #14581, opened 2025-09-17
Improve textmode IO read performance

YO4:
This PR improves IO read performance opened with mode_enc parameter "rt".
That is, it applies when only universal newline conversion is used without encoding conversion.

The benefit varies depending on the reading method used.
It is also affected by the density of CRLF in the source.

### DIFF
--- a/include/ruby/io.h
+++ b/include/ruby/io.h
@@ -119,6 +119,9 @@ struct rb_io_internal_buffer {
 
     /** Designed capacity of the buffer. */
     int capa;
+
+    /** Boundary of read and unget'ed data, for cbuf. Otherwise, not used. */
+    int off_unget;
 } RBIMPL_ATTR_PACKED_STRUCT_UNALIGNED_END();
 
 /** @alias{rb_io_buffer_t} */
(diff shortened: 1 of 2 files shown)


YO4:
Benchmark script I use is here.
```ruby
require 'benchmark/ips'
require 'tempfile'

str1k = "a" * 1024
str26 = (?a..?z).to_a.join + "\r\n"

no_newline = File.join(Dir.tmpdir, "no_lf")
sparse_newline = File.join(Dir.tmpdir, "sparse_lf")
dense_newline = File.join(Dir.tmpdir, "dense_lf")

open(no_newline, "wb")     { |f|  200.times { f.write str1k } }
open(sparse_newline, "wb") { |f| 3000.times { f.write str26 } }
open(dense_newline, "wb")  { |f| 5000.times { f.write "\r\n" } }

{"no" => no_newline, "sparse" => sparse_newline, "dense" => dense_newline}.each do |name, src|
  Benchmark.ips do |x|

    ["r", "rt", "rb"].each do |mode|
      x.report("#{name}_lf:#{mode}") do |times|
        target = open(src, mode)
        i = 0;
        while i < times
          target.readlines
          target.rewind
          i += 1
        end
        target.close
      end
    end

    x.compare!
  end
end

File.unlink no_newline
File.unlink sparse_newline
File.unlink dense_newline
```

Modified banchmark code are used to cbuf invalidate case .
core part changed as below:
```ruby
        while i < times
          target.getc
          target.wait_readable
          target.getc
          target.rewind
          i += 1
        end
```

PR #14583, opened 2025-09-17
ZJIT: Split Insn::Const from Insn::GetBlockParamProxy

k0kubun:
https://github.com/ruby/ruby/pull/14483#discussion_r2349638173

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -5898,7 +5898,7 @@ mod tests {
           v10:BasicObject = InvokeBuiltin dir_s_open, v0, v1, v2
           PatchPoint NoEPEscape(open)
           GuardBlockParamProxy l0
-          v17:BasicObject[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+          v17:BasicObject[BlockParamProxy] = Const Value(VALUE(0x1000))
           CheckInterrupts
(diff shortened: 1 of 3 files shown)


zjit/src/hir.rs
```diff
@@ -8112,11 +8114,12 @@ mod opt_tests {
         assert_snapshot!(hir_string("test"), @r"
         fn test@<compiled>:2:
         bb0(v0:BasicObject, v1:BasicObject):
-          v6:BasicObject = GetBlockParamProxy l0
-          v8:BasicObject = Send v0, 0x1000, :tap, v6
-          v9:BasicObject = GetLocal l0, EP@3
+          GuardBlockParamProxy l0
+          v7:BasicObject[VALUE(0x1000)] = Const Value(VALUE(0x1000))
```
tekknolagi: you can add

```
Specialization::Object(val) if val == unsafe { rb_block_param_proxy } => write!(f, "[BlockParamProxy]"),
```

to `write_spec` in hir_type/mod.rs next to the FrozenCore stuff if you like

k0kubun: :+1: https://github.com/ruby/ruby/pull/14583/commits/c4bd355b357435ea6db90ecb5a75c7c011158399

PR #14589, opened 2025-09-18
Use `ec->interrupt_mask` to prevent interrupts. (#14588)

ioquatix:
Disallow pending interrupts to be checked during `FiberScheduler#unblock`.

Ractors can send signals at any time, so the previous debug assertion can fail if a Ractor sends a signal.

Backport of https://github.com/ruby/ruby/pull/14588

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -16575,6 +16575,7 @@ scheduler.$(OBJEXT): {$(VPATH)}config.h
 scheduler.$(OBJEXT): {$(VPATH)}constant.h
 scheduler.$(OBJEXT): {$(VPATH)}defines.h
 scheduler.$(OBJEXT): {$(VPATH)}encoding.h
+scheduler.$(OBJEXT): {$(VPATH)}eval_intern.h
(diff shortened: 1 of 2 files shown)


ioquatix:
I had use a slightly different example code to get this to fail:

```ruby
require 'async/scheduler'

Signal.trap(:USR1) do
end

pid = Process.pid

child_pid = fork do
  loop do
    Process.kill(:USR1, pid)
    sleep 0.00001
  end
end

at_exit do
  Process.kill(:KILL, child_pid)
end

puts "Child pid: #{child_pid}"

scheduler = Async::Scheduler.new
Fiber.set_scheduler(scheduler)

q = Thread::Queue.new

Fiber.schedule do
  1.upto(1000000) do |i|
    sleep 0.01
    q.pop
    q.push(1)
    puts "1 iter push/pop"
  end
end
Fiber.schedule do
  1.upto(1000000) do |i|
    sleep 0.01
    q.push(i)
    q.pop
    puts "1 iter push/pop#2"
  end
end
Fiber.schedule do
  gets
  exit!
end
```

PR #14592, opened 2025-09-18
ZJIT: Improve `zjit_bisect.rb`

st0012:
1. The command name in the help message is outdated.
2. When the command failed by itself, the message can be clearer.

### DIFF
--- a/tool/zjit_bisect.rb
+++ b/tool/zjit_bisect.rb
@@ -17,9 +17,10 @@ OptionParser.new do |opts|
   end
 end.parse!
 
-RUBY = ARGV[0] || raise("Usage: ruby jit_bisect.rb <path_to_ruby> -- <options>")
+usage = "Usage: zjit_bisect.rb <path_to_ruby> -- <options>"
+RUBY = ARGV[0] || raise(usage)
 OPTIONS = ARGV[1..]
-raise("Usage: ruby jit_bisect.rb <path_to_ruby> -- <options>") if OPTIONS.empty?
+raise(usage) if OPTIONS.empty?
 LOGGER = Logger.new($stdout)
 
 # From https://github.com/tekknolagi/omegastar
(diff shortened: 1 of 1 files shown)


tool/zjit_bisect.rb
```diff
@@ -103,7 +104,7 @@ def run_with_jit_list(ruby, options, jit_list)
 
 # Try running with no JIT list to get a stable baseline
 unless run_with_jit_list(RUBY, OPTIONS, []).success?
-  raise "Command failed with empty JIT list"
+  raise "Command failed with empty JIT list. Make sure the command by itself runs successfully."
```
tekknolagi: I don't think this is clearer. Perhaps we can say something like:

> Expected the JIT to cause a failure but the command failed with an empty JIT list. To reproduce, try running the following:
>
> `[RUBY, "--zjit-allowed-iseqs=/dev/null", *OPTIONS].shelljoin`

tekknolagi (approved):
lgtm other than comment

tool/zjit_bisect.rb
```diff
@@ -103,7 +104,8 @@ def run_with_jit_list(ruby, options, jit_list)
 
 # Try running with no JIT list to get a stable baseline
 unless run_with_jit_list(RUBY, OPTIONS, []).success?
-  raise "Command failed with empty JIT list"
+  cmd = [RUBY, "--zjit-allowed-iseqs=/dev/null", *OPTIONS].shelljoin
+  raise "The command failed expectedly with an empty JIT list. To reproduce, try running the following: `#{cmd}`"
```
tekknolagi:
```suggestion
  raise "The command failed unexpectedly with an empty JIT list. To reproduce, try running the following: `#{cmd}`"
```

tekknolagi (approved):

PR #14594, opened 2025-09-18
ZJIT: Make exception allocation faster

tekknolagi:
- **Make rb_eException directly use default allocator**
- **ZJIT: Wrap comment**
- **ZJIT: Add tests for modified Exception allocator**

Just noticed this as I was looking around allocators.


tekknolagi:
Marshal does weird stuff with allocator pointer comparisons :(

PR #14595, opened 2025-09-18
ZJIT: Mark more allocators as leaf

tekknolagi:
- **ZJIT: Wrap comment**
- **ZJIT: Expand the list of safe allocators**

### DIFF
--- a/zjit/src/cruby.rs
+++ b/zjit/src/cruby.rs
@@ -1218,6 +1218,19 @@ pub fn get_class_name(class: VALUE) -> String {
     }).unwrap_or_else(|| "Unknown".to_string())
 }
 
+pub fn class_has_leaf_allocator(class: VALUE) -> bool {
+    // empty_hash_alloc
+    if class == unsafe { rb_cHash } { return true; }
+    // empty_ary_alloc
+    if class == unsafe { rb_cArray } { return true; }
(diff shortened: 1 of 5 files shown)


zjit/src/cruby.rs
```diff
@@ -1218,6 +1218,14 @@ pub fn get_class_name(class: VALUE) -> String {
     }).unwrap_or_else(|| "Unknown".to_string())
 }
 
+pub fn class_has_safe_allocator(class: VALUE) -> bool {
+    if class == unsafe { rb_cHash } { return true; }
```
k0kubun, line 1223: Could you maybe add a comment for each non-default-allocator class that mentions the name of the allocator function? I don't know where's the allocator function for Hash/Array/String/Regexp and I'm not sure why they're safe, so I would like to look at them.

zjit/src/cruby.rs
```diff
@@ -1218,6 +1218,14 @@ pub fn get_class_name(class: VALUE) -> String {
     }).unwrap_or_else(|| "Unknown".to_string())
 }
 
+pub fn class_has_safe_allocator(class: VALUE) -> bool {
```
k0kubun: The meaning of "safe" seems ambiguous. Could you just call it `class_has_leaf_allocator`?

PR #14607, opened 2025-09-19
ZJIT: Refactor `NewRangeFixnum`

altxtech:
## Context

#14409

https://github.com/ruby/ruby/pull/14409#discussion_r2350238583
>You may have noticed that this doesn't produce a New range instruction. For that you probably need to use a local variable (makes it harder for the bytecode compiler to reason about, but the JIT sees through it). If you have time, please rewrite these tests.

https://github.com/ruby/ruby/pull/14409#discussion_r2350240389
>There's some code above to do this for you: see arguments_likely_fixnums and coerce something or other

### Refactor tests

Modify tests force the usage of `NewRangeFixnum` instruction and make tests names more consistent.  

### Refactor optimize ranges

Simplify the logic and move it to `type_specialize`. Deleted the standalone `optimize_ranges` function.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -1080,7 +1080,8 @@ class TestZJIT < Test::Unit::TestCase
   def test_new_range_fixnum_both_literals_inclusive
     assert_compiles '1..2', %q{
       def test()
-        (1..2)
+        a = 2
+        (1..a)
       end
       test; test
     }, call_threshold: 2
@@ -1089,7 +1090,8 @@ class TestZJIT < Test::Unit::TestCase
   def test_new_range_fixnum_both_literals_exclusive
     assert_compiles '1...2', %q{
       def test()
-        (1...2)
+        a = 2
+        (1...a)
       end
       test; test
     }, call_threshold: 2
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -7163,41 +7163,50 @@ mod opt_tests {
     }
 
     #[test]
-    fn test_optimize_range_fixnum_inclusive_literals() {
+    fn test_optimize_new_range_fixnum_inclusive_literals() {
         eval("
             def test()
-              (1..2)
+              a = 2
+              (1..a)
             end
             test; test
         ");
         assert_snapshot!(hir_string("test"), @r"
         fn test@<compiled>:3:
         bb0(v0:BasicObject):
-          v4:RangeExact[VALUE(0x1000)] = Const Value(VALUE(0x1000))
+          v1:NilClass = Const Value(nil)
+          v5:Fixnum[2] = Const Value(2)
+          v8:Fixnum[1] = Const Value(1)
+          v16:RangeExact = NewRangeFixnum v8 NewRangeInclusive v5
           CheckInterrupts
-          Return v4
+          Return v16
         ");
     }
 
+
     #[test]
-    fn test_optimize_range_fixnum_exclusive_literals() {
+    fn test_optimize_new_range_fixnum_exclusive_literals() {
         eval("
             def test()
-              (1...2)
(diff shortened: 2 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -1982,6 +1982,20 @@ impl Function {
                         self.insn_types[replacement.0] = self.infer_type(replacement);
                         self.make_equal_to(insn_id, replacement);
                     }
+                    Insn::NewRange { low, high, flag, state } => {
+                        let low_is_fix  = self.is_a(low,  types::Fixnum);
+                        let high_is_fix = self.is_a(high, types::Fixnum);
+
+                        if low_is_fix || high_is_fix {
```
altxtech, line 1989: @tekknolagi 

`arguments_likely_fixnums` is not applicable here because that helper checks if *both* arguments are fixnums or profiled to be fixnums.  

Profiling is not enabled for `NewRange`, so this optimization is applied if *either* end can be proven to be fixnum without profiling.  

Relevant test case:
```rust 
  #[test]
  fn test_optimize_new_range_fixnum_inclusive_high_guarded() {
      eval("
          def test(a)
            (1..a)
          end
          test(2); test(3)
      ");
      fn test@<compiled>:3:
        v5:Fixnum[1] = Const Value(1)
        v13:Fixnum = GuardType v1, Fixnum
        v14:RangeExact = NewRangeFixnum v5 NewRangeInclusive v13
        CheckInterrupts
        Return v14
      ");
  }
```

In this case, the optimization fires even without profiling info for `a`.   Because the start of the range is a literal Fixnum.  

We could change this, tough. We could enable profiling and just use `arguments_likely_fixnum`, and fire the optimization if *both* ends are likely fixnum. In effect, making it similar to other Fixnum ops.

st0012 (approved):

test/ruby/test_zjit.rb
```diff
@@ -1080,7 +1080,8 @@ def test(n) = n..10
   def test_new_range_fixnum_both_literals_inclusive
     assert_compiles '1..2', %q{
       def test()
-        (1..2)
+        a = 2
+        (1..a)
       end
       test; test
     }, call_threshold: 2
```
tekknolagi: test for opcode `newrange`

test/ruby/test_zjit.rb
```diff
@@ -1089,7 +1090,8 @@ def test()
   def test_new_range_fixnum_both_literals_exclusive
     assert_compiles '1...2', %q{
       def test()
-        (1...2)
+        a = 2
```
tekknolagi, line 1149: test for opcode `newrange`

zjit/src/hir.rs
```diff
@@ -1985,6 +1985,20 @@ impl Function {
...
+                            let low_fix = self.coerce_to_fixnum(block, low, state);
+                            let high_fix = self.coerce_to_fixnum(block, high, state);
+                            let repl = self.push_insn(block, Insn::NewRangeFixnum{ low: low_fix, high: high_fix, flag, state });
```
```suggestion
                            let repl = self.push_insn(block, Insn::NewRangeFixnum { low: low_fix, high: high_fix, flag, state });
```

```suggestion
                            let replacement = self.push_insn(block, Insn::NewRangeFixnum{ low: low_fix, high: high_fix, flag, state });
```

altxtech: done

altxtech: done

test/ruby/test_zjit.rb
altxtech: done

test/ruby/test_zjit.rb
altxtech: done

tekknolagi:
Thank you!

PR #14608, opened 2025-09-19
ZJIT: Add polymorphism counters

aidenfoxivey:
Add counters to profile dynamic send instructions as being polymorphic or unprofiled.


zjit/src/hir.rs
```diff
@@ -1844,6 +1866,19 @@ impl Function {
...
+                                        Some(true) => {
+                                            self.push_insn(block, Insn::IncrCounter(Counter::dynamic_send_polymorphic));
+                                        },
+                                        Some(false) => {
+                                            panic!("Should not have monomorphic profile at this point in this branch");
+                                        },
+                                        None => {
+                                            self.push_insn(block, Insn::IncrCounter(Counter::dynamic_send_no_profiles));
+                                        }
+                                    }
```
k0kubun, line 1880: I want this counter code to take up less space if possible

```suggestion
                                    match self.is_polymorphic_at(recv, frame_state.insn_idx) {
                                        Some(true) => self.push_insn(block, Insn::IncrCounter(Counter::dynamic_send_polymorphic)),
                                        Some(false) => panic!("Should not have monomorphic profile at this point in this branch"),
                                        None => self.push_insn(block, Insn::IncrCounter(Counter::dynamic_send_no_profiles)),
                                    }
```

k0kubun:
Instead of printing them separately with `print_counters`, could we start creating a prefix for send fallback reasons with `print_counters_with_prefix`? Polymorphic calls and no profile seem like something that should be part of that analysis. 

I feel like that prefix should be named `send_fallback_` and the current `send_fallback_` should be renamed to something else, like `unspecialized_def_type_`, which breaks down a fallback reason like `send_fallback_unspecialized_def_type`.

ref: https://github.com/Shopify/ruby/issues/745

aidenfoxivey:
> Instead of printing them separately with `print_counters`, could we start creating a prefix for send fallback reasons with `print_counters_with_prefix`? Polymorphic calls and no profile seem like something that should be part of that analysis.
> 
> I feel like that prefix should be named `send_fallback_` and the current `send_fallback_` should be renamed to something else, like `unspecialized_def_type_`, which breaks down a fallback reason like `send_fallback_unspecialized_def_type`.
> 
> ref: [Shopify#745](https://github.com/Shopify/ruby/issues/745)

So we'd move from:
```
    dynamic_send_polymorphic,
    dynamic_send_no_profiles,
```

to 

```
    send_fallback_polymorphic,
    send_fallback_no_profiles,
```

zjit/src/hir.rs
```diff
@@ -1840,6 +1862,19 @@ impl Function {
                             // If we know that self is reasonably monomorphic from profile information, guard and use it to fold the lookup at compile-time.
                             // TODO(max): Figure out how to handle top self?
                             let Some(recv_type) = self.profiled_type_of_at(recv, frame_state.insn_idx) else {
+                                if get_option!(stats) {
+                                    match self.is_polymorphic_at(recv, frame_state.insn_idx) {
+                                        Some(true) => {
+                                            self.push_insn(block, Insn::IncrCounter(Counter::send_fallback_polymorphic));
+                                        },
+                                        Some(false) => {
+                                            panic!("Should not have monomorphic profile at this point in this branch");
```
st0012: Can you explain it a bit and potentially add a comment as well?

aidenfoxivey: Certainly!

zjit/src/hir.rs
```diff
@@ -1840,6 +1862,15 @@ impl Function {
                             // If we know that self is reasonably monomorphic from profile information, guard and use it to fold the lookup at compile-time.
                             // TODO(max): Figure out how to handle top self?
                             let Some(recv_type) = self.profiled_type_of_at(recv, frame_state.insn_idx) else {
+                                if get_option!(stats) {
+                                    match self.is_polymorphic_at(recv, frame_state.insn_idx) {
```
st0012, line 1866: So in this case, `is_polymorphic_at` is actually used to tell whether we statically know the type or not?
If my understanding is right, can we use simpler approaches to do that here?

And what would be other places where we can use `is_polymorphic_at` that can use the 3 possible outcomes it returns?

aidenfoxivey: > If my understanding is right, can we use simpler approaches to do that here?

Interesting - I would be happy to have something more simple. Do you have anything in mind?

> And what would be other places where we can use is_polymorphic_at that can use the 3 possible outcomes it returns?

At least for this pull request, I suppose I imagined `is_polymorphic_at` to be mostly for counters? I _think_ this is where all the counters are used here as it stands.

k0kubun (approved):

PR #14609, opened 2025-09-19
Make it easier to reproduce commands from CI

rwstauner:
Expand the interpolated command so that it is easier to see exactly what command was run.

image

### DIFF
--- a/.github/auto_request_review.yml
+++ b/.github/auto_request_review.yml
@@ -11,6 +11,7 @@ files:
   'doc/zjit*': [team:jit]
   'test/ruby/test_zjit*': [team:jit]
   'defs/jit.mk': [team:jit]
+  # Skip github workflow files because the team don't necessarily need to review dependabot updates for GitHub Actions. It's noisy in notifications, and they're auto-merged anyway.
 options:
   ignore_draft: true
   # This currently doesn't work as intended. We want to skip reviews when only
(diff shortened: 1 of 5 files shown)


.github/auto_request_review.yml
```diff
@@ -11,6 +11,7 @@ files:
   'doc/zjit*': [team:jit]
   'test/ruby/test_zjit*': [team:jit]
   'defs/jit.mk': [team:jit]
+  # Skip github workflow files because the team don't necessarily need to review dependabot updates for GitHub Actions. It's noisy in notifications, and they're auto-merged anyway.
```
rwstauner, line 14: https://github.com/ruby/ruby/pull/13912#issuecomment-3079718656

.github/workflows/yjit-ubuntu.yml
```diff
@@ -195,6 +195,7 @@ jobs:
           test -n "${LAUNCHABLE_STDOUT}" && exec 1> >(tee "${LAUNCHABLE_STDOUT}")
           test -n "${LAUNCHABLE_STDERR}" && exec 2> >(tee "${LAUNCHABLE_STDERR}")
 
+          verbose () { set -x && "$@" && set +x; } && verbose \
```
k0kubun: Can we just use `set -x`?

```suggestion
          set -x
```

I would rather want `zjit-*.yml` to use `|` instead of `>-` and call `set -x` there too.

rwstauner: Sounds good to me.

PR #14610, opened 2025-09-19
ZJIT: Fix and run CI with HIR optimizer disabled

tekknolagi:
Fixes https://github.com/Shopify/ruby/issues/758

### DIFF
--- a/.github/workflows/zjit-macos.yml
+++ b/.github/workflows/zjit-macos.yml
@@ -38,6 +38,12 @@ jobs:
             specopts: '-T --zjit-call-threshold=1'
             test_all_opts: '--seed=46450'
 
+          - test_task: 'check'
+            configure: '--enable-zjit=dev'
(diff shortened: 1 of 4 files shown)


tekknolagi:
Might also fix https://github.com/Shopify/ruby/issues/741

tekknolagi:
Sure--will rebase anyway

.github/workflows/zjit-macos.yml
```diff
@@ -38,6 +38,12 @@ jobs:
             specopts: '-T --zjit-call-threshold=1'
             test_all_opts: '--seed=46450'
 
+          - test_task: 'check'
+            configure: '--enable-zjit=dev'
```
k0kubun: Could you put `configure` after the lines of `run_opts` and `specopts` for this job and the above job in both `zjit-ubuntu` and `zjit-macos`? I want to see `--zjit-disable-hir-opt` earlier in the job name to understand the difference between them more quickly.

k0kubun (approved):

PR #14611, opened 2025-09-19
ZJIT: Simplify NewHash HIR and Codegen

st0012:
We can use `Vec<Opnd>` instead of `Vec<(Opnd, Opnd)>` in NewHash HIR as it's the only usage of such type, which requires special handling.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -549,7 +549,7 @@ pub enum Insn {
     ToNewArray { val: InsnId, state: InsnId },
     NewArray { elements: Vec<InsnId>, state: InsnId },
     /// NewHash contains a vec of (key, value) pairs
-    NewHash { elements: Vec<(InsnId,InsnId)>, state: InsnId },
+    NewHash { elements: Vec<InsnId>, state: InsnId },
     NewRange { low: InsnId, high: InsnId, flag: RangeType, state: InsnId },
     NewRangeFixnum { low: InsnId, high: InsnId, flag: RangeType, state: InsnId },
     ArrayDup { val: InsnId, state: InsnId },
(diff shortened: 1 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -2374,9 +2370,11 @@ impl Function {
                 worklist.push_back(state);
             }
             &Insn::NewHash { ref elements, state } => {
-                for &(key, value) in elements {
```
tekknolagi, line 2377: this can be `worklist.extend(elements)` or something

zjit/src/hir.rs
```diff
@@ -3427,12 +3425,17 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     assert!(count % 2 == 0, "newhash count should be even");
                     let exit_id = fun.push_insn(block, Insn::Snapshot { state: exit_state });
                     let mut elements = vec![];
+                    let mut pairs = vec![];
                     for _ in 0..(count/2) {
                         let value = state.stack_pop()?;
                         let key = state.stack_pop()?;
-                        elements.push((key, value));
+                        pairs.push((key, value));
```
```
pairs.push(value);
pairs.push(key);
```

and then just reverse, or something?

tekknolagi (approved):
lgtm

zjit/src/hir.rs
st0012: The order would become `key3: value3, key2: value2, key1: value1` due to the order things are popped. And since Ruby's hash is ordered, we need to preserve the order here too

st0012: Oh wait I'm wrong 🤔

zjit/src/hir.rs
```diff
@@ -3427,11 +3417,14 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
                     let mut elements = vec![];
+
                     for _ in 0..(count/2) {
                         let value = state.stack_pop()?;
                         let key = state.stack_pop()?;
-                        elements.push((key, value));
+                        elements.push(value);
+                        elements.push(key);
                     }
+
```
tekknolagi: nit: no newline

st0012:
@tekknolagi Can you help me merge this?

PR #14612, opened 2025-09-19
ZJIT: Fix disasm tests on release build

k0kubun:
This PR fixes `make zjit-test` on builds that don't have the `disasm` feature.

### DIFF
--- a/zjit/src/asm/arm64/mod.rs
+++ b/zjit/src/asm/arm64/mod.rs
@@ -1247,112 +1247,112 @@ mod tests {
     fn test_add_reg() {
         let cb = compile(|cb| add(cb, X0, X1, X2));
         cb.with_disasm(|disasm| assert_snapshot!(disasm, @"  0x0: add x0, x1, x2"));
-        assert_snapshot!(cb.string(), @"2000028b");
+        assert_snapshot!(cb.hexdump(), @"2000028b");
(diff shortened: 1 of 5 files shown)


zjit/src/asm/arm64/mod.rs
```diff
@@ -1246,129 +1246,129 @@ mod tests {
     #[test]
     fn test_add_reg() {
         let cb = compile(|cb| add(cb, X0, X1, X2));
-        assert_snapshot!(cb.disasm(), @"  0x0: add x0, x1, x2");
-        assert_snapshot!(cb.string(), @"2000028b");
+        cb.with_disasm(|disasm| assert_snapshot!(disasm, @"  0x0: add x0, x1, x2"));
```
st0012, line 1249: Would it be even better if we have a marco like: `assert_disasm_snapshot!(cb, @"  0x0: add x0, x1, x2")`?

k0kubun: My understanding is that `insta` doesn't support updating a custom macro like that

st0012: Ah I see that makes sense. Thanks for the context 👍

st0012 (approved):
Thanks for fixing it!

PR #14613, opened 2025-09-19
ZJIT: Add `PushOpnds` and `PopOpnds` HIR instructions

st0012:
This will give us the ability to optimize those operations away when the elements are empty.

If this is acceptable, I will open another PR to not adding those instructions when not necessary.


st0012:
After discussing with @tekknolagi, we agreed that whether an HIR requires pushing/popping opnds should not be a HIR concern.

PR #14614, opened 2025-09-19
ZJIT: Avoid unnecessary `PopOpnds` and `PushOpnds` codegen

st0012:
If there's no opnds to restore, we don't need to do anything.


k0kubun:
Why don't you skip `sub_into` in `gen_push_opnds` too?

k0kubun:
What instructions use such `opnds` by the way?

st0012:
`NewHash`, `ToRegexp`, `StringConcat`, and `CCallVariadic`

k0kubun (approved):

tekknolagi:
Ooh, nice

PR #14616, opened 2025-09-20
Remove unnecessary `iseq` arg from `vm_get|setclass_variable` functions

st0012:
Since all of its callers retrieve the `iseq`'s value from the same CFP they pass to the functions, we don't need to pass the iseq separately.

As a reference, this is `GET_ISEQ`'s definition:

```
#define GET_ISEQ() (GET_CFP()->iseq)
```

### DIFF
--- a/yjit/src/codegen.rs
+++ b/yjit/src/codegen.rs
@@ -10207,7 +10207,6 @@ fn gen_getclassvariable(
     let val_opnd = asm.ccall(
         rb_vm_getclassvariable as *const u8,
         vec![
-            Opnd::mem(64, CFP, RUBY_OFFSET_CFP_ISEQ),
             CFP,
             Opnd::UImm(jit.get_arg(0).as_u64()),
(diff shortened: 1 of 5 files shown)


yjit/src/codegen.rs
```diff
@@ -10231,7 +10230,6 @@ fn gen_setclassvariable(
     asm.ccall(
         rb_vm_setclassvariable as *const u8,
         vec![
-            Opnd::mem(64, CFP, RUBY_OFFSET_CFP_ISEQ),
```
tekknolagi, line 10234: YJIT and ZJIT could theoretically burn in constants here. I wonder why we do CFP offsets

st0012:
@tekknolagi raised a good point that `reg_cfp->iseq` would still have a memory read. So while this code is a refactor, it doesn't optimzie anything. What we can do instead is to keep this parameter, but in ZJIT & YJIT pass the `iseqPtr` we already have to it, which doesn't need any memory read.
I'll close this PR and open another one to do that.

PR #14624, opened 2025-09-22
Make sure we never context switch while holding VM lock.

luke-gruber:
We were seeing errors in our application that looked like:

```
[BUG] unexpected situation - recordd:1 current:0
/error.c:1097 rb_bug_without_die_internal
/vm_sync.c:275 disallow_reentry
/eval_intern.h:136 rb_ec_vm_lock_rec_check
/eval_intern.h:147 rb_ec_tag_state
/vm.c:2619 rb_vm_exec
/vm.c:1702 rb_yield
/eval.c:1173 rb_ensure
```

We concluded that there was context switching going on while a thread held the VM lock. During the investigation into the issue, we added assertions that we never yield to another thread with the VM lock held. We enabled these VM lock assertions even in single ractor mode. These assertions were failing in a few places, but most notably in finalizers. We were running finalizers with the VM lock held, and they were context switching and causing this issue.

These rules must be held going forward to ensure we don't context switch unexpectedly:

If you have the VM lock held,
    * Don't enter the interpreter loop.
    * Don't call ruby methods whether or not they are defined in ruby
    * Don't call `rb_nogvl`, `rb_mutex_lock`, etc.
    * Don't check interrupts

Rework global variables, don't lock when calling getter or setter.

Add a test that fails without these lock_rec changes.

Add ASSERT_vm_unlocking() to vm_call0_body
This uncovered many more test failures.

Revert changes introduced in 2f6c694977dc0cdc4766a8a921e74290963c19a7

This didn't appear to be a correct fix. We should allow raising NoMemoryError even if we're under the VM lock. It will automatically unlock us.


yjit/src/yjit.rs
```diff
@@ -45,6 +45,14 @@ pub extern "C" fn rb_yjit_init(yjit_enabled: bool) {
 
     // If --yjit-disable, yjit_init() will not be called until RubyVM::YJIT.enable.
     if yjit_enabled {
+
+    // Call YJIT hooks before enabling YJIT to avoid compiling the hooks themselves.
+    // Needs to be done without VM lock held because it can context switch.
+    unsafe {
+        let yjit = rb_const_get(rb_cRubyVM, rust_str_to_id("YJIT"));
+        rb_funcall(yjit, rust_str_to_id("call_jit_hooks"), 0);
+    }
```
k0kubun: indentation seems off and there's an empty line that seems unnecessary

```suggestion
        // Call YJIT hooks before enabling YJIT to avoid compiling the hooks themselves.
        // Needs to be done without VM lock held because it can context switch.
        unsafe {
            let yjit = rb_const_get(rb_cRubyVM, rust_str_to_id("YJIT"));
            rb_funcall(yjit, rust_str_to_id("call_jit_hooks"), 0);
        }
```

k0kubun: Also, you're copying this to two places instead of sharing it through `yjit_init` now, but how about creating a separate function that does this and calling it here and in `rb_yjit_enable`?

k0kubun:
YJIT changes look good to me

vm_sync.h
```diff
@@ -57,6 +67,13 @@ rb_vm_lock(const char *file, int line)
     if (rb_multi_ractor_p()) {
         rb_vm_lock_body(LOCATION_PARAMS);
     }
+    else {
+#if VM_CHECK_MODE > 0
+        if (rb_current_execution_context(false)) {
```
jhawthorn: What's the case that we end up without an EC? During boot? I feel like otherwise we should make sure there is an EC in order to take the VM lock

luke-gruber: Yes 😆 I'll address this.

luke-gruber: @jhawthorn Okay fixed.

PR #14625, opened 2025-09-22
YJIT: Pass iseq pointer to get/set classvariable functions

st0012:
Since we already have the iseq pointer, we can actually save one memory read by passing it directly.


aidenfoxivey (approved):
Looks reasonable!

PR #14629, opened 2025-09-22
ZJIT: Inline attr_accessor/attr_writer to SetIvar

tekknolagi:


tekknolagi:
Welp, I think this causes crashes in benchmarks despite passing tests

tekknolagi:
https://github.com/ruby/ruby/pull/14673

PR #14635, opened 2025-09-23
gc_validate_pc(): Exclude imemos; add a test and a hint in assert message

XrXr:
The validation is relevant only for traceable userland ruby objects ruby code could interact with. ZJIT's use of `rb_vm_method_cfunc_is()` allocates a CC imemo and was failing this validation when it was actually fine. Relax the check.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -971,14 +971,16 @@ rb_gc_obj_slot_size(VALUE obj)
 }
 
 static inline void
-gc_validate_pc(void)
+gc_validate_pc(VALUE obj)
 {
 #if RUBY_DEBUG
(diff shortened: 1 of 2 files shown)


tekknolagi:
I would also like to change the ZJIT optimizer to *not* allocate at compile-time. What is the best replacement method?

XrXr:
In this case the allocation was happening at runtime though, through `Insn::IsMethodCfunc` as part of opt_new.

But all lookup methods do caching in some way so all allocate. We kind of just have to deal with it unless we want to move out of using CMEs wholesale, not realistic.

PR #14638, opened 2025-09-23
ZJIT: Add stats for cfuncs that are not optimized

st0012:
From `lobsters`

```
Top-20 Unoptimized C functions (73.0% of total 15,276,688):
                               Kernel#is_a?: 2,052,363 (13.4%)
                              Class#current: 1,892,623 (12.4%)
                                String#to_s:   975,973 ( 6.4%)
                                  Hash#key?:   677,623 ( 4.4%)
                              String#empty?:   636,468 ( 4.2%)
                              TrueClass#===:   457,232 ( 3.0%)
                                   Hash#[]=:   455,908 ( 3.0%)
                             FalseClass#===:   448,798 ( 2.9%)
         ActiveSupport::OrderedOptions#_get:   377,468 ( 2.5%)
                            Kernel#kind_of?:   339,551 ( 2.2%)
                                 Kernel#dup:   329,371 ( 2.2%)
                                  String#==:   324,286 ( 2.1%)
                            String#include?:   297,528 ( 1.9%)
                                    Hash#[]:   294,561 ( 1.9%)
                             Array#include?:   287,145 ( 1.9%)
                        Kernel#block_given?:   283,633 ( 1.9%)
                             BasicObject#!=:   278,874 ( 1.8%)
                                Hash#delete:   250,951 ( 1.6%)
                               Set#include?:   246,447 ( 1.6%)
                               NilClass#===:   242,776 ( 1.6%)
```

From `liquid-render`

```
Top-20 Unoptimized C functions (99.8% of total 5,195,549):
                          Hash#key?: 2,459,048 (47.3%)
                        String#to_s: 1,119,758 (21.6%)
                       Set#include?:   799,469 (15.4%)
                       Kernel#is_a?:   214,223 ( 4.1%)
                         Integer#<<:   171,073 ( 3.3%)
                          Integer#/:   127,622 ( 2.5%)
          CGI::EscapeExt#escapeHTML:    56,971 ( 1.1%)
                         Regexp#===:    50,008 ( 1.0%)
                      String#empty?:    43,990 ( 0.8%)
                         String#===:    36,838 ( 0.7%)
                          String#==:    21,309 ( 0.4%)
                      Time#strftime:    21,251 ( 0.4%)
                       String#strip:    15,271 ( 0.3%)
                        String#scan:    13,753 ( 0.3%)
                          String#+@:    12,603 ( 0.2%)
                     Array#include?:     8,059 ( 0.2%)
                           String#+:     5,295 ( 0.1%)
                         String#dup:     4,606 ( 0.1%)
                          String#-@:     3,213 ( 0.1%)
                     Class#generate:     3,011 ( 0.1%)
```

### DIFF
--- a/zjit.rb
+++ b/zjit.rb
@@ -43,7 +43,7 @@ class << RubyVM::ZJIT
     print_counters_with_prefix(prefix: 'dynamic_send_type_', prompt: 'dynamic send types', buf:, stats:, limit: 20)
     print_counters_with_prefix(prefix: 'unspecialized_def_type_', prompt: 'send fallback unspecialized def_types', buf:, stats:, limit: 20)
     print_counters_with_prefix(prefix: 'send_fallback_', prompt: 'dynamic send types', buf:, stats:, limit: 20)
-    print_counters_with_prefix(prefix: 'not_optimized_cfuncs_', prompt: 'Unoptimized C functions', buf:, stats:, limit: 20)
+    print_counters_with_prefix(prefix: 'not_optimized_cfuncs_', prompt: 'unoptimized sends to C functions', buf:, stats:, limit: 20)
 
     # Show exit counters, ordered by the typical amount of exits for the prefix at the time
     print_counters_with_prefix(prefix: 'unhandled_yarv_insn_', prompt: 'unhandled YARV insns', buf:, stats:, limit: 20)
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -419,7 +419,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::GetSpecialSymbol { symbol_type, state: _ } => gen_getspecial_symbol(asm, *symbol_type),
         Insn::GetSpecialNumber { nth, state } => gen_getspecial_number(asm, *nth, &function.frame_state(*state)),
         &Insn::IncrCounter(counter) => no_output!(gen_incr_counter(asm, counter)),
-        Insn::CountUnoptimizedCFunc { signature, counter_ptr } => no_output!(gen_count_unoptimized_cfunc(asm, signature, *counter_ptr)),
+        Insn::IncrCounterPtr { counter_ptr } => no_output!(gen_incr_counter_ptr(asm, *counter_ptr)),
         Insn::ObjToString { val, cd, state, .. } => gen_objtostring(jit, asm, opnd!(val), *cd, &function.frame_state(*state)),
         &Insn::CheckInterrupts { state } => no_output!(gen_check_interrupts(jit, asm, &function.frame_state(state))),
         &Insn::HashDup { val, state } => { gen_hash_dup(asm, opnd!(val), &function.frame_state(state)) },
@@ -1530,13 +1530,11 @@ fn gen_guard_bit_equals(jit: &mut JITState, asm: &mut Assembler, val: lir::Opnd,
 }
 
 /// Generate code that records unoptimized C functions if --zjit-stats is enabled
-fn gen_count_unoptimized_cfunc(asm: &mut Assembler, signature: &str, counter_ptr: *mut u64) {
+fn gen_incr_counter_ptr(asm: &mut Assembler, counter_ptr: *mut u64) {
     if get_option!(stats) {
-        unsafe extern "C" {
-            fn rb_zjit_count_unoptimized_cfunc(counter_ptr: *mut u64);
-        }
-        asm_comment!(asm, "count unoptimized cfunc: {}", signature);
-        asm_ccall!(asm, rb_zjit_count_unoptimized_cfunc, Opnd::const_ptr(counter_ptr as *const u8));
+        let ptr_reg = asm.load(Opnd::const_ptr(counter_ptr as *const u8));
+        let counter_opnd = Opnd::mem(64, ptr_reg, 0);
+        asm.incr_counter(counter_opnd, Opnd::UImm(1));
     }
 }
 
@@ -1544,11 +1542,7 @@ fn gen_count_unoptimized_cfunc(asm: &mut Assembler, signature: &str, counter_ptr
 fn gen_incr_counter(asm: &mut Assembler, counter: Counter) {
     if get_option!(stats) {
         let ptr = counter_ptr(counter);
-        let ptr_reg = asm.load(Opnd::const_ptr(ptr as *const u8));
-        let counter_opnd = Opnd::mem(64, ptr_reg, 0);
-
-        // Increment and store the updated value
-        asm.incr_counter(counter_opnd, Opnd::UImm(1));
+        gen_incr_counter_ptr(asm, ptr);
     }
 }
 
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -14,7 +14,7 @@ use std::{
 use crate::hir_type::{Type, types};
 use crate::bitset::BitSet;
 use crate::profile::{TypeDistributionSummary, ProfiledType};
-use crate::stats::{Counter, get_or_create_unoptimized_cfunc_counter_ptr};
+use crate::stats::Counter;
 
 /// An index of an [`Insn`] in a [`Function`]. This is a popular
 /// type since this effectively acts as a pointer to an [`Insn`].
@@ -711,8 +711,8 @@ pub enum Insn {
     /// Increment a counter in ZJIT stats
     IncrCounter(Counter),
 
-    /// Increment a counter in ZJIT stats for the given unoptimized C function
-    CountUnoptimizedCFunc { signature: String, counter_ptr: *mut u64 },
+    /// Increment a counter in ZJIT stats for the given counter pointer
+    IncrCounterPtr { counter_ptr: *mut u64 },
 
     /// Equivalent of RUBY_VM_CHECK_INTS. Automatically inserted by the compiler before jumps and
     /// return instructions.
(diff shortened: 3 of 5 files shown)


zjit/src/hir.rs
```diff
@@ -2122,15 +2122,30 @@ impl Function {
...
-                        return Err(());
+                    let cfunc_properties = ZJITState::get_method_annotations().get_cfunc_properties(method);
+
+                    let (return_type, elidable) = match cfunc_properties {
+                        Some(FnProperties { leaf: true, no_gc: true, return_type, elidable }) => {
+                            (return_type, elidable)
+                        }
+                        Some(FnProperties { leaf: false, no_gc: true, .. }) => {
+                            fun.push_insn(block, Insn::IncrCounter(Counter::not_optimized_cfunc_not_leaf));
+                            return Err(());
```
k0kubun: How about returning a counter (or an enum that represents relevant counters) as part of `Err` (changing the return type to something like `Result<(), Counter>` or `Result<(), Option<Counter>>`) and letting the caller push `Insn::IncrCounter` on error?

It seems wasteful that `Err` type is `()` and `push_insn` code is duplicated while the error is explained in the counter.

st0012: Great idea! Updated

st0012:
Actually, I want to see if I can make `not_leaf_nor_no_gc` contain the method name as well. That'll provide more actionable info.

tekknolagi:
Re: method names: @aidenfoxivey is doing exit counter -> iseq attribution in a separate PR so maybe they can be complementary

gems/bundled_gems
```diff
@@ -10,22 +10,22 @@ minitest            5.25.5  https://github.com/minitest/minitest
...
-rexml               3.4.2   https://github.com/ruby/rexml
+rexml               3.4.4   https://github.com/ruby/rexml
 rss                 0.3.1   https://github.com/ruby/rss
 net-ftp             0.3.8   https://github.com/ruby/net-ftp
 net-imap            0.5.10  https://github.com/ruby/net-imap 71c0288b9a8f78a7125a4ce2980ab421acdf5836
 net-pop             0.1.2   https://github.com/ruby/net-pop
 net-smtp            0.5.1   https://github.com/ruby/net-smtp
 matrix              0.4.3   https://github.com/ruby/matrix
 prime               0.1.4   https://github.com/ruby/prime
-rbs                 3.9.4   https://github.com/ruby/rbs aa22d7ea0c992de7557c3e346c9100b8aa36d945
```
tekknolagi, line 21: This probably shouldn't be part of the PR

zjit/src/hir.rs
```diff
@@ -2271,8 +2277,19 @@ impl Function {
             for insn_id in old_insns {
                 if let send @ Insn::SendWithoutBlock { recv, .. } = self.find(insn_id) {
                     let recv_type = self.type_of(recv);
-                    if reduce_to_ccall(self, block, recv_type, send, insn_id).is_ok() {
-                        continue;
+                    match reduce_to_ccall(self, block, recv_type, send, insn_id) {
+                        Ok(()) => continue,
+                        Err(Some(cme)) => {
+                            let owner = unsafe { (*cme).owner };
```
tekknolagi: Let's guard this on `if get_option!(stats)`

zjit/src/stats.rs
```diff
@@ -188,6 +190,21 @@ pub(crate) use incr_counter;
 /// The number of side exits from each YARV instruction
 pub type ExitCounters = [u64; VM_INSTRUCTION_SIZE as usize];
 
+/// Store signature ("Klass#method") to counter pointer mappings for unoptimized cfuncs
+pub static UNOPTIMIZED_CFUNC_COUNTER_POINTERS: LazyLock<Mutex<HashMap<String, Box<u64>>>> = LazyLock::new(|| Mutex::new(HashMap::new()));
+
+/// Get or create a counter pointer for the given signature
+pub fn get_or_create_unoptimized_cfunc_counter_ptr(signature: String) -> *mut u64 {
+    let mut map = UNOPTIMIZED_CFUNC_COUNTER_POINTERS.lock().unwrap();
```
tekknolagi: I would love if @XrXr weighed in on the locking here

zjit.rb
```diff
@@ -43,6 +43,7 @@ def stats_string
     print_counters_with_prefix(prefix: 'dynamic_send_type_', prompt: 'dynamic send types', buf:, stats:, limit: 20)
     print_counters_with_prefix(prefix: 'unspecialized_def_type_', prompt: 'send fallback unspecialized def_types', buf:, stats:, limit: 20)
     print_counters_with_prefix(prefix: 'send_fallback_', prompt: 'dynamic send types', buf:, stats:, limit: 20)
+    print_counters_with_prefix(prefix: 'not_optimized_cfuncs_', prompt: 'Unoptimized C functions', buf:, stats:, limit: 20)
```
tekknolagi: Nit: unoptimized *sends* to C functions

zjit/src/hir.rs
```diff
@@ -707,6 +707,9 @@ pub enum Insn {
     /// Increment a counter in ZJIT stats
     IncrCounter(Counter),
 
+    /// Increment a counter in ZJIT stats for the given unoptimized C function
+    CountUnoptimizedCFunc { signature: String, counter_ptr: *mut u64 },
```
tekknolagi: Do we care enough about the HIR looking a certain way to make a new opcode that's different from `IncrCounter`? It seems like if not for the printing/asm comment, we could get away with just using that.

tekknolagi: Maybe not, if `Counter` vs `counter_ptr` or something. I dunno. This feels like a slightly funky choice of abstractions

tekknolagi: Maybe `IncrDynamicCounter` or something like that

tekknolagi: nit: this is not a signature but instead the (qualified?) method name. signature would include args, etc

tekknolagi (commented):
mostly lgtm

zjit/src/hir.rs
st0012: Should we do the same for `IncrCounter` tho? I thought we still want to generate the counter HIRs even without stats enabled.
Another option is to move this part into codegen

zjit/src/stats.rs
XrXr: Let's move the map to ZJITState and get rid of the locking. It's not different from other global states WRT locking.

zjit/src/stats.rs
```diff
@@ -188,6 +190,21 @@ pub(crate) use incr_counter;
...
+pub fn get_or_create_unoptimized_cfunc_counter_ptr(qualified_method_name: String) -> *mut u64 {
+    let mut map = UNOPTIMIZED_CFUNC_COUNTER_POINTERS.lock().unwrap();
+    let counter = map.entry(qualified_method_name).or_insert_with(|| Box::new(0));
+    &mut **counter as *mut u64
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn rb_zjit_increment_dynamic_counter(counter_ptr: *mut u64) {
+    unsafe { *counter_ptr += 1 }
+}
```
XrXr: A whole function for this seems unnecessary. LIR can do it.

zjit/src/stats.rs
st0012: Ohh I didn't know we can us ZJIT state for this. It's much cleaner

zjit/src/codegen.rs
```diff
@@ -420,6 +420,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::GetSpecialSymbol { symbol_type, state: _ } => gen_getspecial_symbol(asm, *symbol_type),
         Insn::GetSpecialNumber { nth, state } => gen_getspecial_number(asm, *nth, &function.frame_state(*state)),
         &Insn::IncrCounter(counter) => no_output!(gen_incr_counter(asm, counter)),
+        Insn::IncrDynamicCounter { counter_ptr } => no_output!(gen_incr_counter_ptr(asm, *counter_ptr)),
```
k0kubun: > https://github.com/ruby/ruby/pull/14638#discussion_r2383005725

I got rather confused by the "Dynamic" part of the name. I get that `counter_ptr` is dynamically fetched, but once the instruction is constructed, it seems no more dynamic than `IncrCounter`.

How about calling it `IncrCounterPtr`?

zjit/src/hir.rs
```diff
@@ -2271,8 +2277,21 @@ impl Function {
...
+                            let owner = unsafe { (*cme).owner };
+                            let called_id = unsafe { (*cme).called_id };
+                            let class_name = get_class_name(owner);
+                            let method_name = called_id.contents_lossy();
+                            let qualified_method_name = format!("{}#{}", class_name, method_name);
+                            let unoptimized_cfunc_counter_pointers = ZJITState::get_unoptimized_cfunc_counter_pointers();
+                            let counter_ptr = unoptimized_cfunc_counter_pointers.entry(qualified_method_name.clone()).or_insert_with(|| Box::new(0));
+                            let counter_ptr = &mut **counter_ptr as *mut u64;
+
+                            self.push_insn(block, Insn::IncrDynamicCounter { counter_ptr });
```
k0kubun: Can we insert this only if `get_option!(stats)` is true? It seems inconsistent with `Insn::IncrCounter`. It also seems only confusing to insert them in HIR when it does nothing.

zjit/src/hir.rs
tekknolagi: I think we already guard on `stats` for `IncrCounter`. Do you see a case where it is not guarded?

zjit/src/codegen.rs
tekknolagi: Yeah I suggested Dynamic for lack of a better name :( Sorry. It's "dynamic" in the sense that the counter is not statically known at Rust compile-time, I guess. Hm.

zjit/src/hir.rs
st0012: Ah I got the wrong impression that we're always generating `IncrCounter`. Updated

zjit/src/hir.rs
```diff
@@ -978,6 +981,7 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             },
+            Insn::IncrCounterPtr { .. } => write!(f, "IncrDynamicCounter"),
```
tekknolagi: update string name please

PR #14639, opened 2025-09-23
ZJIT: Allow testing JIT code on zjit-test

k0kubun:
This PR allows `zjit-test` to run JIT code and test the result of it.

* It changes `CodeBlock` to use `SystemAllocator` instead of `TestingAllocator` in tests to mark code executable.
* I added a helper function `inspect` to test the output of a Ruby expression and let a test case use it. It's out of scope in this PR to use it in more test cases.
* I also refactored test code a little to use `set_call_threshold` because setting a call threshold through `internal_set_num_profiles` seemed confusing.

### DIFF
--- a/zjit/src/virtualmem.rs
+++ b/zjit/src/virtualmem.rs
@@ -7,6 +7,8 @@ use std::ptr::NonNull;
 use crate::cruby::*;
 use crate::stats::zjit_alloc_size;
 
+pub type VirtualMem = VirtualMemory<sys::SystemAllocator>;
+
 /// Memory for generated executable machine code. When not testing, we reserve address space for
 /// the entire region upfront and map physical memory into the reserved address space as needed. On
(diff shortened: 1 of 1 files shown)


zjit/src/virtualmem.rs
```diff
@@ -302,160 +325,3 @@ pub mod sys {
         }
     }
 }
-
-
-#[cfg(test)]
-pub mod tests {
-    use super::*;
```
XrXr, line 309: Can we keep these unit tests? Deleting all of this seems extreme, and we can keep them without impacting CodeBlock. CodeBlock could use `VirtualMemory<sys::SystemAllocator>` in all builds.

k0kubun: Fixed in https://github.com/ruby/ruby/pull/14639/commits/5087777963bcb054c4ca81fbc6ed1662f5c49538

XrXr (approved):

PR #14640, opened 2025-09-23
ZJIT: Add --zjit-trace-exits

aidenfoxivey:
Add side exit tracing functionality for ZJIT

### DIFF
--- a/zjit.c
+++ b/zjit.c
@@ -31,6 +31,95 @@
 
 #include <errno.h>
 
+#define PTR2NUM(x) (rb_int2inum((intptr_t)(void *)(x)))
+
+// For a given raw_sample (frame), set the hash with the caller's
+// name, file, and line number. Return the  hash with collected frame_info.
+static void
+rb_zjit_add_frame(VALUE hash, VALUE frame)
+{
+    VALUE frame_id = PTR2NUM(frame);
+
+    if (RTEST(rb_hash_aref(hash, frame_id))) {
+        return;
+    }
+    else {
+        VALUE frame_info = rb_hash_new();
+        // Full label for the frame
+        VALUE name = rb_profile_frame_full_label(frame);
+        // Absolute path of the frame from rb_iseq_realpath
+        VALUE file = rb_profile_frame_absolute_path(frame);
+        // Line number of the frame
+        VALUE line = rb_profile_frame_first_lineno(frame);
+
+        // If absolute path isn't available use the rb_iseq_path
+        if (NIL_P(file)) {
+            file = rb_profile_frame_path(frame);
+        }
+
+        rb_hash_aset(frame_info, ID2SYM(rb_intern("name")), name);
+        rb_hash_aset(frame_info, ID2SYM(rb_intern("file")), file);
+
+        if (line != INT2FIX(0)) {
+            rb_hash_aset(frame_info, ID2SYM(rb_intern("line")), line);
+        }
+
+       rb_hash_aset(hash, frame_id, frame_info);
+    }
+}
+
+// Parses the ZjitExitLocations raw_samples and line_samples collected by
+// rb_zjit_record_exit_stack and turns them into 3 hashes (raw, lines, and frames) to
+// be used by RubyVM::ZJIT.exit_locations. zjit_raw_samples represents the raw frames information
+// (without name, file, and line), and zjit_line_samples represents the line information
+// of the iseq caller.
+VALUE
+rb_zjit_exit_locations_dict(VALUE *zjit_raw_samples, int *zjit_line_samples, int samples_len)
+{
+    VALUE result = rb_hash_new();
+    VALUE raw_samples = rb_ary_new_capa(samples_len);
+    VALUE line_samples = rb_ary_new_capa(samples_len);
+    VALUE frames = rb_hash_new();
+    int idx = 0;
+
+    // While the index is less than samples_len, parse zjit_raw_samples and
+    // zjit_line_samples, then add casted values to raw_samples and line_samples array.
+    while (idx < samples_len) {
+        int num = (int)zjit_raw_samples[idx];
+        int line_num = (int)zjit_line_samples[idx];
+        idx++;
+
+        rb_ary_push(raw_samples, SIZET2NUM(num));
+        rb_ary_push(line_samples, INT2NUM(line_num));
+
+        // Loop through the length of samples_len and add data to the
+        // frames hash. Also push the current value onto the raw_samples
+        // and line_samples arrary respectively.
+        for (int o = 0; o < num; o++) {
+            rb_zjit_add_frame(frames, zjit_raw_samples[idx]);
+            rb_ary_push(raw_samples, SIZET2NUM(zjit_raw_samples[idx]));
+            rb_ary_push(line_samples, INT2NUM(zjit_line_samples[idx]));
+            idx++;
+        }
+
+        rb_ary_push(raw_samples, SIZET2NUM(zjit_raw_samples[idx]));
+        rb_ary_push(line_samples, INT2NUM(zjit_line_samples[idx]));
+        idx++;
+
+        rb_ary_push(raw_samples, SIZET2NUM(zjit_raw_samples[idx]));
+        rb_ary_push(line_samples, INT2NUM(zjit_line_samples[idx]));
+        idx++;
+    }
+
+    // Set add the raw_samples, line_samples, and frames to the results
+    // hash.
+    rb_hash_aset(result, ID2SYM(rb_intern("raw")), raw_samples);
+    rb_hash_aset(result, ID2SYM(rb_intern("lines")), line_samples);
+    rb_hash_aset(result, ID2SYM(rb_intern("frames")), frames);
+
+    return result;
+}
+
 void rb_zjit_profile_disable(const rb_iseq_t *iseq);
 
 void
@@ -217,6 +306,8 @@ VALUE rb_zjit_stats(rb_execution_context_t *ec, VALUE self, VALUE target_key);
 VALUE rb_zjit_reset_stats_bang(rb_execution_context_t *ec, VALUE self);
 VALUE rb_zjit_stats_enabled_p(rb_execution_context_t *ec, VALUE self);
 VALUE rb_zjit_print_stats_p(rb_execution_context_t *ec, VALUE self);
+VALUE rb_zjit_trace_exit_locations_enabled_p(rb_execution_context_t *ec, VALUE self);
+VALUE rb_zjit_get_exit_locations(rb_execution_context_t *ec, VALUE self);
 
 // Preprocessed zjit.rb generated during build
 #include "zjit.rbinc"
--- a/zjit.rb
+++ b/zjit.rb
@@ -9,7 +9,10 @@
 module RubyVM::ZJIT
   # Avoid calling a Ruby method here to avoid interfering with compilation tests
   if Primitive.rb_zjit_print_stats_p
-    at_exit { print_stats }
+    at_exit {
+      print_stats
+      dump_locations
+    }
   end
 end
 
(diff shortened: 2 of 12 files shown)


aidenfoxivey:
I'm torn right now whether to keep the same approach here as was made in YJIT. Right now, I have a pretty much entirely copied implementation from it - but I think there is potentially room for different methods.

aidenfoxivey:
Earlier, I had some discussion about the option of using `OnlyCell`.

zjit.c
```diff
@@ -31,6 +31,95 @@
...
+    }
+}
+
+// Parses the ZjitExitLocations raw_samples and line_samples collected by
+// rb_zjit_record_exit_stack and turns them into 3 hashes (raw, lines, and frames) to
+// be used by RubyVM::ZJIT.exit_locations. zjit_raw_samples represents the raw frames information
+// (without name, file, and line), and zjit_line_samples represents the line information
+// of the iseq caller.
+VALUE
+rb_zjit_exit_locations_dict(VALUE *zjit_raw_samples, int *zjit_line_samples, int samples_len)
```
k0kubun, line 77: WDYT about sharing `rb_zjit_exit_locations_dict` and `rb_zjit_add_frame` among YJIT and ZJIT by moving them to `jit.c`?

zjit.rb
```diff
@@ -19,6 +22,106 @@ def enabled?
     Primitive.cexpr! 'RBOOL(rb_zjit_enabled_p)'
   end
 
+  # Check if `--zjit-trace-exits` is used
+  def trace_exit_locations_enabled?
+      Primitive.rb_zjit_trace_exit_locations_enabled_p
```
k0kubun:
```suggestion
    Primitive.rb_zjit_trace_exit_locations_enabled_p
```

zjit/src/backend/lir.rs
```diff
@@ -1630,6 +1631,14 @@ impl Assembler
                         self.load_into(SCRATCH_OPND, Opnd::const_ptr(exit_counter_ptr_for_opcode(opcode)));
                         self.incr_counter_with_reg(Opnd::mem(64, SCRATCH_OPND, 0), 1.into(), C_RET_OPND);
                     }
+
+                    if !pc.is_null() && get_option!(dump_side_exits) {
```
k0kubun: This is inside `if get_options!(stats) {`, but what part of side-exit dump relies on stats being enabled?

I know `--yjit-trace-exits` requires (and automatically implies) `--yjit-stats` to be enabled, but I'm wondering if we can actually cut the dependency between them.

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+    // Trim at `stack_length` since anything past it is redundant
+    frames_buffer.truncate(stack_length as usize);
+    lines_buffer.truncate(stack_length as usize);
+
+    (stack_length, frames_buffer, lines_buffer)
+}
+
+/// Record a backtrace with ZJIT side exits
+#[unsafe(no_mangle)]
+pub extern "C" fn rb_zjit_record_exit_stack(_exit_pc: *const VALUE)
```
k0kubun: I think `_exit_pc` should be renamed to `exit_pc`.

Also, it might be too much, but I'm tempted to share this under `jit/` too, given how big these functions are and the fact that they likely don't need YJIT/ZJIT-specific logic in the aggregation itself?

zjit/src/backend/lir.rs
k0kubun: Also, now that we've merged https://github.com/ruby/ruby/pull/14653, I think you can remove `pc.is_null()` if you rebase from master.

aidenfoxivey: Sure, yeah, let me move it out.

zjit/src/state.rs
aidenfoxivey: Hmm, actually a good idea - let me try that.

aidenfoxivey:
Weirdly, it was a total pain to try to centralize the code into `jit.c`.

tekknolagi:
I would say maybe don't spend too much more time trying to share all the code between YJIT and ZJIT if it's a huge pain. More important to get this working, landed, and maybe merge down the line

zjit/src/backend/lir.rs
```diff
@@ -1629,6 +1630,14 @@ impl Assembler
                     }
                 }
 
+                if get_option!(dump_side_exits) {
+                    self.load_into(C_ARG_OPNDS[0], Opnd::const_ptr(pc as *const u8));
+                    self.ccall(
+                         rb_zjit_record_exit_stack as *const u8,
+                         vec![]
+                    );
```
tekknolagi, line 1640: unless there is a specific reason you do this, can you put it as an element in the `vec![]`? that is what it's there for

zjit/src/state.rs
```diff
@@ -184,6 +194,16 @@ impl ZJITState {
     pub fn get_function_stub_hit_trampoline() -> CodePtr {
         ZJITState::get_instance().function_stub_hit_trampoline
     }
+
+    /// Get a mutable reference to the yjit raw samples Vec
+    pub fn get_raw_samples() -> Option<&'static mut Vec<VALUE>> {
+        ZJITState::get_instance().exit_locations.as_mut().map(|el| &mut el.raw_samples)
+    }
+
+    /// Get a mutable reference to yjit the line samples Vec.
```
tekknolagi: s/yjit/zjit/g

tekknolagi: also move the `the` around

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
     ZJITState::enable_assert_compiles();
     Qnil
 }
+
+/// Call `rb_profile_frames` and format the result into buffers.
```
tekknolagi: can you add a little more detail about how things are formatted and into what buffers and ...? in the doc comment

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+fn record_profiling_frames() -> (i32, Vec<VALUE>, Vec<i32>) {
+    // Use the same buffer size as Stackprof.
```
tekknolagi: link to where stackprof uses this buffer size and explain if it is critical that we have the same buff size?

zjit/src/backend/lir.rs
```diff
@@ -1629,6 +1630,14 @@ impl Assembler
                     }
                 }
 
+                if get_option!(dump_side_exits) {
```
tekknolagi: can we get consistent dump vs trace terminology for this PR?

tekknolagi: probably trace

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+    }
+
+    let (stack_length, frames_buffer, lines_buffer) = record_profiling_frames();
+
+    // Can safely unwrap since `dump_side_exits` must be true at this point
+    let zjit_raw_samples = ZJITState::get_raw_samples().unwrap();
+    let zjit_lines_samples = ZJITState::get_line_samples().unwrap();
+    assert_eq!(zjit_raw_samples.len(), zjit_lines_samples.len());
+
+    // todo(aidenfoxivey): why is this +3 here
```
tekknolagi: nit: TODO

and didn't you find this out?

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+    let samples_length = (stack_length as usize) + 3;
+
+    // If zjit_raw_samples is less than or equal to the current length of the samples
+    // we might have seen this stack trace previously.
+    if zjit_raw_samples.len() >= samples_length {
```
tekknolagi: can you use the same name as the variable (if you are referring to it)? just basically make sure the comments visibly match the code

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+    zjit_raw_samples.push(VALUE(stack_length as usize));
+    zjit_lines_samples.push(stack_length);
+
+    frames_buffer.iter().zip(lines_buffer.iter()).rev().for_each(|(frame, line)| {
+        zjit_raw_samples.push(*frame);
+        zjit_lines_samples.push(*line);
+    });
+
+    // Get the opcode from instruction handler at exit PC.
+    let insn = unsafe { rb_vm_insn_addr2opcode((*exit_pc).as_ptr()) };
```
tekknolagi: Why is it ok to cast an insn as a VALUE? Or what is going on here? It seems really messy

tekknolagi: Why are there VALUEs all over the place?

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+
+    // Push number of times seen onto the stack, which is 1
+    // because it's the first time we've seen it.
+    zjit_raw_samples.push(VALUE(1_usize));
+    zjit_lines_samples.push(1);
+}
+
+/// Mark `raw_samples` so they can be used by rb_yjit_add_frame.
+pub fn gc_mark_raw_samples() {
+    // Return if YJIT is not enabled
```
tekknolagi: zjit

zjit/src/state.rs
```diff
@@ -219,3 +239,139 @@ pub extern "C" fn rb_zjit_assert_compiles(_ec: EcPtr, _self: VALUE) -> VALUE {
...
+    // We don't know the line that this instruction sits at
+    zjit_lines_samples.push(0);
+
+    // Push number of times seen onto the stack, which is 1
+    // because it's the first time we've seen it.
+    zjit_raw_samples.push(VALUE(1_usize));
+    zjit_lines_samples.push(1);
+}
+
+/// Mark `raw_samples` so they can be used by rb_yjit_add_frame.
```
tekknolagi: zjit

zjit/src/stats.rs
```diff
@@ -400,3 +400,52 @@ pub fn with_time_stat<F, R>(counter: Counter, func: F) -> R where F: FnOnce() ->
...
+    }
+
+    // Can safely unwrap since `dump_side_exits` must be true at this point
+    let zjit_raw_samples = ZJITState::get_raw_samples().unwrap();
+    let zjit_line_samples = ZJITState::get_line_samples().unwrap();
+
+    assert_eq!(zjit_raw_samples.len(), zjit_line_samples.len());
+
+    // zjit_raw_samples and zjit_line_samples are the same length so
+    // pass only one of the lengths in the C function.
```
tekknolagi, line 504: this comment can (also) be an assert

zjit/src/backend/lir.rs
k0kubun: The args work only when the split pass turns it into `ParallelMov`, and `compile_side_exits` is after the split pass, so you can't do that here.

This should probably have a comment to explain that.

zjit/src/backend/lir.rs
```diff
@@ -1629,6 +1630,16 @@ impl Assembler
                     }
                 }
 
+                if get_option!(trace_side_exits) {
+                    // Use `load_into` with `C_ARG_OPNDS` instead of `opnds` argument for ccal, since `compile_side_exits`
```
tekknolagi: nit: `ccall` (two `l`)

zjit/src/backend/lir.rs
```diff
@@ -2080,4 +2091,3 @@ mod tests {
         asm.load_into(mem, mem);
     }
 }
-
```
tekknolagi, line 2083: nit: don't make unrelated whitespace changes

aidenfoxivey:
When approved and ready to ship, I will condense the commits.

Here's evidence that Stackprof can use it:

https://asciinema.org/a/54avfhLgkGPahjTCAY7AV0hPq

tekknolagi:
let's ship it. i can just squash it when merging -- no need to do it manually

tekknolagi:
just make the PR description what you want your commit message to be

aidenfoxivey:
> just make the PR description what you want your commit message to be

Done!

tekknolagi:
why squash?

aidenfoxivey:
> why squash?

Oh, good question - I thought that was preferred.

k0kubun:
In this case, Max was going to squash-merge your PR on auto-merge anyway, and your force-push to squash it cancelled the auto-merge.

Screenshot 2025-09-30 at 09 47 37

---

You might want to squash and force-push changes in the following situations (neither of them applies this time):

* You want to push multiple commits into master from a single PR
* You want to format a long descriptive commit message yourself

but when you just want them to be squashed into a single commit with the commit message of just the PR subject, there's no problem in leaving it for the person merging the PR to do that.

aidenfoxivey:
> In this case, Max was going to squash-merge your PR on auto-merge anyway, and your force-push to squash it cancelled the auto-merge.
> 
> Screenshot 2025-09-30 at 09 47 37
> You might want to squash and force-push changes in the following situations (neither of them applies this time):
> 
> * You want to push multiple commits into master from a single PR
> * You want to format a long descriptive commit message yourself
> 
> but when you just want them to be squashed into a single commit with the commit message of just the PR subject, there's no problem in leaving it for the person merging the PR to do that.

Ohh, thank you for the explanation @k0kubun. I had seen that before but not really known what that meant. The teams I've worked on before were big on "fix up your commits to look pretty and then force push".

zjit/src/options.rs
```diff
@@ -235,6 +241,10 @@ fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
             options.print_stats = false;
         }
 
+        ("trace-exits", "") => {
+            options.trace_side_exits = true;
```
st0012, line 245: Does it make sense to also set `options.stats = true;` here so `--zjit-stats` doesn't need to specified separately?
I feel if it's a hard dependency we can just enable both of them together.

k0kubun: As I raised at https://github.com/ruby/ruby/pull/14640#discussion_r2380501403 (which wasn't really addressed for the whole patch), I still wonder if we really need the dependency between them in the first place. 

If it just relies on `rb_zjit_record_exit_stack` calls and it works without stats, we should be able to just get rid of `get_option!(stats)` guards instead of letting `--zjit-trace-exits` depend on `--zjit-stats`.

PR #14642, opened 2025-09-23
fix lvar_state dump size

ko1:
`ibf_dump_write()` should consider the size of the element.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -13272,7 +13272,7 @@ ibf_dump_lvar_states(struct ibf_dump *dump, const rb_iseq_t *iseq)
     const struct rb_iseq_constant_body *const body = ISEQ_BODY(iseq);
     const int size = body->local_table_size;
     IBF_W_ALIGN(enum lvar_state);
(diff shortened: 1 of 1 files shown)


compile.c
```diff
@@ -13272,7 +13272,7 @@ ibf_dump_lvar_states(struct ibf_dump *dump, const rb_iseq_t *iseq)
     const struct rb_iseq_constant_body *const body = ISEQ_BODY(iseq);
     const int size = body->local_table_size;
     IBF_W_ALIGN(enum lvar_state);
-    return ibf_dump_write(dump, body->lvar_states, body->lvar_states ? size : 0);
+    return ibf_dump_write(dump, body->lvar_states, sizeof(enum lvar_state) * (body->lvar_states ? size : 0));
```
nobu, line 13275: This `if_dump_write` can be skipped totally if `!body->lvar_states`?

ko1: except `IBF_W_ALIGN`?

PR #14643, opened 2025-09-24
ZJIT: Incorporate bb0-prologue and PC check into HIR

k0kubun:
This PR changes PC checks for optional arguments and `bb0-prologue` to HIR-based implementations instead of hard-coding their LIR at `gen_entry_prologue` and `gen_function` for `BlockId(0)` respectively.

This should help us implement separate basic blocks for optional arguments and report their addresses using `JITEntry`. However, given that it's already a huge diff due to HIR tests, supporting optional arguments is out of scope in this PR.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -657,7 +657,7 @@ pub enum Insn {
         return_type: Option<Type>,  // None for unannotated builtins
     },
 
-    /// Set up frame and remember the address as a JIT entry if insn_idx is Some
+    /// Set up frame and remember the address as a JIT entry if jit_entry is true
     EntryPoint { jit_entry: bool },
     /// Control flow instructions
     Return { val: InsnId },
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -4368,37 +4368,46 @@ mod validation_tests {
     #[test]
     fn instruction_appears_twice_in_same_block() {
         let mut function = Function::new(std::ptr::null());
-        let entry = function.entry_block;
-        let jit_entry = function.jit_entry_block;
-        function.push_insn(jit_entry, Insn::Jump(BranchEdge { target: entry, args: vec![] }));
-        let val = function.push_insn(entry, Insn::Const { val: Const::Value(Qnil) });
-        function.push_insn_id(entry, val);
-        function.push_insn(entry, Insn::Return { val });
-        assert_matches_err(function.validate(), ValidationError::DuplicateInstruction(entry, val));
+        let block = function.new_block(0);
+
+        // Add jumps from entry blocks to avoid BlockHasNoterminator
+        function.push_insn(function.entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+        function.push_insn(function.jit_entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+
+        let val = function.push_insn(block, Insn::Const { val: Const::Value(Qnil) });
+        function.push_insn_id(block, val);
+        function.push_insn(block, Insn::Return { val });
+        assert_matches_err(function.validate(), ValidationError::DuplicateInstruction(block, val));
     }
 
     #[test]
     fn instruction_appears_twice_with_different_ids() {
         let mut function = Function::new(std::ptr::null());
-        let entry = function.entry_block;
-        let jit_entry = function.jit_entry_block;
-        function.push_insn(jit_entry, Insn::Jump(BranchEdge { target: entry, args: vec![] }));
-        let val0 = function.push_insn(entry, Insn::Const { val: Const::Value(Qnil) });
-        let val1 = function.push_insn(entry, Insn::Const { val: Const::Value(Qnil) });
+        let block = function.new_block(0);
+
+        // Add jumps from entry blocks to avoid BlockHasNoterminator
+        function.push_insn(function.entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+        function.push_insn(function.jit_entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+
+        let val0 = function.push_insn(block, Insn::Const { val: Const::Value(Qnil) });
+        let val1 = function.push_insn(block, Insn::Const { val: Const::Value(Qnil) });
         function.make_equal_to(val1, val0);
-        function.push_insn(entry, Insn::Return { val: val0 });
-        assert_matches_err(function.validate(), ValidationError::DuplicateInstruction(entry, val0));
+        function.push_insn(block, Insn::Return { val: val0 });
+        assert_matches_err(function.validate(), ValidationError::DuplicateInstruction(block, val0));
     }
 
     #[test]
     fn instruction_appears_twice_in_different_blocks() {
         let mut function = Function::new(std::ptr::null());
-        let entry = function.entry_block;
-        let jit_entry = function.jit_entry_block;
-        function.push_insn(jit_entry, Insn::Jump(BranchEdge { target: entry, args: vec![] }));
-        let val = function.push_insn(entry, Insn::Const { val: Const::Value(Qnil) });
+        let block = function.new_block(0);
+
+        // Add jumps from entry blocks to avoid BlockHasNoterminator
+        function.push_insn(function.entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+        function.push_insn(function.jit_entry_block, Insn::Jump(BranchEdge { target: block, args: vec![] }));
+
+        let val = function.push_insn(block, Insn::Const { val: Const::Value(Qnil) });
         let exit = function.new_block(0);
-        function.push_insn(entry, Insn::Jump(BranchEdge { target: exit, args: vec![] }));
+        function.push_insn(block, Insn::Jump(BranchEdge { target: exit, args: vec![] }));
         function.push_insn_id(exit, val);
         function.push_insn(exit, Insn::Return { val });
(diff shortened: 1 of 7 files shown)


zjit/src/hir_type/mod.rs
```diff
@@ -288,7 +289,7 @@ impl Type {
     /// `Type::from_cint(types::CUInt16, 12)`.
     pub fn from_cint(ty: Type, val: i64) -> Type {
```
tekknolagi, line 277: We should probably have a `from_ptr` instead of `from_cint` and not try and smuggle pointers in that way

zjit/src/hir.rs
```diff
@@ -652,6 +657,8 @@ pub enum Insn {
         return_type: Option<Type>,  // None for unannotated builtins
     },
 
+    /// Set up frame and remember the address as a JIT entry if insn_idx is Some
+    EntryPoint { jit_entry: bool },
```
tekknolagi: There's no `insn_idx` in this struct-like enum so the comment is a little hard to understand / misleading

zjit/src/hir.rs
```diff
@@ -1457,6 +1477,7 @@ impl Function {
             &ArrayMax { ref elements, state } => ArrayMax { elements: find_vec!(elements), state: find!(state) },
             &SetGlobal { id, val, state } => SetGlobal { id, val: find!(val), state },
             &GetIvar { self_val, id, state } => GetIvar { self_val: find!(self_val), id, state },
+            LoadPC => LoadPC,
```
tekknolagi: This could go up top in the shallow clone case

zjit/src/hir.rs
```diff
@@ -1468,6 +1489,7 @@ impl Function {
             &ArrayExtend { left, right, state } => ArrayExtend { left: find!(left), right: find!(right), state },
             &ArrayPush { array, val, state } => ArrayPush { array: find!(array), val: find!(val), state },
             &CheckInterrupts { state } => CheckInterrupts { state },
+            &EntryPoint { jit_entry } => EntryPoint { jit_entry },
```
tekknolagi: as with LoadPC

zjit/src/hir.rs
```diff
@@ -2579,6 +2606,7 @@ impl Function {
     pub fn rpo(&self) -> Vec<BlockId> {
         let mut result = self.po_from(self.entry_block);
         result.reverse();
+        result.insert(1, self.jit_entry_block); // Put jit_entry_block after entry_block
```
tekknolagi, line 2630: Why do we need to do this if the interpreter entry jumps to the jit entry?

zjit/src/hir.rs
```diff
@@ -4315,6 +4391,8 @@ mod validation_tests {
     fn instruction_appears_twice_in_different_blocks() {
         let mut function = Function::new(std::ptr::null());
         let entry = function.entry_block;
+        let jit_entry = function.jit_entry_block;
+        function.push_insn(jit_entry, Insn::Jump(BranchEdge { target: entry, args: vec![] }));
```
tekknolagi: Shouldn't the entry jump to the jit entry instead?

zjit/src/hir.rs
```diff
@@ -4304,6 +4378,8 @@ mod validation_tests {
     fn instruction_appears_twice_with_different_ids() {
         let mut function = Function::new(std::ptr::null());
         let entry = function.entry_block;
+        let jit_entry = function.jit_entry_block;
+        function.push_insn(jit_entry, Insn::Jump(BranchEdge { target: entry, args: vec![] }));
```
tekknolagi: entry -> jit entry

zjit/src/hir.rs
```diff
@@ -4694,10 +4848,16 @@ mod tests {
         assert_contains_opcode("test", YARVINSN_newhash);
         assert_snapshot!(hir_string("test"), @r"
         fn test@<compiled>:1:
-        bb0(v0:BasicObject):
-          v5:HashExact = NewHash
+        bb0(v1:BasicObject):
+          EntryPoint interpreter
+          Jump bb2(v1)
+        bb1(v4):
```
tekknolagi: It's a little odd that the parameters to the JIT entrypoint don't get flow typed. I wonder why that is. They should get `BasicObject`

zjit/src/hir.rs
k0kubun: From other comments in this PR, I'm not sure if you understand the intended relationship between blocks.

It's implemented as:

```
bb0 (entry_block) ---→ bb2 (first_block, main CFG)
                       ↑
bb1 (jit_entry_block) -┘
```

Your understanding seems to be:

```
bb0 (entry_block)      bb2 (first_block, main CFG)
      ↓                ↑
bb1 (jit_entry_block) -┘
```

which is not the case.

Because there are two ways to enter JIT code, retrieving the graph from one entry is not enough.

zjit/src/hir.rs
k0kubun: May I ask the motivation for it?

I just chose the path to minimize the diff in this tests, but I can add another basic block that `entry` and `jit_entry` jump to, which would be closer to the actual relationship between blocks (like https://github.com/ruby/ruby/pull/14643#discussion_r2377061199). Is that what you want (not the one you suggested here)?

zjit/src/hir.rs
k0kubun: ditto https://github.com/ruby/ruby/pull/14643#discussion_r2377068160

zjit/src/hir.rs
k0kubun: Yeah sorry, it was an outdated comment I forgot to fix https://github.com/ruby/ruby/pull/14643/commits/adeb771c9895f3eac4b6b8e5e852015cd22edd3c

zjit/src/hir.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/14643/commits/1b6528aa09ded8e7575287994e52351f8b50783e

zjit/src/hir.rs
k0kubun: fixed in https://github.com/ruby/ruby/pull/14643/commits/1b6528aa09ded8e7575287994e52351f8b50783e

zjit/src/hir.rs
tekknolagi: You might need

```diff
+        for (param, param_type) in std::iter::zip(entry_params, param_types.clone()) {
+            // We know that function parameters are BasicObject or some subclass
+            self.insn_types[param.0] = *param_type;
+        }
+
+        // Fill JIT entry parameter types
+        let entry_params = self.blocks[self.jit_entry_block.0].params.iter();
+        assert_eq!(
+            entry_params.len(),
+            entry_params.len(),
+            "param types should be initialized before type inference"
+        );
```

zjit/src/hir.rs
k0kubun: addressed this in https://github.com/ruby/ruby/pull/14643/commits/99655554130a6fd95d9aa220fae8977e73d67e7d

zjit/src/hir.rs
k0kubun: addressed this in https://github.com/ruby/ruby/pull/14643/commits/99655554130a6fd95d9aa220fae8977e73d67e7d

zjit/src/hir.rs
```diff
@@ -2582,6 +2627,7 @@ impl Function {
     pub fn rpo(&self) -> Vec<BlockId> {
         let mut result = self.po_from(self.entry_block);
         result.reverse();
+        result.insert(1, self.jit_entry_block); // Put jit_entry_block after entry_block
```
tekknolagi, line 2630: Note to self for later: let's do `po_from(vec)` or something

zjit/src/hir_type/mod.rs
k0kubun: https://github.com/ruby/ruby/pull/14643/commits/81366f0fa6cbb0b404ccb6c7a1c31355aa5732ce

zjit/src/hir.rs
```diff
@@ -3214,16 +3261,20 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
         insn_idx_to_block.insert(insn_idx, fun.new_block(insn_idx));
     }
 
     // Check if the EP is escaped for the ISEQ from the beginning. We give up
     // optimizing locals in that case because they're shared with other frames.
     let ep_escaped = iseq_escapes_ep(iseq);
 
+    // Compile an interpreter entry to the first block
+    let first_block = insn_idx_to_block[&0];
+    compile_entry_block(fun.entry_block, first_block, &mut fun, false);
```
tekknolagi, line 3273: For later: let's get rid of the bool params

zjit/src/hir.rs
k0kubun: https://github.com/ruby/ruby/pull/14643/commits/a31c8cbdee4c9007b307338a7d9cbae8484481ad

tekknolagi (approved):

zjit/src/hir.rs
k0kubun: What would you replace it with?

k0kubun: We'll need a "JIT entry index" (index for `opt_table` and `jit_entry_blocks`) in `Insn::EntryPoint`, so `bool` would at least need to be `Option<usize>` I guess.

k0kubun: Well, it's going to be so different between them that it probably doesn't make sense to share them anymore. I guess we'll just implement them in different functions.

PR #14650, opened 2025-09-24
vm_dump.c: unbreak unwind on powerpc

barracuda156:
Fix the build accidentally broken by https://github.com/ruby/ruby/commit/229784d160452dcdaa85ae4b5c5fae5c2728324b

@nobu @ioquatix Do we need something more specific here?

For the reference, `libunwind.h` of 10.6: https://github.com/alexey-lysiuk/macos-sdk/blob/c47669c5dac76e2f15970587c5a7274c09631aa2/MacOSX10.6.sdk/usr/include/libunwind.h

P. S. It will be nice to have this fixed before 3.5 release, if possible.

### DIFF
--- a/vm_dump.c
+++ b/vm_dump.c
@@ -770,7 +770,7 @@ rb_vmdebug_thread_dump_state(FILE *errout, VALUE self)
 #  include <sys/mman.h>
 #  undef backtrace
 
-#  if defined(__arm64__)
+#  if defined(__arm64__) || defined(__POWERPC__)
 static bool
 is_coroutine_start(unw_word_t ip)
 {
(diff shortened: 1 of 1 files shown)


ioquatix:
It looks okay to me but I'm unable to test it.

nobu:
I think we can unify non-x86 cases.

barracuda156:
> I think we can unify non-x86 cases.

@nobu Should this be used on ppc too? `trace[n++] = (void *)(ip & 0x7fffffffffffull);`

nobu:
> > I think we can unify non-x86 cases.
> 
> @nobu Should this be used on ppc too? `trace[n++] = (void *)(ip & 0x7fffffffffffull);`

It is ARM64 specific code; Pointer Authentication Stripping.
So that part needs to be wrapped in a macro, that returns the argument as-is on other than ARM64.

barracuda156:
@nobu Could you please take a look?

barracuda156:
@ioquatix @nobu Are we good to merge this?

ioquatix:
I am okay to merge this. I'm assuming @barracuda156 you've tested it on PowerPC.

ioquatix (approved):

barracuda156:
@ioquatix Yes, it worked.

PR #14651, opened 2025-09-24
ZJIT: Support VM_CALL_ARGS_SPLAT

tekknolagi:


k0kubun:
We probably need a new test to reproduce the issue in binarytrees https://github.com/Shopify/ruby/issues/749#issuecomment-3330657555

PR #14653, opened 2025-09-25
ZJIT: Compile ISEQ with optional arguments (WIP)

k0kubun:


zjit/src/codegen.rs
```diff
@@ -307,8 +308,14 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Resul
         }
     }
     result.map(|(start_ptr, gc_offsets)| {
-        let jit_entry_ptr = jit.jit_entry.unwrap().borrow().start_addr.get().unwrap();
-        (start_ptr, jit_entry_ptr, gc_offsets, jit.iseq_calls)
+        // Make sure jit_entry_ptrs can be used as a parallel vector to jit_entry_insns()
+        jit.jit_entries.sort_by(|a, b|
+            a.borrow().jit_entry_idx.partial_cmp(&b.borrow().jit_entry_idx).unwrap_or(Ordering::Equal)
```
aidenfoxivey: If I understand, this is effectively the `<` operator but expanded. I would imagine the alternative is to derive `PartialEq`.

1. Am I correct in my reasoning?
2. If so, why choose this method?

zjit/src/codegen.rs
```diff
@@ -307,8 +308,14 @@ fn gen_function(cb: &mut CodeBlock, iseq: IseqPtr, function: &Function) -> Resul
...
+        );
+        let jit_entry_ptrs = jit.jit_entries.iter().map(|jit_entry|
+            jit_entry.borrow().start_addr.get().unwrap()
```
aidenfoxivey: How do we know that this unwrap won't cause a panic here?

zjit/src/codegen.rs
```diff
@@ -1775,22 +1781,12 @@ fn compile_iseq(iseq: IseqPtr) -> Result<Function, CompileError> {
 
     let mut function = match iseq_to_hir(iseq) {
         Ok(function) => function,
-        Err(err) => {
-            let name = crate::cruby::iseq_get_location(iseq, 0);
-            debug!("ZJIT: iseq_to_hir: {err:?}: {name}");
```
aidenfoxivey, line 1780: Would having a debug statement like this help with debugging should the error case be taken?

zjit/src/codegen.rs
```diff
@@ -1313,12 +1320,11 @@ fn gen_object_alloc_class(asm: &mut Assembler, class: VALUE, state: &FrameState)
     }
 }
 
-/// Compile a frame setup. If is_jit_entry is true, remember the address of it as a JIT entry.
-fn gen_entry_point(jit: &mut JITState, asm: &mut Assembler, is_jit_entry: bool) {
-    if is_jit_entry {
-        assert!(jit.jit_entry.is_none(), "only one jit_entry is expected");
```
aidenfoxivey, line 1319: Intuitively, this line here doesn't make a lot of sense. (Not that that's a problem now.)

If jit_entry was None, then how would there be one of them?

zjit/src/codegen.rs
```diff
@@ -1775,22 +1781,12 @@ fn compile_iseq(iseq: IseqPtr) -> Result<Function, CompileError> {
...
-            return Err(CompileError::ParseError(err));
-        }
+        Err(err) => return Err(CompileError::ParseError(err)),
     };
     if !get_option!(disable_hir_opt) {
         function.optimize();
     }
     function.dump_hir();
-    #[cfg(debug_assertions)]
-    if let Err(err) = function.validate() {
```
aidenfoxivey, line 1789: Is this call to `validate` not necessary?

aidenfoxivey (approved):
Looks good to me! I left a few questions for my own learning.

zjit/src/codegen.rs
k0kubun: We want them to be sorted explicitly by `jit_entry_idx` here, so deriving the ordering is not appropriate (and not even possible due to `Cell<Option<CodePtr>>`). 

You could still implement `Ord` (and `PartialOrd` required by it) to sort them with `jit.jit_entries.sort()`, but then it becomes less ambiguous that we're sorting them based on `jit_entry_idx` here. I want to make it clear that it's sorted by `jit_entry_idx` here.

I actually realized there was a better function for this, so I fixed it https://github.com/ruby/ruby/pull/14653/commits/ed48ed99cd01662def3e4e724dd91354909b0e74. I looked for and used `sort_by` only because it's the `sort_by_key` for Ruby :p

zjit/src/codegen.rs
k0kubun: I'm using it more like an assertion here. Something's wrong in the implementation if it panics, so we do want the panic when it does. 

Every instance created by `JITEntry::new` is baked in a `pos_marker`. It's _always_ called and sets `Some` to `start_addr`, as long as `asm.compile()` succeeds. We're inside `result.map`, so `asm.compile()` didn't fail. This `unwrap()` is meant to assert that this logic holds, but if you think it's unreasonable, I could add a better panic message here.

zjit/src/codegen.rs
k0kubun: You're right. I meant to just move this to the `debug!` after `fun.validate()` in `iseq_to_hir`, but it doesn't cover other errors, so I should have kept it. I resurrected it and made the code a bit shorter https://github.com/ruby/ruby/pull/14653/commits/219b41f5069e94920914817cc92525b5eeb98e01.

zjit/src/codegen.rs
k0kubun: I removed it because it's a duplicated check.

When `disable_hir_opt` is set, `iseq_to_hir` does the same thing by itself. When `disable_hir_opt` is not set, `optimize()` calls `assert_validates()` on `#[cfg(debug_assertions)]`, so this is unreachable in that case.

zjit/src/codegen.rs
k0kubun: It was there to prevent `gen_entry_point` from overwriting `Some` set by something else, but now that the `Option` is changed to `Vec`, nothing is going to be overwritten by `push`ing an entry to it. So we no longer need to assert no overwrites.

> If jit_entry was None, then how would there be one of them?

by the next `jit.jit_entry = Some(jit_entry.clone());` (in the deleted diff) right? It was called with `is_jit_entry` only once prior to this PR, and now it's called multiple times (see `JIT(0)`, `JIT(1)`, ... in HIR), so yes there would be ones.

zjit/src/codegen.rs
aidenfoxivey: After hearing the explanation, that does make sense to me. I think it should be good as is, though it probably wouldn't hurt to use an `.expect("failure reason")` alternatively.

zjit/src/codegen.rs
aidenfoxivey: Ah right, okay - that makes more sense.

zjit/src/codegen.rs
k0kubun: corrected my laziness https://github.com/ruby/ruby/pull/14653/commits/4c3329cb325f01d40dcdbf94a873bb6bfd6ecc6e 🙂

zjit/src/hir.rs
```diff
@@ -1607,40 +1611,60 @@ impl Function {
...
+    fn copy_param_types(&mut self) {
+        for entry_block in self.entry_blocks() {
+            let entry_params = self.blocks[entry_block.0].params.iter();
+            let param_types = self.param_types.iter();
+            assert_eq!(
+                entry_params.len(),
+                param_types.len(),
+                "param types should be initialized before type inference",
+            );
+            for (param, param_type) in std::iter::zip(entry_params, param_types.clone()) {
```
XrXr:
```suggestion
            for (param, param_type) in std::iter::zip(entry_params, param_types) {
```
Shouldn't need to clone the iterator

zjit/src/hir.rs
```diff
@@ -4116,43 +4158,99 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
-        fun.push_insn(entry_block, Insn::IfTrue {
-            val: test_id,
-            target: BranchEdge { target: first_block, args: entry_state.as_args(self_param) },
-        });
-
-        // Side-exit if any optional argument is supplied
-        let state = fun.push_insn(entry_block, Insn::Snapshot { state: FrameState::new(iseq) }); // no need to spill params
-        fun.push_insn(entry_block, Insn::SideExit { state, reason: SideExitReason::OptionalArgumentsSupplied });
+/// Set param_types of the function. They are copied to the param types of entry blocks.
+fn set_param_types(fun: &mut Function) {
```
XrXr: This seems to be duplicate code with Function::set_param_types. Let's keep just the method.

zjit/src/hir.rs
```diff
@@ -3258,79 +3309,61 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
-    let first_block = insn_idx_to_block[&0];
-    compile_entry_block(fun.entry_block, first_block, &mut fun, false);
 
-    // Compile a JIT entry to the first block
-    compile_entry_block(fun.jit_entry_block, first_block, &mut fun, true);
+    // Compile an entry_block for the interpreter
+    compile_entry_block(&mut fun, &jit_entry_insns);
 
     // Iteratively fill out basic blocks using a queue
-    // TODO(max): Basic block arguments at edges
```
XrXr, line 3279: Don't think you meant to delete this TODO.

XrXr (commented):
Looks pretty good to me on a quick pass. Leaving a more detailed review for later/someone else.

zjit/src/hir.rs
k0kubun: added it back https://github.com/ruby/ruby/pull/14653/commits/23546770af4f3e9ce50730f17b78e80ea64ef856

zjit/src/hir.rs
k0kubun: Good catch https://github.com/ruby/ruby/pull/14653/commits/894b8de6138da9c098ed6b4ee7e62b1dc1d83b1a. I'm kind of surprised that Rust doesn't warn me on it while it's unused and not even pub, but I guess it's due to `![allow(dead_code)]` sitting at lib.rs.

zjit/src/hir.rs
k0kubun: thanks, this was a leftover from the version where it was in a loop. https://github.com/ruby/ruby/pull/14653/commits/4a7d0e4321bbb0fc194c819b3295a9c4d12498e9

zjit/src/hir.rs
```diff
@@ -936,7 +935,11 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             }
-            &Insn::EntryPoint { jit_entry } => write!(f, "EntryPoint {}", if jit_entry { "JIT" } else { "interpreter" }),
+            &Insn::EntryPoint { jit_entry_idx } => if let Some(jit_entry_idx) = jit_entry_idx {
+                write!(f, "EntryPoint JIT({jit_entry_idx})")
+            } else {
+                write!(f, "EntryPoint interpreter")
+            },
```
```suggestion
            &Insn::EntryPoint { jit_entry_idx: Some(idx) } => write!(f, "EntryPoint JIT({jit_entry_idx})"),
             &Insn::EntryPoint { jit_entry_idx: None } => write!(f, "EntryPoint interpreter"),
```

zjit/src/hir.rs
```diff
@@ -3125,16 +3155,36 @@ fn insn_idx_at_offset(idx: u32, offset: i64) -> u32 {
     ((idx as isize) + (offset as isize)) as u32
 }
 
+/// List of insn_idx that starts a JIT entry block
+fn jit_entry_insns(iseq: IseqPtr) -> Vec<u32> {
+    let opt_num = unsafe { get_iseq_body_param_opt_num(iseq) };
+    if opt_num > 0 {
+        // Deduplicate entries with HashSet since opt_table may have duplicated entries, e.g. proc { |a=a| a }
+        let mut result = HashSet::new();
```
tekknolagi: nit: vec sort and dedup might be faster and less alloc-y

tekknolagi (approved):
Can you add some HIR tests for side-exiting or raising in an optional parameter? Also, some zjit.rb tests please (incl `assert_compiles`). Otherwise lgtm. And the other benchmarks still pass?

zjit/src/hir.rs
k0kubun: :+1: https://github.com/ruby/ruby/pull/14653/commits/6c5ac77ca55a8359a1d3a6e3dad362dabf253804

zjit/src/hir.rs
k0kubun: sounds good https://github.com/ruby/ruby/pull/14653/commits/3952cdfcfef27a56f439b584a5c062f04bcfca7c

k0kubun:
> Can you add some HIR tests for side-exiting or raising in an optional parameter? 

side-exiting one was already there (due to `trace_*` being a side exit), but added another one that doesn't rely on TracePoint https://github.com/ruby/ruby/pull/14653/commits/a2c6111ad2ded5e7095417bf2f01a626dafd2ef3.

> Also, some zjit.rb tests please (incl `assert_compiles`)

Now that we have https://github.com/ruby/ruby/pull/14639, I'm thinking we should slowly move away from `test_zjit.rb` and do behavior-level tests in Rust as well, which allows us to automatically fill out results by `zjit-test-update` and probably has better parallelism in the test runner.

But since we don't have such existing test cases yet, I just added test cases in `test_zjit.rb` for now https://github.com/ruby/ruby/pull/14653/commits/f513b9b6ab2cddb4258f9ca38aea483bad537dc5.

> And the other benchmarks still pass?

I didn't test 25+ iterations but `run_benchmarks.rb --once` passed with all benchmarks.

PR #14659, opened 2025-09-26
ZJIT: Incorporate parameter loads into HIR

k0kubun:


zjit/src/codegen.rs
```diff
@@ -537,19 +537,28 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, obj: VALUE,
...
+    };
+    asm.load(local)
 }
 
 /// Set a local variable from a higher scope or the heap. `local_ep_offset` is in number of VALUEs.
 /// We generate this instruction with level=0 only when the local variable is on the heap, so we
 /// can't optimize the level=0 case using the SP register.
-fn gen_setlocal_with_ep(asm: &mut Assembler, val: Opnd, val_type: Type, local_ep_offset: u32, level: u32) {
+fn gen_setlocal(asm: &mut Assembler, val: Opnd, val_type: Type, local_ep_offset: u32, level: u32) {
+    let local_ep_offset = i32::try_from(local_ep_offset).unwrap_or_else(|_| panic!("Could not convert local_ep_offset {local_ep_offset} to i32"));
```
```suggestion
    let local_ep_offset = cint::try_from(local_ep_offset).unwrap_or_else(|_| panic!("Could not convert local_ep_offset {local_ep_offset} to i32"));
```

matching the signature of the function this is passed to

zjit/src/hir.rs
XrXr, line 1: Please mark `LoadSelf` as false in `has_effects` for DCE.

XrXr (commented):
Seems like a nice cleanup. It might temporarily increase what we could compile, because this might lean heavier on having spill support in regalloc, but this will take us to a much better place.

But, doesn't this also makes it so that round tripping through memory is required even for JIT2JIT calls?

zjit/src/codegen.rs
k0kubun: fixed in https://github.com/ruby/ruby/pull/14659/commits/6cf4d5ab620e57c28bedb6252e63f09e4b0b2b40

zjit/src/hir.rs
k0kubun: done https://github.com/ruby/ruby/pull/14659/commits/ee51baa141d560f0ef77af78c7d2c69b1ab002f6

k0kubun:
> this might lean heavier on having spill support in regalloc

Yeah, I'm hoping to land that soon too. For the record, `ratio_in_zjit` on lobsters is still 72.2% on this branch (not decreased), so spills are probably happening in places outside the reach of VRegs used in the interpreter entry.

> doesn't this also makes it so that round tripping through memory is required even for JIT2JIT calls?

I'm not sure where that comes from? This PR changes only the interpreter entry, so ISEQ body blocks and JIT
entry blocks (which JIT2JIT calls into) still use the same registers and stack slots (because of `param_opnd`). So JIT2JIT calls would use the same things for passing arguments.

XrXr:
Ah, I missed the JIT entry blocks not having changed. Not an issue, then.

XrXr (approved):

zjit/src/codegen.rs
```diff
@@ -537,19 +537,28 @@ fn gen_defined(jit: &JITState, asm: &mut Assembler, op_type: usize, obj: VALUE,
...
-fn gen_getlocal_with_ep(asm: &mut Assembler, local_ep_offset: u32, level: u32) -> lir::Opnd {
+fn gen_getlocal(asm: &mut Assembler, local_ep_offset: u32, level: u32, use_sp: bool) -> lir::Opnd {
+    let local_ep_offset = i32::try_from(local_ep_offset).unwrap_or_else(|_| panic!("Could not convert local_ep_offset {local_ep_offset} to i32"));
     if level > 0 {
         gen_incr_counter(asm, Counter::vm_read_from_parent_iseq_local_count);
     }
-    let ep = gen_get_ep(asm, level);
-    let offset = -(SIZEOF_VALUE_I32 * i32::try_from(local_ep_offset).unwrap_or_else(|_| panic!("Could not convert local_ep_offset {local_ep_offset} to i32")));
-    asm.load(Opnd::mem(64, ep, offset))
+    let local = if use_sp {
```
tekknolagi, line 545: Do we use this for any locals other than params? If no, it might be clearer to split this into another instruction. Or maybe we should just add local name printing to all `GetLocal` instructions

tekknolagi: (It feels like having branching in the codegen on a static field of the instruction means it should be split, is all)

zjit/src/hir.rs
```diff
@@ -1601,24 +1611,19 @@ impl Function {
             Insn::ToArray { .. } => types::ArrayExact,
             Insn::ObjToString { .. } => types::BasicObject,
             Insn::AnyToString { .. } => types::String,
+            Insn::GetLocal { rest_param: true, .. } => types::ArrayExact,
```
tekknolagi, line 1614: this `rest_param` feels like something that is duplicate info and could be checked by e.g. `... if local_idx == self.rest_param_idx` or something like that

zjit/src/hir.rs
```diff
@@ -3925,7 +3932,8 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                         // Reload locals that may have been modified by the blockiseq.
                         for local_idx in 0..state.locals.len() {
                             let ep_offset = local_idx_to_ep_offset(iseq, local_idx) as u32;
-                            let val = fun.push_insn(block, Insn::GetLocal { ep_offset, level: 0 });
+                            // TODO: We could use `use_sp: true` with PatchPoint
```
tekknolagi, line 3935: this TODO isn't very actionable or clear. can we expand a little bit please?

zjit/src/hir.rs
k0kubun: good call!

zjit/src/codegen.rs
k0kubun: > It feels like having branching in the codegen on a static field of the instruction means it should be split, is all

sure, makes sense. will fix it.

zjit/src/hir.rs
k0kubun: Filed https://github.com/Shopify/ruby/issues/769. I'll update these comments to link it.

PR #14660, opened 2025-09-26
YJIT: Remove test gate for rb_yjit_record_exit_stack

aidenfoxivey:


aidenfoxivey:
tl;dr for making this:

- _exit_pc is called such since the testing configuration will complain of an unused variable when `cargo test` is run
- I figured that I might have been able to just git rid of it, but using `cargo` to test means that in fact it needs to stay.

PR #14663, opened 2025-09-26
[DOC] Fix rendering of $\ in globals.md

peterzhu2118:


nobu:
This should be fixed by the rdoc master.

PR #14664, opened 2025-09-27
[DOC] Tweaks for String#partition

BurdetteLamar:


doc/string/partition.rdoc
```diff
@@ -1,24 +1,32 @@
 Returns a 3-element array of substrings of +self+.
 
-Matches a pattern against +self+, scanning from the beginning.
-The pattern is:
+Converts +pattern+ to a Regexp, if it is not already one.
```
peterzhu2118: I don't think this is true, it doesn't convert `pattern` to a Regexp. Looking at the code, if `pattern` is not a String or Regexp, it converts it to a String. If it's a String, then it uses `String#index` to get the index at which the match occurs.

doc/string/partition.rdoc
```diff
@@ -1,24 +1,32 @@
...
   'hello'.partition('ll')     # => ["he", "ll", "o"]
   'hello'.partition('h')      # => ["", "h", "ello"]
   'hello'.partition('o')      # => ["hell", "o", ""]
   'hello'.partition(/l+/)     #=> ["he", "ll", "o"]
-  'hello'.partition('')       # => ["", "", "hello"]
+  'hello'.partition('')       # => ["", "", "hello"]  # Note: this post_match is a copy of self.
   'тест'.partition('т')       # => ["", "т", "ест"]
   'こんにちは'.partition('に')  # => ["こん", "に", "ちは"]
 
+Does _not_ set {pattern-matching global variables}[rdoc-ref:globals.md@Pattern+Matching].
```
peterzhu2118: I don't think this is true:

```ruby
'hello'.partition(/l+/)
puts $~ # => "ll"
```

doc/string/partition.rdoc
BurdetteLamar: FIxed.

doc/string/partition.rdoc
BurdetteLamar: Fixed.

PR #14672, opened 2025-09-29
Change test-all suite to enable running each test in separate ractors.

luke-gruber:
This tests for ractor safety issues as well as concurrency issues.

You can enable these tests with RUBY_TESTS_WITH_RACTORS=X when running `make test-all TESTS=test/ruby`. Running tests with ractors is currently only working for tests under the "test/ruby" directory. You should also use the `.excludes-ractor` excludes directory.

For example:

```
EXCLUDES=test/.excludes-ractor RUBY_TESTS_WITH_RACTORS=3 TESTS=test/ruby make test-all
```

This will create 3 ractors for each test method, run the test method inside each ractor and then call `join` on the ractors. Then, it's ready to process the next test. The reason we do this instead of taking all the tests and dividing them between the number of ractors is:

* We want to run each test under concurrency load to test whether or not it is safe under ractors. If it isn't safe, we know it's something to do with this specific test.

* If we get a segfault or error, we know the error is coming from this test alone, not interactions with other tests.

The disadvantage of this approach is:

* It's slower than dividing the tests between the number of ractors running.

* This might not uncover ractor scheduling issues that would show up when you have long-running ractors.


luke-gruber:
This branch is ready to be merged @ko1 . When you run the make command listed in the first commit, don't add the `-j` as this option is not supported with ractors enabled yet.

tool/lib/zombie_hunter.rb
```diff
@@ -3,7 +3,7 @@
 module ZombieHunter
   def after_teardown
     super
-    assert_empty(Process.waitall)
+    assert_empty(Process.waitall) unless multiple_ractors?
```
ko1, line 6: why is it disabled?

tool/lib/test/unit/testcase.rb
```diff
@@ -198,22 +215,24 @@ def run runner
...
+          $0 = progname
+        end
       end
 
       def initialize name # :nodoc:
         @__name__ = name
         @__io__ = nil
         @__passed__ = nil
-        @@__current__ = self # FIX: make thread local
+        Ractor.current[:__test_current__] = self
```
ko1, line 231: The removed comment said it should be thread local. is it okay to use Ractor local?

tool/lib/test/unit/testcase.rb
```diff
@@ -198,22 +215,24 @@ def run runner
 
       RUN_TEST_TRACE = "#{__FILE__}:#{__LINE__+3}:in `run_test'".freeze
       def run_test(name)
-        progname, $0 = $0, "#{$0}: #{self.class}##{name}"
+        progname, $0 = $0, "#{$0}: #{self.class}##{name}" if main_ractor?
```
ko1, line 218: TODO:
if we make main Ractor as a manager Ractor (and other Ractors run tests, a manager Ractor knows which tests are running), we can set the progname with running test names by the main Ractor.

tool/lib/test/unit/testcase.rb
```diff
@@ -50,6 +50,23 @@ def mingw? platform = RUBY_PLATFORM
...
+        return true if !defined?(Ractor)
+        Ractor.current == Ractor.main
+      end
+
+      def non_main_ractor?
+        not main_ractor?
+      end
+
+      def multiple_ractors?
+        ENV["RUBY_TESTS_WITH_RACTORS"].to_i > 1
```
ko1: cache the value or define different methods with the envval?
some tests can remove ENVVAL.

test/test_tmpdir.rb
```diff
@@ -154,10 +154,10 @@ def test_ractor
             Ractor.receive
           end
         end
-        dir = r.take
```
ko1, line 168: why does it work...??

tool/lib/zombie_hunter.rb
luke-gruber: Since `after_teardown` runs after each test, other ractors could still have processes running.

tool/lib/test/unit/testcase.rb
luke-gruber: Yes it's okay 👍

tool/lib/test/unit/testcase.rb
luke-gruber: Yeah, that's a good idea. For future work 😆

test/test_tmpdir.rb
luke-gruber: I'll investigate 🤷 Edit: There was a shim in `assert_ractor` method. I removed it.

luke-gruber:
MMTK CI failed but we found out why. We're waiting on https://github.com/ruby/ruby/pull/14811 to be merged.

luke-gruber:
Ready again for review 🙇‍♂️

ext/-test-/iter/init.c
```diff
@@ -5,6 +5,11 @@
 void
 Init_iter(void)
 {
+
```
aidenfoxivey: nit: Empty line here could be removed to match the others.

ext/-test-/iter/yield.c
```diff
@@ -10,6 +10,11 @@ yield_block(int argc, VALUE *argv, VALUE self)
 void
 Init_yield(VALUE klass)
 {
+
```
aidenfoxivey: nit: Same as earlier - Empty line here could be removed to match the others.

ext/-test-/rb_call_super_kw/rb_call_super_kw.c
```diff
@@ -9,6 +9,10 @@ rb_call_super_kw_m(int argc, VALUE *argv, VALUE self)
 void
 Init_rb_call_super_kw(void)
 {
+#ifdef HAVE_RB_EXT_RACTOR_SAFE
+    // Mark this extension as Ractor-safe.
```
aidenfoxivey: nit: For consistency, maybe either have this comment in all `HAVE_RB_EXT_RACTOR_SAFE` sites, or none of them?

test/.excludes-ractor/TestGCCompact.rb
```diff
@@ -0,0 +1 @@
+exclude(/^test_/, "GC.auto_compact = true then setting it back is racy")
```
aidenfoxivey, line 1: question: does "racy" here mean "not ractor safe"?

aidenfoxivey (commented):
Misc. nits + a question.

hsbt:
Can you separate tests of `fileutils` and `json`? and submit them to https://github.com/ruby/fileutils and https://github.com/ruby/json?

It seems to work independently from this PR.

test/.excludes-ractor/TestGCCompact.rb
luke-gruber: It just means lots of ractors are setting `GC.auto_compact = true` at the beginning of the method, doing stuff, then setting it back to the old value (false) at the end of the method. If lots of ractors are running the same test at once, other ractors can be setting that value out from under you, and you might not be testing what you think you are.

PR #14675, opened 2025-09-30
CI: Fail the dump crash log step for visual reminder

XrXr:
I forgot that this step existed and thought crash reporting wasn't
working when they were simply moved to a different step. Failing these
should give a nice visual hint.


XrXr:
Should we put the YML files to auto request reviews?

k0kubun:
> Should we put the YML files to auto request reviews?

See: https://github.com/ruby/ruby/blob/8ce886b2d8bef0d0bc0edf8c1cd2c1348c1cbfef/.github/auto_request_review.yml#L14

k0kubun (approved):

PR #14676, opened 2025-09-30
ZJIT: Unify fallback counters for send-ish insns

k0kubun:


tekknolagi:
This is really nice

tekknolagi (approved):

tekknolagi:
Clear signal we need to add profiles for `opt_aref` and `send`

PR #14677, opened 2025-09-30
Update bundled_gems

hsbt:
`bigdecimal-3.2.3` is failing with master branch now. I manually update other gems.


hsbt:
I filed https://github.com/ruby/bigdecimal/issues/435

PR #14673, opened 2025-09-29
ZJIT: Revert SetIvar specialization

tekknolagi:
- **Revert "ZJIT: Fix rebase issue with tests"**
- **Revert "ZJIT: Inline attr_accessor/attr_writer to SetIvar (#14629)"**

PR #14678, opened 2025-09-30
ZJIT: Re-apply attr_writer inlining

tekknolagi:
This re-applies https://github.com/ruby/ruby/pull/14629 / 40bb47665d3ff57e0f2eb5a9fd9e0109617015c9 by reverting https://github.com/ruby/ruby/pull/14673 /  d4393772b89dab4f33c118a284d92dc80cd63c39.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -1986,7 +1986,7 @@ impl Function {
                             }
                             let getivar = self.push_insn(block, Insn::GetIvar { self_val: recv, id, state });
                             self.make_equal_to(insn_id, getivar);
-                        } else if let (VM_METHOD_TYPE_ATTRSET, [setter_rhs]) = (def_type, args.as_slice()) {
+                        } else if let (VM_METHOD_TYPE_ATTRSET, [val]) = (def_type, args.as_slice()) {
                             self.push_insn(block, Insn::PatchPoint { invariant: Invariant::MethodRedefined { klass, method: mid, cme }, state });
                             if let Some(profiled_type) = profiled_type {
                                 recv = self.push_insn(block, Insn::GuardType { val: recv, guard_type: Type::from_profiled_type(profiled_type), state });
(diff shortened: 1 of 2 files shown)


tekknolagi:
The change is we have to rewrite the method send into the rhs of the assignment, not the new SetIvar instruction

zjit/src/hir.rs
```diff
@@ -1986,6 +1986,24 @@ impl Function {
...
+                            if let Some(profiled_type) = profiled_type {
+                                recv = self.push_insn(block, Insn::GuardType { val: recv, guard_type: Type::from_profiled_type(profiled_type), state });
+                            }
+                            let id = unsafe { get_cme_def_body_attr_id(cme) };
+
+                            // Check if we're accessing ivars of a Class or Module object as they require single-ractor mode.
+                            // We omit gen_prepare_non_leaf_call on gen_setivar, so it's unsafe to raise for multi-ractor mode.
+                            if unsafe { rb_zjit_singleton_class_p(klass) } {
+                                let attached = unsafe { rb_class_attached_object(klass) };
+                                if unsafe { RB_TYPE_P(attached, RUBY_T_CLASS) || RB_TYPE_P(attached, RUBY_T_MODULE) } {
```
XrXr, line 2000: Note to self: this convoluted check could be replaced with `klass.is_subclass_of(rb_cModule)` in a separate PR

zjit/src/hir.rs
```diff
@@ -1986,6 +1986,24 @@ impl Function {
                             }
                             let getivar = self.push_insn(block, Insn::GetIvar { self_val: recv, id, state });
                             self.make_equal_to(insn_id, getivar);
+                        } else if def_type == VM_METHOD_TYPE_ATTRSET && args.len() == 1 {
```
XrXr:
```suggestion
                        } else if let (VM_METHOD_TYPE_ATTRSET, [setter_rhs]) = (def_type, args.as_slice()) {
```
nit: gets rid of the distance between `len==1` check and `vec[0]` indexing

XrXr (approved):

zjit/src/hir.rs
st0012: Ohh this is neat

PR #14679, opened 2025-09-30
ZJIT: Inline well-known C functions into HIR

tekknolagi:
Add the ability to create a Rust function that "open-codes" HIR
implementations of specific well-known C functions, starting with
`String#to_s`. It supports emitting multiple instructions into a
temporary block, but does not support emitting *new blocks* (yet?).

Fix https://github.com/Shopify/ruby/issues/792

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2431,9 +2431,6 @@ impl Function {
                     // func(int argc, VALUE *argv, VALUE recv)
                     let ci_flags = unsafe { vm_ci_flag(call_info) };
                     if ci_flags & VM_CALL_ARGS_SIMPLE != 0 {
-                        if get_option!(stats) {
-                            count_not_inlined_cfunc(fun, block, method);
-                        }
                         gen_patch_points_for_optimized_ccall(fun, block, recv_class, method_id, method, state);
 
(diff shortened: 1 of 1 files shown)


zjit/src/cruby_methods.rs
```diff
@@ -188,3 +196,23 @@ pub fn init() -> Annotations {
...
+    if args.len() == 0 && fun.likely_a(recv, types::StringExact, state) {
+        // No need to coerce the receiver; that is done by the SendWithoutBlock rewriting.
+        return Some(recv);
+    }
+    None
+}
+
+fn inline_kernel_itself(_fun: &mut hir::Function, _block: hir::BlockId, recv: hir::InsnId, args: &[hir::InsnId], _state: hir::InsnId) -> Option<hir::InsnId> {
+    if args.len() == 0 {
+        // No need to coerce the receiver; that is done by the SendWithoutBlock rewriting.
```
XrXr, line 232:
```suggestion
```

`itself` doesn't do any coercion in any caes

tekknolagi: In this case it does; it's a guard that the lhs is the expected type so that we can continue to inline the same method on it

PR #14661, opened 2025-09-26
ZJIT: Add HIR for calling Cfunc with frame

st0012:
Major changes to `activerecord`:

  | Metric                 | Before            | After             | Change                     |
  |------------------------|-------------------|-------------------|----------------------------|
  | Dynamic Send Count     | 48,528,405        | 41,309,323        | -14.9% reduction           |
  | Guard Type Failures    | 5,185,117 (18.0%) | 6,246,106 (20.9%) | +20.5% increase            |
  | VM Write PC Count      | 80,727,948        | 78,078,612        | -3.3% reduction            |
  | ZJIT Instruction Ratio | 60.0%             | 59.0%             | -1.0% decrease             |
  | Side Exit Count        | 28,826,722        | 29,888,003        | +3.7% increase             |
  | Code Region Bytes      | 5,865,472         | 6,029,312         | +2.8% increase             |

This PR adds `CCallWithFrame` HIR to optimize simple cfuncs that are NOT annotated as `leaf` and `no_gc`. So it covers most of the calls currently fall under `unoptimized sends to C functions`. Therefore, this PR also renames that stat to `full frame C function calls` as they can still be optimized with other approaches.

**Before**

```
RSS: 94.0MiB
MAXRSS: 96.7MiB
Writing file /Users/hung-wulo/src/github.com/Shopify/ruby-bench/data/results-ruby-3.5.0-2025-09-29-211504.json
Average of last 10, non-warmup iters: 403ms
***ZJIT: Printing ZJIT statistics on exit***
Top-5 dynamic send types (100.0% of total 48,528,405):
  send_without_block: 32,175,141 (66.3%)
         invokesuper:  7,631,172 (15.7%)
                send:  7,476,408 (15.4%)
         invokeblock:  1,168,574 ( 2.4%)
        send_forward:     77,110 ( 0.2%)
Top-7 send fallback unspecialized def_types (100.0% of total 13,106,939):
      cfunc: 5,553,554 (42.4%)
       iseq: 4,581,123 (35.0%)
    bmethod: 2,875,890 (21.9%)
  optimized:    55,083 ( 0.4%)
      alias:    33,130 ( 0.3%)
    attrset:     8,146 ( 0.1%)
       null:        13 ( 0.0%)
Top-2 dynamic send types (100.0% of total 5,434,194):
  polymorphic: 3,115,659 (57.3%)
  no_profiles: 2,318,535 (42.7%)
Top-20 unoptimized sends to C functions (97.0% of total 5,631,422):
[...]

PR #14680, opened 2025-09-30
ZJIT: Add `NoSingletonClass` patch point

st0012:
This patch point makes sure that when the object has a singleton class, the JIT code is invalidated. As of now, this is only needed for C call optimization.

In YJIT, the singleton class guard only applies to Array, Hash, and String. But in ZJIT, we may optimize C calls from gems (e.g. #14661 will apply to `sqlite3` methods too). So the patch point needs to be applied to a broader range of classes.

Fix https://github.com/Shopify/ruby/issues/768

### DIFF
--- a/zjit/src/invariants.rs
+++ b/zjit/src/invariants.rs
@@ -71,6 +71,7 @@ impl Invariants {
         self.update_ep_escape_iseqs();
         self.update_no_ep_escape_iseq_patch_points();
         self.update_cme_patch_points();
+        self.update_no_singleton_class_patch_points();
     }
 
     /// Forget an ISEQ when freeing it. We need to because a) if the address is reused, we'd be
@@ -87,6 +88,11 @@ impl Invariants {
     /// Forget a CME when freeing it. See [Self::forget_iseq] for reasoning.
     pub fn forget_cme(&mut self, cme: *const rb_callable_method_entry_t) {
         self.cme_patch_points.remove(&cme);
+    }
+
+    /// Forget a class when freeing it. See [Self::forget_iseq] for reasoning.
+    pub fn forget_klass(&mut self, klass: VALUE) {
+        self.no_singleton_class_patch_points.remove(&klass);
     }
 
     /// Update ISEQ references in Invariants::ep_escape_iseqs
@@ -119,6 +125,17 @@ impl Invariants {
             })
             .collect();
         self.cme_patch_points = updated_cme_patch_points;
+    }
+
+    fn update_no_singleton_class_patch_points(&mut self) {
+        let updated_no_singleton_class_patch_points = std::mem::take(&mut self.no_singleton_class_patch_points)
+            .into_iter()
+            .map(|(klass, patch_points)| {
+                let new_klass = unsafe { rb_gc_location(klass) };
+                (new_klass, patch_points)
+            })
+            .collect();
+        self.no_singleton_class_patch_points = updated_no_singleton_class_patch_points;
     }
 }
 
--- a/gc.c
+++ b/gc.c
@@ -1311,6 +1311,9 @@ rb_gc_obj_free(void *objspace, VALUE obj)
         break;
       case T_MODULE:
       case T_CLASS:
+#if USE_ZJIT
+        rb_zjit_klass_free(obj);
+#endif
         args.klass = obj;
         rb_class_classext_foreach(obj, classext_free, (void *)&args);
         if (RCLASS_CLASSEXT_TBL(obj)) {
--- a/zjit.h
+++ b/zjit.h
@@ -19,6 +19,7 @@ void rb_zjit_profile_enable(const rb_iseq_t *iseq);
 void rb_zjit_bop_redefined(int redefined_flag, enum ruby_basic_operators bop);
 void rb_zjit_cme_invalidate(const rb_callable_method_entry_t *cme);
 void rb_zjit_cme_free(const rb_callable_method_entry_t *cme);
+void rb_zjit_klass_free(VALUE klass);
 void rb_zjit_invalidate_no_ep_escape(const rb_iseq_t *iseq);
 void rb_zjit_constant_state_changed(ID id);
 void rb_zjit_iseq_mark(void *payload);
--- a/zjit/src/gc.rs
+++ b/zjit/src/gc.rs
@@ -148,6 +148,16 @@ pub extern "C" fn rb_zjit_cme_free(cme: *const rb_callable_method_entry_struct) 
     invariants.forget_cme(cme);
 }
 
+/// GC callback for finalizing a class
+#[unsafe(no_mangle)]
+pub extern "C" fn rb_zjit_klass_free(klass: VALUE) {
+    if !ZJITState::has_instance() {
+        return;
+    }
+    let invariants = ZJITState::get_invariants();
+    invariants.forget_klass(klass);
+}
+
 /// GC callback for updating object references after all object moves
 #[unsafe(no_mangle)]
 pub extern "C" fn rb_zjit_root_update_references() {


zjit/src/hir.rs
```diff
@@ -140,6 +140,12 @@ pub enum Invariant {
     NoEPEscape(IseqPtr),
     /// There is one ractor running. If a non-root ractor gets spawned, this is invalidated.
     SingleRactorMode,
+    /// Objects of this class have no singleton class.
+    /// When a singleton class is created for an object of this class, this is invalidated.
+    /// Applies to all classes where CCall optimization is used.
```
tekknolagi: This is a bit misleading: we need this for *any* method call specialization

zjit/src/hir.rs
```diff
@@ -2149,6 +2161,12 @@ impl Function {
         self.infer_types();
     }
 
+    fn gen_patch_points_for_optimized_ccall(&mut self, block: BlockId, recv_class: VALUE, method_id: ID, method: *const rb_callable_method_entry_struct, state: InsnId) {
```
tekknolagi: I don't know that this should necessarily go with the other method invariants. It's a little more insidious than that because we could, be relying on type information that gets invalidated across an intervening call. For example:

```ruby
a.foo()
bar()
a.is_a? C
```

In the first call to `a.foo()` we guard that `a` has class `C`, optimize the method lookup, etc. Then that gets invalidated in the call to `bar()`. If we then fold `a.is_a? C` based on type information from the first guard, we might be folding wrong: it's not just specializing method sends.

This means that we need to patchpoint pretty much after every operation that might run arbitrary Ruby code, similar to EP escape.

tekknolagi: (& for this reason let's hide it with the other really noisy ones)

st0012: I'm not sure if I understand why a singleton class invalidation in `bar()` would affect how `a.is_a? C` behaves. Having a singleton class or not, it should both be `true`?

tekknolagi: Oh, sure, ok, sorry -- pretend the `is_a?` call is a `GuardType(a, C)`. This is an exact type check --- literally checking the `CLASS_OF`

tekknolagi: Also we should skip on classes like Integer for which it's not even allowed to make singleton classes

tekknolagi: And I think we want to start tracking the class every time we insert a `GuardType`? Maybe?

zjit/src/hir.rs
```diff
@@ -10287,7 +10306,8 @@ mod opt_tests {
           PatchPoint MethodRedefined(C@0x1008, new@0x1010, cme:0x1018)
           v43:HeapObject[class_exact:C] = ObjectAllocClass C:VALUE(0x1008)
           PatchPoint MethodRedefined(C@0x1008, initialize@0x1040, cme:0x1048)
-          v45:NilClass = CCall initialize@0x1070, v43
+          PatchPoint NoSingletonClass(C@0x1008)
+          v47:NilClass = CCall initialize@0x1070, v43
```
XrXr, line 10403: Mostly echoing Max's comment at https://github.com/ruby/ruby/pull/14680#discussion_r2391676106

The current logic for inserting the patchpoints seems too relaxed in some cases and not aggressive enough in some cases. This particular patchpoint shouldn't be necessary because there is no call to arbitrary code that could create a singleton class.

st0012: > there is call to arbitrary code that could create a singleton class.

Do you mean "there's no arbitrary code that could create a singleton class"?

XrXr: Ah yes, sorry. There is _no_ arbitrary code between `ObjectAllocClass` and the eliminated `GuardType`.

zjit/src/hir.rs
st0012: I added this PP to more type guards. Not all of them because some are for specific types like `Fixnum`, which can't have singleton class.

zjit/src/hir.rs
st0012: I have 2 questions wrt this:
- Is there already a good way to look up previous insns and say "ok now this patch point/type guard won't be needed"? I can look up insns for this specific insn but it feels brittle.
- Do we need to consider code running on another thread changing the state? In this particular case, it seems possible (tho unlikely) that `C` can have a singleton class created to define `new` right before the `CCall` in another thread?

XrXr: > Do we need to consider code running on another thread changing the state? 

Thanks to ractor's design and the GVLs, this is not possible -- yielding only happens when checking interrupts (directly in the IR or indirectly through call to a runtime function). 

So, getting rid of these PatchPoints should probably be done in an optimization pass since it at least needs to see all the inserted-in-post `CheckInterrupt`s. As a first approximation, for `GuardType v`, if all instructions between the creation of `v` and the `GuardType` are `has_effects() == false`, then the `GuardType` could be eliminated without also putting in a PP.

XrXr: This sounds complicated so maybe don't worry about excessive PP for now. Let's just get it correct first and we can make a ticket to revisit this later.

st0012: Thanks for explaining 👍

st0012: Issue: https://github.com/Shopify/ruby/issues/779

zjit/src/hir_type/mod.rs
```diff
@@ -306,6 +306,10 @@ impl Type {
         self.is_subtype(types::NilClass) || self.is_subtype(types::FalseClass)
     }
 
+    pub fn can_have_singleton_class(&self) -> bool {
+        !self.is_immediate() && !self.is_subtype(types::Numeric) && !self.is_subtype(types::Symbol)
```
st0012: My understanding is that in addition to immediate types, `Float`, `Rational`, and `Symbol` (dynamic too) can't have singleton classes.

tekknolagi:
```
irb(main):001> val = nil
=> nil
irb(main):002* class << val
irb(main):003*   def foo = 5
irb(main):004> end
=> :foo
irb(main):005> val.foo
=> 5
irb(main):006>
```

o.O

tekknolagi: ugh it just adds it to `NilClass`

zjit/src/invariants.rs
```diff
@@ -357,3 +376,21 @@ pub extern "C" fn rb_zjit_tracing_invalidate_all() {
...
+
+    with_vm_lock(src_loc!(), || {
+        let invariants = ZJITState::get_invariants();
+        if let Some(patch_points) = invariants.no_singleton_class_patch_points.remove(&klass) {
+            let cb = ZJITState::get_code_block();
+            debug!("Singleton class created for {:?}", klass);
+            compile_patch_points!(cb, patch_points, "Singleton class created for {:?}", klass);
+            cb.mark_all_executable();
+        }
+
```
tekknolagi: delete empty line

zjit/src/hir.rs
```diff
@@ -2021,13 +2032,19 @@ impl Function {
                                 self.push_insn_id(block, insn_id); continue;
                             }
                             self.push_insn(block, Insn::PatchPoint { invariant: Invariant::MethodRedefined { klass, method: mid, cme }, state });
+                            if self.type_of(recv).can_have_singleton_class() {
```
tekknolagi: we should be looking at `klass` instead: this is the class object we are intending to specialize on, including the profiled class. we should not look at the HIR-known type of the object (which might be `BasicObject` a lot)

zjit/src/hir.rs
```diff
@@ -9635,10 +9664,11 @@ mod opt_tests {
           Jump bb2(v5, v6)
         bb2(v8:BasicObject, v9:BasicObject):
           PatchPoint MethodRedefined(Integer@0x1000, itself@0x1008, cme:0x1010)
-          v21:Fixnum = GuardType v9, Fixnum
-          v22:BasicObject = CCall itself@0x1038, v21
+          PatchPoint NoSingletonClass(Integer@0x1000)
```
tekknolagi: we should be avoiding `Integer` singleton class patchpoints but still have them for some reason. this is probably related to my suggested changes above

tekknolagi (changes requested):
back on your queue -- looking better. the approach feels ad-hoc still. i wonder if we can have a more general solution here

zjit/src/hir.rs
```diff
@@ -8592,13 +8618,15 @@ mod opt_tests {
           Jump bb2(v4)
         bb2(v6:BasicObject):
           PatchPoint MethodRedefined(Object@0x1000, foo@0x1008, cme:0x1010)
-          v22:HeapObject[class_exact*:Object@VALUE(0x1000)] = GuardType v6, HeapObject[class_exact*:Object@VALUE(0x1000)]
-          v23:BasicObject = SendWithoutBlockDirect v22, :foo (0x1038)
+          PatchPoint NoSingletonClass(Object@0x1000)
```
tekknolagi, line 8621: This one is kind of weird because

a) we know that Object has instances with singleton classes, notably top-level self,
b) which is the case here

What should the expected behavior be if a singleton class has already been created? On the one hand, we want to specialize sends to top-level self (and other singleton class instances), but on the other hand, we're kind of asserting that no new singleton classes are created. I don't know that I have a specific request here other than to ask for your thoughts

zjit/src/invariants.rs
```diff
@@ -59,6 +59,10 @@ pub struct Invariants {
 
     /// Set of patch points that assume that the interpreter is running with only one ractor
     single_ractor_patch_points: HashSet<PatchPoint>,
+
+    /// Map from a class to a set of patch points that assume objects of the class
+    /// will have no singleton class.
```
tekknolagi, line 64: this needs `update_references` support

tekknolagi: and maybe support for something like `forget_class`

cc @XrXr ?

tekknolagi (approved):
This looks like a good start. I hope we can come up with some kind of approach that unifies this, NoTracePoints, and NoEPEscape

zjit/src/invariants.rs
XrXr: Yes we want both.

st0012: Where should I invoke `forget_class`? Is it `rb_class_remove_from_super_subclasses` or `rb_class_remove_from_module_subclasses`?

XrXr: Should be in `rb_gc_obj_free()`:

```
      case T_MODULE:
      case T_CLASS:
        args.klass = obj;
        rb_class_classext_foreach(obj, classext_free, (void *)&args);
        if (RCLASS_CLASSEXT_TBL(obj)) {
            st_free_table(RCLASS_CLASSEXT_TBL(obj));
        }
        (void)RB_DEBUG_COUNTER_INC_IF(obj_module_ptr, BUILTIN_TYPE(obj) == T_MODULE);
        (void)RB_DEBUG_COUNTER_INC_IF(obj_class_ptr, BUILTIN_TYPE(obj) == T_CLASS);
        break;
```

We are doing lifecycle tracking for arbitrary class objects now, and we forget on finalization.

k0kubun (approved):

zjit/src/hir.rs
st0012: I added an issue for further discussions on this: https://github.com/Shopify/ruby/issues/783

PR #14681, opened 2025-09-30
ZJIT: Fix "malformed format string" on stats

k0kubun:
Follow-up on https://github.com/ruby/ruby/pull/14638

The modified line fails when `padded_key` is `String#%` for not-optimized C func stats, which fails several headline benchmarks with `--zjit-stats`. This PR changes it to not include `padded_key` as part of the format string.

### DIFF
--- a/zjit.rb
+++ b/zjit.rb
@@ -221,8 +221,8 @@ class << RubyVM::ZJIT
     return if stats.empty?
 
     counters.transform_keys! { |key| key.to_s.delete_prefix(prefix) }
-    left_pad = counters.keys.map(&:size).max
-    right_pad = counters.values.map { |value| number_with_delimiter(value).size }.max
+    key_pad = counters.keys.map(&:size).max
+    value_pad = counters.values.map { |value| number_with_delimiter(value).size }.max
     total = counters.values.sum
 
     counters = counters.to_a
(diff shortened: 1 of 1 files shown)


zjit.rb
```diff
@@ -236,7 +236,7 @@ def print_counters_with_prefix(buf:, stats:, prefix:, prompt:, limit: nil)
     counters.each do |key, value|
       padded_key = key.rjust(left_pad, ' ')
       padded_value = number_with_delimiter(value).rjust(right_pad, ' ')
-      buf << "  #{padded_key}: #{padded_value} (%4.1f%%)\n" % (100.0 * value / total)
+      buf << "  #{padded_key}: #{padded_value} #{'(%4.1f%%)' % (100.0 * value / total)}\n"
```
XrXr: Instead of going deeper in the escape hole, how about popping up one level and handling it with the formatting mini language?
```suggestion
      buf << "  %*s: %*s (%4.1f%%)\n" % [left_pad, key, right_pad, number_with_delimiter(value), (100.0 * value / total)]"
```

k0kubun: TIL `%*`, thanks https://github.com/ruby/ruby/pull/14681/commits/9031036668cbe4cd552b3a4455129137128b820f

PR #14682, opened 2025-09-30
Tweaks for String#replace

BurdetteLamar:
This is NOT a doc-only PR;  it reverses the alias relationship between #replace and #initialize_copy.


BurdetteLamar:
> In `doc/string.rb`, I wouldn't mention `initialize_copy`, it's a private method.
> 
> Otherwise, this looks good.

Revising earlier reply:  I was looking at string.c, not doc/string.rb.  Fixed.

PR #14687, opened 2025-09-30
ZJIT: Rust code refactors

st0012:


zjit/src/hir_type/mod.rs
```diff
@@ -355,8 +355,7 @@ impl Type {
     fn is_builtin(class: VALUE) -> bool {
         types::ExactBitsAndClass
             .iter()
-            .find(|&(_, class_object)| unsafe { **class_object } == class)
-            .is_some()
+            .any(|&(_, class_object)| unsafe { *class_object } == class)
```
st0012, line 358: The `**class_object` difference is caused by [`any`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any) and [`find`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.find) yield different types.

PR #14692, opened 2025-10-01
ZJIT: Use Marshal.dump to handle large writes

aidenfoxivey:
- **ZJIT: Decouple stats and side exit tracing**
- **ZJIT: Use Mashal.dump**


XrXr:
Merged at https://github.com/ruby/ruby/commit/df90a645d41d831431bff97ca53c8b61585c40c8 with a message I wrote. Would appreciate a bit more context in the future -- for the commit message historians out there :p

PR #14695, opened 2025-10-01
Make test-bundled-gems.rb detect test failure propery even when child process timed out on Windows

YO4:
**I posted new PR #14718 and I think that PR supersedes this.**

test-bundled-gems.rb terminates child process with ```Process.kill("KILL", pid)``` if that timed out on Windows.
Unfortunately, child process exits with 0 in that case, test-bundled-gems.rb misses to detect failure.

This PR changes process termination method to ```system("taskkill /F /PID #{pid}")``` and child process is turned into exiting with 1.

Changing the behavior of Process::kill is also a possible option, but it is not addressed in this PR.

### DIFF
--- a/tool/test-bundled-gems.rb
+++ b/tool/test-bundled-gems.rb
@@ -1,5 +1,4 @@
 require 'rbconfig'
-require 'timeout'
 require 'fileutils'
 require_relative 'lib/colorize'
 require_relative 'lib/gem_env'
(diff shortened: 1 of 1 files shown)


YO4:
Additional context refering this PR
[Bug #21621](https://bugs.ruby-lang.org/issues/21621)

hsbt:
#14718 landed on 2025-10-06 and supersedes this, as you noted. Closing.

PR #14696, opened 2025-10-01
ZJIT: Enable sample rate for side exit tracing

aidenfoxivey:


aidenfoxivey:
Perhaps hold off on reviewing at the moment - I cannot find the option to turn this into a draft but I'm rethinking the naming here.

aidenfoxivey:
Hmm seems to be fine on macOS - as per https://github.com/ruby/ruby/actions/runs/18171590523/job/51727416179?pr=14696#step:15:268 - whereas https://github.com/ruby/ruby/actions/runs/18171590500/job/51727366397?pr=14696#step:16:760 is not cooperating.

aidenfoxivey:
I'm also tempted to remove the primality test regarding sampling rate. The goal from what I understand is more to have the sampling rate coprime with each implied periodicity of side exits calls within the program and that each sampling rate doesn't divide the periodicity. As a result, we should probably prevent using 2, 3, and possibly 5 from being sampling rates, as the probability that our prime divides a periodicity goes down the larger the prime is.

That said, I'm unconvinced that this is essential, since we don't know the periodicity of side exits for certain. We also perhaps don't need perfect accuracy here - only to get the "gist" of it all.

If we want to improve the quality of profiling, we should use Poisson sampling.

k0kubun:
> Hmm seems to be fine on macOS - as per https://github.com/ruby/ruby/actions/runs/18171590523/job/51727416179?pr=14696#step:15:268 - whereas https://github.com/ruby/ruby/actions/runs/18171590500/job/51727366397?pr=14696#step:16:760 is not cooperating.

It seems like one of https://github.com/Shopify/ruby/issues/756 issues, not a fault of this PR.

zjit/src/options.rs
st0012: Since the `sample_interval`'s comment suggests:

> `sample_rate` must provide a string that can be validly parsed to a `usize`.

Should we mention this requirement here?

k0kubun (approved):
seems fine to me as is, but feel free to remove the primality test if you want to.

st0012 (approved):
My comment is not blocking 👍

zjit.rb
```diff
@@ -129,6 +129,8 @@ def dump_exit_locations(filename)
     File.open(filename, "wb") do |file|
       Marshal.dump(RubyVM::ZJIT.exit_locations, file)
     end
+
+    File.size(filename)
```
XrXr:
```suggestion
    File.open(filename, "wb") do |file|
      Marshal.dump(RubyVM::ZJIT.exit_locations, file)
      file.size
    end
```

XrXr (approved):

zjit/src/options.rs
aidenfoxivey: Good point - I will get on that.

aidenfoxivey: Done!

zjit.rb
```diff
@@ -275,9 +276,9 @@ def print_stats
   def dump_locations # :nodoc:
     return unless trace_exit_locations_enabled?
 
-    filename = "zjit_exit_locations.dump"
-    dump_exit_locations(filename)
+    filename = "zjit_exits_#{Time.now.to_i}.dump"
```
tekknolagi, line 279: Nit: better generally to use a PID for things like this because they are (more) easily mappable to a given process. I know the PID of a child process, etc

aidenfoxivey: Ah good catch - lemme quickly pr that.

PR #14698, opened 2025-10-01
ZJIT: Allow higher profile num

st0012:
When we investigate guard failure issues, we sometimes need to use profile num around 100k (e.g. `lobsters` in ruby-bench).

This change is to allow that.


zjit/src/options.rs
```diff
@@ -114,7 +114,7 @@ pub const ZJIT_OPTIONS: &[(&str, &str)] = &[
     ("--zjit-call-threshold=num",
                      "Number of calls to trigger JIT (default: 2)."),
     ("--zjit-num-profiles=num",
-                     "Number of profiled calls before JIT (default: 1, max: 255)."),
+                     "Number of profiled calls before JIT (default: 1, max: 2^32-1)."),
```
st0012: The actual number is `4,294,967,295`, but I feel it's not particularly helpful to state it

k0kubun: Can we just drop `, max: 2^32-1` from the description? Unlike 2^16, you effectively never hit it.

zjit/src/options.rs
```diff
@@ -114,7 +114,7 @@ pub const ZJIT_OPTIONS: &[(&str, &str)] = &[
     ("--zjit-call-threshold=num",
                      "Number of calls to trigger JIT (default: 2)."),
```
k0kubun: Not relevant to the change, but now that you're touching help, I'd appreciate it if you can fix it as well :pray:

```suggestion
                     "Number of calls to trigger JIT (default: 30)."),
```

zjit/src/profile.rs
```diff
@@ -87,7 +87,7 @@ fn profile_insn(bare_opcode: ruby_vminsn_type, ec: EcPtr) {
 
     // Once we profile the instruction num_profiles times, we stop profiling it.
     profile.num_profiles[profiler.insn_idx] = profile.num_profiles[profiler.insn_idx].saturating_add(1);
-    if profile.num_profiles[profiler.insn_idx] == get_option!(num_profiles) {
+    if profile.num_profiles[profiler.insn_idx] as u32 == get_option!(num_profiles) {
```
k0kubun: With your change, I think the profile counter just keeps overflowing.

The whole point of `Options::num_profiles` being `u8` was to make `IseqProfiles::num_profiles` a `Vec<u8>` (to minimize memory consumption). If you change one, you should change the other too, i.e. change `IseqProfiles::num_profiles` from `Vec<u8>` to `Vec<u32>`.

st0012: Ah right. Good catch

zjit/src/options.rs
```diff
@@ -6,7 +6,7 @@ use crate::cruby::*;
 use std::collections::HashSet;
 
 /// Default --zjit-num-profiles
-const DEFAULT_NUM_PROFILES: u8 = 5;
+const DEFAULT_NUM_PROFILES: u32 = 5;
```
tekknolagi, line 9: For situations like this it may be worth having a type alias so all the different uses agree on the type width and parsing and ...

`type NumProfiles = u32` or whatever

st0012: https://github.com/ruby/ruby/pull/14777

PR #14699, opened 2025-10-01
Sync to latest prism

kddnewton:


prism/prism.c
```diff
@@ -5325,7 +5324,7 @@ pm_interpolated_string_node_append(pm_interpolated_string_node_t *node, pm_node_
             } else {
                 // In all other cases, we lose the static literal flag here and
                 // become mutable.
-                CLEAR_FLAGS(node);
+                pm_node_flag_unset((pm_node_t *) node, PM_NODE_FLAG_STATIC_LITERAL);
```
Earlopain, line 5327: Seeing this here now, I did not intend to commit this change.

It causes 3 snapshot changes which I think were correct (or at least not wrong) before. I must have mixed things up because the comment in the change above is also gone.

Could you fix this up? If not, I can make a pr to prism tomorrow

Earlopain: https://github.com/ruby/prism/pull/3668

kddnewton: Thanks!

PR #14700, opened 2025-10-01
Yield to scheduler if interrupts are pending.

ioquatix:
When using `Thread.handle_interrupt(::SignalException => :never)` it is possible that the event loop will hang indefinitely if `rb_thread_call_without_gvl` is used in a loop.

This is one idea to solve that problem.


luke-gruber:
Hey Samuel 👋 . Do you have a minimal repro of the issue? I don't quite understand it.

samuel-williams-shopify:
@luke-gruber This PR now has a test case that fails without the fix and is similar to the gRPC hang (without all the gRPC stuff).

PR #14707, opened 2025-10-02
[Bug #21620] Fix strict aliasing in rb_managed_id_table_lookup

peterzhu2118:
We cannot pass &ccs into rb_managed_id_table_lookup because rb_managed_id_table_lookup takes in a VALUE*, so it violates strict aliasing in C. This causes segfaults when compiling with LTO enabled.


jackorp:
Latest commit on master (5b2ec0eb1be663ff6d3bc12660d48b1e25375353) with the PR applied compiles.

I'd say this PR fixes the issue. Thanks!

PR #14708, opened 2025-10-03
ZJIT: Make sure zjit-test-update works in asm tests

k0kubun:
This PR makes the following improvements on assembly tests:

* Make sure `assert_snapshot!` has nothing other than spaces before that macro on the line
    * When a snapshot spans across multiple lines, `zjit-test-update` copies code before that macro (e.g. `cb.with_disasm(|disasm|`) to those lines in the snapshot.
    * To avoid that, this PR uses `#[cfg(feature = "disasm")]` (in the previous line, not in the same line, to avoid such copies) instead of `with_disasm()`.
* Combine multiple `assert_snapshot!`s by introducing `disasms!` and `hexdumps!`
    * Like `hir_strings!`, it minimizes the number of `zjit-test` → `zjit-test-update` loops for snapshot updates.
* Add assembly assertions to tests in `zjit/src/backend/arm64/mod.rs` that are currently missing them
    * Some of them only test the code size like ` assert_eq!(16, cb.get_write_pos());` at the moment, but it's hard to understand what's changed from its failure. So it now tests the generated code instead.

### DIFF
--- a/zjit/src/asm/arm64/mod.rs
+++ b/zjit/src/asm/arm64/mod.rs
@@ -1213,6 +1213,7 @@ fn cbz_cbnz(num_bits: u8, op: bool, offset: InstructionOffset, rt: u8) -> [u8; 4
 mod tests {
     use super::*;
     use insta::assert_snapshot;
+    use crate::assert_disasm_snapshot;
 
     fn compile<R>(run: R) -> CodeBlock where R: FnOnce(&mut super::CodeBlock) {
         let mut cb = super::CodeBlock::new_dummy();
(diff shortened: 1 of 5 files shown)


zjit/src/asm/arm64/mod.rs
```diff
@@ -1246,128 +1246,146 @@ mod tests {
     #[test]
     fn test_add_reg() {
         let cb = compile(|cb| add(cb, X0, X1, X2));
-        cb.with_disasm(|disasm| assert_snapshot!(disasm, @"  0x0: add x0, x1, x2"));
+        #[cfg(feature = "disasm")]
```
XrXr: I tested locally with the following macro and it updates multiline snapshots fine:

```rust
macro_rules! assert_disasm_snapshot {
    ($cb: expr, @$($tt: tt)*) => {
        $cb.with_disasm(|disasm| {
            assert_snapshot!(disasm, @$($tt)*)
        })
    }
}
```
We should use a macro like that instead of requiring to sprinkle `#[cfg(feature = "disasm")]` everywhere.

XrXr (changes requested):
> When a snapshot spans across multiple lines, zjit-test-update copies code before that macro (e.g. cb.with_disasm(|disasm|) to those lines in the snapshot.

This problem wasn't clear to me so here's some notes about exactly what happens.

Given a mismatch with a multi-line snapshot like the following:

```diff
diff --git i/zjit/src/asm/arm64/mod.rs w/zjit/src/asm/arm64/mod.rs
index 176e51ee97..bd9e6c017d 100644
--- i/zjit/src/asm/arm64/mod.rs
+++ w/zjit/src/asm/arm64/mod.rs
@@ -1960,6 +1960,7 @@ mod tests {
             0x0: add x10, x11, x9, uxtx
             0x4: add x30, x30, x30, uxtx
             0x8: add sp, sp, xzr
+            asdfsdf
         "));
         assert_snapshot!(cb.hexdump(), @"6a61298bde633e8bff633f8b");
     }
```

Running `zjit-test-update` updates the snapshot incorrectly like so:

```diff
diff --git i/zjit/src/asm/arm64/mod.rs w/zjit/src/asm/arm64/mod.rs
index 176e51ee97..aab677343d 100644
--- i/zjit/src/asm/arm64/mod.rs
+++ w/zjit/src/asm/arm64/mod.rs
@@ -1956,11 +1956,11 @@ mod tests {
         add_extended(&mut cb, X30, X30, X30);
         add_extended(&mut cb, X31, X31, X31);

-        cb.with_disasm(|disasm| assert_snapshot!(disasm, @"
-            0x0: add x10, x11, x9, uxtx
-            0x4: add x30, x30, x30, uxtx
-            0x8: add sp, sp, xzr
-        "));
+        cb.with_disasm(|disasm| assert_snapshot!(disasm, @r"
+        cb.with_disasm(|disasm| 0x0: add x10, x11, x9, uxtx
+        cb.with_disasm(|disasm| 0x4: add x30, x30, x30, uxtx
+        cb.with_disasm(|disasm| 0x8: add sp, sp, xzr
+        cb.with_disasm(|disasm| "));
         assert_snapshot!(cb.hexdump(), @"6a61298bde633e8bff633f8b");
     }
 }
```

zjit/src/asm/arm64/mod.rs
k0kubun: Thanks for finding that. I added `assert_disasm_snapshot!` to remove `#[cfg(feature = "disasm")]` https://github.com/ruby/ruby/pull/14708/commits/bda1e18dd57a51f7a31f7601f44a1edfa7f21608.

PR #14713, opened 2025-10-03
Documentation "Update making_changes_to_stdlibs.md mirror Example"

oniram88:
CSV is no more part of the standard lib, but the documentation was not updated (the example link was broken for the master branch)

I random selected CGI.


nobu:
Thank you for the report and the patch.
But CGI is not a good example because `cgi.rb` under [ruby/ruby] and [ruby/cgi] are different.
In fact, there is no `cgi.gemspec` file under [ruby/ruby].
You can select another one, e.g., delegate, English, etc.

[ruby/ruby]: https://github.com/ruby/ruby
[ruby/cgi]: https://github.com/ruby/cgi

PR #14715, opened 2025-10-03
Set Ruby parser if the given ruby accepts `--parser`

nobu:
Now envutil.rb is a part of test-unit-ruby-core gem, which is used with older Ruby versions.


rhenium:
I opened https://github.com/ruby/test-unit-ruby-core/pull/14 just a while ago to fix the same issue. Do we want a workaround in ruby/ruby, even though it's only relevant for the gem users?

PR #14717, opened 2025-10-03
ZJIT: Add Insn::ArrayAref stub

aidenfoxivey:


zjit/src/codegen.rs
```diff
@@ -1200,6 +1201,22 @@ fn gen_new_array(
...
+    array: Opnd,
+    index: Opnd,
+    state: &FrameState,
+) -> lir::Opnd {
+    unsafe extern "C" {
+        fn rb_ary_entry(ary: VALUE, off: c_long) -> VALUE;
+    }
+
+    gen_prepare_leaf_call_with_gc(asm, state);
+
```
```suggestion
```

I don't think it allocates?

zjit/src/codegen.rs
```diff
@@ -1200,6 +1201,22 @@ fn gen_new_array(
...
+/// Compile array access (array[index])
+fn gen_array_aref(
+    asm: &mut Assembler,
+    array: Opnd,
+    index: Opnd,
+    state: &FrameState,
+) -> lir::Opnd {
+    unsafe extern "C" {
+        fn rb_ary_entry(ary: VALUE, off: c_long) -> VALUE;
+    }
```
k0kubun: It's declared in the header (`include/ruby/internal/intern/array.h`) already, so in this case I think we prefer bindgen. We do manual declaration when we do it only for JIT.

zjit/src/hir.rs
```diff
@@ -1942,6 +1949,15 @@ impl Function {
...
                     }
+                } else if self.type_of(idx_val).is_subtype(types::Fixnum) {
+                    self.push_insn(block, Insn::PatchPoint { invariant: Invariant::BOPRedefined { klass: ARRAY_REDEFINED_OP_FLAG, bop: BOP_AREF }, state });
+                    let result = self.push_insn(block, Insn::ArrayAref {
+                        array: self_val,
+                        index: idx_val,
+                        state
+                    });
+                    self.make_equal_to(orig_insn_id, result);
+                    return;
```
k0kubun: I think this should be outside `if let Some(array_obj) = self_type.ruby_object() {` (but still inside `if self_type.is_subtype(types::ArrayExact) {`).

And when you do so, you'd need to conditionally (i.e. when the type is not known at compile time) add a `GuardType`.

zjit/src/codegen.rs
aidenfoxivey: Okay, that sounds good!

zjit/src/hir.rs
aidenfoxivey: Ah okay - I will figure out how to do that. Thank you for your guidance.

k0kubun: This PR would need to add `attr bool zjit_profile = true;` to `opt_aref` in insns.def (and re-run bindgen) by the way.

aidenfoxivey:
image
Unsure exactly why CI might be failing.

XrXr:
Looks like you're not running btest with ZJIT enabled locally. The description should contain `+ZJIT`.

zjit/src/codegen.rs
```diff
@@ -1241,6 +1242,15 @@ fn gen_new_array(
     new_array
 }
 
+/// Compile array access (array[index])
+fn gen_aref_fixnum(
+    asm: &mut Assembler,
+    array: Opnd,
+    index: Opnd,
+) -> lir::Opnd {
+    asm_ccall!(asm, rb_ary_entry, array, index)
```
tekknolagi: `rb_ary_entry` expects a `long`, not a `VALUE`, so you need to shift `index` right by one. We may end up having an `UnboxFixnum` instruction in the future

zjit/src/hir.rs
```diff
@@ -883,6 +884,10 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             }
+            Insn::ArrayArefFixnum { array, index, .. } => {
+                write!(f, "ArrayAref {array}[{index}]")?;
```
```suggestion
                write!(f, "ArrayArefFixnum {array}, {index}")?;
```

zjit/src/hir.rs
```diff
@@ -1560,6 +1565,7 @@ impl Function {
             },
             &InvokeBuiltin { bf, ref args, state, return_type } => InvokeBuiltin { bf, args: find_vec!(args), state, return_type },
             &ArrayDup { val, state } => ArrayDup { val: find!(val), state },
+            &ArrayArefFixnum { array, index } => ArrayArefFixnum { array: find!(array), index },
```
```suggestion
            &ArrayArefFixnum { array, index } => ArrayArefFixnum { array: find!(array), find!(index) },
```

zjit/src/hir.rs
```diff
@@ -1963,6 +1970,15 @@ impl Function {
                         return;
                     }
                 }
+            } else if self.type_of(idx_val).is_subtype(types::Fixnum) {
+                self.push_insn(block, Insn::PatchPoint { invariant: Invariant::BOPRedefined { klass: ARRAY_REDEFINED_OP_FLAG, bop: BOP_AREF }, state });
+                self.push_insn(block, Insn::GuardType { val: idx_val, guard_type: types::Fixnum, state });
```
tekknolagi: nit: let's use the result of the guard in the specialized operation. so `let fixnum_idx = guard ...; areffixnum fixnum_idx` or something

insns.def
```diff
@@ -1519,6 +1519,7 @@ opt_aref
  * default_proc.  This is a method call.  So opt_aref is
  * (surprisingly) not leaf. */
 // attr bool leaf = false; /* has rb_funcall() */ /* calls #yield */
+// attr bool zjit_profile = true;
```
tekknolagi, line 1522: This alone won't get the types into the JIT; you also need to catch it in `src/profile.rs`. Let's split this PR into just a `zjit_profile=true` and `src/profile.rs` modification followed by the broader `ArrayArefFixnum` PR.

This will give us an idea of new C function perf and then separately additional delta from not opening a C frame

tekknolagi (changes requested):

tekknolagi:
Also please add some HIR tests with `inspect`

insns.def
aidenfoxivey: Sounds good - I'll open a new one for the smaller changes.

zjit/src/hir.rs
```diff
@@ -1968,6 +1975,15 @@ impl Function {
                         return;
                     }
                 }
+            } else if self.type_of(idx_val).is_subtype(types::Fixnum) {
```
k0kubun: This would miss the case that `self_type.ruby_object().is_some() && !array_obj.is_frozen()`, but I think it shouldn't?

It might be unreachable because of how ISEQ is compiled, but in that case it should use either `unreachable!()` or `assert!(array_obj.is_frozen())`. Otherwise, this should just be `if` instead of `else if` IMO.

k0kubun:
We probably want to have an HIR test that generates `ArrayArefFixnum`.

zjit/src/hir.rs
aidenfoxivey: Alright, I'll go with an `if`.

tekknolagi: I don't know that this should go here. Or at least it should have some additional bits like the SendWithoutBlockDirect part of type_specialize that look at the profiled type of the LHS. We may not know statically that we're looking at an ArrayExact until we look at the profiles and insert a guard.

You can also leave it as-is and this PR may not do a whole lot until we (if we) move the guard generation into the HIR builder or something

zjit/src/hir.rs
```diff
@@ -888,6 +889,10 @@ impl<'a> std::fmt::Display for InsnPrinter<'a> {
                 }
                 Ok(())
             }
+            Insn::ArrayArefFixnum { array, index, .. } => {
+                write!(f, "ArrayAref {array}[{index}]")?;
+                Ok(())
```
tekknolagi, line 894:
```suggestion
            Insn::ArrayArefFixnum { array, index, .. } => {
                write!(f, "ArrayAref {array}[{index}]")
```

zjit/src/hir.rs
tekknolagi: bump

zjit/src/hir.rs
k0kubun: and why are we not printing it as `ArrayArefFixnum`?

To be clear, I'm more than happy to just call it `Insn::ArrayAref`, but I just want the Insn name and Display be consistent.

tekknolagi: (agree; see my other comment)

zjit/src/hir.rs
k0kubun: Given how common `Fixnum` is for an `Array#[]` argument in general, I think we could rename `Insn::ArrayArefFixnum` to `Insn::ArrayAref` instead.

We can still rename it back to `Insn::ArrayArefFixnum` when we decide to introduce something like `Insn::ArrayArefRange`, which I'd argue is a much lower priority to optimize.

zjit/src/hir.rs
tekknolagi: I think probably you should leave it as-is in this PR and we will figure out LHS profiling as we get better infrastructure. This little frozen optimization thingy was a bit of a red herring

zjit/src/hir.rs
aidenfoxivey: Should be fixed!

zjit/src/hir.rs
aidenfoxivey: Hmm, okay, sounds good.

zjit/src/hir.rs
aidenfoxivey: Should be sorted.

k0kubun:
I guess we can merge this as is and work on it separately if we want, but can we consider doing this for the case that the profiled type of a receiver is `Array` (which is what I wanted to mean at https://github.com/ruby/ruby/pull/14717#discussion_r2403384353)? Like `type_specialize` does, you can look at `profiled_type_of_at` when `type_of` doesn't know the type, and add a `GuardType` of the receiver for the profiled case.

In many cases, you don't have the array literal that `Array#[]` reads, so guarding on the receiver type and specializing it into `ArrayArefFixnum` seems far more important than what the scope of this current PR.

aidenfoxivey:
> I'll still approve the PR, but please tell us whether you want to work on [#14717 (comment)](https://github.com/ruby/ruby/pull/14717#issuecomment-3390987744) in this PR or in another one.

I would prefer to work on this in a second PR if possible.

zjit/src/hir.rs
k0kubun: > Should be fixed!

Looks like you fixed `s/ArrayAref/ArrayArefFixnum/` but this thread was not about it (I'm sorry that I made it seem like it though). I want to proceed to merging it at this point, so I'll push a fix myself for you.

k0kubun:
👌 merging

zjit/src/hir.rs
k0kubun: https://github.com/ruby/ruby/commit/2de13f4d094a92099de92b91cde8d5a7da8c38cc

k0kubun:
> I would prefer to work on this in a second PR if possible.

Filed it as https://github.com/Shopify/ruby/issues/792

PR #14718, opened 2025-10-03
Make test-bundled-gems fails property when timed out on Windows

YO4:
Currently test-bundled-gems.rb terminates child process with Process.kill("KILL", pid) if that timed out on Windows.
Unfortunately, child process exits with 0 in that case, test-bundled-gems.rb misses to detect failure.

SIGINT usage is disabled because it does not function properly, but it can be enabled by correctly utilizing spawn.
Use a spawn call that bypasses the shell to make SIGINT work as intended.

This PR supersedes #14695.


hsbt:
👍  It seems useful for me. Thanks.

PR #14737, opened 2025-10-05
Use `ec->interrupt_mask` to prevent interrupts. [Backport #21610]

ioquatix:
Backport of https://github.com/ruby/ruby/pull/14588


hsbt:
Thanks for your work.

Unfortunately, this seems to not security issue and build failure. I don't want to backport normal bugfix to 3.2.

ioquatix:
That makes sense, I didn't realise 3.2 was security maintenance mode only already!

PR #14740, opened 2025-10-05
[DOC] String slices doc

BurdetteLamar:
Looking in general at `Regexp` issues in `String`, I found that the section "String Slices" does not do justice to its three constituent methods: `#[]`, `#[]=`, and `#slice!`.

So I'm proposing to remove the section and make the methods' documentation more fulsome.


doc/string/aref.rdoc
```diff
@@ -0,0 +1,107 @@
+Returns the substring of +self+ specified by the arguments.
+
+Note: The form <tt>self[regexp, capture = 0]</tt>
+updates {Regexp-related global variables}[rdoc-ref:Regexp@Global+Variables];
+the other forms do not.
```
peterzhu2118: We already say this in the "Form self[regexp, capture = 0]` section. I think it's a bit redundant to mention this twice since it's not that important anyways.

string.c
```diff
@@ -5804,16 +5804,14 @@ rb_str_aref(VALUE str, VALUE indx)
...
 /*
  *  call-seq:
+ *    self[substring] -> new_string or nil
+ *    self[regexp, capture = 0] -> new_string or nil
  *    self[index] -> new_string or nil
  *    self[start, length] -> new_string or nil
  *    self[range] -> new_string or nil
- *    self[regexp, capture = 0] -> new_string or nil
- *    self[substring] -> new_string or nil
  *
```
peterzhu2118: What's the reason to change the order here? It seems like the original order was in the order which this method is most commonly used, which I think we should keep (and also change the method documentation to follow it).

doc/string/aref.rdoc
```diff
@@ -0,0 +1,107 @@
...
+updates {Regexp-related global variables}[rdoc-ref:Regexp@Global+Variables]:
+
+  'hello'[/ell/]     # => "ell"
+  [$`, $&, $']       # => ["h", "ell", "o"]
+  'hello'[/l+/]      # => "ll"
+  [$`, $&, $']       # => ["he", "ll", "o"]
+  'hello'[//]        # => ""
+  [$`, $&, $']       # => ["", "", "hello"]
+  'hello'[/nosuch/]  # => nil
+  [$`, $&, $']       # => [nil, nil, nil]
```
peterzhu2118: What's the significance of showing the regexp global variables? It seems just follow regular regexp matching behavior.

doc/string/aref.rdoc
```diff
@@ -0,0 +1,107 @@
...
+  'hello'[-6] # => nil
+
+<b>Form <tt>self[start, length]</tt></b>
+
+With integer arguments +start+ and +length+ given,
+returns a substring of size +length+ characters (as available)
+beginning at character offset specified by +start+.
+
+If argument +start+ is non-negative,
+the offset is +start':
```
peterzhu2118:
```suggestion
the offset is +start+:
```

doc/string/aref.rdoc
```diff
@@ -0,0 +1,107 @@
...
+Special case: if +start+ equals the length of +self+,
+returns a new empty string:
+
+  'hello'[5, 3]  # => ""
+
+<b>Form <tt>self[range]</tt></b>
+
+With Range argument +range+ given,
+forms substring <tt>substring = self[range.start, range.size]</tt>
+and returns <tt>self[substring]</tt>:
```
peterzhu2118: Isn't this just equivalent to `self[range.start, range.size]`? In what cases does `self[substring]` make a difference?

doc/string/aset.rdoc
```diff
@@ -0,0 +1,187 @@
...
+the argument is appended to +self+:
+
+  s = 'hello'
+  s[5, 3] = 'foo' # => "foo"
+  s               # => "hellofoo"
+
+<b>Form <tt>self[range] = other_string</tt></b>
+
+With Range argument +range+ given,
+searches for substring <tt>substring = self[range.start, range.size]</tt>:
```
peterzhu2118: Maybe something like this?

```suggestion
With Range argument +range+ given,
equivalent to <tt>self[range.start, range.size] = other_string</tt>:
```

doc/string/aref.rdoc
BurdetteLamar: I was thinking to remind readers how useful these globals are.  Dr. Johnson tells us, “People need to be reminded more often than they need to be instructed.”  Will remove if you think it best.

doc/string/aref.rdoc
BurdetteLamar: Removed.

string.c
BurdetteLamar: Re-ordered (call-seq, 3 places, and text, 2 places).

doc/string/aref.rdoc
BurdetteLamar: Removed.

doc/string/aref.rdoc
BurdetteLamar: Fixed.

doc/string/aref.rdoc
BurdetteLamar: Fixed.

doc/string/aset.rdoc
BurdetteLamar: Fixed.

doc/string/aset.rdoc
```diff
@@ -0,0 +1,187 @@
+Returns +self+ with all, a substring, or none of its contents replaced;
+returns the argument +other_string+.
+
+Note: The form <tt>self[regexp, capture = 0]=</tt>
+updates {Regexp-related global variables}[rdoc-ref:Regexp@Global+Variables];
+the other forms do not.
```
peterzhu2118: Ditto here on repeating information.

doc/string/aref.rdoc
```diff
@@ -0,0 +1,98 @@
...
+
+Special case: if +start+ equals the length of +self+,
+returns a new empty string:
+
+  'hello'[5, 3]  # => ""
+
+<b>Form <tt>self[range]</tt></b>
+
+With Range argument +range+ given,
+forms substring <tt>substring = self[range.start, range.size]</tt>:
```
peterzhu2118: I don't think we need to set anything here.

```suggestion
forms substring <tt>self[range.start, range.size]</tt>:
```

doc/string/aset.rdoc
BurdetteLamar: Removed.

doc/string/aref.rdoc
BurdetteLamar: Fixed.

PR #14745, opened 2025-10-06
Remove unused variable warning

etiennebarrie:
$ make test/ruby/test_thread.rb RUBYOPT=-w >/dev/null
    test/ruby/test_thread.rb:1595: warning: assigned but unused variable - bug21127

### DIFF
--- a/test/ruby/test_thread.rb
+++ b/test/ruby/test_thread.rb
@@ -1592,7 +1592,6 @@ q.pop
 
   # [Bug #21342]
   def test_unlock_locked_mutex_with_collected_fiber
-    bug21127 = '[ruby-core:120930] [Bug #21127]'
     assert_separately([], "#{<<~"begin;"}\n#{<<~'end;'}")
     begin;
       5.times do


test/ruby/test_thread.rb
```diff
@@ -1593,6 +1593,7 @@ def frame_for_deadlock_test_2
   # [Bug #21342]
   def test_unlock_locked_mutex_with_collected_fiber
     bug21127 = '[ruby-core:120930] [Bug #21127]'
+    bug21127 = bug21127
```
jhawthorn: I think we should probably just remove these two lines. I think it was copied by accident from above (this bug number does not match the `# [Bug #21342]` number

etiennebarrie: Oh right 🤦 
Should I include that bug number instead? Is it documented how to use these tags? I see a lot of local variables with strings with the email id/bug id, but can we just use a comment with simply the bug id instead?

jhawthorn: We can just use a comment, I don't think there's anything special about them. We have them in locals sometimes to pass them to assert so that if the assertion fails it shows up in CI with the issue number.

PR #14747, opened 2025-10-06
Bump RDoc

st0012:
[RDoc 6.15.0](https://github.com/ruby/rdoc/releases/tag/v6.15.0) has some bug fixes and an enhancement that'll directly help docs.ruby-lang.org's search experience.


st0012:
The failing tests are all from `test-bundled-gems-spec`, related to `BigDecimal`. And the [failed RDoc ones](https://github.com/ruby/ruby/actions/runs/18336333447/job/52221217936) have been fixed. It'd be great if we can merge this first

k0kubun (approved):

PR #14761, opened 2025-10-07
Add debug #define to call sched_yield before each pthread_mutex_lock

luke-gruber:
This is useful for debugging mutex issues as it increases contention for locks. It is off by default.


jhawthorn:
I like it!

Would it be useful to do on unlock too? (Probably best to leave as an exercise for whoever is trying to reproduce something that requires that)

PR #14763, opened 2025-10-07
Fix typo in comment in array#zip docs

RicardoTrindade:
Simple fix where I believe the word is duplicated in the documentation for the array#zip method


XrXr:
Thanks!

PR #14919, opened 2025-10-23
Update vendored URI to 1.0.4 for Ruby 3.4

hsbt:
https://www.ruby-lang.org/en/news/2025/10/07/uri-cve-2025-61594/


k0kubun (approved):
Thanks!

PR #14923, opened 2025-10-23
[Feature #21572] Syntax error at capture in alternation pattern

nobu:
Should fail even with `-c` option.

[[Feature #21572]](https://bugs.ruby-lang.org/issues/21572)


prism/prism.c
```diff
@@ -16960,22 +16960,30 @@ parse_strings(pm_parser_t *parser, pm_node_t *current, bool accepts_label, uint1
...
  */
 static void
-parse_pattern_capture(pm_parser_t *parser, pm_constant_id_list_t *captures, pm_constant_id_t capture, const pm_location_t *location) {
+parse_pattern_capture(pm_parser_t *parser, pm_constant_id_list_t *captures, pm_constant_id_t capture, const pm_location_t *location, pm_pattern_capturing_t *capturing) {
     // Skip this capture if it starts with an underscore.
     if (*location->start == '_') return;
 
     if (pm_constant_id_list_includes(captures, capture)) {
         pm_parser_err(parser, location->start, location->end, PM_ERR_PATTERN_CAPTURE_DUPLICATE);
+    } else if (capturing->in_alternative_pattern) {
```
Earlopain: prism can parse multiple ruby versions and the prism gem does get backported to stable branches.
Since this practically already is a syntax error I am unsure of how important that actually is, but I did want to mention it. Same below.

```suggestion
    } else if (capturing->in_alternative_pattern && parser->version >= PM_OPTIONS_VERSION_CRUBY_3_5) {
```

prism/prism.c
```diff
@@ -16960,22 +16960,30 @@ parse_strings(pm_parser_t *parser, pm_node_t *current, bool accepts_label, uint1
...
 #define PM_PARSE_PATTERN_MULTI 2
 
+typedef struct {
+    bool in_alternative_pattern;
+    bool capture_in_pattern;
+} pm_pattern_capturing_t;
+
 static pm_node_t *
-parse_pattern(pm_parser_t *parser, pm_constant_id_list_t *captures, uint8_t flags, pm_diagnostic_id_t diag_id, uint16_t depth);
+parse_pattern(pm_parser_t *parser, pm_constant_id_list_t *captures, uint8_t flags, pm_diagnostic_id_t diag_id, uint16_t depth, pm_pattern_capturing_t *capturing);
```
Earlopain: Can you reorder arguments here (and in other places) so that diag_id and depth come last (that's convention in the code)? Moving it after captures seems most logical to me.

Earlopain (commented):
Thanks! You can also remove a bunch of code in `compile.c` and `prism_compile.c` that currently deals with this, right?

kddnewton:
@nobu I'm not sure if this fixes the problem fully. This appears to only work on the top level, so it still gets a compile error if the variable is nested, as in:

```ruby
a => [b] | 1
```

I think this might need a different approach.

Earlopain:
> I'm not sure if this fixes the problem fully

That specific case was addressed in https://github.com/ruby/ruby/pull/15329

PR #4563, opened 2021-06-11
Make it possible to create instances of `Process::Status`.

ioquatix:
In order to implement process wait in schedulers we need to be able to create `Process::Status` objects.

PR #15213, opened 2025-11-17
Add `rb_process_status_for` to `intern/process.h`.

noteflakes:
https://bugs.ruby-lang.org/issues/21704

This PR exposes `rb_process_status_new` by adding its declaration to `intern/process.h`.

This is needed in order to allow `Fiber::Scheduler` implementations to directly create instances of `Process::Status`, instead of having to call `Process::Status.wait`. An io_uring-based fiber scheduler could implement `process_wait` by using `io_uring_prep_waitid`, which waits for a child to exit and returns the child PID and exit status, which are then be passed to `rb_process_status_new`.

cc @ioquatix

### DIFF
--- a/include/ruby/internal/intern/process.h
+++ b/include/ruby/internal/intern/process.h
@@ -31,6 +31,16 @@ RBIMPL_SYMBOL_EXPORT_BEGIN()
 /* process.c */
 
 /**
+ * Creates a new instance of Process::Status.
+ *
+ * @param[in] pid The process ID.
+ * @param[in] status The process status.
+ * @param[in] error Error code.
+ * @return VALUE An instance of Process::Status.
+ */
+VALUE rb_process_status_new(rb_pid_t pid, int status, int error);
+
+/**
  * Wait for the specified process to terminate, reap it, and return its status.
  *
  * @param[in] pid The process ID to wait for.
--- a/include/ruby/internal/intern/process.h
+++ b/include/ruby/internal/intern/process.h
@@ -34,8 +34,8 @@ RBIMPL_SYMBOL_EXPORT_BEGIN()
  * Creates a new instance of Process::Status.
  *
  * @param[in] pid The process ID.
- * @param[in] status The process status.
- * @param[in] error Error code.
+ * @param[in] status The "waitpid status", as returned by waitpid(2). This is NOT the exit status/exit code, see waitpid(2).
+ * @param[in] error Error code (if waitpid(2) returned -1).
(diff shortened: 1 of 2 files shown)


noteflakes:
See also #4563

include/ruby/internal/intern/process.h
```diff
@@ -30,6 +30,16 @@ RBIMPL_SYMBOL_EXPORT_BEGIN()
 
 /* process.c */
 
+/**
+ * Creates a new instance of Process::Status.
+ *
+ * @param[in] pid The process ID.
+ * @param[in] status The process status.
+ * @param[in] error Error code.
```
eregon: Could this be documented better? It's the error code corresponding to which system call, etc.

include/ruby/internal/intern/process.h
```diff
@@ -30,6 +30,16 @@ RBIMPL_SYMBOL_EXPORT_BEGIN()
 
 /* process.c */
 
+/**
+ * Creates a new instance of Process::Status.
+ *
+ * @param[in] pid The process ID.
+ * @param[in] status The process status.
```
eregon:
```suggestion
 * @param[in] status The "waitpid status", as returned by waitpid(2). This is NOT the exit status/exit code, see waitpid(2).
```
i.e. it's not the "exit status"/"exit code" (which can be accessed with `WEXITSTATUS(wstatus)`, how confusing), it's more than that.

eregon (commented):
LGTM, just needs clearer docs

include/ruby/internal/intern/process.h
noteflakes: Fixed in d30aa0c654dd34e666fb9c493b19bbddda4cce40.

include/ruby/internal/intern/process.h
noteflakes: Fixed in d30aa0c654dd34e666fb9c493b19bbddda4cce40.

eregon (approved):

eregon:
@noteflakes Did you delete this branch on purpose?
It's meant to be discussed at the next dev meeting: https://bugs.ruby-lang.org/issues/21689#note-4

noteflakes:
Ooh sorry @eregon, deleted by mistake.

ioquatix:
I guess we need to discuss the design on the issue tracker in more detail.

ioquatix:
Follow up PR in `io-event`: https://github.com/socketry/io-event/pull/203

PR #15296, opened 2025-11-22
Fix Set#^ to not mutate its argument

gildesmarais:
Great changes on `Set` - thank you, @jeremyevans!

`Set#^` in the new C backend was mutating the right-hand set when the operand was a Set and returned that mutated object. 
This violated the documented contract (“Returns a new set”) and made XOR non-commutative when the right operand was shared.

Fix: always produce a fresh result (`dup` the receiver), and for non-Set enumerables, deduplicate via a temporary Set before toggling so duplicates don’t double-flip.

Added regression test `test_xor_does_not_mutate_other_set` in `test/ruby/test_set.rb`.


jeremyevans (approved):
Thank you for finding and fixing this!

jeremyevans:
@knu This looks correct to me, so I plan to merge it within a couple days unless you have objections.

knu (approved):
Looks correct to me.

test_xor() is not good enough and can be improved, too.

gildesmarais:
Glad to see the approvals. Let me know if anything else is needed to get this merged. Happy to support. 🧡 Thank you!

gildesmarais:
> test_xor() is not good enough and can be improved, too.

Had a look into Set's ruby tests and noticed there likely is a difference to Hash-based Sets when `compare_by_identity` is used. 

Method documentation reads:

> "Makes the set compare its elements by their identity and returns self. This method may not be supported by all subclasses of Set." 
> Source: https://docs.ruby-lang.org/en/3.4/Set.html#method-i-compare_by_identity


In a [separate branch of my fork](https://github.com/ruby/ruby/compare/master...gildesmarais:ruby:test/set-test-gaps), I've added "missing" tests and a fixing attempt when `compare_by_identity` is used. 

You're invited to have a look. I will be happy to push these upstream - just please let me know how you'd like to see these contributed.

test/ruby/test_set.rb
```diff
@@ -703,6 +703,17 @@ def test_xor
     }
   end
 
+  def test_xor_does_not_mutate_other_set
+    a = Set[1]
+    b = Set[1, 2]
```
tekknolagi, line 708: I think we can go further and even `.freeze` `a` and `b`

jeremyevans:
@knu I think we should merge this soon so it makes the next preview release. Unless I hear objections, I plan to squash merge this around 10 AM JST.

knu:
Thank you.  I'm merging this!

PR #15447, opened 2025-12-08
[Misc #21768] Remove deprecated APIs

nobu:
[[Misc #21768]](https://bugs.ruby-lang.org/issues/21768)

ruby-bench tests failures are:
- `RData`
  - deivid-rodriguez/byebug#924
  - socketry/cool.io#85
  - ruby-concurrency/concurrent-ruby#1096
  - ~~gjtorikian/commonmarker#424~~
    `lobsters` vendored in `ruby-bench` and `commonmarker` used by it are outdated
  - tmm1/http_parser.rb#87
  - brianmario/yajl-ruby#232
  - oxidize-rb/rb-sys#688 (a related bug in the test code)
- `rb_postponed_job_register_one`
  - tmm1/stackprof#240

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -185,19 +185,14 @@ define rp
     print (struct RBasic *)($arg0)
   else
   if ($flags & RUBY_T_MASK) == RUBY_T_DATA
-    if ($flags & RUBY_TYPED_FL_IS_TYPED_DATA)
-      set $data = (struct RTypedData *)($arg0)
-      set $type = (const rb_data_type_t *)($data->type & ~1)
-      printf "%sT_DATA%s(%s): ", $color_type, $color_end, $type->wrap_struct_name
-      print *$type
-      if ($data->type & 1)
-        print (void *)&$data->data
-      else
-        print $data
-      end
+    set $data = (struct RTypedData *)($arg0)
+    set $type = (const rb_data_type_t *)($data->type & ~1)
+    printf "%sT_DATA%s(%s): ", $color_type, $color_end, $type->wrap_struct_name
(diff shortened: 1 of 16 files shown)


eregon:
Nice.
Re `Remove struct RData deprecated by struct RTypedData` I know one usage in [concurrent-ruby](https://github.com/ruby-concurrency/concurrent-ruby/blob/master/ext/concurrent-ruby-ext/atomic_reference.c).
I suppose the reason is TypedData was not available yet for the oldest supported version back then, but that seems fine now, so we'll fix it in concurrent-ruby.
Related discussion: https://bugs.ruby-lang.org/issues/19998
I think it's good to remove RData if we can.

k0kubun:
If you decide to merge this PR, please consider merging https://github.com/ruby/ruby-bench/pull/455 to fix the ruby-bench CI. You should have write permission on the repository.

sandstrom:
Nice to see that Ruby is cleaning up deprecated and old methods.

I love Ruby, but there is a bit of cruft in the stdlib, so nice to see some of it go away (still more things that I think could be cut, so that effort can instead be spent on making the frequently used parts of the stdlib more ergonomic, better documented, etc).

larskanis:
Can you please merge this PR early in the release cycle? There are quite some gems out there which still use the untyped data API. I have at least 2 actively maintained gems, which still makes use of it: "fxruby" and "pkcs11".

hsbt:
Now this PR will work with https://github.com/ruby/ruby-bench/pull/476/changes/5893da21261b337ef0e4ccc1358d254548155e09.

PR #15455, opened 2025-12-09
Declare legacy RData as TypedData

jhawthorn:
I was realizing with the new `TYPED_DATA_EMBEDDED` it should be possible to store an RData as a proper valid RTypedData. We do this by marking the T_DATA as embedded, and having `void *data` be the first element of the struct (so that it's in the same location) after that pointer we can store dmark and dfree.

I think there is still usage of the old API in the wild and this should remove most of our maintenance burden internally.

cc @byroot @nobu What do you think? Is this worth doing? I can do more testing and clean this up for merging if you think so.

There is one potential incompatibility with this: RTYPEDDATA_P will return true for all T_DATA. (I suppose we could remove all references to `RTYPEDDATA_P` internally and add a new method to check whether it's this special type of TypedData, but that seems unnecessary)

```
CApiWrappedTypedStruct RTYPEDDATA_P returns false for an untyped data object FAILED
Expected true == false
to be truthy but was false
/Users/jhawthorn/src/ruby/spec/ruby/optional/capi/typed_data_spec.rb:97:in 'block (3 levels) in <top (required)>'
/Users/jhawthorn/src/ruby/spec/ruby/optional/capi/typed_data_spec.rb:21:in '<top (required)>'
```


gc.c
```diff
@@ -1059,17 +1060,67 @@ rb_data_object_check(VALUE klass)
...
+/*
+ * Legacy RData support: RData is implemented as an embedded RTypedData
+ * with the mark/free/data stored in the payload. RDATA(obj)->data still
+ * works because the payload starts at the same offset as RData.data.
+ */
+struct rb_legacy_rdata {
+    void *data;           /* must be first for RDATA(obj)->data compatibility */
+    RUBY_DATA_FUNC dmark;
+    RUBY_DATA_FUNC dfree;
+};
```
nobu: It is possible to keep the order as previous and make extra embed size 0.
You can obtain the container pointer from the member pointer with `ccan_container_of`.

spec/ruby/optional/capi/typed_data_spec.rb
```diff
@@ -86,15 +86,4 @@
...
-    it "returns true for a typed data" do
-      a = @s.typed_wrap_struct(1024)
-      @s.RTYPEDDATA_P(a).should == true
-    end
-
-    it "returns false for an untyped data object" do
-      a = @s.untyped_wrap_struct(1024)
-      @s.RTYPEDDATA_P(a).should == false
-    end
-  end
```
nobu, line 99: Should enclose in `ruby_version_is ""..."4.0" do` / `end`.

nobu:
Apart from this PR, [Always treat encoding as TYPEDDATA](https://github.com/ruby/ruby/pull/15455/commits/47299639bd4105085e20ecb734851abc7792bf36) should be merged, I think.

gc.c
```diff
@@ -1059,17 +1060,67 @@ rb_data_object_check(VALUE klass)
...
+
+/* Verify struct RData fields line up with embedded rb_legacy_rdata in RTypedData */
+STATIC_ASSERT(rdata_data, offsetof(struct RData, data) == offsetof(struct RTypedData, data) + offsetof(struct rb_legacy_rdata, data));
+STATIC_ASSERT(rdata_dmark, offsetof(struct RData, dmark) == offsetof(struct RTypedData, data) + offsetof(struct rb_legacy_rdata, dmark));
+STATIC_ASSERT(rdata_dfree, offsetof(struct RData, dfree) == offsetof(struct RTypedData, data) + offsetof(struct rb_legacy_rdata, dfree));
```
byroot: Do we even need this anymore? If legacy RData is retrofitted inside RTypedData, we don't care about offsets. AFAIK these were only necessary to be able to distinguish the two.

gc.c
```diff
@@ -1059,17 +1060,67 @@ rb_data_object_check(VALUE klass)
...
+    if (rdata->dfree == RUBY_DEFAULT_FREE) {
+        xfree(rdata->data);
+    }
+    else if (rdata->dfree) {
+        rdata->dfree(rdata->data);
+    }
+}
+
+static const rb_data_type_t rb_legacy_rdata_type = {
+    .wrap_struct_name = "legacy RData",
```
byroot:
```suggestion
    .wrap_struct_name = "deprecated RData",
```

gc.c
byroot: Or do you want to keep supporting `RDATA(obj)->dmark` and `RDATA(obj)->dfree` ? I can't imagine how an extension would legitimately need to do that, to me that always was private API.

byroot (commented):
I've been meaning to retrofit RData as RTypedData since about two years, but your solution is much better than the one I was toying with.

I believe this is very valuable as `RData` is very rare now, officially deprecated, and removing many conditionals in hotspots should help with performance (as demonstrated in https://github.com/ruby/ruby/pull/15387). And I think you now can remove the `RTYPEDDATA_P(obj)` condition in `rbimpl_check_typeddata`.

My only worry is that we're perhaps a bit late in the release cycle for such a change, but I'd really really like that change for 4.0 though.

byroot:
> There is one potential incompatibility with this: RTYPEDDATA_P will return true for all T_DATA.

I think after this change, `RTYPEDDATA_P` has almost no more usefulness internally, so we can just remove about all references of it. If you are worried about 3rd party usage, you can define it to be something like:

```c
RTYPEDDATA_TYPE(obj) != &deprecated_rdata_type
```

But we probably deprecate this as well as part of https://bugs.ruby-lang.org/issues/19998

byroot:
You should also be able to get rid of `RUBY_TYPED_FL_IS_TYPED_DATA`.

jeremyevans (commented):
Overall, I think this is good change. I worry it may be a little late in the dev cycle to introduce this change, though. We are only a little over two weeks from release.

I think we may want a developer with more GC knowledge that I have to review the commented out code in gc.c. We may want to wait to commit this until we have a better understanding about why the existing code was wrong, and what the correct code would be. If we already have that understanding, we should remove or replace the commented out code and include the reasoning in the commit message.

gc.c
jhawthorn: We need somewhere to store `dmark` and `dfree` and RTypedData needs the location those used to be in for RData

gc.c
jhawthorn: I had intended to keep `dmark`/`dfree` just in case they were in use, but if we're aiming for Ruby 4.1 I can see removing them. We should probably keep `RDATA(obj)->data` for now

jhawthorn:
I agree it's probably too late in the cycle for this. I'll wait until January.

> I think we may want a developer with more GC knowledge that I have to review the commented out code in gc.c. We may want to wait to commit this until we have a better understanding about why the existing code was wrong, and what the correct code would be. If we already have that understanding, we should remove or replace the commented out code and include the reasoning in the commit message.

The `rb_gc_shutdown_call_finalizer_p` change? The function tells us whether we need to call the finalizer for an object during VM shutdown. For `T_DATA` we generally have to, this was trying to skip them for RData/RTypedData with no `data` or no `dfree` (where we would not call the free function normally). However it was "wrong" for TypedData because `RDATA(obj)->dfree` is going to accidentally read from `RTYPEDDATA(obj)->type` which will never be 0.

There's no reason to test this. `data` being `NULL` is rare, and we can just let the free code later treat that as a no-op. I'll do this in a separate PR.

gc.c
jhawthorn: It does seem like it was a common pattern to check `RDATA(obj)->dmark == some_known_mark_function` in the days before typed data

byroot: Interesting, I don't think I ever saw this.

But all things considered, the only downside is that legacy RData will end up bigger, which is a perfectly fine tradeoff for something that has been deprecated for ages, and that no popular gem I know of still use.

This PR seem very valuable to me, so 💯

byroot (approved):
Shall we merge this?

byroot:
Poke :) I can take care of rebasing this PR if you wish.

jhawthorn:
I found what might be a fundamental problem with this approach on MMTK:

I defined the special deprecated struct for this as:

```
static const rb_data_type_t rb_deprecated_rdata_type = {
    .wrap_struct_name = "RData(deprecated)",
    .function = {
        .dmark = rdata_deprecated_mark,
        .dfree = rdata_deprecated_free,
    },
    .flags = RUBY_TYPED_EMBEDDABLE,
};
```

However that won't work under MMTK. All other RUBY_TYPED_EMBEDDABLE also use RUBY_TYPED_FREE_IMMEDIATELY, and that's important so that we can call the free function on the allocated memory and also do so at finalizer time while holding the GVL.

byroot:
Ah good catch, I forgot about `FREE_IMMEDIATELY` being a requirement for embedded typedata

PR #15529, opened 2025-12-12
Delay GVL handoff on blocking operation

ko1:
When a thread invokes the blocking operation such as I/O, the interpreter currently tries to pass the GVL to another thread immediately. However, if the blocking operation completes in a sufficiently short amount of time, the cost of context switching, especially on multi-core machines outweigh the benefit.

This patch introduces delayed GVL passing for blocking operations.

Scenario1: Short blocking operation

(1) `thread_sched_to_waiting_common()` does not immediately
    schedule another ready thread, but instead set the `waiting` flag.
(2) When the blocking operation finishes,
    `thread_sched_to_running_common()` is called without a context switching
    and `sched->running` is not changed.
    Clear the `waiting` flag and keep going.

Scenario2: Long blocking operation

(1) Same as above.
(2) The timer thread checks the running threads to provide
    timeslice every 10ms and finds the thread (th) performing
    blocking operation.
(3) The timer thread switches the running thread from th to another
    ready thread.
(4) When the blocking operation (on th) finishes,
    `thread_sched_to_running_common()` is called and `sched->running`
    should be another thread (or be NULL) GVL.

fix https://bugs.ruby-lang.org/issues/21685 (and the idea is from discussions)

### DIFF
--- a/thread_pthread.c
+++ b/thread_pthread.c
@@ -506,6 +506,12 @@ ASSERT_ractor_sched_locked(rb_vm_t *vm, rb_ractor_t *cr)
     VM_ASSERT(cr == NULL || vm->ractor.sched.lock_owner == cr);
 }
 
+static void
+ASSERT_ractor_sched_unlocked(rb_vm_t *vm, rb_ractor_t *cr)
+{
+    VM_ASSERT(cr == NULL || vm->ractor.sched.lock_owner != cr);
+}
+
 RBIMPL_ATTR_MAYBE_UNUSED()
 static bool
 ractor_sched_running_threads_contain_p(rb_vm_t *vm, rb_thread_t *th)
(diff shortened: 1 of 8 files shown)


thread_pthread.c
```diff
@@ -3025,14 +3056,74 @@ timer_thread_check_timeout(rb_vm_t *vm)
...
+    if (waiting_ths) {
+        int i=0;
+        rb_thread_t *th;
+        while ((th = waiting_ths[i++]) != NULL) {
+            // TODO: `th` can be freed. how to prevent the freeing?
+
+            struct rb_thread_sched *sched = TH_SCHED(th);
+            thread_sched_lock(sched, th);
+            {
+                if (th->sched.waiting && sched->running == th && sched->readyq_cnt > 0) {
```
luke-gruber, line 3072: Here, I think it would be good to check an `event_serial` from the thread to make sure the `waiting` you are seeing is the right one. The thread could have already context switched to another thread, setting `th->sched.waiting` to false and then the next thread could have context switched back to `th`. So, `th` could only be waiting for a very small amount of time.

PR #15540, opened 2025-12-13
Fix documentation of RB_PASS_CALLED_KEYWORDS in C API

eregon:


jeremyevans (approved):
This seems fine, but we really should document the supported `kw_splat` values in a single place, instead of having duplicate copies for every method using `kw_splat`.

eregon:
> This seems fine, but we really should document the supported `kw_splat` values in a single place, instead of having duplicate copies for every method using `kw_splat`.

Yeah, not sure if Dosygen has a good way to do that.
Also when looking at the source it seems more convenient to have it explained right there, rather than trying to "manually resolve" a documentation link/reference.

eregon:
Merged manually as b423204cb3851d1ad22d76578f7b516f1543c0dc since CI is broken on master and this PR is for sure not causing it.

PR #15795, opened 2026-01-03
New alloc_func Interface

byroot:
[Bug #21818]

Currently exceptions can be sent across ractors, but because of a limitation in the API, the exception backtrace is duped as an empty backtrace.

The problem is that backtraces are embedded objects, hence the classic `rb_class_alloc(klass)` API is insufficient because we need to know the size of the Backtrace object we're duping to instantiate the copy.

This is solved by introducing a replacement for `rb_define_alloc_func` which takes a function that receive both the class and the object being copied.

When used for an initial allocation, the copied object is set to Qundef.

### DIFF
--- a/object.c
+++ b/object.c
@@ -571,9 +571,19 @@ rb_obj_clone_setup(VALUE obj, VALUE clone, VALUE kwfreeze)
 }
 
 static VALUE
+obj_alloc_copy(VALUE obj)
+{
+    VALUE klass = rb_obj_class(obj);
+    if (klass && RCLASS_EXT_PRIME(klass)->copy_allocator) {
+        return RCLASS_COPY_ALLOCATOR(klass)(klass, obj);
+    }
+    return rb_obj_alloc(klass);
+}
+
+static VALUE
 mutable_obj_clone(VALUE obj, VALUE kwfreeze)
 {
-    VALUE clone = rb_obj_alloc(rb_obj_class(obj));
+    VALUE clone = obj_alloc_copy(obj);
     return rb_obj_clone_setup(obj, clone, kwfreeze);
 }
 
@@ -640,7 +650,7 @@ rb_obj_dup(VALUE obj)
     if (special_object_p(obj)) {
         return obj;
     }
-    dup = rb_obj_alloc(rb_obj_class(obj));
+    dup = obj_alloc_copy(obj);
     return rb_obj_dup_setup(obj, dup);
 }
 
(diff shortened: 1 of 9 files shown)


object.c
```diff
@@ -2307,7 +2317,12 @@ class_call_alloc_func(rb_alloc_func_t allocator, VALUE klass)
 
     RUBY_DTRACE_CREATE_HOOK(OBJECT, rb_class2name(klass));
 
-    obj = (*allocator)(klass);
+    if (RCLASS_EXT_PRIME(klass)->copy_allocator) {
```
nobu, line 2320: What `allocator` will come here for a subclass of copy_allocator class?
May not this flag mismatch with it?

byroot: I think `rb_init_alloc_func` handles that.

vm_method.c
```diff
@@ -1637,34 +1637,89 @@ rb_method_entry_set(VALUE klass, ID mid, const rb_method_entry_t *me, rb_method_
 
 #define UNDEF_ALLOC_FUNC ((rb_alloc_func_t)-1)
 
+static VALUE
+copy_allocator_adapter(VALUE klass)
+{
+    if (LIKELY(RCLASS_COPY_ALLOCATOR_P(klass))) {
+        rb_copy_alloc_func_t func = RCLASS_COPY_ALLOCATOR(klass);
+        return func(klass, Qundef);
```
eregon, line 1645: This will likely crash though as the `rb_copy_alloc_func_t` won't expect a Qundef.

eregon: Nevermind, after re-reading the design it's clear all `rb_copy_alloc_func_t` need to handle Qundef (https://bugs.ruby-lang.org/issues/21852#Proposed-new-API)

object.c
eregon, line 2320: This extra check seems a non-trivial overhead added to all object allocations (mentioned in https://bugs.ruby-lang.org/issues/21852#note-8). And same for dup/clone of course.
I think we don't need `rb_copy_alloc_func_t`, at least not for `Thread::Backtrace` where we can define dup & clone (reusing internal functions) or find a more direct way to allow passing it between Ractors.

byroot: > This extra check seems a non-trivial overhead

It's just a pointer dereference.

```c
#define RCLASS_EXT_PRIME(c) (&((struct RClass_and_rb_classext_t*)(c))->classext)
```

I highly doubt it would have a measurable impact compared to the cost of allocating an object.

byroot:
Closing in favor of https://github.com/ruby/ruby/pull/18499

PR #15849, opened 2026-01-11
Add YJIT test for outdated comment

byroot:
As of Ruby 4.0, the YJIT comment isn't quite correct because we now store the reference to the `imemo/fields` object inline.

Which means that a Struct of precisely 78 members (max embeddable on 64bit archs) can cohexist in embedded and heap shapes as demonstrated by the added tests.

If the comment was correct I would expect we could easily crash YJIT, but I'm unable to, so I suspect there is a guard I'm not seeing that already handle that.

cc @eregon @tekknolagi I just stumbled on this, I'm actually surprised it doesn't crash, but can't figure out how? Do any of you know?

I think we should add a test case for that corner case and update the comment, but I'd need help for that.

### DIFF
--- a/test/ruby/test_yjit.rb
+++ b/test/ruby/test_yjit.rb
@@ -644,6 +644,40 @@ class TestYJIT < Test::Unit::TestCase
     RUBY
   end
 
+  STRUCT_MAX_EMBEDDED_MEMBERS = (
+    GC::INTERNAL_CONSTANTS[:RVARGC_MAX_ALLOCATE_SIZE] -
+    GC::INTERNAL_CONSTANTS[:RBASIC_SIZE] -
+    GC::INTERNAL_CONSTANTS[:RVALUE_OVERHEAD]
+  ) / RbConfig::SIZEOF["void*"]
+
+  def test_spilled_struct_aref
+    omit("FIXME: https://github.com/Shopify/ruby/issues/977")
+    assert_compiles(<<~RUBY)
+      LargeStruct = Struct.new(:foo, :bar, *(#{STRUCT_MAX_EMBEDDED_MEMBERS} - 2).times.map { :"m_\#{it}" })
+
+      def foo(obj)
+        foo = obj.foo
+        raise "Expected 1, got: \#{foo}" unless foo == 1
+        bar = obj.bar
+        raise "Expected 2, got: \#{bar}" unless bar == 2
+      end
+
+      embedded_struct = LargeStruct.new(1, 2)
+      # Bump RCLASS_MAX_IV_COUNT for LargeStruct
+      embedded_struct.instance_variable_set(:@test, 1)
+
+      # Next allocation reserves space for the imemo/fields reference.
+      heap_struct = LargeStruct.new(1, 2)
+
+      RubyVM::YJIT.reset_stats!
+
+      foo(embedded_struct)
+      foo(embedded_struct)
+      foo(heap_struct)
+      foo(heap_struct)
+    RUBY
+  end
+
   def test_struct_aset
     assert_compiles(<<~RUBY)
       def foo(obj)
--- a/test/ruby/test_struct.rb
+++ b/test/ruby/test_struct.rb
@@ -41,8 +41,14 @@ module TestStruct
     end
   end
 
+  MAX_EMBEDDED_MEMBERS = (
+    GC::INTERNAL_CONSTANTS[:RVARGC_MAX_ALLOCATE_SIZE] -
+    GC::INTERNAL_CONSTANTS[:RBASIC_SIZE] -
+    GC::INTERNAL_CONSTANTS[:RVALUE_OVERHEAD]
+  ) / RbConfig::SIZEOF["void*"]
+
   def test_larger_than_largest_pool
-    count = (GC::INTERNAL_CONSTANTS[:RVARGC_MAX_ALLOCATE_SIZE] / RbConfig::SIZEOF["void*"]) + 1
+    count = MAX_EMBEDDED_MEMBERS + 1
     list = Array(0..count)
     klass = @Struct.new(*list.map { |i| :"a_#{i}"})
     struct = klass.new(*list)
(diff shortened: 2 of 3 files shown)


k0kubun:
> it doesn't crash, but can't figure out how?

Did you check these `embedded_struct` and `heap_struct` objects have different `RSTRUCT_EMBED_LEN_MASK` results and have values in different locations? If they do but happen to point to valid memory, maybe you could test the return values of `obj.foo` and `obj.bar` and let the test fail, even if the test doesn't crash?

byroot:
> Did you check these `embedded_struct` and `heap_struct` objects have different `RSTRUCT_EMBED_LEN_MASK` results and have values in different locations?

I did via `ObjectSpace.dump`:

```ruby
require "objspace"
require "rbconfig/sizeof"

STRUCT_MAX_EMBEDDED_MEMBERS = (
  GC::INTERNAL_CONSTANTS[:RVARGC_MAX_ALLOCATE_SIZE] -
  GC::INTERNAL_CONSTANTS[:RBASIC_SIZE] -
  GC::INTERNAL_CONSTANTS[:RVALUE_OVERHEAD]
) / RbConfig::SIZEOF["void*"]

LargeStruct = Struct.new(:foo, :bar, *(STRUCT_MAX_EMBEDDED_MEMBERS - 2).times.map { :"m_#{it}" })

def foo(obj)
  obj.foo
  obj.bar
end

embedded_struct = LargeStruct.new(123)
# Bump RCLASS_MAX_IV_COUNT for LargeStruct
embedded_struct.instance_variable_set(:@test, 1)

# Next allocation reserves space for the imemo/fields reference.
heap_struct = LargeStruct.new(123)

puts "embedded_struct: #{ObjectSpace.dump(embedded_struct)}"
puts "heap_struct: #{ObjectSpace.dump(heap_struct)}"

foo(heap_struct)
foo(embedded_struct)
```

```
embedded_struct: {"address":"0x103797800", "type":"STRUCT", "shape_id":10, "slot_size":640, "class":"0x1037d0f10", "references":["0x10386eeb8"], "memsize":640, "flags":{"wb_protected":true, "age":0}}
heap_struct: {"address":"0x10386ee68", "type":"STRUCT", "shape_id":0, "slot_size":40, "class":"0x1037d0f10", "memsize":664, "flags":{"wb_protected":true, "age":0}}
```

test/ruby/test_yjit.rb
```diff
@@ -644,6 +644,33 @@ def foo(obj)
...
+
+      embedded_struct = LargeStruct.new(123)
+      # Bump RCLASS_MAX_IV_COUNT for LargeStruct
+      embedded_struct.instance_variable_set(:@test, 1)
+
+      # Next allocation reserves space for the imemo/fields reference.
+      heap_struct = LargeStruct.new(123)
+
+      foo(heap_struct)
+      foo(embedded_struct)
```
eregon: Maybe check the result of `obj.foo` and `obj.bar`, if it's wrong it might just read other memory

eregon:
I haven't touched that YJIT code, but I did change the ZJIT code for Struct setting and it seems to potentially have the same issue:
https://github.com/ruby/ruby/blob/ce390f19954edda3874ed62768440953b40be0da/zjit/src/hir.rs#L2901-L2902
https://github.com/ruby/ruby/blob/ce390f19954edda3874ed62768440953b40be0da/zjit/src/hir.rs#L2926-L2933

That is, it checks if embedded and assume it will hold without a guard for it.

eregon:
CI is failing like
https://github.com/ruby/ruby/actions/runs/20896252445/job/60035241402?pr=15849
```
    1) Failure:
  TestYJIT#test_spilled_struct_aref [/home/runner/work/ruby/ruby/src/test/ruby/test_yjit.rb:654]:
  Expected no exits, but got:
  {putspecialobject: 1}
  Reasons:
  method call fallback reasons: 
      (all relevant counters are zero)
  invokeblock fallback reasons: 
      (all relevant counters are zero)
  invokesuper fallback reasons: 
      (all relevant counters are zero)
  method call exit reasons: 
      (all relevant counters are zero)
  invokeblock exit reasons: 
      (all relevant counters are zero)
  invokesuper exit reasons: 
      (all relevant counters are zero)
  getblockparamproxy exit reasons: 
      (all relevant counters are zero)
  getinstancevariable exit reasons:
      (all relevant counters are zero)
  setinstancevariable exit reasons:
      (all relevant counters are zero)
  leave exit reasons:
      interp_return:         93 (100.0%)
  left shift (opt_ltlt) exit reasons: 
      (all relevant counters are zero)
  right shift (>>) exit reasons: 
      (all relevant counters are zero)
  invalidation reasons: 
      constant_ic_fill:          2 (100.0%)
```
so it seems to side exit in that case

byroot:
Oh! is there a guard on `shape_id`? by any chance? Because the `shape_id` does include the `slot_size` so that would explain why it's not crashing.

byroot:
@k0kubun I can now make this fail (but not in debug mode):

```
$ make -j test/ruby/test_yjit.rb TESTOPTS="-n /test_spilled_struct_aref/"

[1/1] TestYJIT#test_spilled_struct_aref = 0.02 s
  1) Failure:
TestYJIT#test_spilled_struct_aref [/Users/byroot/src/github.com/byroot/ruby/test/ruby/test_yjit.rb:654]:
exited with status 256, stderr:
-e:7:in 'Object#foo': Expected 1, got: -2.0000000000000067 (RuntimeError)
	from -e:23:in 'block in <main>'
	from -e:30:in '<main>'


Finished tests in 0.026190s, 38.1825 tests/s, 114.5475 assertions/s.
1 tests, 3 assertions, 1 failures, 0 errors, 0 skips
```

So I think there's indeed a bug in YJIT (unlikely to be hit, but still).

k0kubun:
Thanks, filed https://github.com/Shopify/ruby/issues/977. Could you add an omit and merge the failing test?

PR #15865, opened 2026-01-13
Fiber scheduler: add socket hooks

noteflakes:
https://bugs.ruby-lang.org/issues/21837

This PR adds optional Fiber scheduler hooks for socket operations so schedulers can submit socket I/O directly, e.g. through io_uring, instead of only waiting for readiness via the generic `io_wait` hook.

Hooks added:

- `#socket_recv(socket, buffer, length, flags, from = nil)` - invoked by `Socket#recv` and `Socket#recvfrom`. `buffer` is an `IO::Buffer`; `from`, when non-nil, is a mutable `String` the scheduler fills with the packed sockaddr of the sender.
- `#socket_send(socket, buffer, length, flags, destination = nil)` - invoked by `Socket#send`, `BasicSocket#send`, and `UDPSocket#send`. `buffer` is a read-only `IO::Buffer`; `destination`, when non-nil, is a packed sockaddr string.
- `#socket_connect(socket, address)` - invoked by `Socket#connect`. `address` is a packed sockaddr string.
- `#socket_accept(socket, address)` - invoked by `Socket#accept` and `TCPServer#accept`. `address` is a mutable `String` the scheduler fills with the packed sockaddr of the accepted peer. Returns the accepted file descriptor as an integer, or `-errno`.
- `#socket_shutdown(socket, how)` - invoked by `BasicSocket#shutdown`. `how` is an integer (`SHUT_RD=0`, `SHUT_WR=1`, `SHUT_RDWR=2`).

All hooks are optional: if the scheduler does not respond to a hook, the corresponding blocking system call is made directly. All hooks are expected to return the result as an integer (bytes transferred, file descriptor, or `0` on success) or `-errno` on error.

Packed sockaddr strings are raw `struct sockaddr` byte blobs, as produced and consumed by `Socket.pack_sockaddr_in` / `Addrinfo#to_s` etc.

Related split-out prerequisite: #17473 adds the IO::Buffer callback helpers used to safely expose `String` storage to scheduler hooks.

cc @ioquatix


ioquatix:
Exciting!

WDYT: https://man7.org/linux/man-pages/man3/io_uring_prep_shutdown.3.html

samuel-williams-shopify:
I also have another idea, if `socket.accept` is passed a block, can it accept multiple times? We can schedule it into the io_uring once and repeatedly fire the block which may be a little more efficient.

noteflakes:
> WDYT: https://man7.org/linux/man-pages/man3/io_uring_prep_shutdown.3.html

UringMachine does have a `#shutdown` method. I can add a `#socket_shutdown` hook.

FWIW, the current implementation of `BasicSocket#shutdown` doesn't even release the GVL while calling `shutdown(2)`: https://github.com/ruby/ruby/blob/875dce2f5989c67b93e420d08939a74aa11d6199/ext/socket/basicsocket.c#L89-L109

noteflakes:
> I also have another idea, if `socket.accept` is passed a block, can it accept multiple times? We can schedule it into the io_uring once and repeatedly fire the block which may be a little more efficient.

There's already a [`Socket.accept_loop`](https://docs.ruby-lang.org/en/master/Socket.html#method-c-accept_loop) class method that does this, but implemented in pure Ruby and it supports accepting on multiple sockets. UringMachine has an `#accept_each` method that does a multishot accept, so if there was support for this in the fiber scheduler interface it would be trivial to implement.

My concern though is that this is an API change. Maybe it would be better to introduce a `Socket#accept_loop` method, and potentially a separate fiber scheduler hook? Maybe this should be discussed on bugs.ruby-lang.org?

noteflakes:
@ioquatix any idea on what's going on with this error? I can't seem to reproduce this on my machine.

https://github.com/ruby/ruby/actions/runs/20975148213/job/60288017307?pr=15865

```
     1) Error:
  TestFiberScheduler#test_iseq_compile_under_gc_stress_bug_21180:
  Test::Unit::ProxyError: end of file reached
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:87:in 'IO#read_nonblock'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:87:in 'block in Scheduler#run_once'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:82:in 'Array#each'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:82:in 'Scheduler#run_once'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:138:in 'block in Scheduler#run'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:136:in 'Thread.handle_interrupt'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:136:in 'Scheduler#run'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:168:in 'Scheduler#close'
      /Users/runner/work/ruby/ruby/src/test/fiber/scheduler.rb:147:in 'Scheduler#scheduler_close'
```

ioquatix (approved):
LGTM.

samuel-williams-shopify:
A few follow up notes:

- `IO::Buffer` read / write returning `EAGAIN` or blocking - behaves differently depending on running in a scheduler or not. Maybe this needs to be specified more explicitly.
- `IO::Buffer` lock count / locked_p - we may need a count.
- `IO::Buffer` send/recv - we may want to add this interface.

PR #15991, opened 2026-01-28
ZJIT: Emit run-time check for write barrier

### LINKED ISSUE #894: ZJIT: Emit run-time check for write barrier
Right now if we don't know statically that an object being written to the frame/an object is an immediate, we always emit a write barrier. Instead, we should emit a run-time check testing if it is false or an immediate and only conditionally fire the write barrier.

A reasonable first PR would guard if VM_ENV_FLAG_WB_REQUIRED is set before a setlocal and side-exit if so (thereby completely omitting the write barrier in the majority of cases)

jacob-shops:
This PR moves logic out of `Insn::SetLocal` and into `YARVINSN_setlocal`, `YARVINSN_setlocal_WC_0`, and `YARVINSN_setlocal_WC_1` using @39bytes's new `GetEP` and `GuardNoBitsSet` instructions. The overall effect of this change is to replace setlocal side exits with guards.

Co-authored with @jhawthorn

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -19,7 +19,7 @@ use crate::profile::{TypeDistributionSummary, ProfiledType};
 use crate::stats::Counter;
 use SendFallbackReason::*;
 
-mod tests;
+pub(crate) mod tests;
 mod opt_tests;
 
 /// An index of an [`Insn`] in a [`Function`]. This is a popular
@@ -656,6 +656,9 @@ pub enum SendFallbackReason {
     SingletonClassSeen,
     /// The super call is passed a block that the optimizer does not support.
     SuperCallWithBlock,
+    /// When the `super` is in a block, finding the running CME for guarding requires a loop. Not
+    /// supported for now.
+    SuperFromBlock,
     /// The profiled super class cannot be found.
     SuperClassNotFound,
     /// The `super` call uses a complex argument pattern that the optimizer does not support.
@@ -708,6 +711,7 @@ impl Display for SendFallbackReason {
             ComplexArgPass => write!(f, "Complex argument passing"),
             UnexpectedKeywordArgs => write!(f, "Unexpected Keyword Args"),
             SingletonClassSeen => write!(f, "Singleton class previously created for receiver class"),
+            SuperFromBlock => write!(f, "super: call from within a block"),
             SuperCallWithBlock => write!(f, "super: call made with a block"),
             SuperClassNotFound => write!(f, "super: profiled class cannot be found"),
             SuperComplexArgsPass => write!(f, "super: complex argument passing to `super` call"),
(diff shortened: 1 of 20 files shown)


zjit/src/hir.rs
```diff
@@ -6722,6 +6723,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     }
                 }
                 YARVINSN_setlocal_WC_0 => {
+                    // TODO(Jacob): Modify this to use guards
```
tekknolagi: only the actual SetLocal case needs guards

tekknolagi:
can you sink the GetEP into the branches it's used

zjit/src/hir.rs
```diff
@@ -6804,7 +6810,11 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                 }
                 YARVINSN_setlocal_WC_1 => {
                     let ep_offset = get_arg(pc, 0).as_u32();
-                    fun.push_insn(block, Insn::SetLocal { val: state.stack_pop()?, ep_offset, level: 1 });
+                    const level: u32 = 1;
```
jacob-shops, line 6729: Should we condition on `ep_escaped` like in `YARVINSN_setlocal_WC_0`?

This idea came from an in person discussion with @jhawthorn.

tekknolagi:
you should rebase to get ahead of the prism fixes

tekknolagi:
(edit I was wrong and thought https://github.com/ruby/ruby/pull/16052 had already been merged)

tekknolagi:
These are some weird failures

jacob-shops:
> These are some weird failures

Yeah, I'm not entirely sure what is going on. I was hoping it was a prism thing. I'll debug this more next week!

tekknolagi:
I recommend pairing with @jhawthorn if he is available to try running this under wbcheck. Perhaps we are over-eager.

k0kubun:
IIUC, the previous implementation would handle it in the C function and never side-exit on setlocal, but this version could. Unless we assume this side exit is effectively never taken, can we avoid introducing another side-exit reason, or at least recompile it when the side exit is hit?

tekknolagi:
YJIT side exits if WB needed

tekknolagi:
although i guess it does chain guard, so is that what you mean by recompile?

k0kubun:
Exactly, so it doesn't. In YJIT, `asm.jmp(Target::side_exit(...))` is a hard side-exit, but `jit_chain_guard` usually means that there's a final chain that doesn't side-exit.

tekknolagi:
Can you try rebasing this? I patched a couple bugs around local variables yesterday and I wonder if this will still fail

tekknolagi:
Closing for now since this has fallen pretty far behind. If you take this up again, feel free to re-open or make a new PR

PR #16084, opened 2026-02-06
[Bug #14582] Fix unable to use `method-entry` and `method-return` tracing probes since 2.5

jacob-carlborg:
Also  emit DTrace USDT probes for tail calls.

### DIFF
--- a/test/dtrace/test_function_entry.rb
+++ b/test/dtrace/test_function_entry.rb
@@ -18,7 +18,28 @@ ruby$target:::method-entry
 	}
 
 	assert_equal 10, foo_calls.length, probes
-	line = '3'
+	line = '6'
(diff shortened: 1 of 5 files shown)


test/dtrace/test_function_entry.rb
jacob-carlborg, line 173: BTW, is anyone running these tests? Because it fails for me because `__enable` doesn't exist.

jacob-carlborg:
@tenderlove could you please take a look at this DTrace related change? If I understand correctly you added the initial support for DTrace USDT probes to Ruby.

jacob-carlborg:
@ko1 could you please take a look at this pull request which improves the DTrace probes?

PR #16129, opened 2026-02-10
ZJIT: Only reload written locals after send-with-block

### LINKED ISSUE #774: ZJIT: Avoid reloading locals on send that are not referenced by blockiseq
Send-ish instructions with non-null `blockiseq` reloads every local variable at the moment, but we shouldn't need to reload locals that are not referenced by the `blockiseq`.

tekknolagi:
Track recursive writes to parent frames. Also a whole bunch of other stuff this uncovered along the way. This PR is best reviewed commit-by-commit.


tekknolagi:
OH. We may need to also investigate `compile_entry_state` which uses the old `iseq_escapes_ep` check and also needs to check if the iseq has historically escaped EP.

tekknolagi:
Obviated by https://github.com/ruby/ruby/pull/17417

PR #16154, opened 2026-02-12
Backport GH-15908

hsbt:
from https://github.com/ruby/ruby/pull/15908

We should fix to suppress logic for `benchmark-ips`.


hsbt:
I will also backport https://github.com/ruby/ruby/pull/16157 for build failure of cygwin.

k0kubun (approved):

PR #16183, opened 2026-02-17
Stack-allocate small re_registers

jhawthorn:
In rb_reg_search_set_match we would previously always xmalloc two arrays for the beg and end registers. Very often these allocations are few, and for that case we can use stack-allocated memory instead.

After, if the match is successful, we need to copy the registers onto a MatchData. If the existing MatchData fits we memcpy the new match over, otherwise we free the existing MatchData's pointers and replace them with our new ones.

```
❯ make benchmark ITEM=vm_regexp
/home/jhawthorn/.rubies/ruby-4.0.0/bin/ruby --disable=gems -rrubygems -I./benchmark/lib ./benchmark/benchmark-driver/exe/benchmark-driver \
            --executables="compare-ruby::/home/jhawthorn/.rubies/ruby-4.0.0/bin/ruby --disable=gems -I.ext/common --disable-gem" \
            --output=markdown --output-compare -v $(find ./benchmark -maxdepth 1 -name 'vm_regexp' -o -name '*vm_regexp*.yml' -o -name '*vm_regexp*.rb' | sort)
compare-ruby: ruby 4.0.0 (2025-12-25 revision 553f1675f3) +PRISM [x86_64-linux]
built-ruby: ruby 4.1.0dev (2026-02-17T06:09:22Z reg_search_stack_a.. 439c4496cc) +PRISM [x86_64-linux]
# Iteration per second (i/s)

|                  |compare-ruby|built-ruby|
|:-----------------|-----------:|---------:|
|vm_regexp         |     11.154M|   14.095M|
|                  |           -|     1.26x|
|vm_regexp_invert  |     11.144M|   14.301M|
|                  |           -|     1.28x|
```

I implemented this aiming to improve the speed of a regex `gsub` which had no match:

```
Benchmark 1: /home/jhawthorn/.rubies/ruby-base/bin/ruby --disable-gems --enable-frozen-string-literal -e 'str = "foo"; 20_000_000.times { str.gsub(/x/, "") }'
  Time (mean ± σ):      2.027 s ±  0.022 s    [User: 2.023 s, System: 0.004 s]
  Range (min … max):    1.999 s …  2.073 s    10 runs

Benchmark 2: /home/jhawthorn/.rubies/ruby-test/bin/ruby --disable-gems --enable-frozen-string-literal -e 'str = "foo"; 20_000_000.times { str.gsub(/x/, "") }'
  Time (mean ± σ):      1.139 s ±  0.010 s    [User: 1.134 s, System: 0.004 s]
  Range (min … max):    1.131 s …  1.158 s    10 runs

Summary
  /home/jhawthorn/.rubies/ruby-test/bin/ruby --disable-gems --enable-frozen-string-literal -e 'str = "foo"; 20_000_000.times { str.gsub(/x/, "") }' ran
    1.78 ± 0.03 times faster than /home/jhawthorn/.rubies/ruby-base/bin/ruby --disable-gems --enable-frozen-string-literal -e 'str = "foo"; 20_000_000.times { str.gsub(/x/, "") }'
```

I would like to (eventually) allocate the `struct re_registers` as fully embedded within the `T_MATCH` itself, but that requires more reorganization (and figuring out how `struct re_registers` is used directly in gems other than strscan).

### DIFF
--- a/re.c
+++ b/re.c
@@ -1804,7 +1804,14 @@ rb_reg_search_set_match(VALUE re, VALUE str, long pos, int reverse, int set_back
         .pos = pos,
         .range = reverse ? 0 : len,
     };
-    struct re_registers regs = {0};
+
+    OnigPosition stack_beg[ONIG_NREGION], stack_end[ONIG_NREGION];
+    struct re_registers regs = {
+        .allocated = 0, // avoids realloc/free in onig_region_resize/free
+        .num_regs = ONIG_NREGION,
+        .beg = stack_beg,
+        .end = stack_end,
+    };
 
     OnigPosition result = rb_reg_onig_match(re, str, reg_onig_search, &args, &regs);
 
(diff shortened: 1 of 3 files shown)


nobu (approved):
Probably, we should send the PR to onigmo too?

jhawthorn:
> Probably, we should send the PR to onigmo too?

Probably. I'm going to think a little more about this because I want to avoid these two xmalloc always (find a way to store the re_regs directly on the T_MATCH)

jhawthorn:
Closed in favour of #16906

PR #16189, opened 2026-02-17
Move win32-registry from default to bundled gems

larskanis:
Because win32-registry needs fiddle and fiddle is a bundled gem.

[Bug #21855](https://bugs.ruby-lang.org/issues/21855)


CookiePLMonster:
Can we expect to see this backported to 3.4 at some point?

PR #16255, opened 2026-02-26
win32/configure.bat accept empty prefix

YO4:
Currently, running `win32/configure.bat --prefix=` outputs `prefix = \=/` in the Makefile.
Fix this to prevent invalid values from being output.

Other similar processes have not been changed, as specifying an empty string would be meaningless.

### DIFF
--- a/win32/configure.bat
+++ b/win32/configure.bat
@@ -133,7 +133,8 @@ goto :unknown_opt
 goto :loopend ;
 :dir
   if "%eq%" == "" call :take_arg
-  echo>> %config_make% %opt:~2% = %arg:\=/%
+  if defined arg set "arg=%arg:\=/%"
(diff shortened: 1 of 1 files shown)


nobu:
Good catch, and`:path` may have a similar issue?

nobu (approved):

YO4:
`nmake install DESTDIR=./target` installs to ./target/bin/ruby.exe so `win32/configure.bat --prefix=` is useful.
I don't know that `win32/configure.bat --path=` is meaningful or not, so I left it.

nobu:
Doesn’t  `--prefix=.` work instead?

YO4:
Wow! That seemed to be enough for me.
Thanks!

PR #16269, opened 2026-02-27
ZJIT: Remove GetLEP

tekknolagi:
Use get_lvar_level and GetEP instead.


k0kubun (approved):
nice

tekknolagi:
How do I get this to generate a `GetEP k` with `k > 0`?

tekknolagi:
Will do a follow up adding a test if I figure it out

nirvdrum:
What motivated this change? I liked the clarity of `GetLEP`.

tekknolagi:
It's not semantically different enough to require a separate opcode IMO. Even GetEP could just be a series of LoadField in the limit

PR #16298, opened 2026-03-04
ZJIT: Avoid SendDirect when callee ISeq cannot compile

nirvdrum:
While using `zjit_bisect` on the _lobsters_ benchmark in [ruby-bench](https://github.com/ruby/ruby-bench), trying to track down the segfault fixed in #16251, I ended up in an odd situation where the bisect encountered an application exception and halted rather than crash with the segfault I expected. Running ZJIT with `--zjit-allowed-iseqs` with this specific list of methods did indeed result in an application exception and any modification I made to that list made the exception go away; it had to be this specific set of three methods to error out.

I spent a couple of days debugging this and narrowed it down to a confluence of events. I think at some level it has to do with a constructor that has optional positional arguments followed by kwargs and the way we jump into the method. On top of that, we have a `SendDirect` to an ISEQ that cannot be compiled, and I suspect something isn't being cleaned up correctly when we side-exit. Currently, "cannot be compiled" means it's not in the `--zjit-allowed-iseqs` list, so I think this only impacts that narrow use case.

I really wanted to make sure this wasn't something that could occur during typical side-exits in ZJIT. I probed this from many different angles and couldn't induce the behavior outside of `--zjit-allow-iseqs`. But, I also can't conclusively say what's leading to the stack corruption. I could spend more time chasing that down, but I'm not sure it's all that valuable. We shouldn't try to `SendDirect` to an ISEQ we know _a priori_ that we cannot compile.

The test case here was drawn directly from the _lobsters_ code, but eliminates any usage of Rails. We can rename methods if folks would like. I confirmed this test fails with _master_ and passes with this PR applied.

### DIFF
--- a/test/ruby/test_zjit.rb
+++ b/test/ruby/test_zjit.rb
@@ -164,6 +164,50 @@ class TestZJIT < Test::Unit::TestCase
     }, call_threshold: 2, allowed_iseqs: 'entry@-e:2'
   end
 
+  def test_send_direct_disallowed_callee_uses_fallback
+    script = <<~RUBY
+      def fast_string_to_time(string)
+        return unless string.include?("-")
+        if is_utc?
+          target(string, in: "UTC")
+        else
+          target(string)
+        end
+      end
+
+      def is_utc?
+        default_timezone == :utc
+      end
+
+      def default_timezone
+        :utc
+      end
+
+      def target(year = nil, mon = nil, mday = nil, hour = nil, min = nil, sec = nil, zone = nil, in: nil, precision: 9)
+        [year.class, binding.local_variable_get(:in), precision]
+      end
+
+      20.times { fast_string_to_time("2023-05-08 10:21:10.510629") }
+      puts :ok
+    RUBY
+
+    out, err, status = eval_with_jit(
+      script,
+      call_threshold: 3,
+      allowed_iseqs: <<~LIST,
+        fast_string_to_time@-e:2
+        is_utc?@-e:11
+        default_timezone@-e:15
+      LIST
+      extra_args: ['--zjit-dump-hir']
+    )
+
+    assert_success(out, err, status)
+    assert_includes(out, "ok\n")
+    assert_includes(out, "fn fast_string_to_time@-e:2:")
+    refute_match(/SendDirect .*:target/, out)
+  end
+
   def test_opt_new_with_custom_allocator
     assert_compiles '"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"', %q{
       require "digest"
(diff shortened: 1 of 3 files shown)


XrXr (commented):
> We shouldn't try to SendDirect to an ISEQ we know a priori that we cannot compile.

This makes sense. I think we can make the refusal to direct send more general than in case the callee is filtered by `--zjit-allow-iseqs`. Since 231407c251d, we don't attempt to compile an ISeq that we've failed to compiled before. If we see the callee has `IseqStatus::CantCompile` we should also fallback.

XrXr:
Also question, this is not fixing a correctness issue, right? It's avoiding a slow path when we can. The exact purpose of the test is not clear to me.

tekknolagi:
No, Kevin noticed a correctness issue.

tekknolagi:
But I think we should make it so SendDirect to something that failed to compile is not breaking, but only slightly slower

tekknolagi:
reduced repro:

```ruby
def target(zone = nil, inx: nil)
  [inx]
end

def fast_string_to_time
  target(inx: "u-cant-see-me")
end

20.times { fast_string_to_time }
result = fast_string_to_time
puts :ok, result.inspect
```

jitlist.txt

```
fast_string_to_time@tmp/broken.rb:6
```

`../build-dev/miniruby --zjit-call-threshold=3 --zjit-allowed-iseqs=jitlist.txt --zjit-dump-hir tmp/broken.rb`

tekknolagi:
Current guess: we are passing the kw bits wrong so the interpreter doesn't see our kwarg

tekknolagi:
Oh, maybe not. There is no `checkkeyword` in `target`.

k0kubun:
> But, I also can't conclusively say what's leading to the stack corruption. I could spend more time chasing that down, but I'm not sure it's all that valuable. We shouldn't try to SendDirect to an ISEQ we know a priori that we cannot compile.

`!can_compile_iseq`/`--zjit-allowed-iseqs` is just one of the failure modes of `function_stub_hit`. Even if `can_compile_iseq` is true, `SendDirect` (`function_stub_hit`) could still decide to not compile an ISEQ due to OOM. So, while this removes the failure in the `!can_compile_iseq` case, I do think it's valuable to investigate why it fails and fix it directly.

tekknolagi (changes requested):
Sounds like we found the underlying bug and can fix that instead

nirvdrum:
I think there's still some value in not using `SendDirect` when we know we can't compile. In particular, I like the `FilteredOut` fallback reason. But, yes, now @XrXr has tracked down the actual problem we should defer this PR until that fix lands.

We should also add a test for the OOM case. But, I think that better suited for the other PR.

tekknolagi:
Closing for now

PR #16303, opened 2026-03-04
Mark `$_` as box-dynamic to bypass Box gvar_tbl cache

dak2:
`$_` is updated through svar (rb_lastline_set), bypassing rb_gvar_set, so the Box gvar_tbl cache is never invalidated                                              
and returns a stale value. Call rb_gvar_box_dynamic so gvar_use_box_tbl() skips the cache.

Fixes [Bug #21940](https://bugs.ruby-lang.org/issues/21940)

### DIFF
--- a/test/ruby/test_box.rb
+++ b/test/ruby/test_box.rb
@@ -583,6 +583,49 @@ class TestBox < Test::Unit::TestCase
     end;
   end
 
+  def test_lastline_not_cached_in_box
+    assert_separately([ENV_ENABLE_BOX], __FILE__, __LINE__, "#{<<~"begin;"}\n#{<<~'end;'}", ignore_stderr: true)
+    begin;
+      r, w = IO.pipe
+      w.write("first\nsecond\n")
+      w.close
+      STDIN.reopen(r)
+      via_gets = Ruby::Box.new.eval(<<~'CODE')
+        gets
+        _ = $_
+        gets
+        $_
+      CODE
+      assert_equal "second\n", via_gets
+    end;
+  end
+
+  def test_lastline_not_cached_in_nested_boxes
+    assert_separately([ENV_ENABLE_BOX], __FILE__, __LINE__, "#{<<~"begin;"}\n#{<<~'end;'}", ignore_stderr: true)
+    begin;
+      r, w = IO.pipe
+      w.write("outer1\ninner1\ninner2\nouter2\n")
+      w.close
+      STDIN.reopen(r)
+      inner_via_gets, outer_via_gets = Ruby::Box.new.eval(<<~'CODE')
+        gets
+        _ = $_
+
+        inner_result = Ruby::Box.new.eval(<<~'INNER')
+          gets
+          _ = $_
+          gets
+          $_
+        INNER
+
+        gets
+        [inner_result, $_]
+      CODE
+      assert_equal "inner2\n", inner_via_gets
+      assert_equal "outer2\n", outer_via_gets
+    end;
+  end
+
   def test_load_path_and_loaded_features
     setup_box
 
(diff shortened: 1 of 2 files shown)


dak2:
@tagomoris 

As we discussed in person last week, I’ve committed the fixes for `$_` and `$~` as a single issue in this PR.
Could you please review it?

dak2:
@tagomoris

I've also made sure that `$_` isn't cached using `rb_gvar_box_dynamic`, so could you please review this?

https://github.com/ruby/ruby/pull/16644

test/ruby/test_box.rb
```diff
@@ -583,6 +583,40 @@ def test_match_variables_are_not_cached_in_box
...
   end
 
+  def test_lastline_not_cached_in_box
+    assert_separately([ENV_ENABLE_BOX], __FILE__, __LINE__, "#{<<~"begin;"}\n#{<<~'end;'}", ignore_stderr: true)
+    begin;
+      box = Ruby::Box.new
+      box.eval(<<~'CODE')
+        $_ = "first"
+        _ = $_
+        $_ = "second"
```
tagomoris: `$_` is a gvar that is updated when Kernel.#gets or Kernel.#readline read input from STDIN.
So the test case need to check (with input characters from STDIN)
1. call Kernel.#gets and the read data (A) is equal to `$_`
2. call Kernel.#gets again and the read data (B: different from A) is equal to `$_`

test/ruby/test_box.rb
```diff
@@ -583,6 +583,40 @@ def test_match_variables_are_not_cached_in_box
...
+        raise "expected 'second' but got '#{$_}'" unless $_ == "second"
```
tagomoris: Is there any reason to not use `assert_equal`?

test/ruby/test_box.rb
```diff
@@ -583,6 +583,40 @@ def test_match_variables_are_not_cached_in_box
...
+      CODE
+    end;
+  end
+
+  def test_lastline_not_cached_in_multiple_boxes
```
tagomoris: The test requirements are similar to the above one: need to check whether `Kernel.#gets` (or readline) updates `$_` properly.

tagomoris (changes requested):

test/ruby/test_box.rb
dak2: Thanks for pointing that. I addressed it.

https://github.com/ruby/ruby/pull/16303/changes#diff-230b825f93af3aee0e62073aaca69ee5b48c6f00accd81b56c58e4a272e3badcR586-R601

test/ruby/test_box.rb
dak2: There wasn't any particular reason. So, I decided to use `assert_equal`.

https://github.com/ruby/ruby/compare/07ada1a31e4c07f473ff6bc90bd387eb94530f54..5e3c97cfa3e47b38aa09f9c3a45d589c5a73aa3e#diff-230b825f93af3aee0e62073aaca69ee5b48c6f00accd81b56c58e4a272e3badc

test/ruby/test_box.rb
dak2: I handled this in the same way.

https://github.com/ruby/ruby/pull/16303/changes#diff-230b825f93af3aee0e62073aaca69ee5b48c6f00accd81b56c58e4a272e3badcR586-R601

tagomoris:
@dak2 Could you rebase this branch on master? The master branch already has `$(top_srcdir)/prism/internal/strpbrk.h` lines in `depend`, which is now causing CI failure
https://github.com/ruby/ruby/actions/runs/25098017298/job/73539623025?pr=16303

dak2:
@tagomoris 

Thanks for noticing me. Rebased, all checks are green.

tagomoris:
Thank you for the fix!

PR #16322, opened 2026-03-06
Expose rb_int_parse_cstr() as public API

jinroq:
Move `rb_int_parse_cstr()` declaration and `RB_INT_PARSE_*` flags from internal/bignum.h to the public header include/ruby/internal/intern/bignum.h so that C extensions can use this function without writing their own prototype declarations.

The flags are changed from an enum to #define macros to follow the convention used in the public header for similar constants.

see: https://github.com/ruby/strscan/pull/192

### DIFF
--- a/include/ruby/internal/intern/bignum.h
+++ b/include/ruby/internal/intern/bignum.h
@@ -26,6 +26,10 @@
 # include <stddef.h>
 #endif
 
+#ifdef HAVE_SYS_TYPES_H
+# include <sys/types.h>         /* for ssize_t (note: on Windows ssize_t is */
+#endif                          /* `#define`d in ruby/config.h) */
+
 #include "ruby/internal/attr/nonnull.h"
 #include "ruby/internal/dllexport.h"
 #include "ruby/internal/value.h"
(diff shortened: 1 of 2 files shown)


jinroq:
Due to a change in the specification of `StringScanner#integer_at`, there is no longer a need to expose `rb_int_parse_cstr()`. This PR will be closed.

see: https://github.com/ruby/strscan/pull/193#issue-4032942463

nobu:
> see: [ruby/strscan#193 (comment)](https://github.com/ruby/strscan/pull/193#issue-4032942463)

I can't see that comment. What is that change?

And `rb_str_to_inum(rb_str_new(ptr, len), base, 0)` will be less efficient than `rb_int_parse_cstr` because it creates an intermediate string object.

jinroq:
@nobu 

> I can't see that comment. What is that change?

I'm sorry if it's hard to understand.

> The specification for MatchData#integer_at has been defined [here](https://bugs.ruby-lang.org/issues/21932#note-7). StringScanner#integer_at follows this specification.

I just wanted to convey that it follows the same specifications as `MatchData#integer_at`.

> And rb_str_to_inum(rb_str_new(ptr, len), base, 0) will be less efficient than rb_int_parse_cstr because it creates an intermediate string object.

`StringScanner#integer_at` implements fast pass. `rb_str_to_inum(rb_str_new(ptr, len), base, 0)` will be called if the input does not pass fast pass.

jinroq:
@nobu 
BTW, should we reopen this PR?

PR #16334, opened 2026-03-07
[Tests] Add new test case for string with capacity

aguspe:
I added a new test case to cover string with capacity: with non-empty source strings,encoding override, negative/zero
capacity, capacity smaller than source length, and frozen source mutability.


aguspe:
Closing this one myself to avoid piling up too many open PRs at once and saturating the maintainers' review queue. Happy to reopen or rework it later if there is interest, thank you!

PR #16349, opened 2026-03-09
Batch cleanup geniv before sweep

jhawthorn:
Previously we needed to check every object to determine whether or not it might have an entry in the ~id2ref table or~ geniv table. Both of these should now be uncommon.

This commit adds a function, rb_gc_vm_weak_table_essential_p, which allows gc.c to decide which tables to always handle in batch.

MMTK will continue to call rb_gc_vm_weak_table_foreach on all tables. This only changes the default GC to sweep the "essential" tables first, but continue to call rb_gc_obj_free_vm_weak_references on other objects.

This makes the ~id2ref and~ geniv table "essential", as they're usually small and need to be tested on every type of object. It leaves the fstring table, symbol table, callinfo table, and overloaded CME table as-is because those tables are (mostly) larger and are only needed by specific object types.

---

A longer-term goal of this is to be able to allocate objects from a special pages tagged with the fact they will never hit the `rb_gc_obj_needs_cleanup_p == true` case, but I believe there's enough benefit doing this on its own.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1411,11 +1411,6 @@ rb_gc_obj_needs_cleanup_p(VALUE obj)
         return rb_gc_imemo_needs_cleanup_p(obj);
     }
 
-    shape_id_t shape_id = RBASIC_SHAPE_ID(obj);
-    if (rb_shape_has_fields(shape_id) && rb_shape_layout(shape_id) == SHAPE_ID_LAYOUT_OTHER) {
-        return true;
-    }
-
     switch (flags & RUBY_T_MASK) {
       case T_FLOAT:
       case T_RATIONAL:
@@ -2161,10 +2156,6 @@ void
 rb_gc_obj_free_vm_weak_references(VALUE obj)
 {
     ASSUME(!RB_SPECIAL_CONST_P(obj));
-
-    if (rb_obj_gen_fields_p(obj)) {
-        rb_free_generic_ivar(obj);
-    }
 
     switch (BUILTIN_TYPE(obj)) {
       case T_STRING:
@@ -4048,6 +4039,20 @@ vm_weak_table_frozen_strings_foreach(VALUE *str, void *data)
 }
 
 void rb_fstring_foreach_with_replace(int (*callback)(VALUE *str, void *data), void *data);
+
+// Whether this table must be cleaned every GC after marking.
+// Other tables may be skipped cleaned up per-object via rb_gc_obj_free_vm_weak_references.
+bool
+rb_gc_vm_weak_table_essential_p(enum rb_gc_vm_weak_tables table)
+{
+    switch (table) {
+      case RB_GC_VM_GENERIC_FIELDS_TABLE:
+        return true;
+      default:
+        return false;
+    }
+}
+
 void
 rb_gc_vm_weak_table_foreach(vm_table_foreach_callback_func callback,
                             vm_table_update_callback_func update_callback,
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -4232,6 +4232,15 @@ gc_sweep_freeobj_hooks(rb_objspace_t *objspace)
     }
 }
 
+static int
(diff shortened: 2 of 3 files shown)


gc.c
```diff
@@ -1329,29 +1326,28 @@ rb_gc_obj_needs_cleanup_p(VALUE obj)
...
+        return false;
 
       case T_BIGNUM:
         if (!(flags & BIGNUM_EMBED_FLAG)) return true;
-        return rb_shape_has_fields(shape_id);
+        return false;
 
       case T_STRUCT:
         if (!(flags & RSTRUCT_EMBED_LEN_MASK)) return true;
-        if (flags & RSTRUCT_GEN_FIELDS) return rb_shape_has_fields(shape_id);
```
jhawthorn: ~I'm not sure I understand why this was necessary before. I believe this indicates that we stored a reference to an imemo_fields object _inside_ the object, so I don't know why that would require cleanup.~

Ah! I see I had this backwards. `RSTRUCT_GEN_FIELDS` means it's in generic_fields_tbl, so it made sense before and makes sense to remove now 😊

luke-gru:
I guess the only case where this could be slower is if we have _lots_ of old objects with genivs. That seems unlikely to me 👍

byroot:
Deferring to @eightbitraptor on this one. I know I edited these a few times in the past but I'm not particularly knowledgeable on them.

peterzhu2118:
I'm not a huge fan of this PR because it means that the whole id2ref and generic fields tables gets scanned every GC. As @luke-gru said, it might be an issue in certain scenarios. That being said, I think that is a bit of a corner case and it's probably fine merging this PR.

> A longer-term goal of this is to be able to allocate objects from a special pages tagged with the fact they will never hit the rb_gc_obj_needs_cleanup_p == true case, but I believe there's enough benefit doing this on its own.

I think I would prefer something like that or an API to register when an object requires finalization (e.g. generic fields added, external malloc buffer allocated, file descriptor opened). That way, an object defaults to not having any finalization callbacks except when that API has been called.

jhawthorn:
> whole id2ref and generic fields tables gets scanned every GC

The assumption being made here is that tese tables are smaller (or non-existant) vs the number of objects we have to examine when freeing. In most cases this will be a win, other than perhaps folks who store ivars on T_STRING objects (T_DATA and T_STRUCT no longer use the geniv table).

As you know, MMTK iterates all tables here. I'm hoping to find a middleground where we iterate the tables which are cheap here, so that we don't have to during sweeping.

byroot:
> other than perhaps folks who store ivars on T_STRING objects 

So most Action View users? (granted with https://github.com/rails/rails/commit/db27b67bb9ae8194bdb38559cac194707086de87 most `AS::SafeBuffer` shouldn't have an ivar).

> (T_DATA and T_STRUCT no longer use the geniv table).

Well, this is only true for non-embedded `T_DATA` but this isn't not officially public API, that is probably most of them.

> The assumption being made here is that tese tables are smaller (or non-existant) 

I think the keyword here is assumption. Perhaps it would be worth instrumenting this on some workloads. Not sure if you still have the ability to ship modified builds in production, but that would be a good way to prove or disprove the assumption (if you still use sorbet the id2ref table is likely big).

peterzhu2118:
> As you know, MMTK iterates all tables here.

The difference is that MMTk processes these tables in parallel, which (ideally) means that it takes as long as scanning the largest table. Since it's sequential here, it would take the summation of the time to scan the tables.

> I'm hoping to find a middleground where we iterate the tables which are cheap here, so that we don't have to during sweeping.

I think your theory is solid, but are you able to write a microbenchmark where we can observe the benefits and tradeoffs?

eightbitraptor (approved):

jhawthorn:
We've talked this through a bit more offline.

The reason this is desirable isn't so much for an immediate performance gain, but because it allows us to know at object allocation time (rather than only when sweeping) whether cleanup is required for an object.

PR #16359, opened 2026-03-10
Improve performance of String#scrub by skipping ASCII runs with `search_nonascii`

FletcherDares:
## Background
I identified that `String#scrub` and `String#scrub!` were slower than expected when dealing with very large strings with an unknown encoding. We have a use case where we run scrub on HTML pages and the following code change resulted in ~20x performance improvement.
Before:
```ruby
string.force_encoding("UTF-8")
string.scrub!
```
After:
```ruby
string.force_encoding("UTF-8")
string.scrub! unless string.valid_encoding?
```
I was curious about what caused the performance difference between `String#scrub!` and `String#valid_encoding?` when the string is valid utf-8. I decided to dig into the internal implementation of String#scrub and found the opportunity for a performance improvement, see the following suggestion.
## Proposal:
```c
// string.c:11838 (Ruby 4.1.0)
p = search_nonascii(p, e);
if (!p) {
    p = e;
}
// Enter scrub loop.
while (p < e) {
    int ret = rb_enc_precise_mbclen(p, e, enc);
    if (MBCLEN_NEEDMORE_P(ret)) {
        break;
    }
    else if (MBCLEN_CHARFOUND_P(ret)) {
        cr = ENC_CODERANGE_VALID;
        p += MBCLEN_CHARFOUND_LEN(ret);
+       p = search_nonascii(p, e);
+       if (!p) {
+           p = e;
+           break;
+       }
    }
    // MBCLEN_INVALID_P handling unchanged.
```
This patch does not change any observable behaviour, all tests pass.
### Justification:
This adds the pattern used in `String#valid_encoding?` to `String#scrub` which allows us to use `search_nonascii` to skip ascii characters after encountering a multi-byte sequence. This is implemented for `String#valid_encoding?` in string.c:799 (Ruby 4.1.0). 
`search_nonascii` returns a pointer to the address of the next non-ascii character.

#### Upsides:
- Better performance (3-45x) when we have some ascii within the text that we are attempting to scrub. Testing with HTML examples showed 15x-40x speedups depending on the density of multibyte characters.
- A version of this is implemented right now where it saved 1% average processing time and 1.5% at P90 over many requests.
#### Tradeoffs:
- Worse performance (0.7x) if the text is entirely multibyte with no-ascii characters.

### Results:
| Benchmark                                      | Control    | Experiment | Speedup    |
| :--------------------------------------------- | ---------: | ---------: | :--------- |
| scrub_known_valid                              |    29.075M |    29.708M | 1.0x      |
| scrub_known_invalid                            |   247.926k |   740.033k | 2.9x      |
| scrub_unknown_but_valid_coderange              |   274.488k |   889.488k | 3.2x      |
| scrub_unknown_and_invalid_coderange            |   269.232k |   739.194k | 2.8x      |
| scrub_unknown_but_valid_coderange_worst_case   |   234.635k |   164.706k | 0.7x      |
| scrub_unknown_and_invalid_coderange_worst_case |   239.335k |   162.561k | 0.7x      |
| scrub_japanese_html                            |     187.47 |    2927.00 | 15.6x     |
| scrub_english_html                             |     510.22 |   22694.00 | 44.5x     |

#### Methodology:
- Apple M4 Pro macOS 26.3
- Used built in benchmark tooling
- non-HTML benchmark Strings were all ~1KB
- Benchmarked Japanese & English HTML pages pulled from real sites. (I didn't include them as they were large ~2MB)

### DIFF
--- a/benchmark/string_scrub.yml
+++ b/benchmark/string_scrub.yml
@@ -1,0 +1,48 @@
+prelude: |
+
+  STRING_SIZE = 1024
+  def duplicate_to_length(str, target_length)
+    return "" if target_length <= 0
+    return str[0, target_length] if str.length >= target_length
+
+    (str * ((target_length / str.length) + 1))[0, target_length]
+  end
+  base = "Hello \u{1f600} world! \u{00e9}\u{00f1}"
+  padding = duplicate_to_length(base, STRING_SIZE)
+
+  valid_utf8 = (padding.b + "OK".b).force_encoding("UTF-8")
+  valid_utf8.valid_encoding?
+  unknown_but_valid_utf8 = valid_utf8.dup.b.force_encoding("UTF-8")
+  invalid_utf8 = (padding.b + "\x80\xFF".b).force_encoding("UTF-8")
+  invalid_utf8.valid_encoding?
+  unknown_but_invalid_utf8 = (padding.b + "\x80\xFF".b).force_encoding("UTF-8")
+
+  worst_case_utf8 = duplicate_to_length("\u{1f600}\u{00e9}\u{00f1}", STRING_SIZE).b.force_encoding("UTF-8")
+
+  unknown_but_valid_utf8_worst_case = worst_case_utf8.dup.b.force_encoding("UTF-8")
+  unknown_but_invalid_utf8_worst_case = (worst_case_utf8.b + "\x80\xFF".b).force_encoding("UTF-8")
+
+benchmark:
+  scrub_known_valid: |
+    string = valid_utf8.dup
+    string.scrub!
+
+  scrub_known_invalid: |
+    string = invalid_utf8.dup
+    string.scrub!
+
+  scrub_unknown_but_valid_coderange: |
+    string = unknown_but_valid_utf8.dup
+    string.scrub!
+
+  scrub_unknown_and_invalid_coderange: |
+    string = unknown_but_invalid_utf8.dup
+    string.scrub!
+
+  scrub_unknown_but_valid_coderange_worst_case: |
+    string = unknown_but_valid_utf8_worst_case.dup
+    string.scrub!
+
+  scrub_unknown_and_invalid_coderange_worst_case: |
+    string = unknown_but_invalid_utf8_worst_case.dup
+    string.scrub!
--- a/string.c
+++ b/string.c
@@ -11843,10 +11843,17 @@ enc_str_scrub(rb_encoding *enc, VALUE str, VALUE repl, int cr)
             else if (MBCLEN_CHARFOUND_P(ret)) {
                 cr = ENC_CODERANGE_VALID;
                 p += MBCLEN_CHARFOUND_LEN(ret);
-                p = search_nonascii(p, e);
-                if (!p) {
-                    p = e;
-                    break;
+                /*
+                 * After a valid multibyte character, skip the following ASCII run.
+                 * If the next byte is already non-ASCII, search_nonascii would only
+                 * rediscover p after its word-at-a-time setup.
+                 */
+                if (p < e && ISASCII(*p)) {
+                    p = search_nonascii(p, e);
+                    if (!p) {
+                        p = e;
+                        break;
+                    }
                 }
             }
             else if (MBCLEN_INVALID_P(ret)) {
--- a/string.c
+++ b/string.c
@@ -11843,6 +11843,11 @@ enc_str_scrub(rb_encoding *enc, VALUE str, VALUE repl, int cr)
             else if (MBCLEN_CHARFOUND_P(ret)) {
                 cr = ENC_CODERANGE_VALID;
                 p += MBCLEN_CHARFOUND_LEN(ret);
+                p = search_nonascii(p, e);
+                if (!p) {
+                    p = e;
+                    break;
+                }
             }
             else if (MBCLEN_INVALID_P(ret)) {
                 /*


duerst:
I have looked at the request. I'm not exactly sure what is meant by "unknown"; it seems it stands for "unknown encoding, assuming UTF-8", which should be stated explicitly.
'Speedup' makes this look very impressive, but in terms of actual time gained or lost, the inverse is more helpful: If an operation takes e.g. 1s, the maximum that can be saved is 1s, but much more can be lost (e.g. 9s if the operation suddenly takes 10s).
It should be possible to additionally speed up the case of valid UTF-8 (strings or ranges) rather than just ascii if that's desirable. But that would be a bit more code.

FletcherDares:
Hi @duerst, thank you so much for the review! Samuel spoke very highly of your Unicode knowledge.

>  I'm not exactly sure what is meant by "unknown"

"unknown" does mean "unknown encoding assuming UTF-8". In the example I provided this is the purpose of `string.force_encoding("UTF-8")` we have a string that we are now saying should be UTF-8 but have not validated whether that is true yet. This is the core difference between the first two benchmarks and the rest, in these we have cached whether it valid or not. This is why the first benchmark's performance is the same across both implementations (1.0x) as we use this cache to early exit on this [line](https://github.com/ruby/ruby/blob/86049a98754507a480e657274b4d38b07c46e6ea/string.c#L11785).

>  the maximum that can be saved is 1s, but much more can be lost

My understanding of this comes from the following example, given a string `"abcdéfg"` it is faster to perform `String#valid_encoding?` than it is to perform `String#scrub`. It turns out that this is because in the `String#valid_encoding?` implementation the flow is:
1. Skip all of the ASCII characters at the start of the string with `search_nonascii`.
2. Process the multi-byte in our case `'é'` 
3. Skip all the ASCII characters until we hit another multi-byte. ([line](https://github.com/ruby/ruby/blob/86049a98754507a480e657274b4d38b07c46e6ea/string.c#L798) where this happens)
4. Jump to 2 unless at the end of the string.

In `String#scrub` we do the following:
1. Skip all of the ASCII characters at the start of the string with `search_nonascii`.
2. Process (scrub) the multi-byte in our case `'é'` 
3. Jump to 2 unless at the end of the string.

It turns out that processing the multi-bytes is slow as compared to the `search_nonascii` function. My suggested code change here is to add the same ASCII skip to `String#scrub` that is used in step 3 of the `String#valid_encoding?` implementation. If my understanding of this is correct, the pathological case would be strings of full multi-byte characters. This is what the two `worst_case` benchmarks show where we see worse performance than the original implementation of 0.7x across a 1KB string. This is from the extra `search_nonascii` function call per multi-byte which functions as a no-op.

> It should be possible to additionally speed up the case of valid UTF-8 (strings or ranges) rather than just ascii if that's desirable

I'm really curious about any ideas you have here! I'm also happy to answer any questions.

duerst:
@FletcherDares:

I wrote:
> the maximum that can be saved is 1s, but much more can be lost

to which you explained again why things get faster (which I already understand). What I tried to suggest is that you show the inverse of the speedup (call it slowdown or whatever) in your table. That would e.g. be 1.42 for scrub_unknown_but_valid_coderange_worst_case and 0.022 for scrub_english_html. This makes it much easier to compare the different numbers.

> It should be possible to additionally speed up the case of valid UTF-8 (strings or ranges) rather than just ascii if that's desirable. But that would be a bit more code.
SIMD processing can go a long way. For an example (not validating, but transcoding) see https://dl.acm.org/doi/abs/10.1145/1345206.1345222. With 64-bit words, 8 bytes can be handled at the same time, even without SIMD instructions. An example of this is the function count_utf8_lead_bytes_with_word in string.c, which is used in the function enc_strlen, but assumes valid UTF-8.

FletcherDares:
Hi @duerst, sorry for the delay!

Here's an updated table that includes slowdown figures as well:

| Benchmark                                      | Control    | Experiment | Speedup | Slowdown |
| :--------------------------------------------- | ---------: | ---------: | :------ | :------- |
| scrub_known_valid                              |    29.075M |    29.708M | 1.0x    | 1.0x     |
| scrub_known_invalid                            |   247.926k |   740.033k | 2.9x    | 0.34x    |
| scrub_unknown_but_valid_coderange              |   274.488k |   889.488k | 3.2x    | 0.31x    |
| scrub_unknown_and_invalid_coderange            |   269.232k |   739.194k | 2.8x    | 0.36x    |
| scrub_unknown_but_valid_coderange_worst_case   |   234.635k |   164.706k | 0.7x    | 1.42x    |
| scrub_unknown_and_invalid_coderange_worst_case |   239.335k |   162.561k | 0.7x    | 1.47x    |
| scrub_japanese_html                            |     187.47 |    2927.00 | 15.6x   | 0.064x   |
| scrub_english_html                             |     510.22 |   22694.00 | 44.5x   | 0.022x   |

The primary goal of this PR is to unify the approach between `String#valid_encoding?` and `String#scrub` by skipping ASCII characters (the performance improvement is a nice side effect of that change).

Given that scope, do you think it's worth pursuing SIMD or other further performance optimizations as part of this PR, or would you be okay with investigating / tackling those in a follow-up (i.e. is this blocking in your opinion)?

FletcherDares:
I realized that if I gate the `search_nonascii` call behind a check to see if the first character is ASCII we can avoid the overhead from `search_nonascii` with a simple bitmask. This turns out to have done wonders for the performance as can be seen below:

The basis is that most ASCII characters will follow other ASCII character and most multi-byte will follow other multi-bytes.

| Benchmark | Base | New w/ Guard | Speedup | Slowdown |
|---|---:|---:|---:|---:|
| `scrub_known_valid` | 15.582M | 15.794M | 1.01x | 0.99x |
| `scrub_known_invalid` | 250.163k | 730.166k | 2.92x | 0.34x |
| `scrub_unknown_but_valid_coderange` | 272.724k | 965.076k | 3.54x | 0.28x |
| `scrub_unknown_and_invalid_coderange` | 248.361k | 716.855k | 2.89x | 0.35x |
| `scrub_unknown_but_valid_coderange_worst_case` | 250.724k | 245.481k | 0.98x | 1.02x |
| `scrub_unknown_and_invalid_coderange_worst_case` | 227.917k | 226.813k | 1.00x | 1.00x |
| `scrub_english_html` | 0.373k | 16.991k | 45.55x | 0.02x |
| `scrub_japanese_html` | 0.031k | 0.704k | 22.71x | 0.04x |

ioquatix (approved):
LGTM. @nobu can you also review?  Thanks :)

ima1zumi (approved):
LGTM

PR #16360, opened 2026-03-10
fix: raise on shape capacity overflow for large ivar counts

GrantBirki:
This pull request makes shape capacity growth raise `ArgumentError` when the grown capacity would exceed `SHAPE_MAX_FIELDS`, instead of truncating the value back into `attr_index_t`. That lets objects with very large numbers of instance variables fail cleanly instead of overflowing the ivar backing store and crashing the interpreter.

### DIFF
--- a/shape.c
+++ b/shape.c
@@ -487,7 +487,12 @@ shape_grow_capa(attr_index_t current_capa)
         capacities++;
     }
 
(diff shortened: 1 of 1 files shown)


GrantBirki:
bump @tenderlove

hsbt:
Resolved by 678d86f5e46, which removed the `malloc` growth path and now transitions to a COMPLEX shape once `max_capacity` is reached, so the overflow can no longer happen.

PR #16371, opened 2026-03-11
Lazy load error_highlight, did_you_mean, and syntax_suggest

hsbt:
Defer requiring these gems from boot time to first Exception#detailed_message call. They only enhance error display, so loading them eagerly is unnecessary overhead.

On macOS, `ruby -e1` boot time drops from 114ms to 30ms (with user-installed gems), matching `--disable-*` flag performance.

### DIFF
--- a/test/ruby/test_ractor.rb
+++ b/test/ruby/test_ractor.rb
@@ -401,6 +401,23 @@ class TestRactor < Test::Unit::TestCase
     RUBY
   end
 
+  def test_detailed_message_in_ractor
+    # gem_prelude defers loading error_highlight, did_you_mean, and
+    # syntax_suggest until the first error display, which is not possible
+    # in non-main Ractors. Exception#detailed_message must still return
+    # the plain message there. [Feature #21951]
+    assert_ractor(<<~'RUBY', args: ["--enable=gems"], ignore_stderr: true)
+      message = Ractor.new do
+        begin
+          raise "boom"
+        rescue => e
+          e.detailed_message(highlight: false)
+        end
+      end.value
+      assert_include(message, "boom")
+    RUBY
+  end
+
   def assert_make_shareable(obj)
     refute Ractor.shareable?(obj), "object was already shareable"
     Ractor.make_shareable(obj)
--- a/test/ruby/test_ractor.rb
+++ b/test/ruby/test_ractor.rb
@@ -402,19 +402,20 @@ class TestRactor < Test::Unit::TestCase
   end
 
   def test_detailed_message_in_ractor
-    # gem_prelude defers loading error_highlight, did_you_mean, and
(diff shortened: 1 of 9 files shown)


byroot:
I'm really thorn on this. Faster boot time is definitely a good thing for small scripts.

However for larger application, this is a lot of code that would probably end up being loaded too late to benefit from copy on write, and cause various caches to be busted at runtime.

But I guess it's always possible to explicitly eagerly require them?

eregon:
FWIW I was wondering if maybe `autoload` could be used to achieve most of these benefits, and be more transparent.
Also this approach means `ErrorHighlight, DidYouMean, SyntaxSuggest` will be defined but as empty modules (I guess), and code that does `DidYouMean.some_method` might break if it doesn't `require 'did_you_mean'` before, or fail in unexpected way for new code.

hsbt:
> I'm really thorn on this. Faster boot time is definitely a good thing for small scripts.
> 
> However for larger application, this is a lot of code that would probably end up being loaded too late to benefit from copy on write, and cause various caches to be busted at runtime.
> But I guess it's always possible to explicitly eagerly require them?

Maked sense. I only focused boottime for small application at this investigation. For large application, we can introduce some option or configuration.

I don't like a lot of configuration and options, and I don't want hacks like initializing with exception calls to be widespread.

I'll try to find a good solution at a developer meeting or something.

byroot:
Not sure how good an idea it is, but perhaps decorating `Process.warmup` could be an option? It's becoming relatively standard to call it prefork nowadays.

Another idea is just for these libraries to expose a `.eager_load!` method. That's a Rails convention, but is used extensively even outside of the Rails world.

hsbt:
In that sense, I think we can load them by `Ruby::DetailedError.load` in `application.rb`. 

I feel like there are better class and method names now.

test/ruby/test_ractor.rb
```diff
@@ -372,6 +372,23 @@ def test_ractor_new_raises_isolation_error_if_proc_uses_yield
     end
   end
 
+  def test_detailed_message_in_ractor
+    # gem_prelude defers loading error_highlight, did_you_mean, and
+    # syntax_suggest until the first error display, which is not possible
+    # in non-main Ractors. Exception#detailed_message must still return
```
jhawthorn: > which is not possible in non-main Ractors

Why? Non-main ractors support autoloading, and the require is delegated to the main Ractor.

I think we should be able to make this work 😊

hsbt: Thanks. I am not familiar with Ractor, so I had left this out of scope to keep the change small. Prompted by your comment, I asked Claude to try it, and it worked. I adopted this in 427027d0005.

PR #16388, opened 2026-03-13
Describe multiple registrations per rb_gc_register_address

larskanis:
Add a note about multiple registrations of one and the same VALUE address.

It is implemented for more than a decade, but is not documented so far. IMHO it's intended behavior. So let's document and test it, to make it clear.


larskanis:
Not sure why the tests don't run in GHA, but I added a test case for multiple registrations of one and the same address.

This is related to #16097 since Rice registers multiple references in the same way as I [proposed for Swig](https://github.com/swig/swig/pull/3377). Both depend on the behavior documented in this PR.

PR #16409, opened 2026-03-16
ZJIT: Remove WriteBarrier if load-store eliminated corresponding StoreField

tekknolagi:
We don't need these extra write barriers. Track the InsnId of the corresponding StoreField/ArrayAset instruction on the WriteBarrier and keep track of which ones were deleted in load-store elimination.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -872,7 +872,9 @@ pub enum Insn {
     /// Write `val` at an offset of `recv`.
     /// When writing a Ruby object to a Ruby object, one must use GuardNotFrozen (or equivalent) before and WriteBarrier after.
     StoreField { recv: InsnId, id: ID, offset: i32, val: InsnId },
-    WriteBarrier { recv: InsnId, val: InsnId },
+    /// Fire a GC write barrier for writing `val` into `recv`. Hold onto a reference to the
+    /// corresponding `StoreField` instruction to make eliding write barriers easier.
+    WriteBarrier { recv: InsnId, val: InsnId, store: InsnId },
 
     /// Check whether VM_FRAME_FLAG_MODIFIED_BLOCK_PARAM is set in the environment flags.
     /// Returns CBool (0/1).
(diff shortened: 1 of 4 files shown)


tekknolagi:
This is maybe a little icky because we're special-casing StoreField/ArrayAset as "producing an output" in our definite assignment verifier. Too gross? What should we do instead?

XrXr:
Feels fine to me. If we want to completely build this out we could define the output of these stores as producing a zero width type (like `struct A;` in Rust) that has to be consumed. But we're not about to add zero width and linear type to the type system.

zjit/src/hir.rs
```diff
@@ -4953,6 +4960,7 @@ impl Function {
         for block in self.rpo() {
             let old_insns = std::mem::take(&mut self.blocks[block.0].insns);
             let mut new_insns = vec![];
+            let mut deleted_stores = vec![];
```
jacob-shops, line 4963: I keep trying to think of an idiomatic way for the relationship between `LoadField` and `WriteBarrier` to be encoded in this case that could be cleaner. Not because this case is bad, but because I think the optimizer will have other such cases of instruction interactions in the future. A few thoughts for now:

1. One answer is to formally encode such relationships, but I think this becomes sea-of-nodes-ish and unwieldy very quickly. I like your solution.
2. Another answer is to potentially look at some inversion of the current effects system; or a lower bound of effects. Usually we elide checks when we know there are no effects (like Randy's patchpoint patch). Here, we want to elide `WriteBarrier`, `LoadField` and `StoreField` when we know that the desired effect has already occurred in a kind of idempotency check. No action needed for now and I like this change. My gut says there's something here for our optimizer API and to keep things clean in the longer term.

However, I'm not entirely convinced that my point in 2 doesn't end up becoming an ad-hoc reimplementation of 1 and I need to think about it more.

zjit/src/hir.rs
```diff
@@ -4984,7 +4993,12 @@ impl Function {
                         }
                         insn_id
                     }
-                    Insn::WriteBarrier { .. } => {
+                    Insn::WriteBarrier { store, .. } => {
+                        if deleted_stores.contains(&store) {
+                            // We dropped the corresponding store instruction; don't fire the write
+                            // barrier for no reason.
+                            continue;
+                        }
```
jacob-shops, line 5001: I think the answer is no, but I want to gut-check.

Is there ever a situation where it is possible to emit a `WriteBarrier` *before* the deleted `StoreField`? This approach won't catch this case, but I don't think it's a valid scenario anyway.

jacob-shops (approved):
Nice catch!

zjit/src/hir.rs
tekknolagi: Anything's possible but in this case because of data dependencies the WriteBarrier must be referring to a previous instruction

k0kubun (approved):

tekknolagi:
Closing for now.

PR #16430, opened 2026-03-17
ZJIT: Address feedback from @tekknolagi

tenderlove:
I pushed the redundant move elimination to the linearization pass. I'm unsure if `resolve_parallel_mov_pass` is necessary anymore.


k0kubun (approved):

tekknolagi (commented):
What about the other stuff from https://github.com/ruby/ruby/pull/16295#pullrequestreview-3963176637 ? Such as SCRATCH opnd, etc

PR #16469, opened 2026-03-19
[Bug #21301] Fix Time.new month-mday normalization with UTC zone

khasinski:
## Summary

`Time.new` with `"UTC"` zone skips month-mday normalization, producing invalid dates:

```ruby
Time.new(2025, 4, 31, 0, 0, 0, "UTC").to_s   #=> "2025-04-31 00:00:00 UTC" (!)
Time.new(2025, 4, 31, 0, 0, 0, "+00:00").to_s #=> "2025-05-01 00:00:00 +0000"
Time.utc(2025, 4, 31).to_s                     #=> "2025-05-01 00:00:00 UTC"
```

The internal timestamp is correct (all three agree on `to_i`), but the cached `vtm` fields (`mon`, `mday`) are wrong because the UTC path in `time_init_vtm()` sets `tm_got=1` without normalizing month-mday first.

## Fix

- Extract the existing month-mday normalization from `time_arg()` into a reusable `vtm_normalize_month_mday()` function
- Call it in the UTC branch of `time_init_vtm()` before `vtm_day_wraparound()`

This also fixes the string parsing path (`Time.new("2025-04-31T00:00:00Z")`), which flows through the same `time_init_vtm()` function.

https://bugs.ruby-lang.org/issues/21301

### DIFF
--- a/test/ruby/test_time.rb
+++ b/test/ruby/test_time.rb
@@ -1515,4 +1515,35 @@ class TestTime < Test::Unit::TestCase
       assert_equal("-10000-01-01T00:00:00Z", Time.utc(-10000).__send__(method))
     end
   end
+
+  def test_utc_month_mday_normalization
+    bug21301 = "[ruby-core:121801] [Bug #21301]"
+
+    # 30-day months: April 31 -> May 1
+    assert_equal(Time.utc(2025, 5, 1), Time.new(2025, 4, 31, 0, 0, 0, "UTC"), bug21301)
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Thanks you for working on this. Implementation wise, this looks this same as #13246, though with more complete tests.

I'm not sure whether this addresses the leap seconds issue discussed in Bug 21301, but it seems worth merging anyway.

PR #16481, opened 2026-03-20
[Bug #21849] Fix Ripper.lex splitting tstring_content on invalid interpolation

khasinski:
## Summary

`Ripper.lex('"x#$%"')` incorrectly emits two `tstring_content` tokens (`"x"` + `"#$%"`) instead of one (`"x#$%"`). Same issue for `#@` with invalid following characters.

```ruby
# Before (two tokens):
Ripper.lex('"x#$%"')
#=> [... [:on_tstring_content, "x"], [:on_tstring_content, "#$%"], ...]

# After (single token, matches Prism):
Ripper.lex('"x#$%"')
#=> [... [:on_tstring_content, "x#$%"], ...]
```

## Root cause

In `tokadd_string()`, the lexer breaks out of the scanning loop on any `#$` or `#@` without checking if the following character forms a valid variable name. The caller (`parser_peek_variable_name`) then correctly rejects it, but by then the token is already split.

## Fix

Split the `#{`/`#$`/`#@` check: `#{` always breaks (embedded expression), but `#$` and `#@` now only break when `parser_peek_variable_name()` confirms a valid variable follows. Valid interpolations (`#$_`, `#$!`, `#@foo`) are unaffected.

https://bugs.ruby-lang.org/issues/21849

### DIFF
--- a/parse.y
+++ b/parse.y
@@ -6766,6 +6766,7 @@ none		: /* none */
 
 static int regx_options(struct parser_params*);
 static int tokadd_string(struct parser_params*,int,int,int,long*,rb_encoding**,rb_encoding**);
+static enum yytokentype parser_peek_variable_name(struct parser_params *p);
 static void tokaddmbc(struct parser_params *p, int c, rb_encoding *enc);
 static enum yytokentype parse_string(struct parser_params*,rb_strterm_literal_t*);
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Fix looks correct to me, though I'm not a ripper expert.

PR #16523, opened 2026-03-24
ZJIT: Allow non-terminator instructions after LIR terminators

k0kubun:
This PR drops the constraint (added in https://github.com/ruby/ruby/pull/15974 after https://github.com/ruby/ruby/pull/15879) that LIR blocks must end with a terminator instruction. It allows `--zjit-perf=hir` https://github.com/ruby/ruby/pull/16501 to put a `PosMarker` after jump instructions to include them in perf symbols.

With this change, it's also possible to profile `IfFalse`/`IfTrue`/`Jump`, but I made it out of scope for this PR to make reviews easy.

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -139,7 +139,6 @@ impl BasicBlock {
         if self.rpo_index == DUMMY_RPO_INDEX {
             return EdgePair(None, None);
         }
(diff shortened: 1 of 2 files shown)


k0kubun:
converting this to draft again until I fix it to introduce an upper limit to the number of instruction scans per block.

PR #16537, opened 2026-03-25
Use build-mode: none for CodeQL C/C++ scanning

hsbt:
Remove the autobuild step and build-related setup steps since CodeQL now supports scanning C/C++ without builds.


hsbt:
This change resulted in the following changes to the execution time.

Before: 15m 39s at https://github.com/ruby/ruby/actions/runs/23522810179/job/68469469295
After: 4m 16s

Because different tools use different verification methods, the results may not be identical. However, since the SAST check is just guardrail, I will choose the one with the shorter execution time.

PR #16570, opened 2026-03-26
Fix performance of fiber pool stack tests.

samuel-williams-shopify:


ioquatix (approved):
LGTM.

k0kubun:
@ioquatix @samuel-williams-shopify Currently, `TestFiber#test_fiber_pool_stack_acquire_failure` fails (with Timeout or SIGKILL) a lot with Ruby 4.0 on Ruby CI:

Screenshot 2026-05-11 at 21 00 42

For master, it appears to be green, and I assume this PR's changes has some impact on it. Do you have any idea how we can stabilize the test for `ruby_4_0`?

Ideally, we want to avoid backporting `RUBY_SHARED_FIBER_POOL_MAXIMUM_COUNT` and `RUBY_SHARED_FIBER_POOL_MINIMUM_COUNT` environment variables, if they're considered as new features of Ruby 4.1.

ioquatix:
Sorry about this issue.

> Ideally, we want to avoid backporting RUBY_SHARED_FIBER_POOL_MAXIMUM_COUNT and RUBY_SHARED_FIBER_POOL_MINIMUM_COUNT environment variables, if they're considered as new features of Ruby 4.1.

I agree, they are new features but without it, the test is flaky (and basically impossible to fix), IOW, I'd consider deleting the test in 4.0 if that's the only option. Since we have the test in 4.1, I think it's good enough.

PR #16615, opened 2026-03-31
Check for VM_ENV_DATA_INDEX_FLAGS in mid-escape

jhawthorn:
vm_make_env_each iterates over frames on the stack, escaping each one. This can leave the stack in an inconsistent state.

If a sampling profiler calls rb_profile_frames from a signal handler, it can observe the stack in this inconsistent state, so we need to check for it.

This was validated by calling `rb_profile_frames` at [nearly every point inside `vm_make_env_each`](https://github.com/jhawthorn/ruby/commit/f4f8a6f8b93fa805e7804cff8265d2b322a88141). This provides a reliable reproduction of the issue (± a signal arriving more unpredictably because of a compiler potentially rearranging writes).

I really wanted to fix this by removing the the interim state with the forwarding pointer in `ep[0]`, but I wasn't able to make it work. This approach should mitigate the issue for now.

### DIFF
--- a/vm_insnhelper.c
+++ b/vm_insnhelper.c
@@ -800,7 +800,12 @@ rb_vm_frame_method_entry_unchecked(const rb_control_frame_t *cfp)
     const VALUE *ep = cfp->ep;
     rb_callable_method_entry_t *me;
 
-    while (!VM_ENV_LOCAL_P_UNCHECKED(ep)) {
+    for (;;) {
+        if (UNLIKELY(!FIXNUM_P(ep[VM_ENV_DATA_INDEX_FLAGS]))) {
+            // vm_make_env_each mid-escape: flags slot holds the env object.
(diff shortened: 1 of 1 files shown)


vm_insnhelper.c
AlexaCampusano, line 789: Would we need to apply the same change here? In the stacktrace I followed when I was investigating the issue, exceptions were coming from `rb_vm_frame_method_entry`.

AlexaCampusano (commented):
One comment

vm_insnhelper.c
jhawthorn: > Would we need to apply the same change here? In the stacktrace I followed when I was investigating the issue, exceptions were coming from `rb_vm_frame_method_entry`.

I think that was on Ruby 3.4. On 4.0+, since #14648, rb_profile_frames should only be calling `rb_vm_frame_method_entry_unchecked`

PR #16618, opened 2026-03-31
ZJIT: Add missing guard on ivar access on T_{DATA,CLASS,MODULE}

XrXr:
T_DATA, T_MODULE, and T_CLASS objects can share the exact same shape.
The shape on these objects give an index off of the fields array to
get at the ivar. When two objects share the same shape, but differ
in the T_* builtin type, however, the way to get to the fields array
differ.

Previously, we did not guard the builtin type, so the guard allowed
using say, loading `t_string[RCLASS_OFFSET_PRIME_FIELDS_OBJ]`. A classic
type confusion situation that crashed.

Guard the builtin type, in addition to the shape. Note that this is not
necessary for T_OBJECTs since they never have the same shape as other
builtin types.

---

Our pipeline isn't smart enough to figure out that the shape id and the T_* builtin type are in the same 64 bit and collapse down to a mask-and-compare, oh well.

### DIFF
--- a/zjit/src/hir_type/mod.rs
+++ b/zjit/src/hir_type/mod.rs
@@ -1,6 +1,7 @@
 //! High-level intermediate representation types.
 
 #![allow(non_upper_case_globals)]
+use crate::cruby;
 use crate::cruby::{rb_block_param_proxy, Qfalse, Qnil, Qtrue, RUBY_T_ARRAY, RUBY_T_CLASS, RUBY_T_HASH, RUBY_T_MODULE, RUBY_T_STRING, VALUE};
 use crate::cruby::{rb_cInteger, rb_cFloat, rb_cArray, rb_cHash, rb_cString, rb_cSymbol, rb_cRange, rb_cModule, rb_zjit_singleton_class_p};
 use crate::cruby::ClassRelationship;
@@ -213,6 +214,7 @@ impl Type {
             else if val.class_of() == unsafe { rb_cSymbol } { bits::DynamicSymbol }
             else if let Some(bits) = Self::bits_from_exact_class(val.class_of()) { bits }
             else if let Some(bits) = Self::bits_from_subclass(val.class_of()) { bits }
+            else if val.typed_data_p() { bits::TypedTData }
             else {
                 unreachable!("Class {} is not a subclass of BasicObject! Don't know what to do.",
                              get_class_name(val.class_of()))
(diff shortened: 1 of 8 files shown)


XrXr:
This should fix https://github.com/Shopify/ruby/issues/964 and probably some other flaky failures. Running locally shows that these tests are actually SEGVing. On CI they probably get lucky with the out-of-bounds access and the guard page, triggering a `SystemStackError` instead of crashing.

XrXr:
This is also a bug in the polymorphic getivar implementation. I can fix that in a separate PR.

https://github.com/ruby/ruby/blob/0364a612166631ee20663391aefe122d374aff7e/zjit/src/hir.rs#L8172

k0kubun:
Can we remove `test/.excludes-zjit/TestOpenURIProxy.rb`, `test/.excludes-zjit/TestOpenURI.rb`, and `test/.excludes-zjit/TestOpenURISSL.rb` in this PR? (i.e. revert https://github.com/ruby/ruby/pull/16606 and https://github.com/ruby/ruby/pull/16603)

k0kubun (approved):
Thanks for fixing this :pray:

zjit/src/hir_type/mod.rs
```diff
@@ -429,6 +431,24 @@ impl Type {
         matches!(self.spec, Specialization::Object(_))
     }
 
+    /// Find a `T_*` type that is exactly as wide as `self`.
+    pub fn builtin_type_equivalent(&self) -> Option<cruby::ruby_value_type> {
+        if self.bit_equal(types::Array) {
```
tekknolagi, line 436: why do we need it to be exactly as wide? this strikes me as something where we want to do `.class() == c_String` or something. or maybe `is_subtype(types::Array)`.

XrXr: If we allow the T_* type to be wider than the hir type, then guarding with a singleton HIR type would admit other instances that are not in the specialization. If we allow it to be narrower, some values that should pass the guard don't. The only option left is to make it invariant.

XrXr:
Ok, this fixes _a_ bug, but the SystemStackError are still happening. Maybe they're hitting the polymorphic path. I won't ship the revert for now.

tekknolagi:
Coming back to this: in an ideal world, the shape ID should be a reflection of the physical layout in memory. For that reason, ideally T_OBJECT embedded a,b,c,d should (ideally) not have the same shape as T_CLASS extended a,b,c,d. Why was that introduced?

cc @tenderlove

XrXr:
T_OBJECT shapes and T_CLASS shapes don't overlap, but T_DATA, T_MODULE, and generic ivar imemo fields objects can share shape. If the polymorphic decision tree only consists of T_OBJECTs, we could emit different HIR for that that only uses the shape ID.

tekknolagi:
I think they should not share shapes :/

PR #16621, opened 2026-03-31
ZJIT: Optimize send with a nil block to SendWithoutBlockDirect

### LINKED ISSUE #969: ZJIT: Optimize send with nil block arg to SendDirect
When a `send` has `ARGS_BLOCKARG` and the block arg is nil (either statically known or profiled as monomorphically nil), we can strip the block arg and optimize the call to `SendDirect` without a block.

This reduces the `param_block` category in the complex argument-parameter fallback stats (34.2% of total on lobsters as of the burndown in #833).

### Implementation
- Profile the block arg alongside self and regular args for `send` instructions
- If the block arg type is statically `NilClass` or profiled as monomorphically nil, strip it and emit `SendDirect` without a block
- Insert a `GuardBitEquals` side exit when relying on profiled (not static) nil
- Update the frame state snapshot after stripping to keep SP consistent in codegen

### Bug fix
The initial implementation had a stack consistency error (`sp: N+1, bp: N`) when the guard side-exited for a non-nil block. The frame state snapshot still contained the stripped block arg, causing `gen_save_sp` to compute the wrong SP. Fixed by creating a new snapshot with the block arg removed.

Related: #833 (lobsters perf burndown)

### LINKED ISSUE #1002: ZJIT: Specialize send with caller_blockarg
On lobsters,

```
Top-20 send fallback reasons (99.9% of total 4,681,626):
                             invokeblock_not_specialized: 1,962,606 (41.9%)
                            one_or_more_complex_arg_pass:   914,920 (19.5%)
                          send_without_block_polymorphic:   508,454 (10.9%)
                          send_without_block_no_profiles:   316,413 ( 6.8%)
                          send_without_block_megamorphic:   246,494 ( 5.3%)
                             sendforward_not_specialized:   217,984 ( 4.7%)
                                        send_polymorphic:   124,354 ( 2.7%)
  send_without_block_not_optimized_method_type_optimized:    97,655 ( 2.1%)
                                   too_many_args_for_lir:    74,307 ( 1.6%)
```

Next to `invokeblock_not_specialized`, `one_or_more_complex_arg_pass` is the `#2` fallback reason.

```
Top-7 popular complex argument-parameter features not optimized (100.0% of total 1,017,314):
    caller_blockarg: 571,544 (56.2%)
  param_forwardable: 217,399 (21.4%)
         param_rest:  97,788 ( 9.6%)
       caller_kwarg:  67,957 ( 6.7%)
       param_kwrest:  35,286 ( 3.5%)
       caller_splat:  14,132 ( 1.4%)
    caller_kw_splat:  13,208 ( 1.3%)
```

`caller_blockarg` is the most frequent reason for `one_or_more_complex_arg_pass` on lobsters.

rwstauner:
I added tests in the first commit so that you can see the HIR change in the second one.

This mostly just ends up reclassifying send_fallback_one_or_more_complex_arg_pass into the new send_block_arg_not_nil bucket and send_polymorphic (revealing many sites can't convert due to a polymorphic receiver, not the block) but railsbench does increase the direct send count.

relevant stats comparing master `c8c8f2257c` vs branch `e6c6e2f19b`
using ```caffeinate ./run_benchmarks.rb --chruby 'm::ruby-mdev --zjit-stats' --chruby 'b::ruby-bdev --zjit-stats' --warmup 30 --bench 10 railsbench lobsters```

### railsbench
| counter | master | branch | Δ |
|---|---|---|---|
| `send_count` | 596,132,896 | 596,130,244 | −2,652 |
| `complex_arg_pass` | 13,151,900 | 8,186,689 | −4,965,211 |
| `send_block_arg_not_nil` (new) | 0 | 3,843,157 | +3,843,157 |
| `send_polymorphic` | 7,719,140 | 7,930,552 | +211,412 |
| `iseq_optimized` | 111,866,816 | 112,498,067 | +631,251 |
| `optimized_send_count` | 531,393,754 | 532,221,561 | +827,807 |
| `dynamic_send_count` | 64,739,142 | 63,908,683 | −830,459 |
| `exit_block_arg_not_nil` | 0 | 1 | +1 |
| `side_exit_count` | 2,133,396 | 2,133,407 | +11 |

### lobsters
| counter | master | branch | Δ |
|---|---|---|---|
| `send_count` | 306,460,169 | 306,473,895 | +13,726 |
| `complex_arg_pass` | 7,617,096 | 3,383,030 | −4,234,066 |
| `send_block_arg_not_nil` (new) | 0 | 3,987,958 | +3,987,958 |
| `send_polymorphic` | 4,930,077 | 5,066,919 | +136,842 |
| `iseq_optimized` | 55,933,199 | 55,949,658 | +16,459 |
| `optimized_send_count` | 257,870,129 | 257,921,680 | +51,551 |
| `dynamic_send_count` | 48,590,040 | 48,552,215 | −37,825 |
| `exit_block_arg_not_nil` | 0 | 0 | 0 |
| `side_exit_count` | 2,140,879 | 2,141,996 | +1,117 |


k0kubun:
Does lobsters have no `block_arg_not_nil` exits with this?

k0kubun (approved):

zjit/src/hir.rs
```diff
@@ -3704,9 +3708,38 @@ impl Function {
...
+                            if let Some(&block_arg) = args.last() {
+                                let block_arg_type = self.type_of(block_arg);
+                                if block_arg_type.is_subtype(types::NilClass) {
+                                    // Block arg is known to be nil - strip it and treat as no block
+                                    self.push_insn(block, Insn::GuardBitEquals {
+                                        val: block_arg,
+                                        expected: Const::Value(Qnil),
+                                        reason: SideExitReason::BlockArgNotNil,
+                                        state,
+                                    });
```
tekknolagi: If it's known to be nil, we don't need to guard it

tekknolagi: Also we have `self.is_a` I think

zjit/src/profile.rs
```diff
@@ -97,6 +97,8 @@ fn profile_insn(bare_opcode: ruby_vminsn_type, ec: EcPtr) {
             let cd: *const rb_call_data = profiler.insn_opnd(0).as_ptr();
             let argc = unsafe { vm_ci_argc((*cd).ci) };
             // Profile all the arguments and self (+1).
+            // Note: For sends with ARGS_BLOCKARG, the block arg is NOT profiled here.
```
tekknolagi: Could we? Should we?

rwstauner:
I might be doing something wrong but I don't see any nonzero `exit_block_arg_not_nil` when running any of the benchmarks.

zjit/src/profile.rs
rwstauner: I don't know... what could profiling tell us about blocks?

tekknolagi: If it's nil or not I guess

rwstauner: Oh... and then we could do more of this.  So far we do this if we know statically that it's nil, but if we profiled it and found it to be nil we could do the guard and the optimization for cases that aren't statically known. 🤔

tekknolagi: I dunno how common this is though

k0kubun:
I took the liberty to just rebase the branch onto the latest master and push it to the branch. We still need to address feedback from Max, but I think it's ready for review.

zjit/src/hir.rs
k0kubun: well, sorry my "rebase" ended up containing a diff that addresses this but I didn't mean to push. So this is already addressed. Let's start over from the latest diff.

zjit/src/profile.rs
k0kubun: well, similarly to https://github.com/ruby/ruby/pull/16621#discussion_r3502622506, it looks like I ended up pushing a change that removes it. let's revisit that from the latest diff.

zjit/src/hir.rs
```diff
@@ -3864,9 +3868,57 @@ impl Function {
...
+                                    if !statically_nil {
+                                        // Guard needed when relying on profiled type. Uses the original
+                                        // `state` so a side exit re-executes the send with the block
+                                        // arg still on the VM stack.
+                                        self.push_insn(block, Insn::GuardBitEquals {
+                                            val: block_arg,
+                                            expected: Const::Value(Qnil),
+                                            reason: SideExitReason::BlockArgNotNil,
+                                            state,
+                                            recompile: None,
```
k0kubun: Can we recompile on exit? Or is it that the profiler doesn't collect type information for block args?

k0kubun (approved):

zjit/src/hir.rs
rwstauner: Good catch.  This slipped in during a previous rebase.  Changing it to recompile eliminates some of the side exits that the PR had added previously.  I'll update this.

tekknolagi:
How is it that it makes `optimized_send_count` worse for lobsters?

rwstauner:
> How is it that it makes optimized_send_count worse for lobsters?

I need to redo it.  I rebased the branch but not the master I compared it to.  Now I have to resolve conflicts again.

rwstauner:
I redid the stats against latest commits and it looks better

PR #16631, opened 2026-04-01
ZJIT: Remove Option<T> from UnionFind

tekknolagi:
We can have it fixpoint at nodes pointing to themselves instead. This
also saves space.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -36,6 +36,12 @@ impl From<InsnId> for usize {
     }
 }
 
+impl From<usize> for InsnId {
+    fn from(val: usize) -> Self {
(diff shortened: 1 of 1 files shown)


tekknolagi:
cc @alanwu-shopify-inc in particular because I want to be mindful of usize conversion functions

XrXr (approved):

XrXr:
By the way, we're being a bit naughty by panicking inside `impl From<usize> for VRegId {`, since `From` is supposed to be infallible <https://doc.rust-lang.org/std/convert/trait.From.html>. But it's only a tiny bit worse than having our own trait that has docs about container length limits.

PR #16639, opened 2026-04-02
ZJIT: Fix SEGV in struct accessor after polymorphic type dispatch

k0kubun:
This fixes a crash that can be reproduced with `make test-tool RUN_OPTS="--zjit-call-threshold=1" TESTS="testunit/test_sorting"`, which seems to be introduced by https://github.com/ruby/ruby/pull/16634.


k0kubun:
looks like we have another failure 😕

PR #16667, opened 2026-04-06
ZJIT: Move tracing and LIR passes into run_pass! macro

tekknolagi:


zjit/src/backend/arm64/mod.rs
```diff
@@ -1627,8 +1627,8 @@ impl Assembler {
                 }
             }
 
-            let preferred_registers = trace_compile_phase("preferred_registers", || asm.preferred_register_assignments(&intervals));
-            let (assignments, num_stack_slots) = trace_compile_phase("linear_scan", || asm.linear_scan(intervals.clone(), regs.len(), &preferred_registers));
+            let preferred_registers = run_pass!(asm.preferred_register_assignments(&intervals));
+            let (assignments, num_stack_slots) = run_pass!(asm.linear_scan(intervals.clone(), regs.len(), &preferred_registers));
```
k0kubun, line 1631: > https://github.com/ruby/ruby/pull/16666#discussion_r3042130958
> Instead of `run_pass!(asm.pass_name(args))`, I would make it `run_pass!(pass_name, args)`. 
>
> I think the problem is, as I said above, too many regalloc implementation details are leaked into `compile_with_regs`, which used to have only the pipeline of LIR passes before. It should just be a single `asm.alloc_regs` call and it should run sub-passes in it. It's already traced like that and therefore seems to match our mental model. Then every call of passes can be just `run_pass!(pass_name)` without arguments.

PR #16694, opened 2026-04-09
ZJIT: More recognizable JIT return poison, assert only when RUBY_DEBUG

XrXr:
A good poison value gives a recognizable fault address if someone
erroneously dereferences. Asserting the pointer is not poison shouldn't
be necessary, so move it to RUBY_DEBUG builds.

---

FYI the poison looks like `DF BA DF BA DF BA DF BA` in little endian memory dumps.

### DIFF
--- a/zjit.h
+++ b/zjit.h
@@ -66,10 +66,8 @@ static inline void rb_zjit_jit_frame_update_references(zjit_jit_frame_t *jit_fra
 
 #define rb_zjit_enabled_p (rb_zjit_entry != 0)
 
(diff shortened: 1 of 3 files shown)


XrXr:
Unfortunately, inheriting the same flaky failure from `master` on various platform and configs: `[BUG] inconsistent live slot number: expect 256903, but 256901.`

k0kubun (approved):
> BADFrame `0xbadfbadfbadfbadf`

very cool

PR #16695, opened 2026-04-09
Assume `chdir(2)` isn't called and cache `rb_dir_getwd_ospath()`

byroot:
[Feature #21987]

The cache is reset when `Dir.chdir` is invoked.

```
compare-ruby: ruby 4.1.0dev (2026-04-09T12:24:09Z master c919778017) +PRISM [arm64-darwin25]
built-ruby: ruby 4.1.0dev (2026-04-09T16:39:56Z cache-pwd 9946d0eb07) +PRISM [arm64-darwin25]
```

|             |compare-ruby|built-ruby|
|:------------|-----------:|---------:|
|pwd          |    104.703k|   89.698M|
|             |           -|   856.69x|
|expand_path  |    100.734k|    1.098M|
|             |           -|    10.90x|

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -504,6 +504,7 @@ fnmatch(
 }
 
 VALUE rb_cDir;
+static VALUE cwd_ospath;
 static VALUE sym_directory, sym_link, sym_file, sym_unknown;
 
 #if defined(DT_BLK) || defined(S_IFBLK)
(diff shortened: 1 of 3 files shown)


dir.c
```diff
@@ -1241,6 +1242,7 @@ nogvl_chdir(void *ptr)
 static void
 dir_chdir0(VALUE path)
 {
+    cwd_ospath = 0;
```
byroot, line 1245: I'm afraid we may have a race condition here though, as we release the GVL for `chdir`.

PR #16702, opened 2026-04-09
In gc_marks_finish, use avg_slot_size when allowing heaps to expand

luke-gruber:
Before, the heaps could be underprovisioned because it was always using the smallest slot size in its calculation.

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -5509,7 +5509,13 @@ gc_marks_finish(rb_objspace_t *objspace)
             }
 
             if (full_marking) {
-                heap_allocatable_bytes_expand(objspace, NULL, sweep_slots, total_slots, heaps[0].slot_size);
+                /* Use weighted average slot size since total_slots spans all heaps */
+                size_t total_heap_bytes = 0;
+                for (int i = 0; i < HEAP_COUNT; i++) {
+                    total_heap_bytes += heaps[i].total_slots * heaps[i].slot_size;
+                }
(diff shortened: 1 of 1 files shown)


gc/default/default.c
```diff
@@ -5509,7 +5509,13 @@ gc_marks_finish(rb_objspace_t *objspace)
...
 
             if (full_marking) {
-                heap_allocatable_bytes_expand(objspace, NULL, sweep_slots, total_slots, heaps[0].slot_size);
+                /* Use weighted average slot size since total_slots spans all heaps */
+                size_t total_heap_bytes = 0;
+                for (int i = 0; i < HEAP_COUNT; i++) {
+                    total_heap_bytes += heaps[i].total_slots * heaps[i].slot_size;
+                }
+                size_t avg_slot_size = total_slots > 0 ? total_heap_bytes / total_slots : heaps[0].slot_size;
+                heap_allocatable_bytes_expand(objspace, NULL, sweep_slots, total_slots, avg_slot_size);
```
eightbitraptor, line 5518: I'd be interested in seeing if adjusting the allocatable bytes like this results in a change to the heap shape or GC cadence since we'll now be allowed to allocate more pages.

Not a blocker, I think this is a good change as is.

eightbitraptor (approved):

eightbitraptor (approved):

eightbitraptor:
Sorry for the delay in reviewing this @luke-gruber - I don't know why the notification slipped through the net!

PR #16706, opened 2026-04-09
ZJIT: Make `hir::types::Class` not final and have it include metaclasses

XrXr:
Every class boots with a metaclass, and all metaclasses are subclasses
of Class, so `types::Class` has no business in `ExactBitsAndClass`.
In fact, we should never see an object whose RBasic::class is exactly
rb_cClass because classes get a metaclass on boot. So there is no
ClassExact type.

This fixes the side exits on getivar-module.rb that were introduced in
a8f3c34556bac709587ded4e0e4dad08932d5900 ("ZJIT: Add missing guard
on ivar access on T_{DATA,CLASS,MODULE}"). The `GuardType v, Class` checked
for exactly rb_cClass and never passed.

### DIFF
--- a/zjit/src/hir_type/gen_hir_type.rb
+++ b/zjit/src/hir_type/gen_hir_type.rb
@@ -89,7 +89,7 @@ def base_type name, c_name: nil
   [type, exact]
 end
 
-# Define a new type that cannot be subclassed.
+# Define a new type that has no subclasses and cannot be subclassed.
 # If c_name is given, mark the rb_cXYZ object as equivalent to this type.
 def final_type name, base: $object, c_name: nil
   if c_name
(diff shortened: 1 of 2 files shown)


k0kubun:
Is this something an HIR test can check if is working?

k0kubun (approved):

XrXr:
>Is this something an HIR test can check if is working?

We would want to assert that some HIR runs without exiting. Not sure if we have existing examples?

XrXr:
Ok I tried implementing a assert_compiles_and_no_exits(), but the issue is, you also need to enable_zjit_stats() beforehand to get the counter on exit. But by turning on stats, you also gunk up the HIR printout. Solving these issues is out of scope here. I'll make a ticket.

zjit/src/hir_type/gen_hir_type.rb
```diff
@@ -109,7 +109,9 @@ def final_type name, base: $object, c_name: nil
 base_type "Set", c_name: "rb_cSet"
 base_type "Regexp", c_name: "rb_cRegexp"
 module_class, _ = base_type "Module", c_name: "rb_cModule"
-class_ = final_type "Class", base: module_class, c_name: "rb_cClass"
+# Class cannot be subclassed by doing `class Sub < Class`,
+# but every metaclass is a subclass of `Class`. It's not final.
+module_class.subtype "Class"
```
tekknolagi, line 114: don't we want `base_type` here?

tekknolagi: it would at least make singletons visually distinct in their type bits

XrXr: we can't just use `base_type` because it doesn't let you say `Class < Module`. Also, we don't want ClassExact that it'd introduce, since it's actually an empty type.

tekknolagi: https://github.com/ruby/ruby/pull/17054

PR #16707, opened 2026-04-10
Assert shape doesn't change on iteration

jhawthorn:
Debugging a rare (?) CI error


jhawthorn:
https://github.com/ruby/ruby/pull/16737

PR #16709, opened 2026-04-10
dir.c: cache and revalidate working directory

byroot:
`rb_dir_getwd_ospath()` is called quite frequently, but the overwhelming majority of the time, the current directory didn't change.

We can also assume that most of the time, `PATH_MAX` is enough for `getcwd`, hence we can first attempt to use a small stack buffer rather than always allocate on the heap.

This way we can keep the last `pwd` and revalidate it with no allocation.

On macOS syscalls are fairly slow, so the gain isn't very large.

macOS:

```
compare-ruby: ruby 4.1.0dev (2026-04-09T05:19:02Z master c091c186e4) +PRISM [arm64-darwin25]
built-ruby: ruby 4.1.0dev (2026-04-09T06:37:20Z get-cwd-cache ea02126d79) +PRISM [arm64-darwin25]
```

|         |compare-ruby|built-ruby|
|:--------|-----------:|---------:|
|Dir.pwd  |    105.183k|  113.420k|
|         |           -|     1.08x|

Linux (inside virtualized Docker)
compare-ruby: ruby 4.1.0dev (2026-04-07T08:26:25Z master fcd210086c) +PRISM [aarch64-linux]
built-ruby: ruby 4.1.0dev (2026-04-09T06:38:09Z get-cwd-cache 6774af9ba7) +PRISM [aarch64-linux]

|         |compare-ruby|built-ruby|
|:--------|-----------:|---------:|
|Dir.pwd  |      4.157M|    5.541M|
|         |           -|     1.33x|

### DIFF
--- a/dir.c
+++ b/dir.c
@@ -1585,6 +1585,8 @@ dir_chdir(VALUE dir)
 #endif
 }
 
+static VALUE last_cwd;
+
 #ifndef _WIN32
 static VALUE
 getcwd_to_str(VALUE arg)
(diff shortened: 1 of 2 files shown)


dir.c
```diff
@@ -1604,20 +1606,40 @@ getcwd_xfree(VALUE arg)
...
+        return rb_dir_getwd_ospath_slowpath();
+    }
+
+    if (!last_cwd || strcmp(RSTRING_PTR(last_cwd), path) != 0) {
+#ifdef __APPLE__
+        last_cwd = rb_str_normalize_ospath(path, strlen(path));
+#else
+        last_cwd = rb_str_new2(path);
+#endif
+        rb_str_freeze(last_cwd);
```
jhawthorn: I think this is a good idea, but it's accessible to Ractors, so we should use an atomic load from last_cwd and atomic store back into it (we don't need CAS, since we will be validating what we read from the cache, but we do need at least have atomics for the release/acquire semantics to make the string's memory safe to access, also we should probably make sure to store after freezing/make_shareable).

dir.c
```diff
@@ -1604,20 +1606,40 @@ getcwd_xfree(VALUE arg)
...
     return rb_ensure(getcwd_to_str, (VALUE)path, getcwd_xfree, (VALUE)path);
 }
+
+VALUE
+rb_dir_getwd_ospath(void)
+{
+    char buf[PATH_MAX];
+    char *path = getcwd(buf, PATH_MAX);
+    if (!path) {
+        return rb_dir_getwd_ospath_slowpath();
```
jhawthorn, line 1622: If we determine it's acceptable to freeze the string returned from `Dir.pwd` we should do so both in the fast and slowpath.

byroot: `Dir.pwd` dup the string later on. There shouldn't be any user visible change here.

PR #16712, opened 2026-04-10
Ensure version from bundled_gems is used in tool/rdoc-srcdir

st0012:
- `tool/rdoc-srcdir` used `Dir.glob(...).first` to find bundled gems (rdoc, tsort), which picks alphabetically. When multiple versions coexist in `.bundle/gems/` (e.g. `rdoc-7.1.0` and `rdoc-7.2.0`), it selects the wrong (older) one.
- Fix by reading the declared version from `gems/bundled_gems` and constructing the exact path, so the correct version is always loaded.

### DIFF
--- a/tool/rdoc-srcdir
+++ b/tool/rdoc-srcdir
@@ -1,7 +1,16 @@
 #!ruby -W0
 
+srcdir = File.dirname(__dir__)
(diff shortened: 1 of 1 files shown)


st0012:
cc @hsbt

hsbt:
👍

PR #16715, opened 2026-04-10
ZJIT: Assert no side exits in assert_compiles()

### LINKED ISSUE #973: ZJIT: assert_compiles_and_no_exits() test helper
https://github.com/ruby/ruby/pull/16706#issuecomment-4218461051 and the speed drop on getivar-module.rb in ruby-bench showed that we probably want a test helper that asserts there is no side-exits, as I accidentally introduced a never passing `GuardType`. See the link for ergonomic issues to iron out.

XrXr:
Most of the time, we want to assert that we compile and the compiled
code runs without exiting. A small number of tests trigger side exits,
and those are changed to use assert_compiles_allowing_exits().

```console
$ rg -F 'assert_compiles(' -o | wc
     289     289   11862
$ rg -F 'assert_compiles_allowing_exits(' -o | wc
      38      38    2196
```

Fixes: https://github.com/Shopify/ruby/issues/973

### DIFF
--- a/zjit/src/backend/lir.rs
+++ b/zjit/src/backend/lir.rs
@@ -2750,12 +2750,12 @@ impl Assembler
                 }).unwrap_or(false);
 
                 // If enabled, instrument exits first, and then jump to a shared exit.
-                let counted_exit = if get_option!(stats) || should_record_exit {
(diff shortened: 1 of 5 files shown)


k0kubun (approved):
Thanks!

PR #16717, opened 2026-04-10
Go from 2 to 1 allocations in onig_region_resize

luke-gruber:


afurm (approved):
Nice optimization! One thing to double-check - are we sure about memory alignment here? Since both pointers now share one allocation, want to make sure OnigPosition doesn't have any special alignment requirements that could bite us. Other than that, looks good to merge.

luke-gruber:
As far as I understand the situation with Ruby and onigmo, we make patches here and they are backported to onigmo?

luke-gruber:
T_MATCH has embedded onig regions now.

PR #16719, opened 2026-04-11
Speed up Integer#to_s with a two digit lookup table

khasinski:
## Summary

Both `rb_fix2str` (numeric.c) and `big2str_2bdigits` (bignum.c) convert integers to base 10 one digit at a time. This replaces the inner loop with a 200 byte lookup table that emits two digits per iteration. Non base 10 paths are untouched.

## Benchmarks

Apple M series, best of 3 runs, 5M to 10M ops each.

Fixnum (`rb_fix2str`):

| digits | master | patch | delta |
|---|---|---|---|
| 1  (`5`)                  | 64 ns/op  | 64  | 0% |
| 2  (`42`)                 | 64        | 65  | within noise |
| 3  (`400`)                | 66        | 64  | 3% |
| 5  (`12345`)              | 69        | 67  | 3% |
| 10 (`1234567890`)         | 77        | 67  | 13% |
| 19 (`2**62-1`)            | 111       | 75  | 33% |

Bignum (`big2str_2bdigits`, measured against the fixnum patch as baseline so the delta is bignum only):

| digits | base | patch | delta |
|---|---|---|---|
| 20   | 146 ns/op | 124  | 15% |
| 40   | 174       | 152  | 13% |
| 100  | 236       | 213  | 10% |
| 500  | 1119      | 1086 | 3% |
| 1000 | 3490      | 3459 | 1% |

Above ~500 digits the Karatsuba recursion dominates and the digit emission saving drops into the noise.

Break even for fixnum is around 3 digits. `rb_fix_to_s_static[0..9]` already serves single digit results from a cache so those never reach `rb_fix2str`.

## Correctness

- `test/ruby/test_integer.rb`: 38 tests, 421628 assertions, 0 failures
- `test/ruby/test_bignum.rb`: 74 tests, 607 assertions, 0 failures
- `make test-all`: 34694 tests, 0 new failures
- Fuzz: 100k random fixnums and 500 random bignums up to 2**256, cross checked against `sprintf("%d", n)`, plus base 2/8/16/36 round trips

## Why not `ext/json/vendor/jeaiii-ltoa.h`

ruby/json ships jeaiii's itoa and it is already in tree. I tried it in a sibling branch. Results are within 1-3% of the 2 digit table on Apple Silicon, because modern compilers already lower `u / 100` and `u % 100` to a single magic multiply plus shift, which is the same optimization jeaiii hand codes. The 2 digit approach is 20 lines instead of 267, works uniformly for both fixnum and the fixed width leading zero padding required by `big2str_2bdigits`, and achieves the same first order speedup. Happy to switch to jeaiii in a follow up if reviewers prefer it.

## Files touched

- `numeric.c` (+22, -3)
- `bignum.c` (+81, -8) also hosts the shared `ruby_decimal_digit_pairs` constant next to `ruby_digitmap`
- `internal/bignum.h` (+1) declaration
- `benchmark/int_to_s.yml` (new, 25 lines) reproduces the numbers above

Public API and ABI unchanged. Non base 10 conversion, negative handling, and bignum Karatsuba decomposition are all unchanged.

## Test plan

- [x] `make test-all` clean (same pre existing `TestArgf#test_puts` flake as master)
- [x] `make test-all TESTS=ruby/test_integer.rb`
- [x] `make test-all TESTS=ruby/test_bignum.rb`
- [x] 100k random fixnum fuzz vs `sprintf("%d", n)`
- [x] 500 random bignum fuzz up to 2**256 vs `sprintf("%d", n)`
- [x] Base 2/8/16/36 round trip for fix and big
- [x] Edge cases: 0, ±1, ±99, ±100, 2^30-1, 2^62-1, 2^62, 2^100, 10^{19,39,99,499,999}


jeremyevans (approved):
This optimization makes sense to me. Requesting review from @tompng, our resident expert in digits.

tompng (approved):
:+1:

PR #16721, opened 2026-04-11
ZJIT: Annotate Float and Integer predicates

khasinski:
## Summary

Adds method annotations in `zjit/src/cruby_methods.rs` for pure
numeric predicates so ZJIT can:

1. Emit the fast `CCall` path (skip frame push/pop and GC safepoint)
   for `Float#nan?`, `Float#finite?`, `Float#infinite?`.
2. Propagate `BoolExact` return type for `Integer#zero?/even?/odd?`
   and `Float#zero?/positive?/negative?` builtin predicates, so
   downstream HIR passes can reason about boolean results.

All targeted C bodies are pure (`isnan`, `isfinite`, `isinf`, fixnum
bit tests, `FLOAT_ZERO_P`, `RFLOAT_VALUE` comparisons). No allocation,
no Ruby calls, no side effects. Safe to mark `leaf`, `no_gc`,
`elidable`.

## Benchmark

Tight `while` loop, 20M iters, `--zjit-call-threshold=2`, best of 5:

| predicate         | before  | after   | speedup |
|-------------------|---------|---------|---------|
| `Float#nan?`      | 0.140 s | 0.110 s | ~21%    |
| `Float#finite?`   | 0.145 s | 0.114 s | ~21%    |
| `Float#infinite?` | 0.129 s | 0.094 s | ~27%    |

Integer and Float builtin predicates do not show a direct speedup in
isolation (the builtin leaf path was already fast), but the return
type propagation enables guard and branch elimination in downstream
HIR passes.

## Test plan

- [x] `make test-all TESTS=test/ruby/test_zjit.rb` passes (28 tests, 0 failures)
- [x] HIR dump confirms `CCall` (not `CCallWithFrame`) for Float cfunc predicates
- [x] HIR dump confirms `BoolExact` return type on `InvokeBuiltin` for annotated builtin predicates
- [x] Smoke test of all six methods returns correct values under `--zjit`

### DIFF
--- a/benchmark/float_predicate.yml
+++ b/benchmark/float_predicate.yml
@@ -1,0 +1,12 @@
+prelude: |
+  floats = [1.0, -1.0, 0.0, Float::NAN, Float::INFINITY, -Float::INFINITY]
+
+benchmark:
+  float_nan?: floats.each { |f| f.nan? }
+  float_finite?: floats.each { |f| f.finite? }
+  float_infinite?: floats.each { |f| f.infinite? }
+  float_zero?: floats.each { |f| f.zero? }
+  float_positive?: floats.each { |f| f.positive? }
+  float_negative?: floats.each { |f| f.negative? }
+
+loop_count: 1000000
--- a/benchmark/integer_predicate.yml
+++ b/benchmark/integer_predicate.yml
@@ -1,0 +1,9 @@
+prelude: |
+  nums = (0..9).to_a
+
+benchmark:
+  integer_zero?: nums.each { |n| n.zero? }
+  integer_even?: nums.each { |n| n.even? }
+  integer_odd?: nums.each { |n| n.odd? }
+
+loop_count: 1000000
--- a/zjit/src/cruby_methods.rs
+++ b/zjit/src/cruby_methods.rs
@@ -266,9 +266,18 @@ pub fn init() -> Annotations {
     annotate!(rb_cInteger, "[]", inline_integer_aref);
     annotate!(rb_cInteger, "to_s", types::StringExact);
     annotate!(rb_cString, "to_s", inline_string_to_s, types::StringExact);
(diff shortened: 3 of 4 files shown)


khasinski:
The `test-bundled-gems, --zjit-call-threshold=1` failure on `test_debugger_doesnt_complain_about_duplicated_breakpoint` appears to be a flake. I reproduced the test locally with this branch under `--zjit --zjit-call-threshold=1` and it passes consistently (5 out of 5 runs of the specific test, 11 out of 11 of the full `debugger_method_test.rb` file). The failure was a 10s TCP timeout in `setup_remote_debuggee`, which does not touch any of the annotated Float or Integer predicate code paths. Happy to retry the job if a maintainer can kick it.

tekknolagi:
Can you please upload the benchmark? It might be useful to track type propagation at least in HIR tests. Or we can even add it to our benchmark suite.

tekknolagi:
Someone on the team will take a look this coming week but please make sure you add HIR tests for these methods returning the right types, being leaf, etc.

k0kubun (approved):

k0kubun:
Do you think [the CI failure](https://github.com/ruby/ruby/actions/runs/24306181244/job/70967691571) is the one addressed by https://github.com/ruby/ruby/pull/16736?

khasinski:
~~Yes, the `--zjit-call-threshold=1` btest failure (pack with buffer) is the same bug fixed in #16736. Once that lands, this CI should go green on rebase.~~

khasinski:
Correction on my previous comment: I was wrong. The CI failure here is actually `test_sync_default_gems` worker crash in `test-tool`, not the pack/buffer btest. #16736 does not address this. Let me investigate this one separately.

khasinski:
The failure was unrelated, it's resolved after rebasing on top of master.

PR #16729, opened 2026-04-12
[3.2] iseq.c: rb_estimate_iv_count handle no superclass

byroot:
[Bug #21992]

When redefining `BasicObject#initialize` there's no super class to access.


hsbt:
Thanks to file this. Unfortunately, Ruby 3.2 is already EOL status.

PR #16735, opened 2026-04-13
ZJIT: Inline Float arithmetic (+, -, *, /) and Float#to_i

khasinski:
## Summary

Skip `CCallWithFrame` overhead for Float arithmetic by adding
`FloatAdd/Sub/Mul/Div` and `FloatToInt` HIR instructions that lower
to `gen_prepare_leaf_call_with_gc` + direct ccall. This is the same
pattern used by `FixnumDiv`.

The inline functions guard the receiver as Flonum (cheap tag check)
and accept Flonum or Fixnum as the second operand. HeapFloat falls
back to CCallWithFrame via the default Send path.

GC safety: `gen_prepare_leaf_call_with_gc` saves PC and SP to cfp
before the call, so GC can walk the stack if `DBL2NUM` allocates a
heap Float. This matches how `FixnumDiv` calls `rb_jit_fix_div_fix`.

### Why Flonum guard, not Float guard

`GuardType Float` (union of Flonum + HeapFloat) loads the class
pointer from memory and compares to `rb_cFloat`. `GuardType Flonum`
is a single bitwise test: `(val & 3) == 2`. In a tight loop with
multiple Float operations per iteration, the memory loads from Float
guards caused a net regression. Flonum guards gave a net improvement.

## Benchmark

ruby-bench nbody (release build, best of 10+ iterations):

| | Baseline | Patched | Change |
|--|---------|---------|--------|
| nbody | 47ms | 33ms | -30% |

Other benchmarks from ruby-bench and the Ruby repo:

| Benchmark | Change |
|-----------|--------|
| aobench | -4.5% |
| mandelbrot, fib, tak, tarai, pentomino, factorial, strconcat | 0% (no regression) |

I originally discovered this optimization opportunity while profiling
[Ruby Doom](https://github.com/khasinski/doom), a pure-Ruby DOOM
renderer. It is not in the standard benchmarks, but it shows +14% FPS
from these changes. The hot path is a floor/ceiling texture mapping
loop doing ~6 Float operations per pixel across 76,800 pixels per
frame.

## Test plan

- [x] `make zjit-test`: 1602 passed, 0 failed
- [x] `make test-all TESTS=test/ruby/test_zjit.rb`: 29 tests, 0 failures
- [x] Runtime correctness tests for +, -, *, /, *Fixnum, to_i, NaN, Infinity
- [x] HIR snapshot tests for FloatAdd, FloatSub, FloatMul, FloatDiv, FloatToInt, FloatMul with Fixnum
- [x] No regressions on any existing benchmark


khasinski:
The `btest` failure in `make (check, --zjit-call-threshold=1)` is a pre-existing issue on master, not caused by this PR. The failing test (`pack("E*", buffer:)` with Floats) reproduces on current master with `--zjit-call-threshold=1`. The same CI job also [failed on master](https://github.com/ruby/ruby/actions/runs/24302777122) recently.

Fix here https://github.com/ruby/ruby/pull/16736

jit.c
```diff
@@ -827,6 +827,16 @@ rb_jit_fix_div_fix(VALUE recv, VALUE obj)
     return rb_fix_div_fix(recv, obj);
 }
 
+VALUE
+rb_jit_flo_to_i(VALUE num)
```
k0kubun: `jit.c` is for "Glue code shared between YJIT and ZJIT for use from Rust". This should live in `zjit.c` since we're not using it from YJIT.

Also, where did this implementation come from? Can we just define `rb_flo_to_i` that just calls `flo_to_i` and use it to avoid duplicating or deviating from the implementation?

khasinski: Good point on both. Moved the helper to zjit.c and made it just call flo_to_i (exported from numeric.c). The original logic now lives in one place.

k0kubun: You got CI failures because you did not "define `rb_flo_to_i` that just calls `flo_to_i`". Please don't expose a symbol that doesn't have an appropriate prefix. 

Also, since there's a call from `flo_round` to `flo_to_i`, I thought just removing `static` from `flo_to_i` it might impact how `flo_round` is compiled, but it's up to compilers and this might not actually be the case. It's okay to rename `flo_to_i` to `rb_flo_to_i` if it seems like it's producing the same instructions in the binary.

k0kubun: > Small correction on my previous comment: the final structure turned out slightly different. flo_to_i stays static in numeric.c (the leaked-globals check requires the rb_ prefix), and I added a new public rb_flo_to_i wrapper in numeric.c that just delegates to it. ZJIT calls rb_flo_to_i directly via bindgen, so there is no wrapper in zjit.c. Net effect is the same: the truncation logic lives in one place.

yeah, https://github.com/ruby/ruby/pull/16735/changes/e882103e6509862f25d8b09625317eb745ea8935 works. thanks

khasinski: Sorry, we posted almost exactly at the same minute and I've removed mine comment because I've thought you wanted something else and didn't want to add more noise :)

If you're happy with the current solution then we're good!

khasinski:
All green and ready for review 🫡

k0kubun (approved):

tekknolagi:
I think this tanked matmul benchmark

khasinski:
I'll take a look 👀

khasinski:
Confirmed the matmul regression. The issue was that FloatAdd/Sub/Mul/Div returned `types::Float`, so when a result fed into the next Float operation, the Flonum guard was re-checked redundantly. In matmul this produced ~80M guard checks per run. Fix: #16766 (return `types::Flonum` instead).

PR #16736, opened 2026-04-13
ZJIT: Fix getlocal with level=0 reading stale EP data

khasinski:
## Summary

`YARVINSN_getlocal` with `level=0` always loaded from EP memory, but
`setlocal_WC_0` only updates the JIT FrameState without writing to EP.
When the Ruby compiler emits `getlocal` (non-WC variant) for a level=0
local after `setlocal_WC_0`, the EP load returned stale data.

The fix: when `level == 0` and locals are not invalidated, use the
FrameState value. This matches what `getlocal_WC_0` already does.

## Repro

```ruby
def test
  b = +"x"
  v = 2
  [v].pack("C*", buffer: b)
  b.size  # expected 2, got 1
end
```

Under `--zjit-call-threshold=1`, `b` was set via `setlocal_WC_0`
(FrameState only), then read via `getlocal b@0, 0` (EP load). The EP
still had nil/stale data, so `pack` received the wrong buffer VALUE.

This also fixes the flaky `btest` failure for `pack("E*", buffer:)`
with Float values that has been appearing in CI under
`--zjit-call-threshold=1`.

## Test plan

- [x] `make btest RUN_OPTS="--zjit-call-threshold=1"`: 2048/2048 PASS (was 2047)
- [x] `make zjit-test`: 1596 passed, 0 failed
- [x] `make test-all TESTS=test/ruby/test_zjit.rb`: 28 tests, 0 failures

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -7363,8 +7363,18 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                 YARVINSN_getlocal => {
                     let ep_offset = get_arg(pc, 0).as_u32();
                     let level = get_arg(pc, 1).as_u32();
-                    let ep = fun.push_insn(block, Insn::GetEP { level });
-                    state.stack_push(fun.get_local_from_ep(block, ep, ep_offset, level, types::BasicObject));
+                    if level == 0 && !local_inval {
+                        // Same optimization as getlocal_WC_0: use FrameState
+                        let val = state.getlocal(ep_offset);
+                        state.stack_push(val);
+                    } else {
+                        let ep = fun.push_insn(block, Insn::GetEP { level });
+                        let val = fun.get_local_from_ep(block, ep, ep_offset, level, types::BasicObject);
+                        if level == 0 {
+                            state.setlocal(ep_offset, val);
+                        }
+                        state.stack_push(val);
+                    }
                 }
                 YARVINSN_setlocal => {
                     let ep_offset = get_arg(pc, 0).as_u32();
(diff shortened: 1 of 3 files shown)


k0kubun:
Could you add a test in codegen_tests.rs that checks the behavior of the repro please? When an HIR test doesn't have an `assert_snapshot!` on `inspect`, it's not really useful to catch behavior-level bugs.

zjit/src/hir.rs
```diff
@@ -7363,8 +7363,18 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                 YARVINSN_getlocal => {
                     let ep_offset = get_arg(pc, 0).as_u32();
                     let level = get_arg(pc, 1).as_u32();
-                    let ep = fun.push_insn(block, Insn::GetEP { level });
-                    state.stack_push(fun.get_local_from_ep(block, ep, ep_offset, level, types::BasicObject));
+                    if level == 0 && !local_inval {
```
k0kubun, line 7366: question: When do we get `level == 0` here? Isn't `YARVINSN_getlocal_WC_0` supposed to cover all `level == 0` cases?

k0kubun: nvm, I see that the repro code does have `0014 getlocal b@0, 0`, so I guess it proves its reachability.

k0kubun (approved):
Please add a test in codegen_tests.rs, but the fix looks good to me.

khasinski:
I've added a regression test, this should be good to go and I'd be happy to rebase and test #16735 once this is merged

k0kubun (approved):

PR #16737, opened 2026-04-13
Use buffer in rb_ivar_foreach

jhawthorn:
Previously, rb_ivar_foreach would walk up the shape tree, but yield
instance variables to the callback as it went. If the object shape was
modified during this callback, particularly with removing an instance
variable, it could result in reading free'd memory.

This commit solves this by buffering all instance variable names and
values before calling the callback, giving a snapshot of the object at
the time rb_ivar_foreach is called.

The buffer is made with ALLOCV_N, so the performance difference should
be minimal, and I don't think this method is particuarly heavily used.

[[Bug #21996]](https://bugs.ruby-lang.org/issues/21996)

### DIFF
--- a/test/ruby/test_object.rb
+++ b/test/ruby/test_object.rb
@@ -970,6 +970,34 @@ class TestObject < Test::Unit::TestCase
     assert_not_include(s, "@password=")
   end
 
+  def test_inspect_too_complex
+    kernel_inspect = Kernel.instance_method(:inspect)
+
+    klasses = [
+      Class.new,
+      Class.new(String),
+      Class.new(Array),
+      Class.new(Hash),
+      Struct.new(:x),
+      Class.new(Thread::Mutex),
+      # It's very difficult to get a too_complex T_CLASS, to that isn't tested here
+    ]
+
+    klasses.each_with_index do |klass, idx|
+      8.times do |i|
+        klass.new.instance_variable_set(:"@sib_#{rand(999999)}", 1)
+      end
+
+      obj = klass.new
+      obj.instance_variable_set(:@a, 1)
+      obj.instance_variable_set(:@b, 2)
+
+      s = kernel_inspect.bind_call(obj)
+      assert_include(s, "@a=1")
+      assert_include(s, "@b=2")
+    end
+  end
+
   def test_singleton_methods
     assert_equal([], Object.new.singleton_methods)
     assert_equal([], Object.new.singleton_methods(false))
--- a/test/ruby/test_object.rb
+++ b/test/ruby/test_object.rb
@@ -970,6 +970,41 @@ class TestObject < Test::Unit::TestCase
     assert_not_include(s, "@password=")
   end
 
+  def test_inspect_mutating_ivar
+    obj = Object.new
+    evil = Object.new
+    evil.define_singleton_method(:inspect) do
+      obj.instance_variables.each { |v| obj.remove_instance_variable(v) }
+      "evil"
+    end
+    obj.instance_variable_set(:@evil, evil)
+    10.times { |i| obj.instance_variable_set(:"@v#{i}", 0) }
+    # Buffered iteration: inspect sees a snapshot of the original ivars
+    result = obj.inspect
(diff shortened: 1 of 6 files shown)


byroot:
> If the object shape was modified during this callback, particularly with removing an instance variable, it could result in reading free'd memory.

I'll keep reading the diff, and try the repro, but I'm trying to understand this part.

The shape tree is never freed, so I don't get what freed memory we're accessing, clearly, not shape memory right?

test/ruby/test_object.rb
```diff
@@ -970,6 +970,69 @@ def initialize
...
+    kernel_inspect = Kernel.instance_method(:inspect)
+
+    klasses = [
+      Class.new,
+      Class.new(String),
+      Class.new(Array),
+      Class.new(Hash),
+      Struct.new(:x),
+      Class.new(Thread::Mutex),
+      # It's very difficult to get a too_complex T_CLASS, to that isn't tested here
```
byroot:
```suggestion
      # It's very difficult to get a too_complex T_CLASS, so that isn't tested here
```

test/ruby/test_object.rb
```diff
@@ -970,6 +970,69 @@ def initialize
     assert_not_include(s, "@password=")
   end
 
+  def test_inspect_mutating_ivar
```
byroot, line 973: It might be easier to write these tests in `test_shape.rb` which already has a number of helpers etc to create complex shapes and such.

variable.c
```diff
@@ -2353,10 +2362,39 @@ rb_field_foreach(VALUE obj, rb_ivar_foreach_callback_func *func, st_data_t arg,
...
+
 void
 rb_ivar_foreach(VALUE obj, rb_ivar_foreach_callback_func *func, st_data_t arg)
 {
-    rb_field_foreach(obj, func, arg, true);
+    st_index_t count = rb_ivar_count(obj);
+    if (count == 0) return;
+
+    VALUE tmpbuf;
+    struct ivar_buf_entry *buf = ALLOCV_N(struct ivar_buf_entry, tmpbuf, count);
```
byroot, line 2393: What worries me slightly here is that `rb_ivar_foreach` now allocate when called on objects with over 512 ivars, which we're very unlikely to have coverage for, so we could end up calling it in a context were allocations aren't allowed without realizing.

e.g. not sure that's actually the case there, but we call in `ractor.c: obj_traverse_i`.

Perhaps it would be better to keep `rb_ivar_foreach` as is was, and instead add a `_safe` variant? (perhaps not worth it?)

byroot:
> This commit solves this by buffering all instance variable names and values before calling the callback, giving a snapshot of the object at the time rb_ivar_foreach is called.

Maybe I'm dense, but I don't understand how this changed. I get that now we're buffering the list of ivars, but previously `rb_shape_foreach_field(` would basically have the same effect because it recursively walk up the shape tree, but then call the callbacks on its way back. So effectively it already "snapshot" the list of ivars.

Clearly the fix works, I just don't get how.

variable.c
```diff
@@ -2353,10 +2362,39 @@ rb_field_foreach(VALUE obj, rb_ivar_foreach_callback_func *func, st_data_t arg,
...
+    ID name;
+    VALUE val;
+};
+
+static int
+collect_ivar_i(ID id, VALUE val, st_data_t arg)
+{
+    struct ivar_buf_entry **pos = (struct ivar_buf_entry **)arg;
+    (*pos)->name = id;
+    (*pos)->val = val;
```
byroot, line 2375: 🤦 I get it now. We're not just collecting ivar names, but also values. That explains it.

byroot (approved):

variable.c
jhawthorn: Yeah, the values are the problem, which feels a bit silly for the places this is used that don't need the key.

byroot: > which feels a bit silly for the places this is used that don't need the key.

Right. Perhaps a `ivar_foreach_value` would make sense? This way copying values on the stack can be a simple `memcpy`.

PR #16746, opened 2026-04-15
ZJIT: Use SP for reloading locals

### LINKED ISSUE #769: ZJIT: Consider using SP register for reloading locals
send-ish instructions reload locals from the VM stack when they have a blockiseq. The reload uses `cfp->ep` to load stack values, but it could use `SP` register, which should be faster than loading `cfp->ep` from memory, when `efp->ep` is equal to `SP - 1`. It needs to accompany a PatchPoint `Invariant::NoEPEscape`.

nozomemein:
Closes: https://github.com/Shopify/ruby/issues/769

Use `SP` for reloading locals after block-backed send-ish instructions when EP has not escaped.
To make that safe, add a `NoEPEscape` patch point before the reload and use a post-send snapshot for side exits, so exiting does not re-execute the send in the interpreter.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -4678,6 +4678,17 @@ impl Function {
         self.push_insn(block, Insn::PatchPoint { invariant: Invariant::MethodRedefined { klass: recv_class, method: method_id, cme }, state });
     }
 
+    /// Side exit back to the state after a block-backed send.
+    /// Using the pre-send snapshot would re-execute the send in the interpreter.
+    fn gen_post_send_no_ep_escape_patch_point(&mut self, block: BlockId, state: &FrameState, insn_idx: u32) {
+        let iseq = state.iseq;
+        let mut reload_state = state.clone();
+        reload_state.insn_idx = insn_idx as usize;
+        reload_state.pc = unsafe { rb_iseq_pc_at_idx(iseq, insn_idx) };
+        let reload_exit_id = self.push_insn(block, Insn::Snapshot { state: reload_state.without_locals() });
+        self.push_insn(block, Insn::PatchPoint { invariant: Invariant::NoEPEscape(iseq), state: reload_exit_id });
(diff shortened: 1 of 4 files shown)


nozomemein:
I found an issue related to this PR tht also reproduces on master.
Filed: https://github.com/Shopify/ruby/issues/976

zjit/src/hir.rs
```diff
@@ -8101,13 +8087,29 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                         // Reload locals that may have been modified by the blockiseq.
                         // TODO: Avoid reloading locals that are not referenced by the blockiseq
                         // or not used after this. Max thinks we could eventually DCE them.
-                        let ep = fun.push_insn(block, Insn::GetEP { level: 0 });
+                        if !ep_escaped && !state.locals.is_empty() {
+                            // Side exit back to the state after the block-backed send.
+                            // Using the pre-send snapshot would re-execute the send in the interpreter.
+                            let mut reload_state = state.clone();
+                            reload_state.insn_idx = insn_idx as usize;
+                            reload_state.pc = unsafe { rb_iseq_pc_at_idx(iseq, insn_idx) };
```
k0kubun: The `reload_state` construction logic seems to be repeated 4 times. Can we define a function and use it to deduplicate them?

nozomemein: done 👍 
https://github.com/ruby/ruby/pull/16746/changes/d609a4864e4255595967de3b945d1000bdc46de6

tekknolagi:
Hello and thank you for your PR! Many ZJIT members are at RubyKaigi this week and therefore will be slow to respond due to the conference and the travel associated with it. We will take a look soon!

k0kubun (approved):
The HIR optimization seems right to me. Thank you!

PR #16751, opened 2026-04-15
[DOC] Doc for Set.new

BurdetteLamar:


set.c
```diff
@@ -476,22 +476,29 @@ set_initialize_with_block(RB_BLOCK_CALL_FUNC_ARGLIST(i, set))
 }
 
 /*
- *  call-seq:
- *    Set.new -> new_set
- *    Set.new(enum) -> new_set
- *    Set.new(enum) { |elem| ... } -> new_set
+ * call-seq:
+ *   Set.new(object) -> new_set
```
jeremyevans: You can call Set.new without an argument:

```ruby
Set.new # => Set[]
```

```suggestion
 *   Set.new -> new_set 
 *   Set.new(object) -> new_set
```

set.c
```diff
@@ -476,22 +476,29 @@ set_initialize_with_block(RB_BLOCK_CALL_FUNC_ARGLIST(i, set))
...
+ *   Set.new(object) {|element| ... } -> new_set
+ *
+ * Returns a new set based on the given +object+,
+ * which must be an Enumerable.
```
jeremyevans: An argument isn't required, and Set.new supports an explicit nil argument:

```ruby
Set.new(nil) # => Set[]
```

```suggestion
 * which if given and not nil, must be an Enumerable.
```

set.c
BurdetteLamar: I think I can say that +object+ defaults to +nil+?  Made it so.

set.c
BurdetteLamar: Ditto.

set.c
```diff
@@ -476,22 +476,37 @@ set_initialize_with_block(RB_BLOCK_CALL_FUNC_ARGLIST(i, set))
...
- *    Set.new(enum) { |elem| ... } -> new_set
+ * call-seq:
+ *   Set.new(object = nil) -> new_set
+ *   Set.new(object = nil) {|element| ... } -> new_set
+ *
+ * Returns a new \Set object based on the given +object+,
+ * which must be an Enumerable or +nil+.
+ *
+ * With no argument given,
+ * returns a new empty \Set object:
```
jeremyevans: Maybe we should say, "when the argument is `nil`"? The `call-seq` shows `nil` is the default value of the argument, so that also handles the no argument case. We can keep the examples below the same, since the nil value is more likely to be implicit than explicit.

set.c
```diff
@@ -476,22 +476,37 @@ set_initialize_with_block(RB_BLOCK_CALL_FUNC_ARGLIST(i, set))
...
- *  Creates a new set containing the elements of the given enumerable
- *  object.
+ * With a block given and +object+ given as an Enumerable,
+ * calls the block with each element of +object+;
+ * adds the block's return value to the new set:
  *
- *  If a block is given, the elements of enum are preprocessed by the
- *  given block.
+ *   Set.new(4..10) {|i| i * 2 }      # => Set[8, 10, 12, 14, 16, 18, 20]
+ *   Set.new { fail 'Cannot happen' } # => Set[]  # Block not called.
```
jeremyevans: This same example is now used above:

```suggestion
```

jeremyevans (approved):

set.c
BurdetteLamar: I think I have this right, now.

set.c
BurdetteLamar: Removed.

jeremyevans (approved):

peterzhu2118 (approved):

PR #16755, opened 2026-04-15
ZJIT: Stabilize layout-sensitive ivar opt tests

nozomemein:
Currently, some `getivar` / `setivar` snapshot tests produce different HIR between `dev` and `stats` builds.

This mostly comes from build-dependent object layout details. `dev` enables `RUBY_DEBUG`, which changes effective object capacity. 

As a result:
- `getivar` tests can differ depending on whether the profiled receiver is embedded or heap-backed
- `setivar` tests can differ depending on where the capacity boundary is crossed

This change stabilizes those tests by:
- using clearly heap-backed receivers for the polymorphic `getivar` cases
- rewriting the `setivar` shape-transition test to cover the intended specialized path without depending on a capacity boundary
- keeping the heap-upgrade `setivar` snapshot only on builds where that path naturally occurs

(The details are described in each commit.)

Ideally we may also want to have CI coverage for `stats` builds, but that can be done separately in a follow-up PR.


zjit/src/hir/opt_tests.rs
```diff
@@ -7754,12 +7777,13 @@ mod hir_opt_tests {
         set_call_threshold(3);
         eval(r#"
             class C
-              def foo_then_bar
+              def foo_then_many
```
nozomemein, line 7929: We’ve already covered the embedded case in the other tests like `test_optimize_getivar_embedded`, so the tests for polymorphic case use heap-backed receivers instead to avoid depending on the embedded/heap boundary.

zjit/src/hir/opt_tests.rs
```diff
@@ -14953,17 +14978,33 @@ mod hir_opt_tests {
...
-        assert_snapshot!(hir_string("test"), @"
-        fn test@<compiled>:3:
+        // Fill the receiver to the embedded-capacity boundary in initialize so
+        // the first write overflows into the heap and the next write can use
+        // the upgraded heap-backed self type.
+        // See gc/default/default.c for GC::INTERNAL_CONSTANTS and
+        // shape.c::Init_default_shapes for the embedded-capacity calculation
+        // that backs shape capacities.
+        eval(r#"
+            word_size = [0].pack("J").bytesize
```
nozomemein: We can't use `RbConfig::SIZEOF["void*"]` here, as `/test/ruby/test_object.rb` does, because `RbConfig` is not available under `miniruby`. (Not sure if this is right way though.)

zjit/src/hir/opt_tests.rs
```diff
@@ -5557,15 +5557,38 @@ mod hir_opt_tests {
 
     #[test]
     fn test_specialize_multiple_monomorphic_setivar_with_shape_transition() {
-        eval("
-            def test
-              @foo = 1
-              @bar = 2
+        // Compute the embedded-capacity boundary for this build so @foo stays
```
XrXr: I'm surprised that this test was relying on crossing embedded capacity boundary. Two ivars, two different shapes, no? Maybe you can rewrite the test to naturally work on both builds without relying on `::GC`.

XrXr (commented):
This is very tweaky code to get it working on stats build that takes away from readability. These tests exist to show that we can handle objects of various kinds, and them failing on some build kinds shows more a problem with the tests rather than implementation. I'd be happy with an understandable test that passes on one kind of build.

I think it'd be better to have some self consistency check like `obj.embedded?` and give up testing on niche builds rather than trying to pass on all of them. You could do something like:

```rust
let obj = eval(...);
if !obj.embedded_p() {
    return; // test not suitable for this build.
}
```

zjit/src/hir/opt_tests.rs
nozomemein: Yes, fixed it. I rewrote the test without relying embedded capacity boundary. Thanks!
commit: https://github.com/ruby/ruby/pull/16755/changes/bb4ef6b0d9698769d08c4f241263920cd49d82d6

nozomemein:
@XrXr 
Thank you for the review! That totally makes sense to me.
I pushed some changes to apply your suggestions 🙏 

- https://github.com/ruby/ruby/pull/16755/changes/bb4ef6b0d9698769d08c4f241263920cd49d82d6
- https://github.com/ruby/ruby/pull/16755/changes/9e212cbbde5e1ac2e8a3c08633b2f2146865d96d

zjit/src/hir/opt_tests.rs
```diff
@@ -15023,17 +15039,46 @@ mod hir_opt_tests {
...
+        // Make sure the next write is the one that overflows into heap-backed
+        // storage, so this snapshot still exercises the self-type upgrade path.
+        let probe = eval(r#"
+            probe = OBJ.class.new
+            probe.instance_variable_set(:@overflow, 1)
+            probe
+        "#);
+        if probe.embedded_p() {
+            return;
+        }
```
nozomemein, line 15166: Note that `upgrade_self_type_to_heap_after_setivar` may now silently skip on builds where the setup does not hit the heap-upgrade path (e.g. if the effective object capacity changes in debug builds). Failing explicitly here would require checking more internal layout details.

XrXr (approved):

k0kubun (approved):

PR #16760, opened 2026-04-16
Update default gemspecs so that default gems can run

nobu:


st0012:
I think the `require "rbs"` in this PR doesn't work as expected in [doc building workflow](https://github.com/ruby/actions/actions/runs/25895616458/job/76107951475), though it doesn't affect me running `make html` locally. Do you know what's the best way to fix this? Should it be a Ruby side fix or the workflow side 🤔

PR #16762, opened 2026-04-16
Use RB_OBJ_FROZEN_RAW in hash.c instead of RB_OBJ_FROZEN

luke-gruber:


luke-gruber:
I just checked assembler and it gets optimized away anyway, closing

PR #16763, opened 2026-04-16
Remove check for special const in rb_obj_freeze

luke-gruber:
This was introduced a long time ago when lots of internal changes were made to special consts (cc3088ea519).


luke-gruber:
On second thought it's probably nice for misbehaving extensions to get this bug message instead of a random crash. Closing.

PR #16766, opened 2026-04-17
ZJIT: Return Flonum instead of Float from Float arithmetic

khasinski:
## Summary

FloatAdd/Sub/Mul/Div returned `types::Float` (union of Flonum and
HeapFloat). When the result fed into another Float operation, the
Flonum guard was re-checked even though the value was almost
certainly a Flonum. In matmul (200x200), this added ~80M redundant
guard checks per run.

Fix: return `types::Flonum` instead. On 64-bit with USE_FLONUM,
arithmetic results are Flonum for the vast majority of values. If a
result is HeapFloat, the next guard will side-exit to the
interpreter (correct behavior).

## Benchmark

matmul on x86_64 (10 runs, 5 iterations each, best of 10):

| Version | Best | Median |
|---------|------|--------|
| Before #16735 (no Float ops) | 0.487s | 0.583s |
| After #16735 (Float return) | 0.631s | 0.781s |
| This fix (Flonum return) | 0.433s | 0.504s |

## Test plan

- [x] `make zjit-test`: 1611 passed
- [x] `make test-all TESTS=test/ruby/test_zjit.rb`: 29 tests, 0 failures
- [x] `make btest RUN_OPTS="--zjit-call-threshold=1"`: 2048/2048 PASS
- [x] Verified on x86_64 (Ubuntu)


k0kubun:
> If a result is HeapFloat, the next guard will side-exit to the interpreter (correct behavior).

Who generates the next guard? Isn't the return type of `Flonum` supposed to eliminate the redundant type check?

khasinski:
I think that the guard comes from `coerce_to` in `try_inline_float_op`. When the input is a previous FloatMul result, returning `types::Flonum` makes `is_a(val, Flonum)` true so `coerce_to` skips the guard. That is the redundant check this eliminates. If the result happens to be HeapFloat, it still passes through to `rb_float_plus` etc. correctly since those handle both representations.

zjit/src/hir.rs
```diff
@@ -3138,10 +3138,10 @@ impl Function {
...
             Insn::FixnumDiv  { .. } => types::Fixnum,
             Insn::FixnumMod  { .. } => types::Fixnum,
-            Insn::FloatAdd   { .. } => types::Float,
-            Insn::FloatSub   { .. } => types::Float,
-            Insn::FloatMul   { .. } => types::Float,
-            Insn::FloatDiv   { .. } => types::Float,
+            Insn::FloatAdd   { .. } => types::Flonum,
+            Insn::FloatSub   { .. } => types::Flonum,
+            Insn::FloatMul   { .. } => types::Flonum,
+            Insn::FloatDiv   { .. } => types::Flonum,
```
k0kubun, line 3144: Fixnum operations check for overflow and thus guarantee their correctness of their return type (`Fixnum`), but it's not the case for Flonum operations; the C functions we use for these instructions do not guarantee the return type is `Flonum`.

This seems prone to bugs. You're saying in https://github.com/ruby/ruby/pull/16766#issuecomment-4270794053 that it's fine for the things you tested because these Float functions handle both Flonum and non-Flonum Float, but it doesn't necessarily work for instructions that expect a Flonum and don't handle non-Flonum Float if we introduce such a function in the future.

Could we avoid lying the return type?

khasinski: Good point, claiming `Flonum` only works today because every consumer happens to tolerate `HeapFloat`. 🤔 

What if I switch this PR to inline the Flonum arithmetic in `gen_float_add`/etc. (untag, FP op, check encodes as Flonum, retag or side-exit)? That makes the `Flonum` return type honest and drops the `rb_float_*` ccall entirely, so the perf should actually improve over this version.

It's probably a _much_ bigger PR though

khasinski: UPDATE:

I started looking at proper inline Flonum codegen and it's a much bigger lift than I'd hoped, zjit's LIR has no FP register class or FP ops yet. That's a larger PR and I'd rather not block this perf fix on it.

A post-call tag check would make the return type honest but it might costs roughly the same as the guard we're removing, so the perf win disappears.

Would you accept this PR if I add a structural guard against the footgun you raised, e.g. an HIR validator rule that any Flonum-requiring consumer of `FloatAdd/Sub/Mul/Div` output must be on an explicit `tolerates_heap_float` allow-list (currently just the other Float arith ops). That way if someone adds a true Flonum-only instruction later, the build fails loudly instead of silently miscompiling.

I can work on the proper FP inlining in a separate PR.

tekknolagi: I'm offline until next week but will register the opinion that I don't think we should do anything hacky at this stage. It's not worth it. Code lives a long time and lies eventually stagnate and become part of life. I'd much rather gradually build up to float ops, but that's a ways off. As a team, we don't have any plan to do that immediately (we're focusing on general Ruby and web apps). That's not to say we shouldn't have support for float ops, but I'd want to very seriously investigate how other runtimes do this, what parts of the runtime get dramatically more complicated, etc.

khasinski: I'll close this and investigate a proper fix for float ops inlining then. Thank you all for the thorough reviews 🙇

PR #16767, opened 2026-04-17
ZJIT: Iterate infer_types to a fixpoint for cross-block type widening

nirvdrum:
infer_types uses a BFS worklist to compute types. Blocks are re-enqueued only when their own block parameters' types change. This assumes a block's instruction types depend only on its parameters and instructions earlier in the same block.

That invariant does not hold when an instruction references a value defined in a dominator block via SSA rather than through a block parameter. In that case, widening the dominator's value type does not trigger re-processing of the downstream block, leaving stale (too-narrow) types that cause fold_constants to incorrectly eliminate live control-flow edges.

Fix this by wrapping the BFS in an outer fixpoint loop. When any instruction's type changes during a pass, re-seed the worklist with every reachable block and iterate again. The type lattice has finite height and types only widen, so this terminates.

The bug is latent on master because the HIR builder threads cross-block values through block parameters (via merge_stack). It is exposed by inlining, which creates CFG shapes where caller-defined values are consumed inside inlined continuations without being passed as block parameters.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -3216,9 +3216,6 @@ impl Function {
             };
         }
 
-        reachable.insert(self.entries_block);
-        worklist_add!(self.entries_block);
-
         // Helper to propagate types along a branch edge and enqueue the target if anything changed
         macro_rules! enqueue {
             ($self:ident, $target:expr) => {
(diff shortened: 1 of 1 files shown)


nirvdrum:
It turns out #16711 fixed this issue. The test from this PR shows the problem with reading values from the dominator, but I can pull that into the inliner work.

PR #16769, opened 2026-04-18
ZJIT: Guard stack_topn and profile_stack against stack underflow

nirvdrum:
`FrameState::stack_topn` computed its backing index as `self.stack.len() - idx - 1` before checking whether `idx` fell within bounds. When `idx` was greater than or equal to the stack length, the subtraction on `usize` underflowed and panicked, even though the function was supposed to surface the condition as a `ParseError::StackUnderflow`. The panic aborts the Ruby process, so any code path that hit this bug took the interpreter down with it rather than allowing `iseq_to_hir` to fail gracefully.

`ProfileOracle::profile_stack` then unwrapped the `stack_topn` result with `.expect("Unexpected stack underflow in profiling")`, on the premise that the number of profiled operands for an instruction can never exceed the depth of the compile-time virtual stack at that point. That invariant holds for the ISEQs ZJIT normally sees by call-threshold instrumentation, but it does not hold for every ISEQ the inliner hands to `iseq_to_hir`. Once a profiled callee with an operand mismatch reached `profile_stack`, the `expect` fired and crashed the process.

Both failures were encountered while running the liquid-render benchmark with inlining enabled, where the inliner compiles more ISEQs than the normal profiling path would.

Rework `stack_topn` to use `checked_sub` so it returns `ParseError::StackUnderflow` cleanly when the requested index is out of range, and change `profile_stack` to stop mapping profile entries at the first unmappable index instead of panicking. The compile may omit some profile-driven type information in this rare case, but the compilation completes and the rest of the pass can proceed.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -6549,7 +6549,9 @@ impl FrameState {
 
     /// Get a stack operand at idx
     fn stack_topn(&self, idx: usize) -> Result<InsnId, ParseError> {
-        let idx = self.stack.len() - idx - 1;
+        let Some(idx) = self.stack.len().checked_sub(idx + 1) else {
+            return Err(ParseError::StackUnderflow(self.clone()));
+        };
         self.stack.get(idx).ok_or_else(|| ParseError::StackUnderflow(self.clone())).copied()
     }
 
@@ -6717,10 +6719,14 @@ impl ProfileOracle {
         let iseq_insn_idx = state.insn_idx;
         let Some(operand_types) = self.payload.profile.get_operand_types(iseq_insn_idx) else { return };
         let entry = self.types.entry(iseq_insn_idx).or_default();
-        // operand_types is always going to be <= stack size (otherwise it would have an underflow
-        // at run-time) so use that to drive iteration.
+
+        // operand_types is generally <= stack size (otherwise it would underflow at run-time),
+        // but the inliner can reach this code by invoking iseq_to_hir on callee ISEQs whose
+        // profile was collected while the runtime stack held values our compile-time virtual
+        // stack does not yet carry. Skip any profile entries we cannot map without failing the
+        // whole compile.
         for (idx, insn_type_distribution) in operand_types.iter().rev().enumerate() {
-            let insn = state.stack_topn(idx + stack_offset).expect("Unexpected stack underflow in profiling");
+            let Ok(insn) = state.stack_topn(idx + stack_offset) else { break };
             entry.push((insn, TypeDistributionSummary::new(insn_type_distribution)))
         }
     }


zjit/src/hir.rs
```diff
@@ -6549,7 +6549,9 @@ impl FrameState {
 
     /// Get a stack operand at idx
     fn stack_topn(&self, idx: usize) -> Result<InsnId, ParseError> {
-        let idx = self.stack.len() - idx - 1;
+        let Some(idx) = self.stack.len().checked_sub(idx + 1) else {
```
tekknolagi, line 6552: Can you do something like `let idx = ...checked_sub()...ok_or(Underflow)?;`

zjit/src/hir.rs
```diff
@@ -6717,10 +6719,14 @@ impl ProfileOracle {
         let iseq_insn_idx = state.insn_idx;
         let Some(operand_types) = self.payload.profile.get_operand_types(iseq_insn_idx) else { return };
         let entry = self.types.entry(iseq_insn_idx).or_default();
-        // operand_types is always going to be <= stack size (otherwise it would have an underflow
-        // at run-time) so use that to drive iteration.
+
+        // operand_types is generally <= stack size (otherwise it would underflow at run-time),
+        // but the inliner can reach this code by invoking iseq_to_hir on callee ISEQs whose
+        // profile was collected while the runtime stack held values our compile-time virtual
```
tekknolagi, line 6725: This comment does not make a whole lot of sense to me & also talks about a feature we have not landed yet. This should *always* hold, or we should not have profiles for that opcode. What is going amiss?

nirvdrum: You're right. I was seeing SIGABRT with the inliner branch on the _liquid-render_ benchmarks related to a panic in `stack_topn` (`attempt to subtract with overflow`). I tried to work out reasons why that might happen and had convinced myself an assumption made during HIR construction could be different from run time state.

Your comment prompted me to take a second look and it turns out there was an unrelated bug in the inliner that was to blame. I've hit a few `JITState#iseq` vs `FrameState#iseq` that led to SIGABRT. I thought I got them all, but I missed one in `side_exit_with_recompile`. So, a lot of the analysis here is wrong.

The PR is in a weird state now. The reason I opened it was misguided, but the issue it fixes seems like a clear logic error. I could buy that we want to panic on stack underflow, but `stack_topn` returns a `Result` so an `Err` seems to be what's expected. How do you want to proceed? Close the PR and continue to panic in these error cases or fix up `stack_topn`?

tekknolagi: Let's just fix `stack_topn` for now and let caller(s) panic. We can figure out what to do elsewhere later

k0kubun:
> The panic aborts the Ruby process, so any code path that hit this bug took the interpreter down with it rather than allowing iseq_to_hir to fail gracefully.

Is it possible to add a test that would fail without this fix, or does it require the inliner?

zjit/src/hir.rs
tekknolagi: Can we either reduce the scope of this PR or close?

tekknolagi: Closing for now but re-open with narrowed scope if/when you have time

PR #16774, opened 2026-04-20
ZJIT: Remove `JITState#iseq`

nirvdrum:
While working on implementing a new inliner for ZJIT, I keep running into an issue where the wrong iseq pointer is being used. Since `JITState#iseq` and `FrameState#iseq` end up being the same value in the absence of inlining, I don't think we've been all that rigorous about using the "correct" one in a given context. I've been migrating to `FrameState#iseq` in my inliner work and adapting on each rebase. But, having looked at this code for a little while, I think we can remove `JITState#iseq`.

This PR switches to using `FrameState#iseq` in cases where we're working with the code associated with a specific frame (e.g., local-table layout, side-exit resumption target) rather than the overall compilation unit. Other cases are cleaned up by either storing the iseq on the instruction or computing the value that required the iseq and storing that in the instruction.

Without inlining, this change set is really just a clean-up of what we have today. With inlining, however, these changes become semantically meaningful. Making the change now removes the decision point of when to use `JITState#iseq` vs `FrameState#iseq` and simplifies keeping the inliner work current with _master_.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -2854,7 +2854,7 @@ fn gen_spill_locals(jit: &JITState, asm: &mut Assembler, state: &FrameState) {
     gen_incr_counter(asm, Counter::vm_write_locals_count);
     asm_comment!(asm, "spill locals");
     for (idx, &insn_id) in state.locals().enumerate() {
-        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(jit.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id));
+        asm.mov(Opnd::mem(64, SP, (-local_idx_to_ep_offset(state.iseq, idx) - 1) * SIZEOF_VALUE_I32), jit.get_opnd(insn_id));
     }
 }
 
@@ -3028,7 +3028,7 @@ fn side_exit(jit: &JITState, state: &FrameState, reason: SideExitReason) -> Targ
 fn side_exit_with_recompile(jit: &JITState, state: &FrameState, reason: SideExitReason, recompile: Option<Recompile>) -> Target {
     let mut exit = build_side_exit(jit, state);
     exit.recompile = recompile.map(|strategy| SideExitRecompile {
-        iseq: Opnd::Value(VALUE::from(jit.iseq)),
+        iseq: Opnd::Value(VALUE::from(state.iseq)),
         insn_idx: state.insn_idx() as u32,
         strategy,
     });
@@ -3051,7 +3051,7 @@ fn build_side_exit(jit: &JITState, state: &FrameState) -> SideExit {
         pc: Opnd::const_ptr(state.pc),
         stack,
         locals,
-        iseq: jit.iseq,
+        iseq: state.iseq,
         recompile: None,
     }
 }
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -748,7 +748,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::SideExit { state, reason, recompile } => no_output!(gen_side_exit(jit, asm, reason, *recompile, &function.frame_state(*state))),
         Insn::PutSpecialObject { value_type } => gen_putspecialobject(asm, *value_type),
         Insn::AnyToString { val, str, state } => gen_anytostring(asm, opnd!(val), opnd!(str), &function.frame_state(*state)),
-        Insn::Defined { op_type, obj, pushval, v, state } => gen_defined(jit, asm, *op_type, *obj, *pushval, opnd!(v), &function.frame_state(*state)),
+        Insn::Defined { op_type, obj, pushval, v, lep_level, state } => gen_defined(jit, asm, *op_type, *obj, *pushval, opnd!(v), *lep_level, &function.frame_state(*state)),
         Insn::CheckMatch { target, pattern, flag, state } => gen_checkmatch(jit, asm, opnd!(target), opnd!(pattern), *flag, &function.frame_state(*state)),
         Insn::GetSpecialSymbol { symbol_type, state: _ } => gen_getspecial_symbol(asm, *symbol_type),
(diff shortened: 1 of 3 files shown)


zjit/src/codegen.rs
```diff
@@ -733,22 +729,22 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
         Insn::CCallVariadic { cfunc, recv, name, args, cme, state, block, return_type: _, elidable: _ } => {
             gen_ccall_variadic(jit, asm, *cfunc, *name, opnd!(recv), opnds!(args), *cme, *block, &function.frame_state(*state))
         }
-        Insn::GetIvar { self_val, id, ic, state: _ } => gen_getivar(jit, asm, opnd!(self_val), *id, *ic),
+        Insn::GetIvar { self_val, id, ic, iseq, state: _ } => gen_getivar(asm, opnd!(self_val), *id, *ic, *iseq),
```
tekknolagi: Pull it from the FrameState instead

zjit/src/codegen.rs
```diff
@@ -733,22 +729,22 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
...
         &Insn::GetBlockParam { ep_offset, level, state } => gen_getblockparam(jit, asm, ep_offset, level, &function.frame_state(state)),
         &Insn::SetLocal { val, ep_offset, level } => no_output!(gen_setlocal(asm, opnd!(val), function.type_of(val), ep_offset, level)),
         Insn::GetConstant { klass, id, allow_nil, state } => gen_getconstant(jit, asm, opnd!(klass), *id, opnd!(allow_nil), &function.frame_state(*state)),
         Insn::GetConstantPath { ic, state } => gen_get_constant_path(jit, asm, *ic, &function.frame_state(*state)),
-        Insn::GetClassVar { id, ic, state } => gen_getclassvar(jit, asm, *id, *ic, &function.frame_state(*state)),
-        Insn::SetClassVar { id, val, ic, state } => no_output!(gen_setclassvar(jit, asm, *id, opnd!(val), *ic, &function.frame_state(*state))),
-        Insn::SetIvar { self_val, id, ic, val, state } => no_output!(gen_setivar(jit, asm, opnd!(self_val), *id, *ic, opnd!(val), &function.frame_state(*state))),
+        Insn::GetClassVar { id, ic, iseq, state } => gen_getclassvar(jit, asm, *id, *ic, *iseq, &function.frame_state(*state)),
```
tekknolagi: Pull from FrameState instead (here and anything that has `state`)

zjit/src/hir.rs
```diff
@@ -7174,7 +7180,12 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                         // Similar to gen_is_block_given
                         Insn::Const { val: Const::Value(Qnil) }
                     } else {
-                        Insn::Defined { op_type, obj, pushval, v, state: exit_id }
+                        let lep_level = if op_type == DEFINED_YIELD as usize {
```
tekknolagi, line 7185: Can you copy the old (now deleted) comment/assert here?

tekknolagi (approved):
lgtm if you use FrameState

zjit/src/hir.rs
nirvdrum: I've added a new comment. The way the branches are set up in `iseq_to_hir`, I don't think the assertion is needed. It's asserting a situation that always holds.

zjit/src/codegen.rs
```diff
@@ -796,8 +792,7 @@ fn gen_insn(cb: &mut CodeBlock, jit: &mut JITState, asm: &mut Assembler, functio
 
 /// Gets the EP of the ISeq of the containing method, or "local level".
 /// Equivalent of GET_LEP() macro.
-fn gen_get_lep(jit: &JITState, asm: &mut Assembler) -> Opnd {
-    let level = get_lvar_level(jit.iseq);
+fn gen_get_lep(asm: &mut Assembler, level: u32) -> Opnd {
```
rwstauner: I wonder if we still want this function now that it is just a pass through since the level is being provided.  There is only one caller (which could just as easily call gen_get_ep directly).

nirvdrum: Yeah, I think that makes sense. Initially, I was thinking the name `gen_get_lep` was intention-revealing, but we can do that with the name `lep_level`. I'll remove it.

zjit/src/hir.rs
rwstauner, line 2137: Should we add the lep_level to the Display?

rwstauner: Maybe it's not useful, I now realize there are a bunch of fields we don't display here.

rwstauner, line 4116: Is `fun.iseq` the right iseq to use here, or should it come from `state`?

rwstauner, line 8156: Same question.  What is `fun.iseq` set to at this point?  Should we get it from state?

rwstauner: Trying to understand the context this line will have under the inliner.
Will this code be possibly building an inlined function, such that it would need to check frame state and not the original iseq?

rwstauner: I see that after this block ends (line 4140) we get the frame state and use its iseq (compared to the fun.iseq being used here).

zjit/src/hir.rs
tekknolagi: ok ty

zjit/src/hir.rs
tekknolagi: I think you are right, here and elsewhere, should come from FrameState

nirvdrum: This is a good catch, but I think it's out of scope for this PR. Here, I'm just trying to clean up usages of `JITState#iseq`. The inliner work has required other updates to iseq references. I'll try to pull those out as other PRs to make reviewing them easier.

nirvdrum: I think you're right, but I'm going to handle that in a separate changeset.

tekknolagi (approved):
very nice thank you

PR #16780, opened 2026-04-21
Add local_iseq to glossary

rwstauner:
By name I was trying to determine how the local_iseq was more local than the one being worked on... but in actuality it goes in the other direction because that's not what "local" means in this context.

### DIFF
--- a/doc/contributing/glossary.md
+++ b/doc/contributing/glossary.md
@@ -32,6 +32,7 @@ Just a list of acronyms I've run across in the Ruby source code and their meanin
 | JIT | Just In Time compiler |
 | `lep` | Local Environment Pointer. An `ep` which is tagged `VM_ENV_FLAG_LOCAL`. Usually this is the `ep` of a method (rather than a block, whose `ep` isn't "local") |
 | `local` | Local. Refers to a local variable. |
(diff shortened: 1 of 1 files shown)


nirvdrum (approved):
Thanks. I think this is helpful.

doc/contributing/glossary.md
```diff
@@ -32,6 +32,7 @@ Just a list of acronyms I've run across in the Ruby source code and their meanin
 | JIT | Just In Time compiler |
 | `lep` | Local Environment Pointer. An `ep` which is tagged `VM_ENV_FLAG_LOCAL`. Usually this is the `ep` of a method (rather than a block, whose `ep` isn't "local") |
 | `local` | Local. Refers to a local variable. |
+| `local_iseq` | Nearest enclosing `iseq` that owns a local-variable scope. Found by walking up the parent chain (`parent_iseq`) and skipping block iseqs (`ISEQ_TYPE_BLOCK`) until a non-block iseq (method, top, class, eval, etc.) is reached. For a method iseq, the local iseq is itself; for a block, it's the method (or other non-block scope) the block was defined in. Used to resolve local-variable slots, decide what a bare `return` returns from, and similar scope-level questions. |
```
jhawthorn, line 35: I think this might deserve a little more nuance, because don't blocks also have a local variable scope with their own slots? Maybe the distinction is that they share variables _up to_ the "local iseq"?

The case I've usually found needing the local iseq is block parameters and svars

PR #16783, opened 2026-04-22
ZJIT: Add a YarvInsnIdx type for clarity

nirvdrum:
While working with `ProfileOracle`, I found the `types` field definition of `HashMap<usize, Vec<(InsnId, TypeDistributionSummary)>>` was a little obtuse to work with. There's really no indication of what that `usize` is supposed to represent. The type alias is more intention-revealing for no extra cost. There's no additional type safety from having this alias, but I think it makes the code easier to read.


tenderlove (approved):
I like this. I prefer to use more descriptive types when possible, especially when there's no overhead.

PR #16779, opened 2026-04-21
Fix passing frozen_string_literal option to #compile_file backed by Prism

palkan:
There is a regression in `RubyVM::InstructionSequence.compile_file` when Prism is used under the hood (i.e., `iseqw_s_compile_file_prism`). 

Given the following file (`a.rb`):

```ruby
# a.rb
"foo"
```

We can see that the `frozen_string_literal` compile flag is not recognized:

```sh
$ ./ruby -v
ruby 4.1.0dev (2026-04-21T07:40:39Z master 36a22e04b7) +PRISM [aarch64-linux]

$ ./ruby -e "puts RubyVM::InstructionSequence.compile_file('./a.rb', {frozen_string_literal: true}).eval.frozen?"
false

$ ruby -v
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [aarch64-linux]

$ ruby -e "puts RubyVM::InstructionSequence.compile_file('./a.rb', {frozen_string_literal: true}).eval.frozen?"
true
```

This PRs adds setting the frozen_string_literal flag on the result object before parsing the file (similarly to how we do in `pm_iseq_compile_with_option`).

PR #16789, opened 2026-04-26
Add test for compile option support

byroot:
The bootsnap test suite ran into a regression in `compile_file`: https://github.com/rails/bootsnap/pull/535

Apparently it has already been fixed.

### DIFF
--- a/test/ruby/test_iseq.rb
+++ b/test/ruby/test_iseq.rb
@@ -212,6 +212,26 @@ class TestISeq < Test::Unit::TestCase
     end
   end
 
+  def test_compile_file_options
+    Tempfile.create(%w"test_iseq .rb") do |f|
+      f.puts('"test"')
+      f.close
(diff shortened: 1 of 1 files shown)


Earlopain:
It was fixed in https://github.com/ruby/ruby/pull/16779. ruby-dev builds are broken for almost a month already because of ASAN failures https://github.com/ruby/ruby-dev-builder/actions/runs/23763490942/job/69236963083 (maybe you can poke someone?)

byroot:
Ah thanks for the pointer, I think I know what's up with ASAN.

byroot:
I think https://github.com/ruby/ruby/pull/16792 will fix it, but I can't confirm it locally because ASAN won't crash on such bug on macOS it seems.

byroot:
It worked: https://github.com/ruby/ruby-dev-builder/actions/runs/24996114427

PR #16793, opened 2026-04-27
Win32 resolv reads UseDomainNameDevolution (REG_DWORD) as string, emits "Broken registry" warning

mrswadge:
On Windows, ext/win32/lib/win32/resolv.rb reads the registry value UseDomainNameDevolution via the string path (read_s). On normal systems this key is REG_DWORD, so the code emits:

Broken registry, ...\UseDomainNameDevolution was REG_DWORD, ignored

The registry is valid; the warning is caused by a type mismatch in the code.

Environment:
- Ruby branch: ruby_3_4
- File: ext/win32/lib/win32/resolv.rb
- Downstream observed in CINC 19.2.12 packaging Ruby 3.4.x on Windows
- Registry key type:
  HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\Tcpip\Parameters\UseDomainNameDevolution = REG_DWORD (for example value 1)

Actual behavior:
- Warning is emitted saying registry is broken when key is REG_DWORD.
- This issue causes CINC bootstrap failure on Windows devices.

Expected behavior:
- No warning for valid REG_DWORD type.
- UseDomainNameDevolution should be read as integer and honoured.

Root cause:
- get_item_property uses reg.read_s for non-expand reads.
- get_info calls get_item_property(..., "UseDomainNameDevolution").to_i.
- The refined read_s warns for any non-REG_SZ value.

Proposed fix (minimal):
1. Add an integer accessor helper for registry values.
2. Use that helper for UseDomainNameDevolution.
3. Keep existing string behaviour for string-valued keys.

### DIFF
--- a/ext/win32/lib/win32/resolv.rb
+++ b/ext/win32/lib/win32/resolv.rb
@@ -43,7 +43,7 @@ module Win32
 #====================================================================
   module Resolv
     begin
-      require 'win32/registry'
+      require 'win32/registry' unless defined?(Win32::Registry)
       module SZ
         refine Registry do
           # ad hoc workaround for broken registry
@@ -83,7 +83,7 @@ module Win32
 
           unless nvdom.empty?
             @search = [ nvdom ]
-            udmnd = get_item_property(TCPIP_NT, 'UseDomainNameDevolution').to_i
+            udmnd = get_item_property_i(TCPIP_NT, 'UseDomainNameDevolution')
             if udmnd != 0
               if /^\w+\./ =~ nvdom
                 devo = $'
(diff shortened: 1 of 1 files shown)


nobu:
Is this fixed in the latest win32-registory gem?
I don’t think we want to/should maintain old branches differently from the upstream repository.

mrswadge:
> Is this fixed in the latest win32-registory gem? I don’t think we want to/should maintain old branches differently from the upstream repository.

I had found that the implementation of https://github.com/ruby/ruby/blob/ruby_3_4/ext/win32/lib/win32/resolv.rb vs https://github.com/ruby/ruby/blob/master/ext/win32/lib/win32/resolv.rb are substantially different, therefore my decision was to alter the 3_4 branch specifically.

I don't have a version of CINC Client with a newer version of Ruby, so was not able to test the same scenario on the latest ruby versions.

I can however confirm that after editing the embedded ruby in CINC Client 19.2.12 directly, that these changes did fix the underlying issue in Ruby 3.4.

nobu:
It is fixed by ruby/resolv#100.

nobu:
Created the backport ticket: https://bugs.ruby-lang.org/issues/22017.

mrswadge:
> Created the backport ticket: https://bugs.ruby-lang.org/issues/22017.

Thank you.

mrswadge:
> Created the backport ticket: https://bugs.ruby-lang.org/issues/22017.

If I am reading it correctly, I noticed that https://bugs.ruby-lang.org/issues/22017 appears to be closed. Would you be able to confirm if https://github.com/ruby/resolv/pull/100 will still be getting backported to Ruby 3.4?

Thanks.

PR #16800, opened 2026-04-28
st.c: Store `st_table->bins` at the end of `st_table->entries`

byroot:
Same optimization that was applied to `set_table` in 85c52079aa35a1d2e063a5b40eebe91701c8cb9e

This saves one pointer per `st_table`.
Now that GC support 64B slots, if we can save another 8B either in `st_table` or in `struct RHash`, hashes would now fit in 64B slots.

I can imagine a few ways where we could find the remaining 8B:

### `RHash.ifnone`

Most hashes don't have a default value or proc, we could store the `ifnone` field externally (e.g. fields). The downside of course is that it would become a Ractor sync point (`generic_fields_tbl`).
We could also store it after the `st_table`, but since it can be set after initialization, that could only be an optimization done by the compactor.

### `st_table.type`

This is a 8B field which only a handful of different values, given the overwhelming majority of hashes use the same few types.
We could not store it in `st_table`, and instead make the caller responsible to always pass it to all `st_*` function.
That of course would break the existing `st_*` API (so would probably need to have a private copy), and make it less ergonomic, but wouldn't be too hard.

### Make `st_table.num_entries` a `uint32_t`

This of course would limit the biggest possible hash to 4 billion entries, but would save 4B in `num_entries`, `entries_start` and `entries_bound`. Perhaps it is an acceptable tradeoff?

cc @jhawthorn any opinions?

### DIFF
--- a/st.c
+++ b/st.c
@@ -398,7 +398,7 @@ set_bin(st_index_t *bins, int s, st_index_t n, st_index_t v)
 
 /* Mark I-th bin of table TAB as empty, in other words not
    corresponding to any entry.  */
-#define MARK_BIN_EMPTY(tab, i) (set_bin((tab)->bins, get_size_ind(tab), i, EMPTY_BIN))
+#define MARK_BIN_EMPTY(tab, i) (set_bin(st_bins_ptr(tab), get_size_ind(tab), i, EMPTY_BIN))
 
 /* Values used for not found entry and bin with given
    characteristics.  */
@@ -415,7 +415,7 @@ set_bin(st_index_t *bins, int s, st_index_t n, st_index_t v)
    corresponding to deleted entries. */
 #define MARK_BIN_DELETED(tab, i)				\
     do {                                                        \
-        set_bin((tab)->bins, get_size_ind(tab), i, DELETED_BIN); \
+        set_bin(st_bins_ptr(tab), get_size_ind(tab), i, DELETED_BIN); \
     } while (0)
 
 /* Macros to check that value B is used empty bins and bins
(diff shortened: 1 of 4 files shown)


st.c
```diff
@@ -2304,7 +2285,7 @@ st_rehash(st_table *tab)
     int rebuilt_p;
 
     do {
-        if (tab->bin_power <= MAX_POWER2_FOR_TABLES_WITHOUT_BINS)
+        if (tab->entry_power <= MAX_POWER2_FOR_TABLES_WITHOUT_BINS)
```
byroot, line 2288: This is interesting, some places where checking `entry_power`, others where checking `bin_power`, causing an off by one error. Hard to tell for sure which one was really intended, but based on the constant doc, I'd say it was `entry_power`:

```c
/* If the power2 of the allocated `entries` is less than the following
   value, don't allocate bins and use a linear search.  */
#define MAX_POWER2_FOR_TABLES_WITHOUT_BINS 4
```

jhawthorn:
> ifnone

I had the same thought. We could store this in the st_table as Qundef (or in some fixed location like after bins/after entries, though that might be awkward if we have to store it differently for ar tables, or always promote ifnone tables to st_table)

> st_table.type

I like this idea, we only have a couple types for a `T_HASH` I think (standard, identity, maybe cdhash?). I feel like with LTO there might also be a chance to inline better if we're telling the compiler a constant set of function pointers (not 100% sure). I agree the API change/compatibility is awkward though.

> Make `st_table.num_entries` a `uint32_t`

I feel like this should be fine, `2**32 * 3 * 8 = 103GB` just for the entries, but that's probably a discussion that needs to be had on  the tracker.

jhawthorn:
> This of course would limit the biggest possible hash to 4 billion entries, but would save 4B in `num_entries`, `entries_start` and `entries_bound`

Maybe we could move `entries_start` which feels particularly wasteful? That's 0 unless we've shifted off the front of the entries or happened to delete them, in which case maybe we could store the offset in the first tombstoned entry? or we could make it much smaller (16 bits?) and rebuild if that overflows?

byroot:
> We could store [finone] in the st_table as Qundef 

I thought so too, but figure it's be awkward and painful, we'd have to account for it in `RHASH_SIZE`, iteration, etc etc.

> or in some fixed location like after bins/after entries,

Here it will be a bit tricky because of the separate `RHash` & `st_table` APIs. We'd have to make sure not to loose it on resizes, etc. Seems like a pain.

Ivars seems like the only viable solution, except it's a problem for ractors...

> maybe cdhash?)

Honestly I feel `cdhash` (an any other place where we set a an st type for RHash I don't know about) is a bit of an abuse as evidenced by how I broke it recently. I'd like to migrate `cdhash` to be an IMEMO (or TypedData).

PR #16802, opened 2026-04-28
Fixing __has_include(<stdckdint.h>) compilation error

darkain:
The very thing the conditional for `__has_include` is trying to handle, is causing compilation to break.

This file was brought into the ruby/bigdecimal project, and is failing to compile on a platform which doesnt have `__has_include` support in the compiler.

Compile log can be found in this comment:
https://github.com/ruby/bigdecimal/pull/531#issuecomment-4337861950

### DIFF
--- a/missing/dtoa.c
+++ b/missing/dtoa.c
@@ -210,10 +210,14 @@
 #include <locale.h>
 #endif
 
-#if defined(HAVE_STDCKDINT_H) || !defined(__has_include)
-#elif __has_include(<stdckdint.h>)
+#if !defined(HAVE_STDCKDINT_H)
+#if defined(__has_include)
+#if __has_include(<stdckdint.h>)
 # define HAVE_STDCKDINT_H 1
 #endif
+#endif
+#endif
+
 #ifdef HAVE_STDCKDINT_H
 # include <stdckdint.h>
 #endif


zenspider:
This should be ported over to bigdecimal and any other libs that might use this file.

zenspider (approved):

nobu:
What compiler and environment?

darkain:
@nobu This is on Oracle Linux 7.9

nobu:
Thank you.
What is the exact compiler version?

I found that the default compiler on Oracle Linux 7 is gcc 4.8.5, but it is too old.
I think you install a newer version.

darkain:
The specific check in the code that I'm submitting this PR for is explicitly a fix for "old compilers" that wasn't done correctly. Its the only thing that prevents the code from working on these systems.

tompng:
GCC 5 supports `__has_include`, so your GCC version is 4.8.5 ~ 4.9.4. Ruby >= 3.3 dropped support of  GCC <= 6 (https://bugs.ruby-lang.org/issues/19353), and 3.2 is already EOL. So the code is correct for the oldest supported GCC version in the oldest supported Ruby version.

It's unacceptable to merge this as a fix, but I think it's OK as a refactor if it's refactored.

Comparing `empty-if and elif` vs `3-depth nested if`, it's hard to say that it's clearly improved. Can you change the 3-depth nested `if` to 2-depth nested `if` like this? 
```c
#if !defined(HAVE_STDCKDINT_H) && defined(__has_include)
#if __has_include(<stdckdint.h>)
```

PR #16804, opened 2026-04-29
Rename `putstring` instruction as `dupstring`

byroot:
As well as `putchilledstring` as `dupchilledstring`.

This is more consistent with similar `duparray` and `duphash` instructions and better reflect it's behavior.

However I do get that this may not be worth the churn?


jeremyevans:
I'm in favor of this change, as it makes it obvious the instruction is allocating a new object. Having `putstring` allocate but `putobject` not allocate is confusing.

byroot:
Thanks @jeremyevans yhis has been my pet peeve for a while. I wasn't planing to merge this unless I'd get support from someone else, went faster than I expected.

@ruby/jit any strong opposition? (Given you tend to work with instructions more than most)

XrXr:
Makes sense to me.

k0kubun:
:+1:

PR #16805, opened 2026-04-29
Fix coverage support for RubyVM::ISeq.compile

byroot:
[[Bug #22018]](https://bugs.ruby-lang.org/issues/22018)
    
ISeq returned by `RubyVM::InstructionSequene.load_iseq` weren't handled by the coverage module.

Repro:

```ruby
require "coverage"

File.write("/tmp/a.rb", <<~RUBY)
  module CoverableRaw
    def self.call
      "cover up"
    end
  end

  CoverableRaw.call
RUBY

Coverage.start
require "/tmp/a.rb"
p Coverage.result

File.write("/tmp/b.rb", <<~RUBY)
  module Coverable
    def self.call
      "cover up"
    end
  end

  Coverable.call
RUBY

class RubyVM::InstructionSequence
  def self.load_iseq(path)
    compile_file(path)
  end
end

Coverage.start
require "/tmp/b.rb"
p Coverage.result
```

Expected:

```ruby
{"/tmp/a.rb" => [1, 1, 1, nil, nil, nil, 1]}
{"/tmp/b.rb" => [1, 1, 1, nil, nil, nil, 1]}
```

Actual:

```ruby
{"/tmp/a.rb" => [1, 1, 1, nil, nil, nil, 1]}
{}
```

### DIFF
--- a/iseq.c
+++ b/iseq.c
@@ -864,10 +864,6 @@ set_compile_option_from_hash(rb_compile_option_t *option, VALUE opt)
 static rb_compile_option_t *
 set_compile_option_from_ast(rb_compile_option_t *option, const rb_ast_body_t *ast)
 {
-#define SET_COMPILE_OPTION(o, a, mem) \
-    ((a)->mem < 0 ? 0 : ((o)->mem = (a)->mem > 0))
-    SET_COMPILE_OPTION(option, ast, coverage_enabled);
-#undef SET_COMPILE_OPTION
     if (ast->frozen_string_literal >= 0) {
         option->frozen_string_literal = ast->frozen_string_literal;
     }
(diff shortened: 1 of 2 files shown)


iseq.c
```diff
@@ -864,10 +864,6 @@ set_compile_option_from_hash(rb_compile_option_t *option, VALUE opt)
 static rb_compile_option_t *
 set_compile_option_from_ast(rb_compile_option_t *option, const rb_ast_body_t *ast)
 {
-#define SET_COMPILE_OPTION(o, a, mem) \
-    ((a)->mem < 0 ? 0 : ((o)->mem = (a)->mem > 0))
-    SET_COMPILE_OPTION(option, ast, coverage_enabled);
```
byroot, line 869: This was always disabling coverage for `parse.y`, but it's called from several codepaths. Since the intent is to disable coverage for `eval`, I moved the logic to `rb_iseq_new_eval`.

PR #16808, opened 2026-04-29
compile.c: repack `struct ibf_header`

byroot:
`uint8_t` is more than enough to store major and minor versions.

### DIFF
--- a/compile.c
+++ b/compile.c
@@ -12533,8 +12533,10 @@ static const char IBF_ENDIAN_MARK =
 
 struct ibf_header {
     char magic[4]; /* YARB */
-    uint32_t major_version;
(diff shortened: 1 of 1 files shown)


nobu:
https://github.com/ruby/ruby/actions/runs/25108743320/job/73576724560?pr=16808#step:10:1836
>   /tmp/tmp.2h2cWkATwc/ruby: unmatched version file (4.21 for 4.10005) (RuntimeError)

This error is because reading 32bit integers written by host ruby as 8bit integers

byroot:
> because reading 32bit integers written by host ruby as 8bit integers

I suspected as much, but still unclear to me how to reproduce or fix it.

PR #16810, opened 2026-04-29
ZJIT: Specialize direct sends to methods with post-required positional parameters

nirvdrum:
This PR allows `can_direct_send` to specialize calls to methods that mix optional and required positional parameters (e.g. `def m(a = 10, b)` or `def m(a, b = 10, c)`). Previously, post-required parameters were rejected in `can_direct_send`, causing such calls back to the generic megamorphic `Send`. As far as I can tell, the rejection traces to the original optional-parameter support commit (89849f3b4a), where post-param support appears to have been deferred for scope reasons rather than blocked by a deeper correctness concern.

The `can_direct_send` guard now accounts for `post_num` (the count of required positional parameters that appear after optional params). The `num_optionals_passed` formula in `gen_send_iseq_direct` also now accounts for `post_num`, so the right `jit_entry_blocks` index is selected when the caller fills some, but not all, optionals. Other relevant parts of ZJIT already accounted for `post_num`.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -1671,12 +1671,14 @@ fn gen_send_iseq_direct(
         // See vm_call_iseq_setup_normal_opt_start in vm_inshelper.c
         let lead_num = params.lead_num as u32;
         let opt_num = params.opt_num as u32;
+        let post_num = params.post_num as u32;
         let keyword = params.keyword;
         let kw_total_num = if keyword.is_null() { 0 } else { unsafe { (*keyword).num } } as u32;
-        assert!(args.len() as u32 <= lead_num + opt_num + kw_total_num);
+        assert!(args.len() as u32 <= lead_num + opt_num + post_num + kw_total_num);
         // For computing optional positional entry point, only count positional args
+        // and exclude the always-present lead and post slots.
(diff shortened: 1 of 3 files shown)


XrXr (approved):
Sure, but let's drop support for these if they create too much complexity down the line when combined with other parameter argument features.

XrXr:
If these are not popular in practice there is no point adding a line in the matrix to support these.

nirvdrum:
Agreed, we shouldn't keep code that isn't being used. It's hard to say what's being gated by the generic `Send` at the moment. I think the Rubocop benchmark is the most representative in _ruby-bench_.

With that said, I also think there's a delicate balance insomuch as I'd prefer to avoid lint rules and blog posts advising Rubyists to write functions according to a seemingly arbitrary set of rules because in one case they'll be JITted and another they won't. Fortunately, ZJIT is pretty well-structured; this PR is quite small and I don't anticipate the maintenance burden being that large. But, yeah, if it doesn't pay for itself we should yank it.

XrXr:
FWIW post params are straight rejected in YJIT as they just don't seem to be used in practice. I'm a little worried about how when combined with rest params, they make the check a bit more complicated and also prevent simple movement to do the arg transfer. We could reject that combination.

PR #16811, opened 2026-04-29
ZJIT: Inline Class#allocate

jhawthorn:
This adds specialization for `Class#allocate` to perform the same inlining that Class#new will do. This was done conservatively, only inlining when we see a known leaf allocator so that there's no difference between interpreter and JIT if the allocator happens to observe the stack.

`Class#allocate` isn't that common, though it has [one significant use in Rails](https://github.com/rails/rails/blob/83423052fc899315fad91c20b5197a186fc8f0fd/activerecord/lib/active_record/persistence.rb#L313). Though less common than `Class#new`, I think it should not be slower. It's useful for benchmarking and can be useful for performing initialization _before_ a user-defined initialize method.

The check for leaf allocator and calling `Insn::ObjectAllocClass` instead of `Insn::ObjectAlloc` needs to happen in the cfunc inlining because that happens after the type specialization pass.

### DIFF
--- a/zjit/src/hir/opt_tests.rs
+++ b/zjit/src/hir/opt_tests.rs
@@ -4718,6 +4718,94 @@ mod hir_opt_tests {
     }
 
     #[test]
+    fn test_inline_class_allocate() {
+        eval("
+            class C; end
+            def test = C.allocate
+            test
+        ");
+        assert_snapshot!(hir_string("test"), @"
+        fn test@<compiled>:3:
+        bb1():
+          EntryPoint interpreter
+          v1:BasicObject = LoadSelf
+          Jump bb3(v1)
+        bb2():
+          EntryPoint JIT(0)
+          v4:BasicObject = LoadArg :self@0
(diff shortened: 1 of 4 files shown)


k0kubun (approved):
looks good, but it might be nice to have an HIR test?

zjit/src/hir/opt_tests.rs
```diff
@@ -4691,6 +4691,94 @@ mod hir_opt_tests {
         ");
     }
 
+    #[test]
+    fn test_opt_allocate() {
```
tekknolagi: nit: we prefer tests with `test_<predicate>_<outcome>`, close ish to English

tekknolagi: ideally we should be able to look at the name and determine if an update to the HIR is OK or not. this does not give a second bit of information to tell us if we've optimized the right way

tekknolagi (approved):
sheep it

PR #16812, opened 2026-04-30
Use flag to determine when imemo_memo.u3 is a VALUE

jhawthorn:
We always know when `MEMO.u3.value` contains a VALUE and even fire a write barrier for it. We should use this when marking, support compaction, and no longer need conservative marking.

Also remove unused `MEMO.u3.func`

### DIFF
--- a/imemo.c
+++ b/imemo.c
@@ -101,15 +101,26 @@ rb_free_tmp_buffer(volatile VALUE *store)
 }
 
 struct MEMO *
-rb_imemo_memo_new(VALUE a, VALUE b, VALUE c)
+rb_imemo_memo_new(VALUE a, VALUE b, long c)
 {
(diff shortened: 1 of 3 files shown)


imemo.c
```diff
@@ -99,15 +101,26 @@ rb_free_tmp_buffer(volatile VALUE *store)
...
+
+    return memo;
+}
+
+struct MEMO *
+rb_imemo_memo_new_value(VALUE a, VALUE b, VALUE c)
+{
+    struct MEMO *memo = IMEMO_NEW(struct MEMO, imemo_memo, 0);
 
     *((VALUE *)&memo->v1) = a;
```
jhawthorn, line 108: @peterzhu2118 @eightbitraptor Hopefully this helps mmtk a little 😊

peterzhu2118 (approved):
Awesome! Love that it makes this not pinning anymore.

PR #16820, opened 2026-05-01
Add back note about regexp /x comment limitation

Earlopain:
It was dropped in https://github.com/ruby/ruby/commit/932dd9f10e684fa99b059054fbc934607d85b45a. Ref https://github.com/ruby/prism/issues/4097

cc @eregon


eregon (approved):
Thanks!

PR #16828, opened 2026-05-02
ZJIT: Drop redundant type guards via block-local HIR canonicalize

### LINKED ISSUE #978: ZJIT: Redundant GuardType remains at CFG join blocks
## Summary

The current ZJIT HIR optimizer cannot eliminate redundant GuardType instructions at CFG (Control Flow Graph) join blocks.
When a value is already proven to have a narrower type inside each predecessor branch, the merge block re-guards it on every
 call, leaving redundant runtime type checks in the JIT-compiled code.

## Reproduction

```ruby
def test(n, cond)
  if cond
    a = n + 1
  else
    a = n + 2
  end
  n + a
end
```

### Optimized HIR (--zjit-dump-hir-opt):

```
bb3:  # cond == true
  v69:Fixnum = GuardType v14, Fixnum
  v70 = FixnumAdd v69, 1
  Jump bb5(v13, v14, ..., v70)        # passes raw v14
bb4:  # cond == false
  v73:Fixnum = GuardType v38, Fixnum
  v74 = FixnumAdd v73, 2
  Jump bb5(v37, v38, ..., v74)        # passes raw v38
bb5(v51, v52:BasicObject, ..., v54):  # v52 inferred as BasicObject
  v75:Fixnum = GuardType v52, Fixnum  # redundant — both predecessors prove Fixnum
  v79 = FixnumAdd v75, v54
```
                                                                                                                            
The GuardType at v75 is guaranteed to succeed at runtime (both predecessors have already proven n is Fixnum), 
yet it executes on every call.
                                                                                                                   
## Root cause

Each Jump passes the raw pre-guard value (v14 / v38: BasicObject) rather than the GuardType result (v69 / v73: Fixnum). The 
merge-block parameter v52 is therefore inferred as union(BasicObject, BasicObject) = BasicObject, and fold_constants cannot
drop a guard whose input type is wider than the guard type.

dak2:
## Summary

`GuardType` narrows only its result id, not the original input. 
A later guard on the same input therefore looks unfoldable to `fold_constants` even though the side-exit semantics already prove the narrower type.
Canonicalize rewrites later uses to the most recent `Guard*` result, making the redundant guard foldable.
For example, in this CFG-join shape:

 ```ruby
  def test(n, cond)                                                                                                           
    if cond   
      a = n + 1
    else                                                                                                                      
      a = n + 2
    end                                                                                                                       
    n + a   # `n` gets a redundant Fixnum guard here
  end
 ```

This PR adds a block-local HIR `canonicalize` pass that walks each block in RPO and rewrites every operand through union-find plus a per-block `rewrite_map` keyed on the most recent `Guard*` for that value. 
After canonicalization, `infer_types` can narrow merge-block parameter types and `fold_constants` can then drop the redundant guards in both shapes above.

Inspired by Cranelift's canonicalize https://cfallin.org/blog/2026/04/09/aegraph/

Fixes: https://github.com/Shopify/ruby/issues/978

## Benchmarks

Bench (arm64 linux devcontainer, ruby/ruby-bench, warmup=10 bench=20)

```
  Throughput              master/staged
    lobsters              1.021 (+2.1%, within ±3-6% noise)
    railsbench            1.012 (+1.2%, within ±3-4% noise)

  --zjit-stats            lobsters / railsbench
    code_region_bytes     -1.1%   / -1.0%   (redundant CFG-join guards still removed)
    guard_type_count      -29.4%  / +30.1%  (railsbench likely single-run noise) ⚠
    compile_hir_time      +14.6%  / +13.7%  (canonicalize_time: 67ms / 31ms)
    invalidation_time      ±0%    /  ±0%
```

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -4970,11 +4970,8 @@ impl Function {
                     *operand = rewrite_map.get(&canon).copied().unwrap_or(canon);
                 });
 
-                // GuardTypeNot is excluded: its infer_type is BasicObject
-                // (HIR has no negated-type representation), so registering
-                // would widen downstream uses' inferred types. For the
-                // binary guards only `left` is registered because their
-                // infer_type is type_of(left).
+                // For the binary guards only `left` is registered because their infer_type is
+                // type_of(left).
                 match &self.insns[canonical_id.0] {
                     Insn::GuardType      { val:  src, .. }
                     | Insn::GuardBitEquals { val:  src, .. }
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -4144,6 +4144,9 @@ impl Function {
                 }
             }
         }
+        crate::stats::trace_compile_phase("canonicalize", ||
+            crate::stats::with_time_stat(Counter::compile_hir_canonicalize_time_ns, || self.canonicalize())
+        );
         crate::stats::trace_compile_phase("infer_types", || self.infer_types());
     }
 
@@ -4198,6 +4201,9 @@ impl Function {
                 }
             }
         }
+        crate::stats::trace_compile_phase("canonicalize", ||
+            crate::stats::with_time_stat(Counter::compile_hir_canonicalize_time_ns, || self.canonicalize())
+        );
         crate::stats::trace_compile_phase("infer_types", || self.infer_types());
     }
 
@@ -4486,6 +4492,9 @@ impl Function {
                 }
             }
         }
+        crate::stats::trace_compile_phase("canonicalize", ||
+            crate::stats::with_time_stat(Counter::compile_hir_canonicalize_time_ns, || self.canonicalize())
+        );
         crate::stats::trace_compile_phase("infer_types", || self.infer_types());
     }
 
@@ -4816,6 +4825,9 @@ impl Function {
                 self.push_insn_id(block, insn_id);
             }
         }
+        crate::stats::trace_compile_phase("canonicalize", ||
+            crate::stats::with_time_stat(Counter::compile_hir_canonicalize_time_ns, || self.canonicalize())
+        );
         crate::stats::trace_compile_phase("infer_types", || self.infer_types());
     }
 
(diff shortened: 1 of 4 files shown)


tekknolagi:
Thank you for the PR! This feels like it has a more general shape peeking out of it than its current state (something GVN-like (global value numbering)). Would you be open to some guided rewrites to make it more general and more powerful?

dak2:
@tekknolagi 

Yes, I'm definitely open to that! I can see how this could generalize toward something GVN-like.
Please guide me on where you'd like to start.

tekknolagi:
I think what we want is a (for now) local version of `canonicalize` from https://cfallin.org/blog/2026/04/09/aegraph/

That is, a function that walks over each block in `rpo` and for each block does:

1. make a new `rewrite_map` `HashMap`
1. for each instruction
    1. look up its ID in the union-find
    1. rewrite all operands (using `for_each_operand_mut`) to first look through the union-find and then see if there is an available replacement in `rewrite_map`
    1. if the current instruction is a any of our `GuardXYZ` or `RefineType` instructions, insert the union-find-ed instruction ID as the canonical rewrite for the guard operand

This can later (not in this PR, please) be extended to walk the dominator tree using a scoped hashmap or something so that we get cross-block results that don't correspond to block parameters.

Then you should call this before the (traced) call sites of `infer_types`. You'll probably have to update some tests, which is expected (and good!). To do that, you can use `make zjit-test-update` (read the [docs](https://docs.ruby-lang.org/en/master/jit/zjit_md.html) for more).

dak2:
@tekknolagi 
Thanks for the detailed explanation! I’ll use this as a reference while I implement it, so please review it again.

zjit/src/hir.rs
```diff
@@ -5611,6 +5678,8 @@ impl Function {
         run_pass!(optimize_c_calls);
         run_pass!(convert_no_profile_sends);
         run_pass!(optimize_load_store);
+        run_pass!(canonicalize);
```
tekknolagi: I think you can remove this one, at least for now. We will start off running it with type inference

tekknolagi:
Can you update the description to match the new PR contents? Or I'll get to it later. Otherwise I think it looks good to me

tekknolagi:
Out of curiosity @ydah what did you mean when you assigned yourself to this PR?

k0kubun (approved):

XrXr (approved):

dak2:
@tekknolagi
Thank you for the review! I rewrote PR descriptions and title.

tekknolagi:
I'll take over any small edits required to get this landed & avoid rebase conflicts

dak2:
@tekknolagi 

Thanks for your huge help!

tekknolagi:
Thanks for contributing :)

PR #16835, opened 2026-05-03
Add {Method,UnboundMethod,Proc}#source_range

eregon:
[[Feature #21998]](https://bugs.ruby-lang.org/issues/21998)

- [x] I will squash before merging.

### DIFF
--- a/proc.c
+++ b/proc.c
@@ -43,11 +43,201 @@ VALUE rb_cUnboundMethod;
 VALUE rb_cMethod;
 VALUE rb_cBinding;
 VALUE rb_cProc;
+static VALUE rb_cSourceRange;
 
 static rb_block_call_func bmcall;
 static int method_arity(VALUE);
 static int method_min_max_arity(VALUE, int *max);
 static VALUE proc_binding(VALUE self);
+
+struct source_range_data {
+    VALUE path;
+    VALUE absolute_path;
+    int start_line;
+    int start_column;
+    int end_line;
+    int end_column;
+};
+
+static size_t
+source_range_memsize(const void *ptr)
+{
+    return sizeof(struct source_range_data);
+}
+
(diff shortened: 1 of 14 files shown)


test/coverage/test_coverage.rb
```diff
@@ -522,11 +522,11 @@ def baz; end
...
-        [Object, :a, 6, 18, 6, 25] => 2,
-        [Object, :b, 7, 18, 8, 3] => 0,
-        [Object, :bar, 2, 20, 3, 1] => 1,
-        [Object, :baz, 4, 9, 4, 11] => 0,
-        [Object, :foo, 1, 20, 1, 22] => 2,
+        [Object, :a, 6, 0, 6, 25] => 2,
+        [Object, :b, 7, 0, 8, 3] => 0,
+        [Object, :bar, 2, 0, 3, 1] => 1,
+        [Object, :baz, 4, 4, 4, 11] => 0,
+        [Object, :foo, 1, 0, 1, 22] => 2,
```
eregon: @mame WDYT about this?
If we want to return from the `p` of `proc { ... }` (and not from the `{`) then we need to change the `code_location` (see `compile.c`/`prism_compile.c`) and this affects methods Coverage too.
(I don't think maintaining 2 separate `code_location` makes sense)

Given the changes in `compile.c`/`prism_compile.c` which feel a bit messy I think the easier way is to return from the `{` and then we don't need to change this. It also matches what the extended `source_location` used to return and what Charles wants.
So if you think that's fine too then I'll try to convince matz.

eregon: I changed to have the start at `{`/`do`, same as the old extended `source_location` and method coverage.

More details in https://bugs.ruby-lang.org/issues/21998#note-24

eregon:
Approved by matz in https://bugs.ruby-lang.org/issues/21998#note-27

PR #16836, opened 2026-05-03
Fix node used for a call with a block

eregon:
* Use the ITER node_id for a call with a block, otherwise it would cover other arguments but not the block.

https://bugs.ruby-lang.org/issues/22020

### DIFF
--- a/lib/error_highlight/core_ext.rb
+++ b/lib/error_highlight/core_ext.rb
@@ -7,7 +7,7 @@ module ErrorHighlight
         locs = self.backtrace_locations
         return "" if locs.size < 2
         callee_loc, caller_loc = locs
-        callee_spot = ErrorHighlight.spot(self, backtrace_location: callee_loc, point_type: :name)
+        callee_spot = ErrorHighlight.spot(self, backtrace_location: callee_loc, point_type: :args)
         caller_spot = ErrorHighlight.spot(self, backtrace_location: caller_loc, point_type: :name)
         if caller_spot && callee_spot &&
             caller_loc.path == callee_loc.path &&
--- a/compile.c
+++ b/compile.c
@@ -8453,6 +8453,8 @@ compile_iter(rb_iseq_t *iseq, LINK_ANCHOR *const ret, const NODE *const node, in
 {
     const int line = nd_line(node);
     const NODE *line_node = node;
+    const NODE *iter;
+    enum node_type iter_type;
     const rb_iseq_t *prevblock = ISEQ_COMPILE_DATA(iseq)->current_block;
     LABEL *retry_label = NEW_LABEL(line);
     LABEL *retry_end_l = NEW_LABEL(line);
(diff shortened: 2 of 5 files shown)


eregon:
I wonder if
```
> RubyVM::AbstractSyntaxTree.of(lambda { })
=> (ITER@1:30-1:40 (FCALL@1:30-1:36 :lambda nil) (SCOPE@1:37-1:40 tbl: [] args: nil body: (BEGIN@1:38-1:38 nil)))
```
should return the SCOPE instead of the ITER.
That's related to the discussion in https://bugs.ruby-lang.org/issues/21998.
If it did it would be easier to differentiate the "block definition node" from the "call node to which the block is given".

(on Prism this is already the case: `p Prism.find(lambda { })` returns `BlockNode` and not `CallNode`)

---

FWIW a method definition looks like this with RubyVM::AST:
```
irb(main):002> RubyVM::AbstractSyntaxTree.of(method(def m = 42))
=> 
(DEFN@1:37-1:47
 mid: :m
 body:
   (SCOPE@1:37-1:47
    tbl: []
    args:
      (ARGS@1:42-1:42
       pre_num: 0
       pre_init: nil
       opt: nil
       first_post: nil
       post_num: 0
       post_init: nil
       rest: nil
       kw: nil
       kwrest: nil
       block: nil)
    body: (INTEGER@1:45-1:47 42)))
```

eregon:
Maybe a better fix for https://bugs.ruby-lang.org/issues/22020 would be to extend the location of the FCALL to include the block too.
Then we can easily differentiate the call to create the block (FCALL, as before but with longer location) from the block definition (ITER, as before).

mame:
For [Bug #22020], could you get approval from matz or someone familiar with parse.y first?

The procedure for fixing error_highlight should be:

1. Open a PR against the ruby/error_highlight repository so that the tests pass with both the current and the new behavior of `RubyVM::AbstractSyntaxTree.of`.
2. Then change `RubyVM::AbstractSyntaxTree.of` in ruby/ruby.

lib/error_highlight/core_ext.rb
```diff
@@ -7,7 +7,7 @@ module CoreExt
         locs = self.backtrace_locations
         return "" if locs.size < 2
         callee_loc, caller_loc = locs
-        callee_spot = ErrorHighlight.spot(self, backtrace_location: callee_loc, point_type: :name)
+        callee_spot = ErrorHighlight.spot(self, backtrace_location: callee_loc, point_type: :args)
```
mame, line 10: I am not sure why this change is needed?

eregon: It's necessary to not fail the ErrorHighlight tests.
With this PR changes, `ITER` is used in two cases:
```
lambda { }.call(1)
       ^
```
for which we use `point_type: :args`

```
method_does_not_exist { }
^^^^^^^^^^^^^^^^^^^^^
```
for which we use `point_type: :name`

PR #16837, opened 2026-05-03
[DOC] Doc for Pathname#read

BurdetteLamar:
@peterzhu2118, I think since this method deals in bytes that it should have examples in Japanese and binary data.  If you agree, I'll harmonize File#read to do the same (separate PR).

### DIFF
--- a/pathname_builtin.rb
+++ b/pathname_builtin.rb
@@ -11,6 +11,12 @@
 #   Pathname.new('lib')            # => #<Pathname:lib>
 #   Pathname.new('/usr/local/bin') # => #<Pathname:/usr/local/bin>
 #
+# == About the Examples
+#
+# Many examples here use these variables:
+#
+#  :include: doc/examples/files.rdoc
+#
 # == Convenience Methods
 #
 # The class provides *all* functionality from class File and module FileTest,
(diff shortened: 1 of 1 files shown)


pathname_builtin.rb
```diff
@@ -1040,8 +1046,63 @@ def each_line(...) # :yield: line
...
+  # returns +length+ bytes if available, beginning at the given +offset+:
+  #
+  #   Pathname.new('t.txt').read(10, 2)
+  #   # => "rst line\r\n"
+  #   Pathname.new('t.ja').read(10, 2)
+  #   # => "\x93\xE3\x82\x93\xE3\x81\xAB\xE3\x81\xA1"
+  #   Pathname.new('t.dat').read(10, 2)
+  #   # => "\x99\x90\x99\x91\x99\x92\x99\x93\x99\x94"
+  #
+  # Returns +nil+ if the specified number of bytes is unavailable:
```
peterzhu2118: I think the "specified number of bytes is unavailable" is a bit ambiguous. It sounds like the "length is larger than the files size" case (where we are requesting more bytes than there are). Maybe something like this?

```suggestion
  # Returns +nil+ if +offset+ is past the end of the file:
```

peterzhu2118 (approved):

PR #16843, opened 2026-05-04
shape.h: introduce SHAPE_ID_FL_EMBEDDED_FIELDS

byroot:
This flag tells whether the object has its fields embedded inside its slot. This is more or less the same information as `ROBJECT_HEAP`.
    
It should allow reading into a T_OBJECT by checking only the shape. Later on, I'd like all types to have their `heap_id` in their shape, allowing a quick and efficient way to query their slot size.

This is very experimental, I don't know for sure if it's worth it, but the exploratory work is helpful.

Blockers:

  - Since the introduction of the new slot sizes, `RVALUE_OVERHEAD` is substracted from slot sizes rather than added, this breaks a few assumptions in shape.c. I need to check that with @eightbitraptor .
  - ~~Compaction isn't yet supported. The `heap_id` part of the `shape_id` has to be updated after an object has been moved. I'm not yet sure how best do that.~~


tekknolagi:
The JITs would like this, I think

byroot:
> The JITs would like this, I think

Yes, that's the reason for the change. But right now the patch require lots of brittle logic sprinkled everywhere, so I might try getting to the same result with a different approach.

byroot:
This was achieve the the "shape layout" lots introduced a couple months ago.

PR #16844, opened 2026-05-04
Avoid issues with refinement zsuper lookup

jeremyevans:
Without this patch, there are multiple issues when defining a zsuper method via a refinement. These issues are due to the orig_me entry for the zsuper method not being updated when the method it refers to is modified via:

* Method being redefined (in same or closer ancestor)
* Method being removed
* Method being undef-ed (in same or closer ancestor)
* Method being overridden by method in included module
* Method being overridden by method in prepended module
* Method being defined in an included/prepended module at time of refinement

This includes a comprehensive test suite for these cases, using both zsuper methods and iseq methods that call super. Without the changes to refinement method lookup, the following types of errors occur for refinement zsuper methods (all of the iseq methods that call super work correctly with or without these changes):

* Incorrect result (not returning result of expected method, not raising NoMethodError if the method was removed or has been undef-ed)
* SystemStackError: stack level too deep
* NotImplementedError: false() function is unimplemented on this machine

This avoids the issues with refinement zsuper lookup by turning the zsuper into a cfunc that uses rb_call_super_kw. This is not a perfect solution, for two reasons:

* cfuncs are slower than zsuper
* arity/parameters for the method not as helpful

It may possible to avoid these issues by clearing method caches in more cases. However, I think that would require a lot of extra work, since you cannot just clear the method cache for the current class. You would need to clear it for all subclasses that are refined, and if this is a module, do the same for all classes that include/prepend the module, as well as any subclasses of those classes.

Considering the need for refinement zsuper methods is very rare, the performance and arity/parameters issues seem acceptable (to me).

Fixes [Bug #22022]

### DIFF
--- a/test/ruby/test_refinement.rb
+++ b/test/ruby/test_refinement.rb
@@ -1058,6 +1058,586 @@ class TestRefinement < Test::Unit::TestCase
     end;
   end
 
+  {
+    zsuper: "public :a",
+    super: "def a = super"
+  }.each do |desc, method_def|
+    define_method :"test_modify_#{desc}_refinement_method_in_superclass" do
+      assert_separately([], <<-"end;")
+        class A
+          private def a = :a
+          alias a a
+        end
+
+        class B < A
+        end
+
+        module R
+          refine B do
+            #{method_def}
+          end
+        end
+        using R
+        assert_equal(:a, B.new.a)
+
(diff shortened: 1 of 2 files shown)


shugo (approved):
It looks good to me.  Thank you!

PR #16851, opened 2026-05-04
[DOC] Add Japanese and data to IO.read

BurdetteLamar:
This method is byte-oriented, and so should have examples with multi-byte characters and binary data.

Also, I think examples of an IO singleton method should use File, not IO.


BurdetteLamar:
Marking as Draft;  I think I've made some mistakes here.

peterzhu2118 (approved):

PR #16857, opened 2026-05-04
[DOC] for IO.write

BurdetteLamar:
This method is byte-oriented, and so should have examples with multi-byte characters and binary data.

Also, I think examples of an IO singleton method should use File, not IO.

### DIFF
--- a/io.c
+++ b/io.c
@@ -12405,7 +12405,7 @@ io_s_write(int argc, VALUE *argv, VALUE klass, int binary)
 
 /*
  *  call-seq:
- *    IO.write(path, data, offset = 0, **opts)    -> integer
+ *    IO.write(path, data, offset = 0, **opts) -> nonnegative_integer
  *
  *  Opens the stream, writes the given +data+ to it,
  *  and closes the stream; returns the number of bytes written.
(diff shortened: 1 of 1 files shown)


io.c
```diff
@@ -12414,27 +12414,57 @@ io_s_write(int argc, VALUE *argv, VALUE klass, int binary)
...
+ *    path = 't.tmp'
+ *    File.write(path, File.read('t.txt')) # => 47
+ *    File.read(path)
+ *    # => "First line\nSecond line\n\nFourth line\nFifth line\n"
+ *    File.write(path, File.read('t.ja'))  # => 15
+ *    File.read(path)
+ *    # => "こんにちは"
+ *    File.write(path, File.read('t.dat')) # => 12
+ *    File.read(path)
+ *    # => "\xFE\xFF\x99\x90\x99\x91\x99\x92\x99\x93\x99\x94"
```
peterzhu2118: I find this example a bit confusing. Why are we reading a file and writing it to another one? Can we just directly write the string so it's more obvious what it's writing?

BurdetteLamar: Examples simplified.

peterzhu2118 (approved):

PR #16863, opened 2026-05-05
Ruby::Box fix stale cached values for exception-related global variables ($! and $@)

dak2:
Ruby::Box fix stale cached values for exception-related global variables (`$!` and `$@`)

The exception-related virtual variables `$!` (current exception) and `$@` (its backtrace) are stored on the execution context (ec->errinfo and the rescue/ensure frame's local slot accessed via errinfo_place), not in box->gvar_tbl. Caching their values in box->gvar_tbl makes the second read return a stale value from the previous raise/rescue:

```
begin; raise "first";  rescue; p $!; end
begin; raise "second"; rescue; p $!; end
# before: #<RuntimeError: first> / #<RuntimeError: first>
# after:  #<RuntimeError: first> / #<RuntimeError: second>
```

```
begin; raise "first";  rescue; p $@.first; end
begin; raise "second"; rescue; p $@.first; end
# before: same backtrace returned for both
# after:  distinct backtrace per raise
```

Fixes [Bug #21991](https://bugs.ruby-lang.org/issues/21991) 
Related PR: https://github.com/ruby/ruby/pull/16303

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -2224,6 +2224,9 @@ Init_eval(void)
     rb_gvar_ractor_local("$@");
     rb_gvar_ractor_local("$!");
 
+    rb_gvar_box_dynamic("$@");
+    rb_gvar_box_dynamic("$!");
+
     rb_define_global_function("raise", f_raise, -1);
     rb_define_global_function("fail", f_raise, -1);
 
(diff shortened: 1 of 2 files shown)


tagomoris (approved):
👍

PR #16865, opened 2026-05-05
Ensure Symbol#to_proc respects caller's Ruby::Box context

niku:
Ensure Symbol#to_proc respects caller's Ruby::Box context

[Bug #22015](https://bugs.ruby-lang.org/issues/22015)
During a normal method call, the VM creates a control frame, and the Ruby::Box associated with that frame determines the execution context.
However, Symbol#to_proc does not automatically create a frame.
Therefore, we need to manually push a frame associated with the appropriate Box context.

When determining the Box to associate with this new frame, we cannot simply look at the closest Ruby-level control frame.
This is because some built-in methods (like `Enumerable#map`) are implemented in Ruby (e.g., `<internal:enum>`).
These built-in Ruby frames belong to the default Root or Main Box, which would obscure the actual user-defined Box context we need to capture.

To fix this properly, we now traverse the frames and check their inherent Box states.
Since built-in Ruby frames always run in the Root or Main Box, we safely skip them by continuing our traversal until we find a user-defined Box (an Optional Box, indicated by `BOX_OPTIONAL_P`).

### DIFF
--- a/vm_insnhelper.c
+++ b/vm_insnhelper.c
@@ -5197,7 +5197,48 @@ rb_vm_yield_with_cfunc(rb_execution_context_t *ec, const struct rb_captured_bloc
 static VALUE
 vm_yield_with_symbol(rb_execution_context_t *ec,  VALUE symbol, int argc, const VALUE *argv, int kw_splat, VALUE block_handler)
 {
-    return rb_sym_proc_call(SYM2ID(symbol), argc, argv, kw_splat, rb_vm_bh_to_procval(ec, block_handler));
+    VALUE passed_proc = rb_vm_bh_to_procval(ec, block_handler);
+
+    if (!rb_box_available()) {
+        return rb_sym_proc_call(SYM2ID(symbol), argc, argv, kw_splat, passed_proc);
+    }
+
+    const rb_control_frame_t *reg_cfp = ec->cfp;
+    const rb_control_frame_t *ruby_cfp = rb_vm_get_ruby_level_next_cfp(ec, reg_cfp);
+    const rb_box_t *box = NULL;
+
+    /*
+     * Traverse the frames until a user-defined Box (Optional Box) is found.
+     * Frames for built-in methods like <internal:xxx> run in the Root or Main Box,
+     * so we can safely skip them with this condition without doing string comparisons.
+     */
+    while (ruby_cfp) {
+        box = current_box_on_cfp(ec, ruby_cfp);
+        if (BOX_OPTIONAL_P(box)) {
+            break;
+        }
+        ruby_cfp = rb_vm_get_ruby_level_next_cfp(ec, RUBY_VM_PREVIOUS_CONTROL_FRAME(ruby_cfp));
+    }
+
+    // Fallback to the normal call if no user-defined Box (Optional Box) is found
+    if (!ruby_cfp || !box) {
+        return rb_sym_proc_call(SYM2ID(symbol), argc, argv, kw_splat, passed_proc);
+    }
+
+    VALUE filename = rb_iseq_path(CFP_ISEQ(ruby_cfp));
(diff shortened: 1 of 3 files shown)


hsbt:
I rebased this onto current master (only a trivial context conflict in `test/ruby/test_box.rb`); the added test passes. But since the main/root box separation ([Feature #21881]), a plain `RUBY_BOX=1` script runs in the main box, which `BOX_OPTIONAL_P` does not match, so the [Bug #22015] scenario (`["a b"].flat_map(&:shellsplit)`) still raises `NoMethodError` on current master. Changing the traversal condition to `BOX_USER_P` fixes it; with that change `make btest` and `test/ruby/test_box.rb` are green, and rubygems' `bin/rspec` boots past this error under `RUBY_BOX=1`.

Two remaining points: yields from Ruby-defined builtins (e.g. `Array#map`) go through `vm_invoke_symbol_block` and are not covered, so `box.eval(%q{[[1]].map(&:box_method)})` still fails even with this patch. And `rb_iseq_new` on every yield may be a hot-path cost. Since this touches box core, the final call should be @tagomoris's.

A rebased branch with the `BOX_USER_P` change and a main-box test is at https://github.com/hsbt/ruby/tree/pr-16865-rebased-userbox.

vm_insnhelper.c
```diff
@@ -5197,7 +5197,47 @@ rb_vm_yield_with_cfunc(rb_execution_context_t *ec, const struct rb_captured_bloc
...
+    // Fallback to the normal call if no user-defined Box (Optional Box) is found
+    if (!ruby_cfp || !box) {
+        return rb_sym_proc_call(SYM2ID(symbol), argc, argv, kw_splat, passed_proc);
+    }
+
+    VALUE filename = rb_iseq_path(CFP_ISEQ(ruby_cfp));
+    const rb_iseq_t *iseq = rb_iseq_new(Qnil, filename, filename, Qnil, 0, ISEQ_TYPE_TOP);
+    VALUE val;
+
+    vm_push_frame(ec, iseq, VM_FRAME_MAGIC_TOP | VM_ENV_FLAG_LOCAL | VM_FRAME_FLAG_FINISH,
```
tagomoris, line 5231: Inserting `VM_FRAME_MAGIC_TOP` here looks too wild for me because it means the TOP of contexts, but this point is just the middle of method calls. If a code raises an exception on this frame, the backtrace will be really confusing.
(Yes, I used `VM_FRAME_MAGIC_TOP` in `Box#eval`. It is eval...)

How about building and pushing a frame of `VM_FRAME_MAGIC_BLOCK` stead of `MAGIC_TOP`?
We need to create a dummy ENV with a reference to the env of `ruby_cfp` and it is not easy, but the backtrace will be much more meaningful (it is equal to the one from usual Ruby code block argument).

tagomoris (commented):
@niku I'm very sorry for the late review.
And @hsbt I added a review comment about the frame type.
Could you check if it is reasoble?

vm_insnhelper.c
hsbt: I reworked this to push a `VM_FRAME_MAGIC_BLOCK` frame as suggested. The dummy ENV turned out to be simple: passing `VM_GUARDED_PREV_EP(ruby_cfp->ep)` as the specval is enough, and the box resolves through the ep chain. The backtrace is now identical in shape to a literal block call, showing `block in <main>` at the caller's line. `make btest`, `test/ruby/test_box.rb`, and rubygems' `bin/rspec` boot under `RUBY_BOX=1` are green.

hsbt:
Thanks for finding and fixing this. I merged your commits as-is in #18546, with two more fixes on top: matching the main box too after the main/root box separation, and pushing a block frame instead of a TOP frame as suggested in the review here. Closing this one as superseded by that PR.

PR #16868, opened 2026-05-05
[DOC] Doc for Pathname#binread

BurdetteLamar:


pathname_builtin.rb
```diff
@@ -1046,67 +1040,16 @@ def each_line(...) # :yield: line
...
-  #   # => "こんにちは"
-  #   Pathname.new('t.dat').read
-  #   # => "\xFE\xFF\x99\x90\x99\x91\x99\x92\x99\x93\x99\x94"
-  #
-  # On Windows, text mode can terminate reading and leave bytes in the file unread
-  # when encountering certain special bytes.
-  # Consider using #binread if all bytes in the file should be read.
-  #
-  # With argument +length+ given, returns +length+ bytes if available:
+  #   binread(length = nil, offset = 0) -> string or nil
```
peterzhu2118, line 1048: This call-seq is missing `opts`:


```suggestion
  #   binread(length = nil, offset = 0, **opts) -> string or nil
```

peterzhu2118: Ah never mind, `binread` doesn't accept a hash. My mistake.

peterzhu2118: Only `binwrite` accepts an `opts` hash. I fixed that in https://github.com/ruby/ruby/pull/16873

peterzhu2118 (approved):

PR #16871, opened 2026-05-05
Relocate timestamps page

BurdetteLamar:
Eliminates needless structure.

Was in doc/file/; now in doc/, and doc/file/ removed.


peterzhu2118 (commented):
I'm not a huge fan of this change. We should keep specific docs in its own directory rather than putting it in the top level `doc` directory.

PR #16888, opened 2026-05-07
ZJIT: Convert ZJIT HIR Extended basic blocks to traditional basic blocks

### LINKED ISSUE #959: ZJIT: Proposal to convert EBB to conventional BB in HIR
TL;DR: I think converting EBB to regular BB would simplify our current code and also enable us to do more advanced compiler passes more easily

## Background

HIR currently uses extended basic blocks (EBBs). An EBB allows branches (e.g. `IfTrue`, `IfFalse`) to appear anywhere within a block, not just at the end. The LIR, by contrast, uses conventional basic blocks where only the last 0, 1, or 2 instructions can be terminators.

This mismatch creates unnecessary complexity and I'd like to propose converting HIR to use traditional basic blocks.

### 1. Unify the block model between HIR and LIR

LIR already uses traditional basic blocks. The codegen lowering pass must split every HIR EBB at each mid-block branch, creating a new fall-through LIR block.

Multiple LIR blocks end up sharing the same `hir_block_id`, which complicates any mapping between the two IRs.

If HIR used the same block model, lowering would be a straightforward 1:1 block mapping.

### 2. Reuse data flow algorithms between HIR and LIR

LIR has liveness analysis, critical edge splitting , and successor/predecessor computation that all assume traditional basic blocks. HIR has its own `ControlFlowInfo` that must work around EBBs.

With a shared block model, these algorithms could be written once and reused, or at minimum follow the same patterns and assumptions.

### 3. Simplify CFG analysis

With EBBs, computing successors requires [scanning every instruction in a block](https://github.com/ruby/ruby/blob/a897e06db526996c55451c5d7d5049013b76e1a7/zjit/src/hir.rs#L8471-L8483):

```
// Since ZJIT uses extended basic blocks, one must check all instructions
// for their ability to jump to another basic block, rather than just
// the instructions at the end of a given basic block.
```

This affects every pass that needs CFG information:
- `ControlFlowInfo::new()`: successor/predecessor maps
- `po_from()`: reverse post-order traversal
- `clean_cfg()`: CFG cleanup / dead block elimination
- `fold_constants()`: must break on mid-block terminators

With conventional basic blocks, successors are determined solely by the last 2 instructions.

### 4. Enable standard compiler optimizations

Many textbook optimizations assume traditional basic blocks:
- Dominator tree construction
- Loop detection and LICM
- GVN/GCM (Global Code Motion)
- Tail merging / block deduplication

These are simpler to implement and reason about when each block has exactly one entry point and one exit point.

## How hard is it?

I think we mostly need to:

* Change HIR construction to create new blocks as we're emitting (when we emit branching instructions like `IfTrue` / `IfFalse`).
* Eliminate the block splitting code in LIR.
* Update `ControlFlowInfo`
* Fix tests

I think everything else in HIR would stay mostly the same (BranchEdge, SSA param passing, etc).

tenderlove:
This PR changes the basic blocks used in HIR from extended basic blocks to traditional basic blocks.  This means all basic blocks in HIR are guaranteed to end with some kind of control flow (jump / return), and there are no mid-block control flow instructions.

This PR introduces a new instruction called `CondBranch` that replaces `IfTrue` and `IfFalse`.  The new instruction has `if_true` and `if_false` arms.  `CondBranch` is a terminator instruction, so it ends blocks.  All HIR blocks will now end with a terminator instruction, and no blocks will have "mid-block" control flow.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -2695,6 +2695,11 @@ impl Function {
             Insn::Jump(edge) => vec![edge.target],
             Insn::Entries { targets } => targets,
             Insn::Unreachable | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => vec![],
+            // Blocks that don't end with terminators are technically errors,
+            // every block in the CFG should end with a terminator. But we
+            // want to be able to iterate over poorly constructed CFG when
+            // debugging, so we'll return an empty vec.  The validation
+            // routines check for terminators, so we should catch CFG errors there.
             _ => vec![]
         }
     }
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -5721,20 +5721,6 @@ impl Function {
             Ok(())
         };
 
-        // The `rpo` function calls `successors`, and `successors` will blow up
-        // if a block doesn't end with a terminator.  First we'll iterate
-        // through all blocks (regardless of reachability) and assert that they
-        // end with a terminator.  Then we should be safe to use `rpo`.
-        for (idx, block) in self.blocks.iter().enumerate() {
-            let insns = &block.insns;
-            if let Some(insn_id) = insns.last() {
-                let last = self.find(*insn_id);
-                if !last.is_terminator() {
-                    return Err(ValidationError::BlockHasNoTerminator(BlockId(idx)));
-                }
-            }
-        }
-
         for block_id in self.rpo() {
             let insns = &self.blocks[block_id.0].insns;
             for (idx, insn_id) in insns.iter().enumerate() {
@@ -5750,11 +5736,18 @@ impl Function {
                     }
                     _ => {}
                 }
+
                 if insn.is_terminator() {
                     // Blow up if we have a terminator that isn't at the end
                     // of the block.
                     if idx != insns.len() - 1 {
                         return Err(ValidationError::TerminatorNotAtEnd(block_id, *insn_id, idx))
+                    }
+                }
+                // If the last instruction isn't a terminator, return an error
+                if idx == insns.len() - 1 {
+                    if !insn.is_terminator() {
+                        return Err(ValidationError::BlockHasNoTerminator(block_id));
                     }
                 }
             }
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -7131,7 +7131,6 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                     let not_nil_false = fun.push_insn(block, Insn::RefineType { val, new_type: not_nil_false_type });
                     state.replace(val, not_nil_false);
                     queue.push_back((state.clone(), target, target_idx, local_inval));
-
                 }
                 YARVINSN_branchif | YARVINSN_branchif_without_ints => {
                     if opcode == YARVINSN_branchif {
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -7,11 +7,11 @@
 #![allow(clippy::match_like_matches_macro)]
 use crate::{
     backend::lir::C_ARG_OPNDS,
-    cast::IntoUsize, codegen::{local_idx_to_ep_offset, max_iseq_versions}, cruby::*, invariants::{self}, payload::{get_or_create_iseq_payload, IseqPayload}, options::{debug, get_option, DumpHIR}, state::ZJITState, json::Json,
+    cast::IntoUsize, codegen::{local_idx_to_ep_offset, max_iseq_versions}, cruby::*, invariants::{self}, payload::{get_or_create_iseq_payload, IseqPayload}, options::{get_option, DumpHIR}, state::ZJITState, json::Json,
     state,
 };
 use std::{
-    cell::RefCell, collections::{BTreeSet, HashMap, HashSet, VecDeque}, ffi::{c_void, c_uint, c_int, CStr}, fmt::Display, mem::{align_of, size_of}, ptr, slice::Iter,
+    cell::RefCell, collections::{HashMap, HashSet, VecDeque}, ffi::{c_void, c_uint, c_int, CStr}, fmt::Display, mem::{align_of, size_of}, ptr, slice::Iter,
     sync::atomic::Ordering,
 };
 use crate::hir_type::{Type, types};
@@ -952,9 +952,8 @@ pub enum Insn {
     /// Unconditional jump
     Jump(BranchEdge),
 
-    /// Conditional branch instructions
-    IfTrue { val: InsnId, target: BranchEdge },
-    IfFalse { val: InsnId, target: BranchEdge },
+    /// Conditional branch
+    CondBranch { val: InsnId, if_true: BranchEdge, if_false: BranchEdge },
 
     /// Call a C function without pushing a frame
     /// `name` and `owner` are for printing purposes only
@@ -1328,10 +1327,10 @@ macro_rules! for_each_operand_impl {
             Insn::Jump(BranchEdge { args, .. }) => {
                 $visit_many!(args);
             }
-            Insn::IfTrue { val, target: BranchEdge { args, .. } }
-            | Insn::IfFalse { val, target: BranchEdge { args, .. } } => {
+            Insn::CondBranch { val, if_true: BranchEdge { args: true_args, .. }, if_false: BranchEdge { args: false_args, .. } } => {
                 $visit_one!(val);
-                $visit_many!(args);
+                $visit_many!(true_args);
+                $visit_many!(false_args);
             }
             Insn::ArrayDup { val, state }
             | Insn::Throw { val, state, .. }
@@ -1468,7 +1467,7 @@ impl Insn {
         match self {
             Insn::Jump(_)
             | Insn::Entries { .. }
-            | Insn::IfTrue { .. } | Insn::IfFalse { .. } | Insn::EntryPoint { .. } | Insn::Return { .. }
+            | Insn::CondBranch { .. } | Insn::EntryPoint { .. } | Insn::Return { .. }
             | Insn::PatchPoint { .. } | Insn::SetIvar { .. } | Insn::SetClassVar { .. } | Insn::ArrayExtend { .. }
             | Insn::ArrayPush { .. } | Insn::SideExit { .. } | Insn::SetGlobal { .. }
             | Insn::SetLocal { .. } | Insn::Throw { .. } | Insn::IncrCounter(_) | Insn::IncrCounterPtr { .. }
@@ -1482,7 +1481,7 @@ impl Insn {
     /// Return true if the instruction ends a basic block and false otherwise.
     pub fn is_terminator(&self) -> bool {
         match self {
-            Insn::Jump(_) | Insn::Entries { .. } | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => true,
+            Insn::Unreachable | Insn::CondBranch { .. } | Insn::Jump(_) | Insn::Entries { .. } | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => true,
             _ => false,
         }
     }
@@ -1490,7 +1489,7 @@ impl Insn {
     /// Return true if the instruction is a jump (has successor blocks in the CFG).
     pub fn is_jump(&self) -> bool {
         match self {
-            Insn::IfTrue { .. } | Insn::IfFalse { .. } | Insn::Jump(_) | Insn::Entries { .. } => true,
+            Insn::CondBranch { .. } | Insn::Jump(_) | Insn::Entries { .. } => true,
             _ => false,
         }
     }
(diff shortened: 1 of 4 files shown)


zjit/src/hir.rs
```diff
@@ -2700,6 +2697,32 @@ impl Function {
...
+    fn num_successors(&self, block: BlockId) -> usize {
+        let insns = &self.blocks[block.0].insns;
+        let last = self.find(*insns.last().unwrap());
+        match last {
+            Insn::CondBranch { .. } => 2,
+            Insn::Jump(_) => 1,
+            Insn::Entries { targets } => targets.len(),
+            Insn::Unreachable | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => 0,
+            _ => 0
+            // _ => unreachable!("block doesn't end with a jump or terminator")
```
tenderlove: Should this be unreachable or return 0?

zjit/src/hir.rs
```diff
@@ -2700,6 +2697,32 @@ impl Function {
...
+
+    fn successors(&self, block: BlockId) -> Vec<BlockId> {
+        let insns = &self.blocks[block.0].insns;
+        let last = self.find(*insns.last().unwrap());
+        match last {
+            Insn::CondBranch { if_true, if_false, .. } => vec![if_true.target, if_false.target],
+            Insn::Jump(edge) => vec![edge.target],
+            Insn::Entries { targets } => targets,
+            Insn::Unreachable | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => vec![],
+            _ => vec![]
```
tenderlove, line 2703: Same question

zjit/src/hir.rs
nirvdrum: I like unreachable here.

zjit/src/hir.rs
nirvdrum: I like the null object pattern in a lot of cases, but here this looks more exceptional and if we end up in an unexpected state I'd rather the compiler hard fault so we can investigate.

zjit/src/hir.rs
```diff
@@ -5247,6 +5264,9 @@ impl Function {
     }
 
     fn absorb_dst_block(&mut self, num_in_edges: &[u32], block: BlockId) -> bool {
+        if self.num_successors(block) > 1 {
```
tekknolagi: replace with `if != Jump { return false }`

zjit/src/hir.rs
tekknolagi: Torn because we also want to be able to navigate (e.g. print) the graph in exceptional circumstances without panicking. I think I lean toward empty vec (optionally debug print) and lean on validate pass

zjit/src/hir.rs
```diff
@@ -3034,35 +3057,33 @@ impl Function {
                     let insn_id = self.blocks[block.0].insns[i];
                     // Instructions without output, including branch instructions, can't be targets
                     // of make_equal_to, so we don't need find() here.
-                    let insn_type = match &self.insns[insn_id.0] {
-                        &Insn::IfTrue { val, target: BranchEdge { target, ref args } } => {
+                    // Clone the insn so we don't hold a borrow on self.insns
+                    // across calls to set_type() below.
+                    let insn_type = match self.insns[insn_id.0].clone() {
```
tekknolagi: pls figure out how to remove `clone`; is expensive

zjit/src/hir.rs
```diff
@@ -5668,35 +5660,52 @@ impl Function {
...
+        let check_edge = |block_id: BlockId, edge: &BranchEdge| -> Result<(), ValidationError> {
+            let target_len = self.blocks[edge.target.0].params.len();
+            let args_len = edge.args.len();
+            if target_len != args_len {
+                return Err(ValidationError::MismatchedBlockArity(block_id, target_len, args_len));
+            }
+            Ok(())
+        };
+
+        // The `rpo` function calls `successors`, and `successors` will blow up
```
tekknolagi: this can be undone if we make `successors` not panic

zjit/src/hir.rs
```diff
@@ -5668,35 +5660,52 @@ impl Function {
...
-                }
-                block_has_terminator = true;
-                if idx != insns.len() - 1 {
-                    return Err(ValidationError::TerminatorNotAtEnd(block_id, *insn_id, idx));
+                if insn.is_terminator() {
+                    // Blow up if we have a terminator that isn't at the end
+                    // of the block.
+                    if idx != insns.len() - 1 {
+                        return Err(ValidationError::TerminatorNotAtEnd(block_id, *insn_id, idx))
+                    }
```
tekknolagi, line 5750: maybe add a separate check now, end of loop, taht the last instruction passes `is_terminator`

zjit/src/hir.rs
```diff
@@ -7062,14 +7070,21 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
...
+                        if_false: BranchEdge { target, args: iffalse_state.as_args(self_param) }
                     });
+
+                    block = fall_through;
+
                     let not_nil_false_type = types::Truthy;
                     let not_nil_false = fun.push_insn(block, Insn::RefineType { val, new_type: not_nil_false_type });
                     state.replace(val, not_nil_false);
                     queue.push_back((state.clone(), target, target_idx, local_inval));
+
```
tekknolagi: nit: remove

zjit/src/hir.rs
```diff
@@ -8313,10 +8379,16 @@ fn compile_entry_block(fun: &mut Function, jit_entry_insns: &[u32], insn_idx_to_
             val: Const::CPtr(unsafe { rb_iseq_pc_at_idx(fun.iseq, jit_entry_insn) } as *const u8),
         });
         let test_id = fun.push_insn(entry_block, Insn::IsBitEqual { left: pc, right: expected_pc });
-        fun.push_insn(entry_block, Insn::IfTrue {
+
+        // TODO: is this the right insn index??
```
tekknolagi: note to self to come back to this

zjit/src/hir.rs
k0kubun: +1 for unreachable. Do we ever use this 0 associated to non-terminator instructions?

zjit/src/hir.rs
```diff
@@ -8285,15 +8351,15 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
 
     fun.profiles = Some(profiles);
     if let Err(err) = crate::stats::trace_compile_phase("validate", || fun.validate()) {
-        debug!("ZJIT: {err:?}: Initial HIR:\n{}", FunctionPrinter::without_snapshot(&fun));
+        eprintln!("ZJIT: {err:?}: Initial HIR:\n{}", FunctionPrinter::without_snapshot(&fun));
```
k0kubun: Do we need this change? IIRC we don't have anything else in YJIT or ZJIT that is this verbose by default.

tekknolagi: this was made during debugging iirc

zjit/src/hir.rs
k0kubun: I guess ideally we want to assign the next `jit_entry_insn` (or the insn index after optionals for the last `jit_entry_insn`) to the fall_through block, but IIRC we use this only for HIR/LIR/asm dumps, so we shouldn't worry too much about it if it's complicated to fix?

zjit/src/hir.rs
tenderworks: Ah sorry.  I shouldn't have checked this in.  I'll remove it.

k0kubun (approved):
https://github.com/ruby/ruby/pull/16888#pullrequestreview-4275099117 should probably be fixed and I don't think we want this https://github.com/ruby/ruby/pull/16888#discussion_r3229046160, but other than that, it looks good to me.

zjit/src/hir.rs
tenderlove: I deleted this

zjit/src/hir.rs
tenderlove: I deleted this particular function, so I think we should decide in the `successors` method.

zjit/src/hir.rs
tenderlove: This is inside the validation routine though.  I think we _want_ to find all blocks that don't have terminators and return an error.

zjit/src/hir.rs
tenderlove: I don't think we want to do that.  Then we wouldn't catch "mid block jumps" in the validation routine.

zjit/src/hir/opt_tests.rs
```diff
@@ -15539,7 +15584,7 @@ mod hir_opt_tests {
 
         // After profiling via the side exit, rebuilding HIR should now
         // have a SendDirect for greet_recompile instead of SideExit.
-        assert_snapshot!(hir_string("test_no_profile_recompile"), @r"
+        assert_snapshot!(hir_string("test_no_profile_recompile"), @"
```
tekknolagi: can you please upgrade your insta and revert the `r` changes?

zjit/src/hir.rs
tenderlove: Maybe we should have a debugging version?  I'd hate for our normal compilation pass to "happen to work" with an invalid CFG, and I'd prefer if we had a hard crash.

tekknolagi: The debugging version calls `validate` between each opt pass. I think we should be ok

tekknolagi: If you want to split `successors` and maintain two versions of this function, of `rpo`, and whatever else display/etc need, I guess we can do that. I am ok with more letting stuff slide and catching in validate

nirvdrum: That's fair. My concern would be it goes undetected until it causes issues somewhere else far removed. At least if it aborts I know exactly where and I can at that point add whatever logging or debug working backwards. Unless I'm mistaken, these are cases that should never happen.

With that said, if a separate validation pass can catch it, then I can get on board. I'm just paranoid because I've seen the separation of the offender and the validator fall out of sync or accidentally get removed on other projects.

tenderlove: > The debugging version calls `validate` between each opt pass. I think we should be ok

I didn't realize that.  If this is the case then I'm good with this being a null object.  I don't think we should remove the changes to the validator below though.

zjit/src/hir.rs
tekknolagi: not all blocks. non-reachable ones may be all sorts of malformed. i recommend doing `rpo` and only validating those

tekknolagi: basically moving this terminator check into the following loop, either before or after the other insn loop

zjit/src/hir.rs
```diff
@@ -2690,6 +2687,19 @@ impl Function {
...
+    fn successors(&self, block: BlockId) -> Vec<BlockId> {
+        let insns = &self.blocks[block.0].insns;
+        let last = self.find(*insns.last().unwrap());
+        match last {
+            Insn::CondBranch { if_true, if_false, .. } => vec![if_true.target, if_false.target],
+            Insn::Jump(edge) => vec![edge.target],
+            Insn::Entries { targets } => targets,
+            Insn::Unreachable | Insn::Return { .. } | Insn::SideExit { .. } | Insn::Throw { .. } => vec![],
+            _ => vec![]
+            // _ => unreachable!("block doesn't end with a jump or terminator")
```
tekknolagi: if we remove this commented code (maybe add an informative comment expressing intent) we should be good to land

zjit/src/hir.rs
tenderlove: Fixed

zjit/src/hir.rs
tenderlove: I removed the commented code already.  I'll add a comment above the `_` case.

zjit/src/hir.rs
tenderlove: I used a peekabable iterator to solve this.  Thanks!

zjit/src/hir.rs
tekknolagi: looks like we did

tekknolagi:
land ahoy

XrXr:
Did you run benchmarks?

tenderlove:
@XrXr I hadn't, but here are the headline benchmarks. "before" is the commit before this landed, "after" is with this commit:

```
before: ruby 4.1.0dev (2026-05-12T20:39:11Z hir-bb ece14b61f5) +ZJIT +PRISM [arm64-darwin25]
after: ruby 4.1.0dev (2026-05-12T23:16:13Z master be557259b7) +ZJIT +PRISM [arm64-darwin25]

--------------  ------------  ------------  -------------  ------------
bench            before (ms)    after (ms)  after 1st itr  before/after
liquid-c                 N/A           N/A            N/A           N/A
activerecord     81.3 ± 8.8%   80.5 ± 1.5%          0.946         1.009
chunky-png      292.0 ± 0.4%  291.5 ± 0.7%          1.008         1.002
erubi-rails     346.6 ± 0.4%  356.3 ± 0.3%          0.973         0.973
hexapdf         843.9 ± 0.6%  842.4 ± 0.8%          1.005         1.002
liquid-compile   21.6 ± 8.2%   21.4 ± 8.6%          1.002         1.008
liquid-render    49.0 ± 3.2%   49.4 ± 3.2%          0.975         0.994
lobsters        343.2 ± 2.2%  348.9 ± 3.6%          0.983         0.984
mail             55.6 ± 3.3%   55.3 ± 3.5%          0.977         1.006
psych-load      805.0 ± 1.1%  822.4 ± 2.0%          0.991         0.979
railsbench      506.7 ± 2.4%  509.9 ± 1.9%          0.969         0.994
rubocop         75.9 ± 12.9%  78.5 ± 12.0%          0.989         0.967
ruby-lsp         62.4 ± 1.5%   62.7 ± 3.4%          1.003         0.995
sequel           23.6 ± 4.7%   23.9 ± 5.2%          0.881         0.986
shipit          540.6 ± 2.7%  540.5 ± 2.3%          1.074         1.000
--------------  ------------  ------------  -------------  ------------

Legend:
```

PR #16892, opened 2026-05-08
Give the fiber pool its own mutex to allow sweep thread to acquire it

luke-gruber:
Non-ruby threads can't acquire the VM lock, so use another lock. We can't do allocations inside this lock, so use mimmalloc the one time where this is necessary. This change will also benefit MMTk because it uses worker threads for sweeping.

### DIFF
--- a/cont.c
+++ b/cont.c
@@ -291,12 +291,69 @@ rb_free_shared_fiber_pool(void)
     struct fiber_pool_allocation *allocations = shared_fiber_pool.allocations;
     while (allocations) {
         struct fiber_pool_allocation *next = allocations->next;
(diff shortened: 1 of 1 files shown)


ioquatix (changes requested):
For sure the fiber pool is not global, so protecting it with a global lock feels wrong to me.

Also, have you evaluated the performance impact?

luke-gruber:
I assume you're referring to `EXPERIMENTAL_FIBER_POOL`. You're right, I'll change it to be local to a pool.

> Also, have you evaluated the performance impact?

Without concurrent sweeping, it is negligible because it is uncontended. It's on the order of nanoseconds per acquisition.

PR #16896, opened 2026-05-08
Fix gdb rb_ps helper

swebb:
Over time the .gdbinit initializer has drifted from the codebase and the rb_ps helper no longer works. This PR fixes it. The changes that caused it to break were:

* 226f37059ec5f3ea3a1417e0bab630c64dbc8ac3 renamed cfp->iseq to cfp->_iseq.
* 6c24904a690eb7c4e20c3fa8c3751acc03454100 switched from storing the last_id to storing the next_id.
* f7ae32ed3b5b93247f9f62a58e3dd129098d0b27 removed ID_ENTRY_SIZE.

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -979,7 +979,7 @@ end
 
 define print_lineno
   set $cfp = $arg0
-  set $iseq = $cfp->iseq
+  set $iseq = rb_get_cfp_iseq($cfp)
   set $pos = $cfp->pc - $iseq->body->iseq_encoded
   if $pos != 0
     set $pos = $pos - 1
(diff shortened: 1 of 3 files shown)


.gdbinit
```diff
@@ -1083,7 +1083,7 @@ define print_id
           set $aryptr = $ary->as.heap.ptr
           set $arylen = $ary->as.heap.len
         end
-        set $result = $aryptr[($serial % ID_ENTRY_UNIT) * ID_ENTRY_SIZE + $t]
+        set $result = $aryptr[($serial % ID_ENTRY_UNIT) + $t]
```
k0kubun, line 1086: What is this change about? Can you link a reference to help me understand this?

k0kubun: :eyes: https://github.com/ruby/ruby/commit/f7ae32ed3b5b93247f9f62a58e3dd129098d0b27

k0kubun: It at least seems like how the use of `ID_ENTRY_SIZE` is converted, so the change seems sane.

k0kubun (approved):

PR #16898, opened 2026-05-08
Protect th->keeping_mutexes when freeing mutex + thread concurrently

luke-gruber:
When freeing a mutex, we unlink any locked mutex from its th->keeping_mutexes. When we free a thread, we unlink all mutexes from its th->keeping_mutexes. If we free concurrently from 2+ threads, we need to protect th->keeping_mutexes from corruption.

### DIFF
--- a/thread.c
+++ b/thread.c
@@ -446,14 +446,15 @@ rb_threadptr_join_list_wakeup(rb_thread_t *thread)
     }
 }
 
-// This lock is for protecting thread->keeping_mutexes when mutex and thread are concurrently freed
-static void keeping_mutexes_lock_lock(void);
-static void keeping_mutexes_lock_unlock(void);
+// This lock protects thread->keeping_mutexes when a mutex and its owning
+// thread are concurrently freed.
+static void keeping_mutexes_lock_lock(rb_thread_t *th);
+static void keeping_mutexes_lock_unlock(rb_thread_t *th);
 
 void
 rb_threadptr_unlock_all_locking_mutexes(rb_thread_t *th)
 {
-    keeping_mutexes_lock_lock();
+    keeping_mutexes_lock_lock(th);
     {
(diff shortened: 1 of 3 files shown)


thread.c
```diff
@@ -446,18 +446,31 @@ rb_threadptr_join_list_wakeup(rb_thread_t *thread)
...
-        rb_mutex_t *mutex = th->keeping_mutexes;
-        th->keeping_mutexes = mutex->next_mutex;
-
-        // rb_warn("mutex #<%p> was not unlocked by thread #<%p>", (void *)mutex, (void*)th);
-        VM_ASSERT(mutex->ec_serial);
-        const char *error_message = rb_mutex_unlock_th(mutex, th, 0);
-        if (error_message) rb_bug("invalid keeping_mutexes: %s", error_message);
+    keeping_mutexes_lock_lock();
+    {
+        while (th->keeping_mutexes) {
```
samuel-williams-shopify, line 459: We lock the mutex even if `th->keeping_mutexes` is null?

luke-gruber: Most of the time this will be uncontended and it will always be uncontended unless concurrent sweeping. We do need to lock it before touching `th->keeping_mutexes` because it may be modified by another thread. I did change it to make the lock per-thread so that should reduce contention further when concurrent sweeping.

samuel-williams-shopify:
I think you should clearly explain the motivation in the code (e.g. we are introducing additional locking for parallel GC).

Also is this only a matter of free? What about other places where keeping_mutexes is manipulated? I don't think the assumptions are documented clearly.

luke-gruber:
I'll add more documentation, you're right to point it out.  I've also created a [ticket](https://bugs.ruby-lang.org/issues/22067) that adds more context to freeing `T_DATA`s in parallel.

PR #16900, opened 2026-05-08
Add autoload_data_lock to protect autoload_data when free concurrently

luke-gruber:
This lock protects autoload_data->constants from being corrupted when concurrently freeing autoload_data and its associated autoload_const.


samuel-williams-shopify:
Is it really a good idea to introduce a global lock?

luke-gruber:
This global lock is only acquired on free calls, so right now of course there is no contention because there is only 1 thread that frees, which is the Ruby thread that calls GC. This is needed, however, for when there are multiple sweeping threads if we want to free T_DATAs concurrently.

PR #16901, opened 2026-05-08
ZJIT: add an unreachable instruction

tenderlove:
Unreachable instructions terminate blocks.  We'll use this mostly for testing as a terminator instruction (since traditional BB's will require all blocks to end with a terminator)

### DIFF
--- a/zjit/src/backend/arm64/mod.rs
+++ b/zjit/src/backend/arm64/mod.rs
@@ -1561,6 +1561,9 @@ impl Assembler {
                 Insn::Breakpoint => {
                     brk(cb, A64Opnd::None);
                 },
+                Insn::Abort => {
+                    udf(cb, u16::MAX);
+                },
                 Insn::CSelZ { truthy, falsy, out } |
                 Insn::CSelE { truthy, falsy, out } => {
                     csel(cb, out.into(), truthy.into(), falsy.into(), Condition::EQ);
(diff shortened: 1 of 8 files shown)


zjit/src/hir.rs
```diff
@@ -1147,6 +1147,7 @@ pub enum Insn {
     CheckInterrupts { state: InsnId },
 
     BreakPoint,
+    Unreachable,
```
k0kubun, line 1154: Could you leave a comment above this that explains how to use this instruction and that this is not supposed to be generated from `iseq_to_hir`?

k0kubun (approved):

zjit/src/backend/arm64/mod.rs
```diff
@@ -1561,6 +1561,9 @@ impl Assembler {
                 Insn::Breakpoint => {
                     brk(cb, A64Opnd::None);
                 },
+                Insn::Abort => {
+                    udf(cb, 0);
```
XrXr: nit: I'd prefer this to not be all zeros so it doesn't look uninitialized in a memory dump.

```suggestion
                    udf(cb, u16::MAX);
```

XrXr (approved):

zjit/src/hir.rs
tenderlove: Yep, I'll add that!

tekknolagi: We will likely use it in the optimizer though

tekknolagi:
Merge conflict unfortunately

PR #16905, opened 2026-05-08
Set EC in rb_gc_event_hook

peterzhu2118:
The event hook may use the EC and it will be null when it is running from a GC thread.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -242,7 +242,30 @@ rb_gc_event_hook(VALUE obj, rb_event_flag_t event)
     rb_execution_context_t *ec = rb_gc_get_ec();
     if (!ec->cfp) return;
 
+#if USE_MODULAR_GC
+    bool gc_thread_p = false;
+    if (!GET_EC()) {
+        gc_thread_p = true;
+
+# ifdef RB_THREAD_LOCAL_SPECIFIER
+        rb_current_ec_set(ec);
+# else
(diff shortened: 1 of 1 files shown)


jhawthorn:
> The event hook may use the EC

Which event hooks use the EC? I've been trying to push back against us accommodating bad C extensions which use this internal API in unsupported ways (I am happy to have these arguments and suggest fixes to these gems 😊).

From our docs:

> However, in  internal events,  you can  not use any  Ruby APIs  (even object
> creations). [...] Limitations are MRI version specific.

I think "you can  not use any  Ruby APIs" makes it very clear that anything that uses EC is invalid. The one exception to this is I think NEWOBJ should be run with an EC (to allow tracing the allocation location).

peterzhu2118:
> Which event hooks use the EC?

`freeobj_i` calls `rb_tracearg_from_tracepoint`:

https://github.com/ruby/ruby/blob/da9e38460a776dd8b10d7f42d4630a1e5f3dc455/ext/objspace/object_tracing.c#L128

`rb_tracearg_from_tracepoint` uses `get_trace_arg`, which uses the EC to store the `trace_arg`:

https://github.com/ruby/ruby/blob/da9e38460a776dd8b10d7f42d4630a1e5f3dc455/vm_trace.c#L963

Maybe we can store the `trace_arg` elsewhere?

PR #16906, opened 2026-05-09
Allocate T_MATCH storage embedded within object

jhawthorn:
Previously all `T_MATCH` (`MatchData`) objects would require separate malloc memory to store the captured position data. This PR changes this to store this with the T_MATCH object using VWA (Variable Width Allocation) whenever possible. Must of the work necessary was done in #16903, by avoiding internal use of the rmatch->regs struct.

In order to support objects too long to fit within our current size classes (greater than ~60 match groups) or if a C extension calls RMATCH_REGS (which I'd like to deprecate, but will propose that separately), we have a flag to indicate that we'll use `struct re_registers` with malloc for storage as this used to, similar to the "heap" storage for ex. Array and String.

This aims to make allocation faster (avoiding `malloc`), usage faster (avoiding a pointer chase, better cache), and GC faster (avoiding `free`).

Additionally, our matching functions use a stack-allocated buffer (backed by `ALLOCV_N`, so it can use a GC-backed object if very large) for the initial capture group storage, so we can avoid that allocation too (I previously proposed this in #16183).

Co-authored with @byroot

### DIFF
--- a/benchmark/vm_regexp.yml
+++ b/benchmark/vm_regexp.yml
@@ -3,6 +3,12 @@ prelude: |
 benchmark:
   vm_regexp: |
     /hoge/ =~ str
+  vm_regexp_alternating: |
+    /hoge/ =~ str
+    /huge/ =~ str
   vm_regexp_invert: |
     str =~ /hoge/
+  vm_regexp_invert_alternating: |
+    str =~ /hoge/
+    str =~ /huge/
 loop_count: 6000000
--- a/depend
+++ b/depend
@@ -5952,6 +5952,7 @@ gc.$(OBJEXT): $(top_srcdir)/internal/numeric.h
 gc.$(OBJEXT): $(top_srcdir)/internal/object.h
 gc.$(OBJEXT): $(top_srcdir)/internal/proc.h
 gc.$(OBJEXT): $(top_srcdir)/internal/rational.h
+gc.$(OBJEXT): $(top_srcdir)/internal/re.h
 gc.$(OBJEXT): $(top_srcdir)/internal/sanitizers.h
 gc.$(OBJEXT): $(top_srcdir)/internal/serial.h
 gc.$(OBJEXT): $(top_srcdir)/internal/set_table.h
--- a/gc.c
+++ b/gc.c
@@ -100,6 +100,7 @@
 #include "internal/object.h"
 #include "internal/proc.h"
 #include "internal/rational.h"
+#include "internal/re.h"
 #include "internal/sanitizers.h"
 #include "internal/struct.h"
 #include "internal/symbol.h"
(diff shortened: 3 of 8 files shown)


jhawthorn:
Overly simplistic benchmark:

```
$ ips --ruby ./miniruby_before --ruby ./miniruby_after --yjit -e '/o/.match("foo")'
ruby 4.1.0dev (2026-05-11T10:25:29Z master d7ef97204d) +YJIT dev_nodebug +PRISM [x86_64-linux]
    /o/.match("foo"):     9.582M i/s (± 2.6%, GC 22.9%)

ruby 4.1.0dev (2026-05-11T17:20:04Z t_match_embedded 6c144833df) +YJIT dev_nodebug +PRISM [x86_64-linux]
    /o/.match("foo"):    16.519M i/s (± 0.3%, GC  9.0%)


Summary
  ruby ./miniruby_after ran
    1.72 ± 0.05 times faster than ruby ./miniruby_before
```

jhawthorn:
```
$ make benchmark ITEM=vm_regexp COMPARE_RUBY=$(which ruby)
/home/jhawthorn/.rubies/ruby-dev/bin/ruby --disable=gems -rrubygems -I./benchmark/lib ./benchmark/benchmark-driver/exe/benchmark-driver \
            --executables="compare-ruby::/home/jhawthorn/.rubies/ruby-dev/bin/ruby -I.ext/common --disable-gem" \
            --output=markdown --output-compare -v $(find ./benchmark -maxdepth 1 -name 'vm_regexp' -o -name '*vm_regexp*.yml' -o -name '*vm_regexp*.rb' | sort)
compare-ruby: ruby 4.1.0dev (2026-05-11T10:25:29Z master d7ef97204d) +PRISM [x86_64-linux]
built-ruby: ruby 4.1.0dev (2026-05-11T17:20:04Z t_match_embedded 6c144833df) +PRISM [x86_64-linux]
# Iteration per second (i/s)

|                  |compare-ruby|built-ruby|
|:-----------------|-----------:|---------:|
|vm_regexp         |     11.868M|   12.997M|
|                  |           -|     1.10x|
|vm_regexp_invert  |     11.860M|   12.802M|
|                  |           -|     1.08x|
```

Fewer benefits (but still ~10%) from this benchmark likely because it's used in a way where it's able to reuse the existing match.

byroot:
> it's used in a way where it's able to reuse the existing match.

As mentioned during pairing, I think we still could in most case. But perhaps if we manage to make T_MATC fully embedded, then the GC cost will be small enough that this optimization to re-use match object wouldn't be worth it anymore?

luke-gruber:
I believe `match_set_regs` should sometimes `ALLOCV_END` the buffer before it raises OOM.

jhawthorn:
> I believe `match_set_regs` should sometimes `ALLOCV_END` the buffer before it raises OOM.

`ALLOCV_END` is optional. If we don't call it the next GC cycle will clean up.

jhawthorn:
> As mentioned during pairing, I think we still could in most case. But perhaps if we manage to make T_MATC fully embedded, then the GC cost will be small enough that this optimization to re-use match object wouldn't be worth it anymore?

This is still reusing that match. I tested removing it which made vm_regexp 6-10% slower (so still faster than master, but fastest with both optimizations). So seems worth keeping for now unless there is motivation to remove it.

PR #16907, opened 2026-05-09
Improve Pathname by migrating internal methods to pathname.c

nobu:
[[Feature #22060]](https://bugs.ruby-lang.org/issues/22060)

### DIFF
--- a/pathname.c
+++ b/pathname.c
@@ -112,6 +112,24 @@ path_sub(int argc, VALUE *argv, VALUE self)
 }
 
 /*
+ * Predicate method for root directories.  Returns +true+ if the
+ * pathname consists of consecutive slashes.
+ *
+ * It doesn't access the filesystem.  So it may return +false+ for some
+ * pathnames which points to roots such as <tt>/usr/..</tt>.
+ */
+static VALUE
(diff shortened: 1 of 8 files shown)


pathname.c
```diff
@@ -97,6 +120,212 @@ path_sub(int argc, VALUE *argv, VALUE self)
...
+{
+    VALUE path = get_strpath(self);
+    const char *ptr = RSTRING_PTR(path);
+    long len = RSTRING_LEN(path);
+    if (len < 1) return Qfalse;
+    if (drive_letter) {
+        if (len >= 2 && ISALPHA(ptr[0]) && (ptr[1] == ':')) return Qtrue;
+    }
+    return RBOOL(isdirsep(ptr[0]));
+}
```
eregon, line 178: How about using `File.absolute_path?` instead of reimplementing this?

nobu: While `File.absolute_path?` returns `false` on a relative path with the drive letter on Windows, e.g., "c:.", `Pathname#absolute?` returns `true`.
To be honest, I think the behavior of Pathname is questionable, but for now, I’ve made sure not to change how it works.

PR #16909, opened 2026-05-09
Deprecate `Pathname::SEPARATOR_PAT` instead of making it private

Earlopain:
It has been used in bundler for the last 10 years: https://github.com/ruby/rubygems/commit/0e8b29fac744a2cb11e7e42904234930266dd9cd

Let's deprecate first and give remove later to give people time to update.

For example, I have commited lockfiles. Currently there is no released bundler that doesn't use this constant, so my workflows fail when running against ruby-head (and still, I would not want to have to immediatly update bundler because of this).

It's similar to to the CI workaround with `BUNDLER_VERSION: 0` here

Followup to https://github.com/ruby/ruby/pull/16895

### DIFF
--- a/.github/workflows/ubuntu.yml
+++ b/.github/workflows/ubuntu.yml
@@ -251,8 +251,6 @@ jobs:
       - name: Run ruby-bench
         run: ruby run_benchmarks.rb -e "ruby::../build/install/bin/ruby" ${{ matrix.bench_opts }}
         working-directory: ruby-bench
-        env:
-          BUNDLER_VERSION: 0
 
       - uses: ./.github/actions/slack
         with:
--- a/.github/workflows/zjit-macos.yml
+++ b/.github/workflows/zjit-macos.yml
@@ -225,8 +225,6 @@ jobs:
       - name: Run ruby-bench
         run: ruby run_benchmarks.rb -e "zjit::../build/install/bin/ruby ${{ matrix.ruby_opts }}" ${{ matrix.bench_opts }}
         working-directory: ruby-bench
-        env:
-          BUNDLER_VERSION: 0
 
       - uses: ./.github/actions/slack
         with:
--- a/.github/workflows/zjit-ubuntu.yml
+++ b/.github/workflows/zjit-ubuntu.yml
@@ -279,8 +279,6 @@ jobs:
       - name: Run ruby-bench
         run: ruby run_benchmarks.rb -e "zjit::../build/install/bin/ruby ${{ matrix.ruby_opts }}" ${{ matrix.bench_opts }}
         working-directory: ruby-bench
-        env:
-          BUNDLER_VERSION: 0
 
       - uses: ./.github/actions/slack
         with:
--- a/depend
+++ b/depend
@@ -11143,6 +11143,7 @@ parser_st.$(OBJEXT): {$(VPATH)}st.c
 pathname.$(OBJEXT): $(hdrdir)/ruby.h
 pathname.$(OBJEXT): $(hdrdir)/ruby/ruby.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/compilers.h
+pathname.$(OBJEXT): $(top_srcdir)/internal/error.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/file.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/serial.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/static_assert.h
--- a/pathname.c
+++ b/pathname.c
@@ -1,5 +1,6 @@
 #include "ruby.h"
 #include "internal.h"
+#include "internal/error.h"     /* for rb_warn_deprecated_to_remove_at */
 #include "internal/file.h"
 #include "internal/string.h"
 #include "internal/vm.h"
@@ -358,6 +359,26 @@ InitVM_pathname(void)
     rb_define_private_method(rb_cPathname, "has_trailing_separator?", has_trailing_separator, 1);
     rb_define_private_method(rb_cPathname, "add_trailing_separator", add_trailing_separator, 1);
     rb_define_private_method(rb_cPathname, "del_trailing_separator", del_trailing_separator, 1);
+
+    {
+        /* NOTE: Used in bundler up to version 4.0.11. Remove this
+         * later. */
+        static const char fullname[] = "Pathname::SEPARATOR_PAT";
+        static const char source[] =
+#if alt_separator
+            "[/\\\\]"
+#else
+            "/"
+#endif
+            "(?# internal and deprecated constant kept only for old Bundler; do not use)";
+        const char *const name = fullname + rb_strlen_lit("Pathname::");
+
+        VALUE pat = rb_reg_new(source, sizeof(source)-1, 0);
+        OBJ_FREEZE(pat);
+        rb_define_const(rb_cPathname, name, pat);
+        rb_deprecate_constant(rb_cPathname, name);
+        rb_warn_deprecated_to_remove_at(4.2, fullname, NULL);
+    }
 
     rb_provide("pathname.so");
 }


nobu:
When do you think we can remove this constant?

Earlopain:
First release it with 4.1 as deprecated and then removing it early in 4.2 development sounds fine to me honestly. It looks only used by bundler and you fixed that already.

Although bundler right now has a rather high required ruby version (>= 3.2), so I wouldn't mind waiting a while longer since it's so little code.

What do you think?

nobu:
> First release it with 4.1 as deprecated and then removing it early in 4.2 development sounds fine to me honestly. It looks only used by bundler and you fixed that already.

After the release of the new Ruby instead of the new Bundler?
If the old Bundler will be used even after the new Bundler is released, may we have to keep the deprecated constant as far as there are such lock files?

Earlopain:
Ah, yes. I mean ruby versions with those. Just a normal deprecation cycle for it.

Bundler versions are recorded in the lockfile so you will _always_ have some that don't update. A longer deprecation would be better (the code for this is really short so keeping it doesn't do much harm), but I think it doesn't need to be so long.

pathname_builtin.rb
```diff
@@ -351,19 +351,23 @@ def sub_ext(repl)
...
+    SEPARATOR_PAT_INTERNAL = /[#{separator_list}]/
   else
     separator_list = Regexp.quote File::SEPARATOR
-    SEPARATOR_PAT = /#{separator_list}/
+    SEPARATOR_PAT_INTERNAL = /#{separator_list}/
   end
-  SEPARATOR_PAT.freeze
-  private_constant :SEPARATOR_PAT
+  SEPARATOR_PAT_INTERNAL.freeze
+  SEPARATOR_PAT = SEPARATOR_PAT_INTERNAL # :nodoc:
```
nobu:
```suggestion
  SEPARATOR_PAT = /#{SEPARATOR_PAT_INTERNAL}(?# internal and deprecated constant kept only for bad code in old Bundler; do not use)/.freeze # :nodoc:
```

nobu:
The lockfiles in ruby-bench have each different BUNDLED WITH versions, but run with the version specified in the first lockfile, because `BUNDLER_VERSION` env is set at the restart.
Does this restarting really make sense?

nobu (approved):

nobu:
```diff
diff --git a/depend b/depend
index 7faf59f5a44..c5bb9b38186 100644
--- a/depend
+++ b/depend
@@ -11143,6 +11143,7 @@ parser_st.$(OBJEXT): {$(VPATH)}st.c
 pathname.$(OBJEXT): $(hdrdir)/ruby.h
 pathname.$(OBJEXT): $(hdrdir)/ruby/ruby.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/compilers.h
+pathname.$(OBJEXT): $(top_srcdir)/internal/error.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/file.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/serial.h
 pathname.$(OBJEXT): $(top_srcdir)/internal/static_assert.h
diff --git a/pathname.c b/pathname.c
index a294a0d7301..fc74246d4db 100644
--- a/pathname.c
+++ b/pathname.c
@@ -1,5 +1,6 @@
 #include "ruby.h"
 #include "internal.h"
+#include "internal/error.h"     /* for rb_warn_deprecated_to_remove_at */
 #include "internal/file.h"
 #include "internal/string.h"
 #include "internal/vm.h"
@@ -359,6 +360,26 @@ InitVM_pathname(void)
     rb_define_private_method(rb_cPathname, "add_trailing_separator", add_trailing_separator, 1);
     rb_define_private_method(rb_cPathname, "del_trailing_separator", del_trailing_separator, 1);
 
+    {
+        /* NOTE: Used in bundler up to version 4.0.11. Remove this
+         * later. */
+        static const char fullname[] = "Pathname::SEPARATOR_PAT";
+        static const char source[] =
+#if alt_separator
+            "[/\\\\]"
+#else
+            "/"
+#endif
+            "(?# internal and deprecated constant kept only for old Bundler; do not use)";
+        const char *const name = fullname + rb_strlen_lit("Pathname::");
+
+        VALUE pat = rb_reg_new(source, sizeof(source)-1, 0);
+        OBJ_FREEZE(pat);
+        rb_define_const(rb_cPathname, name, pat);
+        rb_deprecate_constant(rb_cPathname, name);
+        rb_warn_deprecated_to_remove_at(4.2, fullname, NULL);
+    }
+
     rb_provide("pathname.so");
 }
 
```

hsbt:
I will backport https://github.com/ruby/rubygems/pull/9529 to 4.0.12 at next week. I'm okay to remove `Pathname::SEPARATOR_PAT`.

Earlopain:
Please consider leaving it around to give people a chance to upgrade naturally. It will break all previous bundler versions on ruby 4.1

Also, I'm not sure if I am supposed to move this into c @nobu. Am I?

Earlopain:
Ah, I guess I am. I didn't notice the conflict, sorry.

nobu:
> Please consider leaving it around to give people a chance to upgrade naturally. It will break all previous bundler versions on ruby 4.1

Ruby 4.1 bundles the latest bundler, so I don't think it will be a problem as long as the auto switching will be removed or disabled.

hsbt:
If someone wants to use Ruby 4.1 with Bundler 2.x, they are completely wrong. They should first test their system with Ruby 3.x and Bundler 4.0/4.1 before upgrading to Ruby 4.1.

Earlopain:
Nobody said anything about bundler 2. I just need to use the current latest released bundler (or any 4.x version) to hit this.

Bundler introduced auto-switching in 2.3 and the rationale is explained in https://blog.rubygems.org/2022/01/23/bundler-v2-3

hsbt:
I didn't understand what you want. I said "I will release 4.0.12 at next week.". Are you saying we should change the direction of 4.1  because you can't wait a week?

Earlopain:
If I run ruby 4.1, and have an existing lockfile, then it will break unless I also upgrade bundler. If you do not consider this important, then this PR is not necessary.

eregon:
IOW any CI testing ruby-head and having a checked-in `Gemfile.lock` with a `BUNDLED WITH` section is currently broken. We see that happened with `thrift` and `prism` linking this PR and working around with `BUNDLER_VERSION=0`, but probably for more CIs too which don't know about this PR.

It seems nicer to deprecate and let some time for these CIs/Gemfile.lock to update to latest Bundler, than breaking them immediately in a rather unclear way.
A new Bundler release is certainly helpful, but it might still not be obvious why it breaks and what's the fix with an existing Gemfile.lock.

(maybe Bundler shouldn't add `BUNDLED WITH` but it's what it does by default, even when there is no `BUNDLED WITH` section, I don't like auto-switching myself, it often causes more problems than it solves IMO)

eregon:
Though to be clear I think it's also OK to wait for a new release and see if it affects many CIs or just a few.

Ultimately this feels like something that should have been caught by ruby/ruby CI but hasn't (and neither by ruby/ruby-dev-builder CI).
Maybe we should add a test that `bundle install` with a Gemfile.lock `BUNDLED WITH` the latest Bundler release should work with ruby/ruby master?
EDIT: I understand we might not want that constraint ideally, but OTOH don't we want to encourage testing with ruby-head?

nobu:
@Earlopain
> If I run ruby 4.1, and have an existing lockfile, then it will break unless I also upgrade bundler.

Even the current master includes the latest bundler, it auto-switches to the lockfile version unless it is disabled by the environment variable or the configuration.
This happens even if that version is known to be vulnerable; I think this feature is harmful or out-of-date, and should be deprecated or should ignore the teeny at least.

Earlopain:
I think it's sensible to record and use the bundler version in the lockfile. It is helpful for static dependency analysis and allows bundler to be updated outside of the ruby release process like any other gem.

IMO what bundler doesn't get right is how difficult it is to actually update it. Just doing `bundle update` is not enough, you need to `bundle update --bundler`. I'd say that many people don't actually know this, or at the very least don't do this as often when compared against `bundle update`

eregon:
> This happens even if that version is known to be vulnerable; I think this feature is harmful or out-of-date, and should be deprecated or should ignore the teeny at least.

Agreed, I think Bundler should stop using the teeny version in `BUNDLED WITH`.

hsbt:
Bundler 4.0.12 is already released. We don't need to restore this constant.

PR #16910, opened 2026-05-09
[Bug #22057] test-bundler: add `--enable-gems` explicitly

tikkss:
This is because `make test-bundler-parallel` command, which runs successfully via turbo_tests, does not have `--disable-gems` option specified.

`--disable-gems` is currently set in `RUN_OPTS`. Removing it would affect not only test-bundler but the entire test suite, so I have configured it specifically for this command instead.

This patch will fix an error in `./spec/bundler/bundler/cli_common_spec.rb` when executing `make test-bundler`.


tikkss:
I modified the PR title, description and commit message to use `[Bug #] ...` instead of the `Fixes [Bug #](https://example.com)` style

PR #16914, opened 2026-05-10
Remove rb_gc_worker_thread_(un)set_vm_context

peterzhu2118:
These functions are no longer used.


YO4:
It also appears that `context->lock`, which is initialized in `rb_gc_initialize_vm_context`, is also not being used.
Is this assessment correct?

peterzhu2118:
Yes, that looks correct. Can you open a PR to remove it?

YO4:
Thank you. I'll give it a try.

YO4:
I made a PR #17697

PR #16917, opened 2026-05-11
[Bug #22063] Reject NaN Regexp timeout values

afurm:
Reject NaN Regexp timeout values

Bug: https://bugs.ruby-lang.org/issues/22063

## Summary

- reject `Float::NAN` in regexp timeout validation
- cover both `Regexp.timeout=` and `Regexp.new(..., timeout:)`

## Validation

- `./ruby --disable-gems -e 'begin; Regexp.new("foo", timeout: Float::NAN); rescue => e; puts [e.class, e.message].join(": "); end; begin; Regexp.timeout = Float::NAN; rescue => e; puts [e.class, e.message].join(": "); end'`
- `make test-all TESTS='test/ruby/test_regexp.rb --name=/timeout_corner_cases/'`
- `make test-all TESTS='test/ruby/test_regexp.rb'`
- `git diff --check`

### DIFF
--- a/re.c
+++ b/re.c
@@ -12,6 +12,7 @@
 #include "ruby/internal/config.h"
 
 #include <ctype.h>
+
 #include "encindex.h"
 #include "hrtime.h"
 #include "internal.h"
--- a/re.c
+++ b/re.c
@@ -12,7 +12,6 @@
 #include "ruby/internal/config.h"
 
 #include <ctype.h>
-
 #include "encindex.h"
 #include "hrtime.h"
 #include "internal.h"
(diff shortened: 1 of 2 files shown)


re.c
```diff
@@ -3971,7 +3972,7 @@ static void
 set_timeout(rb_hrtime_t *hrt, VALUE timeout)
 {
     double timeout_d = NIL_P(timeout) ? 0.0 : NUM2DBL(timeout);
-    if (!NIL_P(timeout) && timeout_d <= 0) {
+    if (!NIL_P(timeout) && (isnan(timeout_d) || timeout_d <= 0)) {
```
nobu: You can do it without `isnan()`.

```suggestion
    if (!NIL_P(timeout) && !(timeout_d > 0)) {
```

afurm: Thanks. Fixed

re.c
```diff
@@ -12,7 +12,6 @@
 #include "ruby/internal/config.h"
 
 #include <ctype.h>
-
 #include "encindex.h"
```
nobu, line 16:
```suggestion

#include "encindex.h"
```

PR #16919, opened 2026-05-11
Use monotonic counters for tracking malloc/free

eightbitraptor:
This PR replaces the single `objspace->malloc_counters.{increase,oldmalloc_increase}` `size_t` fields with pairs of monotonically-increasing counters. 

Snapshots of these counters are taken at each GC, so that the live malloc_increase is computed as

        (malloc - malloc_at_last_gc) - (free - free_at_last_gc)

We update the baselines at each GC. Minor GC's update `malloc` and `free` associated with young objects only (counters). Major GC's update based on "oldcounters" as well.

The malloc/free counters are 64 bits wide which should provide ample headroom (~9 exabytes!). We use `size_t` on 64-bit and `uint64_t` on 32-bit, wrapped by a `gc_counter_t` struct.

However, because updating a `uint64_t` is a multi-instruction operation on 32 bit, we have to introduce a lock to the `malloc_counters` struct to avoid racing.

We introduced 2 new macros `MALLOC_COUNTERS_LOCK` and `MALLOC_COUNTERS_UNLOCK` that use `rb_nativethread_lock_t`.

The lock is initialized in `rb_gc_impl_objspace_init` and destroyed in `rb_gc_impl_objspace_free`. We chose this because it mirrors existing finalizer_lock pattern in wbcheck.

On 64 bit platforms aligned 64 bit loads are atomic, and writes are already using `RUBY_ATOMIC_SIZE_ADD` so the locks are not needed and the macros do nothing.

## Benchmarks

8 years ago, Eric Wong implemented a change like this in [Bug 14767](https://bugs.ruby-lang.org/issues/14767) but reverted it shortly after due to performance and RSS concerns in some benchmarks. Given that this patch touches the allocation fast path it's important to make sure that we haven't regressed behaviour or RSS.

This patch does not replicate the benchmark regressions seen in the original patch. Instead we show a consistent (but small) improvement over the base commit in all the benchmarks tested during the original ticket.

```
ubuntu@ip-172-31-42-127:~/ruby$ taskset -c 1 make benchmark ARGS=benchmark/array_sample* COMPARE_RUBY=$HOME/.rubies/ruby-master-fast/bin/ruby
/usr/bin/ruby --disable=gems -rrubygems -I./benchmark/lib ./benchmark/benchmark-driver/exe/benchmark-driver \
            --executables="compare-ruby::/home/ubuntu/.rubies/ruby-master-fast/bin/ruby -I.ext/common --disable-gem" \
            --output=markdown --output-compare -v benchmark/array_sample*
compare-ruby: ruby 4.1.0dev (2026-05-14T09:26:58Z master 69857ea9d8) +PRISM [x86_64-linux]
built-ruby: ruby 4.1.0dev (2026-05-14T22:54:52Z mvh-monotonic-coun.. d8e8aaca26) +PRISM [x86_64-linux]
warming up..
# Iteration per second (i/s)

|ary.sample               |     21.202M|   21.420M|
|                         |           -|     1.01x|
|ary.sample(2)            |      5.542M|    5.601M|
|                         |           -|     1.01x|
|array_sample_100k_10     |     101.821|   103.691|
|array_sample_100k_11     |      76.301|    77.411|
|                         |           -|     1.01x|
|array_sample_100k__100   |      13.628|    13.956|
|array_sample_100k__1k    |       1.416|     1.442|
|array_sample_100k__6k    |       0.420|     0.434|
|                         |           -|     1.03x|
|array_sample_100k___10k  |       0.289|     0.296|
|array_sample_100k___50k  |       0.075|     0.076|
|                         |           -|     1.00x|
```

We also ran the Headline benchmarks from the ruby-bench suite:

```
master: ruby 4.1.0dev (2026-05-14T09:26:58Z master 69857ea9d8) +YJIT +PRISM [x86_64-linux]
experiment: ruby 4.1.0dev (2026-05-14T22:54:52Z mvh-monotonic-coun.. d8e8aaca26) +YJIT +PRISM [x86_64-linux]

--------------  -------------  ---------  ---------------  ---------  ------------------  -----------------  ---------------------
bench             master (ms)  RSS (MiB)  experiment (ms)  RSS (MiB)  experiment 1st itr  master/experiment  RSS master/experiment
activerecord     148.6 ± 7.1%       84.7     147.1 ± 7.7%       86.2               1.003              1.010                  0.983
chunky-png       614.9 ± 0.1%       72.9     611.6 ± 0.1%       81.0               1.004              1.005                  0.900
erubi-rails      940.5 ± 2.0%      124.8     916.1 ± 3.4%      124.6               0.988              1.027                  1.001
hexapdf         1803.5 ± 3.6%      213.4    1768.4 ± 3.2%      191.1               0.977              1.020                  1.116
liquid-c          57.1 ± 7.1%       66.5      57.5 ± 6.7%       67.5               1.014              0.993                  0.985
liquid-compile    45.9 ± 9.1%       48.7      46.2 ± 8.0%       48.6               1.002              0.994                  1.002
liquid-render     76.7 ± 5.2%       60.2      78.3 ± 4.9%       58.0               1.002              0.980                  1.037
lobsters         837.8 ± 1.4%      351.5     843.1 ± 3.2%      334.4               0.936              0.994                  1.051
mail             114.3 ± 4.6%       65.5     114.9 ± 4.6%       67.2               1.004              0.994                  0.976
psych-load      1636.0 ± 1.4%       32.5    1642.3 ± 1.3%       32.3               0.997              0.996                  1.005
railsbench      1615.0 ± 0.1%      134.2    1626.6 ± 1.1%      139.0               0.998              0.993                  0.966
rubocop         148.0 ± 15.1%      140.8    146.4 ± 15.5%      139.8               1.000              1.011                  1.007
ruby-lsp         142.1 ± 3.1%       71.7     143.2 ± 2.6%       72.8               1.018              0.993                  0.985
sequel            63.5 ± 0.9%       45.9      63.5 ± 5.1%       52.5               0.978              0.999                  0.874
shipit          1234.3 ± 1.7%      196.5    1201.6 ± 3.3%      192.5               1.037              1.027                  1.021
--------------  -------------  ---------  ---------------  ---------  ------------------  -----------------  ---------------------

Legend:
- experiment 1st itr: ratio of master/experiment time for the first benchmarking iteration.
- master/experiment: ratio of master/experiment time. Higher is better for experiment. Above 1 represents a speedup.
- RSS master/experiment: ratio of master/experiment RSS. Higher is better for experiment. Above 1 means lower memory usage.
```

These results show small statistically insignificant changes in both directions. None are consistently outside the margin of error.

For example: 
 
- Apparent wins: shipit 1.037, erubi-rails 1.027, hexapdf 1.020. However variance is ±1.7-3.6%, so all comfortably inside noise.
 - Apparent losses: liquid-render 0.980, raisbench 0.993, but a 1.1% variance
 - rubocop ±15.1%/15.5% variance, so again, the 0.7% regression is noise. 
 
 Median ratio is ~1.0. No benchmark shows a regression that's both directional and outside its own stderr.

### DIFF
--- a/gc/default/default.c
+++ b/gc/default/default.c
@@ -495,11 +495,29 @@ enum gc_mode {
     gc_mode_compacting,
 };
 
+#if SIZEOF_SIZE_T >= 8
+typedef size_t gc_counter_t;
+#else
+typedef uint64_t gc_counter_t;
+#endif
+
+struct gc_malloc_bytes {
+    gc_counter_t malloc;
+    gc_counter_t free;
+
+    /* Snapshots of `malloc` / `free` taken at the end of the last GC */
+    gc_counter_t malloc_at_last_gc;
+    gc_counter_t free_at_last_gc;
+};
+
 typedef struct rb_objspace {
     struct {
-        size_t increase;
+        struct gc_malloc_bytes counters;
 #if RGENGC_ESTIMATE_OLDMALLOC
-        size_t oldmalloc_increase;
+        struct gc_malloc_bytes oldcounters;
+#endif
+#if SIZEOF_SIZE_T < 8
+        rb_nativethread_lock_t lock;
 #endif
     } malloc_counters;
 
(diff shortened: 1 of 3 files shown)


nobu:
> The malloc/free counters are 64 bits wide which should provide ample headroom (~9 exabytes!). We use `size_t` on 64-bit and `uint64_t` on 32-bit, wrapped by a `gc_counter_t` struct.

Is it necessary to use `uint6_t` on 32-bit platforms?

> However, because updating a `uint64_t` is a multi-instruction operation on 32 bit, we have to introduce a lock to the `malloc_counters` struct to avoid racing.

Some platforms, even 32-bit, may be able to implement 64-bit atomic operations, like as `rbimpl_atomic_u64_load_relaxed`.

eightbitraptor:
> Is it necessary to use `uint6_t` on 32-bit platforms?

If we use a `uint32_t` then we'd overflow the counters after 2GB of allocated bytes.
 
> Some platforms, even 32-bit, may be able to implement 64-bit atomic operations, like as `rbimpl_atomic_u64_load_relaxed`.

I'll look more into this, thank you.

nobu:
> If we use a `uint32_t` then we'd overflow the counters after 2GB of allocated bytes.

It is 4GiB, because of unsigned.

eightbitraptor:
> It is 4GiB, because of unsigned.

Yes, absolutely - my apologies, I was investigating a signed approach briefly and got mixed up. I still think it's far too easy to overflow 4GiB of malloc given that `total_malloc_bytes` and `total_free_bytes` are monotonically accumulating for the lifetime of the process.

> Some platforms, even 32-bit, may be able to implement 64-bit atomic operations, like as rbimpl_atomic_u64_load_relaxed.

Done - I've updated this PR that restructures this so that on 32-bit platforms with `HAVE_GCC_ATOMIC_BUILTINS_64` we use atomic operations in the same way that `rbimpl_atomic_u64_load_relaxed` does, otherwise we fallback to a mutex (although I don't think that we officially support any targets that don't `HAVE_GCC_ATOMIC_BUILTINS_64`).

I'll update with benchmarks.

eightbitraptor:
Updated the description with benchmark results.

gc/default/default.c
```diff
@@ -516,11 +517,38 @@ enum gc_mode {
     gc_mode_compacting,
 };
 
+#if SIZEOF_SIZE_T >= 8
+typedef size_t gc_counter_t;
+#else
+typedef RBIMPL_ALIGNAS(8) uint64_t gc_counter_t;
```
nobu: Isn't this same as `rbimpl_atomic_uint64_t`?

```suggestion
typedef rbimpl_atomic_uint64_t gc_counter_t;
```

If this works, casts to `rbimpl_atomic_uint64_t` won't be needed.

eightbitraptor: Thank you, you're right :bow: 

I've refactored to remove the casts and switch to `rbimpl_atomic_uint64_t`

PR #16920, opened 2026-05-11
ZJIT: Infer block param types with parallel phi semantics

tekknolagi:
In `Function::infer_types`, snapshot the types of all branch arguments before writing to any target block param. Previously, when a self-loop's branch args referenced the target's own params (e.g. a phi swap), updating params sequentially could observe values updated earlier in the same loop iteration rather than the pre-update values that phi parallel assignment semantics require.

Also modify `changed` if we widen a parameter's type.

Unfortunately I don't have any concrete bug examples here.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -3021,6 +3021,17 @@ impl Function {
         // Fill entry parameter types
         self.copy_param_types();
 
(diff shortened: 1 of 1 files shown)


k0kubun (approved):

XrXr (approved):

tekknolagi:
What is with these Windows builds??

PR #16921, opened 2026-05-11
Replace subclasses linked list with weakref array

jhawthorn:
This converts the previous subclasses linked list to a new imemo type which is a weakref array. 

The primary motivation for this is to allow classes to be freed concurrently, which this accomplishes by removing the linked-list removal.

It's probably also possible to achieve this by locking or other techniques on the existing linked list design, but this is simpler and should also use a little less memory.

The concern with this weakref approach is that we have to examine these objects after marking and before sweeping. Because this object is only created for classes with subclasses (with `require "rails/all"` there are ~500 of these), and it can benefit from generational GC, I hope this is reasonable.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -1294,13 +1294,28 @@ rb_gc_handle_weak_references(VALUE obj)
         break;
 
       case T_IMEMO: {
-        GC_ASSERT(imemo_type(obj) == imemo_callcache);
-
-        struct rb_callcache *cc = (struct rb_callcache *)obj;
-        if (cc->klass != Qundef &&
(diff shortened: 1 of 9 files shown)


gc.c
```diff
@@ -1328,6 +1343,7 @@ rb_gc_imemo_needs_cleanup_p(VALUE obj)
       case imemo_ment:
       case imemo_iseq:
       case imemo_callinfo:
+      case imemo_subclasses:
```
byroot: You could return `FL_TEST_RAW(obj, IMEMO_SUBCLASSES_HEAP)` instead, skipping sweeping in most cases.

PR #16923, opened 2026-05-11
[Bug #22064] Fix GC compaction for compare-by-identity sets

peterzhu2118:
Compare-by-identity sets use the address for hashing, so we must pin it so the object does not move in GC compaction. Objects in a compare-by-identity set is not currently pinned, causing the set to be broken if the object is moved.

For example:

```ruby
set = Set.new.compare_by_identity

o = Object.new
set.add(o)

puts set.include?(o)

GC.verify_compaction_references(expand_heap: true, toward: :empty)

puts set.include?(o)
```

It should output true twice, but it outputs true and false.

### DIFF
--- a/set.c
+++ b/set.c
@@ -124,6 +124,14 @@ struct set_object {
 };
 
 static int
+mark_and_pin_key(st_data_t key, st_data_t data)
+{
+    rb_gc_mark((VALUE)key);
+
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Thank you for fixing this!

PR #16927, opened 2026-05-12
ZJIT: Remove GuardType deduplication

tekknolagi:
Now that ece14b61f505eea1ebefb3b8295df0fcf4d22567 has landed, we get
this for free.

We will eventually add value numbering, which will do more, but for now
remove this ad-hoc code.

structaref benchmark is 5ms/iteration before and after this commit, and 9ms with `canonicalize` disabled.


k0kubun:
ref: https://github.com/ruby/ruby/pull/16654

k0kubun (approved):

PR #16940, opened 2026-05-13
Use rbs-integrated RDoc

st0012:
This will allow docs.ruby-lang.org to include method rbs sigs.

### DIFF
--- a/gems/bundled_gems
+++ b/gems/bundled_gems
@@ -37,7 +37,7 @@ ostruct             0.6.3   https://github.com/ruby/ostruct
 pstore              0.2.1   https://github.com/ruby/pstore
 benchmark           0.5.0   https://github.com/ruby/benchmark
 logger              1.7.0   https://github.com/ruby/logger
-rdoc                7.2.0   https://github.com/ruby/rdoc bc0baee787609f2205e8f1103f3e1d36b923184a
+rdoc                7.2.0   https://github.com/ruby/rdoc 1f93543615928b6d45357432f16ec6006e2d8b1e
 win32ole            1.9.3   https://github.com/ruby/win32ole
 irb                 1.18.0  https://github.com/ruby/irb
 reline              0.6.3   https://github.com/ruby/reline


st0012:
@nobu https://github.com/ruby/ruby/pull/16760 makes `make html` not picking up the version specified in `gems/bundled_gems` anymore. I found this as I'm trying to make ruby/ruby adopt the latest RDoc commit.
I think we need to revert the `tool/rdoc-srcdir` changes so we can make sure it picks up the right RDoc.

nobu:
> https://github.com/ruby/ruby/pull/16760 makes `make html` not picking up the version specified in `gems/bundled_gems` anymore.

It is intentional, since the built extensions may not be loadable by baseruby.
If you don't use baseruby, e.g., `make HAVE_BASERUBY=no`, the gems under `.bundle` will be used.

> I think we need to revert the `tool/rdoc-srcdir` changes so we can make sure it picks up the right RDoc.

RBS contains the extension library, so it cannot work when cross-compiling.

nobu:
I'm guessing you want to use `$(XRUBY)` when running `make rdoc`, right?

```patch
index b1debdebc97..81e0505b2b1 100644
--- a/common.mk
+++ b/common.mk
@@ -69,8 +69,8 @@ LIBRUBY_EXTS  = ./.libruby-with-ext.time
 REVISION_H    = ./.revision.time
 PLATFORM_D    = $(TIMESTAMPDIR)/.$(PLATFORM_DIR).time
 ENC_TRANS_D   = $(TIMESTAMPDIR)/.enc-trans.time
-X_$(no_baseruby:no=NO_)BASERUBY = $(BASERUBY)
-X_$(yes_baseruby:yes=NO_)BASERUBY = $(XRUBY)
+X_$(CROSS_COMPILING:yes=)BASERUBY = $(BASERUBY)
+X_$(CROSS_COMPILING:no=)BASERUBY = $(XRUBY)
 RDOC          = $(X_BASERUBY) --enable-gems "$(tooldir)/rdoc-srcdir"
 RDOCOUT       = $(EXTOUT)/rdoc
 HTMLOUT       = $(EXTOUT)/html
```

Earlopain:
There were a few PRs to rdoc here lately, so I don't know which is responsible (probably not this one), but https://docs.ruby-lang.org/en/master/ is now being generated by rdoc 6.5.0 from ruby 3.2. Just wanted to let you know

st0012:
Oh that's bad. @Earlopain Thanks for raising it

PR #16944, opened 2026-05-13
ZJIT: Stop using def_ids! for HIR dump purposes

k0kubun:
This PR introduces `FieldName` enum that is used for `LoadField` identifiers in HIR dump.

Currently, we intern `ID`s (`def_ids!` uses `rb_intern2`) for HIR dump purposes. Since these IDs are not used by CRuby, we shouldn't insert these IDs into CRuby's symbol table, which would consume extra memory.

Also this PR prefers the use of canonical names for the index used by CRuby, e.g. `VM_ENV_DATA_INDEX_FLAGS`. Previously, we used inconsistent names for these things, so we ended up having multiple names for the same thing: `_ep_specval`/`_env_data_index_specval` and `_ep_flags`/`_env_data_index_flags`, which leads to confusion.

### DIFF
--- a/zjit/src/codegen.rs
+++ b/zjit/src/codegen.rs
@@ -21,7 +21,7 @@ use crate::stats::{counter_ptr, with_time_stat, trace_compile_phase, Counter, Co
 use crate::{asm::CodeBlock, cruby::*, options::debug, virtualmem::CodePtr};
 use crate::backend::lir::{self, Assembler, C_ARG_OPNDS, C_RET_OPND, CFP, EC, NATIVE_STACK_PTR, Opnd, SP, SideExit, SideExitRecompile, Target, asm_ccall, asm_comment};
 use crate::hir::{iseq_to_hir, BlockId, Invariant, RangeType, SideExitReason::{self, *}, SpecialBackrefSymbol, SpecialObjectType};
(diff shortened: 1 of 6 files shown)


tekknolagi (approved):
Seems reasonable to me. I will caution that this probably makes `Insn` even bigger as it increases field name sizes from 64 bit to 128 bit.

tekknolagi:
We may want to consider maintaining our own intern ID pool and index by u8 or u16 or something instead (& copy CRuby IDs into it).

XrXr (approved):

k0kubun:
> I will caution that this probably makes Insn even bigger as it increases field name sizes from 64 bit to 128 bit.
>
> We may want to consider maintaining our own intern ID pool and index by u8 or u16 or something instead (& copy CRuby IDs into it).

It's a good point to pay attention to. `size_of::<hir::Insn>()`, however, stays 120 before and after this PR.

Currently, `CCallWithFrame` and `CCallVariadic` use 113 bytes, which becomes 120 with alignment. `LoadField` is 60 bytes with the new `FieldName` of 16 bytes. So it's a bit far from being a problem at the moment, but yeah, let's consider doing it when it becomes a problem.

PR #16948, opened 2026-05-13
ZJIT: Add back return type for .class

### LINKED ISSUE #850: ZJIT: Something is wrong with types::Class in the HIR type system
Currently breaking rubocop: https://github.com/ruby/ruby/pull/14945#issuecomment-3465974031

tekknolagi:
We fixed the underlying bug in https://github.com/ruby/ruby/pull/15268
(f9c51ac5ea42382a22fb8a2cc5d7aeb78e77c962).


XrXr (approved):

XrXr:
Just curious: That bug had nothing to do with `Class` being effectively an empty type? https://github.com/ruby/ruby/pull/16706

tekknolagi:
I don't think so, no

tekknolagi:
I could be wrong though

k0kubun (approved):

PR #16949, opened 2026-05-13
ZJIT: Replace Test(BoxBool(x)) with x

tekknolagi:
We don't need to first turn a CBool into a Ruby bool then check that bit
pattern; we can go straight to working with the CBool.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -5318,6 +5318,14 @@ impl Function {
                     Insn::Test { val } if self.type_of(val).is_known_truthy() => {
                         self.new_insn(Insn::Const { val: Const::CBool(true) })
                     }
+                    Insn::Test { val: test_val } => {
+                        if let Insn::BoxBool { val: bool_val } = self.find(test_val) {
+                            self.make_equal_to(insn_id, bool_val);
+                            continue;
+                        } else {
+                            insn_id
+                        }
+                    }
                     Insn::CondBranch { val, if_true, .. } if self.is_a(val, Type::from_cbool(true)) => {
                         self.new_insn(Insn::Jump(if_true))
                     }
--- a/zjit/src/hir/opt_tests.rs
+++ b/zjit/src/hir/opt_tests.rs
@@ -16604,4 +16604,49 @@ mod hir_opt_tests {
           Return v27
         ");
     }
+
+    #[test]
+    fn test_elide_test_of_box_bool() {
+        eval(r#"
+            def test(a, b)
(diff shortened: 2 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -5262,6 +5262,14 @@ impl Function {
                     Insn::Test { val } if self.type_of(val).is_known_truthy() => {
                         self.new_insn(Insn::Const { val: Const::CBool(true) })
                     }
+                    Insn::Test { val: test_val } => {
+                        if let Insn::BoxBool { val: bool_val } = self.find(test_val) {
+                            self.make_equal_to(insn_id, bool_val);
+                            continue;
```
XrXr, line 5324: So a few lines above, it says "// Don't bother re-inferring the type of index; we already know it." because it uses `type_of`. We don't check the type here, so maybe we should return `bool_val` instead of `continue`?

zjit/src/hir/opt_tests.rs
```diff
@@ -16461,4 +16461,48 @@ mod hir_opt_tests {
...
+          v7:BasicObject = LoadArg :self@0
+          v8:BasicObject = LoadArg :a@1
+          v9:BasicObject = LoadArg :b@2
+          Jump bb3(v7, v8, v9)
+        bb3(v11:BasicObject, v12:BasicObject, v13:BasicObject):
+          PatchPoint MethodRedefined(Symbol@0x1008, ==@0x1010, cme:0x1018)
+          v48:StaticSymbol = GuardType v12, StaticSymbol
+          v49:CBool = IsBitEqual v48, v13
+          v50:BoolExact = BoxBool v49
+          CheckInterrupts
```
XrXr, line 16640: OT: I assume v50 is solely for the framestate in `CheckInterrupts`. We have some wiggle room WRT timing and can probably check earlier to avoid the extra work.

XrXr (approved):

zjit/src/hir.rs
tekknolagi: Oh the other difference is that we would re-push `bool_val` to the instructions, which we don't want to do. That would lead to having the instruction twice in the block/code unit.

I have a PR to make a better API here but I haven't spent enough time making sure it's the right API. Maybe I should just move it forward because it makes a lot of this clearer. In my proposed API this would be `replace_with` as opposed to `replace_with_new` or `drop` or `skip` (the other potential actions, IIRC).

k0kubun (approved):

PR #16950, opened 2026-05-13
ZJIT: Run canonicalize before fold_constants

tekknolagi:
This helps get rid of repeated type guards in a loop such as structaref
or structaset benchmark. @dak2 originally had this in his PR but it
wasn't tested so I removed it. This brings it back with a test that
catches it.

### DIFF
--- a/zjit/src/hir/opt_tests.rs
+++ b/zjit/src/hir/opt_tests.rs
@@ -9761,7 +9761,7 @@ mod hir_opt_tests {
           v33:ArrayExact = GuardType v10, ArrayExact
           v34:CUInt64 = LoadField v33, :_rbasic_flags@0x1040
           v35:CUInt64 = GuardNoBitsSet v34, RUBY_FL_FREEZE=CUInt64(2048)
-          v37:CUInt64 = GuardNoBitsSet v34, RUBY_ELTS_SHARED=CUInt64(4096)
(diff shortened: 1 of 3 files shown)


tekknolagi:
Unfortunately this does not make structaref, for example, any faster than it was before. Still ~5ms/iter. Fixnum guards are cheap.

zjit/src/hir/opt_tests.rs
```diff
@@ -16461,4 +16461,62 @@ mod hir_opt_tests {
...
+          v17:Fixnum[0] = Const Value(0)
+          PatchPoint NoSingletonClass(C@0x1008)
+          PatchPoint MethodRedefined(C@0x1008, var@0x1010, cme:0x1018)
+          v138:ObjectSubclass[class_exact:C] = GuardType v12, ObjectSubclass[class_exact:C]
+          v139:BasicObject = LoadField v138, :var@0x1040
+          PatchPoint MethodRedefined(Integer@0x1048, +@0x1050, cme:0x1058)
+          v178:Fixnum = GuardType v139, Fixnum
+          v179:Fixnum = FixnumAdd v17, v178
+          PatchPoint NoEPEscape(test)
+          v183:Fixnum = FixnumAdd v179, v178
```
tekknolagi, line 16509: This previously had a GuardType in between each

k0kubun (approved):

PR #5445, opened 2022-01-14
Add Thread.each_caller_location

jeremyevans:
This method takes a block and yields Thread::Backtrace::Location
objects to the block.  It does not take arguments, and always
starts at the default frame that caller_locations would start at.

Implements [Feature #16663]

PR #16951, opened 2026-05-13
[Bug #22070] Fix segfault in `Thread.each_caller_location`

amomchilov:
Fixes https://bugs.ruby-lang.org/issues/22070


amomchilov:
cc @jeremyevans, who might have context from #5445

jeremyevans (approved):
Thanks for fixing this!

jeremyevans:
@mame requesting your review, because it appears that 10767283dd0277a1d780790ce6bde67cf2c832a2 introduced this regression. The change looks reasonable to me, but you probably understand the code it modifies better than I do.

PR #16952, opened 2026-05-13
ZJIT: Specialize monomorphic getivar on recompile

k0kubun:
ref: https://github.com/Shopify/ruby/issues/984

This PR compiles monomorphic getivar into an optimized load followed by a fallback, instead of using a fallback in all cases, like polymorphic getivar already does.

### DIFF
--- a/zjit/src/hir.rs
+++ b/zjit/src/hir.rs
@@ -3331,14 +3331,17 @@ impl Function {
         self.resolve_receiver_type_from_profile(recv, insn_idx)
     }
 
-    fn polymorphic_summary(&self, profiles: &ProfileOracle, recv: InsnId, insn_idx: YarvInsnIdx) -> Option<TypeDistributionSummary> {
+    fn polymorphic_summary(&self, profiles: &ProfileOracle, recv: InsnId, insn_idx: YarvInsnIdx, include_monomorphic: bool) -> Option<TypeDistributionSummary> {
         let Some(entries) = profiles.types.get(&insn_idx) else {
             return None;
         };
         let recv = self.chase_insn(recv);
         for (entry_insn, entry_type_summary) in entries {
             if self.union_find.borrow().find_const(*entry_insn) == recv {
-                if entry_type_summary.is_polymorphic() || entry_type_summary.is_skewed_polymorphic() {
+                let usable = entry_type_summary.is_polymorphic()
+                    || entry_type_summary.is_skewed_polymorphic()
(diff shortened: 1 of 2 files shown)


zjit/src/hir.rs
```diff
@@ -3331,14 +3331,17 @@ impl Function {
         self.resolve_receiver_type_from_profile(recv, insn_idx)
     }
 
-    fn polymorphic_summary(&self, profiles: &ProfileOracle, recv: InsnId, insn_idx: YarvInsnIdx) -> Option<TypeDistributionSummary> {
+    fn polymorphic_summary(&self, profiles: &ProfileOracle, recv: InsnId, insn_idx: YarvInsnIdx, include_monomorphic: bool) -> Option<TypeDistributionSummary> {
```
tekknolagi: Let's avoid a profusion of boolean parameters. This feels like it's plumbing through a bunch of logic to do something that should be simpler

zjit/src/hir.rs
```diff
@@ -4405,9 +4408,8 @@ impl Function {
                         }
                         if self.policy.no_side_exits {
```
tekknolagi, line 4430: I don't understand why the `no_side_exits` is different here than in HIR generation

tekknolagi (commented):
This feels a little off, like something deeper is wrong. I want to revisit policy and policy API

zjit/src/hir.rs
k0kubun: Could you elaborate? Which parts do you think are different?

zjit/src/hir.rs
k0kubun: dropped the parameter at https://github.com/ruby/ruby/pull/16952/changes/d463c3ba85ddeed2b2c6308c4a5cd999ced42bb9. I suspect we'd need to change how ivars are compiled to unify things more than the current state.

k0kubun:
> This feels a little off, like something deeper is wrong. I want to revisit policy and policy API

The impression I get is that the policy interface itself is very basic and there's nothing wrong about it, but the way monomorphic calls and polymorphic calls are compiled is so different between them that the policy can't be uniformly applied to these different logics.

If monomorphic calls are compiled in the same way as polymorphic calls as if the size of polymorphism is 1, we didn't need to change how the policy is applied to these. We indeed want to compile the last chain in the same way across different versions, but somehow they are handled very differently right now.

PR #16960, opened 2026-05-14
Run RDoc with baseruby when cross-compling

nobu:
To run with the revision specified in gems/bundled_gems file.


st0012:
Thank you!

PR #16964, opened 2026-05-14
Fix UAF in IO::Buffer#& when self or mask is an invalidated slice

himura467:
`io_buffer_and` accessed `buffer->base` and `mask_buffer->base` directly without validating that the buffers were still live. A slice whose parent had been freed retained its stale base pointer, so calling `&` on it caused a UAF.

Use `io_buffer_get_bytes_for_reading` for both operands, which raises `IO::Buffer::InvalidatedError` before any memory access if either buffer has been invalidated.

```ruby
inner = IO::Buffer.new(IO::Buffer::PAGE_SIZE)
slice = inner.slice(0, 8)
inner.free
slice & IO::Buffer.for("ABCDEFGH")  # => was UAF, now raises InvalidatedError
```

### DIFF
--- a/io_buffer.c
+++ b/io_buffer.c
@@ -3378,14 +3378,14 @@ io_buffer_pwrite(int argc, VALUE *argv, VALUE self)
 }
 
 static inline void
-io_buffer_check_mask(const struct rb_io_buffer *buffer)
+io_buffer_check_mask_size(size_t size)
 {
-    if (buffer->size == 0)
+    if (size == 0)
         rb_raise(rb_eIOBufferMaskError, "Zero-length mask given!");
 }
 
 static void
-memory_and(unsigned char * restrict output, unsigned char * restrict base, size_t size, unsigned char * restrict mask, size_t mask_size)
+memory_and(unsigned char * restrict output, const unsigned char * restrict base, size_t size, const unsigned char * restrict mask, size_t mask_size)
 {
     for (size_t offset = 0; offset < size; offset += 1) {
         output[offset] = base[offset] & mask[offset % mask_size];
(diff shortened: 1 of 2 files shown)


io_buffer.c
```diff
@@ -3413,13 +3413,21 @@ io_buffer_and(VALUE self, VALUE mask)
...
 
+    const void *base;
+    size_t size;
+    io_buffer_get_bytes_for_reading(buffer, &base, &size);
+
+    const void *mask_base;
+    size_t mask_size;
+    io_buffer_get_bytes_for_reading(mask_buffer, &mask_base, &mask_size);
+
     io_buffer_check_mask(mask_buffer);
```
kou: Can we use `mask_base` and `mask_size` instead of `mask_buffer` here?

himura467: Good point. `|`, `^`, `and!`, `or!`, and `xor!` have the same issue, and `io_buffer_check_mask` is called in all of them. I was planning to fix those first in follow-up PRs and then refactor `io_buffer_check_mask` (changing its argument from `const struct rb_io_buffer *` to `size_t`) in a single PR. Should I fold that into this PR instead?

kou: Let's do it in this PR but it may be better that we rename `io_buffer_check_mask()` to `io_buffer_check_mask_size()`.

himura467: Refactored.

io_buffer.c
```diff
@@ -3378,14 +3378,14 @@ io_buffer_pwrite(int argc, VALUE *argv, VALUE self)
 }
 
 static inline void
-io_buffer_check_mask(const struct rb_io_buffer *buffer)
+io_buffer_check_mask_size(size_t mask_size)
 {
-    if (buffer->size == 0)
+    if (mask_size == 0)
```
kou: The `mask_` prefix is redundant in this context because this function name is `io_buffer_check_mask_size()`. `size` must be mask size.

himura467: Thanks, updated.

ioquatix (approved):
Thanks!

kou (approved):
+1

PR #16966, opened 2026-05-14
ZJIT: Add method inliner

### LINKED ISSUE #967: ZJIT: Method Inlining
# Background

Method call inlining is a common and profitable optimization for compilers. In Ruby, nearly all operations are method calls giving us an opportunity to affect performance widely. At the most basic level, avoiding the setup and clean-up of method calls saves execution time and memory. Perhaps more importantly, however, is that inlining a method gives the optimizer more context to make better optimization decisions. Interprocedural optimization is expensive and so its application is quite limited. Inlining a method body into another method turns that into intraprocedural optimization and that is something ZJIT does extensively today.

## Constraining Polymorphism

Inline caches have been an immensely powerful tool for speeding up applications in dynamic programming languages. Research has shown that at any given call site, the set of receiver types that a method is called on is typically very low (often, only 1). An inline cache avoids the expensive method resolution phase by caching the method handle locally and guarding it by a quick type check (e.g., reference comparison on cached shape and current receiver shape).

However, a compiled method must function correctly at all call sites. Across many call sites, we tend to see more diversity in both the receiver type and the argument types. So, while a given call site may be monomorphic, the compiled call target may be polymorphic or even megamorphic.

Previous work, focused on Ruby, has demonstrated that call target splitting can reduce the level of polymorphism. Here, we duplicate a method and allow the multiple copies to be compiled, only needing to support a smaller subset of call sites. Taken to the extreme, each call site would have its own copy of the call target. That, in turn, very frequently makes the call target monomorphic, which tends to have a cascading effect. The compiler can make bigger bets, producing more optimized code.

Method inlining provides a natural opportunity to support splitting. I propose that ZJIT inlines unoptimized methods. I.e., instead of taking the body a method that has already been processed by the optimization loop, we create a new copy of the uninitialized method by going back to the ISEQ list. Once the body of that method has been inlined into the caller and the CFG cleaned up to meet our (extended) basic block invariants, the combined list of instructions can be processed all at once.

There is a trade-off here: each inlined method will require memory space and, being compiled independently, will spend more time in the compiler. The benefit of this approach, however, is we can specialize the inlined method precisely for its use by the caller. This will typically allow the compiled code to be smaller than it would if it needed to support other call sites. There's a snowball effect as we inline more of the call graph.

# Inlining Policy

Inlining can be a finicky optimization. In many cases it improves performance, while in others it can have no impact at all, or worse, can slow down execution. The more we inline, the slower we make the compilation process. If we inline too much, we start to have a noticeable impact on the process's memory usage. Larger methods can cause thrashing in the CPU's instruction cache. Unbounded recursion could prevent the application from ever running at all.

To limit this effect, we will need an inlining policy. The policy will make the determination of whether a method should be inlined or not. We can make that decision based on data known to us at compile time, such as:

* Size of the target method in instructions
* Profile info that could tell us how frequently a method is called or a branch taken
* Method name and receiver type
* Total amount inlined so far

The inlining policy will need to be tuned over time. I suggest as a first step that we adopt the heuristics of other optimizing Ruby runtimes. Presumably, they've done some amount of work in devising their heuristics. While different VMs and compilers have different concerns, at a high level, inlining is more about the runtime behavior of Ruby code rather than the VM's execution of that code.

# Considerations

Inlining in Ruby can be a bit tricky. We have methods defined statically and others defined dynamically at runtime. We allow classes to be reopened and methods to be redefined at run time. We have blocks and procs, which are special types of methods, and in addition to returning, they can `break`. We have `set_trace_func` and `TracePoint`. We have the ability to fetch and manipulate backtraces. We have `binding`. Anything that needs access to the frame is going to need special handling since inlining logically removes that data. We have a rich set of metaprogramming methods. We have `throw` and `raise`, which affect control flow. We also have inspection tools, such as debuggers and profilers.

We also have methods implemented in C and Ruby &mdash; we will only inline Ruby methods into Ruby methods. That means to make the most of inlining, we'll want to have Ruby-based alternative implementations of common core library methods (e.g., we've previously done this with `Array#each`). As a community, and at Shopify in particular, we'll want to shift away from native extensions where a gem will suffice. For native extensions that exist only to bind to native libraries, we'll want to look hard at FFI and Fiddle.

I suggest we ignore most, if not all, of the special cases in this first pass to the extent that we can detect them, leaving them as regular method calls. Put another way, our initial inlining policy should be quite conservative in what it admits. We will inline simple methods and blocks with straightforward control flow and that don't need to materialize frames. We can always tweak from there. Boiling the ocean, as it were, would result in difficult to debug code propagating all throughout. Handling simple cases first will give us a firm foundation to build upon.

# Compilation Strategy

If all we do is take a method's body and insert into it the caller, we haven't gained a whole lot. Yes, we can reduce method call overhead and that has value. But, inlining gives us extra context to make better decisions in speculative optimization. Moreover, we should loop our optimization pass so we can revisit earlier phases after later phases have simplified or transformed the HIR for a method. Classically, we'd loop until we reach a fixed point, but we could establish an upper-bound (either hard-coded or config option) on loop iteration count. Notably, we must defer inlining until after type specialization in order to correctly identify the call target. However, we'd be inlining an unspecialized method body, so we'll ultimately need to run type specialization a second time to properly process the inlined method.

When the runtime state changes such that our optimized code is invalid (e.g., an inlined method is redefined), we need to unwind our inlined method, restore state, and side-exit to the interpreter. This is a process we call deoptimization. It implies that we have a way to detect an invalidation and a mechanism to recover our state. Beyond a performance hit, there should be no observable change to the run time behavior.

Once we've thrown away compilation, we can either alter our existing profile or start from scratch (it's unclear which would be most profitable at this time). With the new data, we may find it makes sense to attempt compiling again. In pathological cases, the runtime behavior never stabilizes or cycles through so many states that we'd label it megamorphic. To avoid decompilation loops, we'll want to know how many times we've inlined a method and had to throw it away.

# Related Projects

In order for inlining to really make sense, we need to have [lightweight frames](https://github.com/Shopify/ruby/issues/909). We can start the inlining project without lightweight frames, but having to inline the heavy frame logic will constrain performance.

Having a Ruby-based core library will be necessary to get the full advantage of ZJIT and method inlining. The Rubinius project started down this path many years ago. TruffleRuby has carried on from there. Chris Seaton [made the case](https://youtu.be/-iVh_8_J-Uo?si=ErMXheVT1o1lvMlU) for a Ruby-based core library at RubyConf 2022. It may make sense to coordinate with the JRuby and TruffleRuby teams. One challenge there, however, is each VM is likely to need a different set of primitives in different places, even if the overall implementation is Ruby.

Carrying on from migrating native code to Ruby, we'll want to have a faster FFI. Currently, binding to a native library and invoking a foreign function is simply slower than using a native extension. Often that performance hit isn't significant, but when it is, FFI becomes a poor choice. Even in cases where it wouldn't be a hot spot, it's hard to expect a slower solution to gain traction.

nirvdrum:
# Overview

This PR contains a new method inliner for ZJIT. While this is a fully functional method inliner, this work is intended to be a starting point rather than a production-ready inliner. In order to make the code easy to understand, this is a very straightforward inliner. The goal here is to get the infrastructure into place so other JIT work can be inlining-aware. Moreover, other optimization work is currently blocked by the lack of an inliner.

# Review Notes

The history for this PR is pretty clean. I took an incremental approach to development and that's reflected by logically cohesive commits. It's a fairly large PR. Reviewing commit-by-commit will probably make the process easier. Where I thought warranted, the commit message has additional context to understand why a particular change is made.

In places where I initially excluded something only to include later, I squashed those commits together. In other places, I've retained both commits so we can evaluate whether we want the before or after state. Notably, this work began when ZJIT used extended basic blocks and needed to be adapted after #16888 was merged. Rather than pretend this work began after the move to traditional basic blocks, I've retained both approaches because I think it was instructive to see.

I also tried to be comprehensive, but judicious, with comments in both tests and implementation. If any comment is unclear or seems unnecessary, please flag it. The target audience for the comments is someone reasonably familiar with ZJIT's architecture and the process of inlining, but that may not have a complete mental map of all the state involved or how it interacts with other parts of the compiler. Where design decisions met a fork in the road, I indicated which path we took and why.

# Optimize Loop

Simply eliminating method dispatch by inlining method bodies is interesting, but undersells what inlining enables. The real benefit is the extra context we have for making optimization decisions. Accordingly, I wrapped ZJIT's `optimize` pipeline in a loop that will repeatedly run type specialization, inlining, and subsequent optimizations until inlining reaches a fixed-point. In order to avoid runaway compilation I added a configurable upper-bound.

While working with _ruby-bench_ I found we rarely looped more than 2 or 3 times, at least with the currently defined defaults for inline threshold and budget. As those defaults were chosen arbitrarily, that's not a very strong claim. But, I think the intuition is correct: large callers exhaust their budget quickly and large callees fail the threshold, so a fixed-point is reached after only a few passes.

Be that as it may, both inlining and the optimization loop will incur additional compilation time. I don't think it's worth spending a lot of time tuning the inlining parameters until we land block inlining. As such, the inliner is disabled by default. Those that want to use it can control compilation time indirectly by setting appropriate inliner options for their application.

# Limitations / Future Work

* The biggest limitation at the moment is that there is no block inlining. I plan to do that in a follow-up PR. But, I reached a natural checkpoint once method inlining was completed.
* The default values for the various configurable inlining options were all chosen arbitrarily. I plan to look at representative workloads and tune for better defaults.
* I have profiled the inliner to deal with the most egregious compilation issues, but the inlining pass is not optimized. It is written to be straightforward and reasonably performant, but doesn't do anything terribly sophisticated to share/reuse compilations across multiple call sites. Given the inliner is disabled in this PR, I'd prefer to save potentially complex optimizations for a follow-up PR where they can be more easily evaluated.
* The inlining policy is incredibly basic: once a method is eligible for compilation, all of its `SendDirect` instructions will be evaluated for inlining. If the method meets the configured threshold and there's inlining budget available, it will be inlined. That process will continue in a loop until either no more inlining is possible or an upper bound is hit. We do not selectively inline a subset of `SendDirect` based on any profiling data. I expect this policy will be refined or replaced once exposed to more varied workloads.
* Our current approach to recompiling polymorphic ivar accesses does not play well with inlining. An inlined monomorphic ivar read that later goes polymorphic causes a shape-failure side exit, which invalidates the ISEQ. After enough invalidations the ISEQ's next recompile lands in `no_side_exits` mode, which forces every ivar access in that ISEQ onto the dynamic path. Before inlining only the affected ISEQ was caught; with inlining the cascade propagates invalidation to every caller that inlined the callee, so the entire caller (including its own unrelated ivars) is recompiled in `no_side_exits` mode. We'll need to revisit this (re)compilation policy for inlining.

# New Configuration Options

| Option | Description | Unit |
|--------|--------|--------|
| `--zjit-inline-threshold` | Max size for method to be eligible for inlining | YARV bytecode encoded size |
| `--zjit-inline-budget` | Per-caller cumulative size for inlining | Caller's HIR instruction length |
| `--zjit-inline-max-iterations` | The upper-bound on optimize loop if it doesn't reach a fixed-point | Count | 
| `--zjit-inline-deny` | CSV list of qualified method names that will not inline any calls | String |

# Sample Run

To give a rough idea of what changes we see with the inliner, I ran the _lobsters_ benchmark on this branch with `--zjit-inline-threshold=0` (i.e., inlining disabled) and `--zjit-inline-threshold=30`. The harness defaults (`WARMUP_ITRS=15 MIN_BENCH_ITRS=10 MIN_BENCH_TIME=10`) were used. I ran this on an M4 Pro MacBook Pro, so numbers should be taken with a grain of salt.

<details><summary>Benchmark Results</summary>
<p>

| Metric | Baseline | Inline-30 | Δ |
|---|---:|---:|---:|
| `send_count` | 183,177,341 | 153,580,860 | −16.2% |
| `optimized_send_count` | 159,489,694 (87.1%) | 130,708,793 (85.1%) | −18.0% |
| `dynamic_send_count` | 23,687,647 (12.9%) | 22,872,067 (14.9%) | −3.4% |
| `inline_method_count` | 0 | 6,708 | new |
| `iseq_optimized_send_count` | 54,067,730 | 23,613,869 | −56.3% |
| `inline_cfunc_optimized_send_count` | 74,812,509 | 75,187,772 | +0.5% |
| `inline_iseq_optimized_send_count` | 6,298,842 | 6,562,848 | +4.2% |
| `non_variadic_cfunc_optimized_send_count` | 13,813,825 | 14,743,103 | +6.7% |
| `variadic_cfunc_optimized_send_count` | 10,496,788 | 10,601,201 | +1.0% |
| `dynamic_getivar_count` | 11,448,240 | 18,044,644 | +57.6% |
| `dynamic_setivar_count` | 4,273,637 | 7,859,945 | +83.9% |
| `dynamic_definedivar_count` | 1,278,221 | 2,368,919 | +85.3% |
| `compiled_iseq_count` | 5,711 | 5,111 | −10.5% |
| `compiled_side_exit_count` | 73,639 | 105,471 | +43.2% |
| `failed_iseq_count` | 3 | 3 | 0 |
| `compile_time` | 2,806 ms | 5,369 ms | +91.3% |
| `compile_hir_time` | 1,560 ms | 3,580 ms | +129.5% |
| `compile_lir_time` | 1,140 ms | 1,682 ms | +47.6% |
| `code_region_bytes` | 31,080,448 | 45,285,376 | +45.7% |
| `zjit_alloc_bytes` | 22,784,791 | 24,815,675 | +8.9% |
| `total_mem_bytes` | 53,865,239 | 70,101,051 | +30.1% |
| `side_exit_count` | 8,644,965 | 7,851,332 | −9.2% |
| `side_exit_size` | 12,392,832 | 17,623,092 | +42.2% |
| `guard_type_count` | 143,754,515 | 121,335,155 | −15.6% |
| `guard_shape_count` | 36,007,056 | 32,029,543 | −11.0% |
| `vm_insn_count` | 114,646,306 | 108,789,358 | −5.1% |
| `zjit_insn_count` | 879,149,176 | 883,749,075 | +0.5% |
| `ratio_in_zjit` | 88.46% | 89.04% | +0.6 pp |
| avg non-warmup iter | 602 ms | 619 ms | +2.8% |

</p>
</details> 

The results show a 56% reduction in `iseq_optimized_send_count` and 18% reduction in overall `optimized_send_count`. Compile time increased, which is expected. Just to reiterate, optimization of the compilation pass is deferred to a follow-up PR and the tuning of inlining parameters (not explored here) will also impact compilation time. The rise in dynamic ivar counts warrants more exploration, but I think it's a symptom of ivar recompilation having a larger blast radius in the presence of inlining. This is noted as something we generally need to revisit in the "Limitations / Future Work" section. Polymorphic ivar reads occur on some hot paths in Rails.


zjit/src/codegen.rs
```diff
@@ -1501,6 +1539,129 @@ fn gen_send_without_block(
...
+    asm_comment!(asm, "initialize cfp->pc for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_PC), Opnd::const_ptr(callee_pc));
+    asm_comment!(asm, "initialize cfp->iseq for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_ISEQ), VALUE::from(iseq).into());
+    asm_comment!(asm, "clear cfp->jit_return for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_JIT_RETURN), 0.into());
+    asm_comment!(asm, "initialize cfp->sp for inlined frame");
+    let ep_offset = state.stack().len() as i32 + local_size as i32 - args.len() as i32 + VM_ENV_DATA_SIZE as i32 - 1;
+    let new_sp = asm.lea(Opnd::mem(64, SP, (ep_offset + 1) * SIZEOF_VALUE_I32));
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_SP), new_sp);
```
tenderlove: We have to materialize these fields up front because our stack materialization code can't handle code without an entry point?  Am I reading this correctly that inlined frame pushes are punished in comparison to non-inlined callees?

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+            return false;
+        }
+
+        // User-supplied denylist of qualified method names (e.g. `User#name` or
+        // `Foo.bar`). Lets us isolate the inliner's contribution from individual
+        // problem methods without committing to a heuristic. Anonymous code paths
+        // (blocks, procs without a stable method binding) can't be expressed in this
+        // format, so they aren't matched and fall through to the rest of the checks.
+        // Unsafe deref of `cme` is safe here because `inline_methods` only calls
+        // `should_inline` for `SendDirect` instructions, which carry a non-null cme.
```
tenderlove, line 4458: Is this just for debugging purposes?  Do we expect to allow users to do this?

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+        let mut pc_idx: u32 = 0;
+        while pc_idx < callee_encoded_size {
+            let pc = unsafe { rb_iseq_pc_at_idx(callee_iseq, pc_idx) };
+            let opcode: u32 = unsafe { rb_iseq_opcode_at_pc(callee_iseq, pc) }
+                .try_into()
+                .unwrap();
+            if opcode == YARVINSN_invokeblock || opcode == YARVINSN_trace_invokeblock {
+                incr_counter!(inline_reject_invokeblock);
+                return false;
+            }
```
tenderlove, line 4415: It seems like we profile invoke block, does this handle the profiled instruction names (like `zjit_invokeblock`)?

zjit/src/options.rs
```diff
@@ -114,6 +114,54 @@ pub struct Options {
...
+    pub inline_budget: usize,
+
+    /// Set of qualified method names (e.g. `Class#method`, `Module::Class.method`) that
+    /// `should_inline` will refuse to inline. Populated from `--zjit-inline-deny=…` as a
+    /// comma-separated list. Used as a debugging/experimentation knob: callees on this
+    /// list are rejected before any of the other `should_inline` checks run, so we can
+    /// isolate the contribution of specific methods to the inliner's overall effect.
+    /// Empty by default; matching uses the same string format produced by
+    /// `qualified_method_name`. Only named methods are matched today; anonymous code
+    /// (blocks, procs without a stable method binding) cannot be denied this way.
```
tenderlove, line 174: Ah ok, so this is for debugging.

zjit/src/hir.rs
```diff
@@ -6281,6 +6773,12 @@ pub struct FrameState {
 
     stack: Vec<InsnId>,
     locals: Vec<InsnId>,
+
+    // For inlined frames, the caller's state at the call site. This allows
+    // side exits from inlined code to reconstruct the full frame chain.
+    // Rc is used because all Snapshots within an inlined callee share the
+    // same caller state, avoiding redundant deep clones during remapping.
+    caller_state: Option<Rc<FrameState>>,
```
tekknolagi: Why not refer to parent by `InsnId`? It seems natural to walk pointers that way

zjit/src/options.rs
nirvdrum: The inlining policy isn't terribly refined at the moment. This option is an escape hatch should the existing heuristics and/or inlining options result in inlining a method that you don't want to inline. But, yes, I added it to satisfy a need I had to set pathological cases aside while working with the _lobsters_ benchmark. I didn't invest in a more comprehensive pattern because that need hadn't arisen yet.

zjit/src/codegen.rs
```diff
@@ -1501,6 +1539,129 @@ fn gen_send_without_block(
...
+    fn cfp_opnd(offset: i32) -> Opnd {
+        Opnd::mem(64, CFP, offset - (RUBY_SIZEOF_CONTROL_FRAME as i32))
+    }
+    let callee_pc = unsafe { rb_iseq_pc_at_idx(iseq, 0) };
+    asm_comment!(asm, "initialize cfp->pc for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_PC), Opnd::const_ptr(callee_pc));
+    asm_comment!(asm, "initialize cfp->iseq for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_ISEQ), VALUE::from(iseq).into());
+    asm_comment!(asm, "clear cfp->jit_return for inlined frame");
+    asm.mov(cfp_opnd(RUBY_OFFSET_CFP_JIT_RETURN), 0.into());
```
k0kubun: > Inlined HIR will install a fresh JITFrame via gen_save_pc_for_gc before
> any subsequent GC-triggering operation.

If this comment is true, can we remove this write into `RUBY_OFFSET_CFP_JIT_RETURN` here? Isn't it supposed to be overwritten before any call that would read it?

```suggestion
```

k0kubun: Same applies to PC and ISEQ. We shouldn't need to write them eagerly. If it fails, it likely means we're missing `gen_save_pc_for_gc` in such cases.

zjit/src/hir.rs
```diff
@@ -1093,6 +1093,25 @@ pub enum Insn {
         state: InsnId,
     },
 
+    /// Push an interpreter frame for an inlined callee. Initially does a full frame push;
+    /// will be replaced with a lightweight frame push when that project lands.
+    PushLightweightFrame {
```
tekknolagi: I think we need to ratchet down effects of these instructions otherwise we can't do e.g. load elimination around them

zjit/src/codegen.rs
nirvdrum: I intended for the comment from L1599 - L1608 to answer your first question. It sounds like it doesn't so I'll try to make it clearer. But, yes, it's a short-term penalty. Perhaps my approach was flawed, but I expect this entire method to be overhauled/rewritten as part of the lightweight frames work so I made this little more than a copy-and-pasted version of `gen_push_frame` with necessary modifications to allow inlining to work.

Now that I'm looking at it, I think we can do better by allocating a `JITFrame` at compile time and installing that. As long as that doesn't conflict with the lightweight frames work, I'm happy to do that. It may be possible to defer `cfp->sp` as well, but I'd need to take a a deeper look at that.

zjit/src/hir.rs
```diff
@@ -3013,6 +3064,9 @@ impl Function {
             // as a reference for FrameState, which we use to generate side-exit code.
             Insn::Snapshot { .. } => types::Any,
             Insn::IsA { .. } => types::BoolExact,
+            // Frame instructions have no output; the type is never used.
```
tekknolagi: put in `has_output` returning false and then these go at the start of `infer_type`

zjit/src/hir.rs
nirvdrum: I suppose that's to be decided. I like it as an escape hatch if you've got a pathological case you don't want to inline but can't exclude via tuning params because that would exclude things you want to inline. It's similar in spirit to `--zjit-allowed-iseqs`, albeit with different value format (file path vs CSV list). We can certainly unify that -- the CSV line on CLI was just very convenient when running benchmarks.

With that said, if it's really not useful to others, we can drop it.

zjit/src/hir.rs
nirvdrum: Nope, good catch. I'll fix that. Likewise for block params below.

zjit/src/codegen.rs
tenderlove: Makes sense.  I think I understood it correctly from your comments, but I wanted to double check.

zjit/src/codegen.rs
nirvdrum: Great. I didn't want to conflict with anything you were doing. I've pushed two new commits (5b1a93af7b57fd12011ed071c13e3666396bfa58 and 149ebb14f4fc21ae8f94b5caaf3a7eaf39a52fe7) to clean this up. I resolved the overall problem a little bit differently than you suggested, so I'd appreciate if you could take another look.

zjit/src/codegen.rs
nirvdrum: @k0kubun chimed in so I went ahead and made the change (5b1a93af7b57fd12011ed071c13e3666396bfa58 and 149ebb14f4fc21ae8f94b5caaf3a7eaf39a52fe7). Please take another look and let me know if there's anything you'd like to see changed.

zjit/src/options.rs
```diff
@@ -114,6 +114,54 @@ pub struct Options {
...
+    /// `should_inline` runs in a given fixed-point iteration, `self.insns.len()` has
+    /// already been bumped by `iseq_to_hir`'s initial build, then by `type_specialize`
+    /// and the trivial `inline` pass, and (in iterations 2+) by every prior pass in
+    /// the loop including the previous round's `inline_methods`. It is a useful proxy
+    /// for "compile work done" but not for "size of the compiled output".
+    ///
+    /// Note: this is a different unit than `inline_threshold` — that field is callee
+    /// YARV bytecode words; this one is caller HIR instructions, allocation high-water
+    /// mark. They aren't directly comparable; YARV → HIR typically expands roughly 1-3x.
+    pub inline_budget: usize,
```
tekknolagi: Why `insns` vs keeping a tally of YARV opcodes or something? Mostly I ask because this and hte following limit are in different units (and I would expect HIR insn inline budget to behave less stably over time). Neither is amazing but other compilers seem to use something more like AST nodes inlined, or interpreter opcodes inlined, or things like that.

zjit/src/options.rs
```diff
@@ -142,6 +190,10 @@ impl Default for Options {
             allowed_iseqs: None,
             log_compiled_iseqs: None,
             max_versions: 2,
+            inline_threshold: 0,
```
tekknolagi: Can you make these top-level constants so that we can e.g. use them in option doc output automatically

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
         crate::stats::trace_compile_phase("infer_types", || self.infer_types());
     }
 
+    /// Check whether a callee ISEQ is eligible for inlining into this function.
+    /// The checks implemented here are conservative: simple positional parameters only,
+    /// no escaping EP, no invokeblock/yield, no getblockparam. Callee body shape
+    /// restrictions (single return, etc.) live in inline_methods since they need
+    /// the built HIR to evaluate.
+    fn should_inline(&self, callee_iseq: IseqPtr, cme: *const rb_callable_method_entry_t) -> bool {
```
tekknolagi, line 4434: Can we split this into `can_inline` and `should_inline` with the former only holding things that are hard limits and the latter being only heuristics for optimization driven

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+                    continue;
+                }
+
+                // Snapshot the caller's HIR state so we can roll back if compiling
+                // the callee fails or its body has no return paths. The caller is
+                // mutated in place by add_iseq_to_hir, so on rejection we truncate
+                // the side-tables back to these lengths. The original block isn't
+                // touched until we commit to the inline below, so rejection paths
+                // only need to advance the search cursor without restoring it.
+                let pre_insns_len = self.insns.len();
```
tekknolagi, line 4551: I don't love this bit. Union-find especially is really tricky to roll back if we're not careful. Do we even use union-find in HIR construction?

Why do we need the rollback?

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+                // can serve as the return_block argument; the callee's leaves become
+                // Jumps to this block directly during translation. The continuation
+                // is included in the pre-state snapshot above so a rollback also
+                // discards it.
+                let continuation = self.new_block(self.blocks[block.0].insn_idx);
+
+                // The callee's ISEQ may still have profiling instrumentation active
+                // (zjit_opt_* bytecodes). Disable it so add_iseq_to_hir sees the
+                // original opcodes.
+                unsafe { crate::cruby::rb_zjit_profile_disable(iseq) };
```
tekknolagi: I feel like the caller should not be dictating profile policy (accidentally) for callees. This might mean that our `iseq_to_hir` should be looking over all opcodes via a `bare_opcode` map so we don't have to do this.

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+                        search_start = send_pos + 1;
+                        continue;
+                    }
+                };
+
+                // A callee with no return paths never reaches the continuation: it
+                // diverges or always side-exits. Inlining it would leave the caller's
+                // SendDirect result aliased to a Param in an unreachable block, so
+                // reject and roll back.
+                if add_result.num_returns == 0 {
```
tekknolagi: I think that's fine if that happens? What is the problem? That would just mean the thing leading to the send is now terminated by something else and this should be the only predecessor to that (new) block.

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+                // post-default body. SendDirect emission already guarantees
+                // args.len() lies in `lead_num + post_num + kw_num..=lead_num +
+                // opt_num + post_num + kw_num`, with `kw_num` being the callee's
+                // full keyword count (zero when the callee has no keywords). After
+                // `prepare_direct_send_args` runs, the caller's arg list has a slot
+                // for every callee keyword (filled in callee table order, padded
+                // with defaults for omitted optional keywords), so the keyword tail
+                // is a fixed-size addition we subtract off before recovering the
+                // optional-positional count.
+                let callee_params = unsafe { iseq.params() };
```
tekknolagi, line 4614: Is there some shared way we can do this? I feel like we have this copied three ways across the codebase now.

cc @XrXr

zjit/src/hir.rs
```diff
@@ -4198,6 +4270,400 @@ impl Function {
...
+                //   * any remaining non-parameter locals are nil-initialized.
+                let callee_body_params: Vec<InsnId> = self.blocks[callee_entry_body_block.0].params.clone();
+
+                // First param is self.
+                if !callee_body_params.is_empty() {
+                    self.make_equal_to(callee_body_params[0], recv);
+                }
+
+                // Next params are locals.
+                let num_locals = callee_body_params.len() - 1; // -1 for self
```
tekknolagi, line 4661: Someone who understands the calling convention better than me is going to have to take a look at this closely

zjit/src/hir.rs
```diff
@@ -5674,19 +6140,40 @@ impl Function {
...
-        run_pass!(remove_redundant_patch_points);
-        run_pass!(remove_duplicate_check_interrupts);
-        run_pass!(eliminate_dead_code);
+        // The optimization pipeline runs in a fixed-point loop so that inlining and
+        // type specialization can feed each other: the first iteration inlines direct
+        // calls and specializes the inlined code, and subsequent iterations can inline
+        // calls that only became monomorphic after the previous round of specialization.
+        // Termination is guaranteed because each iteration either inlines at least one
+        // call (growing the function toward the inlining budget) or reaches a fixed point.
+        for _ in 0..get_option!(inline_max_iterations) {
```
tekknolagi, line 6282: How did you arrive at this optimizer/inliner shape?

zjit/src/hir.rs
```diff
@@ -5674,19 +6140,40 @@ impl Function {
...
+            // Function is assumed to have types inferred already
+            run_pass!(type_specialize);
+            // The trivial inliner runs first to handle simple cases (constant returns,
+            // parameter returns, etc.) without frame push/pop overhead. The general
+            // inliner then handles more complex methods that require full inlining.
+            run_pass!(inline);
+            let did_inline = self.inline_methods();
+
+            #[cfg(debug_assertions)] self.assert_validates();
```
tekknolagi: Can you make `inline_methods` a "normal" pass (with the counter, etc) so that it becomes just another `run_pass!(...)`

zjit/src/hir.rs
```diff
@@ -6622,28 +7190,85 @@ fn invalidates_locals(opcode: u32, operands: *const VALUE) -> bool {
...
 pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
     if !ZJITState::can_compile_iseq(iseq) {
         return Err(ParseError::NotAllowed);
     }
     let payload = get_or_create_iseq_payload(iseq);
-    let mut profiles = ProfileOracle::new(payload);
     let mut fun = Function::new(iseq);
     fun.was_invalidated_for_singleton_class_creation = payload.was_invalidated_for_singleton_class_creation;
 
+    let result = add_iseq_to_hir(&mut fun, iseq, true, None)?;
```
tekknolagi: these new boolean and option parameters do not spark joy. Given that we have precisely two configurations (`true, None` and `false, Some(...)`) can we make this an enum or something reasonable

zjit/src/hir.rs
```diff
@@ -7825,7 +8458,7 @@ pub fn iseq_to_hir(iseq: *const rb_iseq_t) -> Result<Function, ParseError> {
                         let stack_count = state.stack.len();
                         let recv = state.stack_topn(argc as usize)?;  // args are on top
                         let entry_args = state.as_args(self_param);
-                        if let Some(summary) = fun.polymorphic_summary(&profiles, recv, exit_state.insn_idx) {
+                        if let Some(summary) = fun.polymorphic_summary(&profiles, recv, exit_state.iseq, exit_state.insn_idx) {
```
tekknolagi: how is your use of `exit_state.iseq` here different from `iseq` elsewhere?

zjit/src/payload.rs
```diff
@@ -13,6 +13,14 @@ pub struct IseqPayload {
...
     /// JIT code versions. Different versions should have different assumptions.
     pub versions: Vec<IseqVersionRef>,
+    /// Caller versions whose compiled code contains an inlined copy of this ISEQ.
+    /// Populated when a caller version is committed in `gen_iseq` and the inliner
+    /// pulled this ISEQ's HIR into the caller. Read by `exit_recompile` to invalidate
+    /// callers when this ISEQ has no standalone version to invalidate but its profile
+    /// changed (e.g. shape guard failure inside an inlined body). Entries to
+    /// already-invalidated versions are skipped at read time and pruned on caller
+    /// invalidation; see `invalidate_iseq_version`.
+    pub inlined_into: Vec<IseqVersionRef>,
```
tekknolagi: I need the `inlined_into` and `inlined_callees` and wiring up stuff explained to me at some point

zjit/src/codegen.rs
k0kubun: Looks good :+1: I'll make another point in another thead.

zjit/src/codegen.rs
```diff
@@ -1501,6 +1539,130 @@ fn gen_send_without_block(
...
+        recv,
+        iseq: Some(iseq),
+        cme,
+        frame_type,
+        specval,
+        write_block_code: iseq_may_write_block_code(iseq),
+    });
+
+    // Materialize the inlined callee's CFP so async observers (GC scans, signal-driven
+    // backtrace walkers) see correct PC/ISEQ before the inlined body runs. Unlike
```
k0kubun: > GC scans

My understanding is that GC scans happen only under the VM lock. If so, GC scans happen only after every thread goes past `gen_save_pc_for_gc`. So I believe "GC scans" do not require this `cfp->jit_return` write here. If you agree with it, can we drop "GC scans" from this comment?

> signal-driven backtrace walkers

We currently don't write `cfp->jit_return` in `gen_push_frame` (for non-inlined calls in master), but I think this is a valid point to make this `cfp->jit_return` write (whether the call is inlined or not), for now.

YJIT doesn't write `cfp->pc` on `gen_push_frame`, but it doesn't crash on signal-driven async `rb_profile_frames` calls because it checks if `cfp->pc` is valid, based on `cfp->iseq`. (It's possible that the `cfp->pc` slot has PC from past calls for the same ISEQ and it returns a wrong profiling result though.)

For ZJIT, we don't have a way to tell if `cfp->jit_return` is valid. So it's not safe to leave `cfp->jit_return` uninitialized. For that reason, writing `cfp->jit_return` on frame push seems right.

zjit/src/codegen.rs
k0kubun: Posted it at https://github.com/ruby/ruby/pull/16966#pullrequestreview-4294623443. I ended up suggesting no change there though.

k0kubun:
Looked at the codegen changes so far. The frame handling looks good :+1:

zjit/src/options.rs
nirvdrum: I went with this for the following reasons:

* Checking instruction length is a constant time operation
* It's state we already have on hand
* Using instruction length accounts for the growth of a method through inlining
* It's similar to Graal's usage of IR node count for inlining budget

But, I'm not married to it. I don't like how it penalizes us for adding HIR instructions that a future pass may very well strip out. Rather than get hung up on it, I just did something that seemed sensible to me and moved on. Changing it should be trivial.

zjit/src/hir.rs
nirvdrum: I'm sorry, I can't tell if you're asking specifically about this loop form or something deeper. The "Optimize Loop" section in the PR description hits upon the rationale -- mostly that inlining alone wasn't super interesting. Repeated application of inlining and optimization is informed by Truffle's partial evaluation process. From an implementation point of view, it doesn't mess with the pipeline and is easily disabled.

I anticipated differences in opinion on how to do it or whether to do it at all, which is why I called it out in the PR overview. If you have a preferred design, please let me know. What's there is functional, simple, and demonstrative, but not really load-bearing. If it's a problem, the whole thing can be thrown away without any impact on the rest of the inliner.

zjit/src/hir.rs
nirvdrum: I started off by vastly restricting what was a candidate for inlining and incrementally expanding it. I started with single returns only, then multi-returns. This just isn't a case I had handled yet. I can drop the defensive guard.

zjit/src/hir.rs
nirvdrum: Is that something you want to address pre-merge or in a follow-up PR?

tekknolagi: Ideally pre. Otherwise this has weird second-order effects I think. We have a function `vm_zjit_insn_to_bare_insn`

zjit/src/hir.rs
tekknolagi: I am asking about the specific shape of the loop. Other compilers have different ways they do this. There's a bunch of academic research. I'm asking if this was an intentional BFS-like design? If it's modeled after something else? The easiest first implementation? etc

There's a bunch of stuff in the literature like the [Prokopec paper](https://www.researchgate.net/profile/Aleksandar-Prokopec-2/publication/331408280_An_Optimization-Driven_Incremental_Inline_Substitution_Algorithm_for_Just-in-Time_Compilers/links/5c77f4cd92851c695047e765/An-Optimization-Driven-Incremental-Inline-Substitution-Algorithm-for-Just-in-Time-Compilers.pdf) too.

zjit/src/hir.rs
nirvdrum: My mental model at the time was an ancestry of `FrameState`. I see now that following `Snapshot` is nicer so thanks for pointing me in that direction. I didn't realize that [Rc::make_mut](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.make_mut) was clone-on-write either, so my attempt at having a cheap, shared copy wasn't working as intended.

I wish the type system were a little tighter here. We use some variant of `state: InsnId`, which works for any `Insn`, but only `Insn::Snapshot` should be permitted. I realize we handle that by matching on the instruction type in `Function::frame_state`, but I'd love for the type system to handle that. I think it would make the intent of the code easier to follow.

zjit/src/hir.rs
nirvdrum: It isn't. I was maintaining parity with `exit_state.insn_idx`, but I can change it to `iseq` easily enough.

zjit/src/hir.rs
```diff
@@ -4196,6 +4268,409 @@ impl Function {
...
+                let pre_insn_types_len = self.insn_types.len();
+                let pre_blocks_len = self.blocks.len();
+                let pre_uf_len = self.union_find.borrow().forwarded.len();
+
+                // Create the continuation block before translating the callee so it
+                // can serve as the return_block argument; the callee's leaves become
+                // Jumps to this block directly during translation. The continuation
+                // is included in the pre-state snapshot above so a rollback also
+                // discards it.
+                let continuation = self.new_block(self.blocks[block.0].insn_idx);
```
tekknolagi: This should probably be call frame state insn idx

tekknolagi: errr the next insn idx after that I guess

test/ruby/test_zjit.rb
```diff
@@ -382,6 +382,258 @@ def test
...
+      def callee(arr)
+        arr.each do |x|
+          break 7 if x > 5
+        end
+      end
+      def test(a) = callee(a)
+
+      test([1, 6, 99])
+    RUBY
+  end
```
k0kubun: Could you move all these tests to codegen_tests instead? JIT code tests that do not rely on external libraries should be put in codegen_tests.rs going forward. ref: https://github.com/ruby/ruby/pull/16548

zjit/src/hir.rs
```diff
@@ -1091,6 +1091,25 @@ pub enum Insn {
...
+        state: InsnId,
+    },
+    /// Pop the interpreter frame for an inlined callee, restoring the caller's frame.
+    /// Carries the callee's iseq and argument count so codegen can compute the SP offset
+    /// to restore.
+    PopLightweightFrame {
+        iseq: IseqPtr,
+        argc: usize,
+        state: InsnId,
+    },
```
k0kubun: Can we call them `PushJITFrame`/`PopJITFrame`? Despite the project name, the actual type name in the Rust code is `JITFrame`, so it seems confusing to use names inconsistent with it.

```suggestion
    PushJITFrame {
        iseq: IseqPtr,
        cme: *const rb_callable_method_entry_t,
        recv: InsnId,
        args: Vec<InsnId>,
        blockiseq: Option<IseqPtr>,
        state: InsnId,
    },
    /// Pop the interpreter frame for an inlined callee, restoring the caller's frame.
    /// Carries the callee's iseq and argument count so codegen can compute the SP offset
    /// to restore.
    PopJITFrame {
        iseq: IseqPtr,
        argc: usize,
        state: InsnId,
    },
```

zjit/src/invariants.rs
```diff
@@ -211,9 +211,16 @@ pub extern "C" fn rb_zjit_invalidate_no_ep_escape(iseq: IseqPtr) {
...
-        // Remember that this ISEQ may escape EP
+        // Remember that this ISEQ may escape EP. If we have already processed
+        // an EP escape for this ISEQ, the JIT version was already invalidated
+        // and any patch points were already patched out. Redoing that work
+        // would just re-walk the same patch points and overwrite them with
+        // identical jumps. CRuby calls this on every env heap-promotion,
+        // regardless of whether ZJIT has anything left to invalidate.
         let invariants = ZJITState::get_invariants();
-        invariants.ep_escape_iseqs.insert(iseq);
+        if !invariants.ep_escape_iseqs.insert(iseq) {
```
alanwu-shopify-inc: Please drop this commit. It's superseded by https://github.com/ruby/ruby/commit/070ebb3f0e302233be2f434e6c69114ffb96d06e

zjit/src/codegen.rs
nirvdrum: I'm updating to integrate changes from #17188.

zjit/src/hir.rs
nirvdrum: Good catch.

zjit/src/codegen.rs
```diff
@@ -1537,6 +1581,143 @@ fn gen_send_without_block(
...
+    asm: &mut Assembler,
+    iseq: IseqPtr,
+    argc: usize,
+    state: &FrameState,
+) {
+    let local_size = unsafe { get_iseq_body_local_table_size(iseq) }.to_usize();
+    let sp_offset = (state.stack().len() + local_size - argc + VM_ENV_DATA_SIZE.to_usize()) * SIZEOF_VALUE;
+
+    asm_comment!(asm, "restore caller SP after inline");
+    let old_sp = asm.sub(SP, sp_offset.into());
```
tekknolagi: nit: use `sub_into`

zjit/src/codegen.rs
```diff
@@ -1537,6 +1581,143 @@ fn gen_send_without_block(
...
+    asm.mov(SP, old_sp);
+
+    asm_comment!(asm, "restore caller CFP after inline");
```
tekknolagi, line 1679: nit: `add_into`

zjit/src/hir.rs
```diff
@@ -2744,6 +2816,19 @@ impl Function {
...
     }
 
+    /// Return the deepest inlining nesting present in this function's frames.
+    /// The top-level frame is depth 0, so a function that inlines nothing
+    /// returns 0. Codegen uses this to size the per-function JITFrame slot
+    /// region, which needs one slot per simultaneously live frame, i.e.
+    /// `inlining_depth() + 1` slots. This is a measurement of what the function
+    /// actually contains, not a configured limit.
+    pub fn inlining_depth(&self) -> InlineDepth {
+        self.insns.iter().filter_map(|insn| match insn {
```
tekknolagi, line 2808: I think this needs to walk blocks in RPO and then look at those instructions. We should not make a habit of reading dead instructions

zjit/src/hir.rs
tekknolagi: cc @nirvdrum please address

zjit/src/hir.rs
```diff
@@ -4262,7 +4366,419 @@ impl Function {
...
+            incr_counter!(inline_reject_ep_escapes);
+            return false;
+        }
+
+        // Reject callees that use yield/invokeblock (block inlining is future work).
+        let callee_encoded_size = unsafe { get_iseq_encoded_size(callee_iseq) };
+        let mut pc_idx: u32 = 0;
+        while pc_idx < callee_encoded_size {
+            let pc = unsafe { rb_iseq_pc_at_idx(callee_iseq, pc_idx) };
+            let opcode: u32 = unsafe { rb_iseq_opcode_at_pc(callee_iseq, pc) }
```
tekknolagi: use bare opcode

zjit/src/hir.rs
tekknolagi: cc @nirvdrum please address

zjit/src/hir.rs
tekknolagi: Do we still need rollback?

zjit/src/hir.rs
```diff
@@ -4262,7 +4366,419 @@ impl Function {
...
+                //
+                // Each callee Snapshot also inherits a frame depth one greater than
+                // the call-site frame's. Because a callee's own calls only become
+                // inlinable in a later fixed-point iteration (after type
+                // specialization promotes them to SendDirect), frames are always
+                // woven in outermost-first, so the call-site frame's depth is already
+                // final here.
+                let caller_depth = self.frame_state(state).depth;
+                for idx in pre_insns_len..self.insns.len() {
+                    if let Insn::Snapshot { state: inlined_state } = &mut self.insns[idx] {
```
tekknolagi: I feel like this should happen instead on FrameState construction

zjit/src/options.rs
```diff
@@ -115,8 +115,35 @@ pub struct Options {
     /// Maximum number of versions per ISEQ
     pub max_versions: usize,
 
-    /// Maximum callee bytecode size for inlining (0 disables inlining).
+    /// Per-callee size threshold for inlining, measured as the callee's YARV bytecode
+    /// encoded size in u32 words (`get_iseq_encoded_size`). Callees larger than this
```
alanwu-shopify-inc: The "encoded size in u32 words" part is inaccurate. `get_iseq_encoded_size` is `ISEQ_BODY(iseq)->iseq_size` which is a count of VALUEs.

zjit/src/hir.rs
```diff
@@ -5785,9 +6302,9 @@ impl Function {
         }
 
         macro_rules! run_pass {
-            ($name:ident) => {
+            ($name:ident) => {{
                 let counter = counter_for!($name);
-                crate::stats::trace_compile_phase(stringify!($name), ||
+                let ret = crate::stats::trace_compile_phase(stringify!($name), ||
```
tekknolagi: nit: `result`

zjit/src/hir.rs
tekknolagi: this doesn't look resolved

zjit/src/hir.rs
```diff
@@ -6410,6 +6955,20 @@ pub struct FrameState {
...
+    /// propagates here automatically, and so the caller's state has a single
+    /// source of truth in the IR.
+    caller: Option<InsnId>,
+
+    /// Inlining nesting depth of this frame. The top-level (non-inlined) frame
+    /// is depth 0; each level of inlining increments it by one. Codegen uses
+    /// this to pick a distinct JITFrame slot per active inlined frame so that
+    /// `cfp->jit_return` values do not alias across the shared native stack frame.
+    /// This value's upper bound is the `inline_max_iterations` value.
+    pub depth: InlineDepth,
```
tekknolagi, line 6917: perhaps it is easier to leave this computed instead of as a field; then we avoid having to do a bunch a bunch of stuff on set-up

zjit/src/hir.rs
```diff
@@ -6718,14 +7327,31 @@ impl ProfileOracle {
...
         }
         let self_type_distribution = &operand_types[0];
         entry.push((self_param, TypeDistributionSummary::new(self_type_distribution)))
     }
+
+    /// Merge profile entries from an inlined callee's oracle into this one. Callee InsnIds are
+    /// remapped by `insn_base` so that they refer to the caller's namespace after the inliner
+    /// copies the callee's instructions into the caller. Entries are inserted under the callee's
+    /// (iseq, insn_idx) key, which does not collide with the caller's own entries.
+    fn merge_callee(&mut self, callee: &ProfileOracle, insn_base: usize) {
```
tekknolagi: `insn_base` seems unused and you changed how this worked I think

zjit/src/hir.rs
tekknolagi: Let's drop

zjit/src/hir.rs
```diff
@@ -4338,12 +4338,12 @@ impl Function {
             return false;
         }
 
-        // Only inline callees with simple positional parameters. Optional params,
-        // keyword params, rest/post params, block params, and forwardable params all
-        // have local table layouts that our parameter mapping doesn't handle yet.
+        // Inline callees with required and/or optional positional parameters. Keyword
+        // params, rest/post params, block params, and forwardable params are still
+        // rejected because their local-table layouts and argument-shaping prologues
+        // need additional work to map cleanly into HIR.
```
alanwu-shopify-inc: The reason given here is true but unclear. The deeper reason is that our JIT-to-JIT calling convention doesn't support these parameter types and we can only inline calls that are fully supported in the calling convention.

zjit/src/invariants.rs
tekknolagi: Do we need to change this still? Why early return?

tekknolagi:
> The results show a 56% reduction in iseq_optimized_send_count and 18% reduction in overall optimized_send_count.

I think we should count inlining as optimizing a send. This is otherwise a "bad" signal.

zjit/src/codegen.rs
```diff
@@ -1632,26 +1632,32 @@ fn gen_push_lightweight_frame(
         write_block_code: iseq_may_write_block_code(iseq),
     });
 
-    // Fully materialize the inlined callee's CFP. Unlike gen_send_iseq_direct
-    // where the callee's JIT entry writes cfp->pc/iseq on demand (or
-    // gen_save_pc_for_gc lazily installs a JITFrame pointer), inlined code has
-    // no entry point, so we must initialize these fields up front.
+    // Publish the inlined callee's entry JITFrame before the inlined body runs.
+    // Unlike gen_send_iseq_direct, where the callee's separate gen_entry_point
```
alanwu-shopify-inc: The "unlike" part now inaccurate after https://github.com/ruby/ruby/pull/17188

alanwu-shopify-inc:
Consider collapsing all the commits that refresh HIR snapshots to one that lives at the top.

tekknolagi:
> Consider collapsing all the commits that refresh HIR snapshots to one that lives at the top.

To make it easier to rebase or?

XrXr:
> Consider collapsing all the commits that refresh HIR snapshots to one that lives at the top.

>> To make it easier to rebase or?

To make the log easier to follow, because the intention is to not merge with squashing. It may or may not make the rebase easier. There's other commits that are arguable candidates for squashing together (e.g. [ZJIT: Initialize cfp->pc and cfp->sp for inlined frames](https://github.com/ruby/ruby/pull/16966/commits/d4cb77eb49ba5c34335ca49f61c7d3c5e4ad8876) and [ZJIT: Defer cfp->sp write at inlined frame push to match the non-inlined path](https://github.com/ruby/ruby/pull/16966/changes/9a975c5d95c42e90e25ba2e36309b6fc00a147e8), [ZJIT: Convert inliner tests without a dependency on external libraries into codegen tests](https://github.com/ruby/ruby/pull/16966/changes/0cdc4ede5463fdaacec621e49a23dddb6d268cb0) and [ZJIT: Add tests for inling methods with catch tables](https://github.com/ruby/ruby/pull/16966/changes/e5f41ea5c058b75e563bf15228d3d11b54949deb)) but the refresh commits are clear noise.

zjit/src/hir.rs
nirvdrum: The basic inliner shape was drawn from a couple of compiler books I have (_Engineering a Compiler_ and _Modern Compiler Implementation in ML_) with some influence from other compilers (this [inlining survey blog post](https://bernsteinbear.com/blog/inlining-heuristics/) is a great resource). I biased toward simplicity for the first iteration in order to establish a baseline. For tuning options I looked at what Graal offers and tried to translate that into something ZJIT can support today (e.g., we don't do background compilation and have more limited profiling).

For the optimization loop, I had seen Ruby code optimize very well in TruffleRuby, where its partial evaluator interleaves inlining and other optimization. The goal was to let one level of inline expansion unlock simplification before considering the next level. The Prokopec paper you linked noted a similar approach of alternating between call tree simplification and inlining yielding better results than just inlining alone. However, the process described in Prokopec et al. also includes an inline exploration phase that this inliner does not have -- after eligibility checks, the inline heuristic is just threshold and budget (there's no cost/benefit analysis at the moment). So I don't want to over-claim from that paper; it's inspiration rather than evidence that this PR's implementation has the same properties.

That limited heuristic is why I went with a BFS-like approach. Inlining visits all `SendDirect`s at a particular call level and then moves on to the rest of the optimization pipeline. If that exposes more `SendDirect`s before we hit a fixed point or the iteration cap, the next pass can inline those. I initially implemented inlining without alternation and found the results weren't particularly interesting since the actual call overhead isn't that pronounced; inlining in ZJIT is most useful for unlocking other optimizations.

[Steiner et al.](https://scispace.com/pdf/adaptive-inlining-and-on-stack-replacement-in-the-cacao-2a8cmi22nl.pdf) compares DFS, DFS + cancel, BFS, and knapsack exploration. That work shows DFS can be better, but absent better heuristics, I didn't want a deeply nested or recursive call to cause the inliner to go really deep on one callee while starving others. The BFS-like traversal is less likely to starve sibling call sites and is a bit more predictable. Ultimately, I expect we'll replace this with a policy that can make more informed inlining decisions.

For this particular implementation, I went with something intentionally simple. It's an amalgamation of ideas rather than a direct implementation of any one paper I found. Looping the current optimization pipeline is easy to disable and keeps the non-inlining phase order intact. I expect this is an area that we'll want to experiment with. The simple optimization loop yields demonstrable benefits and serves as a reasonable baseline that we can compare against for future work.

zjit/src/hir.rs
tekknolagi: It's doing more than just JITFrame operations so this would be a misnomer. Let's leave as-is for now until it's actually just doing 1-3 movs

tekknolagi: cc @nirvdrum what do you think? I am not a huge fan of either name rn

nirvdrum: Assuming we're not looking to merge `gen_push_frame` and `gen_push_lightweight_frame`, I'm partial to the "lightweight" moniker. I think it clearly establishes the intended difference between the two operations.

I could get on board with `PushInlinedFrame`/`PopInlinedFrame`, but I'm not a fan of `PushJITFrame`/`PopJITFrame`. Handling the `JITFrame` metadata is only one part of what these operations do, so naming the HIR instruction after the data structure feels a bit too narrow.

zjit/src/invariants.rs
nirvdrum: Thanks for double-checking this. It turns out we do not use union-find in construction. I can't recall if it ever did in local work or if it was an overzealous guard I inserted while merging #16066. But I've verified we don't need it and have removed that rollback mechanism.

As for rollback in general: compiling an ISEQ can legitimately fail and since we're compiling directly into the caller, we need to unwind whatever we've done. Fortunately, since this is all append-only we can just truncate the various lists back to their original length.

alanwu-shopify-inc: I think you replied to the wrong thread.

zjit/src/hir.rs
nirvdrum: I'm going to leave the name as-is for now. It's a straightforward change to make if we really don't like the name.

zjit/src/invariants.rs
nirvdrum: Whoops. You're right. Thanks for flagging it. I meant to reply to [this comment](https://github.com/ruby/ruby/pull/16966#discussion_r3381865066). I'll carry it over there. Apologies for the confusion.

zjit/src/hir.rs
nirvdrum: Thanks for double-checking this. It turns out we do not use union-find in construction. I can't recall if it ever did in local work or if it was an overzealous guard I inserted while merging #16066. But I've verified we don't need it and have removed that rollback mechanism.

As for rollback in general: compiling an ISEQ can legitimately fail and since we're compiling directly into the caller, we need to unwind whatever we've done. Fortunately, since this is all append-only we can just truncate the various lists back to their original length.

zjit/src/hir.rs
nirvdrum: Sorry. I pushed a half solution. It should be fixed now.

zjit/src/payload.rs
nirvdrum: This code has all been removed. It existed as a way for an inlined callee to invalidate/recompile the method it was inlined into. I've solved that more straightforwardly with 664e0bf52c12b821fe4deec6ea120a89ee52f8ac.

zjit/src/codegen.rs
nirvdrum: Thanks. I didn't know about that one.

zjit/src/options.rs
nirvdrum: I wanted to use the same terminology as the function being used: `rb_iseq_encoded_size`. I realize that underlying it's a count of `VALUE`s, but how would you phrase that in a way that makes sense to a Ruby user passing a CLI option to set the inline threshold?

zjit/src/hir.rs
nirvdrum: Thanks for flagging. This was removed/reworked in 3288613bede6226374ee7040485bc4b1d40d9d75.

zjit/src/codegen_tests.rs
```diff
@@ -6,12 +6,22 @@ use crate::backend::lir::Assembler;
...
 
+#[track_caller]
+fn with_inlining<T>(f: impl FnOnce() -> T) -> T {
+    with_rubyvm(|| {});
+    let old_threshold = get_option!(inline_threshold);
+    unsafe { OPTIONS.as_mut().unwrap().inline_threshold = 30; }
+    let result = f();
+    unsafe { OPTIONS.as_mut().unwrap().inline_threshold = old_threshold; }
+    result
+}
```
tekknolagi, line 24:
```suggestion
#[track_caller]
fn with_inlining<T>(f: impl FnOnce() -> T) -> T {
    with_rubyvm(|| {
      let old_threshold = get_option!(inline_threshold);
      unsafe { OPTIONS.as_mut().unwrap().inline_threshold = 30; }
      let result = f();
      unsafe { OPTIONS.as_mut().unwrap().inline_threshold = old_threshold; }
      result
    })
}
```

zjit/src/invariants.rs
tekknolagi: Bumping this thread tho

zjit/src/hir.rs
tekknolagi: Alternatively we can keep track of this in the inliner by walking up the tree each time we inline and keeping the `max()`

zjit/src/codegen_tests.rs
nirvdrum: Thanks. I don't know what happened there. Apologies for not catching it before pushing.

zjit/src/hir.rs
tekknolagi: (don't disable profiles)

zjit/src/hir.rs
```diff
@@ -4283,7 +4367,409 @@ impl Function {
...
+
+        // Reject callees that use yield/invokeblock (block inlining is future work).
+        let callee_encoded_size = unsafe { get_iseq_encoded_size(callee_iseq) };
+        let mut pc_idx: u32 = 0;
+        while pc_idx < callee_encoded_size {
+            let pc = unsafe { rb_iseq_pc_at_idx(callee_iseq, pc_idx) };
+            let opcode: u32 = unsafe { rb_iseq_opcode_at_pc(callee_iseq, pc) }
+                .try_into()
+                .unwrap();
+            // This scan runs before `rb_zjit_profile_disable`, so check the
```
tekknolagi: Why do we disable `invokeblock` completely vs just not optimizing it?

zjit/src/hir.rs
tekknolagi: not addressed?

zjit/src/hir.rs
nirvdrum: It's just how I scoped the PR. I started by restricting what we support and systematically whittled that list down as I added support for them to the inliner, wrote tests, and verified everything works with _ruby-bench_. You can see that play out in chronological order in the git history. I opened the PR before going through that process with `invokeblock`. If you want that included now rather than in the planned follow-up PR, I can move that up.

zjit/src/hir.rs
nirvdrum: Nifty. I didn't know about `rb_iseq_bare_opcode_at_pc`. That cleans things up nicely.

zjit/src/codegen_tests.rs
nirvdrum: Thanks for the suggestion. I needed to adjust the type of `f` because of ownership rules so I couldn't apply the suggestion via GitHub. It's been fixed, however.

zjit/src/hir.rs
tekknolagi: It's fine; just leave TODO I guess

tekknolagi: As is it looks like something is broken rather than an "idk, will investigate later"

zjit/src/hir.rs
nirvdrum: I just double-checked and it's `let result` in the latest version. Maybe there was a lag? Or I'm finding GitHub is having difficulty keeping up with the PR. I've had to refresh a few times.

zjit/src/hir.rs
nirvdrum: I reworked this in e9359348e2ec8902b89a0b0875cc3c066fb4d4b4. Introducing the `AddIseqMode` enum made it a lot nicer to pass inliner-specific data around. Let me know if this substantially addresses your concern.

zjit/src/hir.rs
nirvdrum: I eliminated the loop in e9359348e2ec8902b89a0b0875cc3c066fb4d4b4 as part of a larger clean-up on tracking the caller and inlining depth.

zjit/src/hir.rs
nirvdrum: I think e9359348e2ec8902b89a0b0875cc3c066fb4d4b4 addresses the concern about set-up activity. I looked at computing the depth on demand and it's a bit involved because multiple consumers of the `depth` field don't have the corresponding `Function` reference to follow the `caller` chain. Unless you feel strongly about it, I'm inclined to leave `depth` as a field.

zjit/src/hir.rs
nirvdrum: I adopted `vm_zjit_insn_to_bare_insn ` in 461589a97cd855ed7bacdd52ecef3ec194714af6.

zjit/src/hir.rs
nirvdrum: This was fixed in 15b43935e9eb798cc2ccdd888171ef265b934739.

test/ruby/test_zjit.rb
nirvdrum: These were moved in 8bd8e5eec3bd921616b64d84ebf4452627e9ac84.

zjit/src/invariants.rs
nirvdrum: I've dropped the commit.

zjit/src/hir.rs
XrXr: I think this is fine for now. We can add methods to the params struct later to clean this up but that should be for another PR.

zjit/src/hir.rs
XrXr: This part is fine, but the comment above is confusing.

`                // The callee's body entry block has params: [self, local0, local1, ..., stack0, stack1, ...]
` shouldn't mention stack: On entry, stack size is 0 and the comment itself makes no mention of stack parameters. The quoted code also assumes that there are no values on stack.

zjit/src/options.rs
XrXr: For a user facing description, I'd probably say it's count of pointers, or not even bother with committing to a specific unit, as the value passed isn't comparable to other values.

But this is a field description, and it's not user facing. For this, I would say "measured in YARV bytecode encoded size (`ISEQ_BODY(iseq)->iseq_size`)". Both `get_iseq_encoded_size` and `u32` are implementation details subject to change. The C definition is the ground truth.

k0kubun: I would just say "the callee's YARV bytecode size", or "the total number of instructions and operands in the callee's YARV bytecode" if you want to be more specific.

zjit/src/options.rs
```diff
@@ -114,6 +132,54 @@ pub struct Options {
...
+    /// (blocks, procs without a stable method binding) cannot be denied this way.
+    pub inline_deny: HashSet<String>,
+
+    /// Upper bound on how many times the `optimize` fixed-point loop will iterate
+    /// before giving up. Each iteration runs `type_specialize` → `inline` →
+    /// `inline_methods` → the rest of the HIR pipeline; in steady state the loop
+    /// terminates as soon as an iteration fails to inline anything new. The cap
+    /// exists to bound compile time when something pathological prevents the loop
+    /// from reaching a fixed point.
+    pub inline_max_iterations: InlineDepth,
```
k0kubun, line 183: As Alan pointed out above, field descriptions are not user-facing. If you mean to let end-users use any of the command-line options about inlining, consider updating `ZJIT_OPTIONS` to document them in `ruby --help`.

I think `--zjit-inline-deny` should be an undocumented developer option like `--zjit-allowed-iseqs`, but we might want to consider exposing others.

zjit/src/hir.rs
```diff
@@ -4283,7 +4379,379 @@ impl Function {
...
+                    } else if Some(i) == kw_bits_local_idx {
+                        // Hidden kw_bits slot: alias to the SendDirect's compile-time
+                        // value as a fixnum, the same encoding the interpreter uses for
+                        // this hidden local. checkkeyword's FixnumBitCheck will read
+                        // this constant directly inside the inlined body.
+                        let bits_const = self.push_insn(block, Insn::Const {
+                            val: Const::Value(VALUE::fixnum_from_usize(kw_bits as usize)),
+                        });
+                        self.make_equal_to(param_id, bits_const);
+                    } else if i < num_locals {
```
k0kubun, line 4684: I think this works fine, but I'm confused by the fact that we're checking it. Since `num_locals = callee_body_params.len() - 1`, `callee_body_params[1..]` would have the size of `num_locals`. I would imagine `i` would go only up to `num_locals - 1`.

Isn't `i < num_locals` always true? (what's the point in checking it?)

zjit/src/hir.rs
k0kubun: The definition seems fine, but I left my feedback on how it's used at https://github.com/ruby/ruby/pull/16966#discussion_r3404800750

zjit/src/hir.rs
k0kubun: In my mind "Lightweight Frame" == "JITFrame", so if it's not named `PushJITFrame`/`PopJITFrame` because it's doing more than that, I don't think it should be named `PushLightweightFrame`/`PopLightweightFrame` either.

> I could get on board with PushInlinedFrame/PopInlinedFrame

This makes sense to me. Can we do `PushInlineFrame`/`PopInlinedFrame` then?

zjit/src/options.rs
```diff
@@ -361,6 +431,35 @@ fn parse_option(str_ptr: *const std::os::raw::c_char) -> Option<()> {
...
+            // duplicated commas don't introduce a "" sentinel into the set.
+            for entry in csv.split(',') {
+                let trimmed = entry.trim();
+                if !trimmed.is_empty() {
+                    options.inline_deny.insert(trimmed.to_string());
+                }
+            }
+        }
+
+
```
k0kubun, line 463: Is this extra blank line intentional?

```suggestion
```

tenderlove (approved):

zjit/src/hir.rs
nirvdrum: Added a TODO to clarify what future work to complete.

tekknolagi:
Enabling auto merge. Blast radius is small because it's disabled by default. Only one new CI run and it's optional.

XrXr:
Switched to "auto-merge (rebase)" since effort was spent writing commit messages :)

tekknolagi:
After offline conversation, back to squash.

nirvdrum:
> Switched to "auto-merge (rebase)" since effort was spent writing commit messages :)

FYI, despite having a clean history, not every commit builds. I opted not to go back and rewrite every commit with the EBB->BB change. The history is useful for the evolution of the PR and I think the commits are logical demarkation points, but making every commit build was too time-consuming with the frequent rebases and ensuing breakages.

PR #16972, opened 2026-05-14
[DOC] Update refinement documentation after disallowing include in refinements

jeremyevans:
When this documentation was written, refinements allowed the use of include. However, support for that was removed in Ruby 3.2.


shugo (approved):
It looks good to me.  Thank you!

PR #16973, opened 2026-05-14
Extract bundled-gem C extensions before configuring exts

st0012:
To support rbs-integrated RDoc, we need to make sure rbs is built properly and available. This PR does that as well as bumping RDoc.

Did a smoke-test at [my fork](https://github.com/st0012/actions/actions/runs/25978792739/job/76363687335). So this time it should work 🤞


nobu:
Re the PR title, "Extract" instead of "Build"?

nobu (approved):

PR #16983, opened 2026-05-15
Drop rbs require in rdoc-srcdir

st0012:
This is currently blocking Ruby's doc generation:
https://github.com/ruby/actions/actions/runs/25895616458/job/76107951475

Related to https://github.com/ruby/ruby/pull/16760

We'll need to make rbs work for the process so we can adopt the rbs-integrated RDoc. But for now let's drop it to unblock doc generation.

### DIFF
--- a/tool/rdoc-srcdir
+++ b/tool/rdoc-srcdir
@@ -2,7 +2,6 @@
 
 require 'rubygems'
 require 'rdoc/rdoc'
-require 'rbs'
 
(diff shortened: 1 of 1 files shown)


nobu (approved):
Forgot to remove it, thanks.

st0012:
@nobu do you know how we should build rbs for rdoc’s generation, while making it work for cross compilation?

PR #16988, opened 2026-05-15
ZJIT: Share a single JITFrame across all C method frames

k0kubun:
This PR writes a special value (`1`) into `cfp->jit_return` for C method frames instead of allocating `JITFrame` with `pc = 0, iseq = 0` for each C method callsite. We may do callsite-specific optimizations for C methods later, but for a little while, our focus will be on ISEQ methods and stack maps.

Carved out from a patch for ISEQ frames to simplify its diff.

### DIFF
--- a/zjit.h
+++ b/zjit.h
@@ -27,6 +27,7 @@ typedef struct zjit_jit_frame {
 
 #if USE_ZJIT
 extern void *rb_zjit_entry;
+extern const zjit_jit_frame_t rb_zjit_c_frame;
 extern uint64_t rb_zjit_call_threshold;
 extern uint64_t rb_zjit_profile_threshold;
 void rb_zjit_compile_iseq(const rb_iseq_t *iseq, rb_execution_context_t *ec, bool jit_exception);
@@ -48,6 +49,21 @@ void rb_zjit_tracing_invalidate_all(void);
 void rb_zjit_invalidate_no_singleton_class(VALUE klass);
 void rb_zjit_invalidate_root_box(void);
 void rb_zjit_jit_frame_update_references(zjit_jit_frame_t *jit_frame);
+
+// Special value for cfp->jit_return that means "this is a C method frame, use
+// rb_zjit_c_frame as the JITFrame". We don't control the native stack layout
+// for C frames, so there's no per-call JITFrame storage; we set this sentinel
+// instead of a heap-allocated JITFrame pointer.
+#define ZJIT_JIT_RETURN_C_FRAME 0x1
+
+static inline const zjit_jit_frame_t *
+CFP_ZJIT_FRAME(const rb_control_frame_t *cfp)
+{
+    if ((VALUE)cfp->jit_return == ZJIT_JIT_RETURN_C_FRAME) {
+        return &rb_zjit_c_frame;
+    }
+    return (const zjit_jit_frame_t *)cfp->jit_return;
+}
 #else
 #define rb_zjit_entry 0
 static inline void rb_zjit_compile_iseq(const rb_iseq_t *iseq, rb_execution_context_t *ec, bool jit_exception) {}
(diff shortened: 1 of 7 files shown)


zjit.h
```diff
@@ -48,6 +49,21 @@ void rb_zjit_tracing_invalidate_all(void);
 void rb_zjit_invalidate_no_singleton_class(VALUE klass);
 void rb_zjit_invalidate_root_box(void);
 void rb_zjit_jit_frame_update_references(zjit_jit_frame_t *jit_frame);
+
+// Special value for cfp->jit_return that means "this is a C method frame, use
+// rb_zjit_c_frame as the JITFrame". We don't control the native stack layout
+// for C frames, so there's no per-call JITFrame storage; we set this sentinel
+// instead of a heap-allocated JITFrame pointer.
+#define ZJIT_JIT_RETURN_C_FRAME 0x1
```
XrXr, line 57: Could you try if bindgen is happy with this, this way we don't need to `as usize` on the rust side? Also, may need `#include "ruby/internal/config.h"` for size_t definition.

```suggestion
#define ZJIT_JIT_RETURN_C_FRAME ((size_t)0x1)
```

zjit.h
```diff
@@ -62,40 +78,39 @@ static inline void rb_zjit_tracing_invalidate_all(void) {}
 static inline void rb_zjit_invalidate_no_singleton_class(VALUE klass) {}
 static inline void rb_zjit_invalidate_root_box(void) {}
 static inline void rb_zjit_jit_frame_update_references(zjit_jit_frame_t *jit_frame) {}
+static inline const zjit_jit_frame_t* CFP_ZJIT_FRAME(const rb_control_frame_t *cfp) { return NULL; }
```
XrXr:
```suggestion
static inline const zjit_jit_frame_t *CFP_ZJIT_FRAME(const rb_control_frame_t *cfp) { return NULL; }
```

XrXr (approved):
LGTM: Just nits you can ignore at will

tekknolagi:
What/how much does this buy us?

zjit.h
k0kubun: Unfortunately, bindgen didn't work with that change.

```diff
diff --git a/zjit.h b/zjit.h
index c3594b0529..b3d96fa426 100644
--- a/zjit.h
+++ b/zjit.h
@@ -54,7 +54,7 @@ void rb_zjit_jit_frame_update_references(zjit_jit_frame_t *jit_frame);
 // rb_zjit_c_frame as the JITFrame". We don't control the native stack layout
 // for C frames, so there's no per-call JITFrame storage; we set this sentinel
 // instead of a heap-allocated JITFrame pointer.
-#define ZJIT_JIT_RETURN_C_FRAME 0x1
+#define ZJIT_JIT_RETURN_C_FRAME ((size_t)0x1)

 static inline const zjit_jit_frame_t *
 CFP_ZJIT_FRAME(const rb_control_frame_t *cfp)
diff --git a/zjit/src/cruby_bindings.inc.rs b/zjit/src/cruby_bindings.inc.rs
index 1194a590df..559134a979 100644
--- a/zjit/src/cruby_bindings.inc.rs
+++ b/zjit/src/cruby_bindings.inc.rs
@@ -234,7 +234,6 @@ pub const VM_ENV_DATA_INDEX_SPECVAL: i32 = -1;
 pub const VM_ENV_DATA_INDEX_FLAGS: u32 = 0;
 pub const VM_BLOCK_HANDLER_NONE: u32 = 0;
 pub const SHAPE_ID_NUM_BITS: u32 = 32;
-pub const ZJIT_JIT_RETURN_C_FRAME: u32 = 1;
 pub const ZJIT_JIT_RETURN_POISON: i64 = -4981057192772781345;
 pub type rb_alloc_func_t = ::std::option::Option<unsafe extern "C" fn(klass: VALUE) -> VALUE>;
 pub const RUBY_Qfalse: ruby_special_consts = 0;
```

zjit.h
k0kubun: https://github.com/ruby/ruby/pull/16988/commits/4267d9a7995d1095a7bdb72767f4837930a6e924

k0kubun:
> What/how much does this buy us?

It doesn't seem to have a significant impact on `zjit_alloc_bytes`: it's `14,585,480` -> `14,575,364` on lobsters.

The point of this PR, however, is more about removing the write of an actual `JITFrame` for C methods. For stack maps, we will start writing `JITFrame` into a native stack slot for ISEQ frames, and it's hard for C method frames since we cannot control the stack layout of C functions.

PR #16993, opened 2026-05-15
ZJIT: Split test_dont_inline_integer_xor_with_bignum_or_boolean

k0kubun:
This PR splits `test_dont_inline_integer_xor_with_bignum_or_boolean` into three tests.

Having multiple `assert_snapshots!` in a single test function means that you need to run multiple iterations of `zjit-test` + `zjit-test-update` loops. This function has three pairs of Ruby code + HIR test, so there's no good reason to group them in the same test function.


tekknolagi (approved):
thank you!

PR #16995, opened 2026-05-16
Prohibit super in refined module method

jeremyevans:
Before, this super could result in a call to a method in BasicObject (or a module included in or prepended to BasicObject), or a call to method_missing for the receiver's class.

This explicitly disallows the use of super in such a case, with a specific error message.

Fixes [Bug #22071]


shugo (approved):
It looks good to me 👍

PR #16996, opened 2026-05-16
Bundled gem extensions are out of scope of update-deps

nobu:
Follow up of commit d5c5fcb80a432e2078139c460230dc7952216c35.
`target` may be the basename only, skip all rules under bundled gem.


st0012:
This unblocked the RDoc bump. Thank you!

PR #16997, opened 2026-05-16
Handle refinements correctly in {Method,UnboundMethod}#super_method

jeremyevans:
A `Method#super_method` chain should return the methods that `super` would call if the method was called normally. Previously, there were multiple problems:

* There was an infinite `super_method` loop for refined methods.

* Lookup considered the refinements activated at the call site of `super_method`, and not all the call site of `super` inside the method.

This tries to recreate the logic that `super` uses inside `super_method`. It avoids the loop. It also considers the refinements activated for the method itself, not for the caller of `super_method`, correctly handling refinements in outer scopes of the method. This requires avoiding the use of `rb_callable_method_entry_with_refinements`, which implicitly will consider refinements activated in the caller of `super_method`.

The added tests attempt to ensure that the `super_method` lookup chain matches the methods that `super` calls if you call the method.

This adds an `RICLASS_FOR_REFINEMENT_P` helper method, for logic that is used a couple times in the new code and once in `vm_search_normal_superclass`.

### DIFF
--- a/internal/class.h
+++ b/internal/class.h
@@ -646,6 +646,14 @@ RICLASS_OWNS_M_TBL_P(VALUE iclass)
     return RCLASSEXT_ICLASS_IS_ORIGIN(ext) && !RCLASSEXT_ICLASS_ORIGIN_SHARED_MTBL(ext);
 }
 
+static inline bool
+RICLASS_FOR_REFINEMENT_P(VALUE iclass)
+{
+    return BUILTIN_TYPE(iclass) == T_ICLASS &&
+            RB_TYPE_P(RBASIC(iclass)->klass, T_MODULE) &&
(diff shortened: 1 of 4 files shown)


shugo (commented):
Thank you.

It looks good execept that super_method doesn't return the right method for BMETHOD.

```ruby
class Base
  def b = "Base"
end
module M
  refine(Base) { def b = "M -> " + super }
end
using M
class A < Base
  define_method(:b) { "A ->" + super() }
end

a = A.new
p a.b                             # => "A -> M -> Base"
p a.method(:b).super_method.owner # => exepected #<refinement:Base@M>, but Base
```

jeremyevans:
@shugo Thank you for your review. I added a patch to handle bmethods. Those require a different approach to find the CREF to use to determine the activated refinements.

shugo (approved):
Thank you.  It looks good to me :+1:

PR #16999, opened 2026-05-16
ZJIT: Remove dead SideExitReasons

a5-stable:
`ArrayMax` (#16411) and `FixnumDiv` (#15452) have codegen implemented
and no longer return `Err` from `gen_insn`.


XrXr (approved):
Thanks!

PR #17004, opened 2026-05-16
[DOC] Tweaks for Pathname#children

BurdetteLamar:


pathname_builtin.rb
```diff
@@ -744,26 +744,33 @@ def join(*args)
...
   #
-  # For example:
-  #   pn = Pathname("/usr/lib/ruby/1.8")
-  #   pn.children
-  #       # -> [ Pathname:/usr/lib/ruby/1.8/English.rb,
-  #              Pathname:/usr/lib/ruby/1.8/Env.rb,
-  #              Pathname:/usr/lib/ruby/1.8/abbrev.rb, ... ]
-  #   pn.children(false)
-  #       # -> [ Pathname:English.rb, Pathname:Env.rb, Pathname:abbrev.rb, ... ]
+  # With `full_paths` given as `true` (the default),
```
peterzhu2118: A bit of a nit but I'm not sure I like calling it `full_path`. It implies to me that it's an absolute path (which it is not necessarily). It's also inconsistent with the argument name in the source code (which is `with_directory`).

BurdetteLamar: Good point about the parameter name.  How about `with_dirname`?  `dirname` is a well-established notion.

(We don't necessarily need to be guiided by the code's parameter names;  sometimes they're pretty weak.)

BurdetteLamar: Adjusted.

peterzhu2118 (approved):

PR #17010, opened 2026-05-17
Fix UAF in IO::Buffer#| when self or mask is an invalidated slice

himura467:
Same fix as #16964, applied to `IO::Buffer#|`.

`io_buffer_or` accessed `buffer->base` and `mask_buffer->base` directly without validating that the buffers were still live. A slice whose parent had been freed retained its stale base pointer, so calling `|` on it caused a UAF.

Use `io_buffer_get_bytes_for_reading` for both operands, which raises `IO::Buffer::InvalidatedError` before any memory access if either buffer has been invalidated.

```ruby
inner = IO::Buffer.new(IO::Buffer::PAGE_SIZE)
slice = inner.slice(0, 8)
inner.free
slice | IO::Buffer.for("ABCDEFGH")  # => was UAF, now raises InvalidatedError
```

### DIFF
--- a/io_buffer.c
+++ b/io_buffer.c
@@ -3433,7 +3433,7 @@ io_buffer_and(VALUE self, VALUE mask)
 }
 
 static void
-memory_or(unsigned char * restrict output, unsigned char * restrict base, size_t size, unsigned char * restrict mask, size_t mask_size)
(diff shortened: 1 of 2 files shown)


kou (approved):
+1

PR #17012, opened 2026-05-17
ZJIT: StringExact/NilClass for GetSpecialSymbol and GetSpecialNumber

a5-stable:
According to `rb_reg_nth_match`, `rb_reg_match_pre(post/last)`,  
`GetSpecialSymbol` and `GetSpecialNumber` always returns `StringExact` or `NilClass`.

Fix infer_type to StringExact|NilClass instead of BasicObject.


tekknolagi (changes requested):
Can you add HIR tests for these please? We'd like to see them returning String/Nil. Surprised we don't already have tests.

a5-stable:
@tekknolagi 
Thank you for your review! 
I added test cases for `GetSpecialSymbol` and `GetSpecialNumber.` Both tests confirm they return `StringExact|NilClass`.

tekknolagi (approved):
Thanks!

